PR #4419, opened 2017-02-03
Transaction timestamp condition

arkpar:


ethcore/src/types/transaction.rs
```diff
@@ -52,6 +52,16 @@ impl Decodable for Action {
 	}
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "ipc", binary)]
+/// Transaction activation condition.
+pub enum Condition {
```
rphmeier: nit: doc over attrs

rphmeier: and a `From<BlockNumber>`, `From<u64>` would be nice as well

rpc/src/v1/types/transaction_condition.rs
```diff
@@ -0,0 +1,67 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use ethcore;
+
+/// Represents rpc api block number param.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
```
rphmeier: this doc seems outdated

rpc/src/v1/types/transaction_request.rs
```diff
@@ -41,9 +41,9 @@ pub struct TransactionRequest {
 	pub data: Option<Bytes>,
 	/// Transaction's nonce
 	pub nonce: Option<U256>,
-	/// Delay until this block if specified.
-	#[serde(rename="minBlock")]
-	pub min_block: Option<BlockNumber>,
+	/// Delay until this block condition.
+	#[serde(rename="condition")]
+	pub condition: Option<TransactionCondition>,
```
rphmeier: rename no longer necessary

rpc/src/v1/types/confirmations.rs
```diff
@@ -197,8 +197,8 @@ pub struct TransactionModification {
 	/// Modified gas
 	pub gas: Option<U256>,
 	/// Modified min block
```
tomusdrw: Comment & rename not needed.

ethcore/src/miner/transaction_queue.rs
```diff
@@ -1082,7 +1083,12 @@ impl TransactionQueue {
 			if delayed.contains(&sender) {
 				continue;
 			}
-			if tx.min_block.unwrap_or(0) > best_block {
+			let delay = match tx.condition {
+				Some(Condition::Number(n)) => n > best_block,
+				Some(Condition::Timestamp(t)) => t > time::get_time().sec as u64,
```
rphmeier: This could be based off of the best block's timestamp rather than current time, since that's all that contracts can inspect. It's additionally often set into the future and not consistent with "real" time, but can be agreed upon unanimously just like block numbers

arkpar: Waiting for the block timestamp would result in transactions included no in the first available block, but in the next one. But at least the transaction is guaranteed not to fail, so I guess that's reasonable

ethcore/light/src/client/mod.rs
```diff
@@ -110,7 +112,11 @@ impl Client {
 		let best_num = self.chain.best_block().number;
 		self.tx_pool.lock()
 			.values()
-			.filter(|t| t.min_block.as_ref().map_or(true, |x| x <= &best_num))
+			.filter(|t| match t.condition {
+				Some(TransactionCondition::Number(ref x)) => x <= &best_num,
+				Some(TransactionCondition::Timestamp(ref x)) => *x <= time::get_time().sec as u64,
```
rphmeier: this condition check needs updating as well. probably would be best done after changing `HeaderChain` to use `ethcore::encoded` (done locally).

rphmeier: given that it's "dead" code I don't mind merging this PR without it

arkpar: Ok, I'll leave it to you then :)

PR #4412, opened 2017-02-02
web browser dapp: prompt to install extension

derhuerst:
Fixes #4300. Tries inline installation first, falls back to opening a popup with the Chrome Web Store.

Will need a better description why to install the extension.

PR #4423, opened 2017-02-03
Extension installation overlay

jacogr:
TODO: Test once extension is published with new/correct extension URL.

Displays application prompt for extension installation. (Closes https://github.com/ethcore/parity/issues/4300)

Based off #4412 with changes -

- Doesn't display on hidden Web view, rather displays as part of application
- Proper store set for backing the component, rather than loose-standing calls

Future work -

- Create store & component tests
- Integrate prompt with Home view once available

parity 2017-02-03 15-43-08

### DIFF
--- a/js/src/views/Application/Extension/extension.js
+++ b/js/src/views/Application/Extension/extension.js
@@ -44,7 +44,7 @@ export default class Extension extends Component {
         <p>
           <FormattedMessage
             id='extension.intro'
-            defaultMessage='Parity now has an extension available for Chrome that allows for the safe browsing of Ethereum-enabled distributed applications. It is hightly recommended that you install this extension to further enhance your Parity experience.'
+            defaultMessage='Parity now has an extension available for Chrome that allows safe browsing of Ethereum-enabled distributed applications. It is highly recommended that you install this extension to further enhance your Parity experience.'
           />
         </p>
         <p className={ styles.buttonrow }>
--- a/js/src/views/Application/Extension/extension.js
+++ b/js/src/views/Application/Extension/extension.js
@@ -1,0 +1,74 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import { observer } from 'mobx-react';
+import React, { Component } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import { Button } from '~/ui';
+import { CloseIcon, CheckIcon } from '~/ui/Icons';
+
+import Store from './store';
+import styles from './extension.css';
+
+@observer
+export default class Extension extends Component {
+  store = new Store();
+
+  render () {
+    const { shouldShowWarning } = this.store;
+
+    if (!shouldShowWarning) {
+      return null;
+    }
+
+    return (
+      <div className={ styles.body }>
+        <CloseIcon
+          className={ styles.close }
+          onClick={ this.onClose }
+        />
+        <p>
+          <FormattedMessage
+            id='extension.intro'
+            defaultMessage='Parity now has an extension available for Chrome that allows the viewing of Ethereum identities and the safe browsing of Ethereum-enabled distributed applications. It is hightly recommended that you install this extension to further enhance your Parity experience.'
+          />
+        </p>
+        <p className={ styles.buttonrow }>
+          <Button
+            className={ styles.button }
+            icon={ <CheckIcon /> }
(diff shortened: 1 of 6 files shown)


js/src/views/Application/Extension/extension.js
```diff
@@ -0,0 +1,74 @@
...
+
+import Store from './store';
+import styles from './extension.css';
+
+@observer
+export default class Extension extends Component {
+  store = new Store();
+
+  render () {
+    const { shouldShowWarning } = this.store;
```
ngotchac: Why not `showWarning` ? Being polite is good though

js/src/views/Application/Extension/extension.js
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <div className={ styles.body }>
+        <CloseIcon
+          className={ styles.close }
+          onClick={ this.onClose }
+        />
+        <p>
+          <FormattedMessage
+            id='extension.intro'
+            defaultMessage='Parity now has an extension available for Chrome that allows the viewing of Ethereum identities and the safe browsing of Ethereum-enabled distributed applications. It is hightly recommended that you install this extension to further enhance your Parity experience.'
```
ngotchac: For now, the published extension won't have the augmentation part. Thus we can get rid of `the viewing of Ethereum identities`

js/src/views/Application/Extension/store.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @action hideWarning = (sleep = A_DAY) => {
+    this.nextDisplay = Date.now() + sleep;
+    store.set(NEXT_DISPLAY, this.nextDisplay);
+  }
+
+  @action testInstall = () => {
+    this.shouldInstall = this.readStatus();
+    console.log('testInstall', this.shouldInstall);
```
ngotchac: Not sure if logging is needed here. We already have too many logs by default. If needed, could use the new `logger` feature

js/src/views/Application/Extension/store.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  readStatus = () => {
+    const hasExtension = Symbol.for('parity.extension') in window;
+    const ua = browser.analyze(navigator.userAgent || '');
+
+    console.log('readStatus', hasExtension, ua);
```
ngotchac: Same as above

js/src/views/Application/Extension/store.js
```diff
@@ -0,0 +1,92 @@
...
+      }
+    })
+    .catch((error) => {
+      console.warn('Unable to perform direct install', error);
+      window.open(EXTENSION_PAGE, '_blank');
+
+      this.hideWarning(TEN_MINUTES + A_MINUTE);
+      setTimeout(() => {
+        this.testInstall();
+      }, TEN_MINUTES);
```
ngotchac: Shouldn't we have a way of closing the Warning only when the extension gets installed ? It seems bizarre to close it even if the installation failed (if the user clicked on `install` not on purpose, he can click on close).

Also, I'm not sure that the extension would actually be detected without a page refresh by just looking at the window object... To be verified

js/src/views/Application/Extension/store.js
jacogr: You are right, that one is a misplaced debugging comment. Thanks.

js/src/views/Application/Extension/store.js
jacogr: Same as above as well. Sorry.

js/src/views/Application/Extension/store.js
jacogr: Yes, 100% - would need to test this, updating the URL now to run through.

js/src/views/Application/Extension/extension.js
```diff
@@ -44,7 +44,7 @@ export default class Extension extends Component {
         <p>
           <FormattedMessage
             id='extension.intro'
-            defaultMessage='Parity now has an extension available for Chrome that allows the viewing of Ethereum identities and the safe browsing of Ethereum-enabled distributed applications. It is hightly recommended that you install this extension to further enhance your Parity experience.'
+            defaultMessage='Parity now has an extension available for Chrome that allows for the safe browsing of Ethereum-enabled distributed applications. It is hightly recommended that you install this extension to further enhance your Parity experience.'
```
ngotchac: 1. "it is *highly* recommended"
2. could get rid of "allows *for the* safe"

jacogr: Erk. Damnit.

ngotchac:
Works fine, but I think we want to use the inline install : https://support.google.com/webmasters/answer/34592?hl=en

@tomusdrw Localhosts sites must be added to the extension config in the Chrome Store
There should be at least (127.0.0.1/localhost):300 and (127.0.0.1/localhost):8180

jacogr:
*.web3.site is added to webmaster tools. However 1.5.1 doesn't have full support for it (yet). (And even in those cases we are not yet re-directing people, and cannot redirect all)

Inline install does run through, obviously just pops up.

For 1.5.1, fine as it, works, the extra auto-browser window popup is not the end of the world.

ngotchac:
Sure it's not the end of the world, but if it's only a one/few-line(s) to add to the config of the extension, I don't see why we shouldn't do it

jacogr:
100%, will check once the beta is build in a short while, hopefully the config is updated and it is slightly smoother. As it stands, no reason not to go. (Unless that config change breaks the process in some way)

jacogr:
Merging as-is so testing with beta build can be done.

jacogr:
s/grumble/looksgood/ (non code related concerns)

PR #4426, opened 2017-02-03
Build embedded Parity JS properly and separatly 

ngotchac:
For https://github.com/ethcore/parity-extension/issues/19

Updates Webpack and build the embedded parity bar as a separate build.
Emits a `embed.json` file that contains the needed assets.

Some components of the `~/ui` have been deleted from the `index.js` file : the editor and the PasswordStrength ones. This is because their dependencies are quite big, and Webpack 2 tree-shaking doesn't work that well : it wouldn't exclude them from the embed build, whereas the components are not used.

It seems that the production build was in fact not working (minification wasn't occurring) coming from the fact that we didn't transpile EthereumJS-Tx. Now it's fixed.
Webpack has been updated to the latest stable release.


js/src/ui/index.js
```diff
@@ -14,118 +14,47 @@
...
-  TxList,
-  TypedInput,
-  Warning
-};
+export AccountCard from './AccountCard';
+export Actionbar from './Actionbar';
+export ActionbarExport from './Actionbar/Export';
+export ActionbarImport from './Actionbar/Import';
+export ActionbarSearch from './Actionbar/Search';
+export ActionbarSort from './Actionbar/Sort';
```
jacogr: Aligning with the exports as done in `Form` & `Container` would actually be quire good -

```
export Actionbar, { Export as ActionbarExport, ...} from './Actionbar;
```

Alternatively since these are sub components, making them static on `Actionbar`, only exporting it and then using `Actionbar.Export` could also work.

js/src/ui/Form/index.js
```diff
@@ -14,35 +14,19 @@
...
+export InputAddress from './InputAddress';
+export InputAddressSelect from './InputAddressSelect';
+export InputChip from './InputChip';
+export InputDate from './InputDate';
+export InputInline from './InputInline';
+export InputTime from './InputTime';
+export Label from './Label';
+export RadioButtons from './RadioButtons';
+export Select from './Select';
+export TypedInput from './TypedInput';
```
jacogr: +1

js/src/redux/providers/workerWrapper.js
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+if (!process.env.EMBED) {
+  const worker = require('./worker');
+
+  module.exports = { setupWorker: worker.setupWorker };
```
jacogr: Since nothing else is used, why not -

```
const setupWorker = require('./worker').setupWorker;

module.exports = { setupWorker }
```

Would probably add a `TODO` here for when babel finally supports non top-level imports

jacogr:
Build is failing.

ngotchac:
Updated : tests passing

jacogr:
Looks good, however only to be merged after 1.5.2 is released. (Makes backporting much more difficult)

PR #4432, opened 2017-02-04
Fix eth_sign/parity_postSign

gavofyork:
`eth_sign` (and `parity_postSign`) used to return concatenated `r ++ s ++ v` with `v` being 0 or 1. it now agrees with geth as `v ++ r ++ s` with `v` being 27 or 28.


gentlejack:
accordion_arrow

PR #4436, opened 2017-02-05
Use secure websocket from HTTPS clients

skddc:
Currently, the unsecure `ws://` scheme is hardcoded. With this change, the scheme will dynamically change to the secure `wss://` when the frontend is requested from an HTTPS origin.

I might have missed something, and I'm entirely new to this codebase, so please excuse me if this doesn't make sense. We just stumbled into `ws://` failing for our parity node after enabling HTTPS and this is what I found as the most likely cause of the issue.


jacogr:
Looks good and above-board. I just want to do a test run-through to verify.

skddc:
[clabot:check]

skddc:
Cool, thanks.

PR #4438, opened 2017-02-05
Arbitrary labels for extended keys (u32, H256 built-in)

NikVolf:
Can be basically any byte sequence (but caller must be sure what is he doing since key is compacted into a 512-bit hash after all, no matter of original data length). This slightly increase complexity of the signatures, but probably worth it.


ethkey/src/extended.rs
```diff
@@ -21,6 +21,46 @@ use Public;
 use bigint::hash::{H256, FixedHash};
 pub use self::derivation::Error as DerivationError;
 
+pub trait Label {
```
rphmeier: document!

rphmeier: particularly the implied invariant on `store` that `target.len()` must be `> T::len()`, and that breaking this invariant will lead to panic.

ethkey/src/extended.rs
```diff
@@ -228,24 +266,24 @@ mod derivation {
 		// curve point (compressed public key) --  index
 		//             0.33                    --  33..37
 		data[0..33].copy_from_slice(&public_serialized);
-		BigEndian::write_u32(&mut data[33..37], index);
+		index.store(&mut data[33..37]);
```
rphmeier: should use `T::len()`

gavofyork:
looks like this only supports hard derivation for > 31 bit indexes, no? it's important to support soft derivation for 256-bit indexes...

NikVolf:
@gavofyork Nope, for any :)
There is a test for it `h256_soft_match`

gavofyork:
cool

ethkey/src/extended.rs
```diff
@@ -21,6 +21,53 @@ use Public;
...
+			Derivation::Hard(index)
+		}
+	}
+}
+
+impl Label for H256 {
+	fn len() -> usize { 32 }
+
+	fn store(&self, target: &mut [u8]) {
+		self.copy_to(&mut target[0..4]);
```
gavofyork: what's happening here? only copying the first 4 bytes?

NikVolf: yeah right
i added it more like an example here, didn't payed much attention, thanx

PR #4440, opened 2017-02-06
Open popup without attempting inline

tomusdrw:
Opening popup outside of a frame processing user action (click event) will cause the popup to be blocked.

This fix only attempts inline installation when on the right domain.


js/src/views/Application/Extension/store.js
```diff
@@ -68,6 +68,19 @@ export default class Store {
   installExtension = () => {
     this.setInstalling(true);
 
+    if (window.location.hostname === 'home.web3.site') {
```
jacogr: Should probably just check for <anything).web3.site, however it is good getting around the issue until we figure out a way forward.

js/src/util/dapplink.js
```diff
@@ -16,7 +16,9 @@
 
 import base32 from 'base32.js';
 
-const BASE_URL = '.web.web3.site';
+import { DOMAIN } from './constants';
+
+const BASE_URL = `.web${DOMAIN}`;
```
jacogr: Thanks.

PR #4461, opened 2017-02-07
Fixing CORS headers for parity.web3.site

tomusdrw:


gavofyork:
might need patching?

tomusdrw:
`*.web3.site` is not in `1.5.2` at all afair

PR #4477, opened 2017-02-08
Signer provenance

tomusdrw:
Exposes signer provenance in RPC and displays it in the UI.

Currently IPC doesn't support metadata extraction, so we cannot properly identify IPC requests yet.
It should be added when https://github.com/ethcore/jsonrpc/issues/17 is implemented.

GUI:
screenshot from 2017-02-08 13-42-27


ethcore/src/trace/db.rs
```diff
@@ -305,7 +305,7 @@ impl<T> TraceDatabase for TraceDB<T> where T: DatabaseExtras {
 	}
 
 	fn trace(&self, block_number: BlockNumber, tx_position: usize, trace_position: Vec<usize>) -> Option<LocalizedTrace> {
-		let trace_position_deq = trace_position.into_iter().collect();
+		let trace_position_deq = trace_position.into_iter().collect::<VecDeque<usize>>();
```
rphmeier: can be more efficiently done with a simple `trace_position.into()`

tomusdrw: :+1: Good point. Had to touch this line since it wasn't compiling on latest nightly.

js/src/views/Signer/components/RequestOrigin/requestOrigin.js
```diff
@@ -0,0 +1,119 @@
...
+
+import styles from './requestOrigin.css';
+
+export default class RequestOrigin extends Component {
+  static contextTypes = {
+    api: PropTypes.object.isRequired
+  };
+
+  static propTypes = {
+    origin: PropTypes.oneOfType([
```
derhuerst: I'd prefer having two props, e.g. `type` and `origin` here, one containing `rpc`/`ipc`/etc and one containing the data necessary to display it. Better to read and cleaner than passing an object as a prop.

js/src/views/Signer/components/RequestOrigin/requestOrigin.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      <RequestOrigin origin={ { ipc: '0x1234' } } />,
+      context
+    ).text()).to.equal('Requested by IPC session<Connect(IdentityIcon) />');
+  });
+
+  it('renders signer', () => {
+    expect(shallow(
+      <RequestOrigin origin={ { signer: '0x12345' } } />,
+      context
+    ).text()).to.equal('Requested by UI session<Connect(IdentityIcon) />');
```
derhuerst: IMO "Requested by <user-here> from the UI" would be more intuitive.

gavofyork:
marked inprogress as build is failing.

js/src/views/Signer/components/RequestOrigin/requestOrigin.js
tomusdrw: K, will pre-process RPC response.

js/src/views/Signer/components/RequestOrigin/requestOrigin.spec.js
tomusdrw: so `by` or `from`?

derhuerst: both. "requested by <identity icon> from the UI".

tomusdrw: But `<identity icon>` doesn't mean an account, but UI session.

rphmeier:
Core portion LGTM

PR #4479, opened 2017-02-08
simplify npm release process

derhuerst:
When trying to use `@parity/parity.js`, I noticed a lot of significant shortcomings of our npm-published libraries as well as some minor problems. **This PR intends to create a starting point to improve the shape of these libraries.**

I tried to make the build & publish process more transparent. Instead of copying files using the Webpack copy plugin into a temp directory, I now does a plain `cp -r src/something npm/library-name/src` and then transpiles these files using plain Babel  (see `npm run scripts/dryrun-npm.sh`). 

This means that we publish ES5 source code as individual files instead of a bundle including all dependencies & polyfills. There's a few advantages involved:

- consuming `@parity/parity.js`, you can now see what's going on in a call to the library, e.g. if you passed invalid parameters and the error messages are not 100% helpful
- stack traces, including source maps should work
- having the bare transpiled source code, the bundler of *my consuming project* can decide wether to bundle for browsers or for node, and which libraries to bundle or to exclude.

Because this PR deprecates the `.npmjs` temp directory, it would make sense if we handle `npm/library-name` as the canonical root directory of our npm-published libraries.

- just like before, we need to make sure all dependencies are properly declared in its `package.json`, especially as we'll (probably) remove more polyfills & tooling from the published library (see https://github.com/ethcore/parity/pull/4448#issuecomment-277748842)
- [we pickup the version of our root `package.json`, put it into the library `package.json`](https://github.com/ethcore/parity/blob/4c6f20d654f9fe6924ebd7eb846c85d5b8d70d21/js/scripts/release.sh#L77) and commit these changes. this makes it transparent which library will be published and gives us a better chance to move to a server-friendly versioning scheme in the future.

I also changed the smoke tests inside `npm/library-name/test` to work with the core `assert` module. This has the advantage that the tests can live where the library gets copied to and we don't need to install additional dependencies in a subfolder.

----

Moving on from this PR, I see the following improvements to be made (but this is IMO):

- remove all polyfills from the libraries and state in their readme that these are required. this prevents duplicate code of consuming users and enables them to choose how & where to put these dependencies in place. this is how the majority of the well-known maintained libraries on NPM does it.
- we might fully move the code into `npm/library-name`. this has several advantages:
  - stability: we don't randomly publish new libraries whenever a UI build succeeds
  - testability: independent from the large UI codebase, with a simpler setup
  - discoverability by 3rd parts users: people will be able to find & report bugs & shortcomings, as they may feel overwhelmed by the size of `src`.

----

Things still to be done:

- [x] remove `~` from `src/api` and `src/abi`. it's only a few occurrences that don't seem to be worth the build setup.
- [ ] test `scripts/release.sh` and `npm run clean` more. i'm sure i missed things
- [x] figure out where to put the E2E tests in `test/e2e`. there's also `test/npm*.js` that seem partially obsolete. – 0465034e9acbcd15ff062d460cdae90cb94d4a77
- [x] solve the issue with requiring/importing `parity.js` twice (it is impossible right now) – 	de56706
- [x] deal with #4448 – de56706
- [x] `npm/etherscan` depends on `api/util`

----

Tradeoffs made in this PR:

- As of de56706, we expect the users consuming our libraries to bundle & calls the required poly fills by themselves.
- ffd960c duplicates `abi/util/address` into `3rdparty/etherscan/util` in order to make `@parity/etherscan` completely independent from `@parity/parity.js`. A more elegant solution would be to dynamically include the required file, but this seems hard to do without a Webpack-like setup. (We dropped the Webpack setup because it lacks transparency and is very complicated).


js/scripts/dryrun-npm.sh
```diff
@@ -1,36 +1,46 @@
...
+pushd .; cd npm/jsonrpc
+npm test
+popd
+
+echo "*** Building etherscan for NPM"
+cp -r src/3rdparty/etherscan npm/etherscan/src
+env LIBRARY=etherscan npm run ci:build:npm
 
-  cd $DIRECTORY
+pushd .; cd npm/jsonrpc
```
ngotchac: `etherscan`

js/scripts/dryrun-npm.sh
```diff
@@ -1,36 +1,46 @@
...
-  echo "*** Publishing $PACKAGE from $DIRECTORY"
-  echo "npm publish --access public || true"
-  cd ../..
+echo "*** Building shapeshift for NPM"
+cp -r src/3rdparty/shapeshift npm/shapeshift/src
+env LIBRARY=shapeshift npm run ci:build:npm
 
-done
-cd ..
+pushd .; cd npm/jsonrpc
```
ngotchac: `shapeshift`

js/scripts/dryrun-npm.sh
```diff
@@ -1,36 +1,46 @@
...
+npm test
+popd
+
+echo "*** Building parity.js for NPM"
+cp -r src/abi npm/parity/src/abi
+cp -r src/api npm/parity/src/api
+env LIBRARY=parity npm run ci:build:npm
+
+pushd .; cd npm/jsonrpc
```
ngotchac: `cd npm/parity`

derhuerst:
Note: We need to properly test `scripts/release.sh` before this gets merged, as we have automatic publishing to npm.

derhuerst:
> Why not having one script for the `dryrun` and not `dryrun` version of the npm scripts, one with one extra-line for publishing ? Right now one of them has some issues, so it would prevent that and having to maintain two scripts

@ngotchac Did this in 5500e21.

ngotchac:
I updated a bit the PR. Everything seems to work fine, but we might want to test every possibility before merging. Might also want to update as a Major release, since we got rid of the babel + promise polyfills

jacogr:
Don't do this - 

`
ffd960c duplicates abi/util/address into 3rdparty/etherscan/util
`

We cannot maintain stuff (untested) in 2 places.

ngotchac:
@jacogr Aha we were talking about this with @derhuerst , knowing someone would point this out. So what would be the solution ? Have the whole `parity.js` dependency for only one util function ? I don't think that it's a good solution

ngotchac:
Ok so I added a symlink in the Etherscan folder that points to the right file in the `abi/util` folder. The file gets copied by following the link during the script execution.

derhuerst:
Manually checked the npm libs to be published with `npm pack` and fixed a few things in 0760a85. Should work now. Haven't checked the publishing process yet though.

ngotchac:
Looks good to me !

jacogr:
1. `npm/*/LICENSE` - just a copy of the main LICENSE - I would suggest just copying the in since it serves no purpose to duplicate the full text. If the license ever changes, the folder can contain a own version. (As of now, it just adds bloat)

2. Additional comments made as per the sources.

js/npm/etherscan/test/smoke.spec.js
```diff
@@ -14,21 +14,8 @@
...
-}
-
-if (isNode) {
-  global.fetch = require('node-fetch');
-}
-
-import ShapeShift from './3rdparty/shapeshift';
-
-export default ShapeShift;
+assert.doesNotThrow(() => {
```
jacogr: I am really not sure why we are doing it this way around. I would strongly suggest using `describe` and `it` blocks to be consistent with the rest of the codebase instead of using clever helpers.

```
describe('@parity.js/etherscan', () => {
  it('load the built library', () => {
    expect(() => {
      const etherscan = require('..');
    }).to.not.throw;
  });
});
```

works as well, is consistent and readable by all.

js/npm/jsonrpc/test/smoke.spec.js
```diff
@@ -14,21 +14,12 @@
...
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import 'babel-polyfill/dist/polyfill.js';
-import es6Promise from 'es6-promise';
-es6Promise.polyfill();
+const assert = require('assert');
 
-const isNode = typeof global !== 'undefined' && typeof global !== 'undefined';
-const isBrowser = typeof self !== 'undefined' && typeof self.window !== 'undefined';
+assert.doesNotThrow(() => {
```
jacogr: As above.

js/npm/parity/test/smoke.spec.js
```diff
@@ -14,13 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-const parity = require('../');
+const assert = require('assert');
 
-describe('load the Parity library', function () {
-  it('should no throw any error', () => {
-    expect(parity).to.be.ok;
+assert.doesNotThrow(() => {
```
jacogr: As above.

`describe` & `it` as per the rest.

js/npm/shapeshift/test/smoke.spec.js
```diff
@@ -0,0 +1,21 @@
...
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+const assert = require('assert');
```
jacogr: As above.

js/package.json
```diff
@@ -36,12 +36,12 @@
     "ci:build:app": "NODE_ENV=production webpack --config webpack/app",
     "ci:build:lib": "NODE_ENV=production webpack --config webpack/libraries",
     "ci:build:dll": "NODE_ENV=production webpack --config webpack/vendor",
-    "ci:build:npm": "NODE_ENV=production webpack --config webpack/npm",
-    "ci:build:jsonrpc": "babel-node ./scripts/build-rpc-json.js --output .npmjs/jsonrpc",
+    "ci:build:npm": "NODE_ENV=production babel npm/$LIBRARY/src --out-dir npm/$LIBRARY/dist --ignore spec.js --source-maps",
+    "ci:build:jsonrpc": "babel-node ./scripts/build-rpc-json.js > npm/jsonrpc/index.json",
```
jacogr: Please no redirect. Do it in the script so it is obvious where things end up according to the actual source.

js/scripts/build-rpc-json.js
```diff
@@ -67,4 +60,4 @@ Object.keys(interfaces).sort().forEach((group) => {
   });
 });
 
-fs.writeFileSync(INDEX_JSON, JSON.stringify({ methods: methods }, null, 2), 'utf8');
+process.stdout.write(JSON.stringify({ methods: methods }, null, 2));
```
jacogr: As above.

js/scripts/prepare-npm-libraries.sh
```diff
@@ -0,0 +1,63 @@
...
+mkdir -p npm/shapeshift/src
+cp -R src/3rdparty/shapeshift/* npm/shapeshift/src
+env LIBRARY=shapeshift npm run ci:build:npm
+
+pushd .; cd npm/shapeshift
+npm test
+popd
+
+# exit with exit code
+exit 0
```
jacogr: +1 like this approach

js/src/3rdparty/etherscan/util/address.js
```diff
@@ -0,0 +1 @@
+../../../abi/util/address.js
```
jacogr: ? 

1. Why are we adding another file here?
2. There is no differences between having a reference to API vs. ABI. When using an exposed utility (unless you are the API), best to point to the API, not ABI.

js/test/npmJsonRpc.js
```diff
@@ -15,13 +15,13 @@
...
 try {
-  var JsonRpc = require('../.npmjs/jsonrpc/library.js').default;
+  const JsonRpc = require('../npm/jsonrpc').default;
 
   if (typeof JsonRpc !== 'object') {
     throw new Error('JsonRpc');
   }
 
-  console.log(JsonRpc);
+  console.log('JSON RPC Endpoints:', Object.keys(JsonRpc));
```
jacogr: Think the original was actually a debug that slipped through, however this makes (slightly) more sense.

js/webpack/libraries.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     // library
     'inject': ['./inject.js'],
     'web3': ['./web3.js'],
-    'parity': ['./parity.js']
+    'parity': ['./library.parity.js']
```
jacogr: We have things mixed up here. This version of parity is supposed to be the version served via `/parity-utils/{inject,parity}.js` - library refers to the npm version.

js/npm/etherscan/test/smoke.spec.js
ngotchac: I think it's just not to introduce dependencies as this a the built-in `assert` package

js/src/3rdparty/etherscan/util/address.js
ngotchac: This is Github way of showing a Symlink. This is related to what was discussed above.

js/package.json
derhuerst: Well in the script it hasn't been obvious before. I changed it to redirecting because it seems infinitely more flexible, e.g. when debugging, and follows unix principles.

Will change it if you insist. 😉

js/src/3rdparty/etherscan/util/address.js
derhuerst: See https://github.com/ethcore/parity/pull/4479#issuecomment-278604405

js/webpack/libraries.js
derhuerst: Fully agree that it's not a for the library. @ngotchac and I discussed renaming it to `dev.parity.js`. What do you think?

js/npm/etherscan/test/smoke.spec.js
jacogr: It is a dev dependency. I'd rather not have things done in disparate ways all across the codebase.

js/webpack/libraries.js
jacogr: It is not dev.parity either :)

It is actually the version of parity.js that gets used by all dapps running inside Parity. e.g. https://github.com/gavofyork/gavcoin/blob/master/index.html#L11

js/src/3rdparty/etherscan/util/address.js
jacogr: Ok, so why is the symlink not `../../../api/util/something` ?

In all honestly, the address stuff does not actually belong in the ABI, it just "landed" there.

js/webpack/libraries.js
derhuerst: If I don't confuse things, we can move `src/parity.js` to `npm/parity/src/index.js` and rename `src/library.parity.js` to `src/parity.js`.

js/npm/etherscan/test/smoke.spec.js
derhuerst: IMO i'd rather not have mocha just for making sure things don't crash. Will change it anyway.

js/src/3rdparty/etherscan/util/address.js
derhuerst: > There is no differences between having a reference to API vs. ABI. When using an exposed utility (unless you are the API), best to point to the API, not ABI.

I'm curious, why is it better?

js/npm/etherscan/test/smoke.spec.js
ngotchac: (The only issue I see is that the error message is not displayed with `assert`, which is not the best)

js/src/3rdparty/etherscan/util/address.js
jacogr: It is not part of the public interface of the ABI. The API explicitly exposes it and will maintain it. Basically referring it from the ABI is like using `_prop` on an interface - only really meant for internal consumption, it can change.

The API crosses that line to not duplicate code, but this ABI/API interaction still needs to find a clean way of handling this transgression - the reason it was done this way around was to only have dependencies from API -> ABI, not the other way around. (So the API import is wrong and it is a known wrong, but at least only a single wrong... we should not add more.)

js/webpack/libraries.js
jacogr: Yes, the changes here

https://github.com/ethcore/parity/pull/4479/files#diff-5a5d605811790e35c3e21f577912e89eR17

and here 

https://github.com/ethcore/parity/pull/4479/files#diff-161120f1d7582520482df9363dee21acR17

Are the ones where you went wrong. You tried to convert the one into the other, instead of understanding the purpose. (Changes basically copies the one to the other - more or less) Should have kept it, moved `library.parity.js` to `index.js` (like done elsewhere) and have been done, but rather these changes got them mixed.

js/npm/etherscan/test/smoke.spec.js
derhuerst: @ngotchac Not as nicely, yet but it says "requiring `@parity/foo` fails".

js/webpack/libraries.js
derhuerst: 5da69d7 renames `parity.js` to `parity.npm.js` and reverts `library.parity.js` to `parity.js`.

js/src/api/util/address.js
```diff
@@ -0,0 +1 @@
+../../abi/util/address.js
```
ngotchac: Why not moving the file in `api` ?

derhuerst: @ngotchac Maybe i missed something, but then `abi` would depend on `api`, which we try to prevent.

ngotchac: Not sure I fully followed the discussion about what goes in ABI vs. API. However it just looks weird to have a symlink to symlink.

jacogr: `~/abi/util` is not part of the exposed interfaces (i.e. stuff on Abi.*), so don't use it. 

Could kick the can for this further down the road and not fix it here since this part will break when that is eventually changed, but since this this whole PR is a non-critical addition (especially when it comes to sources for 3rd party), don't see the need to try rush without doing things properly.

jacogr:
Moving to got issues. 

First the grumble part -

- We are still duplicating the `LICENCE` files - copy them in as per original.
- We are still doing the redirects
- We are still not complying with code-base wide standards for testing

Secondly the has-issues part -

- Symlinks are a no-no on Windows. Possible suggestions (first thing that pops to mind)
  - pass an Api class/instance to etherscan so it can use `{Api,api}.util.*` - not perfect, but since we expect users to be in our ecosystem and this one is not bound to get millions of users, a trade-off.
  - Copy the file in when bundling (distant second)
- We are not dealing sufficiently well with the polyfill issue here. This PR brings back the first user-reported issue after the 1.4 release, the dreaded "Array.includes is not a function". As library providers, we can expect dep libraries to be there, polyfills are a slippery slope. Suggestions -
  - No ES2015/2016 type constructs allowed in the 3rdparty stuff (? refactoring where applicable)
  - We need to have something that deals with fetch. Users should just include and it works. (As it did with the webpack-compiled version - which initially had exactly the same issues as reported here)

derhuerst:
> First the grumble part -
>
> - We are still duplicating the `LICENCE` files - copy them in as per original.
> - We are still doing the redirects
> - We are still not complying with code-base wide standards for testing

Ok.

> - Symlinks are a no-no on Windows. Possible suggestions (first thing that pops to mind)
>   - pass an Api class/instance to etherscan so it can use `{Api,api}.util.*` - not perfect, but since we expect users to be in our ecosystem and this > one is not bound to get millions of users, a trade-off.

The idea behind this PR is to get clean, decoupled libraries published to NPM. What you suggest would move us in the opposite direction.

Maybe that sounds too harsh, but arguing that only because people use our Ethererum node we can force them to use our other stuff seems to be awfully close to [*embrace extend extinguish*](https://en.wikipedia.org/wiki/Embrace,_extend_and_extinguish). Arguing that the lib doesn't need to have a good quality (in a PR that tries to improve it) because it's never going to be often-used also seems weird.

> - Symlinks are a no-no on Windows. Possible suggestions (first thing that pops to mind)
> […]
>   - Copy the file in when bundling (distant second)

Will do that as a trade-off.

> - We are not dealing sufficiently well with the polyfill issue here. This PR brings back the first user-reported issue after the 1.4 release, the dreaded "Array.includes is not a function". As library providers, we can expect dep libraries to be there, polyfills are a slippery slope.

Polyfills are basically shared globals, `require`ing them is in many cases considered bad practice, especially if they're environment-specific. `require`ing `babel-polyfill` breaks any codebase that includes it somewhere else.

Regarding `Array.prototype.includes`: Shouldn't babel transpile that to `Array.prototype.indexOf`?

> - We need to have something that deals with fetch. Users should just include the library without having to polyfill anything and it works.

Strongly disagree. It is neither very hard nor uncommon to have users to `require`/`import` polyfills before using a library. In our case it would look like [this](https://github.com/ethcore/parity/blob/fb1cdc18ac53f7e9d0cb258ff6459a29a3c138aa/js/npm/parity/README.md#usage).

> (As it did with the webpack-compiled version - which initially had exactly the same issues as reported here)

As stated, the Webpack version doesn't work in codebases that `require` babel-polyfill somewhere else. It effectively makes it impossible to use parity.js with ES2015+ code. It contains the wrong `fetch` polyfill if run in non-browser environments.

jacogr:
Not sure what Babel does, `Array.includes` always creates an issue, hence using it as an example. (Which is actually very real, first support ticket on parity.js)

And sorry, but I'm not going to support anything that is going to create yet more overhead in "this doesn't work". We need to find an elegant polyfill solution if the current is not it. It is always very easy to throw out the baby with the bathwater, but you need to understand why the baby got dirty in the first place.

Dependent libraries are one thing, having to deal with other stuff lands us in the category of "this is too much effort". Steps should be -

1. npm install
2. it just works

And no, we are not at #2.

js/npm/etherscan/package.json
```diff
@@ -21,13 +22,16 @@
     "Parity",
     "Promise"
   ],
-  "scripts": {
+  "peerDependencies": {
+    "es6-promise": "^4.0.5",
+    "isomorphic-fetch": "^2.2.1",
+    "babel-polyfill": "^6.22.0"
```
jacogr: Really wondering if https://github.com/es-shims are maybe not a more elegant (can include it without issues) solution. Worth a go.

js/src/3rdparty/etherscan/util/address.js
```diff
@@ -0,0 +1 @@
+../../../api/util/address.js
```
jacogr: This should probably be removed now?

js/scripts/prepare-npm-libraries.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# build all packages
+echo "$NPM_TOKEN" >> ~/.npmrc
+
+printf "\n\n"
+printf "\n***************************************"
+printf "\n***** Building jsonrpc for NPM ********"
+printf "\n***************************************\n\n"
+npm run ci:build:jsonrpc
+cp LICENSE npm/jsonrpc/LICENSE
```
jacogr: +1

js/.gitignore
```diff
@@ -8,3 +8,7 @@ docs
 .happypack
 .npmjs
 .eslintcache
+
+npm/*/src
+npm/*/dist
+npm/jsonrpc/index.json
```
jacogr: Shouldn't the LICENCE files be here now as well since it is added after the fact?

js/npm/etherscan/README.md
```diff
@@ -1,34 +1,38 @@
...
 
-Usage -
+## usage
+
+You will need to include [`babel-polyfill`](https://github.com/babel/babel/tree/master/packages/babel-polyfill) as well as [a `Promise` polyfill](https://github.com/stefanpenner/es6-promise#readme) and [a `fetch` polyfill](https://github.com/matthew-andrews/isomorphic-fetch) if your target platform doesn't support them.
+
+```js
+require('babel-polyfill');
+require('es6-promise').polyfill();
+require('isomorphic-fetch');
```
jacogr: Really ridiculously complicated just to get started. Not an option. Been there, done that, got the scars.

(As indicated, we may get away with fetch, we may - indicating that you need to include it for it in a Node environment.)

js/npm/etherscan/package.json
jacogr: Scratch that.

Babel actually has the solution - the transform runtime plugin. Basically means only `babel-runtime` is added as a package dep, and the runtime includes the polyfill without polluting the global namespace. No Promise, polyfill and other include-at-head crainess. (Well, according to them - this is their library mode)

js/npm/etherscan/README.md
jacogr: For fetch - could... maybe... use the package.json browser field (all modern bundler support this), and for the node entry include `isometric-fetch` for the browser one `whatwg-fetch`.

(Together with `babel-runtime` it could get us to a "just works" point.)

derhuerst: [That is what `isomorphic-fetch` does.](https://github.com/matthew-andrews/isomorphic-fetch/blob/c22fcda/package.json#L5-L6). But people consuming the library want to be able to bundle it themselves. That's why it's in the readme.

gavofyork:
@jacogr can we do something with this?

jacogr:
Closing. Re-open PR when actually being worked on.

PR #4482, opened 2017-02-08
Fix key.meta.vault for root dir keys && read vault.meta without vault key

svyatonik:
Details:
1. When key is moved from vault back to the root directory, it's meta still contains the `vault` field. Fixed + added test on RPC level
2. As vault meta is intended to contain hint for vault password, it was a bad idea to allow reading meta for opened vaults only => added support for reading unopened vaults meta + updated test.

### DIFF
--- a/ethstore/src/dir/disk.rs
+++ b/ethstore/src/dir/disk.rs
@@ -234,6 +234,10 @@ impl<T> VaultKeyDirectoryProvider for DiskDirectory<T> where T: KeyFileManager {
 			})
 			.collect())
 	}
+
+	fn vault_meta(&self, name: &str) -> Result<String, Error> {
+		VaultDiskDirectory::meta_at(&self.path, name)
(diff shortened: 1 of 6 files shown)


svyatonik:
@jacogr

svyatonik:
3. Added support for vaults, created before vault.meta appeared (since it is 'no cost' && it fails if there is no meta field in vault files)

rphmeier:
Did these "old vaults" exist in any published release? Why bother supporting them if not?

svyatonik:
@rphmeier Nope, but meta field for vaults is actually optional (this is also true for key files: `KeyFile` works the same way:
https://github.com/ethcore/parity/blob/master/ethstore/src/json/key_file.rs#L101 ). So it seems logical to me to react as `return meta=None` if there's `"meta": null` in json, than to fail RPC request.

Issue #2235, opened 2016-09-21
Implement RPCs for light client


rphmeier:
Should figure out which APIs to implement.

`eth_call`, `eth_estimateGas`, and transaction replay all require fetching state data from the network -- this needs to be handled asynchronously

---

A tempting option is coroutines/context switching:
- Create each call request in a coroutine
- State backend which yields the coroutine with a description of a network request
- resume coroutines when network requests are ready.

With this strategy we could execute multiple transactions at a time using a maximum number of stacks, with a queue of pending transactions.

Unfortunately, [libfringe](https://github.com/nathan7/libfringe) is the best context-switching library Rust has, and the platform support is disappointing. `libmill` and `libdill` represent C alternatives.

---

Another option is to use asynchronous I/O, where we transform transaction execution into a finite-state machine:

``` rust
enum State {
    Partial(NetworkRequest, Executive),
    Finished,
}
```

and use something like futures to dispatch the request and continue execution until another needs to be made. This is more flexible than coroutines but implementation is more involved, but probably will move in a different direction than state backends.

---

...or just block on the network until each and every request is answered. If only one transaction is being executed at a time from RPC, this might be the best option, but it doesn't scale at all for any other use case.

rphmeier:
There's also the issue of how to abstract across multiple implementations of the RPCs -- My current run at the issue was https://github.com/ethcore/parity/pull/2072 but it's definitely not the best solution.

rphmeier:
The LES protocol is being extended with the ability to request peers to execute transactions. This will make most of these adjustments unnecessary.

rphmeier:
Now that the network implementation is more ready, it's time to revisit how stuff like state queries will be done.

Likely it will all be through the `light::Client` structure which will first look in its local cache and then fall back to the network. Given that a very high proportion of state/chain queries will need to go to the network, I think it's best to design these APIs with async use-cases in mind.

I think the best way to do this is with a `light::net::Handler` implementation with a handle stored within the client (independent from the `LightSync` handler which will have one-directional data flow). The functions on the `light::Client` will generally be simple forwarders to the `OnDemand` light protocol handler.

This handler should look for a peer which can serve the request, dispatch it, and save the callback for later once that request has a suitable response, retrying until a set timeout has passed.

Batching requests to peers using this mechanism will be tricky without an enforced timeout but might be possible. In general I think batching is only really feasible while syncing (where required data is predictable).

Issue #3958, opened 2016-12-23
Get rid of `Weak` in RPC


tomusdrw:
Afair we used `Weak` because the servers were not closed correctly upon dropping and it kept the whole `Client` running.

I'm pretty sure that the servers are closing correctly now, so usage of `Weak` shouldn't be necessary.

rphmeier:
It should actually be possible to use references instead given that the services are all created by the time the RPC is initialized.

5chdn:
This _refactoring_ labelled issue is inactive since last year. 

@paritytech/core-devs Please, confirm within 7 days if this is still relevant. If so, please comment below and add an assignee, or close it otherwise.

tomusdrw:
Yes, would be nice to have. Labelled as `Q2-easy` - it's a good task for hackathon or a beginner wanting to contribute.

andrebeat:
I think issue can be closed.

tomusdrw:
Correct, thanks! Resolved by #5688

PR #4485, opened 2017-02-08
Preparation for Light client RPC

rphmeier:
This PR gets us most of the way to RPCs for the current light client, and ports the existing full RPCs over to more generic abstractions instead of `MiningBlockChainClient` and `MinerService`.

@tomusdrw This touches a lot of code in the signer implementation (particularly to move it to futures). The tests pass, but let me know if you see anything suspicious. There will also undoubtedly be conflicts with #4477.

Key points:
- Skeleton of `eth` API
- Generic `Dispatcher` with `FullDispatcher` implementation
- `OnDemand` improvements

Next up:
  - Light client transaction queue
  - OnDemand service request batching
  - `LightDispatcher` (fetch nonce, historic gas price via OnDemand)

Part of #2235


ethcore/light/src/client/mod.rs
```diff
@@ -109,12 +107,12 @@ impl Client {
...
 	pub fn ready_transactions(&self) -> Vec<PendingTransaction> {
-		let best_num = self.chain.best_block().number;
+		let best = self.chain.best_header();
 		self.tx_pool.lock()
 			.values()
 			.filter(|t| match t.condition {
-				Some(TransactionCondition::Number(ref x)) => x <= &best_num,
-				Some(TransactionCondition::Timestamp(ref x)) => *x <= time::get_time().sec as u64,
+				Some(TransactionCondition::Number(x)) => x <= best.number(),
+				Some(TransactionCondition::Timestamp(x)) => x <= best.timestamp(),
```
tomusdrw: I think this should be a separate PR to have some discussion about it. 
IMHO it's not the right way to do it, we may start propagating transaction way too late.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -38,20 +41,142 @@ use v1::types::{
...
+		future::done(inner()).boxed()
+	}
+
+	fn dispatch_transaction(&self, signed_transaction: PendingTransaction) -> Result<H256, Error> {
+		let hash = signed_transaction.transaction.hash();
+
+		take_weak!(self.miner).import_own_transaction(&*take_weak!(self.client), signed_transaction)
+			.map_err(errors::from_transaction_error)
+			.map(|_| hash)
+		}
```
tomusdrw: Incorrect indentation

rpc/src/v1/impls/eth.rs
```diff
@@ -356,113 +356,144 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> Eth for EthClient<C, SN, S, M, EM> where
...
+					match client.balance(&address, id.into()) {
+						Some(balance) => Ok(balance.into()),
+						None => Err(errors::state_pruned()),
+					}
 				}
 			}
-		}
+		};
+
+		future::done(inner()).boxed()
```
tomusdrw: Maybe it's worth turning this pattern into macro? It will be greatly simplified when `Carrier` is stabilized anyway.

rpc/src/v1/impls/personal.rs
```diff
@@ -106,28 +95,41 @@ impl<C, M> Personal for PersonalClient<C, M> where
 	}
 
 	fn send_transaction(&self, meta: Metadata, request: TransactionRequest, password: String) -> BoxFuture<RpcH256, Error> {
-		let sign_and_send = move || {
-			let client = take_weak!(self.client);
-			let miner = take_weak!(self.miner);
+		let setup = || {
+			let dispatcher = self.dispatcher.clone();
 			let accounts = take_weak!(self.accounts);
```
tomusdrw: Perhaps it's worth to introduce `take_weak` alike macro for futures to avoid such `setup` steps.

rpc/src/v1/impls/signing.rs
```diff
@@ -50,105 +49,103 @@ enum DispatchResult {
...
-		let default_account = match default_account {
-			DefaultAccount::Provided(acc) => acc,
-			DefaultAccount::ForDapp(dapp) => take_weak!(self.accounts).default_address(dapp).ok().unwrap_or_default(),
+			Ok((self.dispatcher.clone(), accounts, default_account))
 		};
-		let payload = dispatch::from_rpc(payload, default_account, &*client, &*miner);
-		self.add_to_queue(payload)
+
+		let weak_signer = self.signer.clone();
+		future::done(setup())
```
tomusdrw: Identation here looks bad.
Could be simplified with `take_weak` for futures

```rust
let accounts = take_weakf!(self.accounts);
let signer = take_weakf!(self.signer);
let default_account = match default_account {
  DefaultAccount::Provided(acc) => acc,
  DefaultAccount::ForDapp(dapp) => accounts.default_address(dapp).ok().unwrap_or_default(),
};

dispatch::from_rpc(payload, default_account, &self.dispatcher)
  .and_then(|payload| {
     ...
   })
   .boxed()
```

rpc/src/v1/impls/signing.rs
```diff
@@ -170,67 +167,78 @@ impl<C: 'static, M: 'static> ParitySigning for SigningQueueClient<C, M> where
...
-				e => ready.complete(Err(errors::internal("Unexpected result.", e))),
-			}
-		});
-
-		p.then(|result| futures::done(result.expect("Ready is never dropped nor canceled."))).boxed()
+
+		// when dispatch is complete
+		res.then(move |res| {
+			// register callback via the oneshot sender.
+			handle_dispatch(res, move |response| {
```
tomusdrw: Most probably `handle_dispatch` could now return Future instead of accepting a callback (but it's a separate issue)

rpc/src/v1/impls/signing.rs
```diff
@@ -170,67 +167,78 @@ impl<C: 'static, M: 'static> ParitySigning for SigningQueueClient<C, M> where
...
+		res.then(move |res| {
+			handle_dispatch(res, move |response| {
+				match response {
+					Ok(RpcConfirmationResponse::Signature(sig)) => ready.complete(Ok(sig)),
+					Err(e) => ready.complete(Err(e)),
+					e => ready.complete(Err(errors::internal("Unexpected result.", e))),
+				}
+			});
+
+			p.then(|result| futures::done(result.expect("Ready is never dropped nor canceled."))).boxed()
```
tomusdrw: `.boxed()` unecessary

rpc/src/v1/impls/signing.rs
```diff
@@ -170,67 +167,78 @@ impl<C: 'static, M: 'static> ParitySigning for SigningQueueClient<C, M> where
...
+			handle_dispatch(res, move |response| {
+				match response {
+					Ok(RpcConfirmationResponse::Decrypt(data)) => ready.complete(Ok(data)),
+					Err(e) => ready.complete(Err(e)),
+					e => ready.complete(Err(errors::internal("Unexpected result.", e))),
+				}
+			});
+
+			// and wait for that to resolve.
+			p.then(|result| futures::done(result.expect("Ready is never dropped nor canceled."))).boxed()
```
tomusdrw: Unecessary `.boxed()`

rpc/src/v1/impls/signing_unsafe.rs
```diff
@@ -38,106 +37,101 @@ use v1::types::{
...
 	}
 
-	fn handle(&self, payload: RpcConfirmationPayload, account: DefaultAccount) -> Result<RpcConfirmationResponse, Error> {
-		let client = take_weak!(self.client);
-		let miner = take_weak!(self.miner);
-		let accounts = take_weak!(self.accounts);
+	fn handle(&self, payload: RpcConfirmationPayload, account: DefaultAccount) -> BoxFuture<RpcConfirmationResponse, Error> {
+		let setup = move || {
+			let accounts = take_weak!(self.accounts);
+			let default_account = account;
```
tomusdrw: Why is this line needed? Similar to `dispatch` could be greatly simplified with `take_weak` for futures

rpc/src/v1/impls/signing.rs
```diff
@@ -50,105 +49,103 @@ enum DispatchResult {
...
-					on_response(result.unwrap_or_else(|| Err(errors::request_rejected())))
-				})
-			},
-			Err(e) => on_response(Err(e)),
-		}
-	}
+	fn dispatch(&self, payload: RpcConfirmationPayload, default_account: DefaultAccount) -> BoxFuture<DispatchResult, Error> {
+		let setup = move || {
+			let accounts = take_weak!(self.accounts);
+			let default_account = default_account;
```
tomusdrw: Superflous line?

rpc/src/v1/tests/mocked/signing.rs
```diff
@@ -91,9 +95,17 @@ fn should_add_sign_to_queue() {
 
 	// then
 	let promise = tester.io.handle_request(&request);
-	assert_eq!(tester.signer.requests().len(), 1);
-	// respond
-	tester.signer.request_confirmed(1.into(), Ok(ConfirmationResponse::Signature(0.into())));
+
+	// the future must be polled at least once before request is queued.
+	let signer = tester.signer.clone();
+	::std::thread::spawn(move || loop {
```
tomusdrw: We need to figure out how to do it better :/ Will think about it :)

rpc/src/v1/impls/signing.rs
rphmeier: rustc can't deduce the move of `default_account` without it.

rpc/src/v1/impls/signing_unsafe.rs
rphmeier: forces rustc to recognize the fact that `account` has been moved into the closure, otherwise we'd have to `clone()` it.

ethcore/light/src/client/mod.rs
rphmeier: This is just the change discussed here: https://github.com/ethcore/parity/pull/4419#discussion_r99390569

What's wrong with it?

rpc/src/v1/impls/eth.rs
rphmeier: Yes, might be worth doing. I don't like to hide the allocation but I suppose it's evident in the return type.

rpc/src/v1/impls/personal.rs
rphmeier: No, the `setup` phase explicitly allows us to avoid multi-level `BoxFuture`s where possible. If we used a `take_weak`-like macro in an `and_then` combinator, we'd be restricted to having the `and_then` closure return `BoxFuture`.

ethcore/light/src/client/mod.rs
tomusdrw: Pending transactions might be first propagated a little bit too late (we need to first import the block with given timestamp and only after that the propagation starts).
It's probably OK, but we need to make sure to indicate this in docs/ui so that people understand the consequences.

rphmeier: Yes, but it protects against issues with local time being out-of-sync with "chain time". Documentation is definitely key.

rpc/src/v1/impls/signing.rs
rphmeier: right, that'll involve a bit more digging into the internals of the signer queue (to get rid of the old async callback stuff) but should be done eventually.

tomusdrw: Yup, probably it's best to log it (https://github.com/ethcore/parity/issues/4490) and leave as a separate PR.

rpc/src/v1/impls/eth.rs
rphmeier: @tomusdrw Can we leave as-is for now? The PR which finishes up the `eth` + `eth_filter` implementation can clean things up. For now, I want to see what patterns actually emerge and need visual optimizing.

rphmeier:
@tomusdrw Grumbles (except for the test stuff) should be addressed now.

rpc/src/v1/tests/mocked/signing.rs
rphmeier: Can we log that as a separate issue, at least?

tomusdrw: Sure.

rpc/src/v1/impls/eth.rs
```diff
@@ -356,113 +356,144 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> Eth for EthClient<C, SN, S, M, EM> where
...
 	}
 
-	fn block_transaction_count_by_hash(&self, hash: RpcH256) -> Result<Option<RpcU256>, Error> {
-		Ok(
-			take_weak!(self.client).block(BlockId::Hash(hash.into()))
-				.map(|block| block.transactions_count().into())
-		)
+	fn block_transaction_count_by_hash(&self, hash: RpcH256) -> BoxFuture<Option<RpcU256>, Error> {
+		let inner = || {
```
gavofyork: what's the point of this `inner` pattern? doesn't it just add an extra 3 lines of useless code into every function?

rpc/src/v1/impls/eth.rs
gavofyork: i'm really not very keen on introducing such a swathe of boilerplate.

gavofyork:
not very happy about the additional of boilerplate. is there nothing that can be done?

rphmeier:
It serves a purpose: convert result-using code into future-using code. Two options:
  - use the just-written `take_weakf!` and pay more interleaved "boilerplate" cost with stuff like `return future::err(e).boxed()` instead of `return Err(e)`
  - write a macro to wrap the function body and compile down to the `let inner = || {...}` pattern used explicitly here. Still introduces the visual noise + rightwards drift of the status quo.

tomusdrw:
There is no additional cost with `teak_weakf!`, in both examples we return one `.boxed()` future. There is an issue to get rid of those `Weak` references: #3958  that would also reduce the boilerplate.

rphmeier:
I understand there's no more runtime cost, I mean you still pay a boilerplate cost in places, just more spread out.

rphmeier:
Ok, no more `let inner` pattern via `take_weakf` and `try_bf` macros.

tomusdrw:
Some conflicts to resolve.

rphmeier:
(resolved locally, but can't build after #4486)

PR #4486, opened 2017-02-08
Ledger wallet support

arkpar:
@General-Beck `libudev-dev` is now required to build on linux
@tomusdrw please review RPC/signer changes


hw/src/ledger.rs
```diff
@@ -0,0 +1,364 @@
...
+	EthereumClassic,
+}
+
+/// Hardware waller error.
+#[derive(Debug)]
+pub enum Error {
+	/// Ethereum wallet protocol error.
+	ProtocolError(&'static str),
+	/// Hidapi error.
+	UsbError(hidapi::HidError),
```
rphmeier: style: no Error suffix

hw/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+
+pub use ledger::KeyPath;
+
+/// Hardware waller error.
+#[derive(Debug)]
+pub enum Error {
+	/// Ledger device error.
+	LedgerDeviceError(ledger::Error),
+	/// USB error.
+	UsbError(libusb::Error),
```
rphmeier: style: no Error suffix

parity/account.rs
```diff
@@ -18,7 +18,7 @@ use std::path::PathBuf;
 use ethcore::ethstore::{EthStore, SecretStore, import_accounts, read_geth_accounts};
 use ethcore::ethstore::dir::RootDiskDirectory;
 use ethcore::ethstore::SecretVaultRef;
-use ethcore::account_provider::AccountProvider;
+use ethcore::account_provider::{AccountProvider, AccountProviderSetttings};
```
rphmeier: `Setttings` -> `Settings`

parity/params.rs
```diff
@@ -184,6 +185,7 @@ impl Default for AccountsConfig {
 			testnet: false,
 			password_files: Vec::new(),
 			unlocked_accounts: Vec::new(),
+			enable_hardware_wallets: false,
```
rphmeier: shouldn't this be `true` if it's supposed to be opt-out?

arkpar: It is not needed for most use cases of account management such as creating or listing accounts.

ethcore/src/account_provider/mod.rs
```diff
@@ -278,14 +328,36 @@ impl AccountProvider {
 		Ok(r)
 	}
 
+	/// Returns each hardware account along with name and meta.
+	pub fn hardware_accounts_info(&self) -> Result<HashMap<Address, AccountMeta>, Error> {
+		let r: HashMap<Address, AccountMeta> = self.hardware_accounts()?
```
tomusdrw: Is type annotation necessary here?

hw/src/ledger.rs
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	fn read_wallet_address(handle: &hidapi::HidDevice, key_path: KeyPath) -> Result<Address, Error> {
+		let ver = Self::send_apdu(handle, commands::GET_APP_CONFIGURATION, 0, 0, &[])?;
+		if ver.len() != 4 {
+			return Err(Error::Protocol("Version packet size mismatch"));
+		}
+
+		let (major, minor, patch) = (ver[1], ver[2], ver[3]);
+		if major <= 1 && minor == 0 && patch < 3 {
```
tomusdrw: Is this condition correct? Should it mean that anything greater then 1.0.3 is supported?
If so then it needs to be:
```rust
if major < 1 || (major == 1 && minor == 0 && patch < 3) 
```
Currently version `0.x.0` (`x > 0`) will be treated as correct, no?

rpc/src/v1/impls/parity.rs
```diff
@@ -138,6 +138,22 @@ impl<C, M, S: ?Sized, U> Parity for ParityClient<C, M, S, U> where
 		)
 	}
 
+	fn hardware_accounts_info(&self) -> Result<BTreeMap<String, BTreeMap<String, String>>, Error> {
```
tomusdrw: Could be `BTreeMap<H160, BTreeMap<String, String>>` or even better `BTreeMap<H160, HwAccountInfo>`

hw/src/ledger.rs
```diff
@@ -0,0 +1,364 @@
...
+		}
+
+		let eth_path = &ETH_DERIVATION_PATH_BE[..];
+		let etc_path = &ETC_DERIVATION_PATH_BE[..];
+		let derivation_path = match key_path {
+			KeyPath::Ethereum => eth_path,
+			KeyPath::EthereumClassic => etc_path,
+		};
+		let key_and_address = Self::send_apdu(handle, commands::GET_ETH_PUBLIC_ADDRESS, 0, 0, derivation_path)?;
+		if key_and_address.len() != 107 { // 1 + 65 PK + 1 + 20 Addr
```
svyatonik: Is it correct? 1+65+1+20 = 87, not 107. A bit confusing :)

arkpar: Address actually takes 40 because it comes ascii-hex encoded. Will update the comment.

hw/src/ledger.rs
```diff
@@ -0,0 +1,364 @@
...
+	/// Device with request key is not available.
+	KeyNotFound,
+	/// Signing has been cancelled by user.
+	UserCancel,
+}
+
+impl fmt::Display for Error {
+	fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+		match *self {
+			Error::Protocol(ref s) => write!(f, "Ledget protocol error: {}", s),
```
rphmeier: typo: Ledget

General-Beck:
`libudev-dev` added

gavofyork:
need `README` altering for linux build instructions, too?

arkpar:
already altered

hw/src/ledger.rs
```diff
@@ -0,0 +1,364 @@
...
+
+		// read response
+		chunk_index = 0;
+		let mut message_size = 0;
+		let mut message = Vec::new();
+		loop {
+			let mut chunk: [u8; HID_PACKET_SIZE] = [0; HID_PACKET_SIZE];
+			let chunk_size = handle.read(&mut chunk)?;
+			trace!("read {:?}", &chunk[..]);
+			if chunk_size < 5 || chunk[1] != 0x01 || chunk[1] != 0x01 || chunk[2] != APDU_TAG {
```
karalabe: If you copy our code, at least copy correctly. First should be `chunk[0]`.

arkpar: Ehm, I'm pretty sure this is correct. Why do you think it should be `[0]`?

arkpar: This is rust code btw

tomusdrw: It's checking `chunk[1]` twice so either one should be `chunk[0]` or second one is redundant.

karalabe: > This is rust code btw

You don't say.

The communication channel part of the header is 2 bytes, both `0x01`.

karalabe: @arkpar Take another look at the code you copied, it's correct there.

arkpar: hidapi sends HID report id in the first byte. I've noticed you are using `libusb` for communication. This is a poor choice because it does not allow communicating with HID devices on macOS.  I'm only using `libusb` for hotplug notifications - a feature that your code does not have at all. Perhaps you should be more careful with such accusations.

karalabe: I didn't say where you copied it from. Seems you're awfully familiar with it though.

karalabe: But ok, let's play. I dare you to publicly say you did not write this code based on the go-ethereum implementation.

arkpar: I've sure seen your PR  but I did not copy it. I've looked into the reference scala code mostly. In particular https://github.com/LedgerHQ/ledger-wallet-ethereum-chrome/blob/67864c09fc88fd744cd17a28ba9460e20987b898/src/main/scala/co/ledger/wallet/web/ethereum/core/device/LedgerTransportHelper.scala#L39
There's only a few ways you can write a simple function that splits data into 64 byte chunks and appends a simple header.

PR #4489, opened 2017-02-09
Fix Token Reg Dapp issues in Firefox

### LINKED ISSUE #4348: TokenReg validation doesn't work in Firefox
Also it doesn't display quite nice.
screenshot from 2017-01-30 10-31-15

ngotchac:
Fixes #4348

@tomusdrw I couldn't reproduce the modal layout issue, but fixed one. Let me know if it still breaks on your side.


tomusdrw:
Yes, layout looks fine although validation is still not working (Firefox 50.1.0).
To reproduce:
1. Click "Register a new token"
2. Enter any 3 characters to "Token TLA"

It will always indicate that TLA should be 3 characters long, bluring input doesn't help.

js/src/dapps/tokenreg/Inputs/Text/input-text.js
```diff
@@ -105,22 +105,22 @@ export default class InputText extends Component {
     const { validationType, contract } = this.props;
     const validation = validate(value, validationType, contract);
 
-    if (validation instanceof Promise) {
```
derhuerst: 👍

ngotchac: Yep, this wasn't working in Firefox...

derhuerst: and we shouldn't do it anyway. there's `Promise.resolve` and `npm i is-promise`

jacogr:
Issue for the TLA validation - https://github.com/ethcore/parity/issues/4334

js/src/dapps/tokenreg/Inputs/Text/input-text.js
gavofyork: why doesn't it work in firefox? it's valid JS isn't it?

jacogr: It doesn't work reliably with the `es6-promise` shim, which I'm guessing is included in FF. (`Bluebird` has the same issue.) 

Chrome doesn't show it since it has native Promises.

derhuerst: @gavofyork In JS, it is considered bad practice to assert `instanceof` in these cases. We only care about getting a thenable interface, which `Promise.resolve` should happily accept. See https://github.com/then/is-promise/issues/6

js/src/dapps/tokenreg/Inputs/Text/input-text.js
```diff
@@ -105,22 +105,22 @@ export default class InputText extends Component {
...
+    }, 50);
 
-      return validation
-        .then(validation => {
-          this.setValidation({
-            ...validation,
-            disabled: false,
-            loading: false
-          });
+    return Promise.resolve(validation)
```
jacogr: +1

js/src/dapps/tokenreg/Inputs/Text/input-text.js
```diff
@@ -105,22 +105,22 @@ export default class InputText extends Component {
     const { validationType, contract } = this.props;
     const validation = validate(value, validationType, contract);
 
-    if (validation instanceof Promise) {
+    const loadingTimeout = setTimeout(() => {
       this.setState({ disabled: true, loading: true });
+    }, 50);
```
jacogr: Doing this always feels very hacky, i.e. you are working around something you shouldn't be doing :) Only a mustn't-grumble.

ngotchac: Hmmm... It's more that it's not necessary to change the state twice if the request resolves in less than 50ms, no ?

PR #4491, opened 2017-02-09
Fix address and accounts links

### LINKED ISSUE #4314: Make address click-through to Parity view (non-external)
As requested by @tomusdrw 

TxList links for addresses to go through to `#/address/<address>` for accounts to go through to `#/accounts/<address>`

ngotchac:
Closes #4314 

In Signer and Transaction lists, redirect the links to Parity UI and not Etherscan

### DIFF
--- a/js/src/ui/TxList/TxRow/txRow.js
+++ b/js/src/ui/TxList/TxRow/txRow.js
@@ -16,8 +16,10 @@
 
 import moment from 'moment';
 import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import { Link } from 'react-router';
 
-import { txLink, addressLink } from '~/3rdparty/etherscan/links';
+import { txLink } from '~/3rdparty/etherscan/links';
 
 import IdentityIcon from '../../IdentityIcon';
 import IdentityName from '../../IdentityName';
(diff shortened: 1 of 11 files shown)


jacogr:
Failing tests.

ngotchac:
Fix test (and added 2 actually)

js/src/views/Signer/store.js
```diff
@@ -17,13 +17,16 @@
 import { isEqual } from 'lodash';
 import { action, observable } from 'mobx';
 
-export default class Store {
+export default class SignerStore {
   @observable balances = {};
   @observable localHashes = [];
 
-  constructor (api, withLocalTransactions = false) {
+  externalLink = '';
```
jacogr: _externalLink for non-exposed internal variables. (As per rest of the class)

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -138,4 +139,30 @@ export default class TxRow extends Component {
...
+
+    return `/addresses/${address}`;
+  }
 }
+
+function mapStateToProps (initState) {
+  const { accounts } = initState.personal;
+  const accountAddresses = Object.keys(accounts);
+
+  return () => {
```
jacogr: Why not a simple

```
return {
  accountAddresses
};
```

(As opposed to the function?)

jacogr:
I'm just wondering if `externalLink` is anything but `''` anywhere? Adds a lot of complexity (especially in terms of yet more change listeners), however I'm not seeing where it is set to any value that is not the default.

(I may be going blind)

js/src/ui/TxList/TxRow/txRow.js
ngotchac: It's only so it doesn't re-render on every Redux dispatched action (since the same Object is thus returned)

ngotchac:
It's mainly for the embedded Parity Bar for the Chrome Extension ! ;)

PR #4492, opened 2017-02-09
Fix ethstore build

arkpar:
Cargo 0.16.0 no longer builds ethstore binary with `cargo -p ethstore`. 
`cargo -p ethstore --bin ethstore` does not work either.


tomusdrw:
I have another fix for this:
```
# ethstore/Cargo.toml

[lib]

[[bin]]
name = "ethstore"
path = "src/bin/main.rs"
doc = false
```

arkpar:
Turns out the binary is built anyway, just to a different path.

PR #4501, opened 2017-02-09
Light Client transaction queue, initial LightDispatcher

rphmeier:
Goes after #4485

The TXQ only manages local transactions, but still distinguishes between current and future.
Doesn't do intensive verification because someone only hurts their own node by triggering pathological behavior. Nonces are initially guessed at based on what's provided in the transactions for an account (which will usually be accurate/fetched from network), but ancient TXs can still be culled.

The dispatcher uses the `Dispatcher` interface from #4485, but it's sort of incomplete: needs stuff like gas price corpus and gas amount estimation to be fully complete (this needs the request cache, on_demand request batching, and the very scary Remote Transaction Execution)


rpc/src/v1/helpers/dispatch.rs
```diff
@@ -129,31 +134,15 @@ impl<C: MiningBlockChainClient, M: MinerService> Dispatcher for FullDispatcher<C
...
-				);
-				let signed = try_bf!(
-					SignedTransaction::new(t.with_signature(signature, network_id))
-						.map_err(|e| {
-						  debug!(target: "miner", "Hardware wallet has produced invalid signature: {}", e);
-						  errors::account("Invalid signature generated", e)
-						})
-				);
-				WithToken::No(signed)
+				hardware_signature(&*accounts, address, t, network_id).map(WithToken::No)
```
NikVolf: how is this related to light client?

rphmeier: Just factoring out common code shared between the `FullDispatcher` and the `LightDispatcher`. The dispatcher's role is to fill optional fields of transaction requests, fulfill signer requests, and dispatch transactions onto the network.

rphmeier: (the reason for the split is that the light node must fetch next nonce, expected gas price, estimated gas from the network, while a full node has all this information available locally)

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -167,6 +156,119 @@ impl<C: MiningBlockChainClient, M: MinerService> Dispatcher for FullDispatcher<C
...
+				nonce: nonce,
+				action: filled.to.map_or(Action::Create, Action::Call),
+				gas: filled.gas,
+				gas_price: filled.gas_price,
+				value: filled.value,
+				data: filled.data,
+			};
+
+			if accounts.is_hardware_address(address) {
+				return hardware_signature(&*accounts, address, t, network_id).map(WithToken::No)
```
rphmeier: @NikVolf here you can see the invocation of the `hardware_signature` bit from the light dispatcher.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -167,6 +156,119 @@ impl<C: MiningBlockChainClient, M: MinerService> Dispatcher for FullDispatcher<C
...
+			condition: request.condition,
+		}).boxed()
+	}
+
+	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
+		-> BoxFuture<WithToken<SignedTransaction>, Error>
+	{
+		let network_id = None; // TODO: fetch from client.
+		let address = filled.from;
+		let best_header = self.client.block_header(BlockId::Latest)
```
NikVolf: seing a lot of the same expectation over and over again
worse drying it in the common function?

ethcore/light/src/client/mod.rs
```diff
@@ -16,25 +16,24 @@
...
 use ethcore::verification::queue::{self, HeaderQueue};
-use ethcore::transaction::{PendingTransaction, Condition as TransactionCondition};
 use ethcore::blockchain_info::BlockChainInfo;
 use ethcore::spec::Spec;
 use ethcore::service::ClientIoMessage;
 use ethcore::encoded;
 use io::IoChannel;
 
-use util::hash::{H256, H256FastMap};
+use util::hash::H256;
```
NikVolf: can be just `util::H256` with the other guys below i believe

rphmeier:
@NikVolf grumbles addressed

NikVolf:
lgtm 
but other guys might have too look also

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+impl TransactionQueue {
+	/// Import a pending transaction to be queued.
+	pub fn import(&mut self, tx: PendingTransaction) -> Result<TransactionImportResult, TransactionError>  {
+		let sender = tx.sender();
+		let hash = tx.hash();
+		let nonce = tx.nonce;
+
+	    let res = match self.by_account.entry(sender) {
+			Entry::Vacant(entry) => {
+				entry.insert(AccountTransactions {
```
tomusdrw: It seems that we will store a lot of old transactions, cause if all transactions from particular sender are imported we will remove the entry here.
But then if some node propagates it again we will insert it to the queue with `Assumed` nonce, and will never `cull` it unless the sender actually sends another transaction.

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+	}
+}
+
+// transactions associated with a specific account.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct AccountTransactions {
+	// believed current nonce (gotten from initial given TX or `cull` calls).
+	cur_nonce: CurrentNonce,
+	current: Vec<PendingTransaction>, // ordered "current" transactions (cur_nonce onwards)
+	future: BTreeMap<U256, PendingTransaction>, // "future" transactions.
```
tomusdrw: Could be just `HashMap`, no? Doesn't seem that we use ordering anywhere

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+	/// If the address has no queued transactions, then `None` will be returned
+	/// and the next nonce will have to be deduced via other means.
+	pub fn next_nonce(&self, address: &Address) -> Option<U256> {
+		self.by_account.get(address).map(AccountTransactions::next_nonce)
+	}
+
+	/// Get all transactions ready to be propagated.
+	/// `best_block_number` and `best_block_timestamp` are used to filter out conditionally
+	/// propagated transactions.
+	pub fn ready_transactions(&self, best_block_number: u64, best_block_timestamp: u64) -> Vec<PendingTransaction> {
```
tomusdrw: Should indicate that the transactions are not sorted by any priority (actually it will batch transactions from single sender togethe).
Since we propagate transactions in chunks some senders will always be a bit more priviledged by some node.

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+							for future in old_cur {
+								let future_nonce = future.nonce;
+								acct_txs.future.insert(future_nonce, future);
+							}
+
+							TransactionImportResult::Current
+						} else if idx == cur_len && acct_txs.current.last().map_or(false, |f| f.nonce + 1.into() != nonce) {
+							trace!(target: "txqueue", "Queued future transaction for {}, nonce={}", sender, nonce);
+							let future_nonce = nonce;
+							acct_txs.future.insert(future_nonce, tx.clone());
```
tomusdrw: Future queue should have a hard limit (probably current queue also), otherwise you can use arbitrary amount of nodes memory just by propagating future transactions that will never be included in the blockchain.

ethcore/light/src/transaction_queue.rs
rphmeier: Ordering relied on here: https://github.com/ethcore/parity/pull/4501/files#diff-e1ba00eb97ba4534d6c62db683387edbR229

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+			next_nonce = next_nonce + 1.into();
+		}
+	}
+}
+
+/// Light transaction queue. See module docs for more details.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TransactionQueue {
+	by_account: HashMap<Address, AccountTransactions>,
+	by_hash: H256FastMap<PendingTransaction>,
```
rphmeier: That's fair. I suppose the transactions can be shared with an internal `Rc`?

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+}
+
+impl TransactionQueue {
+	/// Import a pending transaction to be queued.
+	pub fn import(&mut self, tx: PendingTransaction) -> Result<TransactionImportResult, TransactionError>  {
+		let sender = tx.sender();
```
rphmeier: Seems right to me. Weirdness in online viewer is just line wrapping.

ethcore/light/src/transaction_queue.rs
tomusdrw: You're putting `TransactionQueue` into `RwLock`, so I'm afraid it needs to be `Arc`.

ethcore/light/src/transaction_queue.rs
rphmeier: Light nodes don't have the capability to validate transactions, so this queue isn't intended to store anything propagated from the network, only local stuff from RPC. The intended usage is to fetch actual nonces at the best block for all `queued_senders` and call `cull` using those.

ethcore/light/src/transaction_queue.rs
```diff
@@ -0,0 +1,421 @@
...
+						return Err(TransactionError::Old)
+					}
+
+					// lower our assumption until corrected later.
+					acct_txs.cur_nonce = CurrentNonce::Assumed(nonce);
+				}
+
+				match acct_txs.current.binary_search_by(|x| x.nonce.cmp(&nonce)) {
+					Ok(idx) => {
+						trace!(target: "txqueue", "Replacing existing transaction from {} with nonce {}",
```
rphmeier: With the intended use-case (storing transactions only of local origin), it makes sense to replace whenever the same nonce is encountered.

ethcore/light/src/transaction_queue.rs
rphmeier: Haven't implemented the transaction propagation code for the light node just yet. I'll add docs that it's in "arbitrary" order.

ethcore/light/src/transaction_queue.rs
tomusdrw: Right, but if you ever propagate transaction with lower gas price it will never be replaced in full nodes transaction queue (at least the ones using default implementations) giving the user false feeling that the transaction was replaced, although this could be left to be displayed in UI.

rphmeier: If the user wants to try with a lower gas price then they're welcome to, the least we can do is allow our implementation to propagate it to the network. What the peers do with the transaction is beyond our control. The fact that it would probably not work should be left to documentation instead of behavior.

rphmeier:
@tomusdrw Grumbles should be addressed

ethcore/light/src/transaction_queue.rs
```diff
@@ -202,14 +223,20 @@ impl TransactionQueue {
...
-		}).collect()
+		self.by_account.values()
+			.flat_map(|acct_txs| {
+				acct_txs.current.iter().take_while(|tx| match tx.condition {
+					None => true,
+					Some(Condition::Number(blk_num)) => blk_num <= best_block_number,
+					Some(Condition::Timestamp(time)) => time <= best_block_timestamp,
+				}).map(|info| info.hash)
+			})
+			.filter_map(|hash| self.by_hash.get(&hash))
```
tomusdrw: Might be good to put `.expect` here instead to track potential issues (had issues like this in standard transaction pool)

rphmeier: I chose to use `filter_map` specifically to avoid panics in this case. I think it's better for something not to be propagated than for the whole node to go down. Adding a warning here "Inconsistency detected between `by_account` and `by_hash`" makes sense to me, though.

PR #4504, opened 2017-02-09
Several fixes to the Wallet in general

### LINKED ISSUE #4304: Multisig wallet contract doesn't throw when sub-call throws.
Why i cant send ICN from my contract? Its says out of gas
Parity 1.5.0

ngotchac:
Closes #4304 

Other fixes (asked by @tomusdrw)
- [x] No warning in the Transfert modal if it would need confirmations (above remaining daily limit)
- [x] Page needs a refresh to show pending transactions (could not reproduce...)
- [x] Improve the Settings modifications
- [x] Click on notification when Wallet receives something doesn't go to the wallet page (but the account page)

### DIFF
--- a/js/src/modals/WalletSettings/walletSettings.js
+++ b/js/src/modals/WalletSettings/walletSettings.js
@@ -54,7 +54,7 @@ class WalletSettings extends Component {
           visible
           title={
             <FormattedMessage
-              id='ui.walletSettings.rejected.title'
+              id='walletSettings.rejected.title'
               defaultMessage='rejected'
             />
           }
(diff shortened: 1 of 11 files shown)


js/src/modals/WalletSettings/walletSettings.js
```diff
@@ -51,12 +52,27 @@ class WalletSettings extends Component {
       return (
         <Modal
           visible
-          title='rejected'
+          title={
+            <FormattedMessage
+              id='ui.walletSettings.rejected.title'
```
jacogr: Can we just make it `walletSettings.*`? `ui.*` is for components under `~/ui` - will just flatten out the produced files as well.

ngotchac: Sure, don't know why I put this...

ngotchac: Done

PR #4505, opened 2017-02-10
Fix TnC overflows on small screens

ngotchac:
TnC on startup were displaying two scroll bars on small screen

image


jacogr:
Related issue https://github.com/ethcore/parity/issues/4245

PR #4507, opened 2017-02-10
Fixed a warning and bumped libusb-sys

arkpar:


jacogr:
Confirmed that compilation on Mac OS X now works (for me) with this version of libusb. Thanks @arkpar

PR #4515, opened 2017-02-10
Key derivation in ethstore & rpc

NikVolf:
Now client can create derived address for any of his accounts and optionally save it as a new account (or just get raw address and it's up to him how to use it)

Two types of derivation:
- hierarchical deterministic wallet (kind of bitcoin-compatible standard of bip0032), where you provide a series of indices for one or more sequental derivations (for example, Soft(0)/Soft(1)/Hard(0))

- hash derivation where you provide just `H256` (for example, invoice or other content hash) and derivation type (soft/hard)

Right now it's up to client how to use derived keys - as an full account or just save it in address book after derivation (to use it later - client must then be derived again from the same account with the same parameters and `save: true`).

Another option to sign transaction with the derived key directly coming in following pr to keep things small here


ethstore/src/ethstore.rs
```diff
@@ -350,6 +382,54 @@ impl SimpleSecretStore for EthMultiStore {
...
+	fn sign_derived(&self, account_ref: &StoreAccountRef, password: &str, derivation: Derivation, message: &Message)
+		-> Result<Signature, Error>
+	{
+		let accounts = self.get(&account_ref)?;
+		for account in accounts {
+			// Skip if password is invalid
+			if !account.check_password(password) {
+				continue;
+			}
+			let extended = self.generate(account.crypto.secret(password)?, derivation)?;
```
rphmeier: is this a bottleneck? can we cache extended key-pairs?

ethstore/src/ethstore.rs
```diff
@@ -350,6 +382,54 @@ impl SimpleSecretStore for EthMultiStore {
...
+			let secret = extended.secret().secret();
```
rphmeier: would a `Deref` implementation make this less weird?

rphmeier:
It seems like we could unify the `_derived` and regular interfaces under the same function names with a `Derivation::None`/`Derivation::Identity` enum. Fewer functions to look up branches in consumers of this API.

arkpar:
Would be nice to adopt BIP-44 string format. Maybe in another PR.
These RPCs should also be implemented for hardware wallets eventually.

rpc/src/v1/impls/parity_accounts.rs
```diff
@@ -25,7 +25,7 @@ use ethcore::account_provider::AccountProvider;
 use jsonrpc_core::Error;
 use v1::helpers::errors;
 use v1::traits::ParityAccounts;
-use v1::types::{H160 as RpcH160, H256 as RpcH256, DappId};
+use v1::types::{H160 as RpcH160, H256 as RpcH256, DappId, Derivate, DerivateHierarchical, DerivateHash};
```
rphmeier: `Derivate` not a word, should be `Derive`

NikVolf:
@rphmeier Not sure if i want to break existing api just to introduce key derivation which is rather small and humble part of it

ethstore/src/ethstore.rs
NikVolf: it's not in release build at all

ethstore/src/ethstore.rs
NikVolf: well it's not that weird
first `secret()` is `ExtendedSecret` of the extended key pair and second `secret()` is a raw `H256` secret of this extended secret, i don't see logical reason to `Deref` them in something they are not in any way equal to

rphmeier: It's a strange API design for sure. Doesn't help readability that you have to dig into the internals to know what's happening.

```rust
let secret = extended.secret(); // I have the secret!
let secret = secret.secret(); // I have the...secret secret?
```

Would definitely make more sense (to read) as `x.secret().as_raw()` or something

NikVolf: yeah i didn't like it either, but when it comes to individual function names, their names are legit, it's just when they are used one after another weirdness shows :)

`as_raw` looks nice, yeah, will rename

NikVolf:
@arkpar 
Looks like it can be adopted by the wallet app using this rpc api
it will just derive address with specified node indices and that's it

PR #4532, opened 2017-02-13
Added vaults support to `ethstore-cli`

svyatonik:
following #4493


svyatonik:
This will also help in dealing with issues like this one: https://github.com/ethcore/parity/issues/4533

PR #4534, opened 2017-02-13
Fix contract queries bug

### LINKED ISSUE #4520: Contract constant return function of bytes32 array show wrong data
I'm interacting to a contract that returns a bytes32 array, when I query this function, parity shows wrong data.

Parity result:
image

Expected result:
image

For debug: https://testnet.etherscan.io/address/0x22240d285d8e7e2d38f5257298e93e29a74163c0#code

ngotchac:
Fixes #4520

Add an option to the `call` method (for functions of a contract) to output the raw tokens (`Token` class, with `type` and `value` attributes) instead of `token.value`.

Fixes some UI and logic bugs in the Address Selector too.

### DIFF
--- a/js/src/views/Contract/Queries/inputQuery.js
+++ b/js/src/views/Contract/Queries/inputQuery.js
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import BigNumber from 'bignumber.js';
 import React, { Component, PropTypes } from 'react';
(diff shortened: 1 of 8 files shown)


js/src/views/Contract/Queries/inputQuery.js
```diff
@@ -54,18 +54,33 @@ export default class InputQuery extends Component {
           className={ styles.methodTitle }
           title={ name }
         />
+        { this.renderError() }
         { this.renderContent() }
       </Card>
     );
   }
 
+  renderError () {
```
jacogr: We should really be using `newError` so we UI-wide consistent display.

PR #4540, opened 2017-02-14
Close on double-click for Signer Account selection

### LINKED ISSUE #4525: Add close on double-click for Signer account selection

ngotchac:
Closes #4525

Close only on double-click. Change default on simple click.
Also update the default in an optimistic way (not waiting for network request). The request will be valid 99% of the time. If not, it will just be reverted.

### DIFF
--- a/js/src/views/ParityBar/accountStore.js
+++ b/js/src/views/ParityBar/accountStore.js
@@ -24,7 +24,9 @@ export default class AccountStore {
   constructor (api) {
     this._api = api;
 
-    this.loadAccounts();
+    this.loadDefaultAccount()
+      .then(() => this.loadAccounts());
+
     this.subscribeDefaultAccount();
   }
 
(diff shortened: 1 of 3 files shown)


js/src/views/ParityBar/accountStore.js
```diff
@@ -33,7 +35,15 @@ export default class AccountStore {
   }
 
   @action setDefaultAccount = (defaultAccount) => {
-    this.defaultAccount = defaultAccount;
+    transaction(() => {
+      this.accounts = this.accounts.map((account) => {
+        account.default = account.address === this.defaultAccount;
```
jacogr: Is this correct? It basically sets the default to the previous default, i.e. this.defaultAccount is previous version since default is only set afterwards. Looks wrong. (If correct, a NOTE explanation may be in order here)

js/src/views/ParityBar/accountStore.js
```diff
@@ -47,13 +57,22 @@ export default class AccountStore {
...
+    this.setDefaultAccount(address);
+
     return this._api.parity
       .setNewDappsWhitelist(accounts)
       .catch((error) => {
         console.warn('makeDefaultAccount', error);
       });
   }
 
+  loadDefaultAccount () {
```
jacogr: It would probably make sense to add this to the spec.js as well (since the store is more-or-less tested already)

js/src/views/ParityBar/accountStore.js
ngotchac: Yes right...

js/src/views/ParityBar/accountStore.js
ngotchac: Added

Issue #4420, opened 2017-02-03
Parity sometimes crashes on account deletion


3esmit:
Sometimes I create accounts for testing porpouses, like testing recovery, or just to help someone with screenshots, and I don't even set a password for them.
Happened for me twice, that when I pressed Yes to delete, parity crashed in that moment, no error message displayed. I just reopened it using Parity Ethereum Windows shortcut. When parity reopens the account is gone, as expected, and all runs normal. 

The first time it happened I tried to delete with wrong password, it told me could not delete, then I put the right password and parity gone.
The second time I created an account to test if MEW accepted Parity recovery phrase and didn't set any password. tested there, come back to parity, deleted and parity gone. 

I'm running Windows 10 and Parity 1.5.0.

3esmit:
I was wondering about Windows telemetry/indexing services and imagined the situation where parity tries to delete a file when Windows is reading it. So I created an account and opened with notepad the account file. When I try to delete it in parity, parity crashes but is able to delete the account.
I tried again doing the same and parity didn't crashed. Also tried to mark "read only" but then parity says it cannot delete the account with no crashes.

rphmeier:
Have you got any console output? My guess is we're just hitting a panic of some kind in the deletion code.

3esmit:
Found out how to reproduce the bug.
Open more then one Parity UI tab. While other Tabs are looking at accounts page, use one of the UI to delete a account.

````
2017-02-08 10:24:52  Invalid key file: "C:\\Users\\Ricardo\\AppData\\Roaming\\Parity\\Ethereum\\keys\\ethereum\\UTC--2017-02-08T12-23-58Z--925abb68-93e0-8f48-7620-ef1abed3a2f6" (O sistema não pode encontrar o arquivo especificado. (os error 2))
thread '<unnamed>' panicked at 'Entry exists, because it was returned by `get`; qed', ../src/libcore\option.rs:705
stack backtrace:
   0:     0x7ff6bb2f68ba - <unknown>
   1:     0x7ff6bb2f3aca - <unknown>
   2:     0x7ff6bb2f44ad - <unknown>
   3:     0x7ff6bb2f4348 - <unknown>
   4:     0x7ff6bb2f4264 - <unknown>
   5:     0x7ff6bb2f41f9 - <unknown>
   6:     0x7ff6bb302d57 - <unknown>
   7:     0x7ff6bb302dbf - <unknown>
   8:     0x7ff6bb113709 - <unknown>
[...]

PR #4542, opened 2017-02-14
Remove invalid expectation (beta)

NikVolf:
#4420

code looks like a mess there, and invalid `expect` is like an icy on the cake

### DIFF
--- a/ethstore/src/ethstore.rs
+++ b/ethstore/src/ethstore.rs
@@ -271,7 +271,9 @@ impl SimpleSecretStore for EthMultiStore {
 			// Remove from cache
 			let mut cache = self.cache.write();
 			let is_empty = {
-				let mut accounts = cache.get_mut(address).expect("Entry exists, because it was returned by `get`; qed");
+				let mut accounts = cache.get_mut(address)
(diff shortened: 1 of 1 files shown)


rphmeier:
Build of rocksdb-sys failing with the gcc on the CI? @General-Beck

General-Beck:
@rphmeier no, build succeful

tomusdrw:
Can you elaborate why the expectation was incorrect? Seems that `fn get` always reads the account from cache, no?
So the only reason I see is that there is a race condition when trying to remove the same account twice, was there any other issue?

NikVolf:
@tomusdrw 
exactly this
with concurrent access the expectation can't be 100% correct, because there is no single lock  

and ui/user might duplicate delete account request so that it results in a panic

rphmeier:
@General-Beck this one looks pretty failed to me: https://gitlab.ethcore.io/parity/parity/builds/46155

arkpar:
I thought we don't support armv6. Where's the PR check for that target coming from?

NikVolf:
test fail is unrelated to the pr

PR #4544, opened 2017-02-14
Fix console dapp

tomusdrw:
- Enables linting for `console.js` and fixes all errors.
- Fixes `web3` RPC endpoint, adding CORS headers for signer.
- Exposes all possible custom methods as web3 extensions.


jacogr:
Really focussed only on the fixing part, which seems to cover the instances that were problematic.

Thanks for taking the time to do the eslint fixes as well, gave that a walk-though, much better. (The reformat closes https://github.com/ethcore/parity/issues/4157)

Issue #4172, opened 2017-01-13
Feature - expose state and chaindata via ipfs-http-api


kumavis:
This is a recommendation for an **MVP of Ethereum-IPFS integration** by exposing all hash-referenced ethereum objects (state trie and chaindata) via the `ipfs-http-api` spec. Lookups are performed by hash, results are returned as binary with no metadata.

### Requirements
http server that implements the "block.get" method of the `ipfs-http-api`
example uri:
```
http://localhost:5001/api/v0/block/get?arg=z43AaGF5tmkT9SEX6urrhwpEW5ZSaACY73Vw357ZXTsur2fR8BM
```

### Request handling flow
1. expose route:  `http://localhost:5001/api/v0/block/get?arg=${contentId}`

2. parse request url for `contentId`

3. parse contentId into { `type`, `hash` } via `multiformats/rust-cid`

4. lookup `type` by `hash`, return result as binary

where `type` is any of: 

| Type                 |      Name                                             |  Code |
|----------------------|-------------------------------------------------------|-------|
|`eth-block`           | Ethereum Block (block header)                                  | `0x90`|
|`eth-block-list`      | Ethereum Block List (block ommers)                             | `0x91`|
|`eth-tx-trie`         | Ethereum Transaction Trie Node             | `0x92`|
|`eth-tx`              | Ethereum Transaction                            | `0x93`|
|`eth-tx-receipt-trie` | Ethereum Transaction Receipt Trie Node     | `0x94`|
|`eth-tx-receipt`      | Ethereum Transaction Receipt                    | `0x95`|
|`eth-state-trie`      | Ethereum State Trie Node            | `0x96`|
|`eth-storage-trie`    | Ethereum Contract Storage Trie Node | `0x98`|

#### ipfs-http-api
spec repo: https://github.com/ipfs/http-api-spec
`block.get` spec (note spec doc has not been updated to use content-id (cid) instead of multihash)
https://github.com/ipfs/http-api-spec/blob/master/apiary.apib#L807-L921

#### content-id (CID)
multibase encoded ( see multibase )
format: `<cid version><data type><multihash>`
spec repo: https://github.com/ipld/cid
rust implementation: https://github.com/ipld/rust-cid

#### multibase
format: `<varint-base-encoding-code><base-encoded-data>`
spect repo: https://github.com/multiformats/multibase
cargo package: https://crates.io/crates/multibase

#### multihash
format: `<varint hash function code><varint digest size in bytes><digest>`
[...]

PR #4545, opened 2017-02-14
IPFS MVP

maciejhirsz:
Listening on port 5001 for IPFS-compatible calls as listed in #4172, supported thus far:
- `eth-block`
- `eth-block-list`
- `eth-tx`
- `eth-state-trie`


NikVolf:
no tests ;)

ipfs/src/handler.rs
```diff
@@ -0,0 +1,145 @@
...
+		.map(|part| &part[name.len() + 1..])
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+   #[test]
+	fn test_get_param() {
+		let query = "foo=100&bar=200&qux=300";
```
NikVolf: can't it be encoded like 
`foo%20bar=200?`

ipfs/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+					.expect("Static content type; qed")));
+
+				Next::write()
+			},
+			NotFound(reason) => {
+				res.set_status(StatusCode::NotFound);
+
+				res.headers_mut().set(ContentLength(reason.len() as u64));
+				res.headers_mut().set(ContentType("text/plain".parse()
+					.expect("Static content type; qed")));
```
NikVolf: i don't think that any request handling that can be triggered by arbitrary third party should ever contain `expect` or `unwrap`

NikVolf: But yeah... that's surely cant fail 

But there should be test that triggers this execution path

parity/run.rs
```diff
@@ -420,6 +421,10 @@ pub fn execute(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>) -> R
 	};
 	let signer_server = signer::start(cmd.signer_conf.clone(), signer_deps)?;
 
+
```
NikVolf: redundant empty line

ipfs/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+		.handle(move |_| IpfsHandler::new(client.clone()))
+		.map(|(listening, srv)| {
+
+			::std::thread::spawn(move || {
+				srv.run();
+			});
+
+			listening
+		})
+		.map_err(Into::into)
```
NikVolf: with the introduction of `Carrier` maybe construction a-la
```rust
	Ok(hyper::Server::http(&addr)?
		.handle(move |_| IpfsHandler::new(client.clone()))
		.map(|(listening, srv)| {

			::std::thread::spawn(move || {
				srv.run();
			});

			listening
		})?)
```

is better?

ipfs/src/lib.rs
maciejhirsz: I can change that to use a macro for mime types.

ipfs/src/handler.rs
rphmeier: indentation on `#[test]` seen here is also wrong

arkpar:
There should also be a CLI option disabling the server. I also suggest it should be disabled by default as most people don't really need this and it only increases attack surface.

rphmeier:
Agreed, the overhead of starting up the server here shouldn't be incurred unless in-use (and AFAIK Metamask will be the one user of this feature)

ipfs/src/handler.rs
```diff
@@ -0,0 +1,145 @@
...
+		self.out = Out::OctetStream(data);
+
+		Ok(())
+	}
+}
+
+/// Get a query parameter's value by name.
+pub fn get_param<'a>(query: &'a str, name: &str) -> Option<&'a str> {
+	query.split('&')
+		.find(|part| part.starts_with(name) && part[name.len()..].starts_with("="))
```
maciejhirsz: @tomusdrw no, it will create an empty slice and `starts_with` can't panic. I added a test for it now just in case.

ipfs/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "1.6.0"
+license = "GPL-3.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+
+[dependencies]
+ethcore = { path = "../ethcore" }
+ethcore-util = { path = "../util" }
+rlp = { path = "../util/rlp" }
+hyper = { default-features = false, git = "https://github.com/ethcore/hyper" }
+cid = "~0.2.0"
```
NikVolf: i believe we use like `cid = "0.2"` everywhere which is almost equivalent to `cid = "~0.2.0"` ?

maciejhirsz:
Think I got most of the grumbles so far, TODO:

- [ ] CLI switch, disabled by default.
- [ ] Origin header checking.

ipfs/src/handler.rs
maciejhirsz: @NikVolf I've switched the function to be private, so that it can only be consumed within the module. 

I think that as long as the consumer is aware that no URL encoding or decoding is performed it should be quite fine. It's unlikely that this crate will ever have a need for URL-encoded query strings by the virtue of all things IPFS going through multibase, specifically base58btc. Even if it ever does, it shouldn't be difficult to replace this by an external crate that can do it.

NikVolf:
looks cool to me
other guys have to look also

PR #4549, opened 2017-02-15
Tooltips with react-intl

### LINKED ISSUE #4551: Tooltip for Tabbar does not show up (only accounts show)

jacogr:
(Split from https://github.com/ethcore/parity/pull/4539)

- Allow react-intl messages with `~/ui/Tooltips/Tooltip`
- Change text elements to `FormattedMessage` where Tooltips are constructed
- Fix main Application Tabbar Tooltip not showing (Closes #4551)
- Simple smoke-tests for Tooltips/Tooltip

### DIFF
--- a/js/src/ui/Tooltips/Tooltip/tooltip.js
+++ b/js/src/ui/Tooltips/Tooltip/tooltip.js
@@ -15,12 +15,13 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 import React, { Component, PropTypes } from 'react';
+import { FormattedMessage } from 'react-intl';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { FlatButton } from 'material-ui';
-import ActionDoneAll from 'material-ui/svg-icons/action/done-all';
-import ContentClear from 'material-ui/svg-icons/content/clear';
-import NavigationArrowForward from 'material-ui/svg-icons/navigation/arrow-forward';
+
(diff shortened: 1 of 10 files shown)


js/src/ui/Tooltips/Tooltip/tooltip.js
```diff
@@ -30,15 +31,15 @@ let tooltipId = 0;
...
 class Tooltip extends Component {
   static propTypes = {
-    title: PropTypes.string,
-    text: PropTypes.string,
-    right: PropTypes.bool,
+    className: PropTypes.string,
     currentId: PropTypes.number,
     maxId: PropTypes.number,
-    className: PropTypes.string,
+    right: PropTypes.bool,
```
ngotchac: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md

Should we ? (PS, `r` is after `o` :laughing:)

jacogr: Funnily enough, I played with that some weeks ago - it does some nice stuff, e.g. required first, then normal, then handlers (`on`Something).

ngotchac: So shouldn't we use it ?

jacogr: We should, the only reason I have not pulled it in as part of one of my eslint updates (yet) is that we are basically wrong in 70% of the cases. So lot of small little niggly updates, just didn't have the time to fiddle.

Medium-term, I would love to.

Issue #4548, opened 2017-02-15
Extraneous certifier logging


jacogr:
Dev console fills up with warning logs -

```
Error: Certifier 0 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 1 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 2 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 3 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 4 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 6 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 7 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 10 does not exist.
(anonymous) @ middleware.js:183
middleware.js:183 Error: Certifier 11 does not exist.
```

from

```
.catch((err) => {
                  if (/does not exist/.test(err.toString())) {
                    return console.warn(err.toString());
                  }

                  console.warn(`Could not fetch certifier ${id}:`, err);
                });
```

Probably best to squash these "does not exist" errors, the normal error logged afterwards is a worth-while concern.

PR #4555, opened 2017-02-15
Fix pasting of value in Input fields

### LINKED ISSUE #4553: Pasting value into an input field erases the previous entered text
e.g. trying to post `_base` into TokenReg ABI resets all other values already entered

As reported by @tomusdrw

ngotchac:
Fixes #4553 

Also fixes some issues with the logger, and removes default logging of `certification ... not found`


jacogr:
Closes #4553 & Closes #4548

PR #4559, opened 2017-02-15
Explicitly set seconds to 0 from selector

jacogr:
Be explicit rather than assuming that timeselector zeros seconds. Sets -

- ms to 0 (ignored, not passed through to Parity)
- sec to 0 (be explicit in expectation)

If we do change the time selector component to be granular, the seconds setting can be removed.

### DIFF
--- a/js/src/ui/GasPriceEditor/store.js
+++ b/js/src/ui/GasPriceEditor/store.js
@@ -85,7 +85,12 @@ export default class GasPriceEditor {
           break;
 
         case CONDITIONS.TIME:
-          this.condition = Object.assign({}, this.condition, { time: new Date() });
+          const time = new Date();
+
+          time.setMilliseconds(0); // ignored by/not passed to Parity
+          time.setSeconds(0); // current time selector doesn't allow seconds
+
+          this.condition = Object.assign({}, this.condition, { time });
           break;
 
         case CONDITIONS.NONE:
--- a/js/src/ui/GasPriceEditor/store.js
+++ b/js/src/ui/GasPriceEditor/store.js
@@ -103,7 +103,9 @@ export default class GasPriceEditor {
     });
   }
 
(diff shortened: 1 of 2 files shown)


js/src/ui/GasPriceEditor/store.spec.js
```diff
@@ -162,9 +162,11 @@ describe('ui/GasPriceEditor/Store', () => {
     });
 
     describe('setConditionDateTime', () => {
-      it('sets the datatime', () => {
-        store.setConditionDateTime('testingDateTime');
-        expect(store.condition.time).to.equal('testingDateTime');
+      it('sets the datetime', () => {
+        const testTime = new Date('1973-06-11 07:52');
```
ngotchac: Can we test that it actually sets the seconds to 0 ?

js/src/ui/GasPriceEditor/store.js
```diff
@@ -104,6 +104,9 @@ export default class GasPriceEditor {
   }
 
   @action setConditionDateTime = (time) => {
+    time.setMilliseconds(0); // ignored by/not passed to Parity
```
derhuerst: Potentially dangerous as it mutates `time` instead of cloning it using `new Date(time)`.

js/src/ui/GasPriceEditor/store.spec.js
derhuerst: What about this?

```js
const testTime = new Date('1973-06-11 07:52:17');
```

jacogr: Sure, can test additionally.

js/src/ui/GasPriceEditor/store.js
jacogr: Verified that it doesn't affect the components used with actual testing prior to pushing. So not going to "fix" it.

Issue #1966, opened 2016-08-19
TransactionQueue improvements


gavofyork:
- [x] Keep transactions in persistent storage
- [ ] Future becomes a fixed-size rotating pool
- [ ] Current retains present priority mechanism
- [x] RPC for rebroadcasting (a) transaction(s)
- [x] RPC for inspecting queue contents [txpool API](https://github.com/ethereum/go-ethereum/wiki/Management-APIs#txpool) 
- [x] RPC for getting effective next nonce, taking into account queue

tomusdrw:
My thoughts:
```rust

// List of all transactions
transactions: HashMap<Address, BTreeMap<U256, Rc<VerifiedTransaction>>>

// Transaction and it's location.
// We store copies of handles to easily find transactions in queues (for removal / re-insertion)
LocalizedTransaction {
	transaction: Rc<VerifiedTransaction>,
	location: Current(TransactionHandle)/Future(FutureTransactionHandle)
}

// Current (with limit)
// Ordered by:
// 1. Sender priority (local transactions / reimported transactions / penalized transactions)
// 2. Nonce Height (nonce - state_nonce)
// 3. Gas Price
// 4. Hash
current: PriorityQueue<TransactionHandle>

TransactionHandle {
	transaction: Rc<VerifiedTransaction>,
	nonce_height: U256,
	priority: Local/RetractedBlock/External/GasLimitReached,
}

// Future (with limit)
// Ordered by:
// 1. Sender priority
// - stalled senders (occupies future, but transactions were not moved to Current for a long time) should be penalized
// 2. Nonce Height
// 3. Gas Price
// 4. Hash
future: PriorityQueue<FutureTransactionHandle>

FutureTransactionHandle {
	transaction: Rc<VerifiedTransaction>,
	nonce_height: U256,
	priority: u64 //(block number at insertion time; higher is better)
}
/*
Notes:
1. Priorities (in C & F) always have to be updated for all transactions from particular sender.
2. Each import/removal needs to update `nonce_heights` (perhaps could be optimized if it doesn't change).
*/
```

hermanjunge:
+1. (Following closely the issue). Need an implementer?

gavofyork:
help always welcome :)

rphmeier:
[...]

PR #4566, opened 2017-02-15
Save pending local transactions in the database

### LINKED ISSUE #2336: Initialize database outside of Client
Client shouldn't have the responsibility of initializing a database. This should be done in the `Service` at the initialization step.

rphmeier:
Now you can schedule transactions, restart your node, and still have them propagate correctly!

Adds a "node info" column family which will be used for general local node information (in the future, sync security level and anything else we might come up with). Introduces a migration for this purpose, solidifying the implicit incompatibility between 1.5 and 1.6-era DBs which already exists due to changes in receipt format.

All non-mined (pending and future) local transactions stored, along with their conditions and restored upon startup.

Also abstracts over the `Database` using a `KeyValueDB` trait implemented for an in-memory store as well, so we no longer require I/O for most DB tests.

Part of #1966 
Fixes #2336


parity/run.rs
```diff
@@ -314,6 +329,31 @@ pub fn execute(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>) -> R
 	let client = service.client();
 	let snapshot_service = service.snapshot_service();
 
+	// initialize the local node information store.
+	// TODO: tick it every once in a while?
```
rphmeier: in the case of a panic or other dirty exit where destructors aren't run, the list of transactions in the database won't be updated. we might want to tick periodically (once every 5 minutes?) to ensure a higher chance of them making it in.

local-store/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+		}
+		{
+			// pending removed, will write nothing.
+			let store = super::create(db.clone(), None, Dummy(vec![]));
+			assert_eq!(store.pending_transactions().unwrap(), vec![])
+		}
+	}
+
+	#[test]
+	#[should_panic]
```
tomusdrw: I thought that it shouldn't panic?

ethcore/src/migrations/v11.rs
```diff
@@ -0,0 +1,46 @@
...
+use util::kvdb::Database;
+use util::migration::{Batch, Config, Error, Migration, Progress};
+use std::sync::Arc;
+
+/// Copies over data for all existing columns.
+#[derive(Default)]
+pub struct ToV11(Progress);
+
+
+impl Migration for ToV11 {
```
tomusdrw: Is this migration necessary? I thought that you an just append new columns without copying the whole database.

> On-the-fly adding new Column Families and dropping them. Both operations are reasonably fast.
https://github.com/facebook/rocksdb/wiki/Column-Families#introduction

local-store/src/lib.rs
rphmeier: It'll panic because of the assertion that the loaded set is equal to the stored. They aren't equal because loading skips all transactions with an invalid signature.

ethcore/src/migrations/v11.rs
rphmeier: Right, the existing behavior is an artifact of our migration system, which was written before we had column families. We can probably add an in-place upgrade mode,  but we'd still have to create a backup copy.

Not sure if that's within the scope of the PR.

local-store/src/lib.rs
tomusdrw: So maybe it would be better to have `assert(expected != actual, "They aren't equal because loading skips all transactions with an invalid signature.")` instead? Should panic means that it can actually panic anywhere - imho it's suitable only for simple function calls not tests like this.

ethcore/src/migrations/v11.rs
tomusdrw: But you can just skip the migration at all right? (just alter `v10` to return more columns and that's it) Since if I understand correctly opening a database specifying more columns should work fine.

rphmeier: AFAIK you have to open the database with `N` columns and then add the `N + 1`th. Our migration manager doesn't support this operation currently. It should be simple enough to add, though.

rphmeier:
Now updates the store once every 15 minutes.

ethcore/src/migrations/v11.rs
rphmeier: @tomusdrw Separate issue filed for that: #4581

ethcore/src/db.rs
```diff
@@ -212,7 +214,7 @@ impl Writable for DBTransaction {
 	}
 }
 
-impl Readable for Database {
```
gavofyork: `KeyValueDB` and `KVDB`? really no better naming than this?

rphmeier: ideally we'd have the trait called `kvdb::Database`, but our rocksdb wrapper has already taken that name and I didn't want to break all existing usages of it (although the trait name `KeyValueDB` is arguably more descriptive). I could use a single-letter name for the generic param, but I don't think it really matters.

PR #4567, opened 2017-02-16
Secret store - initial version

svyatonik:
As proposed by @gavofyork :
```
first milestone is a RESTful API for getting keys encrypted with the requesting identity's public key
i.e. something like localhost:8082/<content_hash>/<request_signature> -> <encrypted_key>
```
Unfortunately, you'll be unable to test it, as there's no currently way to fill database with some contents. But if you fill it, it will work like that:
```
curl http://localhost:8082/0000000000000000000000000000000000000000000000000000000000000001/a199fb39e11eefb61c78a4074a53c0d4424600a3e74aad4fb9d93a26c30d067e1d4d29936de0c73f19827394a1dd049480a0d581aee7ae7546968da7d3d1c2fd01
8add166961bba5e20be6580109c1b59e1179b13ebc310ad024fffcc70d1fb379e230b33e7e842c3e2d12e4ad3ff9c866c471b054f8a0c69604c04de482a76f66e3254a13
```

PR outline:
1) added separate directory/package at the root of the project. I could move it somewhere else,  but it is  kinda similar to stratum (it also listens for some incoming requests) && stratum is also there. So maybe it is a good choice.
2) it was suggested that key server should be an ipc module. @NikVolf has told me that IPCs are turned off in current master (after I spend some time trying to beat compilation errors :) ). So now it will work in the same process. But there is some IPC stuff left, that I have already made. Idk - if it is better to remove or left it now.
3) I've added new compilation feature to parity (sstore), as I suppose most of parity users do not want to run key server. I do not  know if it is a good idea ( @arkpar has told some time ago that you guys had some troubles with features). So I can remove it if you think it must be removed. For now, to build/test with sstore support:
```
cargo build --features sstore
cargo test --features store -p ethcore-secstore
```
4) finally - the 'core' key server implementation is just a hyper-based http server, listening for GET requests. Then recovering public key from passed signature, checking ACLs (now it has some dummy implementation, as it was not a part of first milestone), then searching for the key in sstore-database (I use util::Database to store mapping: content_hash -> encryption key), encrypting this key with recovered public && returning this encrypted key as text/plain.


gavofyork:
as a minor point on naming, best to stick to one of "sstore", "secstore" and "secretstore" throughout the codebase. i like "secretstore".

different IPC module, even if it's in the same process for now, is nice because it forces clear interfaces.

svyatonik:
@gavofyork thanks, will update PR

Cargo.lock
```diff
@@ -619,6 +620,23 @@ dependencies = [
...
+version = "1.0.0"
+dependencies = [
+ "ethcore-devtools 1.6.0",
+ "ethcore-ipc 1.6.0",
+ "ethcore-ipc-codegen 1.6.0",
+ "ethcore-ipc-nano 1.6.0",
+ "ethcore-util 1.6.0",
+ "ethcrypto 0.1.0",
+ "ethkey 0.2.0",
+ "hyper 0.9.14 (registry+https://github.com/rust-lang/crates.io-index)",
```
tomusdrw: hyper is a bit heavy and slow, have you considered our async version or for instance `tokio-minihttp`?

Cargo.toml
```diff
@@ -81,6 +82,7 @@ evm-debug = ["ethcore/evm-debug"]
 evm-debug-tests = ["ethcore/evm-debug-tests"]
 slow-blocks = ["ethcore/slow-blocks"]
 final = ["ethcore-util/final"]
+sstore = ["ethcore-secstore"]
```
tomusdrw: I think it would be best to enable it by default (or build it on the CI), otherwise it's hard to detect any possible breaking changes causing compilation errors in that crate, since people don't alter the features and builds are not including this as well.

secret_store/src/http_listener.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl<T> KeyServer for KeyServerHttpListener<T> where T: KeyServer + 'static {
+	fn document_key(&self, signature: &RequestSignature, document: &DocumentAddress) -> Result<DocumentEncryptedKey, Error> {
+		self.handler.key_server.document_key(signature, document)
+	}
+}
+
+impl<T> HttpHandler for KeyServerHttpHandler<T> where T: KeyServer + 'static {
+	fn handle(&self, req: HttpRequest, mut res: HttpResponse) {
+		if req.method != HttpMethod::Get {
```
tomusdrw: Might be good to limit the requests depending on the planned usage:
1. If we're not intending to use this server from browser context we can deny requests with `Origin` header.
2. If we do intend to use it from the browser we need to add CORS support, cause currently the website will be able to perform a request, but won't be able to read the response.

Cargo.lock
svyatonik: Not actually :( By 'our async version' do you mean https://github.com/ethcore/hyper ? I'll take a look into it. tokio-minihttp sounds very good, as it has 'mini' in its name :) I'll check it out - thanks

Cargo.toml
svyatonik: Maybe it will be better to enable it later? When this functionality will be finished && polished. Now it has only basic features && it is not usable at all. I don't mind fixing any compilation issues by myself until it is ready for production, if you're worried about it.

secret_store/src/http_listener.rs
svyatonik: I also do not yet fully understand final shape of this functionality - who will be responsible for IPFS/Swarm integration. Is it key server, or some other component. If that's a key server, then most probably it will have its own UI. Or maybe it is user himself && it is his duty to save document to IPFS and then upload key here (that's a strange decision, but it can be so). I think @gavofyork could help us here. Or I can just add TODO about your comment here && update it later.

parity/sstore.rs
```diff
@@ -0,0 +1,106 @@
...
+	// the only dependency will be BlockChainClient
+}
+
+#[cfg(not(feature = "sstore"))]
+mod server {
+	use super::{Configuration, Dependencies};
+
+	/// Noop key server implementation
+	pub struct KeyServer {
+	}
```
rphmeier: braced empty struct?

parity/sstore.rs
```diff
@@ -0,0 +1,106 @@
...
+
+	impl KeyServer {
+		/// Create new noop key server
+		pub fn new(_conf: Configuration, _deps: Dependencies) -> Result<Self, String> {
+			Ok(KeyServer {})
```
rphmeier: `Ok(KeyServer)`

secret_store/src/http_listener.rs
rphmeier: @svyatonik Probably we could have a trait "DocumentStore" with a single function "fn fetch(H256) -> BoxFuture<Option<Document>>`, and this could be trivially implemented for any swarm/ipfs/disk/memory store. This server should be concerned solely with the logic above that.

as an aside, couldn't this be a separate executable?

svyatonik: @rphmeier Yes - we can go that way, but I'm not sure that this would be a part of this particular server/service :) Maybe user/external app will access this keyserver just for storing/retrieving keys && will upload/download files to this `DocumentStore` itself. Another option that this keyserver would be extended to something bigger && will have this functionality itself - then yes, I'll add something like `DocumentStore` here, thanks :)

It will need access to the blockchain data, so it can't be **totally** separate application. But it is intended to be separate executable in future (after IPC will be enabled) - yes.

tomusdrw: You could actually use `rust-web3` and existing IPC transport :)

svyatonik: Thanks, I like this idea :) So @gavofyork maybe create a completely separate executable for key-server and talk to Parity using RPC calls (it is enough for calling contract methods)? I think that's a great idea

rphmeier: @svyatonik Yes, the `eth_call` RPC is suitable for calling contracts. `rust-web3` also has `ethabi` integration for easy usage of contracts, abstracting over any direct `eth_call` nastiness.

Cargo.lock
svyatonik: agreed to use `tokio-minihttp` with @tomusdrw but left it for another PR, as now it doesn't have a way to be closed programatically. so first fix `tokio-minihttp`, then switch to it

secret_store/src/http_listener.rs
svyatonik: Left as is, until @gavofyork approves

parity/sstore.rs
svyatonik: done

parity/sstore.rs
svyatonik: done

svyatonik:
tried to make `KeyServer` trait IPC-ready + renamed sstore + secstore to secretstore everywhere

secret_store/src/http_listener.rs
gavofyork: IPC transport layer was effectively meant to be a very efficient RPC interface, supporting efficient underlying transport paradigms.

this would allow adding new, potentially non-ethereum, functionality without compromising the integrity of the existing codebase. whether the functionality is added as a separately spawned instance of `parity` (as with sync and stratum) or a completely separate executable, isn't so important.

unless there's a particularly good reason to go with the eminently inefficient JSONRPC, i would prefer to stick to using (and improving) our trait-based IPC module architecture.

i would lightly favour the same executable for now since it will make system deployment easier; that said, if you're desperate to have it in a different executable (rather than just a different crate/module) i'm not going to stand in the way.

svyatonik: ok, I don't feel myself confident enough to make such decisions. So let it be an IPC module (in future).

arkpar:
For this one I'd also suggest disabling it by default

gavofyork:
@svyatonik will leave the disabled-by-default for a separate PR...

svyatonik:
@gavofyork The feature is disabled by default from the beginning.

PR #4577, opened 2017-02-16
Deprecate eth_compile* RPCs

rphmeier:
See https://github.com/ethereum/EIPs/issues/209

We don't use this in the UI AFAIK, the contract editor just uses browser-solidity, right @ngotchac?

### DIFF
--- a/rpc/src/v1/impls/eth.rs
+++ b/rpc/src/v1/impls/eth.rs
@@ -16,8 +16,6 @@
 
 //! Eth rpc implementation.
 
-use std::io::{Write};
-use std::process::{Command, Stdio};
 use std::thread;
 use std::time::{Instant, Duration};
 use std::sync::{Arc, Weak};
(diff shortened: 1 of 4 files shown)


rpc/src/v1/impls/eth.rs
```diff
@@ -642,37 +629,14 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> Eth for EthClient<C, SN, S, M, EM> where
 	}
 
 	fn compile_lll(&self, _: String) -> Result<Bytes, Error> {
-		rpc_unimplemented!()
+		Err(errors::deprecated("Compilation of LLL via RPC is deprecated".to_string()))
 	}
 
 	fn compile_serpent(&self, _: String) -> Result<Bytes, Error> {
```
maciejhirsz: When deprecating RPC methods, marking them as `@deprecated` in the trait definition will help us keep the wiki in sync, since the markdown builder checks for those. [Example](https://github.com/ethcore/parity/blob/ed09a76c918077e7ef9e5bdfe557fff200701eef/rpc/src/v1/traits/parity_set.rs#L77).

rphmeier: Thanks, wasn't aware of this feature.

gavofyork:
looks like they've been more than deprecated - they've been removed...

ngotchac:
Yep, these RPC are not used by the UI

rphmeier:
@gavofyork 

Yes, I think it makes sense to remove them. Only solidity compilation was ever supported, and then only if `solc` was installed on the system and in the search path.

PR #4580, opened 2017-02-16
Proper default accounts RPCs

tomusdrw:
Proper APIs for default account.

Currently we have two layers, global:
`setNewDappsAddresses / setNewDappsDefaultAccount`

and then dapp-specific methods:
`setDappAddresses / setDappDefaultAccount`

You can specify dapp-specific default account, but still rely on global whitelist for list of accounts.

Default account is always returned as the first one on the list of accounts.
NOTE: setting a list of accounts will also override default account.

CC @jacogr


jacogr:
JS looks good. (Also started building off it for enhanced defaultAccount functionality, awaiting merge.)

Issue #4443, opened 2017-02-06
Account store is reloading accounts to often


tomusdrw:
Each time we try to access accounts they are reloaded from disk.
https://github.com/ethcore/parity/blob/c2ceefbc22915feb77ea012cdbb057bea1ea863c/ethstore/src/ethstore.rs#L255

Should not happen that often. May explain gitter reported slowness when a lot of accounts is managed by Parity.

PR #4583, opened 2017-02-16
Optimize key directory reloads

NikVolf:
Only reloads if something changed in the directory (including from outside of the app)

addresses #4443

### DIFF
--- a/ethstore/src/dir/mod.rs
+++ b/ethstore/src/dir/mod.rs
@@ -16,7 +16,6 @@
 
 use std::path::{PathBuf};
 use {SafeAccount, Error};
-use util::{H256, FixedHash};
 
 mod disk;
 mod geth;
@@ -64,7 +63,7 @@ pub trait KeyDirectory: Send + Sync {
 	/// Return vault provider, if available
 	fn as_vault_provider(&self) -> Option<&VaultKeyDirectoryProvider> { None }
 	/// Returns hash of the directory content, if supported
-	fn hash(&self) -> Result<H256, Error> { Ok(H256::zero()) }
+	fn hash(&self) -> Result<u64, Error> { Ok(0u64) }
 }
 
 /// Vaults provider
--- a/ethstore/src/dir/mod.rs
+++ b/ethstore/src/dir/mod.rs
@@ -62,8 +62,8 @@ pub trait KeyDirectory: Send + Sync {
 	fn path(&self) -> Option<&PathBuf> { None }
 	/// Return vault provider, if available
 	fn as_vault_provider(&self) -> Option<&VaultKeyDirectoryProvider> { None }
-	/// Returns hash of the directory content, if supported
-	fn hash(&self) -> Result<u64, Error> { Ok(0u64) }
+	/// Unique representation of directory account collection
+	fn unique_repr(&self) -> Result<u64, Error>;
 }
 
 /// Vaults provider
--- a/ethstore/src/dir/mod.rs
+++ b/ethstore/src/dir/mod.rs
@@ -16,6 +16,7 @@
 
 use std::path::{PathBuf};
 use {SafeAccount, Error};
+use util::{H256, FixedHash};
 
 mod disk;
 mod geth;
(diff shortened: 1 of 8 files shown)


util/src/sha3.rs
```diff
@@ -68,6 +68,24 @@ impl<T> Hashable for T where T: AsRef<[u8]> {
 	}
 }
 
+impl<T> Hashable for [T] where T: Hashable {
+	fn sha3(&self) -> H256 {
```
rphmeier: This implementation seems strange: intuitively I'm not sure what it would mean to call
`vec!["a, "b", "c"].sha3()`, but I don't think I would expect it to be 

`sha3("a") ^ sha3("b") ^ sha3("c") ^ sha3("")`.

Why not just replace any invocation of this with
`things_to_hash.into_iter().map(|x| x.sha3()).fold(Default::default(), BitXor::bitxor)`?

ethstore/src/dir/mod.rs
```diff
@@ -62,6 +63,8 @@ pub trait KeyDirectory: Send + Sync {
 	fn path(&self) -> Option<&PathBuf> { None }
 	/// Return vault provider, if available
 	fn as_vault_provider(&self) -> Option<&VaultKeyDirectoryProvider> { None }
+	/// Returns hash of the directory content, if supported
+	fn hash(&self) -> Result<H256, Error> { Ok(H256::zero()) }
```
rphmeier: this default implementation would lead to accounts never being reloaded?

NikVolf: key directory does not operate the concept of "reloading accounts" at all
it just returns hash of all account headers, if it able to

ethstore/src/ethstore.rs
```diff
@@ -244,11 +246,23 @@ impl EthMultiStore {
...
 	}
 
+	fn reload_if_changed(&self) -> Result<(), Error> {
+		let mut last_dir_hash = self.dir_hash.write();
+		let dir_hash = Some(self.dir.hash()?);
+		if *last_dir_hash == dir_hash {
+			return Ok(())
+		}
+		self.reload_accounts()?;
+		*last_dir_hash = dir_hash;
```
rphmeier: This could be set before reloading to prevent parallel reloads.

util/src/sha3.rs
NikVolf: Hmm.. actually, I think i won't use sha3 at all

ethstore/src/dir/mod.rs
rphmeier: Even so, there should not be a valid implementation of a key directory which causes `reload_if_changed` to think the data has never changed.

If we broaden the description of the function to something like "Returns a unique 32-byte representation of the directory's contents.", then I don't imagine any implementation which couldn't do it.

(in that case, a more descriptive name like `unique_repr(&self) -> Result<H256, Error>` might make sense)

ethstore/src/ethstore.rs
NikVolf: What? Set and release a lock before actual reloading?

ethstore/src/dir/mod.rs
NikVolf: Yeah, seems legit

ethstore/src/ethstore.rs
rphmeier: hmm, yeah, disregard that :)

NikVolf:
@rphmeier updated using simple std hasher & addressed another issue

util/src/sha3.rs
NikVolf: and btw, just fyi, xoring is the preferred way to combine multiple hashes (that is proven to be as good as the individual hashes are themselves), so that's quite natural to expect this implementation when hashing list of hashes

ethstore/src/ethstore.rs
```diff
@@ -244,11 +245,23 @@ impl EthMultiStore {
...
 			iterations: iterations,
 			cache: Default::default(),
+			dir_hash: Default::default(),
 		};
 		store.reload_accounts()?;
 		Ok(store)
 	}
 
+	fn reload_if_changed(&self) -> Result<(), Error> {
+		let mut last_dir_hash = self.dir_hash.lock();
```
svyatonik: this will work for root dir only (not for vaults), but maybe that's a good idea for next PR :)

NikVolf: @svyatonik 
yeah but i guess users are not supposed to mess with the vault directories themselves

svyatonik: they can use ethstore cli utility to modify vaults

PR #4588, opened 2017-02-17
Display ETH balance in overlay account selector

jacogr:
- Closes https://github.com/ethcore/parity/issues/4524
- Add balance display to AccountCard for defaultAccount selectors
- Don't wrap account header (if name is longer ...)

parity 2017-02-17 16-18-58

parity 2017-02-17 16-31-37

parity 2017-02-21 10-56-09


jacogr:
Good catch, happens when we have multiple branches going. Will add for consistency.

PR #4590, opened 2017-02-17
Revert Double Click on Accounts to close in Signer Bar

ngotchac:


jacogr:
Not quite revert, rather "close on click". Feels good though.

jacogr:
Ummm, it used to be click to select without closing. (Which is why the original was logged, somewhere either the close didn't work or was added.) Either way, this way around does tick the boxes.

PR #4594, opened 2017-02-17
Remainder of RPC APIs implemented for the light client

### LINKED ISSUE #4435: Light client data cache
`OnDemand` and `rpc` should each store a handle.

rphmeier:
Closes #4435 

Some are just stubs (like traces), lots of mining/sealing-related functionality not available yet.
Data fetched from the network is now cached. We fetch block bodies and compute a gas price corpus to estimate gas price. This corpus can be reused until it "expires" after a set amount of time, which will probably be 6 or 12 hours.


gavofyork:
lgtm, but only skimmed it.

PR #4608, opened 2017-02-20
More CLI settings for IPFS API

maciejhirsz:
- Allow to set different address binding via `--ipfs-api-interface`.
- Allow to set a list of Host headers to filter via `--ipfs-api-hosts`.
- Allow to set CORS via `--ipfs-api-cors`.
- Requests will be killed if invalid `Host` or `Origin` header is sent to avoid side effects.
- DRY-ed up and removed some unnecessary cloning in `parity/configuration.rs`.


ipfs/src/lib.rs
```diff
@@ -82,25 +166,26 @@ impl Handler<HttpStream> for IpfsHandler {
...
 
 				res.headers_mut().set(ContentLength(reason.len() as u64));
 				res.headers_mut().set(ContentType(mime!(Text/Plain)));
-
-				Next::write()
 			}
 		}
+
+		if let Some(cors_header) = cors::get_cors_header(&self.cors_domains, &self.origin) {
+			res.headers_mut().set(cors_header);
```
maciejhirsz: `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers` are only necessary in preflight request, currently non-GET request get killed outright so there is no need for them. I've added `Vary`.

ipfs/src/lib.rs
```diff
@@ -23,31 +23,113 @@ extern crate cid;
...
+		};
+
+		let cors_domains = match self.cors_domains {
+			Some(ref domains) => domains,
+			None => return false,
+		};
+
+		cors_domains.iter().any(|domain| match *domain {
+			AccessControlAllowOrigin::Value(ref allowed) => origin == allowed,
+			AccessControlAllowOrigin::Any => true,
```
rphmeier: @maciejhirsz ?

maciejhirsz: This is already fixed, you just can't see it on the diff here (it's one line below).

PR #4621, opened 2017-02-21
Align list displays with SectionList (UI consistency)

jacogr:
monosnap 2017-02-21 12-58-30

monosnap 2017-02-21 12-56-48

monosnap 2017-02-21 12-57-18


derhuerst:
Not sure I agree with the UX. But that aside, it lags significantly on my laptop.

MacBook Pro (Retina, 13-inch, Early 2015)
Chromium 56.0.2890.0 (64-bit)

jacogr:
Transitions. Great feedback, will remove those.

derhuerst:
Also, consider that there won't necessarily be mouse support (hover) on all clients. It's uncommon among devs, but I can imagine people that want to use the UI on their Surface with touch.

js/src/ui/Container/container.css
```diff
@@ -48,7 +51,8 @@ $transitionAll: all 0.75s cubic-bezier(0.23, 1, 0.32, 1);
 
     .hoverOverlay {
       background: $backgroundHover;
-      transform: scale(1, 1);
+      /*transform: scale(1, 1);*/
```
ngotchac: Shouldn't they be removed ? (The commented out properties)

js/src/ui/DappCard/dappCard.js
```diff
@@ -65,34 +71,29 @@ export default class DappCard extends Component {
...
+  getLink (app) {
+    const { showLink } = this.props;
+
+    if (!showLink) {
+      return null;
+    }
+
+    return app.url === 'web'
+      ? '/web'
+      : `/app/${app.id}`;
```
ngotchac: Can't we have this piece of logic somewhere else where it could be accessible from every component? I can imagine that it could be useful to have an easy access to this function.

js/src/ui/SectionList/sectionList.css
```diff
@@ -38,13 +43,14 @@
...
       display: flex;
-      flex: 0 1 33.33%;
-      max-width: 33.33%;
+      flex: 0 1 $widthNormal;
+      max-width: $widthNormal;
       opacity: 0.85;
       padding: 0.25em;
-      transition: all 0.75s cubic-bezier(0.23, 1, 0.32, 1);
+      /* https://www.binarymoon.co.uk/2014/02/fixing-css-transitions-in-google-chrome/ */
+      transform: translateZ(0);
```
ngotchac: :+1:

js/src/views/Accounts/Summary/summary.js
```diff
@@ -108,52 +110,71 @@ export default class Summary extends Component {
...
+            { this.renderOwners() }
+            { this.renderCertifications() }
+          </div>
+        }
+        link={ this.getLink() }
+      >
+        <Tags
+          className={ styles.tags }
+          tags={ tags }
+          handleAddSearchToken={ handleAddSearchToken }
```
ngotchac: This doesn't seem to work (cannot click on Tags anymore to add them to the Search)

js/src/views/Accounts/Summary/summary.js
```diff
@@ -163,63 +184,74 @@ export default class Summary extends Component {
...
-              </div>
-              <ReactTooltip id={ `owner_${owner.address}` }>
-                <strong>{ owner.name } </strong><small> (owner)</small>
-              </ReactTooltip>
-            </div>
-          ))
+                <div
+                  data-tip
+                  data-for={ `owner_${owner.address}` }
+                  data-effect='solid'
```
ngotchac: What are these for again ?

js/src/views/Application/Container/container.js
```diff
@@ -33,7 +33,10 @@ export default class Container extends Component {
     const { children, onCloseFirstRun, showFirstRun, upgradeStore } = this.props;
 
     return (
-      <ParityBackground className={ styles.container }>
+      <ParityBackground
+        attachDocument
```
ngotchac: I'm not sure we want this (at least not for every single dapp). Some dapp devs might built their app thinking that the background would be white. There is actually one example of this on Ropsten, which is just some text, black. If the Parity Background is black/dark, the dapp won't render correctly.

js/src/ui/Container/container.css
jacogr: I left it in on purpose - want to revisit the actual animations, so need guidance as to what it was. (There are a couple)

js/src/ui/DappCard/dappCard.js
jacogr: Sure, logged as https://github.com/ethcore/parity/issues/4660

js/src/views/Application/Container/container.js
jacogr: Good point. Will make the iframe bg white.

js/src/views/Accounts/Summary/summary.js
jacogr: Ummm.... Not intended :)

js/src/views/Accounts/Summary/summary.js
jacogr: You added them :) It is actually quite nice, shows the name info on hover. (Have it on my todos to replication for Vaults)

js/src/views/Accounts/Summary/summary.js
jacogr: Fixed.

js/src/views/Application/Container/container.js
jacogr: Fixed. (iframe bg white)

js/src/views/Accounts/Summary/summary.js
ngotchac: Ah, right

ngotchac:
Looks good, only last point remaining:

> when hovering a line of Accounts (for example) where the number of item is less than the max (eg. 2), even if the mouse is not over an Account, there is an transformation/animation going on

jacogr:
Not going to fix that, it actually is part of the SectionList component - feel free to give it a stab, but it is maybe annoying, but really not anywhere near the critical path, so I'm not considering it. 

I have discussed this - Best work-around I could muster a couple of PRs ago - need to move the unselected sizes down to 29% when hovering the row, keeping it at 33.33% and makes at least Chrome not resize all equally - and then we end up with an issue since overflows start to occur.

PR #4631, opened 2017-02-22
Vault Management UI (round 2)

jacogr:
Extending the work from https://github.com/ethcore/parity/pull/4446, working towards completing https://github.com/ethcore/parity/issues/3501

Includes & requires (there should be merged first, relevant parts in here) -
  - ~~SelectionList component introduced in https://github.com/ethcore/parity/pull/4639 (used for selection of vaults)~~

Future work -
  - Enhancement to allow for Vault selection upon account creation, requires availability of https://github.com/ethcore/parity/pull/4637
  - Display of vault in summary hover info, requires availability of https://github.com/ethcore/parity/pull/4621

Edit vault -
  - Meta data edit (description & passwordHint)
  - Password edit

parity 2017-02-22 11-52-22

Account Vault info -
  - Show vault information when attached

parity 2017-02-23 09-38-22

Edit Meta -
  - Show current associated vault
  - Allow editing of vault (based on which are open)

parity 2017-02-23 11-38-17
parity 2017-02-23 11-38-44


js/src/modals/EditMeta/editMeta.js
```diff
@@ -39,6 +41,11 @@ class EditMeta extends Component {
   }
 
   store = new Store(this.context.api, this.props.account);
+  vaultStore = VaultStore.get(this.context.api);
+
+  componentWillMount () {
+    this.vaultStore.loadVaults();
```
ngotchac: Can't it be loaded only once (the Vault Store) when the app starts ?

js/src/modals/EditMeta/editMeta.js
```diff
@@ -154,22 +163,87 @@ class EditMeta extends Component {
...
+
+    if (!isAccount) {
+      return null;
+    }
+
+    return (
+      <InputAddress
+        allowCopy={ false }
+        allowInvalid
+        disabled
```
ngotchac: `readOnly`?

js/src/modals/VaultMeta/vaultMeta.js
```diff
@@ -0,0 +1,293 @@
...
+import PasswordStrength from '~/ui/Form/PasswordStrength';
+import { CheckIcon, CloseIcon } from '~/ui/Icons';
+
+import styles from '../VaultCreate/vaultCreate.css';
+
+@observer
+class VaultMeta extends Component {
+  static propTypes = {
+    newError: PropTypes.func.isRequired,
+    vaultStore: PropTypes.object.isRequired
```
ngotchac: Why is the `VaultStore` passed as a prop here, and not in the `EditMeta` component?

js/src/modals/VaultMeta/vaultMeta.js
```diff
@@ -0,0 +1,293 @@
...
+            disabled={ isBusyMeta }
+            icon={ <CloseIcon /> }
+            key='close'
+            label={
+              <FormattedMessage
+                id='vaults.editMeta.button.close'
+                defaultMessage='close'
+              />
+            }
+            onClick={ this.onClose }
```
ngotchac: I know this has been discussed earlier, but I think that at least for new modals using the Portal, we shouldn't add this useless `Close` button

js/src/modals/VaultSelector/vaultSelector.js
```diff
@@ -0,0 +1,99 @@
...
+
+import { Portal, SelectionList, VaultCard } from '~/ui';
+
+@observer
+export default class VaultSelector extends Component {
+  static propTypes = {
+    onClose: PropTypes.func.isRequired,
+    onSelect: PropTypes.func.isRequired,
+    selected: PropTypes.string,
+    vaultStore: PropTypes.object.isRequired
```
ngotchac: Same as above regarding the `vaultStore` as a prop

js/src/views/Account/Header/header.js
```diff
@@ -154,4 +155,25 @@ export default class Header extends Component {
...
+    const { account } = this.props;
+    const { meta } = account;
+
+    if (!meta || !meta.vault) {
+      return null;
+    }
+
+    return (
+      <div className={ styles.vault }>
+        <IdentityIcon
```
ngotchac: Matter of personal taste here, but I find the Identity Icon a bit big here

js/src/modals/VaultMeta/vaultMeta.js
jacogr: At this point it is done for the sake of consistency :(

js/src/modals/EditMeta/editMeta.js
jacogr: The reason it is done here is to ensure we have the current info. It is loaded already at this point, basically a refresh if/when it was changed outside of the window. (It _should_ be ok, but rather safe than sorry)

js/src/modals/VaultMeta/vaultMeta.js
jacogr: Basically not needed until it is needed, so don't want to pull it in in the parent.

Even where the info is displayed on the Account Header, this is still the case - all the required info is already available as part of meta to render the icon. 

In the case of VaultMeta it gets called from `~/views/Vaults` so already available, no need to create.

jacogr:
Fixed the password bug.

As for your question - would like to see both rename & delete.

js/src/modals/VaultMeta/vaultMeta.js
ngotchac: But isn't it a Singleton? The instance is saved in memory, so there's no need to re-create it, right? Just seems unbalanced to have the two approaches. No big deal though

ngotchac:
Have some errors in the console when modifying the password: 

```Warning: Failed prop type: Invalid prop `disabled` of type `object` supplied to `EnhancedButton`, expected `boolean`.```

ngotchac:
Last comments:

- Find the Identity Icon in the Account Summary a bit big
- Not sure about the Vault Store being passed as props sometimes, and sometimes not

PR #4632, opened 2017-02-22
Generic state backend

### LINKED ISSUE #4616: Minimal state backend abstraction
cc @arkpar 

With remote transaction execution for the light client, we have three possible state backends:
  - Block enactment (current implementation)
  - Transaction proving: never skip trie lookups
  - Execute on top of raw HashDB.

These cases can be unified under a `Backend` trait, which we assume has an API more or less similar to `StateDB` as it exists.

rphmeier:
Closes #4616.

This is intended to allow us to enact state changes on general databases/caches, rather than just the single `StateDB`.

Also opens the door to executing EVMs over different account storage models, by moving the trie lookups fully into the backend, which may be done in the future.

### DIFF
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -37,6 +37,8 @@ use util::trie::recorder::Recorder;
 mod account;
 mod substate;
 
+pub mod backend;
+
 pub use self::account::Account;
 pub use self::substate::Substate;
 
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -194,7 +194,7 @@ impl AccountEntry {
 /// checkpoint can be discateded with `discard_checkpoint`. All of the orignal
 /// backed-up values are moved into a parent checkpoint (if any).
 ///
-pub struct State<B> {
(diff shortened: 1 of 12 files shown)


ethcore/src/state/mod.rs
```diff
@@ -186,8 +194,8 @@ impl AccountEntry {
 /// checkpoint can be discateded with `discard_checkpoint`. All of the orignal
 /// backed-up values are moved into a parent checkpoint (if any).
 ///
-pub struct State {
-	db: StateDB,
+pub struct State<B> {
```
arkpar: why not `B: Backend` ?

rphmeier: More general > less general. This is a common pattern among e.g. structures in `std`: constrain the generic parameters only when introducing functionality that depends on those constraints.

arkpar: I'd agree if there was some generic functionality, but `State` is pretty much useless without a backend. Attempting to use it with unsupported type will leave you with a bare struct which does not really represent the `State` abstraction. I think it would be better to explicitly disallow that.

PR #4637, opened 2017-02-22
Enhance dialog layouts (round 1)

jacogr:
Ongoing effort to enhance dialogs with better layouts. (Raised in https://github.com/ethcore/parity/pull/4625). This does not cover all the modals and does not address every thing on each modal - it is part of an ongoing process. (Currently trying to go wide with first-touch before going deep.)

Includes & depends on (these to be merged first, relevant parts in here) -
- ~~`~/ui/SelectionList` component https://github.com/ethcore/parity/pull/4639~~
- ~~Geth account import fix https://github.com/ethcore/parity/pull/4641~~

CreateAccount/Type Selection -

parity 2017-02-22 22-09-14

CreateAccount/Geth import -

parity 2017-02-22 21-30-41
parity 2017-02-22 21-24-01

CreateAccount/Done -

parity 2017-02-23 16-57-26

AddAddress -

parity 2017-02-23 09-12-27


ngotchac:
Really like the new layout !

Some small issues though:

[ ] The imports are not working (JSON File and Presale Wallet)
[ ] The inputs are not cleaned when you change the Import type (eg. put a name in New Account, go back, click on JSON File : the same name is pre-filled) ; not really an issue but I find it misleading.

One thing that would be great (even though could be outside of this PR), is to use the same upload component that for the Import in Accounts (click or drop). I think it would look especially great in the Portal.

jacogr:
Point #1 - that is not good, I tested all of them. Will re-look.

Point #2 is a good catch - I actually fixed that in the vaults 3 PR. (Could backport that in here)

js/src/modals/AddAddress/addAddress.js
```diff
@@ -98,64 +98,74 @@ export default class AddAddress extends Component {
...
+              />
+            }
+            label={
+              <FormattedMessage
+                id='addAddress.input.address.label'
+                defaultMessage='network address'
+              />
+            }
+            onChange={ this.onEditAddress }
+            ref='inputAddress'
```
ngotchac: Why are there some `ref`s here?

jacogr: Sorry, didn't remove that one - it was a toy in dev. Will remove, unused.

jacogr: Nope thinking of another one. No idea, not used there or in the tests, so unsure.

jacogr:
Import fix, thanks @ngotchac - https://github.com/ethcore/parity/pull/4674

js/src/modals/AddAddress/addAddress.js
ngotchac: So can we remove them ?

jacogr: They have been. (Unless I didn't push?)

PR #4641, opened 2017-02-22
Fix Geth account import

jacogr:
Fixes https://github.com/ethcore/parity/issues/4279

### DIFF
--- a/js/src/modals/CreateAccount/store.js
+++ b/js/src/modals/CreateAccount/store.js
@@ -32,6 +32,7 @@ export default class Store {
   @observable description = '';
   @observable gethAccountsAvailable = [];
   @observable gethAddresses = [];
+  @observable gethImported = [];
   @observable isBusy = false;
   @observable isWindowsPhrase = false;
   @observable name = '';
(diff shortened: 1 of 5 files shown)


js/src/modals/CreateAccount/store.js
```diff
@@ -234,20 +239,23 @@ export default class Store {
...
-        return Promise.all(this.gethAddresses.map((address) => {
-          return this._api.parity.setAccountName(address, 'Geth Import');
-        }));
-      })
-      .then(() => {
-        return Promise.all(this.gethAddresses.map((address) => {
-          return this._api.parity.setAccountMeta(address, {
-            timestamp
+      .then((gethImported) => {
+        console.log('createAccountFromGeth', gethImported);
```
derhuerst: could be `console.info` to reduce noise.

jacogr: On purpose - very specific "noise" in this case for an operation that happens once-off. So going to "ignore" the comment, but understand the concern very much for anything that is not basically a once-off :)

derhuerst: `console.info` helps devs filter out logs irrelevant to their current problem.

PR #4649, opened 2017-02-23
nl translations

h3ll0fr13nd:
Tested on linux


jacogr:
Sorry for the delay, just want to do a run through with a pull-down. (I glanced over it, looks very good.) We just need make sure everything is extracted (thumb-suck I think we are 80%+ done, maybe even more) and have some extra tools, i.e. something that can check for missing & non-existent messages.

Really do appreciate your efforts on this.

PR #4446, opened 2017-02-06
Vault Management UI (first round)

jacogr:
Integrates Vault management capabilities into the UI, closes #4452 (UI portion of #3501). Includes -

- `modals/VaultCreate` - create vaults with their metadata
- `modals/VaultAccounts` - assign/de-assign accounts to a vault
- `modals/VaulLock` - closes/locks a previously opened vault
- `modals/VaultUnlock` - opens/unlocks a closed vault
- `views/Vaults` accessible from the Accounts view
    - includes the store for managing vault information
- `ui/VaultCard` card to display a single vault
- `ui/{Container,Title,ConfirmDialog}` tie-in changes for display via Portal

Future work & enhancements before 1.6 -

- Change of password for Vaults
- Edit of metadata for Vaults (`meta.{description, passwordHint}`)
- Display of associated vault under Account properties & summary (`meta.vault`)
- Allow moving account to vault in account summary pages
- Selection of vault on account creation

In action -

https://youtu.be/JWc01XWPf9U

parity 2017-02-10 14-37-42

PR #4652, opened 2017-02-23
Vault Management UI (round 3)

jacogr:
Extending the work from #4446 & #4631, finally Closes https://github.com/ethcore/parity/issues/3501

Adds the following functionality -
  - Selection of vaults on default account creation screen
  - Display of vault info on list summaries

Future work -
- Currently the Vault attach on create is only available on the single account creation mechanisms 

CreateAccount vault selection -
parity 2017-02-23 16-47-46

Vault information on account summary -
parity 2017-02-23 17-06-35

Depends on the following to be merged -
  - ~~SelectionList component https://github.com/ethcore/parity/pull/4639~~
  - ~~ModalBox component https://github.com/ethcore/parity/pull/4637~~
  - ~~Hover info cards https://github.com/ethcore/parity/pull/4621~~
  - ~~Previous round of Vault UI changes https://github.com/ethcore/parity/pull/4631~~


jacogr:
Marked in-progress to resolve conflicts. (Should not be there, but alas, they are...)

ngotchac:
I'm not sure I really like the UI. The vault names are in gray, right under the tokens and certifications that are in white... Why not putting the vaults in a sort of *Tag* like the tokens and the certifications ? Also there is no margin between the tokens and the vault. (I am talking about the Accounts Summary card)

jacogr:
Could def. tag it like we have with Balances/Certifications, should actually look more consistent.

jacogr:
Tag-ified

monosnap 2017-02-28 16-28-38

js/src/modals/CreateAccount/NewImport/newImport.js
```diff
@@ -18,14 +18,16 @@ import { observer } from 'mobx-react';
 import React, { Component, PropTypes } from 'react';
 import { FormattedMessage } from 'react-intl';
 
-import { Form, FileSelect, Input } from '~/ui';
+import { ChangeVault, Form, FileSelect, Input } from '~/ui';
```
ngotchac: `ChangeVault` is not in `~/ui`

js/src/modals/CreateAccount/ChangeVault/changeVault.js
```diff
@@ -0,0 +1,51 @@
...
+
+import { observer } from 'mobx-react';
+import React, { Component, PropTypes } from 'react';
+
+import { VaultSelect } from '~/ui';
+
+@observer
+export default class ChangeVault extends Component {
+  static propTypes = {
+    store: PropTypes.object.isRequired,
```
ngotchac: It's a bit hard to know which store we're talking about here. Could use the same naming as `vaultStore`

js/src/ui/Form/VaultSelect/vaultSelect.js
```diff
@@ -0,0 +1,103 @@
...
+    onSelect: PropTypes.func.isRequired,
+    value: PropTypes.string.isRequired,
+    vaultStore: PropTypes.object
+  };
+
+  state = {
+    isOpen: false
+  };
+
+  vaultStore = this.props.vaultStore || VaultStore.get(this.context.api);
```
ngotchac: Wouldn't it be better to choose whether it should be passed as prop or not? I find it a bit confusing

js/src/modals/CreateAccount/store.js
```diff
@@ -220,7 +226,28 @@ export default class Store {
     this.stage--;
   }
 
-  createAccount = () => {
+  createAccount = (vaultStore) => {
```
ngotchac: Wouldn't it make sense to attach the `vaultStore` to this Store before, instead of passing this as an argument (might be used for other methods)?

ngotchac:
The Vaults are showing twice...

image

js/src/ui/Form/VaultSelect/vaultSelect.js
jacogr: Actually only done this way for testing.

js/src/modals/CreateAccount/store.js
jacogr: No. We specifically don't want it in FirstRun. So it is an option, not a requirement.

js/src/modals/CreateAccount/NewImport/newImport.js
jacogr: Ok, I mucked-up a merge with a move.

js/src/modals/CreateAccount/store.js
ngotchac: Could be optional without being an argument, no?

js/src/modals/CreateAccount/ChangeVault/changeVault.js
jacogr: Makes sense, should have renamed with the others.

jacogr:
Last one is another botched merge. :(

ngotchac:
Also, could we also use the tagged version of the Vault for the Account Summary in an Account page. It's not that great right now

js/src/modals/CreateAccount/ChangeVault/changeVault.js
jacogr: On looking at the code. Store is the CreateAccount/store - used like that in all components in CreateAccount. (Not renaming it here accross all)

jacogr:
Not doing that in here. I will as soon as we have a proper tag mechanism in-place, as it stands it is just more duplication which I'm not up for.

js/src/modals/CreateAccount/ChangeVault/changeVault.js
ngotchac: Well, it's obvious what store it is inside the `CreateAccount` context. It's less obvious in the `ChangeVault` context (you cannot just look at the imports)

js/src/ui/Form/VaultSelect/vaultSelect.js
ngotchac: So.... ?

js/src/modals/CreateAccount/ChangeVault/changeVault.js
jacogr: `~/ui/CreateAccount/ChangeVault`

js/src/ui/Form/VaultSelect/vaultSelect.js
jacogr: It only gets use by the tests here.

js/src/modals/CreateAccount/store.js
jacogr: I prefer passing it here when we create, only used at this point and set by the caller as an option.

js/src/modals/CreateAccount/ChangeVault/changeVault.js
ngotchac: But it could be used elsewhere : `views/Account/Header` is not only used for `Accounts` but also for `Address` for example. Don't want to make a big thing, but I think that renaming it doesn't cost much and adds some readability.

js/src/ui/Form/VaultSelect/vaultSelect.js
ngotchac: So as a prop ? Thus we don't need the `VaultStore.get(this.context.api)` ?

ngotchac:
Looks ok, still some code-style gumbles, not blocking though.

js/src/ui/Form/VaultSelect/vaultSelect.js
jacogr: The tests (e.g. vaultSelect.spec.js) passes thrhough as prop to tightly control the conditions. Normal operations is via getter.

ngotchac: Is there any other components that has this kind of logic ? I think that if we really want to have clean logic, we would have one _pure_ component that gets everything via its props and that get tested, and one wrapper component that passes to this _pure_ component the right props from whatever (a store instance, context, ...)

js/src/modals/CreateAccount/ChangeVault/changeVault.js
jacogr: No issues either way, but not going to do it just for the component and the others are out of scope for the PR. We should probably do it everywhere, we pass it everywhere as store. (Not just this component). Issues -

1. As you said, it may not be  obvious from reading what we are dealing with (fails a bit in "write code for reading")
2. It conflicts with Redux store naming, i.e. when a MobX store & Redux is used (couple of places), the MobX store cannot be called 'store' in props

Should be logged for a global refactor/cleanup - https://github.com/ethcore/parity/issues/4748

PR #4654, opened 2017-02-23
Fix Tendermint deadlock

keorn:


rphmeier:
Are there any race conditions which could occur during handling of the message now that the step isn't protected within that logic? Seems that now a writer could squeeze in between readers in `handle_valid_message` and `generate_message`.

rphmeier:
(this could be averted by threading the read lock through; unfortunately there still isn't a seamless `upgrade` which prevents races against other potential writers)

keorn:
`handle_valid_message` should run according to the latest state, "valid" is invariant to state

PR #4655, opened 2017-02-23
Propagate trie errors upwards from State

rphmeier:
Goes after #4632

Allows us to execute transactions on incomplete tries without issue.

### DIFF
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -573,7 +573,7 @@ impl<B: Backend> State<B> {
 				let addr_hash = account.address_hash(address);
 				{
 					let mut account_db = factories.accountdb.create(db.as_hashdb_mut(), addr_hash);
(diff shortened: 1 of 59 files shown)


rphmeier:
(To support non trie-based backends we might want an associated `Error` type on `state::Backend` and `Ext`.

ethcore/src/state/mod.rs
```diff
@@ -186,8 +195,8 @@ impl AccountEntry {
 /// checkpoint can be discateded with `discard_checkpoint`. All of the orignal
 /// backed-up values are moved into a parent checkpoint (if any).
 ///
-pub struct State {
-	db: StateDB,
+pub struct State<B> {
```
arkpar: why not add `B: Backend` here too?

rphmeier: That's the same commit as #4632 as this builds upon that, we can ice this until it's in.

PR #4657, opened 2017-02-23
Misc. small UI fixes

### LINKED ISSUE #4644: Accounts Bar could be more helpful on smal screens
- Vaults could have a label
- "New" could be changed to "Account" and "Wallet", I think plus icon and a name is descriptive enough to figure out that it means "New"

screenshot from 2017-02-22 20-19-52

jacogr:
- Shorten menubar entries to relevant words (Closes #4644)
  - Accounts, Addresses, Contracts
  - Account, Address, Contract
- Fields
  - Input - add onDefaultAction callback prop
  - AddressInput - add autoFocus prop
- Delete Account
  - Auto focus password field
  - Create on enter in field
- Open Vault
  - Auto focus on password field
  - Fix default action (not on tab, Enter only)
  - Add Form around inputs with extra padding
- Create Account
  - Normal account - Auto-focus name field
  - JSON imports - Auto-focus name field
  - Raw key - Auto-focus key field
  - Recovery phrase - Auto-focus field & clear phrase between use
- Create Wallet
  - MultiSig - put entry fields first, auto-focus name
  - Watch - Auto-focus name
- Add Address
  - Auto-focus on address field
- Add Contract
  - Auto-focus address field (all ABI types)
- Deploy contract
  - Auto-focus name field
  - Re-arrange fields to have focussable field first
- Edit Meta
  - Auto-focus name field
- Wallet Settings
  - Auto-focus modifications field
- Verification
  - Auto-focus first data field
  - Auto-focus verification number field

### DIFF
--- a/js/src/modals/AddAddress/addAddress.js
+++ b/js/src/modals/AddAddress/addAddress.js
@@ -65,7 +65,7 @@ export default class AddAddress extends Component {
   renderDialogActions () {
     const { hasError } = this.store;
 
-    return ([
+    return [
       <Button
         icon={ <CancelIcon /> }
         label={
@@ -89,7 +89,7 @@ export default class AddAddress extends Component {
         onClick={ this.onAdd }
         ref='addButton'
       />
-    ]);
+    ];
   }
 
   renderFields () {
@@ -99,6 +99,7 @@ export default class AddAddress extends Component {
       <Form>
         <InputAddress
           allowCopy={ false }
+          sutoFocus
           disabled={ !!this.props.address }
           error={ addressError }
           hint={
--- a/js/src/modals/AddAddress/addAddress.js
+++ b/js/src/modals/AddAddress/addAddress.js
@@ -68,6 +68,7 @@ export default class AddAddress extends Component {
     return [
       <Button
         icon={ <CancelIcon /> }
+        key='cancel'
         label={
           <FormattedMessage
             id='addAddress.button.close'
@@ -80,6 +81,7 @@ export default class AddAddress extends Component {
       <Button
         disabled={ hasError }
         icon={ <AddIcon /> }
+        key='save'
         label={
           <FormattedMessage
             id='addAddress.button.add'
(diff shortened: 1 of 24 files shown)


js/src/modals/AddAddress/addAddress.js
```diff
@@ -89,7 +89,7 @@ export default class AddAddress extends Component {
         onClick={ this.onAdd }
         ref='addButton'
       />
-    ]);
+    ];
```
ngotchac: The `Button`s miss the `key` prop. And while we're at it, we might want to get rid of every `Cancel` button for modals that use the Portal. There is already this big `X` button to cancel/close the modal/Portal.

js/src/views/Address/address.js
```diff
@@ -153,7 +153,7 @@ class Address extends Component {
       <Button
         key='delete'
         icon={ <ActionDelete /> }
-        label='forget address'
+        label='forget'
```
ngotchac: `FormattedMessage` ?

jacogr: I have an i18n branch going, starting my way up from the bottom of the views, will get to that in there. Not doing in here, minor tweaks before release, nothing more.

js/src/modals/AddAddress/addAddress.js
jacogr: Not trying to fix the all button layouts here. Wait for modal updates round 2, 3, etc. (Need to work through one-by-one to see what works and what doesn't - help is appreciated there, it is a lot of ground to cover to say the least.)

Key is valuable though.

jacogr: Ummm... I actually fixed this in my modals update branch (along with a ton of others, we didn't seem to be as vigilant as we should have), however re-doing it here.

jacogr:
Comments addressed, merging.

jacogr:
As for the i18n comments, see my comments on the NL branch that went in - we need some extra checking tools, duplicates both ways, summary reports, etc. (However... once again, need to try and see if there are already existing helpers, we are not the first people with having to keep these updated)

For now keeping it as a manual process to run these until everything is sorted out. (Once manual goes 100%, can make that output part of the normal dev builds.)

PR #4661, opened 2017-02-24
Max gas limit and min gas price

arkpar:


gavofyork:
might be sensible to force gas ceil target to be at most the max gas limit, otherwise a misconfigured miner could start creating blocks that it can't import.

PR #4663, opened 2017-02-24
Extract remaining i18n strings (views)

jacogr:


jacogr:
Will re-open when active/working on the branch again.

PR #4667, opened 2017-02-24
Slightly different background for default Account

### LINKED ISSUE #4666: SelectionList should highlight defaults better
Possibly try (despite different components with backgrounds) to bring back a slight background highlight.

ngotchac:
Fixes #4666


js/src/ui/SelectionList/selectionList.css
```diff
@@ -47,6 +47,14 @@
   filter: none;
 }
 
+.default {
+  background: rgba(255, 255, 255, 0.75);
+
+  &:hover {
+    background: initial;
```
derhuerst: afaict this is not a valid value. `background` -> `background-color` and `initial` -> `transparent` would be spec-compliant.

jacogr:
Like the idea, however now it looks very, very close to what we have for disabled - instead of popping up it pops down. (If that makes any sense) Maybe the border would work more sensibly?

ngotchac:
What about this :

image

jacogr:
Ok, will pull down and take a peek. (Along with unselected ones and the impacts on other modals where we don't have defaults, the shot shown does look like it moves in the right direction.)

jacogr:
Just looking at the Dapps, the previous looks much better (make no mistake, not perfect, but it seems like the previous bleed-through has some benefits) -

Old -
parity 2017-02-24 16-50-34

New -
parity 2017-02-24 16-48-38

Basically it entered the "it is too gray now" territory. (The same effect should be seen on the Accounts, but alas, they don't yet use the Container yet)

In the account selector, find it difficult to focus on what I'm looking at, still too close together now. It feels like the selected & defaults don't "fit" with the way things generally look.

parity 2017-02-24 16-52-29

ngotchac:
Okay, let's try with the border then:
image

jacogr:
The problem with the current implementation is that the "hover effects are not properly supported.

Pre (border basically disappears, seamless hover expand) -
parity 2017-02-24 20-00-03

This (border is being overwritten by the hover expand) -
parity 2017-02-24 20-01-14

ngotchac:
Closing this in favor of #4688

PR #4669, opened 2017-02-24
Re-Introducing HappyPack

ngotchac:
Happy Pack is now compatible with Webpack 2!

This PR updates dev dependencies (all of them or minor or patch updates).
It introduces HappyPack again (rebuilts are now taking 4 to 5s on my machine).

(Also added a Yarn lock file. This isn't use by anything, but could be if we want to)


jacogr:
Could we on-ice this until 1.6 is out? I'd prefer not pulling this in at the end of a release cycle.

Generally happy with the HappyPack being back and packing for us, not quite convinced on yarn. (Don't need to sell it to me, I know full well the benefits.) Just don't believe in pulling stuff in "just because of maybe".

Overall -

1. Put on-ice until 1.6 is out (marked as such)
2. Pull yarn lock until we are actually using it

ngotchac:
Fine with putting this on-ice. Just wondering what harm could the yarn lock file make though?

jacogr:
Nothing really. Apart from the fact that we are not using yarn consistently - as soon as I do the first `npm install ...` it will be outdated, for you using `yarn install ...` it will update.  

So no "harm" (i.e. breakage), rather let us be consistent instead of wishful. No rush, let's just do it properly and consistently.

ngotchac:
Okay, deleted the Yarn lock file, and actually added it to the gitignore until we want to use it (like this anyone can use `yarn install` without worrying about not comiting the lock file).

PR #4671, opened 2017-02-24
Custom dev chain presets

### LINKED ISSUE #4494: --chain=dev doesn't seal when transactions are sent in parallel
Possible workaround is to use `--reaseal-min-period`, but probably we should just check if there is anything in the queue after the block is mined to trigger sealing again.

keorn:
Closes #4494 
Seems to work as expected, all transactions get included.


rphmeier:
Seems to override user-specified CLI options? Defaults are good but this seems to restrict freedom.

keorn:
Can add a warning that `--chain dev` is a special preset. Otherwise per Engine defaults would be needed, which would require larger changes to the current setup.

rphmeier:
(logged as separate issue)

I think this might make sense to do as "preset" configs that could come bundled with Parity.

PR #4684, opened 2017-02-26
Remote transaction execution

rphmeier:
(For `eth_call` and `eth_estimateGas`)

Peers can request execution of a transaction from each other, returning a subset of the state proving the (failed or successful) execution, up to a certain amount of gas.

This is accomplished through two state backends: `Proving` and `ProofCheck`.
The `OnDemand` service also now supports transaction execution requests.


ethcore/light/src/net/mod.rs
```diff
@@ -1182,6 +1193,90 @@ impl LightProtocol {
 		Ok(())
 	}
 
+	// Receive a request for proof-of-execution.
+	fn get_transaction_proof(&self, peer: &PeerId, io: &IoContext, raw: UntrustedRlp) -> Result<(), Error> {
+		// refuse to execute more than this amount of gas at once.
+		// this is appx. the point at which the proof of execution would no longer fit in
+		// a single Devp2p packet.
+		const MAX_GAS: usize = 50_000_000;
```
NikVolf: make it pub const and document?

ethcore/light/src/net/mod.rs
```diff
@@ -1182,6 +1193,90 @@ impl LightProtocol {
...
+				return Ok(())
+			}
+		};
+		let mut peer = peer.lock();
+
+		let req_id: u64 = raw.val_at(0)?;
+
+		let req = {
+			let req_rlp = raw.at(1)?;
+			request::TransactionProof {
```
NikVolf: shouldn't the `rlp::Decodable` be implemented for `TransactionProof` instead of reading it just here?

NikVolf: nvm seems like a lot of this in the same module
still might be a question for future

ethcore/light/src/net/mod.rs
rphmeier: there are also similar constants within all functions which serve data, `MAX_HEADERS`, `MAX_RECEIPTS`, etc..

ethcore/light/src/net/mod.rs
rphmeier: I'm rewriting the requests module next to use PIP requests + batching. Cleaner usage of RLP will be part of that.

PR #4688, opened 2017-02-27
Less agressive grayscale/opacity in SelectionList

jacogr:
@ngotchac Pulls in the ideas from your fix as well. Could you check that 

(a) the consistency between selected/unselected is right
(b) the consistency between default/select is right

I'm thinking it probably still needs some small tweaks.

### DIFF
--- a/js/src/ui/SelectionList/selectionList.css
+++ b/js/src/ui/SelectionList/selectionList.css
@@ -16,6 +16,7 @@
 */
 
 .item {
+  border: 2px solid transparent;
   display: flex;
   flex: 1;
   height: 100%;
@@ -23,7 +24,9 @@
   width: 100%;
 
   &:hover {
-    box-shadow: inset 0 0 0 2px rgb(255, 255, 255);
+    border-color: transparent;
+    filter: none;
+    opacity: 1;
   }
 
   .content {
@@ -31,7 +34,7 @@
     width: 100%;
 
     &:hover {
-      background: rgba(255, 255, 255, 0.25);
+      background-color: rgba(255, 255, 255, 0.5);
     }
   }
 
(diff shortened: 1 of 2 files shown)


js/src/ui/SelectionList/selectionList.css
```diff
@@ -16,22 +16,25 @@
...
   flex: 1;
   height: 100%;
   position: relative;
   width: 100%;
 
   &:hover {
-    box-shadow: inset 0 0 0 2px rgb(255, 255, 255);
+    border-color: transparent;
+    filter: none;
+    opacity: 1;
```
ngotchac: Isn't `filter` and `opacity` only needed for the `.unselected` class ?

js/src/ui/SelectionList/selectionList.css
```diff
@@ -43,13 +46,17 @@
...
+
+  &.default {
+    border-color: rgba(255, 255, 255, 0.75);
+  }
 }
 
 .unselected {
-  filter: grayscale(100%);
-  opacity: 0.5;
+  filter: grayscale(10%);
```
ngotchac: I liked it a bit more gray actually... Could increase this a bit

js/src/ui/SelectionList/selectionList.css
jacogr: This just resets everything. Could be specific, i.e. `&.unlselected:hover { ... }` - simpler this way, no impact either way.

js/src/ui/SelectionList/selectionList.css
jacogr: Me too, actually. Had to contain it a bit to not look totally disabled, yet, but more subtle.

PR #4689, opened 2017-02-27
Display badges on summary view

jacogr:
parity 2017-02-27 13-59-25

Normal view with description available on expansion.

### DIFF
--- a/js/src/ui/Balance/balance.css
+++ b/js/src/ui/Balance/balance.css
@@ -28,6 +28,7 @@
 }
 
 .empty {
+  min-height: 24px;
   line-height: 24px;
   opacity: 0.25;
   overflow: hidden;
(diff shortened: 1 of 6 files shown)


ngotchac:
The badge icons are a bit big IMO, and they might be a little bit off (too high by 1 or 2px)... ?

jacogr:
I actually put a background (red) on the container while aligning, it matches up with the top of the badge & currency, i.e. container. (Unless my eyes are giving in.)

As for the sizes - I kept them exactly what we have currently, so they match on simple and expanded views. What actually needs to happen -

(a) We need to get real badge-y images (small, colourful - this was discussed elsewhere)
(b) the badges needs to float over the account icon

Not doing the above yet. (Which reminds me, actually need to log it, came up yesterday - https://github.com/ethcore/parity/issues/4707)

ngotchac:
Okay, fair enough then !

PR #4696, opened 2017-02-27
EC math functions

svyatonik:
Will need this for secret store.
There are no tests as these will be very artificial. Like - I do not know the answer, but my functions shows result A, so let it be A. I could add such tests, if you think these would be helpful.

### DIFF
--- a/ethkey/src/lib.rs
+++ b/ethkey/src/lib.rs
@@ -28,6 +28,7 @@ mod brain;
 mod error;
 mod keypair;
 mod keccak;
+pub mod math;
 mod prefix;
 mod random;
 mod signature;
--- a/ethkey/src/lib.rs
+++ b/ethkey/src/lib.rs
@@ -28,7 +28,6 @@ mod brain;
 mod error;
 mod keypair;
 mod keccak;
-pub mod math;
 mod prefix;
 mod random;
 mod signature;
@@ -44,6 +43,8 @@ pub trait Generator {
 	/// Should be called to generate new keypair.
 	fn generate(self) -> Result<KeyPair, Error>;
 }
+
+pub mod math;
 
 pub use self::brain::Brain;
 pub use self::error::Error;
--- a/ethkey/src/math.rs
+++ b/ethkey/src/math.rs
@@ -16,10 +16,11 @@
 
 use super::{SECP256K1, Public, Secret, Error};
 use secp256k1::key;
+use secp256k1::constants::{GENERATOR_X, GENERATOR_Y};
 
 /// Inplace multiply public key by secret key (EC point * scalar)
 pub fn public_mul_secret(public: &mut Public, secret: &Secret) -> Result<(), Error> {
-	let key_secret = key::SecretKey::from_slice(&SECP256K1, &secret[..])?;
+	let key_secret = secret.to_secp256k1_secret()?;
 	let mut key_public = to_secp256k1_public(public)?;
 	key_public.mul_assign(&SECP256K1, &key_secret)?;
 	set_public(public, &key_public);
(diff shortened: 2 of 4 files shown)


svyatonik:
P.S.: at least these work fine in secretstore

ethkey/src/math.rs
```diff
@@ -0,0 +1,122 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use super::{SECP256K1, Public, Secret, Error};
+use secp256k1::key;
+
+/// Inplace multiply public key by secret key (EC point * scalar)
+pub fn public_mul_secret(public: &mut Public, secret: &Secret) -> Result<(), Error> {
```
NikVolf: why not to make this methods on corresponding structs?

ethkey/src/math.rs
```diff
@@ -0,0 +1,122 @@
...
+	let mut key_secret = to_secp256k1_secret(secret)?;
+	key_secret.inv_assign(&SECP256K1)?;
+
+	*secret = key_secret.into();
+	Ok(())
+}
+
+/// Compute power of secret key inplace (secret ^ pow).
+/// This function is not intended to be used with large powers.
+pub fn secret_pow(secret: &mut Secret, pow: usize) -> Result<(), Error> {
```
rphmeier: the code of this function could be improved with `match`:

```rust
match pow {
    0 => {}
    1 => {}
    pow => {}
}
Ok(())
```

ethkey/src/math.rs
svyatonik: I considered this, but these are very specific (i.e. won't be used by most of users) so I decided that it is better to place math functions in separate module (to avoid garbage in IntelliSense && etc). I could move, if you think it is better

ethkey/src/math.rs
```diff
@@ -0,0 +1,122 @@
...
+	let secret_copy = secret.clone();
+	for _ in 1..pow {
+		secret_mul(secret, &secret_copy)?;
+	}
+
+	Ok(())
+}
+
+/// Return base point of secp256k1
+pub fn generation_point() -> Public {
```
NikVolf: isn't it the same as taking GENERATOR_X, GENERATOR_Y from here using less cpu cycles?

https://github.com/ethcore/rust-secp256k1/blob/master/src/constants.rs#L53

ethkey/src/math.rs
```diff
@@ -0,0 +1,122 @@
...
+	let public_data = {
+		let mut temp = [4u8; 65];
+		(&mut temp[1..65]).copy_from_slice(&public[0..64]);
+		temp
+	};
+
+	Ok(key::PublicKey::from_slice(&SECP256K1, &public_data)?)
+}
+
+fn to_secp256k1_secret(secret: &Secret) -> Result<key::SecretKey, Error> {
```
NikVolf: could be a consuming method on `Secret` struct as well

ethkey/src/math.rs
NikVolf: i already use them, see extended.rs

ethkey/src/math.rs
svyatonik: cool, but idk how to combine  these :) could you point the proper function?

ethkey/src/lib.rs
```diff
@@ -28,6 +28,7 @@ mod brain;
 mod error;
 mod keypair;
 mod keccak;
+pub mod math;
```
rphmeier: style: public modules should be separated from private.

ethkey/src/math.rs
NikVolf: something like this

```
		let mut public_sec_raw = [0u8; 65];
		public_sec_raw[0] = 4;
		public_sec_raw[1..33].copy_from_slice(&GENERATOR_X);
		public_sec_raw[33..65].copy_from_slice(&GENERATOR_Y);
		let public_sec = PublicKey::from_slice(&SECP256K1, &public_sec_raw).map_err(|_| Error::InvalidPoint)?;
```

ethkey/src/math.rs
rphmeier: rather than as methods, it should use `std::ops::{Add, Mul}` and maybe their `_Assign` variants. I assume the failure condition is just that the context wasn't created with the correct capabilities. Since our context is global and always created with full capabilities, it should be safe to assert that no errors occur.

rphmeier: Failure conditions like invalid scalar also shouldn't be possible; if they are then it's a bug in the type allowing us to create invalid secrets.

NikVolf: it's up to you, but i would like to refactor my `extended.rs` code to use these functions also, would be more handy to just use structs

svyatonik: ok, I'll change what is appropriate to ops && make another functions to be methods of Public/Secret

ethkey/src/math.rs
rphmeier: or a `From` implementation

ethkey/src/math.rs
svyatonik: @rphmeier actually I have met such situations while working on secretstore. For example - I have been able to construct this Secret basing on this constant:
https://github.com/ethcore/rust-secp256k1/blob/master/src/key.rs#L47
but it was an incorrect Secret :)
Another example is trying to calculate s1 * inv(s1 - s1) also leads to incorrect Secret :)

ethkey/src/math.rs
svyatonik: cool, will add these :)

ethkey/src/math.rs
NikVolf: yeah if there are situations where errors possible, `std::ops::*` is no go

rphmeier: in that case, then it's probably best not to use ops, just methods on the type which can error.

svyatonik: @NikVolf @rphmeier I have just remembered that Public is alias for H512. I think it is not ok to add EC-specific math to hash type. What do you guys think?

svyatonik: still can move some of functions to Secret:: methods, though

NikVolf: yeah
and as far as i remember @debris wanted to wrap `H512` to `Public` the same as `H256` wrapped to `Secret`, so maybe leave it until then

svyatonik: moved appropriate functions to Secret methods

ethkey/src/math.rs
svyatonik: done

ethkey/src/math.rs
svyatonik: done

ethkey/src/math.rs
svyatonik: These are inplace methods => consuming does not fits here.
`From` must not fail. This could fail.

ethkey/src/lib.rs
svyatonik: done

NikVolf:
Any tests against well-known vectors?

svyatonik:
@NikVolf seed first comment. If you have any 'well-known' vector, please share - I'll add

NikVolf:
Fair enough, we can hardcode some samples later in case to be confident with later changes

Issue #4474, opened 2017-02-08
Cleanup Dapps & UI servers


tomusdrw:
Currently we support different ways to access our UI, like:
1. Configured proxy.pac
2. Serving from localhost/127.0.0.1
3. Supporting `parity.web3.site` domain
4. Customizing `--*-interface` and serving from external IPs

There is a lot of weird edge cases and convoluted code to support all those schemes. Different schemes have also different security consequences (properly identifying dapp is sometimes difficult see: #3382 )
I would like to propose a revised/simplified approach for accesing UI.

## CLI options summary:
`--ui-port 8180`, `--ui-interface 127.0.0.1`, `--ui-domain web3.site`
`--dapps-port 8080`
(Get rid of `--dapps-user`, `--dapps-pass`, `--dapps-hosts`, `--dapps-interface`)

## Addressing
We should support following addressing scheme for both UI and Dapps:
`(<params>.)*.<handler>.<domain>:<port>`

### Examples:
`parity.web3.site:8180` - opens up UI
`<dapphash>.web3.site:8080` - opens up a network dapp/content
`<token-url>.web.web3.site:8080` - opens a web-proxy
`api.web3.site:8080/content/<hash>` - exposes a REST API 

`<domain>` should be configurable, so when set to `localhost` it becomes
`parity.localhost:8180` - opens UI
`<dapphash>.localhost:8080` - opens a network dapp/content

## External access
To expose Parity and listen on remote addresses one can run with:
`--ui-interface <extip> --ui-domain <customdomain>`

And then access `parity.<customdomain>:8180`.

In case one doesn't want to register traditional `<customdomain>`, he can use either:
1. `/etc/hosts` - configure your own domain there
2. `proxy.pac` - served by Parity as it's now
It's already a very advanced use case and I think it's ok to do it this way

## Required modifications
Both UI and core should be updated to support only this kind of addressing scheme.

5chdn:
From #5339:

>Due to a mistake in a script, I called parity with the following (note that shell expansion >didn't occur for the code between the backticks):
>```
>parity --ui-interface `hostname -I`
>```
>The error printed to stdout was `Invalid port specified: 8180`.
>
>Expected error message would be:
>```
>Invalid interface: `hostname -I`
>```

Tbaut:
Stale issue. Please reopen if you think it's still relevant for Parity UI.

PR #4710, opened 2017-02-28
Allow specifying extra cors headers for dapps

tomusdrw:
Closes: https://github.com/ethcore/parity/issues/4618 (solution is to use `--dapps-cors all`)

It's a temporary fix until #4474 will be solved (that will happen when pub/sub will be implemented)


gavofyork:
@todr this important for 1.6?

tomusdrw:
It allows to use full-featured Parity UI when running in docker container, I would say it'd be nice to have it in 1.6

PR #4713, opened 2017-03-01
provide uncle size where available in RPC

rphmeier:
cc @arkpar 

Also adds a test for the `Block` and `RichBlock` types that `size: None` goes to `null` as expected.


tomusdrw:
I think it should actually be 0, if we want to be aligned with geth.

gavofyork:
geth is arbitrary and, in any sensible reading, flat-out wrong. `null` is the most correct answer here and we'll stick with it.

tomusdrw:
@gavofyork @rob do I guess the change here should be reverted since it will return non-null values if the full.block is available for an uncle.

rphmeier:
@tomusdrw Not sure if I follow, that's the whole point of the PR. The RPC isn't that well designed, so it is supposed to return the same result as `eth_getBlockBy*`, which includes a `size`. The size isn't always available, but we should return it when it is.

tomusdrw:
So I guess it should consistently return null for uncles. Don't see a point in returning non-reliable size in any case.

PR #4716, opened 2017-03-01
Add registrar fields

keorn:
Also put [registrar contract](https://github.com/ethcore/contracts/blob/master/SimpleRegistry.sol) in the genesis of `--chain dev`.

Can also move the `"registrar"` to `CommonParams`.


jacogr:
Tested with new POA chain, registry deployed & configured, it is pulled in and available.

PR #4719, opened 2017-03-02
Enable --warp by default

arkpar:


gavofyork:
does this interact well with `--tracing` &c.?

arkpar:
`--tracing`, `--fatdb`, etc would start collecting data from the restoration point. I agree that this makes little sense. Fixed in the latest commit.

gavofyork:
ok - one final thing: if the user states `--warp` while also configuring it (either through CLI or past options) `tracing`, `fat db` or `archive` then it should warn them that warp is disabled due to these.

PR #4723, opened 2017-03-02
Update Uglify (fix to 2.8.2) to fix binary builds

jacogr:
Fixes https://github.com/ethcore/parity/issues/4722


jacogr:
Tested locally with `cargo build --release --no-default-features --features ui` 

Merging to get a `js-precompiled` built to re-run the local master tests. (Equivalent to the above, final check)

PR #4725, opened 2017-03-02
Secretstore DKG protocol draft + distributed encryption/decryption tests

svyatonik:
For DKG protocol, encryption && decryption details, see:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.124.4128&rep=rep1&type=pdf

For fully-functional prototype we are currently missing following things (will be added in next PRs):
1) decryption protocol (should be quite easy to implement)
2) network transport for both protocols
3) address an issue (or just live with it) with decryption when number of nodes, participating in decryption is odd (I'll add an issue for tracking this when/if this PR will be merged)
4) add implementation for cases when scheme is 1-of-* or if there is only one KeyServer acting (this protocol is currently not supporting such cases)
5) store secret shares + encrypted document secret in DB (after milestone 1 db contains only plain document secret)
6) add tests for session when some nodes are misbehaving && there are complaints sent. As I suppose there could be some things I have missed during this draft implementation


svyatonik:
7. ACL checker implementation

svyatonik:
OK - I'm going to extend this PR with decryption protocol today/tomorrow. So marking it as in-progress.

svyatonik:
Implemented decryption protocol (no commutative enc yet + lack of tests).
Next steps: add tests for decryption protocol + continue with the list above.

gavofyork:
did you contact the authors of the paper? any idea why it doesn't work with odd sets?

svyatonik:
@gavofyork not yet. will try tomorrow. no any ideas

secret_store/src/key_server_cluster/decryption_session.rs
```diff
@@ -102,8 +102,10 @@ pub enum SessionState {
 
 impl Session {
 	/// Create new decryption session.
-	pub fn new(id: SessionId, access_key: Secret, self_node_id: Public, encrypted_data: EncryptedData, acl_storage: Arc<AclStorage>, cluster: Arc<Cluster>) -> Self {
-		Session {
+	pub fn new(id: SessionId, access_key: Secret, self_node_id: Public, encrypted_data: EncryptedData, acl_storage: Arc<AclStorage>, cluster: Arc<Cluster>) -> Result<Self, Error> {
```
rphmeier: this function is getting quite long...maybe extract out to a `Params` struct for readability in invocations?

rphmeier: e.g. 
```rust
new(Params {
    id: id,
    access_key: my_secret,
    // ...
}
```

PR #4734, opened 2017-03-03
Support both V1 & V2 DataChanged events in registry

jacogr:
- Select registry ABI based on code hash (V1 - Ropsten & Homestead, V2 otherwise)
- Render V2 DataChanged events (owner field not available)


gavofyork:
New chain info merged.

PR #4736, opened 2017-03-03
Update SelectionList indicators

jacogr:
- Closes https://github.com/ethcore/parity/issues/4737
- Move selection indicators (checked & default) to the left - no overlapping with tags
- Remove select tick, keep default/star tick

parity 2017-03-03 13-50-30
parity 2017-03-03 13-50-03
parity 2017-03-03 13-49-40


js/src/ui/SelectionList/selectionList.css
```diff
@@ -58,7 +79,3 @@
   filter: grayscale(10%);
   opacity: 0.75;
```
ngotchac: Can we set it to `0.5`? It would be more obvious IMO

jacogr: ? Not sure it is for this PR? 

1. As-is in master currently?
2. Remember the original complaint, it looked "disabled" so users didn't know what to do.

Agreed though, I'm "ok" but not "happy" with the selections, warrants a bit more playing with the colors, borders & opacity.

ngotchac: Well it's somehow related to this PR since it removes the `tick`, thus makes non-selected accounts less obvious : that's why I'm suggesting an opacity of 0.5... After all there are no such things as disabled accounts, so I'm not sure if that's a real issue ?

jacogr: There are disabled accounts. Hardware wallets show up as disabled when they are not plugged in. (Same on the signer - when not in it is disabled.)

My preference for the borders (and actual icon colours) would be the "almost yellow" as in the CreateAccount dialog icons. But that needs some playing. (Basically the opposite of the link blue)

PR #4749, opened 2017-03-03
Fix SectionList hovering issue

ngotchac:
Fixes issue where items stretch on a SectionList line while hovering the line but on no item (thus when on a line with < 3 items): eg. in Accounts, Dapps...  

Closes https://github.com/ethcore/parity/issues/4665


js/src/ui/SectionList/sectionList.css
```diff
@@ -39,41 +38,33 @@ $widthExpanded: 42%;
...
-          flex: 0 1 $widthShrunk;
-          max-width: $widthShrunk;
-
-          &:hover {
-            flex: 0 0 $widthExpanded;
-            max-width: $widthExpanded;
-          }
-        }
-      }
+    .item.stretchOn:hover {
```
jacogr: There is a reason for the shrinking. When the name (Title) is at exactly 33% and you hover over the row the affected item does not shrink. (This is the original implementation, sadly the simplest, sadly broken in edge-cases)

ngotchac:
Again, I played with all lengths of titles and all, and could not find any bug... Please be a bit more specific if possible

jacogr:
Will need to do it after the release crunch is over - will first need to create the issue (as it manifested, could probably mean rolling back to be 100% sure it is what it is) and then go from there. 

As I said very valid as to why, what the specific circumstances are need to find again. Simpler is always better, which is why we started exactly where you are atm. Also possibly other changes could have squashed it, with no need for the work-around anymore. (So either way, rooting for this, just not taking any risks.)

jacogr:
Ok, did a hacky rollback to an old local WIP branch. My ropsten chain still had problematic instances (under vaults), so got lucky. Something changed somewhere in the containers, even with the 33.33% original, no original non-stretches when updated. 

However could also spot what you meant with "strange overflow issues, hence the 0 width". (e.g. some additional cases where issues crops up, tested this with accounts - obviously easier since the names are editable)

So we _seems_ to cater for both, happy that original/simple can be back.

PR #4753, opened 2017-03-03
Optimistic snapshot

### LINKED ISSUE #4727: Optimistic snapshot creation
1. Take snapshot with `pruning-history` confirmations (store in "pending" directory)
2. Store until `SNAPSHOT_PROPAGATE` confirmations.
3. At that point, If snapshot block hash == canonical block hash, then move to "local" directory.
    otherwise, delete.

This should enable basically every parity node on the network to create a snapshot with a very small chance of failure, as opposed to solely archive nodes.

@gavofyork one for 1.6?

rphmeier:
Closes #4727 

cc @keorn

This will need a bit of testing before merge, so will ice this for now.

Changes snapshot creation behavior. Previously, snapshots were created into `snapshot_dir` after a delay and immediately propagated after that. The delay was too long for pruned nodes to be able to create the snapshot.

Now, snapshots are initially created at `--pruning-history` blocks behind, and are only propagated after a certain amount of time.

### DIFF
--- a/ethcore/src/service.rs
+++ b/ethcore/src/service.rs
@@ -20,7 +20,7 @@ use util::*;
 use io::*;
 use spec::Spec;
 use error::*;
-use client::{Client, ClientConfig, ChainNotify};
+use client::{BlockChainClient, Client, ClientConfig, ChainNotify};
 use miner::Miner;
 
 use snapshot::ManifestData;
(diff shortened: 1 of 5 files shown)


rphmeier:
In-progress because I'm actually not a huge fan of the approach I took here. I think it might be better not to do so much more directory shuffling and just prevent propagation in the sync code until the snapshot is X blocks old + canonical. The annoying part then is to delete the "bad" snapshot in the event of reorg, which can't easily be done from the sync code and is more convenient to do via the "solidification" mechanism.

rphmeier:
Closing for now; memory-based pruning history is enough to ensure most nodes take a snapshot now, anyway.

PR #4757, opened 2017-03-04
EIP-168, 169: Dust protection

arkpar:
This implements 
https://github.com/ethereum/EIPs/issues/168 Variant C
https://github.com/ethereum/EIPs/issues/169 Variant C

### DIFF
--- a/ethcore/src/state/account.rs
+++ b/ethcore/src/state/account.rs
@@ -315,6 +315,11 @@ impl Account {
 		self.code_hash == SHA3_EMPTY
 	}
 
+	/// Check if account is basic (Has no code).
+	pub fn is_basic(&self) -> bool {
+		self.code_hash == SHA3_EMPTY
+	}
+
 	/// Return the storage root associated with this account or None if it has been altered via the overlay.
 	pub fn storage_root(&self) -> Option<&H256> { if self.storage_is_clean() {Some(&self.storage_root)} else {None} }
 
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -74,8 +74,6 @@ pub enum ProvedExecution {
 	Complete(Executed),
 }
 
-const RIPEMD_BUILTIN: Address = H160([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3]);
-
 #[derive(Eq, PartialEq, Clone, Copy, Debug)]
 /// Account modification state. Used to check if the account was
 /// Modified in between commits and overall.
@@ -275,13 +273,13 @@ enum RequireCache {
 
 /// Mode of dealing with null accounts.
 #[derive(PartialEq)]
-pub enum CleanupMode {
+pub enum CleanupMode<'a> {
 	/// Create accounts which would be null.
 	ForceCreate,
 	/// Don't delete null accounts upon touching, but also don't create them.
 	NoEmpty,
-	/// Add encountered null accounts to the dirty-set, to be deleted later.
-	KillEmpty,
+	/// Mark all touched accounts.
+	TrackTouched(&'a mut HashSet<Address>),
 }
 
 const SEC_TRIE_DB_UNWRAP_STR: &'static str = "A state can only be created with valid root. Creating a SecTrieDB with a valid root will not fail. \
@@ -563,26 +561,30 @@ impl<B: Backend> State<B> {
 		let is_value_transfer = !incr.is_zero();
 		if is_value_transfer || (cleanup_mode == CleanupMode::ForceCreate && !self.exists(a)?) {
 			self.require(a, false)?.add_balance(incr);
-		} else if cleanup_mode == CleanupMode::KillEmpty && self.exists(a)? {
-			self.touch(a)?;
+		} else if let CleanupMode::TrackTouched(set) = cleanup_mode {
+			if self.exists(a)? {
+				set.insert(*a);
+				self.touch(a)?;
+			}
 		}
-
 		Ok(())
 	}
 
 	/// Subtract `decr` from the balance of account `a`.
-	pub fn sub_balance(&mut self, a: &Address, decr: &U256) -> trie::Result<()> {
+	pub fn sub_balance(&mut self, a: &Address, decr: &U256, cleanup_mode: &mut CleanupMode) -> trie::Result<()> {
 		trace!(target: "state", "sub_balance({}, {}): {}", a, decr, self.balance(a)?);
-		if !decr.is_zero() || !self.exists(a)? {
+		if !decr.is_zero() {
 			self.require(a, false)?.sub_balance(decr);
 		}
-
+		if let CleanupMode::TrackTouched(ref mut set) = *cleanup_mode {
+			set.insert(*a);
+		}
 		Ok(())
 	}
 
 	/// Subtracts `by` from the balance of `from` and adds it to that of `to`.
-	pub fn transfer_balance(&mut self, from: &Address, to: &Address, by: &U256, cleanup_mode: CleanupMode) -> trie::Result<()> {
-		self.sub_balance(from, by)?;
+	pub fn transfer_balance(&mut self, from: &Address, to: &Address, by: &U256, mut cleanup_mode: CleanupMode) -> trie::Result<()> {
+		self.sub_balance(from, by, &mut cleanup_mode)?;
 		self.add_balance(to, by, cleanup_mode)?;
 		Ok(())
 	}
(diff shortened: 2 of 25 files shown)


gavofyork:
added a variant C to 168, which might be worth considering

gavofyork:
c.c.2 is definitely not redundant.

gavofyork:
will need resolving

arkpar:
resolved

ethcore/src/state/account.rs
```diff
@@ -315,6 +315,11 @@ impl Account {
 		self.code_hash == SHA3_EMPTY
 	}
 
+	/// Check if account is basic (Has no code).
+	pub fn is_basic(&self) -> bool {
+		self.code_hash == SHA3_EMPTY
```
gavofyork: don't think this will work for newly created accounts

ethcore/src/state/mod.rs
```diff
@@ -95,6 +97,7 @@ enum AccountState {
 /// account (`None`)
 struct AccountEntry {
 	account: Option<Account>,
+	clean_balance: Option<U256>,
```
gavofyork: probably worth documenting these members now - `clean_balance` isn't immediately obvious in meaning. would probably rename to `old_balance` or something similar to emphasise its underlying semantics.

ethcore/src/state/mod.rs
```diff
@@ -121,12 +128,14 @@ impl AccountEntry {
 		AccountEntry {
 			account: self.account.as_ref().map(Account::clone_dirty),
 			state: self.state,
+			clean_balance: self.clean_balance,
```
gavofyork: appears after other fields (others constructors appear prior)

ethcore/src/state/mod.rs
```diff
@@ -537,13 +548,8 @@ impl<B: Backend> State<B> {
...
 			self.require(a, false)?.add_balance(incr);
-		} else {
-			match cleanup_mode {
-				CleanupMode::KillEmpty(set) => if !is_value_transfer && self.exists(a)? && !self.exists_and_not_null(a)? {
-					set.insert(a.clone());
-				},
-				_ => {}
-			}
+		} else if cleanup_mode == CleanupMode::KillEmpty && self.exists(a)? {
+			self.require(a, false)?;
```
gavofyork: consider introducing a function `fn touch(&self, a: ...) -> ... { self.require(a, false) }` and using that in order to make clear the intent here.
this doesn't make use of `is_value_transfer`; i am skeptical that precise semantics have been preserved.

ethcore/src/state/mod.rs
```diff
@@ -672,17 +672,30 @@ impl<B: Backend> State<B> {
...
 	/// Clear state cache
 	pub fn clear(&mut self) {
 		self.cache.borrow_mut().clear();
 	}
 
+	/// Remove any touched empty or dust accounts.
+	pub fn kill_garbage(&mut self, remove_empty_touched: bool, min_balance: &Option<U256>, kill_contracts: bool) -> trie::Result<()> {
+		let to_kill: HashSet<_> = {
+			self.cache.borrow().iter().filter_map(|(address, ref entry)|
+			if (entry.is_dirty() || address == &RIPEMD_BUILTIN) &&
```
gavofyork: `RIPEMD` is very ugly - can we avoid having it here?

also, not entirely sure what these conditions are supposed to achieve.

ethcore/src/state/mod.rs
```diff
@@ -672,17 +672,30 @@ impl<B: Backend> State<B> {
...
+			((remove_empty_touched && entry.is_null())
+			|| min_balance.map_or(false, |ref balance| entry.account.as_ref().map_or(false, |account|
```
gavofyork: indentation off by one (still have an open paren here from previous line)

ethcore/src/state/mod.rs
```diff
@@ -672,17 +672,30 @@ impl<B: Backend> State<B> {
...
+				(account.is_basic() || kill_contracts)
+				&& account.balance() < balance && entry.clean_balance.as_ref().map_or(false, |b| account.balance() < b)))) {
```
gavofyork: could do with some commentary.

ethcore/src/state/substate.rs
```diff
@@ -28,9 +28,6 @@ pub struct Substate {
 	/// Any accounts that have suicided.
 	pub suicides: HashSet<Address>,
 
-	/// Any accounts that are tagged for garbage collection.
-	pub garbage: HashSet<Address>,
```
gavofyork: i'm a bit concerned that the alteration of semantics (from explicit collection of gardbage-collected accounts to the implicit reliance on touching the cache) here will lead to subtle consensus flaws. i don't remember exactly, but when i coded this originally i chose this means of doing it for a reason.

ethcore/src/verification/verification.rs
```diff
@@ -392,6 +400,12 @@ mod tests {
 		verify_block_family(&header, bytes, engine, bc)
 	}
 
+	fn unordered_test(bytes: &[u8], engine: &Engine) -> Result<(), Error> {
```
gavofyork: mark with `#[test]` if only used in the test?

gavofyork:
not at all keen on the removal of `garbage` from substate - it departs significantly from the structure of the YP spec and relies on largely undocumented and unspecified behaviour of the account cache.

arkpar:
Current implementation does not handle state reverts properly. This is what caused the original RIPEMD built-in bug, for which there's going to be an exception added to the YP. `State` already tracks touched/dirty states for entries, why repeat that logic elsewhere? Also, how's `State` is less documented and specified than `Substate`/`Executive`?  As far as I can see it has even better documentation.

ethcore/src/verification/verification.rs
arkpar: It is part of the `tests` mod already, but not a test itself.

ethcore/src/state/account.rs
arkpar: If it's  newly created account that does not have code it is a basic account, right? It stops being basic only after it gets the code.

ethcore/src/state/mod.rs
arkpar: I've added this because this new implementation revealed a subtle consensus flaw in the previous one. If a built-in runs out of gas it's account still gets deleted. This is now in sync with the go-ethereum and other clients which also have a special case for this particular built-in account.

ethcore/src/state/mod.rs
arkpar: This branch only executes when `is_value_transfer == false`

ethcore/src/state/mod.rs
rphmeier: that said, this special-casing is only for the mainnet; our code should reflect that.

ethcore/src/state/mod.rs
arkpar: This is in line with the declaration order. Other constructors are exceptions because they use `account` before it is moved into `self.account`. Will fix here for consistency.

gavofyork:
Well I guess the point is that while the notion of "in the cache" may be similar to the definition of "touched" within the strict context of EIP/YP, because it was not explicitly implemented or specified to be the same concept, we run the risk that what might reasonably be considered an internal change or optimisation to our state-caching code will accidentally and subtly break consensus.

gavofyork:
i now remember the reason we don't want to do it this way: reversion of an EVM call-path should also revert the touchedness of an account, in much the same way it reverts other aspects of substate.

arkpar:
This has been updated to track touched accounts in the substate

rphmeier:
Could use a merge

ethcore/src/state/account.rs
gavofyork: i guess it's fine - this particular function should work at least (though at one point `new_contract` actually gave a non-basic account which just happened to have no code yet, but looks like it's now just an uncached basic account).

gavofyork:
Variant C of 168 purposefully doesn't mention "touched" accounts:

> At the end of the transaction, any account created or whose balance is decreased by the execution of that transaction which is now dust SHALL instead become non-existent (i.e. deleted).

Yet this PR seems to operate solely on them.

gavofyork:
Variant C of 168 purposefully doesn't mention "touched" accounts:

> At the end of the transaction, any account created or whose balance is decreased by the execution of that transaction which is now dust SHALL instead become non-existent (i.e. deleted).

Yet this PR seems to operate solely on them.

arkpar:
@rphmeier I'll merge it once @gavofyork is happy with the implementation.

arkpar:
@gavofyork Quoting EIP-161:
> An account is considered to be touched when it is involved in any potentially state-changing operation

Clearly this includes all EIP-168C accounts as well. A check for "Created or whose balance is decreased by the execution of that transaction which is now dust" is performed "at the end of the transaction" in `kill_garbage` as specified.

Could you point out what's exactly wrong with that?

I don't think there's any sense in maintaining yet another additional list of accounts that were "Created or whose balance is decreased by any non-zero value" as this is, again, just wasting memory.

ethcore/src/state/mod.rs
```diff
@@ -538,31 +550,30 @@ impl<B: Backend> State<B> {
...
 		}
-
 		Ok(())
 	}
 
 	/// Subtract `decr` from the balance of account `a`.
-	pub fn sub_balance(&mut self, a: &Address, decr: &U256) -> trie::Result<()> {
+	pub fn sub_balance(&mut self, a: &Address, decr: &U256, cleanup_mode: &mut CleanupMode) -> trie::Result<()> {
 		trace!(target: "state", "sub_balance({}, {}): {}", a, decr, self.balance(a)?);
-		if !decr.is_zero() || !self.exists(a)? {
```
gavofyork: reason for removal?

arkpar: This function is never called for non-existing accounts. However it is indeed incorrect to assume it here. I'll restore it.

gavofyork:
> Could you point out what's exactly wrong with that?

the concern was with the prefilter of "touched" being applied to to EIP168's "any account created or whose balance is decreased". just need to prove that EIP168's condition strictly implies the property of being "touched". i suspect it does, but it does need a reasonably formal proof before we can be sure this properly implements the EIP.

ethcore/src/state/mod.rs
```diff
@@ -106,6 +110,10 @@ impl AccountEntry {
 		self.state == AccountState::Dirty
 	}
 
+	fn is_null(&self) -> bool {
```
arkpar: better name: `exists_and_is_null`

PR #4762, opened 2017-03-04
Update README.md 

jdetychey:
Fix for Cargo fail when building from source

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,11 @@ $ cargo build --release
 ```
 
 This will produce an executable in the `./target/release` subdirectory.
+Note: if cargo fails to parse manifest try:
 
+```bash
+$ ~/.cargo/bin/cargo build --release
+```
 ----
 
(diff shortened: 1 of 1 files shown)


jdetychey:
Build from source with:
```
$ cargo build --release
```
Can lead to this error:
```
failed to parse manifest at `~/parity/Cargo.toml`
Caused by:
  could not parse input as TOML
Cargo.toml:58:9 expected a key but found an empty string
Cargo.toml:58:9-58:10 expected `.`, but found `'`
```
The one line installer of rust leads to multiple version of cargo.:
```
$ curl https://sh.rustup.rs -sSf | sh
```
Notably:
```
cargo --version
cargo 0.8.0 (built 2016-03-22)
```
and:
```
$ ~/.cargo/bin/cargo build --version
cargo 0.16.0-nightly (6e0c18c 2017-01-27)
```
Therefore using instead:
```
~/.cargo/bin/cargo build --release
```
Fix it

PR #4771, opened 2017-03-05
Update Settings/Proxy view to match entries in proxy.pac

bjornwgnr:
Replaces 'home.parity' with 'parity.web3.site' in Settings/Proxy view

### DIFF
--- a/js/src/views/Settings/Proxy/proxy.js
+++ b/js/src/views/Settings/Proxy/proxy.js
@@ -51,9 +51,9 @@ export default class Proxy extends Component {
               <div>
                 <FormattedMessage
                   id='settings.proxy.details_0'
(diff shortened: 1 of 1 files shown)


gavofyork:
not actually true any more - we bought web3.site so it all just works, as long as they go to 8081. the entire proxy.pac thing needs removing completely.

tomusdrw:
Sort of. If we switch to support only domain-based access using `proxy.pac` will be necessary to easily support remote-accesed nodes (other options include altering `/etc/hosts` file or setting up local DNS entry).

I would keep it for now and re-visit that when #4474 is being done.

ddorgan:
[clabot:check]

ddorgan:
[clabot:check]

ddorgan:
[clabot:check]

debris:
@tomusdrw any update on this?

tomusdrw:
The page is still there and proxy.pac is still supported so best to update it.

PR #4800, opened 2017-03-07
Calibrate step before rejection

### LINKED ISSUE #4797: no sync on kovan in daemonized mode
parity: v1.5.4 and v1.5.6
changed from ropsten to kovan

running command `parity`: sync concordant to http://kovan-stats.parity.io/

issue: running daemonized mode with `parity daemon /tmp/parity_daemon.pid` 
- get's latest block, if /tmp/parity_daemon.pid not exists, 
- or last sync block after parity stopped and /tmp/parity_daemon.pid exists

on start but no further blocks and sync fails

keorn:
Closes #4797


ethcore/src/engines/authority_round.rs
```diff
@@ -83,6 +83,8 @@ pub struct AuthorityRound {
 	client: RwLock<Option<Weak<EngineClient>>>,
 	signer: EngineSigner,
 	validators: Box<ValidatorSet + Send + Sync>,
+	/// Is this Engine just for testing (prevents step calibration).
+	test: bool,
```
arkpar: Would be better to name it according to what it does. E.g. `disable_step_calibration`. Or, even better, invert it and name as `calibrate_step`

gavofyork:
@keorn need resolving.

PR #4804, opened 2017-03-07
Better icon on windows

arkpar:
Matches Mac icon now. 
screen shot 2017-03-07 at 17 55 59

Tray icon still requires some work. We need someone to draw it in 16x16 from scratch. Scaling down does not work so well.
screen shot 2017-03-07 at 17 59 51


jacogr:
For the scaled-down tray icon version, would probably use the one without the round background, just the "P".

arkpar:
This is beyond my photoshop skill. It needs to have some kind of outline (maybe in the alpha-channel only) so that it looks good on dark and white background.

PR #4805, opened 2017-03-07
Update Wallet to new Wallet Code

ngotchac:
Updates the UI to use the new Wallet.
Now one can deploy contracts with a Wallet.

The Wallet Library should be deployed to any network where we want to have it.
The bytecode + ABI are in `js/src/contracts/code/wallet.js` : `walletLibrary` and `walletLibraryABI`.
The deployed address should then be added to the registry with `walletLibrary` as a key (the value of `walletLibraryRegKey`). It has been done for Kovan already.


jacogr:
Will add the above instructions to my auto-deploy tool as well. (So it just gets done by click of a button) https://github.com/ethcore/parity/pull/4726

js/src/util/tx.js
```diff
@@ -84,6 +73,114 @@ export function postTransaction (_func, _options, _values = []) {
...
 
+export function deploy (contract, _options, values, metadata = {}, statecb = () => {}) {
+  const options = { ..._options };
+  const { api } = contract;
+  const address = options.from;
+
+  return WalletsUtils
+    .isWallet(api, address)
+    .then((isWallet) => {
+      if (!isWallet) {
```
jacogr: +1 Really happy with this approach. (For both deploy & estimate)

js/src/ui/MethodDecoding/methodDecoding.js
```diff
@@ -120,6 +120,18 @@ class MethodDecoding extends Component {
         <span className={ styles.highlight }>
           { gas.toFormat(0) } gas ({ gasPrice.div(1000000).toFormat(0) }M/<small>ETH</small>)
         </span>
+        {
+          transaction.gasUsed
+            ? (
+              <span>
+                <span>used</span>
```
jacogr: Would have made sense to i18n these additions, but no matter, busy with this actual file in my conversion branch, so will be ok.

js/src/views/Signer/store.js
```diff
@@ -41,8 +41,9 @@ export default class SignerStore {
     this.balances = Object.assign({}, this.balances, balances);
   }
 
-  @action setLocalHashes = (localHashes) => {
-    if (!isEqual(localHashes, this.localHashes)) {
+  @action setLocalHashes = (localHashes = []) => {
+    // Use slice to make sure they are both Arrays (MobX uses Objects for Observable Arrays)
+    if (!isEqual(localHashes.slice(), this.localHashes.slice())) {
```
jacogr: Not sure which one is more performant from a MobX perspective - `.peek()` or `.slice()` (Typically do use the former when it is a known (i.e. on `this.`) value

jacogr:
Ok, happy with it from a code perspective, it needs to go for 1.6 - full end-to-end scenario testing will happen once merged.

PR #4808, opened 2017-03-07
Add replay protection

keorn:


arkpar:
Looks like it breaks compatibility with existing AuthorityRound networks, E.g. kovan. There should be a check for a transition block number `eip155_transition`. It is currently part of the ethash spec for some reason, but can be moved to `CommonParams`

keorn:
This is only one way protection, it does not add validation yet. We can introduce that later down the line.

PR #4832, opened 2017-03-09
CI: test coverage (for core)

General-Beck:
https://github.com/ethcore/parity/issues/4445

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -508,6 +508,7 @@ test-coverage:
   only:
     - coverage
   script:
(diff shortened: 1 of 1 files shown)


rphmeier:
AFAIK `cov.sh` only does coverage for core. What about coverage for UI code?

General-Beck:
@rphmeier do that, then js

rphmeier:
Ok, makes sense.

General-Beck:
@gavofyork need to check some tests. and empower

Issue #4836, opened 2017-03-09
Cant create dylib using parity.


Radagast-red:
When using parity as a library for dylib project. Project cant be build.

Please try to build attached project this with:
rustc 1.15.1 (021bd294c 2017-02-08)
cargo 0.16.0-nightly (6e0c18c 2017-01-27)

: cargo build
[parityIssue.tar.gz](https://github.com/ethcore/parity/files/831026/parityIssue.tar.gz)

and error occur during build.

rphmeier:
Gist here: https://gist.github.com/rphmeier/8179b3da8c7023eb2bb7641af237d9f6

rphmeier:
The only linker error I get is because of https://github.com/servo/heapsize requiring `jemalloc` (presumably, dylibs only link against the system allocator)

rphmeier:
Everything works with `crate_type="lib"`.

My take on this is that it would be far too much work to try and change out our memory instrumentation to work on the system allocator as well for such a niche use-case. If you'd like to make a PR, would gladly review and merge it.

paulperegud:
This method of linking was working fine against Parity v1.5.7 and below.

Radagast-red:
Can you post your error log?

Radagast-red:
Doesn't make sense. I have found commit:
https://github.com/ethcore/parity/commit/b5219bc7238006c75bf5d79784d1786a4e677e8f

This decorates snappy to be linked statically despite everything. So its copy ends up in each dynamic lib i believe. Simply removing static link fix this, works for lib and dylib.

debris:
@arkpar do we need to link snappy statically?

arkpar:
We do, but it looks like the `link` directive is  superfluous here since libsnappy is already linked by one of dependencies of ethcore-util

arkpar:
@Radagast-red does #4841 resolve an issue for you?

Radagast-red:
Yes, thank you.

PR #4841, opened 2017-03-09
Don't link libsnappy explicitly

arkpar:
It is already linked by `rocksdb-sys` which is a dependency.
Should help with #4836

### DIFF
--- a/util/src/snappy.rs
+++ b/util/src/snappy.rs
@@ -23,7 +23,6 @@ const SNAPPY_OK: c_int = 0;
 const SNAPPY_INVALID_INPUT: c_int = 1;
 const SNAPPY_BUFFER_TOO_SMALL: c_int = 2;
 
-#[link(name = "snappy", kind = "static")]
 extern {
 	fn snappy_compress(
 		input: *const c_char,


gavofyork:
if we switch from rocksdb to lmdb, wouldn't this break our linking?

debris:
> if we switch from rocksdb to lmdb, wouldn't this break our linking? 

Yes

what do you think about small *hack* with features?

**Cargo.toml**

```toml
[features]
default = ["rocksb"]
```

**snappy.rs**

```rust
// snappy is already linked by rocksdb
#[cfg(non(features='rocksb'))]
mod non_rocksb {
    #[link(name = "snappy", kind = "static")]
    extern {}
}
```

arkpar:
`#[link(name = "snappy", kind = "static")]` is wrong anyway, because it assumes that there's a library already available in the system. If at some point we'd remove rocksdb it would break in even worse way. It would use the system library that some systems have and some don't and it won't work with cross-compiling scenarios.
Proper way is to have `libsnappy-sys` and `libsnappy` rust bindings is a separate crate. if we switch from rocksdb to lmdb this will indeed break things and we'll have add that crate eventually.

PR #4844, opened 2017-03-09
Fix paste in Inputs

### LINKED ISSUE #4751: Input onPaste does not take selected text into account
e.g. GitHubHint contract, field for unhint is bytes32 - pasting a hex value e.g. `0x123` yields `0x0x123`

ngotchac:
Fixes #4751

### DIFF
--- a/js/src/ui/Form/Input/input.js
+++ b/js/src/ui/Form/Input/input.js
@@ -214,6 +214,7 @@ export default class Input extends Component {
 
   onChange = (event, value) => {
     event.persist();
+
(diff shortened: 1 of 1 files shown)


jacogr:
As discussed, we need to think about the timeouts, there must be a cleaner way. As it stands, does solve the issue at hand.

ngotchac:
I actually added what we talked about @jacogr ; seems to work and is a bit cleaner. I also implemented the TODO : revert to initial value on ESC

PR #4847, opened 2017-03-09
Fix TxViewer when no `to` (contract deployment)

### LINKED ISSUE #4799: TxViewer re-submitting contract fails
```
eth_sendTransaction([{"from":"<removed>","to":"0x","nonce":"0x0","value":"0x0","data":"0x606060405234610000575b60008054600160a060020a03191633600160a060020a03161790555b5b610190806100366000396000f300606060405263ffffffff60e060020a6000350416630900f0108114610045578063445df0ac146100605780638da5cb5b1461007f578063fdacd576146100a8575b610000565b346100005761005e600160a060020a03600435166100ba565b005b346100005761006d61012d565b60408051918252519081900360200190f35b346100005761008c610133565b60408051600160a060020a039092168252519081900360200190f35b346100005761005e600435610142565b005b6000805433600160a060020a03908116911614156101275781905080600160a060020a031663fdacd5766001546040518263ffffffff1660e060020a02815260040180828152602001915050600060405180830381600087803b156100005760325a03f115610000575050505b5b5b5050565b60015481565b600054600160a060020a031681565b60005433600160a060020a039081169116141561015f5760018190555b5b5b505600a165627a7a7230582046c837d8b8391beea3ca635cba119de618813eadaec7ddff2214888d1711b6db0029","gasPrice":"0x174876e801","gas":"0x2dc6c0"}]): -32602: Invalid params index.e19a53db46.js:1:344466
```
When `to` is empty, re-submitting transaction leads to `to` field set to `0x` and it's rejected with invalid params message.

ngotchac:
Fixes #4799

### DIFF
--- a/js/src/dapps/localtx/Transaction/transaction.js
+++ b/js/src/dapps/localtx/Transaction/transaction.js
@@ -227,18 +227,18 @@ export class LocalTransaction extends BaseTransaction {
 
   toggleResubmit = () => {
     const { transaction } = this.props;
-    const { isResubmitting, gasPrice } = this.state;
+    const { isResubmitting } = this.state;
 
-    this.setState({
+    const nextState = {
       isResubmitting: !isResubmitting
-    });
+    };
 
-    if (gasPrice === null) {
-      this.setState({
-        gasPrice: api.util.fromWei(transaction.gasPrice, 'shannon').toNumber(),
-        gas: transaction.gas.div(1000000).toNumber()
-      });
+    if (!isResubmitting) {
+      nextState.gasPrice = api.util.fromWei(transaction.gasPrice, 'shannon').toNumber();
+      nextState.gas = transaction.gas.div(1000000).toNumber();
     }
+
+    this.setState(nextState);
   };
 
   setGasPrice = el => {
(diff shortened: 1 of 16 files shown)


js/src/dapps/localtx/Transaction/transaction.js
```diff
@@ -256,23 +259,29 @@ export class LocalTransaction extends BaseTransaction {
 
     const newTransaction = {
       from: transaction.from,
-      to: transaction.to,
```
tomusdrw: That doesn't seem to be right. What if the transaction actually has valid `to`?

tomusdrw: Ah, just noticed that we're adding it later on. I don't understand how does it fix the problem though. Previously `transaction.to` was set to `0x` so we will set it anyway, no?

ngotchac: It's added after :) (L283)

ngotchac: So the issue was that the `to` field was set, and thus formatted to `hex`: `'' => '0x'`.
I guess the API should be modified in order to take this empty `to` case into account.

jacogr: Yeap. Current fix does the trick, i.e. not encoding it.

PR #4850, opened 2017-03-10
Update parity-ui-precompiled with branch

jacogr:
Closes https://github.com/ethcore/parity/issues/4813

Explicitly specify branch in Cargo.toml and update based on where we are. (So will reflect stable, beta or master). Saves manual work and mistakes.

### DIFF
--- a/js/scripts/release.sh
+++ b/js/scripts/release.sh
@@ -89,7 +89,7 @@ fi
 
 echo "*** Updating cargo parity-ui-precompiled#$PRECOMPILED_HASH"
 git submodule update
-sed -i "s/^parity-ui-precompiled = { git =.*/parity-ui-precompiled = { git = \"https:\/\/github.com\/ethcore\/js-precompiled.git\", optional = true, branch = \"$BRANCH\" }/" dapps/ui/Cargo.toml
+sed -i "/^parity-ui-precompiled/ { s/branch = \".*\"/branch = \"$BRANCH\"/g; }" dapps/ui/Cargo.toml
 cargo update -p parity-ui-precompiled
 # --precise "$PRECOMPILED_HASH"
 
--- a/js/scripts/release.sh
+++ b/js/scripts/release.sh
@@ -89,8 +89,7 @@ fi
 
 echo "*** Updating cargo parity-ui-precompiled#$PRECOMPILED_HASH"
 git submodule update
-cat dapps/ui/Cargo.toml | sed "s/^parity-ui-precompiled = { git =.*/parity-ui-precompiled = { git = \"https:\/\/github.com\/ethcore\/js-precompiled.git\", optional = true, branch = \"$BRANCH\" }/" >dapps/ui/Cargo.toml.update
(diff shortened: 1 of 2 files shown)


js/scripts/release.sh
```diff
@@ -89,11 +89,14 @@ fi
 
 echo "*** Updating cargo parity-ui-precompiled#$PRECOMPILED_HASH"
 git submodule update
+cat dapps/ui/Cargo.toml | sed "s/^parity-ui-precompiled = { git =.*/parity-ui-precompiled = { git = \"https:\/\/github.com\/ethcore\/js-precompiled.git\", optional = true, branch = \"$BRANCH\" }/" >dapps/ui/Cargo.toml.update
```
tomusdrw: `sed -i dapps/ui/Cargo.toml "s/^parity-ui-precompiled = { git =.*/parity-ui-precompiled = { git = \"https:\/\/github.com\/ethcore\/js-precompiled.git\", optional = true, branch = \"$BRANCH\" }/"`
Should suffice.

jacogr: Will update.

Actually reason for the split was that Mac sed is not 100% compatible, `-i` takes additional arguments. (And I wanted to test locally)

jacogr:
Ok, will try `--precice branch` locally. (Definately cleaner - should have done so already, @arkpar mentioned it as well)

ngotchac:
Why not simply:

```bash
sed -i '/^parity-ui-precompile/ { s/dev/master/g; }' dapps/ui/Cargo.toml
```

and have in `dapps/ui/Cargo.toml`:

```
parity-ui-precompiled = { git = "https://github.com/ethcore/js-precompiled.git", optional = true, branch = "master" }
```

jacogr:
Cannot do s/master/$BRANCH/ - on beta or stable it is not `master`. It may be when newly moved, but for e.g. stable it could be beta when just moved. (Actually this is why we had a bum 1.4.12 release)

We get this wrong all the time since we do it manually.

ngotchac:
So 

```
sed -i '/^parity-ui-precompile/ { s/branch = ".*"/branch = "dev"/g; }' dapps/ui/Cargo.toml
```

?

js/scripts/release.sh
```diff
@@ -89,11 +89,13 @@ fi
 
 echo "*** Updating cargo parity-ui-precompiled#$PRECOMPILED_HASH"
 git submodule update
+sed -i "/^parity-ui-precompiled/ { s/branch = \".*\"/branch = \"$BRANCH\"/g; }" dapps/ui/Cargo.toml
```
ngotchac: Just out of curiosity : why `"` instead of `'` (which would permit having `branch = "$BRANCH"`) ?

tomusdrw:
Can we also add a  warning comment in `dapps/ui/Cargo.toml` that it should always be a single line and that it's overwritten by build?

js/scripts/release.sh
jacogr: With ' the shell doesn't expand the variables, with " it does. (Hence the escape jiggle)

Issue #4833, opened 2017-03-09
Metropolis release


arkpar:
- [x] [EIP-208 (86): Account abstraction](https://github.com/ethereum/EIPs/pull/208) | #4697 
- [x] [EIP-98: Receipt state root removal](https://github.com/ethereum/EIPs/issues/98) |  #4296 
- [x] [EIP-210 (96): Blockhash and state root changes](https://github.com/ethereum/EIPs/pull/210) | #5505 
- [x] [EIP-100: Change difficulty adjustment to target mean block time including uncles](https://github.com/ethereum/EIPs/issues/100) | #5027 
- [x] [EIP-198 (101): Precompile for bigint modular exponentiation](https://github.com/ethereum/EIPs/pull/198) | #4926 
- [x] [EIP-206 (140): REVERT opcode](https://github.com/ethereum/EIPs/pull/206) | #5477 
- [x] [EIP-214 (116): STATIC_CALL opcode](https://github.com/ethereum/EIPs/pull/214) | #4851
- [x] [EIP-211 (5/8) RETURNDATACOPY and RETURNDATASIZE](https://github.com/ethereum/EIPs/pull/211) | #5678
- [x] [EIP-212 Precompiled contract for pairing check](https://github.com/ethereum/EIPs/pull/212) | #5307 
- [x] [EIP-213 Precompiled contracts for addition and scalar multiplication on the elliptic curve alt_bn128](https://github.com/ethereum/EIPs/pull/213) | https://github.com/paritytech/parity/pull/4999
- [x] [EIP-649 Difficulty bomb delay and issuance reduction](https://github.com/ethereum/EIPs/pull/669) | #5855
- [x] [EIP-658 Embedding transaction return data in receipts](https://github.com/ethereum/EIPs/pull/658) | #5855

--- 

For reference - Byzantium Meta: [EIP-609](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-609.md)

debris:
I will take  `EIP-116: STATIC_CALL opcode`

rphmeier:
I'll take `EIP-198 (101): Precompile for bigint modular exponentiation`

NikVolf:
Will take EIP-213 Precompiled contracts for addition and scalar multiplication on the elliptic curve alt_bn128

debris:
I will take 

`EIP-100: Change difficulty adjustment to target mean block time including uncles`
`EIP-206 (140): REVERT opcode`

[...]

PR #4697, opened 2017-02-27
EIP-86

arkpar:
Implements https://github.com/ethereum/EIPs/pull/208

PR #4851, opened 2017-03-10
EIP-116 (214), #4833

debris:
Everything should be working, but there are no tests. Opened a pr to test all existing tests && to get second pair of eyes on this code.

side-notes, executive / externalities types need to be decoupled from ipc types. I got the impression that some global imports are misused and it's quite annoying

### DIFF
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -55,7 +55,7 @@ pub struct TestBlockChainClient {
 	/// Blocks.
 	pub blocks: RwLock<HashMap<H256, Bytes>>,
 	/// Mapping of numbers to hashes.
- 	pub numbers: RwLock<HashMap<usize, H256>>,
+	pub numbers: RwLock<HashMap<usize, H256>>,
 	/// Genesis block hash.
 	pub genesis_hash: H256,
 	/// Last block hash.
@@ -353,7 +353,7 @@ pub fn get_temp_state_db() -> GuardedTempResult<StateDB> {
 
 impl MiningBlockChainClient for TestBlockChainClient {
 	fn latest_schedule(&self) -> Schedule {
-		Schedule::new_post_eip150(24576, true, true, true, true)
+		Schedule::new_post_eip150(24576, true, true, true)
 	}
 
 	fn prepare_open_block(&self, author: Address, gas_range_target: (U256, U256), extra_data: Bytes) -> OpenBlock {
--- a/ethcore/src/evm/ext.rs
+++ b/ethcore/src/evm/ext.rs
@@ -53,25 +53,24 @@ pub enum CreateContractAddress {
 }
 
 /// Externalities interface for EVMs
-// TODO: [rob] associated error type instead of `trie::Result`. Not all EVMs are trie powered.
 pub trait Ext {
 	/// Returns a value for given key.
-	fn storage_at(&self, key: &H256) -> trie::Result<H256>;
+	fn storage_at(&self, key: &H256) -> evm::Result<H256>;
 
 	/// Stores a value for given key.
-	fn set_storage(&mut self, key: H256, value: H256) -> trie::Result<()>;
+	fn set_storage(&mut self, key: H256, value: H256) -> evm::Result<()>;
 
 	/// Determine whether an account exists.
-	fn exists(&self, address: &Address) -> trie::Result<bool>;
+	fn exists(&self, address: &Address) -> evm::Result<bool>;
 
 	/// Determine whether an account exists and is not null (zero balance/nonce, no code).
-	fn exists_and_not_null(&self, address: &Address) -> trie::Result<bool>;
+	fn exists_and_not_null(&self, address: &Address) -> evm::Result<bool>;
 
 	/// Balance of the origin account.
-	fn origin_balance(&self) -> trie::Result<U256>;
+	fn origin_balance(&self) -> evm::Result<U256>;
 
 	/// Returns address balance.
-	fn balance(&self, address: &Address) -> trie::Result<U256>;
+	fn balance(&self, address: &Address) -> evm::Result<U256>;
 
 	/// Returns the hash of one of the 256 most recent complete blocks.
 	fn blockhash(&mut self, number: &U256) -> H256;
@@ -99,13 +98,13 @@ pub trait Ext {
 	) -> MessageCallResult;
 
 	/// Returns code at given address
-	fn extcode(&self, address: &Address) -> trie::Result<Arc<Bytes>>;
+	fn extcode(&self, address: &Address) -> evm::Result<Arc<Bytes>>;
 
 	/// Returns code size at given address
-	fn extcodesize(&self, address: &Address) -> trie::Result<usize>;
+	fn extcodesize(&self, address: &Address) -> evm::Result<usize>;
 
 	/// Creates log entry with given topics and data
-	fn log(&mut self, topics: Vec<H256>, data: &[u8]);
+	fn log(&mut self, topics: Vec<H256>, data: &[u8]) -> evm::Result<()>;
 
 	/// Should be called when transaction calls `RETURN` opcode.
 	/// Returns gas_left if cost of returning the data is not too high.
@@ -113,7 +112,7 @@ pub trait Ext {
 
 	/// Should be called when contract commits suicide.
 	/// Address to which funds should be refunded.
-	fn suicide(&mut self, refund_address: &Address) -> trie::Result<()> ;
+	fn suicide(&mut self, refund_address: &Address) -> evm::Result<()> ;
 
 	/// Returns schedule.
 	fn schedule(&self) -> &Schedule;
--- a/ethcore/src/evm/interpreter/mod.rs
+++ b/ethcore/src/evm/interpreter/mod.rs
@@ -199,6 +199,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 
 		if (instruction == instructions::DELEGATECALL && !schedule.have_delegate_call) ||
 			(instruction == instructions::CREATE2 && !schedule.have_create2) ||
+			(instruction == instructions::STATICCALL && !schedule.have_static_call) ||
 			(instruction == instructions::REVERT && !schedule.have_revert) {
 
 			return Err(evm::Error::BadInstruction {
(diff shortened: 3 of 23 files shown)


ethcore/src/client/test_client.rs
```diff
@@ -352,7 +352,7 @@ pub fn get_temp_state_db() -> GuardedTempResult<StateDB> {
 
 impl MiningBlockChainClient for TestBlockChainClient {
 	fn latest_schedule(&self) -> Schedule {
-		Schedule::new_post_eip150(24576, true, true, true)
```
tomusdrw: IMHO this is reaching a point where a struct would look much better than the 4 boolean flags.
```rust
Schedule::new_post_eip150(PostEip150 {
   max_code_size: 24576,
   fix_exp: true,
   no_empty: true,
   kill_empty: true,
   static_call: true,
});
```

ethcore/src/evm/ext.rs
```diff
@@ -94,15 +94,15 @@ pub trait Ext {
 	fn extcodesize(&self, address: &Address) -> trie::Result<usize>;
```
tomusdrw: Shouldn't all functions here return `evm::Result` instead of `trie::Result`?

ethcore/src/evm/interpreter/mod.rs
```diff
@@ -292,14 +292,21 @@ impl<Cost: CostType> Interpreter<Cost> {
...
-			instructions::CALL | instructions::CALLCODE | instructions::DELEGATECALL => {
+			instructions::CALL | instructions::CALLCODE | instructions::DELEGATECALL | instructions::STATICCALL => {
 				assert!(ext.schedule().call_value_transfer_gas > ext.schedule().call_stipend, "overflow possible");
+				if instruction == instructions::STATICCALL && !ext.schedule().static_call {
+					// STATICCALL is not enabled, yet called
+					return Err(evm::Error::BadInstruction {
+						instruction: instruction
+					});
+				}
+
```
debris: after looking at Arkadiy eip86, I guess this should be under `verify_instruction`

ethcore/src/executive.rs
```diff
@@ -253,13 +259,17 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 		// backup used in case of running out of gas
 		self.state.checkpoint();
 
+		trace!("Executive::call(params={:?}) self.env_info={:?}", params, self.info);
+		if (params.call_type == CallType::StaticCall || self.static_flag) && params.value.value() > 0.into() {
+			return Err(evm::Error::MutableCallInStaticContext);
```
tomusdrw: Two issues:
1. Shouldn't such calls be traced anyway?
2. Seems that `state.checkpoint` is being made, but since we are not `enacting_result` it will not be reverted, is that correct?

ethcore/src/executive.rs
```diff
@@ -359,6 +369,10 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 		// backup used in case of running out of gas
 		self.state.checkpoint();
 
+		if params.call_type == CallType::StaticCall || self.static_flag {
+			return Err(evm::Error::MutableCallInStaticContext);
```
tomusdrw: Same issues with tracing and `state.checkpoint()`

ethcore/src/evm/interpreter/mod.rs
```diff
@@ -292,14 +292,21 @@ impl<Cost: CostType> Interpreter<Cost> {
 					}
 				};
 			},
-			instructions::CALL | instructions::CALLCODE | instructions::DELEGATECALL => {
+			instructions::CALL | instructions::CALLCODE | instructions::DELEGATECALL | instructions::STATICCALL => {
```
tomusdrw: It seems that gas calculation for `staticcall` is not implemented (I suppose it should be the same as for `CALL`)

ethcore/src/evm/ext.rs
debris: I agree, that visually it looks better 👍

ethcore/src/evm/interpreter/mod.rs
debris: thanks for pointing that out!

debris: imo, it actually should be the same as `DELEGATECALL`, since we are not creating new accounts, nor transfering value

ethcore/src/client/test_client.rs
debris: I will change that after #4697 is merged, cause involves the same changes

arkpar: #4757 also changes this. https://github.com/ethcore/parity/pull/4757/files#diff-ef2687136b4e032043a492e9e53dcb7cR177

ethcore/src/evm/ext.rs
rphmeier: Really it should be an associated type `Error`. Externalities are lower level than the machine itself, so it doesn't make that much sense to return an `evm::Result` there.

ethcore/res/ethereum/eip150_test.json
```diff
@@ -23,6 +23,7 @@
 		"maximumExtraDataSize": "0x20",
 		"minGasLimit": "0x1388",
 		"networkID" : "0x1",
+		"eip116Transition": "0x7fffffffffffff",
```
gavofyork: stick to `0x7fffffffffffffff` for consistency

ethcore/res/ethereum/eip161_test.json
```diff
@@ -23,6 +23,7 @@
 		"maximumExtraDataSize": "0x20",
 		"minGasLimit": "0x1388",
 		"networkID" : "0x1",
+		"eip116Transition": "0x7fffffffffffff",
```
gavofyork: as above

ethcore/src/evm/interpreter/mod.rs
```diff
@@ -292,14 +298,15 @@ impl<Cost: CostType> Interpreter<Cost> {
...
 				assert!(ext.schedule().call_value_transfer_gas > ext.schedule().call_stipend, "overflow possible");
+
 				stack.pop_back();
 				let call_gas = provided.expect("`provided` comes through Self::exec from `Gasometer::get_gas_cost_mem`; `gas_gas_mem_cost` guarantees `Some` when instruction is `CALL`/`CALLCODE`/`DELEGATECALL`/`CREATE`; this is one of `CALL`/`CALLCODE`/`DELEGATECALL`; qed");
 				let code_address = stack.pop_back();
 				let code_address = u256_to_address(&code_address);
 
-				let value = if instruction == instructions::DELEGATECALL {
+				let value = if instruction == instructions::DELEGATECALL || instruction ==  instructions::STATICCALL {
```
gavofyork: double ` ` after `==`

ethcore/src/executive.rs
```diff
@@ -250,6 +256,13 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 		tracer: &mut T,
 		vm_tracer: &mut V
 	) -> evm::Result<U256> where T: Tracer, V: VMTracer {
+		trace!("Executive::call(params={:?}) self.env_info={:?}", params, self.info);
+		if (params.call_type == CallType::StaticCall || self.static_flag) && params.value.value() > 0.into() {
```
gavofyork: `params` are for this call, no? in which case, if this call type is `StaticCall` (or `DelegateCall`, for that matter) then it should be fine even if it is a static context (`static_flag == true`)

gavofyork:
@tomusdrw we'll have to see how the yellow paper prefers it. i think it'll end up being easier to consider `STATIC_CALL` as "touching" accounts, even if it's not allowing direct state changes.

debris:
@gavofyork touching accounts has been already clarified -> https://github.com/ethereum/EIPs/pull/214#issuecomment-288697580

> my Yellow Paper PR would also kill dead accounts that are touched during a STATICCALL ethereum/yellowpaper#270

ethcore/src/evm/schedule.rs
```diff
@@ -163,13 +166,36 @@ impl Schedule {
...
+		let mut schedule = Schedule::new_post_eip150(usize::max_value(), true, true, true);
+		schedule.apply_params(block_number, params);
+		schedule
+ 	}
+
+	/// Apply common spec config parameters to the schedule.
+ 	pub fn apply_params(&mut self, block_number: u64, params: &CommonParams) {
+		self.have_create2 = block_number >= params.eip86_transition;
+		self.have_revert = block_number >= params.eip140_transition;
+		self.have_static_call = block_number >= params.eip214_transition;
```
rphmeier: nice, this is better than `have_metropolis_instructions`

ethcore/src/externalities.rs
```diff
@@ -99,33 +101,40 @@ impl<'a, T: 'a, V: 'a, B: 'a, E: 'a> Externalities<'a, T, V, B, E>
...
-		self.state.storage_at(&self.origin_info.address, key)
+	fn storage_at(&self, key: &H256) -> evm::Result<H256> {
+		self.state.storage_at(&self.origin_info.address, key).map_err(Into::into)
 	}
 
-	fn set_storage(&mut self, key: H256, value: H256) -> trie::Result<()> {
-		self.state.set_storage(&self.origin_info.address, key, value)
+	fn set_storage(&mut self, key: H256, value: H256) -> evm::Result<()> {
+		if self.static_flag {
+			Err(evm::Error::MutableCallInStaticContext)
```
rphmeier: it would be interesting if the static mode just had immutable references so mutable calls could literally not be made -- but this works too.

rphmeier:
LGTM, but tests would be nice.

ethcore/src/evm/schedule.rs
```diff
@@ -163,13 +166,36 @@ impl Schedule {
...
+		if block_number >= params.eip210_transition {
+			self.blockhash_gas = 350;
```
pirapira: I think the new gas cost is 800.  https://github.com/ethereum/EIPs/pull/210/files#diff-e02a92c2fb96c1a1bfb05e4c6e2ef5daR42

arkpar: thanks, added to #5855

PR #4857, opened 2017-03-10
Allow v1 get & v2 getData for registry

jacogr:
Currently the fetch data for Kovan paritynews is non-operational.


jacogr:
Closed in favour of cherry picking from dappreg overhaul.

PR #4859, opened 2017-03-10
Chain-selection from UI

gavofyork:


arkpar:
Looks good for now. However hypervisor approach will make it hard to implement changing settings such as chain, ports, etc. in [parity-as-a library](https://github.com/ethcore/parity/issues/4842) setting. We'd probably have to reimplement this so it does not require process shutdown eventually. This would require careful accounting for any shutdown leaks. AFAIK some of the services currently (IPC/RPC/stratum/whatever) don't cleanup 100% after themselves and may leave a thread hanging or some memory stuck in a pointer loop. This will require quite a bit of work and maybe automated tests that would check that there are no memory/thread/FD leaks.

js/src/views/Settings/Parity/parity.spec.js
```diff
@@ -94,5 +100,26 @@ describe('views/Settings/Parity', () => {
         expect(instance.store.changeMode).to.have.been.calledWith('dark');
       });
     });
+    describe('chain selector', () => {
```
jacogr: Blank line before.

js/src/views/Settings/Parity/store.spec.js
```diff
@@ -33,23 +33,35 @@ describe('views/Settings/Parity/Store', () => {
...
+  });
+
   describe('@action', () => {
     describe('setMode', () => {
       it('sets the mode', () => {
         store.setMode('offline');
         expect(store.mode).to.equal('offline');
       });
     });
+    describe('setChain', () => {
```
jacogr: Blank line before.

jacogr:
2 tiny comments, otherwise code looks good.

ethcore/src/client/client.rs
```diff
@@ -276,6 +278,25 @@ impl Client {
...
+
+	/// Hypervisor should restart this client.
+	pub fn restart(&self) {
+		if let Some(ref h) = *self.exit_handler.lock() {
+			(*h)(true, None);
+		}
+	}
+
+	/// Hypervisor should kill this client.
+	pub fn exit(&self) {
```
arkpar: Where is this called from? Why not just implement `Drop`?

parity/main.rs
```diff
@@ -250,7 +266,12 @@ fn main_direct(can_restart: bool) -> i32 {
 		match start(can_restart) {
 			Ok(result) => match result {
 				PostExecutionAction::Print(s) => { println!("{}", s); 0 },
-				PostExecutionAction::Restart => PLEASE_RESTART_EXIT_CODE,
+				PostExecutionAction::Restart(spec_name_override) => {
+					if let Some(spec_name) = spec_name_override {
+						set_spec_name_override(spec_name);
```
arkpar: What about all other CLI options? Are they not preserved?

parity/main.rs
```diff
@@ -208,6 +208,22 @@ fn latest_exe_path() -> Option<PathBuf> {
 		.and_then(|mut f| { let mut exe = String::new(); f.read_to_string(&mut exe).ok().map(|_| updates_path(&exe)) })
 }
 
+fn set_spec_name_override(spec_name: String) {
+	if let Err(e) = File::create(updates_path("spec_name_overide"))
```
arkpar: Why is there a need for a file at all? Can't the hypervisor just pass CLI params to a new process?

tomusdrw: Why not introduce `--chain=auto` as default setting and just update `user_defaults` like we do with `pruning`, `tracing` and `fat_db`?

gavofyork: > Why is there a need for a file at all? Can't the hypervisor just pass CLI params to a new process?

two reasons:

1. getting data out a child process back to the parent cleanly is non-trivial. i used a return-code thus far but it's no good for information-rich content. i guess we could use std-out but then it wouldn't work properly in all circumstances.

2. this method works well for existing auto-update installations; using the hypervisor to relay information (even if it were easy) would mean everyone manually re-installing before this feature worked, since the hypervisor process is always the most recent _manually_ installed executable. 

> Why not introduce --chain=auto as default setting and just update user_defaults like we do with pruning, tracing and fat_db?

(I actually implemented it this way originally.) `user_defaults` is per-chain, which obv. won't work for the setting of which chain to use. we could use another non-per-chain user-defaults, but then that's a lot of extra baggage for what is really just temporary signalling.

parity/main.rs
gavofyork: yes, but that happens as part of the standard hypervisor process.

ethcore/src/client/client.rs
gavofyork: `exit` and `restart` are not yet called from anywhere, however, they seem like useful public api to have; i plan on implementing a UI endpoints for them to all the user to kill or restart parity easily.

rphmeier: I don't see how it belongs in `Client`, whose role is just to manage and provide blockchain data. Hypervisor stuff being at a higher level would make it a lot simpler to manage different kinds of clients, parity-as-a-library, non-hypervised instances, etc.

gavofyork: will remove for the present PR.

gavofyork: i guess we might want to look into an alternative "hypervisor" IPC endpoint that could host such calls, however in the context of the present PR that would be too much extra baggage.

PR #4862, opened 2017-03-10
Attach external accounts via Qr

jacogr:
- Allow the creation of external accounts (via normal create account)
- Scan account QR via camera, create entry, mark as external
- Display account in accounts list
- Allow editing inside account view (meta + forget)

Not included here -

- transacting from external accounts


gavofyork:
doesn't seem to work, at least not with QR codes generated from marek's app.

jacogr:
Only tested with QR codes generated from other iOS wallets (works for those), still trying to get the app actually compiling on my side and available on my device. So discrepancies could be quite possible :(

jacogr:
Ok, figured it out, the QR from our app is without the leading 0x for the address so I don't see it as a valid address.

I'm going to put in a fix for this (append the 0x when not found), however I but would prefer is we stick to having the `0x<hex>` format. It also aligns with what other wallets are doing and makes interoperability possible, i.e. show me your QR and I'll send to it from my iOS wallet.

cc @debris @gavofyork

jacogr:
Closing in favour of https://github.com/paritytech/parity/pull/4881 (which does contain these changes and didn't turn into a monster as initially feared). Can be reviewed as one.

PR #4864, opened 2017-03-10
Reload UI on network switch

### LINKED ISSUE #4814: Reload UI after software update & chain switches
Currently we only prompt for chain switches (reload stop-gap until we have proper refresh in place), for updates the reload is 100% needed, software changed.

ngotchac:
Fixes #4814

### DIFF
--- a/js/src/redux/providers/chainMiddleware.js
+++ b/js/src/redux/providers/chainMiddleware.js
@@ -18,10 +18,6 @@ import BalancesProvider from './balances';
 import { showSnackbar } from './snackbarActions';
 import { DEFAULT_NETCHAIN } from './statusReducer';
 
-// Reload the page on network switch after
-// 5 seconds
-const RELOAD_TIMEOUT = 5000;
-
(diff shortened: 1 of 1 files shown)


jacogr:
Better than what we have. We however need to actually just reload the affected accounts - refresh accounts, balances, contracts, etc. (Clear cache, start over.) For network itself we really don't need to reload the UI.

js/src/redux/providers/chainMiddleware.js
```diff
@@ -18,6 +18,10 @@ import BalancesProvider from './balances';
 import { showSnackbar } from './snackbarActions';
 import { DEFAULT_NETCHAIN } from './statusReducer';
 
+// Reload the page on network switch after
+// 5 seconds
+const RELOAD_TIMEOUT = 5000;
```
gavofyork: any reason this is not instant? just looks to me like 5 of incorrectness.

jacogr:
Tested.

PR #4867, opened 2017-03-11
Fix auto-updater

gavofyork:


arkpar:
Looks good, although relying on the FS attributes seems a bit flaky. Version data could go into an additional file or the binary file name.

PR #4873, opened 2017-03-11
rlp serialization refactor

debris:
for now, it's only serialization refactor. it's still WIP, but wanted to know, if you like interface changes I made:

changes:

- removed `BytesEncodable` trait
- removed `Encodable` trait
- serialization do not use `ToBytes`
- `Encodable` is no longer implemented to `T where T: ToBytes`
- `RlpStream` does not implement `Stream` trait. All functions where moved to public interface. I find additional `Stream` import to be really annoying
- used `byteorder` for serialization
- new api method `append_list` which should be used to append list of encodable items
- new api method `append_internal` which should be used for wrapper types to encoding item without incrementing list elements counter
- serialization of lists preallocates 1 additional byte for payload. This allows us avoid some unnecessary shifts for short lists

benchmarks:

- **old**

```
test bench_stream_1000_empty_lists   ... bench:       9,839 ns/iter (+/- 1,207)
test bench_stream_nested_empty_lists ... bench:         223 ns/iter (+/- 23)
test bench_stream_u256_value         ... bench:         453 ns/iter (+/- 52)
test bench_stream_u64_value          ... bench:         119 ns/iter (+/- 16)
```

- **new (update)**

```
test bench_stream_1000_empty_lists   ... bench:       9,143 ns/iter (+/- 1,209)
test bench_stream_nested_empty_lists ... bench:         181 ns/iter (+/- 22)
test bench_stream_u256_value         ... bench:         344 ns/iter (+/- 53)
test bench_stream_u64_value          ... bench:          90 ns/iter (+/- 10)
```

so encoding integers is **almost 25% faster**


util/rlp/src/stream.rs
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	pub fn new_list(len: usize) -> Self {
+		let mut stream = RlpStream::new();
+		stream.begin_list(len);
+		stream
+	}
+
+	pub fn append<'a, E>(&'a mut self, value: &E) -> &'a mut Self where E: RlpEncodable {
```
rphmeier: One nice change we could make would be to `impl` on `[u8]` instead of `&'a [u8]` (likewise for str), and then do `E: RlpEncodable + ?Sized` so we don't have to double-reference slices (which currently leads to lots of `&&*my_vec` in the codebase)

debris: It will be included in separate pr, this one is big enough ;)

ethcore/src/engines/tendermint/message.rs
```diff
@@ -162,7 +162,8 @@ impl Decodable for Step {
 
 impl Encodable for Step {
 	fn rlp_append(&self, s: &mut RlpStream) {
-		RlpEncodable::rlp_append(&self.number(), s);
+		//s.append(&self.number());
```
keorn: Old comment.

ethcore/src/engines/tendermint/message.rs
```diff
@@ -278,6 +279,8 @@ mod tests {
 			::rlp::encode(&H520::default()).to_vec(),
 			Vec::new()
 		];
+
+		println!("seal: {:?}", seal);
```
keorn: unnecessary print

util/rlp/src/stream.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+/// Appendable rlp encoder.
+pub struct RlpStream {
+	unfinished_lists: ElasticArray16<ListInfo>,
+	buffer: ElasticArray1024<u8>,
+	finished_list: bool,
+}
+
+impl Default  for RlpStream {
```
rphmeier: double space after Default

ethcore/light/src/net/status.rs
```diff
@@ -16,7 +16,7 @@
 
 //! Peer status and capabilities.
 
-use rlp::{DecoderError, RlpDecodable, RlpEncodable, RlpStream, Stream, UntrustedRlp, View};
+use rlp::{DecoderError, RlpDecodable, Encodable, RlpStream, UntrustedRlp, View};
```
arkpar: Why was `RlpEncodable` renamed, but not the rest?

debris: `RlpDecodable` was renamed accordingly in the next pr. The rest of data structures names remained unchanged. I didn't want to introduce to many changes in a single pr

util/rlp/src/impls.rs
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+impl Encodable for Vec<u8> {
+	fn rlp_append(&self, s: &mut RlpStream) {
+		s.encoder().encode_value(self);
+	}
+}
+
+impl<T> Encodable for Option<T> where T: Encodable {
```
arkpar: This is dangerous and should be removed IMO. Why does `Option` has to be a list?

debris: It's not a question to me. It's already like that on master,

https://github.com/ethcore/parity/blob/8a67a0a80aae506233f65cc41962fe483fc10c15/util/rlp/src/rlpstream.rs#L340-L350

arkpar: I know. Might be a good time to get rid of it.

debris: I will file an issue and fix in the next pr

arkpar:
lgtm

util/rlp/src/impls.rs
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+macro_rules! impl_encodable_for_u {
+	($name: ident, $func: ident, $size: expr) => {
+		impl Encodable for $name {
+			fn rlp_append(&self, s: &mut RlpStream) {
+				let leading_empty_bytes = self.leading_zeros() as usize / 8;
+				let mut buffer = [0u8; $size];
+				(&mut buffer as &mut [u8]).$func::<BigEndian>(*self).expect("buffer.len() == sizeof(*self); qed");
```
NikVolf: can be just
`BigEndian::$func(&mut buffer, *self);`

util/rlp/src/stream.rs
```diff
@@ -0,0 +1,335 @@
...
+			buffer: &mut stream.buffer
+		}
+	}
+
+	fn insert_size(&mut self, size: usize, position: usize) -> u8 {
+		let size = size as u32;
+		let leading_empty_bytes = size.leading_zeros() as usize / 8;
+		let size_bytes = 4 - leading_empty_bytes as u8;
+		let mut buffer = [0u8; 4];
+		(&mut buffer as &mut [u8]).write_u32::<BigEndian>(size).expect("buffer.len() == sizeof(value); qed");
```
NikVolf: can be just
`BigEndian::write_u32(&mut buffer)`

util/rlp/src/impls.rs
gavofyork: how else would you encode an `Option` in RLP?

gavofyork:
reduction of code size and number of traits is good.

addition of `_list` everywhere less so.

NikVolf:
lgtm

gavofyork:
is it the case that the only underlying reason for the `append`/`append_list` API divergence is the need to distinguish between `Vec<u8>` as a single piece of data and as a list of numbers?

debris:
yes

gavofyork:
i seem to remember there was something we could do with traits, no? Have a `Listable` trait which is good for all types except `u8`?

util/rlp/src/impls.rs
arkpar: Manually. My point is that changing e.g. a struct member type `T` to `Option<T>` silently changes encoding of that member to a list. I recall fixing a bug caused by this.

debris:
> i seem to remember there was something we could do with traits, no? Have a Listable trait which is good for all types except u8?

Nothing less complicated than what we had before (with at least 3 different traits) comes to my mind. And I think the solution from this pr is less complicated and error prone

gavofyork:
well the fact i now have two different functions to call in order to do the same conceptual thing, depending on some arbitrary distinction between datatypes, adds to the complexity.

that said, i won't block it if y'all prefer this way of doing it.

debris:
thanks, @arkpar @NikVolf @rphmeier @tomusdrw @keorn ?

PR #4907, opened 2017-03-14
Detect rust compiler version in Parity build script, closes 4742

debris:


build.rs
```diff
@@ -0,0 +1,25 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+extern crate rustc_version;
+
+const MIN_RUSTC_VERSION: &'static str = "1.15.1";
+
+fn main() {
+	let is = rustc_version::version().unwrap();
+	let required = MIN_RUSTC_VERSION.parse().unwrap();
+	assert!(is >= required, format!("Detected incompatible compiler version: {}, you need at least {} to compile parity", is, required));
```
tomusdrw: Maybe suggest running `rustup update stable`?

debris: not everyone uses `rustup`. Also, it won't help in cases when someone has local or global rustup/multirust overrides

debris: I would prefer to keep the message brief

tomusdrw: But 90% of those users (that are probably 5% of total users) that are using those features will know how to update the compiler anyway even if the message doesn't apply to them.

I'm not pushing but IMHO it would be better to have even bigger multiline message that will be easy to spot even when mixed with stack trace (and that's probably what you'll get right know).
Have a look at this: https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build.rs#L98 it's a bunch of quite informative messages helping you to solve the problem imho.

gavofyork: agree with tomek here; a bigger multiline message might well reduce our support burden, which is currently rather large.

arkpar:
The message only shows when parity binary begins building, right?
If one of the dependencies requires a new version of the compiler it will still fail with a regular error message. And this is almost always the case since 99% of our codebase is in dependencies.

NikVolf:
@arkpar 
Yeah, i think issue can be properly addressed only on cargo level, there is one proposal there
https://github.com/rust-lang/cargo/issues/2751

build.rs
```diff
@@ -0,0 +1,35 @@
...
+
+const MIN_RUSTC_VERSION: &'static str = "1.15.1";
+
+fn main() {
+	let is = rustc_version::version().unwrap();
+	let required = MIN_RUSTC_VERSION.parse().unwrap();
+	assert!(is >= required, format!("
+
+It looks like you are compiling Parity with an old rustc compiler {}.
+Parity requires newer version {}. Please update you compiler.
```
rphmeier: "Parity requires version {}" (adding "newer" makes it read strangely)
you -> your

debris:
@arkpar parity's `build.rs` is executed in parallel at the same time as the other libraries are being built. We can't guarantee that our error message will be thrown before compiler fails with a different error, but there is nothing better we can do now

PR #4916, opened 2017-03-15
Add Token image from URL

ngotchac:
No need to go the the GHH dapp for setting a Token icon: hash the given URL content and post the GHH transaction for registering it directly from the Token Reg dapp


gavofyork:
(generic component for this would be nice ;)

ngotchac:
100% agreed ! However we don't share component betweens dapps yet.

Issue #4885, opened 2017-03-13
Non-ETH token icons don't show up (correctly) in default account selection (signer bar) when using chrome extension


bjornwgnr:
![icon](https://cloud.githubusercontent.com/assets/17831185/23864780/e3e68376-0813-11e7-81fe-0fe2c3344de1.jpg)

They show up when using the parity browser though.

ngotchac:
This should be fixed from https://github.com/ethcore/parity-extension/pull/66

PR #4922, opened 2017-03-15
Add ability to configure Secure API

ngotchac:
Useful for #4885 ( see https://github.com/ethcore/parity-extension/pull/66 )


jacogr:
Probably want to backport this as well since if we have the check in the extension, we just want to minimise the number of "configuring it via private" out there. Non-invasive as well.

PR #4926, opened 2017-03-15
EIP198 and built-in activation

rphmeier:
Implements [EIP198](https://github.com/ethereum/EIPs/pull/198) with exponentiation by squaring.

Changes to built-in declaration: optional `activate_at` field, defaults to 0. Built-ins won't be available  from the `Engine` until activated. Replaces the `builtin_cost`, `execute_builtin`, and `is_builtin` of engine with a single `builtin(&Address, block_number) -> Option<&Builtin>`, reducing the amount of lookups done.

Could use a few more test cases, although I expect the consensus tests will cover this in detail.

`num::BigUint` is known to be much slower than GMP. `ramp` is a faster pure-Rust alternative, but requires nightly. If this calculation starts to become a bottleneck, we can move to GMP.

### DIFF
--- a/ethcore/res/ethereum/foundation.json
+++ b/ethcore/res/ethereum/foundation.json
@@ -189,6 +189,7 @@
 		"0000000000000000000000000000000000000002": { "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
 		"0000000000000000000000000000000000000003": { "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
 		"0000000000000000000000000000000000000004": { "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
+		"0000000000000000000000000000000000000005": { "builtin": { "name": "modexp", "activate_at": "0x7fffffffffffff", "pricing": { "modexp": { "divisor": 20 } } } },
 		"3282791d6fd713f1e94f4bfd565eaa78b3a0599d": {
 			"balance": "1337000000000000000000"
 		},
--- a/ethcore/src/builtin.rs
+++ b/ethcore/src/builtin.rs
@@ -71,7 +71,7 @@ impl Pricer for Modexp {
 		let mod_len = read_len();
 
 		// floor(max(length_of_MODULUS, length_of_BASE) ** 2 * max(length_of_EXPONENT, 1) / GQUADDIVISOR)
-		// TODO: is saturating the best bahavior here?
+		// TODO: is saturating the best behavior here?
 		let m = max(mod_len, base_len);
 		match m.overflowing_mul(m) {
 			(_, true) => U256::max_value(),
@@ -260,10 +260,18 @@ impl Impl for ModexpImpl {
 
 		// calculate modexp: exponentiation by squaring.
 		fn modexp(mut base: BigUint, mut exp: BigUint, modulus: BigUint) -> BigUint {
-			if base == BigUint::zero() || modulus <= BigUint::one() { return BigUint::zero() }
+			match (base == BigUint::zero(), exp == BigUint::zero()) {
+				(_, true) => return BigUint::one(), // n^0 % m
+				(true, false) => return BigUint::zero(), // 0^n % m, n>0
+				(false, false) if modulus <= BigUint::one() => return BigUint::zero(), // a^b % 1 = 0.
+				_ => {}
+			}
 
 			let mut result = BigUint::one();
 			base = base % &modulus;
+
+			// fast path for base divisible by modulus.
+			if base == BigUint::zero() { return result }
 			while exp != BigUint::zero() {
 				// exp has to be on the right here to avoid move.
 				if BigUint::one() & &exp == BigUint::one() {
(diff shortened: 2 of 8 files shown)


ethcore/res/ethereum/foundation.json
```diff
@@ -189,6 +189,7 @@
 		"0000000000000000000000000000000000000002": { "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
 		"0000000000000000000000000000000000000003": { "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
 		"0000000000000000000000000000000000000004": { "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
+		"0000000000000000000000000000000000000005": { "builtin": { "name": "modexp", "activate_at": "0x7fffffffffffff", "pricing": { "modexp": { "divisor": 20 } } } },
```
rphmeier: Unfortunately, the pricing scheme for modexp isn't a simple function of the size like the others, so we have to create a special pricer which can't really be reused.

ethcore/src/builtin.rs
```diff
@@ -166,9 +230,63 @@ impl Impl for Ripemd160 {
...
 }
 
+impl Impl for ModexpImpl {
+	fn execute(&self, input: &[u8], output: &mut BytesRef) {
+		let mut reader = input.chain(io::repeat(0));
+		let mut buf = [0; 32];
+
+		// read lengths as usize.
+		// ignoring the first 24 bytes might technically lead us to fall out of consensus,
+		// but so would running out of addressable memory!
```
tomusdrw: Why impose such artificial limitation? Is it part of the spec?

ethcore/src/builtin.rs
```diff
@@ -166,9 +230,63 @@ impl Impl for Ripemd160 {
...
+			BigUint::from_bytes_be(&buf[..len])
+		};
+
+		let base = read_num(base_len);
+		let exp = read_num(exp_len);
+		let modulus = read_num(mod_len);
+
+		// calculate modexp: exponentiation by squaring.
+		fn modexp(mut base: BigUint, mut exp: BigUint, modulus: BigUint) -> BigUint {
+			if base == BigUint::zero() || modulus <= BigUint::one() { return BigUint::zero() }
```
tomusdrw: `0**0` will return `0`, there is a comment in original EIP suggesting that it should be `1` for consistency with existing evm `EXP` opcode.

ethcore/src/builtin.rs
```diff
@@ -166,9 +230,63 @@ impl Impl for Ripemd160 {
...
+
+			let mut result = BigUint::one();
+			base = base % &modulus;
```
tomusdrw: This may lead to `base == 0` and we sould short circuit this case as well.

ethcore/src/builtin.rs
rphmeier: Yeah, the spec favors 32 bytes over 8 to encode the lengths because that's the native EVM word size and would require much less fiddling with bits to pass arguments, although really nobody will be able to pass lengths anywhere near that size.

I'd argue that this is a physical rather than an artificial limitation: I don't think a machine will exist for a long time which can handle 2^256-bit numbers.

ethcore/src/executive.rs
```diff
@@ -261,17 +261,23 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
 
-		if self.engine.is_builtin(&params.code_address) {
-			// if destination is builtin, try to execute it
+		// if destination is builtin, try to execute it
+		if let Some(builtin) = self.engine.builtin(&params.code_address, self.info.number) {
+			// Engines aren't supposed to return builtins until activation, but
+			// prefer to fail rather than silently break consensus.
+			if !builtin.is_active(self.info.number) {
+				let msg = format!("Engine returned unactivated built-in at {}", params.code_address);
+				return Err(evm::Error::Internal(msg));
```
rphmeier: Nice catch. It should be done with a `Drop` guard to prevent mistakes like that.

rphmeier: Although, looking closer, this branch already indicates a consensus failure of some kind.

ethcore/src/builtin.rs
tomusdrw: Sure, we can't handle such memory now as well that's why I see no point in this limit at all. Why 8 and not 7?

rphmeier: two reasons: it makes decoding the size much easier with the byteorder library, and it's an amount of memory actually addressable on 64-bit machines.

tomusdrw: Right, sorry. Makes sense!

rphmeier:
@tomusdrw Grumbles addressed; I changed the detected consensus failure in executive (which should never happen) to panic instead of issue an error which would be silently swallowed.

ethcore/src/builtin.rs
```diff
@@ -260,10 +260,18 @@ impl Impl for ModexpImpl {
...
-			if base == BigUint::zero() || modulus <= BigUint::one() { return BigUint::zero() }
+			match (base == BigUint::zero(), exp == BigUint::zero()) {
+				(_, true) => return BigUint::one(), // n^0 % m
+				(true, false) => return BigUint::zero(), // 0^n % m, n>0
+				(false, false) if modulus <= BigUint::one() => return BigUint::zero(), // a^b % 1 = 0.
+				_ => {}
+			}
 
 			let mut result = BigUint::one();
 			base = base % &modulus;
```
tomusdrw: You could do this line at the very beginning.

rphmeier: At the very least you have to check if the modulus is zero beforehand or the code will panic.

gavofyork:
looks reasonable in design. haven't reviewed algo yet, but tests should cover that.

debris:
merged with master to trigger the CI build

PR #4930, opened 2017-03-16
Ensure sealing work enabled in miner once subscribers added

NikVolf:


gavofyork:
why are both conditions needed?

NikVolf:
well afaik, first one starts sealing, and the second one doesn't allow it to turn off after some blocks

debris:
merged with master to trigger the CI build

PR #4931, opened 2017-03-16
Simplify QR-Code component

CraigglesO:
Control everything via props.

You can review the npm module here:
https://github.com/CraigglesO/qr-code-react

Essentially creates a react component that avoids `dangerouslySetInnerHTML` and updates as the props do.

Uses `<table />` rather than ``.

Set a border so that cameras/scanners don't get mad. (Margin property also decides border thickness)

Also, I added support to choose the background and foreground colors.


CraigglesO:
@jacogr - This is ready for review.

jacogr:
Good -

- No innerHTML setting

Bad -

- There are other components (just searching npm for react-qr), why are we using new one? Would always prefer to stay closest to well maintained and popular components not not have to split efforts.
- I would not get rid of the playground at all, we lose the ability to do quick tests and seeing effects
- It has been updated elsewhere to detect the size of the QR (transactions are quite a bit larger than the current address-only use - https://github.com/paritytech/parity/pull/4881/files#diff-03558fec29a7e98f8cc36fe70f66fd36R74) so the listening for props will still need to come into play.

CraigglesO:
Ah, I thought that there was a specific reason that the current npm package was being used, so I just edited it partially. If there is not a reason then let's kill this PR.

I originally thought just deleting the folder entirely would help clean up the ui folder a bit. But I just saw a bunch of your work in other PR's with QR codes. SO it's probably better to keep it all.

jacogr:
The reason why the npm package was being used is just historic - component came from pre-React, then converted to React and then included here. (Only benefit is some slight control and obviously staying as close to the actual industry-wide JS QR solution as possible. We know that will continue to be maintained.)

Closing this for now as you indicated, however I think it is worth investigating other potentials (that we can adapt to grow/shrink based on amount of data), once the other stuff is in.

Issue #4801, opened 2017-03-07
Health check API endpoint


tomusdrw:
`curl -v http://localhost:8080/api/status` should return basic health of the node: number of peers, syncing status

adrianbrink:
Is there a tag for low hanging fruit or easy? This is probably a good intro. Also, in which crate do all the end points live?

tomusdrw:
Hi @adrianbrink! I've tagged this as `Q2-easy`.

The right place to put this would be here:
https://github.com/ethcore/parity/blob/81b57a57c713d9bc67efd8950ef1c94e0a0238a5/dapps/src/api/api.rs#L147-L152

It will be required to push through additional dependencies like `SyncStatus` and `Client` to get peers and some other stats (like latest block number perhaps).

adrianbrink:
Awesome. I'll send a PR by Sunday. :-)

adrianbrink:
@tomusdrw sorry, I'll send it tomorrow. Didn't have enough time today.

andrebeat:
I think this can be closed now.

PR #4932, opened 2017-03-16
Add skeleton for status api.

adrianbrink:
#4801 

This is the pull request tracking my implementation of the status api.

@tomusdrw : This isn't done yet. I'll update the PR as I get further. :-)

### DIFF
--- a/dapps/src/api/api.rs
+++ b/dapps/src/api/api.rs
@@ -54,6 +54,11 @@ impl RestApi {
 			e.info().map(|ref info| App::from_info(k, info))
 		}).collect()
 	}
+
+  fn list_number_of_peers(&self) -> Vec<i32> {
+      // return the count from EthSync::peers()
+      vec![1]
+  }
 }
 
 impl Endpoint for RestApi {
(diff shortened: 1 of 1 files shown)


tomusdrw:
@adrianbrink Looks promising! :) BTW: Make sure to use tabs.

rphmeier:
@adrianbrink As a suggestion, you may want to implement actually getting peer values via a generic method. This will make it easier to set up in tests, easier to support other network implementations than `EthSync` (e.g. the light client), and it won't require bringing in a `sync` crate dependency on `dapps`.

You could do that via a `Box<Fn() -> Vec<i32>>` or a `Box<GetPeerCount>`, where `GetPeerCount` has a function for getting peer count.

tomusdrw:
You can actually make `RestApi` generic, since we `Box` it anyway in `to_handler` method.

tomusdrw:
Closing since there is no progress for a long time and the dapps API has been changed substantially so it will require a merge or a re-write.
@adrianbrink Feel free to create a new PR when it's ready :)

PR #4942, opened 2017-03-17
Minimise transactions progress

ngotchac:
Show transactions progress in the UI on the bottom-right corner, so the user can continue working while the transaction is in progress.

Still need to :

- [x] Clean-up every transaction dialog (CreateWallet, DeployContract, ExecuteContract, WalletSettings, WalletActions)
- [x] Check Verification dialog
- [x] Add infos on completion of the transaction (contract address for deploy, wallet info for wallet...)
- [x] Waiting for #4933 to be merged

We might want to add this also for dapps. It would be basically a one line change, but not sure we want to add it though...


js/src/modals/CreateWallet/createWalletStore.js
```diff
@@ -125,24 +112,18 @@ export default class CreateWalletStore {
...
+      .filter((step) => this.walletType === 'WATCH' || step.key !== 'INFO');
   }
 
-  @computed get waiting () {
-    this.steps
-      .map((s, idx) => ({ idx, waiting: s.waiting }))
-      .filter((s) => s.waiting)
-      .map((s) => s.idx);
-  }
+  constructor (api, props) {
```
jacogr: Would prefer being a bit more explicit here instead of just "props". Maybe the right approach would be to have

```
constructor (api, { accounts, onClose })
```

(Small difference, but always visible what is expected)

js/src/api/rpc/parity/parity.js
```diff
@@ -374,9 +374,21 @@ export default class Parity {
...
-  postTransaction (options) {
+  postTransaction (_options = {}) {
+    const { metadata = {}, ...options } = _options;
+    const parsedOptions = inOptions(options);
+
     return this._transport
-      .execute('parity_postTransaction', inOptions(options));
+      .execute('parity_postTransaction', parsedOptions)
+      .then((requestId) => {
+        this._transport.emit('request', {
```
jacogr: This would make more sense on the logging layer where events are already fired. As it stands it now does the "same" thing in 2 different places.

js/src/modals/WalletSettings/walletSettingsStore.js
```diff
@@ -203,7 +194,9 @@ export default class WalletSettingsStore {
     });
   }
 
-  constructor (api, wallet) {
+  constructor (api, props) {
```
jacogr: `constructor (api, { wallet, onClose })`

(Just so it is obvious what to expect, as opposed to passing anything in)

js/src/redux/providers/requestsReducer.js
```diff
@@ -0,0 +1,43 @@
...
+
+const initialState = {};
+
+export default handleActions({
+  setRequest (state, action) {
+    const { requestId, requestData } = action;
+
+    const nextState = {
+      ...state,
+      [ requestId ]: {
```
jacogr: `[requestId]: {...}` would make more sense, not an array/spaced, rather a key.

js/src/ui/ScrollableText/scrollableText.js
```diff
@@ -0,0 +1,41 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import React, { PropTypes } from 'react';
+
+import styles from './scrollableText.css';
+
+export default function ScrollableText (props) {
```
jacogr: Would prefer ({ small = false, text }) (especially in this case where the signature is on top and the properties only follow much later - linter still catches non-defined use in these cases)

js/src/ui/Title/title.js
```diff
@@ -76,7 +76,7 @@ export default class Title extends Component {
   renderSteps () {
     const { activeStep, steps } = this.props;
 
-    if (!steps) {
+    if (!steps || steps.length < 2) {
```
jacogr: Why the steps.length < 2?

jacogr: Scratch that. (I read <= 2, makes sense)

js/src/util/tx.js
```diff
@@ -82,116 +101,103 @@ export function deploy (contract, _options, values, metadata = {}, statecb = ()
...
+  }
+
+  if (transactionLog) {
+    // Set the contract address in the receipt
+    receipt.contractAddress = transactionLog.params.created.value;
+  }
+
+  const contractAddress = receipt.contractAddress;
+
+  return api.eth.getCode(contractAddress)
```
jacogr:
```
api.eth
  .getCode(...)
  .then(...)
```

(Just align with what we have elsewhere)

js/src/redux/providers/requestsActions.js
```diff
@@ -0,0 +1,257 @@
...
+import BigNumber from 'bignumber.js';
+import store from 'store';
+
+import { ERROR_CODES } from '~/api/transport/error';
+import { trackRequest as trackRequestUtil, parseTransactionReceipt } from '~/util/tx';
+
+const LS_REQUESTS_KEY = '_parity::requests';
+
+class CachedRequests {
+  load (api) {
```
jacogr: I really would have preferred if these actions were tested, i.e. the full class. (It will become a pivotal piece, so rather get the base right and build up from there)

js/src/api/format/input.js
```diff
@@ -139,46 +139,48 @@ export function inOptionsCondition (condition) {
...
+
+  Object.keys(options).forEach((key) => {
+    switch (key) {
+      case 'to':
+        // Don't encode the `to` option if it's empty
+        // (eg. contract deployments)
+        if (options[key]) {
+          options.to = inAddress(options[key]);
+        } else {
+          options.to = '';
```
jacogr: Do we need `''`? (Why not just leave it out?)

jacogr:
Mostly small-ish grumbles. However, do really believe we need the tests as commented on, no reason not to do it with new actions (especially when this pivotal) and judging by the rest as time goes on it become more difficult to add.

jacogr:
Not convinced about the transparency and general UX here. It may have been better to 

(a) have a white background (more-or-less like the amount deposited pop-up)
(b) be explicit in the number of confirms (show count)

parity 2017-03-21 11-43-27

ngotchac:
> (d) animations for disappearing is not great, instead of the JS solution, add a class to animate height down to 0 and then remove it.
(e) not sure we actually need the tx-hash there, at least not in the long cannot-read it format (shorten?)

- (d) The animation is actually fading-out, then animating height down to 0. The animation is CSS only, JS is only there to set the height before the animation starts, otherwise you wouldn't be able to animate a variable height.
- (e) The Tx Hash is actually shortened already, into this `ScrollableText` component so you can still see it entirely by scrolling horizontally and select it to copy/paste

CraigglesO:
I LOVE this. This was really well done Nicolas. 

I have 3 grumbles:
1) Some portions of the div don't show the pointer cursor
2) Doesn't autohide from what I could tell (pending transaction), maybe it's better not to change that though.
3) If you refresh the page, pending transactions keep showing up again.
3b) Also, the pending transactions info does not fit inside the window... maybe take the address and do the common "fe31f2...ad20f5"?

jacogr:
Actually didn't see the action buttons at all - maybe it was just the transparency that got me completely confused since I couldn't see what is top/bottom/where. 

Will play again now that it is updated.

jacogr:
Ok, the overlay of dialogs needs to go, it really is super annoying when I actually try to get work done. (Which is what this improves, i.e. you can do multiples all at once.) Colours much better here, couldn't test the minimize (since I have the transactions overlaying my dialogs and have to keep moving them out of the way), but seems like a step in the right direction.

The txhash still seems way too long where. Possibly would make sense to make that (txhash) a click-through to etherscan and have it available somehow in the actual list as well. (Real scenario - just sent ETH and wanted to send the txhash to somebody, couldn't easily find it without going back into my account. Just having it shortened and clickable will allow that)

ngotchac:
So to test the animation, you can create a transaction without signing it, close the signer, create a new one, and then close the bottom pending transaction popup.

I will update the z-indexes so that modals are always on top.
I will make the tx hashes shorter, but I don't think there's really a need for a link here. If you need to retrieve it, you can always go to the Signer page :)

@CraigglesO The transactions will auto-hide when they are validated (after being included into 6 blocks). This might be wrong for transactions with conditions though... But you can always hide those by clicking on them. The fact that they are persistent across page reloads is wanted, and it is especially useful when you deploy a contract!

jacogr:
@ngotchac I'm not 100% sure transactions with conditions show up here. (I could be wrong, but I submitted one this morning and it didn't show - then again, was looking at 20 things in this PR at once, so could be wrong - briefly bugged me, I would just double-check the intent and if you are happy with the way it is handled.)

ngotchac:
@jacogr They are indeed displayed even with conditions. This is IMO the expected behavior so it's fine, but if it doesn't work on your side, there is an issue...

jacogr:
As discussed, getting there, but a couple of things remaining -

1. Use the ShortenedHash for the hash display (the overflow/cutting is not 100% "nice" - and as per your comment above, no real need to click or copy from there, it is available)
2. Would make sense to show a condensed method decoding (i.e. method, recipient, value, no parameters)
3. Adjust the right/bottom margins slightly to be closer to the edges.

ngotchac:
I updated every gumbles.

PR #4947, opened 2017-03-17
Swap out ethcore.io url for parity.io

chmac:
It throws a certificate error on HTTPS because the certificate is only valid on parity.io.


debris:
[clabot:check]

rphmeier:
Thanks for the fix. Unfortunately, it seems that our CLA bot is temporarily down, so I'll ice this until it's back up.

debris:
[clabot:check]

PR #4952, opened 2017-03-17
systemd: Start parity after network.target

gcarq:
This allows parity to cleanly terminate connections before going down, instead of abruptly losing connectivity for ongoing connections, leaving them in an undefined state ([NetworkTarget - freedesktop.org](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/)).

### DIFF
--- a/scripts/parity.service
+++ b/scripts/parity.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Parity Daemon
+After=network.target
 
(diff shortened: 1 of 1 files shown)


debris:
Thanks for the fix. Unfortunately, it seems that our CLA bot is temporarily down, so I'll ice this until it's back up.

PR #4954, opened 2017-03-18
EIP-100

debris:
https://github.com/ethereum/EIPs/issues/100


arkpar:
As far as I can see this implements EIP-100a. EIP-100b looks simpler and this is what go-ethereum implements.

debris:
done

arkpar:
For eip100b you don't really need to read and pass around uncle count. You can just check if `uncles_root` is empty or not in the header.

debris:
you are absolutely right! I will close this one and create a new pr ;)

PR #4958, opened 2017-03-18
Cancel tx JS

CraigglesO:
Javascript updates from #4949 

* updated jsonrpc->parity
* updated api->rpc->parity
* spelling/grammar fixes

img1
img2
img3


npm-debug.log
```diff
@@ -0,0 +1,24 @@
+0 info it worked if it ends with ok
```
jacogr: Please remove.

jacogr:
Remove `npm-debug.log` (should probably be added to `.gitignore`), update with latest master (to pull in the changes from @tomusdrw , currently they still show in this PR)

CraigglesO:
FIXED! Good eye! I added to `.gitignore` for the future.

I'll definitely refactor the code when the edit button comes in.

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +142,46 @@ class TxRow extends Component {
...
   }
 
+  renderCancelToggle () {
+    const { isCancelOpen, canceled } = this.state;
+    const { parity } = this.context.api;
+    const { hash } = this.props.tx;
+
+    if (canceled) {
+      return (
+        <div className={ styles.pending }>CANCELED</div>
```
jacogr: FormattedMessage?

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +142,46 @@ class TxRow extends Component {
...
+      );
+    }
+
+    if (!isCancelOpen) {
+      return (
+        <div className={ styles.pending }>
+          <div>SCHEDULED</div>
```
jacogr: FormattedMessage?

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +142,46 @@ class TxRow extends Component {
...
+          <a
+            className={ styles.cancel }
+            onClick={ () => this.setState({ isCancelOpen: true }) }
```
jacogr: Would prefer not using in-line closures, aligning with the codebase. (I actually thought we had a linting rule that warns about these, weird that it wasn't picked up)

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +142,46 @@ class TxRow extends Component {
...
+      );
+    }
+
+    return (
+      <div className={ styles.pending }>
+        <div>ARE YOU SURE?</div>
+        <a onClick={ () => this.cancelTransaction(parity, hash) }>Cancel</a>
+        <span> | </span>
+        <a onClick={ () => this.setState({ isCancelOpen: false }) }>Nevermind</a>
+      </div>
```
jacogr: FormattedMessage & inline closures.

js/src/ui/TxList/txList.css
```diff
@@ -42,10 +42,11 @@
...
 
     &.timestamp {
-      padding-top: 1.5em;
-      text-align: right;
+      max-width: 5em;
+      padding-top: 2.25em;
+      text-align: center;
       line-height: 1.5em;
-      opacity: 0.5;
+      color: grey;
```
jacogr: Not 100% sure we use this anywhere, would prefer to keep it where it was to keep consistent (We really need to externalise all colours and just pull in CSS variables - no duplication, consistency is maintained)

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -643,6 +643,23 @@ export default {
...
+    params: [
+      {
+        type: Data,
+        desc: 'transactionId, 32-byte hex',
+        example: '0x1db2c0cf57505d0f4a3d589414f0a0025ca97421d2cd596a9486bc7e2cd2bf8b'
+      }
+    ],
+    returns: {
+      type: Boolean,
+      desc: 'returns `true` if the upgrade to the new release was successfully executed, `false` if not.',
```
rphmeier: Description needs updating

js/src/views/Signer/containers/RequestsPage/requestsPage.js
```diff
@@ -100,7 +100,7 @@ class RequestsPage extends Component {
           <div className={ styles.noRequestsMsg }>
             <FormattedMessage
               id='signer.requestsPage.noPending'
-              defaultMessage='There are no requests requiring your confirmation.'
+              defaultMessage='There are no transactions or requests requiring your confirmation.'
```
rphmeier: Not sure if I 100% agree with this, although getting heavily into bikeshed territory :)
Transaction signing requests are still signer requests. "transactions or requests" implies that they aren't.

js/src/views/Signer/containers/RequestsPage/requestsPage.js
```diff
@@ -114,7 +114,7 @@ class RequestsPage extends Component {
         title={
           <FormattedMessage
             id='signer.requestsPage.pendingTitle'
-            defaultMessage='Pending Requests'
+            defaultMessage='Pending Transactions'
```
rphmeier: I believe we can have more than just pending transaction signing requests available here. @tomusdrw?

CraigglesO: Ah! I am super glad you saw these. I meant to point it out to be disputed. I feel Requests is simply somewhat misleading. We need better terminology, correct? Otherwise, I will drop it.

js/src/ui/TxList/txList.css
CraigglesO: The problem is, if opacity is dropped, the "cancel" button will also be 50% opacity LOL. Also, all font is either black or white, 0.5 opacity for both black and white revert to grey anyways.

js/src/views/Signer/containers/RequestsPage/requestsPage.js
rphmeier: They are requests as in "requests for something to be signed". A transaction isn't valid until it's signed. But this interface is also used for more general "sign this random bit of data" as well. I agree that the terminology is a little muddy and clarification would lead to a better UX, but we also have to be careful to remain accurate.

CraigglesO:
TODO:
* After editing, txRow does not fully update
* former conditions do not yet get passed into edit
* I don't yet like how time is printed

Updates:
* Added Editing
* Added Time/Blocks till submit
* fixed FormattedMessage & inline closures
* fixed description


alt image

js/src/views/Signer/containers/RequestsPage/requestsPage.js
```diff
@@ -100,7 +103,7 @@ class RequestsPage extends Component {
           <div className={ styles.noRequestsMsg }>
             <FormattedMessage
               id='signer.requestsPage.noPending'
-              defaultMessage='There are no requests requiring your confirmation.'
+              defaultMessage='There are no transactions requiring your confirmation.'
```
tomusdrw: Your password in signer might be required for couple of things (as Rob mentioned):
1. Sign transaction `eth_sendTransaction`
2. Sign transaction for offline use `eth_signTransaction`
3. Sign aribtrary data `eth_sign`
4. Decrypt arbitrary data `parity_decryptMessages`

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +147,98 @@ class TxRow extends Component {
...
+            id='ui.txList.txRow.verify'
+            defaultMessage='ARE YOU SURE?'
+          />
+        </div>
+        <a onClick={ (isCancelOpen) ? this.cancelTransaction : this.editTransaction }>
+          <FormattedMessage
+            id='ui.txList.txRow.verify.cancelEdit'
+            defaultMessage='{ which }'
+            values={ {
+              which: `${(isCancelOpen) ? 'Cancel' : 'Edit'}`
```
tomusdrw: We should get a big red/yellow warning for transaction cancellation since it's not reliable as mentioned in RPC docs.

There are two separate cases:
1. Scheduled transactions and transactions that were not propagated to other peers - such transactions can be safely canceled, it will be removed from our local queue and never propagated to the network so it's fine.
2. Other transactions - those transactions cannot be simply "canceled" since they are already floating around a network and there is a very high chance that other peers have accepted the transaction and are relaying it further.
For those transactions we should display a warning (but still allow that) and offer another "cancelation" mechanism decribed here: https://github.com/paritytech/parity/issues/4519, so: Increase gas price, remove data, lower gas and submit transaction again.

When editing transaction we should also differentiate those two cases. (1) is safe to edit, but (2) should always increase the gas price when edited (it increases the chance of actually replacing the transaction that was already propagated).

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -165,6 +262,61 @@ class TxRow extends Component {
...
+
+    parity.removeTransaction(hash);
+    this.setState({ canceled: true });
+  }
+
+  editTransaction = () => {
+    const { parity } = this.context.api;
+    const { hash, gas, gasPrice, to, from, value, input, condition } = this.props.tx;
+
+    parity.removeTransaction(hash);
```
CraigglesO: I would consider the fact that you can't use the same gas price as a bug. In this case, you can still use the same gas price, which is automated anyways. If the user didn't have the choice for gas price, but rather the GP was automated, than I would just GP++ and purely edit.

js/src/views/Signer/containers/RequestsPage/requestsPage.js
CraigglesO: Gah, this one is tricky. I would appreciate thoughts of what to change it to.

To me, "requests" makes absolutely no sense from an economics standpoint.
When I first downloaded this app, I was initially thinking, "I have to approve people sending me money? That's silly". But now I know what it does.

How about something akin to,
"There is nothing requiring your confirmation."

Simple. Really lacks specificity, however.

BETTER YET, hide the freakin div all together. Say nothing.

tomusdrw: I like the simple approach or we can use the opportunity to explain why there is a Signer in the first place :)

CraigglesO:
TODO:
* Update store.spec.js
* Improve parity.removeTransaction description
* move two functions to store

CraigglesO:
`js/src/ui/TxList/store.js` Recieved a complete rewrite.
1) It was subscribing a second time to the api, but the requestsPage store was already handling that so it was pushing a new state 2x per update.
2) It did not allow for the removal of hashes.
3) Sometimes the component would update too fast and not all the promises were fulfilled so there were errors.
4) It was way more complex than it needed to be.

I still think signer store should eventually be subscribed to the api in a manner rather than update every 1.5 seconds... but I don't think it's worth posting an issue about currently.

Lemme know what you think I should edit.

js/src/ui/TxList/TxRow/txRow.js
gavofyork: (2) should also zero-out the `value` and, for sanity, replace `to` with `0x000...000`.

js/src/api/rpc/parity/parity.js
```diff
@@ -395,6 +395,11 @@ export default class Parity {
       .execute('parity_removeReservedPeer', encode);
   }
 
+  removeTransaction (transactionId) {
```
tomusdrw: `transactionHash` would be more adequate

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -643,6 +643,34 @@ export default {
...
+    params: [
+      {
+        type: Data,
+        desc: 'transactionId, 32-byte hex',
+        example: '0x1db2c0cf57505d0f4a3d589414f0a0025ca97421d2cd596a9486bc7e2cd2bf8b'
+      }
+    ],
+    returns: {
+      type: Object,
+      desc: 'the removed transaction if cancellation was successful, an error if not.',
```
tomusdrw: note: it's rather "the removed transaction if found, an error if not" (no guarantees about success state here as well)

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-    this.subscribe();
+    this._newError = newError;
   }
 
-  @action addBlocks = (blocks) => {
-    this.blocks = Object.assign({}, this.blocks, blocks);
+  @action addHash = (hash) => {
+    if (!this.sortedHashes.includes(hash)) {
+      this.sortedHashes.push(hash);
+      this.sortHashes();
```
tomusdrw: Could use a sorted data structure (some kind of `SortedList`) for greater efficiency. But it's most probably not on the hot path.

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-          } else if (bnA.eq(0)) {
-            return -1;
-          }
-
-          return bnB.comparedTo(bnA);
-        });
-
-      this._pendingHashes = this.sortedHashes.filter((hash) => this.transactions[hash].blockNumber.eq(0));
-    });
+  @action removeHash = (hashes) => {
```
tomusdrw: I think it should be rather called `setHashes` ?

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
+    _txhashes
+      .map((txhash) => {
+        eth.getTransactionByHash(txhash)
+          .then((tx) => {
+            this.transactions[txhash] = tx;
+            // If the tx has a blockHash, let's get the blockNumber, otherwise it's ready to be added
+            if (tx.blockHash) {
+              eth.getBlockByNumber(tx.blockNumber)
+                .then((block) => {
+                  this.blocks[tx.blockNumber] = block;
```
tomusdrw: This structure will be leaking memory, probably `removeHash/setHashes` should update `this.blocks` as well and remove old entries.

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-    const blockNumbers = uniq(_blockNumbers).filter((bn) => !this.blocks[bn]);
+  cancelTransaction = (txComponent, tx) => {
+    const { parity } = this._api;
+    const { hash } = tx;
 
-    if (!blockNumbers || !blockNumbers.length) {
-      return;
-    }
+    parity.removeTransaction(hash)
+    .then(() => {
```
tomusdrw: Indent.

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -1622,7 +1650,14 @@ export default {
         example: {
           from: '0xb60e8dd61c5d32be8058bb8eb970870f07233155',
           to: '0xd46e8dd67c5d32be8058bb8eb970870f07244567',
-          value: fromDecimal(2441406250)
+          gas: fromDecimal(30400),
```
jacogr: Would just remove the `fromDecimal` wrappers here, i.e. the response should rather be what we receive via the actual RPC, raw.

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -1622,7 +1650,14 @@ export default {
...
+          gasPrice: fromDecimal(10000000000000),
+          value: fromDecimal(2441406250),
+          data: '0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675',
+          condition: {
+            block: 354221,
+            time: new Date()
```
jacogr: Same with these.

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-          blocks.reduce((blocks, block, index) => {
-            blocks[blockNumbers[index]] = block;
-            return blocks;
-          }, {})
-        );
+  editTransaction = (txComponent, tx) => {
+    const { parity } = this._api;
+    const { hash, gas, gasPrice, to, from, value, input, condition } = tx;
+
+    parity.removeTransaction(hash)
```
tomusdrw: Why not:
```js
parity.removeTransaction(hash)
  .then(() => parity.postTransaction({ ... }))
  .then(() => txComponent.setState({editing: true})) // just swap with line above if the order should be different.
  .catch(message => this._newError({ message }));
```

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +149,113 @@ class TxRow extends Component {
...
+            defaultMessage='EDITING'
+          />
+        </div>
+      );
+    }
+
+    if (!isCancelOpen && !isEditOpen) {
+      const pendingStatus = this.getCondition();
+
+      if (pendingStatus === 'submitting...') {
```
jacogr: Not sure I'm 100% on-board with the `submitting...` - would rather make the state checking against a defined constant (e.g. `const STATUS_PENDING = 'submitting'`) and not have the `...` in there

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +149,113 @@ class TxRow extends Component {
...
+        </div>
+      );
+    }
+
+    if (editing) {
+      return (
+        <div className={ styles.pending }>
+          <FormattedMessage
+            id='ui.txList.txRow.editing'
+            defaultMessage='EDITING'
```
jacogr: It probably would make more sense having the `defaultMessage` as `editing/Editing` and using text-transform to make it uppercase. (That way we don't need to edit the text and get translations to update when we change the way we want the display the look)

(Same for `SUBMITTING` & `CANCEL`)

js/src/ui/TxList/TxRow/txRow.js
jacogr: These should be FormattedMessage as well (Should also be translatable)

js/src/views/Signer/containers/RequestsPage/requestsPage.js
```diff
@@ -85,7 +88,9 @@ class RequestsPage extends Component {
       >
         <TxList
           address=''
+          blockNumber={ blockNumber }
           hashes={ localHashes }
+          newError={ newError }
```
tomusdrw: Isn't there an agreement that fired events should be always prefixed with `on*`? Imho thinking in terms of `props` and `events` helps to understand components architecture better (without imperative: "`createNewErrorNow` stuff)

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -137,11 +149,113 @@ class TxRow extends Component {
...
+          <FormattedMessage
+            id='ui.txList.txRow.verify.cancelEdit'
+            defaultMessage='{ which }'
+            values={ {
+              which: `${(isCancelOpen) ? 'Cancel' : 'Edit'}`
+            } }
+          />
+        </a>
+        <span>{' | '}</span>
+        <a onClick={ this.nevermind }>
```
jacogr: Would name this something else, e.g. not `nevermind` (`revertEdit` or `cancelEdit` or ...)

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -165,6 +279,48 @@ class TxRow extends Component {
...
+
+    if (time) {
+      if ((time.getTime() - Date.now()) >= 0) {
+        return `${dateDifference(new Date(), time, { compact: true })} left`;
+      } else {
+        return 'submitting...';
+      }
+    } else if (blockNumber) {
+      block = blockNumber.minus(block);
+      return (block.toNumber() < 0)
```
jacogr: FormattedMessage here

js/src/views/Signer/store.js
```diff
@@ -95,7 +95,11 @@ export default class SignerStore {
     this._api.parity
       .localTransactions()
       .then((localTransactions) => {
-        this.setLocalHashes(Object.keys(localTransactions));
+        const keys = Object
+          .keys(localTransactions)
+          .filter((key) => localTransactions[key].status !== 'invalid');
```
tomusdrw: Why do we skip `invalid` transactions here? I think it would be quite usefull to re-submit transaction cancelled by mistake. Maybe just change the way those are displayed?

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-  loadBlocks (_blockNumbers) {
-    const blockNumbers = uniq(_blockNumbers).filter((bn) => !this.blocks[bn]);
+  cancelTransaction = (txComponent, tx) => {
+    const { parity } = this._api;
+    const { hash } = tx;
 
-    if (!blockNumbers || !blockNumbers.length) {
-      return;
-    }
+    parity.removeTransaction(hash)
```
jacogr: Slight style grumble, would prefer `.then` & `.catch` indented, i.e.

```
parity
  .removeTransaction(...)
  .then(...)
  .catch(...)
```

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-            return blocks;
-          }, {})
-        );
+  editTransaction = (txComponent, tx) => {
+    const { parity } = this._api;
+    const { hash, gas, gasPrice, to, from, value, input, condition } = tx;
+
+    parity.removeTransaction(hash)
+      .then(() => {
+        parity.postTransaction({
```
jacogr: Indentation of this block and the containing one. In this case (and I generally do exactly what you've done myself), it could aid readability to have `transaction` as a `const` and passing it in.

js/src/ui/TxList/txList.css
```diff
@@ -42,10 +42,11 @@
...
 
     &.timestamp {
-      padding-top: 1.5em;
-      text-align: right;
+      max-width: 5em;
+      padding-top: 0.75em;
+      text-align: center;
       line-height: 1.5em;
-      opacity: 0.5;
+      color: grey;
```
jacogr: Would keep the original to align with the rest of the UI. (Close enough, but not 100% the same in terms of colors here, we don't use gray elsewhere.) Yes... we need to externalize all color and related vars and just pull them in.

js/src/ui/TxList/txList.js
```diff
@@ -35,19 +35,16 @@ class TxList extends Component {
...
-    netVersion: PropTypes.string.isRequired
+    blockNumber: PropTypes.object,
+    netVersion: PropTypes.string.isRequired,
+    newError: PropTypes.func
   };
 
-  store = new Store(this.context.api);
-
   componentWillMount () {
+    this.store = new Store(this.context.api, this.props.newError);
```
jacogr: Any reason for moving this? (They are equivalent)

js/src/ui/TxList/txList.css
CraigglesO: All links and others (sumbitting/ cancel/ edit ...) will be 50% as well. Which is a problem to see/links look strange. I don't mind adding more css, so that the main is 0.5 opacity. But I can't keep the css the way it is. Suggestions are welcome.

js/src/ui/TxList/store.js
CraigglesO: No, this literally removes the hash from the list.

Remember above there is "addHash"

Ex. I click cancel, so the whole tx disappears from the list... This function ensures the tx disappears and the state updates.

js/src/ui/TxList/store.js
CraigglesO: Instead of arrays? V8 arrays used in Chrome are faster. I don't know how it is for FF or whatever.

js/src/ui/TxList/store.js
CraigglesO: Absolutely, I was originally going to give two different error reports with more information depending on where it failed, but I couldn't think of a way to purposely create a failure so.... LOL.

js/src/views/Signer/containers/RequestsPage/requestsPage.js
CraigglesO: I kept it like this to avoid confusion. That's a redux function I am passing in. I think it's more obvious to keep the name.

js/src/views/Signer/store.js
CraigglesO: Huh. I don't know how else to remove canceled transactions. When removeTransaction is called, it does not actually remove the tx from localTransactions, but rather gives 'invalid', so editing and cancelling tx, they don't go away...

Like we discussed before. To avoid having to change the gasPrice, I completely cancel the TX and then resubmit. It works beautifully. If you hit cancel, the submittal bar pops up with every single detail from the just cancelled tx still there. So you just can sign and resubmit for the exact tx back again.

js/src/ui/TxList/txList.js
CraigglesO: newError won't get loaded in until CWM due to lifetimes. I should really remove:
`this.store.loadTransactions(this.props.hashes);` and have that part of the constructor method too....

js/src/ui/TxList/txList.css
jacogr: Ahh, ok, the links as well, makes sense. Maybe be more specific then with `>span` (FormattedMessage)

js/src/views/Signer/store.js
tomusdrw: Maybe keep them but display them in UI differently? Actually I will prepare a special `status` for canceled transactions. `invalid` can happen in some other cases as well and the whole `localTransactions` thing was created to never loose any of transactions that were sent from local node.

js/src/views/Signer/containers/RequestsPage/requestsPage.js
tomusdrw: Ok, it's not inline with some other functions in the same file though, we map redux actions to component events like this:
```js
onConfirm={ startConfirmRequest }
onReject={ startRejectRequest }
```

js/src/ui/TxList/store.js
tomusdrw: Arrays are fast, but instead of sorting in `O(n log(n))` each time you instert an element you can just insert with `O(log(n))` complexity and for large number of elements this will matter.

But nevermind, now I see this is just used to store local transactions so this will be 25 transactions at most afair so it's not even worth thinking about :)

js/src/ui/TxList/store.js
tomusdrw: Ok, I think I get the context in which this function is used, but `removeHash` seems like a function that should do the invert of `addHash` (given hash add it to the list if not present) so it should do: "given hash remove it from the list if present". This function is actually just overriding all hashes with new ones that's why I though it should be named `setHashes` instead of `removeHash`.

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,121 @@
...
-      .then((_transactions) => {
-        const blockNumbers = [];
-        const transactions = _transactions
-          .filter((tx) => tx && tx.hash)
-          .reduce((txs, tx) => {
-            txs[tx.hash] = tx;
-
-            if (tx.blockNumber && tx.blockNumber.gt(0)) {
-              blockNumbers.push(tx.blockNumber.toNumber());
+    _txhashes
```
tomusdrw: Imagine this:
```js
this.sortedHashes = ['A', 'B']
_txhashes = ['A', 'C']
```
It has the same length, but it's not the same array so it will fall into this branch and you'll end up with
```
this.sortedHashes = ['A', 'B', 'C' ]
```
I believe this could be a case when for instance `B` is canceled in favour of `C` and we'll fail to register cancelation first.

Actually it does way more work then required, cause it fetches `eth.getTransactionByHash` for each transaction even if those transactions are already in `this.transactions`.

js/src/ui/TxList/store.js
tomusdrw: On a second glance: `this.transactions` will be leaking as well, we're never removing any entries from it.

CraigglesO: Remember that react components have lifetimes. This is a Mobx store, so it dies with the component. Like you said earlier, 25ish tops are going to exist. So as soon as the component dies, so does all the data. No reason to add more code.

js/src/ui/TxList/store.js
CraigglesO: I couldn't think at the time of a case where in the ui this could happen. 

But I believe you if you think this will be a problem in the future, so I will change it.

edit - If anything changes, I resubmit for all tx just incase there were any changes along the way. I was worried if I filtered the old tx there wouldn't be changes (future edits)

CraigglesO: Relooking at my code now, this is actually not an issue.
this.sameHashList checks the equality of the two (this.containsAll both directions),
so:
(['A','B'], ['B','A']) === true
(['A','B'], ['A','C']) === false

It will be a two step process... if the scenario you explained happened, the first thing it would do is add 'C' to this.sortedHashes, than it would see that it has more variables than its counterpart and remove B.

CraigglesO:
- [ ] removeHash => setHashes
- [ ] update parity.removeTransaction details
- [ ] transactionId => transactionHash (js/src/api/rpc/parity/parity.js)
- [ ] formatMessages
- [ ] never mind => something better
- [ ] text-transform => uppercase
- [ ] submitting… => Submitting
- [ ] remove fromDecimal wrappers (js/src/jsonrpc/interfaces/parity.js)
- [ ] newError => onNewError
- [ ] filter invalid => filter canceled
- [ ] update then catch formatting

js/src/ui/TxList/store.js
tomusdrw: Yes, but if I leave my page open on transaction list it will consume all memory after some time. IMHO its an obvious memory leak and we shouldn't tell our users to just refresh the page from time to time.

js/src/ui/TxList/store.js
tomusdrw: Yes, only if on the second step this condition would be satisfied: `_txhashes.length < this.sortedHashes.length`.
I suppose it will reach eventualy consistency, but the fact is that it's not working 100% correctly in all cases and may cause some heisenbugs.

js/src/ui/TxList/txList.js
tomusdrw: I think you should also cater for `newError` being changed during the lifetime of the component (or when component will be re-used with different props). Isn't it possible to avoid passing the error directly?
Like:
```js
store: new Store(this.context.api, (...args) => this.props.newError(...args))
// or
//store: new Store(this.context.api, (...args) => this.props.newError ? this.props.newError(...args) : null)
```

js/src/ui/TxList/TxRow/txRow.js
CraigglesO: All string returns pulled from `getCondition` as seen here are formatted on line `205`

js/src/ui/TxList/txList.js
CraigglesO: newError is a redux action. So I don't really know what you mean. Basically it goes like this:

action->reducer->store->component.

passing in an argument to newError is the action phase.

tomusdrw: Ok, so the issue was that this:
```js
store: new Store(this.context.api, this.props.newError)
```
doesn't work cause `this.props.newError` is not available yet because of the lifetime issues you mentioned. So your solution was to do:
```js
componentWillMount () {
  this.store = new Store(this.context.api, this.props.newError);
}
```
And that works correctly. My point is that it's still not working for all cases (it's not generic enough), cause `this.props.newError` can change if the parent component decides to change it. So proper implementation should look like this:
```js
componentWillMount () {
  this.store = new Store(this.context.api, this.props.newError);
}
componentWillReceiveProps (newProps) {
  if (this.props.newErrror !== newProps.newError) {
    this.store = new Store(this.context.api, newProps.newError);
    // + copy state of the previous store most probably.
  }
}
```
but that's way to complicated. So instead I proposed a simpler solution:
```js
store: new Store(this.context.api, (...args) => this.props.newError(...args))
```

So store has a reference to a lambda function that will dynamically dispatch to `this.props.newError` so even if `this.props.newError` is changed the Store doesn't need to be updated.

IMHO even if `this.props.newError` is not subject to change currently we shouldn't rely on that cause in future it may and such code may cause a wierd errors in the future that are easy to avoid now.

Hope this explains my motivations and solution ;)

ngotchac: So I think there's a misunderstanding here. The `newError` prop is actually added in the `connect` part of the Component, managed by `react-redux`. This just adds the `newError` function prop from the imported function in `./redux/.../errorActions` or something similar. It is not passed from a parent (or at least shouldn't be). As it's an import, the function won't change during the lifetime of the application.

CraigglesO:
Instead of bulk handling change, this version will handle individual changes noticed.

I also stumbled upon hashes that have null returned via `eth.getTransactionByHash`. So cleaned that up too.

If there are any other minute details you guys see, let me know.

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -1674,9 +1674,9 @@ export default {
         example: {
           from: '0xb60e8dd61c5d32be8058bb8eb970870f07233155',
           to: '0xd46e8dd67c5d32be8058bb8eb970870f07244567',
-          gas: fromDecimal(30400),
-          gasPrice: fromDecimal(10000000000000),
-          value: fromDecimal(2441406250),
+          gas: 30400,
```
tomusdrw: Just curious, why was it changed? It won't work with decimals on RPC layer.

js/src/ui/TxList/store.js
```diff
@@ -14,150 +14,131 @@
...
-
-          return bnB.comparedTo(bnA);
-        });
-
-      this._pendingHashes = this.sortedHashes.filter((hash) => this.transactions[hash].blockNumber.eq(0));
-    });
+  @action removeHash = (hash) => {
+    this.sortedHashes.remove(hash);
+    delete this.transactions[hash];
+    delete this.blocks[hash];
```
tomusdrw: This should be:
```
let tx = this.transactions[hash];
if (tx) {
  delete this.transactions[hash];
  delete this.blocks[tx.blockNumber];
}
```
looks good otherwise.

js/src/jsonrpc/interfaces/parity.js
CraigglesO: Jaco wanted them removed. If you are sure they are required than absolutely I will put them back.

tomusdrw: I think he meant to convert the decimals into hex values so that the example is not using any functions (but I'm unsure though).

CraigglesO:
This should do it. 👍

js/src/ui/TxList/TxRow/txRow.js
```diff
@@ -165,6 +282,48 @@ class TxRow extends Component {
...
+      if ((time.getTime() - Date.now()) >= 0) {
+        return `${dateDifference(new Date(), time, { compact: true })} left`;
+      } else {
+        return 'submitting';
+      }
+    } else if (blockNumber) {
+      block = blockNumber.minus(block);
+      return (block.toNumber() < 0)
+        ? block.abs().toFormat(0) + ' blocks left'
+        : 'submitting';
```
jacogr: Just make the strings here translatable. (Also line 247 is still, as part of the value for `which`, non-translatable)

jacogr:
Couple of last i18n issues, but apart from that looking good.

CraigglesO:
Let me know 💃

jacogr:
One last conflict.

CraigglesO:
DONE.

5chdn:
Can we have this backported to beta? It's causing a support nightmare :)

PR #4961, opened 2017-03-20
Switching ValidatorSet

keorn:
A new `ValidatorSet` which switches between different `ValidatorSet`s on given blocks.

### DIFF
--- a/ethcore/src/engines/validator_set/multi.rs
+++ b/ethcore/src/engines/validator_set/multi.rs
@@ -1,0 +1,139 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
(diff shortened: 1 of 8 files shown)


rphmeier:
We might want to extend this to "multi" fields in general for more customizability.

ethcore/src/engines/validator_set/multi.rs
```diff
@@ -0,0 +1,139 @@
...
+	fn correct_set(&self, bh: &H256) -> Option<&Box<ValidatorSet>> {
+		match self
+			.block_number
+			.read()(bh)
+			.map(|parent_block| self
+					 .sets
+					 .iter()
+					 .rev()
+					 .find(|&(block, _)| *block <= parent_block + 1)
+					 .expect("First ValidatorSet has to start at block 0!")
```
rphmeier: Should be documented, and the example starts at block 1.

ethcore/src/engines/validator_set/multi.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::sync::Weak;
+use util::{H256, Address, RwLock};
+use ids::BlockId;
+use header::BlockNumber;
+use client::{Client, BlockChainClient};
+use super::ValidatorSet;
+
+pub struct Multi {
+	sets: BTreeMap<BlockNumber, Box<ValidatorSet>>,
+	block_number: RwLock<Box<Fn(&H256) -> Result<BlockNumber, String> + Send + Sync + 'static>>,
```
rphmeier: a typedef of the closure would really help readability here

ethcore/src/engines/validator_set/multi.rs
rphmeier: Speaking more broadly, it makes sense to do these kinds of checks at deserialization time, and then the expectation can be infallible.

PR #4964, opened 2017-03-20
eth_sign where account === undefined

jacogr:
Fixes https://github.com/paritytech/parity/issues/4962

### DIFF
--- a/js/src/views/Signer/components/TransactionPendingForm/TransactionPendingFormConfirm/transactionPendingFormConfirm.js
+++ b/js/src/views/Signer/components/TransactionPendingForm/TransactionPendingFormConfirm/transactionPendingFormConfirm.js
@@ -36,6 +36,7 @@ export default class TransactionPendingFormConfirm extends Component {
   };
 
   static defaultProps = {
+    account: {},
     focus: false
   };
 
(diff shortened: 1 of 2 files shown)


tomusdrw:
Would be good to have a test case for `eth_sign` format.

js/src/views/Signer/components/TransactionPendingForm/TransactionPendingFormConfirm/transactionPendingFormConfirm.js
```diff
@@ -27,7 +27,7 @@ import styles from './transactionPendingFormConfirm.css';
 
 export default class TransactionPendingFormConfirm extends Component {
   static propTypes = {
-    account: PropTypes.object.isRequired,
+    account: PropTypes.object,
```
tomusdrw: Perhaps you can use `defaultProps` instead of `account &&`?

jacogr: The funny thing here was that (a) the `isRequired` was not true & (b) the tests actually covered up the failure. But yes, making it default empty is probably a better approach than checking each access. 

Will update, thanks.

Issue #4824, opened 2017-03-08
Backup/Restore keys instead of Export on Accounts Tab


tomusdrw:
Current "Export" button on Accoutns tab is a bit misleading. One may expect that it exports all accounts managed by Parity but it's not the case.
![screenshot from 2017-03-08 13-23-40](https://cloud.githubusercontent.com/assets/581548/23703916/a28317b0-0402-11e7-865d-1881a663b6ce.png)

The button actually exports only accounts data that seem kind of useless (NOTE: there is no "Import" button)
```
{
    "0x4d6Bb4ed029B33cF25D0810b029bd8B1A6bcAb7B": {
        "name": "Icarus Rain",
        "uuid": "434e0792-bda9-19e0-bd60-6d90e5235cac",
        "meta": {
            "timestamp": 1483691260203,
            "passwordHint": ""
        },
        "address": "0x4d6Bb4ed029B33cF25D0810b029bd8B1A6bcAb7B"
    }
}
```

I think it would rather make more sense to expose full key objects from the UI so that they can be imported on a different node/chain without touching any file system.

thekevinb:
I second this.  
An fast usability fix could include 
"Export"=>"Metadata Export"
and a warning dialog about private keys not being included.

ngotchac:
Closed via #4973

PR #4967, opened 2017-03-20
Export account RPC

tomusdrw:
Related #4824 

Also adds documentation to `ethstore` and cleans up some outdated functions of `ethstore` CLI.


ethstore/src/dir/paths.rs
```diff
@@ -0,0 +1,96 @@
...
+	home.push("Ethereum");
+	home
+}
+
+#[cfg(not(any(target_os = "macos", windows)))]
+fn geth_base() -> PathBuf {
+	let mut home = home();
+	home.push(".ethereum");
+	home
+}
```
debris: it would be good to make a standalone module from it. in #4956 I moved `util/src/path.rs → util/path/src/lib.rs` would be nice to merge this two together

ethstore/src/bin/ethstore.rs
```diff
@@ -116,10 +116,13 @@ fn main() {
...
-		"parity" => Box::new(ParityDirectory::create(DirectoryType::Main)?),
-		"parity-test" => Box::new(ParityDirectory::create(DirectoryType::Testnet)?),
-		"geth" => Box::new(GethDirectory::create(DirectoryType::Main)?),
-		"geth-test" => Box::new(GethDirectory::create(DirectoryType::Testnet)?),
+		"geth" => Box::new(RootDiskDirectory::create(paths::geth(false))?),
+		"geth-test" => Box::new(RootDiskDirectory::create(paths::geth(true))?),
+		path if path.starts_with("parity") => {
+			let chain = path.split('-').nth(1).unwrap_or("ethereum");
+			let path = paths::parity(chain);
+			Box::new(RootDiskDirectory::create(path)?)
```
debris: this is not really backwards compatible, is it?

tomusdrw: It works for `parity` (loads mainnet), `parity-testnet` is not really well defined currently, we can alias it to `parity-kovan` if you think it would be better.

ethstore/src/dir/paths.rs
tomusdrw: Since both of those PRs are not merged yet I think it would be best to do it separately. Logged: https://github.com/paritytech/parity/issues/4972

CraigglesO:
Dunno if this was intended but the meta data is the only thing that is stringified:
```
{
    "address": "007e101452124aae9e21f25d4e05f0411835ec75",
    "crypto": {
        "cipher": "aes-128-ctr",
        "cipherparams": {
            "iv": "7968ac68247490d108db682814eee446"
        },
        "ciphertext": "fd7fd2b862d8c6a38de5df34c9404bafa3c17a0f416ca07c8f4056ef32cbcb8f",
        "kdf": "pbkdf2",
        "kdfparams": {
            "c": 10240,
            "dklen": 32,
            "prf": "hmac-sha256",
            "salt": "e33c027fe579f359f86aff90e222b875060fcd8f6d7e83cfcd182c94d08243e7"
        },
        "mac": "6a8bf7e3dadd59d914f775f69907d23fc5ac86711ec6845c668998f4edfb29e2"
    },
    "id": "5394726e-06f4-9bb0-384d-170235296a6b",
    "meta": "{\"passwordHint\":\"E\",\"timestamp\":1490065563273}",
    "name": "CONNOR-DEV",
    "version": 3
}
```

tomusdrw:
That's actually expected. We don't really parse `meta` at all in Parity anywhere, it's only used by UI.

PR #4973, opened 2017-03-21
Export acc js

CraigglesO:
Export accounts as json files
JS from #4967

some typos and need to finish adding in the api. Super close.

alt-img


CraigglesO:
[clabot:check]

CraigglesO:
Done.

Ice Ice baby ;)

fail:
alt
success:
alt-2

jacogr:
Marked onice for merging until https://github.com/paritytech/parity/pull/4967 merged/available.

ngotchac:
Closes #4824

jacogr:
As discussed - We need this on the Account page as well, i.e. going into account, being a able to export from there.

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,195 @@
...
+class ExportAccount extends Component {
+  static contextTypes = {
+    api: PropTypes.object.isRequired
+  };
+
+  static propTypes = {
+    accounts: PropTypes.object.isRequired,
+    balances: PropTypes.object.isRequired,
+    newError: PropTypes.func.isRequired,
+    personalAccountsInfo: PropTypes.func.isRequired,
```
jacogr: Could we keep this as `accountsInfo` (just consistent with what we have everywhere else)

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,195 @@
...
+    onClose: PropTypes.func.isRequired
+  };
+
+  static ExportStore = ExportStore;
+
+  componentDidMount () {
```
jacogr: Remove if not used.

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,195 @@
...
+
+  static propTypes = {
+    accounts: PropTypes.object.isRequired,
+    balances: PropTypes.object.isRequired,
+    newError: PropTypes.func.isRequired,
+    personalAccountsInfo: PropTypes.func.isRequired,
+    onClose: PropTypes.func.isRequired
+  };
+
+  static ExportStore = ExportStore;
```
jacogr: Is there a specific reason this is not on the instance? Basically doing it twice will (on the second run), have the same selection/values as on the first.

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,195 @@
...
+
+  exportAddress (account, password) {
+    const { parity } = this.context.api;
+    const { accounts, newError } = this.props;
+
+    parity.exportAccount(account, password)
+      .then((content) => {
+        const text = JSON.stringify(content, null, 4);
+        const blob = new Blob([ text ], { type: 'application/json' });
+        const filename = accounts[account].name;
```
jacogr: We should stick to the UUID naming as per Parity for the sake of consistency.

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,195 @@
...
+          this.exportAddress(account, inputValue[account]);
+        }
+      });
+  }
+
+  exportAddress (account, password) {
+    const { parity } = this.context.api;
+    const { accounts, newError } = this.props;
+
+    parity.exportAccount(account, password)
```
jacogr: Would move the interaction to the store. (As well as onExport, i.e. all logic in the actual store as opposed to the component)

js/src/ui/AccountCard/accountCard.js
```diff
@@ -32,19 +34,22 @@ export default class AccountCard extends Component {
     className: PropTypes.string,
     disableAddressClick: PropTypes.bool,
     onClick: PropTypes.func,
-    onFocus: PropTypes.func
+    onFocus: PropTypes.func,
+    password: PropTypes.bool,
```
jacogr: `showPassword`? (On first read-through it looked like passing the actual password)

js/src/ui/AccountCard/accountCard.js
```diff
@@ -88,6 +93,7 @@ export default class AccountCard extends Component {
             className={ styles.balance }
             showOnlyEth
           />
+          { (password) ? this.renderPassword() : null }
```
jacogr: @ngotchac Happy with the approach or would passing children from the renderer be more appropriate? (Not unhappy, just concerned that this may blow up as we add more stuff - and currently the password is a single-instance use as well)

jacogr:
Some grumbles around the store and moving the actual RPCs there, in addition one question for @ngotchac around the password, may possibly be better to have a more generic `this.props.children` renderer instead.

js/src/ui/AccountCard/accountCard.js
CraigglesO: I was thinking about this too. Should really just do children, as I could see in the future using accountCards in this way.

js/src/modals/ExportAccount/ExportAccount.js
CraigglesO: pesonalAccountsInfo? It's a redux action pulled in connect->dispatch. Lol, it does look like I added it in though.

CraigglesO:
Should be ready. MUCH better and cleaner. 

* AccountCard now has a child prop
* all proactive code is moved to the store
* other refactoring/better ui-ux
* Export individual accounts

alt-img

js/src/modals/ExportAccount/ExportAccount.js
```diff
@@ -0,0 +1,182 @@
...
+import { FormattedMessage } from 'react-intl';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import { newError } from '~/redux/actions';
+import { personalAccountsInfo } from '~/redux/providers/personalActions';
+import { AccountCard, Button, Portal, SelectionList } from '~/ui';
+import { Input } from '~/ui/Form';
+import { CancelIcon, CheckIcon } from '~/ui/Icons';
+import ExportStore from './ExportStore';
```
jacogr: I would import, and create an instance per Component here. (Aligns with what we do elsewhere as well), i.e.

```
... extends Component {
  ...
  exportStore = new ExportStore();
  ...
  render () {
    const { canExport, ... } = this.exportStore;
...
}
```

js/src/modals/ExportAccount/ExportAccount.js
jacogr: We are not using this anywhere? (Should be removed)

js/src/modals/ExportAccount/ExportStore.js
```diff
@@ -0,0 +1,89 @@
...
+
+import { action, observable } from 'mobx';
+import FileSaver from 'file-saver';
+
+class ExportStore {
+  @observable canExport = false;
+  @observable selectedAccounts = {};
+  @observable inputValue = {};
+
+  insertProps (api, accounts, newError) {
```
jacogr: As per above comment, would make more sense passing this in the constructor.

js/src/modals/ExportAccount/ExportStore.js
```diff
@@ -0,0 +1,89 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import { action, observable } from 'mobx';
+import FileSaver from 'file-saver';
+
+class ExportStore {
```
jacogr: Generally, when adding new stores, it is good practice to add tests for them. (Since most/all of the logic are actually contained within.)

js/src/modals/ExportAccount/ExportStore.js
```diff
@@ -0,0 +1,89 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
jacogr: Just rename the file as `exportStore.js` to fit in with the rest of the codebase.

js/src/ui/AccountCard/accountCard.js
```diff
@@ -27,6 +27,7 @@ import styles from './accountCard.css';
 
 export default class AccountCard extends Component {
   static propTypes = {
+    children: PropTypes.node.isRequired,
```
jacogr: Should be optional? Not all AccountCard users have children to supply.

js/src/ui/AccountCard/accountCard.js
```diff
@@ -36,15 +37,16 @@ export default class AccountCard extends Component {
   };
 
   static defaultProps = {
-    disableAddressClick: false
+    disableAddressClick: false,
+    showPassword: false
```
jacogr: `showPassword` is not on props, remove.

js/src/views/Account/account.js
```diff
@@ -50,15 +51,20 @@ class Account extends Component {
...
   store = new Store();
   hwstore = HardwareStore.get(this.context.api);
 
   componentDidMount () {
+    const { accounts, newError, params } = this.props;
+    const { address } = params;
+
     this.props.fetchCertifiers();
     this.setVisibleAccounts();
+    this.store.insertProps(this.context.api, accounts, address, newError);
```
jacogr: As per above comment and the use of the stores. In this case, I would not change the store, rather use the already created ExportStore to export the individual accounts.

js/src/views/Account/store.js
```diff
@@ -52,4 +70,27 @@ export default class Store {
   @action toggleVerificationDialog = () => {
     this.isVerificationVisible = !this.isVerificationVisible;
   }
+
+  onExport = () => {
```
jacogr: Duplicated, DRY and keep one store that is used by all the different components.

js/src/ui/AccountCard/accountCard.js
CraigglesO: That's so bizarre, my local branch doesn't have that code... hopefully the next up fixes that.

CraigglesO:
- [x] Added specs
- [x] Changed file name to be consistent
- [x] DRY refactoring
- [x] Child props for accountCard
- [x] All logic moved to inside the store(s)
- [x] Added constructor method
- [x] fixed some other small things

js/src/api/rpc/parity/parity.js
```diff
@@ -137,6 +137,11 @@ export default class Parity {
       .execute('parity_executeUpgrade');
   }
 
+  exportAccount (account, password) {
+    return this._transport
+      .execute('parity_exportAccount', inHex(account), password);
```
jacogr: `inAddress(account)` as opposed to `inHex(account)`?

js/src/modals/ExportAccount/exportAccount.js
```diff
@@ -0,0 +1,182 @@
...
+    balances: PropTypes.object.isRequired,
+    newError: PropTypes.func.isRequired,
+    personalAccountsInfo: PropTypes.func.isRequired,
+    onClose: PropTypes.func.isRequired
+  };
+
+  componentWillMount () {
+    const { accounts, newError } = this.props;
+
+    this.exportStore = new ExportStore(this.context.api, accounts, newError, null);
```
jacogr: Not a major issue, but could be done directly in the class body as well, don't need to rely on componentWillMount

js/src/modals/ExportAccount/exportStore.spec.js
```diff
@@ -0,0 +1,149 @@
...
+      expect(multiAccountStore._address).to.deep.equal(null);
+    });
+
+    it('newError created', () => {
+      expect(multiAccountStore._newError).to.deep.equal(null);
+    });
+  });
+
+  describe('@action', () => {
+    it('toggleSelectedAccount', () => {
```
jacogr: Makes more sende to `describe` the functions and hen have the `it` blocks saying exactly what the expectations are, i.e.

```
describe('toggleSelectedAccount', () => {
  it('updates the selected accounts', () => {
...
```

(Same for the rest of the functions)

js/src/views/Account/account.js
```diff
@@ -302,6 +324,63 @@ class Account extends Component {
...
   }
 
+  renderExportDialog () {
+    const { toggleExportDialog } = this.store;
+    const { changePassword, accountValue, onExport } = this.exportStore;
+
+    if (!this.store.isExportVisible) {
+      return null;
+    }
+    return (
```
jacogr: Tiny style grumble, would prefer an empty line between blocks, just to aid readability somewhat.

jacogr:
Getting close, couple of comments -

1. I would expect to be able to export multiple accounts, currently can only do one at a time
2. We cannot deselect export entries, i.e. clicking selects, clicking again does not deselect
3. When opening the export (inside account) again, it remembers my previous password
4. When pressing eneter on the password field, it doesn't export
5. (As you suggested) it doesn't auto-close on export success, would make sense doing that

js/src/modals/ExportAccount/exportAccount.js
CraigglesO: Actually, due to lifetimes, props are not guaranteed to be loaded until componentWillMount. this.context.api will work. Accounts and newError do not. Just incase you were right and I was wrong, I tried it outside CWM and it didn't work.

js/src/modals/ExportAccount/exportStore.spec.js
CraigglesO: AHHH, ok!!! I was a tape user before. LOL, mocha is definitely interesting though.

CraigglesO:
Pretty slider!!!

alt-img

I had to add the slider because the hierarchy was:
SelectionList-> SectionList -> AccountCard -> Child
but the selection control is in selectionList,  so I had no way controlling the toggle in the Child component. Changing a fuck load of code for this seemed like a bad idea.

The toggle looks great though, right?! ^_^

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+  @action setSelectedAccount = (addr) => {
+    this.selectedAccounts[addr] = true;
+    this.canExport = true;
+  }
+
+  @action toggleSelectedAccount = (addr) => {
+    this.selectedAccounts[addr] = !this.selectedAccounts[addr];
+    if (!this.selectedAccounts[addr]) {
+      delete this.selectedAccounts[addr];
+    }
```
ngotchac: Might be more understandable like this:
```js
    if (this.selectedAccounts[addr]) {
      delete this.selectedAccounts[addr];
    } else {
      this.selectedAccounts[addr] = true;
    }
```

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+    this.canExport = false;
+    Object
+      .keys(this.selectedAccounts)
+      .forEach((address) => {
+        if (this.selectedAccounts[address]) {
+          this.canExport = true;
+        }
+      });
```
ngotchac: As the non-selected accounts are deleted from `this.selectedAccounts`, we could just check for `Object.keys(this.selectedAccounts).length`, right?

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  onExport = (event) => {
+    const { parity } = this._api;
+    const accounts = (this._address) ? [this._address] : Object.keys(this.selectedAccounts);
```
ngotchac: Same 3 lines split grumble as above

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+
+    accounts.forEach((account) => {
```
ngotchac: `accounts.forEach((address) => {` might be more relevant

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+
+          this.accountValue = '';
+          if (event) { event(); }
+        })
+        .catch((err) => {
+          const { name, meta } = this._accounts[account];
+          const { passwordHint } = meta;
+
+          this._newError({
+            message: `[${err.code}] Account "${name}" - Incorrect password. (Password Hint: ${passwordHint})`
```
ngotchac: Not sure if possible, but could use `FormattedMessage` here

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,107 @@
...
+  @observable selectedAccount = '';
+  @observable selectedAccounts = {};
+  @observable accountValue = '';
+  @observable inputValue = {};
+
+  constructor (api, accounts, newError, address) {
+    this._api = api;
+    this._accounts = accounts;
+    this._newError = newError;
+    this._address = address;
```
ngotchac: I think a variable to toggle between single and multiple exports should be sufficient here. It seems strange to have these two different variables : `inputValue` for multiple exports and `accountValue` for a single one. I think it could be simpler.

js/src/modals/ExportAccount/exportAccount.js
```diff
@@ -0,0 +1,201 @@
...
+    const { balances } = this.props;
+    const balance = balances[account.address];
+    const { changePassword, getPassword, onFocus } = this.exportStore;
+    const inputValue = getPassword(account);
+
+    return (
+      <AccountCard
+        account={ account }
+        balance={ balance }
+        onClick={ () => onFocus(account.address) }
```
ngotchac: Why not having the `onSelect` here? Plus I think that `onFocus` doesn't do anything really.

js/src/modals/ExportAccount/exportAccount.js
```diff
@@ -0,0 +1,201 @@
...
+      </Portal>
+    );
+  }
+
+  renderList () {
+    let { accounts } = this.props;
+
+    accounts = Object
+      .keys(accounts)
+      .map((address) => accounts[address]);
```
CraigglesO: LOL, at first glance it looks confusing.

SelectionList requires an array of objects, rather than an object holding objects.

before:
```
{
  'A': {}, 'B': {}, 'C': {} 
}
```

after:
```
[
  0: {}, 1: {}, 2: {} 
]
```

js/src/modals/ExportAccount/exportAccount.js
CraigglesO: This is big struggle for me. I didn't want to rewrite the components since I can't easily access the account prop from exportAccount.js, but I totally know what I can do now ... I'll add spread syntax to the onClick function in AccountCard

js/src/modals/ExportAccount/exportStore.js
CraigglesO: This is the spookiest bug I've come across. If you know the answer you're a god send.

If I don't iterate over the keys, mobx won't trigger an state update for react. I did it your way the first time... I don't remember where I read about this bug and I don't know what's happening under the hood. If you have advice about how to avoid it in the future I would be super appreciative.

js/src/modals/ExportAccount/exportStore.js
CraigglesO: If you go to the redux action, message is a string prop passed into ui->error->error, which is in turn passed to a material-ui component.. Should be consider handling errors differently?

js/src/modals/ExportAccount/exportAccount.js
ngotchac: Oh right. So you could use `Object.values(accounts)` if needed. Also, could use a new `const` variable/

js/src/views/Account/account.js
```diff
@@ -74,10 +74,15 @@ class Account extends Component {
   componentWillReceiveProps (nextProps) {
     const prevAddress = this.props.params.address;
     const nextAddress = nextProps.params.address;
+    const { accounts } = nextProps;
 
     if (prevAddress !== nextAddress) {
       this.setVisibleAccounts(nextProps);
     }
+
+    if (!Object.keys(this.exportStore._accounts).length) {
```
ngotchac: Might be good to whether have a getter, or rename the `_accounts` variable `accounts`. We try to use `_` for variables to shouldn't be accessed from outside of the classe

CraigglesO: Yeah, still playing... had to sit on this one. Really the best idea is to wait to create the exportStore until componentWillReceiveProps.

Why not, theres no benefit to componentWillMount

ngotchac:
Still some quirks... About the strange MobX thing, you could use something like this :

```diff
diff --git a/js/src/modals/ExportAccount/exportAccount.js b/js/src/modals/ExportAccount/exportAccount.js
index b57a8dc..6c19136 100644
--- a/js/src/modals/ExportAccount/exportAccount.js
+++ b/js/src/modals/ExportAccount/exportAccount.js
@@ -92,13 +92,18 @@ class ExportAccount extends Component {
   }
 
   renderList () {
-    let { accounts } = this.props;
+    const { accounts } = this.props;
+    const { selectedAccounts } = this.exportStore;
+
+    const accountList = Object.values(accounts)
+      .map((account) => {
+        account.checked = !!(selectedAccounts[account.address]);
 
-    const accountList = Object.values(accounts);
+        return account;
+      });
 
     return (
       <SelectionList
-        isChecked={ this.isSelected }
         items={ accountList }
         noStretch
         onSelectClick={ this.onSelect }
@@ -130,12 +135,6 @@ class ExportAccount extends Component {
     );
   }
 
-  isSelected = (account) => {
-    const { selectedAccounts } = this.exportStore;
-
-    return selectedAccounts[account.address];
-  }
-
   onSelect = (account) => {
     this.exportStore.toggleSelectedAccount(account.address);
   }
diff --git a/js/src/modals/ExportAccount/exportStore.js b/js/src/modals/ExportAccount/exportStore.js
index e3cedb6..c8af7e6 100644
--- a/js/src/modals/ExportAccount/exportStore.js
+++ b/js/src/modals/ExportAccount/exportStore.js
@@ -14,11 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import { action, observable } from 'mobx';
+import { action, computed, observable } from 'mobx';
 import FileSaver from 'file-saver';
 
 export default class ExportStore {
-  @observable canExport = false;
   @observable selectedAccount = '';
   @observable selectedAccounts = {};
   @observable passwordInputs = {};
@@ -33,6 +32,10 @@ export default class ExportStore {
     }
   }
 
+  @computed get canExport () {
+    return Object.keys(this.selectedAccounts).length > 0;
+  }
+
   @action changePassword = (event, password) => {
     const selectedAccount = this.selectedAccount;
 
@@ -61,23 +64,18 @@ export default class ExportStore {
 
   @action setSelectedAccount = (addr) => {
     this.selectedAccounts[addr] = true;
-    this.canExport = true;
   }
 
   @action toggleSelectedAccount = (addr) => {
+    const nextSelectedAccounts = { ...this.selectedAccounts };
+
     if (this.selectedAccounts[addr]) {
-      delete this.selectedAccounts[addr];
+      delete nextSelectedAccounts[addr];
     } else {
-      this.selectedAccounts[addr] = true;
+      nextSelectedAccounts[addr] = true;
     }
-    this.canExport = false;
-    Object
-      .keys(this.selectedAccounts)
-      .forEach((address) => {
-        if (this.selectedAccounts[address]) {
-          this.canExport = true;
-        }
-      });
+
+    this.selectedAccounts = nextSelectedAccounts;
   }
 
   onExport = (event) => {
```

Basically, you get the `selectedAccounts` variable from the render function, that will trigger the re-render when updating it (it seems that it is boggus otherwise). The `canExport` can be a `computed` value that gets updated whenever `selectedAccounts` is updated.

Another issue is that the export component lists the Multisig (might list also hardware) wallets. These cannot be exported and shouldn't be listed.

js/src/modals/ExportAccount/exportStore.js
jacogr: Think the this.* updates should be in a `transaction`. The bug I think you are referring to is due to the fact that state changes are generally not trigerred (unless the object explicitly setup thet way) for key changes, but rather for value changes, i.e. `this.something = { newObject: value }` updates, `this.something[newValue]` doesn't.

js/src/modals/ExportAccount/exportStore.js
```diff
@@ -0,0 +1,106 @@
...
+
+          this.accountValue = '';
+          if (event) { event(); }
+        })
+        .catch((err) => {
+          const { name, meta } = this.accounts[address];
+          const { passwordHint } = meta;
+
+          this._newError({
+            message: `[${err.code}] Account "${name}" - Incorrect password. (Password Hint: ${passwordHint})`
```
jacogr: Should rather have this string in the view or make it and explicit FormattedMessage here

jacogr:
LGTM, just one message to move.

js/src/modals/ExportAccount/exportStore.js
CraigglesO: Dunno. None of this code works anymore with the latest updates.

1) `_pollStatus TypeError: Cannot convert undefined or null to object`
2) `Unable to update callback for subscriptionId 2 TypeError: Cannot convert undefined or null to object`
3) `Warning: Failed prop type: The prop 'balances' is marked as required in 'ExportAccount', but its value is 'undefined'.`

So. It's a little more complicated then just a newError. 

Although. I tried FormattedMessage and the newError ui does not handle it. Nor multiple errors.

https://github.com/paritytech/parity/blob/master/js/src/ui/Errors/errors.js#L59-L77

CraigglesO:
All errors are fixed. As discussed on riot, I don't know how to fix newError, it's an action that takes an object and the errors ui component requires message to be a string....

PR #4974, opened 2017-03-21
Secretstore over network

svyatonik:
PR outline:
- [x] encryption protocol works over network
- [x] encryption results are stored in KS database
- [x] decryption protocol works over network
- [x] polish code

Open questions (please comment if you have another opinion/if I've misunderstood requirements):
1) network protocol is implemented with assumption that a set of KeyServers (aka cluster) is preconfigured and every KS knows IP + Public key of every other KS. It is arguable, but anyway:
1.1) removing KS from a cluster is a dangerous operation, since after it, some secrets could be lost (if threshold is < cluster.size())
1.2) adding KS to a cluster is a dangerous operation, since if someone could add > threshold malicious KS-es, then these later could be used to restore secrets
...So for me cluster configuration is a heavy 'administrative' operation, which is possibly out-of-scope of ours implementation. So the question is - is it enough to provide CLI-option to provide file with cluster configuration?
2) at every moment some nodes of cluster could be down (or in other network segment). So if someone tries to generate key with scheme 3-of-N in that moment, total number of nodes (N) could be less than the cluster size => only portion of KS-es could serve decryption request later.
3) any error, which occurs during encryption/decryption session on one of the nodes leads to error in session. So if connection is lost to one of nodes, or it fails to serve requests in given timeout, session fails.
4) there are two network ports open at each KS-node: one for serving client HTTP requests (like GenerateDocumentKey or RestoreDocumentKey) and the second one is for this custom protocol. The reason is that enc/dec protocol are not stateless && I don't want to implement such type of protocols on the top of HTTP.
5) all connections messages (except the handshake-messages) are encrypted (**this is not yet done**)

Running example:
Currently cluster configuration is hardcoded in rs file, so the following is the only way to run DKG from command line:
1) start 3 KS:
```
RUST_LOG=secretstore=trace,secretstore_net=trace ./parity -d db.ropsten_ss1 --no-dapps --no-jsonrpc --no-ui --warp --chain ropsten --port 30303 --secretstore-port 8082
RUST_LOG=secretstore=trace,secretstore_net=trace ./parity -d db.ropsten_ss2 --no-dapps --no-jsonrpc --no-ui --warp --chain ropsten --port 30304 --secretstore-port 8083
RUST_LOG=secretstore=trace,secretstore_net=trace ./parity -d db.ropsten_ss3 --no-dapps --no-jsonrpc --no-ui --warp --chain ropsten --port 30305 --secretstore-port 8084
```
2) ask KS to generate key using scheme 2-of-N, where N (after cluster connections are established) is equal to 3:
```
curl -X POST http://localhost:8082/0000000000000000000000000000000000000000000000000000000000000001/a199fb39e11eefb61c78a4074a53c0d4424600a3e74aad4fb9d93a26c30d067e1d4d29936de0c73f19827394a1dd049480a0d581aee7ae7546968da7d3d1c2fd01/1
```

**commits before b75846c are already in master**


svyatonik:
Network code is based on @debris p2p implementation, as I've spent a lot of time learning how it works :)

secret_store/src/key_server.rs
```diff
@@ -14,111 +14,207 @@
...
-pub struct KeyServerImpl<T: AclStorage, U: KeyStorage> {
-	acl_storage: T,
-	key_storage: U,
+pub struct KeyServerImpl {
+	data: Arc<Mutex<KeyServerCore>>,
 }
 
-impl<T, U> KeyServerImpl<T, U> where T: AclStorage, U: KeyStorage {
+unsafe impl Send for KeyServerImpl {}
+unsafe impl Sync for KeyServerImpl {}
```
rphmeier: Seems: a) unnecessary, and b) very dangerous!

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -27,13 +59,844 @@ pub trait Cluster {
...
+	data: Arc<ClusterData>,
+}
+
+/// Network cluster view. It is a communication channel, required in single session.
+pub struct ClusterView {
+	core: Arc<Mutex<ClusterViewCore>>,
+}
+
+unsafe impl Send for ClusterView {}
+unsafe impl Sync for ClusterView {}
```
rphmeier: Again: very dangerous!

rphmeier:
As far as I can see, there is nothing in the code/problem specification which requires unsafe code.

Unsafe code should only be used in very specific circumstances, and only with a type-safe wrapper around it. Remember that it pollutes an entire module, so you have to protect at module boundaries.

I don't see the motivation or justification for any unsafety here. It doesn't look like it's necessary for optimization. Is this heavily optimized code? Which optimizations are the unsafe bits necessary for?

rphmeier:
To be clear, we should have a moratorium on new `unsafe` code being introduced to the codebase at all. Our goal here is security, and we can do everything we need to without unsafe code. No need to introduce the risk of subtle data races.

NikVolf:
second hand for `unsafe impl Sync/Send`
it's only suitable for rare cases (bindings for external c/c++ libs that are known to manage synchronization)

rphmeier:
I pulled this down to review it better; it builds and tests fine without the unsafe impls. I hope you don't mind that I removed them.

I'll take a close look now at the logic.

secret_store/src/key_server_cluster/io/handshake.rs
```diff
@@ -0,0 +1,320 @@
...
+					shared_key: self.shared_key.clone().expect("shared_key is filled in Send/ReceivePublicKey; ReceivePrivateKeySignature follows Send/ReceivePublicKey; qed"),
+				}))))
+			},
+			HandshakeState::Finished => panic!("poll Handshake after it's done"),
+		};
+
+		self.state = next;
+		match result {
+			// by polling again, we register new future
+			Async::NotReady => self.poll(),
```
rphmeier: couldn't this lead to unbounded recursion? If we're waiting on some network response, we should just let the event loop decide when to poll again.

secret_store/src/key_server_cluster/io/read_header.rs
```diff
@@ -0,0 +1,44 @@
...
+use std::io;
+use futures::{Future, Poll, Async};
+use tokio_core::io::{ReadExact, read_exact};
+use key_server_cluster::Error;
+use key_server_cluster::io::message::{MESSAGE_HEADER_SIZE, MessageHeader, deserialize_header};
+
+/// Create future for read single message header from the stream.
+pub fn read_header<A>(a: A) -> ReadHeader<A> where A: io::Read {
+	ReadHeader {
+		reader: read_exact(a, vec![0; MESSAGE_HEADER_SIZE]),
```
rphmeier: Could avoid the allocation with a `[u8; 4]`

secret_store/src/key_server_cluster/io/message.rs
```diff
@@ -0,0 +1,247 @@
...
+	/// Message payload size (without header).
+	pub size: u16,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+/// Serialized message.
+pub struct SerializedMessage(Vec<u8>);
+
+impl Deref for SerializedMessage {
+	type Target = Vec<u8>;
```
rphmeier: ought to be `type Target = [u8]` as most generic code expects this and you don't gain anything but `.capacity()` from `Vec<u8>`

secret_store/src/key_server_cluster/io/read_message.rs
```diff
@@ -0,0 +1,86 @@
...
+				let (read, payload) = try_ready!(future.poll());
+				(ReadMessageState::Finished, Async::Ready((read, payload)))
+			},
+			ReadMessageState::Finished => panic!("poll ReadMessage after it's done"),
+		};
+
+		self.state = next;
+		match result {
+			// by polling again, we register new future
+			Async::NotReady => self.poll(),
```
rphmeier: Same here: unbounded recursion until ready?

secret_store/src/key_server_cluster/io/handshake.rs
rphmeier: I see, the `try_ready!` calls will (probably) prevent that from happening.

secret_store/src/key_server_cluster/io/deadline.rs
```diff
@@ -0,0 +1,85 @@
...
+	let timeout = try!(Timeout::new(duration, handle)).map(|_| DeadlineStatus::Timeout).boxed();
+	let future = future.map(DeadlineStatus::Meet).boxed();
+	let deadline = Deadline {
+		future: timeout.select(future),
+	};
+	Ok(deadline)
+}
+
+#[derive(Debug, PartialEq)]
+/// Deadline future completion status.
```
rphmeier: Also, just style-wise but not super important: the rest of the codebase puts derives/attributes under docs.

secret_store/src/key_server.rs
```diff
@@ -14,111 +14,204 @@
...
-	key_storage: U,
+pub struct KeyServerImpl {
+	data: Arc<Mutex<KeyServerCore>>,
 }
 
-impl<T, U> KeyServerImpl<T, U> where T: AclStorage, U: KeyStorage {
+/// Secret store key server data.
+pub struct KeyServerCore {
+	close: Option<futures::Complete<()>>,
+	_handle: Option<thread::JoinHandle<()>>,
```
rphmeier: The [JoinHandle](https://doc.rust-lang.org/std/thread/struct.JoinHandle.html) doesn't join the thread on `Drop`. If you want that behavior, you'll have to trigger it manually. If not, then it's not necessary to keep around.

secret_store/src/key_server.rs
```diff
@@ -14,111 +14,204 @@
...
+			allow_connecting_to_higher_nodes: config.allow_connecting_to_higher_nodes,
+			encryption_config: config.encryption_config.clone(),
+			acl_storage: acl_storage,
+			key_storage: key_storage,
+		};
+
+		let (stop, stopped) = futures::oneshot();
+		let (tx, rx) = mpsc::channel();
+		let handle = thread::spawn(move || {
+			let mut el = Core::new().expect("Creating an event loop should not fail.");
```
rphmeier: Can this error be propagated upwards instead of panicking?

rphmeier:
A couple questions and style grumbles, but overall looks great! The futures stuff looks really well done, very clean and seems to tend to avoid boxing egregiously.

Some comments on the open questions:

> is it enough to provide CLI-option to provide file with cluster configuration

A bunch of CLI options could be fine too, then the user can just set up a parity `config.toml` to configure it correctly and doesn't have to manage multiple configuration files.

> if someone tries to generate key with scheme 3-of-N in that moment

Wouldn't it be better to fail/block until all nodes in the cluster are up? A user requesting a key from the cluster wants one from the whole cluster, not to silently get one from whatever subset of nodes are currently up. In practice, the whole cluster should be up almost all of the time.

> So if connection is lost to one of nodes, or it fails to serve requests in given timeout, session fails.

Does this mean that if I'm trying to decrypt and one of the connections fails then the decryption code just gives up? Ideally, it would retry/move on but that can be beyond the scope of this PR.

> the second one is for this custom protocol. 

That makes sense to me. Do you think it might be possible to abstract it over _any_ `T: Stream` (or maybe Box<Stream> for heterogeneous peers) instead of just a TCP stream? That would make things really nice if we wanted to integrate it with IPFS libp2p once that work is closer to completion. 
Then the message level encryption in point 5 would just be done at the transport level and not at the protocol level.
Is it worth doing in your opinion?

svyatonik:
Thanks for taking your time to review :) I'll take a look at your code comments a bit later. For now, about general questions:
1) the problem is that I must provide mapping { Public: Ip+Port } in command line && I do not know the number of entries in the map (in theory it is unlimited), so I suppose this particular part should reside in file, but the filename itself must be provided via CLI-option. Or maybe force users to use config.toml when they are starting secretstore (of course if config.toml is suitable for storing mappings)?
Another point of question - is it ok to provide this information via CL at all? I mean is it static, or dynamic? I do not know exact requirements, but I suppose that cli-implementation (== this info is static) *at least for now* it is ok
2) actually, the user expects to get key from any **t**-of-N cluster nodes (not the whole cluster or some specific nodes). So t is more important for him than N. But I also doubt that silently generating key when there are M nodes (M < N) is ok. So maybe your solution is better than mine, but it follows that 'maintenance windows' for cluster nodes = 'maintenance windows' for the whole cluster. One node is down => whole cluster is down. So maybe it is better still to generate key, but show current cluster state in GUI? Like "warning: only 5 of 7 nodes are up".
3) exactly - the whole decryption session will fail. I agree - it is arguable - is it better to fail immideately or retry. Also please keep in mind that the whole this process runs as a response to HTTP request and response = result of this process (distributely generated key). So there are also HTTP timeouts => maybe it is better to fail && retry on upper level to make UI more responsive (some indicator like "Last attempt failed with 'connection lost'. Retrying 2/3.").
4) I do not know anything about IPFS p2p now :) But I suppose it is easy to refactor it, as the only point where TCP is involved is establishing connection. All other code just works over io::Read, io::Write (which is TcpStream in this case). Maybe there are also couple of fields, used for debugging (like ip address is used for tracing), but it is not crucial.

secret_store/src/key_server_cluster/io/handshake.rs
debris: this won't lead to unbound recursion if `state(n) = poll(state(n-1))`

Without this additional `self.poll()` this future would never be polled again after returning `Async::NotReady`. And as you noticed `try_ready!` prevents `self.poll()` to be called more than once for the same state 😄

gavofyork:
1. yeah, for now static is fine.

2. if one node is down, that presumably means that that node is useless in later a later decryption operation, and so "M of N" actually becomes "M of (N-1)"? if so, keys should only be generated when all nodes are operational otherwise the integrity of the whole system is put into question.

we can introduce failover mechanisms for production servers - i don't think "node maintenance"  should mean new keys cannot be created.

3. since only M of the N nodes are required for the operation to succeed, it should keep retrying until it becomes clear that there are fewer than M functional nodes.

4. i'm sure eventual IPFS integration will be easy if it's just reading and writing to a socket.

svyatonik:
@rphmeier Thanks for the review :) Fixed all issues, except for documentation/attributes placement - I was using this pattern from the very beginning && there are now a lot of places where it is in this order (ethstore/ethcore). I will (later) make a script to change this order automatically && in next will try to use opposite pattern in my next PRs

secret_store/src/key_server_cluster/io/message.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+/// Encrypt serialized message.
+pub fn encrypt_message(_key: &Secret, message: SerializedMessage) -> Result<SerializedMessage, Error> {
+	Ok(message) // TODO: implement me
+}
+
+/// Decrypt serialized message.
+pub fn decrypt_message(_key: &Secret, payload: Vec<u8>) -> Result<Vec<u8>, Error> {
+	Ok(payload) // TODO: implement me
```
debris: are these `TODO` still needed?

secret_store/src/key_server_cluster/io/read_payload.rs
```diff
@@ -0,0 +1,64 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::io;
+use futures::{Poll, Future};
+use tokio_core::io::{read_exact, ReadExact};
```
debris: would be good to use `tokio_io::io` instead

secret_store/src/key_server_cluster/io/write_message.rs
```diff
@@ -0,0 +1,70 @@
...
+		error: error,
+		future: future,
+	}
+}
+
+/// Write encrypted message to the channel.
+pub fn write_encrypted_message<A>(a: A, key: &Secret, message: Message) -> WriteMessage<A> where A: io::Write {
+	let (error, future) = match serialize_message(message)
+		.and_then(|message| encrypt_message(key, message))
+		.map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{}", e))) {
```
debris: `e.to_string()` is shorter and more idiomatic than `format!("{}", e)`

secret_store/src/key_server_cluster/io/write_message.rs
```diff
@@ -0,0 +1,70 @@
...
+use futures::{Future, Poll};
+use tokio_core::io::{WriteAll, write_all};
+use ethkey::Secret;
+use key_server_cluster::message::Message;
+use key_server_cluster::io::{serialize_message, encrypt_message};
+
+/// Write plain message to the channel.
+pub fn write_message<A>(a: A, message: Message) -> WriteMessage<A> where A: io::Write {
+	let (error, future) = match serialize_message(message)
+		.map_err(|e| io::Error::new(io::ErrorKind::Other, format!("{}", e))) {
```
debris: `e.to_string()` is shorter and more idiomatic than `format!("{}", e)`

parity/secretstore.rs
```diff
@@ -64,10 +66,35 @@ mod server {
...
+					threads: 4,
+					self_private: (***key_pairs[(conf.port - 8082) as usize].secret()).into(),
+					listener_address: ethcore_secretstore::NodeAddress {
+						address: conf.interface.clone(),
+						port: conf.port + 10,
+					},
+					nodes: key_pairs.iter().enumerate().map(|(i, kp)| (kp.public().clone(),
+						ethcore_secretstore::NodeAddress {
+							address: conf.interface.clone(),
+							port: 8082 + 10 + (i as u16),
```
svyatonik: Yes - it is decribed couple of lines above :)
```
// TODO: this is test configuration. how it will be configured in production?
```
It is question №1 in this PR. Basically - I didn't knew whether configuration should be dynamic or static. But still wanted to make some kind of demonstration in real environment (not in tests). So here's test configuration for 3 nodes, which have 2 ports with diff = 10.
Now as I know it must be static (at least for now), I have already removed this test config in my local branch, but still waiting for this PR to be closed before making next one :)

secret_store/src/key_server_cluster/io/message.rs
svyatonik: yes

secret_store/src/key_server_cluster/io/read_payload.rs
svyatonik: Can it be one of next PRs - I have already a bunch of changes done for my next PR (like config, encryption, 'shadow encryption' scheme) :) Are you ok with that?

parity/secretstore.rs
```diff
@@ -64,10 +66,35 @@ mod server {
...
-				data_path: conf.data_path,
+				listener_address: ethcore_secretstore::NodeAddress {
+					address: conf.interface.clone(),
+					port: conf.port,
+				},
+				data_path: conf.data_path.clone(),
+				// TODO: this is test configuration. how it will be configured in production?
+				cluster_config: ethcore_secretstore::ClusterConfiguration {
+					threads: 4,
+					self_private: (***key_pairs[(conf.port - 8082) as usize].secret()).into(),
```
svyatonik: In future KeyServer will be an IPC module. I've found that types which are passed to IPC modules  are marked with #[binary] attribute. So I did this also :) And when I'm adding a `Secret` to such type it compiles with an error. 
Maybe I'm wrong && I could simply remove this attribute from `Configuration` :) It would be great if you could help me with this :)

svyatonik:
@debris Thank for the review :) Will upload changes in a minute

@rphmeier @gavofyork I still have doubts about autodecreasing N in M-of-N scheme. I still think it is ok to decrease N, since:
1) single malicious node could break DKG protocol just by ignoring messages or sending incorrect replies. And if we can't 'ban' this node (at least for this session), that would mean that key generation procedure won't be completed at all. Ban means autodecrease N here;
2) autodecreasing N would have negative effects only if this node is excluded from cluster && after excluding number of active nodes, holding the same secret is < M. The secret will be lost then. But this could also happen in normal scenario => excluding node from cluster is a very serious admin-action.
Last note on this - ECDKG protocol itself is able to autodecrease N, when one of nodes is working in improper way.

@gavofyork "since only M of the N nodes are required for the operation to succeed, it should keep retrying until it becomes clear that there are fewer than M functional nodes" - so do you think that if the node is only connected to < M nodes (and all other connections have failed), it is the sign that there are < than M functional nodes. Or are you suggesting to introduce some kind of timeout for this case? I mean - no connection - try to connect for anther minute && then finally inform user that you're unable to connect.
Another example if at the beginning there were connections to N nodes, then we have selected M nodes to restore the secret && then (in the process of restoring) one of the M nodes goes down. Do you want to autorestart decryption process? Or fail with an error? If autorestart - what is criteria for stopping autorestarts? Timeout? Or we should never stop autorestarting && again - a couple of malicious nodes could fail all decryption sessions?
I do not like retry-until-timeout idea on protocol level, but still do not see another solution for "should keep retrying until it becomes clear..."

secret_store/src/key_server_cluster/io/read_payload.rs
debris: sure!

gavofyork:
A algorithm would be:
1. select M unmarked nodes from N
2. attempt secret de-/re-crypt
3. if failed:
   - mark the unresponsive node
   - if >= M nodes unmarked then go to 1
   - else return FAIL
4. return OK

svyatonik:
OK - no timeouts => that's cool. The rest is currently TODO :)
Still have doubts about auto-decreasing N, though :) Why it is bad to share the secret among a small subset of nodes - is it because of couple of nodes can go down => secret is irrecoverable lost? I see no other reasons. But anyway - you guys are insisting => I'll add (in next PRs) constant to encryption session, so it will be not simply M-of-N, but N-of-L where L is >= N && <= M. For current project I'll add constraint L = M => encryption session will fail if at least 1 node is down. But I suppose it will be useful for potential usage of SecretStore in future + it won't add any additional complexity, except that in 2 places comparison expression will slightly change.

gavofyork:
so the reason it's M-of-N in the first place is redundancy: it's so that you can have `N-M` nodes fail and you'll still be able to recover the secret in a time-critical emergency. this is absolutely foundational to the system, since we don't provide any other means of secret recovery - there are no fallbacks here.

encryption (i.e. secret insertion) is typically not an operation you'd do in a time-critical emergency since you already have the plain text to use. furthermore by "auto-reducing" N, you're actually "auto-reducing" the redundancy.

to give you a simple example, if we have M=2, N=3, the system is designed to be resilient against the situation where one node is Byzantine. if the evil node managed to DoSed one of the other 2 during a victim's secret-insertion operation with auto-reduce-N in place, the operation would apparently succeed (and thus we can no longer assume the plain text exists), yet the secret is now held under a 2-of-2 scheme, meaning the evil node can hold it to random.

in short, if we state it's M-of-N, and the secret-insertion operation succeeds, then as part of the specification, it must be possible to /eventually/ recover the secret given that `N-M` nodes fail indefinitely.

gavofyork:
ok - so are there any other changes to go in?

svyatonik:
I do not want to put any changes here, as ther's already another PR based on this + 3rd one prepared. So maybe just some changes after @debris review

PR #4976, opened 2017-03-21
Try to fix WS race condition connection

ngotchac:
@arkpar found some issues while switching network on Windows (the saved Signer Token was detected as a wrong token by the UI). This is an attempt to fix this, but has to be tested since I cannot reproduce this issue...


arkpar:
Gave it a test and can confirm that the issue is fixed.

Issue #4886, opened 2017-03-13
[Chrome Extension] Signer bar can't be moved


bjornwgnr:
Signer bar seems to be fixed to the bottom right unlike the drag&drop solution in the parity browser.

tomusdrw:
The difficulty is that the signer bar is in an iframe, so the process will be a bit complicated:
1. On mousedown generates the event to upper frame that iframe should be expanded to entire viewport
2. Moving the sidebar is allowed now (within the iframe)
3. On mouseup generates another event to fix the position of the iframe (after the animation is done)

ngotchac:
Couldn't it be (a bit hacky) : two states for the iframe : open & close
Open should take the whole viewport. When closing, just check with JS the position and dimensions of the content and then style the iframe accordingly.

tomusdrw:
@ngotchac Could you elaborate? I'm not sure if I understand this.
What would happen if we want to re-position the signer bar (let's assume it's minimized (closed state?)) and user starts to drag it?

ngotchac:
So the idea would be : 

- The Signer Bar would emit two events : `expand` and `shrink`
These events would be emitted when the user starts dragging the Signer Bar, when the Signer opens to sign transactions, when the user wants to change the default account, etc...
- The iframe handler would respond to the `expand` event by giving the iframe full width and full height
- The iframe handler would respond to the `shrink` event by waiting for let's say 200ms for the Signer Bar to shrink to its correct width/height and move to the wanted position (while the iframe is still with full width and height). Then the iframe handler would query the position and the sizes of the iframe's content, and set the iframe position and size accordingly.

tomusdrw:
So the problem is that if iframe is expanded to full width and height you cannot click anything on the website. We currently emit those events when signer bar is being opened/closed - but they are just expanding the iframe to required size.
So my proposal is re-using the old events and also adds some new ones
1. On events: `signer_bar::open` & `signer_bar::close` set the size to "large" and "small" in cradle
2. On events `signer_bar::drag_start` & `signer_bar::position` & `signer_bar::drag_stop`: maximize iframe, remember new position of an iframe, set iframe size & position to "small".

ngotchac:
That would work but seems a bit complicated no ? Is it a big issue if the website cannot get interactions from the user when the Signer Bar is expanded ? It's more like when you open a modal, you cannot interact with the background. This would be the same, and would make this easier to implement and maintain.

tomusdrw:
[...]

PR #4987, opened 2017-03-21
Dispatch an open event on drag of Parity Bar

ngotchac:
This is needed for #4886

The first step is the get the  size and position of the Parity Signer Bar in the chrome extension right. This is done here paritytech/parity-extension#68

### DIFF
--- a/js/src/views/ParityBar/parityBar.js
+++ b/js/src/views/ParityBar/parityBar.js
@@ -78,15 +78,23 @@ class ParityBar extends Component {
 
     // Hook to the dapp loaded event to position the
     // Parity Bar accordingly
-    DappsStore.get(api).on('loaded', (app) => {
-      this.app = app;
-      this.loadPosition();
-    });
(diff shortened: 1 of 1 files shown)


js/src/views/ParityBar/parityBar.js
```diff
@@ -647,6 +646,13 @@ class ParityBar extends Component {
         return DEFAULT_POSITION;
     }
   }
+
+  setMovingState (moving, extras = {}) {
+    this.setState({ moving, ...extras });
+
+    // Trigger an open event if it's moving
+    this.dispatchOpenEvent(moving);
```
tomusdrw: It might be useful to distinguish between open for moving and open in the extension in future (for instance to use different animation).

ngotchac:
Updated the PR ! Seems to work ok. To test :

Run NODE_ENV=production npm run build && node webpack/build.server.js for the UI
Set 127.0.0.1:3000 in the Extension options

PR #4988, opened 2017-03-21
Strict validation transitions

keorn:


ethcore/src/engines/authority_round.rs
```diff
@@ -64,6 +66,7 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 			block_reward: p.block_reward.map_or_else(U256::zero, Into::into),
 			registrar: p.registrar.map_or_else(Address::new, Into::into),
 			start_step: p.start_step.map(Into::into),
+			validate_score_transition: p.validate_score_transition.map_or(0, Into::into),
```
arkpar: Does the default at 0 break compatibility with existing PoA chains?

arkpar:
Needs a rebase

ethcore/src/engines/authority_round.rs
keorn: Only ones started with 1.5, but the default eip98 transition breaks them anyway.

PR #4999, opened 2017-03-22
EIP-213 (bn128 curve operations)

NikVolf:
- utilizes modified version of `bn` lib
- there maybe some minor alteration and correctness tests added later once EIP gets more specified
- also introduces fail-ability for built-ins (since this operations are first ones of the kind)


ethcore/src/builtin.rs
```diff
@@ -102,7 +111,10 @@ impl Builtin {
 	pub fn cost(&self, input: &[u8]) -> U256 { self.pricer.cost(input) }
 
 	/// Simple forwarder for execute.
-	pub fn execute(&self, input: &[u8], output: &mut BytesRef) { self.native.execute(input, output) }
+	pub fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), Error> { 
+		self.native.execute(input, output)?;
+		Ok(())
```
rphmeier: body can be unchanged: `self.native.execute()`

ethcore/src/builtin.rs
```diff
@@ -294,9 +321,77 @@ impl Impl for ModexpImpl {
...
+		if px == Fq::zero() && py == Fq::zero() {
+			G1::zero()
+		} else {
+			AffineG1::new(px, py).map_err(|_| Error::from("Invalid curve point"))?.into()
+		}
+	)
+}
+
+impl Impl for Bn128AddImpl {
+	fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), Error> {
```
rphmeier: Given that these are fallible, it makes sense to document the failure conditions.

ethcore/Cargo.toml
```diff
@@ -45,6 +45,7 @@ ethcore-bloom-journal = { path = "../util/bloom" }
 hardware-wallet = { path = "../hw" }
 stats = { path = "../util/stats" }
 num = "0.1"
+bn = { git = "https://github.com/paritytech/bn" }
```
rphmeier: any plans to upstream these changes?

NikVolf: yep after i'm done with pairing

arkpar:
We should not really use "alt_" prefix anywhere. As far as I can tell this is only used in libsnark to distinguish two implementations of the same curve apart.

NikVolf:
@arkpar done

ethcore/src/evm/evm.rs
```diff
@@ -70,6 +73,12 @@ impl From<Box<trie::TrieError>> for Error {
 	}
 }
 
+impl From<builtin::Error> for Error {
+	fn from(err: builtin::Error) -> Self {
+		Error::BuiltIn(err.0)
```
rphmeier: Why not just have builtins return a `String` as error rather than go through the trouble of making a `builtin::Error` type which is just destructured here? Not sure it's worth allocating the string at all, actually, since
  - it's easy to make builtins fail
  - builtins are on a hot path
  - errors are silently swallowed anyway

gavofyork:
@NikVolf worth addressing @rphmeier 's suggestion now? or another PR?

ethcore/src/evm/evm.rs
NikVolf: Yeah, makes sense

ethcore/src/executive.rs
```diff
@@ -276,7 +276,7 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 
 			let cost = builtin.cost(data);
 			if cost <= params.gas {
-				builtin.execute(data, &mut output);
+				builtin.execute(data, &mut output)?;
```
tomusdrw: This can fail without `discarding` or `reverting_to_checkpoint`.

Currently we don't have any state-altering builtins, but might be good to have that in mind. Also I'm not entirely sure if one leaking state snapshot would ever be a problem, but IMHO would be good to address that.

NikVolf: not sure i follow
why shouldn't it fail the same way as not enough gas fails, but with different reason?

tomusdrw: I think every other exit from this function is either doing: `state.discard_snapshot()`, `state.revert_to_snapshot()` or `enact_result` (which does this stuff internally afair).

This could exit the function after creating `state.checkpoint()` but without reseting it. Also seems that tracing for top-level failed builtin call might be missing.

NikVolf: You are right, I misjudged the program flow here
the errors and rollback/discard logic should be decoupled from each other though

ethcore/src/executive.rs
```diff
@@ -276,25 +276,31 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
-
-				// trace only top level calls to builtins to avoid DDoS attacks
-				if self.depth == 0 {
-					let mut trace_output = tracer.prepare_trace_output();
-					if let Some(mut out) = trace_output.as_mut() {
-						*out = output.to_owned();
+				if let Err(e) = builtin.execute(data, &mut output) {
+					self.state.revert_to_checkpoint();
+					let evm_err: evm::evm::Error = e.into();
+					tracer.trace_failed_call(trace_info, vec![], evm_err.clone().into());
```
tomusdrw: One last tiny grumble: We're tracing only calls on `self.depth == 0 to avoid DoS attacks for tracing nodes, probably worth doing the same check here.

NikVolf: But out-of-gas is tracing error on any `self.depth` below?
Which seems right, because you never know what caused contract to fail otherwise

tomusdrw: Fair enough, probably the problematic part is `trace_output` anyway.

Issue #5001, opened 2017-03-22
Make Dapps server light-client friendly


rphmeier:
Specifically regarding the registrar: the light client can make calls to it, but they may take an indeterminate amount of time to resolve. 

Maybe `hash_fetch::urlhint::ContractClient` should yield futures?
Will be optimized by having the light client attempt to cache all storage of certain contracts, but that's not implemented yet.

cc @tomusdrw

tomusdrw:
Returning `Futrue` sounds good although it might complicate dapps router (if using our current `hyper` version; next version is future-based), but it's doable :)

PR #5002, opened 2017-03-22
Quick'n'dirty CLI for the light client

### LINKED ISSUE #5012: Poll nonces for light client transaction queue senders periodically
...otherwise their transactions will never be culled from the queue

rphmeier:
Summary of changes:
  - Use database in the light client (particularly relevant to `header_chain.rs`)
  - Add column to the database specifically for the light client.
  - Add `execute_light` function to `parity/run.rs`, triggered on `--light` flag.
Runs RPC servers, but no DApps yet: #5001 
  - Transaction propagation and transaction queue culling.
  - Various minor enhancements to the dispatch/request credits mechanism
  - Fixing some message encoding issues
  - Serve light clients by default, opt out with `--no-serve-light`.

Next needed changes:
  - Integration with main informant.
  - Filters and other missing RPCs needed by the UI
  - Better dispatch mechanism for `on_demand`
  - Better defaults for request credits: 10 headers/sec/peer is too slow if there's only one peer!

Closes #5012


parity/cli/usage.txt
```diff
@@ -70,6 +70,11 @@ Operating Options:
   --keys-path PATH               Specify the path for JSON key files to be found
                                  (default: {flag_keys_path}).
   --identity NAME                Specify your node's name. (default: {flag_identity})
+  --light                        Experimental: run in light client mode. Light clients
+                                 synchronize a bare minimum of data and fetch necessary
+								 data on-demand from the network. Much lower in storage,
```
gavofyork: spaces should be used here to guarantee output alignment.

parity/cli/usage.txt
```diff
@@ -70,6 +70,11 @@ Operating Options:
...
+								 potentially higher in bandwidth. Has no effect with
+								 subcommands (default: {flag_light})
```
gavofyork: Don't forget the full-stop `.`.

gavofyork:
is this really 1,522 −388 ? seems a lot for the documented additional functionality...

rphmeier:
It's about +450 for the header chain stuff, +150 for the transaction propagation, +700 for the CLI (generalizing RPC APIs among other things), and then about +200 for miscellania to make the UI function a bit better.

ethcore/light/src/client/header_chain.rs
```diff
@@ -75,42 +81,142 @@ impl HeapSizeOf for Entry {
...
+		let canon_hash = candidates[0].hash;
+		Ok(Entry {
+			candidates: candidates,
+			canonical_hash: canon_hash,
+		})
+	}
+}
+
+fn cht_key(number: u64) -> String {
+	format!("canonical_{}", number)
```
arkpar: Key comparison should work faster if you put the number in the front. Might be also a good idea to use a 4-byte big endian representation instead, so that enumeration would produce a sorted list.

ethcore/light/src/client/header_chain.rs
```diff
@@ -230,7 +390,7 @@ impl HeaderChain {
 					best.hash
 				};
 
-				self.headers.read().get(&hash).cloned()
+				load_from_db(hash)
```
arkpar: Probably worth caching the best header in memory.

rphmeier: The DB should do this automatically, right? I plan to add customized caching in the future but I think it's beyond the scope of this PR.

ethcore/light/src/client/header_chain.rs
rphmeier: Are the keys compared during regular lookup? Also, is there a minimum length on keys?

arkpar: > Are the keys compared during regular lookup?

Pretty sure they are. Database basically performs a binary search.

> Also, is there a minimum length on keys? 

I believe there is no such minimum.

ethcore/light/src/client/header_chain.rs
arkpar: Happy if this is added later. The closer is the cache to the code that uses it, the better is performance. Introducing multiple levels of caching can bring huge benefits. Higher level cache usually holds less data but is accessed much more frequently. CPU hardware design generally follows this principle and it often applies to the software as well.

rphmeier: Logged in https://github.com/paritytech/parity/issues/5398

rphmeier:
@arkpar headerchain grumbles addressed

PR #5015, opened 2017-03-24
On-chain ACL checker for secretstore

svyatonik:
on top of #4974 
@keorn as agreed yesterday - here's the chain-based AclChecker prototype. Are you going to develop the contract itself? I'll need a contract ABI to generate something like this:
https://github.com/paritytech/parity/blob/master/ethcore/src/miner/service_transaction_checker.rs#L58
Also - are we going to deploy registry contract && get ACL checker contract address from there, or would it be hardcoded/cli-option address?


gavofyork:
would prefer to have the registry throughout.

svyatonik:
Checked success && failure cases on Kovan

rphmeier:
LGTM; I wonder if we could put the contract into the `native_contracts` crate at some point?

svyatonik:
I thought about it, but as I understand it is for 'core' functionality, while this one is secretstore-specific && secretstore itself is very specific. So do not know if this is a good idea to.
Maybe it would be better to use macros from `native_contracts` to generate this file?

gavofyork:
i think it's fine to have in there. it's not a huge maintenance burden and it's super convenient.

svyatonik:
ok, will move there

svyatonik:
done

rphmeier:
> it is for 'core' functionality

I meant it to be a bin for all contracts we use from Parity (which I suppose includes this)

svyatonik:
Ok, got it :) I thought that, because it doesn't include updater contract. But I suppose it is in TODO list.

PR #5018, opened 2017-03-24
Resilient warp sync

arkpar:
On a slow connection it still timeouts trying to download a 16mb chunk. This will be fixed in another PR.


Cole128:
The same thing is happening to me still. Running parity was going fine, and then it started downloading all the blocks one by one.

Cole128:
So now I got this error while syncing the snapshots.

```
2018-04-10 20:48:36  Encountered error during state restoration: Chunk size is too large.
2018-04-10 20:48:36  Encountered error during state restoration: Chunk size is too large.
2018-04-10 20:48:39  Failed to initialize snapshot restoration: IO error: While lock file: /home/colecmac/.local/share/io.parity.ethereum/chains/ethereum/db/906a34e69aec8c0d/snapshot/restoration/db/LOCK: No locks available
```

5chdn:
@Cole128 see #7177

PR #5019, opened 2017-03-24
Fine grained snapshot chunking

arkpar:


ethcore/src/snapshot/mod.rs
```diff
@@ -372,8 +374,10 @@ pub fn chunk_state<'a>(db: &HashDB, root: &H256, writer: &Mutex<SnapshotWriter +
 
 		let account_db = AccountDB::from_hash(db, account_key_hash);
 
-		let fat_rlp = account::to_fat_rlp(&account, &account_db, &mut used_code)?;
-		chunker.push(account_key, fat_rlp)?;
+		let fat_rlps = account::to_fat_rlps(&account, &account_db, &mut used_code, PREFERRED_CHUNK_SIZE)?;
+		for (i, fat_rlp) in fat_rlps.into_iter().enumerate() {
+			chunker.push(account_key.clone(), fat_rlp, i > 0)?;
```
rphmeier: why force chunks on subsequent accounts?

arkpar: So that we would only have to store a few intermediate storage roots during restoration in memory and avoid expensive db lookups for each account.

ethcore/src/snapshot/account.rs
```diff
@@ -62,53 +63,50 @@ impl CodeState {
...
+	const AVERAGE_BYTES_PER_STORAGE_ENTRY: usize = 47;
 	if acc == &ACC_EMPTY {
-		return Ok(::rlp::NULL_RLP.to_vec());
+		return Ok(vec![::rlp::NULL_RLP.to_vec()]);
 	}
 
 	let db = TrieDB::new(acct_db, &acc.storage_root)?;
 
-	let mut pairs = Vec::new();
+	let pair_chunks: Vec<Vec<_>> = db.iter()?.chunks(preferred_size / AVERAGE_BYTES_PER_STORAGE_ENTRY).into_iter().map(|chunk| chunk.collect()).collect();
```
rphmeier: Using a constant here seems tenuous, prone to change, and hard to make the spec match. Couldn't we just pass a size (= remaining space in chunk) and a storage seek target?

`collect` here is an unnecessary allocation considering it's just turned `into_iter` in the next line.

ethcore/src/snapshot/account.rs
```diff
@@ -62,53 +63,50 @@ impl CodeState {
...
-				warn!("code lookup failed during snapshot");
-				account_stream.append(&false).append_empty_data();
+		let pairs_rlp = stream.out();
+
+		let mut account_stream = RlpStream::new_list(5);
+		account_stream.append(&acc.nonce)
+					  .append(&acc.balance);
+
+		// [has_code, code_hash].
+		if acc.code_hash == SHA3_EMPTY {
```
rphmeier: code section needs to be altered, otherwise we'll write duplicate code.

rphmeier:
A couple grumbles, and the spec will need to be updated.

ethcore/src/snapshot/account.rs
arkpar: The problem is that RLP list size has to be specified in advance. You can't precisely estimate size after RLP serialization without doing the serialization.  The spec should not really put a restriction on chunks size, so I'll leave it it for future optimization (which can be done without breaking compatibility).

ethcore/src/snapshot/account.rs
arkpar: Code will be written for first instance only. Following records only store  code hashes.

arkpar:
cc @Nashatyrev @romanman @zilm13. We'll be upgrading to a new snapshot format in the next release. See updated spec here: https://github.com/paritytech/parity/wiki/Warp-Sync-Snapshot-Format
The upgrade is urgent because current  (v1) format does not allow a single account to be split between chunks. And now there's an account on the main network that produces a chunk over 16mb  that overflow devp2p network protocol limit for a single packet.

ethcore/src/snapshot/account.rs
```diff
@@ -62,53 +63,50 @@ impl CodeState {
 
 // walk the account's storage trie, returning an RLP item containing the
```
gavofyork: doc comment need updating.

ethcore/src/snapshot/account.rs
gavofyork: spec states that the hash may only be given when an account with a SHA3 hash strictly less has the same code. that's not (necessarily) the case here, so the spec implies it may be the case that the full code should be repeated.

ethcore/src/snapshot/account.rs
```diff
@@ -62,53 +63,50 @@ impl CodeState {
...
 // account properties and the storage.
-pub fn to_fat_rlp(acc: &BasicAccount, acct_db: &AccountDB, used_code: &mut HashSet<H256>) -> Result<Bytes, Error> {
+pub fn to_fat_rlps(acc: &BasicAccount, acct_db: &AccountDB, used_code: &mut HashSet<H256>, max_storage_items: usize) -> Result<Vec<Bytes>, Error> {
```
gavofyork: what will happen if the current chunk is half-full and the first `Bytes` in the returned `Vec` doesn't fit into the 16MB limit? does it just get closed like that and the next chunk it used?

arkpar: Yes, exactly. This also ensures there won't be 12.01MB full chunks to begin with. If the chunk is 3MB and partial account record is another 3MB it gets written to the next chunk.

ethcore/src/snapshot/account.rs
arkpar: I've updated the spec to specify ordering of same hash account records.

ethcore/src/snapshot/account.rs
gavofyork: well, presumably it can only prevent an account from being much bigger than 4MB. given the docs i would have assumed that if the current chunk file is 13MB and we're dealing with an account with large storage, then 3MB of the storage would be placed in the current file (to get it up to 16MB). it sounds like that wouldn't be the case - rathe the 13MB file would be closed and a new file started with the first 4MB of the account.

arkpar: Added another clarification to the spec:
> Each split record except the one containing the highest storage address hash contains exactly `MAX_STORAGE_ENTRIES_PER_ACCOUNT_RECORD` storage entries.

gavofyork:
ok - i think my issues have been addressed. @rphmeier should give it a final check though.

rphmeier:
LGTM. I think this will be fine as a stopgap until V3, where I'd like to take a look at generalizing block chunks to consensus-dependent "proof chunks".

PR #5025, opened 2017-03-24
node kind RPC

rphmeier:
cc @jacogr for JS side of things.

Adds `parity_nodeKind` which returns an object 
```json
{
    "availability": "availability_str",
    "capability": "capability_str"
}
```

`availability_str` one of `public`, `personal`
`capability_str` one of `full`, `light`

We may add more variants later, but this is enough to cover current use cases of web and personal nodes.


jacogr:
Just to get my ducks in a row -

1. public = web server, public = nodes we all run
2. full = normal node, light = light client (these 2 are pretty self-explanatory)

rphmeier:
Yep, exactly

gavofyork:
. might have broken protocol there - feel free to revert if the JS isn't up to scratch (LGTM though).

PR #5027, opened 2017-03-25
eip100b

debris:
#4833


ethcore/src/ethereum/ethash.rs
```diff
@@ -406,6 +412,8 @@ impl Ethash {
 			panic!("Can't calculate genesis block difficulty");
 		}
 
+		let parent_has_uncles = &sha3::SHA3_EMPTY_LIST_RLP == parent.uncles_hash();
```
gavofyork: i don't think this is what you mean.

debris: hah, you are right :)

PR #5290, opened 2017-03-27
Add lint:i18n to find missing & extra keys

jacogr:
Add `npm run lint:i18n` reporting the missing keys (found in original, not in translation) as well as any extra keys (found in translation, not found in original)

```
*** Checking translations for de
  Missing account.button.delete
  Missing account.button.edit
  Missing account.button.password
...
  Missing walletSettings.rejected.busyStep.title
  Missing walletSettings.rejected.title
  Missing web.requestToken
Found 41 keys, missing 465 keys, 0 extraneous keys

*** Checking translations for nl
Found 506 keys, missing 0 keys, 0 extraneous keys
```

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -53,6 +53,7 @@
     "lint:cached": "npm run lint:css && npm run lint:js:cached",
     "lint:css": "stylelint ./src/**/*.css",
     "lint:fix": "npm run lint:js:fix",
+    "lint:i18n": "babel-node ./scripts/lint-i18n.js",
     "lint:js": "eslint --ignore-path .gitignore ./src/",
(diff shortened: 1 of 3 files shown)


ngotchac:
Looks good. Would it make sense to output it as a JSON though?

jacogr:
I'll log an issue for that - honestly still unsure about exactly what would be useful here, so just started with identifying the gaps. (CSV could be useful as well)

ngotchac:
Okay. Yes I guess JSON would be easy to read if we want to have some kind of metric integration (could display somewhere every languages with the percentage of translation on master)

PR #5304, opened 2017-03-27
Public node with accounts and signing in Frontend

maciejhirsz:
Disclaimer: The PR includes a copy of the dictionary for the frontend, hence the crazy diff. It's not _that_ big.

Start parity with `--public-node` (suggestions for a better flag open):

- Disables account managing RPCs.
- UI account management is replicated in frontend. Accounts are stored in local storage in [keythereum](https://github.com/ethereumjs/keythereum) dump format.
- Creating accounts, recovering accounts from phrases and signing transactions all happens in the browser.

Some concerns and general notes:
- Generating keys in pure JS tends to be slow, and although there are avenues for optimization open there, a solution that re-uses Rust codebase via wasm or asm.js could potentially be better (less code to maintain). For the most part the UX feels fine as-is, although that's on a reasonable CPU, wouldn't try it on a phone.
- UI currently does a lot of requests, which might be both noticeable over the wire and reduce the amount of traffic the node can handle. Throttling things would be nice.
- Missing features, e.g. vaults, should either be hidden or implemented, although there seems to be little point to vaults in local storage?


rphmeier:
While throttling the amount of requests the UI can make to the web node is definitely important, we should also be able to throttle the amount (and intensity) of RPC requests in general. It's fair to assume that any publicly-available RPC will be pummeled mercilessly sometimes by malicious actors.

parity/configuration.rs
```diff
@@ -128,6 +128,7 @@ impl Configuration {
 			Some(true) if pruning == Pruning::Specific(Algorithm::Archive) => writeln!(&mut stderr(), "Warning: Warp Sync is disabled because pruning mode is set to archive").expect("Error writing to stderr"),
 			_ => {},
 		};
+		let public_node = self.args.flag_public_node;
```
tomusdrw: Public node should have only subset of APIs enabled:
`eth,net,parity,rpc,web3`

Some of the disabled APIs are already useless because `AccountProvider` is missing, but stuff like `parity_stopNetwork` should also be disabled.

js/src/api/local/accounts/accounts.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  get (address) {
+    address = address.toLowerCase();
+
+    this._last = address;
+
+    const account = this._store.find((account) => account.address === address);
+
+    if (account == null) {
```
maciejhirsz: I tend to use `== null` as an existence check, since the only values it can coerce to is `null` and `undefined`. `if (!account)` would also work. Is there a specific way to catch existence of a value that fits the code style?

js/src/api/local/accounts/accounts.js
```diff
@@ -0,0 +1,127 @@
...
+  remove (address) {
+    address = address.toLowerCase();
+
+    const index = this._store.findIndex((account) => account.address === address);
+
+    if (index === -1) {
+      return false;
+    }
+
+    if (address === this._last) {
```
maciejhirsz: It should be guaranteed to be lowercase as any inputs where address is passed are forced to lowercase before anything is done with it. That said, I'll add a setter to make it clearer it's an expected behavior.

parity/configuration.rs
maciejhirsz: Resolved now: https://github.com/paritytech/parity/blob/fc31fdf3041c8b013b6e0a6fd794ff5378e935b0/parity/configuration.rs#L724-L731

js/src/api/local/accounts/accounts.js
tomusdrw: I think `if (!account) {` is used everywhere.

gavofyork:
tests failing.

maciejhirsz:
Yeah, not sure what's up with the tests:

```
ERROR: Job failed (system failure): Error response from daemon: invalid header field value "oci runtime error: container_linux.go:247: starting container process caused \"process_linux.go:258: applying cgroup configuration for process caused \\\"failed to write 15099 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/docker/8aa02970dbeed05350837980f150e91c801cdfde378419435f2ece4eb1b7af21/cgroup.procs: invalid argument\\\"\"\n"
```

@tomusdrw I talked about this with Jaco, will bundle it with network change in the FE, which triggers a full reload. UX shouldn't really be an issue as that's pretty far from standard behavior.

PR #5312, opened 2017-03-28
Refresh UI on nodeKind changes, e.g. personal -> public

jacogr:
Part of https://github.com/paritytech/parity/issues/5007

- Poll parity_nodeKind
- Adjust timing factor based on node type (long polls)
- Refresh UI when kind availability/capability changes are detected (as we do with chain changes, e.g. node goes from public -> personal or full -> light)


js/src/redux/providers/chainMiddleware.js
```diff
@@ -24,20 +22,28 @@ export default class ChainMiddleware {
...
+
+          if (newChain) {
+            const { nodeStatus } = store.getState();
+            const { netChain, nodeKind } = nodeStatus;
+
+            // force reload when chain has changed
+            let forceReload = newChain !== netChain && netChain !== DEFAULT_NETCHAIN;
+
+            // force reload when nodeKind (availability/capability) has changed
+            if (!forceReload && collection.nodeKind !== null && nodeKind !== null) {
```
maciejhirsz: Shouldn't `nodeKind` checks be outside of the `if (newChain) {`?

jacogr: Will double-check, thanks. Think you are correct though.

js/src/redux/providers/chainMiddleware.js
```diff
@@ -24,20 +22,34 @@ export default class ChainMiddleware {
...
+          const newNodeKind = collection.nodeKind;
+          let reloadChain = false;
+          let reloadType = false;
+
+          // force reload when chain has changed and is not initial value
+          if (newChain) {
+            const hasChainChanged = newChain !== netChain;
+            const isInitialChain = netChain === DEFAULT_NETCHAIN;
+
+            reloadChain = !isInitialChain && hasChainChanged;
```
maciejhirsz: I don't really follow why a reload does not happen if current chain is a default chain? (I know the logic here didn't change, just curious).

jacogr: Maybe it is slightly wrongly named. On startup we don't know the chain, so we set it to `DEFAULT_CHAIN` (`DEFAULT_CHAIN = '(unknown)'`). So when the value changes from this, we don't do the reload, but on any subsequent changes, we do.

PR #5318, opened 2017-03-28
Snapshot chunks packed by size

arkpar:


util/rlp/src/stream.rs
```diff
@@ -133,14 +133,27 @@ impl RlpStream {
...
 			},
 		}
 
 		// return chainable self
 		self
 	}
 
+
+	/// Declare appending the list of unknown size, chainable.
+	pub fn begin_unbound_list(&mut self) -> &mut RlpStream {
```
rphmeier: "unbounded"

ethcore/src/snapshot/account.rs
```diff
@@ -105,9 +95,47 @@ pub fn to_fat_rlps(acc: &BasicAccount, acct_db: &AccountDB, used_code: &mut Hash
...
-	}).collect()
+		account_stream.begin_unbounded_list();
+		if account_stream.len() > target_chunk_size {
+			// account does not fit, push an empty record to mark a new chunk
+			target_chunk_size = max_chunk_size;
+			chunks.push(Vec::new());
+		}
+
+		if let Some(pair) = leftover.take() {
+			account_stream.append_raw_checked(&pair, 1, target_chunk_size);
```
debris: can you put an `assert` here?

util/rlp/src/stream.rs
```diff
@@ -273,6 +317,17 @@ impl RlpStream {
...
+	/// Finalize current ubnbound list. Panics if no unbounded list has been opened.
+	pub fn complete_unbounded_list(&mut self) {
+		let list = self.unfinished_lists.pop().expect("No open list.");
+		if list.max.is_some() {
+			panic!("List type mismatch.");
+		}
+		let len = self.buffer.len() - list.position;
+		self.encoder().insert_list_payload(len, list.position);
+		self.note_appended(1);
+	}
```
debris: Maybe not in this pr, but what do you think about moving this logic to a new structure - `UnboundedRlpStream`? imo, it would reduce complexity

ethcore/src/snapshot/account.rs
```diff
@@ -105,9 +95,47 @@ pub fn to_fat_rlps(acc: &BasicAccount, acct_db: &AccountDB, used_code: &mut Hash
...
+		}
+
+		loop {
+			match db_iter.next() {
```
debris: one of those loops could be moved to a new structure -> `RlpChunkedStream` which returns vec of chunks

arkpar: Chunking implemented here is too snapshot specific. `Itertools::chunks` can be used in simpler cases.

util/rlp/src/stream.rs
rphmeier: It would only make sense if we forced each list to be a separate stream. What would it mean to call `begin_list(X)` on an `UnboundedRlpStream` or `begin_list_unbounded()` on a regular `Stream`?

arkpar: There is a use case when both bounded and unbounded lists are added to the same stream. I'd redesign `RlpStream` to accept tuples instead of fixed size `begin_list()`/`end_list()` and add a variant of `append()` that accepts an iterator. But that's out of the scope of this PR

ethcore/src/snapshot/account.rs
```diff
@@ -105,9 +95,48 @@ pub fn to_fat_rlps(acc: &BasicAccount, acct_db: &AccountDB, used_code: &mut Hash
...
+		account_stream.begin_unbounded_list();
+		if account_stream.len() > target_chunk_size {
+			// account does not fit, push an empty record to mark a new chunk
+			target_chunk_size = max_chunk_size;
+			chunks.push(Vec::new());
+		}
+
+		if let Some(pair) = leftover.take() {
+			let leftover_appended = account_stream.append_raw_checked(&pair, 1, target_chunk_size);
+			assert!(leftover_appended);
```
gavofyork: prove or remove.

gavofyork:
`assert!` can panic, can't it?

arkpar:
Assert here indicates that this function is used incorrectly (`max_chunk_size` < ~300 bytes)

gavofyork:
feel free to keep the assert, but just ensure you prove why it can never happen :)

arkpar:
Since this is a public function and the `max_chunk_size` constant is defined elsewhere the proof would include assumptions external to this module.

gavofyork:
if a public function can be called with invalid data such that it'll end up returning meaningless results (or worse having arbitrary side-effects), then it should be detecting such invalid input and failing fast with a `Result` (invalid input -> `Result.Err`...).

rphmeier:
@gavofyork 

That is now the behavior. Also, I'm not sure it really counts as a "public" function since it's only accessible from the module where the target chunk size constant is declared (and this is what the assertion depends on).

I think either way is fine, either as an `assert!(checked_append, "leftover only pushed directly after beginning fresh chunk; qed")` or to return an error.

PR #5320, opened 2017-03-28
Filters and block RPCs for the light client

rphmeier:
Adds minimal implementations of log/filter RPCs for the light client.
Doesn't fetch any headers from network, just bails if they're not available locally.


rpc/src/v1/impls/light/eth.rs
```diff
@@ -484,19 +615,98 @@ impl Eth for EthClient {
...
+		Vec::new()
+	}
+
+	fn logs(&self, filter: EthcoreFilter) -> BoxFuture<Vec<Log>, Error> {
+		use std::collections::BTreeMap;
+
+		use futures::stream::{self, Stream};
+		use util::H2048;
+
+		// early exit for "to" block before "from" block.
```
gavofyork: comment is wrong.

rpc/src/v1/impls/light/eth.rs
```diff
@@ -484,19 +615,98 @@ impl Eth for EthClient {
...
+		match filter.from_block {
+			BlockId::Latest | BlockId::Pending => {
+				let best = self.client.best_block_header();
+				let chain_info = self.client.chain_info();
+				if best.number() != chain_info.best_block_number || best.hash() != chain_info.best_block_hash {
```
gavofyork: am i missing something here? looks to just be comparing `self.client.chain_info()` with `self.client.best_block_header()`...

rpc/src/v1/impls/light/eth.rs
rphmeier: it should now.

rpc/src/v1/impls/light/eth.rs
rphmeier: Good catch, seems like it got mangled a bit when I was running over that part of the code.

rpc/src/v1/impls/light/eth.rs
```diff
@@ -484,19 +615,99 @@ impl Eth for EthClient {
...
+		// early exit for "to" block before "from" block.
+		let best_number = self.client.chain_info().best_block_number;
+		let block_number = |id| match id {
+			BlockId::Earliest => Some(0),
+			BlockId::Latest | BlockId::Pending => Some(best_number),
+			BlockId::Hash(h) => self.client.block_header(BlockId::Hash(h)).map(|hdr| hdr.number()),
+			BlockId::Number(x) => Some(x),
+		};
+
+		if block_number(filter.to_block) < block_number(filter.from_block) {
```
gavofyork: does this really play well with `Option`'s ordering? what if `to_block` is `None`?

rphmeier: It's sort of a trade-off, but actually neither `from_block` nor `to_block` is ever a hash where these are invoked, so these are both always `Some`.

The alternative would be to pass a `rpc::Filter` instead of an `EthcoreFilter`, but then the logic for setting the to and from blocks would have to be duplicated.

gavofyork: would suggest using either an alternative type which doesn't given the option of providing a `Hash`, or otherwise bomb early with an `Err` if either is `None` .

PR #5332, opened 2017-03-29
Rename Status/Status -> Status/NodeStatus

jacogr:


ngotchac:
The Class in `js/src/views/Status/NodeStatus/store.js` could be renamed `NodeStatusStore`

PR #5336, opened 2017-03-29
Shared wordlist for frontend

maciejhirsz:
Analog to #5331 - use external repo with single source for the wordlist recovery phrases are generated from.


js/webpack/app.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
       },
       {
         test: /\.js$/,
-        include: /node_modules\/(material-chip-input|ethereumjs-tx)/,
```
tomusdrw: Is this necessary? I think the code should work in any browser supporting ES6. Do we support other browsers anyway?

maciejhirsz: It shouldn't be. We are still using UglifyJS to pack the bundle, and UglifyJS only understands ES5 syntax. I think the only thing that is ES6 there are `const` and `let` declarations, so we could either kick those out, or change the minifier to Babili or something else.

tomusdrw: Ah, I see. So it's ok then, was just making sure that it's needed.

jacogr: I have something in the back of my head to look at the new Babel uglifier, thanks for the reminder in logging that - just need to see if there are any benefits/drawbacks.

PR #5337, opened 2017-03-29
Kovan warp sync fixed

arkpar:


arkpar:
Changes in `ethsync` are required to support non-unique receipts. With EIP-98 multiple blocks can have the the same non-empty receipt that needs to be matched against multiple headers.

ethcore/src/client/client.rs
```diff
@@ -388,12 +388,16 @@ impl Client {
 			let db = self.state_db.lock().boxed_clone_canon(header.parent_hash());
 
 			let enact_result = enact_verified(block, engine, self.tracedb.read().tracing_enabled(), db, &parent, last_hashes, self.factories.clone());
-			let locked_block = enact_result.map_err(|e| {
+			let mut locked_block = enact_result.map_err(|e| {
 				warn!(target: "client", "Block import failed for #{} ({})\nError: {:?}", header.number(), header.hash(), e);
 			})?;
 
+			if header.number() < self.engine().params().validate_receipts_transition && header.receipts_root() != locked_block.block().header().receipts_root() {
+				locked_block.strip_receipts();
```
rphmeier: I feel like this violates the immutable interface that `LockedBlock` presents -- it would be fine internal to `block.rs` but exposing and using `strip_receipts` externally seems to discard invariants we previously had.

gavofyork:
lgtm. @rphmeier should do a final review.

rphmeier:
Seems fine to me

sync/src/blocks.rs
```diff
@@ -366,23 +370,26 @@ impl BlockCollection {
 			let receipts = UntrustedRlp::new(&r);
 			ordered_trie_root(receipts.iter().map(|r| r.as_raw().to_vec())) //TODO: get rid of vectors here
 		};
-		match self.receipt_ids.get(&receipt_root).cloned() {
-			Some(h) => {
-				self.receipt_ids.remove(&receipt_root);
+		match self.receipt_ids.entry(receipt_root) {
+			Entry::Occupied(mut entry) => {
+				let h = entry.get_mut().pop().expect("Empty vectors are not allowed in insert_receipt; qed");
```
rphmeier: actually, if the receipts roots are all the same here then we should be able to drain all block hashes in the entry as soon as we get the correct set.

arkpar: It's not that simple. Every response must be matched against a request, otherwise we can't tell good behaviour from bad. I've made an additional commit that tracks unique receipts and only issues one download request for each receipt root.

gavofyork:
@rphmeier sign off?

gavofyork:
@rphmeier sign off?

PR #5341, opened 2017-03-30
Futures-based native wrappers for contract ABIs

rphmeier:
Motivations:
  - asynchronous calls needed to use the registry with the light client
  - the previous situation of checked-in generated code scattered in various places seemed less maintainable.
  - a public Rust-based binding generation will be useful to the ecosystem at large

Code generation done at compile time in `ethcore/native_contract/generator`. The generation is currently `String`y because procedural macros don't support this kind of thing yet, but it shouldn't be too hard to port over afterwards.

The code generation is pretty good; it handles nested lists and multiple outputs without a problem.
The downside is just that you can't see the generated code very easily. Taking a guess at input/output types and seeing what the type hints tell you is usually good enough.

`Registrar` and `ServiceTransactionChecker` have been ported, and their wrappers now live in `ethcore/native_contracts` which will be the new home for contracts useful to Parity.


ethcore/native_contracts/generator/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+		ParamType::FixedBytes(len) if len <= 32 =>
+			format!("Token::FixedBytes({}.0.to_vec())", name),
+		ParamType::FixedBytes(len) => {
+			needs_mut = true;
+			format!("{}.resize({}, 0); Token::FixedBytes({})", name, len, name)
+		}
+		ParamType::Int(width) => match width {
+			8 => format!("let mut r = [0xff; 32]; r[31] = {}; Token::Int(r)", name),
+			16 | 32 | 64 =>
+				format!("let mut r = [0xff; 32]; BigEndian::write_i{}(&mut r[{}..], {}); Token::Int(r))",
```
rphmeier: @debris I'm not all that familiar with ethabi -- but assuming the int width is fixed, does it matter what the left side of the array holds? is `0xff` right?

ethcore/src/miner/service_transaction_checker.rs
```diff
@@ -14,17 +14,20 @@
...
+use native_contracts::ServiceTransactionChecker as Contract;
 use util::{U256, Uint, Mutex};
 
 const SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME: &'static str = "service_transaction_checker";
 
 /// Service transactions checker.
 #[derive(Default)]
 pub struct ServiceTransactionChecker {
-	contract: Mutex<Option<provider::Contract>>,
+	contract: Mutex<Option<Contract>>,
```
rphmeier: @svyatonik Please review the changes here to make sure it's all ok.

rphmeier:
This could almost be used in `rust-web3` as is.

ethcore/native_contracts/build.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn build_file(name: &str, abi: &str, filename: &str) {
+	let code = ::native_contract_generator::generate_module(name, abi).unwrap();
+
+	let out_dir = ::std::env::var("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join(filename);
```
tomusdrw: Something wrong with indentation here.

ethcore/native_contracts/generator/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	output_type.push_str(")");
+	decode_outputs.push_str("))");
+	Ok((output_type, decode_outputs))
+}
+
+// create code for an argument type from param type.
+fn rust_type(input: ParamType) -> Result<String, ParamType> {
+	Ok(match input {
```
tomusdrw: Borked indentation again.

ethcore/src/miner/service_transaction_checker.rs
svyatonik: Checked it on Ropsten - everything works :)

PR #5346, opened 2017-03-30
Add default eip155 validation

keorn:


gavofyork:
this is only going in now?!

PR #5351, opened 2017-03-30
Update expanse json with fork at block 600000

chrisfranko:
We are updating to the latest exp protocol at block 600k

### DIFF
--- a/ethcore/res/ethereum/expanse.json
+++ b/ethcore/res/ethereum/expanse.json
@@ -15,11 +15,11 @@
 				"difficultyHardforkTransition": "0x59d9",
 				"difficultyHardforkBoundDivisor": "0x0200",
 				"bombDefuseTransition": "0x30d40",
-				"eip150Transition": "0x7fffffffffffffff",
-				"eip155Transition": "0x7fffffffffffffff",
-				"eip160Transition": "0x7fffffffffffffff",
-				"eip161abcTransition": "0x7fffffffffffffff",
-				"eip161dTransition": "0x7fffffffffffffff"
+				"eip150Transition": "0x927C0",
+				"eip155Transition": "0x927C0",
+				"eip160Transition": "0x927C0",
+				"eip161abcTransition": "0x927C0",
+				"eip161dTransition": "0x927C0"
 			}
 		}
 	},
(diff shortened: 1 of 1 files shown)


chrisfranko:
[clabot:check]

ethcore/res/ethereum/expanse.json
```diff
@@ -15,11 +15,11 @@
 				"difficultyHardforkTransition": "0x59d9",
 				"difficultyHardforkBoundDivisor": "0x0200",
 				"bombDefuseTransition": "0x30d40",
-				"eip150Transition": "0x7fffffffffffffff",
-				"eip155Transition": "0x7fffffffffffffff",
-				"eip160Transition": "0x7fffffffffffffff",
-				"eip161abcTransition": "0x7fffffffffffffff",
-				"eip161dTransition": "0x7fffffffffffffff"
+				"eip150Transition": "0x927C0",
+				"eip155Transition": "0x927C0",
```
rphmeier: EIP155 needs a chain/network ID to be specified

rphmeier: With EIP155 enabled, you may want to specify a "chainID" in params. Otherwise, transactions from the Ethereum mainnet will still be replayable onto expanse and vice-versa as they both specify network ID 1, defeating the point a bit.

rphmeier: For reference, `chainID` is used only for transaction signing and defaults to the value of `networkID` if not specified.

chrisfranko: Sure, I get what you are saying, however Expanse uses a completely different network identifier so nodes dont even have the ability to talk to each other and fail at the handshake. I suppose people could copy over a raw tx and try to replay those txs on EXP but its highly unlikely any EXP addresses are also used for ETH.

chrisfranko: oh man I just went full potato, sorry, doing 100 things at once today. Ill choose a chainID and submit it in a few, sorry.

chrisfranko: @rphmeier updated to chainID 2, since etc chose 61.

PR #5364, opened 2017-03-31
Don't queue requests when WS is not connected

jacogr:
Optimise, don't flood node when WS connection is made (especially with public & light nodes)


jacogr:
Closed in favour of https://github.com/paritytech/parity/pull/5367

PR #5374, opened 2017-04-01
Work around mismatch for QR checksum

jacogr:
Don't merge, just here to allow testing, working around checksum mismatch.

(Should be closed once native-signer update is released)


gavofyork:
merging - please revert once signer works.

PR #5383, opened 2017-04-03
parity_getBlockHeaderByNumber and LightFetch utility

rphmeier:
Hand-off to @jacogr 

The motivation behind this method is to lower the strain on nodes being polled for latest block info by the UI. Recent headers and total difficulty are always available on the light client, while recent block bodies are not.

Moves some utilities of the `eth` RPCs for fetching data either from the local chain or from the network into a `LightFetch` which can be instantiated from various different RPCs.

Uses `LightFetch` from the `parity` RPC implementation to implement `parity_getBlockHeaderByNumber`.


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use light::cache::Cache;
+use light::client::LightChainClient;
+use light::cht;
+use light::on_demand::{OnDemand, request};
+
+use ethsync::LightSync;
+use util::{Address, Mutex, Uint, U256};
+
+use v1::helpers::{CallRequest as CRequest, errors, dispatch};
```
gavofyork: consider `CRequest` -> `CallRequestHelper` to aid readability.

gavofyork:
marking inprogress until resolved.

PR #5385, opened 2017-04-03
Allow empty-encoded values from QR encoding

jacogr:
As reported by @tomusdrw


jacogr:
Confirmed fixed for zero nonce (local testing & @tomusdrw 's testing)

PR #5386, opened 2017-04-03
Fixing secretstore TODOs - part 1

svyatonik:
On top of #5015 

PR outline:
1) get rid of hardcoded secretstore configuration. Now it is possible to configure via command line (inconvenient) or via config file (thanks for suggestion, @rphmeier ). Example config:
```
[parity]
chain = "kovan"
base_path = "db.kovan_ss1"

[ui]
disable = true

[rpc]
disable = true

[ipc]
disable = true

[dapps]
disable = true

[network]
port = 30303

[secretstore]
disable = false
self_secret = "6c26a76e9b31048d170873a791401c7e799a11f0cefc0171cc31a49800967509"
nodes = ["cac6c205eb06c8308d65156ff6c862c62b000b8ead121a4455a8ddeff7248128d895692136f240d5d1614dc7cc4147b1bd584bd617e30560bb872064d09ea325@127.0.0.1:8085", "6b8d9b9ecde2f0d8c7b2685dd10d181000fa0b26462674dece3096488d6b8d6e3c0e4e1262e3f0eb3b997783b3d6471c281905c5fafeb7908f3aeb7326274db6@127.0.0.1:8087"]
interface = "local"
port = 8083
http_interface = "local"
http_port = 8082
path = "db.kovan_ss1/secretstore"

[ipfs]
enable = false

[snapshots]
disable_periodic = true
```
2) get rid of obsolete tokio_core::io in favor of tokio_io (as @debris requested in previous PR)
3) added some tests && fixed errors for edge cases (e.g. when decryption is performed by KeyServer, which itself denies access to the requested document to the requestor)
4) spent some time thinking of how to implement "secret is never revealed to any KeyServer"-option. So here's my proposal: in current algorithm secret is calculated by 'master' KeyServer (the one which has received decryption request from user) by calculating expression:
```
DECRYPTED_SECRET = ENCRYPTED_SECRET + COMMON_POINT * SUM(SECRET_SHARE)
where
ENCRYPTED_SECRET && COMMON_POINT - shared artifacts of encryption session
SECRET_SHARE - is calculated on every node during decryption session 
```
The idea is that every node calculates two parts of SECRET_SHARE, so that SECRET_SHARE = (SECRET_SHARE1 * SECRET_SHARE2) mod n. Then it returns SECRET_SHARE1 to master KeyServer, which uses it to compute result of above expression. SECRET_SHARE2 is encrypted with requestor public key && is also returned to master server, but its value is unknown. So at the end of decryption session, master KS returns next data to requestor:
4.1) partially decrypted secret
4.2) M SECRET_SHARE2 values, encrypted with requestor public key
4.3) COMMON_POINT
Requestor then finally decrypts secret using decrypted(SECRET_SHARE2) values as an input to the same expression as above.
In theory, we can omit calculating expression on KeyServer && just return encrypted SECRET_SHARE to requestor, but in practice expression on KeyServer is slightly different && involves more EC math actions. And I wanted to keep EC math, required on client as simple as possible.
Important note: it is not the last place where secret is revealed on KeyServer, as it is currently generated on the master KeyServer during encryption session.
Also - I have left two decryption session options for now - with revealing secret on KeyServer (as it was working before - GET request to /DOCUMENT/SIGNATURE) && without revealing (GET request of form /shadow/DOCUMENT/SIGNATURE).


svyatonik:
5) messages are now sent encrypted over network

debris:
lgtm

PR #5388, opened 2017-04-03
Keep to default signer position when not in dapp

jacogr:
Fixes https://github.com/paritytech/parity/issues/5387


ngotchac:
This doesn't work. As the Parity Bar is never unmounted, if the UI isn't opened in a dapp view, the hook to the dapp loading will never be created. Thus, Parity Bar positioning doesn't work in dapps anymore.

jacogr:
So fix one, introduce another. Ok, will try and find an alternative approach.

jacogr:
Closing.

PR #5390, opened 2017-04-03
Public node: perf and fixes

maciejhirsz:
- Introduced a worker pool for webworkers, which speeds up re-generating of preview avatars (apparently each worker has to JIT the js on it's own, over and over again).
- All keythereum related stuff is now done in workers as well, so the UI never hangs.
- Regenerating the avatars on the new account modal now shows a spinner.
- Restoring accounts from private keys, testing passwords and changing passwords now works as intended.

### DIFF
--- a/js/src/api/local/accounts/account.js
+++ b/js/src/api/local/accounts/account.js
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import { keythereum } from '../ethkey';
+import { createKeyObject, decryptPrivateKey } from '../ethkey';
 
 export default class Account {
   constructor (persist, data) {
@@ -31,12 +31,14 @@ export default class Account {
   }
 
   isValidPassword (password) {
-    try {
-      keythereum.recover(Buffer.from(password), this._keyObject);
-      return true;
-    } catch (e) {
-      return false;
-    }
+    return decryptPrivateKey(this._keyObject, password)
+      .then((privateKey) => {
+        if (!privateKey) {
+          return false;
+        }
+
+        return true;
+      });
   }
 
   get address () {
@@ -68,21 +70,23 @@ export default class Account {
   }
 
   decryptPrivateKey (password) {
-    return keythereum.recover(Buffer.from(password), this._keyObject);
+    return decryptPrivateKey(this._keyObject, password);
+  }
+
+  changePassword (key, password) {
+    return createKeyObject(key, password).then((keyObject) => {
+      this._keyObject = keyObject;
+
+      this._persist();
+    });
   }
 
   static fromPrivateKey (persist, key, password) {
-    const iv = keythereum.crypto.randomBytes(16);
-    const salt = keythereum.crypto.randomBytes(32);
+    return createKeyObject(key, password).then((keyObject) => {
+      const account = new Account(persist, { keyObject });
 
-    // Keythereum will fail if `password` is an empty string
-    password = Buffer.from(password);
-
-    const keyObject = keythereum.dump(password, key, salt, iv);
-
-    const account = new Account(persist, { keyObject });
-
-    return account;
+      return account;
+    });
   }
 
   toJSON () {
--- a/js/src/api/local/accounts/accounts.js
+++ b/js/src/api/local/accounts/accounts.js
@@ -39,7 +39,8 @@ export default class Accounts {
   create (secret, password) {
     const privateKey = Buffer.from(secret.slice(2), 'hex');
 
-    return Account.fromPrivateKey(this.persist, privateKey, password)
+    return Account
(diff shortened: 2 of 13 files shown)


js/src/api/local/accounts/account.js
```diff
@@ -68,21 +70,23 @@ export default class Account {
...
+    return createKeyObject(key, password).then((keyObject) => {
+      this._keyObject = keyObject;
 
-    const keyObject = keythereum.dump(password, key, salt, iv);
+      this._persist();
+    });
+  }
 
-    const account = new Account(persist, { keyObject });
+  static fromPrivateKey (persist, key, password) {
```
jacogr: Not sure if you tested this 100%, if not there is a private key we use in tests on the Rust side inside tests (with a known account). If already done, ignore :)

js/src/api/local/accounts/accounts.js
```diff
@@ -38,14 +38,22 @@ export default class Accounts {
 
   create (secret, password) {
     const privateKey = Buffer.from(secret.slice(2), 'hex');
-    const account = Account.fromPrivateKey(this.persist, privateKey, password);
 
-    this._store.push(account);
-    this.lastAddress = account.address;
+    return Account.fromPrivateKey(this.persist, privateKey, password)
```
jacogr: Small nitpick, would prefer 

```
return Account
  .fromPrivateKey(...)
  .then(...)
```

Just aligns with what we have in the rest of the codebase.

js/src/api/local/accounts/accounts.js
```diff
@@ -73,28 +81,40 @@ export default class Accounts {
...
-    const index = this._store.findIndex((account) => account.address === address);
+    const account = this.get(address);
 
-    if (index === -1) {
+    if (!account) {
       return false;
     }
 
-    const account = this._store[index];
+    return account.isValidPassword(password)
```
jacogr: Same samll alignment comment as above.

js/src/api/local/ethkey/index.js
```diff
@@ -14,31 +14,34 @@
...
-// Allow a web worker in the browser, with a fallback for Node.js
-const hasWebWorkers = typeof Worker !== 'undefined';
-const KeyWorker = hasWebWorkers ? require('worker-loader!./worker')
-                                : require('./worker').KeyWorker;
+import workerPool from './workerPool';
 
-// Local accounts should never be used outside of the browser
-export let keythereum = null;
+export function createKeyObject (key, password) {
+  return workerPool.getWorker().action('createKeyObject', { key, password })
```
jacogr: As per above, prefer

```
return ...getWorker()
  .action(...)
  .then(...)
```

(Next function has it that way, just align)

js/src/api/local/ethkey/worker.js
```diff
@@ -14,68 +14,117 @@
...
+
+      // Convert to array to safely send from the worker
+      return Array.from(key);
+    } catch (e) {
+      return null;
     }
   }
+};
+
+function bytesToHex (bytes) {
```
jacogr: Not 100% sure why this is "duplicated", i.e. not used from `/api/util/format`? (I looked at this the previous round as well, could see a slight difference, but non-duplication is obviously preferred)

js/src/api/local/middleware.js
```diff
@@ -57,6 +57,21 @@ export default class LocalAccountsMiddleware extends Middleware {
       });
     });
 
+    register('parity_changePassword', ([address, oldPassword, newPassword]) => {
+      const account = accounts.get(address);
+
+      return account.decryptPrivateKey(oldPassword)
```
jacogr: Small nitpick as above.

jacogr:
Generally looks good, one testing comment (probably already taken care of), some small style grumbles, just one function duplication where we can re-use. No major grumbles.

Generally though, think we should try and invest in tests for the middleware at some time sooner rather than later, just ensuring that what we do is correct. (And that the future holds to that promise)

js/src/modals/CreateAccount/NewAccount/newAccount.spec.js
```diff
@@ -58,12 +58,13 @@ describe('modals/CreateAccount/NewAccount', () => {
         return instance.componentWillMount();
       });
 
-      it('creates initial accounts', () => {
-        expect(Object.keys(instance.state.accounts).length).to.equal(7);
+      it('resets the accounts', () => {
+        expect(instance.state.accounts).to.be.null;
+        // expect(Object.keys(instance.state.accounts).length).to.equal(7);
```
tomusdrw: Is it still needed?

js/src/api/local/accounts/account.js
maciejhirsz: All the actual logic for this is in Keythereum, and the main issue here is that the tests take a while and can timeout in CI :|.

PR #5395, opened 2017-04-04
YARN

CraigglesO:
Adding a yarn.lock file.

Step one of a three step process.


jacogr:
What is step 2 & 3?

Any reason for updating the run commands to yarn as well?

CraigglesO:
step 2) flow
step 3) 
 a) update to react-router-dom
 b) update webpack config

yeah, yarn just has some improvements on speed in some of those commands... 

derp, but 47 and 51 should be fixed though.

jacogr:
Really not sure the 10ms on a command is worth it. (Until it comes packed with Node, would like to minimise "stuff to do & install before contributing")

So my concern is just that we are making it slightly more difficult for users/developers/translators to participate with the run commands. i.e. previously only node was required. Now running `npm start` will do a `yarn install & ...` - basically unless somebody has yarn globally it will fail at this point. 

This one specifically is a bit of a tougher one actually, since we want the yarn benefits on install, but not mandate it from a startup/setup perspective. Ideas?

CraigglesO:
No worries, It's changed back!

Any other concerns? I'll change em!

PR #5397, opened 2017-04-04
Don't use port 8080 anymore

ngotchac:


ngotchac:
LG ? ;)

codygriffin:
Does the wiki need to be updated to reflect this?  Specifically, the section about remote access to wallets via SSH forwarding?

tomusdrw:
@codygriffin Yes, although 1.7 containing this change hasn't been released yet. I'll update the wiki after that.

PR #5402, opened 2017-04-05
APIs wildcards and simple arithmetic.

### LINKED ISSUE #5394: Add support for --jsonrpc-apis all
It would be nice to be able to use:
`parity --jsonrpc-apis all`

with all being the entire list of available apis.

tomusdrw:
Closes #5394

### DIFF
--- a/parity/rpc_apis.rs
+++ b/parity/rpc_apis.rs
@@ -83,9 +83,17 @@ impl FromStr for Api {
 
 #[derive(Debug, Clone)]
 pub enum ApiSet {
+	// Safe context (like token-protected WS interface)
 	SafeContext,
+	// Unsafe context (like jsonrpc over http)
 	UnsafeContext,
+	// Public context (like public jsonrpc over http)
+	PublicContext,
+	// All possible APIs
+	All,
+	// Local "unsafe" context and accounts access
 	IpcContext,
+	// Fixed list of APis
 	List(HashSet<Api>),
 }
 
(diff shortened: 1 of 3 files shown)


parity/rpc_apis.rs
```diff
@@ -152,21 +180,41 @@ fn to_modules(apis: &[Api]) -> BTreeMap<String, String> {
...
+	pub fn retain(self, set: Self) -> Self {
+		ApiSet::List(&self.list_apis() & &set.list_apis())
+	}
+
 	pub fn list_apis(&self) -> HashSet<Api> {
-		let mut safe_list = vec![Api::Web3, Api::Net, Api::Eth, Api::Parity, Api::Traces, Api::Rpc]
-			.into_iter().collect();
+		let mut public_list = vec![
+			Api::Web3, Api::Net, Api::Eth, Api::Parity, Api::Rpc,
+		].into_iter().collect();
```
rphmeier: in places like these I wish we had an "iter" macro in the standard library: http://play.integer32.com/?gist=d0bf3fcbfcbb35ad8752ed5be30bda64&version=undefined

maciejhirsz: Could skip the `vec!` and just use an array.

rphmeier: yep, that too. but then you have to either have a `HashSet<&Api>` or add a `Clone` impl

Issue #5398, opened 2017-04-05
Specific cache for `light::HeaderChain`


rphmeier:
Perhaps it can just be given a handle to the `LightDataCache` other services share.

@adrianbrink You might be interested in this one, I can mentor.

adrianbrink:
@rphmeier Okay. That would be great :-)

I've opened a PR fix the initial setup and also merged the CLI changes. All tests are still passing.

adrianbrink:
[clabot:check]

adrianbrink:
Should we cache theses things:

- status - &H256 -> BlockStatus
- cht_root - usize -> Option<H256>
- block_header - BlockId -> Option<encoded::Header>

I am asking because the Cache in cache.rs doesn't support those mappings yet.

Should I attempt to make the cache generic (ideally using a macro) or just implement a cache specific to HeaderChain and later on make it generic.

@rphmeier

rphmeier:
@adrianbrink Just headers should be cached for now. The `Cache` has a mapping `H256` -> `encoded::Header`, and headers are stored on disk also keyed by hash. 

Basically, we just need to replace anywhere we go to the DB for headers to first check the cache. If we do go to the DB, it should cache the header after it loads it.

We could cache CHT roots also, but they're really tiny and rarely used, so it's probably not worth the effort for now. 

Computing block statuses should become faster as a result of caching the headers, as we basically just check if we know whether the header exists by looking it up.

rphmeier:
The `Cache` structure is also fine as is, for now. These are the main things which need caching, although we will want to cache recent state values in the future.

adrianbrink:
I've updated the PR.

adrianbrink:
@rphmeier You can close the issue :-)

PR #5403, opened 2017-04-05
Add caching to HeaderChain struct

adrianbrink:
This is the initial PR for adding caching to the HeaderChain struct.

- spelling fixes in ethcore-light
- add reference to docs for LRU cache
- add member variable ```cache``` on HeaderChain

Left todo:
- use the cache in the methods

#5398

### DIFF
--- a/ethcore/light/src/client/header_chain.rs
+++ b/ethcore/light/src/client/header_chain.rs
@@ -427,7 +427,7 @@ impl HeaderChain {
 		}
 	}
 
-	/// Get the nth CHT root, if it has been computed.
+	/// Get the nth CHT root, if it's been computed.
 	///
 	/// CHT root 0 is from block `1..2048`.
 	/// CHT root 1 is from block `2049..4096`
--- a/ethcore/light/src/client/header_chain.rs
+++ b/ethcore/light/src/client/header_chain.rs
@@ -365,19 +365,24 @@ impl HeaderChain {
 	/// will be returned.
 	pub fn block_header(&self, id: BlockId) -> Option<encoded::Header> {
 		let load_from_db = |hash: H256| {
-			let cached = {
-				let mut cache = self.cache.lock();
-				cache.block_header(&hash)
-			};
+			let mut cache = self.cache.lock();
+			let header = cache.block_header(&hash);
 
-			if cached.is_some() {
-				return cached
-			} else {
-				match self.db.get(self.col, &hash) {
-					Ok(val) => val.map(|x| x.to_vec()).map(encoded::Header::new),
-					Err(e) => {
-						warn!(target: "chain", "Failed to read from database: {}", e);
-						None
+			match header {
+				Some(header) => Some(header),
+				None => {
+					match self.db.get(self.col, &hash) {
+						Ok(dbValue) => {
+							dbValue.map(|x| x.to_vec()).map(encoded::Header::new)
+								.and_then(|header| {
+									cache.insert_block_header(hash.clone(), header.clone());
+									Some(header)
+								 })
+						},
+						Err(e) => {
+							warn!(target: "chain", "Failed to read from database: {}", e);
+							None
+						}
 					}
 				}
 			}
@@ -423,7 +428,7 @@ impl HeaderChain {
 		}
 	}
 
-	/// Get the nth CHT root, if it has been computed.
+	/// Get the nth CHT root, if it's been computed.
 	///
 	/// CHT root 0 is from block `1..2048`.
 	/// CHT root 1 is from block `2049..4096`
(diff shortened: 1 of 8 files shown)


ethcore/light/src/client/header_chain.rs
```diff
@@ -408,7 +414,7 @@ impl HeaderChain {
 		}
 	}
 
-	/// Get the nth CHT root, if it's been computed.
+	/// Get the nth CHT root, if it has been computed.
```
rphmeier: technically not wrong before but reads better here :)

ethcore/light/src/net/mod.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-//! PIP Protocol Version 1 implementation.
+//! PLP Protocol Version 1 implementation.
```
rphmeier: No, it really is PIP...

maybe I'll just change the name if it's really confusing :)

rphmeier:
Thanks! The key method to look at is `block_header` for utilizing the cache.

ethcore/light/src/net/mod.rs
adrianbrink: But isn't it **P**arity **L**ightlient **P**rotocol? I have to say that I find it pretty confusing :-)

adrianbrink:
I've rebased and now update the cache.

It looks a build verbose due to the fact that ```insert_block_header()``` doesn't return the inserted item and hence the extended function block in ```and_then()```.

ethcore/light/src/client/header_chain.rs
adrianbrink: Reversed

ethcore/light/src/net/mod.rs
adrianbrink: Also reversed.

ethcore/light/src/client/header_chain.rs
```diff
@@ -359,11 +365,24 @@ impl HeaderChain {
...
-					warn!(target: "chain", "Failed to read from database: {}", e);
-					None
+			let mut cache = self.cache.lock();
+
+			match cache.block_header(&hash) {
+				Some(header) => Some(header),
+				None => {
+					match self.db.get(self.col, &hash) {
+						Ok(dbValue) => {
+							dbValue.map(|x| x.to_vec()).map(encoded::Header::new)
```
rphmeier: last thing: in Rust we usually use `snake_case` style. So `dbValue` should be `db_value`

rphmeier:
Okay, LGTM except for the capitalization issue

ethcore/light/src/client/service.rs
```diff
@@ -56,6 +60,8 @@ pub struct Service {
 impl Service {
 	/// Start the service: initialize I/O workers and client itself.
 	pub fn start(config: ClientConfig, spec: &Spec, path: &Path) -> Result<Self, Error> {
+		let cache = Arc::new(Mutex::new(Cache::new(Default::default(), Duration::hours(6))));
```
rphmeier: This should be given a cache, which should be the same one given to `OnDemand` and RPC in `parity/run.rs`

adrianbrink:
@rphmeier Rebased and fixed up parity/run.rs to now use the same Cache. However, currently we are passing the cache size down but I don't think it is used. I can try and fix that in another PR.

rphmeier:
Thanks!

LGTM, seems like some random other changes have gotten mixed into the diff, though.
Also, the CI doesn't appear to have run...

adrianbrink:
Anything that I should fix? Aren't those other changes from me rebasing?

gavofyork:
will push this as a branch in our repo to get the fire

gavofyork:
see #5428

rphmeier:
Looks like there are some test build failures.
Can you run `test.sh` until it at builds and completes successfully? Maybe set `CARGO_INCREMENTAL=1` for faster compilation.

adrianbrink:
Just noticed. I'll fix it now and then I'll update the PR. And also, I'll run the test script. Before I only ran the tests for the light_client crate.

adrianbrink:
@gavofyork @rphmeier  
The tests in ```./test.sh``` are all passing now locally. Could you trigger a 2nd CI build.

Also, do any of you have a good guide on how to handle rebasing when using a forking approach to git?

adrianbrink:
Also, is there a way to speed up the tests for the entire parity build?

I'm running them with ```CARGO_INCREMENTAL=1 ./test.sh``` on a MBP 13" and it takes around 5 min. :-)

rphmeier:
Usually we just `git merge` against the latest master rather than rebase because it's clean.

And that's about as fast as the tests get :)

adrianbrink:
Okay, that's good to know. I'll switch to ```git merge``` next time :-)

PR #5405, opened 2017-04-05
Store the pending requests per network version

### LINKED ISSUE #5359: Tx minimised shows transactions cross-chain
e.g. tx in-progress on Homestead shows up in-progress on Kovan with the additional issues that 

(a) cannot be resolved/authenticated via Signer, 
(b) addresses are not mapped to names & 
(c) MethodDecoding may not work

ngotchac:
Closes #5359

### DIFF
--- a/js/src/views/Application/Requests/savedRequests.spec.js
+++ b/js/src/views/Application/Requests/savedRequests.spec.js
@@ -19,6 +19,7 @@ import store from 'store';
 
 import SavedRequests, { LS_REQUESTS_KEY } from './savedRequests';
 
+const NETWORK_ID = 42;
 const DEFAULT_REQUEST = {
   requestId: '0x1',
   transaction: {}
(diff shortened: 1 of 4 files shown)


js/src/views/Application/Requests/savedRequests.spec.js
```diff
@@ -19,27 +19,37 @@ import store from 'store';
...
-  beforeEach(() => {
+  beforeEach((done) => {
     store.set(LS_REQUESTS_KEY, {
-      [DEFAULT_REQUEST.requestId]: DEFAULT_REQUEST
+      [NETWORK_ID]: {
+        [DEFAULT_REQUEST.requestId]: DEFAULT_REQUEST
+      }
     });
+
+    savedRequests.load(api)
```
jacogr: Could just return the Promise here instead of calling done. (Mocha is promise-aware)

PR #5414, opened 2017-04-06
Add CIDv0 RPC

keorn:


rpc/src/v1/impls/light/parity.rs
```diff
@@ -342,4 +342,8 @@ impl Parity for ParityClient {
 			capability: Capability::Light,
 		})
 	}
+
+	fn ipfs_cid(&self, _content: Bytes) -> Result<String, Error> {
+		Err(errors::light_unimplemented(None))
```
rphmeier: why not implement this for the light client as well?

rpc/src/v1/helpers/errors.rs
```diff
@@ -224,6 +225,14 @@ pub fn encryption_error<T: fmt::Debug>(error: T) -> Error {
 	}
 }
 
+pub fn encoding_error<T: fmt::Debug>(error: T) -> Error {
```
tomusdrw: `_error` suffix is not really necessary

keorn: Should I change all the other functions in this module too?

gavofyork: if so, worth a separate PR.

gavofyork:
does "add it for the light client as well" really mean a verbatim duplicate of that function? suffice it to say i consider this rather bad practice.

rphmeier:
It's mostly a symptom of the fact that the `parity_` API does just about everything in a single trait. Doing an `Eth`, `EthFilter` split into `ParityChain` and `ParityMisc` would let us at least make `ParityMisc` generic.

keorn:
@rphmeier Dont try to defend, that was just embarrassing...

PR #5415, opened 2017-04-06
Tendermint fixes

keorn:
Fix proposal sync and commit race condition.
Changes the seal format


gavofyork:
any chance of a unit test?

keorn:
Will try, but it has to do with timing.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -260,8 +241,17 @@ impl Tendermint {
 		self.validators.contains(&*self.proposal_parent.read(), address)
 	}
 
-	fn is_above_threshold(&self, n: usize) -> bool {
-		n > self.validators.count(&*self.proposal_parent.read()) * 2/3
+	fn is_above_threshold(&self, n: usize) -> Result<(), EngineError> {
```
tomusdrw: Functions with `is_` prefix usually return `bool`. Maybe `check_` or `validate_if` would be a better name?

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -260,8 +241,17 @@ impl Tendermint {
...
+		let threshold = self.validators.count(&*self.proposal_parent.read()) * 2/3;
```
tomusdrw: That won't really work for a single validator, right? If you ever reach to such state you will never be able to mint any new blocks.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -515,50 +522,45 @@ impl Engine for Tendermint {
...
-			let address = match self.votes.get(&precommit) {
-				Some(a) => a,
-				None => public_to_address(&recover(&precommit.signature.into(), &precommit_hash)?),
-			};
-			if !self.validators.contains(header.parent_hash(), &address) {
-				Err(EngineError::NotAuthorized(address.to_owned()))?
+		if let Ok(proposal) = ConsensusMessage::new_proposal(header) {
+			let proposer = proposal.verify()?;
+			if !self.is_authority(&proposer) {
+				Err(EngineError::NotAuthorized(proposer))?
```
tomusdrw: I think `return Err(EngineError::NotAuthorized(proposer))` would be way clearer, but I saw that you're using this pattern in many places already.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -515,50 +522,45 @@ impl Engine for Tendermint {
...
 			}
+			self.is_view_proposer(header.parent_hash(), proposal.vote_step.height, proposal.vote_step.view, &proposer)?;
+		} else {
+			let vote_step = VoteStep::new(header.number() as usize, consensus_view(header)?, Step::Precommit);
+			let precommit_hash = message_hash(vote_step.clone(), header.bare_hash());
+			let ref signatures_field = header.seal()[2];
```
tomusdrw: `let signatures_field = &header.seal()[2];` would be enough. Do we ever prove that seal has at least 2 elements? explicit `expect` would have been better.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -491,14 +497,15 @@ impl Engine for Tendermint {
 	fn verify_block_basic(&self, header: &Header, _block: Option<&[u8]>) -> Result<(), Error> {
 		let seal_length = header.seal().len();
 		if seal_length == self.seal_fields() {
+			let proposal_len = header.seal()[1].len();
 			let signatures_len = header.seal()[2].len();
-			if signatures_len >= 1 {
+			if (proposal_len == 1) ^ (signatures_len == 1) {
```
tomusdrw: I don't really get that change here.
So the block is valid either if:
1. there is one proposal and the number of signatures is not one (so 0 or 2 sigs is ok)
2. or there is one signature and the number of proposals is not one (so 0 or 2 proposals is ok),

Why is that correct? (Shouldn't it be either (a proposal and 0 sigs) or (a signature and 0 proposals)?)

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -515,50 +522,45 @@ impl Engine for Tendermint {
...
+					warn!(target: "engine", "verify_block_unordered: Duplicate signature from {} on the seal.", address);
+					Err(BlockError::InvalidSeal)?;
+				}
 			}
-			self.is_view_proposer(header.parent_hash(), proposal.vote_step.height, proposal.vote_step.view, &proposer)?;
-		}
 
-		if header.number() == 0 {
-			Err(BlockError::RidiculousNumber(OutOfBounds { min: Some(1), max: None, found: header.number() }))?;
+			self.is_above_threshold(signature_count)?
```
tomusdrw: `origins.len()` would work here, `signature_count` is redundant.

ethcore/src/engines/tendermint/mod.rs
keorn: Integer arithmetic, `threshold` will be 0, so 1 is enough.

ethcore/src/engines/tendermint/mod.rs
keorn: This is sort of wonky, since its just the length of the seal field (1 means no signatures). Will change it to comparing actual RLPs.

ethcore/src/engines/tendermint/mod.rs
keorn: You have to do `into` as well, but can change.

ethcore/src/engines/tendermint/mod.rs
keorn: In `verify_block_basic`, sure can do.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -337,18 +327,32 @@ impl Tendermint {
...
+					Some(Step::Propose)
+				},
 				Step::Precommit if self.has_enough_aligned_votes(message) => {
-					if message.block_hash.is_none() {
-						self.increment_view(1);
-						Some(Step::Propose)
-					} else {
-						Some(Step::Commit)
+					let bh = message.block_hash.expect("previous guard ensures is_some; qed");
+					if *self.last_proposed.read() == bh {
```
arkpar: Are `self.step` and `self.last_proposed` always locked in the same order?

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -462,12 +468,12 @@ impl Engine for Tendermint {
...
 			let sender = public_to_address(&recover(&message.signature.into(), &rlp.at(1)?.as_raw().sha3())?);
 			if !self.is_authority(&sender) {
-				Err(EngineError::NotAuthorized(sender))?;
+				return Err(From::from(EngineError::NotAuthorized(sender)));
 			}
 			self.broadcast_message(rlp.as_raw().to_vec());
 			if self.votes.vote(message.clone(), &sender).is_some() {
 				self.validators.report_malicious(&sender);
-				Err(EngineError::DoubleVote(sender))?
+				return Err(From::from(EngineError::DoubleVote(sender)));
```
rphmeier: `into()`?

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -491,22 +497,19 @@ impl Engine for Tendermint {
...
+			if (header.seal()[1] == ::rlp::NULL_RLP.to_vec())
+				^ (header.seal()[2] == ::rlp::EMPTY_LIST_RLP.to_vec()) {
 				Ok(())
 			} else {
-				Err(From::from(EngineError::BadSealFieldSize(OutOfBounds {
-					min: Some(1),
-					max: None,
-					found: signatures_len
-				})))
+				warn!(target: "engine", "verify_block_basic: Block is neither a Commit nor Proposal.");
```
rphmeier: this branch is also taken when the seal advertises both, right?

ethcore/src/engines/tendermint/mod.rs
arkpar: `!=` instead of `^` reads better for bools

ethcore/src/engines/tendermint/mod.rs
keorn: They are never `write` in the same context.

ethcore/src/engines/tendermint/mod.rs
keorn: Yes, but currently this is never possible.

PR #5416, opened 2017-04-06
Fixing secretstore TODOs - part 2

svyatonik:
On top of #5386 

Most of still pending TODOs are fixed here, including:
1) encryption session will now fail if any cluster node is down. It will also fail if some node is misbehaving. So - N in M-of-N is never autodecreased
2) added 'session timeout' for cases when node is still up (connection is established), but it does not answer to some messages, crucial for the session
3) sessions are finally removed after completion
4) decryption session is now able to auto-restart when node goes down even if it is in the middle of decryption process

I want to add some tests for these fixes, so marking as in-progress.


svyatonik:
5. Added more tests
6. Fixed 1-of-1
7. Fixed closing Parity by ^C when secretstore was working

keorn:
lgtm and seems to work.

secret_store/src/http_listener.rs
```diff
@@ -93,6 +93,13 @@ impl<T> KeyServer for KeyServerHttpListener<T> where T: KeyServer + 'static {
 	}
 }
 
+impl<T> Drop for KeyServerHttpListener<T> where T: KeyServer + 'static {
+	fn drop(&mut self) {
+		// ignore error as we are dropping anyway
+		let _ = self._http_server.close();
```
rphmeier: does this not close on drop already?

svyatonik: Nope. That was the rason behind "Fixed closing Parity by ^C when secretstore was working" issue. You must call `close()` explicitely

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -17,26 +17,54 @@
...
 	SessionParams as EncryptionSessionParams, Session as EncryptionSession};
 use key_server_cluster::io::{DeadlineStatus, ReadMessage, SharedTcpStream, read_encrypted_message, WriteMessage, write_encrypted_message};
 use key_server_cluster::net::{accept_connection as net_accept_connection, connect as net_connect, Connection as NetConnection};
 
-pub type BoxedEmptyFuture = BoxFuture<(), ()>;
+/// Maintain interval (seconds). Every MAINTAN_INTERVAL seconds node:
+/// 1) checks if connected nodes are responding to KeepAlive messages
+/// 2) tries to connect to disconnected nodes
+/// 3) checks if enc/dec sessions are time-outed
+const MAINTAN_INTERVAL: u64 = 10;
```
rphmeier: `MAINTAN` -> `MAINTAIN`

PR #5417, opened 2017-04-06
Tests and tweaks for public node middleware

maciejhirsz:
- When confirming transactions, the request will be "locked" while it awaits for account to be unlocked, thus hiding the signer dialog as soon as password has been entered.
- Webpack builds and Node.js (mocha) now use proper (browser and node respectively) versions of `keythereum` and `secp256k1`.
- Added tests for (at least some) middleware RPC implementations in JS. 
- Whenever crypto is involved the timeouts for tests are upped to 30s, although they were always below 1s on my machine since proper libraries are being used now.
- Exceptions thrown in web workers will be propagated up as rejected promises in the `WorkerPool`. Workers in the pool that encountered an exception will get unblocked and can continue to be used.


js/src/api/local/localAccountsMiddleware.js
```diff
@@ -170,13 +161,21 @@ export default class LocalAccountsMiddleware extends Middleware {
...
+
       const account = accounts.get(from);
 
       return Promise.all([
         this.rpcRequest('parity_nextNonce', [from]),
         account.decryptPrivateKey(password)
       ])
       .then(([nonce, privateKey]) => {
+        if (!privateKey) {
+          transactions.unlock(id);
```
maciejhirsz: Yap, good catch!

PR #5418, opened 2017-04-06
New state tests

arkpar:


ethcore/src/tests/helpers.rs
```diff
@@ -359,12 +359,23 @@ pub fn get_temp_state() -> GuardedTempResult<State<::state_db::StateDB>> {
...
+	State::new(journal_db, U256::from(0), Default::default())
+}
+
 pub fn get_temp_state_db_in(path: &Path) -> StateDB {
 	let db = new_db(path.to_str().expect("Only valid utf8 paths for tests."));
 	let journal_db = journaldb::new(db.clone(), journaldb::Algorithm::EarlyMerge, ::db::COL_STATE);
 	StateDB::new(journal_db, 5 * 1024 * 1024)
 }
 
+pub fn get_temp_mem_state_db() -> StateDB {
```
rphmeier: Why not just replace the other functions with this implementation? `State` and `StateDB` They don't have access to the underlying database anyway, only the `KeyValueDB` trait interface.

arkpar: This requires a lot of changes that are out of scope of this PR. I'll see if I can do that in another one.

gavofyork:
what a mess `state.rs` is now - can't we ditch the huge list in favour of some more automated system?

Issue #5311, opened 2017-03-28
OnDemand dispatch improvements


rphmeier:
Currently using `OnDemand` is a bit of a nuisance due to the required network context on all dispatch calls, and it isn't very efficient, dispatching only one request per packet.

A couple of small changes which would go a long way towards addressing this:
1. decouple submitting requests and dispatch to network. this will make chaining results easier.
2. have dispatch fetch groups of unrelated requests at a time when sending to peers (to sidestep the problem of back-reference dispatching for the time being)

5chdn:
This issue is labelled with _annoyance_: The client behaves within expectations, however this "expected behaviour" itself is at issue. It's inactive since April, neither assigned nor linked to any milestone.

@paritytech/core-devs I see there is a (partial) patch available. Will this make it into 1.7? Please decide on a deadline and add an assignee within 7 days, thanks!

rphmeier:
necessary improvements (although not exactly the ones described here) have been made in #5419 and #5573 and this issue will be closed when those are reviewed and merged.

PR #5419, opened 2017-04-06
Improve on-demand dispatch and add support for batch requests

### LINKED ISSUE #4593: Tests for light `OnDemand` service
Should be doable just by moving the general network test harness in `ethsync` up a bit.

rphmeier:
After #5383, #5320
Closes #4593
Partially addresses #5311

This PR adds batch-request support to the `OnDemand` service. Batch responses are handled up to the point where they are incomplete or incorrect, and the remainder will be reassigned to another peer.

There's still a lot of room for improvement in terms of multiplexing different logical batches into different packets, along with submitting portions of batches at a time instead of submitting the whole thing (which will cause heavy requests to potentially never be served) but this come later.

Requests are now prioritized by submission order.

A strongly-typed API for `OnDemand` has been added; for tuples up to a certain size, you can submit
```rust
(request_type_a, request_type_b, request_type_c)
``` 
and get back a 
```rust
(response_type_a, response_type_b, response_type_c)
```

The RPCs haven't been ported to use the new system yet, so they should be as inefficient as ever. The next PR will do this, along with implementing back-referencing for on-demand requests, where verification data can be extracted from prior responses.


rphmeier:
Strongly-typed API in place:

```rust
let (header, code, account) = on_demand.request(&ctx, (hdr_req, code_req, account_req))
    .expect("all back-references ok");
```

rphmeier:
> Back-references in on-demand requests.

I'll leave this for another PR, when we port the RPCs and other `on_demand` users over.
Tests coming soon.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -413,183 +467,68 @@ impl Handler for OnDemand {
...
 
-	#[test]
-	fn detects_hangup() {
-		let cache = Arc::new(Mutex::new(Cache::new(Default::default(), Duration::hours(6))));
-		let on_demand = OnDemand::new(cache, 0.into());
-		let result = on_demand.header_by_hash(&FakeContext, request::HeaderByHash(H256::default()));
+		// update pending fields and re-queue.
+		let capabilities = guess_capabilities(&pending.requests.requests()[num_answered..]);
+		pending.net_requests = builder.build();
+		pending.required_capabilities = capabilities;
```
tomusdrw: Shoudn't you update `pending.requests` here as well (i.e. remove the answered ones?)

ethcore/light/src/on_demand/request.rs
```diff
@@ -26,17 +26,383 @@ use ethcore::receipt::Receipt;
...
+use request::{self as net_request, IncompleteRequest, Output, OutputKind};
+
 use rlp::{RlpStream, UntrustedRlp};
-use util::{Address, Bytes, DBValue, HashDB, H256, U256};
+use util::{Address, Bytes, DBValue, HashDB, Mutex, H256, U256};
 use util::memorydb::MemoryDB;
 use util::sha3::Hashable;
 use util::trie::{Trie, TrieDB, TrieError};
 
+const SUPPLIED_MATCHES: &'static str = "supplied responses always match produced requests; qed";
```
tomusdrw: why?

ethcore/light/src/on_demand/request.rs
```diff
@@ -26,17 +26,383 @@ use ethcore::receipt::Receipt;
...
+				prover.check_response(cache, &res.receipts).map(Response::Receipts),
+			(&CheckedRequest::Body(ref prover, _), &NetResponse::Body(ref res)) =>
+				prover.check_response(cache, &res.body).map(Response::Body),
+			(&CheckedRequest::Account(ref prover, _), &NetResponse::Account(ref res)) =>
+				prover.check_response(cache, &res.proof).map(Response::Account),
+			(&CheckedRequest::Code(ref prover, _), &NetResponse::Code(ref res)) =>
+				prover.check_response(cache, &res.code).map(Response::Code),
+			(&CheckedRequest::Execution(ref prover, _), &NetResponse::Execution(ref res)) =>
+				prover.check_response(cache, &res.items).map(Response::Execution),
+			_ => Err(Error::WrongKind),
```
tomusdrw: I think it would be more future proof to enumerate all `CheckedRequest` (so that when the protocol is extended you immediatelly see all places requiring implementation), but I know it would make the code here a bit more ugly (and macro would be required):

```rust
match *self {
  CheckedRequest::Account(..) => match!(NetResponse::Account(), response) {
  }
}
```

ethcore/light/src/types/request/mod.rs
```diff
@@ -350,6 +364,33 @@ impl IncompleteRequest for Request {
 			Request::Execution(req) => req.complete().map(CompleteRequest::Execution),
 		}
 	}
+
+	fn adjust_refs<F>(&mut self, mapping: F) where F: FnMut(usize) -> usize {
+		match *self {
```
rphmeier: Yes, although there are currently tests for behavior of the individual types as well as their enum-wrapped counterparts

ethcore/light/src/on_demand/mod.rs
rphmeier: no, `pending.requests` is intentionally not shortened. although that approach would work, I thought it made more sense to keep the original requests batch around and only adjust back-references when building the `net_requests`, which should lead to slightly less work over-all.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -57,16 +57,16 @@ pub type ExecutionResult = Result<Executed, ExecutionError>;
 
 impl LightFetch {
 	/// Get a block header from the on demand service or client, or error.
-	pub fn header(&self, id: BlockId) -> BoxFuture<Option<encoded::Header>, Error> {
+	pub fn header(&self, id: BlockId) -> BoxFuture<encoded::Header, Error> {
```
rphmeier: yep, although I think optimizing this stuff is beyond the scope of the PR. Most of these RPC helpers are redundant with batch requests after #5573

ethcore/light/src/types/request/builder.rs
```diff
@@ -65,44 +73,28 @@ impl RequestBuilder {
 
 /// Requests pending responses.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Requests {
+pub struct Requests<T: IncompleteRequest> {
```
rphmeier: I'll fix that, it's definitely unidiomatic

ethcore/light/src/on_demand/request.rs
rphmeier: https://github.com/paritytech/parity/pull/5419/files/0fd3e36c23123ddd7f33b3891979214e3f23ceb0#diff-abb081d3b17a1ecf465e845b3fef0728R354

this match ensures that a response is invalid if it doesn't have the correct type for the request.

ethcore/light/src/on_demand/request.rs
rphmeier: honestly, that looks cleaner and I never even considered it :)

ethcore/light/src/types/request/builder.rs
```diff
@@ -130,12 +138,35 @@ impl Requests {
 
 		self.answered += 1;
 
-		// fill as much of the next request as we can.
-		if let Some(ref mut req) = self.requests.get_mut(self.answered) {
+		// fill as much of each remaining request as we can.
+		for req in self.requests.iter_mut().skip(self.answered) {
```
rphmeier: We do iterate over all remaining requests upon each supplied response, but as you guessed it will only request outputs that haven't already been filled.

An optimization would be to just fill the next request upon each supplied response and then do a sweep of all unanswered requests after supplying all responses.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -22,174 +22,191 @@
...
 	/// Returns the chain score.
-	pub fn chain_score_by_number(&self, ctx: &BasicContext, req: request::HeaderProof) -> Receiver<U256> {
-		let (sender, receiver) = oneshot::channel();
+	pub fn chain_score_by_number(&self, ctx: &BasicContext, req: request::HeaderProof) -> BoxFuture<U256, Canceled> {
 		let cached = {
 			let mut cache = self.cache.lock();
 			cache.block_hash(&req.num()).and_then(|hash| cache.chain_score(&hash))
 		};
 
 		match cached {
```
rphmeier: All these methods are removed and replaced with a more general `answer_from_cache` in #5573, would prefer to leave this refactor specifically to that PR.

ethcore/light/src/on_demand/request.rs
tomusdrw: Might be worth including the function name in the proof then. Something like:
`check_request called earlier so supplied response matches produced request; qed`

PR #5422, opened 2017-04-06
Add decryption to the UI (in the Signer)

ngotchac:
Using `parity_decryptMessage` RPC call would crash the Signer.

NB : The QR Signer code might not work with these methods (signing and decrypting)

### DIFF
--- a/js/src/api/format/output.js
+++ b/js/src/api/format/output.js
@@ -216,8 +216,8 @@ export function outSignerRequest (request) {
           break;
 
         case 'payload':
-          request[key].decrypt = outTransaction(request[key].decrypt);
-          request[key].sign = outTransaction(request[key].sign);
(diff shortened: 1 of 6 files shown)


js/src/api/format/output.js
```diff
@@ -216,6 +216,8 @@ export function outSignerRequest (request) {
           break;
 
         case 'payload':
+          request[key].decrypt = outTransaction(request[key].decrypt);
+          request[key].sign = outTransaction(request[key].sign);
```
tomusdrw: `sign` or `decrypt` is not really a transaction, is it?

ngotchac: That's true, it was out of laziness... Will update !

ngotchac: Updated!

PR #5423, opened 2017-04-07
Add raw hash signing

keorn:
`parity_signMessage(address, password, message_hash)` in `parity_accounts`

### DIFF
--- a/js/src/jsonrpc/interfaces/parity.js
+++ b/js/src/jsonrpc/interfaces/parity.js
@@ -1697,7 +1697,7 @@ export default {
   },
 
   signMessage: {
-    desc: 'Sign a hashed-message byt the given account.',
+    desc: 'Sign the hashed message bytes with the given account.',
     params: [
       {
         type: Address,
--- a/rpc/src/v1/helpers/dispatch.rs
+++ b/rpc/src/v1/helpers/dispatch.rs
@@ -474,7 +474,7 @@ pub fn execute<D: Dispatcher + 'static>(
 					.map(ConfirmationResponse::SignTransaction)
 				).boxed()
 		},
-		ConfirmationPayload::Signature(address, mut data) => {
+		ConfirmationPayload::PostSignTransaction(address, mut data) => {
 			let mut message_data =
 				format!("\x19Ethereum Signed Message:\n{}", data.len())
 				.into_bytes();
@@ -574,8 +574,8 @@ pub fn from_rpc<D>(payload: RpcConfirmationPayload, default_account: Address, di
 		RpcConfirmationPayload::Decrypt(RpcDecryptRequest { address, msg }) => {
 			future::ok(ConfirmationPayload::Decrypt(address.into(), msg.into())).boxed()
 		},
-		RpcConfirmationPayload::Signature(RpcSignRequest { address, data }) => {
-			future::ok(ConfirmationPayload::Signature(address.into(), data.into())).boxed()
+		RpcConfirmationPayload::PostSignTransaction(RpcSignRequest { address, data }) => {
+			future::ok(ConfirmationPayload::PostSignTransaction(address.into(), data.into())).boxed()
 		},
 	}
 }
--- a/rpc/src/v1/helpers/dispatch.rs
+++ b/rpc/src/v1/helpers/dispatch.rs
@@ -474,7 +474,7 @@ pub fn execute<D: Dispatcher + 'static>(
 					.map(ConfirmationResponse::SignTransaction)
 				).boxed()
 		},
-		ConfirmationPayload::PostSignTransaction(address, mut data) => {
+		ConfirmationPayload::EthSignMessage(address, mut data) => {
 			let mut message_data =
 				format!("\x19Ethereum Signed Message:\n{}", data.len())
 				.into_bytes();
@@ -574,8 +574,8 @@ pub fn from_rpc<D>(payload: RpcConfirmationPayload, default_account: Address, di
 		RpcConfirmationPayload::Decrypt(RpcDecryptRequest { address, msg }) => {
 			future::ok(ConfirmationPayload::Decrypt(address.into(), msg.into())).boxed()
 		},
-		RpcConfirmationPayload::PostSignTransaction(RpcSignRequest { address, data }) => {
-			future::ok(ConfirmationPayload::PostSignTransaction(address.into(), data.into())).boxed()
+		RpcConfirmationPayload::EthSignMessage(RpcSignRequest { address, data }) => {
+			future::ok(ConfirmationPayload::EthSignMessage(address.into(), data.into())).boxed()
 		},
 	}
 }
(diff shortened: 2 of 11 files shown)


rpc/src/v1/helpers/dispatch.rs
```diff
@@ -474,7 +474,7 @@ pub fn execute<D: Dispatcher + 'static>(
 					.map(ConfirmationResponse::SignTransaction)
 				).boxed()
 		},
-		ConfirmationPayload::Signature(address, mut data) => {
+		ConfirmationPayload::PostSignTransaction(address, mut data) => {
```
tomusdrw: It doesn't necessarily need to be a transaction here. It's just a raw data to sign.

keorn: It does return it in a pretty specific way though, not sure what that should be called: https://github.com/paritytech/parity/blob/64a724092ebda0bea5b1f8e6a60ec6d0fe4869ec/rpc/src/v1/helpers/dispatch.rs#L477

tomusdrw: The return type looks like it's just `H520` signature, no? And to sign arbitrary data you hash it first.
Not sure I follow, though...

keorn: Doesn't it `"\x19Ethereum Signed Message:\n{}"` and turns `rsv` into `vrs`? Its definitely not a signature of an arbitrary hash. (return type is still indeed `Signature`)

tomusdrw: Yeah, but apparently that's how the signature is suppose to look like to be compatible with other clients.
So it's a signature of some data with prepended text and definitely not a transaction hash as the change you did suggests. (`Signature -> PostSignTransaction`)

tomusdrw:
What would be this method used for?

rpc/src/v1/helpers/dispatch.rs
gavofyork: +1 - definitely shouldn't be called `PostSignTransaction` - it's definitely not signing a transaction of any variety.

keorn: Thus "not sure what that should be called". Renamed to `EthSignMessage`.

keorn:
@tomusdrw For interacting with secret store for now.

MicahZoltu:
Why is the passphrase passed through the RPC instead of using Parity's existing signing mechanism (which prompts the user for passphrase)?  Is this intended to bypass Parity's signer?  If so, is this addition intended for development/testing only?  If so, are there any safeguards put into place to make it so dApps can't/don't try to use this?

If this is intended for dApp use, then I would argue that having dApps ask for user secrets is a really bad pattern to fall into.  Users should be trained to never give their secrets to a dApp directly and as soon as some legitimate dApps start using this, it will open the doors to malicious dApps leveraging the poorly trained users.

I _would_ like to see the ability to have Parity sign arbitrary messages that is exposed to dApps.  However, I believe that such a method **MUST** go through Parity's existing signing scheme and not some separate signing scheme (such as this proposal).

js/src/jsonrpc/interfaces/parity.js
```diff
@@ -1694,5 +1694,31 @@ export default {
       desc: 'Decrypted message.',
       example: withComment('0x68656c6c6f20776f726c64', 'hello world')
     }
+  },
+
+  signMessage: {
+    desc: 'Sign a hashed-message byt the given account.',
```
gavofyork: typo

gavofyork:
absolutely agree with @MicahZoltu this should not be exposed to dapps. if it really is necessary to provide a password for signing, then it belongs in a sensitive API set such as personal or accounts and certainly not under the safe set, parity: no dapp-visible RPCs should access a user's password.

keorn:
@gavofyork Not sure what you mean, its under `parity_accounts` as agreed. This is for server use.
@MicahZoltu We have multiple methods like this i.e. `personal_sendTransaction`. We will most likely add another version of this RPC that does not take the password and goes through the signer.

rpc/src/v1/helpers/requests.rs
```diff
@@ -113,8 +113,8 @@ pub enum ConfirmationPayload {
 	SendTransaction(FilledTransactionRequest),
 	/// Sign Transaction
 	SignTransaction(FilledTransactionRequest),
-	/// Sign request
-	Signature(Address, Bytes),
+	/// Sign complete transaction
```
gavofyork: not a transaction: it's an arbitrary message with a security prefix.

rpc/src/v1/impls/signing.rs
```diff
@@ -137,10 +137,10 @@ impl<D: Dispatcher + 'static> SigningQueueClient<D> {
 impl<D: Dispatcher + 'static> ParitySigning for SigningQueueClient<D> {
 	type Metadata = Metadata;
 
-	fn post_sign(&self, meta: Metadata, address: RpcH160, data: RpcBytes) -> BoxFuture<RpcEither<RpcU256, RpcConfirmationResponse>, Error> {
+	fn post_sign_transaction(&self, meta: Metadata, address: RpcH160, data: RpcBytes) -> BoxFuture<RpcEither<RpcU256, RpcConfirmationResponse>, Error> {
```
gavofyork: incorrect name

rpc/src/v1/impls/signing_unsafe.rs
```diff
@@ -121,7 +121,7 @@ impl<D: Dispatcher + 'static> ParitySigning for SigningUnsafeClient<D> {
 			.boxed()
 	}
 
-	fn post_sign(&self, _: Metadata,  _: RpcH160, _: RpcBytes) -> BoxFuture<RpcEither<RpcU256, RpcConfirmationResponse>, Error> {
+	fn post_sign_transaction(&self, _: Metadata,  _: RpcH160, _: RpcBytes) -> BoxFuture<RpcEither<RpcU256, RpcConfirmationResponse>, Error> {
```
gavofyork: incorrect name.

rpc/src/v1/traits/parity_signing.rs
```diff
@@ -28,7 +28,7 @@ build_rpc_trait! {
 		/// Posts sign request asynchronously.
 		/// Will return a confirmation ID for later use with check_transaction.
 		#[rpc(meta, name = "parity_postSign")]
-		fn post_sign(&self, Self::Metadata, H160, Bytes) -> BoxFuture<Either<U256, ConfirmationResponse>, Error>;
+		fn post_sign_transaction(&self, Self::Metadata, H160, Bytes) -> BoxFuture<Either<U256, ConfirmationResponse>, Error>;
```
gavofyork: incorrect rename

gavofyork:
@keorn you're right there - my bad, i got confused with the parity.js that was referenced.

however the PR is still wrong; it seems to do three things - add `parity_signMessage` (fine), rename `Signature` -> `EthSignMessage` (fine) and then rename `post_sign` -> `post_sign_transaction` (incorrect).

please revert the latter set of renames, and ensure docs stay correct.

keorn:
Indeed, sry.

Issue #5427, opened 2017-04-07
Warp/Light-friendly Authority Contract chains


rphmeier:
cc @keorn 

Needs to be introduced soon-ish, but there are still a couple small problems I'd like to iron down first.
Opening this issue for discussion.

## Specification

Add a seal field: "authority_transition":
If the authorities set is the same as the prior block, the field's value is empty.
If the authorities set has changed since the last block, the field's value is set to:
    `sha3([nonce: P, [authority1: B_20, authority2: B_20, ...] ])`
`nonce` must be N for the Nth transition, and the authorities' addresses must be in lexicographical order.

This will allow light clients to indicate when they need to get a fraud proof of the transition, which consists of two parts:
  1. fraud proof that the authority set at the previous block is equal to the last known one
  2. fraud proof that the authority set at the current block matches the `authorities_hash` field.

Additionally, light clients will need to store the block number of the last transition along with each CHT root. Then, when re-organizing outside of recent history, a proof of the authority set at that block can be fetched.

Warp sync would work with "consensus proof chunks", which in this case would contain
  1. `transition_headers`: a list of all block headers where the authority set has ever changed
  2. fraud proofs as mentioned above for the transitions in the list.

`len(transition_headers)` must be equal to `transition_headers.last().authority_transition.nonce`
In other words, the snapshot must provably contain proofs for all transitions since the genesis.
This could be optimized with hardcoded checkpoints in the future.

## Problems

Once an authority set has been put in place, any authority within it can continue producing a chain light clients will trust for as long as they like, regardless of whether the "main" chain has forked away. Chain score will have to favor chains with more authorities signing on them.

Bad authorities could silently change and revert the authorities set without being detected, but that's not really a problem with a good authority contract. Also, blocks from the un-mentioned authorities wouldn't be trusted.

We'll also have to save all authority sets (or maybe deltas) to disk on the light client.
If the set changes a lot, this could lead to a high storage requirement.

## Alternatives

Instead of adding a seal field, signal changes in the authority set with log events. This will save block space on an additional field, but would require going to the network to check receipts, and handling false positives. Potentially easy to manufacture false positives to inconvenience light clients, but if the log events had to include the parent hash I don't think people would be able to compute bloom collisions quickly enough.

In this case, the authorities list can be put into the log data. Then light clients can recover ancient, discarded authority sets from block receipts without re-doing the state proofs.

[...]

PR #5454, opened 2017-04-13
Groundwork for generalized warp sync

rphmeier:
Makes changes to the `engines` module that will allow warp sync, ancient block download, and light client sync to work for even complex engines.

Engines now detect and generate data for epoch changes. State proofs validating this data, along with the transitions themselves, are stored in the blockchain extras DB.

Snapshot block chunks are repurposed to a more general "secondary chunks" which engine-specific  snapshot components will create and interpret as a means to corroborate the data within the snapshot state chunks. For proof-of-work engines, these chunks will be as specified in the wiki.

PoA engines like Aura and BasicAuthority will include all transitions in the snapshot, as described in #5427.
Ancient block download/light sync will make use of an `EpochVerifier` that caches an in-memory list of validators derived from the validated transitions obtained in the snapshot (or in the light client case, from the network). When epoch changes occur, we simply obtain a new `EpochVerifier` for that epoch. Ethash epochs don't currently make use of this mechanism, and just pretend that the chain is under one "epoch" so we never need to swap, but could be altered to in the future to ensure our DAG caching is completely optimal.

This approach isn't completely resilient (and indeed, won't really work for something like rinkeby's clique consensus; for that I'd wait on introducing engine-specific storage and have `clique` snapshots contain the entire headerchain.), but it will get the job done for now.

The "genesis epoch" has number 0, so any epochs beyond that _must_ start with `1`, otherwise warp and light sync will break.


ethcore/src/engines/authority_round.rs
```diff
@@ -520,7 +586,33 @@ mod tests {
...
 		assert!(engine.verify_block_family(&header, &parent_header, None).is_ok());
+		assert!(engine.verify_block_external(&header, None).is_ok());
 		header.set_seal(vec![encode(&5usize).to_vec(), encode(&(&*signature as &[u8])).to_vec()]);
+		assert!(engine.verify_block_family(&header, &parent_header, None).is_ok());
+		assert!(engine.verify_block_external(&header, None).is_err());
+	}
+
+	#[test]
+	fn rejects_step_backwards() {
+				let tap = AccountProvider::transient_provider();
```
keorn: indent

rphmeier:
To give a quick run-down of what's still left to do:
  - Implementation of snapshot chunk and restore traits for an authority-based method.
  - Implementation of ancient block restoration which uses an `EpochVerifier` that changes whenever a header is a transition. Perhaps this will be done via another trait hanging off of `SnapshotComponents`
  - Ensure restoration database always contains genesis epoch data. Otherwise, contracts seeded with initial validators won't work.
  - Tests!

rphmeier:
@keorn 

I'm mildly concerned about how `Multi` validator sets will play; there's a requirement that (for correct behavior) epochs should increase monotonically. My solution for now was to add the "inner" set's epoch number to the block number at which the transition occurred, although this will break for nested multi-sets and seems fragile.

ethcore/src/engines/validator_set/multi.rs
```diff
@@ -40,42 +41,73 @@ impl Multi {
...
+			EpochChange::Yes(num, proof) => EpochChange::Yes(set_block + num, proof),
+			other => other,
+		}
+	}
+
+	fn epoch_proof(&self, header: &Header, caller: &Call) -> Result<Vec<u8>, String> {
+		self.correct_set_by_number(header.number()).1.epoch_proof(header, caller)
+	}
+
+	fn epoch_set(&self, header: &Header, proof: &[u8]) -> Result<(u64, super::SimpleList), ::error::Error> {
```
keorn: Maybe pass the previous block epoch nonce and return `ValidatorSet` instead of `SimpleList`? Then you can bump the nonce on recursive call.

rphmeier: main problem there is that it's hard to say what the previous block epoch nonce would be without recovering it; meaning you have to package two state proofs for each transition in warp/light rather than one.

rphmeier: right now, I think it's pretty fair to say that nested multi-sets don't make any sense -- we might just want to rewrite the validator set constructor to take a flag indicating whether we're at the top level (set false before recursing) and throw an error on nested multi set.

keorn: Makes sense, though this makes `SimpleList` a special fundamental `ValidatorSet` rather than just another trait implementation. But I guess that's fine since there are no other scenarios yet.

rphmeier: I'm fine with generalizing a bit later, but all current validator sets can express themselves as a `SimpleList` at any given block, and we need the property of fast verification of blocks out-of-order for the epoch verifier.

ethcore/src/client/client.rs
```diff
@@ -380,6 +390,12 @@ impl Client {
 			return Err(());
 		};
 
+		let verify_external_result = self.verifier.verify_block_external(header, &block.bytes, engine);
+		if let Err(e) = verify_external_result {
+			warn!(target: "client", "Stage 4 block verification failed for #{} ({})\nError: {:?}", header.number(), header.hash(), e);
```
rphmeier: (changed that to stage 5)

rphmeier:
(tagging on behalf of @keorn)

PR #5456, opened 2017-04-14
Generic PubSub implementation

tomusdrw:
This allows to subscribe to any RPC method. Parity does the polling internally (every 1second).

In future we can optimize it, most of the cases can be polled for instance when new block arrives.

@ngotchac Maybe UI can switch for that, might be much better especially for hosted environments.


rpc/src/v1/traits/pubsub.rs
```diff
@@ -0,0 +1,39 @@
...
+
+build_rpc_trait! {
+	/// Parity-specific PUB-SUB rpc interface.
+	pub trait PubSub {
+		type Metadata;
+
+		#[pubsub(name = "parity_subscription")] {
+			/// Subscribe to changes of any RPC method in Parity.
+			#[rpc(name = "parity_subscribe")]
+			fn parity_subscribe(&self, Self::Metadata, Subscriber<Value>, String, Params);
```
gavofyork: Shouldn't `SubscriptionId` be returned somewhere?

rpc/src/v1/metadata.rs
```diff
@@ -14,20 +14,41 @@
...
+	fn eq(&self, other: &Metadata) -> bool {
+		if self.origin != other.origin {
+			return false;
+		}
+
+		if self.session.is_some() != self.session.is_some() {
+			return false;
+		}
+
+		true
```
gavofyork: `self.session`'s value doesn't have to be equal?

rpc/src/v1/traits/pubsub.rs
tomusdrw: The way it works is through `Subscriber::assign_id` method. Two reasons for that:
1. You can do it asynchronously (that's a shitty argument, cause it could just return `Future`)
2. Only after you assign ID you can start sending notification and it's enforced by the signature:
`fn assign_id(self, id: SubscriptionId) -> Sink<T>`

Dropping `Subscriber` rejects the subscription as does `Subscriber::reject(self, Reason)` method.

rpc/src/v1/metadata.rs
tomusdrw: Actually `ParitalEq` implementation is not needed at all. Thanks for spotting!

rpc/src/v1/metadata.rs
```diff
@@ -14,20 +14,40 @@
...
+#[derive(Clone, Default, Debug)]
 pub struct Metadata {
 	/// Request origin
 	pub origin: Origin,
+	/// Request PubSub Session
+	pub session: Option<Arc<Session>>,
 }
 
+// impl cmp::PartialEq for Metadata {
+// 	fn eq(&self, other: &Metadata) -> bool {
```
rphmeier: if it can be removed, better to remove it outright rather than comment it out

rpc/src/v1/impls/pubsub.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+impl<S: core::Middleware<Metadata>> PubSubClient<S> {
+	/// Creates new `PubSubClient`.
+	pub fn new(rpc: MetaIoHandler<Metadata, S>, remote: Remote) -> Self {
+		let poll_manager = Arc::new(RwLock::new(GenericPollManager::new(rpc)));
+		let pm2 = poll_manager.clone();
+
+		let timer = tokio_timer::wheel()
+			.tick_duration(Duration::from_millis(500))
```
rphmeier: Any reason to set the tick duration to 500ms when the only timer spawned has a granularity of 1 second?

rpc/src/v1/impls/pubsub.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl<S: core::Middleware<Metadata>> PubSub for PubSubClient<S> {
+	type Metadata = Metadata;
+
+	fn parity_subscribe(&self, mut meta: Metadata, subscriber: Subscriber<core::Value>, method: String, params: core::Params) {
+		// Make sure to get rid of PubSub session otherwise it will never be dropped.
+		meta.session = None;
+
+		let mut poll_manager = self.poll_manager.write();
+		let (id, receiver) = poll_manager.subscribe(meta, method, params);
```
rphmeier: Any protection against subscribing to `parity_subscribe`?

rpc/src/v1/metadata.rs
tomusdrw: Obviously, sorry!

rpc/src/v1/impls/pubsub.rs
tomusdrw: In worst case it would cause 1 second delay before first poll is triggered, I thought that 500ms is a bit better tradeoff.

rpc/src/v1/impls/pubsub.rs
tomusdrw: Here: https://github.com/paritytech/parity/pull/5456/files#diff-a1469f4029621b0ad3f044a1020baa19R278

rphmeier:
I have a couple questions about how we can improve pub-sub in future PRs:
  - Being able to subscribe to any method is great. But the periodic polling mechanism isn't optimal for certain subscriptions (new block, new logs, getWork) -- how can we generalize over methods which can be updated only but immediately upon external events?
  - Are there plans to add configurable polling times per subscription or at least overall?

tomusdrw:
I've made a list of all RPC methods we support, and groupped them by possible cause of change:
https://gist.github.com/tomusdrw/5d727863c7eb90ce2b64dcea3fe0310d

We can later on optimize polling if following events are exposed by Parity:
- New block (header)
- Change in accounts (new/remove/setName/etc)
- Sync events (peers, syncing, listening)
- Pending Transactions (added/removed from the queue/pending block)

This will cover most of the cases, for the rest we can still rely on polling, maybe polling time could be configured depending on the methods, but I don't really want to expose polling timeout to RPC user.

To implement proper poll-on-event system we can start with a list of string method names, and later maybe allow some metadata in RPC method definition (like: `#[rpc(name="parity_upgradeReady", meta="event:upgrade")]`)?

PR #5460, opened 2017-04-16
Chinese Translation

btceth:
Chinese translation is done, please merge it.


rphmeier:
Conflicts need resolving.

jacogr:
Brilliant, thank you very much. I've tested both versions and it seems to work as expected. The easiest way to test before we enable it in master (which we will do shortly), is to run Parity with

`parity --ui-no-validation`

Then start the UI with

`npm start` in the `js` folder

Changing the language can be done in Settings->Parity

Some small issues remain -

1. As it stands is fails linting, mostly it seems to to comment spacing (running `npm run lint` throws up a number of these small issues)
2. The good news is that these are easily fixable, you can just run `npm run lint:fix` in the `js` folder and it will automatically fix the issues.

Overall - great work, once the linting is resolved can merge.

jacogr:
Will merge as-is and have a follow-up PR to restore failing linting issues.

PR #5455, opened 2017-04-13
Update links to etherscan.io

dstarcev:
- Update links to etherscan.io to reflect recent changes (e.g. separate domain for ropsten)
- Remove duplicated code
- Add some tests

PR #5466, opened 2017-04-18
Change etherscan testnet to ropsten links

### LINKED ISSUE #5464: Parity wallet running in ropsten should use ropsten.etherscan.io
Hi!

I'm using Parity and Parity Walletwith Ropsten, but in the wallet i cannot see the related transactions in etherscan because looks like is searching in testnet.etherscan instead of ropsten.etherscan. 
It will be posible to detect the network and use the appropiate 'blockexplorer'?

Thanks

5chdn:
resolves #5464


tomusdrw:
Duplicate of #5455 ?

5chdn:
Yes. We can close this in favor of #5455 if we can get the CLA website up again. :)

PR #5477, opened 2017-04-19
EIP-140 REVERT opcode

arkpar:
https://github.com/ethereum/EIPs/pull/206

### DIFF
--- a/ethcore/src/evm/evm.rs
+++ b/ethcore/src/evm/evm.rs
@@ -104,8 +104,25 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 pub enum GasLeft<'a> {
 	/// Known gas left
 	Known(U256),
-	/// Return instruction must be processed.
-	NeedsReturn(U256, &'a [u8]),
+	/// Return or Revert instruction must be processed.
+	NeedsReturn {
+		/// Amount of gas left.
+		gas_left: U256,
+		/// Return data buffer.
+		data: &'a [u8],
+		/// Apply or revert state changes on revert.
+		apply_state: bool
+	},
+}
+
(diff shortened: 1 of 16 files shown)


ethcore/src/evm/evm.rs
```diff
@@ -104,8 +104,25 @@ pub type Result<T> = ::std::result::Result<T, Error>;
...
+		data: &'a [u8],
+		/// Apply or revert state changes on revert.
+		apply_state: bool
+	},
+}
+
+/// Finalization result. Gas Left: either it is a known value, or it needs to be computed by processing
+/// a return instruction.
+#[derive(Debug)]
+pub struct FinalizationResut {
```
rphmeier: `Result`

ethcore/src/evm/interpreter/mod.rs
```diff
@@ -84,8 +84,13 @@ enum InstructionResult<Gas> {
 	Ok,
 	UnusedGas(Gas),
 	JumpToPosition(U256),
-	// gas left, init_orf, init_size
-	StopExecutionNeedsReturn(Gas, U256, U256),
+	// gas left, init_off, init_size
+	StopExecutionNeedsReturn {
```
rphmeier: comment here could indicate that `apply` means "apply changes"

ethcore/src/evm/evm.rs
```diff
@@ -114,14 +131,18 @@ pub enum GasLeft<'a> {
 /// `Reult<GasLeft<'a>>`.
 pub trait Finalize {
 	/// Consume the externalities, call return if necessary, and produce a final amount of gas left.
```
NikVolf: Finalization result is more than just gas now, doc comment should reflect this?

PR #5486, opened 2017-04-21
UI 2

jacogr:
- Package split @parity/{api, abi, dapps, jsonrpc, shared, shell, ui, views}
- Wallet split into seperate dapps
- Material-ui replaced with Semantic-UI

Known issues -

- [ ] Chrome extension embed.html does not get built
- [ ] Signer dapp does not use the new components from js-ui yet (aka it is non-operational)
- [ ] Dapps are not published to the dapp registry (yet)
- [ ] Method selection dapp needs a _lot_ of work
- [ ] CI does not (yet) publish the npm packages (manual process atm)
- [ ] Bonds have an issue - it checks for the `transport._url` field (needs updating, providers used only now)
- [ ] Public node interface needs to convert to new shell middleware
- [ ] ...


js/src/api/provider/promise.js
```diff
@@ -0,0 +1,38 @@
...
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+export default class PromiseProvider {
```
maciejhirsz: Nothing wrong with this, but wouldn't it be better to just have promise based providers instead of callback based ones, especially since this is a massive refactor anyway?

js/src/shared/redux/providers/walletActions.js
```diff
@@ -0,0 +1,520 @@
...
+  const promise = owners
+    ? Promise.resolve({ value: owners })
+    : fetchWalletOwners(contract);
+
+  return promise
+    .then((result) => {
+      const owners = result.value;
+
+      return Promise
+        .all(owners.map((owner) => walletInstance.hasConfirmed.call({}, [ operation, owner ])))
```
maciejhirsz: Why not `walletInstance.hasConfirmed([operation, owner])`? Is changing the `this` context to empty object a quirk?

js/src/shared/util/wallets.js
```diff
@@ -0,0 +1,405 @@
...
+  }
+
+  static fetchDailylimit (walletContract) {
+    const walletInstance = walletContract.instance;
+
+    return Promise
+      .all([
+        walletInstance.m_dailyLimit.call(),
+        walletInstance.m_spentToday.call(),
+        walletInstance.m_lastDay.call()
```
maciejhirsz: Again, any particular reason for using `call` instead of just calling the method?

js/src/api/provider/promise.js
jacogr: In principle yes. However, we are aligning the providers with what is now becoming the standard, i.e. allowing providers from anywhere to be attached to dapps. So in principle we only expose providers that "conform to the standard" (and we can even use them, even if provided by other libraries such as web3).

5chdn:
No pressure :-p

PR #5488, opened 2017-04-21
PoA warp implementation

### LINKED ISSUE #5426: Indicate when Engine can warp or light sync
Just add `can_warp` and `can_light_sync` functions.
On startup, if we can't warp, fall back to full sync.
If we can't light sync, fail gracefully.

rphmeier:
Fixes #5426, #5427

This is the logic for warp sync chunking and restoration for PoA chains.

Snapshot components implementation can be found in `snapshot/consensus/authority.rs`.
Corresponding tests are in `snapshot/tests/proof_of_authority.rs`.

I found that some code from `Client::call` was being duplicated (specifically that for ensuring the caller's balance is high enough to cover fees), so I extracted that logic to `Executive::transact_virtual`.

Multi validator set changes also now count as epoch transitions.


keorn:
The most important transitions are `"list"` to `"contract"` and `"contract"` to `"contract"`.

rphmeier:
This is just about ready, but there's an outstanding issue where 
a) Genesis contract constructors aren't executed until `ensure_db_good`
and b) `ensure_db_good` is only called on first run

The combination of the two means that we have an invalid genesis block (with wrong state root) on every non-initial run.

This means that we'll have trouble finding peers, validating the genesis epoch transition's state proof, and more due to genesis block.

I have altered spec loading to take care of this in a separate branch, but it's inexplicably broken some tests we'll have to track down...

rphmeier:
(Currently blocked on #5523)

ethcore/src/state/mod.rs
```diff
@@ -599,7 +599,7 @@ impl<B: Backend> State<B> {
 	pub fn apply(&mut self, env_info: &EnvInfo, engine: &Engine, t: &SignedTransaction, tracing: bool) -> ApplyResult {
 //		let old = self.to_pod();
```
keorn: Left over.

ethcore/src/state/mod.rs
```diff
@@ -599,7 +599,7 @@ impl<B: Backend> State<B> {
...
 
-		let e = self.execute(env_info, engine, t, tracing)?;
+		let e = self.execute(env_info, engine, t, tracing, false)?;
 //		trace!("Applied transaction. Diff:\n{}\n", state_diff::diff_pod(&old, &self.to_pod()));
```
keorn: Left over.

ethcore/src/snapshot/tests/proof_of_authority.rs
```diff
@@ -0,0 +1,251 @@
...
+			push_block(&cur_signers, number + 1, vec![]);
+		}
+	}
+
+	client
+}
+
+#[test]
+fn fixed_to_contract() {
+	//::ethcore_logger::init_log();
```
keorn: Left over.

ethcore/native_contracts/build.rs
```diff
@@ -43,10 +45,16 @@ fn build_file(name: &str, abi: &str, filename: &str) {
 	f.write_all(code.as_bytes()).unwrap();
 }
 
+fn build_test_contracts() {
```
arkpar: Could this be built/included only in test configuration?

arkpar:
looks good in general

ethcore/native_contracts/build.rs
rphmeier: A feature would have to be added since `cfg(test)` isn't propagated down to dependencies.

rphmeier: (IMO I don't think it's worth doing since this crate barely ever needs to be recompiled)

rphmeier:
@keorn could you give this a last run-through and merge if you're happy with it?

PR #5489, opened 2017-04-21
Bump jsonrpc version

NikVolf:
should address https://github.com/paritytech/parity/issues/5470


rphmeier:
Calling this an `insubstantial` isn't really accurate. Obviously the change here corresponds to some changes in code, but the linked issue also doesn't give any indication of _what_ changed. I think it would be better to give a description of changes made and a link to code for this kind of PR in the future, since as-is it's pretty clear that nobody reviewed it but it's now in master.

NikVolf:
Well it was reviewed in jsonrpc crate
Here the diff shows nothing, so there is nothing to discuss and comment really

rphmeier:
Right, my point is that the corresponding PR should be linked to when we do bumps like this since there's usually more eyes on the PRs here.

PR #5496, opened 2017-04-24
Add the packaging metadata to build the parity snap

elopio:
This package will let you publish the latest parity in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

### DIFF
--- a/scripts/snapcraft.yaml
+++ b/scripts/snapcraft.yaml
@@ -17,6 +17,6 @@ apps:
 
 parts:
   parity:
-    source: .
+    source: ..
     plugin: rust
     build-packages: [g++, libudev-dev, libssl-dev, make, pkg-config]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,22 @@
+name: parity
+version: master
+summary: Fast, light, robust Ethereum implementation
+description: |
+  Parity's goal is to be the fastest, lightest, and most secure Ethereum
+  client. We are developing Parity using the sophisticated and cutting-edge
+  Rust programming language. Parity is licensed under the GPLv3, and can be
+  used for all your Ethereum needs.
+
+grade: devel
+confinement: strict
+
+apps:
+  parity:
+    command: parity
+    plugs: [network, network-bind]
+
+parts:
+  parity:
+    source: .
+    plugin: rust
+    build-packages: [g++, libudev-dev, libssl-dev, make, pkg-config]


elopio:
Hello!

Here at Ubuntu we are working on snaps, a new packaging format that will let you publish your software directly to the official store without having to make a complex debian package, without having to land it in debian and without having to wait for the 6 months release cycle of ubuntu. You can automate the delivery of parity in a secure way, so your users get new versions following your own timeline. And very important to your use case is that all the users will get an automatic update every time you push a new revision.

You can read more about snaps here: http://snapcraft.io/

To test this in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/parity
    $ cd parity
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous
    $ parity

(The --dangerous flag is because the snap that you build locally is not signed by the store. Once you push it to the store, your users will just have to do `$ sudo snap install parity`)

If you have any questions, please let me know.

elopio:
[clabot:check]

arkpar:
Thank you for the contribution

elopio:
Thanks for trying it @5chdn !

It's not a hard requirement. I moved the yaml to scripts, but now you have to call snapcraft from that directory. I'll update the instructions in this PR.

We are building a service for automatic builds in build.snapcraft.io. There it's a requirement to have the yaml either in the root, or in the snap dir. But we can talk about that later, and move back the yaml if you are interested in using it.

For now, I would recommend you to push an initial snap to the edge channel in the store, so you can start experimenting: https://snapcraft.io/docs/build-snaps/publish

The snaps in the edge channel are not visible in user searches, so it's great for early adopters and quick feedback. We can make a call for testing in the ubuntu community too to get more testers.

pura vida

elopio:
Thanks for the merge @arkpar :D

elopio:
Oh, hey @General-Beck! I didn't notice you had already pushed the parity snap to the store back then in 15.04. Many things have changed when we reached the LTS. Is there something I can do to help you with the new snap?

General-Beck:
@elopio Tomorrow I will add to CI. And if there are questions, I will certainly turn to you for advice

General-Beck:
@elopio see pls https://build.snapcraft.io/user/paritytech/parity
Do I create an issue on https://github.com/canonical-websites/build.snapcraft.io/issues ?

elopio:
Sorry for the late reply @General-Beck. Take a look at #5781.

5chdn:
@General-Beck @devops-parity can we please use `scripts/snapcraft.yaml` - there is no point in having this script in multiple locations.

elopio:
@5chdn the idea of moving it to snap/snapcraft.yaml is to use the build.snapcraft.io service for continuous delivery, instead of having to take care of building and publishing on your own CI system.

General-Beck:
@5chdn https://build.snapcraft.io/user/paritytech/parity
@elopio https://github.com/canonical-websites/build.snapcraft.io/issues/833

5chdn:
That's fine as long as we remove the script from `scripts/` again. This repository is already confusing enough.

PR #5498, opened 2017-04-25
[ui-2] Button

CraigglesO:
Pretty much the exact same consistency of size and structure as mui, other than primary is no longer usable.


jacogr:
Wouldn't it make more sense to just import https://react.semantic-ui.com/elements/button and convert ~/ui/Button to that for consistency? (Instead of re-doing our own UI lib)

PR #5499, opened 2017-04-25
icon-set

CraigglesO:
I set it up in 2 strides..

Advice, should I just setup all the svgs in the svg folder OR:

```
const FontAwesome = require('react-fontawesome');

class X extends Component {

  render() {
    return (
      <div className="component">

          <FontAwesome name="home" />

      </div>
    );
  }
}
```

make the necessary exports for ui/icons using react components.


jacogr:
Same comment as for Button PR, it probably makes more sense just converting the functions to use https://react.semantic-ui.com/elements/icon (create element, specify name)

PR #5505, opened 2017-04-25
EIP-210 BLOCKHASH changes

arkpar:
https://github.com/ethereum/EIPs/pull/210


ethcore/src/state/mod.rs
```diff
@@ -310,6 +311,11 @@ impl<B: Backend> State<B> {
 		Ok(state)
 	}
 
+	/// Get a VM factory that can execute on this state.
+	pub fn vm_factory(&self) -> EvmFactory {
```
NikVolf: what was the motivation to move vm factory in the state?
how the decision of choosing the vm executor depends on state?

arkpar: I did not move it in the state, it is there already. The state is currently constructed with all the `Factories` that are going to operate on it. I would agree that the state should not be concerned about factories, but refactoring this is out of the scope of this PR.

rphmeier: Yep, it's mostly there because of "State::apply" which should really be a `fn apply(&mut State, VMFactory)` in some other module.

ethcore/src/block.rs
```diff
@@ -278,10 +282,55 @@ impl<'x> OpenBlock<'x> {
 		let gas_floor_target = cmp::max(gas_range_target.0, engine.params().min_gas_limit);
 		let gas_ceil_target = cmp::max(gas_range_target.1, gas_floor_target);
 		engine.populate_from_parent(&mut r.block.header, parent, gas_floor_target, gas_ceil_target);
+		Self::push_last_hash(&mut r.block, last_hashes, engine, &parent.hash())?;
```
rphmeier: shouldn't this be something to be done in `engine.populate_from_parent`?

rphmeier: a utility for engines to call contract functions before and after the transactions within a block as a part of the state transitions seems like it would be useful generally

ethcore/src/block.rs
```diff
@@ -278,10 +282,55 @@ impl<'x> OpenBlock<'x> {
...
+				value: ActionValue::Transfer(0.into()),
+				code: state.code(&contract_address)?,
+				code_hash: state.code_hash(&contract_address)?,
+				data: Some(hash.to_vec()),
+				call_type: CallType::Call,
+			};
+			let mut ex = Executive::new(&mut state, &env_info, engine);
+			let mut substate = Substate::new();
+			let mut output = [];
+			if let Err(e) = ex.call(params, &mut substate, BytesRef::Fixed(&mut output), &mut NoopTracer, &mut NoopVMTracer) {
```
rphmeier: applying the substate, collecting garbage, etc?

ethcore/src/evm/evm.rs
```diff
@@ -83,13 +83,13 @@ impl fmt::Display for Error {
...
 		let message = match *self {
-			OutOfGas => "Out of gas",
-			BadJumpDestination { .. } => "Bad jump destination",
-			BadInstruction { .. } => "Bad instruction",
-			StackUnderflow { .. } => "Stack underflow",
-			OutOfStack { .. } => "Out of stack",
-			BuiltIn { .. } => "Built-in failed",
-			Internal(ref msg) => msg,
+			OutOfGas => "Out of gas".into(),
+			BadJumpDestination { destination } => format!("Bad jump destination {:x}", destination),
```
rphmeier: all of these should be `write!(f, "...", fmt_args)`

ethcore/src/block.rs
arkpar: As far as I understand the EIP this is more of a call with preserved state changes. Finalization is not required.

ethcore/src/block.rs
arkpar: `populate_from_parent` is supposed to populate just the header. It is a good idea to move it to the engine though.

gavofyork:
code looks fine. did anyone cross check this against either the YP or the EIP spec?

gavofyork:
@NikVolf / @rphmeier ^^^ ?

NikVolf:
@gavofyork yep!

apart from the contract bytecode which cannot be reviewed obviously

arkpar:
The bytecode is in the EIP. You can just check that it matches :)

ethcore/src/engines/mod.rs
```diff
@@ -53,6 +53,9 @@ use evm::CreateContractAddress;
 use ethkey::Signature;
 use util::*;
 
+/// Default EIP-210 contrat code
+pub const DEFAULT_BLOCKHASH_CONTRACT: &'static str = "73fffffffffffffffffffffffffffffffffffffffe33141561007a57600143036020526000356101006020510755600061010060205107141561005057600035610100610100602051050761010001555b6000620100006020510714156100755760003561010062010000602051050761020001555b610161565b436000351215801561008c5780610095565b623567e0600035125b9050156100a757600060605260206060f35b610100600035430312156100ca57610100600035075460805260206080f3610160565b62010000600035430312156100e857600061010060003507146100eb565b60005b1561010d576101006101006000350507610100015460a052602060a0f361015f565b63010000006000354303121561012d576000620100006000350714610130565b60005b1561015357610100620100006000350507610200015460c052602060c0f361015e565b600060e052602060e0f35b5b5b5b5b";
```
rphmeier: could a comment include a permanent link (to commit diff, not PR diff) of the EIP?

arkpar: done

ethcore/src/externalities.rs
```diff
@@ -131,20 +128,51 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
...
-				let r = self.env_info.last_hashes[index as usize].clone();
-				trace!("ext: blockhash({}) -> {} self.env_info.number={}\n", number, r, self.env_info.number);
-				r
-			},
-			false => {
-				trace!("ext: blockhash({}) -> null self.env_info.number={}\n", number, self.env_info.number);
-				H256::zero()
-			},
+	fn blockhash(&mut self, number: &U256) -> H256 {
+		if self.env_info.number + 256 >= self.engine.params().eip210_transition {
```
pirapira: I think `+ 256` should appear in the right hand side, not in the left hand side.

ethcore/src/externalities.rs
```diff
@@ -131,20 +128,51 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
...
+			let blockhash_contract_address = self.engine.params().eip210_contract_address;
+			let code_res = self.state.code(&blockhash_contract_address)
+				.and_then(|code| self.state.code_hash(&blockhash_contract_address).map(|hash| (code, hash)));
+
+			let (code, code_hash) = match code_res {
+				Ok((code, hash)) => (code, hash),
+				Err(_) => return H256::zero(),
+			};
+
```
pirapira: According to this comment https://github.com/ethereum/EIPs/pull/210#issuecomment-300357695 it was decided that `BLOCKHASH` instruction should return zero for blocks more-than-256 blocks old. So here should be a check on `number`.

PR #5506, opened 2017-04-25
EIP-86 update

arkpar:
A few minor changes to EIP-86 have been clarified/introduced recently.

1. New contract address gets generated from code hash for contracts created with create transactions only. `CREATE` instruction retains old behaviour (address generated from sender and nonce).
2. Contract can be created even if existing account is non-empty (has balance). The only requirement is that the code does not exist.
3. `CREATE_P2SH` renamed to `CREATE2`


gavofyork:
even now, a contract can be created at an address that exists (with a non-zero balance or even code - though the latter is practically impossible).

PR #5507, opened 2017-04-25
Semantic ui

CraigglesO:
I like the idea of fully loading in the ui. Notice how it changes the whole theme.

Primary's are a blue.


jacogr:
Couple of small cleanups done, will merge.

PR #5509, opened 2017-04-25
Icons

CraigglesO:
Notice the way semantic works in this case. It's better to inline the icons for buttons. There 164 reported instances left... Just something to think about.


jacogr:
Becomes a bit of maintenance hell... especially with 164 instances left.

For button, you can pass an `<Icon />` - it would be much simpler changing `~/ui/Icons` with each export an `Icon` and then letting it filter down. Also bear in mind that icons may be on buttons or elsewhere.

So doing

```
const FileDownloadIcon = <Icon name='download' />

...

export {
  FileDownloadicon
};
```

PS: Not sure why we have 8k lines of yarn here. (Early days, since both the yarn & semantic stuff has just gone in)

CraigglesO:
Did you notice the css issues when playing with it? Thats what was kind of holding me back...

jacogr:
Definately have a bit of css issues all-over, but don't think it is caused by this. Rather, we have overrides upon overrides everywhere. Things of these nature always look worse before they look better :)

jacogr:
Merging as-is, concept ok, additional icons still to be converted.

PR #5519, opened 2017-04-27
add snap to CI

General-Beck:
initial build snap package for parity

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -60,6 +60,20 @@ linux-stable:
     - target/release/ethstore
     - target/release/ethkey
     name: "stable-x86_64-unknown-linux-gnu_parity"
+linux-snap:
+  stage: build
+  only:
+    - snap
+  script:
+    - git clone https://github.com/paritytech/parity
+    - git submodule update --init --recursive
+    - cd scripts
+    - snapcraft
+    - sudo snap install *.snap --dangerous
+    - parity -v
+  tags:
+    - snap
+  allow_failure: true
 linux-stable-debian:
   stage: build
   image: ethcore/rust-debian:latest
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,17 +62,20 @@ linux-stable:
     name: "stable-x86_64-unknown-linux-gnu_parity"
 linux-snap:
   stage: build
+  image: ubuntu:latest
   only:
     - snap
+  before_script:
+    - apt install snapcraft
   script:
-    #- git clone https://github.com/paritytech/parity
     - git submodule update --init --recursive
     - cd scripts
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -60,6 +60,20 @@ linux-stable:
     - target/release/ethstore
     - target/release/ethkey
     name: "stable-x86_64-unknown-linux-gnu_parity"
+linux-snap:
+  stage: build
+  only:
+    - snap
+  script:
+    #- git clone https://github.com/paritytech/parity
+    - git submodule update --init --recursive
```
arkpar: This can be removed

.gitlab-ci.yml
```diff
@@ -60,6 +60,23 @@ linux-stable:
...
+  image: ubuntu:latest
+  only:
+    - snap
+  before_script:
+    - apt install snapcraft
+  script:
+    - git submodule update --init --recursive
+    - cd scripts
+    - snapcraft
+    - sudo snap install *.snap --dangerous
```
rphmeier: what is the `--dangerous` flag used for?

arkpar: see https://github.com/paritytech/parity/pull/5496#issuecomment-296789185

General-Beck: @rphmeier We will not use this flag, it is only for installation

General-Beck: need for test snap package https://gitlab.ethcore.io/parity/parity/builds/53332/artifacts/download

.gitlab-ci.yml
```diff
@@ -60,6 +60,23 @@ linux-stable:
     - target/release/ethstore
     - target/release/ethkey
     name: "stable-x86_64-unknown-linux-gnu_parity"
+linux-snap:
+  stage: build
+  image: ethcore/snapcraft:latest
```
rphmeier: ethcore should be migrated to paritytech at some point

General-Beck: @rphmeier I'm work on this https://hub.docker.com/r/parity

elopio:
@General-Beck you are using gitlab? That's very nice!

I think you could configure this also to automate the deployment like we do in travis. You would have to encrypt the credentials [1], and then do something like this: https://github.com/elopio/ipfs-snap/blob/master/.travis.yml#L10

I haven't experimented with gitlab ci for this yet, so I'd love to help.

And there's also launchpad as an alternative, which we are currently migrating to a nicer UI in build.snapcraft.io: https://snapcraft.io/docs/build-snaps/ci-integration#using-launchpad

[1] https://docs.gitlab.com/ce/ci/variables/#secret-variables

General-Beck:
@elopio Thanks, I'll try

PR #5523, opened 2017-04-28
Calculate post-constructors state root in spec at load time

rphmeier:
...fixes the issue described in https://github.com/paritytech/parity/pull/5488#issuecomment-297440063

Also uncovered a trie bug in the process. I just worked around for now, but I'll look into fixing it ASAP.

### DIFF
--- a/ethcore/src/account_db.rs
+++ b/ethcore/src/account_db.rs
@@ -78,6 +78,7 @@ pub struct AccountDB<'db> {
 
 impl<'db> AccountDB<'db> {
 	/// Create a new AccountDB from an address.
+	#[allow(dead_code)]
 	pub fn new(db: &'db HashDB, address: &Address) -> Self {
 		Self::from_hash(db, address.sha3())
 	}
@@ -131,6 +132,7 @@ pub struct AccountDBMut<'db> {
 
 impl<'db> AccountDBMut<'db> {
 	/// Create a new AccountDB from an address.
+	#[allow(dead_code)]
 	pub fn new(db: &'db mut HashDB, address: &Address) -> Self {
 		Self::from_hash(db, address.sha3())
 	}
--- a/ethcore/src/account_db.rs
+++ b/ethcore/src/account_db.rs
@@ -77,6 +77,12 @@ pub struct AccountDB<'db> {
 }
 
 impl<'db> AccountDB<'db> {
+	/// Create a new AccountDB from an address.
+	#[cfg(test)]
+	pub fn new(db: &'db HashDB, address: &Address) -> Self {
+		Self::from_hash(db, address.sha3())
+	}
+
 	/// Create a new AcountDB from an address' hash.
 	pub fn from_hash(db: &'db HashDB, address_hash: H256) -> Self {
 		AccountDB {
@@ -125,12 +131,23 @@ pub struct AccountDBMut<'db> {
 }
 
 impl<'db> AccountDBMut<'db> {
+	/// Create a new AccountDB from an address.
+	#[cfg(test)]
+	pub fn new(db: &'db mut HashDB, address: &Address) -> Self {
+		Self::from_hash(db, address.sha3())
+	}
+
 	/// Create a new AcountDB from an address' hash.
 	pub fn from_hash(db: &'db mut HashDB, address_hash: H256) -> Self {
 		AccountDBMut {
 			db: db,
 			address_hash: address_hash,
 		}
+	}
+
+	#[cfg(test)]
+	pub fn immutable(&'db self) -> AccountDB<'db> {
+		AccountDB { db: self.db, address_hash: self.address_hash.clone() }
 	}
 }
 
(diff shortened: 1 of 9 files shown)


ethcore/src/account_db.rs
```diff
@@ -78,6 +78,7 @@ pub struct AccountDB<'db> {
 
 impl<'db> AccountDB<'db> {
 	/// Create a new AccountDB from an address.
+	#[allow(dead_code)]
 	pub fn new(db: &'db HashDB, address: &Address) -> Self {
```
arkpar: Why not just remove it?

ethcore/src/spec/spec.rs
```diff
@@ -119,39 +121,42 @@ pub struct Spec {
...
+		gas_used: g.gas_used,
+		timestamp: g.timestamp,
+		extra_data: g.extra_data,
+		seal_rlp: seal_rlp,
+		constructors: s.accounts.constructors().into_iter().map(|(a, c)| (a.into(), c.into())).collect(),
+		state_root_memo: RwLock::new(Default::default()), // will be overwritten right after.
+		genesis_state: s.accounts.into(),
+	};
+
+	let _ = s.run_constructors(&Default::default(), BasicBackend(MemoryDB::new()))?;
```
arkpar: This introduces a significant hit on startup time, especially in debug build. Is it absolutely necessary to have a state root at this point? Why is it required to be calculated at load time? In most cases parity starts with a non-empty state database and genesis state root won't be required at all.

arkpar:
Test are failing

ethcore/src/spec/spec.rs
rphmeier: There are two options:
  - always compute the state root at load time, so calling `genesis_header` always gets you the correct genesis header.
  - document that `spec.genesis_header()` only returns correct values after `ensure_db_good` is called, and always extract the genesis header from the DB in those cases.

This PR implements the first option, which is more simple to understand and IMO will lead to less error-prone usage of `Spec`.

rphmeier:
Tests should pass, just looks like they were cancelled.

ethcore/src/spec/spec.rs
arkpar: I'd rather put an optional pre-computed state root into the json. Alternatively light client could cache the genesis header. Full client only needs to call `spec.genesis_header()` when the database is empty IIRC.

rphmeier: You can already put an optional state root in the spec but I don't really see the purpose, since the only way to know that state root is to calculate it in the first place. And if it's not specified, then you still have to calculate it. I don't see any situation in which it's beneficial to silently return the wrong genesis header on subsequent runs.

It's true that the `Client` only calls `spec.genesis_header` on the first run (actually, now it also calls `generate_epoch_proof` using it and will yield warnings due to the bug this PR fixes), but the sync service and snapshot service get the genesis hash and header respectively. This leads to two kinds of errors when using specs with constructors. 
1. Clients are negotiating over the wrong genesis hash at sync time, partitioning private networks into two sections: nodes which have been restarted and those which haven't yet. 
2. If snapshot restoration is done in a non-first run, the wrong genesis block is restored to disk.

IMO it's easier just to address the root of the problem and ensure we never return garbage data than make it difficult to deploy chains which need to construct contracts in the genesis.

A possible fix could be to restrict access to the genesis header from the spec without calling `ensure_db_good`.

arkpar: > You can already put an optional state root in the spec but I don't really see the purpose, since the only way to know that state root is to calculate it in the first place. 

The purpose it to avoid calculating it on each start as I've stated in the beginning of this thread.

rphmeier: Yes, but it's not really _useful_ until there's a CLI for computing the state root ahead of time. It's not like people deploying chains will work out the root with their own hacked version of Parity.

I think it's premature optimization to worry about the overhead of constructing the genesis state because it operates on an initially empty trie (meaning minimal amount of hashing relative to changes) and generally makes very few changes in the first place. I would expect it to run in fractions of a second even in debug mode, and it's only a cost at startup;

Putting that aside, I think the way to go here if we do want to make that optimization is to make `ensure_db_good` the only way to access genesis block data and then force every other service to go through the `Client`/`BlockChain` if they don't want to incur the cost.

ethcore/src/account_db.rs
rphmeier: There are other constructors here which were marked in the same way. I'm not opposed to removing.

rphmeier: It seems they are used in tests.

rphmeier:
@arkpar concerns should now be addressed

ethcore/src/state_db.rs
```diff
@@ -448,7 +448,8 @@ impl state::Backend for StateDB {
 	fn is_known_null(&self, address: &Address) -> bool {
 		trace!(target: "account_bloom", "Check account bloom: {:?}", address);
 		let bloom = self.account_bloom.lock();
-		!bloom.check(&*address.sha3())
+		let is_null = !bloom.check(&*address.sha3());
```
gavofyork: redundant change?

rphmeier: yeah, I put some tracing there when in the process of figuring out there was a trie bug, but I guess I didn't fully revert

PR #5527, opened 2017-04-28
Pass additional data when reporting

keorn:
This changes the reporting validator set interface. [Updated contracts](https://github.com/paritytech/contracts/pulls/keorn).
Not 100% sure if this is the right balance between structured and unstructured args, malicious proofs will take on very different forms most likely.


ethcore/src/engines/vote_collector.rs
```diff
@@ -319,9 +338,9 @@ mod tests {
 		let collector = VoteCollector::default();
 		let round = 3;
 		// Vote is inserted fine.
-		assert!(full_vote(&collector, H520::random(), round, Some("0".sha3()), &Address::default()).is_none());
+		assert!(full_vote(&collector, H520::random(), round, Some("0".sha3()), &Address::default()));
 		// Returns the double voting address.
-		full_vote(&collector, H520::random(), round, Some("1".sha3()), &Address::default()).unwrap();
+		assert!(!full_vote(&collector, H520::random(), round, Some("1".sha3()), &Address::default()));
```
rphmeier: these two `full_vote` calls have different block hashes. can you clarify why it's still a double vote?

keorn: this is what a double vote is: voting for two different hashes in the same voting round.

PR #5528, opened 2017-04-29
Change all occurrences of ethcore.io into parity.io

R13N:
`find . -type f -not -path './.*' -exec sed -i -e "s/ethcore\.io/parity.io/g" {} \;`
I hope I didn't break anything important.


R13N:
[clabot:check]

PR #5544, opened 2017-05-03
option to disable persistent txqueue

rphmeier:
Introduces the `--no-persistent-txqueue` flag to both clear the existing persistent transaction queue and prevent local transactions created during the current session from being saved to disk.


parity/cli/usage.txt
```diff
@@ -75,6 +75,9 @@ Operating Options:
                                    data on-demand from the network. Much lower in storage,
                                    potentially higher in bandwidth. Has no effect with
                                    subcommands (default: {flag_light}).
+  --no-persistent-txqueue          Don't save pending local transactions to disk to be
```
tomusdrw: Shouldn't it go closer to other transaction queue options?

PR #5439, opened 2017-04-11
Secretstore RPCs + integration

svyatonik:
On top of #5416 

This is PR for RPCs && small changes required for easier secretstore integration.
First change, as suggested by @ngotchac is to use ethcrypto::encrypt/decrypt with known MAC instead of ethcrypto::{encrypt/decrypt}_single_message. This will help with using RPCs like `parity_decryptMessage`.
@ngotchac I've slightly changed your commit && moved DEFAULT_MAC to ethcrypto from bith RPC && SecretStore modules. Hope you're ok with that :)

PR #5545, opened 2017-05-03
SecretStore: remove session on master node

svyatonik:
On top of #5439 
Previously successfully completed session was only removed on 'slave' nodes


ngotchac:
This fixes the hanging issue, thanks @svyatonik !

keorn:
I think the diff covers lots of old changes, if its only `805040b5a121f8ae96bf30d13c0db111eb54b90d` then its good.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1043,6 +1065,64 @@ impl ClusterClient for ClusterClientImpl {
 	}
 }
 
+impl EncryptionSessionWrapper {
+	pub fn new(cluster: &Arc<ClusterData>, session_id: SessionId, session: Arc<EncryptionSession>) -> Arc<Self> {
```
rphmeier: typical style is to pass a `Weak` directly, this way it's more evident from the function signature that it's the caller's responsibility to keep the `Arc` alive.

PR #5546, opened 2017-05-03
Mui-be-gone-3

CraigglesO:
Can't figure out why yet, but the snackbar tab won't reveal itself like it did in the test app -_-

Also it's really only used in two places currently

shell->Applicaitons

ui->errors

So, I need to update the styles to change color... will do soon.


CraigglesO:
Oops. I pulled this from a mui-be-gone to just keep working.. but it still has the dapp changes..

CraigglesO:
To fix:

- [x]  Dapps
- [ ]  Styles

PR #5554, opened 2017-05-04
Connection dialog timestamp info

jacogr:
- Add note that node/browser timestamps should be in-sync with each other and a timestamp server
- Auto-focus token input field

parity 2017-05-04 13-14-41

### DIFF
--- a/js/src/views/Connection/connection.css
+++ b/js/src/views/Connection/connection.css
@@ -49,7 +49,7 @@
 }
 
 .info {
-  margin-top: 2em;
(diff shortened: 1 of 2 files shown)


ngotchac:
Should we put a link to https://time.is/ ?

5chdn:
> Should we put a link to https://time.is/ ?

time.is offers an api on request. we could directly check the offset in the client.

tomusdrw:
Related #5604

PR #5558, opened 2017-05-04
reorg into blocks before minimum history

rphmeier:
by using the dynamic `earliest_state` counter rather than the fixed minimum history size. Also does the same for RPCs and snapshot creation.

Should backport to beta and perhaps stable if memory-based pruning is implemented there.


rphmeier:
Although this isn't a problem during import because no pruning will occur until the block is fully enacted, "riding the edge" of the pruned states with RPC calls could lead to spurious failures and perhaps console warnings if the inspected state is pruned in the middle of the call, but will lead to no panics since we bubble up trie/db errors from `State`.

PR #5559, opened 2017-05-05
dropdowns

CraigglesO:
There is a problem hopefully you can help fix.

in optionsStep.js lines 62-67:

keeps hiding behind the dropdown. Like I said before... Lappy just doesn't want to run the js today. Takes minutes to load each refresh.


CraigglesO:
updates... fixed css and added card. UI got a little worse but should be easily fixable in the future.

ddorgan:
[clabot:check]

js/src/ui/Dropdown/dropdown.js
```diff
@@ -0,0 +1,83 @@
...
+
+import styles from './dropdown.css';
+
+export default function Dropdown ({ defaultValue, disabled, fluid, icon, name, onBlur, onChange, onClick, onClose, onFocus, options, placeholder, scrolling, search, selection, text, value }) {
+  return (
+    <DropdownUI
+      className={ styles.dropdown }
+      defaultValue={ defaultValue }
+      disabled={ disabled }
+      fluid={ fluid }
```
jacogr: Prefer `fullWidth` (more descriptive) to `fluid` - in addition, should default to true.

js/src/ui/Dropdown/dropdown.js
```diff
@@ -0,0 +1,83 @@
...
+      value={ value }
+    />
+  );
+}
+
+Dropdown.propTypes = {
+  defaultValue: PropTypes.number,    // Initial value via index.
+  disabled: PropTypes.bool,          // A disabled dropdown menu or item does not allow user interaction.
+  fluid: PropTypes.bool,             // A dropdown can take the full width of its parent.
+  icon: PropTypes.any,               // Shorthand for Icon.
```
jacogr: Node or String?

js/src/views/Account/Shapeshift/OptionsStep/optionsStep.js
```diff
@@ -54,30 +54,26 @@ export default class OptionsStep extends Component {
...
-                defaultMessage='the type of crypto conversion to do'
-              />
-            }
-            label={
-              <FormattedMessage
-                id='shapeshift.optionsStep.typeSelect.label'
-                defaultMessage='fund account from'
-              />
-            }
+          <div style={ { display: 'block', paddingTop: '15px', paddingBottom: '5px', fontSize: '16px', pointerEvents: 'none', userSelect: 'none', color: 'rgba(0, 0, 0, 0.298039)' } }>
```
jacogr: Prefer that the styles are a constant and included.

js/src/ui/Dropdown/dropdown.js
```diff
@@ -0,0 +1,83 @@
...
+  name: PropTypes.func,              // Name of the hidden input which holds the value.
+  onBlur: PropTypes.func,            // Called on blur.
+  onChange: PropTypes.func,          // Called when the user attempts to change the value.
+  onClick: PropTypes.func,           // Called on click.
+  onClose: PropTypes.func,           // Called on close.
+  onFocus: PropTypes.func,           // Called on focus.
+  options: PropTypes.any,            // Array of Dropdown.Item props e.g. `{ text: '', value: '' }`
+  placeholder: PropTypes.string,     // Placeholder text.
+  scrolling: PropTypes.bool,         // A dropdown can have its menu scroll.
+  search: PropTypes.bool,            // A selection dropdown can allow a user to search through a large list of choices.
```
jacogr: Is this used? (Along with next property?)

Looking at the current actual usage, prefer fewer options (what is needed) than the kitchen-sink. Having sane defaults just makes things simpler, i.e. we now pass `fluid` everywhere it seems. Same with selection, etc.

js/src/views/Account/Shapeshift/OptionsStep/optionsStep.js
```diff
@@ -54,30 +54,26 @@ export default class OptionsStep extends Component {
...
+              id='shapeshift.optionsStep.typeSelect.label'
+              defaultMessage='Fund account from'
+            />
+          </div>
+          <Dropdown
+            placeholder='Choose a crypto-currency'
+            fluid
+            search
+            selection
+            options={ coins }
```
jacogr: As per the comment above, not convinced we need any of `fluid`, `search` & `selection` passed in for our current usage? (We can always expand, but try and keep the footprint of the UI API as small as possible)

js/src/ui/Dropdown/dropdown.js
```diff
@@ -0,0 +1,83 @@
...
+      fluid={ fluid }
+      icon={ icon }
+      name={ name }
+      onBlur={ onBlur }
+      onChange={ onChange }
+      onClick={ onClick }
+      onClose={ onClose }
+      onFocus={ onFocus }
+      options={ options }
+      placeholder={ placeholder }
```
jacogr: We use `hint` as the prop for all inputs everwhere, so it is always label, value & hint. May want to align it.

jacogr:
Only funky bit here is that the actual dropdown doesn't show the icon once it is selected (looking at ShapeShift)

Container is very borked atm, lost all padding and the hover effect makes the hover show up way off base.

CraigglesO:
I'm glad you want to remove cards where they are not necessary. That helps a lot. As for fluid, search, and selection I will make them default to true.

The rest is just small stuff. Shall be done at once!

js/src/ui/Dropdown/dropdown.js
CraigglesO: I originally had `import { nodeOrStringProptype } from '~/util/proptypes';`

but it kept error-ing until I changed it to any. No clue why. Works in every other situation I've used it.

jacogr:
1. Making sure we still pass exactly the same properties through as previously (just simplifies conversion completely)
2. Simplified, we only expose properties we need and set sane defaults for others (we can expand, but rather keep the API footprint small until such time extensions are needed)
3. Added LabelComponent that adds a label to the top (also useful for the actual Input, this now wraps the component and should do the same for input)
4. Moved Dropdown to Form, i.e. all input components are here
5. Fixup tests

Feel free to merge if happy.

PR #5564, opened 2017-05-05
Check pending request on Node local transactions

### LINKED ISSUE #5349: Tx display doesn't handle syncing very well
1. Node was syncing
2. Send tx to contract (contract interface)
3. Tx display doesn't update, even after node indicated Tx has been mined
4. UI refresh still doesn't make the display update

parity 2017-03-30 16-15-37

### LINKED ISSUE #5358: Tx minimised should not use localSorage
Get transactions for local/available from the Node as per the Signer views.

ngotchac:
Closes #5358

Might also close #5349 (@jacogr might want to try, I couldn't reproduce it)

### DIFF
--- a/js/src/views/Application/Requests/savedRequests.js
+++ b/js/src/views/Application/Requests/savedRequests.js
@@ -16,22 +16,11 @@
 
 import store from 'store';
 
-import { ERROR_CODES } from '~/api/transport/error';
-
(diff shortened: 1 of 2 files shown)


jacogr:
Should fix the issues indeed since we are using parity_localTransactions now, which is the correct approach.

Issue #5572, opened 2017-05-07
Error: "Events not supported"


lfrickel:
I take an example from the "Events" section of the OFFICIAL guide to Solidity (http://solidity.readthedocs.io/en/develop/contracts.html?highlight=event#events):
```
pragma solidity ^0.4.11;

contract ClientReceipt {
    event Deposit(address indexed _from, bytes32 indexed _id, uint _value);

    function deposit(bytes32 _id) {
        // Any call to this function (even deeply nested) can
        // be detected from the JavaScript API by filtering
        // for `Deposit` to be called.
        Deposit(msg.sender, _id, msg.value);
    }
}
```
I compile this through the Parity editor and get a number of errors:
```
"msg.value" used in non-payable function. Do you want to add the "payable" modifier to this function?
        Deposit(msg.sender, _id, msg.value);
                                 ^-------^
**9: 34
Events not supported.
    event Deposit(address indexed _from, bytes32 indexed _id, uint _value);
    ^---------------------------------------------------------------------^**
3: 5
Parameter type "bytes32" not supported.
    function deposit(bytes32 _id) {
                     ^---------^
5: 22
Only internal function calls supported.
        Deposit(msg.sender, _id, msg.value);
        ^---------------------------------^
9: 9
Not supported.
        Deposit(msg.sender, _id, msg.value);
        ^-----^
9: 9
Member access: Only call and array length supported.
        Deposit(msg.sender, _id, msg.value);
                ^--------^
9: 17
Member access: Only call and array length supported.
        Deposit(msg.sender, _id, msg.value);
                                 ^-------^
```

Someone can explain to me: HOW CAN IT BE?

**I MUST have events for my contract!**

Thankful in advance for the answer!

rphmeier:
@ngotchac aren't these just warnings from the static analyzer as opposed to the solidity compilation  itself?

5chdn:
Duplicate of #5409

ngotchac:
@rphmeier You are right. Maybe the static analyser messages should be hidden by default (could be/is misleading)

lfrickel:
First, thanks to @rphmeier, @5chdn and @raingloom for the comments.
Secondly, it seems strange to me that the validator, the compiler and the official manual have three different and conflicting opinions, based on such basic functions as events.
Thirdly, I, of course, immediately checked the work of this example, regardless of the messages of the validator.
[...]

PR #5574, opened 2017-05-08
Remove formal verification messages in Dev Contract

ngotchac:
It seems that (part of) the formal verification in Solc will be removed soon anyway (ethereum/solidity#2202), those messages are misleading: #5572 

(Keeping a reference if we want the show them in a special panel)


jacogr:
Definately has had me confused more than once. Think in the future a seperate area would be appropriate.

PR #5577, opened 2017-05-09
input change

CraigglesO:
Input change.

The way input was originally used in some dapps means that it's going to look a little weird. But I think what would be best is to update them to labels and dropdowns, rather than inputs.

Semantic dropdown has a feature where you can search. And old non-editable textfields can be changed to a label.


js/src/ui/Form/Input/input.js
```diff
@@ -66,6 +45,7 @@ export default class Input extends Component {
       'default',
       'initial'
     ]),
+    fluid: PropTypes.bool,
```
jacogr: `fullWidth` (i.e. more descriptive)?

js/src/ui/Form/Input/input.js
```diff
@@ -128,75 +118,48 @@ export default class Input extends Component {
...
+      type
+    } = this.props;
 
     return (
       <div className={ styles.container } style={ style }>
         { this.renderCopyButton() }
-        <TextField
-          autoComplete='off'
-          autoFocus={ autoFocus }
+        <InputUI
```
jacogr: I find `InputUI` slightly misleading since we have the ui package. You had `SemButton` elsewhere before changing it which was more descriptive imho, maybe `SemanticInput` if you don't like the shorter versio or `InputSUI` (as they call it)

js/src/ui/Form/Input/input.js
```diff
@@ -128,75 +118,48 @@ export default class Input extends Component {
...
           onKeyDown={ this.onKeyDown }
           onKeyUp={ this.onKeyUp }
           onFocus={ this.onFocus }
           onPaste={ this.onPaste }
-          readOnly={ readOnly }
-          ref='input'
-          rows={ rows }
-          step={ step }
-          style={ textFieldStyle }
+          placeholder={ defaultValue }
```
jacogr: placeholder === hint in the original version? (We use them everywhere as `hint`)

jacogr:
Definately needs a bit more in regards to the actual labels & hints. From Accounts -> create account
parity 2017-05-09 13-56-47

js/src/ui/Form/Input/input.js
CraigglesO: Sure, I was really insecure about which was better. But I can change them all to SemanticX

CraigglesO:
Yeah, where is this problem located? Why are they displaying like this? Something wrong with the css? It does that everywhere.

js/src/ui/Form/Input/input.js
CraigglesO: I got default value and hint mixed up -_-

js/src/ui/Form/Input/input.js
CraigglesO: I'm going to add both than. Because it's easier to just lookup in semantics docs on what to do, but meanwhile obviously its good to be backward compatible.

I just don't see the value in using a built UI and not being able to consistently refer to the docs rather than have to go digging into ui->forms to remember.

jacogr:
This one is funny, it really seems like they are not doing a great job. Fiddling with the props slightly gets it to stretch. Comments after looking at it closer -

1. We do lose critical capabilities, i.e. readOnly (readOnly !== disabled) - to get around this, we can probably pass that specific prop to the actual input field nested
2. We would need to warp the label as we do with the LabelComponent in the Dropdown
3. Remove the new props, keep to what we had (we just want a straight conversion and do the mapping as needed, i.e. no need to add fluid (at all), fullWidth is not used anywhere so can default without passing a prop, style is currently not used anywhere, so no need to pass)
4. Would probably need to do some magic on hint to pass the value to placeholder, i.e. something like done here https://github.com/paritytech/parity/pull/5562/files#diff-c2d77e92125f470e72dcfb79e2006adbR36

jacogr:
Pulled into different branch, closing.

PR #5580, opened 2017-05-09
UI server refactoring

### LINKED ISSUE #2080: Deprecate separate Signer interface
After we have:
#884 (which will expose WebSockets connection anyway)

and 
#1235 (per-request for HTTP and per-session for WS)

and
few changes in `jsonrpc-core` to support permissioned methods

Separate API for Signer will not be needed.

### LINKED ISSUE #4389: High CPU usage generated by hyper after closing connection without response
To reproduce:
```
curl --data '{"method":"eth_sendTransaction","params":[{"from":"0xb60e8dd61c5d32be8058bb8eb970870f07233155","to":"0xd4e8dd67c5d32be8058bb8eb970870f072445675","gas":"0x76c0","gasPrice":"0x9184e72a000","value":"0x9184e72a","data":"0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675"}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
```
This will trigger a popup in GUI, but instead of confirming it just `CTRL+C` curl awaiting the answer. It causes `hyper` to consume 100% of single core.

### LINKED ISSUE #5448: web3.site forwards to home.web3.site:8180
[Wiki](https://github.com/paritytech/parity/wiki/Parity-Wallet) states that one can navigate to http://web3.site/ in order to access UI but instead it forwards to home.web3.site:8180.

capture

Going straight to `home.web3.site:8080` works for me. (afaik no custom proxy settings)

tomusdrw:
- Removing separate `ethcore-signer` crate
- UI server is now standard Hyper server (with `/api` and RPC)
- UI is working as a RPC server middleware, similar to dapps.
- UI is connecting to WebSockets server using token-based scheme (as previously)
- `/api/apps` removed in favour of RPC call

Why?
- Removes special casing for domains, and CORS headers
- Clear separation
- UI works just like dapp (but it uses separate port for security reasons)
- WS server is now first class citizen - you can connect without token to get standard set of APIs or with token to get a sudo access.

Closes #4389 
Closes #2080
Closes #5448
Related #4474


js/src/index.js
```diff
@@ -54,16 +54,15 @@ if (process.env.NODE_ENV === 'development') {
...
+const parityUrl = process.env.PARITY_URL || '127.0.0.1:8546';
 
 let token = null;
 
 if (window.location.hash && window.location.hash.indexOf(AUTH_HASH) === 0) {
   token = qs.parse(window.location.hash.substr(AUTH_HASH.length)).token;
 }
 
-const api = new SecureApi(`${urlScheme}${parityUrl}`, token);
+const api = new SecureApi(parityUrl, token);
```
jacogr: Not convinced we want to lose the scheme here, it was done due to users running over HTTPS on the network. Seems like we only support localhost now?

tomusdrw: The scheme is resolved inside `SecureApi`, note that currently `parityUrl` will be just `<host>:<port>`

jacogr: Ok, I missed a step although I did see the stuff below. However with line 57 above now, we only do localhost and cannot connect to anything that is on the network?

tomusdrw: Hmm, right. So it's possible to override (in constructor) `protocol` returned by `SecureApi` in case server protocol should be different than the one the UI is running. I can refactor it and detect if protocol was specified in the URL and in such case always use that one if you prefer.

jacogr: Just worried since the connection to http was logged by a user. (Previously we did ignore protocol here as well)

tomusdrw: But now the protocol is handled correctly for all possible cases (including dapps server that had protocol hardcoded).
The only uncovered case is when someone is using `SecretApi` as a library and wants to connect to `https` server on a website hosted on `http`.

dapps/src/page/builtin.rs
```diff
@@ -27,6 +27,7 @@ pub struct PageEndpoint<T : WebApp + 'static> {
 	/// Safe to be loaded in frame by other origin. (use wisely!)
 	safe_to_embed_on: Option<(String, u16)>,
 	info: EndpointInfo,
+	html5_routing: bool,
```
maciejhirsz: something like `fallback_to_index_html` might be a better name?

dapps/src/page/builtin.rs
```diff
@@ -38,6 +39,20 @@ impl<T: WebApp + 'static> PageEndpoint<T> {
 			prefix: None,
 			safe_to_embed_on: None,
 			info: EndpointInfo::from(info),
+			html5_routing: false,
+		}
+	}
+
+	/// Creates a new `PageEndpoint` for builtin (compile time) Dapp.
+	/// Instead of returning 404 this endpoint will always server index.html.
+	pub fn with_fallback(app: T) -> Self {
```
maciejhirsz: :+1: on alternative constructor

jacogr:
JS side LGTM. Some conflicts slipped in.

maciejhirsz:
Rust side looks good to me too.

tomusdrw:
@paritytech/ci Any ideas why there is no build?

General-Beck:
@tomusdrw fixed https://github.com/paritytech/parity/commit/3ad00132c5c4c0149da356978d1034471bf00eae

tomusdrw:
Thanks!

maciejhirsz:
Still LGTM.

arkpar:
needs a rebase

NikVolf:
... and needs rebase again

PR #5583, opened 2017-05-09
Report missing author in Aura

keorn:
Detect if a block by a designated primary is missing from the chain and call `report_benign`. `report_benign` should not trigger any severe action since it can not be proven to be the fault of the given validator.


ethcore/src/engines/authority_round.rs
```diff
@@ -377,13 +376,19 @@ impl Engine for AuthorityRound {
...
 		let parent_step = header_step(parent)?;
+		// Ensure header is from the step after parent.
 		if step <= parent_step {
 			trace!(target: "engine", "Multiple blocks proposed for step {}.", parent_step);
 			self.validators.report_malicious(header.author(), header.number(), Default::default());
 			Err(EngineError::DoubleVote(header.author().clone()))?;
 		}
+		// Report if a primary was skipped.
+		if step > parent_step + 1 {
+			let skipped_primary = self.step_proposer(&parent.hash(), parent_step + 1);
```
rphmeier: why not do this in a loop for all skipped steps?

PR #5588, opened 2017-05-10
Use default account for sending transactions

### LINKED ISSUE #3135: Add ability to define which is the default account
Currently, the default account (ie appearing in the list input address field by default) is the last account that was created.

It would be nice to:
1. let the user define its default account
OR
2. let the user define a custom ordering of the accounts and consider the default account as the first according to this ordering
OR
3. use by default the last account that was used

Opinions welcome on what option you would prefer from the 3 above.

ngotchac:
Closes #3135 

In deploy/contract execution/create wallet => use the default account (set in dapps), but switch to the latest account used for those transaction if it changes

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -36,7 +36,7 @@ import ContextProvider from '~/ui/ContextProvider';
 import muiTheme from '~/ui/Theme';
 import MainApplication from './main';
 
-import { patchApi } from '~/util/tx';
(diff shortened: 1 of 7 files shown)


jacogr:
Ok as a stop-gap, but not really future-proof where we move all views to dapps. (Where we set and use it is in different contexts)

PR #5591, opened 2017-05-10
relicense RLP to MIT/Apache2

rphmeier:
As discussed with @gavofyork by request of a community member.
Published to crates.io under version `0.1.1` -- I will add the core-devs group as owners as soon as a cargo issue is patched.


rphmeier:
(quickly relicensing all deps as well)

rphmeier:
It's annoying to have two copies each of `LICENSE-MIT` and `LICENSE-APACHE` but I wanted to keep them only where really relevant to avoid confusion. Perhaps a "licenses" directory would make sense.

PR #5594, opened 2017-05-10
Proper signer Pub-Sub for pending requests.

tomusdrw:


rpc/src/v1/impls/signer.rs
```diff
@@ -18,38 +18,59 @@
...
 		signer: &Arc<SignerService>,
 	) -> Self {
+		let subscribers = Arc::new(Mutex::new(Subscribers::default()));
+		let subs = Arc::downgrade(&subscribers);
+		let s = Arc::downgrade(signer);
+		signer.queue().on_event(move |_event| {
+			if let (Some(s), Some(subs)) = (s.upgrade(), subs.upgrade()) {
+				for subscription in subs.lock().values() {
+					let subscription: &Sink<_> = subscription;
+					let _ = subscription.notify(s.requests()
```
rphmeier: this seems like it could be a pretty heavy operation. couldn't we perform the conversion once and pass a slice to to `notify` instead?

tomusdrw: Good point, thanks. Updated to clone the results instead and also added an event loop instead of `.wait()`.

Issue #5431, opened 2017-04-07
Change eth_sign to expect the user to prepend the message, not Parity.


MicahZoltu:
Since https://github.com/paritytech/parity/issues/4459 was "fixed", Parity has begun prepending the message passed to `eth_sign` with `\x19Ethereum Signed Message:\n`.  This leads to incredibly confusing behavior for the developer since the expectation is that when one callse `eth_sign(address, message)` Parity will sign `message` with the private key associated with `address`.  However, this is _not_ the behavior that actually occurs.

I recognize that this was added as a security feature to prevent dApps from getting valid transactions signed by Parity, but I think this is the wrong solution to the problem from a developer ergonomics perspective.  Instead, `eth_sign` should expect `message` to already have the magic string prependended to it.  If the magic string was not prepended, then `eth_sign` should return an error to the user clearly stating that they must prepend the `eth_sign` request with the magic string.  With this change, the system is still secure against dApps signing transactions, but it also makes it very clear to the developer (and future readers of that code) what exactly is being signed.

I just spent quite a bit of time troubleshooting why the signature I was getting back from Parity wasn't the signature I expected, only to find out that it is because Parity was not signing the message I sent it, it was signing a _different_ message.

If this issue is rejected for whatever reason, then the [documentation](https://github.com/paritytech/parity/wiki/JSONRPC-eth-module#eth_sign) should be updated to be _really_ clear about what is happening here.

On a side note, I would like to better understand the actual risks that `eth_sign` presents.  It seems that if a dApp wants a transaction signed, it can just call `eth_signTransaction`.  It is unclear to me how making `eth_sign` sign an enveloped message is really solving anything?

tomusdrw:
The reason the message is prepended is to protect your private key from [chosen-plaintext attack](https://en.wikipedia.org/wiki/Chosen-plaintext_attack).
As you noticed signing arbitrary transactions can be achieved via `eth_signTransaction`.

Documentation needs to be updated for sure, the behaviour of the message not containing the prepended text was implemented to keep it compatible with geth.

MicahZoltu:
I'm guessing since this is designed to be geth compatible there is no hope in getting it changed?

Is there some other method that can be used that will allow me to use Parity to sign a message (without a prefix)?

It seems like this attack vector only applies to unlocked accounts, and if you are running an untrusted dApp with an unlocked account they can already steal everything from you so allowing for arbitrary signing isn't really increasing attack surface area meaningfully.

[...]

PR #5597, opened 2017-05-11
Fix eth_sign signature encoding.

### LINKED ISSUE #5490: eth_sign behavior differences between Parity and other clients
Hello, we were having issues with a script running between different nodes using eth_sign and on-chain ecrecover().  Tracked down the differences and created a diagram illustrating the differences we observed here:

signature generation validation

tomusdrw:
Closes #5490

### DIFF
--- a/ethkey/src/signature.rs
+++ b/ethkey/src/signature.rs
@@ -45,27 +45,23 @@ impl Signature {
 		self.0[64]
 	}
 
-	/// Encode the signature into VRS array (V altered to be in "Electrum" notation).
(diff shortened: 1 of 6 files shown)


rphmeier:
awesome, this has been long-needed :)

debris:
lgtm!

tomusdrw:
There is a separate issue for that #5431 I think it should be an EIP/interfaces issues instead to decide together with other client implementations. Now at least we conformc to the JSONRPC spec on ethereum/wiki and we're also geth-compatible.

PR #5599, opened 2017-05-11
Don't show invalid ABI interfaces

ngotchac:
And don't show execute button on contract view if no functions to execute from the contract

### DIFF
--- a/js/src/api/contract/contract.js
+++ b/js/src/api/contract/contract.js
@@ -29,7 +29,9 @@ export default class Contract {
     }
 
     this._api = api;
-    this._abi = new Abi(abi);
+    this._abi = abi instanceof Abi
+      ? abi
+      : new Abi(abi);
 
     this._subscriptions = {};
     this._constructors = this._abi.constructors.map(this._bindFunction);
--- a/js/src/util/abi.js
+++ b/js/src/util/abi.js
@@ -16,6 +16,8 @@
 
 import { range } from 'lodash';
 
+import Abi from '~/abi';
+
 const ARRAY_TYPE = 'ARRAY_TYPE';
 const ADDRESS_TYPE = 'ADDRESS_TYPE';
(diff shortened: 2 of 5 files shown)


js/src/util/abi.js
```diff
@@ -147,3 +149,31 @@ export function parseAbiType (type) {
...
+    return new Abi(_abi);
+  } catch (error) {
+    // If there is a wrong type, filter it out
+    if (/Cannot convert .+ valid ParamType/i.test(error.message)) {
+      const validInterfaces = [];
+
+      _abi.forEach((interf) => {
+        try {
+          Abi.parseABI([ interf ]);
+          validInterfaces.push(interf);
```
jacogr: A filter would probably have been more appropriate since we are filtering?

js/src/api/contract/contract.js
```diff
@@ -29,7 +29,9 @@ export default class Contract {
     }
 
     this._api = api;
-    this._abi = new Abi(abi);
+    this._abi = abi instanceof Abi
```
jacogr: I prefer the original approach where we are explicit on what we expect in as opposed to having either one or the other? No other way around it?

js/src/util/abi.js
```diff
@@ -147,3 +149,31 @@ export function parseAbiType (type) {
...
+  try {
+    return new Abi(_abi);
+  } catch (error) {
+    // If there is a wrong type, filter it out
+    if (/Cannot convert .+ valid ParamType/i.test(error.message)) {
+      const validInterfaces = [];
+
+      _abi.forEach((interf) => {
+        try {
+          Abi.parseABI([ interf ]);
```
jacogr: When doe this become a problem and why is this ont on the actual ABI layer instead of just benefitting the UI?

jacogr:
Tests failing?

debris:
@paritytech/ui-devs any updates on this?

PR #5600, opened 2017-05-11
[beta] Fix ethsign

tomusdrw:
Not labelling as `A8-backport` since the code is completely different.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -557,7 +557,7 @@ test-rust-stable:
     - export RUST_FILES_MODIFIED=$(git --no-pager diff --name-only $CI_BUILD_REF^ $CI_BUILD_REF | grep -v -e ^js -e ^\\. -e ^LICENSE -e ^README.md -e ^appveyor.yml -e ^test.sh -e ^windows/ -e ^scripts/ -e^mac/ -e ^nsis/ | wc -l)
   script:
     - export RUST_BACKTRACE=1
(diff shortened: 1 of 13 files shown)


rphmeier:
@paritytech/ci -- where's gitlab?

gavofyork:
this is a little problematic. it blindly copies geth, yet:
- the documentation on the ethereum and parity RPC wiki contains an example which states the original ordering;
- oo7-parity.js assumes the original ordering;
- at least one dapp the same.

tomusdrw:
- Example here: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign contains a RSV signature.
- Parity wiki is a bit outdated (was auto-updated, but some manual changes were lost it seems), doesn't even mention that the data is altered before signing: https://github.com/paritytech/parity/wiki/JSONRPC-eth-module#eth_sign
- It brings back the ordering we had since the very 1.0 version and is used by other providers as well (Metamask, TestRPC). I don't see a reason to keep it as VRS but see a point in having a cross-client compatible API.

gavofyork:
ok the EF changed the wiki spec (unilaterally - how surprising) to fit the geth impl.

https://github.com/ethereum/wiki/wiki/JSON-RPC/_compare/c4321dedf743c6e0d9dd5d3679b028cae46eef7c...3dea1229fb5a23188798075d6127406a1ab70690

may as well play along.

tomusdrw:
Seems that build failure is not related to my changes. @paritytech/ci could you guys have a look? `kcov` is not installed?

PR #5610, opened 2017-05-12
React router dom

CraigglesO:
Before I move forward, I have to make adjustments for this to truly work:

1) Switch how the dapps are loaded... (move them to Application)
This is the new route schema. So if you go to, say "localhost:3000/accounts", it's not going to load.
```
ReactDOM.render(
  <ContextProvider api={ api } muiTheme={ muiTheme } store={ store }>
    <Router>
      <Application>
        <Route path='/' exact component={ Dapps } />
        <Route path='/:id' component={ Dapp } onEnter={ onEnterDapp } />
        <Route path='/:id/:details' component={ Dapp } onEnter={ onEnterDapp } />
      </Application>
    </Router>
  </ContextProvider>,
  document.querySelector('#container')
);
```

2. No visual cue initially upon click. Ideally, the speed of change from one to another will improve, so I am not sure if I should worry about that yet?

3. Some of the views used react-router stuff like `indexRoute` and `redirect`. These are no longer applicable in react-router 4.0.
I'll just update these myself..

Any suggestions about what you do and do not want changed moving forward let me know.


js/src/shell/index.js
```diff
@@ -85,13 +85,12 @@ function onEnterDapp ({ params }) {
 
 ReactDOM.render(
   <ContextProvider api={ api } muiTheme={ muiTheme } store={ store }>
-    <Router history={ hashHistory }>
-      <Route path='/' component={ Application }>
-        <Redirect from='/auth' to='/' />
+    <Router>
+      <Application>
+        <Route path='/' exact component={ Dapps } />
```
jacogr: We can't lose the /auth route, used by the shell to generate and specify new tokens.

CraigglesO: Yep. Havn't solved that one yet -_- There is another issue with clicking an account [accounts -> account] It pushes a '#/' prior to the account. 

I really think it would be better to have accounts and account all in one dapp if that is possible.

debris:
@paritytech/ui-devs any updates on this?

CraigglesO:
YES! When the new Webpack config is pulled in, this will be finished as well. Again, too many changes happening right now, so I'll keep it closed. Would like to keep it though.

debris:
This pr hasn't been updated for > 28 days and I'm closing it as a part of pull request cleanup. If it's still relevant, please rebase and reopen it.

PR #4414, opened 2017-02-03
ease using secrets in tests

debris:
related #4402

PR #5612, opened 2017-05-12
Create an account for chain=dev

### LINKED ISSUE #5335: Make it easier to use Parity for continuous integration testing.
At the moment, Parity is a bit awkward to use for integration testing because it is designed to be secure (generally a good thing).  This means things like `unlockAccount` RPC call only last for a single request and creating a new account (`parity_newAccountFromPhrase`) needs to be run against a live node.

The ideal outcome of this issue would be for Parity to have a well established pattern for using it as part of a headless dApp integration testing suite.  This means a clear, supported, and well documented workflow for using Parity in a non-secure but easy to test against way.

One potential solution would be to make it so `--chain dev` came with several accounts baked in and unlocked, with one of them as the coinbase and all of them having various amounts of ETH available (at least one should have no ETH and at least one should have a ton of ETH).  By default, the one with a lot of ETH should be the current coinbase.  This would allow integration test authors to spin up a node with a new chain, upload contracts, test contract/dApp behavior, and then tear down the chain.

Whatever the solution, for security purposes it may make sense to make the solution *only* work for dev chains or otherwise be disabled for main network chain (not sure how feasible this is).

tomusdrw:
Closes #5335 
Also contains #4414 - I've noticed that `Secret` semantics was changed anyway (it doesn't necessarily store a valid secret any more), so I've implemented convenient conversions and replaced occurrences in code.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -1548,7 +1548,7 @@ mod tests {
 	}
 
 	fn secret() -> Secret {
-		Secret::from_slice(&"".sha3()).unwrap()
+		"".sha3().into()
 	}
 
 	#[test]
--- a/ethcore/src/engines/authority_round.rs
+++ b/ethcore/src/engines/authority_round.rs
@@ -453,7 +453,6 @@ mod tests {
 	use util::*;
 	use header::Header;
 	use error::{Error, BlockError};
-	use ethkey::Secret;
 	use rlp::encode;
 	use block::*;
 	use tests::helpers::*;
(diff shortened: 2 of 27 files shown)


rphmeier:
> it doesn't necessarily store a valid secret any more
I think this is really bad. There isn't a point in having a `Secret` type which can store invalid secrets.

tomusdrw:
Yeah, I agree. Maybe I'm mistaken, but afaiu the code, secret-arithmetic may result in some invalid secrets, we can log it as a separate issue then.

debris:
afaik, it can only result in invalid secrets if secp context is not initialized properly

tomusdrw:
`0` is not a valid secret and you can get `0` by subtracting a secret from itself.

gavofyork:
would be nice to have an option in the chain spec file that allows you to provide a secret rather than an account as a key to the genesis accounts section; in this case it would auto-create the account for you.

but this is a good first effort in this direction.

PR #5624, opened 2017-05-15
run cargo fmt, (not meant to be merged)

guanqun:
This is only for preview purpose.


build.rs
```diff
@@ -21,7 +21,8 @@ const MIN_RUSTC_VERSION: &'static str = "1.15.1";
 fn main() {
 	let is = rustc_version::version().unwrap();
 	let required = MIN_RUSTC_VERSION.parse().unwrap();
-	assert!(is >= required, format!("
+	assert!(is >= required,
```
NikVolf: already discussed, we will not align code this way, but

```rust
assert!(is >= required,
    other_para,
    final_param));
```

parity/blockchain.rs
```diff
@@ -188,33 +188,13 @@ fn execute_import(cmd: ImportBlockchain) -> Result<(), String> {
...
 	// build client
-	let service = ClientService::start(
-		client_config,
-		&spec,
-		&client_path,
-		&snapshot_path,
-		&cmd.dirs.ipc_path(),
-		Arc::new(Miner::with_spec(&spec)),
-	).map_err(|e| format!("Client service error: {:?}", e))?;
+	let service = ClientService::start(client_config, &spec, &client_path, &snapshot_path, &cmd.dirs.ipc_path(), Arc::new(Miner::with_spec(&spec)))
```
NikVolf: no way :)

build.rs
NikVolf: and almost everything below is about this

guanqun: probably, the `rustfmt` is confused by the multiple lines in `format!`. Let me check if `rustfmt` has an option to format like this.

parity/blockchain.rs
guanqun: not sure which option controls the behavior...

PR #5634, opened 2017-05-16
Light friendly dapps

rphmeier:
Takes the first steps towards a futures-based dapps server and activates it for the light client.


dapps/src/tests/fetch.rs
```diff
@@ -60,7 +60,7 @@ fn should_return_503_when_syncing_but_should_make_the_calls() {
 
 	// then
 	response.assert_status("HTTP/1.1 503 Service Unavailable");
-	assert_eq!(registrar.calls.lock().len(), 4);
+	assert_eq!(registrar.calls.lock().len(), 2);
```
tomusdrw: Can you update the test name as well?

updater/src/updater.rs
```diff
@@ -338,9 +340,13 @@ impl fetch::urlhint::ContractClient for Updater {
 			.ok_or_else(|| "Registrar not available".into())
 	}
 
-	fn call(&self, address: Address, data: Bytes) -> Result<Bytes, String> {
-		self.client.upgrade().ok_or_else(|| "Client not available".to_owned())?
-			.call_contract(BlockId::Latest, address, data)
+	fn call(&self, address: Address, data: Bytes) -> BoxFuture<Bytes, String> {
+		let res = (move || {
+			self.client.upgrade().ok_or_else(|| "Client not available".to_owned())?
```
tomusdrw: I'd go with `map` instead of a closure.

rphmeier: I used the closure because of the `?` operator usage here, it seemed easier than doing an `and_then`

dapps/src/tests/fetch.rs
rphmeier: is it wrong now? "the calls" is pretty ambiguous: it's definitely making some calls, although not as many as before.

tomusdrw: Right, makes sense.

updater/src/updater.rs
tomusdrw: IMHO would be cleaner with `and_then`:
```rust
futures::done(
    self.client.upgrade()
        .ok_or_else(|| "Client not available".to_owned())
        .and_then(|client| client.call_contract(BlockId::Latest, address, data))
).boxed()
```

rphmeier: done

PR #5647, opened 2017-05-17
use copy_from_slice instead of for loop

guanqun:
According to the doc here: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice

It will use memcpy under the hood, so I think it has slightly better performance.


rphmeier:
in general, LLVM will lift simple loops like this to memcpy/memmove, but this also gives a readability bonus. thanks!

Issue #884, opened 2016-04-04
Pub/Sub RPC feature


tymat:
See: https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB

gavofyork:
despite the go guys insistence, full duplex operations is not within the JSONRPC v2 standard. that said, we might support it if it's not too much effort.

rphmeier:
There seems to be quite a bit of demand for this feature. :+1: to get it in for 1.4

jacogr:
We do support a subset of this (obviously implemented via polling, not via events from Parity) in our own JS API. So we need to align the 2 sides when doing this in parallel - in addition, need to carefully look at the actual use-case and not (imho) just do a straight as-is implementation.

gituser:
would be nice to have a transaction notification feature on all wallets stored in the client like in bitcoin.
cheers

gituser:
Guess this has been postponed. Would be nice to have feature analog of [bitcoin's walletnotify](http://bitcoin.stackexchange.com/questions/24457/how-do-i-use-walletnotify) where you can pass incoming txid to any of the wallet addresses to the external script or notify backend instead of querying parity every minute.

5chdn:
This issue is labelled with _sometimesoon_: Issue is worth doing soon. But it is inactive for several months.

It was initially scheduled for 1.4, now it's attached to the 1.7 milestone. Is this a realistic deadline?

tomusdrw:
Yup, initial part ready in #5456 we still may want to implement geth-compatible subscriptions (I'll work on that soon)

5chdn:
Is eth_pubsub complete now?

tomusdrw:
Still missing less important methods, I'll implement them next week to finally close this:
- [ ] syncing
- [ ] Pending (local?) transactions

debris:
@tomusdrw what is the status of it?

tomusdrw:
Still missing those two methods, but on my list.

tomusdrw:
It's actually tracked in #6426, let's close this one.

PR #5655, opened 2017-05-18
Latest headers Pub-Sub

tomusdrw:
Related #884 

The first subscription from geth-compatible interface.


rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -0,0 +1,153 @@
...
+		_proposed: Vec<Bytes>,
+		_duration: u64,
+	) {
+		const EXTRA_INFO_PROOF: &'static str = "Object exists in in blockchain (fetched earlier), extra_info is always available if object exists; qed";
+		let blocks = enacted
+			.into_iter()
+			.filter_map(|hash| self.client.block_header(BlockId::Hash(hash)))
+			.map(|header| {
+				let hash = header.hash();
+				(header, self.client.block_extra_info(BlockId::Hash(hash)).expect(EXTRA_INFO_PROOF))
```
rphmeier: headers are pruned out of the light client very quickly, so I wouldn't expect this to always pass.

tomusdrw: That's an implementation for full client, light implementation doesn't have `extra_info` at all.

rphmeier:
What about handling `retracted` blocks for reorg? Does the geth RPC not do that?

tomusdrw:
In case of re-org `enacted` should contain all the blocks that become canonical, the client will receive headers with numbers that were received in the past. It should be the same behaviour as for geth.

Issue #5435, opened 2017-04-10
docker versions do not match tags


5chdn:
related to #5354, docker hub images versions do not match:

    docker pull ethcore/parity:stable  # 1.6.3 - should be 1.5.12
    docker pull ethcore/parity:v1.5.12 # 1.7.0 - should be 1.5.12
    docker pull ethcore/parity:beta    # 1.7.0 - should be 1.6.5-beta
    docker pull ethcore/parity:v1.6.5  # 1.6.5 - correct \o/
    docker pull ethcore/parity:latest  # 1.6.3 - should be 1.7.0-unstable

5chdn:
Also, what's the difference between stable and stable-release?

    $ curl -sS 'https://registry.hub.docker.com/v2/repositories/ethcore/parity/tags/'  | jq '."results"[]["name"]' |sort
    "beta"
    "latest"
    "master"
    "nightly"
    "stable"
    "stable-release"
    "v1.5.12"
    "v1.6.4"
    "v1.6.5"
    "v1.6.6"

5chdn:
This issue is labelled with _annoyance_: The CI behaves within expectations, however this "expected behaviour" itself is at issue. It's neither assigned nor linked to any milestone.

@paritytech/ci Please decide on a deadline and add an assignee within 7 days, thanks!

5chdn:
Actually, this is more than just _annoyance_.

General-Beck:
@5chdn we move to https://hub.docker.com/r/parity/parity/ with reorganisation

debris:
imo, the problem is caused by [this line](https://github.com/paritytech/parity/blob/2f757babb9d29607682fcb5a8c1f6af18033281c/docker/ubuntu/Dockerfile#L32)

- Dockerfile:

```dockerfile
RUN git clone https://github.com/paritytech/parity
```

Cloning repo always takes the latests version of given repo, whereas we should use [local copy](https://github.com/paritytech/parity-bitcoin/blob/228c97ad85a4dd1de530804a18d293f315a97df3/docker/ubuntu/Dockerfile#L3) of the repo 

- Dockerfile

```dockerfile
ADD . /build/parity
```

- build command:

```
docker build -t parity-ubuntu -f docker/ubuntu/Dockerfile .
```

General-Beck:
[...]

PR #5659, opened 2017-05-19
improved dockerfile builds

debris:
As noticed in #5435, our docker versions do not match tags. As described by @General-Beck this issue is currently solved by using custom build argument - `BUILD_TAG`. This pr simlifies and improves build process further.

Current build commands:

```bash
# branch that we want to build 
branch_or_tag_name="1.7"

# first clone of the repo
git clone https://github.com/paritytech/parity
cd parity
git checkout hub

# run docker build
# second clone of the repo occurs inside the container
docker build --build-arg BUILD_TAG=branch_or_tag_name --no-cache=true --tag ethcore/parity:branch_or_tag_name .
```

New build commands:

```bash
# branch that we want to build 
branch_or_tag_name="1.7"

# run docker build
# the only clone of the repo
git clone https://github.com/paritytech/parity
git checkout branch_or_tag_name

docker build -f docker/ubuntu/Dockerfile .
```

This pr also:
- makes every parity branch/tag buildable with stable/beta/nightly on centos/arm
- removes unneeded `ubuntu-dev` and `ubuntu-stable` dockerfiles

PS
*Needs to be backported, if merged


docker/centos/Dockerfile
```diff
@@ -1,29 +1,32 @@
 FROM centos:latest
 WORKDIR /build
+ADD . /build/parity
```
tomusdrw: Add command should be moved right before `RUN cd parity` to utilized Docker cache for all deps and rust compilation.

General-Beck:
We will also need to add readme and make corrections to the [wiki](https://github.com/paritytech/parity/wiki/Docker)

5chdn:
> We will also need to add readme and make corrections to the wiki

@General-Beck Can do it, what exactly changed for the user?

gavofyork:
make sure this doesn't break the auto-updating integration please!

docker/README.md
```diff
@@ -0,0 +1,3 @@
+Usage
+
+```docker build --no-cache=true -f docker/ubuntu/Dockerfile .```
```
asymmetric: I would tag the image too by adding `-t parity`. This makes the output of `docker images` a bit more friendly:

```
❯ docker images
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
parity                         latest              9f8e2b065950        7 minutes ago       1.74 GB
```

asymmetric: Also, why the `--no-cache` option?

debris: I also don't know. `--no-cache` was here before (the file was moved from `docker/hub/README.md` -> `docker/README.md` and I simply kept it unchanged

General-Beck: `--no-cache`can be removed

PR #5671, opened 2017-05-20
Guanqun add unit tests

gavofyork:


gavofyork:
not building properly. @General-Beck ?

guanqun:
I used `cargo test -p ethjson` and it worked fine locally. What's the issue? I see "All checks have passed" from CI.

rphmeier:
@guanqun https://gitlab.parity.io/parity/parity/builds/54720 indicates that the build was skipped because it modified no Rust files, which definitely isn't true.

guanqun:
aha, I think I know the problem, here's the fix for it. https://github.com/paritytech/parity/pull/5668/commits/14b715bdc7e746801ed4e83af8654445d2d7366b

The original script only detects for "^js", interesting enough, I'm modifying "json"...

But I don't have permission to push to `guanqun-add-unit-tests`, probably someone can cherry-pick this commit.

arkpar:
This is really great, thanks a lot!

PR #5676, opened 2017-05-22
Parity-PubSub as a separate API.

tomusdrw:
Also enables generic pubsub for light client as well (I missed it in earlier).


arkpar:
needs a rebase

PR #5677, opened 2017-05-22
Adding CLI options: port shift and unsafe expose.

tomusdrw:
1. Allows to run two instances easily.
2. Allows to expose a node easily (for instance when running in Docker container or just playing around).
3. `--*-interface all` now implies `--*-hosts all` unless configured differently.

Most probably will need some merging with #5580

### DIFF
--- a/parity/cli/usage.txt
+++ b/parity/cli/usage.txt
@@ -83,7 +83,7 @@ Convenience Options:
                                    Includes network port and all servers (RPC, WebSockets, UI, IPFS, SecretStore).
                                    (default: {flag_ports_shift})
   --unsafe-expose                  All servers will listen on external interfaces and will
-                                   be remotely accessible. It's equivallent with setting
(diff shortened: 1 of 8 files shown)


parity/cli/usage.txt
```diff
@@ -76,6 +76,18 @@ Operating Options:
...
                                    subcommands (default: {flag_light}).
 
+Convenience Options:
+  -c --config CONFIG               Specify a filename containing a configuration file.
+                                   (default: {flag_config})
+  --ports-shift SHIFT              Add SHIFT to all port numbers Parity is listening on.
+                                   Includes network port and all servers (RPC, WebSockets, UI, IPFS, SecretStore).
+                                   (default: {flag_ports_shift})
+  --unsafe-expose                  All servers will listen on external interfaces and will
+                                   be remotely accessible. It's equivallent with setting
```
rphmeier: typo: equivalent

PR #5678, opened 2017-05-22
EIP-211 RETURNDATACOPY and RETURNDATASIZE

arkpar:
https://github.com/ethereum/EIPs/pull/211

### DIFF
--- a/ethcore/src/evm/evm.rs
+++ b/ethcore/src/evm/evm.rs
@@ -97,10 +97,45 @@ impl fmt::Display for Error {
 /// A specialized version of Result over EVM errors.
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+
+/// Return data buffer. Holds memory from a previous call and a slice into that memory.
+#[derive(Debug)]
+pub struct ReturnData {
+	mem: Vec<u8>,
+	offset: usize,
+	size: usize,
+}
+
+impl ::std::ops::Deref for ReturnData {
+	type Target = [u8];
+	fn deref(&self) -> &[u8] {
+		&self.mem[self.offset..self.offset + self.size]
+	}
+}
+
+impl ReturnData {
+	/// Create empty `ReturnData`.
+	pub fn empty() -> Self {
+		ReturnData {
+			mem: Vec::new(),
+			offset: 0,
+			size: 0,
+		}
+	}
+	/// Create `ReturnData` from give buffer and slice.
+	pub fn new(mem: Vec<u8>, offset: usize, size: usize) -> Self {
+		ReturnData {
+			mem: mem,
+			offset: offset,
+			size: size,
+		}
+	}
+}
+
 /// Gas Left: either it is a known value, or it needs to be computed by processing
 /// a return instruction.
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum GasLeft<'a> {
+#[derive(Debug)]
+pub enum GasLeft {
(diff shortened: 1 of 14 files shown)


ethcore/src/evm/evm.rs
```diff
@@ -97,18 +97,53 @@ impl fmt::Display for Error {
...
+	pub fn empty() -> Self {
+		ReturnData {
+			mem: Vec::new(),
+			offset: 0,
+			size: 0,
+		}
+	}
+	/// Create `ReturnData` from give buffer and slice.
+	pub fn new(mem: Vec<u8>, offset: usize, size: usize) -> Self {
+		ReturnData {
```
rphmeier: should probably truncate the allocation here to just `size` to alleviate peak memory usage concerns. I'm not aware of a fast way to do it safely, though.

`let _ = mem.drain(..offset); mem.truncate(size); mem.shrink_to_fit()` might be a lot of work to do on every return.

rphmeier:
where is the implementation of `RETURNDATASIZE`? I don't see it when searching the diff

ethcore/src/evm/evm.rs
arkpar: Updated. Now it reallocates but only if too much memory is wasted.

arkpar:
@rphmeier It was missing :) Added now

ethcore/src/evm/interpreter/memory.rs
```diff
@@ -109,6 +114,20 @@ impl Memory for Vec<u8> {
...
 	}
+
+	fn into_return_data(mut self, offset: U256, size: U256) -> ReturnData {
+		let offset = offset.low_u64() as usize;
+		let size = size.low_u64() as usize;
+		if !is_valid_range(offset, size) {
+			return ReturnData::empty()
+		}
+		if self.len() - size > MAX_RETURN_WASTE_BYTES {
+			let mem = self.drain(offset..).take(size).collect();
```
rphmeier: i really doubt the optimizer is smart enough to turn this into a single malloc + memcpy, so it ends up paying both the O(n) iteration and having to allocate `self.len() + size` bytes in total.

i would suggest either using a `drain`, `truncate`, `shrink_to_fit` combo or allocating a `vec![0; size]` and using `copy_from_slice` to pay only one of those costs.

rphmeier: (unless the optimized IR/ASM shows that this is actually a non-issue)

ethcore/src/evm/interpreter/memory.rs
```diff
@@ -109,6 +114,20 @@ impl Memory for Vec<u8> {
 			Memory::resize(self, size)
 		}
 	}
+
+	fn into_return_data(mut self, offset: U256, size: U256) -> ReturnData {
+		let offset = offset.low_u64() as usize;
+		let size = size.low_u64() as usize;
+		if !is_valid_range(offset, size) {
```
rphmeier: this just checks whether `offset + size` would overflow, but they should also be bounds-checked against `self.len()` at this point. that would also make the check within `ReturnData::new` redundant, so maybe it would be better as a precondition of that function.

arkpar: Memory is guaranteed to be allocated at this point. This is assured by mem/gas requirements checking before each instruction. That's why none of the methods in this struct check for bounds.

NikVolf:
lgtm as well

PR #5679, opened 2017-05-22
WASM contracts MVP

NikVolf:
- If wasm contracts are enabled by the engine, evm executor will choose wasm interpreter to run the code if also the contract code starts with wasm program header.

- Wasm contracts can be called, but cannot call other contracts in the MVP

- Wasm contracts can suicide

- Wasm contracts can read and update storage, allocate memory and use most of the rust std lib (the only implication of it is the resulting size of the contract)

- Build process for the contracts and use of the state api both can be found in `wasm-tests` repo:
https://github.com/paritytech/wasm-tests (the parity also runs tests against them, compiled). Rust wasm support is not very mature still, requires nightly and some linker args magic to produce result we need (contributions and suggestions to improve the above are welcome)

- `parity-wasm` repo is the general purpose crate to serialize/deserialize wasm and also hosts (probably temporary) the interpreter

- `wasm-utils` repo contains utilities to manipulate wasm for parity-specific targets (optimize symbols, substitute allocators, inject gas counter)


.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = ethcore/res/ethereum/tests
 	url = https://github.com/ethereum/tests.git
 	branch = develop
+[submodule "ethcore/res/wasm-tests"]
+	path = ethcore/res/wasm-tests
+	url = https://github.com/NikVolf/wasm-tests
```
rphmeier: a paritytech repo might be better

rphmeier:
Could we use paritytech repos instead of `nikvolf` for wasm-utils and wasm-tests?

ethcore/src/evm/wasm/call_args.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub sender: [u8; 20],
+
+    /// Original transaction initiator
+    pub origin: [u8; 20],
+
+    /// Transfer value
+    pub value: [u8; 32],
+
+    /// Reserved space / alignment to 256 bytes
+    _reserved: [u8; 164],
```
arkpar: Is this really required? `data` makes this struct over 256 bytes anyway, so it is not really aligned.

ethcore/src/evm/wasm/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+		let env_instance = self.program.module("env")
+			.ok_or(evm::Error::Wasm("Env module somehow does not exist in wasm runner runtime"))?;
+
+		// todo: prefer panic?
+		let env_memory = env_instance.memory(interpreter::ItemIndex::Internal(0))
+			.map_err(|_| evm::Error::Wasm("Linear memory somehow does not exist in wasm runner runtime"))?;
+
+		if params.gas > ::std::u64::MAX.into() {
+			// todo: prefer panic?
+			return Err(evm::Error::Wasm("Wasm interpreter cannot run contracts with gas > 2^64"));
```
arkpar: `>=` to be precise

ethcore/src/engines/authority_round.rs
```diff
@@ -58,6 +58,8 @@ pub struct AuthorityRoundParams {
 	pub eip155_transition: u64,
 	/// Monotonic step validation transition block.
 	pub validate_step_transition: u64,
+	/// If wasm contracts are supported.
+	pub wasm: bool,
```
rphmeier: this could be in `CommonParams`

ethcore/src/engines/mod.rs
```diff
@@ -309,6 +309,11 @@ pub trait Engine : Sync + Send {
 		self.snapshot_components().is_some()
 	}
 
+	/// If this engine supports wasm contracts.
+	fn supports_wasm(&self) -> bool {
+		false
```
rphmeier: then this would just be `self.params().supports_wasm`

ethcore/src/evm/evm.rs
```diff
@@ -79,6 +82,12 @@ impl From<builtin::Error> for Error {
 	}	
 }
 
+impl From<wasm::RuntimeError> for Error {
+	fn from(_err: wasm::RuntimeError) -> Self {
+		Error::Wasm("Runtime error")
```
rphmeier: This isn't a really useful message. Why not format the wasm error into a `String`?

ethcore/src/evm/evm.rs
```diff
@@ -89,6 +98,7 @@ impl fmt::Display for Error {
 			StackUnderflow { .. } => "Stack underflow",
 			OutOfStack { .. } => "Out of stack",
 			BuiltIn { .. } => "Built-in failed",
+			Wasm { .. } => "Wasm error",
```
rphmeier: this should use the inner string

ethcore/src/evm/wasm/call_args.rs
rphmeier: Alignment can also be done with a `PhantomData`, but nothing will naturally require align greater than system word length until `#[repr(align(_))]` attribute stabilizes.

ethcore/src/evm/wasm/result.rs
```diff
@@ -0,0 +1,47 @@
...
+//! Wasm evm results helper
+
+use byteorder::{LittleEndian, ByteOrder};
+
+use parity_wasm::interpreter;
+
+use super::ptr::WasmPtr;
+use super::runtime::Error as RuntimeError;
+
+pub struct WasmResult {
```
arkpar: Struct and functions could use some documentation

ethcore/src/evm/wasm/runtime.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub struct Runtime<'a> {
+	gas_counter: u64,
+	gas_limit: u64,
+	dynamic_top: u32,
+	ext: &'a mut evm::Ext,
+	memory: Arc<interpreter::MemoryInstance>,
+}
+
+impl<'a> Runtime<'a> {
```
arkpar: Missing documentation for pub functions

rphmeier:
How are non `UserFunction` operations gas-priced?

ethcore/src/evm/evm.rs
NikVolf: to avoid heap allocations which can be triggered by the very cheap contract?

ethcore/src/evm/wasm/runtime.rs
```diff
@@ -0,0 +1,274 @@
...
+
+	pub fn storage_write(&mut self, context: interpreter::CallerContext) 
+		-> Result<Option<interpreter::RuntimeValue>, interpreter::Error>
+	{
+		let mut context = context;
+		let val = self.pop_h256(&mut context)?;
+		let key = self.pop_h256(&mut context)?;
+		trace!(target: "wasm", "storage_write: value {} at @{}", &val, &key);
+
+		// todo: return a runtime error contract can handle or as it is now - general failure?
```
arkpar: I suppose external errors should not be propagated to contracts since they can't handle them in any meaningful way. External error == consensus error here.

ethcore/src/evm/wasm/call_args.rs
NikVolf: sure, on second thought, it is not required
I thought it might be useful for contracts to be compatible when the header changes (extends or shrinks), but this can be achieved by other more consensus-friendly means (versioned header)

ethcore/src/evm/wasm/mod.rs
NikVolf: oh but why u64::MAX is not good enough to be a max here?

ethcore/src/evm/wasm/runtime.rs
NikVolf: updated

ethcore/src/evm/wasm/mod.rs
rphmeier: `u64::MAX` is `2^64 - 1`, so `>=` just corrects the statement.

ethcore/src/evm/evm.rs
rphmeier: I see, but there are also a lot of wasm errors with allocated strings that will be discarded here. Perhaps an inner `enum WasmError { Runtime, Other(String) }` so that extra info isn't lost?

NikVolf:
@rphmeier 
in the MVP, any single wasm opcode just costs 1 gas
runtime is free to multiply it by any coefficient though:

https://github.com/paritytech/parity/pull/5679/files#diff-8bf882394d19ca210e0b45ddfc7b775aR151

ethcore/src/evm/evm.rs
NikVolf: I think i will do as you originally suggested for now, more proper way would require interpreter to be generic over error type that user functions can return

rphmeier: yeah, that makes sense.

ethcore/src/evm/wasm/mod.rs
```diff
@@ -151,49 +151,106 @@ impl evm::Evm for WasmInterpreter {
...
+	use parity_wasm::elements::ValueType::*;
+
 	interpreter::UserFunctions {
 		executor: runtime,
 		functions: vec![
 			interpreter::UserFunction {
 				name: "_storage_read".to_owned(),
-				params: vec![elements::ValueType::I32, elements::ValueType::I32],
-				result: Some(elements::ValueType::I32),
+				params: vec![I32, I32],
```
rphmeier: since these all seem to be statically known, why not have it take a `&'static [ValueType]` instead of requiring allocation on each one?

NikVolf: Yeah, this optimisation can be made right away

ethcore/src/evm/evm.rs
```diff
@@ -66,6 +67,8 @@ pub enum Error {
 	MutableCallInStaticContext,
 	/// Likely to cause consensus issues.
 	Internal(String),
+	/// Wasm runtime error
+	Wasm(String),
```
Vurich: Is there something better we can give here than `String`? Maybe a `WasmError` enum with some number of "proper" error variants and one `Unknown(String)` variant.

ethcore/src/evm/wasm/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+					0, 
+					len,
+				),
+				apply_state: true,
+			})
+		}
+	}
+}
+
+fn native_bindings<'a>(runtime: &'a mut Runtime) -> interpreter::UserFunctions<'a> {
```
Vurich: Unnecessary lifetime parameters

ethcore/src/evm/wasm/runtime.rs
```diff
@@ -0,0 +1,356 @@
...
+		if prev + amount > self.gas_limit {
+			// exceeds gas
+			false
+		} else {
+			self.gas_counter = prev + amount;
+			true
+		}		
+	}
+
+	fn h256_at(&self, ptr: WasmPtr) -> Result<H256, interpreter::Error> {
```
Vurich: Can this lead to type confusion bugs? Can we have a `WasmPtr` carry a `PhantomData` to ensure that we deserialize to the correct type? If not, can we mark these functions as `unsafe`?

ethcore/src/evm/wasm/runtime.rs
```diff
@@ -0,0 +1,356 @@
...
+		// write call descriptor
+		// call descriptor is [args_ptr, args_len, return_ptr, return_len]
+		//   all are 4 byte length, last 2 are zeroed
+		let mut d_buf = [0u8; 16];
+		LittleEndian::write_u32(&mut d_buf[0..4], args_ptr);
+		LittleEndian::write_u32(&mut d_buf[4..8], args_len);
+		self.memory.set(d_ptr, &d_buf)?;
+
+		// write call args to memory
+		self.memory.set(args_ptr, &call_args.address)?;
```
Vurich: Hard-coded byte offsets here - can we use `mem::size_of_value` or abstract this some other way?

ethcore/src/evm/wasm/runtime.rs
```diff
@@ -0,0 +1,356 @@
...
+			"_suicide" => {
+				self.suicide(context)
+			},
+			"_create" => {
+				self.create(context)
+			},
+			"_debug" => {
+				self.debug_log(context)
+			},
+			"gas" => {
```
Vurich: Should this have an underscore like the rest? (I don't know, but it just sticks out so I thought I'd mention it).

ethcore/src/executive.rs
```diff
@@ -34,6 +34,8 @@ pub use types::executed::{Executed, ExecutionResult};
 /// Maybe something like here: `https://github.com/ethereum/libethereum/blob/4db169b8504f2b87f7d5a481819cfb959fc65f6c/libethereum/ExtVM.cpp`
 const STACK_SIZE_PER_DEPTH: usize = 24*1024;
 
+const WASM_MAGIC_NUMBER: [u8; 4] = [0x00, 0x61, 0x73, 0x6d];
```
Vurich: This is just personal taste but if a byte array is supposed to be ASCII I'd rather write this as `*b"\0asm"` instead. It doesn't matter, but it might be something to keep in mind going forward.

ethcore/src/evm/wasm/runtime.rs
NikVolf: I think we'd better keep it this way for now
to distinguish injected functions from runtime
"gas" function is unique in a way it cannot be called ever directly

ethcore/src/evm/wasm/runtime.rs
NikVolf: No, it's crossplatform offset, we cannot rely on Rust memory layout, despite it may happen to have same value.

ethcore/src/evm/wasm/runtime.rs
NikVolf: It's perfectly safe, because it works on bound-checked memory api

ethcore/src/evm/evm.rs
NikVolf: I would avoid to wrap `parity_wasm::interpreter::Error`, because it cannot be later meaningfully encoded as trace error and put in trace db, that's why it's encoded as string

Also I cannot imagine any code that would rely on handling specific wasm interpreter errors

Vurich: As long as it's strictly for human-readable messages then that's fine. If we ever finding ourself matching on it we can refactor to make it an error.

ethcore/src/evm/wasm/runtime.rs
Vurich: Bound-checked, sure, but that doesn't mean that it won't cause type confusion. We're for all intents and purposes doing unchecked `mem::transmute` to synthesise these values. It's not going to access uninitialized memory but it may cause us to interpret data as the wrong type.

ethcore/src/evm/wasm/runtime.rs
Vurich: Ok, can we at least extract it to be more self-documenting, or use `debug_assert_eq!(mem::size_of_value(something), 20)` to make sure that we're not truncating data? Also, we can use non-target-dependent datatypes in a `#[repr(C)]` struct, which also has the benefit of allowing us to blit it to memory in a single call (even if it requires doing `mem::transmute::<Foo, [u8; 100]>(my_struct)` or whatever).

ethcore/src/evm/wasm/runtime.rs
Vurich: What happens if you compile a C file with a function called `gas` to wasm and run it as a contract? If we want it to be impossible (for some value of "impossible") to accidentally link a file with a conflicting name we can use non-ASCII or unprintable characters, like `\0` or 💩.

Obviously if defining `gas` and compiling to wasm does nothing then you can just give it whatever name you like, I don't actually know the semantics here I'm just pointing out something that might cause a problem.

NikVolf: You cannot have named functions in webassembly, only imports or exports are named
And if C program will import `gas` function explicitly, it will be invalid as a contract and will fail instantly, before even running (as wasm module validation fails when there are duplicated imports, there should be rule about it in webassembly official testsuite)

ethcore/src/evm/wasm/mod.rs
NikVolf: not working, btw, since `interpreter::UserFunctions` has an explicit lifetime
see https://gitlab.parity.io/parity/parity/-/jobs/59538

ethcore/src/evm/wasm/runtime.rs
Vurich: Sweet, LGTM then.

ethcore/src/evm/wasm/mod.rs
Vurich: It's not because `UserFunctions` has a lifetime parameter, it's because `Runtime` has a lifetime parameter. `&mut Runtime` has two elided lifetime parameters, so when you had the lifetime parameter to the function it elided one of them and not the other. You do need the parameters in this case, sorry. I didn't know `Runtime` took a lifetime parameter.

Basically, it desugars to:

```rust
fn native_bindings<'a, 'b>(runtime: &'a mut Runtime<'b>) -> interpreter::UserFunctions<'???>
```

Since there are two parameters to choose from, it doesn't know which to pass to `UserFunctions`.

ethcore/src/evm/wasm/runtime.rs
NikVolf: Still worth adding a explicit test for this case, thanks

ethcore/src/executive.rs
```diff
@@ -34,7 +34,7 @@ pub use types::executed::{Executed, ExecutionResult};
 /// Maybe something like here: `https://github.com/ethereum/libethereum/blob/4db169b8504f2b87f7d5a481819cfb959fc65f6c/libethereum/ExtVM.cpp`
 const STACK_SIZE_PER_DEPTH: usize = 24*1024;
 
-const WASM_MAGIC_NUMBER: [u8; 4] = b"\0asm";
+const WASM_MAGIC_NUMBER: &'static [u8; 4] = b"\0asm";
```
Vurich: Alternatively (it doesn't matter in this case, but it might in the future):

```rust
const WASM_MAGIC_NUMBER: [u8; 4] = *b"\0asm";
```

Also, Rust will automatically infer `'static` in `const` and `static` items (so you can write `&[u8; 4]` for the type)

ethcore/src/evm/wasm/runtime.rs
NikVolf: ok, let's chat about this and `WasmPtr` later in more detail
I will update in following prs if the chat will produce some useful outcome

Vurich: Sounds like a great idea. Don't get me wrong, the code looks good, I'm just pointing to some areas that might deserve a second look. There's nothing that I would argue against merging because of.

PR #5686, opened 2017-05-23
Fix from/into electrum in ethkey

NikVolf:


ethkey/src/signature.rs
```diff
@@ -54,7 +54,7 @@ impl Signature {
 	/// Parse bytes as a signature encoded as RSV (V in "Electrum" notation).
 	/// May return empty (invalid) signature if given data has invalid length.
 	pub fn from_electrum(data: &[u8]) -> Self {
-		if data.len() != 65 || data[0] < 27 {
+		if data.len() != 65 || data[64] < 27 {
```
tomusdrw: It was VRS before and we wanted to validate `v`. Currently it's `rsv` again.

PR #5688, opened 2017-05-24
rpc -> weak to arc

CraigglesO:
Without a circular dependency, can just use arc now.


tomusdrw:
Closes #3958

rpc/src/v1/impls/eth.rs
```diff
@@ -124,11 +124,11 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> EthClient<C, SN, S, M, EM> where
 		options: EthClientOptions
 	) -> Self {
 		EthClient {
-			client: Arc::downgrade(client),
-			snapshot: Arc::downgrade(snapshot),
-			sync: Arc::downgrade(sync),
-			miner: Arc::downgrade(miner),
-			accounts: accounts.as_ref().map(Arc::downgrade),
+			client: client.clone(),
```
tomusdrw: Instead of cloning here we could actually change the signature from `&Arc<C>` to `Arc<C>`, it would require more changes though and the result would be the same.

CraigglesO: You read my mind.

Issue #5502, opened 2017-04-25
SecretStore: dynamically generate KeyPairs


svyatonik:
During encryption/decryption session, every KS[i] -> KS[j] message is encrypted with SHARED_KEY=GENERATE(KS[i].KeyPair.Public, KS[j].KeyPair.Public).
KeyPair.Private of KS[i] is currently stored in plain text in configuration file on KS[i].

As @keorn suggested, it would be good not to store this in plain text => suggestion is to generate these values when encryption session is initializing, so we do not need to store it in configuration files.

svyatonik:
The same key is used to confirm that we are contacting with exactly same server we want to contact => by dynamically generating this key, there will be no way to make sure this happens.
There must be another way not to store this in plain text.

keorn:
You can store the current list in a contract. Similar to `ValidatorSet` or enodes that @arkpar is working on.

svyatonik:
Discussed this in private.
1) this issue is about storing private keys in configuration files. This must be resolved as:
1.1) every KeyServer must have corresponding account from ethstore (insert new account if not provided)
1.2) on startup, password for this account must be provided
1.3) use this account as KS KeyPair
2) next issue is about maintaining list of KeyServers. It is what @keorn comment is about. I'll add a separate issue with explanation.

PR #5694, opened 2017-05-25
SecretStore: generating signatures

svyatonik:
### Requirements:
a set of N KeyServers, sharing KeyPair (public is known to every node, secret could be restored by t KeyServers) must be able to sign message (message hash).

### Options:
1) it is possible to generate N-of-N ECDSA-compatible signature
2) it is possible to generate t-of-N ECDSA-compatible signature, but you must run t-of-t algorithm for every possible subset of t nodes
3) it is possible to generate t-of-N EC-Schnorr signature

Of these options, 3rd one has been choosen for now.

### Paper, on which current implementation is based:
https://www.researchgate.net/profile/A_Chronopoulos/publication/224386852_Efficient_multi-party_digital_signature_using_adaptive_secret_sharing_for_low-power_devices_in_wireless_networksPLEASE_REFERENCE_IN_YOUR_PAPERS/links/00b7d52c2a3fb5bf8e000000/Efficient-multi-party-digital-signature-using-adaptive-secret-sharing-for-low-power-devices-in-wireless-networksPLEASE-REFERENCE-IN-YOUR-PAPERS.pdf
also (important): http://www.wu.ece.ufl.edu/mypapers/notesMulti-partyDigitalSignature.pdf

### PR outline:
1) added relevant EC-math functions to secret_store/src/key_server_cluster/math.rs.
1.1) there's same issue as with earlier ECDKG implementation - currently signing only works on t-of-N groups, where t is even. Support for odd t would be added in next PRs, as I have to find the error in paper. This time it must be easier, because there are notes on the paper (see link above), in which this case is described.
1.2) I have slightly changed signature calculation, as it was not compatible with some non-distributed EC-Schnorr implementation, I've found on github. Now it is compatible. And non-distributed code is also in math.rs (local_compute_signature())
2) part of signing session is distributed key generation session (which is used as nonce when computing signature), so I had to separate key generation session from encryption session.
3) as an side effect of (2), secret encryption now can be completed in three steps:
3.1) client asks KeyServer to generate server key. KeyServer returns JointServerPublic to the client
3.2) client generates and encrypts DocumentSecret (which is point on EC in this case) using following procedure:
3.2.1) client generates random EC-field scalar k
3.2.2) client computes CommonPoint = k * ECGenerationPoint
3.2.3) client computes EncryptedDocumentSecret = DocumentSecret + k * JointServerPublic
3.3) client posts pair { CommonPoint, EncryptedDocumentSecret } to KeyServer.
3.4) in combination with 'shadow key retrieval', which was introduced before, this guarantees that KeyServer will never see unencrypted DocumentSecret. Previously DocumentSecret was generated on KeyServer.
4) only 'author' of (3.2.1) request is allowed to submit (3.2.3) request. So I've made some changes to internal SecretStore database to support this. Previous versions of db are incompatible with this new version.
5) both decryption and signng sessions behave in the same manner: at the very beginning they're selecting t nodes out of all nodes, which have agreed to restore the secret/sign message for requester. So I've created consensus_session.rs (which is message exchange for selecting this subset) and consensus.rs (which is subset itself). There's still a lot of almost same code in decryption_session.rs && signing_session.rs - so I'll make additional changes later (in next PRs) to eliminate duplicated code.
6) active sessions container now is a separate entity, living in the cluster_sessions.rs (previously there was a lot of duplicated code for getting/inserting/removing each kind of session)

### API changes:
Previous KeyServer API remains active - you still can simultaneously generate ServerKey && DocumentKey, retrieve DocumentKey && DocumentKey shadow.
In addition, following HTTP requests are now supported:
1) to generate server key (without storing DocumentKey in db), caller must POST to the /shadow/{server_key_id}/{signature}/{threshold} URL:
```
curl -v -X POST http://localhost:8082/shadow/0000000000000000000000000000000000000000000000000000000000000000/de12681e0b8f7a428f12a6694a5f7e1324deef3d627744d95d51b862afc13799251831b3611ae436c452b54cdf5c4e78b361a396ae183e8b4c34519e895e623c00/1
```
result is JointServerPublic:
```
"0x07d6c0e4cecbe7a3ab3a8f5bd6439bb78ecd41f3e7487d34981668c50c62976e6f57eda20c9bd7026b4defa31141520a3480561b3c5bec5275d9d7f4f1e9d50a"
```
2) to store DocumentKey, encrypted by client with JointServerPublic, caller must POST to the /shadow/{server_key_id}/{signature}/{common_point}/{encrypted_key}:
```
curl -v -X POST http://localhost:8082/shadow/0000000000000000000000000000000000000000000000000000000000000000/de12681e0b8f7a428f12a6694a5f7e1324deef3d627744d95d51b862afc13799251831b3611ae436c452b54cdf5c4e78b361a396ae183e8b4c34519e895e623c00/368244efaf441c2dabf7a723355a97b3b86f27bdb2827ae6f34ddece5132efd37af4ba808957b7113b4296bc4ae9ec7be38f9de6bae00504e775883a50d4658a/b7ad0603946987f1a154ae7f074e45da224eaa83704aac16a2d43a675d219654cf087b5d7aacce0790a65abbc1a495b26e71a5c6e9a4a71b543bf0048935bc13
```
3) to sign message, caller must issue GET request to the /{server_key_id}/{signature}/{message_hash}
```
curl -v http://localhost:8082/0000000000000000000000000000000000000000000000000000000000000000/de12681e0b8f7a428f12a6694a5f7e1324deef3d627744d95d51b862afc13799251831b3611ae436c452b54cdf5c4e78b361a396ae183e8b4c34519e895e623c00/0000000000000000000000000000000000000000000000000000000000000001
```
result is signature of message_hash, encrypted with caller' public key:
```
"0x0410fe210ffdd38611b252e24c61b610d6ea8ec70f9f7186184441871839562cc18c94174af0c804973ce88a6f07ec7604b4b91821a6ca68ca348d16cd647fba494f8132ff36ad4654abd5cba6fb1fd98cacc12aaf3a0fe70010be91f5ab7b69db8f80e528be3baba70341721d6058ad66665363033f0cee963476da746e843d4006170ab549fa82f14eec7b420d542c3415a61fc3075f39e1529bc4e6f54f1170496393485f6ef271ad8b587c8fae8a18"
```

### Next PRs (TODOs for me):
1) return error text (as plain json string) if request has failed (@ngotchac request)
2) support t-of-N signatures, when t is odd
3) refactoring to eliminate almost duplicate code from signing && decryption sessions (and, possibly, generation/encryption session)
4) add more tests
5) #5502 + security review
6) stress testing + potential performance fixes
6) try to find an easy way to generate ECDSA-compatible signatures


svyatonik:
Test fail seems unrelated, as I haven't changed anything in ethkey in this PR and ethkey is not dependent on secret_store.

tomusdrw:
Please merge latest master to avoid this failure.

PR #5695, opened 2017-05-25
SecretStore: return error text along with HTTP status code

svyatonik:
On top of #5694
@ngotchac Here's response when key already exists:
```
> POST /shadow/0000000000000000000000000000000000000000000000000000000000000000/de12681e0b8f7a428f12a6694a5f7e1324deef3d627744d95d51b862afc13799251831b3611ae436c452b54cdf5c4e78b361a396ae183e8b4c34519e895e623c00/1 HTTP/1.1
> Host: localhost:8082
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json
< Content-Length: 68
< Date: Thu, 25 May 2017 11:17:27 GMT
< 
* Connection #0 to host localhost left intact
"\"Internal error: session with the same id is already registered\""
```

Is it enough? Or should I work on more appropriate error message for this specific case? :)


ngotchac:
It's think it's good enough :) Just useful to get something back

PR #5697, opened 2017-05-26
Fixed --datadir option

arkpar:


tomusdrw:
I thought that datadir is supposed to change only database location, no? That exactly what the next line is doing.

arkpar:
Help says it is equivalent to `--base-path`

parity/configuration.rs
```diff
@@ -881,7 +881,7 @@ impl Configuration {
 		use path;
 
 		let local_path = default_local_path();
-		let base_path = self.args.flag_base_path.as_ref().map_or_else(|| default_data_path(), |s| s.clone());
+		let base_path = self.args.flag_base_path.as_ref().or_else(|| self.args.flag_datadir.as_ref()).map_or_else(|| default_data_path(), |s| s.clone());
 		let data_path = replace_home("", self.args.flag_datadir.as_ref().unwrap_or(&base_path));
```
tomusdrw: You should update this to just use `base_path` then.
Currently running with `--base-path /base --datadir /data` will put everything to `/data`

PR #5698, opened 2017-05-26
Recover from empty phrase in dev mode

ngotchac:


js/src/views/Signer/components/RequestOrigin/requestOrigin.js
```diff
@@ -29,7 +29,9 @@ export default class RequestOrigin extends Component {
   static propTypes = {
     origin: PropTypes.shape({
       type: PropTypes.oneOf(['unknown', 'dapp', 'rpc', 'ipc', 'signer']),
-      details: PropTypes.string.isRequired
+      details: PropTypes.shape({
```
tomusdrw: Actually it can be either string (RPC origin, Dapp id, IPC session) or object with `details` in it.

codewiz:
I am building from head, but the bug is still present. Could it be that that the parity-ui-precompiled needs to be updated?

PR #5700, opened 2017-05-27
two small changes in evm

guanqun:
- add one unit test for `is_push` in `instructions.rs`
- split `VMType` into a separate file as suggested by the TODO in `factory.rs` (no functional changes)

### DIFF
--- a/ethcore/src/evm/mod.rs
+++ b/ethcore/src/evm/mod.rs
@@ -19,10 +19,11 @@
 pub mod ext;
 pub mod evm;
 pub mod interpreter;
-mod vmtype;
 #[macro_use]
 pub mod factory;
 pub mod schedule;
+
+mod vmtype;
 mod instructions;
 #[cfg(feature = "jit" )]
 mod jit;
--- a/ethcore/src/evm/mod.rs
+++ b/ethcore/src/evm/mod.rs
@@ -19,6 +19,7 @@
 pub mod ext;
 pub mod evm;
 pub mod interpreter;
+mod vmtype;
 #[macro_use]
 pub mod factory;
 pub mod schedule;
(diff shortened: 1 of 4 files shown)


guanqun:
evmbin doesn't seem to use `VMType`, I did a quick grep:

```
➜  parity $ cd evmbin
➜  evmbin $ git grep VMType
```
The main two folder that are using `VMType` are `ethcore` and `parity`

```
$ git grep VMType | tr "/" " " | awk '{ print $1}' | sort | uniq
ethcore
parity
```
And these two components can `cargo build` and `cargo test` properly. So I think it's fine.

Code-wise, in `evm/mod.rs`, I'm pulling the `VMType` directly into `evm` module, and that's the same behavior as before. As long as other code is using `evm::VMType' instead of `evm::factory::VMType`, then we're fine. :)

guanqun:
Do we have other external repo that depends on this?

BTW: How can I trigger an integration build for this PR?

ethcore/src/evm/mod.rs
```diff
@@ -19,6 +19,7 @@
 pub mod ext;
 pub mod evm;
 pub mod interpreter;
+mod vmtype;
```
rphmeier: style: separate public modules from private modules

ethcore/src/evm/vmtype.rs
```diff
@@ -0,0 +1,75 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::fmt;
+
+#[derive(Debug, PartialEq, Clone)]
+/// Type of EVM to use.
```
rphmeier: style: docs over attributes

ethcore/src/evm/mod.rs
guanqun: Sure, moved down a bit.

ethcore/src/evm/vmtype.rs
guanqun: Fixed.

NikVolf:
https://github.com/paritytech/parity/pull/5718
for the CI check

gavofyork:
谢谢!

guanqun:
不用谢 :)

PR #5716, opened 2017-05-30
Fix local transactions without condition.

### LINKED ISSUE #5715: Can't send transaction, can't "Reveal" on ENS - nothing happens!
In the last few hours I haven't been able to reveal any bids, it just says "Revealing" and when I come back to it I need to do it again, nothing happens.

I also can't make any new bids, or even send transactions to my other ETH accounts. It just says "Waiting for confirmations...." forever.

Anyone knows what's going on?

tomusdrw:
It was not possible to cancel transaction that didn't have a condition.
In fact no transaction was displayed, cause `TxRow` expected every transaction to always have a condition.

Closes #5715

### DIFF
--- a/js/src/ui/TxList/TxRow/txRow.js
+++ b/js/src/ui/TxList/TxRow/txRow.js
@@ -210,31 +210,36 @@ class TxRow extends Component {
 
     if (!isCancelOpen && !isEditOpen) {
       const pendingStatus = this.getCondition();
+      const isPending = pendingStatus === 'pending';
 
-      if (pendingStatus === 'submitting') {
-        return (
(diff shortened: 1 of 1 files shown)


CraigglesO:
Yeah, this looks good to me. Real shame that mistake. It's hard to check every use case sometimes. 

I wanted to test the branch myself, but I keep getting errors even using the branches rust compiled. Which branch of rust are you using to make the --ui-no-valdiation work?

tomusdrw:
@CraigglesO No worries, we didn't catch it during the review either, and it wasn't that obvious that this functionality is actually required.
You can run with `--ws-origins all` as a workaround for now.

PR #5731, opened 2017-05-31
Require phrase confirmation.

tomusdrw:
I've decided to allow users to go back on the confirmation page - I think it would be very frustrating if you went through the process and couldn't finish it cause you didn't backup the phrase (especially on test networks).

On non-test networks pasting the whole phrase is also blocked (you can paste up to 8 characters - so pasting word by word should work fine).

### DIFF
--- a/js/src/modals/CreateAccount/AccountDetails/accountDetails.js
+++ b/js/src/modals/CreateAccount/AccountDetails/accountDetails.js
@@ -102,51 +102,47 @@ export default class AccountDetails extends Component {
       />
     );
     const label = (
-     <FormattedMessage
+      <FormattedMessage
         id='createAccount.accountDetails.phrase.label'
         defaultMessage='owner recovery phrase'
       />
     );
 
     if (!isConfirming) {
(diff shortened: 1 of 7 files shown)


CraigglesO:
I don't like that people can't paste in the full thing. Forcing them to copy paste in the first place already gets the user thinking about the importance.
I also think they should just paste in the security phrase.
I mean realistically, if they make a mistake and don't have the security phrase, just make another. it's incredibly easy.

tomusdrw:
Ok, so the flow is a bit changed now:
1. When the phrase is displayed user needs to type in "I have written down the phrase" to go to the next step.
2. On the next step she needs to type the fully recovery phrase (and there is no back button).

Pasting is disabled on non-test networks. If we allow pasting there people will just store the phrase in the clipboard and won't back it up.
IMHO there has been too many people that didn't back up the phrase earlier (even though we outline the importance of that) and lost their ETH because of that. We really should do whatever is (reasonably) possible to force them to back it up.

PR #5734, opened 2017-05-31
Public node using WASM

maciejhirsz:
- Replace secp256k1 and pharse to address generation with a WASM module.
- Worker pool starts with 2 workers and never grows - incoming actions wait for a worker to become free. This is due to the fact that not just workers are heavy, but WASM itself has some requirements (Emscripten seems to insist on minimum 15mb of memory being available).
- Starting Parity with `--public-node` disables signer token.


js/src/api/local/localAccountsMiddleware.spec.js
```diff
@@ -32,7 +32,7 @@ class MockedTransport extends JsonRpcBase {
   }
 }
 
-describe('api/local/LocalAccountsMiddleware', function () {
+describe.only('api/local/LocalAccountsMiddleware', function () {
```
tomusdrw: Should be just `describe`.

parity/rpc.rs
```diff
@@ -141,6 +141,7 @@ pub struct WsConfiguration {
 	pub origins: Option<Vec<String>>,
 	pub hosts: Option<Vec<String>>,
 	pub signer_path: PathBuf,
+	pub require_token: bool,
```
tomusdrw: I think `support_full_api` or `support_token_api` would be a bit better name.

js/wasm/ethkey/src/main.rs
```diff
@@ -0,0 +1,153 @@
...
+use core::intrinsics;
+
+// Pull in the system libc library for what crt0.o likely requires.
+extern crate libc;
+extern crate tiny_keccak;
+extern crate tiny_secp256k1;
+
+use tiny_secp256k1::{is_valid_secret, create_public_key, ECPointG};
+
+#[link_args = "-s EXPORTED_FUNCTIONS=['_input_ptr','_secret_ptr','_public_ptr','_address_ptr','_ecpointg','_verify_secret','_brain']"]
```
NikVolf: `EXPORTED_FUNCTIONS` emcc arg is no longer required afaik, all pub functions in binary with `[no_mangle]` will be available as exports if i'm not mistaken

js/wasm/ethkey/build.sh
```diff
@@ -0,0 +1,16 @@
+# Remove previous build to avoid name conflicts
```
NikVolf: Not sure how it will run on windows?

maciejhirsz: It doesn't need to. The built wasm is in the repo, this script isn't being run when building Parity.

js/wasm/ethkey/src/main.rs
rphmeier: they should also be marked `extern` to force them to adhere to C calling convention. otherwise they are only safe to call from Rust code

jacogr:
lgtm from a JS perspective

PR #5742, opened 2017-06-01
use NULL_RLP, remove NULL_RLP_STATIC

guanqun:
I thought `NULL_RLP_STATIC` was created for a reason, but I can't think of the specific reason, even the Rust book suggests we should prefer `const`. So this PR, but maybe it has other reasons, then feel free to close this PR.

`cargo build` and `cargo test` both seem ok.

### DIFF
--- a/ethcore/src/account_db.rs
+++ b/ethcore/src/account_db.rs
@@ -18,8 +18,6 @@
 use util::*;
 use rlp::NULL_RLP;
 
-static NULL_RLP_STATIC: [u8; 1] = [0x80; 1];
-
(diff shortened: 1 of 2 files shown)


rphmeier:
LGTM, still working on getting CI up for out-of-ord PRs! Thanks for your patience and repeated contributions :)

guanqun:
Thank you all to come up with this great project. For me, I'm only able to work on this part time (limited time per day), so bear with me to create these small/trivial PRs. Hopefully when I get more familiar with the whole codebase, I can send bigger PR.

PR #5746, opened 2017-06-01
No bueno null account

CraigglesO:
I think this is the safest patch for now. Just for foundation, completely ignore address `0x00a329c0648769A73afAc7F9381E08FB43dBEA72`.

Another 1ish ETH was lost today. Just for now, completely ignore the address. Save people from the mistake of thinking it's theirs to use.


tomusdrw:
There is another PR pending that removes this account on RPC level: #5730

CraigglesO:
That's even better.

Issue #4194, opened 2017-01-17
Automatic dapps refreshing


gavofyork:
Currently you have to restart parity to get any new local dapps listed (or the details of existing ones updated). Would be nice if this were automated - ideally just by watching the dapps path.

PR #5752, opened 2017-06-02
Dapp refresh

CraigglesO:
Closes: #4194

alt1
alt2

If adding or removing a dapp, this will update the list upon:
* removing a dapp from the dapps folder
* linking a dapp
* a 404.

If you want to quickly test it:
`git clone https://github.com/paritytech/skeleton mydapp`
`git cd mydapp`

`ln -s $PWD/dist $HOME/Library/Application\ Support/io.parity.ethereum/dapps/mydapp`
(If you are using windows go to the tutorial... :P)

**Click refresh**

and to test removal:
`cd ~/Library/Application Support/io.parity.ethereum/dapps`
or `cd /Users/$USER/Library/Application Support/io.parity.ethereum/dapps`

`rm -rf mydapp`


dapps/src/endpoint.rs
```diff
@@ -39,7 +40,7 @@ pub struct EndpointInfo {
 	pub icon_url: String,
 }
 
-pub type Endpoints = Arc<BTreeMap<String, Box<Endpoint>>>;
+pub type Endpoints = Arc<RwLock<BTreeMap<String, Box<Endpoint>>>>;
```
rphmeier: why add `RwLock` if it's only read through? doesn't seem to need any lock at all

CraigglesO: Well, the end goal is to push to the BTreeMap new dapp entries as a js developer adds. Currently you have to restart parity to get the new dapp.

dapps/src/router.rs
```diff
@@ -61,14 +62,16 @@ impl http::RequestMiddleware for Router {
 		let is_head_request = *req.method() == hyper::Method::Head;
 
 		trace!(target: "dapps", "Routing request to {:?}. Details: {:?}", url, req);
+		let refresh = self.endpoints.clone();
```
tomusdrw: Why is it cloned here?

dapps/src/router.rs
```diff
@@ -109,6 +112,7 @@ impl http::RequestMiddleware for Router {
 				=>
 			{
 				trace!(target: "dapps", "Resolving to 404.");
+				refresh.unwrap().refresh_local_dapps();
```
tomusdrw: `self.endpoints` can be empty so you should unwrap.
It should be just `self.endpoints.as_ref().map(|endpoints| endpoints.refresh_local_dapps())`.

Even though the dapps are refreshed we should still re-try resolving the request.
Most likely a recursive call should be done here with some additional parameter preventing a refresh again.

js/src/api/rpc/parity/parity.js
```diff
@@ -95,6 +95,11 @@ export default class Parity {
       .execute('parity_dappsList');
   }
 
+  dappsRefresh () {
```
tomusdrw: I don't think the RPC method (and JS code) will be useful at all if the router can auto-refresh missing dapps.

CraigglesO: Ah! Thanks for the review! I'll reply when I wake up in a few hours. The JS code just ensures an instant refresh.. otherwise you have to reload the page (<super>+R), and I don't think we should require that. It would be confusing to the end user.

So for instance, if you link a new dapp... how would JS know unless it constantly checked if there was a new one... [When pub-sub is ready, I can change it] but this is incredibly inefficient since the number of times a developer is going to add a new dapp is like... once a month at most?

dapps/src/apps/mod.rs
```diff
@@ -80,7 +81,34 @@ pub fn all_endpoints<F: Fetch>(
...
+		}
+	}
+	// remove those dead dapps
+	for k in v.iter() {
+        pages.write().remove(k.as_str());
+    }
+	// new dapps to be added
+	for (k, v) in new_pages {
+		if pages.read().contains_key(&k) != true {
+			pages.write().insert(k, v);
```
CraigglesO: I must have bad luck, but the reason I used a vector and the reason I use short term locks, (and maybe this is a unique issue to parking_lot) is that if you have a read lock, the write lock will wait indefinitely for the read lock to finish before creating it's own lock. Has no one else run into this?

tomusdrw: You should only take one `write()` lock at the top (you will be able to read from it as well).

js/src/api/rpc/parity/parity.js
tomusdrw: Yeah, I think refreshing the page is actually ok when you add a new dapp. As you said it will happen once a month at most.
IMHO not worth supporting additional RPC method and this additional JS code.

But I don't have very strong opinion on that. Since the code is already there we can probably merge it.

dapps/src/apps/mod.rs
rphmeier: it's actually a safety feature to deadlock instead of aliasing writeable pointers

CraigglesO: Ok, so:
a) if the function is not too complex, it's ok to have a write lock for the time of the function
b) you can add reads after a write.

dapps/src/apps/ui.rs
```diff
@@ -22,6 +22,7 @@ use endpoint::{Endpoint, Handler, EndpointPath};
 use {address, handlers};
 
 /// Redirection to UI server.
+#[derive(Clone)]
```
CraigglesO: Didn't clean it after playing with the code -_-

gavofyork:
not building?

gavofyork:
definitely not building
https://gitlab.parity.io/parity/parity/-/jobs/59584

CraigglesO:
Fixed all the test cases. All good now.

gavofyork:
@tomusdrw would be good to get your eyes on this before you jet off :)

js/src/views/Dapps/dappsStore.js
```diff
@@ -277,6 +291,12 @@ export default class DappsStore extends EventEmitter {
 
       this.apps = this.apps
         .filter((app) => !app.id || !newAppsIds.includes(app.id))
+        .filter((app) => { // check local dapp changes
+          if (app.type === 'local' && _local && apps.indexOf(app) === -1) {
```
tomusdrw: This additional filter seems redundant, local apps that are not part of `apps` array will be removed by previous condition anyway `!newAppsIds.includes(app.id)` (unless local apps doesn't have `id` field?)

parity/dapps.rs
```diff
@@ -305,5 +305,10 @@ mod server {
 				})
 				.collect()
 		}
+
+		fn refresh_local_dapps(&self) -> bool {
+			self.endpoints.clone().refresh_local_dapps();
```
tomusdrw: Cloning here is invalid since we were just modifying data in clone.

rpc/src/v1/helpers/dapps.rs
```diff
@@ -22,12 +22,18 @@ use v1::types::LocalDapp;
...
+/// TODO: Update this trait to a struct
 impl<F> DappsService for F where
-	F: Fn() -> Vec<LocalDapp> + Send + Sync + 'static
+	F: Fn(bool) -> (Vec<LocalDapp>, bool) + Send + Sync + 'static
 {
 	fn list_dapps(&self) -> Vec<LocalDapp> {
-		(*self)()
+		(*self)(true).0
 	}
+    fn refresh_local_dapps(&self) -> bool {
```
tomusdrw: Bad indentation.

rpc/src/v1/helpers/dapps.rs
```diff
@@ -22,12 +22,18 @@ use v1::types::LocalDapp;
...
 	/// List available local dapps.
 	fn list_dapps(&self) -> Vec<LocalDapp>;
+	/// Refresh local dapps list
+	fn refresh_local_dapps(&self) -> bool;
 }
 
+/// TODO: Update this trait to a struct
 impl<F> DappsService for F where
-	F: Fn() -> Vec<LocalDapp> + Send + Sync + 'static
+	F: Fn(bool) -> (Vec<LocalDapp>, bool) + Send + Sync + 'static
```
tomusdrw: Why is this implementation even needed? Is it only for tests? If yes it would be best to follow the todo above.

dapps/src/lib.rs
```diff
@@ -101,31 +101,51 @@ impl<F> WebProxyTokens for F where F: Fn(String) -> Option<Origin> + Send + Sync
...
 	pub fn list(&self) -> Vec<apps::App> {
-		self.endpoints.iter().filter_map(|(ref k, ref e)| {
+		self.endpoints.read().iter().filter_map(|(ref k, ref e)| {
 			e.info().map(|ref info| apps::App::from_info(k, info))
 		}).collect()
 	}
+	/// Check for any changes in the local dapps folder and update.
+	pub fn refresh_local_dapps(&mut self) {
+		let new_local = apps::fs::local_endpoints(self.dapps_path.clone(), self.embeddable.clone());
+		let (local_endpoints, to_remove) = self.local_endpoints.clone().into_iter().partition(|k| new_local.contains_key(&k.clone()));
```
tomusdrw: Could be `let (local_endpoints, to_remove) = mem::replace(&mut self.local_endpoints, vec![]).into_iter().parition(...)` to avoid cloning.

dapps/src/lib.rs
```diff
@@ -101,31 +101,51 @@ impl<F> WebProxyTokens for F where F: Fn(String) -> Option<Origin> + Send + Sync
...
+		self.local_endpoints = local_endpoints;
+		let mut pages = self.endpoints.write();
+		// remove the dead dapps
+		for k in to_remove {
+			pages.remove(&k);
+		}
+		// new dapps to be added
+		for (k, v) in new_local {
+			if pages.contains_key(&k) != true {
```
tomusdrw: `!pages.contains_key(&k)` seems clearer.

dapps/src/router.rs
```diff
@@ -88,9 +90,9 @@ impl http::RequestMiddleware for Router {
...
 			},
 			// Then delegate to dapp
-			(Some(ref path), _, _) if self.endpoints.as_ref().map(|ep| ep.contains_key(&path.app_id)).unwrap_or(false) => {
+			(Some(ref path), _, _) if endpoints.as_ref().map(|ep| ep.contains_key(&path.app_id)).unwrap_or(false) => {
 				trace!(target: "dapps", "Resolving to local/builtin dapp.");
-				Some(self.endpoints
+				Some(endpoints
 					.as_ref()
 					.expect("endpoints known to be set; qed")
 					.get(&path.app_id)
```
tomusdrw: Seems that refreshing on 404 is gone now, so it's only manual afaict.

dapps/src/apps/mod.rs
```diff
@@ -64,9 +65,10 @@ pub fn all_endpoints<F: Fetch>(
 	web_proxy_tokens: Arc<WebProxyTokens>,
 	remote: Remote,
 	fetch: F,
-) -> Endpoints {
+) -> (Vec<String>, Endpoints) {
 	// fetch fs dapps at first to avoid overwriting builtins
-	let mut pages = fs::local_endpoints(dapps_path, embeddable.clone());
+	let mut pages = fs::local_endpoints(dapps_path.clone(), embeddable.clone());
+	let local_endpoints: Vec<String> = pages.iter().map(|(k,_)| k.clone()).collect();
```
tomusdrw:
```
let local_endpoints = pages.keys().cloned().collect();
```

dapps/src/lib.rs
```diff
@@ -101,31 +101,51 @@ impl<F> WebProxyTokens for F where F: Fn(String) -> Option<Origin> + Send + Sync
...
 impl Endpoints {
 	/// Returns a current list of app endpoints.
 	pub fn list(&self) -> Vec<apps::App> {
-		self.endpoints.iter().filter_map(|(ref k, ref e)| {
+		self.endpoints.read().iter().filter_map(|(ref k, ref e)| {
 			e.info().map(|ref info| apps::App::from_info(k, info))
 		}).collect()
 	}
+	/// Check for any changes in the local dapps folder and update.
+	pub fn refresh_local_dapps(&mut self) {
```
tomusdrw: This method shouldn't be `&mut` the `local_endpoints` should be wrapped in `Arc<RwLock` instead (like `endpoints`).

CraigglesO: Even though it's only used in this function?

rpc/src/v1/helpers/dapps.rs
CraigglesO: https://github.com/paritytech/parity/blob/04bd4d9559e1f99259fcc41b5a5fd7e4e69f2c7e/parity/dapps.rs#L294

Yeah, I'll change it.

rpc/src/v1/helpers/dapps.rs
CraigglesO: the indentation is correct in my code. So what do I do to fix this?

parity/dapps.rs
CraigglesO: Except there is a cascading effect. If I don't clone here, I'll have to make it mutable. And then I have to change `rpc::src::v1::helpers::dapps::DappsService` which means all mutable borrows will have to change.

https://doc.rust-lang.org/std/sync/struct.Arc.html

Unless changing the rwlock right away, isn't it standard procedure to clone an arc before use?

rpc/src/v1/helpers/dapps.rs
rphmeier: it's spaces instead of tabs, you should replace them

dapps/src/router.rs
CraigglesO: I didn't want to dispute this, but now that I am re-incorporating it, I want to do it as so SINCE the code is only run upon the condition `self.endpoints.is_some()`

CraigglesO:
The crazy thing is that the bug would only rear it's ugly head some cases. Simplify filter push proves that. Anyways, the test cases would freeze on a `let mut pages = self.endpoints.write()` SOOOO Unless someone knows how to make that stop I'm making it verbose:
```
pub fn refresh_local_dapps(&self) {
		let mut local = self.local_endpoints.write();
		let new_local = apps::fs::local_endpoints(self.dapps_path.clone(), self.embeddable.clone());
		let (_, to_remove): (Vec<String>, Vec<String>) = local.clone()
			.into_iter()
			.partition(|k| new_local.contains_key(&k.clone()));
		// remove the dead dapps
		for k in to_remove {
			let index = local.iter().position(|x| *x == k).unwrap();
			local.remove(index);
			self.endpoints.write().remove(&k);
		}
		// new dapps to be added
		for (k, v) in new_local {
			if self.endpoints.write().contains_key(&k) != true {
				local.push(k.clone());
				self.endpoints.write().insert(k, v);
			}
		}
	}
```

js/src/views/Dapps/dappsStore.js
CraigglesO: No, that filter is not designed to actually remove old apps, only add usable apps.. 

 don't want to rewrite this module for this, but for UI-2 I'll redo the module because ideally ALL apps should be removable and addable. But that's a PR for another time. This is minimally invasive.

tomusdrw:
Waiting for #6192

tomusdrw:
I believe ` js/package-lock.json` is produced by new NPM, right? I suppose we should finally check that in: @jacogr ?

PR #5756, opened 2017-06-04
Use randomized subscription ids for PubSub

tomusdrw:


rpc/src/v1/helpers/subscribers.rs
```diff
@@ -16,45 +16,75 @@
...
+	pub fn as_string(&self) -> String {
+		format!("{:?}", self.0)
+	}
+}
+
+#[derive(Clone)]
 pub struct Subscribers<T> {
-	next_id: u64,
-	subscriptions: HashMap<u64, T>,
+	rand: StdRng,
```
rphmeier: why use random IDs? is it a reason that's defeated by not using cryptographic-strength randomness?

tomusdrw: When you run public node it would be possible to cancel other peoples subscriptions. We could instead have an id per session, but I think it would be a bit more complicated.

rphmeier: i see, but you could still guess and get lucky and cancel others'?

tomusdrw: Yes, you could even try to fire milions of requests and cancel all of them. But that should be that should be handled by some rate limiting (that is needed anyway).

PR #5763, opened 2017-06-05
Bump mio

arkpar:
Fixes crashes on windows

### DIFF
--- a/util/network/src/lib.rs
+++ b/util/network/src/lib.rs
@@ -54,6 +54,9 @@
 //! }
 //! ```
 
+//TODO: use Poll from mio
+#![allow(deprecated)]
+
 extern crate ethcore_io as io;
 extern crate ethcore_util as util;
 extern crate parking_lot;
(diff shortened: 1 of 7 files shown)


util/network/src/lib.rs
```diff
@@ -54,6 +54,9 @@
 //! }
 //! ```
 
+//TODO: use Poll from mio
+#![allow(deprecated)]
```
NikVolf: how about limit this allowance to deprecated mio imports only?
this todo might live for ages there

arkpar: Is there a way to do that? Setting an attribute on `extern crate mio` does not work.

NikVolf: you have to mark every `use::mio::<DeprecatedItem>` with this `#[allow(deprecated)]` I believe

but let me double-check

NikVolf: Ok, this doesn't work, it supresses only `use` itself and still warns when actually used
Disregard this remark, then

PR #5774, opened 2017-06-06
PoA: Wait for transition finality before applying

rphmeier:
Summary of changes:
  - Engine::on_new_block gets a flag indicating if the block was the first of an epoch. For PoA engines this will lead to the contracts `finalizeChange` function being called.
  - Epoch changes split into two modes: signals and instant changes. Switching to a new fixed list is an instant change, but switching to a contract requires finality so it just signals a change.
  - Signals lead to instant changes at the block where they are finalized.
  - `authority_round::RollingFinality` keeps track of the current finality state for Aura engines.
  - Warp snapshots for PoA include a proof of the change and a proof of finality if necessary.
  - `EpochVerifier` gains a `check_finality_proof` function which will check a finality proof under the epoch.
  - `AuthorityRound` configuration gains a `immediateTransitions` flag to retain the old behavior and disregard 

Although the API looks complicated, it's largely just to support light clients easily -- checking on `epoch_end_signal` requires nothing more than the block and receipts, as generating a proof is done only on full nodes. 

`is_epoch_end` doesn't have access to the state, which enforces that any epoch changes requiring secure access to state are only applied on finality.

Tendermint hasn't been ported yet, but it should be pretty quick


rphmeier:
Core logic is done now, but tests still need patching up. I'll leave a list of things I'd like to have automated tests against before merging:
  - [ ] `epoch_transition_for` on canonical/non-canonical chains
  - [ ] test that transition to fixed list is instant
  - [ ] test that transition to contract is not instant
  - [ ] test that transition within contract is also not instant
  - [ ] test snapshot on contract chain.
  - [ ] test snapshot on chain that transitions to contract.

Also the snapshot warp target needs to be changed to target the most recent finalized block to prevent an unfinalized pending transitions to be missed when there is a gap.

rphmeier:
I'd also really like to have "import tests" run as part of the CI: 
basically, we'll have a submodule containing tests. each test will consist of a chain spec and blocks to import and we'll test (using freshly compiled parity) `parity import --chain spec_file blocks_file` completes without issue. We can do something similar with snapshots.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -413,27 +497,160 @@ impl Engine for AuthorityRound {
...
+				x @ &mut None => {
+					// epoch_transition_for can be an expensive call, but in the absence of
+					// forks it will only need to be called for the block directly after
+					// epoch transition, in which case it will be O(1) and require a single
+					// DB lookup.
+					let last_transition = match client.epoch_transition_for(chain_head.hash()) {
+						Some(t) => t,
+						None => {
+							// this really should never happen :)
+							debug!(target: "engine", "No genesis transition found.");
```
NikVolf: why not panic here, then

json/src/spec/authority_round.rs
```diff
@@ -49,6 +49,9 @@ pub struct AuthorityRoundParams {
 	/// Block from which monotonic steps start.
 	#[serde(rename="validateStepTransition")]
 	pub validate_step_transition: Option<Uint>,
+	/// Whether transitions should be immediate.
+	#[serde(rename="immediateTransitions")]
+	pub immediate_transitions: Option<bool>,
```
keorn: Shouldn't that be only relevant to `ValidatorSet`s that do transitions (unlike `SimpleList`)?

rphmeier: Transitions to `SimpleList` in a `Multi` set are instant regardless of the setting of this flag. It could be a per-validator-set flag, but it makes more sense to have it global because
  a) this flag influences whether finality checking is done, which is at the engine level
  b) if any single validator set doesn't finalize transitions as necessary, then warp sync and light sync won't work and it's the same effect as having none of the sets finalize transitions.

ethcore/src/snapshot/consensus/authority.rs
```diff
@@ -413,84 +332,57 @@ impl Rebuilder for ChunkRebuilder {
...
-		if self.last_proofs.len() != self.unverified_firsts.len() + 1 {
-			return Err(Error::WrongChunkFormat("More than one 'last' chunk".into()).into());
-		}
-
 		// verify the first entries of chunks we couldn't before.
-		let lasts_iter = self.last_proofs.iter().map(|&(_, ref hdr, ref proof)| (hdr, &proof[..]));
-		let firsts_iter = self.unverified_firsts.iter().map(|&(_, ref hdr)| hdr);
+		// we store all last verifiers, but not all firsts.
+		// match each unverified first epoch with a last epoch verifier.
+		let mut lasts_reversed = self.last_epochs.iter().rev();
```
arkpar: Is this efficient enough?

ethcore/src/snapshot/consensus/authority.rs
```diff
@@ -413,84 +332,57 @@ impl Rebuilder for ChunkRebuilder {
...
+		for &(ref header, ref finality_proof, hash) in self.unverified_firsts.iter().rev() {
+			let mut found = false;
+			while let Some(last) = lasts_reversed.next() {
```
arkpar: More idiomatic version: `while let Some((header, verifier)) =`

ethcore/src/snapshot/consensus/authority.rs
rphmeier: a reversed `[T]` iterator is just pointer-bumping, same as an in-order iterator. The loop itself just walks two vectors in reverse, which should be pretty fast.

arkpar:
LGTM, I've looked into changes outside of `authority_round` mostly.

arkpar:
needs merging

ethcore/src/snapshot/tests/test_validator_contract.json
```diff
@@ -39,11 +43,11 @@
...
...0910390a25b50565b6103276104c7565b60008054806020026020016040519081016040528092919081815260200182805480156103a957602002820191906000526020600020905b8160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001906001019080831161035f575b505050505090505b90565b5b505050565b5b5050565b8280548282559060005260206000209081019282156104005760005260206000209182015b828111156103ff5782548255916001019190600101906103e4565b5b50905061040d91906104db565b5090565b81548183558181151161043857818360005260206000209182019101610437919061051e565b5b505050565b8280548282559060005260206000209081019282156104b6579160200282015b828111156104b55782518260006101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff1602179055509160200191906001019061045d565b5b5090506104c391906104db565b5090565b602060405190810160405280600081525090565b61051b91905b8082111561051757600081816101000a81549073ffffffffffffffffffffffffffffffffffffffff0219169055506001016104e1565b5090565b90565b61054091905b8082111561053c576000816000905550600101610524565b5090565b905600a165627a7a7230582041ce7e5c820bc89b1a330a3233c4f3013e77433ecba368fa234adf758d87fe1d0029"
```
afck: Where can I find the source code for this?

afck: Ah, sorry, found the link to the gist.

ethcore/src/snapshot/tests/test_validator_contract.json
```diff
@@ -39,11 +43,11 @@
...
```
afck: Is that really the code from https://gist.github.com/anonymous/2a43783647e0f0dfcc359bd6fd81d6d9?
https://ethervm.io/decompile says this code contains a `finalizeChange` function, but the gist has a `finalizeSignal`. Also, the signature of `reportMalicious` seems to be different.

Issue #4401, opened 2017-02-02
Duration limit required even if Homestead transition is zero


lgpawel:
Minor annoyance: if using a chain spec with `homesteadTransition` at 0, `durationLimit` should not be required, as the Homestead difficulty evolution does not use this parameter at all. Yet in such cases Parity quits with `Spec json is invalid`. (I presume this may be fixed easily by just letting the `durationLimit` to default to, say, `0xd` whenever it's not present, regardless of the `homesteadTransition` value.)

5chdn:
This issue is labelled with _annoyance_: The client behaves within expectations, however this "expected behaviour" itself is at issue. It's inactive since February, neither assigned nor linked to any milestone.

@paritytech/ui-devs Please decide on a deadline and add an assignee within 7 days, thanks!

5chdn:
Yes, sorry for the noise.

PR #5777, opened 2017-06-06
Duration limit made optional for EthashParams

grbIzl:
#4401 Making duration limit parameter optional as it's not used in case of homestead transition. I didn't force any additional checks in tests. Just modify them in order to reflect the change.

### DIFF
--- a/json/src/spec/ethash.rs
+++ b/json/src/spec/ethash.rs
@@ -39,7 +39,7 @@ pub struct EthashParams {
 	pub metropolis_difficulty_increment_divisor: Option<Uint>,
 	/// See main EthashParams docs.
 	#[serde(rename="durationLimit")]
-	pub duration_limit: Uint,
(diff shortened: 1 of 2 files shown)


NikVolf:
@rphmeier it's Anton Gavrilov, no CLA needed

rphmeier:
ah, sorry Anton :)

json/src/spec/ethash.rs
```diff
@@ -39,7 +39,7 @@ pub struct EthashParams {
 	pub metropolis_difficulty_increment_divisor: Option<Uint>,
 	/// See main EthashParams docs.
 	#[serde(rename="durationLimit")]
-	pub duration_limit: Uint,
+	pub duration_limit:Option<Uint>,
```
arkpar: a space is missing

grbIzl: My bad (

Issue #5739, opened 2017-06-01
Problem with connection between nodes in PoA chain


Ochir:
I created 5 instances of Parity (1.6.7) nodes in Azure cloud and created PoA chain. 
When there are few connections, it works OK. 
However, I constantly see problems with connection, when  more then 20 transactions are sent to one of nodes. 
Below is log. 

    _2017-06-01 11:31:00 UTC  DEBUG own_tx  Imported to Current (hash 8bd465f22caf4e3e1a3575c516c6c83e5120beeae7e713b9d919e41ea63fce5b)
    2017-06-01 11:31:00 UTC  TRACE own_tx  Status: TransactionQueueStatus { pending: 3, future: 22 }
    2017-06-01 11:31:00 UTC  TRACE own_tx  Importing transaction: PendingTransaction { transaction: SignedTransaction { transaction: UnverifiedTransaction { unsigned: Transaction { nonce: 1048581, gas_price: 0, gas: 400000, action: Call(8
    a1c79cd0f1a3989c37f3e2a5bb64a7877627c57), value: 0, data: [191, 253, 151, 184, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
    , 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 17, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 9, 136, 124, 216, 102, 1, 140, 96, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 80, 210, 119, 25
[...]

PR #5778, opened 2017-06-06
A CLI flag to allow fast transaction signing when account is unlocked.

tomusdrw:
Sending 250 txs: ~20s
Sending 250 txs with --fast-unlock: ~1.4s


5chdn:
Does this address #5739 ?

tomusdrw:
Yes. But it doesn't really explain the difference in behaviour between 1.6.6 and 1.6.7. I'm in favour of closing this and re-opening if the problem is not fixed in 1.7.

ethcore/src/account_provider/mod.rs
```diff
@@ -116,15 +116,22 @@ type AccountToken = String;
...
+	/// Dapps settings.
 	dapps_settings: RwLock<DappsSettingsStore>,
 	/// Accounts on disk
 	sstore: Box<SecretStore>,
 	/// Accounts unlocked with rolling tokens
 	transient_sstore: EthMultiStore,
 	/// Accounts in hardware wallets.
 	hardware_store: Option<HardwareWalletManager>,
+	/// When unlocking permanently on for some time store a raw secret instead of password.
+	fast_unlock: bool,
```
debris: it needs better description. I can't understand this one

NikVolf:
Why providing another flag and not just storing private key instead of password, always, as we originally did for this unsafe unlocks?
Isn't storing private key provides less attack surface comparing to storing the password (that can be used to access multiple private keys)?

tomusdrw:
> Isn't storing private key provides less attack surface comparing to storing the password (which can be used to access multiple private keys)?

Only if you assume that the attacker has read access to the filesystem (and knows where the private keys are). I think storing a password is strictly more secure, since it's not enough to scan memory/swapfile to get full access to your account.

In a very paranoid scenario you can even copy or mount a volume with private keys (encrypted or over the network) only when some transactions are supposed to be signed, while still having the account unlocked permantently.

@debris description updated.

PR #5781, opened 2017-06-07
update the source for the snapcraft package

elopio:
The source attribute of the snapcraft.yaml is relative to the directory where snapcraft is called.
When there's a snap directory in the root of the proyect, snapcraft should be called from the root, and thus we were going back one extra directory.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,6 +17,6 @@ apps:
 
 parts:
   parity:
-    source: ..
+    source: .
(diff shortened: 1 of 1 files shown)


elopio:
Now that you moved the snapcraft.yaml to the snap directory, there's a duplicated one in the scripts directory. Should we remove that one?

5chdn:
Putting this on ice until the @paritytech/ci decides which script they use as discussed in #5496

PR #5784, opened 2017-06-07
Block invalid account name creation

jacogr:
- Sets name error properly on empty name between next/prev
- Fixes clicking rapidly on next that creates invalid accounts in case where name was previously set
- Fixed token balance checks (currently Geth key imports are not working)
- Additional safety checks for empty phrase (fromNew) and canCreate in actual action handler

### DIFF
--- a/js/src/modals/CreateAccount/createAccount.js
+++ b/js/src/modals/CreateAccount/createAccount.js
@@ -332,12 +332,10 @@ class CreateAccount extends Component {
     return this.createStore
       .createAccount(this.vaultStore)
       .then(() => {
-        this.createStore.setBusy(false);
         this.createStore.nextStage();
         this.props.onUpdate && this.props.onUpdate();
       })
       .catch((error) => {
-        this.createStore.setBusy(false);
         this.props.newError(error);
       });
   }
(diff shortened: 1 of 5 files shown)


js/src/modals/CreateAccount/store.js
```diff
@@ -91,7 +91,11 @@ export default class Store {
   }
 
   @computed get hasAddress () {
-    return !!(this.address);
+    return this.address.length !== 0;
```
tomusdrw: `this.address` is never `null/undefined`?

js/src/modals/CreateAccount/createAccount.js
```diff
@@ -329,8 +329,6 @@ class CreateAccount extends Component {
   }
 
   onCreate = () => {
-    this.createStore.setBusy(true);
```
tomusdrw: Seems that we don't need to `setBusy(false)` as well (since it's done inside `createAccount` in store)

jacogr: +1

js/src/modals/CreateAccount/store.js
jacogr: You are right, will revert - extra safety. (It _should_ be empty, but no need to muck about)

PR #5786, opened 2017-06-07
Disable compression for RLP strings

### LINKED ISSUE #5779: getTransactionreceipt returns null for some transaction on ETC blockchain
Hi
I'm currently running ETC with Parity/v1.4.10-stable-4e6329a-20170118/x86_64-linux-gnu/rustc1.15.0

At block 3862300,
Calling eth_getTransactionReceipt to 0x0e914f05f01f3aa28c092bd1140094dffc32917a06352517ff0e96aa396ce901
returns null which should give some valid value.

gastracker also outputs error on viewing 3862300 block.
http://gastracker.io/block/3862300
( http://gastracker.io/block/3862301 works well. Seems like gastracker is using parity behind )

I can explore that block from etcchain block explorer.
https://etcchain.com/block/number/3862300

0x0e914f05f01f3aa28c092bd1140094dffc32917a06352517ff0e96aa396ce901
transaction has 0 amount transfer.
I think this caused some bug retrieving eth_getTransactionReceipt.

arkpar:
Fixes #5779


keorn:
better just change [this](https://github.com/paritytech/parity/blob/9aff1b3ce73cfdcb6c5bf6651d57421a275a969e/util/rlp/src/compression.rs#L123) to `simple_compress`

tomusdrw:
lgtm

splix:
What version of Parity is safe to use now? It seems that 1.4.x and 1.5.x are prone to this bug. What about 1.3.x and 1.6.x?

arkpar:
1.6.8 Contains the fix. A re-sync is required though. Versions 1.3.0-1.6.7 are affected.

splix:
@arkpar do I need to resync from the genesis, or it's enough to resync from block 3855567, which is seems to be a first block having this issue?

splix:
Also, are there plans to include this fix into 1.5.x?

PR #5790, opened 2017-06-07
Content Security Policy

### LINKED ISSUE #4315: Dapps server should return ContentSecurityPolicy headers
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

tomusdrw:
Closes #4315 

Also fixes issue on firefox when accessing dapps on `http://localhost:8180`

### DIFF
--- a/js/src/ui/DappIcon/dappIcon.js
+++ b/js/src/ui/DappIcon/dappIcon.js
@@ -41,7 +41,7 @@ export default class DappIcon extends Component {
         src={
           app.type === 'local'
             ? `${dappsUrl}/${app.id}/${app.iconUrl}`
-            : `${dappsUrl}${app.image}`
+            : `${app.image}`
         }
       />
     );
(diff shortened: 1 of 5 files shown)


js/src/ui/DappIcon/dappIcon.js
```diff
@@ -41,7 +41,7 @@ export default class DappIcon extends Component {
         src={
           app.type === 'local'
             ? `${dappsUrl}/${app.id}/${app.iconUrl}`
-            : `${dappsUrl}${app.image}`
+            : `${app.image}`
```
tomusdrw: `/api/content` is now available on `:8180` as well, so we don't need to go to `:8545`.

arkpar:
Is this compatible with all major browsers?

tomusdrw:
Yup:
screenshot from 2017-06-28 08-23-14

Anyway it's increased security for the browser that support it and it's ignored by the ones that don't. I didn't remove old security headers we used although they are now obsolete (e.g. `x-frame-options`) for the exact same reasons. The spec says that if both are present CSP will take precedence.

PR #5795, opened 2017-06-08
use the git version for the parity snap

elopio:
A new feature landed recently in snapcraft that will use information from the git repo to set the snap version.
If the snap is built from a tagged revision, the tag will be used as the version.


elopio:
This works only with annotated tags.
I've just noticed that you are not using them, so I've closed the PR.

PR #5797, opened 2017-06-08
Fix terminology distributed -> decentralized applications

5chdn:
As discussed on riot some weeks ago. The terminology introduced by [Paul Baran (1962)](https://www.rand.org/content/dam/rand/pubs/papers/2005/P2626.pdf) does not apply to DApps in 2017 which are decentralized rather than _distributed_.

This is _insubstantial_, however, I don't know how this affects translations. I noticed the dutch translation still has _gedistribueerde applicaties_, etc.


rphmeier:
Seems fine to leave the translations slightly broken, but if you like we could hold off on merging until we get updated i18n as well.

h3ll0fr13nd:
For Dutch i18n for this item, see https://github.com/paritytech/parity/pull/5802

PR #5801, opened 2017-06-08
use rust 1.18's retain to boost the purge performance

guanqun:
Therefore, we don't need to create a temporary vector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ below to build from source.
 
 ## Build dependencies
 
-**Parity requires Rust version 1.17.0 to build**
+**Parity requires Rust version 1.18.0 to build**
 
 We recommend installing Rust through [rustup](https://www.rustup.rs/). If you don't already have rustup, you can install it like this:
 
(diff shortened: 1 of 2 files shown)


rphmeier:
excellent, glad that's stabilized.

guanqun:
Why can't we merge this? :) Because we don't want to require a recent rust version?

rphmeier:
glad to merge it, there were just conflicts before.

guanqun:
Hmm. the CI is still using the old version...

```
stable-x86_64-unknown-linux-gnu (default)
rustc 1.17.0 (56124baa9 2017-04-24)
```

Probably it's just because it's still using the cached version. How can we re-trigger the CI?

guanqun:
Checking the `.gitlab-ci.yml`, seems like we're using a specific image `parity/rust:gitlab-ci`, I believe we have to upgrade this default image in order to pick up the latest rust version.

guanqun:
It passed the CI, should be ready for merge. :)

PR #5825, opened 2017-06-12
Fixed mem write for empty slice

arkpar:
Also removed side_effects field that is not used anywhere

### DIFF
--- a/ethcore/src/evm/interpreter/memory.rs
+++ b/ethcore/src/evm/interpreter/memory.rs
@@ -89,9 +89,10 @@ impl Memory for Vec<u8> {
 	}
 
 	fn write_slice(&mut self, offset: U256, slice: &[u8]) {
-		let off = offset.low_u64() as usize;
-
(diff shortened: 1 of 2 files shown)


arkpar:
Also removed `side_effects` field that is not used anywhere

ethcore/src/evm/interpreter/memory.rs
```diff
@@ -89,9 +89,10 @@ impl Memory for Vec<u8> {
 	}
 
 	fn write_slice(&mut self, offset: U256, slice: &[u8]) {
-		let off = offset.low_u64() as usize;
-
-		self[off..off+slice.len()].copy_from_slice(slice);
+		if !slice.is_empty() {
+			let off = offset.low_u64() as usize;
+			self[off..off+slice.len()].copy_from_slice(slice);
```
arkpar: Apparently `copy_from_slice` checks for bounds even if slice is empty.

arkpar:
Merged by mistake. Reverted.

PR #5830, opened 2017-06-13
PubSub for parity-js

kaikun213:
In progress. Still needs testing and documentation


js/src/api/pubsub/pubsub.js
```diff
@@ -0,0 +1,605 @@
...
+  // newPendingTransactions (callback) {
+  //   throw Error('not supported yet');
+  // }
+  //
+  // syncing (callback) {
+  //   throw Error('not supported yet');
+  // }
+
+  // parity API
+  accountsInfo (callback) {
```
tomusdrw: How about splitting this APIs into separate namespaces, like:
`api.pubsub().parity().accountsInfo()` and `api.pubsub().eth().newHeads()` ?

js/src/api/pubsub/pubsub.js
```diff
@@ -0,0 +1,605 @@
...
+        : callback(null, outNumber(data));
+    }, [inBlockNumber(blockNumber)]);
+  }
+
+  getCode (callback, address, blockNumber = 'latest') {
+    return this.addListener(this._api, 'eth_getCode', callback, [inAddress(address), inBlockNumber(blockNumber)]);
+  }
+
+  sendRawTransaction (callback, data) {
+    return this.addListener(this._api, 'eth_sendRawTransaction', callback, [inData(data)]);
```
tomusdrw: Doesn't really make sense to subscribe to that.

js/src/api/pubsub/pubsub.js
```diff
@@ -0,0 +1,605 @@
...
+  }
+
+  submitTransaction (callback, data) {
```
tomusdrw: Doesn't really make sense to subscribe to that.

js/src/api/pubsub/pubsub.js
```diff
@@ -0,0 +1,605 @@
...
+        ? callback(error)
+        : callback(null, (logs) => logs.map(outLog));
+    }, [inFilter(options)]);
+  }
+
+  getWork (callback) {
+    return this.addListener(this._api, 'eth_getWork', callback);
+  }
+
+  submitWork (callback, nonce, powHash, mixDigest) {
```
tomusdrw: Doesn't really make sense to subscribe to that.

js/src/api/pubsub/pubsub.js
```diff
@@ -0,0 +1,605 @@
...
+    return this.addListener(this._api, 'eth_submitWork', callback, [inNumber16(nonce), powHash, mixDigest]);
+  }
+
+  submitHashrate (callback, hashrate, clientId) {
```
tomusdrw: Doesn't really make sense to subscribe to that.

js/src/api/transport/ws/ws.js
```diff
@@ -81,13 +79,23 @@ export default class Ws extends JsonRpcBase {
...
-    this._ws = new WebSocket(this._url, hash);
+    // rpc secure API
+    if (this._token) {
+      const time = parseInt(new Date().getTime() / 1000, 10);
+      const sha3 = keccak_256(`${this._token}:${time}`);
+      const hash = `${sha3}_${time}`;
+
+      this._sessionHash = sha3;
+      this._ws = new WebSocket(this._url, hash);
+    // pubsub API
```
tomusdrw: Technically both secure and non-secure APIs suport pubsub, so I think the comment should rather say "non-secure API"

js/src/api/transport/ws/ws.js
```diff
@@ -194,31 +202,66 @@ export default class Ws extends JsonRpcBase {
     }, 50);
   }
 
+  _extract = (result) => {
+    // initial pubsub ACK
+    if (result.id && this._messages[result.id].subscription) {
+      this._messages[result.id].method === 'eth_subscribe'
```
tomusdrw: Assigning `this._messages[result.id]` to a variable would help in readability of the whole method.

tomusdrw: I would add `const { result, id, method } = result;` at the top as well.

js/src/api/transport/ws/ws.js
```diff
@@ -194,31 +202,66 @@ export default class Ws extends JsonRpcBase {
...
+        ? this._subscriptions['eth_subscription'][result.result] = result.id
+        : this._subscriptions['parity_subscription'][result.result] = result.id;
+      // resolve promise with messageId because subId's can collide (eth/parity)
+      this._messages[result.id].resolve(result.id);
+      // save subId for unsubscribing later
+      this._messages[result.id]['subId'] = result.result;
```
tomusdrw: why not `.subId` ?

js/src/api/transport/ws/ws.js
```diff
@@ -194,31 +202,66 @@ export default class Ws extends JsonRpcBase {
...
+      return this._messages[result.id];
+    // normal message
+    } else if (result.id) {
```
tomusdrw: Imho code like this:
```js
if (condition) {
   ...
   return something;
}

if (secondCondition) {
   return somethingElse;
}
```
is easier to read than a series of `else ifs`

js/src/api/transport/ws/ws.js
```diff
@@ -194,31 +202,66 @@ export default class Ws extends JsonRpcBase {
...
+      const { method, params, json, resolve, reject, callback } = this._extract(result);
 
       Logging.send(method, params, { json, result });
 
-      if (result.error) {
+      if (result.error || (result.params && result.params.error)) {
         this.error(event.data);
 
+        if (result.params.error) {
+          result[error] = result.params.error;
```
tomusdrw: I'm pretty sure that's not what you want (should be `result.error =` instead).

Actually I think it might be better to write it like this:
```js
result.error = (result.params && result.params.error) || result.error;
if (result.error) {
  ...
}
```

js/src/api/transport/ws/ws.js
```diff
@@ -194,31 +202,66 @@ export default class Ws extends JsonRpcBase {
...
+
         // Don't print error if request rejected or not is not yet up...
         if (!/(rejected|not yet up)/.test(result.error.message)) {
           console.error(`${method}(${JSON.stringify(params)}): ${result.error.code}: ${result.error.message}`);
         }
 
         const error = new TransportError(method, result.error.code, result.error.message);
 
-        reject(error);
+        result.id ? reject(error) : callback(error);
```
tomusdrw: Imho better to write an `if-else` instead.

js/src/api/transport/ws/ws.js
```diff
@@ -249,6 +292,38 @@ export default class Ws extends JsonRpcBase {
...
+  unsubscribe (messageId) {
+    return new Promise(() => {
+      const id = this.id;
+      const { subId, method } = this._messages[messageId];
+      const params = [subId];
+      const uMethod = method === 'eth_subscribe' ? 'eth_unsubscribe' : 'parity_unsubscribe';
+      const json = this.encode(uMethod, params);
+
+      this._messages[id] = { id, 'method': uMethod, params, json,
+        resolve: _ => {
```
tomusdrw: I would extract those methods to `const resolve = () => {...}` and then leave `this.messages[id]` as simple as in `subscribe`.

js/src/api/transport/ws/ws.js
```diff
@@ -249,6 +292,38 @@ export default class Ws extends JsonRpcBase {
...
+      const subscription = true;
+
+      this._messages[id] = { id, method, params, json, resolve, reject, callback, subscription };
+
+      this._send(id);
+    });
+  }
+
+  unsubscribe (messageId) {
+    return new Promise(() => {
```
tomusdrw: Why this promise is never resolved? I think it should be resolved in `this._messages[id]`, no?

js/src/api/transport/ws/ws.js
```diff
@@ -249,6 +292,38 @@ export default class Ws extends JsonRpcBase {
...
+      const params = [subId];
+      const uMethod = method === 'eth_subscribe' ? 'eth_unsubscribe' : 'parity_unsubscribe';
+      const json = this.encode(uMethod, params);
+
+      this._messages[id] = { id, 'method': uMethod, params, json,
+        resolve: _ => {
+          delete this._messages[messageId];
+          method === 'eth_subscribe' ? delete this._subscriptions['eth_subscription'][subId] : delete this._subscriptions['parity_subscription'][subId];
+        }, reject: e => {
+          throw Error('Error unsubscribing.' + e);
```
tomusdrw: Instead of crashing when usubscribe fails you should rather `reject` the upper promise.

js/src/api/transport/ws/ws.js
```diff
@@ -249,6 +292,38 @@ export default class Ws extends JsonRpcBase {
...
+      const id = this.id;
+      const { subId, method } = this._messages[messageId];
+      const params = [subId];
+      const uMethod = method === 'eth_subscribe' ? 'eth_unsubscribe' : 'parity_unsubscribe';
+      const json = this.encode(uMethod, params);
+
+      this._messages[id] = { id, 'method': uMethod, params, json,
+        resolve: _ => {
+          delete this._messages[messageId];
+          method === 'eth_subscribe' ? delete this._subscriptions['eth_subscription'][subId] : delete this._subscriptions['parity_subscription'][subId];
```
tomusdrw: `delete this._subscriptions[method === 'eth_subscribe' ? ...][subId];`

Also I think this check is done more often than necessary, can't we store the subscription names and unsubscribe method names in `this_messages`? I think it would simplify a lot of code.

js/src/api/pubsub/pubsub.js
kaikun213: We could, I had them split before but that makes it more difficult for the user again to keep track which subscriptionId belongs to which api for unsubscribing. 
We could split them again but keep that the Id's are non-correlating.
Also `eth`  only supports one method so far so the extra work for the user seemed unnecessary.

js/src/api/transport/ws/ws.js
kaikun213: Yes thought about it, now I realized it makes it much cleaner.

js/src/api/pubsub/pubsub.js
jacogr: Actually prefer @tomusdrw approach in this case - it also makes sure that we never overlap, and aligns better with teh way the APIs are structured?

kaikun213: Ok, `api().pubsub.eth` , `api().pubsub.parity` and `api().pubsub.net` are supported now. Subscription IDs are namespace unrelated, thus unsubscribing can be done either by calling `api().pubsub.unsubscribe(id)` or one of the modules e.g. `api().pubsub.eth.unsubscribe(id)`

js/src/api/pubsub/eth/eth.js
```diff
@@ -0,0 +1,236 @@
...
+  constructor (transport) {
+    super(transport);
+    this._api = 'parity';
+  }
+
+  newHeads (callback) {
+    return this.addListener('eth', 'newHeads', callback);
+  }
+  //
+  // logs (callback) {
```
tomusdrw: Logs are now available: #5705 but this can come in a separate PR.

Imho we should either uncomment the code or remove it.

js/src/api/rpc/trace/trace.js
```diff
@@ -28,9 +28,9 @@ export default class Trace {
       .then(outTraces);
   }
 
-  call (options, blockNumber = 'latest', whatTrace = ['trace']) {
+  call (options, whatTrace = ['trace'], blockNumber = 'latest') {
```
tomusdrw: Good catch :+1:

js/src/api/transport/ws/ws.js
tomusdrw: @kaikun213 This comment is still valid.

js/src/api/transport/ws/ws.js
tomusdrw: @kaikun213 That one is still valid too.

js/src/api/transport/ws/ws.js
```diff
@@ -307,19 +327,19 @@ export default class Ws extends JsonRpcBase {
...
 
-      this._messages[id] = { id, 'method': uMethod, params, json,
-        resolve: _ => {
-          delete this._messages[messageId];
-          method === 'eth_subscribe' ? delete this._subscriptions['eth_subscription'][subId] : delete this._subscriptions['parity_subscription'][subId];
-        }, reject: e => {
-          throw Error('Error unsubscribing.' + e);
-        } };
-
+      this._messages[id] = { id, 'method': uMethod, params, json, resolve, reject };
```
tomusdrw: `'` around `method` are not required

js/src/api/transport/ws/ws.js
kaikun213: Had some missunderstanding there. Fixxed now!

tomusdrw:
Looks good. @jacogr Could you have a look and merge if you consider it ready?

PR #5838, opened 2017-06-14
Fix wrongly called timeouts

ngotchac:
Avoid doing `catch(nextTimeout)` since some `nextTimeout` functions have the actual timeout as first argument. Calling this would pass the error as the timeout, which would make it instantaneous:

```js
console.log(Date.now());
setTimeout(() => {
  console.log(Date.now());
}, new Error('This is an error'));
```

The consequences being sending thousands of requests on continuous failure.


ngotchac:
@maciejhirsz Yes that's write. If you call `setTimeout` or `setInterval` with something else than a Number, it takes `0` as this argument

PR #5840, opened 2017-06-14
Whitelist APIs for generic Pub-Sub

tomusdrw:
Ping @kaikun213


parity/rpc_apis.rs
```diff
@@ -566,6 +570,13 @@ impl ApiSet {
 				public_list.insert(Api::Personal);
 				public_list
 			},
+			ApiSet::PubSub => vec![
```
maciejhirsz: can be just a slice or a fixed array? (Same for line 539)

maciejhirsz: Small grumble: could be just a slice or a fixed array instead of a vec? Same for line 539.

tomusdrw: You cannot consume elements from an array or slice afaict. `into_iter()` is implemented only for `&` and returns `IntoIterator<Item=&T>`, so we would need to clone the items.

rphmeier: definitely not a bottleneck in startup -- but cloning N 1-byte enum values should be really quick.

tomusdrw: Fixed.

PR #5858, opened 2017-06-16
Allow IPFS server to accept POST requests

maciejhirsz:


tomusdrw:
You should probably allow `OPTIONS` as well for CORS to work.

PR #5866, opened 2017-06-18
convert try!() to ?

guanqun:
I didn't change the `try!()` usages in `ipc/` folder, as I see from the commit log `restore try in ipc codegen` in `8125b5690c7f2d08ffab72706b14a0ef85b6c5a9`, not sure if there's anything special. However if you think it's good time to convert these, I'm willing to fix these as well.


rphmeier:
The IPC codegen currently can't handle the `?` operator, or at least it couldn't when I initially ported the codebase. The issue was that it treated each `?` as just a `)`. If it works now, feel free to change it.

guanqun:
Thank you @rphmeier  Good to know the reason. I'll probably give it a try some time later.

PR #5867, opened 2017-06-18
ArchiveDB and other small fixes

guanqun:
Add test cases, spaces to tab etc.

### DIFF
--- a/ethstore/src/dir/vault.rs
+++ b/ethstore/src/dir/vault.rs
@@ -296,6 +296,7 @@ mod test {
 		assert!(check_vault_name("vault_with_underscores"));
 		assert!(check_vault_name("vault-with-dashes"));
 		assert!(check_vault_name("vault-with-digits-123"));
+		assert!(check_vault_name("vault中文名字"));
 	}
 
 	#[test]
--- a/util/src/journaldb/archivedb.rs
+++ b/util/src/journaldb/archivedb.rs
@@ -243,6 +243,20 @@ mod tests {
 	}
 
 	#[test]
+	#[should_panic]
+	fn multiple_owed_removal_not_allowed() {
+		let mut jdb = ArchiveDB::new_temp();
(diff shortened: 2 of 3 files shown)


ethstore/src/dir/vault.rs
```diff
@@ -296,6 +296,7 @@ mod test {
 		assert!(check_vault_name("vault_with_underscores"));
 		assert!(check_vault_name("vault-with-dashes"));
 		assert!(check_vault_name("vault-with-digits-123"));
+		assert!(check_vault_name("vault中文名字"));
```
rphmeier: good idea to check that unicode works fully :)

util/src/journaldb/archivedb.rs
```diff
@@ -239,6 +239,21 @@ mod tests {
...
+	#[should_panic]
+	fn multiple_owed_removal_not_allowed() {
+		let mut jdb = ArchiveDB::new_temp();
+		let h = jdb.insert(b"foo");
+		jdb.commit_batch(0, &b"0".sha3(), None).unwrap();
+		assert!(jdb.contains(&h));
+		jdb.remove(&h);
+		jdb.remove(&h);
+		// commit_batch would call journal_under(),
+		// and we don't allow multiple owned removals.
```
rphmeier: semantics of journaldb are a little weird as it never deleted, so it actually wouldn't panic if the two removals were done on opposite sides of a `journal_under` operation...but as this test it will panic if you remove twice in a single batch.

guanqun: Yes, I deliberately remove twice in a single batch and so mark this test as `should_panic`. This basically test the assert logic, if you think  it's un-necessary, I can revert this change.

rphmeier: can't hurt to have the test case -- it will let us know if that behavior changes.

PR #5881, opened 2017-06-20
Use standard paths for Ethash cache

### LINKED ISSUE #4565: Use standard cache directory for cached files.
Cached files should use XDG cache path on linux, `~/Library/Caches` on Mac, and a separate directory on windows (`%TEMP%` maybe?). This includes:
* Ethash light DAG cache
* Fetched DApps and files

tomusdrw:
Closes #4565


ethash/src/compute.rs
```diff
@@ -94,8 +95,8 @@ pub struct Light {
 /// Light cache structure
 impl Light {
 	/// Create a new light cache for a given block number
-	pub fn new(block_number: u64) -> Light {
-		light_new(block_number)
+	pub fn new(cache_dir: &Path, block_number: u64) -> Light {
```
NikVolf: probably `new<P: AsRef<Path>> (cache_dir: P, ...)` is more public-api friendly, along with methods down the stack

ethcore/src/spec/spec.rs
```diff
@@ -185,18 +185,26 @@ fn load_from(s: ethjson::spec::Spec) -> Result<Spec, Error> {
...
+		).expect(concat!("Chain spec ", $e, " is invalid."))
 	};
 }
 
 impl Spec {
 	/// Convert engine spec into a arc'd Engine of the right underlying type.
 	/// TODO avoid this hard-coded nastiness - use dynamic-linked plugin framework instead.
-	fn engine(engine_spec: ethjson::spec::Engine, params: CommonParams, builtins: BTreeMap<Address, Builtin>) -> Arc<Engine> {
+	fn engine(
+		cache_dir: &Path,
```
NikVolf: Anything preventing to use the same generic `AsRef<Path>` here?

ethcore/src/spec/spec.rs
```diff
@@ -148,15 +148,15 @@ pub struct Spec {
 	genesis_state: PodState,
 }
 
-fn load_from(s: ethjson::spec::Spec) -> Result<Spec, Error> {
+fn load_from(cache_dir: &Path, s: ethjson::spec::Spec) -> Result<Spec, Error> {
```
NikVolf: `AsRef<Path>` here also?

ethcore/src/spec/spec.rs
tomusdrw: Those are private functions and they are only called with `&Path` as argument, don't see a point in making those generic.
Following this logic we should replace all `&str` instances with `<T: AsRef<str>>` in the entire codebase.

NikVolf: No, we should not replace `&str` instances with `<T: AsRef<str>>` because there is no such type which is used more than `String` itself while implementing `AsRef<&str>`. If it were the case, like with `&str` and `AsRef<Path>`, that could have made sense actually.

tomusdrw:
Fair point. Updated.

arkpar:
CI fails: https://gitlab.parity.io/parity/parity/-/jobs/58451

tomusdrw:
The tests should be fixed earlier, the job is from some previous commit.

@ddorgan any idea why CI is not visible on this PR?

akvadrako:
So where is ethash located now? I don't see it documented.

tomusdrw:
It's in `$BASE/cache`.

PR #5897, opened 2017-06-21
Proper light client informant and more verification of imported headers

### LINKED ISSUE #3901: Parameterize Informant with traits
For integration with light client, stuff like best block + queue info + peer info is shared. MGas/Tx per second is not relevant. Most of the coloring logic can be reused for sure.

rphmeier:
Closes #3901


ethcore/light/src/cache.rs
```diff
@@ -153,6 +153,22 @@ impl Cache {
...
+}
+
+impl HeapSizeOf for Cache {
+	fn heap_size_of_children(&self) -> usize {
+		self.headers.current_size()
+			+ self.canon_hashes.current_size()
+			+ self.bodies.current_size()
+			+ self.receipts.current_size()
+			+ self.chain_score.current_size()
+			// + corpus?
```
NikVolf: `todo:` or remove ?

ethcore/light/src/client/mod.rs
```diff
@@ -263,6 +266,30 @@ impl Client {
...
+				warn!(target: "client", "Stage 3 block verification failed for #{} ({})\nError: {:?}",
+					verified_header.number(), verified_header.hash(), e);
+				bad.push(hash);
+				continue;
+			};
+
+			// "external" verification.
+			let verify_external_result = self.engine.verify_block_external(&verified_header, None);
+			if let Err(e) = verify_external_result {
+				warn!(target: "client", "Stage 4 block verification failed for #{} ({})\nError: {:?}",
```
NikVolf: prefer one parameter per line if not all parameters can fit one line?

rphmeier: I think this is more readable as an exception, where the format string is on one line and the arguments are below it, in order. 
It makes for easy visual matching of format args to values without much vertical scanning.

The difference between it and a regular function call with many params is that you tend to look back and forth much more with a log.

parity/informant.rs
```diff
@@ -190,25 +331,24 @@ impl Informant {
...
+			.filter_map(|h| client.block(BlockId::Hash(*h)))
 			.map(|b| b.transactions_count())
 			.sum();
 
 		if ripe {
-			if let Some(block) = imported.last().and_then(|h| self.client.block(BlockId::Hash(*h))) {
+			if let Some(block) = imported.last().and_then(|h| client.block(BlockId::Hash(*h))) {
 				let header_view = block.header_view();
 				let size = block.rlp().as_raw().len();
 				let (skipped, skipped_txs) = (self.skipped.load(AtomicOrdering::Relaxed) + imported.len() - 1, self.skipped_txs.load(AtomicOrdering::Relaxed) + txs_imported);
```
NikVolf: how about checking this `self.skipped.load(AtomicOrdering::Relaxed) + imported.len() - 1` won't overflow or commenting why won't ever happen?

parity/informant.rs
```diff
@@ -68,24 +57,181 @@ impl MillisecondDuration for Duration {
...
+
+		let (importing, sync_info) = match (self.sync.as_ref(), self.net.as_ref()) {
+			(Some(sync), Some(net)) => {
+				let status = sync.status();
+				let net_config = net.network_config();
+
+				cache_sizes.insert("sync", status.mem_used);
+
+				let importing = is_major_importing(Some(status.state), queue_info.clone());
+				(importing, Some(SyncInfo {
```
NikVolf:
```rust
let info = SyncInfo {
        ...
};

(importing, Some(info))
```

might be better visually

NikVolf:
looks good, style and out-of-diff comments  can be ignored

PR #5899, opened 2017-06-22
Cancel propagated TX

CraigglesO:


js/src/ui/TxList/store.js
```diff
@@ -125,6 +126,30 @@ export default class Store {
...
+  killTransaction = (txComponent, tx) => {
+    const { hash, gasPrice, from } = tx;
+
+    this._api.parity
+      .removeTransaction(hash)
+      .then(() => {
+        return this._api.parity.postTransaction({
+          from: from,
+          to: from,                    // set to owner
+          gas: new BigNumber(21000),   // set default gas
```
CraigglesO: I really wanted to pull from `this._api.parity.minGas().then((gas) => {...})` just incase the value changes one day...
but I was always getting strange numbers. Not really sure I actually understand the use case for minGas.

PR #5904, opened 2017-06-22
Information about used paths added to general output block

grbIzl:
Following paths are printed out:
- To DB
- To keys
- To dapps


rphmeier:
It should also print this information for the light client (see `execute_light` function)

grbIzl:
CI pipeline https://gitlab.parity.io/parity/parity/pipelines/13578

PR #5908, opened 2017-06-22
Implement skeleton for transaction index and epoch transition proof PIP messages

rphmeier:
This puts the protocol in a situation where we can provide the implementations of these methods (which may be complex to verify on the client side) after 1.7 without breaking protocol compatibility with Parity-1.7 nodes.


rphmeier:
https://gitlab.parity.io/parity/parity/commit/3c562ff110ed69acc5bf8119faadbfa10d52027a/pipelines

PR #5909, opened 2017-06-22
Prioritize accounts over address book

maciejhirsz:
If a user creates an entry in the address book, and later tries to import an account that produces the same address, the account never shows up in the UI, no error is presented.

The account actually does get created, but since `parity_allAccountsInfo` returns a map, and the map is constructed such that the address book can overshadow actual accounts. This is a not-so-elegant solution that guarantees that accounts always take precedence over address book.

Also replaced `BTreeMap<String, String>` with a dedicated struct.

### DIFF
--- a/rpc/src/v1/impls/parity_accounts.rs
+++ b/rpc/src/v1/impls/parity_accounts.rs
@@ -16,7 +16,7 @@
 
 //! Account management (personal) rpc implementation
 use std::sync::Arc;
-use std::collections::BTreeMap;
+use std::collections::btree_map::{BTreeMap, Entry};
 use util::Address;
 
 use ethkey::{Brain, Generator, Secret};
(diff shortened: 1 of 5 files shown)


rpc/src/v1/impls/parity_accounts.rs
```diff
@@ -50,26 +50,33 @@ impl ParityAccountsClient {
...
+				meta: v.meta,
+				uuid: v.uuid.map(|uuid| uuid.to_string())
+			}));
+
+		let mut accounts: BTreeMap<RpcH160, ExtAccountInfo> = BTreeMap::new();
+
+		for (address, map) in account_iter {
+			match accounts.get(&address) {
+				// Skip if an account with UUID is already present.
+				Some(ref occupied) if occupied.uuid.is_some() => continue,
```
rphmeier: `Entry` API could be used here to coalesce the "get or insert" logic, but I don't know if it would be cleaner...

gavofyork:
build failing.

Issue #5907, opened 2017-06-22
Error: DoubleVote


phahulin:
Hi everyone,
we're trying to set up a PoA test network where list of validators is provided by a contract.
Initially there is a single node and then 12 more are added one by one via the contract. These nodes connect to the initial one as a bootnode, they don't connect to each other directly. So there is 12 to 1 correspondence.

We run on `Parity/v1.6.8-beta-c396229-20170608/x86_64-linux-gnu/rustc1.17.0`.
At some transaction (can't tell exactly which one, seems to be random) the following error occurs:
```
...
2017-06-22 12:39:04 UTC Verifier #0 TRACE ethcore::state_db  Cache lookup skipped for 9154f5f8c33dfa92cc49a5cdae504cd82d4b0952: no parent hash
2017-06-22 12:39:04 UTC Verifier #0 TRACE account_bloom  Check account bloom: 9154f5f8c33dfa92cc49a5cdae504cd82d4b0952
2017-06-22 12:39:04 UTC Verifier #0 TRACE state  add_balance(9154…0952, 0): 0
2017-06-22 12:39:04 UTC Verifier #0 TRACE ethcore::state_db  Cache lookup skipped for 9154f5f8c33dfa92cc49a5cdae504cd82d4b0952: no parent hash
2017-06-22 12:39:04 UTC Verifier #0 DEBUG engine  Set of validators obtained: [dd0bb0e2a1594240fed0c2f2c17c1e9ab4f87126, 9154f5f8c33dfa92cc49a5cdae504cd82d4b0952, 12a71bc07e8ae18019f343261e017d07a5217494, ad42a938bef94b804149bf35a854eb85505116df, 2b1d1b8cb055d300acf065b357be9e3e308a0daf, 9184a9e9d098b91ad5c2a64ceade8fe33f59eedb, abeeee1006e2a2a016e1cb668bbfbfacf40008ca, 1d5366bb34d4837991f7192e0f53093bb76d8c85, 1f3b85de5b011d6ff931aed076417b2642afbdc6, a8f7eaab62744dac30e636bf875ea456de097456, 9b16beff3641f0c96bc3cb1b01637baa64fd161b, 96e58e6c9a0d1ed9b52dc67eaf586143997edae0, f7eed6f9592d5c99d184af6bd9aae05046b2ca6d]
2017-06-22 12:39:04 UTC Verifier #0 TRACE engine  Multiple blocks proposed for step 299617384.
[...]

PR #5910, opened 2017-06-22
fix minor race condition in aura seal generation

rphmeier:
Might be related to #5907


rphmeier:
https://gitlab.parity.io/parity/parity/commit/9b3068479c048bfcc8ad79a91113866e68504ef5/pipelines

PR #5917, opened 2017-06-23
TransactionQueue improvements

tomusdrw:
3 important transaction queue and propagation improvements:
- [x] Instead of prioritizing transactions with low hashes prioritize the ones that were first seen.
- [x] Require a minimal gas price increase to prevent network stress.
- [x] Prevent constructing oversized packets with transactions


sync/src/chain.rs
```diff
@@ -2059,15 +2061,33 @@ impl ChainSync {
...
-							packet.append(&tx.transaction);
-							// update stats
-							let id = io.peer_session_info(peer_id).and_then(|info| info.id);
-							stats.propagated(tx.transaction.hash(), id, block_number);
+
+					let mut to_send = to_send;
+					let mut packet;
+					loop {
+						// Construct RLP
+						packet = RlpStream::new_list(to_send.len());
```
rphmeier: rather than halving the amount of propagated every time, I think we have a `RlpStream::new_list_unbounded` or similar that's used in snapshot creation. Then we can always fit as many as possible into the packet.

rphmeier:
Other than packet construction grumble, LGTM

sync/src/chain.rs
tomusdrw: Thanks! Addressed. The only issue I have is that it incurs (small) additional overhead when constructing all packages, whereas the previous approach was only more expensive when the limit is actually reached.

arkpar:
needs a merge

PR #5942, opened 2017-06-27
[WIP] Add get_with method to KeyValueDB and HashDB

Vurich:
This adds a `get_with` method to both `KeyValueDB` and `HashDB`, which takes a reference-taking closure and passes the borrowed value to the function (to avoid cloning). I did some quick benchmarking and this actually seems _slower_ than the previous code. I imagine this is because I reimplemented `get` in terms of `get_with` and the extra dynamic dispatch causes this to be slower. There are some places in the code where the cloned value is only ever used as a borrow where the `get_with` method would be an instant win, so it might be better to have both `get` and `get_with` implemented and use the latter only where necessary.

### DIFF
--- a/util/src/hashdb.rs
+++ b/util/src/hashdb.rs
@@ -25,12 +25,12 @@ pub type DBValue = ElasticArray128<u8>;
 
 // TODO: Should this include a generic version of `HashDB::insert`?
 pub trait HashDBExt {
-	fn get_with<Out, F: FnOnce(&DBValue) -> Out>(&self, _: &H256, _: F) -> Option<Out>;
+	fn get_with<Out, F: FnOnce(&[u8]) -> Out>(&self, _: &H256, _: F) -> Option<Out>;
 }
 
 macro_rules! get_with_fn_def {
 	() => {
-		fn get_with<Out, F: FnOnce(&DBValue) -> Out>(
+		fn get_with<Out, F: FnOnce(&[u8]) -> Out>(
 			&self,
 			key: &H256,
 			f: F,
@@ -39,7 +39,7 @@ macro_rules! get_with_fn_def {
 			let mut output: Option<Out> = None;
 
 			{
-				let mut wrapper = |key: &DBValue| {
+				let mut wrapper = |key: &[u8]| {
 					output = Some((o_func.take().unwrap())(key));
 				};
 
@@ -96,11 +96,11 @@ pub trait HashDB: AsHashDB + Send + Sync {
 	/// }
 	/// ```
 	fn get(&self, key: &H256) -> Option<DBValue> {
-		let mut o_func = Some(Clone::clone);
+		let mut o_func = Some(DBValue::from_slice);
 		let mut output = None;
 
 		{
-			let mut wrapper = |key: &DBValue| { output = Some((o_func.take().unwrap())(key)); };
+			let mut wrapper = |key: &[u8]| { output = Some((o_func.take().unwrap())(key)); };
 
 			self.get_exec(key, &mut wrapper);
 		}
@@ -113,7 +113,7 @@ pub trait HashDB: AsHashDB + Send + Sync {
 	/// default implementation of `get_with` will panic if `f` is called more than
 	/// once and in general there is no sensible behavior if `f` is called more
 	/// than once.
-	fn get_exec(&self, key: &H256, f: &mut FnMut(&DBValue));
+	fn get_exec(&self, key: &H256, f: &mut FnMut(&[u8]));
 
 	/// Check for the existance of a hash-key.
 	///
--- a/util/src/hashdb.rs
+++ b/util/src/hashdb.rs
@@ -30,7 +30,7 @@ pub trait HashDBExt {
 
 macro_rules! get_with_fn_def {
 	() => {
-		fn get_with<Out, F: FnOnce(&[u8]) -> Out>(
+		fn get_with<Out, F: for<'a> FnOnce(&'a [u8]) -> Out>(
 			&self,
 			key: &H256,
 			f: F,
@@ -113,7 +113,11 @@ pub trait HashDB: AsHashDB + Send + Sync {
 	/// default implementation of `get_with` will panic if `f` is called more than
 	/// once and in general there is no sensible behavior if `f` is called more
 	/// than once.
-	fn get_exec(&self, key: &H256, f: &mut FnMut(&[u8]));
+	fn get_exec<'this>(
+		&'this self,
+		key: &'this H256,
+		f: &'this mut for<'a: 'this> FnMut(&'a [u8])
+	);
 
 	/// Check for the existance of a hash-key.
 	///
--- a/util/src/hashdb.rs
+++ b/util/src/hashdb.rs
@@ -23,8 +23,10 @@ use std::collections::HashMap;
 /// `HashDB` value type.
 pub type DBValue = ElasticArray128<u8>;
 
-// TODO: Should this include a generic version of `HashDB::insert`?
+/// Supplemental trait to `HashDB` supplying generic methods for trait objects, internally
+/// calling monomorphic forms of these methods.
 pub trait HashDBExt {
+	/// Get a reference, transform it, and return the result (prevents cloning).
 	fn get_with<Out, F: FnOnce(&[u8]) -> Out>(&self, _: &H256, _: F) -> Option<Out>;
 }
 
@@ -40,6 +42,8 @@ macro_rules! get_with_fn_def {
 
 			{
 				let mut wrapper = |key: &[u8]| {
+					// Unwrap here because `get_exec` is required to call the closure no more than
+					// once.
 					output = Some((o_func.take().unwrap())(key));
 				};
 
--- a/util/src/hashdb.rs
+++ b/util/src/hashdb.rs
@@ -42,9 +42,13 @@ macro_rules! get_with_fn_def {
 
 			{
 				let mut wrapper = |key: &[u8]| {
-					// Unwrap here because `get_exec` is required to call the closure no more than
-					// once.
-					output = Some((o_func.take().unwrap())(key));
+					output = Some(
+						(
+							o_func.take().expect(
+								"The implementation of `get_exec` called its argument twice - this \
+								 is a bug!"
+							)
+						)(key));
 				};
 
 				self.get_exec(key, &mut wrapper);
--- a/util/src/hashdb.rs
+++ b/util/src/hashdb.rs
@@ -112,10 +112,10 @@ pub trait HashDB: AsHashDB + Send + Sync {
 	/// default implementation of `get_with` will panic if `f` is called more than
 	/// once and in general there is no sensible behavior if `f` is called more
 	/// than once.
-	fn get_exec<'this>(
-		&'this self,
-		key: &'this H256,
-		f: &'this mut for<'a: 'this> FnMut(&'a [u8])
+	fn get_exec(
+		&self,
+		key: &H256,
+		f: &mut FnMut(&[u8])
 	);
 
 	/// Check for the existance of a hash-key.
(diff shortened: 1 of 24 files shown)


util/src/hashdb.rs
```diff
@@ -42,7 +95,29 @@ pub trait HashDB: AsHashDB + Send + Sync {
 	///   assert_eq!(m.get(&hash).unwrap(), hello_bytes);
 	/// }
 	/// ```
-	fn get(&self, key: &H256) -> Option<DBValue>;
+	fn get(&self, key: &H256) -> Option<DBValue> {
+		let mut o_func = Some(DBValue::from_slice);
+		let mut output = None;
+
+		{
+			let mut wrapper = |key: &[u8]| { output = Some((o_func.take().unwrap())(key)); };
```
rphmeier: could be simplified just to `DBValue::from_slice(key)`

codebase style replaces unwrap with `expect("qualify why it never will panic; qed")`

util/src/hashdb.rs
```diff
@@ -15,13 +15,66 @@
...
+			&self,
+			key: &H256,
+			f: F,
+		) -> Option<Out> {
+			let mut o_func: Option<F>   = Some(f);
+			let mut output: Option<Out> = None;
+
+			{
+				let mut wrapper = |key: &[u8]| {
+					output = Some((o_func.take().unwrap())(key));
```
rphmeier: qualify `unwrap` and ensure invariant is documented in trait

util/src/journaldb/archivedb.rs
```diff
@@ -58,7 +58,16 @@ impl ArchiveDB {
...
 	fn payload(&self, key: &H256) -> Option<DBValue> {
-		self.backing.get(self.column, key).expect("Low-level database error. Some issue with your hard disk?")
+		self.backing.get(self.column, key).expect(
+			"Low-level database error. Some issue with your hard disk?"
+		)
+	}
+
+	fn payload_exec(&self, key: &H256, f: &mut FnMut(&[u8])) {
+		// TODO: Fix this once `KeyValueDB` has a `get_exec` method
+		if let Some(val) = self.payload(key) {
```
rphmeier: outdated TODO? and I think `payload` is now dead code

util/src/journaldb/overlayrecentdb.rs
```diff
@@ -430,8 +434,51 @@ impl HashDB for OverlayRecentDB {
...
+		// Possible bug in the Rust compiler.
+		let mut wrapper = |d: &[u8]| {
+			was_called.set(true);
+			f(d);
+		};
+
+		let journal_overlay = self.journal_overlay.read();
+		let short_key = to_short_key(key);
+
+		journal_overlay.backing_overlay.get_exec(&short_key, &mut wrapper);
```
rphmeier: Looks like the wrapper can be avoided if we just match on `get_ref` for `backing_overlay`

util/src/journaldb/overlayrecentdb.rs
```diff
@@ -430,8 +434,51 @@ impl HashDB for OverlayRecentDB {
...
+			}
+		}
+
+		let was_called = Cell::new(false);
+
+		// This type annotation is necessary, the inferred lifetime is too restrictive otherwise.
+		// Possible bug in the Rust compiler.
+		let mut wrapper = |d: &[u8]| {
+			was_called.set(true);
+			f(d);
```
rphmeier: something seems fishy about this. is it not soundness hole that we can call this with data borrowed from a lock acquired within the function body?

rphmeier: It seems that while it's possible to specify bounds on an HRTB when accepting one as a parameter, it's only possible to declare closures that fulfill universal HRTBs.

Vurich: Where could wrapper store the ref that would be observable? The read lock must outlive the call to `get_exec`.

util/src/journaldb/overlayrecentdb.rs
Vurich: Great point. `get_ref` doesn't exist on `MemoryDB` at the moment but it could be trivially implemented.

Vurich:
[clabot:check]

util/src/kvdb.rs
```diff
@@ -151,7 +197,29 @@ pub trait KeyValueDB: Sync + Send {
 	fn transaction(&self) -> DBTransaction { DBTransaction::new() }
 
 	/// Get a value by key.
-	fn get(&self, col: Option<u32>, key: &[u8]) -> Result<Option<DBValue>, String>;
+	fn get(&self, col: Option<u32>, key: &[u8]) -> Result<Option<DBValue>, String> {
+		let mut o_func = Some(DBValue::from_slice);
+		let mut output = None;
+
+		{
+			let mut wrapper = |key: &[u8]| { output = Some((o_func.take().unwrap())(key)); };
```
rphmeier: `DBValue::from_slice` would be simpler here

util/src/journaldb/overlayrecentdb.rs
Vurich: HRTBs are a bit of a mess right now, it'll probably be something that gets fixed as part of the Chalk work

util/src/kvdb.rs
Vurich: Yeah, this was a copy-paste thing because originally I tried to use `get_with`. Good catch.

util/src/journaldb/overlayrecentdb.rs
rphmeier: sure, but the definition of the function is `for <'a: 'this> FnMut(&'a [u8])`, so it should only be able to take references that strictly outlive the borrow of self. Data borrowed from a read lock doesn't. Soundness seems like it's only preserved here by the fact that the closure isn't allowed to make (valid) assumptions about the input data: https://is.gd/ZfJOfc

Vurich: It's outlives, not strictly outlives, right? Pretty sure the borrow from the read lock is guaranteed to live as long as `self`.

Vurich: If you try to make it fully generic over all lifetimes it fails because it borrows the `&mut FnMut`

rphmeier: these are the assumptions I am operating on:
  1. `'this` (lifetime of self borrow) can be safely coerced to a smaller lifetime, but must cover the entire function body
  2. the read guard itself may live for `'this' (why you can return a `ReadGuard<'this, T>`)
  3. borrows of `ReadGuard<'this, T>` will be strictly smaller than `'this` (why you can't return `&'a [u8]` from a `ReadGuard<'this, Vec<u8>>`)

But somehow we are taking data which is borrowed from an internal `ReadGuard` and calling a function that says "I only take data that lives at least as long as 'this" with it.

Vurich: I think it's `'a >= 'this` and `'this >= 'guard`, I don't know why borrows of some variable can't last the entire lifetime of the variable.

rphmeier: borrows of a variable can last the entire lifetime of that variable, but my understanding the lifetime of the inner `ReadGuard` value has to be less than than `'this`, because it is declared within the function body.

Vurich: Agreed, but importantly the `'this` lifetime is not the full lifetime of the borrow that calls it. It can be as short as you like as long as the inequalities are satisfied, and there's nothing that says that anything has to strictly outlive anything else.

Honestly, I trust the compiler on this one. When I tried to do a universal quantification or a parameterised concrete lifetime it failed, the only one it would accept was a bounded forall quantification. If you can write up an example of unsoundness caused by this then I'd be happy to submit it to the lang team as an unsoundness bug, but certainly as far as my (possibly faulty) intuition on lifetimes go this makes sense.

util/src/journaldb/overlayrecentdb.rs
Vurich: I actually prefer the version that has a wrapper and consistently uses `get_exec` compared to one that calls `{...}.get_ref({...}).or_else(|| {...}.get())` and then calls `get_exec` on `self.backing` if that fails. They're equivalent, performance-wise, and it's really a case of personal taste. I'm happy to change it if you feel strongly about it.

util/src/hashdb.rs
Vurich: The invariant is documented already

util/src/hashdb.rs
```diff
@@ -15,13 +15,70 @@
...
+			f: F,
+		) -> Option<Out> {
+			let mut o_func: Option<F>   = Some(f);
+			let mut output: Option<Out> = None;
+
+			{
+				let mut wrapper = |key: &[u8]| {
+					// Unwrap here because `get_exec` is required to call the closure no more than
+					// once.
+					output = Some((o_func.take().unwrap())(key));
```
rphmeier: could you move the comment into an `expect`? 
`Thread N panicked at "Called Option::unwrap on a None value", Option.rs:999`
is way worse than
`Thread N panicked at "this exact, expected invariant was violated!", Option.rs:999`.

We don't always get backtraces from our issue reports.

util/src/journaldb/overlayrecentdb.rs
rphmeier: doesn't `MemoryDB::get_exec` lead to two virtual indirections (wrapper + wrapped), while `get_ref` leads to only one?

the state trie will typically have at least the top few layers in the backing overlay.

Vurich: Damn, excellent point. I think a better solution might be to implement `get_with` for monomorphic code so that you get this speedup for free. Maybe I'll stick a TODO there, put in a performance issue and write the `get_ref` version.

Vurich: What I'll actually do is implement a `get_ref` for every `HashDB` and `KeyValueDB` that returns a `Cow`, we can extract that to a trait later. That means I can implement both `get` and `get_exec` in terms of that, which makes the former more performant (compared to delegating to `get_exec`) and the latter less error-prone (since you can't accidentally call it twice).

Vurich: I think that a nicer thing to do here (but work for a future PR) would be to have `get` return a `Cow` and then use `get_cloned` or even just `get({...}).cloned()` for when you _really_ want to clone it. Avoids hiding costs.

Vurich: I'm wondering if we can have a `HashDB` method that gets a value that returns a `Cow` so any code that wraps doesn't have to care if a `HashDB`/`KeyValueDB` has to return an owned value (because it's doing some kind of locking) or whether it can return a borrow. Then we can just pass whatever you happen to get up the stack and only pay the indirection cost from the trait objects.

rphmeier: Having that method makes sense, but a significant proportion of reads are of data held within locks

Vurich: Right, it's something that would need looking into

Vurich:
Original issue here https://github.com/paritytech/parity/issues/3993

gavofyork:
@rphmeier / @Vurich any activity here?

Vurich:
Waiting on my full sync to finish so I can profile this against master. It looks like it led to a regression in our benchmark suite so I want to see if it causes it to be faster or slower on the import. If it's slower then I'll have to do more work before a merge

NikVolf:
is it WIP? If so, there is a special label `A1-inprogress` for it
and it is mutually exclusive with `A0-pleasreview`

5chdn:
branch has conflicts and tests fail. updated labels as per latest comment.

gavofyork:
looks stale. closing.

PR #5949, opened 2017-06-28
Upgrade elastic array

Vurich:
This is a huge change, which includes some changes to replace code that originally cloned to reuse allocations instead. The updated `elastic-array` crate renames its consuming `Vec`-conversion method to `into_vec`, which means that I can do a simple `sed -i 's/to_vec/into_vec/'` and then fix the compilation errors.

This commit is probably a minor performance win and definitely a significant readability win.


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -192,7 +192,7 @@ impl LightFetch {
 			let action = req.to.map_or(Action::Create, Action::Call);
 			let gas = req.gas.unwrap_or(U256::from(10_000_000)); // better gas amount?
 			let value = req.value.unwrap_or_else(U256::zero);
-			let data = req.data.map_or_else(Vec::new, |d| d.to_vec());
+			let data = req.data.unwrap_or(Vec::new());
```
debris: `unwrap_or_default` might be a better choice

Vurich:
I need to fix the build for tests (they don't get built by default so I only saw the problems when CI tried it out), I'll fix the minor grumble too (I 100% agree with it).

Vurich:
Tests passing now, ready to merge?

PR #5951, opened 2017-06-28
Node Health warnings

tomusdrw:
- [x] /api/health endpoint (can be used for automatic monitoring of the node) returns 200OK when 3 checks are passing, 412 otherwise.
- [x] UI displays health status in always visible place.
- [x] Currently uses REST service to get latest time, will switch to use NTP instead (@andrebeat)

UI code is ready to review, core code will change a bit.
Couple of screenshots:

A warning dialog:
screenshot from 2017-06-28 17-05-37

Home Page:
screenshot from 2017-06-28 17-06-18

Status Page:
screenshot from 2017-06-28 17-05-56

Dapp:
screenshot from 2017-06-29 08-42-53

If everything is ok:
screenshot from 2017-06-28 17-08-26


dapps/src/api/response.rs
```diff
@@ -16,17 +16,25 @@
...
 }
 
-pub fn as_json_error<T: Serialize>(val: &T) -> Box<Handler> {
+pub fn as_json<T: Serialize>(status: StatusCode, val: &T) -> ContentHandler {
+	let json = serde_json::to_string(val)
+		.expect("serialization to string is infallible; qed");
+	ContentHandler::new(status, json, mime!(Application/Json))
+}
+
+pub fn as_json_error<T: Serialize>(status: StatusCode, val: &T) -> ContentHandler {
```
andrebeat: I think you can remove this function.

js/src/ui/StatusIndicator/statusIndicator.js
```diff
@@ -0,0 +1,67 @@
...
+
+const statuses = ['bad', 'needsAttention', 'ok'];
+
+export default class StatusIndicator extends Component {
+  static propTypes = {
+    type: PropTypes.oneOf(['radial', 'signal']),
+    id: PropTypes.string.isRequired,
+    status: PropTypes.oneOf(statuses).isRequired,
+    title: PropTypes.arrayOf(PropTypes.node),
+    tooltipPlacement: PropType.oneOf(['left', 'top', 'bottom', 'right'])
```
andrebeat: Should be `PropTypes.oneOf`.

dapps/src/api/response.rs
tomusdrw: It's used to return 404 error in API.

js/src/ui/StatusIndicator/statusIndicator.js
tomusdrw: Good catch! Thanks. Not sure why build didn't pick it up.
@ddorgan Seems that CI is not reporting back to github again and JS build was skipped even though this PR is altering JS files.

gavofyork:
you can remove the peers from the bottom right and place the status indicator in it place. (peers was only added there as a makeshift status indication.)

possibly for another PR, it would be good to move a lot of the cruft from the version string and just place the version (e.g. "version 1.7.0"). also removing the update status altogether if it's unknown would be good.

PR #5956, opened 2017-06-30
NTP time sync

andrebeat:
Updates the `TimeChecker` from `ui-warnings` to use an NTP client instead of using an HTTP time API.


dapps/src/api/time.rs
```diff
@@ -56,27 +64,72 @@ impl From<io::Error> for Error {
...
+	fn now(&self) -> Timespec;
 }
 /// Default system time provider.
 #[derive(Clone)]
 pub struct StdTimeProvider;
 impl TimeProvider for StdTimeProvider {
 	fn new() -> Self where Self: Sized { StdTimeProvider }
-	fn utc_timestamp_millis(&self) -> i64 {
-		let time = ::time::now_utc().to_timespec();
+	fn now(&self) -> Timespec {
```
tomusdrw: We can inline this function since we  don't really test internals of `SimpleNtp`.

dapps/src/api/time.rs
```diff
@@ -169,40 +192,39 @@ impl<F: Fetch, T: TimeProvider> TimeChecker<F, T> {
...
-			} else {
-				let data = ::std::io::Cursor::new(b"1".to_vec());
-				futures::future::ok(fetch::Response::from_reader(data))
-			}
+	struct FakeNtp(RefCell<Vec<Duration>>, Arc<Mutex<u64>>);
+	impl FakeNtp {
+		fn new() -> FakeNtp {
+			FakeNtp(
+				RefCell::new(vec![Duration::milliseconds(150)]),
+				Arc::new(Mutex::new(0)))
```
tomusdrw: A `Cell` would do here.

dapps/src/api/time.rs
```diff
@@ -169,40 +192,39 @@ impl<F: Fetch, T: TimeProvider> TimeChecker<F, T> {
...
-	impl TimeProvider for FakeTime {
-		fn new() -> Self where Self: Sized { FakeTime(RefCell::new(vec![150, 0])) }
-		fn utc_timestamp_millis(&self) -> i64 {
-			self.0.borrow_mut().pop().expect("Expecting only two calls to utc_timestamp_millis.")
+
+	impl Ntp for FakeNtp {
+		fn drift(&self) -> BoxFuture<Duration, Error> {
+			let mut val = self.1.lock();
+			*val = *val + 1;
+			futures::future::ok(self.0.borrow_mut().pop().expect("Expecting only one call to now().")).boxed()
```
tomusdrw: since `self.0` can be modified outside I would rather write something like: "Unexpected call to `drift()` instead.

js/src/ui/StatusIndicator/statusIndicator.js
```diff
@@ -37,6 +37,11 @@ export default class StatusIndicator extends Component {
 
   render () {
     const { id, status, title, type, tooltipPlacement } = this.props;
+    const tooltip = title.find(x => !x.isEmpty) ? (
```
tomusdrw: you can just supply this to `data-tip`, if it's false the tooltip won't be shown.
I tried to check if `title.length != 0` but it seems that we sometimes put empty strings there?

Also `isEmpty` is kind of fishy, I'm pretty sure it will result `undefined` for any possible builtin data type in JS.

dapps/src/api/time.rs
andrebeat: Agreed, I'll remove the `TimeProvider` trait.

js/src/ui/StatusIndicator/statusIndicator.js
andrebeat: Oh, I didn't notice the `data-tip`. Yes, in the case of health success we return an empty message. Alternatively, we could update the `HealthInfo` message to be an `Option`.

Yes, I agree it's probably not a good idea to use `isEmpty`. Maybe we can use lodash's `_.isEmpty`, or something like that which handles all the corner cases.

dapps/src/api/time.rs
```diff
@@ -56,27 +64,72 @@ impl From<io::Error> for Error {
...
+	fn drift(&self) -> BoxFuture<Duration, Error>;
+}
 
-		1_000 * time.sec + time.nsec as i64 / 1_000_000
+/// NTP client using the SNTP algorithm for calculating drift.
+#[derive(Clone)]
+pub struct SimpleNtp<T> {
+	address: Arc<String>,
+	time_provider: T,
+	pool: CpuPool,
```
tomusdrw: One more grumble. Could we share this `CpuPool` with `fetch`? Otherwise we'll spawn 8 threads and most of them will do nothing 99% of the time.

util/fetch/src/client.rs
```diff
@@ -126,6 +126,10 @@ impl Client {
 		*self.client.write() = (time::Instant::now(), client.clone());
 		Ok(client)
 	}
+
+	pub fn pool(&self) -> CpuPool {
```
andrebeat: If you prefer I could change it to create the pool outside `Fetch` and pass it to its constructor (and `Ntp`'s).

js/src/ui/StatusIndicator/statusIndicator.js
andrebeat: I could not get the tooltip to disable, even by setting the `data-tip` explicitly to `false` or `''`.

andrebeat:
I think I have addressed all your grumbles (for Rust code).

tomusdrw:
Perfect! Thanks a lot!

Issue #5429, opened 2017-04-07
Have `Handler::on_connect` return a flag indicating whether the peer is kept


rphmeier:
Could be an enum or a simple bool.
If no `Handlers` keep the peer, `LightProtocol` should disconnect them.

Ensure that at least one of the full node handlers says to keep peers.

@adrianbrink Maybe one you'd be interested in?

adrianbrink:
@rphmeier I'll take this one too :-)

rphmeier:
Thanks!

Vurich:
This can be closed now

PR #5958, opened 2017-06-30
Report whether a peer was kept from `Handler::on_connect`

Vurich:
This looked like a quick feature that was still up for grabs. I returned a custom enum instead of a bool and used `BitOr::bitor` instead of an inherent `or` (or similarly-named) method, both of those are decisions ripe for bikeshedding. I also have no idea if the message I chose for `Error::display` is remotely meaningful.

### DIFF
--- a/ethcore/light/src/net/mod.rs
+++ b/ethcore/light/src/net/mod.rs
@@ -821,12 +821,15 @@ impl LightProtocol {
 		}));
 
 		let any_kept = self.handlers.iter().map(
-			|handler|
-			handler.on_connect(&Ctx {
-				peer: *peer,
-				io: io,
-				proto: self,
-			}, &status, &capabilities)
+			|handler| handler.on_connect(
+				&Ctx {
+					peer: *peer,
+					io: io,
+					proto: self,
+				},
+				&status,
+				&capabilities
+			)
 		).fold(PeerStatus::Kept, PeerStatus::bitor);
 
 		if any_kept == PeerStatus::Unkept {
--- a/ethcore/light/src/net/mod.rs
+++ b/ethcore/light/src/net/mod.rs
@@ -31,6 +31,7 @@ use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
+use std::ops::{BitOr, BitAnd, Not};
 
 use provider::Provider;
 use request::{Request, NetworkRequests as Requests, Response};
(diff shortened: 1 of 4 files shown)


ethcore/light/src/net/mod.rs
```diff
@@ -766,15 +820,20 @@ impl LightProtocol {
 			awaiting_acknowledge: None,
 		}));
 
-		for handler in &self.handlers {
+		let any_kept = self.handlers.iter().map(
+			|handler|
```
rphmeier: style: could you indent the body of the closure as it's on multiple lines?

ethcore/light/src/on_demand/mod.rs
```diff
@@ -29,7 +29,17 @@ use futures::sync::oneshot::{self, Sender, Receiver, Canceled};
 use network::PeerId;
 use util::{RwLock, Mutex};
 
-use net::{self, Handler, Status, Capabilities, Announcement, EventContext, BasicContext, ReqId};
+use net::{
```
rphmeier: better to keep it on a few, indented lines as here: https://github.com/paritytech/parity/pull/5958/files#diff-e8c26f6188b2b873a7467abd3041b843R42

sync/src/light_sync/mod.rs
```diff
@@ -229,26 +230,34 @@ pub struct LightSync<L: AsLightClient> {
...
-		let chain_info = ChainInfo {
-			head_td: status.head_td,
-			head_hash: status.head_hash,
-			head_num: status.head_num,
-		};
 
-		{
-			let mut best = self.best_seen.lock();
-			*best = ::std::cmp::max(best.clone(), Some(chain_info.clone()));
+			PeerStatus::Unkept
```
rphmeier: it should not disconnect the peer before returning `Unkept`

ethcore/light/src/net/mod.rs
```diff
@@ -766,15 +820,20 @@ impl LightProtocol {
...
 			handler.on_connect(&Ctx {
 				peer: *peer,
 				io: io,
 				proto: self,
 			}, &status, &capabilities)
-		}
+		).fold(PeerStatus::Kept, PeerStatus::bitor);
```
rphmeier: not sure if this is intentional, but it does correctly handle the caveat where having no attached handlers should keep the peer (as light servers don't have any handlers but shouldn't disconnect incoming connections).

rphmeier:
closes #5429

ethcore/light/src/net/mod.rs
Vurich: That was intentional, but it might be better to add a comment. I meant to ask about it.

sync/src/light_sync/mod.rs
Vurich: Oh right, so at the moment any handler missing those capabilities will disconnect the peer?

ethcore/light/src/net/mod.rs
Vurich: RFC style is to have it start on the same line as the arguments if it's a single expression, I'll change it to that. I'm OK with indenting it if it's got braces but there's no way that I can set up my editor to indent it after arguments without braces so I have to manually reindent any time I change it.

ethcore/light/src/net/mod.rs
```diff
@@ -157,6 +158,54 @@ pub struct Peer {
 	awaiting_acknowledge: Option<(SteadyTime, Arc<FlowParams>)>,
 }
 
+/// Whether or not a peer was kept by a handler
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PeerStatus {
```
NikVolf: how is this enum better than just `bool`?

NikVolf: oh it's merged already, still it's overengineering to me.. :)

Vurich: Because by reading the signature there is no way to know if `true` means kept or dropped. If it was called `is_dropped` I would return a `bool`. I don't know about you, but I hit "what does this bool _even mean_" all the time so I try to make sure my code doesn't fall into the same trap.

NikVolf: agree, true for kept (positive result of operation) and false for dropped (negative result) is pretty natural to assume though

Vurich: Even if you can assume it, both would be reasonable choices and there's no way to know your assumption is correct without reading the code. If I can put all the information you need in the type signature then I will. We could use `u8` instead of `bool` and instead of type-safe enums like in C, but we don't because they are semantically different. Likewise here.

PR #5975, opened 2017-07-03
Fix underflow in gas calculation

Vurich:
This doesn't affect the value of the calculation itself since it's lower-clamped to an unsigned value anyway. It simply prevents an underflow in the intermediate values - although the input and output must be strictly positive there is nothing that prevents intermediate values from being negative. This only panicked before, it didn't cause incorrect behaviour.

### DIFF
--- a/ethcore/src/ethereum/ethash.rs
+++ b/ethcore/src/ethereum/ethash.rs
@@ -453,11 +453,14 @@ impl Ethash {
 		let parent_has_uncles = parent.uncles_hash() != &sha3::SHA3_EMPTY_LIST_RLP;
 
 		let min_difficulty = self.ethash_params.minimum_difficulty;
+
 		let difficulty_hardfork = header.number() >= self.ethash_params.difficulty_hardfork_transition;
-		let difficulty_bound_divisor = match difficulty_hardfork {
-			true => self.ethash_params.difficulty_hardfork_bound_divisor,
-			false => self.ethash_params.difficulty_bound_divisor,
+		let difficulty_bound_divisor = if difficulty_hardfork {
+			self.ethash_params.difficulty_hardfork_bound_divisor
+		} else {
+			self.ethash_params.difficulty_bound_divisor
 		};
+
 		let duration_limit = self.ethash_params.duration_limit;
 		let frontier_limit = self.ethash_params.homestead_transition;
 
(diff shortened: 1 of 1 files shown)


ethcore/src/ethereum/ethash.rs
```diff
@@ -453,11 +453,14 @@ impl Ethash {
...
 
 		let min_difficulty = self.ethash_params.minimum_difficulty;
+
 		let difficulty_hardfork = header.number() >= self.ethash_params.difficulty_hardfork_transition;
-		let difficulty_bound_divisor = match difficulty_hardfork {
-			true => self.ethash_params.difficulty_hardfork_bound_divisor,
-			false => self.ethash_params.difficulty_bound_divisor,
+		let difficulty_bound_divisor = if difficulty_hardfork {
+			self.ethash_params.difficulty_hardfork_bound_divisor
+		} else {
```
rphmeier: codebase style is to use match here (and honestly it's more compact)

Vurich: Just when it's an expression? We don't do it when we use if as a statement.

Sure, I'll change it here though.

Vurich: A quick grep through the codebase shows only 2 or 3 uses of `= match ... { true => ..., false => ... }` and over a hundred uses of `= if` (I might have missed a couple because I have to use `-A 1` and eyeball it for the `match` case)

gavofyork: `true =>` 94 cases, `false =>` 95 cases. some of those will be the same `= if` fragment, but still, it's a lot more than 2 or 3. general convention is to use the `match` syntax since it's more compact. it's not a desperately hard rule, but generally preferable unless there are mitigating circumstances.

Vurich: I totally get your point, and I'm OK conforming to any style that you want, but I don't understand how I could have predicted what the style should be and even if I eventually learn it we're just going to have this same issue when someone else joins. It's a waste of my time and it's a waste of yours. Evangelising for rustfmt is a waste of both our time too but it's a waste of time that happens once instead of every time a new contributor starts submitting PRs.

```
➤ # boolean match-as-expression
➤ rg '= match ' -A 1 -t rust | rg '(true|false) => ' | wc -l
21
➤ # boolean match-as-statement
➤ rg '^\s*match ' -A 1 -t rust | cat | rg '(true|false) =>'  | wc -l
51
➤ # multi-line if-as-expression
➤ rg '= if .*\{$' | wc -l
73
➤ # single-line if-as-expression
➤ rg '= if .*\{.*\}' | wc -l
37
➤ # multi-line boolean match
➤ # (doesn't include where the 2nd branch is multi-line but the 1st is not)
➤ rg 'match ' -A 1 -t rust | cat | rg '(true|false) => \{$' | wc -l
11
```

Again, I'm OK with inconsistency too, but if we're going to have an inconsistent style then there's no point discussing style at all on PRs unless it's utterly egregious.

PR #5978, opened 2017-07-03
Generate transition proof on uncommitted state

rphmeier:
Sometimes it wasn't possible to generate proofs on state committed immediately prior. This works around that and reduces the amount of DB commits necessary.

Also does a little bit of cleanup in `block.rs`, since I tried to make these changes there earlier.


keorn:
Close in favour of #5991

PR #5991, opened 2017-07-04
New contract PoA sync fixes

keorn:
Generate proof on uncommitted state and dont report skipped proposer if not signer.
Also prevents re-sealing on same step when validator set changes.


rphmeier:
@gavofyork I'm probably not the best person to review this impartially as Peter and I made these fixes together

ethcore/src/engines/validator_set/contract.rs
```diff
@@ -107,14 +107,14 @@ impl ValidatorSet for ValidatorContract {
 	fn report_malicious(&self, address: &Address, _set_block: BlockNumber, block: BlockNumber, proof: Bytes) {
 		match self.provider.report_malicious(&*self.transact(), *address, block.into(), proof).wait() {
 			Ok(_) => warn!(target: "engine", "Reported malicious validator {}", address),
-			Err(s) => warn!(target: "engine", "Validator {} could not be reported {}", address, s),
+			Err(_) => {} // warn!(target: "engine", "Validator {} could not be reported {}", address, s),
```
arkpar: Reason for warning removal?

rphmeier: good catch. it was removed before because the checks weren't in place and it made testing very spammy

gavofyork:
@rphmeier doesn't build

rphmeier:
Added a few more commits to address two issues:
  1. ensure that pending transition proofs are always flushed, since finding the correct epoch transition proof can make use of a database iterator, which is only available on flushed data
  2. ensure that the amount of pending transition proof lookups are minimized by caching them between blocks. This actually also fixes the symptom that lead me to find the first issue, so that commit may not be necessary.

I would suggest keeping the additional flush for now since it's very rare, but then we can adjust `kvdb::Database` to support ordered iteration over the backing and buffered storage. See #6048

PR #5992, opened 2017-07-04
Add missing CSP for web3.site

tomusdrw:
cc @kaikun213


gavofyork:
will need resolving and @kaikun213 please take a look.

kaikun213:
Tested it and it works fine and from my small rust knowledge the code looks good. Only request left is that we enable port 3000 too e.g. when parity is run with `--no-ui-validation` flag. (Disable CSP)

gavofyork:
ok - will leave that to @tomusdrw then

tomusdrw:
@kaikun213 updated!

Issue #5988, opened 2017-07-04
Migrate to serde 1.0


debris:
- [x] migrate crates which utilize serde `0.9` to serde `1.0`
- [x] migrate executables which use docopt `0.7` to docopt `0.8`
- [x] migrate toml from `0.2` to `0.4`
- [x] migrate code that uses `rustc_serialize::json::Json`
  - [x] `ethcore/src/miner/price_info.rs`
  - [x] `util/network/src/node_table.rs`

0x7CFE:
What is the status for this issue? I think I could handle it, if it's not taken.

debris:
I believe we still use `rustc_serialize` in several places

```
ethcore/benches/evm.rs:24:extern crate rustc_serialize;
ethcore/benches/evm.rs:78:      use rustc_serialize::hex::FromHex;
util/network/src/lib.rs:72:extern crate rustc_serialize;
util/network/src/node_table.rs:35:pub use rustc_serialize::json::Json;
```

0x7CFE:
What about `ethcore/src/miner/price_info.rs`? I couldn't see this path on `master`.

debris:
it was relocated to `price-info/src/lib.rs` and it's already updated and uses serde

0x7CFE:
Looks like we've finished all of the related work. Should this issue be closed for now?

debris:
yes, definitely!

PR #5996, opened 2017-07-04
migration to serde 1.0

debris:
Partly addresses #5988


tomusdrw:
@paritytech/ci Where are the tests?

rphmeier:
#6009 still uses serde 0.9 but upgrading should be pretty trivial. This should be merged first and then I'll port the whisper RPC over.

PR #6000, opened 2017-07-05
Fix first run wizard.

### LINKED ISSUE #5999: New account wizard on fresh install is broken.
You cannot skip it and you cannot go back in case you didn't back up the phrase.

tomusdrw:
Closes #5999


jacogr:
I actually stumbled across this yesterday in my branch as well (at least the +2 vs +3 issue, was going to PR today). LGTM.

PR #6003, opened 2017-07-05
Refactor price_info

### LINKED ISSUE #4205: Use `reqwest/FetchService` in `price_info.rs`
Currently it's using Hyper and spawns a new thread for each call:
https://github.com/ethcore/parity/blob/8517369372f2e151dedf5d6f1b6c32bd372453da/ethcore/src/miner/price_info.rs#L67-L81

It would be better if `Client` could be re-used and `Handler` would be rewritten using either `reqwest` or our global `FetchService`.

andrebeat:
This PR updates `price_info` to use `Fetch` instead of using hyper directly.
Additionally, the `Fetch` client is now cached to avoid creation on each
request.

I've tried to minimize the impact of these changes on the rest of the `ethcore`
module. The `price_info::Client` uses the `fetch::Client` directly instead of
being parameterized on `F: Fetch`, this is to avoid proliferating this type
parameter through the rest of `ethcore`. The `fetch::Client` is also created by
`price_info::Client` since it's not particularly easy to inject this dependency.
Currently this spawns 4 threads which will be mostly idle. I could add a
constructor to `fetch::Client` that allows defining the number of threads used
in the `CpuPool`.

Fixes #4205.


rphmeier:
I think @axelchalon was also working on this (you guys may want to comment on issues you are starting work on to avoid running into each other in the future)

andrebeat:
@rphmeier You're right, I'm sorry 😕 I should definitely have done that, especially considering that also happened on my last PR. @axelchalon let me know what's your progress on this and if necessary I'll close this PR.

ethcore/src/miner/price_info.rs
```diff
@@ -15,94 +15,95 @@
...
-			let url = FromStr::from_str("http://api.etherscan.io/api?module=stats&action=ethprice")
-				.expect("string known to be a valid URL; qed");
-
-			if let Err(e) = Self::request(url, set_price) {
-				warn!("Failed to auto-update latest ETH price: {:?}", e);
-			}
-		});
+impl fmt::Debug for Client {
+	fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+		write!(fmt, "Client {{ api_endpoint: {} }}", self.api_endpoint)
```
rphmeier: maybe have the debug info indicate it's a price fetch client? We have another `Client` which is already very prolific throughout the codebase.

ethcore/src/miner/price_info.rs
```diff
@@ -15,94 +15,95 @@
...
+	}
+}
 
-		// Wait for exit
-		let _ = rx.recv().map_err(|e| format!("Request interrupted: {:?}", e))?;
-		client.close();
+impl Client {
+	pub fn new() -> Client {
+		let api_endpoint = "http://api.etherscan.io/api?module=stats&action=ethprice".to_owned();
+		let fetch = FetchClient::new().expect("Error starting fetch client");
```
rphmeier: it should return an error upon failure to start rather than aborting the whole program

andrebeat: I agree, and the only reason I did this was to minimize the API changes in `ethcore`. Specifically, the `GasPricer` (which holds the `price_info`) is created from a `GasPricerConfig` using `Into` which assumes no failures.

debris:
Imo, if we are refactoring `price_info` it would be good to split it from `ethcore` to it's own crate.

axelchalon:
@andrebeat I hadn't started working on it, no worries! I'll make sure to comment on the issues more in the future!

ethcore/src/miner/price_info.rs
```diff
@@ -15,94 +15,95 @@
...
-	channel: mpsc::Sender<()>,
+/// Price info error.
+#[derive(Debug, Clone)]
+pub enum Error {
+	/// The API returned an unexpected status code or content.
+	UnexpectedResponse(&'static str, String),
+	/// There was an error when trying to reach the API.
+	Fetch(String),
+	/// IO error when reading API response.
+	Io(String),
```
tomusdrw: Why doesn't the errors just wrap `fetch::Error` and `io::Error`?

ethcore/src/miner/price_info.rs
```diff
@@ -15,94 +15,95 @@
...
-			let url = FromStr::from_str("http://api.etherscan.io/api?module=stats&action=ethprice")
-				.expect("string known to be a valid URL; qed");
-
-			if let Err(e) = Self::request(url, set_price) {
-				warn!("Failed to auto-update latest ETH price: {:?}", e);
-			}
-		});
+impl fmt::Debug for Client {
+	fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+		write!(fmt, "price_info::Client {{ api_endpoint: {} }}", self.api_endpoint)
```
tomusdrw: `fmt.debug_struct` instead?

ethcore/src/miner/price_info.rs
tomusdrw: Actually it would be best if the `FetchClient` was shared with other services.

andrebeat:
@debris should this be moved to a root crate or should I move it to `util`? (or some other place).

andrebeat:
@tomusdrw @rphmeier I'll refactor this to inject the `FetchClient` dependency.

debris:
@andrebeat root directory is ok

andrebeat:
I think I have addressed all your comments. I also changed `price_info` to use `serde_json` instead of `rustc_serialize`.

parity/params.rs
```diff
@@ -226,15 +227,17 @@ impl Default for GasPricerConfig {
...
-				GasPricer::new_calibrated(GasPriceCalibratorOptions {
-					usd_per_tx: usd_per_tx,
-					recalibration_period: recalibration_period,
-				})
+				GasPricer::new_calibrated(
+					GasPriceCalibratorOptions {
+						usd_per_tx: usd_per_tx,
+						recalibration_period: recalibration_period,
+					},
+					fetch)
```
tomusdrw: nitpick: imho it looks bettter like this:
```rust
  ...
  },
  fetch,
)
...
```

ethcore/Cargo.toml
```diff
@@ -32,6 +32,7 @@ ethjson = { path = "../json" }
 ethkey = { path = "../ethkey" }
 ethstore = { path = "../ethstore" }
 evmjit = { path = "../evmjit", optional = true }
+fetch = { path = "../util/fetch" }
```
tomusdrw: To reduce the number of dependencies IMHO we could re-export fetch from `price-info`.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ ethcore-light = { path = "ethcore/light" }
 ethcore-logger = { path = "logger" }
 ethcore-stratum = { path = "stratum" }
 ethkey = { path = "ethkey" }
+fetch = { path = "util/fetch" }
```
tomusdrw: Main crate is already able to instantiate `FetchClient` using a re-export from `hash-fetch` crate.

price-info/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+#![warn(missing_docs)]
+#![cfg_attr(feature="benches", feature(test))]
+#![cfg_attr(feature="dev", feature(plugin))]
```
tomusdrw: No features are not listed in the `Cargo.toml` so this seems to be redundant.

price-info/src/price_info.rs
```diff
@@ -0,0 +1,145 @@
...
+
+	use self::ethcore_logger::init_log;
+	use self::util::{Condvar, Mutex};
+	use std::sync::Arc;
+	use std::time::Duration;
+	use fetch::{Client as FetchClient, Fetch};
+	use price_info::{Client, PriceInfo};
+
+	#[test] #[ignore]
+	fn should_get_price_info() {
```
tomusdrw: Actually if `Client` would be generic over `Fetch` you would be able to write proper tests for this service.

andrebeat: Sure, I'll update it 👍

andrebeat:
@tomusdrw I've fixed the re-exports. I also made the `Client` generic over `Fetch` and added some tests. I added a default value to the `Fetch` type parameter to avoid "polluting" the `ethcore` API.

util/fetch/src/client.rs
```diff
@@ -61,6 +61,14 @@ pub trait Fetch: Clone + Send + Sync + 'static {
 		f.boxed()
 	}
 
+	/// Spawn the future in context of this `Fetch` thread pool as "fire and forget", i.e. dropping this future without
+	/// canceling the underlying future.
+	/// Implementation is optional.
+	fn forget<F, I, E>(&self, _: F) where
```
rphmeier: why is implementation optional?

gavofyork: @andrebeat ?

andrebeat: @rphmeier You're right, it should not be optional. I think the same may be true for the `process` method.

PR #6009, opened 2017-07-06
Initial Whisper implementation

### LINKED ISSUE #4685: Whisper Support
Support the whisper subprotocol and associated RPCs

rphmeier:
Closes #4685

Summary of changes:
  - `whisper::net` module contains the network handler, which maintains a pool of messages up to a given size. Messages with a higher PoW will push out those with lower.
  - We respect other peers' topic bloom filters, but do not issue them ourselves (TODO)
  - The RPC layer expects a common payload format, as well as three kinds of encryption:
    1. Asymmetric: ECIES
    2. Symmetric AES-GCM with multi-use key, random nonce appended to ciphertext
    3. Broadcast: Symmetric AES-GCM with single use key and global nonce, with key XOR'd into topic hashes. "salted topics" prepended to ciphertext.
  - The payload format consists of the message body, arbitrary padding, and an optional signature.
  - There are RPCs for creating ephemeral identities, either asymmetric (Secp256k1) or symmetric (AES-256-GCM).
   - Enable Whisper with the `--whisper` flag, "shh" RPC API, and choose target memory (MB) with `--whisper-pool-size`. This could be made dynamic in the future.
   - Subscribe or poll filters for messages with certain topics, signatures, and encryption/broadcast  mode


rphmeier:
I'm OK with adding a scheme for giving out topic bloom filters to peers, but I'd like the "darkness" to be configurable.

The rough idea is to take the list of each bloom relevant to a local subscriptions and give each of N peers `1 / N` of them xor'd together, with a few random bits set. The hard part is making sure we always have coverage of all the subscriptions we care about even in the face of a changing or unreliable peer set.

rphmeier:
(I'll ice this to post 1.7.0)

ethkey/src/random.rs
```diff
@@ -15,18 +15,30 @@
...
+		}
+	}
+}
+
+impl<'a> Generator for &'a mut OsRng {
+	type Error = ::Void;
+
+	fn generate(self) -> Result<KeyPair, Self::Error> {
+		let (sec, publ) = SECP256K1.generate_keypair(self)
+			.expect("context always created with full capabilities; qed");
```
arkpar: Is invalid capabilities the only possible source of failure here? Why not just pass the error as it was before the change?

whisper/src/message.rs
```diff
@@ -0,0 +1,476 @@
...
+		(leading_zeros * 8) + hash.get(leading_zeros + 1).map_or(0, |b| b.leading_zeros() as usize)
+	};
+	let spacetime = size as f64 * ttl as f64;
+
+	(1u64 << leading_zeros) as f64 / spacetime
+}
+
+/// A topic of a message.
+#[derive(Debug, Default, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Topic(pub [u8; 4]);
```
arkpar: Why not use `H32`? It supports RLP serialization and bloom filters already.

whisper/src/message.rs
```diff
@@ -0,0 +1,476 @@
...
+
+impl Message {
+	/// Create a message from creation parameters.
+	/// Panics if TTL is 0.
+	pub fn create(params: CreateParams) -> Self {
+		use byteorder::{BigEndian, ByteOrder};
+		use rand::{Rng, SeedableRng, XorShiftRng};
+
+		let mut rng = {
+			let mut thread_rng = ::rand::thread_rng();
```
arkpar: Does it not require a crypto-secure rng?

whisper/src/message.rs
```diff
@@ -0,0 +1,476 @@
...
+	// does checks for validity.
+	fn from_components(envelope: Envelope, size: usize, hash: H256, now: SystemTime)
+		-> Result<Self, Error>
+	{
+		const LEEWAY_SECONDS: u64 = 2;
+
+		if envelope.expiry <= envelope.ttl { return Err(Error::LivesTooLong) }
+		if envelope.ttl == 0 { return Err(Error::ZeroTTL) }
+
+		let issue_time_adjusted = Duration::from_secs(envelope.expiry - envelope.ttl - LEEWAY_SECONDS);
```
arkpar: can this underflow?

whisper/src/net.rs
```diff
@@ -0,0 +1,638 @@
...
+	// reserve space for additional elements.
+	fn reserve(&mut self, additional: usize) {
+		self.slab.reserve_exact(additional);
+		self.sorted.reserve(additional);
+		self.known.reserve(additional);
+	}
+
+	// whether a message is known or within the bounds of PoW.
+	fn may_accept(&self, message: &Message) -> bool {
+		!self.known.contains(message.hash()) && {
```
arkpar: Comment above says that a known message should return `true`

whisper/src/message.rs
rphmeier: The bloom algorithm is slightly different, I think. It's not too much overhead to have a separate type, and it conveys the usage better IMO

whisper/src/message.rs
rphmeier: it's only used to find a nonce which brings the PoW below the threshold. Speed here would be the critical factor. From an outside observer's perspective it doesn't matter whether the method used to choose the nonce is high or low entropy. It could even be consecutive

whisper/src/net.rs
```diff
@@ -0,0 +1,638 @@
...
+	fn partial_cmp(&self, other: &SortedEntry) -> Option<Ordering> {
+		Some(self.cmp(other))
+	}
+}
+
+// stores messages by two metrics: expiry and PoW rating
+// when full, will accept messages above the minimum stored.
+struct Messages {
+	slab: ::slab::Slab<Message>,
+	sorted: Vec<SortedEntry>,
```
arkpar: perhaps ` std::collections::BinaryHeap` would fit better?

whisper/src/message.rs
rphmeier: at least the subtraction of leeway can. I'll make that saturating.

whisper/src/net.rs
```diff
@@ -0,0 +1,638 @@
...
+				None => {
+					debug!(target: "whisper", "Received message from unknown peer.");
+					return Err(Error::UnknownPeer(*peer));
+				}
+			};
+
+			let mut peer = peer.lock();
+
+			if !peer.can_send_messages() {
+				return Err(Error::UnexpectedMessage);
```
arkpar: Can connection recover after this? Maybe it would be better to disconnect on any such error?

whisper/src/net.rs
rphmeier: a binary heap will be optimized for the use case when we tend to pop off the first element. we tend to remove elements by two metrics: when they expire (~random access) and also by pushing out low PoW elements (sorted access). My intuition was that vector pops (removing low work entries) are very cheap `O(n)` worst case, and a vector `retain` periodically to prune expired messages is also relatively cheap `O(n)`. 

Doing an arbitrary number of pops from a binary heap to remove low work entries is `O(nlogn)`, and the retainment logic for live messages would have to reconstruct the heap each time from the iterator, again `O(nlogn)`

ethkey/src/random.rs
rphmeier: it's the only source of failure, yeah. @debris and I discussed making this function infallible before, and I figured this was a good time for it.

whisper/src/net.rs
rphmeier: the error will lead to a punishment in the caller.

whisper/src/net.rs
rphmeier: yeah, should be is _not_ known (I had it as `is_known` before but forgot to update when I generalized it)

PR #6035, opened 2017-07-11
fixed invalid blockies identicons

debris:
mobile wallet blockies and parity blockies sometimes were different, because parity uses Adam's fork of blockies library from a year ago

https://www.npmjs.com/package/blockies

https://www.npmjs.com/package/ethereum-blockies

# **needs backport**


jacogr:
Looks correct. My only concern would be that if the icons were different (I have not delved into the differences between the 0.0.2 version and 0.1.1), we would be giving the UI users different icons (in some cases?).

js/src/api/util/identity.js
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import blockies from 'blockies';
+import blockies from 'ethereum-blockies';
```
jacogr: Interface is not 100% the same - on line 25 where we call `blockies({..})` the new interface is `blockies.create({..})` - not sure about the other parameters

jacogr:
Ok, something is seriously wrong. I changed (locally) the `blockies({...})` to `blockies.create({...})` (otherwise it shows just an empty block and this is what I get -

parity 2017-07-11 11-10-30

Notice the `Jaco` on top and the `Jaco` in the transaction list is not the same. Then on Etherscan where the icons historically matched up, it doesn't match with either the ones above, nor what is displayed in teh accounts list (which is yet again different from what is above) -

ethereum account 0x00d1f07af5501156a3dc81ed93f9eebd81d3e472 info 2017-07-11 11-10-55

(I see there is aPR out for a random fix in the 0.1.1 version, wondering if that is not the cause of this)

debris:
### ethereum-blockies
screen shot 2017-07-11 at 11 10 57

### blockies (we use currently)
screen shot 2017-07-11 at 11 11 25

debris:
https://github.com/ethereum/blockies/pull/4

petejkim:
The identicons still don't match up with Etherscan though

PR #6038, opened 2017-07-11
Limit transaction queue memory & limit future queue

tomusdrw:
- Disables 20x current block gas limit by default
- Increases no of trasnactions in queue to 8192 by default
- Limits the queue heap usage to 2MB by default

Changes the limits of future queue to be around 10x smaller than the current queue.


ethcore/src/miner/transaction_queue.rs
```diff
@@ -557,7 +571,7 @@ pub struct TransactionQueue {
 	/// The maximum amount of gas any individual transaction may use.
 	tx_gas_limit: U256,
 	/// Current gas limit (block gas limit * factor). Transactions above the limit will not be accepted (default to !0)
-	gas_limit: U256,
+	block_gas_limit: U256,
```
arkpar: Why the rename? This is still a queue gas limit, not the block gas limit. `total_gas_limit` might be a better alternative.

rphmeier:
LGTM other than rename grumble

PR #6044, opened 2017-07-12
`--config` option handles bundled presets  

sjeohp:
https://github.com/paritytech/parity/issues/4673


parity/cli/presets/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+pub fn preset_config_string(arg: &str) -> Result<&'static str, &str> {
+    match arg.to_lowercase().as_ref() {
+        "poa" => Ok(include_str!("./config.poa.toml")),
+        "pow" => Ok(include_str!("./config.pow.toml")),
```
rphmeier: prefer to call these "kovan" and "foundation", PoA and PoW are much more general terms

rphmeier:
What exactly happens after it loads the config? does it act as a "run" command?
One of the best presets to have would be for the dev chain with `--force-sealing`, `--max-reseal-period`, etc.

sjeohp:
Yeah, once loaded it behaves like a user-specified config, which (I think) becomes a run command.

keorn:
Would be good to get rid of [this]( https://github.com/paritytech/parity/blob/master/parity/configuration.rs#L335).
Might be also worth it to make the PR more general and include presets like in [the generator](https://paritytech.github.io/parity-config-generator/), maybe @tomusdrw has some more suggestions.

sjeohp:
I'm replacing the dev special case at the moment.  The "legacy" CLI options (eg. --gasprice) are not currently included in configs so need to tweak some things there.

Yep open to suggestions for other presets.

parity/cli/mod.rs
```diff
@@ -542,6 +544,7 @@ struct Mining {
 	tx_gas_limit: Option<String>,
 	tx_time_limit: Option<u64>,
 	relay_set: Option<String>,
+	min_gas_price: Option<String>,
```
rphmeier: shouldn't it be a number (probably u64 or U256) instead of a string?

sjeohp: yeah it definitely should be, I just copied the type, will tidy it still.

sjeohp: `--gasprice` is a string for some reason and it has to play nice with that.

rphmeier: could leave `--gasprice` in legacy and introduce a new flag. our naming convention would be `--gas-price` anyway

keorn: or `--min-gas-price` to better convey what it is

sjeohp: I did introduce `--min-gas-price` as a sealing/mining option and left `--gasprice` the same.  I only meant that the two have to be interchangeable in one place so I used the same type (string) initially.

rphmeier:
Should not remove `Cargo.lock` -- we need it to be checked in.

keorn:
The convention for `--chain` is simply to provide a name instead of file, can this be the same but for `--config`?

sjeohp:
`--config` is already used for user-defined configs, these are predefined and bundled.  Could make it `--preset` but then I don't think it's intuitive that other options override it, e.g. if user specifies `--config` and `--preset`, `--config` should take precedence.

Also this does take a name, e.g. `parity preset dev` replaces the previous special case for `--chain=dev`

sjeohp:
Changed `preset` command to an additional `--config` check.
- if arg matches a local config, use that
- if not, check for matching presets

eg. from above: 
> `--config config.toml` starts with custom config, while `--config mining` starts with a bundled one

parity/cli/usage.rs
```diff
@@ -154,32 +149,26 @@ macro_rules! usage {
...
-					}
-				};
-
-				Ok(raw_args.into_args(config))
+				let config_file = raw_args.flag_config.clone().unwrap_or_else(|| raw_args.clone().into_args(Config::default()).flag_config);
+				let config_file = replace_home(&::dir::default_data_path(), &config_file);
+				match (fs::File::open(&config_file), raw_args.flag_config.clone()) {
+					// Load config file
+					(Ok(mut file), _) => {
+						println_stderr!("Loading config file from {}", &config_file);
```
rphmeier: side note: with Rust 1.19.0 we can kill this macro in favor of `eprintln!` from `std`. but we shouldn't bump the required version just yet

keorn:
Documentation similar to `--chain` missing [here](https://github.com/paritytech/parity/blob/17e6d2d51cec9e349768bf82125dbcb6e20b6038/parity/cli/usage.txt#L80).

parity/cli/usage.txt
```diff
@@ -294,6 +296,10 @@ Sealing/Mining Options:
                                    means we relay nothing if not mining);
                                    lenient - Same as strict when mining, and cheap
                                    when not (default: {flag_relay_set}).
+  --min-gas-price WEI              Minimum amount of Wei per GAS to be paid for a
```
gavofyork: since this is exactly equivalent to `--gasprice`, `--gasprice`'s definition should be changed to `Equivalent to --min-gas-price WEI`.

Issue #5957, opened 2017-06-30
SecretStore: store KeyServers list in contract


svyatonik:
Following https://github.com/paritytech/parity/issues/5502#issuecomment-312192703

Currently list of key servers it is stored in each KS configuration file => it is hard to maintain it. So suggestion is to add contract, which will store { KeyServerPub: KeyServerAddr } map. This contract could be an **alternative way** to provide a list of KeyServers on startup.

PR #6057, opened 2017-07-14
SecretStore: do not cache ACL contract address

### LINKED ISSUE #5853: SecretStore: do not cache ACL contract address
Currently ACL contract address is cached here: https://github.com/paritytech/parity/blob/8d04dffe697e9b41f9893487e220ec30623885e2/secret_store/src/acl_storage.rs#L53

=> after changing contract address in registry (which was the case, as @keorn said), it is not updated. We need a way to update it (the most obvious way is not to cache address at all).

svyatonik:
closes #5853


svyatonik:
After #5957 (work in progress) this code will be slightly updated to read contract address only when new block comes (along with new KeyServers list). But behavior will remain the same, so imho this PR could be separately reviewed/merged.

svyatonik:
closed in favor of #6107

PR #6080, opened 2017-07-18
Unroll loops in light_compute

Vurich:
Godbolt was showing that this loop (or rather, a representative facsimile of this loop) was not being unrolled, so I used an unrolling macro to force the unrolling to happen. This allows better const-folding and avoids jumps (also, since we make a bunch of function calls that stomp on the registers it avoids having to load from the stack). This is a micro-optimization but benchmarks show almost a 10% improvement.

### DIFF
--- a/ethash/src/lib.rs
+++ b/ethash/src/lib.rs
@@ -23,6 +23,8 @@ extern crate sha3;
 extern crate parking_lot;
 
 #[macro_use]
+extern crate crunchy;
+#[macro_use]
 extern crate log;
 mod compute;
 
--- a/ethash/src/lib.rs
+++ b/ethash/src/lib.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "benches", feature(test))]
+
 // Copyright 2015-2017 Parity Technologies (UK) Ltd.
 // This file is part of Parity.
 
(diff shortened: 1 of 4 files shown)


ethash/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "benches", feature(test))]
```
rphmeier: place this under the license header

rphmeier:
LGTM other than style grumble. It would be nice to have the `crunchy` crate under the `paritytech:core-devs` group as authors.

ethash/src/lib.rs
rphmeier: also under the docs :)

Vurich:
The grumbles on this PR have been fixed, can this get a rereview?

ethash/src/lib.rs
rphmeier: @Vurich this grumble is still an issue

rphmeier:
LGTM modulo the style nit

ethash/src/lib.rs
Vurich: Oh, sorry. I missed this.

Vurich:
De-nitted

PR #6085, opened 2017-07-18
Tendermint epoch transitions

### LINKED ISSUE #6046: Epoch transitions for tendermint chains
1. `epoch_end_signal` function which functions the same as `authority_round`
2. `is_epoch_end` function which checks if there is a pending transition for current block.
3. finality proof is simply the block header with a valid seal under the given validator set
4. `snapshot_components` that returns `PoASnapshot`.

Warp sync will now work seamlessly for contract-based tendermint chains.

feynmanliang:
Resolves #6046


feynmanliang:
[clabot:check]

gavofyork:
many thanks for the contribution!

rphmeier:
@feynmanliang Thanks! It's almost there (and sorry for being a bit ambiguous about this), but there are just a couple things missing:
  - There should be an `EpochVerifier` which has a `verify_light` function that checks a block's seal under the validator set
  - The `EpochVerifier` should also have a `check_finality_proof` which does the same.
  - The validator set proof should be combined with the block header as a finality proof, and in conjunction can be used to produce an `EpochVerifier`. Look at how this is done in `AuthorityRound` for guidance.

How to check the seal:
In tendermint blocks have instant finality, so any block which is correctly signed is finalized. as such, only each block can only finalize itself, unlike authority round where any block can finalize an unbounded amount of others.

For checking the seal of a header, take a look at the `verify_block_family` function of `Tendermint`. The basic steps are: 
  - iterate over `header.seal.get(2)` (fail if `None`) as `UntrustedRlp`
  - extract the signatures from each `Precommit` 
  - The block is sealed if the number of unique signers is a majority of the epoch set, and unsealed otherwise.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +101,53 @@ pub struct Tendermint {
...
+			Err(EngineError::BadSealFieldSize(OutOfBounds {
+				min: Some(threshold),
+				max: None,
+				found: n
+			}).into())
+		}
+	}
+
+	fn check_finality_proof(&self, proof: &[u8]) -> Option<Vec<H256>> {
+		panic!("Tendermint blocks have instant finality, so `check_finality_proof` should never get called.")
```
rphmeier: tendermint blocks do have instant finality, so the "proof" here would just be an encoded header (which we check for seal validity).

the way warp sync for PoA works is by providing a list of validator set changes along with proofs of finality of the change. A change is signalled, and the old validator set approves the signal by finalizing the block where the signal occurred. From this proved list of validator set handoffs we can get to the most recent finalized block with full security. `check_finality_proof` is the way we check that signals were finalized, so it can't panic.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +101,53 @@ pub struct Tendermint {
...
+	))
+}
+
+impl super::EpochVerifier for EpochVerifier {
+	fn verify_light(&self, header: &Header) -> Result<(), Error> {
+		let mut signatures = HashSet::new();
+		let ref header_signatures_field = header.seal().get(2).ok_or(BlockError::InvalidSeal)?;
+		for rlp in UntrustedRlp::new(header_signatures_field).iter() {
+			let signature = rlp.as_val()?; // TODO: why is `rlp` have type ()?
+			signatures.insert(signature);
```
rphmeier: this should also ensure that
  a) each signature is a valid signature of `header.bare_hash()`
  b) the public key recovered from each signature corresponds to an address in `self.subchain_validators`

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -590,6 +637,24 @@ impl Engine for Tendermint {
...
+		let (signal_number, set_proof) = match destructure_proofs(proof) {
+			Ok(x) => x,
+			Err(e) => return ConstructedVerifier::Err(e),
+		};
+
+		let first = signal_number == 0;
+		match self.validators.epoch_set(first, self, signal_number, set_proof) {
+			Ok((list, _)) => {
+				// Tendermint blocks have instant finality, so no need to check finality proof
+				ConstructedVerifier::Trusted(Box::new(EpochVerifier {
```
rphmeier: should still return `Untrusted` if some block hash is returned by `epoch_set`, and finality proofs may be checked (but not here). The "proof" passed here may be invalid, so this is also something of a verification function under the consensus logic.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +101,53 @@ pub struct Tendermint {
...
+		rlp.at(1)?.data()?,
+	))
+}
+
+impl super::EpochVerifier for EpochVerifier {
+	fn verify_light(&self, header: &Header) -> Result<(), Error> {
+		let mut signatures = HashSet::new();
+		let ref header_signatures_field = header.seal().get(2).ok_or(BlockError::InvalidSeal)?;
+		for rlp in UntrustedRlp::new(header_signatures_field).iter() {
+			let signature = rlp.as_val()?; // TODO: why is `rlp` have type ()?
```
rphmeier: i don't understand this comment. `rlp` has type `UntrustedRlp`

rphmeier:
Some comments I just made are shown as outdated, but they are still relevant

ethcore/src/engines/tendermint/mod.rs
feynmanliang: My bad, this was left over from my debugging >.<

ethcore/src/engines/tendermint/mod.rs
feynmanliang: :+1:

To clarify, you're referring to https://github.com/paritytech/parity/blob/master/ethcore/src/snapshot/consensus/authority.rs?utf8=%E2%9C%93#L365?

rphmeier: yes, and https://github.com/paritytech/parity/blob/master/ethcore/src/snapshot/consensus/authority.rs?utf8=%E2%9C%93#L199

ethcore/src/engines/tendermint/mod.rs
feynmanliang: :+1: will add the logic for checking finality of the block hash returned by `epoch_set`.

Does something need to be done wrt checking validity of the "proof" passed here, or is it sufficient to have that happen at https://github.com/paritytech/parity/blob/master/ethcore/src/snapshot/consensus/authority.rs?utf8=%E2%9C%93#L199?

rphmeier: it should decode correctly into the two portions (signal proof and finality proof), and each portion should be checked. signal proof is checked if `epoch_set` returns `Ok`, and `finality_proof` will be checked at that point long as we return `ConstructedVerifier::Unconfirmed` from `epoch_verifier`

ethcore/src/engines/tendermint/mod.rs
feynmanliang: I am a little bit lost here. Looking at `verify_block_family`, it pattern matches on first converting the `Header` into either a `Proposal` or `VoteStep`. Then for `Proposal`s it checks if the `proposer` is an authority in the parent block and is the `proposer` of the current `proposal`, and for `VoteStep` it does the counting of signatures.

Are we to assume here that the `Header` will always correspond to a `VoteStep`?

feynmanliang:
@rphmeier thank you for your comments and helping me through this!

ethcore/src/engines/tendermint/mod.rs
feynmanliang: Is the finality proof for tendermint just `::rlp::encode(header)`?

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -37,11 +37,12 @@ use ethkey::{recover, public_to_address, Signature};
 use account_provider::AccountProvider;
 use block::*;
 use spec::CommonParams;
-use engines::{Engine, Seal, EngineError};
+use engines::{Engine, Seal, EngineError, ConstructedVerifier};
+use engines::authority_round::{combine_proofs, destructure_proofs};
```
rphmeier: could the `tendermint` module get its own versions of these functions (as the engines will be split into seperate crates in the future)? also, `tendermint` proofs don't need "signal_number" because `signal_number` is always the number of the header passed to `epoch_verifier`. That means the proofs can be slightly more compact at 2 elements instead of 3.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -329,13 +329,13 @@ fn verify_external<F: Fn(Report)>(header: &Header, validators: &ValidatorSet, st
...
 
-fn combine_proofs(signal_number: BlockNumber, set_proof: &[u8], finality_proof: &[u8]) -> Vec<u8> {
+pub fn combine_proofs(signal_number: BlockNumber, set_proof: &[u8], finality_proof: &[u8]) -> Vec<u8> {
 	let mut stream = ::rlp::RlpStream::new_list(3);
 	stream.append(&signal_number).append(&set_proof).append(&finality_proof);
 	stream.out()
 }
 
-fn destructure_proofs(combined: &[u8]) -> Result<(BlockNumber, &[u8], &[u8]), Error> {
+pub fn destructure_proofs(combined: &[u8]) -> Result<(BlockNumber, &[u8], &[u8]), Error> {
```
rphmeier: if these are kept private it gives us more freedom to upgrade their internal behavior

ethcore/src/engines/tendermint/mod.rs
rphmeier: Also, for tendermint, `combine_proofs` signature can be `fn(&[u8], Header) -> Vec<u8>` and `destructure_proofs` would be `fn(&[u8]) -> Result<(&[u8], Header), _>`

rphmeier:
LGTM except for `combine_proofs`/`destructure_proofs` grumble. I'll also check out the branch and make sure everything's working before merging. Adding tests for the `tendermint::EpochVerifier` behavior would be good too.

ethcore/src/engines/authority_round/mod.rs
feynmanliang: :+1:

ethcore/src/engines/tendermint/mod.rs
feynmanliang: :+1:

feynmanliang:
Copied `{combine,destructure}_proofs` over; I'm happy to do the tests (can you point me to a reference of what these might look like?).

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +102,60 @@ pub struct Tendermint {
...
+			let address = public_to_address(&recover(&signature.into(), &message)?);
+
+			if !self.subchain_validators.contains(header.parent_hash(), &address) {
+				return Err(EngineError::NotAuthorized(address.to_owned()).into());
+			}
+			addresses.insert(address);
+		}
+
+		let n = addresses.len();
+		let threshold = self.subchain_validators.len() * 2/3;
```
feynmanliang: Should this be a majority (i.e. 1/2) or 2/3 (for Byzantine fault tolerance)

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +102,60 @@ pub struct Tendermint {
...
+		if n > threshold {
```
feynmanliang: Does it matter if it's ">" or ">="? I'd imagine in the case of three `subchain_validators` we'd want verification to pass when we get 2 of them

ethcore/src/engines/tendermint/mod.rs
rphmeier: Tendermint requires 2/3

rphmeier:
One way you could test is by making `EpochVerifier` generic over a `Recover` trait (with `fn recover(&self, H520) -> Result<Address, Error>` -- in the actual engine this would be instantiated with an implementation that actually does the recovery, but in a test you can mock it with some custom mapping. Then you can test behavior of `EpochVerifier<MockRecover>` to ensure that `check_finality_proof` and `verify_light` work correctly.

ethcore/src/engines/tendermint/mod.rs
rphmeier: should always be `>`. In the case of 3 subchain validators you would actually need all 3 to be honest. With an amount of byzantine players `f`, we have to have at least `3f + 1` players total for safety.

feynmanliang: 👍

ethcore/src/engines/tendermint/mod.rs
feynmanliang: 👍

feynmanliang:
@rphmeier I added a `recover` closure parameter to `EpochVerifier`'s constructor so I could stub it out in the tests.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -103,6 +103,7 @@ pub struct Tendermint {
 
 struct EpochVerifier {
 	subchain_validators: SimpleList,
+	recover: Box<Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync>,
```
rphmeier: Can we make this a generic parameter instead or even one thing in `cfg(test)` and another in `cfg(not(test))`? Doing several virtual calls per block will probably be a performance hit we don't have to take.

feynmanliang: I think we might need a virtual here if we want to have the `signature`s used in the mock be grabbed from the `tap` provided by `setup()`. The alternative would be to hard code those addresses. I'll push a commit for this and see what you think

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,6 +101,61 @@ pub struct Tendermint {
 	validators: Box<ValidatorSet>,
 }
 
+struct EpochVerifier {
+	subchain_validators: SimpleList,
+	recover: Box<Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync>,
```
rphmeier: @feynmanliang 

Try
```rust
struct EpochVerifier<F> 
    where F: Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync 
{
    subchain_validators: SimpleList,
    recover: F,
}
```

and just don't box the closures when you instantiate `EpochVerifier`.

feynmanliang: :+1:

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,12 +101,14 @@ pub struct Tendermint {
 	validators: Box<ValidatorSet>,
 }
 
-struct EpochVerifier {
+struct EpochVerifier<F>
+		where F: Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync {
```
rphmeier: style: brace on next line, unindented

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -101,12 +101,14 @@ pub struct Tendermint {
...
 	subchain_validators: SimpleList,
-	recover: Box<Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync>,
+	recover: F
 }
 
-impl super::EpochVerifier for EpochVerifier {
+impl <F> super::EpochVerifier for EpochVerifier<F>
+		where F: Fn(&Signature, &Message) -> Result<Address, Error> + Send + Sync {
```
rphmeier: style: brace on next line, unindented

ethcore/src/engines/tendermint/mod.rs
rphmeier: also the where clause should be indented one level, not two

ethcore/src/engines/tendermint/mod.rs
feynmanliang: :+1:

ethcore/src/engines/tendermint/mod.rs
feynmanliang: :+1:

rphmeier:
@feynmanliang 

LGTM, just will ask for another quick look-over by @keorn

gavofyork:
@keorn ?

PR #6100, opened 2017-07-19
fix #6052. honor --no-color for signer command

### LINKED ISSUE #6052: The --no-color option is only partly implemented
The --no-color option implemented in #1135 no longer covers all output by the parity CLI. An example of output not covered is:
`./parity signer new-token`
which uses white and bold styling by running
`Colour::White.bold().paint(...)`
instead of redefining paint and using
`paint(Colour::White.bold(), ...)`
as has been done in a0bc1f9dae00ee8a84bd466632b78862970caf89.

nicolasochem:


parity/signer.rs
```diff
@@ -48,13 +49,18 @@ pub fn codes_path(path: &Path) -> PathBuf {
...
-	Ok(generate_token_and_url(&ws_conf, &ui_conf)?.message)
+pub fn execute(ws_conf: rpc::WsConfiguration, ui_conf: rpc::UiConfiguration, logger_config: LogConfig) -> Result<String, String> {
+	Ok(generate_token_and_url(&ws_conf, &ui_conf, &logger_config)?.message)
 }
 
-pub fn generate_token_and_url(ws_conf: &rpc::WsConfiguration, ui_conf: &rpc::UiConfiguration) -> Result<NewToken, String> {
+pub fn generate_token_and_url(ws_conf: &rpc::WsConfiguration, ui_conf: &rpc::UiConfiguration, logger_config: &LogConfig) -> Result<NewToken, String> {
 	let code = generate_new_token(&ws_conf.signer_path).map_err(|err| format!("Error generating token: {:?}", err))?;
 	let auth_url = format!("http://{}:{}/#/auth?token={}", ui_conf.interface, ui_conf.port, code);
+	let paint = |c: Style, t: String| match &logger_config.color {
```
rphmeier: given that `paint` is only used once, the call to it could be replaced with the body. also, you can `match logger_config.color` directly, without borrowing it (bools are always copied).

parity/signer.rs
```diff
@@ -77,6 +83,6 @@ fn generate_new_token(path: &Path) -> io::Result<String> {
 	codes.clear_garbage();
 	let code = codes.generate_new()?;
 	codes.to_file(&path)?;
-	trace!("New key code created: {}", Colour::White.bold().paint(&code[..]));
+	trace!("New key code created: {}", White.bold().paint(&code[..]));
```
rphmeier: shouldn't this also be a call to `paint`? if so, keep `paint` as a closure.

PR #6107, opened 2017-07-20
SecretStore: do not cache ACL contract + on-chain key servers configuration

### LINKED ISSUE #5853: SecretStore: do not cache ACL contract address

### LINKED ISSUE #5957: SecretStore: store KeyServers list in contract

svyatonik:
1) acl contract address is now re-read when new block is enacted/retracted => there's no need to restart KeyServers when address is changed. closes #5853
2) it is possible to read KeyServer-s set from on-chain contract. Previously it was possible only via configuration file => when this set had to be changed, all configurations files had to be edited + all KeyServers had to be restarted. closes #5957
2.1) behavior remains the same until contract with name `secretstore_server_set` is registered on registry. Contract is here: https://github.com/paritytech/contracts/pull/45
2.2) when contract is registered, nodes set read from configuration file (previous behavior) is replaced with nodes from this contract
2.3) when nodes are added/removed from contract' set, connection is removed/established. If connection is removed, when there is active session involving 'obsolete node', this session is either fails with an error (keygen, encryption), or tries to continue without this node (decryption, signing)

Testing is currently in progress


svyatonik:
`KeyServerSet.sol` contract (see link above) must be reviewed before this => leaving `inprogress` label here

svyatonik:
Contract PR is approved => this PR is also ready for review

PR #6114, opened 2017-07-21
fix docker build in stable & add Aura test

General-Beck:
+ https://github.com/paritytech/parity/commit/d7eb3cb51ef6fed23e135b3aecc5fe1c8de66a8e


rphmeier:
Aura test should be taken out. It tests functionality that only can exists in beta/master.

PR #6124, opened 2017-07-22
Optimize to_hex for fun

Vurich:
I was just poking around the code this morning and noticed somewhere where an allocation could be removed. It's only a minor change but it's worth keeping in mind that every time you allocate, somewhere a Linux kernel programmer spills their coffee.

### DIFF
--- a/ethash/src/compute.rs
+++ b/ethash/src/compute.rs
@@ -108,8 +108,8 @@ impl Light {
 
 	pub fn file_path<T: AsRef<Path>>(cache_dir: T, seed_hash: H256) -> PathBuf {
 		let mut cache_dir = cache_dir.as_ref().to_path_buf();
-		cache_dir.push(to_hex(&seed_hash));
-		cache_dir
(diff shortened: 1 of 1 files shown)


rphmeier:
Is there a benchmark showing this is a bottleneck? Our goal is to have zero `unsafe` blocks in Parity, so without good reason I am against this being merged.

Vurich:
As far as I can tell it is not a bottleneck, I'm just messing around with it. However, this file is absolutely full of `unsafe` and this is a blatantly safe use here, so I don't think that's a strong argument against it. I would agree if you were to say that it was frivolous, though.

rphmeier:
this file is full of unsafe, and it's historically been one of the few places we haven't been able to purge it from the codebase because it degrades performance. Look at the rest of the thousands of lines of Rust code we have and see how much `unsafe` you find

Vurich:
I just found a case of unsound-but-it-works in ethash's `unsafe`, so now I'm on your side that if we're going to maintain this we should keep it safe.

PR #6125, opened 2017-07-24
rlp_derive

debris:
this pr introduces 4 new derives: 

- `RlpEncodable` - encodes a structure a list
- `RlpDecodable` - decodes structure from a  list
- `RlpEncodableWrapper` - encodes structure with a single field, without declaring new list
- `RlpDecodableWrapper` decodes structure with a single field

Possible improvements:
encoding and decoding enums


debris:
requires: rust `1.19` which has https://github.com/rust-lang/rfcs/pull/1506

rphmeier:
@General-Beck ^

@debris What makes me worry about this is that we will know that the structs serialize and deserialize to RLP, but it's a lot harder to know if it's the _correct_ RLP now. This is important for stuff like protocol structures, and existing structures we've written to old databases.

debris:
@rphmeier I agree, that's why this pr needs careful review. Long term, I believe it's much needed, cause it simplifies our code and reduces amount of boilerplate, the same way `serde_derive` reduces it for `serde`.

General-Beck:
@debris @rphmeier After the release we will switch CI to version 1.19

ethcore/light/src/lib.rs
```diff
@@ -76,6 +76,8 @@ extern crate futures;
 extern crate itertools;
 extern crate rand;
 extern crate rlp;
+#[macro_use]
+extern crate rlp_derive;
```
rphmeier: could you toss this at the end of the imports?

rphmeier:
There are already tests for the light client which directly serialize the messages with manual RLP streaming (for this reason exactly) so I am not worried about PIP breaking.

Generally, having a nice suite import/regression tests (parity-import-tests repo) will protect against regressions here as well.

PR #6127, opened 2017-07-24
Use cached prime sieve instead of recalculating each time

Vurich:
This is _much_ faster. I don't know how much speed we gain with this alone, but combined with the unrolling work we get down to consistently ~8% below the C implementation (better than before, which was inconsistently around 3% below). A possible improvement would be to use `UnsafeCell`, since doing so would be sound, but benching it shows that aquiring the lock is not a bottleneck (it's not a real lock, after all).

### DIFF
--- a/ethash/src/compute.rs
+++ b/ethash/src/compute.rs
@@ -19,7 +19,7 @@
 
 // TODO: fix endianess for big endian
 
-use primal::is_prime;
+use primal::Sieve;
 use std::cell::Cell;
 use std::mem;
 use std::ptr;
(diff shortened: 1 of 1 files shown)


arkpar:
Would be even faster to cache `data_size`/`cache_size` per epoch

Vurich:
Yeah, this is what `libethash` does. This uses less disk space, though (in theory). If we're going down that path I would rather store the whole set of epoch->`data_size`/`cache_size` in a static array so it gets read from disk instead of RAM. I don't know whether an anonymous memmap would help here too, I _think_ that would write to disk but it might just act like a memmap while actually using RAM. What we essentially want is to force allocation in swap space.

arkpar:
`epoch_number`->`size` is relatively small. Mainnet only has 133 epochs so far and there probably won't be more than 300 before the switch to PoS. As far as I know `memmap` is used for the DAG data, not for the DAG size map. This is an important memory optimization indeed, especially for the light client.

Vurich:
A constant-size solution would be to just store the last-seen one, since it's very common for us to just call this with monotonically-increasing epochs.

NikVolf:
need rebase and @arkpar approval

Vurich:
I think that on balance it might be better to have both this _and_ cached cache/data sizes. Both are relatively small memory costs but this speeds up initial calculation (since getting `is_prime` from a sieve is way faster than getting it using the standalone function) and then cached sizes means that we don't have to recalculate very often. I would say that this could be merged and then we could add a thread-local `MemoryLruCache` for those two functions. That's my two cents though, I'd appreciate feedback from @arkpar

arkpar:
@Vurich agreed.

Vurich:
This is waiting for https://github.com/paritytech/parity/pull/6193 to merge, since this will conflict with that. I'll include both this and the `MemoryLruCache` changes in the new version of the PR.

PR #6128, opened 2017-07-24
Filter tokens logs from current block, not genesis

ngotchac:
(note that filters are only used to trigger a balance update and notifications)

### DIFF
--- a/js/src/redux/providers/balancesActions.js
+++ b/js/src/redux/providers/balancesActions.js
@@ -144,34 +144,37 @@ export function updateTokensFilter (_addresses, _tokens, options = {}) {
       promises.push(api.eth.uninstallFilter(tokensFilter.filterToId));
     }
 
-    const promise = Promise.all(promises);
+    Promise
+      .all([
+        api.eth.blockNumber(),
+      ].concat(promises))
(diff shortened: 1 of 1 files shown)


js/src/redux/providers/balancesActions.js
```diff
@@ -144,34 +144,37 @@ export function updateTokensFilter (_addresses, _tokens, options = {}) {
...
-    const newFilters = Promise.all([
-      api.eth.newFilter(optionsFrom),
-      api.eth.newFilter(optionsTo)
-    ]);
-
-    promise
-      .then(() => newFilters)
+    Promise
+      .all([
+        api.eth.blockNumber()
```
jacogr: Do we need to retrieve the blockNumber again when we are already triggering this because of a blockNumber update?

ngotchac: So this doesn't get triggered on new block, only when the filters need to be updated, which is when you need to update the tokens value of the displayed addresses, ie. when switching views between your account, your address book, contracts, etc. We don't update the tokens balances for accounts we don't show, except those we own (so we can get a notification)

PR #6132, opened 2017-07-24
WASM MVP continued

NikVolf:
- call/call code/static call externs
- modified runtime to allow only contracts compiled with `panic=abort`
- examples are now compiled on stable rust 1.19 (strict requirement)


NikVolf:
would be nice to backport, because there are some changes in runtime and it's not versioned atm

NikVolf:
Also added some test refactoring because of evm-vm-wasm split pr before

PR #6133, opened 2017-07-24
Don't return 412 in health status

NikVolf:


gavofyork:
@tomusdrw please review this on return...

PR #6134, opened 2017-07-24
Move more params to the common section.

rphmeier:
- `gasLimitBoundDivisor`
- `registrar`
- `blockReward`
- `eip155Transition`

Also, Miner will now only seal the first block of a chain if there are no hard forks included.


keorn:
Since we plan to make Parity more generic over state machines is there be a better place for those?

rphmeier:
@keorn actually, that is the entire purpose of this PR. Now `AuthorityRound` and `Tendermint` have no configuration other than consensus-specific. `CommonParams` and `Builtins` will become part of the `EthereumMachine`, and engines will own only a `Machine ` and their internal members for consensus.

To ease integration, I plan to have a blanket `impl<E: Engine<EthereumMachine> + ?Sized> E { ... }` in the `ethcore` crate.

keorn:
True, `CommonParams` is mostly `Ethereum` specific anyway.

PR #6139, opened 2017-07-25
Pub sub blocks

CraigglesO:
@kaikun213 - 
1) Not entirely sure I am using unsubscribe properly, would like you to verify,
2) I changed localTransactions, if that causes any problems let me know, just needed more control.


Cargo.lock
```diff
@@ -418,6 +418,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "error-chain"
```
tomusdrw: Seems like those changes are unnecessary.

js/src/views/Status/NodeStatus/store.js
```diff
@@ -37,37 +41,65 @@ export default class StatusStore {
...
 
   startPolling () {
-    this._pollStatus();
-    this._pollLongStatus();
+    this._startPolling();
   }
 
   stopPolling () {
-    Object.keys(this._timeoutIds).forEach((key) => clearTimeout(this._timeoutIds[key]));
+    if (this.subscriptionId) {
```
tomusdrw: Should wait for `subscriptionId`, otherwise if you `stopPolling` before the promises in `startPolling` are resolved you may end up with dangling subscription.

js/src/views/Signer/store.js
```diff
@@ -49,8 +49,8 @@ export default class SignerStore {
   }
 
   @action unsubscribe () {
-    if (this._timeoutId) {
-      clearTimeout(this._timeoutId);
+    if (this.subscriptionId) {
```
tomusdrw: It should always wait for subscription id, otherwise you may end up with a dangling subscription.

js/src/views/Signer/store.js
```diff
@@ -88,20 +88,18 @@ export default class SignerStore {
   }
 
   fetchLocalTransactions = () => {
```
tomusdrw: The name suggests that the transactions will be fetched, but actually it sets up a subscription.
The name should be altered and it should be ensured that the method is never called more than once.

js/src/views/Signer/store.js
```diff
@@ -88,20 +88,18 @@ export default class SignerStore {
...
+      }
+      const keys = Object
+        .keys(transactions)
+        .filter((key) => transactions[key].status !== 'canceled');
+
+      this.setLocalHashes(keys);
+    })
+    .then((subscriptionId) => {
+      this.subscriptionId = subscriptionId;
+    });
```
tomusdrw: Would be good to at least log errors here.

js/src/views/Status/NodeStatus/store.js
```diff
@@ -37,37 +41,65 @@ export default class StatusStore {
...
+
+  _startPolling () {
+    this.api.pubsub.parity.getBlockHeaderByNumber((err, block) => {
+      if (err) {
+        return;
+      }
+      this.blockNumber = block.number;
+      this.blockTimestamp = block.timestamp;
+      this._pollMinerSettings();
+      return Promise
```
tomusdrw: `return` here is misleading, it looks a bit like `then(subscriptionId)` is chained to this promise, but it's not.

js/src/views/Status/NodeStatus/store.js
```diff
@@ -37,37 +41,65 @@ export default class StatusStore {
...
+            chain, defaultExtraData, enode, netPeers, netPort, rpcSettings, hashrate
+          });
+        })
+        .catch((error) => {
+          console.error('_pollStatuses', error);
+        });
+    })
+    .then((subscriptionId) => {
+      this.subscriptionId = subscriptionId;
+    });
```
tomusdrw: Would be good to at least log errors.

CraigglesO:
cleaned up and handles all errors/use cases. Lemme know what you think.

js/src/views/Signer/store.js
```diff
@@ -49,8 +49,13 @@ export default class SignerStore {
   }
 
   @action unsubscribe () {
-    if (this._timeoutId) {
-      clearTimeout(this._timeoutId);
+    const self = this;
+
+    if (self.subscribed) {
+      if (!self.subscriptionId) {
```
kaikun213: You can simplify the code by saving the promise itself as subscription and then resolving it in unsubscribe.
This works fine when the id does not need to get exposed.
[E.g. like here in the bonds library](https://github.com/paritytech/oo7-parity/blob/kaikun/src/index.js#L105)

CraigglesO: Ah, I understand your design now and how nice that is. Sorry I didn't fully understand your first comment.

gavofyork:
@tomusdrw could you take a look at this?

js/src/api/pubsub/parity/parity.js
```diff
@@ -185,14 +185,7 @@ export default class Parity extends PubsubBase {
...
         ? callback(error)
-        : callback(null, transactions => {
-          Object.values(transactions)
-            .filter(tx => tx.transaction)
-            .map(tx => {
-              tx.transaction = outTransaction(tx.transaction);
-            });
-          return transactions;
-        });
+        : callback(null, transactions);
```
ngotchac: Why do we get rid of the `outTransaction` here?

tomusdrw: Good catch. Brought it back.

tomusdrw:
@ngotchac Fixed, please have a look.

PR #6140, opened 2017-07-25
Fix unsoundness in ethash's unsafe code

Vurich:
`some_vec.resize(n, mem::uninitialized())` is unsound since `resize` reads the value it is passed. Since it just writes it to the allocated block, this will currently (probably) just lead to LLVM optimising only the copies out and leaving the `self.len = n`, making this call equivalent to `set_len`. Since theoretically LLVM could do anything with these reads using `set_len` is more correct. This also relies on `resize` not, for example, branching on the passed value.

Additionally, reading beyond the end of an `&mut [T; N]` is unsound, although there is data there in this case. This is less likely to cause issues but if we're going to use unsafety it'd be better to keep it sound. I think that the Rust team is within rights to take advantage of this as an optimisation opportunity (which would, of course, make this problematic), but I don't know.


ethash/src/compute.rs
```diff
@@ -269,57 +271,81 @@ pub fn light_compute(light: &Light, header_hash: &H256, nonce: u64) -> ProofOfWo
...
+		ptr::copy_nonoverlapping(
+			header_hash.as_ptr(),
+			s_mix.get_unchecked_mut(0).bytes.as_mut_ptr(),
+			32,
+		);
+		ptr::copy_nonoverlapping(
+			mem::transmute(&nonce),
+			s_mix.get_unchecked_mut(0).bytes[32..].as_mut_ptr(),
+			8
+		);
```
debris: maybe use [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) instead?

ethash/src/compute.rs
```diff
@@ -269,57 +271,81 @@ pub fn light_compute(light: &Light, header_hash: &H256, nonce: u64) -> ProofOfWo
...
+
+		// compute sha3-512 hash and replicate across mix
 		sha3::sha3_512(s_mix.get_unchecked_mut(0).bytes.as_mut_ptr(), NODE_BYTES, s_mix.get_unchecked(0).bytes.as_ptr(), 40);
-		let (f_mix, mut mix) = s_mix.split_at_mut(1);
+		let (f_mix, mix) = s_mix.split_at_mut(1);
+		let mix_words: &mut [u32; MIX_WORDS] = make_const_array!(MIX_WORDS, mix);
+
 		for w in 0..MIX_WORDS {
```
debris: can you also `unroll!` this one?

ethash/src/compute.rs
```diff
@@ -269,57 +271,81 @@ pub fn light_compute(light: &Light, header_hash: &H256, nonce: u64) -> ProofOfWo
...
-		unroll! {
-			// ETHASH_ACCESSES
-			for i_usize in 0..64 {
-				let i = i_usize as u32;
-
-				let index = fnv_hash(
-					f_mix.get_unchecked(0).as_words().get_unchecked(0) ^ i,
-					*mix.get_unchecked(0).as_words().get_unchecked(i as usize % MIX_WORDS)
-				) % num_full_pages;
+		for i in 0..ETHASH_ACCESSES as u32 {
```
debris: why did you remove `unroll!` here?

ethash/src/compute.rs
```diff
@@ -332,21 +358,38 @@ fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64)
...
+		ptr::copy_nonoverlapping(
+			mix.get_unchecked_mut(0).bytes.as_ptr(),
+			buf[64..].as_mut_ptr(),
+			32,
+		);
+		ptr::copy_nonoverlapping(
+			mix.get_unchecked_mut(0).bytes.as_ptr(),
+			mix_hash.as_mut_ptr(),
+			32,
+		);
```
debris: copy_from_slice?

ethash/src/lib.rs
```diff
@@ -143,9 +143,11 @@ mod benchmarks {
 
 	#[bench]
 	fn bench_light_compute(b: &mut Bencher) {
+		use ::std::env;
+
 		let hash = [0xf5, 0x7e, 0x6f, 0x3a, 0xcf, 0xc0, 0xdd, 0x4b, 0x5b, 0xf2, 0xbe, 0xe4, 0x0a, 0xb3, 0x35, 0x8a, 0xa6, 0x87, 0x73, 0xa8, 0xd0, 0x9f, 0x5e, 0x59, 0x5e, 0xab, 0x55, 0x94, 0x05, 0x52, 0x7d, 0x72];
 		let nonce = 0xd7b3ac70a301a249;
-		let light = Light::new(486382);
+		let light = Light::new(env::temp_dir(), 486382);
```
debris: It's better to use [tempdir](https://crates.io/crates/tempdir) crate

ethash/src/compute.rs
```diff
@@ -269,57 +271,81 @@ pub fn light_compute(light: &Light, header_hash: &H256, nonce: u64) -> ProofOfWo
...
 
 fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64) -> ProofOfWork {
+	macro_rules! make_const_array {
+		($n:expr, $value:expr) => {{
+			// We use explicit lifetimes to ensure that val's borrow is invalidated until the
+			// transmuted val dies.
+			unsafe fn make_const_array<'a, T, U>(val: &'a mut [T]) -> &'a mut [U; $n] {
+				use ::std::mem;
+
+				debug_assert_eq!(val.len() * mem::size_of::<T>(), $n * mem::size_of::<U>());
```
rphmeier: TODO: dependent typing :)

ethash/src/compute.rs
Vurich: Because `ETHASH_ACCESSES` is 64 and 64 * 32 = 2048. We don't want 2048 copies of the same code. Best to leave the outer loop as a real loop.

ethash/src/lib.rs
Vurich: I just copied this from another place in the code that did this.

ethash/src/compute.rs
Vurich: A lot of our `ptr::copy_nonoverlapping` calls can be replaced with `copy_from_slice`, I'll rewrite all our code to use it in a seperate PR (unless you'd consider this a blocker).

ethash/src/compute.rs
```diff
@@ -269,57 +271,81 @@ pub fn light_compute(light: &Light, header_hash: &H256, nonce: u64) -> ProofOfWo
...
+
+		// compute sha3-512 hash and replicate across mix
 		sha3::sha3_512(s_mix.get_unchecked_mut(0).bytes.as_mut_ptr(), NODE_BYTES, s_mix.get_unchecked(0).bytes.as_ptr(), 40);
-		let (f_mix, mut mix) = s_mix.split_at_mut(1);
+		let (f_mix, mix) = s_mix.split_at_mut(1);
+		let mix_words: &mut [u32; MIX_WORDS] = make_const_array!(MIX_WORDS, mix);
+
 		for w in 0..MIX_WORDS {
-			*mix.get_unchecked_mut(0).as_words_mut().get_unchecked_mut(w) = *f_mix.get_unchecked(0).as_words().get_unchecked(w % NODE_WORDS);
+			*mix_words.get_unchecked_mut(w) =
```
rphmeier: to clarify: what was happening before was that it read into `mix[1]` from a pointer to the words of   `mix[2]`? and now it explicitly indexes into the one it should?

ethash/src/compute.rs
Vurich: Yes, I can! It's probably a good candidate, too.

ethash/src/compute.rs
```diff
@@ -391,9 +434,11 @@ fn light_new<T: AsRef<Path>>(cache_dir: T, block_number: u64) -> Light {
 	assert!(cache_size % NODE_BYTES == 0, "Unaligned cache size");
 	let num_nodes = cache_size / NODE_BYTES;
 
-	let mut nodes = Vec::with_capacity(num_nodes);
-	nodes.resize(num_nodes, Node::default());
+	let mut nodes: Vec<Node> = Vec::with_capacity(num_nodes);
 	unsafe {
+		// Use uninit instead of unnecessarily writing `size_of::<Node>() * num_nodes` 0s
+		nodes.set_len(num_nodes);
```
rphmeier: the loop that sets the `bytes` of each node to the hash of the previous' will be broken now if `nodes[0].bytes` is anything but default.

ethash/src/compute.rs
Vurich: It got a pointer to `mix[1]`, transmuted it to a `[u32; NODE_WORDS]`, and then read between index 0 and `NODE_WORDS * 2`. Basically reading into `mix[2]`, yeah. Now it transmutes to a `[u32; NODE_WORDS * 2]`.

ethash/src/compute.rs
rphmeier: this code was written before `copy_from_slice` was stabilized, but now we can use it.

ethash/src/compute.rs
Vurich: If only, if only.

ethash/src/compute.rs
Vurich: The only thing is that unsafe code (`mem::uninitialized`) cannot trust safe code `copy_from_slice`. `copy_from_slice` can assume that its data is initialized, whereas `copy_nonoverlapping`, as an unsafe function, cannot.

ethash/src/lib.rs
debris: I think we do use it in several places, but you are right `env::temp_dir` is way more common. I will create a separate issue for this

ethash/src/compute.rs
```diff
@@ -210,17 +210,18 @@ pub fn slow_get_seedhash(block_number: u64) -> H256 {
...
 }
 
-#[inline]
 fn sha3_512(input: &[u8], output: &mut [u8]) {
 	unsafe { sha3::sha3_512(output.as_mut_ptr(), output.len(), input.as_ptr(), input.len()) };
 }
 
-#[inline]
+fn sha3_512_inplace(input: &mut [u8]) {
+	unsafe { sha3::sha3_512(input.as_mut_ptr(), input.len(), input.as_ptr(), input.len()) };
```
rphmeier: with a comment explaining why it's safe? couldn't a `sha3` implementation technically optimize by using the output buffer for absorbing + squeezing instead of an internal one which is later copied?

ethash/src/compute.rs
debris: copy_from_slice is exactly the same thing as copy_nonoverlapping [link](https://doc.rust-lang.org/src/core/slice/mod.rs.html#667)

ethash/src/compute.rs
```diff
@@ -238,25 +242,34 @@ fn get_data_size(block_number: u64) -> usize {
...
+	unsafe {
+		// This is safe - the `sha3_512` call below reads the first 40 bytes (which we explicitly set
+		// with two `copy_nonoverlapping` calls) but writes the first 64, and then we explicitly write
+		// the next 32 bytes before we read the whole thing with `sha3_256`.
+		//
+		// This cannot be elided by the compiler as it doesn't know the implementation of
+		// `sha3_512`.
+		let mut buf: [u8; 64 + 32] = mem::uninitialized();
+
+		ptr::copy_nonoverlapping(header_hash.as_ptr(), buf.as_mut_ptr(), 32);
```
debris: copy_from_slice is exactly the same thing as copy_nonoverlapping [link](https://doc.rust-lang.org/src/core/slice/mod.rs.html#667)

Vurich: Not for uninit. The docs for `mem::uninitialized` specifically mention that the `ptr::*` functions are the only valid way to initialise uninitialised memory

arkpar:
please revert changes to `ethcore/res/ethereum/tests`

Vurich:
Fixed, should be ready to merge after CI. Sorry about that.

Issue #5872, opened 2017-06-19
Parity does not filter `100.64.0.0/10` ip range when `--allow-ips=public`


offlinehacker:
All these private bogon ip addresses must be blocked https://www.team-cymru.org/Services/Bogons/bogon-bn-agg.txt

5chdn:
It seems Parity only blocks RFC1918 private network connections, but it should also block reserved IPs. This is not really a bug but an annoyance as 100.64.0.0/10 is not a private network.

We should consider blocking all reserved IP address spaces. https://en.wikipedia.org/wiki/IPv4#Special-use_addresses

I personally use iptables to circumvent this issue. https://ethereum.stackexchange.com/a/9964

sjeohp:
Just confirming that the solution here is to filter all [reserved ips](https://en.wikipedia.org/wiki/Reserved_IP_addresses) from the public list?

sjeohp:
Also it looks like the only current criteria for the `--allow-ips=private` list is being not-[global](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_global), so some reserved IPs (loopback, link-local, etc) can be included.  Should probably filter those too?

arkpar:
@sjeohp This should really be addressed by allowing custom ranges in addition to predefined "public" and "private" sets. E.g. `--allow-ips="100.64.0.0/10 2.3.0.0/32"`

5chdn:
On a practical note, it might be easier (from a user perspective) blacklisting rather than whitelisting sets of ip adresses, i.e., `--block-ips="private,100.64.0.0/10,2.3.0.0/32"`

sjeohp:
I've refactored to exclude more of the reserved/special-use IPs including `100.64.0.0`, but there are still ambiguities in the official reserved lists, hence why Rust's `std::net` IP functions have been unstable for so long.

I can add both `--allow-ips` and `--block-ips`.  Only issue I can see would be potential conflicts,  e.g. passing `--allow-ips="100.64.0.0/10 2.3.0.0/16" --block-ips="private,100.64.0.0/10,2.3.0.0/32"`, but I think an error in that case makes sense.

sjeohp:
On second thought maybe something like `--allow-ips="private,100.64.0.0/10,-2.3.0.0/16"`?  Where that would allow all in the private list and the 100.64.0.0/10 range, and block 2.3.0.0/16, with order indicating precedence, either ascending or descending.

[...]

PR #6144, opened 2017-07-25
Refactor --allow-ips to handle custom ip-ranges

sjeohp:
Filter some additional reserved ranges from the predefined `public` and `private` lists, as in [#5872](https://github.com/paritytech/parity/issues/5872).  Also:

`--allow-ips FILTER` supports everything it did before but accepts more args:
> FILTER can be one of:
                                   private - connect to private network IP addresses only;
                                   public - connect to public network IP addresses only;
                                   all - connect to any IP address;
                                   none - block all (for use with a custom filter as below);
                                   a custom filter list in the format: "private,ip_range1,-ip_range2,...".
                                   Where ip_range1 would be allowed and ip_range2 blocked;
                                   Custom blocks ("-ip_range") override custom allows ("ip_range");
                                   (default: {flag_allow_ips}).

Accepts one predefined symbol (public, private, all, or none) and unlimited custom ranges.  Could potentially support combinations of predefined like "public private ..." but not sure it would be useful/worthwhile.


arkpar:
Please re-format changed files to use tabs instead of spaces. We use tabs consistently in our codebase.

util/network/src/lib.rs
```diff
@@ -145,14 +148,61 @@ impl NonReservedPeerMode {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "ipc", binary)]
+pub struct IpFilter {
+    pub predefined: AllowIP,
+    pub custom_allow: Vec<IpNetwork>,
+    pub custom_block: Vec<IpNetwork>,
+} 
+
+impl IpFilter {
+    pub fn new() -> IpFilter {
```
arkpar: I'd name it `default`  or `none` or `empty` to show that it creates an empty filter

util/network/src/lib.rs
```diff
@@ -145,14 +148,61 @@ impl NonReservedPeerMode {
...
+        let mut filter = IpFilter::new();
+        for f in s.split_whitespace() {
+            match f {
+                "all" => filter.predefined = AllowIP::All,
+                "private" => filter.predefined = AllowIP::Private,
+                "public" => filter.predefined = AllowIP::Public,
+                "none" => filter.predefined = AllowIP::Non,
+                custom => {
+                    if custom.starts_with("-") {
+                        match IpNetwork::from_str(&custom.to_owned().split_off(1)) {
```
arkpar: Could be just `filter.custom_block.push(IpNetwork::from_str(&custom.to_owned().split_off(1)?)`; Same for the `match` below.

util/network/src/lib.rs
```diff
@@ -145,14 +148,61 @@ impl NonReservedPeerMode {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub enum AllowIP {
 	/// Connect to any address
 	All,
 	/// Connect to private network only
 	Private,
 	/// Connect to public network only
 	Public,
+
+    Non,
```
arkpar: `None`?

arkpar:
Looks great apart from minor style issues.

util/network/src/lib.rs
sjeohp: No, I didn't want it to be confused with `Option::None` but reading it now I can see that it might be more confusing :-p

PR #6145, opened 2017-07-25
Added cargo-fuzz tests for Rlp crate

onicslabs:
Set up some fuzz test harnesses for the UntrustedRlp struct
and RlpStream's append_raw function.

---
This change is [Reviewable](https://reviewable.io/reviews/paritytech/parity/6145)


util/rlp/fuzz/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+
+[package]
+name = "rlp-fuzz"
+version = "0.0.1"
+authors = ["Automatically generated"]
```
rphmeier: can you set this to `Parity Technologies <admin@parity.io>`?

onicslabs: can do :)

onicslabs:
[clabot:check]

rphmeier:
A `README.md` in the `rlp-fuzz` crate explaining how to run it (unless it's just `cargo test -p rlp-fuzz`?) would be great too. Do you recommend us to run this as part of the CI, or our nightly tests?

onicslabs:
This addition won't pass under a stable build, since cargo-fuzz requires nightly to compile correctly. Should I remove this pull request?

Is it possible to exclude fuzz directories from stable compilation, then check fuzz directories compile correctly under a nightly build?

onicslabs:
I'd be happy to do a small readme.md

I'll include a basic cargo-fuzz installation guide, and run through steps for the entire fuzz testing cycle.

rphmeier:
@onicslabs You could replace the crate with a stub when a `cfg(feature = "nightly")` is not true. You can set this feature in a build script that checks using the `rustc_version` crate.

onicslabs:
Just added conditional compilation for the two fuzzing tests I wrote.
Long-term, cargo-fuzz should integrate a feature/option to automatically add conditional compilation.

May start working on this, and see what the cargo-fuzz community thinks about adding a feature like that.

util/rlp/fuzz/fuzz_targets/append_raw.rs
```diff
@@ -0,0 +1,18 @@
+#![no_main]
+#[macro_use] extern crate libfuzzer_sys;
+#[cfg(feature = "nightly"])
```
rphmeier: shouldn't it be `#![cfg(feature = "nightly")]` to be applied to the whole source file? to me this looks like it will only import the `rlp` crate on nightly, but still try to do everything else on stable.
same for `untrusted.rs`

util/rlp/fuzz/fuzz_targets/append_raw.rs
```diff
@@ -0,0 +1,19 @@
+#![no_main]
+#[macro_use] extern crate libfuzzer_sys;
+#[cfg(feature = "nightly")]
```
rphmeier: there is a distinction between `#[attr]` and `#![attr]`. One is applied to the following item, and the other is applied to the enclosing item.

onicslabs: I'm having trouble getting features to work nicely with cargo-fuzz. For example, to enable the feature, one must pass the `--features <feature-name` flag to `cargo build`.

I haven't been able to find documentation on how to enable a feature from the build script, or through cargo-fuzz options. The entire project compiles fine with this feature inclusion, but now the fuzz tests will not run with `cargo fuzz run untrusted`.

Currently I am trying to require the nighlty feature to build the binary targets with `required-feature` option under `parity/util/rlp/fuzz/Cargo.toml`. This stops the binaries from being built, but it is unclear how to enable these features when using a nightly compiler.

onicslabs: figured out enabling features in build scripts:
`println!("cargo:rustc-cfg=feature=\"nightly\"")`

rphmeier: > does this mean that #![cfg(feature = "nightly")] at the top of the file will include the entire file as a nightly feature?

yep

util/rlp/fuzz/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    match version_meta().unwrap().channel {
+        Channel::Stable => {
+            println!("cargo:rustc-cfg=RUSTC_IS_STABLE");
+        }
+        Channel::Beta => {
+            println!("cargo:rustc-cfg=RUSTC_IS_BETA");
+        }
+        Channel::Nightly => {
+            println!("cargo:rustc-cfg=RUSTC_IS_NIGHTLY");
+            println!("cargo:rustc-cfg=feature=\"nightly\"")
```
rphmeier: what's the purpose of all the `RUSTC_IS_*`? can't you just `if let Channel::Nightly = version_meta().unwrap() { /* set feature flag */ }`

onicslabs: you're right, I took this code from an example on using rustc_version (not super familiar with conditional compilation).

I'll remove the code from the other cases, and only set the feature flag for nightly

util/rlp/fuzz/fuzz_targets/append_raw.rs
```diff
@@ -0,0 +1,19 @@
...
+use rand::Rng;
+
+fuzz_target!(|data: &[u8]| {
+    let mut rls = RlpStream::new();
+
+    // Generate random item count to further fuzz append_raw's parsing
+    let mut rng = rand::thread_rng();
+    let rand_item_count = rng.gen_range::<usize>(1 as usize, 4096 as usize);
+
+    let _ = rls.append_raw(data, rand_item_count);
```
rphmeier: `append_raw` doesn't do any checking of the data passed AFAIK. but it does check that `item_count` is within the number of remaining items in the list (which in this case, would be 1).

what do you expect to find by fuzz-testing this? `RlpStream` is typically not exposed to external input, whereas `UntrustedRlp` is.

util/rlp/fuzz/fuzz_targets/untrusted.rs
```diff
@@ -0,0 +1,11 @@
+#![no_main]
+#![cfg(feature = "nightly")]
+#[macro_use] extern crate libfuzzer_sys;
+extern crate rlp;
+
+use rlp::UntrustedRlp;
+
+fuzz_target!(|data: &[u8]| {
+    let _ = UntrustedRlp::new(data);
```
rphmeier: creating `UntrustedRlp` just wraps the data given. probably this fuzz test should try and iterate over all lists recursively.

util/rlp/fuzz/fuzz_targets/append_raw.rs
onicslabs: trying to get full fuzz-coverage of the classes in the rlp crate, and I hadn't considered where the data may be coming from.

I agree it makes more sense to fully fuzz the UntrustedRlp struct, and am focusing more on that now that the foundational stuff is fixed.

Also still becoming more familiar with the code base, another expectation/goal of writing fuzz tests for all the code.

If for some reason or another, a developer decides to use a normally internal function/struct, it would be beneficial to know if that feature is buggy.

util/rlp/fuzz/fuzz_targets/untrusted.rs
onicslabs: ok, what do you mean by all the lists?

rphmeier: RLP is a data format consisting mostly of nested lists.

onicslabs: I added another fuzz test that creates a list (prepending `0xc + data_len` to the fuzz data).

> probably this fuzz test should try and iterate over all lists recursively.

Do you mean creating lists of various types from the fuzz data, and recursively calling `UntrustedRlp::new(fuzz-list-data)` on those lists?

rphmeier: `UntrustedRlp` has an `iter()` method yielding an iterator over other `UntrustedRlp` elements within it. You could write a function
```rust
fn iter_recursive(rlp: UntrustedRlp) {
    for x in rlp.iter() {
        iter_recursive(x)
    }
}
```

this would essentially walk the whole RLP structure recursively

onicslabs: this sounds like it requires building a minimally well-formed list in the RLP structure. is this what you are talking about building from the fuzz data?

i don't fully understand your thinking on this. i get the code coverage, just trying to figure out guide the fuzzer without constraining it too much.

onicslabs: after reading through `parity/util/rlp/src/impls.rs`, think i get what you were talking about with recursive lists. Make a nested list, where each item is a nested list of each type and subtype en/decoded by the UntrustedRLP structure.

```
rlp-root
|- hash-list
 -|-hash64
 -|-hash128
 -|-hash160
 -|-etc
```
Do you think throwing fuzz data into these list items will get the right code/crash coverage (using your recursive function to traverse the RLP structure)?

util/rlp/fuzz/gen_untrusted_corpus.rs
```diff
@@ -0,0 +1,94 @@
...
+    rlp.append(&u8b);
+    rlp.append(&u16b);
+    rlp.append(&u32b);
+    rlp.append(&u64b);
+    rlp.append(&u128b);
+    rlp.append(&u256b);
+
+    // Read in base path to fuzzing corpus directory from `RLPCORPUS` environment var
+    let corp = env::var("RLPCORPUS").unwrap();
+    let fp = format!("{}{}", corp, "/untrusted_data/well-formed-list-uint");
```
rphmeier: should use `PathBuf` and appending methods for this

util/rlp/fuzz/fuzz_targets/untrusted.rs
rphmeier: So the interesting thing about RLP is that actually, each item is either one of a few things:
  - A list of items, prefixed by its length
  - Special byte for empty data (treated as empty list)
  - A list of bytes, prefixed by its length

This is why it's called "recursive list prefix".
Take a look at [the spec](https://github.com/ethereum/wiki/wiki/RLP) for more details.
All of our strongly-typed decoding is implemented on top of those primitives.

What I'm suggesting is that we have the fuzzer try and generate data, which we treat as a bunch of nested lists of bytes (and try to walk accordingly). This will allow the fuzzer to attempt to explore almost all code paths in the decoder, possibly finding some well-formed RLP that leads to panic.

onicslabs: > What I'm suggesting is that we have the fuzzer try and generate data which we treat as a bunch of nested lists of bytes (and try to walk accordingly)

Currently, I'm using the `gen_untrusted_corpus.rs` tool to build some basic, well-structured RLP streams. I'm still working on generating all the data types `UntrustedRlp` can normally handle, and then I'll build some RLP streams with mostly good, but a little messed up data. Hopefully, this will give the fuzzer more to mutate over, and we'll get deeper coverage.

> Isn't this the usual strategy with fuzz testing? e.g. if you have a PNG decoder, you try to have the fuzzer generate data that breaks the decoder, not to make the fuzzer generate a picture of a horse.

Absolutely agree. I'm trying to add more to the fuzzer corpus, so that its mutation algorithms can be more efficient in breaking the RLP decoder. Like sending in a PNG of a horse with a trojan encoded inside :)

onicslabs:
Getting much better results with the latest improvements. Before, got no crashes even after multiple days use. Now, after only a few minutes testing, all workers finish after reaching a crash threshold. Not sure how much is due to errors on my part, or if the fuzzer is actually finding bugs now :)

digging in deeper to find out what is what, wish me luck

util/rlp/fuzz/std/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "rlp-std"
```
rphmeier: what's the purpose of this crate? we will add the given test case as a unit test when we prepare the fix.

onicslabs: just a quick implementation to confirm the panic works under a stable build. I will retract it since it will be a unit test.

util/rlp/fuzz/build.rs
```diff
@@ -0,0 +1,19 @@
...
+fn main() {
+    assert!(version().unwrap().major >= 1);
+
+    match version_meta().unwrap().channel {
+        Channel::Stable => {
+        }
+        Channel::Beta => {
+        }
+        Channel::Nightly => {
+            println!("cargo:rustc-cfg=feature=\"nightly\"")
```
debris:
```rust
if let Channel::Nighlty = version_meta().unwrap().channel {
  println!("cargo:rustc-cfg=feature=\"nightly\"")
}
```

util/rlp/fuzz/gen_untrusted_corpus.rs
```diff
@@ -0,0 +1,113 @@
...
+            panic!("couldn't write to file: {}", why.description())
+        },
+        Ok(_) => println!("successfully wrote to file")
+    }
+
+}
+
+fn write_corpus(corpdir: &str, rlp: &RlpStream) {
+    // Read in base path to fuzzing corpus directory from `RLPCORPUS` environment var
+    let corp = env::var("RLPCORPUS").unwrap();
```
debris: what is fuzzing corpus? does it need to be persistent? also:

1. It will crash on everyone's machine, cause `RLPCORPUS` is no defined
2. I see no reason for tests to spoil my filesystem.

onicslabs: > what is fuzzing corpus? does it need to be persistent?

the fuzzing corpus starts as an initial set of valid input data, and cargo-fuzz mutates the initial data into a larger, potentially crash-causing input. it needs to be somewhat persistent, but you could generate the corpus from scratch each time.

generating from scratch is computationally more expensive (depends on size of corpus), and some amount of state is lost in code penetration. cargo-fuzz also has a command `cmin` to minify the corpus.

> It will crash on everyone's machine, cause RLPCORPUS is no defined

True, I wasn't sure on the preferred way to do this. I'll change the path to a constant in the test, and use relative path so its all maintained in the `fuzz/corpus/<test-name>` directory.

> I see no reason for tests to permanently spoil my filesystem. can you use tempdir?

What about adding tempdir as an option, for testers that don't want to have a persistent fuzzing corpus? 

Tempdir definitely seems to be the right way to go for automated build tests, since they just need basic fuzzing coverage. 

Then in contexts that require deeper coverage, tests can still be run with a persistent corpus.

gavofyork:
@debris / @onicslabs looking stale... any intention to sort this?

onicslabs:
I'm willing to correct any remaining grumbles. The most recent two commits were meant to resolve code cleanup around requiring a nightly build for fuzz tests, and config variables being defined with environment variables.

Is there anything I can do about integrating the build into your CI? Wasn't sure if this was something internal or not.

onicslabs:
@debris I'm looking through `gitlab-ci.yml`, would a new build entry for something like `test-fuzz-nightly` work for CI integration? 

Where the test build switches to nightly, then runs a script that checks for/runs fuzz tests for a configured time or number of iterations.

tomusdrw:
@onicslabs any progress?

onicslabs:
@tomusdrw was still waiting on the preferred method of CI testing. Not sure of best way to add fuzz tests, so that they only run for 100 rounds, and only when changes are made to the crate they are fuzzing.

tomusdrw:
@onicslabs I'm fine with running fuzzing every time during nightly builds, not sure how to achieve that though. @paritytech/ci ?

debris:
It's stale, so I'm closing it

PR #6146, opened 2017-07-25
SecretStore: encrypt messages using private key from key store

### LINKED ISSUE #5502: SecretStore: dynamically generate KeyPairs

svyatonik:
on top of #6107 

closes #5502 

Previously secret key, used to encrypt communication between two nodes was stored as plain text in parity configuration file. Now you can use secret key of any account from key store (similar to `--engine-signer` as Peter suggested). Previous configuration is also compatible - if len(configuration_file.secretstore.self_secret) == 64, then it is parsed as plain-text secret, if len is equal to 40, then it is parsed as account' address && password for this account is required (configuration_file.account.password).


svyatonik:
inprogress because #6107 is also inprogress

svyatonik:
This PR is for last 6 commits only

ethstore/src/lib.rs
```diff
@@ -35,6 +35,7 @@ extern crate ethcore_bigint as bigint;
 extern crate ethcrypto as crypto;
 extern crate ethkey as _ethkey;
 extern crate parity_wordlist;
+extern crate ethcore_util as util;
```
debris: I can't see where it's used. Isn't it redundant? Besides that `ethstore` is also compiled for android and ios. Bringing `util` dependency here, breaks that.

ethstore/src/ethstore.rs
```diff
@@ -509,6 +513,14 @@ impl SimpleSecretStore for EthMultiStore {
 		Err(Error::InvalidPassword)
 	}
 
+	fn agree(&self, account: &StoreAccountRef, password: &str, other: &Public) -> Result<Secret, Error> {
+		let accounts = self.get_matching(account, password)?;
+		for account in accounts {
+			return account.agree(password, other);
```
debris: imo, it would be more idiomatic to write

```rust
match accounts.first() {
  Some(ref account) => account.agree(password, other),
  None => Err(Error::InvalidPassword),
}
```

ethstore/src/lib.rs
svyatonik: Thanks. Checked it again - ethstore compiles without it. Looks like an artifact of some experiment. Will update PR

ethstore/src/ethstore.rs
svyatonik: also fixed couple of other functions above

PR #6153, opened 2017-07-26
Add support for ConsenSys multisig wallet

ngotchac:
This PR adds support for [this](https://github.com/ConsenSys/MultiSigWallet/blob/master/contracts/solidity/MultiSigWallet.sol) multisig wallet. It should also be easier to add support for other Wallets now.

It also fixes a bug in the Transfer modal when sending a token from a Wallet that has no ETH.

NB : there might be some gas estimation issues when for example confirming an owner addition. As this version of the Multisig Wallet doesn't throw when the executed call fails, the gas estimation is a bit undervalued.


jacogr:
Looks good. Spend a lot of time on the new calculations as well, just double-checking the logic - was not 100% sure why it was re-done here without splitting it out, makes the PR bigger and the logic much more fragile and difficult to follow. 

Would really need your help to port this to UI-2 as well, where everything now lives in separate repos (`dapp-wallet` is mostly affected, contract stubs would end with `js-shared`, nothing should change in the parity repo itself, i.e. all affected files here are now externally). Not sure it is possible for me to do this port without breaking things.

PR #6156, opened 2017-07-26
Light client improvements

rphmeier:
- a few bug fixes in `on_demand` and `parity_rpc`
- blockchain import command for light clients.

### DIFF
--- a/ethcore/light/src/on_demand/mod.rs
+++ b/ethcore/light/src/on_demand/mod.rs
@@ -392,7 +392,7 @@ impl OnDemand {
 							self.in_transit.write().insert(req_id, pending);
 							return None
 						}
-						Err(net::Error::NoCredits) => {}
+						Err(net::Error::NoCredits) | Err(net::Error::NotServer) => {}
 						Err(e) => debug!(target: "on_demand", "Error dispatching request to peer: {}", e),
 					}
 				}
--- a/ethcore/light/src/on_demand/mod.rs
+++ b/ethcore/light/src/on_demand/mod.rs
@@ -266,7 +266,6 @@ impl OnDemand {
 		-> Result<Receiver<Vec<Response>>, basic_request::NoSuchOutput>
 	{
 		let (sender, receiver) = oneshot::channel();
-
 		if requests.is_empty() {
 			assert!(sender.send(Vec::new()).is_ok(), "receiver still in scope; qed");
 			return Ok(receiver);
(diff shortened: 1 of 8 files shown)


rphmeier:
@arkpar d42d798 and 3ee5390 may be worth backporting before 1.7

arkpar:
Contains commits from https://github.com/paritytech/parity/pull/6148, probably needs a rebase

ethcore/light/src/on_demand/mod.rs
```diff
@@ -54,13 +54,13 @@ struct Peer {
...
 impl Peer {
-	// whether this peer can fulfill the
+	// whether this peer can fulfill the necessary capabilities
 	fn can_fulfill(&self, c: &Capabilities) -> bool {
 		let caps = &self.capabilities;
 
-		caps.serve_headers == c.serve_headers &&
-			caps.serve_chain_since >= c.serve_chain_since &&
-			caps.serve_state_since >= c.serve_chain_since
+		caps.serve_headers >= c.serve_headers &&
```
arkpar: `serve_headers` is a bool as far as I can see. Why `>=` ?

ethcore/light/src/on_demand/mod.rs
```diff
@@ -54,13 +54,13 @@ struct Peer {
...
+			(c.serve_chain_since.is_none() || caps.serve_chain_since <= c.serve_chain_since) &&
+			(c.serve_state_since.is_none() || caps.serve_state_since <= c.serve_state_since)
```
arkpar: Why does `is_none` allow a request?
Shouldn't  chain and state numbers be strictly lower?

ethcore/light/src/on_demand/mod.rs
rphmeier: the purpose of this function is to check whether the capabilities `c` are encompassed by the capabilities `caps`. If `c.serve_headers` is `false`, `caps.serve_headers` can be `false` or `true. If `c.serve_headers` is `true`, `caps.serve_headers` can only be `true`.

ethcore/light/src/on_demand/mod.rs
rphmeier: should probably be rewritten as `c.serve_chain_since >= caps.serve_chain_since`

rphmeier: if the request doesn't require any state or chain serving, then it doesn't matter what the peer capabilities are. If the request does require chain or state serving, then the block number this peer supports beyond should be `<=` the request required block number. but this should also guard against `caps.serve_since.is_none()`. I will rewrite for clarity

PR #6167, opened 2017-07-27
Fix connecting to wildcard addresses.

### LINKED ISSUE #6166: ws connection failure in 1.7.0
Hello,
I'm testing 1.7.0 using binary from gitlab artifacts (https://gitlab.parity.io/parity/parity/pipelines/14209 "linux stable").

After the node is started, I open http://$IP:8180 to access dapps. However I get stuck there, because of the popup that asks for a security token. After I generate a token and enter it into the input in the bottom of the popup, it just stays there. [Here is a screenshot](http://take.ms/vyTKd). Looks like it tries to establish ws connection to the node, but fails.

Seems like in 1.7.0 there's a new section `[websockets]` in toml configuration file. Here is my config.toml in full

```
[parity]
chain = "spec.json"
base_path = "parity"
[network]
nat="extip:$IP"
port = 30300
discovery=true
[rpc]
cors = "all"
interface = "all"
hosts = ["all"]
port = 8540
[ui]
interface = "all"
port = 8180
[dapps]
cors = "all"
interface = "all"
hosts = ["all"]
port = 8080

[websockets]
disable = false
port = 8546
interface = "all"
origins = ["all"]
apis = ["web3", "eth", "net", "parity", "traces", "rpc", "secretstore"]
hosts = ["all"]

[account]
password = ["node.pwd"]
[mining]
force_sealing = true
engine_signer = "0xDd0BB0e2a1594240fED0c2f2c17C1E9AB4F87126"
reseal_on_txs = "none"
[footprint]
tracing = "on"
pruning = "archive"
fat_db = "on"
```
The problem in my case is that I run parity on Azure VM and the VM itself does not have a public ip address. So I can't  use it in `interface`s as parity won't be able to bind to it
```
$ ifconfig -a
eth0      Link encap:Ethernet  HWaddr 00:0d:3a:19:e7:23  
          inet addr:10.0.0.4  Bcast:10.0.0.255  Mask:255.255.255.0
          inet6 addr: fe80::20d:3aff:fe19:e723/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:353816 errors:0 dropped:2 overruns:0 frame:0
          TX packets:147233 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:387496544 (387.4 MB)  TX bytes:157183388 (157.1 MB)

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:449861 errors:0 dropped:0 overruns:0 frame:0
          TX packets:449861 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1 
          RX bytes:64818567 (64.8 MB)  TX bytes:64818567 (64.8 MB)
```

I had a similar connection problem before, but was able to solve it in 1.6.x by adding `nat="extip:$IP"` into `[network]` section. But now it looks like ws tries to use `0.0.0.0` instead of actual ip from `nat`.

Please advise how to fix this.

tomusdrw:
Closes #6166

### DIFF
--- a/js/src/secureApi.js
+++ b/js/src/secureApi.js
@@ -101,13 +101,7 @@ export default class SecureApi extends Api {
       return 'dapps.parity';
     }
 
-    const { host } = this._dappsAddress;
-
-    if (!host || host === '0.0.0.0') {
-      return window.location.hostname;
-    }
-
-    return host;
+    return this._dappsAddress.host;
   }
 
   get isConnecting () {
@@ -171,6 +165,25 @@ export default class SecureApi extends Api {
         this.emit('disconnected');
         console.error('unhandled error in secureApi', error);
       });
+  }
+
+  /**
+   * Resolves a wildcard address to `window.location.hostname`;
+   */
+  _resolveHost (url) {
+    const parts = url ? url.split(':') : [];
+    const port = parts[1];
+    let host = parts[0];
+
+    if (!host) {
+      return host;
+    }
+
+    if (host === '0.0.0.0') {
+      host = window.location.hostname;
+    }
+
+    return port ? `${host}:${port}` : host;
   }
 
   /**
@@ -316,8 +329,8 @@ export default class SecureApi extends Api {
         this._uiApi.parity.wsUrl()
       ])
       .then(([dappsUrl, wsUrl]) => {
-        this._dappsUrl = dappsUrl;
-        this._wsUrl = wsUrl;
+        this._dappsUrl = this._resolveHost(dappsUrl);
+        this._wsUrl = this._resolveHost(wsUrl);
       });
   }
 


js/src/secureApi.js
```diff
@@ -174,6 +168,25 @@ export default class SecureApi extends Api {
   }
 
   /**
+   * Resolves a wildcard address to `window.location.hostname`;
+   */
+  _resolveHost (url) {
+    const parts = url ? url.split(':') : [];
```
jacogr: Would have probably gone for 

```
let [host, port] = url ? url.split(':') :: [];
```

But tiny preferences thing, nothing else.

PR #6168, opened 2017-07-28
SecretStore: bunch of fixes and improvements

svyatonik:
on top of #6146

1) improved logging.
2) fixed generation session failure due to network lags.
3) cli option to disable SecretStore HTTP API, so that you can have less than N (in M-of-N scheme) external 'public' nodes, which process client requests. It also decreases startup time (which has been very important in testing) + frees network port in cases where it is not needed to have exactly N public nodes.
4) cli option to disable ACL check. It is extremely useful in testing (previously I had to recompile parity with slightly changed source file to achieve the same effect). However, it could be a security hole, if misused (i.e. used in production). So please argue if you feel this is a bad option.


svyatonik:
`inprogress` as 'parent' #6146 is also `inprogress`

secret_store/src/key_server_cluster/signing_session.rs
```diff
@@ -572,7 +572,7 @@ mod tests {
 	use std::collections::{BTreeMap, VecDeque};
 	use ethkey::{self, Random, Generator, Public};
 	use util::H256;
-	use super::super::super::acl_storage::tests::DummyAclStorage;
+	use super::super::super::acl_storage::DummyAclStorage;
```
debris: `use acl_storage::DummyAclStorage;` should work just fine

secret_store/src/key_server_cluster/decryption_session.rs
```diff
@@ -467,7 +467,7 @@ impl Ord for DecryptionSessionId {
 mod tests {
 	use std::sync::Arc;
 	use std::collections::BTreeMap;
-	use super::super::super::acl_storage::tests::DummyAclStorage;
+	use super::super::super::acl_storage::DummyAclStorage;
```
debris: `use acl_storage::DummyAclStorage;` should work just fine

secret_store/src/http_listener.rs
```diff
@@ -74,19 +74,26 @@ struct KeyServerSharedHttpHandler<T: KeyServer + 'static> {
...
-	pub fn start(listener_address: &NodeAddress, key_server: T) -> Result<Self, Error> {
+	pub fn start(listener_address: Option<NodeAddress>, key_server: T) -> Result<Self, Error> {
 		let shared_handler = Arc::new(KeyServerSharedHttpHandler {
 			key_server: key_server,
 		});
-		let handler = KeyServerHttpHandler {
+		/*let handler = KeyServerHttpHandler {
 			handler: shared_handler.clone(),
-		};
+		};*/
```
debris: it should be removed

debris:
> cli option to disable ACL check. It is extremely useful in testing (previously I had to recompile parity with slightly changed source file to achieve the same effect). However, it could be a security hole, if misused (i.e. used in production). So please argue if you feel this is a bad option.

When starting parity with this option enabled, I think we should at least display warning (with bold yellow/red letters).

svyatonik:
Thanks, for suggestion, will update PR a little bit later

svyatonik:
Fixed style + added warning message

secret_store/src/acl_storage.rs
```diff
@@ -47,6 +48,12 @@ struct CachedContract {
 	contract: Option<SecretStoreAclStorage>,
 }
 
+#[derive(Default, Debug)]
+/// Dummy ACL storage implementation (check always passed).
```
rphmeier: nit: attributes under docs

PR #6173, opened 2017-07-28
Using multiple NTP servers

tomusdrw:
And introducing exponential backoff for the servers if they start to return errors (or KoD packets).

On ice until we get a vendor-specific ntp pool url.


tomusdrw:
ntp servers are now available, ready for review and merge

5chdn:
labeling this for backport.

PR #6177, opened 2017-07-28
Realloc test added

fckt:
Need to update wasm-tests submodule first (have no permissions ATM). Not to be merged until.


fckt:
wasm-tests submodule was updated, tests are ok

PR #6178, opened 2017-07-28
Fix dapps CSP when UI is exposed externally

### LINKED ISSUE #6174: DApps not accessible in 1.7.0
I use this artifact 'https://gitlab.parity.io/parity/parity/-/jobs/61863/artifacts/download to test 1.7.0
I created a single node and open http://$IP:8180 to access DApps.

1. It notifies me that I have a single node and probably there's a connection issue - which is fine, but in chrome console every ~ 1 sec I see these requests ending in errors
```
Request URL:http://$IP:8180/api/health
Request Method:GET
Status Code:412 Precondition Failed
Remote Address:40.71.41.241:8180
Referrer Policy:no-referrer-when-downgrade
```
The response is returned though: `{"peers":{"status":"bad","message":"You are not connected to any peers. ...`
Is this OK?

2. When I open my DApp I get this error:
```
Refused to display 'http://$IP:8540/KeysGenerator/' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 0.0.0.0:8180 *.web3.site:8180 home.web3.site localhost:3000".
```

Here's part of config.toml
```
[rpc]
cors = "all"
interface = "all"
hosts = ["all"]
port = 8540
[ui]
interface = "all"
port = 8180
```

tomusdrw:
Closes #6174

Also contains a fix to dapps development using latest inject.js, cc @kaikun213


rphmeier:
I wonder if this also fixes the issues with DApps that I had on the light client where the page was completely blank.

parity/configuration.rs
```diff
@@ -572,11 +580,26 @@ impl Configuration {
...
-				vec![]
+			extra_embed_on: {
+				let mut extra_embed = dev_ui.clone();
+				match self.ui_hosts() {
+					// In case host validation is disabled allow all frame ancestors
+					None => extra_embed.push(("*".to_owned(), ui_port)),
+					Some(hosts) => extra_embed.extend(hosts.into_iter().filter_map(|host| {
+						let mut it = host.split(":");
+						let host = it.next();
+						let port = it.next().and_then(|v| u16::from_str(v).ok());
```
maciejhirsz: I guess we don't really care if the settings are invalid? (more than one colon, port not being valid uint?)

tomusdrw: Yeah, we always attempt to parse hosts/origins in the best way we can.

PR #6184, opened 2017-07-28
Decouple virtual machines

NikVolf:
- Now there is a common crate vm (/ethcore/vm)
- EVM-specific code in evm crate (/ethcore/evm)
- WASM-specific code in wasm crate (/ethcore/wasm)
- A lot of code now does not use evm at all, depends only on vm (can be further factored out)

No new logic here, just a lot of tree changes


debris:
@NikVolf wasm && evm test s

debris:
they are not run by our CI, I fixed that in #6187 but can't merge the pr when all of them are broken

debris:
fixed in #6132

Issue #6165, opened 2017-07-27
MMAP in ethash DAG used for verification


rphmeier:
Would significantly reduce memory usage in light client.
@vurich one for you?

andrebeat:
Can I take this?

Vurich:
Sorry @andrebeat, [a PR is open already](https://github.com/paritytech/parity/pull/6193).

5chdn:
Please always link issues to PRs. https://github.com/blog/1506-closing-issues-via-pull-requests

PR #6193, opened 2017-07-31
Use memmap for dag cache

Vurich:
I did some refactoring in the process here - I wanted `compute.rs` to ideally know nothing about the precise storage situation of the cache itself. There's obviously some bikeshedding to be done regarding that choice. I also allowed the cache to failover to a memory-backed cache in case the memmap fails. I think that we probably want to have some kind of way to specify that we want to optimize for CPU time over RAM and force the failover to memory-backed cache. I've got a TODO comment with more specifics there.

Originally I had it always build the memory-backed cache and only access the filesystem on `flush`, so that the `new` method didn't have to access the filesystem and return an IO error, but then I realised I could failover instead. This behaviour might be a good solution for "optimize for CPU time", or we could have "optimize for CPU time", "optimize for peak memory" and "balanced", the latter being the behaviour I just described.


ethash/src/cache.rs
```diff
@@ -0,0 +1,282 @@
+// TODO: Mmap failing to build is transparent, we want to add a possibility to use RAM over a mmap
```
rphmeier: needs license header

ethash/src/cache.rs
```diff
@@ -0,0 +1,282 @@
...
+fn cache_path<P: ToOwned<Owned = PathBuf> + ?Sized>(path: &P, ident: &H256) -> PathBuf
+where
+	PathBuf: Borrow<P>,
+{
+	let mut buf = path.to_owned();
+	buf.push(to_hex(ident));
+	buf
+}
+
+fn consume_cache<P: AsRef<Path>>(cache: &mut Cache, path: &P) -> io::Result<()> {
```
rphmeier: there are some leftover `println!`s in this function

ethash/src/cache.rs
```diff
@@ -0,0 +1,282 @@
...
+	let mut file = File::open(path)?;
+
+	let mut nodes: Vec<u8> =
+		Vec::with_capacity(file.metadata().map(|m| m.len() as _).unwrap_or(1_000_000));
+	file.read_to_end(&mut nodes)?;
+	let out: Vec<Node> = unsafe {
+		Vec::from_raw_parts(
+			nodes.as_mut_ptr() as *mut _,
+			nodes.len() / NODE_BYTES,
+			nodes.capacity() / NODE_BYTES,
```
rphmeier: do we know the capacity will always be divisible by `NODE_BYTES`?

Vurich: Good catch, this isn't a safety issue since the pointer is the same and the capacity is always less than or equal to whatever the fractional capacity would be (`free` doesn't care about size, only the pointer) but this should definitely be `assert`'d. We can afford it.

Originally we asserted that the file size was equal to the calculated cache size, but I don't believe it's a safety issue and the size of the file being correct is no sign that the file contains sensical data.

ethash/src/cache.rs
Vurich: Thank you. That was a stupid mistake.

ethash/src/cache.rs
rphmeier: well, it may try to `je_dealloc ` with the wrong size given...AFAIK this can lead to UB or at the very least unexpected program termination.

rphmeier: why not read in the file `NODE_BYTES` at a time, initializing each one as we go?

Vurich: That's a great solution, I think I'll go with that.

ethash/src/shared.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl Clone for Node {
+	fn clone(&self) -> Self {
+		Node { bytes: *&self.bytes }
+	}
+}
+
+impl Node {
+	pub fn as_words(&self) -> &[u32; NODE_WORDS] {
```
NikVolf: Not sure about `safe`-`unsafe` conventions here
If you are unsafely transmuting public member of the structure in the method, shouldn't the method be `unsafe` as well?

rphmeier: intuition about whether a method should be tagged unsafe: when passing data constructed completely in `safe` code can lead to unsafety, then the function would be unsafe. 

In this case, the lifetimes and fact that we are only dealing with raw bytes makes things fine. an assertion that `NODE_WORDS * 4 <= NODE_BYTES` would be good, though.

rphmeier: Actually, this is not correct. 
@Vurich `Node` should be a `#[repr(C)]`, have a `_marker: [u64; 0]` and an assertion at each transmute site that `align_of_val(&self.bytes) == align_of::<[u64/u32; SIZE]>()`

Vurich: Yeah, the way I'd go is that if safe code can not cause unsafety, no matter what data it passes to the function, then then function is safe. Otherwise it's `unsafe`. Whether or not the code inside uses `unsafe` is immaterial. `Vec::set_len` is unsafe despite being 100% "safe" code. As for the alignment issue, I thought that one-field structs were guaranteed to have the same representation as their inner type. The assertion should be `mem::size_of::<[u8; NODE_BYTES]>() == mem::size_of::<[u64; NODE_DWORDS]/[u32; NODE_WORDS]>()`, I guess that's not handled by the transmute since it's transmuting pointers. I should note that if this is a problem then it's a problem in the current version of the software, this is no worse than it (unlike the file reading issue, which is a regression), but that's not a reason not to change it.

Vurich:
Probably worth noting that this is more-or-less unacceptably slow. Probably significantly faster for `from_file` (since the number of copies are massively reduced, it just reads from the page cache) but new caches have suffered an 100% slowdown (on my HDD). Maybe keep it behind a feature gate for now until `OptimizeFor` (or some equivalent) is implemented.

ethash/src/shared.rs
rphmeier: > I thought that one-field structs were guaranteed to have the same representation as their inner type

I have never seen this guarantee, and I don't see why it would hold true. Please link to the place where a core Rust team member states this in a comment above the code.

rphmeier: >  I should note that if this is a problem then it's a problem in the current version of the software

Fair enough, but it's better to fix instead of leave as-is

ethash/src/cache.rs
```diff
@@ -0,0 +1,310 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+// TODO: Mmap failing to build is transparent, we want to add a possibility to use RAM over a mmap
+//       for maximum speed (since the OS needs to zero the file when we do `set_len` - although I
+//       think it can occasionally prevent that from happening - and it has to page the data into
+//       the page cache from disk). I suggest an `OptimizeFor` struct that gets passed into an
```
rphmeier: Could we make this part of the same PR? 

We don't want to suffer any performance regression on the full nodes, while we do want to limit memory usage on the light node.

Vurich: It would delay the PR more than it already has been, but I would be OK with doing that.

rphmeier: There has to be a way to set it "on" on light clients and "off" on full nodes anyway, so it might as well go in the

Vurich: Fair enough, I'll get on it.

Vurich:
Now `./test.sh` works again and it will automatically use a mmap for the light client.

ethash/src/cache.rs
gavofyork: fyi asserts should never fire, particularly not if the data coming from the user or filesystem happens to be "wrong". in this case, use error returns and bomb out cleanly if it's not trivially fixable/recoverable.

asserts are only there to encode logical tautologies and should always be accompanied with an explanation as to why they will *never* fire.

Vurich: I'll change this to an error return and replace the `assert!`s with `debug_assert!`s

ethcore/src/spec/spec.rs
```diff
@@ -240,7 +278,11 @@ fn load_from<T: AsRef<Path>>(cache_dir: T, s: ethjson::spec::Spec) -> Result<Spe
 		timestamp: g.timestamp,
 		extra_data: g.extra_data,
 		seal_rlp: seal_rlp,
-		constructors: s.accounts.constructors().into_iter().map(|(a, c)| (a.into(), c.into())).collect(),
+		constructors: s.accounts
+		               .constructors()
```
Vurich: [This is "smart tabs"](https://www.emacswiki.org/emacs/SmartTabs), it's just how it gets laid out automatically for me. I'll change it so the default is closer to our style.

5chdn:
Closes #6165

ethash/src/shared.rs
```diff
@@ -0,0 +1,119 @@
...
+		sz = sz - 2 * ETHASH_MIX_BYTES as u64;
+	}
+	sz as usize
+}
+
+pub type NodeBytes = [u8; NODE_BYTES];
+pub type NodeWords = [u32; NODE_WORDS];
+pub type NodeDwords = [u64; NODE_DWORDS];
+
+pub union Node {
```
rphmeier: please add a test that `::std::mem::size_of::<Node>() == NODE_BYTES`, since some code depends on it. Also, does it need to be `#[repr(C)]`? If not, please add a comment describing why, since it's not immediately obvious.

ethash/src/shared.rs
```diff
@@ -0,0 +1,119 @@
...
+
+static CHARS: &'static [u8] = b"0123456789abcdef";
+pub fn to_hex(bytes: &[u8]) -> String {
+	let mut v = Vec::with_capacity(bytes.len() * 2);
+	for &byte in bytes.iter() {
+		v.push(CHARS[(byte >> 4) as usize]);
+		v.push(CHARS[(byte & 0xf) as usize]);
+	}
+
+	unsafe { String::from_utf8_unchecked(v) }
```
rphmeier: given that this isn't a bottleneck, can we just have it be checked? i would like to have the unsafe code here reduced to what is absolutely necessary.

ethash/src/seed_compute.rs
```diff
@@ -0,0 +1,103 @@
...
+	pub fn resume_compute_seedhash(mut hash: H256, start_epoch: u64, end_epoch: u64) -> H256 {
+		for _ in start_epoch..end_epoch {
+			sha3_256::inplace(&mut hash);
+		}
+		hash
+	}
+}
+
+#[test]
+fn test_seed_compute_once() {
```
rphmeier: tests in their own module

ethash/src/cache.rs
rphmeier: does it actually need to return an error? just round up the capacity to nearest multiple of `NODE_BYTES` and zero-extend. doesn't matter that it'll break the calculations, but would be easier than passing an error all the way up.

ethash/src/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+
+	Ok(memmap)
+}
+
+fn make_memory_cache(num_nodes: usize, ident: &H256) -> Vec<Node> {
+	let mut nodes: Vec<Node> = Vec::with_capacity(num_nodes);
+	// Use uninit instead of unnecessarily writing `size_of::<Node>() * num_nodes` 0s
+	unsafe {
+		nodes.set_len(num_nodes);
+		initialize_memory(nodes.as_mut_ptr(), num_nodes, ident);
```
rphmeier: out of curiosity, what would happen if `initialize_memory` panicked before writing in all the node data? would the destructor for `Vec<Node>` try and set drop flags, ending up writing uninit memory?

rphmeier:
Other than unsafety questions, clarifications, LGTM. Please address those, and then we can merge (once @arkpar gives it a review as well, since this is _extremely_ sensitive code)

ethash/src/cache.rs
Vurich: Even if `Node` contained `Drop` types (it doesn't), `union`s inhibit destructors.

ethash/src/shared.rs
Vurich: This is existing code, I just moved it. I wouldn't agree that having this be unchecked is a problem, since it's absolutely clear that it is impossible to write non-ascii data here. I understand your concern about unsafety, but I would disagree that it needs changing. If you feel very strongly about this or if another reviewer also thinks this needs fixing then I'll accept being outvoted.

ethash/src/cache.rs
Vurich: Actually, I guess that it's possible that `Vec`'s destructor would read its elements. I don't think it does, but there's nothing stopping it from doing so. The right thing to do here is to add `set_len` after the call to `initialize_memory`

ethash/src/cache.rs
```diff
@@ -0,0 +1,327 @@
...
+			let buf = unsafe {
+				slice::from_raw_parts_mut(vec.as_mut_ptr() as *mut u8, vec.len() * NODE_BYTES)
+			};
+
+			// Return early if this fails. We only use this for writing to the memmap. I would do
+			// this more explicitly but I need to return it from this scope to fix lifetime issues.
+			// If you're seeing this comment in the future when we have non-lexical lifetimes then
+			// refactor this to use a `match` or something, like below. Maybe you even have monads,
+			// future time-traveller, that would make this significantly cleaner.
+			file.write(buf).map(|_| ())?;
```
andrebeat: Shouldn't this be a `write_all` instead?

rphmeier: nice catch!

ethash/src/shared.rs
Vurich: It does need to be `#[repr(C)]`, good catch. I've never used untagged unions before, it's quite hard to find information on them. I think in general it's invalid to use non-`#[repr(C)]` unions to reinterpret data.

ethash/src/compute.rs
```diff
@@ -467,159 +315,310 @@ fn calculate_dag_item(node_index: u32, cache: &[Node]) -> Node {
...
 
-static CHARS: &'static [u8] = b"0123456789abcdef";
-fn to_hex(bytes: &[u8]) -> String {
-	let mut v = Vec::with_capacity(bytes.len() * 2);
-	for &byte in bytes.iter() {
-		v.push(CHARS[(byte >> 4) as usize]);
-		v.push(CHARS[(byte & 0xf) as usize]);
+	#[test]
+	fn test_difficulty_test() {
+		let hash = [
```
arkpar: This kind of formatting does not really make the code more readable. Could you please revert it?

ethash/src/cache.rs
```diff
@@ -0,0 +1,320 @@
...
+		&self.cache_path
+	}
+
+	pub fn flush(&mut self) -> io::Result<()> {
+		if let Some(last) = self.epoch.checked_sub(2).map(|ep| {
+			cache_path(self.cache_dir.as_ref(), &self.builder.epoch_to_ident(ep))
+		})
+		{
+			fs::remove_file(last)
+				.unwrap_or_else(|error| warn!("Error removing stale DAG cache: {:?}", error));
```
arkpar: Will there be a warning printed if the file does not exist? Is such case it is better to silence it.

arkpar:
lgtm except for a couple of minor grumbles

ethash/src/compute.rs
Vurich: Sorry, I use rustfmt because I don't like manually formatting. I didn't mean to commit this at all. The next version of rustfmt will support the visual block formatting that we had before (so first element on the next line and tabbed in, consecutive elements on same line). Thanks for catching this.

ethash/src/cache.rs
Vurich: The error is being passed all the way up anyway if, for example, the filesystem is corrupted (i.e. `File::open` or `File::read_to_end` fail). In the case that the file is the wrong length it will attempt to regenerate it, same as if it doesn't exist.

Vurich:
Can anyone rereview once the CI's finished?

PR #6194, opened 2017-07-31
Rewards on closing blocks

grbIzl:
Added tracing of rewards on closing blocks. The following major changes made:

- Add new action Reward with reward type (Block or Uncle)
- Extended engine's API in order to return traces on closing block.
- New parameter for engine's spec params added (apply_reward). By default it's true
- New engine spec null_morden_with_reward created (in order to distinguish with default null_morden without reward)
- Db code of trace filtering was modified in order to reflect, that traces can be block-related not only transaction-related
- Localisation trace (passed outside by rpc) modified. Transaction id and hash in trace made optional. So it can be null now. This should be expressed in release notes somehow.
- Integration and unit tests added

The work was done in terms of https://github.com/paritytech/parity/issues/4858


NikVolf:
needs a rebase to master

ethcore/src/spec/spec.rs
```diff
@@ -166,9 +168,17 @@ impl From<ethjson::spec::Params> for CommonParams {
 			nonce_cap_increment: p.nonce_cap_increment.map_or(64, Into::into),
 			remove_dust_contracts: p.remove_dust_contracts.unwrap_or(false),
 			wasm: p.wasm.unwrap_or(false),
+<<<<<<< HEAD
```
keorn: Leftovers

ethcore/src/block.rs
```diff
@@ -196,6 +204,9 @@ pub trait IsBlock {
 
 	/// Get all uncles in this block.
 	fn uncles(&self) -> &[Header] { &self.block().uncles }
+
+	/// Get tracing enabled flag for this block.
+	fn tracing_enabled(&self) -> &bool { &self.block().tracing_enabled }
```
rphmeier: could just return `bool`, it's actually smaller than the ref type

ethcore/src/engines/null_engine.rs
```diff
@@ -17,10 +17,15 @@
 use std::collections::BTreeMap;
 use util::Address;
 use builtin::Builtin;
-use engines::Engine;
+use block::*;
+use util::*;
+use engines::{Engine, CloseOutcome};
```
rphmeier: Better not to use glob imports where avoidable

ethcore/src/ethereum/ethash.rs
```diff
@@ -271,40 +272,57 @@ impl Engine for Arc<Ethash> {
...
 		for u in fields.uncles.iter() {
+			let uncle_miner = u.author().clone();
+			let result_uncle_reward: U256;
+
 			if eras == 0 {
+				result_uncle_reward = reward * U256::from(8 + u.number() - current_number) / U256::from(8);
 				fields.state.add_balance(
 					u.author(),
-					&(reward * U256::from(8 + u.number() - current_number) / U256::from(8)),
+					&(result_uncle_reward),
```
rphmeier: parentheses no longer necessary

grbIzl:
Add relnotes label in order to reflect, that transaction id and hash for trace can be null after this change

ethcore/src/block.rs
```diff
@@ -429,7 +453,7 @@ impl<'x> OpenBlock<'x> {
 			s.block.header.set_transactions_root(ordered_trie_root(s.block.transactions.iter().map(|e| e.rlp_bytes().into_vec())));
 		}
 		let uncle_bytes = s.block.uncles.iter().fold(RlpStream::new_list(s.block.uncles.len()), |mut s, u| {s.append_raw(&u.rlp(Seal::With), 1); s} ).out();
-		if s.block.header.uncles_hash().is_zero() {
+		if s.block.header.uncles_hash().is_zero() || s.block.header.receipts_root() == &SHA3_NULL_RLP {
```
tomusdrw: Why is this necessary and related?

ethcore/src/spec/spec.rs
```diff
@@ -98,6 +98,8 @@ pub struct CommonParams {
 	pub block_reward: U256,
 	/// Registrar contract address.
 	pub registrar: Address,
+	/// Apply reward
+	pub apply_reward: bool,
```
tomusdrw: IMHO this deserves a bit more description.

ethcore/src/trace/db.rs
```diff
@@ -363,17 +375,29 @@ impl<T> TraceDatabase for TraceDB<T> where T: DatabaseExtras {
 						.map(Into::<Vec<FlatTrace>>::into)
 						.enumerate()
 						.flat_map(|(tx_position, traces)| {
-							let tx_hash = self.extras.transaction_hash(block_number, tx_position)
-								.expect("Expected to find transaction hash. Database is probably corrupted");
+							let trace_tx_number;
+							let trace_tx_hash;
+
+							match self.extras.transaction_hash(block_number, tx_position) {
```
tomusdrw: IMHO:
```rust
let (trace_tx_number, trace_tx_hash) = match ... {
  Some(hash) => (Some(tx_position), Some(hash.clone())),
  ...
}
```
looks nicer.

ethcore/src/block.rs
```diff
@@ -137,6 +142,7 @@ impl ExecutedBlock {
 			transactions_set: Default::default(),
 			state: state,
 			traces: if tracing {Some(Vec::new())} else {None},
+			tracing_enabled: tracing,
```
tomusdrw: Can't we just check if `trace.is_some()` instead of storing it here?

ethcore/src/engines/null_engine.rs
```diff
@@ -64,4 +69,42 @@ impl Engine for NullEngine {
...
+			fields.state.add_balance(
+				u.author(),
+				&(result_uncle_reward),
+				CleanupMode::NoEmpty
+			)?;
+			tracer.trace_reward(uncle_miner, result_uncle_reward, RewardType::Uncle);
+		}
+
+		fields.state.commit()?;
+		match *fields.tracing_enabled {
```
tomusdrw: Why not alter the block traces directly here? We modify the state anyway?

ethcore/src/spec/spec.rs
arkpar: Why is this even needed? Can't we just assume `apply_reward == block_reward != 0`?

ethcore/src/block.rs
grbIzl: Yes, we can. I thought about it. But IMHO relying on the explicit flag is better than basing on the assumption. The logic of working with traces can be changed and this assumption may become false

ethcore/src/block.rs
grbIzl: I incorrectly fixed the bug with this check. Thanks! I missed it (((

ethcore/src/block.rs
arkpar: You could just add a member function that would return the flag instead of introducing another state variable.

ethcore/src/block.rs
```diff
@@ -392,9 +403,16 @@ impl<'x> OpenBlock<'x> {
 
 		let unclosed_state = s.block.state.clone();
 
-		if let Err(e) = s.engine.on_close_block(&mut s.block) {
-			warn!("Encountered error on closing the block: {}", e);
+		match s.engine.on_close_block(&mut s.block) {
+			Ok(outcome) => match outcome.trace {
```
arkpar: `if let(Some(t)) = outcome.trace` could be used here

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -543,18 +544,30 @@ impl Engine for AuthorityRound {
...
 		let fields = block.fields_mut();
+		let mut tracer = ExecutiveTracer::default();
 		// Bestow block reward
 		let reward = self.params().block_reward;
 		let res = fields.state.add_balance(fields.header.author(), &reward, CleanupMode::NoEmpty)
 			.map_err(::error::Error::from)
 			.and_then(|_| fields.state.commit());
+
+		// Trace it
+		let block_miner = fields.header.author().clone();
```
arkpar: `block_author` is a more generic name for the miner

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -538,18 +539,31 @@ impl Engine for Tendermint {
...
+
 		// Commit state so that we can actually figure out the state root.
 		if let Err(ref e) = res {
 			warn!("Encountered error on closing block: {}", e);
 		}
-		res
+		match res {
+			Ok(_) => match *fields.tracing_enabled {
+				true => Ok(CloseOutcome{trace: Some(tracer.traces())}),
+				false => Ok(CloseOutcome{trace: None})
```
arkpar: spaces around `{` here and in other places

ethcore/src/spec/spec.rs
grbIzl: Also an assumption. I agree, that not zero bock_reward without its application doesn't make much sense. But is it safe to rely on such things?

ethcore/src/trace/executive_tracer.rs
```diff
@@ -162,6 +162,21 @@ impl Tracer for ExecutiveTracer {
...
+			subtraces: 0,
+			action: Action::Reward(Reward {
+				author: author,
+				value: value,
+				reward_type: reward_type,
+			}),
+			result: Res::None,
+			trace_address: Default::default(),
+		};
+		debug!(target: "trace", "Traced failed reward {:?}", trace);
```
arkpar: Why "failed"?

ethcore/src/trace/executive_tracer.rs
```diff
@@ -162,6 +162,21 @@ impl Tracer for ExecutiveTracer {
 		debug!(target: "trace", "Traced failed suicide {:?}", trace);
```
arkpar: "failed" also looks wrong here

ethcore/src/spec/spec.rs
arkpar: IMO it is ok and keeps spec structure simple. Having `block_reward` and `apply_reward` in the spec just adds more confusion. You can make a 
`pub fn apply_reward(&self) -> bool` in the code and use that everywhere in case the logic changes.

ethcore/src/trace/executive_tracer.rs
grbIzl: Doubled copy-paste :-)

ethcore/src/spec/spec.rs
rphmeier: Yeah, it should be in the `engines::common` module for code reuse. But not part of the `Engine` trait itself!

ethcore/src/engines/null_engine.rs
grbIzl: I don't know the reasoning for this but traces are not mutable from outside. Actually I've repeated the logic made for tracing transactions - traces are added inside block already

rphmeier:
As mentioned in a now hidden diff comment, would it be possible to create a common function for the duplicated code here? It will also making "state-machine" refactoring easier in the future.

ethcore/src/engines/null_engine.rs
tomusdrw: I think it's because `Executive` operates only on state and transaction not on full blocks.
IMHO it would be cleaner to actually alter the traces directly here.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -545,18 +546,32 @@ impl Engine for AuthorityRound {
 	}
 
 	/// Apply the block reward on finalisation of the block.
-	fn on_close_block(&self, block: &mut ExecutedBlock) -> Result<(), Error> {
+	fn on_close_block(&self, block: &mut ExecutedBlock) -> Result<CloseOutcome, Error> {
+		let tracing_enabled = block.tracing_enabled();
 		let fields = block.fields_mut();
+		let mut tracer = ExecutiveTracer::default();
```
tomusdrw: nitpick: The tracing happens in `Engine` not `Executive` so we could have a separate struct for tracing the rewards.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -545,18 +546,32 @@ impl Engine for AuthorityRound {
...
 			.and_then(|_| fields.state.commit());
+
+		if tracing_enabled {
+			let block_author = fields.header.author().clone();
+			tracer.trace_reward(block_author, self.params().block_reward, RewardType::Block);
+		}
+
 		// Commit state so that we can actually figure out the state root.
 		if let Err(ref e) = res {
 			warn!("Encountered error on closing block: {}", e);
```
tomusdrw: This could be inside `match` block, but if we decide to add traces directly it won't be needed.

ethcore/src/ethereum/ethash.rs
```diff
@@ -274,40 +275,61 @@ impl Engine for Arc<Ethash> {
...
+	fn on_close_block(&self, block: &mut ExecutedBlock) -> Result<CloseOutcome, Error> {
 		let reward = self.params().block_reward;
+		let tracing_enabled = block.tracing_enabled();
 		let fields = block.fields_mut();
 		let eras_rounds = self.ethash_params.ecip1017_era_rounds;
 		let (eras, reward) = ecip1017_eras_block_reward(eras_rounds, reward, fields.header.number());
+		let mut tracer = ExecutiveTracer::default();
 
 		// Bestow block reward
+		let result_block_reward = reward + reward / U256::from(32) * U256::from(fields.uncles.len());
```
tomusdrw: perf: Could use shift instead of division here

ethcore/src/ethereum/ethash.rs
```diff
@@ -274,40 +275,61 @@ impl Engine for Arc<Ethash> {
...
+		}
+
 		// Bestow uncle rewards
 		let current_number = fields.header.number();
 		for u in fields.uncles.iter() {
+			let uncle_author = u.author().clone();
+			let result_uncle_reward: U256;
+
 			if eras == 0 {
+				result_uncle_reward = reward * U256::from(8 + u.number() - current_number) / U256::from(8);
```
tomusdrw: Same here...

ethcore/src/ethereum/ethash.rs
```diff
@@ -274,40 +275,61 @@ impl Engine for Arc<Ethash> {
...
 				fields.state.add_balance(
 					u.author(),
-					&(reward * U256::from(8 + u.number() - current_number) / U256::from(8)),
+					&result_uncle_reward,
 					CleanupMode::NoEmpty
-				)
+				)	
 			} else {
+				result_uncle_reward = reward / U256::from(32);
```
tomusdrw: ...and here.

ethcore/src/trace/executive_tracer.rs
```diff
@@ -159,7 +159,22 @@ impl Tracer for ExecutiveTracer {
...
-		debug!(target: "trace", "Traced failed suicide {:?}", trace);
+		debug!(target: "trace", "Traced suicide {:?}", trace);
+		self.traces.push(trace);
+	}
+	
+	fn trace_reward(&mut self, author: Address, value: U256, reward_type: RewardType) {
+		let trace = FlatTrace {
+			subtraces: 0,
+			action: Action::Reward(Reward {
+				author: author,
```
tomusdrw: You can use rust1.19 shorthads:
```
Reward { author, value, reward_type }
```

rpc/src/v1/types/trace.rs
```diff
@@ -299,6 +299,53 @@ impl From<trace::Call> for Call {
 	}
 }
 
+/// Reward type.
+#[derive(Debug, Serialize)]
+pub enum RewardType {
+	/// None
+	#[serde(rename="none")]
+	None,
```
tomusdrw: Why do we need this?

grbIzl:
@rphmeier  what code do you mean? I assume, that you're speaking about tracing status in block. I've removed the flag and left only method with logic. But I don't remember any questions about code duplication.

grbIzl:
After the discussion with @tomusdrw and @debris access to the traces made mutable, and the work with it moved from block to the engine itself.
Also other comments were fixed.

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -541,12 +542,21 @@ impl Engine for Tendermint {
...
+		let tracing_enabled = block.tracing_enabled();
 		let fields = block.fields_mut();
+		let mut tracer = ExecutiveTracer::default();
 		// Bestow block reward
 		let reward = self.params().block_reward;
 		let res = fields.state.add_balance(fields.header.author(), &reward, CleanupMode::NoEmpty)
 			.map_err(::error::Error::from)
 			.and_then(|_| fields.state.commit());
+
+		if tracing_enabled {
```
rphmeier: @grbIzl This block here is exactly the same for authority round and tendermint, and probably for any new engine we would add in the future. It's also not really the concern of the consensus engine to give out block rewards, so we plan to split that kind of thing out as well. Moving this to a function in `engines::common` module would help us keep the code cleaner and assist in the refactor in the future.

grbIzl: Got it! Thanks for explanation. Will try to create a suitable solution for this.

ethcore/src/engines/mod.rs
```diff
@@ -477,4 +470,28 @@ pub mod common {
 		}
 		Ok(())
 	}
+
+	/// Trace rewards on closing block
+	pub fn bestow_block_reward<E: Engine + ?Sized>(block: &mut ExecutedBlock, engine: &E) -> Result<(), Error> {
+		let tracing_enabled = block.tracing_enabled();
+		let fields = block.fields_mut();
+		let mut tracer = ExecutiveTracer::default();
```
tomusdrw: `tracer` can be moved inside `if`

ethcore/src/engines/mod.rs
```diff
@@ -477,4 +470,28 @@ pub mod common {
...
+		// Bestow block reward
+		let reward = engine.params().block_reward;
+		let res = fields.state.add_balance(fields.header.author(), &reward, CleanupMode::NoEmpty)
+			.map_err(::error::Error::from)
+			.and_then(|_| fields.state.commit());
+
+		if tracing_enabled {
```
tomusdrw: I would inline the variable here `block.tracing_enabled()`
Actually it could even look like this:
```rust
fields.traces.as_mut().map(|mut traces| {
  let mut tracer = ExecutiveTracer::new();
  tracer.trace_reward(...);
  traces.push(tracer.traces())
})
```
and neither the if statement nor the `tracing_enabled` function in block are needed in such case.

ethcore/src/engines/null_engine.rs
```diff
@@ -105,9 +106,9 @@ impl Engine for NullEngine {
 		}
 
 		fields.state.commit()?;
-		match tracing_enabled {
```
tomusdrw: Actually it would be better to avoid collecting traces at all if tracing is disabled, not only discard them afterwise.

ethcore/src/ethereum/ethash.rs
```diff
@@ -326,10 +327,10 @@ impl Engine for Arc<Ethash> {
 
 		// Commit state so that we can actually figure out the state root.
 		fields.state.commit()?;
-		match tracing_enabled {
-			true => Ok(CloseOutcome { trace: Some(tracer.traces()) } ),
-			false => Ok(CloseOutcome { trace: None } )
-		}		
+		if tracing_enabled {
```
tomusdrw: Same here (avoid to produce traces when tracing is disabled)

grbIzl: Are you sure? In this case it seems to be covered with if.

tomusdrw: Oh, you're right sorry!

ethcore/src/block.rs
```diff
@@ -91,7 +91,7 @@ pub struct ExecutedBlock {
 	receipts: Vec<Receipt>,
 	transactions_set: HashSet<H256>,
 	state: State<StateDB>,
-	traces: Option<Vec<Vec<FlatTrace>>>,
```
gavofyork: style: don't kill the trailing `,`

ethcore/src/block.rs
```diff
@@ -393,8 +396,9 @@ impl<'x> OpenBlock<'x> {
 		let unclosed_state = s.block.state.clone();
 
 		if let Err(e) = s.engine.on_close_block(&mut s.block) {
-			warn!("Encountered error on closing the block: {}", e);
+ 			warn!("Encountered error on closing the block: {}", e);
```
gavofyork: dodgy space added

ethcore/src/block.rs
```diff
@@ -419,7 +423,7 @@ impl<'x> OpenBlock<'x> {
 		let mut s = self;
 
 		if let Err(e) = s.engine.on_close_block(&mut s.block) {
-			warn!("Encountered error on closing the block: {}", e);
+ 			warn!("Encountered error on closing the block: {}", e);
```
gavofyork: dodgy space added (beginning of line)

ethcore/src/block.rs
```diff
@@ -429,7 +433,7 @@ impl<'x> OpenBlock<'x> {
 			s.block.header.set_transactions_root(ordered_trie_root(s.block.transactions.iter().map(|e| e.rlp_bytes().into_vec())));
 		}
 		let uncle_bytes = s.block.uncles.iter().fold(RlpStream::new_list(s.block.uncles.len()), |mut s, u| {s.append_raw(&u.rlp(Seal::With), 1); s} ).out();
-		if s.block.header.uncles_hash().is_zero() {
+		if s.block.header.uncles_hash().is_zero() || s.block.header.uncles_hash() == &SHA3_EMPTY_LIST_RLP {
```
gavofyork: any particular reason for this change?

ethcore/src/engines/mod.rs
```diff
@@ -229,7 +229,7 @@ pub trait Engine : Sync + Send {
 
 	/// Block transformation functions, after the transactions.
 	fn on_close_block(&self, _block: &mut ExecutedBlock) -> Result<(), Error> {
-		Ok(())
+ 		Ok(())
```
gavofyork: dodgy space added

ethcore/src/ethereum/ethash.rs
```diff
@@ -24,6 +24,7 @@ use block::*;
 use builtin::Builtin;
 use vm::EnvInfo;
 use error::{BlockError, Error, TransactionError};
+use trace::{Tracer, ExecutiveTracer,RewardType};
```
gavofyork: inconsistent spaces.

ethcore/src/ethereum/ethash.rs
```diff
@@ -275,38 +276,60 @@ impl Engine for Arc<Ethash> {
...
+
 			if eras == 0 {
+				result_uncle_reward = (reward * U256::from(8 + u.number() - current_number)).shr(3);
 				fields.state.add_balance(
 					u.author(),
-					&(reward * U256::from(8 + u.number() - current_number) / U256::from(8)),
+					&result_uncle_reward,
 					CleanupMode::NoEmpty
-				)
+				)
```
gavofyork: trailing whitespace

ethcore/src/spec/spec.rs
```diff
@@ -170,7 +170,7 @@ impl From<ethjson::spec::Params> for CommonParams {
 			wasm: p.wasm.unwrap_or(false),
 			gas_limit_bound_divisor: p.gas_limit_bound_divisor.into(),
 			block_reward: p.block_reward.map_or_else(U256::zero, Into::into),
-			registrar: p.registrar.map_or_else(Address::new, Into::into),
+			registrar: p.registrar.map_or_else(Address::new, Into::into)
```
gavofyork: don't kill the trailing `,`

ethcore/src/tests/trace.rs
```diff
@@ -0,0 +1,206 @@
...
+	let mut last_hashes = vec![];
+	let mut last_header = genesis_header.clone();
+	last_hashes.push(last_header.hash());
+
+	let kp = KeyPair::from_secret_slice(&"".sha3()).unwrap();
+	let author = kp.address();
+
+	// Add root block first
+	let mut root_block = OpenBlock::new(
+			engine,
```
gavofyork: only one indentation needed

ethcore/src/tests/trace.rs
```diff
@@ -0,0 +1,206 @@
...
+
+	last_header = BlockView::new(&root_block.rlp_bytes()).header();
+	let root_header = last_header.clone();
+	db = root_block.drain();
+
+	last_hashes.push(last_header.hash());
+
+	// Add parent block
+	let mut parent_block = OpenBlock::new(
+			engine,
```
gavofyork: only one indentation needed

ethcore/src/tests/trace.rs
```diff
@@ -0,0 +1,206 @@
...
+		Default::default(), 
+		true, 
+		db, 
+		&last_header, 
+		Arc::new(last_hashes.clone()),
+		author.clone(),
+		(3141562.into(), 31415620.into()), 
+		vec![],
+		false
+		).unwrap();
```
gavofyork: should be one indent fewer

ethcore/src/tests/trace.rs
```diff
@@ -0,0 +1,206 @@
...
+		panic!("error importing block: {:#?}", res.err().unwrap());        
+	}
+
+	block.drain();
+	client.flush_queue();
+	client.import_verified_blocks();
+
+	// Test0. Check overall filter
+	let filter = TraceFilter {
+			range: (BlockId::Number(1)..BlockId::Number(3)),
```
gavofyork: should be one indentation fewer

ethcore/src/trace/types/filter.rs
```diff
@@ -128,6 +128,10 @@ impl Filter {
 				let from_matches = self.from_address.matches(&suicide.address);
 				let to_matches = self.to_address.matches(&suicide.refund_address);
 				from_matches && to_matches
+			},
+			Action::Reward(ref reward) => {
+				let to_matches = self.to_address.matches(&reward.author);
```
gavofyork: what's the point of `let X = ...; X` ? why not just `...`?

ethcore/src/trace/types/flat.rs
```diff
@@ -162,8 +162,9 @@ impl Into<Vec<FlatTransactionTraces>> for FlatBlockTraces {
 mod tests {
 	use rlp::*;
 	use super::{FlatBlockTraces, FlatTransactionTraces, FlatTrace};
-	use trace::trace::{Action, Res, CallResult, Call, Suicide};
+	use trace::trace::{Action, Res, CallResult, Call, Suicide, Reward};
 	use evm::CallType;
+    use trace::RewardType;
```
gavofyork: tabs! not spaces.

gavofyork:
more just did a style review. quite a lot of issues. some were probably just misconfigured editor, others misunderstanding about our house style.

one or two, however, were carelessness, which is a bad omen given this is utterly consensus-critical code that you're hacking on.

the logic here should be positively and deeply reviewed by at least three experienced core devs before merge.

ethcore/src/block.rs
grbIzl: Yes. From my point of view it's a logical flaw. If block is initialized using default construction, uncle_hash won't be zero and won't be set in this method. I assume, that original intent was to set it in this case. You can see the same pattern (checking default values and zero) in this function for other fields. As a result simple integration test with close_and_seal for block failed.

grbIzl:
@gavofyork I provided the reasoning in the question about logic's change. All other style comments will be fixed with the next commit.

ethcore/src/block.rs
gavofyork: ok that's fair enough.

debris:
I will fix build failures and merge to master

grbIzl:
Thanks @debris !

debris:
Done. I'll merge once CI is green 👍

PR #6195, opened 2017-07-31
Multi-call RPC

tomusdrw:
Removes duplicated state-touching methods in miner, now client can forward all pending-block queries correctly if pending state is available and otherwise falls back to `latest` state.

Introduces `parity_call` and `trace_callMany` allowing to run many non-persistent, but dependent calls on a state (for instance deploy contract + query a function from that contract).

CC @kaikun213 

@paritytech/core-devs Let me know if you prefer to review miner-client refactoring separetly from the feature I can split the PRs.


rpc/src/v1/impls/light/parity.rs
```diff
@@ -389,4 +389,8 @@ impl Parity for ParityClient {
 	fn ipfs_cid(&self, content: Bytes) -> Result<String, Error> {
 		ipfs::cid(content)
 	}
+
+	fn call(&self, _requests: Vec<CallRequest>, _block: Trailing<BlockNumber>) -> BoxFuture<Vec<Bytes>, Error> {
+		future::err(errors::light_unimplemented(None)).boxed()
```
rphmeier: implementation of this function will be tricky as discussed w/ @tomusdrw

tomusdrw: Logged #6202

ethcore/src/client/client.rs
```diff
@@ -907,7 +907,7 @@ impl Client {
 	pub fn state_at(&self, id: BlockId) -> Option<State<StateDB>> {
 		// fast path for latest state.
 		match id.clone() {
-			BlockId::Pending => return self.miner.pending_state().or_else(|| Some(self.state())),
+			BlockId::Pending => return self.miner.pending_state(self.chain.read().best_block_number()).or_else(|| Some(self.state())),
```
arkpar: Why is block number required here? Does not the miner know the best pending state already?

tomusdrw: The issue is that the pending state might be stalled, so we check that by comparing block numbers.
We had a similar thing earlier (at least for RPC), but the behaviour was very inconsistent - some stuff was going via the client and some via RPC.

PR #6197, opened 2017-07-31
Chainspec validation

debris:
changes:
- new executable which verifies chainspec json files
- `scrips/validate_chainspec.sh` which validates our chainspecs
- fixed invalid chainspecs

`validate_chainspec.sh` output looks like this:

```
ethcore/res/authority_round.json is valid
ethcore/res/basic_authority.json is valid
ethcore/res/constructor.json missing field `gasLimitBoundDivisor` at line 11 column 2
ethcore/res/instant_seal.json is valid
ethcore/res/null.json is valid
ethcore/res/null_morden.json is valid
ethcore/res/tendermint.json is valid
ethcore/res/validator_contract.json is valid
ethcore/res/validator_multi.json is valid
ethcore/res/validator_safe_contract.json unexpected field `engine.?.params.gasLimitBoundDivisor`
ethcore/res/validator_safe_contract.json unexpected field `engine.?.params.gasLimitBoundDivisor2`
ethcore/res/ethereum/classic.json is valid
ethcore/res/ethereum/eip150_test.json is valid
ethcore/res/ethereum/eip161_test.json is valid
ethcore/res/ethereum/expanse.json is valid
ethcore/res/ethereum/foundation.json is valid
ethcore/res/ethereum/frontier_like_test.json is valid
ethcore/res/ethereum/frontier_test.json is valid
ethcore/res/ethereum/homestead_test.json is valid
ethcore/res/ethereum/kovan.json is valid
ethcore/res/ethereum/metropolis_test.json is valid
ethcore/res/ethereum/morden.json is valid
ethcore/res/ethereum/olympic.json is valid
ethcore/res/ethereum/ropsten.json is valid
ethcore/res/ethereum/transition_test.json is valid
```


debris:
According to gitlab, rust job succeeded

scripts/validate_chainspecs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+ERR=0
+cargo build --release -p chainspec
+
+for spec in `ls ethcore/res/*.json`; do
```
Vurich: Unnecessary `ls`, just do `for spec in path/to/dir/*; do`. Iterating over `ls` output is unstable anyway. Also, backticks are deprecated and also smell and are bad.

Hey, instead of me telling you this stuff, go install [Shellcheck](https://github.com/koalaman/shellcheck/) since I think both of those notes are implemented as lints.

debris: thanks! I will check it out :D

chainspec/src/main.rs
```diff
@@ -0,0 +1,48 @@
+extern crate serde_json;
+extern crate serde_ignored;
+extern crate ethjson;
+
+use std::collections::BTreeSet;
+use std::{fs, env, process};
+use ethjson::spec::Spec;
+
+fn quit<S: AsRef<str>>(s: S) -> ! {
```
Vurich: I think that `Foo: AsRef<Bar>` as a type parameter is unnecessary because of Rust's automatic coercion (for example, `foo(&bar)` will succeed if `foo` takes a `&T` and `bar` implements `AsRef<T>`). I think I've made this same mistake in my latest PR with `AsRef<Path>`. So this could just take `&str` with the same API.

arkpar:
But we already have a unit test that does that. See https://github.com/paritytech/parity/pull/6028

debris:
So there are few reasons to actually have `chainspec` executable:

1. currently there is no way to validate a json file which is not a part of parity repo. I needed this
2. unused fields are not checked 
3. currently not every json file in `ethcore/res/` and `ethcore/res/ethereum/` is checked and we have to remember to update the test if we add new spec

2 && 3 can be addressed with improved unit tests, but 1 not

debris:
@arkpar I removed existing tests

rphmeier:
Agreed with @debris on this -- the broken chainspecs in the repo were all not covered by the tests when I made my PR.

PR #6208, opened 2017-08-01
removed unused code from util/path

debris:


debris:
will reopen when issues are fixed

PR #6214, opened 2017-08-02
Transaction queue reservations

sjeohp:
Parallelizes transaction verification to help with race condition from [6096](https://github.com/paritytech/parity/issues/6096).


tomusdrw:
Just out of curiosity, why can't we have an atomic nonce update when importing local transactions that require nonce (it could be even managed outside of txqueue lock)?

I looked at the code, but don't really see how this solves the nonce issue, it seems that to make a reservation I need a transaction hash and to compute the hash I need the nonce.

sjeohp:
My understanding was that multiple threads can try to add the same transactions but at the moment can only verify serially.  So this just parallelizes it by reserving the unverified hash and moving verification outside the lock.

`Error importing transaction: Transaction(AlreadyImported)` as far as I can tell is an indication of that serial verification/adding race and is always returned if another thread is importing the same transactions.

sjeohp:
Can see I've missed the point with the nonces.

PR #6215, opened 2017-08-02
Replace usages of ::std::env::temp_dir with tempdir crate in tests

### LINKED ISSUE #6147: Replace usages of ::std::env::temp_dir with tempdir crate in tests
https://github.com/paritytech/parity/pull/6140#discussion_r129265892

nicolasochem:
Fix #6147

### DIFF
--- a/ethash/src/compute.rs
+++ b/ethash/src/compute.rs
@@ -528,98 +528,112 @@ fn to_hex(bytes: &[u8]) -> String {
 	unsafe { String::from_utf8_unchecked(v) }
 }
 
-#[test]
-fn test_get_cache_size() {
-	// https://github.com/ethereum/wiki/wiki/Ethash/ef6b93f9596746a088ea95d01ca2778be43ae68f#data-sizes
-	assert_eq!(16776896usize, get_cache_size(0));
-	assert_eq!(16776896usize, get_cache_size(1));
-	assert_eq!(16776896usize, get_cache_size(ETHASH_EPOCH_LENGTH - 1));
-	assert_eq!(16907456usize, get_cache_size(ETHASH_EPOCH_LENGTH));
-	assert_eq!(16907456usize, get_cache_size(ETHASH_EPOCH_LENGTH + 1));
-	assert_eq!(284950208usize, get_cache_size(2046 * ETHASH_EPOCH_LENGTH));
-	assert_eq!(285081536usize, get_cache_size(2047 * ETHASH_EPOCH_LENGTH));
-	assert_eq!(285081536usize, get_cache_size(2048 * ETHASH_EPOCH_LENGTH - 1));
-}
+#[cfg(test)]
+mod test {
+	extern crate tempdir;
+	use super::get_cache_size;
+	use super::get_data_size;
+	use super::ETHASH_EPOCH_LENGTH;
+	use super::Light;
+	use super::light_compute;
+	use super::SeedHashCompute;
+	use super::quick_get_difficulty;
+	use std::fs;
+	use self::tempdir::TempDir;
 
-#[test]
-fn test_get_data_size() {
-	// https://github.com/ethereum/wiki/wiki/Ethash/ef6b93f9596746a088ea95d01ca2778be43ae68f#data-sizes
-	assert_eq!(1073739904usize, get_data_size(0));
-	assert_eq!(1073739904usize, get_data_size(1));
-	assert_eq!(1073739904usize, get_data_size(ETHASH_EPOCH_LENGTH - 1));
-	assert_eq!(1082130304usize, get_data_size(ETHASH_EPOCH_LENGTH));
-	assert_eq!(1082130304usize, get_data_size(ETHASH_EPOCH_LENGTH + 1));
-	assert_eq!(18236833408usize, get_data_size(2046 * ETHASH_EPOCH_LENGTH));
-	assert_eq!(18245220736usize, get_data_size(2047 * ETHASH_EPOCH_LENGTH));
-}
+	#[test]
+	fn test_get_cache_size() {
+		// https://github.com/ethereum/wiki/wiki/Ethash/ef6b93f9596746a088ea95d01ca2778be43ae68f#data-sizes
+		assert_eq!(16776896usize, get_cache_size(0));
+			assert_eq!(16776896usize, get_cache_size(1));
+		assert_eq!(16776896usize, get_cache_size(ETHASH_EPOCH_LENGTH - 1));
+		assert_eq!(16907456usize, get_cache_size(ETHASH_EPOCH_LENGTH));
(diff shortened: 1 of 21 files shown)


ethash/src/lib.rs
```diff
@@ -118,20 +118,27 @@ impl EthashManager {
...
-	assert_eq!(ethash.cache.lock().recent_epoch.unwrap(), 1);
-	assert_eq!(ethash.cache.lock().prev_epoch.unwrap(), 0);
-	ethash.compute_light(1, &hash, 1);
-	assert_eq!(ethash.cache.lock().recent_epoch.unwrap(), 0);
-	assert_eq!(ethash.cache.lock().prev_epoch.unwrap(), 1);
-	ethash.compute_light(70000, &hash, 1);
-	assert_eq!(ethash.cache.lock().recent_epoch.unwrap(), 2);
-	assert_eq!(ethash.cache.lock().prev_epoch.unwrap(), 0);
+#[cfg(test)]
+mod test {
```
rphmeier: convention would call this "mod tests", here and other places

ethash/src/compute.rs
```diff
@@ -528,98 +528,112 @@ fn to_hex(bytes: &[u8]) -> String {
...
-	assert_eq!(16776896usize, get_cache_size(ETHASH_EPOCH_LENGTH - 1));
-	assert_eq!(16907456usize, get_cache_size(ETHASH_EPOCH_LENGTH));
-	assert_eq!(16907456usize, get_cache_size(ETHASH_EPOCH_LENGTH + 1));
-	assert_eq!(284950208usize, get_cache_size(2046 * ETHASH_EPOCH_LENGTH));
-	assert_eq!(285081536usize, get_cache_size(2047 * ETHASH_EPOCH_LENGTH));
-	assert_eq!(285081536usize, get_cache_size(2048 * ETHASH_EPOCH_LENGTH - 1));
-}
+#[cfg(test)]
+mod test {
+	extern crate tempdir;
```
rphmeier: @paritytech/core-devs any opinion on including the crate within the module? I'd personally prefer it in the crate root.

ethash/src/compute.rs
```diff
@@ -528,98 +528,112 @@ fn to_hex(bytes: &[u8]) -> String {
...
+	use super::SeedHashCompute;
+	use super::quick_get_difficulty;
+	use std::fs;
+	use self::tempdir::TempDir;
+
+	#[test]
+	fn test_get_cache_size() {
+		// https://github.com/ethereum/wiki/wiki/Ethash/ef6b93f9596746a088ea95d01ca2778be43ae68f#data-sizes
+		assert_eq!(16776896usize, get_cache_size(0));
+			assert_eq!(16776896usize, get_cache_size(1));
```
tomusdrw: indentation

ethash/src/compute.rs
```diff
@@ -528,98 +528,112 @@ fn to_hex(bytes: &[u8]) -> String {
...
-	assert_eq!(result.value[..], boundary[..]);
-}
+	#[test]
+	fn test_light_compute() {
+		let hash = [0xf5, 0x7e, 0x6f, 0x3a, 0xcf, 0xc0, 0xdd, 0x4b, 0x5b, 0xf2, 0xbe, 0xe4, 0x0a, 0xb3, 0x35, 0x8a, 0xa6, 0x87, 0x73, 0xa8, 0xd0, 0x9f, 0x5e, 0x59, 0x5e, 0xab, 0x55, 0x94, 0x05, 0x52, 0x7d, 0x72];
+		let mix_hash = [0x1f, 0xff, 0x04, 0xce, 0xc9, 0x41, 0x73, 0xfd, 0x59, 0x1e, 0x3d, 0x89, 0x60, 0xce, 0x6b, 0xdf, 0x8b, 0x19, 0x71, 0x04, 0x8c, 0x71, 0xff, 0x93, 0x7b, 0xb2, 0xd3, 0x2a, 0x64, 0x31, 0xab, 0x6d];
+		let boundary = [0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x3e, 0x9b, 0x6c, 0x69, 0xbc, 0x2c, 0xe2, 0xa2, 0x4a, 0x8e, 0x95, 0x69, 0xef, 0xc7, 0xd7, 0x1b, 0x33, 0x35, 0xdf, 0x36, 0x8c, 0x9a, 0xe9, 0x7e, 0x53, 0x84];
+		let nonce = 0xd7b3ac70a301a249;
+		// difficulty = 0x085657254bd9u64;
+		let light = Light::new(TempDir::new("").unwrap(), 486382);
```
tomusdrw: I don't think it will really work, it's dropped right after start so if the test passes it means that it doesn't really use the directory anyway or it creates it and never removes in which case it doesn't make it any better than `::std::env::temp_dir()`
(here and other occurrences)

ethstore/tests/util/transient_dir.rs
```diff
@@ -14,17 +14,18 @@
...
 use ethstore::dir::{KeyDirectory, RootDiskDirectory};
 use ethstore::{Error, SafeAccount};
+use self::tempdir::TempDir;
 
 pub fn random_dir() -> PathBuf {
 	let mut rng = OsRng::new().unwrap();
-	let mut dir = env::temp_dir();
-	dir.push(format!("{:x}-{:x}", rng.next_u64(), rng.next_u64()));
-	dir
+    let temp_path = TempDir::new("").unwrap();
```
tomusdrw: 1. Identation
2. The directory is not really removed after the test so don't see any point in using the crate here (in the way it's currently used)

debris: This module can be removed entirely, cause it duplicates TempDir logic

ethash/src/compute.rs
debris: I agree, but only if it's used in more than 1 place by this crate

```rust
#[cfg(test)]
extern crate tempdir;
```

rphmeier: yep, i was leaning that direction as well. although as Rust will (in a few versions) infer `extern crate` declarations, I thought it might be best to leave them as easy to hunt down as possible

debris: really? is there an rfc for that?

Vurich: [Yes, there is](https://github.com/rust-lang/rfcs/pull/2088). I personally hope it doesn't happen, although that might just be status quo bias talking.

It's not a certainty though, the PR is yet to be merged. We have this pattern all over the codebase already:

```
➤ rg '#\[cfg' -A 2 | rg '\bmod\b' -A 1 | rg 'extern crate' | wc -l
10
```

tomusdrw:
Any progress on that one?

gavofyork:
stale. closing.

PR #6231, opened 2017-08-04
price-info does not depend on util

debris:
tiny pr, 

- first commit removes dependency on util and moves price-info.rs -> lib.rs
- second commit makes a code a bit more idiomatic


util/fetch/src/client.rs
```diff
@@ -297,7 +297,7 @@ impl Response {
 
 	/// Returns `true` if response status code is successful.
 	pub fn is_success(&self) -> bool {
-		self.status() == reqwest::StatusCode::Ok
+		self.status().is_success()
```
tomusdrw: Why was this changed? AFAIR `is_sucess()` returns true for any 2xx status code and I think we should be more strict in that case.

debris: I see, will revert it

debris:
CI fail is unrelated

PR #6233, opened 2017-08-04
native-contracts crate does not depend on util any more

debris:


debris:
js test failure is unrelated

PR #6237, opened 2017-08-05
use Math.floor instead of Math.round when displaying balance

### LINKED ISSUE #6236: Parity UI - Balances should not round (up) [suggestion]
If you want to round balances on the Accounts tab, please round down, NOT UP.

Even better would be not to round at all; you have what you have, if you only want to show 3 decimal places, fine, show them, but please don't alter them. Also it would be nice to be able to configure how many decimal places to display.

debris:
fixes #6236


tomusdrw:
The original issue is about "Accounts" tab, this fixes it only partially in TokenReg dapp.

gavofyork:
would merge, but it doesn't close the issue

PR #6247, opened 2017-08-07
Nonce future - reserve and dispatch

sjeohp:


rpc/src/v1/helpers/nonce.rs
```diff
@@ -0,0 +1,196 @@
...
+                    drop(reserved);
+                    // prospective nonce is available
+                    if let Some(mut reserved) = self.reserved.try_write() {
+                        reserved.push_back(self.value);
+                        // use this nonce for signing
+                        // then poll for dispatch readiness
+                        // dispatch may fail if another thread drops its nonce
+                        Ok(Async::Ready(self.value))
+                    } else {
+                        Ok(Async::NotReady)
```
tomusdrw: I don't think it's valid to return `Async::NotReady` in case the lock is not available. I think it might never be polled again since `RwLock` won't notify the `Task` that the state changed.

rpc/src/v1/helpers/nonce.rs
```diff
@@ -0,0 +1,196 @@
...
+    Prospective,
+    Reserved,
+    Dispatch,
+    Error(NonceError),
+}
+
+pub struct Nonce {
+    state: NonceState,
+    value: U256,
+    reserved: Arc<RwLock<VecDeque<U256>>>,
```
tomusdrw: Probably could be just `Arc<RwLock<U256>>` since we require that nonces are consecutive anyway.

sjeohp: Issue there is how to check that the nonce is the oldest (ie. front of the queue) before dispatching.  Otherwise if it is dispatched and then an older reserved nonce is dropped it would create a gap, right?

tomusdrw: Ah, right so maybe `Arc<RwLock<(U256, U256)>>` will be enough then? One for the first expected nonce and the second element for last nonce.

sjeohp: Yep that would do it.

rpc/src/v1/helpers/nonce.rs
sjeohp: Yeah I'm not entirely sure what to do about this yet but I have some ideas.

debris: Honestly, I doubt it's possible. I've been playing with futures for several months and got stuck like this a couple of times.

tomusdrw:
Any progress on this one? Would be good to enable multi-threaded RPC by default, but that one is a blocker.

PR #6271, opened 2017-08-09
Use git for the snap version

elopio:
For commits with an annotated tag, that tag will be the version. For other commits, a git identifier will be used.
This way, there is no need to manually edit the snap version.


ddorgan:
[clabot:check]

gavofyork:
[clabot:check]

PR #6274, opened 2017-08-10
Expose health status over RPC

tomusdrw:
Prevents unecessary http calls (and clears the console from those).

Brought back `/api/health` endpoint returning 412 status in case of failure (useful for automatic health monitoring).

Health status extracted to a separate crate.


dapps/node-health/src/types.rs
```diff
@@ -0,0 +1,57 @@
...
+	/// Everything's OK.
+	#[serde(rename = "ok")]
+	Ok,
+	/// Node health need attention
+	/// (the issue is not critical, but may need investigation)
+	#[serde(rename = "needsAttention")]
+	NeedsAttention,
+	/// There is something bad detected with the node.
+	#[serde(rename = "bad")]
+	Bad
```
NikVolf: trailing comma?

PR #6305, opened 2017-08-15
WASM math test

NikVolf:
+ auxiliary runtime changes

### DIFF
--- a/ethcore/wasm/src/runtime.rs
+++ b/ethcore/wasm/src/runtime.rs
@@ -72,24 +72,26 @@ impl RuntimeContext {
 }
 
 /// Runtime enviroment data for wasm contract execution
-pub struct Runtime<'a> {
+pub struct Runtime<'a, 'b> {
 	gas_counter: u64,
 	gas_limit: u64,
 	dynamic_top: u32,
 	ext: &'a mut vm::Ext,
 	memory: Arc<interpreter::MemoryInstance>,
 	context: RuntimeContext,
+	instance: &'b interpreter::ProgramInstance,
 }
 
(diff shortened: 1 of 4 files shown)


ethcore/wasm/src/runtime.rs
```diff
@@ -449,9 +452,58 @@ impl<'a> Runtime<'a> {
...
+	}
+
+	fn bitswap_i64(&mut self, context: interpreter::CallerContext)
+		-> Result<Option<interpreter::RuntimeValue>, interpreter::Error>
+	{
+		let x1 = context.value_stack.pop_as::<i32>()?;
+		let x2 = context.value_stack.pop_as::<i32>()?;
+
+		let result = ((Runtime::bswap_32(x2 as u32) as u64) << 32
+			| Runtime::bswap_32(x1 as u32) as u64) as i64;
```
rphmeier: is that actually how it's supposed to work? 
[a b c d e f g h] -> [d c b a h g f e]

instead of [h g f e d c b a]

NikVolf: Stack reverses the parameter ordering, so first parameter is popped last

ethcore/wasm/src/tests.rs
```diff
@@ -414,3 +414,37 @@ fn storage_read() {
 	assert_eq!(gas_left, U256::from(99682));
 	assert_eq!(Address::from(&result[12..32]), address);
 }
+
+
+// Tests that contract's ability to read from a storage
```
fckt: looks like there should be a different commentary..

PR #6312, opened 2017-08-16
Add pagination for trace_filter rpc method

grbIzl:
- This work is done in terms of https://github.com/paritytech/parity/issues/4858 issue.

- trace_filter method was extended with two optional parameters offset and count in order to paginate through output traces.

- For offset the new type was created TraceIndex in order to identify the trace. Inside the client it's converted into TraceId

- New tests were not created as we have only trivial implementation for these methods in test_client.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1658,7 +1658,7 @@ impl BlockChainClient for Client {
 		self.chain.read().logs(blocks, |entry| filter.matches(entry), filter.limit)
 	}
 
-	fn filter_traces(&self, filter: TraceFilter, after: Option<&TraceId>, count: Option<usize>) -> Option<Vec<LocalizedTrace>> {
+	fn filter_traces(&self, filter: TraceFilter, after: Option<usize>, count: Option<usize>) -> Option<Vec<LocalizedTrace>> {
 		let start = self.block_number(filter.range.start);
 		let end = self.block_number(filter.range.end);
 
@@ -1675,26 +1675,7 @@ impl BlockChainClient for Client {
 					return None;
 				}
 
-				let mut traces_iter = traces.into_iter();
-				if let Some(after) = after {
-					match after.transaction {
-						TransactionId::Location(block_id, transaction_number) => {
-							if let Some(block_number) = self.block_number(block_id) {
-								match traces_iter.find(|trace| trace.block_number == block_number
-									&& trace.transaction_number == transaction_number) {
-									Some(_) => {},
-									None => return None,
-								}
-							}
-						},
-						TransactionId::Hash(transaction_hash) => {
-							match traces_iter.find(|trace| trace.transaction_hash == transaction_hash) {
-								Some(_) => {},
-								None => return None,
-							}
-						},
-					}
-				}
+				let traces_iter = traces.into_iter().skip(after.unwrap_or(0));
 				Some(match count {
 					Some(count) => traces_iter.take(count).collect(),
 					None => traces_iter.collect(),
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1677,18 +1677,28 @@ impl BlockChainClient for Client {
 
 				let mut traces_iter = traces.into_iter();
 				if let Some(after) = after {
-					if let TransactionId::Location(BlockId::Hash(block_hash), transaction_number) = after.transaction {
-						match traces_iter.find(|ref trace| trace.block_hash == block_hash
-							&& trace.transaction_number == transaction_number) {
-							Some(_) => {},
-							None => return None,
-						}
+					match after.transaction {
+						TransactionId::Location(block_id, transaction_number) => {
+							if let Some(block_number) = self.block_number(block_id) {
+								match traces_iter.find(|trace| trace.block_number == block_number
+									&& trace.transaction_number == transaction_number) {
+									Some(_) => {},
+									None => return None,
+								}
+							}
+						},
+						TransactionId::Hash(transaction_hash) => {
+							match traces_iter.find(|trace| trace.transaction_hash == transaction_hash) {
+								Some(_) => {},
+								None => return None,
+							}
+						},
 					}
 				}
-				if let Some(count) = count {
-					traces_iter = traces_iter.take(count as usize).collect::<Vec<_>>().into_iter();
-				}
-				Some(traces_iter.collect())
+				Some(match count {
+					Some(count) => traces_iter.take(count).collect(),
+					None => traces_iter.collect(),
+				})
 			},
 			_ => None,
 		}
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1677,12 +1677,11 @@ impl BlockChainClient for Client {
 
 				let mut traces_iter = traces.into_iter();
 				if let Some(after) = after {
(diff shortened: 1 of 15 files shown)


ethcore/src/client/client.rs
```diff
@@ -1641,7 +1641,7 @@ impl BlockChainClient for Client {
 		self.chain.read().logs(blocks, |entry| filter.matches(entry), filter.limit)
 	}
 
-	fn filter_traces(&self, filter: TraceFilter) -> Option<Vec<LocalizedTrace>> {
+	fn filter_traces(&self, filter: TraceFilter, after: Option<&TraceId>, count: Option<u64>) -> Option<Vec<LocalizedTrace>> {
```
tomusdrw: Why not making it part of `TraceFilter`?

ethcore/src/client/traits.rs
```diff
@@ -195,7 +195,7 @@ pub trait BlockChainClient : Sync + Send {
 	fn replay(&self, t: TransactionId, analytics: CallAnalytics) -> Result<Executed, CallError>;
 
 	/// Returns traces matching given filter.
-	fn filter_traces(&self, filter: TraceFilter) -> Option<Vec<LocalizedTrace>>;
+	fn filter_traces(&self, filter: TraceFilter, after: Option<&TraceId>, count: Option<u64>) -> Option<Vec<LocalizedTrace>>;
```
tomusdrw: `usize` will be enough for `count`

ethcore/src/client/client.rs
grbIzl: Actually there were 3 options how to add this here:
- Modify TraceFIlter
- Add optional parameters to filter_traces
- Instead add new method like filter_traces_with_pagination
I've discussed it with @arkpar  and we chosen the second one. But frankly all of them pretty similar. There is no decisive advantage for any of them.

ethcore/src/client/traits.rs
grbIzl: Agree

ethcore/src/client/client.rs
```diff
@@ -1671,7 +1671,24 @@ impl BlockChainClient for Client {
...
 
 				let traces = self.tracedb.read().filter(&filter);
-				Some(traces)
+				if traces.is_empty() {
+					return None;
+				}
+
+				let mut traces_iter = traces.into_iter();
+				if let Some(after) = after {
+					if let TransactionId::Location(BlockId::Hash(block_hash), transaction_number) = after.transaction {
```
svyatonik: Why does it quietly ignores `after` argument if it is `Some(TransactionId::Hash(_))`? I can see that in RPC only `TransactionId::Location` is used, but it is not good to have public method in separate (not RPC crate) which accepts two options and quietly works for single option only.

ethcore/src/client/client.rs
```diff
@@ -1671,7 +1671,24 @@ impl BlockChainClient for Client {
...
+						match traces_iter.find(|ref trace| trace.block_hash == block_hash
```
svyatonik: ref is not needed in `ref trace` here

ethcore/src/client/client.rs
```diff
@@ -1671,7 +1671,24 @@ impl BlockChainClient for Client {
...
+							&& trace.transaction_number == transaction_number) {
+							Some(_) => {},
+							None => return None,
+						}
+					}
+				}
+				if let Some(count) = count {
+					traces_iter = traces_iter.take(count as usize).collect::<Vec<_>>().into_iter();
```
svyatonik: `count` is already `usize`, so `count as usize` is not needed here

ethcore/src/client/client.rs
```diff
@@ -1671,7 +1671,24 @@ impl BlockChainClient for Client {
...
+				if let Some(after) = after {
+					if let TransactionId::Location(BlockId::Hash(block_hash), transaction_number) = after.transaction {
+						match traces_iter.find(|ref trace| trace.block_hash == block_hash
+							&& trace.transaction_number == transaction_number) {
+							Some(_) => {},
+							None => return None,
+						}
+					}
+				}
+				if let Some(count) = count {
```
svyatonik: You could avoid extra allocation here like that:
```
				Some(match count {
					Some(count) => traces_iter.take(count as usize).collect(),
					None => traces_iter.collect(),
				})
```

ethcore/src/client/client.rs
```diff
@@ -1677,18 +1677,28 @@ impl BlockChainClient for Client {
...
+									&& trace.transaction_number == transaction_number) {
+									Some(_) => {},
+									None => return None,
+								}
+							}
+						},
+						TransactionId::Hash(transaction_hash) => {
+							match traces_iter.find(|trace| trace.transaction_hash == transaction_hash) {
+								Some(_) => {},
+								None => return None,
```
debris: it's more idiomatic to write:

```rust
if traces_iter.find(|trace| trace.transaction_hash == transaction_hash).is_none() {
  return None;
}
```

rpc/src/v1/impls/light/trace.rs
```diff
@@ -31,7 +31,7 @@ pub struct TracesClient;
 impl Traces for TracesClient {
 	type Metadata = Metadata;
 
-	fn filter(&self, _filter: TraceFilter) -> Result<Option<Vec<LocalizedTrace>>, Error> {
+	fn filter(&self, _filter: TraceFilter, _offset: Option<TraceIndex>, _count: Option<usize>) -> Result<Option<Vec<LocalizedTrace>>, Error> {
```
debris: wouldn't offset being `usize` be easier to use (and to implement)?

With offset being `TraceIndex` you can't fetch traces in parallel and/or from multiple clients. I imagine this might be a common practice for exchanges.

// thread 0, 
ethereum_client_0.filter(filter_options, 0, 1000);

// thread 1,
ethereum_client_1.filter(filter_options, 1000, 2000);

// thread 2,
ethereum_client_2.filter(filter_options, 2000, 3000);

ethcore/src/client/client.rs
tomusdrw: Using separate parameters breaks backward compatibility, we already support `limit` parameter for `eth_getLogs` filter object and that would be inline with that.

Would be good if those parameters could be omitted (using `Trailing`), but we don't suport two optional parameters.

IMHO would be better to make it part of `TraceFilter` object.

ethcore/src/client/client.rs
```diff
@@ -1664,14 +1664,22 @@ impl BlockChainClient for Client {
...
 					range: s as usize..e as usize,
 					from_address: From::from(filter.from_address),
 					to_address: From::from(filter.to_address),
 				};
 
-				let traces = self.tracedb.read().filter(&filter);
-				Some(traces)
+				let traces = self.tracedb.read().filter(&db_filter);
+				if traces.is_empty() {
+					return None;
```
tomusdrw: Why not `Some(vec![])` - I think `None` was used to indicate invalid block range.

gavofyork:
@grbIzl still waiting for this (fairly minor) change...

SinErgy84:
Please keep your wiki up2date ... 
mentioned changes are missing on https://github.com/paritytech/parity/wiki/JSONRPC-trace-module#trace_filter

5chdn:
FYI https://github.com/paritytech/wiki/blob/master/JSONRPC-trace-module.md#trace_filter

PR #6313, opened 2017-08-16
parity-evm --chain flag fix

cdetrio:
`parity-evm --chain chain_file.json` was not loading the spec from chain_file.json.


cdetrio:
replaced by https://github.com/paritytech/parity/pull/6314

PR #6318, opened 2017-08-16
Update README.md to fix SSL error

ltfschoen:
Show commands that when executed in sequence should overcome an error that is encountered when running only the simple one-liner on Mac


5chdn:
Added it to the wiki if you don't mind. :)

https://github.com/paritytech/parity/wiki/FAQ:-Building,-Installation,-and-Testing#the-one-line-installer-fails-with-ssl-peer-handshake-failed-on-macos

PR #6322, opened 2017-08-17
Itertools are no longer reexported from util, optimized triedb iter

debris:
Triedb benches

old:

```
test trie_iter                      ... bench:   6,571,603 ns/iter (+/- 732,647)
```

new:

```
test trie_iter                      ... bench:   6,181,846 ns/iter (+/- 461,696)
```


util/src/trie/triedb.rs
```diff
@@ -293,7 +292,17 @@ impl<'a> TrieDBIterator<'a> {
 	/// The present key.
 	fn key(&self) -> Bytes {
 		// collapse the key_nibbles down to bytes.
-		self.key_nibbles.iter().step(2).zip(self.key_nibbles.iter().skip(1).step(2)).map(|(h, l)| h * 16 + l).collect()
+		unsafe {
```
tomusdrw: Does it really require such big block of `unsafe` code to be fast?
what about:
```rust
let nibbles = self.key_nibbles;
let len = nibbles.len() - 1;
let mut i = 0;
while i < len {
  result.push(nibbles[i] * 16 + nibbles[i + 1]);
  i += 2;
}
```

rphmeier: maybe the checked indexing is too slow? i would guess that using unsafe only for `get_unchecked` would be good enough.

rphmeier:
LGTM other than @tomusdrw grumble

util/src/trie/triedb.rs
```diff
@@ -293,7 +292,15 @@ impl<'a> TrieDBIterator<'a> {
 	/// The present key.
 	fn key(&self) -> Bytes {
 		// collapse the key_nibbles down to bytes.
-		self.key_nibbles.iter().step(2).zip(self.key_nibbles.iter().skip(1).step(2)).map(|(h, l)| h * 16 + l).collect()
+		let nibbles = &self.key_nibbles;
+		let mut i = 0;
+		let mut result = Bytes::with_capacity(nibbles.len() / 2);
+		let len = nibbles.len() - 1;
```
arkpar: Is there a guarantee that `nibbles` is never empty and this can't underflow?

debris: no, just pushed a fix for that

PR #6329, opened 2017-08-18
Add RPC eth_chainId for querying the current blockchain chain ID

sorpaas:
Currently although we can use `net_version` RPC call to get the
current network ID, there's no RPC for querying the chain ID. This
makes it impossible to determine the current actual blockchain using
the RPC. An ETH/ETC client can accidentally connect to an ETC/ETH RPC
endpoint without knowing it unless it tries to sign a transaction or
it fetch a transaction that is known to have signed with a chain
ID. This has since caused trouble for application developers, such as
MetaMask, to add multi-chain support.

The same RPC endpoint is also about to be merged for ETC's
go-ethereum: https://github.com/ethereumproject/go-ethereum/pull/336


sorpaas:
[clabot:check]

tcz001:
+1 on this PR, I faced with this problem when using web3.js which is using the getId by default

tcz001:
Just created a EIP694(https://github.com/ethereum/EIPs/issues/694)

rpc/src/v1/impls/eth.rs
```diff
@@ -284,6 +284,11 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> Eth for EthClient<C, SN, S, M, EM> where
 		Ok(format!("{}", version))
 	}
 
+	fn chain_id(&self) -> Result<String, Error> {
```
tomusdrw: Chain id in the transactions is serialized as `Option<u64>`,  would be best to keep it constistent.
IMHO using a generic `String` response is a bit misleading. If we want to return a hex-encoded value then `Option<U64>` would be a right choice.

Note that using `Option` indicates that in case the chain id is missing `null` will be returned.

rphmeier:
I also disagree with polluting the `eth` namespace just yet, but am definitely not opposed to the idea in general.

tcz001:
Agree, an int in string type would be better for extension and compatibility here, how about changing it in EIP694 @sorpaas & @whilei, any thoughts?

whilei:
@tcz001 I've created an [EIP-rpc_chain_id as a PR](https://github.com/ethereum/EIPs/pull/695) (since it should be a PR rather than an issue) which is almost exactly like your issue [EIP694](https://github.com/ethereum/EIPs/issues/694) -- maybe we can switch to using the PR as a reference?

I used a bigint in hexadecimal because that's in line with the rest of the `eth_`-namespaced returnables, eg. https://github.com/ethereumproject/go-ethereum/wiki/JSON-RPC#eth_blocknumber.

tcz001:
Thx, will close that issue

tomusdrw:
Build failure seems unrelated. Will be happy to merge under `parity_chainId`. After the method is accepted via EIP we can rename it to `eth_chainId`.

gavofyork:
Tests fail with `RPC_API_Test...error: An unknown error occurred`.

rpc/src/v1/impls/parity.rs
```diff
@@ -200,6 +200,10 @@ impl<C, M, U> Parity for ParityClient<C, M, U> where
 		Ok(self.settings.chain.clone())
 	}
 
+	fn chain_id(&self) -> Result<Option<U256>, Error> {
```
tomusdrw: 1. minor: `U64` would be enough
2. As mentioned earlier `chain_id` is already returned as decimal (`u64`) in transactions and I'd prefer consistency (either all `U64` (hex) or all `u64` (decimal)).

sorpaas: @tomusdrw Where's the U64 struct defined?

tomusdrw: Thought we already have it, but seems we only have U128 and U256. You can create `U64` in `types/uint.rs`.

rpc/src/v1/types/uint.rs
```diff
@@ -104,7 +98,25 @@ macro_rules! impl_uint {
...
 
+impl serde::Serialize for U256 {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: serde::Serializer {
+		serializer.serialize_str(&format!("0x{}", self.0.to_hex()))
+	}
+}
+
+impl serde::Serialize for U64 {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: serde::Serializer {
+		serializer.serialize_str(&format!("0x{}", self.0))
```
rphmeier: i don't think this will actually format as hex. maybe `{:h}`? @tomusdrw should it be left-padded with zeros?

sorpaas:
`gitlab-test-rust-stable` was failed due to a connection timeout. Glad if someone can help to restart it.

gavofyork:
@tomusdrw @rphmeier please verify all good.

PR #6331, opened 2017-08-18
Better check the created accounts before showing Startup Wizard

ngotchac:
The Startup Wizard would still display even if accounts has been created in Parity. This PR fixes it.

### DIFF
--- a/js/src/views/Application/store.js
+++ b/js/src/views/Application/store.js
@@ -28,11 +28,18 @@ export default class Store {
     this._migrateStore();
 
     this._api = api;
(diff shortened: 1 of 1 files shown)


tomusdrw:
First run requires T&C to be accepted, this assumes that they were accepted if user already have some accounts created. (Just making sure that this is what we want).

ngotchac:
@tomusdrw Yes that's right. That might be an issue if an account has been created from the CLI. However, a user could only use the CLI and thus never accept the T&C anyway, right? Not sure if that's an issue though

PR #6354, opened 2017-08-22
Wasm math tests extended

NikVolf:


ethcore/wasm/src/tests.rs
```diff
@@ -415,36 +415,121 @@ fn storage_read() {
 	assert_eq!(Address::from(&result[12..32]), address);
 }
 
+macro_rules! reqrep_test {
```
arkpar: This could be just a function that accepts `code` and `input`

NikVolf: but it uses macro, the testfile content should be known at compile time, no runtime loading there

arkpar: Right, that's exactly why I was suggesting making testfile content (`code`) an argument to the function.

PR #6355, opened 2017-08-22
Running state test using parity-evm

### LINKED ISSUE #6286: Add a comandline option to execute a custom stateTest
With random state test generation we need a support option from the clients to run a custom .json file that is a StateTest.

The file format is here:
https://github.com/ethereum/tests/blob/develop/GeneralStateTests/stExample/add11.json

The usage should be like following:
./parity --runfile "/GeneralStateTests/stExample/add11.json" --format "StateTest"

the client should return whether the test execution was success or not.
if it is a failure it should provide information on which fork rule there was an error.
so then we could debug it more carefully.

tomusdrw:
Closes #6286 

Should be carefuly reviewed cause it changes some stuff in `State` and `Executive`.


gavofyork:
@rphmeier and @arkpar please give it a look

arkpar:
Merged this before seeing @gavofyork's comment. LGTM. Changes in `State` and `Executive` are not consensus-related. PR does not touches transactions, only calls.

PR #6370, opened 2017-08-24
Light-client sync for contract-based PoA

rphmeier:
Closes https://github.com/paritytech/parity/issues/5427

Summary of changes:
  - Generalize state-dependent proofs
  - Check for signals.
  - Store pending and finalized epoch changes in the light client's header chain.
  - Separate `EngineClient` from `BlockChainClient` and implement `EngineClient` for light client.
  - Fetch and check epoch change proofs from the network on demand during light sync.


rphmeier:
just a note: the only proof that the light client actually needs to go to the network for is when the transition to the contract first begins. Transitions within the contract can have their proofs computed fully on the client-side.

ethcore/light/src/client/header_chain.rs
```diff
@@ -53,6 +59,9 @@ const HISTORY: u64 = 2048;
 /// The best block key. Maps to an RLP list: [best_era, last_era]
 const CURRENT_KEY: &'static [u8] = &*b"best_and_latest";
 
+/// Key storing the last canonical epoch transition.
+const LAST_CANONICAL_TRANSITION: &'static [u8] = &*b"canonical_transition";
```
arkpar: Does this require a database upgrade?

rphmeier: No, the most recent version of the code instantiates this key if it doesn't exist already. Only chains which don't use epochs can be synced by the light client right now anyway. When running the new version, their genesis epoch data will be populated correctly.

PR #6389, opened 2017-08-26
Trie optimizations

debris:
changes:
- triedb seek and descend is not recursive
- optimized memorydb `insert`, `remove` and `emplace`
- optimized hashdb `keys()`

all trie insert benchmarks are about 10% faster


util/src/memorydb.rs
```diff
@@ -200,16 +203,17 @@ impl HashDB for MemoryDB {
 			return SHA3_NULL_RLP.clone();
 		}
 		let key = value.sha3();
-		if match self.data.get_mut(&key) {
-			Some(&mut (ref mut old_value, ref mut rc @ -0x80000000i32 ... 0)) => {
-				*old_value = DBValue::from_slice(value);
+		match self.data.entry(key) {
+			Entry::Occupied(mut entry) => {
+				let &mut (ref mut old_value, ref mut rc) = entry.get_mut();
+				if *rc >= -0x80000000i32 && *rc <= 0 {
```
debris: `rc >= -0x80000000i32` is probably not needed, right @rphmeier ?

util/src/memorydb.rs
```diff
@@ -181,7 +181,10 @@ impl HashDB for MemoryDB {
 	}
 
 	fn keys(&self) -> HashMap<H256, i32> {
-		self.data.iter().filter_map(|(k, v)| if v.1 != 0 {Some((k.clone(), v.1))} else {None}).collect()
+		self.data.iter()
```
arkpar: Why remove a `filter_map`? Clippy produces a warning on `.filter().map()` sequence

util/src/trie/triedb.rs
```diff
@@ -217,59 +217,76 @@ impl<'a> TrieDBIterator<'a> {
...
-						status: Status::At,
-						node: node.clone().into(),
-					});
-				} else {
-					self.trail.push(Crumb {
-						status: Status::Exiting,
-						node: node.clone().into(),
-					});
-				}
+	fn seek_descend<'key>(&mut self, mut node_data: DBValue, mut key: NibbleSlice<'key>) -> super::Result<()> {
```
arkpar: Should be renamed to just `seek` since it is not recursive anymore.

util/src/trie/triedb.rs
```diff
@@ -217,59 +217,76 @@ impl<'a> TrieDBIterator<'a> {
...
+		enum Step {
+			Descend(DBValue, usize),
+			Return,
```
arkpar: This could be just a `return` statement

util/src/memorydb.rs
debris: ah, ok

I just found it more readable than this long one-liner

debris:
@arkpar fixed

util/src/memorydb.rs
debris: fixed

debris:
js failure is unrelated

PR #6391, opened 2017-08-27
fixed pending block panic

### LINKED ISSUE #6239: Thread '<unnamed>' panicked at 'index out of bounds: the len is 0 but the index is 1', ethcore/src/ethereum/ethash.rs:537
Hey, all is in the title

I got this error when I launch Parity and a pool software at the same time : 

`stack backtrace:
   0:     0x55f790a2939c - backtrace::backtrace::trace::h18b17b72e62d3e01

Thread '<unnamed>' panicked at 'index out of bounds: the len is 0 but the index is 1', ethcore/src/ethereum/ethash.rs:537

Aborted`

Where can it come from ?

Thanks

debris:
closes #6239


tomusdrw:
That's a quick for current issue, long term solution should either distinguish the type or return `Result/Option` from nonce to indicate that it's not always present.

Issue #4500, opened 2017-02-09
Trezor hardware wallet support


mcgravier:
While quite popular in cryptocurrency community, the only service that supports Trezor for Ethereum at the  time is Myetherwallet - it would be far better to have direct implementation in Parity

rphmeier:
#4486 lays the foundation for other HW wallets, probably will come decently soon afterwards. The main blocker is just that we haven't got a Trezor wallet to test with.

3esmit:
@prusnak @slush0 could donate a Trezor wallet to Parity team, and even collaborate in the development.

prusnak:
Has Ethereum foundation run out of all its money?

rphmeier:
@prusnak Not too familiar with the foundation's financial situation; they might be able to answer that better. The donation bit wasn't meant to be interpreted literally, just a little tongue-in-cheek comment.

prusnak:
I have no problem with donating a dev unit, hit me with an email to stick@satoshilabs.com with the address of a core dev.

I just hate that we live in a world where every cryptocoin raises million dollars in crowdfunding that are burnt on useless stuff, yet everyone thinks we hardware wallet producers are obliged to give away free devices to everyone developer out there.

rphmeier:
Thanks @prusnak, AFAIK we have already made plans to order a Trezor, it just hasn't arrived yet.

mcgravier:
@prusnak I understand the frustration - There was commitment from multiple parties to integrate it, yet noone did since firmware support has been released (6 months ago). AFAIK you had to do integration for MyEtherwallet by yourself which is just sad. 

Since I'm not software developer,  all I can do is submitting issues at relevant github repos :(

gavofyork:
> Has Ethereum foundation run out of all its money

fyi Parity Technologies (fka Ethcore) has no connection whatsoever with Ethereum Foundation. We have never received any aid of any kind from them. We have no relationship with them, professional or otherwise.

5chdn:
This issue is labelled with _sometimesoon_: Issue is worth doing soon. But it's inactive for several months and neither assigned nor linked to a milestone.

Do we have a Trezor yet? It probably needs a re-prioritization, or a proper deadline and an assignee.

rphmeier:
We have a Trezor, but this will have to be done UI-side as opposed to node-side since Trezor requires showing a visual pattern to the user.

klungler:
[...]

PR #6403, opened 2017-08-28
Trezor Support

folsen:
This PR addresses #4500 and extends what already exists for Ledgers to work for Trezor devices as well. Among other things it adds

* Pin Matrix modal to the UI that pops up when a device is plugged in
* RPC method to send requests to the Trezor device (such as pin)
* Library code to talk to the Trezor over USB HID


js/src/jsonrpc/interfaces/parity.js
```diff
@@ -360,6 +360,32 @@ export default {
     }
   },
 
+  keepkey: {
```
jacogr: Does not match up with the method exposed (e.g. `parity_trezor`)

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,149 @@
...
+        <div className={ styles.modal }>
+          <div className={ styles.body }>
+            <div id={ styles.title }>
+              <FormattedMessage
+                id='pinMatrix.enterPin'
+                defaultMessage='{enterPin}'
+                values={ {
+                  enterPin: (
+                    <span>
+                      Please enter the <i>pin</i> for your { device.manufacturer } hardware wallet
```
jacogr: Needs to be `FormattedMessage` for translation as well (as used lower down)

folsen:
[clabot:check]

hw/src/trezor/gen/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod types;
+pub mod messages;
```
debris: is it possible to generate those files at compile time rather than committing 25k loc? :)

hw/src/lib.rs
```diff
@@ -16,39 +16,58 @@
...
 use std::sync::{Arc, Weak};
+use std::sync::atomic;
 use std::sync::atomic::AtomicBool;
+use std::thread;
 use std::time::Duration;
-use parking_lot::Mutex;
-use ethkey::{Address, Signature};
-
-pub use ledger::KeyPath;
+use util::{Bytes, U256};
```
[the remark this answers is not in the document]
folsen: Ah, I didn't know it was there too, will change!

hw/src/trezor/gen/mod.rs
NikVolf: doubling request, this code should be committed only if compile-time generation is absolutely impossible

folsen: While not impossible, there is no existing way to do it. Down the line we could write a macro to do some compile-time generation, but it's pretty hard to do without requiring the protobuf C-library from google. An acceptable intermediate solution that I've now pushed is to put all the generated code in the `trezor-sys` crate and import what we need from there.

hw/Cargo.toml
```diff
@@ -9,9 +9,13 @@ authors = ["Parity Technologies <admin@parity.io>"]
 [dependencies]
 log = "0.3"
 parking_lot = "0.4"
+protobuf = "1.4"
+serde = "1.0"
+serde_json = "1.0"
 hidapi = { git = "https://github.com/paritytech/hidapi-rs" }
 libusb = { git = "https://github.com/paritytech/libusb-rs" }
 ethkey = { path = "../ethkey" }
+ethcore-util = { path = "../util" }
```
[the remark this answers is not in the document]
folsen: Currently `ethcore/src/account_provider/mod.rs` imports the following from the hardware wallet crate:
```
use hardware_wallet::{Error as HardwareError, HardwareWalletManager, KeyPath, TransactionInfo};
```
and it's using it for providing accounts given by hardware wallets as well as signing transactions with hardware wallets (the `sign_with_hardware` function). I don't think you could remove the dependency without abstracting out both account-fetching and tx-signing to their own crates.

hw/src/lib.rs
```diff
@@ -16,39 +16,58 @@
...
+	TrezorDevice(trezor::Error),
 	/// USB error.
 	Usb(libusb::Error),
 	/// Hardware wallet not found for specified key.
 	KeyNotFound,
 }
 
-/// Hardware waller information.
+/// This is the transaction info we need to supply to Trezor message. It's more
+/// or less a duplicate of ethcore::transaction::Transaction, but we can't
```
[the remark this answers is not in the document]
folsen: Currently `ethcore/src/account_provider/mod.rs` imports the following from the hardware wallet crate:
```
use hardware_wallet::{Error as HardwareError, HardwareWalletManager, KeyPath, TransactionInfo};
```
and it's using it for providing accounts given by hardware wallets as well as signing transactions with hardware wallets (the `sign_with_hardware` function). I don't think you could remove the dependency without abstracting out both account-fetching and tx-signing to their own crates.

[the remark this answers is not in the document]
rphmeier: those things probably can be taken out into their own crates anyway. I would suggest making a single crate for account management, which encompasses both software and hardware wallets.

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,147 @@
...
+                <FormattedMessage
+                  id='pinMatrix.enterPin'
+                  defaultMessage='Please enter the pin for your {manufacturer} hardware wallet'
+                  values={ {
+                    manufacturer: device.manufacturer
+                  } }
+                />
+              </span>
+            </div>
+            <div id={ styles.passcodeBoxes }>
```
tomusdrw: should be `className` not `id`

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+/* but WITHOUT ANY WARRANTY; without even the implied warranty of
+/* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+/* GNU General Public License for more details.
+/*
+/* You should have received a copy of the GNU General Public License
+/* along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+.bold {
+  font-weight: bold !important;
```
tomusdrw: The rule doesn't seem to be used.

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+.pinMatrix {
```
tomusdrw: Empty rule

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background: rgba(255, 255, 255, 0.75);
+  z-index: 20000
+}
+
+.modal {
+  position: fixed;
```
tomusdrw: Can share some stuff with `.overlay`

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+  padding: 2em 4em;
+  text-align: center;
+  max-width: 30em;
+  min-height: 200px;
+  background: rgba(25, 25, 25, 0.75);
+  color: rgb(208, 208, 208);
+  box-shadow: rgba(0, 0, 0, 0.25) 0px 14px 45px, rgba(0, 0, 0, 0.22) 0px 10px 18px
+}
+
+#title {
```
tomusdrw: Empty rule

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+
+}
+
+#passcodeBoxes {
```
tomusdrw: Please avoid using id-based rules.

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+  border-bottom-right-radius: 3px;
+  border-top-right-radius: 3px;
+  position: absolute;
+  line-height: 20px;
+}
+#clearThik:hover {
+  cursor: pointer;
+}
+#clearThik:after {
+	right: 100%;
```
tomusdrw: Indentation is broken here.

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#pin {
+  margin-left: 2px;
+  min-height: 20px;
+  position: relative;
+  letter-spacing: 2px;
+}
+
+#clearThik {
```
tomusdrw: id rule -> class rule

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+.passcodeBox {
+  position: relative;
+  width: 25px;
+  height: 25px;
+  line-height: 25px;
+  margin: 2px;
+  background-color: rgba(0, 0, 0, 0.5);
+  transition: background-color 500ms;
+}
+.passcodeBox:hover {
```
tomusdrw: please merge the rules to avoid duplication
```css
.passcodeBox {
   ...

   &:hover {
      ...
   }
}

File: js/src/modals/PinMatrix/pinMatrix.js [hunk 16]
```diff
@@ -0,0 +1,147 @@
...
+          </div>
+        </div>
+
+      </div>
+    );
+  }
+
+  renderPasscodeBox () {
+    return Array.apply(null, Array(9)).map((box, index) => {
+      let addDigit = () => this.addDigit(this.pinMatrix[index]);
```
tomusdrw: IMHO you can do with a single handler here and custom attribute:
```
handleAddDigit = (ev) => {
  const index = ev.target.getAttribute('key');
}
```

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,147 @@
...
+
+      return (
+        <div
```
tomusdrw: Clickable elements should be `buttons` to handle focus correctly.

js/src/modals/PinMatrix/pinMatrix.css
```diff
@@ -0,0 +1,159 @@
...
+#passcodeBoxes {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: center;
+  width: 100px;
+  padding: 20px 0;
+  margin: auto;
+}
+
+.passcodeBox {
```
tomusdrw: please add `cursor: pointer`

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,147 @@
...
+              </span>
+            </div>
+            <div id={ styles.passcodeBoxes }>
+              {this.renderPasscodeBox()}
+            </div>
+
+            <div id={ styles.pin }>
+              {passcode.replace(/./g, '*')}
+              {
+                (passcode.length)
```
tomusdrw: braces not needed

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,147 @@
...
+                  ? <div id={ styles.clearThik } onClick={ this.removeDigit } />
+                  : null
+              }
+            </div>
+            <div>
+              <span className={ styles.button } id={ styles.submit } onClick={ this.submit }>Submit</span>
```
tomusdrw: Multi-attribute elements should have each attribute in a separate line.

tomusdrw: event handlers should start with `handle*`

js/src/modals/PinMatrix/pinMatrix.js
```diff
@@ -0,0 +1,147 @@
...
+              />
+            )
+          });
+        } else {
+          this.setState({
+            passcode: '',
+            failureMessage: ''
+          });
+        }
+      });
```
tomusdrw: What if it fails?

rpc/src/v1/impls/light/parity.rs
```diff
@@ -132,6 +132,11 @@ impl Parity for ParityClient {
 		)
 	}
 
+	fn trezor(&self, message_type: String, device_path: Option<String>, message: Option<String>) -> Result<String, Error> {
```
tomusdrw: Imho the API is quite unsafe - it allows you to send arbitrary messages to a connected trezor device by any dapp.

We should:
1. Move that API to `parity_set` (protected set of methods) or even `parity_accounts`
2. Restrict the RPC and introduce enum for allowed messages + the messages should be constructed on parity side.

js/src/mobx/hardwareStore.js
```diff
@@ -51,6 +52,21 @@ export default class HardwareStore {
     }, HW_SCAN_INTERVAL);
   }
 
+  scanTrezor () {
+    return this._api.parity
+      .trezor('get_devices')
+      .then((message) => {
+        this.pinMatrixRequest = JSON.parse(message).map((path) => {
```
tomusdrw: I'm not really a mobx expert, but shouldn't modification of this be done in some kind of `@action`? @ngotchac ?

js/src/mobx/hardwareStore.js
```diff
@@ -51,6 +52,21 @@ export default class HardwareStore {
...
+          return { path: path, manufacturer: 'Trezor' };
```
tomusdrw: Don't we need a name here?

js/src/mobx/hardwareStore.js
```diff
@@ -147,6 +164,19 @@ export default class HardwareStore {
...
+  pinMatrixAck (device, passcode) {
+    return this._api.parity
+      .trezor('pin_matrix_ack', device.path, passcode)
+      .then((message) => {
+        this.scan();
+        // true/false is returned
+        return JSON.parse(message);
+      })
+      .catch((err) => {
+        return err;
```
tomusdrw: This converts an error into a positive response, was that intended? Is the calling code aware of that?

hw/src/trezor.rs
```diff
@@ -0,0 +1,424 @@
...
+				})
+			}
+			Ok(None) => Err(Error::ClosedDevice(dev_info.path.clone())),
+			Err(e) => Err(e),
+		}
+	}
+
+	pub fn message(&self, message_type: String, device_path: Option<String>, message: Option<String>) -> Result<String, Error> {
+		match message_type.as_ref() {
+			"get_devices" => {
```
tomusdrw: As mentioned on the RPC layer, we should make it an enum.

hw/src/trezor.rs
```diff
@@ -0,0 +1,424 @@
...
+				Ok(Some(From::from(response.get_address())))
+			}
+			_ => Ok(None),
+		}
+	}
+
+	/// Sign transaction data with wallet managing `address`.
+	pub fn sign_transaction(&self, address: &Address, t_info: &TransactionInfo) -> Result<Signature, Error> {
+		let device = self.devices.iter().find(|d| &d.info.address == address).ok_or(Error::KeyNotFound)?;
+		let usb = self.usb.lock();
```
tomusdrw: Is the lock required for the entire function? Seems that `HidDevice` does not require the lock, does it?
(same comment applies everywhere where `self.usb.lock()` is done).

hw/src/trezor.rs
```diff
@@ -0,0 +1,424 @@
...
+				} else {
+					let v = resp.get_signature_v();
+					let r = H256::from_slice(resp.get_signature_r());
+					let s = H256::from_slice(resp.get_signature_s());
+					if let Some(c_id) = *chain_id {
+						// If there is a chain_id supplied, Trezor will return a v
+						// part of the signature that is already adjusted for EIP-155,
+						// so v' = v + 2 * chain_id + 35, but code further down the
+						// pipeline will already do this transformation, so remove it here
+						Ok(Signature::from_rsv(&r, &s, (v - (35 + 2 * c_id as u32)) as u8))
```
tomusdrw: can overflow in case of invalid `v`, might be worth checking that.

hw/src/trezor.rs
```diff
@@ -0,0 +1,424 @@
...
+				let resp: EthereumTxRequest = protobuf::core::parse_from_bytes(&bytes)?;
+				if resp.has_data_length() {
+					let mut msg = EthereumTxAck::new();
+					let len = resp.get_data_length() as usize;
+					msg.set_data_chunk(data[..len].to_vec());
+					self.send_device_message(handle, &MessageType::MessageType_EthereumTxAck, &msg)?;
+					self.signing_loop(handle, chain_id, &data[len..])
+				} else {
+					let v = resp.get_signature_v();
+					let r = H256::from_slice(resp.get_signature_r());
```
tomusdrw: iirc `from_slice` panics in case the length is invalid - might be worth checking the length here.

hw/src/trezor.rs
```diff
@@ -0,0 +1,431 @@
...
+		if first_chunk < 9 || buf[0] != '?' as u8 || buf[1] != '#' as u8 || buf[2] != '#' as u8 {
+			return Err(protocol_err);
+		}
+		let msg_type = MessageType::from_i32(((buf[3] as i32 & 0xFF) << 8) + (buf[4] as i32 & 0xFF)).ok_or(protocol_err)?;
+		let msg_size = ((buf[5] as u32 & 0xFF) << 24) + ((buf[6] as u32 & 0xFF) << 16) + ((buf[7] as u32 & 0xFF) << 8) + (buf[8] as u32 & 0xFF);
+		let mut data = Vec::new();
+		data.extend_from_slice(&buf[9..]);
+		while data.len() < (msg_size as usize) {
+			device.read_timeout(&mut buf, 10_000)?;
+			data.extend_from_slice(&buf[1..]);
```
tomusdrw: How do we know that the whole buffer is filled?

hw/src/lib.rs
```diff
@@ -16,39 +16,58 @@
...
+/// import ethcore here as that would be a circular dependency.
+#[derive(Debug)]
+pub struct TransactionInfo {
+	pub nonce: U256,
+	pub gas_price: U256,
+	pub gas_limit: U256,
+	pub to: Option<Address>,
+	pub value: U256,
+	pub data: Vec<u8>,
+	pub network_id: Option<u64>,
```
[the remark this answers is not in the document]
folsen: Ah good point, this was written before the terminology was changed in the account provider, I will update!

hw/src/trezor.rs
```diff
@@ -0,0 +1,431 @@
...
+						address: addr,
+					},
+				})
+			}
+			Ok(None) => Err(Error::ClosedDevice(dev_info.path.clone())),
+			Err(e) => Err(e),
+		}
+	}
+
+	pub fn message(&self, message_type: String, device_path: Option<String>, message: Option<String>) -> Result<String, Error> {
```
rphmeier: do any of these passed parameters actually have to be owned data?

js/src/mobx/hardwareStore.js
folsen: The manufacturer is used to say "Please enter the pin for your {manufacturer} device", which I thought was good enough. If you have multiple trezors plugged in you might need a name to distinguish though, but I didn't think it would be all that important. You could plug them in and unlock one at a time to distinguish. But open to other suggestions, we'd need to propagate the name from the backend though, right now only USB path is pushed out.

hw/src/trezor.rs
folsen: The HidDevice does require a lock because multiple threads trying to read from the HID interface results in an error (something like "HID Interface is already in use").

hw/src/trezor.rs
folsen: Such is the protocol of the device, if it were to return fewer bytes, that would be a protocol error. The bytes returned from this function would be somewhat "mangled" and thus unable to parse into a protobuf message later, resulting in a protocol error being thrown.

hw/src/lib.rs
```diff
@@ -121,54 +173,97 @@ impl libusb::Hotplug for EventHandler {
...
+		}
 	}
 }
 
 impl HardwareWalletManager {
 	pub fn new() -> Result<HardwareWalletManager, Error> {
 		let usb_context = Arc::new(libusb::Context::new()?);
-		let ledger = Arc::new(Mutex::new(ledger::Manager::new()?));
-		usb_context.register_callback(None, None, None, Box::new(EventHandler { ledger: Arc::downgrade(&ledger) }))?;
+		let hidapi = Arc::new(Mutex::new(hidapi::HidApi::new().map_err(|_| libusb::Error::Io)?));
```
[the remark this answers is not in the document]
folsen: The error returned is just a plain string, so there wasn't a good place to add it, but I've added a new error type now that just wraps the string.

rpc/src/v1/impls/light/parity.rs
folsen: It doesn't allow sending arbitrary messages _to the device_. There's plenty of steps to validate the data before it hits the device, ultimately the only thing that is sent to the device is a "pin" which is sent through protobuf and is "supposed" to be safe, however as an extra safety precaution we could make sure that the input string isn't more than 9 characters, this is however already done by the trezor firmware and IMO should be their responsibility to make sure that the input pin is safety-checked.

hw/src/trezor.rs
folsen: No, changing it!

js/src/modals/PinMatrix/pinMatrix.js
folsen: Yeah, this should really be a message like "Failed to communicate with device", a failure here means it wasn't able to even send the pin to the device.

js/src/mobx/hardwareStore.js
CraigglesO: by adding an `@observable` and then in the module using it a `@observer`, as soon as there is a value change, the value updates. An action is not necessary.

hw/src/trezor.rs
rphmeier: why does it always discard the first read byte on each iteration of the loop?

CraigglesO: The first byte of each 64 byte increment from the device is always `32` as an error check to make sure each send isn't messed up. Ideally, each time new data comes it, it's worthwhile to authenticate.

folsen: Yeah, part of the protocol that the first byte is '?', maybe I should throw in a quick check that it is that? If the data is wrong and it can't be parsed into whatever protobuf it's supposed to be, it will throw an error anyway though? Not exactly sure what adding a check gives us.

hw/src/trezor.rs
folsen: I've checked and don't believe `from_slice` panics.

folsen:
So in my quest to debug this on Linux I discovered that it wasn't a bug in our software. Apparently on linux you need to modify a udev file according to these instructions: https://doc.satoshilabs.com/trezor-user/settingupchromeonlinux.html

With this setting changes, everything works as expected on Linux.

I will make `open_path` return a guard as indicated above and change the input to the RPC method to an Enum to make it less stringly typed, then I'll change the status to pleasereview again.

hw/src/ledger.rs
```diff
@@ -139,7 +134,8 @@ impl Manager {
 	}
 
 	fn read_device_info(&self, dev_info: &hidapi::HidDeviceInfo) -> Result<Device, Error> {
-		let mut handle = self.open_path(&dev_info.path)?;
+		let usb = self.usb.lock();
+		let mut handle = self.open_path(|| usb.open_path(&dev_info.path))?;
```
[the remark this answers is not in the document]
folsen: Actually upon further inspection I'm not entirely clear on what the goal of returning the guard from `open_path` would be, or indeed what that would even mean, `open_path` isn't really aware that it's handling a locked item?

To clear up the lock "issues" I propose (in a separate PR) that we refactor the whole HW wallet interface, and implement a device-dispatcher and unified trait for all kinds of devices, so adding new devices in the future just means implementing the wire protocol.

folsen:
@tomusdrw In the latest commit I implemented the change we talked about, moving from stringly typed message type to an actual Enum. Although with all of the extra import and cross-crate dependencies now necessary I'm not 100% sure it's worth it, please let me know what you think.

Overall I think the hardware wallet structure and interface could use some love. Like mentioned in a comment above, there's a way to unify all of this and make it generic over any device type. The device-specific code should really just be the wire-protocol. Even though there are many imaginable message types that we would want to send to a device, currently there is only a way to get locked devices and a way to send a pin, perhaps they should just be two separate RPC methods and be generic over device type (so no `trezor_pin_ack` and `keepkey_pin_ack` etc, just `hw_wallet_pin_ack` or whatever).

I'm not sure how much of a priority it would be to clean this API up though, and if it should be done in this PR or another. Feedback welcome!

CraigglesO:
hw_wallet_pin_ack is confusing because IMO because more hardware wallets will come with more unique communication protocols.
Trezor is the 1st generation from the author. There is a Trezor 2.0 on its way, so when decisions are made on your end about the structure of the project, I'll make the minor modifications to make it work for the keepkey without (ideally) changing really anything (will just use the trezor functions)

folsen:
I'm changing this back to pleasereview, @tomusdrw if you want to keep working on the design please change it back to inprogress again.

gavofyork:
needs resolving. and another review eventually...

tomusdrw:
Sorry, forgot about this. I will submit the fixes in a separate PR.

hw/src/trezor.rs
```diff
@@ -0,0 +1,462 @@
...
+			}
+			Ok(None) => Err(Error::ClosedDevice(dev_info.path.clone())),
+			Err(e) => Err(e),
+		}
+	}
+
+	pub fn message(&self, message_type: &TrezorMessageType, device_path: &Option<String>, message: &Option<String>) -> Result<String, Error> {
+		match *message_type {
+			TrezorMessageType::GetDevices => {
+				serde_json::to_string(&self.closed_devices).map_err(Error::SerdeError)
```
tomusdrw: Would really prefer to have a serializable enum return type as well, passing the strings around feels a bit hacky.
TBH since we only have two possible messages why not split it into separate methods?

hw/src/trezor.rs
```diff
@@ -0,0 +1,462 @@
...
+const ETH_DERIVATION_PATH: [u32; 4] = [0x8000002C, 0x8000003C, 0x80000000, 0]; // m/44'/60'/0'/0
+const ETC_DERIVATION_PATH: [u32; 4] = [0x8000002C, 0x8000003D, 0x80000000, 0]; // m/44'/61'/0'/0
+
+
+/// Type of message that can be sent to a Trezor device
+#[derive(Debug, Serialize, Deserialize)]
+pub enum TrezorMessageType {
+	/// Get a list of locked devices.
+	#[serde(rename="getDevices")]
+	GetDevices,
```
tomusdrw: `GetLockedDevices` might be a better name to distinguish it from `hardwareAccounts` that we already have (and where trezor shows up as well after it's unlocked).

hw/src/trezor.rs
```diff
@@ -0,0 +1,462 @@
...
+	/// Supply a pin for the Pin Matrix Ack
+	#[serde(rename="pinMatrixAck")]
+	PinMatrixAck,
```
tomusdrw: Since the parameters (device and pin) are required only for this variant why not encode them here in the enum?

rpc/src/v1/impls/parity.rs
```diff
@@ -151,6 +152,11 @@ impl<C, M, U> Parity for ParityClient<C, M, U> where
 		)
 	}
 
+	fn trezor(&self, message_type: TrezorMessageType, device_path: Option<String>, message: Option<String>) -> Result<String, Error> {
```
tomusdrw: Should be moved to `parity_accounts` to prevent unauthorized access.

hw/src/trezor.rs
folsen: Yeah that's what I was asking above as well. If we change it to two methods, then `getLockedDevices` could be `lockedHardwareAccountsInfo` to match with the `hardwareAccountsInfo` method and a `hardwarePinMatrixAck` method (that could live under `parity_accounts` to be more protected than just getting a list of locked devices).

That would also solve the "enum problem" by removing it altogether. I ended up not doing an enum return type because the workaround for lack of `impl Trait` felt hackier than returning a string.

If you agree then I think I'll split it into two methods and remove this whole message-passing thing.

tomusdrw: Awesome, makes sense!

hw/src/ledger.rs
```diff
@@ -187,24 +182,25 @@ impl Manager {
...
 	pub fn device_info(&self, address: &Address) -> Option<WalletInfo> {
-		self.devices.iter().find(|d| &d.info.address == address).map(|d| d.info.clone())
+		self.devices.read().iter().find(|d| &d.info.address == address).map(|d| d.info.clone())
 	}
 
 	/// Sign transaction data with wallet managing `address`.
 	pub fn sign_transaction(&self, address: &Address, data: &[u8]) -> Result<Signature, Error> {
-		let device = self.devices.iter().find(|d| &d.info.address == address)
-			.ok_or(Error::KeyNotFound)?;
+		let devices = self.devices.read();
```
tomusdrw: Swap the lock order to prevent deadlock.

hw/src/trezor.rs
```diff
@@ -0,0 +1,440 @@
...
+			}
+			_ => Ok(None),
+		}
+	}
+
+	/// Sign transaction data with wallet managing `address`.
+	pub fn sign_transaction(&self, address: &Address, t_info: &TransactionInfo) -> Result<Signature, Error> {
+		let devices = self.devices.read();
+		let device = devices.iter().find(|d| &d.info.address == address).ok_or(Error::KeyNotFound)?;
+		let usb = self.usb.lock();
```
tomusdrw: Swap the lock order to prevent deadlock.

js/src/mobx/hardwareStore.js
```diff
@@ -55,6 +58,21 @@ export default class HardwareStore {
     }, HW_SCAN_INTERVAL);
   }
 
+  scanTrezor () {
+    return this._api.parity
+      .lockedHardwareAccountsInfo()
+      .then((paths) => {
+        this.pinMatrixRequest = paths.map((path) => {
```
tomusdrw: As @jacogr pointed out in one of my requests Mobx may not update the listeners in case the assignment doesn't happen in an action:

```js
@action
setPinMatrixRequest= (req) => {
  this.pinMatrixRequest = req
};

[...]
.then((paths) => {
  this.setPinMatrixRequest(paths.map(...));
});
```

js/src/mobx/hardwareStore.js
```diff
@@ -55,6 +58,21 @@ export default class HardwareStore {
     }, HW_SCAN_INTERVAL);
   }
 
+  scanTrezor () {
+    return this._api.parity
+      .lockedHardwareAccountsInfo()
```
tomusdrw: This could be moved to pubsub API, but we can keep it for a separate PR.

PR #6404, opened 2017-08-28
remove the redundant hasher in Bloom

Hawstein:
Fix the TODO stuff in Bloom.

### DIFF
--- a/util/bloom/src/lib.rs
+++ b/util/bloom/src/lib.rs
@@ -24,9 +24,6 @@ use std::hash::{Hash, Hasher};
 use std::collections::HashSet;
 use siphasher::sip::SipHasher;
 
-// TODO [ToDr] Both hashers are exactly the same - no point to keep two.
-const NUMBER_OF_HASHERS: usize = 2;
-
 /// BitVec structure with journalling
 /// Every time any of the blocks is getting set it's index is tracked
 /// and can be then drained by `drain` method
@@ -80,8 +77,6 @@ pub struct Bloom {
 	bitmap: BitVecJournal,
 	bitmap_bits: u64,
 	k_num: u32,
-	// TODO [ToDr] Both hashers are exactly the same - no point to keep two.
-	sips: [SipHasher; NUMBER_OF_HASHERS],
 }
 
 impl Bloom {
@@ -93,12 +88,10 @@ impl Bloom {
 		let bitmap_bits = (bitmap_size as u64) * 8u64;
 		let k_num = Bloom::optimal_k_num(bitmap_bits, items_count);
 		let bitmap = BitVecJournal::new(bitmap_bits as usize);
-		let sips = [SipHasher::new(), SipHasher::new()];
 		Bloom {
 			bitmap: bitmap,
 			bitmap_bits: bitmap_bits,
 			k_num: k_num,
-			sips: sips,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


util/bloom/src/lib.rs
```diff
@@ -178,11 +174,11 @@ impl Bloom {
...
+	fn bloom_hash<T>(&self, hashes: &mut [u64; 2], item: &T, k_i: u32) -> u64
 		where T: Hash
 	{
-		if k_i < NUMBER_OF_HASHERS as u32 {
-			let mut sip = self.sips[k_i as usize].clone();
+		if k_i < 2 {
+			let mut sip = self.sip.clone();
 			item.hash(&mut sip);
 			let hash = sip.finish();
 			hashes[k_i as usize] = hash;
```
tomusdrw: The thing is that the `hashes` array will contain exactly the same number twice, so the whole thing is redundant as well.

I think we should initialize two different `SipHashers` and keep the logic here, but that will break backward compatibility.
If we want to use only a single hasher then this code should be simplified as well, and I think we should review if that hashing method doesn't generate too many collisions.

Hawstein: Hi, could you explain in which way using different hashers will break backward compatibility, thank you.

Without the bc problem, using two different independent hashers is indeed the best choice according to Kirsch and Mitzenmacher's paper.

tomusdrw: A bloom filter of non-empty accounts is stored as part of the state and loaded via `Bloom::from_parts`. If we change the `bloom_hash` method it will affect `check` and therefore the restored `bitmap` will have no correlation to accounts we are checking it against.
The change is possible but will require a database migration similar to: https://github.com/paritytech/parity/blob/2b81b982197d27efeea80c8a176c385114d95fdb/ethcore/src/migrations/v10.rs#L17

tomusdrw: After speaking with some team members we decided that it would be best to leave the hashing logic as is (i.e. avoid migration) but simplify it as much as we can (perhaps gaining some performance).
So you can probably avoid calling `item.hash(&mut sip)` twice (just copy the value for the next iteration) and avoid storing the `SipHasher` alltogether (as it now suggests that the hasher state is actually maintained, which is not true because it's being cloned before use).

Hawstein: @tomusdrw Thanks for explaining it.

I just updated the PR according to your suggestion.

NikVolf:
I can't think of any such test besides constant comparison like this:

```
let bloom = Bloom::new(128, 16);
bloom.set(<anything in 0..128 range>, <anything T: Hash>)
assert_eq!(bloom.drain_journal().drain(), <constant from previous version>)
```

and It would be nice if it will be added before merge

Hawstein:
Sure, test added. @NikVolf @tomusdrw

NikVolf:
@Hawstein 

Great! Awesome set of samples also 👍

Hawstein:
Follow up:

I guess we should add backward compatibility test for the `Bloom::from_parts` too since it's another way to create the Bloom instance. Adding test for it avoids people change the implementation in `from_parts`. And just as @tomusdrw said, non-empty accounts state will be loaded via `Bloom::from_parts`.

Sorry for not considering the comprehensive aspects and add it in this PR. If you guys consider it necessary, I can create another PR for it.  @tomusdrw @NikVolf

tomusdrw:
@Hawstein it would be good to have such test indeed. Feel free to submit another PR if you are game.

Hawstein:
@tomusdrw PR created.

PR #6420, opened 2017-08-30
eth_call returns output of contract creations

tomusdrw:
Required to optimize token balances fetch in a light client friendly mode. @ngotchac please have a look.


ethcore/src/executive.rs
```diff
@@ -294,7 +308,8 @@ impl<'a, B: 'a + StateBackend, E: Engine + ?Sized> Executive<'a, B, E> {
 					data: None,
 					call_type: CallType::None,
 				};
-				(self.create(params, &mut substate, &mut tracer, &mut vm_tracer), vec![])
+				let mut out = if output_from_create { Some(vec![]) } else { None };
+				(self.create(params, &mut substate, &mut tracer, &mut vm_tracer, &mut out), out.unwrap_or_else(Vec::new))
```
ngotchac: We might want to keep the same prototype as for `call`? (`out` between `substate` and `tracer`)

tomusdrw: Good point. Fixed.

ngotchac:
I think it's mostly for light client compatibility. And it's only about 20 more lines, doesn't look that complexe, does it ?

tomusdrw:
@debris The call is run without a tracer (we don't need it), but we are interested in the output (that's what `eth_call` is all about - you submit a call and see what it returned).

Also as @ngotchac tracing is not (and most likely won't be) available for light clients, so we cannot use that in the UI if we want it to be compatible.

debris:
If you don't want to / can't trace everything, we could introduce just another type of tracer, which traces only transaction output

ngotchac:
Wouldn't that add a lot more complexity ?

debris:
ok, after talking with @tomusdrw I understand the code better and understand why the code is implemented as it is

I am ok with the changes, but it would be good to get at least 1 more person to review this code

ethcore/src/executive.rs
```diff
@@ -528,7 +544,7 @@ impl<'a, B: 'a + StateBackend, E: Engine + ?Sized> Executive<'a, B, E> {
 		let mut subvmtracer = vm_tracer.prepare_subtrace(params.code.as_ref().expect("two ways into create (Externalities::create and Executive::transact_with_tracer); both place `Some(...)` `code` in `params`; qed"));
 
 		let res = {
-			self.exec_vm(params, &mut unconfirmed_substate, OutputPolicy::InitContract(trace_output.as_mut()), &mut subtracer, &mut subvmtracer)
+			self.exec_vm(params, &mut unconfirmed_substate, OutputPolicy::InitContract(output.as_mut().or(trace_output.as_mut())), &mut subtracer, &mut subvmtracer)
```
rphmeier: does this mean that whenever saving creation output, trace output won't be saved?

tomusdrw: No. Because `trace_output` is always the same as `output` (if both are selected) we just re-use the same value to get it. Couple of lines below this you can see that trace_output is copied either from `output` or `trace_output`.

PR #6422, opened 2017-08-30
Private transactions integration pr

grbIzl:
This pr is used for integrating all the changes in terms of building private transactions system.
We tried to review changes made on every step, but didn't request the merge into the main brunch till the functionality was ready and tested
So it's ready now. The documentation with detailed description of the solution is available on wiki: https://paritytech.github.io/wiki/Private-Transactions.html


ethcore/src/private_transactions/private_transactions.rs
```diff
@@ -0,0 +1,45 @@
...
+	pub fn new() -> Self {
+		PrivateTransactions {
+			transactions: RwLock::new(Vec::new()),
+		}
+	}
+
+	pub fn import(&self, transaction: UnverifiedTransaction, _peer_id: usize) -> Result<(), Error> {
+		{
+			let mut transactions = self.transactions.write();
+			transactions.push(transaction.clone());
```
rphmeier: why clone?

ethcore/src/private_transactions/private_transactions.rs
```diff
@@ -0,0 +1,45 @@
...
+		}
+		Ok(())
+	}
+
+	pub fn get_list(&self) -> Vec<UnverifiedTransaction> {
+		let transactions = self.transactions.read();
+		transactions.clone()
```
rphmeier: `self.transactions().read().clone()`

ethcore/src/service.rs
```diff
@@ -50,7 +50,9 @@ pub enum ClientIoMessage {
 	/// Take a snapshot for the block with given number.
 	TakeSnapshot(u64),
 	/// New consensus message received.
-	NewMessage(Bytes)
+	NewConsensusMessage(Bytes),
+	/// New private transaction received
+	NewPrivateTransaction(Bytes, usize)
```
rphmeier: trailing comma

sync/src/chain.rs
```diff
@@ -157,8 +157,9 @@ const SNAPSHOT_MANIFEST_PACKET: u8 = 0x12;
 const GET_SNAPSHOT_DATA_PACKET: u8 = 0x13;
 const SNAPSHOT_DATA_PACKET: u8 = 0x14;
 const CONSENSUS_DATA_PACKET: u8 = 0x15;
+const PRIVATE_TRANSACTION_PACKET: u8 = 0x16;
 
-pub const SNAPSHOT_SYNC_PACKET_COUNT: u8 = 0x16;
+pub const SNAPSHOT_SYNC_PACKET_COUNT: u8 = 0x17;
```
rphmeier: that requires a protocol upgrade, doesn't it? @arkpar

rphmeier:
How are private transactions differentiated from regular transactions if they both serialize to the `UnverifiedTransaction` type? I don't really see why this requires custom messages.

ethcore/src/client/chain_notify.rs
```diff
@@ -18,6 +18,12 @@ use ipc::IpcConfig;
 use bigint::hash::H256;
 use bytes::Bytes;
 
+pub enum ChainMessageType {
```
rphmeier: missing documentation

ethcore/src/client/client.rs
```diff
@@ -1215,6 +1218,10 @@ impl BlockChainClient for Client {
 		Ok(results)
 	}
 
+	fn get_private_transactions_provider(&self) -> Arc<PrivateTransactionsProvider> {
```
rphmeier: style: generally we don't prefix getters with `get_`

ethcore/src/private_transactions/private_transactions.rs
```diff
@@ -0,0 +1,54 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use transaction::UnverifiedTransaction;
+use error::Error;
+use parking_lot::RwLock;
+
+/// Storage for private transactions
+pub struct PrivateTransactions {
+	transactions: RwLock<Vec<UnverifiedTransaction>>,
+	signed_transactions:  RwLock<Vec<UnverifiedTransaction>>,
```
rphmeier: I'm a little confused why it needs locks here. Isn't it protected by a mutex within the `Provider`?

ethcore/src/private_transactions/private_transactions.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+impl PrivateTransactions {
+	pub fn new() -> Self {
```
rphmeier: documentation on methods

sync/src/tests/private_transactions.rs
```diff
@@ -0,0 +1,90 @@
...
+use ethcore::spec::Spec;
+use ethcore::miner::MinerService;
+use ethcore::transaction::*;
+use ethcore::account_provider::AccountProvider;
+use ethkey::KeyPair;
+use tests::helpers::*;
+use SyncConfig;
+
+#[test]
+pub fn send_private_transaction() {
```
rphmeier: i don't think test functions need to be public

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+			if let Some(n) = np.upgrade() {
+				f(&*n);
+			}
+		}
+	}
+
+	/// Add private transaction into the store
+	pub fn import_private_transaction(&self, rlp: &[u8], peer_id: usize) -> Result<(), EthcoreError> {
+		let tx: UnverifiedTransaction = UntrustedRlp::new(rlp).as_val()?;
+		// TODO: notify engines about private transactions
```
rphmeier: why?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+		let validator_accounts = self.config.lock().validator_accounts.clone();
+		trace!("Private transaction received");
+		if validator_accounts.is_empty() {
+			self.broadcast_private_transaction(rlp.into());
+			return Ok(());
+		}
+		let private_tx: PrivateTransaction = UntrustedRlp::new(rlp).as_val()?;
+		let contract = private_tx.contract;
+		let contract_validators = self.get_validators(BlockId::Latest, &contract)?;
+		let addresses_intersection = contract_validators
```
kirushik: `iter().find(<predicate>).is_some()` can be written more efficiently: `iter().any(<predicate>)`.

Also, you don't need `filter` and `collect` here, get an `Option` for the first element -- and just check if it is Some.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+			let client = self.client.read().clone().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+			let client = client.upgrade().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+			let original_tx = self.extract_original_transaction(private_tx.clone(), &contract)?;
+			let details_provider = TransactionDetailsProvider::new(&*client as &MiningBlockChainClient);
+			let insertion_time = client.chain_info().best_block_number;
+			// Verify with the first account available
+			let validation_account = addresses_intersection[0].clone();
+			trace!("The following account will be used for verification: {:?}", validation_account);
+			let add_res = self.transactions_for_verification.lock().add_transaction(original_tx, contract, validation_account, private_tx.hash(), &details_provider, insertion_time);
+			match add_res {
```
kirushik: More "rusty" way to write this would be `add_res.and_then(|_| <your Ok branch here>)`
This will keep the `Err` branch without explicit rewriting.

Please also note that `add_res` can be trivially inlined after those changes.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+		SignedPrivateTransaction {
+			private_transaction_hash: private_transaction_hash,
+			r: sig.r().into(),
+			s: sig.s().into(),
+			v: sig.v() as u64 + if let Some(n) = chain_id { 35 + n * 2 } else { 27 },
+		}
+	}
+
+	/// 0 if `v` would have been 27 under "Electrum" notation, 1 if 28 or 4 if invalid.
+	pub fn standard_v(&self) -> u8 { match self.v { v if v == 27 || v == 28 || v > 36 => ((v - 1) % 2) as u8, _ => 4 } }
```
kirushik: I would prefer to have some explicit branches for 27 and 28. The `(v-1)%2` formula is confusing enough by itself...

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+				desc.original_transaction.nonce,
+				desc.original_transaction.gas_price
+			)?;
+			trace!("Last required signature received, public transaction created: {:?}", public_tx);
+			//Sign and add it to the queue
+			let accounts = self.accounts.read().clone().ok_or_else(|| PrivateTransactionError::AccountProviderIsMalformed)?;
+			let accounts = accounts.upgrade().ok_or_else(|| PrivateTransactionError::AccountProviderIsMalformed)?;
+			let chain_id = desc.original_transaction.chain_id();
+			let hash = public_tx.hash(chain_id);
+			let signer = self.config.lock().signer_account.clone();
```
kirushik: You'd better not to `clone()` the whole signer_account here — it should be enough to put a MutexGueard into a variable, and then clone only required fields out of it.

(This approach would work best with a new explicit lexical scope, out of which you would return a tuple `(address, password)`.)

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+			//Add signature to the store
+			match self.transactions_for_signing.lock().add_signature(&private_hash, tx.signature()) {
+				Ok(_) => trace!("Signature stored for private transaction"),
+				Err(err) => trace!("Failed to add signature to signing store, error: {:?}", err),
+			}
+		}
+		Ok(())
+	}
+
+	fn last_required_signature(&self, desc: &PrivateTransactionSigningDesc, sign: Signature) -> Result<bool, EthcoreError>  {
```
kirushik: I challenge you to rewrite this function without a single use of `clone()`!

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+	}
+
+	fn encrypt(&self, _contract_address: &Address, data: &[u8]) -> Result<Bytes, EthcoreError> {
+		//TODO: retrieve key from secret store using contract
+		let init_key: Bytes = "cac6c205eb06c8308d65156ff6c862c62b000b8ead121a4455a8ddeff7248128d895692136f240d5d1614dc7cc4147b1bd584bd617e30560bb872064d09ea325".from_hex().unwrap();
+		let key: Bytes = init_key[..INIT_VEC_LEN].into();
+
+		let iv = initialization_vector();
+		let mut encrypted = Vec::with_capacity(data.len() + iv.len());
+		encrypted.extend(repeat(0).take(data.len()));
```
kirushik: Having iterator here seems excessive, maybe just do `vec![0; data.len()+iv.len()]`, and then use `split_at_mut` to get that buffer for encryption?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+		encrypt(&key, &iv, &data, &mut encrypted);
+		encrypted.extend_from_slice(&iv);
+
+		Ok(encrypted)
+	}
+
+	fn decrypt(&self, _contract_address: &Address, data: &[u8]) -> Result<Bytes, EthcoreError> {
+		let mut encrypted: Bytes = data.to_vec().clone();
```
kirushik: `data.into()` seems more ergonomic, and (possibly) marginally more efficient.

ethcore/src/private_transactions/mod.rs
grbIzl: This was done intentionally in order to release mutex.

tomusdrw:
@grbIzl any progress?

grbIzl:
@tomusdrw I'm working on integration with sstore now. The last commit in terms of this PR made 2 days ago. Why was it put to stale?

tomusdrw:
Oh, sorry. I thought they were done 16 Nov:
image

tomusdrw:
Is it possible to merge this PR partially? It's really huge already and impossible to review.

grbIzl:
Yes, it's possible. What is our strategy for such things? I've asked @debris about partial merges, and we agreed, that's it's better to make it for working code. I use this PR in order integrate and test all changes for this feature till the all components comply to at least the basic scenario. Actually several people already contributed to it. So the last part is sstore integration made by @svyatonik I'm trying to make it working now :-)

sync/src/chain.rs
```diff
@@ -116,10 +116,11 @@ known_heap_size!(0, PeerInfo);
...
 
-const PROTOCOL_VERSION_63: u8 = 63;
-const PROTOCOL_VERSION_62: u8 = 62;
-const PROTOCOL_VERSION_1: u8 = 1;
-const PROTOCOL_VERSION_2: u8 = 2;
+pub const ETH_PROTOCOL_VERSION_63: u8 = 63;
+pub const ETH_PROTOCOL_VERSION_62: u8 = 62;
+pub const PAR_PROTOCOL_VERSION_1: u8 = 1;
+pub const PAR_PROTOCOL_VERSION_2: u8 = 2;
+pub const PAR_PROTOCOL_VERSION_3: u8 = 3;
```
rphmeier: doc strings please

arkpar:
Not for this review, but this module is getting quite big. Better move it to a separate crate if possible.

ethcore/src/error.rs
```diff
@@ -34,6 +34,73 @@ use transaction::Error as TransactionError;
...
+	/// State is not available.
+	StatePruned,
+	/// State is incorrect.
+	StateIncorrect,
+	/// Wrong private transaction type.
+	BadTransactonType,
+	/// Contract does not exist or was not created.
+	ContractDoesNotExist,
+	/// Reference to the client is corrupted.
+	ClientIsMalformed,
```
arkpar: Is this used anywhere? Such errors should just panic anyway

sync/src/tests/helpers.rs
```diff
@@ -184,30 +230,57 @@ pub trait Peer {
 
 	/// Restart sync for a peer.
 	fn restart_sync(&self);
+
+	/// Process the queue of pending io messages
+	fn prune_io_messages(&self);
```
tomusdrw: Maybe `process_all` instead?

sync/src/tests/helpers.rs
```diff
@@ -184,30 +230,57 @@ pub trait Peer {
...
+
+	/// Process selected message from the queue
+	fn process_io_message(&self, message: IOMessage);
```
tomusdrw: Why does it need to be part of the trait? Doesn't look used outside.

sync/src/tests/helpers.rs
```diff
@@ -378,8 +502,32 @@ impl<P> TestNet<P> where P: Peer {
 		}
 	}
 
+	pub fn deliver_io_messages(&mut self) {
+		for peer in 0..self.peers.len() {
+			if !self.peers[peer].is_io_queue_empty() {
```
tomusdrw: Since we always have a pattern of `is_*_empty` followed by `prune_*` maybe it's better to allow calls to `prune_io_messages` and do the check internally?

That would make the trait much leaner.

sync/src/tests/helpers.rs
```diff
@@ -126,23 +132,63 @@ impl<'p, C> SyncIo for TestIo<'p, C> where C: FlushingBlockChainClient, C: 'p {
...
+	if let Err(e) = client.import_block(b.rlp_bytes()) {
+		panic!("error importing block which is valid by definition: {:?}", e);
+	}
+
+	client.flush_queue();
+	client.import_verified_blocks();
+}
+
+/// Mock for emulution of async run of new blocks
+pub struct NewBlockMessage {
```
tomusdrw: Why public? It's not possible to access any fields nor to create such structure outside of this module.

sync/src/tests/helpers.rs
```diff
@@ -126,23 +132,63 @@ impl<'p, C> SyncIo for TestIo<'p, C> where C: FlushingBlockChainClient, C: 'p {
...
+		&*self.private_provider
+	}
+
 	fn peer_session_info(&self, _peer_id: PeerId) -> Option<SessionInfo> {
 		None
 	}
 
 	fn eth_protocol_version(&self, _peer: PeerId) -> u8 {
-		63
+		ETH_PROTOCOL_VERSION_63
```
tomusdrw: :+1: for constants :)

sync/src/tests/helpers.rs
```diff
@@ -378,8 +502,32 @@ impl<P> TestNet<P> where P: Peer {
...
+
+	pub fn deliver_new_block_messages(&mut self) {
+		for peer in 0..self.peers.len() {
+			if !self.peers[peer].is_new_blocks_queue_empty() {
+				self.peers[peer].prune_new_block_messages();
+			}
+		}
+	}
+
+	pub fn io_queue_empty(&self) -> bool {
```
tomusdrw: Does it have to be public? Seems to be used only in `done`

sync/src/tests/helpers.rs
```diff
@@ -53,11 +57,12 @@ pub struct TestIo<'p, C> where C: FlushingBlockChainClient, C: 'p {
 	pub to_disconnect: HashSet<PeerId>,
 	pub packets: Vec<TestPacket>,
 	pub peers_info: HashMap<PeerId, String>,
+	pub private_provider: Arc<PrivateTransactionProvider>,
```
tomusdrw: Why do we need `Arc`?

ethcore/src/executive.rs
```diff
@@ -573,13 +573,16 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
 		match res {
-			Ok(ref res) if res.apply_state => tracer.trace_create(
-				trace_info,
-				gas - res.gas_left,
-				trace_output.map(|data| output.as_ref().map(|out| out.to_vec()).unwrap_or(data)),
-				created,
-				subtracer.drain()
-			),
+			Ok(ref res) if res.apply_state => {
+				substate.contracts_created.push(created.clone());
```
arkpar: Not really sure about this change. This was a hack initially. As understand the initial design `Substate` is supposed to hold changes made in the context of execution. While creating a contract a result of creation transaction happens after that. There should be a better way to get contract address after execution.

ethcore/Cargo.toml
```diff
@@ -13,9 +13,12 @@ bn = { git = "https://github.com/paritytech/bn" }
 byteorder = "1.0"
 common-types = { path = "types" }
 crossbeam = "0.3"
+env_logger = "0.4"
```
tomusdrw: Why is it needed?

ethcore/src/executive.rs
tomusdrw: Is it used only for tests? Might be better to set up a tracer for tests instead.

ethcore/src/executive.rs
```diff
@@ -803,7 +805,8 @@ mod tests {
 
 		let FinalizationResult { gas_left, .. } = {
 			let mut ex = Executive::new(&mut state, &info, &machine);
-			ex.create(params, &mut substate, &mut None, &mut NoopTracer, &mut NoopVMTracer).unwrap()
+			let output = BytesRef::Fixed(&mut[0u8;0]);
```
tomusdrw: Why is it required?

ethcore/src/executive.rs
tomusdrw: I see it was moved from `externalities`. Seems it makes sense.

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+	/// Generate unique contract key && encrypt passed data. Encryption can only be performed once.
+	fn encrypt(
+		&self,
+		contract_address: &Address,
+		contract_address_signature: &Signature,
+		requester: &Address,
+		unlocked_accounts: Arc<AccountProvider>,
+		initialisation_vector: &H128,
+		plain_data: &[u8]
+	) -> Result<Bytes, EthcoreError>;
```
tomusdrw: Can we have some more specificc error here? Handling all possible `EthcoreErrors` is not feasible.

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+	base_url: Option<String>,
+	threshold: u32,
+	sessions: Mutex<HashMap<Address, EncryptionSession>>,
+}
+
+impl SecretStoreEncryptor {
+	/// Create new encryptor with empty parameters.
+	pub fn empty() -> Result<Self, EthcoreError> {
+		Ok(SecretStoreEncryptor {
+			client: FetchClient::new()
```
tomusdrw: Fetch client should be re-used, otherwise it spawns a bunch of threads iirc which is suboptimal.

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+			threshold: 0,
+			sessions: Mutex::new(HashMap::new()),
+		})
+	}
+
+	/// Create new encryptor
+	pub fn new(key_server_url: Option<String>, threshold: u32) -> Result<Self, EthcoreError> {
+		Ok(SecretStoreEncryptor {
+			client: FetchClient::new()
+				.map_err(|e| EthcoreError::PrivateTransaction(PrivateTransactionError::Encrypt(format!("{}", e))))?,
```
tomusdrw: Is it really an `Encrypt` error?

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+			return Err(EthcoreError::PrivateTransaction(PrivateTransactionError::Encrypt(response.status().canonical_reason().unwrap_or("unknown").into())));
+		}
+
+		// read HTTP response
+		let mut result = String::new();
+		response.read_to_string(&mut result)?;
+
+		// response is JSON string (which is, in turn, hex-encoded, encrypted Public)
+		let result_len = result.len();
+		if result_len == 0 || &result[0..1] != "\"" || &result[result_len - 1..result_len] != "\"" {
```
tomusdrw: Please no. If it's a JSON please use `serde_json` to deserialize it.

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+			return Err(EthcoreError::PrivateTransaction(PrivateTransactionError::Encrypt(format!("Invalid SecretStore response: {}", result))));
+		}
+		let encrypted_bytes: ethjson::bytes::Bytes = result[1..result_len-1].parse().map_err(|e| EthcoreError::PrivateTransaction(PrivateTransactionError::Encrypt(e)))?;
+
+		// decrypt Public
+		let decrypted_bytes = unlocked_accounts.decrypt(*requester, None, &ethcrypto::DEFAULT_MAC, &encrypted_bytes)?;
+		let decrypted_key = Public::from_slice(&decrypted_bytes);
+
+		// and now take x coordinate of Public as a key
+		let key: Bytes = (*decrypted_key)[..INIT_VEC_LEN].into();
```
tomusdrw: What if it's shorter than `INIT_VEC_LEN`?

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+		});
+		Ok(key)
+	}
+
+	fn obtained_key(&self, contract_address: &Address) -> Option<Bytes> {
+		let mut sessions = self.sessions.lock();
+		let stored_session = sessions.entry(*contract_address);
+		match stored_session {
+			Entry::Occupied(session) => {
+				if Instant::now() > session.get().end_time {
```
tomusdrw: But it's not removed from `self.sessions`, is it? Perhaps we need some kind of garbage collection here, otherwise it opens up an attack vector (where multiple sessions are expired, but never gced)

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+		contract_address_signature: &Signature,
+		requester: &Address,
+		unlocked_accounts: Arc<AccountProvider>
+	) -> Result<Bytes, EthcoreError> {
+		// check if the key was already cached
+		if let Some(key) = self.obtained_key(contract_address) {
+			return Ok(key);
+		}
+		// key id in SS is H256 && we have H160 here => expand with assitional zeros
+		let mut contract_address_extended = Vec::with_capacity(32);
```
arkpar:
```
let contract_address_extended: H256 = contract_address.into();
```
should also work

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+//! Private transactions module.
+/// Export encryptor module.
+pub mod encryptor;
+/// Export the private_transactions module.
+pub mod private_transactions;
+
+extern crate ethabi;
+
+pub use self::encryptor::{Encryptor, SecretStoreEncryptor, DummyEncryptor};
+pub use self::private_transactions::*;
```
tomusdrw: We're trying to avoid such re-exports. Makes it a bit difficult to analyze.

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	fn obtained_key(&self, contract_address: &Address) -> Option<Bytes> {
+		let mut sessions = self.sessions.lock();
+		let stored_session = sessions.entry(*contract_address);
+		match stored_session {
+			Entry::Occupied(session) => {
+				if Instant::now() > session.get().end_time {
+					session.remove_entry();
+					return None;
```
arkpar: superfluous `return` here and below

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+use transaction::{SignedTransaction, Transaction, Action, UnverifiedTransaction};
+use client::{Client, BlockChainClient, ChainNotify, ChainMessageType, BlockId, MiningBlockChainClient};
+use account_provider::AccountProvider;
+use service::ClientIoMessage;
+use error::{PrivateTransactionError};
+use ethcore_miner::transaction_queue::{TransactionDetailsProvider as TransactionQueueDetailsProvider, AccountDetails};
+use miner::MinerService;
+use native_contracts::Private as Contract;
+use trace::{Tracer, VMTracer};
+use rustc_hex::FromHex;
```
tomusdrw: Seems like a lot of imports, perhaps it's possible to split remainder of this file into separate files?

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+				}
+				else {
```
arkpar: should be on a previous line

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+					hash: tx_hash,
+					contract_address: None,
+					status_code: 0,
+				})
+			}
+		}
+	}
+
+	/// Try to unlock account using stored passwords
+	fn unlock_account(&self, account: &Address) -> Result<bool, EthcoreError> {
```
tomusdrw: Why do we need account unlocking? It opens up a bunch of race condition attacks.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+use account_provider::AccountProvider;
+use service::ClientIoMessage;
+use error::{PrivateTransactionError};
+use ethcore_miner::transaction_queue::{TransactionDetailsProvider as TransactionQueueDetailsProvider, AccountDetails};
+use miner::MinerService;
+use native_contracts::Private as Contract;
+use trace::{Tracer, VMTracer};
+use rustc_hex::FromHex;
+
+// Source avaiable at https://gist.github.com/arkpar/5c8fda407c491163a38aeb90c7fac1d2
```
arkpar: Is this still true? I suppose the contract has changed since then.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+	pub hash: H256,
+	/// Created contract address if any.
+	pub contract_address: Option<Address>,
+	/// Execution status.
+	pub status_code: u8,
+}
+
+/// Manager of private transactions
+pub struct Provider {
+	encryptor: RwLock<Arc<Encryptor>>,
```
tomusdrw: Since every part of this struct requires synchronization, why not wrap the entire struct into `Mutex/RwLock` instead?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+
+/// Initialization vector length.
+const INIT_VEC_LEN: usize = 16;
+
+/// Private transaction message call to the contract
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct PrivateTransaction {
```
arkpar: Would nice to move all transaction definitions to a separate module.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+		*self.accounts.write() = Some(accounts);
+	}
+
+	/// Sets encryptor
+	pub fn set_encryptor(&self, encryptor: Arc<Encryptor>) {
+		*self.encryptor.write() = encryptor.clone();
+	}
+
+	/// Sets provider's config.
+	pub fn set_config(&self, config: ProviderConfig) -> Result<(), EthcoreError> {
```
arkpar: Is there a need to be able to `set_config` after creation? Why not just pass config to `new`?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+		let client = client.upgrade().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+		let ready_transactions = self.transactions_for_verification.lock().ready_transactions();
+		let fetch_nonce = |a: &Address| client.latest_nonce(a);
+		for transaction in ready_transactions {
+			let transaction_hash = transaction.hash();
+			match self.transactions_for_verification.lock().private_transaction_descriptor(&transaction_hash) {
+				Ok(desc) => {
+					match self.config.read().validator_accounts.iter().find(|&&account| account == desc.validator_account) {
+						Some(account) => {
+							let accounts = self.accounts.read().clone().ok_or_else(|| PrivateTransactionError::AccountProviderIsMalformed)?;
```
tomusdrw: That's pretty deeply nested. Perhaps it's possible to use some early exists instead?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+	}
+
+	/// Register client reference
+	pub fn register_client(&self, client: Weak<Client>) {
+		*self.client.write() = Some(client);
+	}
+
+	/// Register accounts provider
+	pub fn register_account_provider(&self, accounts: Weak<AccountProvider>) {
+		*self.accounts.write() = Some(accounts);
```
tomusdrw: Why can't we require that in the constructor?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+				.any(|&validator| validator == address))
+			.next() {
+			None => {
+				// Not for verification, broadcast further to peers
+				self.broadcast_private_transaction(rlp.into());
+				return Ok(());
+			},
+			Some(&validation_account) => {
+				trace!("Private transaction taken for verification");
+				let client = self.client.read().clone().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
```
tomusdrw: Perhpas pass `client` as a parameter instead to avoid circular dependencies? (Similar to how it's done in miner (not saying it's perfect, but I think it's more clear than having `Mutex<Option<Weak<Client>>>`)

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+
+	/// Retrieve and verify the first available private transaction for every sender
+	fn process_queue(&self) -> Result<(), EthcoreError> {
+		let client = self.client.read().clone().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+		let client = client.upgrade().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+		let ready_transactions = self.transactions_for_verification.lock().ready_transactions();
+		let fetch_nonce = |a: &Address| client.latest_nonce(a);
+		for transaction in ready_transactions {
+			let transaction_hash = transaction.hash();
+			match self.transactions_for_verification.lock().private_transaction_descriptor(&transaction_hash) {
```
arkpar: Should not `self.transactions_for_verification` be locked once in the beginning? I can think of all sorts of races happening here.

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+		}
+	}
+
+	/// Register client reference
+	pub fn register_client(&self, client: Weak<Client>) {
+		*self.client.write() = Some(client);
+	}
+
+	/// Register accounts provider
+	pub fn register_account_provider(&self, accounts: Weak<AccountProvider>) {
```
arkpar: Would it be possible to make these registration functions mutable and get rid of `RwLock`s?

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+		let address = keccak(&contract_address.rlp_bytes());
+		let (iv, _) = address.split_at(INIT_VEC_LEN);
+		H128::from_slice(iv)
+	}
+
+	fn sign_contract_address(&self, contract_address: &Address) -> Result<Signature, EthcoreError> {
+		let accounts = self.accounts.read().clone().ok_or_else(|| PrivateTransactionError::AccountProviderIsMalformed)?;
+		let accounts = accounts.upgrade().ok_or_else(|| PrivateTransactionError::AccountProviderIsMalformed)?;
+		// key id in SS is H256 && we have H160 here => expand with assitional zeros
+		let mut contract_address_extended = Vec::with_capacity(32);
```
arkpar: A simple H160 to H256 should do here

ethcore/src/private_transactions/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+
+	fn execute_private<T, V>(&self, transaction: &SignedTransaction, options: TransactOptions<T, V>, block: BlockId) -> Result<PrivateExecutionResult<T, V>, EthcoreError>
+		where
+			T: Tracer,
+			V: VMTracer,
+	{
+		let client = self.client.read().clone().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+		let client = client.upgrade().ok_or_else(|| PrivateTransactionError::ClientIsMalformed)?;
+		let mut env_info = client.env_info(block).ok_or(PrivateTransactionError::StatePruned)?;
+		env_info.gas_limit = U256::max_value();
```
arkpar: Should be a sensible limit, otherwise a DOS is possible.

sync/src/chain.rs
```diff
@@ -2198,6 +2215,53 @@ impl ChainSync {
...
+	/// Called when peer sends us new private transaction packet
+	fn on_private_transaction(&self, io: &mut SyncIo, peer_id: PeerId, r: &UntrustedRlp) -> Result<(), PacketDecodeError> {
+		if !self.peers.get(&peer_id).map_or(false, |p| p.can_sync()) {
+			trace!(target: "sync", "{} Ignoring packet from unconfirmed/unknown peer", peer_id);
+			return Ok(());
+		}
+
+		trace!(target: "sync", "Received private transaction packet from {:?}", peer_id);
+		if let Err(e) = io.private_transactions_provider().import_private_transaction(r.as_raw()) {
+			debug!("Ignoring the message, error queueing: {}", e);
```
arkpar: `target: "sync"`

sync/src/chain.rs
```diff
@@ -2198,6 +2215,53 @@ impl ChainSync {
...
+	/// Called when peer sends us signed private transaction packet
+	fn on_signed_private_transaction(&self, io: &mut SyncIo, peer_id: PeerId, r: &UntrustedRlp) -> Result<(), PacketDecodeError> {
+		if !self.peers.get(&peer_id).map_or(false, |p| p.can_sync()) {
+			trace!(target: "sync", "{} Ignoring packet from unconfirmed/unknown peer", peer_id);
+			return Ok(());
+		}
+
+		trace!(target: "sync", "Received signed private transaction packet from {:?}", peer_id);
+		if let Err(e) = io.private_transactions_provider().import_signed_private_transaction(r.as_raw()) {
+			debug!("Ignoring the message, error queueing: {}", e);
```
arkpar: `target: "sync"`

ethcore/src/private_transactions/encryptor.rs
svyatonik: The owner of `SecretStoreEncryptor` instance (`PrivateTransactionsProvider`) is reused (encryptor is only created when `set_config` is called here: https://github.com/paritytech/parity/pull/6422/files#diff-471e80c21627e94e8648551d9c5364f2R286 and `set_config` is only called once). So looks like it isn't a problem

ethcore/src/private_transactions/encryptor.rs
```diff
@@ -0,0 +1,300 @@
...
+/// Duration of storing retrieved keys (in ms)
+const ENCRYPTION_SESSION_DURATION: u64 = 30 * 1000;
+
+/// Trait for encryption/decryption operations.
+pub trait Encryptor: Send + Sync + 'static {
+	/// Generate unique contract key && encrypt passed data. Encryption can only be performed once.
+	fn encrypt(
+		&self,
+		contract_address: &Address,
+		contract_address_signature: &Signature,
```
svyatonik: It seems that `contract_address_signature` could be generated inside these methods (`encrypt` and `decrypt`) since both `AccountProvider` and `requester` are passed here. It will decrease number of arguments && localize all `Address` -> `H256` conversions + private key operations.

ethcore/src/private_transactions/encryptor.rs
svyatonik: @grbIzl I don't have a test env for this functionality, but looks like this should work:
```
let encrypted_bytes: ethjson::bytes::Bytes = result.parse().map_err(|e| EthcoreError::PrivateTransaction(PrivateTransactionError::Encrypt(e)))?;
```
Don't know - why I've added this initially. Could you please check if this works?

ethcore/src/private_transactions/encryptor.rs
svyatonik: `decrypted_key` is `Public` (64-bytes) and `INIT_VEC_LEN` is only 16 (bytes). Don't think additional check is required, because both code (range access) and constant are module-local.

grbIzl:
Comments fixed:

- Functionality moved to the separate crate. 
- Lib was refactored in order to get rid off complex references
- New contract with nonces (for replay attack protection) integrated
- Other fixes

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+	/// Create a new provider.
+	pub fn new(client: Arc<Client>, accounts: Arc<AccountProvider>, encryptor: Arc<Encryptor>, config: ProviderConfig) -> Result<Self, PrivateTransactionError> {
+		Ok(Provider {
+			config: config,
+			notify: RwLock::new(Vec::new()),
+			transactions_for_signing: Mutex::new(SigningStore::new()),
+			transactions_for_verification: Mutex::new(VerificationStore::new()),
+			client: client,
+			accounts: accounts,
+			encryptor: encryptor,
```
debris: recent version of rust has nice syntax sugar, so you can write:

```rust
Provider {
  config,
  client,
  accounts,
  encryptor,
  ..Default::default()
}
```

and it will initialize fields in exactly the same way

private_transactions/src/encryptor.rs
```diff
@@ -0,0 +1,318 @@
...
+		plain_data.extend(repeat(0).take(cypher_len - INIT_VEC_LEN));
+		ethcrypto::aes::decrypt(&key, &iv, cypher, &mut plain_data);
+
+		Ok(plain_data)
+	}
+}
+
+/// Dummy encryptor.
+#[derive(Default)]
+pub struct DummyEncryptor;
```
debris: this structure is `pub` and therefore a part of the crate. It's used in tests located in multiple files. Maybe instead of exposing it a part of crate, move all tests to `tests/` directory and define this structure over there

private_transactions/src/error.rs
```diff
@@ -0,0 +1,157 @@
...
+use ethereum_types::Address;
+use trie::TrieError;
+use ethcore::account_provider::SignError;
+use ethcore::error::ExecutionError;
+use transaction::Error as TransactionError;
+use ethkey::Error as KeyError;
+
+#[derive(Debug)]
+/// Errors concerning private transaction processing.
+pub enum PrivateTransactionError {
```
debris: use [error_chain](https://crates.io/crates/error-chain). We don't do manual error conversion in parity any more

debris: error_chain will also add `std::error::Error` and `std::fmt::Display` error implementations so it will be easier integrate it with external modules

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+			ethabi::Param { name: "s".into(), kind: ethabi::ParamType::Bytes },
+		]};
+
+		let tokens = [
+			ethabi::Token::Array(validators.iter().map(|a| ethabi::Token::Address(a.clone())).collect()),
+			ethabi::Token::Bytes(code),
+			ethabi::Token::Bytes(storage),
+		];
+
+		constructor.encode_input(constructor_code, &tokens).expect("Input is always valid")
```
debris: it's better to use `private::PrivateContract` which is typesafe. Ebove function could be replaced with:

```rust
let private = private::PrivateContract::default();
private.constructor(code, validators)
```

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+			ethabi::Token::Array(signatures.iter().map(|s| {
+				let mut v: [u8; 32] = [0; 32];
+				v[31] = s.v();
+				ethabi::Token::Uint(v.into())
+			}).collect()),
+			ethabi::Token::Array(signatures.iter().map(|s| ethabi::Token::FixedBytes(s.r().to_vec())).collect()),
+			ethabi::Token::Array(signatures.iter().map(|s| ethabi::Token::FixedBytes(s.s().to_vec())).collect()),
+		];
+
+		function.encode_input(&tokens).expect("Input is always valid")
```
debris: same, don't use raw `ethabi` if you have a generated code

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+			gas_price: gas_price,
+			value: 0.into(),
+			data: Self::generate_set_state_call(signatures, state)
+		})
+	}
+
+	/// Call into private contract.
+	pub fn private_call(&self, block: BlockId, transaction: &SignedTransaction) -> Result<Executed, PrivateTransactionError> {
+		let result = self.execute_private(transaction, TransactOptions::with_no_tracing(), block)?;
+		Ok(result.result)
```
debris: no need for `?`. one line is enought
```rust
self.execute_private(transaction, TransactOptions::with_no_tracing(), block)
```

private_transactions/src/messages.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Message with private transaction encrypted
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct PrivateTransaction {
+	/// Encrypted data
+	pub encrypted: Bytes,
+	/// Address of the contract
+	pub contract: Address,
+}
+
+impl Decodable for PrivateTransaction {
```
debris: use `#[derive(RlpDecodable)]` instead

private_transactions/src/messages.rs
```diff
@@ -0,0 +1,127 @@
...
+			return Err(DecoderError::RlpIncorrectListLen);
+		}
+		Ok(PrivateTransaction {
+			encrypted: d.val_at(0)?,
+			contract: d.val_at(1)?,
+		})
+	}
+}
+
+impl Encodable for PrivateTransaction {
```
debris: use `#[derive(RlpEncodable)]` instead

private_transactions/src/messages.rs
```diff
@@ -0,0 +1,127 @@
...
+	/// Signature of the validator
+	/// The V field of the signature
+	v: u64,
+	/// The R field of the signature
+	r: U256,
+	/// The S field of the signature
+	s: U256,
+}
+
+impl Decodable for SignedPrivateTransaction {
```
debris: use `#[derive(RlpDecodable)]` instead

private_transactions/src/messages.rs
```diff
@@ -0,0 +1,127 @@
...
+		Ok(SignedPrivateTransaction {
+			private_transaction_hash: d.val_at(0)?,
+			v: d.val_at(1)?,
+			r: d.val_at(2)?,
+			s: d.val_at(3)?,
+		})
+	}
+}
+
+impl Encodable for SignedPrivateTransaction {
```
debris: use `#[derive(RlpEncodable)]` instead

private_transactions/src/private_transactions.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Storage for private transactions for signing
+pub struct SigningStore {
+	/// Transactions and descriptors for signing
+	transactions: HashMap<H256, PrivateTransactionSigningDesc>,
+}
+
+impl SigningStore {
+	/// Creates new store
+	pub fn new() -> Self {
```
debris: if `new()` takes no params, it's more idiomatic to use `derive(Default)`

sync/src/api.rs
```diff
@@ -337,28 +344,29 @@ impl NetworkProtocolHandler for SyncProtocolHandler {
...
 	}
 
 	fn timeout(&self, io: &NetworkContext, _timer: TimerToken) {
-		self.sync.write().maintain_peers(&mut NetSyncIo::new(io, &*self.chain, &*self.snapshot_service, &self.overlay));
-		self.sync.write().maintain_sync(&mut NetSyncIo::new(io, &*self.chain, &*self.snapshot_service, &self.overlay));
-		self.sync.write().propagate_new_transactions(&mut NetSyncIo::new(io, &*self.chain, &*self.snapshot_service, &self.overlay));
+		let mut io = NetSyncIo::new(io, &*self.chain, &*self.snapshot_service, Some(self.private_tx_provider.clone()), &self.overlay);
+		self.sync.write().maintain_peers(&mut io);
+		self.sync.write().maintain_sync(&mut io);
+		self.sync.write().propagate_new_transactions(&mut io);
```
debris: previously there were 3 different structs created whereas now, there is only one being used. Does it affect `sync` in any way?

sync/src/sync_io.rs
```diff
@@ -66,19 +70,22 @@ pub struct NetSyncIo<'s, 'h> where 'h: 's {
...
 		chain: &'s BlockChainClient,
 		snapshot_service: &'s SnapshotService,
+		private_tx_provider: Option<Arc<PrivateTransactionProvider>>,
 		chain_overlay: &'s RwLock<HashMap<BlockNumber, Bytes>>) -> NetSyncIo<'s, 'h> {
 		NetSyncIo {
 			network: network,
 			chain: chain,
 			snapshot_service: snapshot_service,
 			chain_overlay: chain_overlay,
+			private_tx_provider: private_tx_provider,
```
debris: why is it optional? It's always initialized to some, even in tests

ethcore/src/client/private_notify.rs
```diff
@@ -0,0 +1,26 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use error::TransactionImportError;
+
+/// Represent private transactions handler inside the client
+pub trait PrivateNotify : Send + Sync {
+	/// fires when private transaction message queued via client io queue
+	fn private_transaction_queued(&self) -> Result<(), TransactionImportError> {
+		// does nothing by default
+		Ok(())
```
debris: this trait has only one function. I believe it's not a good pattern to have a default implementation for the only one function

ethcore/src/test_helpers.rs
```diff
@@ -14,12 +14,14 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
+//! Set of different helpers for client tests
```
debris: this should not be a part of public ethcore interface

private_transactions/src/encryptor.rs
```diff
@@ -0,0 +1,318 @@
...
+		plain_data: &[u8]
+	) -> Result<Bytes, PrivateTransactionError> {
+		// retrieve the key, try to generate it if it doesn't exist yet
+		let key = match self.retrieve_key("", false, contract_address, accounts.clone()) {
+			Ok(key) => Ok(key),
+			Err(PrivateTransactionError::EncryptionKeyNotFound(_)) => {
+				trace!("Key for account wasnt found in sstore. Creating. Address: {:?}", contract_address);
+				self.retrieve_key(&format!("/{}", self.config.threshold), true, contract_address, accounts.clone())
+			}
+			Err(err) => Err(err),
```
debris:
```rust
if let Err(PrivateTransactionError::EncryptionKeyNotFound(_)) = key {
  // do what you do
}
```

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+			}
+		}
+	}
+
+	/// Calculate hash from united private state and contract nonce
+	fn calculate_state_hash(&self, state: &Bytes, nonce: U256) -> H256 {
+		let state_hash = keccak(state);
+		let mut state_buf = [0u8; 64];
+		state_buf[..32].clone_from_slice(&state_hash);
+		state_buf[32..].clone_from_slice(&H256::from(nonce).0);
```
debris: `.0` at the end is redundant cause `H256` implements `Deref<[u8]>`

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+		keccak(&state_buf.as_ref())
```
debris: `as_ref()` is redundant

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+	}
+
+	/// Try to unlock account using stored passwords
+	fn unlock_account(&self, account: &Address) -> Result<bool, PrivateTransactionError> {
```
debris: this function never returns `Err` therefore `Result` is redundant

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+		let private_tx: PrivateTransaction = UntrustedRlp::new(rlp).as_val()?;
+		let contract = private_tx.contract;
+		let contract_validators = self.get_validators(BlockId::Latest, &contract)?;
+
+		match contract_validators
+			.iter()
+			.filter(|&&address| validator_accounts
+				.iter()
+				.any(|&validator| validator == address))
+			.next() {
```
debris: 4 lines above should be replaced with:

```rust
.find(|address| validator_accounts.contains(address))
```

ethcore/src/client/client.rs
```diff
@@ -909,6 +911,11 @@ impl Client {
 		*self.io_channel.lock() = io_channel;
 	}
 
+	/// Get io channel
+	pub fn get_io_channel(&self) -> Arc<Mutex<IoChannel<ClientIoMessage>>> {
```
debris: Exposing `Mutex`'s in public interface is a bad pattern. Use [channels](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) instead

private_transactions/src/lib.rs
```diff
@@ -0,0 +1,782 @@
...
+		}
+	}
+
+	/// Private transaction for validation added into queue
+	pub fn on_private_transaction_queued(&self) -> Result<(), PrivateTransactionError> {
+		self.process_queue()
+	}
+
+	/// Retrieve and verify the first available private transaction for every sender
+	fn process_queue(&self) -> Result<(), PrivateTransactionError> {
```
debris: i do not understand why some errors are propagated upstream while the others are (not)silently ignored with traces

ethcore/src/lib.rs
```diff
@@ -150,6 +151,8 @@ pub mod snapshot;
 pub mod spec;
 pub mod state;
 pub mod state_db;
+// Test helpers made public for usage outside ethcore
+pub mod test_helpers;
```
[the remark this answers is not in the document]
grbIzl: I tried that. Unfortunately there are a lot of uses of ethcore classes inside test_helpers. So it's not possible directly to move it to the separate crate and include as dependency for ethcore (because of cyclic dependencies). Implemented solution with public mod (I agree not the best one) was a compromise for complete refactoring of all the helpers

sync/src/sync_io.rs
grbIzl: No, you're not correct. See chain.rs tests. All of them don't have Private Provider and initialize it with None.

ethcore/src/lib.rs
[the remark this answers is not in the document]
debris: than `ethcore` tests which is `test_helpers` should be moved to `ethcore/tests/` directory. And you can have a dependency cycle like this:

`test_helpers [dependencies] ` contains `ethcore`
`ethcore [dev-dependencies]` contains `test_helpers`
`private-transactions [dev-dependencies]` contains `test_helpers`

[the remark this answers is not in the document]
grbIzl: Unfortunately, it seems, that Rust compiler doesn't allow such structure. The problem in details: let's say, we have struct A in ethcore crate and fn func() -> A in test_helpers crate. If to call func inside ethcore crate, the compiler returns the error: "expected struct A, found struct ethcore::A". Discussed the issue with @debris offline and decided to leave this as for now, other possible alternatives like massive copy-pasting of helpers seem worse.

ethcore/src/client/client.rs
```diff
@@ -1175,6 +1182,22 @@ impl Client {
 			BlockId::Pending => Some(self.chain.read().best_block_number() + 1),
 		}
 	}
+
+	/// Sets handler for private messages
+	pub fn set_private_notify(&self, notify: Arc<PrivateNotify>) {
+		*self.private_notify.write() = Some(Arc::downgrade(&notify));
```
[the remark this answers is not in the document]
grbIzl: It's done through the setter because PrivateProvider also requires Client in its constructor.

private_transactions/src/encryptor.rs
grbIzl: I've moved tests to the separate directory, it's good for code's structure anyway. But DummyEncryptor left public for the crate, because it's used outside (in sync tests)

private_transactions/src/lib.rs
grbIzl: error[E0277]: the trait bound `[u8; 64]: std::convert::AsRef<[u8]>` is not satisfied
   --> private_transactions/src/lib.rs:238:3
    |
238 |   keccak(&state_buf)
    |   ^^^^^^ the trait `std::convert::AsRef<[u8]>` is not implemented for `[u8; 64]`
    |
    = help: the following implementations were found:
              <[T; <unevaluated[]>] as std::convert::AsRef<[T]>>
              <[T; <unevaluated[]>] as std::convert::AsRef<[T]>>
              <[T; <unevaluated[]>] as std::convert::AsRef<[T]>>
              <[T; <unevaluated[]>] as std::convert::AsRef<[T]>>
            and 30 others
    = note: required because of the requirements on the impl of `std::convert::AsRef<[u8]>` for `&[u8; 64]`
    = note: required by `hash::keccak`

private_transactions/src/encryptor.rs
grbIzl: It's not equivalent to the current code, because the error needs to be propagated

private_transactions/src/lib.rs
grbIzl: I need to return result.result in this method and abort execution on the first line, if result is not calculated

ethcore/private-tx/src/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+	Client, ChainNotify, ChainMessageType, ClientIoMessage, BlockId,
+	MiningBlockChainClient, ChainInfo, Nonce, CallContract
+};
+use ethcore::account_provider::AccountProvider;
+use ethcore_miner::transaction_queue::{TransactionDetailsProvider as TransactionQueueDetailsProvider, AccountDetails};
+use ethcore::miner::MinerService;
+use ethcore::trace::{Tracer, VMTracer};
+use rustc_hex::FromHex;
+
+// Source avaiable at https://github.com/paritytech/contracts/blob/master/contracts/PrivateContract.sol
```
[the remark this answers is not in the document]
grbIzl: Replaced with the actual link

ethcore/private-tx/src/messages.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+impl SignedPrivateTransaction {
+	/// Construct a signed private transaction message
+	pub fn new(private_transaction_hash: H256, sig: Signature, chain_id: Option<u64>) -> Self {
+		SignedPrivateTransaction {
+			private_transaction_hash: private_transaction_hash,
+			r: sig.r().into(),
+			s: sig.s().into(),
+			v: sig.v() as u64 + if let Some(n) = chain_id { 35 + n * 2 } else { 27 },
```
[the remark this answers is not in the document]
grbIzl: Not sure. Only with some refactoring of Transaction struct (like creating of TransactionSignature trait and so on). Didn't want to touch so vital code for that.

ethcore/private-tx/src/lib.rs
```diff
@@ -0,0 +1,659 @@
...
+		let contract_validators = self.get_validators(BlockId::Latest, &contract)?;
+
+		let validation_account = contract_validators
+			.iter()
+			.find(|address| self.validator_accounts.contains(address));
+
+		match validation_account {
+			None => {
+				// Not for verification, broadcast further to peers
+				self.broadcast_private_transaction(rlp.into());
```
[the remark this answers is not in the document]
grbIzl: Good point. Some basic validation should be introduced. Is it okay to do that in the separate PR?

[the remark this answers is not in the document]
grbIzl: Added an issue for that https://github.com/paritytech/parity/issues/8250

ethcore/private-tx/src/messages.rs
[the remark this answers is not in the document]
tomusdrw: I specifically meant the chain-id encoding within `v` I think the rest is fine.

ethcore/transaction/src/transaction.rs
```diff
@@ -78,6 +78,25 @@ pub enum Condition {
...
+	/// Adds chain id into v
+	pub fn add_chain_replay_protection(v: u64, chain_id: Option<u64>) -> u64 {
+		v + if let Some(n) = chain_id { 35 + n * 2 } else { 27 }
+	}
+
+	/// Returns refined v
+	/// 0 if `v` would have been 27 under "Electrum" notation, 1 if 28 or 4 if invalid.
+	pub fn check_replay_protection(v: u64) -> u8 {
+		match v {
+			v if v == 27 => 0 as u8,
```
niklasad1: 27 => 0 as u8

ethcore/transaction/src/transaction.rs
```diff
@@ -78,6 +78,25 @@ pub enum Condition {
...
+			v if v == 28 => 1 as u8,
```
niklasad1: 28 => 1 as u8

ethcore/transaction/src/transaction.rs
grbIzl: removed

ethcore/transaction/src/transaction.rs
grbIzl: removed

debris:
to simplify merge with master, I removed I made `ethcore` depend on `kvdb_rocksb` again... cc #8320

Issue #6418, opened 2017-08-30
Tracking issue util refactor


debris:
- [ ] @Hawstein journaldb and rocksdb dependencies should be separated from util.
- [ ] @Hawstein all trie traits from util should be separated from util.
- [ ] migrations should be a library
- [x] @debris -> replace sha3::Hashable trait and all of usages of it with a simple function called `hash`
- [x] @debris -> this `hash` function should be a standalone library importing C impl of sha3 and tiny-keccak (for updatable sha3s)
- [x] @folsen util should not reexport bigint
- [x] @Hawstein util should not reexport ansi_term
- [x] @axelchalon -> util should not reexport HeapSizeOf
- [x] @Hawstein util should not reexport ParkingLot
- [x] @debris util should not reexport `vector::*`
- [x] @debris triehash should be separated from util
- [x] @Hawstein semantic_version should be separated from util
- [x] @Hawstein timer::PerfTimer should be a part of ethcore
- [x] @debris error::UtilError should be replaced by error_chain
- [ ] remove public reexports of traits modules
- ~flatten `util` subcrates~ (controversial)

rphmeier:
I'm not convinced about flattening util subcrates. Why not group non-blockchain-specific utilities within a single folder?

debris:
ok, let's keep all subcrates in util directory for now. We'll discuss it again someday

debris:
I'll take

```
- [ ] replace sha3::Hashable trait and all of usages of it with a simple function called `hash`
- [ ] this `hash` function should be a standalone library importing C impl of sha3 and tiny-keccak (for updatable sha3s)
```

NikVolf:
Also fixed hash should be crates.io library dependant on bigint and little more

axelchalon:
I'll take `util should not reexport HeapSizeOf` for now

debris:
I will take `triehash should be separated from util`

folsen:
I'll take `util should not reexport bigint`

Hawstein:
I'll take `util should not reexport ansi_term` if you guys don't mind:)

Hawstein:
I'll take one more: `util should not reexport ParkingLot`.

debris:
@Hawstein thanks, merged!

Hawstein:
I will take `timer::PerfTimer should be a part of ethcore`.

Hawstein:
I will take `semantic_version should be separated from util`.

Hawstein:
[...]

PR #6423, opened 2017-08-31
replace trait Hashable with fn keccak

debris:
part of #6418 

changes:
- renamed const variables `SHA3_XXX` to `KECCAK_XXX`
- `TransactionView`, `BlockView` and `HeaderView` has now `pub fn hash()` which returns their hash. Usage of it does not require import of `Hashable` trait
- removed `trait Hashable`
- added `fn keccak`. Using function is more idiomatic. Also it's more flexible and can be easily used with options and iterators. We can now write:
  ```rust
  fn old_usage(option: Option<H256>, vector: Vec<H256>) {
    let x = option.map(|o| o.sha3());
    // or 
    let x = option.as_ref().map(Hashable::sha3);

    let y = vector.into_iter().map(|i| i.sha3()).collect();
    // or
    let y = vector.iter().map(Hashable::sha3).collect();
  }

  fn new_usage(option: Option<H256>, vector: Vec<H256>) {
    let x = option.map(keccak);
    // or
    let x = option.as_ref().map(keccak);

    let y = vector.into_iter().map(keccak).collect();
    // or
    let y = vector.iter().map(keccak).collect();
  }
  ```
- `fn keccak` is no longer a part of `util`. This is a critical step required to make modules independent

@paritytech/core-devs can you please review the code? The sooner we agree on those changes and merge them, the less hassle it will be to keep it up-to-date


Vurich:
This is just a renaming pass and still uses the C implementation, right? There is unsafe code that relies on the precise semantics of the C SHA3 (specifically, that it uses an internal buffer and so the input and output arrays can be the same/overlapping). I think that it might be better if we don't do that (and use the output buffer directly to reduce `memcpy` calls) but for now that's the assumption we make.

util/hash/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+pub fn keccak<T: AsRef<[u8]>>(s: T) -> H256 {
+	let mut result = [0u8; 32];
+	keccak_into(s, &mut result);
+	H256(result)
+}
+
+pub fn keccak_into<T: AsRef<[u8]>>(s: T, dest: &mut [u8]) {
+	let input = s.as_ref();
+	unsafe {
+		keccak_256(dest.as_mut_ptr(), dest.len(), input.as_ptr(), input.len());
```
Vurich: Why do we ignore the return value of `keccak_256`? Probably at least worth putting a comment explaining that.

util/network/src/connection.rs
```diff
@@ -312,16 +312,16 @@ impl EncryptedConnection {
 		}
 		let mut key_material = H512::new();
 		shared.copy_to(&mut key_material[0..32]);
-		nonce_material.sha3_into(&mut key_material[32..64]);
-		key_material.sha3().copy_to(&mut key_material[32..64]);
-		key_material.sha3().copy_to(&mut key_material[32..64]);
+		keccak_into(&nonce_material, &mut key_material[32..64]);
```
Vurich: Is this really more ergonomic? What's wrong with using a trait here? I wouldn't have an opinion either way if this was original code but since it's the stated purpose of the PR it deserves analysis.

util/hash/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+	pub fn keccak_512(out: *mut u8, outlen: usize, input: *const u8, inputlen: usize) -> i32;
+}
+
+pub fn keccak<T: AsRef<[u8]>>(s: T) -> H256 {
+	let mut result = [0u8; 32];
+	keccak_into(s, &mut result);
+	H256(result)
+}
+
+pub fn keccak_into<T: AsRef<[u8]>>(s: T, dest: &mut [u8]) {
```
Vurich: Bikeshed: `_into` implies taking ownership of `T`'s data and transforming it into something rather than writing it to a buffer, this confused me when first reading this code. I recommend `write_keccak`, but it's a matter of taste. I think it's good to avoid names that already have a common, conflicting use in Rust code.

debris:
> This is just a renaming pass and still uses the C implementation, right? 

Yes. Cause the C implementation is still ~10% faster than we achieved with rust.

> There is unsafe code that relies on the precise semantics of the C SHA3 (specifically, that it uses an internal buffer and so the input and output arrays can be the same/overlapping). I think that it might be better if we don't do that (and use the output buffer directly to reduce memcpy calls) but for now that's the assumption we make.

I'm not sure I understand your point. Do you want to return internal buffer from keccak function? It's not possible with this type of hash function.

Vurich:
Essentially all I'm saying is that some unsafe code relies on the hash function allowing input and output to overlap, so as long as that still holds we're all good. It looks like it does, I just wanted to make certain.

debris:
Yes, nothing has changed internally

util/network/src/connection.rs
debris: It's one of the purposes :) imo, there is no point in using the trait if simple function is enough

ethash/src/compute.rs
```diff
@@ -336,7 +336,7 @@ fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64)
 			);
 
 			// compute sha3-512 hash and replicate across mix
```
folsen: s/sha3/keccak ?

ethash/src/compute.rs
```diff
@@ -214,14 +214,14 @@ fn fnv_hash(x: u32, y: u32) -> u32 {
...
 
-fn sha3_512(input: &[u8], output: &mut [u8]) {
-	unsafe { sha3::sha3_512(output.as_mut_ptr(), output.len(), input.as_ptr(), input.len()) };
+fn keccak_512(input: &[u8], output: &mut [u8]) {
+	unsafe { hash::keccak_512(output.as_mut_ptr(), output.len(), input.as_ptr(), input.len()) };
 }
 
-fn sha3_512_inplace(input: &mut [u8]) {
+fn keccak_512_inplace(input: &mut [u8]) {
 	// This is safe since `sha3_*` uses an internal buffer and copies the result to the output. This
```
folsen: s/sha3/keccak

rphmeier:
LGTM now that grumbles have been addressed

PR #6424, opened 2017-08-31
fix modexp bug: return 0 if base is zero

Hawstein:
It currently returns `result` which is `BigUint::one()`.

### DIFF
--- a/ethcore/src/builtin.rs
+++ b/ethcore/src/builtin.rs
@@ -267,6 +267,34 @@ impl Impl for Ripemd160 {
 	}
 }
 
+// calculate modexp: exponentiation by squaring. the `num` crate has pow, but not modular.
+fn modexp(mut base: BigUint, mut exp: BigUint, modulus: BigUint) -> BigUint {
+	use num::Integer;
+
+	match (base.is_zero(), exp.is_zero()) {
+		(_, true) => return BigUint::one(), // n^0 % m
+		(true, false) => return BigUint::zero(), // 0^n % m, n>0
+		(false, false) if modulus <= BigUint::one() => return BigUint::zero(), // a^b % 1 = 0.
+		_ => {}
+	}
+
+	let mut result = BigUint::one();
+	base = base % &modulus;
+
+	// fast path for base divisible by modulus.
+	if base.is_zero() { return BigUint::zero() }
+	while !exp.is_zero() {
+		if exp.is_odd() {
+			result = (result * &base) % &modulus;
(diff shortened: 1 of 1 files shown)


NikVolf:
indeed

ethcore/src/builtin.rs
```diff
@@ -309,7 +309,7 @@ impl Impl for ModexpImpl {
 			base = base % &modulus;
 
 			// fast path for base divisible by modulus.
-			if base.is_zero() { return result }
+			if base.is_zero() { return BigUint::zero() }
```
rphmeier: isn't this done here?

https://github.com/paritytech/parity/pull/6424/files#diff-f01cea79b5ea6391c260249a4cf801a5R303

and tested here

https://github.com/paritytech/parity/pull/6424/files#diff-f01cea79b5ea6391c260249a4cf801a5L660

Hawstein: No. the place you mentioned is the original base. But here the base is calculated via `base % &modulus`. See: 

https://github.com/paritytech/parity/pull/6424/files#diff-f01cea79b5ea6391c260249a4cf801a5R309

rphmeier: Ah, right. A test would be good

Hawstein: But I think we can just move `base = base % &modulus;` up to the first line in this function. and remove the test here. All the condition will be tested at:

https://github.com/paritytech/parity/pull/6424/files#diff-f01cea79b5ea6391c260249a4cf801a5R303

WDYT? @rphmeier @debris

rphmeier: Then it would panic if the modulo was zero, which is protected at the given point.

Hawstein: Ah, right. 👍

Hawstein: I am struggling for the test. There are more stuffs besides the `modexp` function. I can easily test the `modexp` function, however since it is just a inner function, I have to test the `ModexpImpl::execute`. I will be really appreciated if there are some instructions.

gavofyork:
waiting for a test

ethcore/src/builtin.rs
NikVolf: oh just move the function outside to the mod level and add a test, there are plenty of support functions there already

Hawstein:
@gavofyork @NikVolf The PR is updated according to your suggestion.

PR #6428, opened 2017-08-31
triehash is separated from util

debris:
part of #6418


ethcore/src/snapshot/block.rs
```diff
@@ -23,7 +23,7 @@ use hash::keccak;
 use views::BlockView;
 use rlp::{DecoderError, RlpStream, UntrustedRlp};
 use util::{Bytes, H256};
-use util::triehash::ordered_trie_root;
+use triehash::ordered_trie_root;
```
NikVolf: used only once in the code, could be referenced by global path instead of `use`, as in the test

Hawstein: Is there some rules for the usage of `use` in parity project? (it currently seems not though)
For example:
* make the self-host libs a group and third party libs a group
* sort it alphabetically within the group
* make it clear when to use `*`
* make it clear when to use global path
* others

I think we can make those rules in the CONTRIBUTION.md and make contributors to follow those rules.

debris: We don't have such rules currently, although I totally agree that we should have them

PR #6435, opened 2017-09-01
remove re-export of parking_lot in util

Hawstein:
Part of #6418


gavofyork:
doesn't build.

Hawstein:
@gavofyork Build problem is fixed.

PR #6441, opened 2017-09-02
Transaction permissioning

arkpar:
This adds a smart contract that governs transaction permissions. The contract accepts sender address and returns a mask of allowed transaction types for this address.

### DIFF
--- a/ethcore/src/ethereum/ethash.rs
+++ b/ethcore/src/ethereum/ethash.rs
@@ -17,7 +17,7 @@
 use std::path::Path;
 use std::cmp;
 use std::collections::{BTreeMap, HashMap};
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use hash::{KECCAK_EMPTY_LIST_RLP};
 use ethash::{quick_get_difficulty, slow_get_seedhash, EthashManager};
 use util::*;
@@ -29,13 +29,15 @@ use trace::{Tracer, ExecutiveTracer, RewardType};
 use header::{Header, BlockNumber};
 use state::CleanupMode;
 use spec::CommonParams;
-use transaction::UnverifiedTransaction;
+use transaction::{UnverifiedTransaction, SignedTransaction};
 use engines::{self, Engine};
 use evm::Schedule;
 use ethjson;
 use rlp::{self, UntrustedRlp};
 use vm::LastHashes;
 use semantic_version::SemanticVersion;
+use tx_filter::{TransactionFilter};
+use client::{Client, BlockChainClient};
 
 /// Parity tries to round block.gas_limit to multiple of this constant
 pub const PARITY_GAS_LIMIT_DETERMINANT: U256 = U256([37, 0, 0, 0]);
@@ -141,6 +143,7 @@ pub struct Ethash {
 	ethash_params: EthashParams,
 	builtins: BTreeMap<Address, Builtin>,
 	pow: EthashManager,
+	tx_filter: Option<TransactionFilter>,
 }
 
 impl Ethash {
@@ -152,6 +155,7 @@ impl Ethash {
 		builtins: BTreeMap<Address, Builtin>,
 	) -> Arc<Self> {
 		Arc::new(Ethash {
+			tx_filter: TransactionFilter::from_params(&params),
 			params,
 			ethash_params,
 			builtins,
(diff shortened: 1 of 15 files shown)


NikVolf:
How about short description of the feature? :)

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	/// Clear cached permissions.
+	pub fn clear_cache(&self) {
+		self.permission_cache.lock().clear();
+	}
+
+	pub fn register_client(&self, client: Weak<BlockChainClient>) {
+		*self.client.write() = Some(client);
+
```
NikVolf: redundant empty line

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+			}
+		)
+	}
+
+	/// Clear cached permissions.
+	pub fn clear_cache(&self) {
+		self.permission_cache.lock().clear();
+	}
+
+	pub fn register_client(&self, client: Weak<BlockChainClient>) {
```
NikVolf: Undocumented public method

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+		*self.client.write() = Some(client);
+
+	}
+
+	pub fn transaction_allowed(&self, parent_hash: &H256, transaction: &SignedTransaction) -> bool {
```
NikVolf: Undocumented public method

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+use futures::{self, Future};
+use spec::CommonParams;
+use transaction::{Action, SignedTransaction};
+use hash::KECCAK_EMPTY;
+
+const MAX_CACHE_SIZE: usize = 4096;
+
+mod tx_permissions {
+    pub const _ALL: u32 = 0xffffffff;
+    pub const NONE: u32 = 0x0;
```
NikVolf: spaces instead of tabs

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+use util::{Mutex, Address, H256, Bytes, RwLock};
+use futures::{self, Future};
+use spec::CommonParams;
+use transaction::{Action, SignedTransaction};
+use hash::KECCAK_EMPTY;
+
+const MAX_CACHE_SIZE: usize = 4096;
+
+mod tx_permissions {
+    pub const _ALL: u32 = 0xffffffff;
```
NikVolf: why `_ALL` and not `ALL` ?

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub const NONE: u32 = 0x0;
+	pub const BASIC: u32 = 0b00000001;
+	pub const CALL: u32 = 0b00000010;
+	pub const CREATE: u32 = 0b00000100;
+	pub const _PRIVATE: u32 = 0b00001000;
```
NikVolf: why `_PRIVATE` and not `PRIVATE` ?

ethcore/src/tx_filter.rs
rphmeier: could use the bitflags crate for this

ethcore/src/tx_filter.rs
arkpar: To suppress compiler warning. This is currently unused

ethcore/src/tx_filter.rs
arkpar: I did not want to introduce another dependency just for this. This type is used internally so a simple `u32` is enough.

ethcore/src/tx_filter.rs
```diff
@@ -0,0 +1,233 @@
...
+					trace!("Permissions required: {}, got: {}", tx_type, permissions);
+					permissions & tx_type != 0
+				}
+			}
+		})
+	}
+}
+
+impl ChainNotify for TransactionFilter {
+	fn new_blocks(&self, imported: Vec<H256>, _invalid: Vec<H256>, _enacted: Vec<H256>, _retracted: Vec<H256>, _sealed: Vec<H256>, _proposed: Vec<Bytes>, _duration: u64) {
```
rphmeier: now that we're adding so many new chain-notify implementations, would be nice to not be cloning each of these vectors each time, especially since most of them only read.

arkpar: Agreed, although most of the time they only contain 0 or 1 elements.

ethcore/src/ethereum/ethash.rs
```diff
@@ -437,13 +441,28 @@ impl Engine for Arc<Ethash> {
...
 		engines::ConstructedVerifier::Trusted(Box::new(self.clone()))
 	}
 
 	fn snapshot_components(&self) -> Option<Box<::snapshot::SnapshotComponents>> {
 		Some(Box::new(::snapshot::PowSnapshot::new(SNAPSHOT_BLOCKS, MAX_SNAPSHOT_BLOCKS)))
 	}
+
+	fn register_client(&self, client: Weak<Client>) {
+		if let Some(ref filter) = self.tx_filter {
+			filter.register_client(client as Weak<BlockChainClient>);
```
rphmeier: engines are used in light clients too. requiring a full node to run consensus logic should have been a blocker for this PR.

rphmeier: furthermore, I don't see why TX filtering would be coupled with the Ethash memory-hard PoW consensus system in the first place.

rphmeier: it makes more sense in a permissioned environment anyway

Issue #6443, opened 2017-09-02
Define contribution rules


debris:
and place them in CONTRIBUTION.md

[discussion](https://github.com/paritytech/parity/pull/6428#discussion_r136693669)

5chdn:
Can help drafting the file if someone roughly explains what's about to go into that file.

Tabs not spaces? :-p

https://github.com/paritytech/parity/commits/714f01fda737b24f2be3ff20837899c7f0a213f5/CONTRIBUTING.md

folsen:
I will create a draft and share for commenting then PR

folsen:
@5chdn i went looking for this quite hard and somehow missed it! Maybe link from the readme as well or something

5chdn:
It's automatically integrated into the Github interface. Happy to link it somewhere else.

![screenshot at 2018-06-06 16-05-21](https://user-images.githubusercontent.com/15729797/41043140-8c90b7e8-69a3-11e8-994d-9acd5e344152.png)

folsen:
@5chdn man I feel so old, I remember a time when github was barely helpful at all

PR #6459, opened 2017-09-04
Refactor: Don't reexport bigint from util

folsen:
Part of #6418

This PR removes
```
pub use bigint::prelude::*;		
pub use bigint::hash;
```
from `util/src/lib.rs` and then changes all the affected places to directly import what's needed from bigint.


NikVolf:
So now you are reexporting things from ethcore_bigint?

I think better way is to use U256 directly from `bigint` crate on crates.io where possible

debris:
> I think better way is to use U256 directly from bigint crate on crates.io where possible

I believe this could be a separate pr. For now we don't need this that much, cause in most places we use both anyway :)

NikVolf:
@debris 
Ok, but almost the same job will be performed twice then ;)

debris:
@NikVolf nah, second time will be easier, cause there won't be any wildcard imports

something like that

```
s/bigint::hash::H256/ethcore_bigint::H256
s/bigint::prelude::U256/bigint::U256 
```

+ filling out missing import should be enough

folsen:
> second time will be easier, cause there won't be any wildcard imports

Yeah actually one of the most annoying things here were glob imports, should we put in a rule to never use wildcard imports? Would make sense to me.

debris:
@folsen Definitely! Another one for #6443

rphmeier:
We've cut down a ton on the number of glob imports, but `use util::*` is still pretty common. I don't think any new glob imports are being introduced but a bunch are still hanging around.

debris:
I'll merge before more conflicts are introduced

PR #6465, opened 2017-09-05
fixed master

debris:
bug introduced in #6459, cause we have merged 2 exclusive prs at the same time


debris:
temporarily reverted #6370 (as discussed with @arkpar). The pr should be revisited once it's compatible with master

PR #6467, opened 2017-09-05
WASM runtime update

NikVolf:
- WASM contracts will be able to utilize various blockchain runtime externs (gas_limit, difficulty, coinbase, etc..)
- corresponding happy test
- also updates to use latest parity-wasm and corresponding refactoring


NikVolf:
@svyatonik 
I want gas price to match exactly when we don't change wasm itself, it's consensus-critical issue

rphmeier:
@NikVolf I agree. Related: what made the gas costs for examples smaller? I was worried that they aren't remaining consistent.

NikVolf:
@rphmeier 
oh we change the `pwasm-std` currently a lot, so the way wasm produced in tests repository will change from time to time

PR #6468, opened 2017-09-05
light-poa round 2: allow optional casting of engine client to full client

rphmeier:
The TX filter will now only allow transactions on full nodes by attempting to upgrade the `EngineClient`. But I still don't think that this logic belongs in the consensus layer at all.


rphmeier:
@arkpar only the diff of the last 2 commits is important. the rest is just the changes you've looked at before.

PR #6470, opened 2017-09-05
SecretStore: use random key to encrypt channel + session-level nonce

svyatonik:
Given: SecretStore = set of KeyServers. Third-party KeyServer can't just 'enter' this set - it must be added to the set, using some administrative protocol. To ensure that we're connecting to servers from this set only, every KeyServer has its own KeyPair. By signing some_data with this KeyPair, KS can prove that it is the one that we're trying to connect.

Previously: during handshake, KS1 && KS2 a new Channel.KeyPair was computed from KS1.KeyPair && KS2.KeyPair. This Channel.KeyPair was used to encrypt the channel between these two key servers. So even after restart key was the same => it was possible to *successfully* replay previously captured messages.

This PR outline:
1) random encryption keys: during handshake *random* keys are generated and these are used to compute shared KeyPair
2) naive inter-session replay protection: every node has its own `session_counter: usize`, which is used as session-level nonce. When session is created on master node, `session_counter` is increased, and passed along with every session message. Every message for the same session must contain this number, or else session will fail. When node receives next session initialization message, it checks if its nonce is *larger* than previous nonce, received from the same node: if so, session is created, else creation fails (this implies that this KeyServer could be master for at most `usize::MAX` sessions between consequent KeyServer restarts, which I suppose is enough).
3) added missing header version check

So, in assumption that trying to replay same-session messages could only lead to session failure (this is out of scope of this PR, but *mostly* true), there's no way to *successfully* use messages from previous sessions.

### DIFF
--- a/secret_store/src/key_server_cluster/cluster_sessions.rs
+++ b/secret_store/src/key_server_cluster/cluster_sessions.rs
@@ -16,7 +16,7 @@
 
 use std::time;
 use std::sync::{Arc, Weak};
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::collections::{VecDeque, BTreeSet, BTreeMap};
 use parking_lot::RwLock;
 use ethkey::{Public, Secret, Signature};
@@ -68,6 +68,18 @@ pub struct ClusterSessions {
 	acl_storage: Arc<AclStorage>,
 	/// Make faulty generation sessions.
 	make_faulty_generation_sessions: AtomicBool,
+	/// Always-increasing sessions counter. Is used as session nonce to prevent replay attacks:
+	/// 1) during handshake, KeyServers generate new random key to encrypt messages
+	/// => there's no way to use messages from previous connections for replay attacks
+	/// 2) when session (of any type) is started, master node increases its own session counter and broadcasts it
+	/// 3) when slave KeyServer receives session initialization message, it checks that new nonce is larger than previous (from the same master)
+	/// => there's no way to use messages from previous sessions for replay attacks
+	/// 4) KeyServer checks that each session message contains the same nonce that initialization message
+	/// Given that: (A) handshake is secure and (B) session itself is initially replay-protected
+	/// => this guarantees that sessions are replay-protected.
+	session_counter: AtomicUsize,
+	/// Maximal session nonce, received from given connection.
+	max_session_nonce: RwLock<BTreeMap<NodeId, u64>>,
 }
 
 /// Active sessions container.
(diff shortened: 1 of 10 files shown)


secret_store/src/key_server_cluster/cluster_sessions.rs
```diff
@@ -317,6 +342,21 @@ impl ClusterSessions {
 		}
 		Ok(encrypted_data)
 	}
+
+	/// Check or generate new session nonce.
+	fn check_session_nonce(&self, master: &NodeId, session_nonce: Option<u64>) -> Result<u64, Error> {
```
NikVolf: it's already in a session-ish mod, could be named just `check_nonce`

secret_store/src/key_server_cluster/decryption_session.rs
```diff
@@ -59,6 +59,8 @@ struct SessionCore {
 	pub key_share: DocumentKeyShare,
 	/// Cluster which allows this node to send messages to other nodes in the cluster.
 	pub cluster: Arc<Cluster>,
+	/// Session-level nonce.
+	pub session_nonce: u64,
```
NikVolf: could be named just `nonce`, since it's inside a `SessionCore` struct

secret_store/src/key_server_cluster/decryption_session.rs
```diff
@@ -95,8 +97,10 @@ pub struct SessionParams {
 	pub key_share: DocumentKeyShare,
 	/// ACL storage.
 	pub acl_storage: Arc<AclStorage>,
-	/// Cluster
+	/// Cluster.
 	pub cluster: Arc<Cluster>,
+	/// Session nonce.
+	pub session_nonce: u64,
```
NikVolf: ditto

secret_store/src/key_server_cluster/encryption_session.rs
```diff
@@ -71,6 +73,8 @@ pub struct SessionParams {
 	pub key_storage: Arc<KeyStorage>,
 	/// Cluster
 	pub cluster: Arc<Cluster>,
+	/// Session nonce.
+	pub session_nonce: u64,
```
NikVolf: and here also

secret_store/src/key_server_cluster/encryption_session.rs
```diff
@@ -252,6 +265,14 @@ impl SessionImpl {
 
 		Ok(())
 	}
+
+	/// Check session nonce.
+	fn check_session_nonce(&self, message_session_nonce: u64) -> Result<(), Error> {
```
NikVolf: it's `SessionImpl`, makes sense to call method just `check_nonce`

NikVolf:
Small grumble about avoiding long naming, logic looks ok, but would like another look from @keorn maybe

PR #6471, opened 2017-09-05
Fix slow balances

ngotchac:
Better fetching of the token balances.

1. The UI now makes one call for about 64 balances updates, thanks to this https://github.com/paritytech/contracts/pull/76
2. It caches the tokens info it got from the registry (only the META data can be changed, ie. the images)
3. A bunch of fixes on how the balances are retrieved.


js/src/redux/providers/tokensActions.js
```diff
@@ -14,56 +14,225 @@
...
+      })
+      .catch((error) => {
+        console.warn('tokens::fetchTokens', error);
+      });
+  };
+}
+
 export function fetchTokens (_tokenIndexes, options = {}) {
-  const tokenIndexes = uniq(_tokenIndexes || []);
+  const tokenIndexes = uniq(_tokenIndexes || []).sort();
```
tomusdrw: Why does it need to be sorted?

js/src/redux/providers/tokensActions.js
```diff
@@ -14,56 +14,225 @@
...
 
-    return tokenReg.getInstance()
-      .then((tokenRegInstance) => {
-        const promises = tokenIndexes.map((id) => fetchTokenInfo(api, tokenRegInstance, id));
+    return tokenReg.getContract()
+      .then((tokenRegContract) => {
+        let promise = Promise.resolve();
 
-        return Promise.all(promises);
+        tokenChunks.forEach((tokenChunk) => {
```
tomusdrw:
```js
return tokenChunks.reduce((promise, tokenChunk) => {
  return promise.then(() => ...);
}, Pormise.resolve());
```
?

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+        const tokenData = tokenReg.instance.token
+          .decodeOutput(rawTokenData)
+          .map((t) => t.value);
+
+        const [ address, tag, format, name ] = tokenData;
+        const image = images[index];
+
+        const token = {
+          address,
+          id: sha3(address + tokenIndex).slice(0, 10),
```
tomusdrw: might be good to extract `id` to some function, it's the 3rd time it's used

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+  });
+
+  const calls = requests.map((req) => api.eth.call(req));
+  const imagesPromise = fetchTokensImages(api, tokenReg, tokenIndexes);
+
+  let results;
+
+  return Promise.all(calls)
+    .then((_results) => {
+      results = _results;
```
tomusdrw: Looks a bit hacky, why not:
```js
return imagesPromise.then(images => {
  return results.map(...) 
});
```
or
```js
  return imagesPromise.then(images => ({ images, results }));
}).then(({images, results}) => {

});
```

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+}
+
+export function fetchTokensImages (api, tokenReg, tokenIndexes) {
+  const requests = tokenIndexes.map((tokenIndex) => {
+    const metaCalldata = tokenReg.getCallData(tokenReg.instance.meta, {}, [tokenIndex, 'IMG']);
+
+    return { to: tokenReg.address, data: metaCalldata };
+  });
+
+  const calls = requests.map((req) => api.eth.call(req));
```
tomusdrw: Fetching all images and all info looks a bit heavy.

I think we should further optimize it and only fetch the info/images of tokens that are actually displayed (but in a separate PR)

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+ *   {
+ *     [ who ]: { [ tokenId ]: BigNumber } // The balances of `who`
+ *   }
+ */
+export function fetchAccountsBalances (api, tokens, updates) {
+  const accountAddresses = Object.keys(updates);
+
+  // Updates for the ETH balances
+  const ethUpdates = accountAddresses
+    .map((accountAddress) => {
```
tomusdrw: `map/reduce` here seems overcomplicated, `tokenIds` will always have either 0 or 1 element if I'm not mistaken:
```js
const ethUpdates = accountAddresses
   // only accounts fetching eth 
  .filter(address => updates[address].find(tokenId => ETH_TOKEN.id))
  .reduce((nextUpdates, address) => {
     nextUpdates[address] = [ETH_TOKEN.id];
   }, {});
```

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+        nextUpdates[accountAddress] = tokenIds;
+      }
+
+      return nextUpdates;
+    }, {});
+
+  // Updates for Tokens balances
+  const tokenUpdates = Object.keys(updates)
+    .map((accountAddress) => {
+      return updates[accountAddress].filter((tokenId) => tokenId !== ETH_TOKEN.id);
```
tomusdrw: This could be simplified as well:
```js
const tokenUpdates = Object.keys(updates).reduce((nextUpdates, address) => {
   const tokens = updates[address].filter(tokenId => tokenId !== ETH_TOKEN.id);
   if (tokens.length) {
      nextUpdates[address] = tokens;
   }
   return nextUpdates;
}, {});
```

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+
+      return nextUpdates;
+    }, {});
+
+  let ethBalances = {};
+  let tokensBalances = {};
+
+  const ethPromise = fetchEthBalances(api, Object.keys(ethUpdates))
+    .then((_ethBalances) => {
+      ethBalances = _ethBalances;
```
tomusdrw:
```js
const ethPromise = fetchEthBalances(api, ...);
const tokenPromise = Object.keys(tokenUpdates).reduce((promise, address) => {
  return promise.then(balances => {
     return fetchTokensBalances(api, updateTokens, ...).then(newBalances => {
       balances[address] = newBalances[address];
       return balances;
     });
  });
}, Promise.resolve({}));

return Promise.all([ethPromise, tokenPromise]).then(([ethBalances, tokensBalances]) => {
 ...
});
```

tomusdrw: Now that I wrote the code I think there is a bit of redundancy in calling `fetchTokensBalances`:

Why can't we do a single query for all tokens seen in any account and all accounts at once?
Something like:
```js
const updateTokens = uniq(Object.values(tokenUpdates).reduce((all, ids) => all.concat(ids), []));
const accounts = Object.keys(tokenUpdates);
const tokenPromise = fetchTokensBalances(api, updateTokens, accounts);
```
?

js/src/util/tokens/index.js
```diff
@@ -0,0 +1,293 @@
...
+    });
+}
+
+function fetchEthBalances (api, accountAddresses) {
+  const promises = accountAddresses
+    .map((accountAddress) => api.eth.getBalance(accountAddress));
+
+  return Promise.all(promises)
+    .then((balancesArray) => {
+      return balancesArray.reduce((balances, balance, index) => {
```
tomusdrw: In the future it might actually be better to display balances as they come instead of waiting for all of them.

gavofyork:
is this still polling on the JS side? what about using filters?

tomusdrw:
1. It uses filter for transfer even on any non-zero address token in the registry.
2. When the event is received we check if `to` or `from` is one of our accounts
3. If it is we refresh a balance of this particular token for affected accounts.

Another filter is set up to monitor the token registry and update the other filters if there are any changes there.

The main issue was with initial fetch that now is working much better:
1. The token ids are cached (so we don't need to query the token registry 600 times just to get the addresses)
2. Initial balances are fetched in chunks for all accounts - for each chunk of 64 tokens we produce `eth_call` that is able to fetch balances for all accounts. (Fetching all tokens at once takes too long and causes noticable lag).
3. The rest of polling methods is now using `parity_pubsub`, so we don't need to periodicaly check for:
 - best block
 - pending transactions (signer)
 - nodeHealth
 - peers
 - syncStatus

That's from the top of my head, @ngotchac correct me if I'm wrong on anything.

js/src/api/transport/logger.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  get logs () {
+    return this._logs.sort(Logger.sorter);
+  }
+
+  get methods () {
+    const methods = this.logs.reduce((methods, log) => {
+      if (!methods[log.method]) {
+        methods[log.method] = [];
```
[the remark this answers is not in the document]
ngotchac: I'm not sure how this would be better?

js/src/redux/providers/balancesActions.js
```diff
@@ -100,41 +102,98 @@ function notifyBalanceChange (who, fromValue, toValue, token) {
...
+    const tokensAddressMap = Object.values(tokens).reduce((map, token) => {
+      map[token.address] = token;
+      return map;
+    }, {});
 
-    const tokensToUpdate = _tokens || Object.values(tokens);
     const tokenAddresses = tokensToUpdate
       .map((t) => t.address)
-      .filter((address) => address)
+      .filter((address) => address && !/^(0x)?0*$/.test(address))
```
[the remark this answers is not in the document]
ngotchac: It also filters out `0x` and `0x0` addresses, do you think it's a big overhead?

js/src/api/transport/logger.js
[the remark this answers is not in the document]
tomusdrw: Single line instead of 3 = less visual clutter and it's also a common pattern in JS

js/src/redux/providers/balancesActions.js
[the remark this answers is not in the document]
tomusdrw: How come we ever end up with such addresses? Not really an overhead, but I think the code would look a bit cleaner.

js/src/redux/providers/balancesActions.js
```diff
@@ -100,41 +102,98 @@ function notifyBalanceChange (who, fromValue, toValue, token) {
...
+        return;
       }
     }
 
-    log.debug('updating the token filter', addresses, tokenAddresses);
     const promises = [];
+    const updates = {};
+
+    // Fetch all tokens for each new address
+    newAddresses.forEach((address) => {
```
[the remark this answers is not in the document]
ngotchac: It's used to fetch the new tokens / addresses balances, a few lines below :)

js/src/redux/providers/balancesActions.js
```diff
@@ -194,82 +221,129 @@ export function updateTokensFilter (_addresses, _tokens, options = {}) {
...
+        // No accounts to update
+        if (Object.keys(updates).length === 0) {
           return;
         }
 
-        const tokensToUpdate = Object.values(tokens)
-          .filter((t) => tokenAddresses.includes(t.address));
+        Object.keys(updates).forEach((who) => {
+          // Keep non-empty token addresses
+          updates[who] = uniq(updates[who]);
```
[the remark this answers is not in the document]
ngotchac: If multiple transactions for the same token happens on in the same block, I guess we can have some duplicates.

js/src/redux/providers/balancesActions.js
```diff
@@ -194,82 +221,129 @@ export function updateTokensFilter (_addresses, _tokens, options = {}) {
...
+
+        return updates;
+      }, {});
     }
 
-    const updates = addresses.reduce((updates, who) => {
-      updates[who] = tokensToUpdate.map((token) => token.id);
-      return updates;
-    }, {});
+    let start = Date.now();
```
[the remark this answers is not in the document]
ngotchac: It's re-used bellow.

js/src/redux/providers/balancesActions.js
```diff
@@ -194,82 +221,129 @@ export function updateTokensFilter (_addresses, _tokens, options = {}) {
...
+
+            return tokenIds.concat(nextTokenIds);
+          }, []);
+
+        const tokenIndexesToFetch = uniq(tokenIdsToFetch)
+          .filter((tokenId) => tokens[tokenId] && tokens[tokenId].index && !tokens[tokenId].fetched)
+          .map((tokenId) => tokens[tokenId].index);
+
+        if (tokenIndexesToFetch.length > 0) {
+          start = Date.now();
```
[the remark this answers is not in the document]
ngotchac: You mean at the very beginning of the file?

js/src/redux/providers/tokensActions.js
ngotchac: I thought about doing this in a different way, which would have needed a sorted array.... Forgot to remove it!

js/src/util/tokens/index.js
ngotchac: That's actually what's done ! :)

js/src/util/tokens/index.js
ngotchac: Yeah, right now it's first the ETH balances, then tokens. But with Redux, the more we dispatch a new state, the more it re-renders a *lot* of Components...

js/src/util/tokens/index.js
ngotchac: So it's because not all accounts need to update the same tokens.

ngotchac:
Grumbles should be addressed!

js/src/util/tokens/index.js
ngotchac: So I spoke/wrote too fast. It shouldn't re-render components, but will run the dispatcher methods and compare the results for every rendered components, so it's anyway better to do it once than multiple times.

js/src/util/tokens/index.js
ngotchac: Yeah I agree that it's hard to read, and the whole thing could be simplified if we didn't really care about having a fine-tuned fetching. But I guess the scope of the PR was to be able to make _just_ the right number of calls.

js/src/util/tokens/index.js
ngotchac: So it's in `/src/redux/providers/balancesActions.js` > `fetchTokensBalances` : the `fetchTokens` method is called only for the tokens we don't have meta-data/name yet.

`loadTokensBasics` in `tokensActions` is the method that is called on load, and that fetches only the tokens addresses. When the token cache is already filled, it actually only updates the images that might have changed since the last load.

PR #6472, opened 2017-09-06
Random cleanups / improvements to a state

debris:


gavofyork:
until we have CI set up to label according to files changed, please label consensus-sensitive PRs as such.

ethcore/src/state/mod.rs
```diff
@@ -853,27 +849,29 @@ impl<B: Backend> State<B> {
 
 	// load required account data from the databases.
 	fn update_account_cache(require: RequireCache, account: &mut Account, state_db: &B, db: &HashDB) {
-		match (account.is_cached(), require) {
```
rphmeier: is changing this actually a cleanup or improvement? I think it was fine before

rphmeier:
other changes LGTM

gavofyork:
@debris can you address the grumble?

ethcore/src/state/mod.rs
tomusdrw: IMHO it's cleaner now (one less identation, clear indication that already cached accounts and `RequireCache::None` is not handled at all).

debris: @tomusdrw I was about to write exactly the same thing ;)

gavofyork:
@arkpar review?

PR #6483, opened 2017-09-08
elliptic pairing special cases for zero points

cdetrio:
This should fix some elliptic curve test cases. See also https://github.com/paritytech/bn/pull/3

### DIFF
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -14,7 +14,7 @@ build = "build.rs"
 ansi_term = "0.9"
 bit-set = "0.4"
 bloomchain = "0.1"
-bn = { git = "https://github.com/paritytech/bn" }
+bn = { git = "https://github.com/cdetrio/bn", branch = "mont-repr-fix" }
 byteorder = "1.0"
 clippy = { version = "0.0.103", optional = true}
 common-types = { path = "types" }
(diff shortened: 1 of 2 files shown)


ethcore/Cargo.toml
```diff
@@ -14,7 +14,7 @@ build = "build.rs"
 ansi_term = "0.9"
 bit-set = "0.4"
 bloomchain = "0.1"
-bn = { git = "https://github.com/paritytech/bn" }
+bn = { git = "https://github.com/cdetrio/bn", branch = "mont-repr-fix" }
```
rphmeier: could you make a PR to the upstream version so we still only have a paritytech dependency?

ethcore/src/builtin.rs
```diff
@@ -474,22 +474,37 @@ impl Impl for Bn128PairingImpl {
 				let b_a_y = Fq::from_slice(&input[idx*192+160..idx*192+192])
 					.map_err(|_| Error::from("Invalid b argument real coeff y coordinate"))?;
 
-				vals.push((
-					G1::from(
+				let g1_point;
+				if a_x == Fq::zero() && a_y == Fq::zero() {
+					g1_point = G1::zero();
+				} else {
+					//g1_point = G1::from(bn::AffineG1::new(a_x, a_y).expect("Invalid a argument - not on curve"));
```
rphmeier: commented-out code

arkpar:
Thanks, but this is already fixed in https://github.com/paritytech/parity/pull/5855/commits/22269b634d6646ffc5034eb31b1a07eea9b4045c

PR #6498, opened 2017-09-11
Fix broken JavaScript tests

tomusdrw:


tomusdrw:
Rust tests failure unrelated (seems like timing issue).

PR #6510, opened 2017-09-13
Fix detecting hardware wallets.

tomusdrw:


tomusdrw:
Closing, since this will be addressed in #6403

PR #6512, opened 2017-09-13
WASM sha3 test

fckt:


NikVolf:
To whom it may concern, gas used by test contracts is changed again, because we had introduced a couple of changes in tests before (see https://github.com/paritytech/wasm-tests/pull/15, for example)

Issue #6476, opened 2017-09-06
Migrate to latest jsonrpc and futures


tomusdrw:
Includes:
- Get rid of `BoxFuture`
- Get rid of `.boxed()`
- Rewrite `ipfs` server to hyper@latest
- Rewrite `parity-dapps` to hyper@latest

0x7CFE:
What's the status for this issue?

tomusdrw:
Done.

PR #6522, opened 2017-09-15
Update jsonrpc dependencies and rewrite dapps to futures.

### LINKED ISSUE #6261: ignoring connection above current capacity: NoSlabSpace(<hidden seed>)
I'm running:

- **Parity version**: Parity/v1.7.0-beta-5f2cabd-20170727/x86_64-linux-gnu/rustc1.18.0
- **Operating system**: Ubuntu 16.04.3 LTS (GNU/Linux 4.4.0-89-generic x86_64) 
- **And installed**: bash <(curl https://get.parity.io -kL)

This node is one in three of a PoA private setup. On top of the RPC of this node we are running a very complex, and active multi-tenant API layer using web3 to interface with the node. There are quite some watches for events going on.

It has been working splendidly for a while, and then all of a sudden, the api loses connection to the node which displays a streaming list of the same message:

```
ignoring connection above current capacity: NoSlabSpace(<hidden seed>)
```

- Searched the code to see where and why this message was thrown, but could not find it
- Looked for a max connection parameter for the RPC, could not find one (only between nodes)
- Increased the server size from 1GB/1CPU to 4GB/2CPU, it still happens 

I'm looking for 

- a cause, i assume the RPC connections, but i'm not sure. 
- what is the limit, and how to increase it
- other solutions i have not thought about (and preferably not "set less watches" since i need them)

For reference, all relevant configs below
---

toml
```
[parity]
chain = "spec.json"
base_path = "/root/parity0"
[network]
port = 30300

[rpc]
interface = "xx.xx.xx.xx"
hosts = ["all"]
port = 8545

[ui]
interface = "xx.xx.xx.xx"
port = 8180

[websockets]
interface = "xx.xx.xx.xx"

[dapps]
interface = "xx.xx.xx.xx"
hosts = ["all"]
port = 8080

[account]
password = ["node.pwds"]

[mining]
engine_signer = "0x00..."
reseal_on_txs = "none"
gas_floor_target = "47000000"
gas_cap = "62831840"
tx_gas_limit = "62831840"
```
 Note: I'm aware of the depreciation notices on the dapps etc parameters

spec.json

```
{
    "name": "MyNet",
    "engine": {
        "authorityRound": {
            "params": {
                "gasLimitBoundDivisor": "0x400",
                "stepDuration": "1",
                "blockReward": "0x4563918244F40000",
                "validators" : {
                    "list": [
                        "0x00...",
                        "0x00...",
                        "0x00..."
                    ]
                }
            }
        }
    },
    "params": {
        "minGasLimit": "0x1388",
        "networkID" : "0x2323"
    },
    "genesis": {
        "seal": {
            "authorityRound": {
                "step": "0x0",
                "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
            }
        },
        "difficulty": "0x20000",
        "gasLimit": "0x5B8D80"
    },
    "accounts": {
        "0x0000000000000000000000000000000000000001": { "balance": "1", "builtin": { "name": "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
        "0x0000000000000000000000000000000000000002": { "balance": "1", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
        "0x0000000000000000000000000000000000000003": { "balance": "1", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
        "0x0000000000000000000000000000000000000004": { "balance": "1", "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } }
    }
}
```

tomusdrw:
- The HTTP (RPC/Dapps & UI) servers are now based on latest version of hyper (which also makes them way more performant).
- minihttp is not needed any more, http supports multiple server threads as well.
- Updates `futures` to latest version and removes all `.boxed()` and `BoxFuture` occurrences (deprecated).

~Awaiting for https://github.com/paritytech/jsonrpc/pull/199 to be reviewed and merged first (so that we can branch out as `parity-1.8`)~

Closes #6261, #6476


gavofyork:
doesn't build..

dapps/src/handlers/fetch.rs
```diff
@@ -296,100 +316,54 @@ impl<H: ContentValidator, F: Fetch> ContentFetcherHandler<H, F> {
...
+		loop {
+			trace!(target: "dapps", "Polling status: {:?}", self.status);
+			self.status = match mem::replace(&mut self.status, FetchState::Empty) {
+				FetchState::Error(error) => {
+					return Ok(futures::Async::Ready(error.into()));
+				},
+				FetchState::Streaming(response) => {
+					return Ok(futures::Async::Ready(response));
+				},
+				any => any,
```
debris: imo, it's better to write: 

```rust
if let FetchState::Error(ref mut error) = self.status {
  // here do mem::replace on error
  // and return Ok(..)
}

if let FetchState::Streaming(ref mut response) = self.response {
  // here do mem::replace on response
  // and return Ok(..)
}

```

dapps/src/handlers/reader.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use futures::{self, sink, Sink, Future};
+use futures::sync::mpsc;
+use hyper;
+
+type Sender = mpsc::Sender<Result<hyper::Chunk, hyper::Error>>;
+
+const MAX_CHUNK_SIZE: usize = 32 * 1024;
+
+pub struct Reader<R: io::Read> {
```
debris: This structure requires very descriptive documentation. I've spent over 10 minutes looking at it to understand why it's done like that.

And if I understand correctly, `sending` is a sink, which sends data directly to a server client. Is that right?

debris: Also, I can see that `hyper::Response` is generic over `Body`, so maybe there is a way to use it with every structure which implements stream of `Chunks`? Or anything that implements `AsRef<u8>`. Anyway, even if this is not possible now, maybe we should just open an issue in hyper repo. Just to avoid unnecessary allocation, when read value is moved to a heap when creating a `Chunk.

dapps/src/handlers/fetch.rs
tomusdrw: That doesn't give you ownership of `error` required for `return`.

abitrolly:
There are a lot of warnings about `.boxed()` and `BoxFuture` in current builds. Making their removal a separate PR can speed up review.

tomusdrw:
The warnings are most likely coming from `jsonrpc` libraries, that need to be bumped as well - and this refactoring is required for that.
I can split the PR, but I think there is not so many instances of `.boxed()` and `BoxFuture` in current codebase and it wouldn't really make this PR much smaller.

abitrolly:
@tomusdrw exact list of libs with warnings, mined from this log - https://travis-ci.org/cyberFund/cybernode/builds/281880085#L1321

```
✗ ./warncounter.py
('hash', 1)
('reqwest', 8)
('fetch', 5)
('patricia_trie', 169)
('rocksdb', 6)
('vm', 9)
('evm', 398)
```
[warncounter.py](https://gist.github.com/abitrolly/6ab1af736760d78ef6b03a2307a67c55)

abitrolly:
With versions:
```
✗ ./warncounter.py
    1  hash v0.1.0
    8  node-health v0.1.0
    5  fetch v0.1.0
  169  patricia_trie v0.1.0
    6  parity-hash-fetch v1.8.0
    5  vm v0.1.0
    4  parity-whisper v0.1.0
    3  parity-local-store v0.1.0
  368  parity-rpc v1.8.0
   10  parity-rpc-client v1.4.0
   17  rpc-cli v1.4.0
```

tomusdrw:
So because most of the lines are in `parity-rpc` the PR wouldn't be smaller at all. Removing `.boxed()` from that part of code was done more elegantly because of the changes in `jsonrpc-core` (allowing to return `T: IntoFuture` instead of `BoxFuture` everywhere) that is being updated in this PR. So we can fix `.boxed()`, but this PR will change exactly the same lines to further remove it.

That said, I'm fine with doing that if @paritytech/core-devs believe it will help with getting this PR through.

rpc/src/v1/impls/eth.rs
```diff
@@ -456,15 +447,15 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> Eth for EthClient<C, SN, S, M, EM> where
...
+		Box::new(future::done(res))
 	}
 
 	fn block_by_hash(&self, hash: RpcH256, include_txs: bool) -> BoxFuture<Option<RichBlock>, Error> {
-		future::done(self.block(BlockId::Hash(hash.into()), include_txs)).boxed()
+		Box::new(future::done(self.block(BlockId::Hash(hash.into()), include_txs)))
 	}
 
 	fn block_by_number(&self, num: BlockNumber, include_txs: bool) -> BoxFuture<Option<RichBlock>, Error> {
-		future::done(self.block(num.into(), include_txs)).boxed()
```
gavofyork: why the change from `.boxed()` to `Box::new`?

debris: `boxed()` has been deprecated. https://github.com/alexcrichton/futures-rs/issues/228

tomusdrw: The motivation is that `.boxed()` requires `Future + Send + 'static` and sometimes it's not needed. Wrapping with `Box::new` covers both cases.

grbIzl: And usage of boxed causes warning during the compilation. It litters the build log significantly

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -274,10 +274,6 @@ impl ClusterCore {
 		Box::new(net_connect(&node_address, handle, data.self_key_pair.clone(), disconnected_nodes)
 			.then(move |result| ClusterCore::process_connection_result(data, Some(node_address), result))
 			.then(|_| finished(())))
-<<<<<<< HEAD
```
tomusdrw: Oh, thanks!

abitrolly:
> overall, looks fine, but the pr of that size is simply unreviewable

It is still `+2,065 −2,881` with non-trivial changes mixed in.

PR #6552, opened 2017-09-19
Fix number of confirmations for transaction

ngotchac:


jacogr:
Failures unrelated to change (maybe not merged with master or current broken master).

PR #6556, opened 2017-09-19
Fix #6540

### LINKED ISSUE #6540: Parity.js cannot deploy contracts from an unlocked node
*Moved from #6536*

---

I'm running:
- **Parity version**: Parity/v1.8.0-unstable-b731ccea1-20170903/x86_64-linux-gnu/rustc1.19.0
- **And installed**: via `pacman -S parity-git`
- **parity.js** installed via `npm i @parity/parity.js`

---

What I tried:
```js
// adapted from the README.md
const Api = require('@parity/parity.js').Api;
const transport = new Api.Transport.Http('http://localhost:8545');
const api = new Api(transport);
const abi = [{"constant":true,"inputs":[],"name":"owner","outputs":[{"name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"},{"inputs":[],"payable":false,"stateMutability":"nonpayable","type":"constructor"}] // from a small contract
const bcode = '6060604 ... a0029' // large bytecode string
const contract = new api.newContract(abi);
contract.deploy({data: bcode});
```

Expected result: Contract deployed on Kovan network.

Actual result:

The promise is rejected with "request not found", but the transaction [appears to have succeeded](https://kovan.etherscan.io/tx/0x76161b35c213a8ce6d9073f6ece2be7fc5ce31ad9cf910670788811df4925a46):

```
parity_checkRequest(["0x76161b35c213a8ce6d9073f6ece2be7fc5ce31ad9cf910670788811df4925a46"]): -32042: Request not found.                     
pollMethod { e: parity_checkRequest: -32042: Request not found.                                
    at /home/travis/prg/protocol/node_modules/@parity/parity.js/library.js:1:161621            
    at <anonymous>     
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)                       
  code: -32042,        
  type: 'REQUEST_NOT_FOUND',                   
  method: 'parity_checkRequest',               
  text: 'Request not found.' }                 
(node:15338) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 7): e: parity_checkRequest: -32042: Request not found.
```

I am running parity as an unlocked node with the following options:
```sh
ACCOUNT_LIQPROV='0x00360d2b7D240Ec0643B6D819ba81A09e40E5bCd'

parity \
  --chain kovan \
  --author $ACCOUNT_LIQPROV \
  --unlock $ACCOUNT_LIQPROV \
  --password ~/prg/keys/kovan/account_liqprov \
  --rpcport 8545 \
  --force-ui
```

### Update:

Can confirm this also occurs on version `Parity/v1.7.0-beta-5f2cabd6e-20170727/x86_64-linux-gnu/rustc1.19.0`

travs:
Skips the `checkRequest` call if result of `postTransaction` is a tx ID.
If there is a better way to check whether `result` is a tx ID than looking at its length, let's do that.

Made a similar PR over [here](https://github.com/paritytech/parity.js/pull/3), but I can't even get the module installed from that repo, and the latest changes are not as new as those in this repo, so it looks like this one makes more sense to merge to.

### DIFF
--- a/js/src/api/contract/contract.js
+++ b/js/src/api/contract/contract.js
@@ -133,11 +133,11 @@ export default class Contract {
 
         return this._api.parity
           .postTransaction(encodedOptions)
-          .then((result) => {
-            if (result.length !== 66) {
-              statecb(null, { state: 'checkRequest', result });
-              return this._pollCheckRequest(result);
-            } else { return result; }
+          .then((requestId) => {
+            if (requestId.length !== 66) {
+              statecb(null, { state: 'checkRequest', requestId });
+              return this._pollCheckRequest(requestId);
+            } else { return requestId; }
           })
           .then((txhash) => {
             statecb(null, { state: 'getTransactionReceipt', txhash });
--- a/js/src/api/contract/contract.js
+++ b/js/src/api/contract/contract.js
@@ -133,9 +133,11 @@ export default class Contract {
 
         return this._api.parity
           .postTransaction(encodedOptions)
-          .then((requestId) => {
-            statecb(null, { state: 'checkRequest', requestId });
-            return this._pollCheckRequest(requestId);
+          .then((result) => {
+            if (result.length !== 66) {
+              statecb(null, { state: 'checkRequest', result });
+              return this._pollCheckRequest(result);
+            } else { return result; }
           })
           .then((txhash) => {
             statecb(null, { state: 'getTransactionReceipt', txhash });


travs:
Not sure why I keep getting this failing build.
I'm branching from `master`; is this the wrong thing to do?

jacogr:
Failures fixed in https://github.com/paritytech/parity/pull/6553

travs:
@jacogr 
Copy that. Will rebase this guy when #6553 merges

gavofyork:
done

travs:
Rebased and ready to merge (if it passes 🤞 )

js/src/api/contract/contract.js
```diff
@@ -133,9 +133,11 @@ export default class Contract {
 
         return this._api.parity
           .postTransaction(encodedOptions)
-          .then((requestId) => {
-            statecb(null, { state: 'checkRequest', requestId });
-            return this._pollCheckRequest(requestId);
+          .then((result) => {
+            if (result.length !== 66) {
+              statecb(null, { state: 'checkRequest', result });
```
jacogr: Why did we lose the `requestId` naming? This is is that the holder of the statecb checks the `requestId` field which does not exist anywhere anymore.

travs: oh yeah that's my mistake, one sec I will revise

travs: revised 9deac99a718fa351a28ab5b254928ecac9e4522b

jacogr:
LGTM

As for the parity.js repo, it will be removed in favour of the js-{abi, api, jsonrpc} repos. (I do keep those updated from master, at this point they are direct match for what is found there, in the short term they will become the only source of truth.)

travs:
@jacogr Ok great. I want to use `api` from node so I can just install via the instructions in the readme [here](https://github.com/paritytech/js-api)?

jacogr:
@travs Will update the instructions. 

But basically, not published to npm (yet, we are busy working through that with CI, will be in the next week or so), so for now would just need to have a github package.json entry: (Also the way we use is atm with out next-generation UI)

```
{
  ...
  "dependencies" {
    "@parity/api": "paritytech/js-api"
    ...
  }
  ...
}
```

Then just use it as you would the normal parity.js, i.e.

```
const ParityApi = require('@parity/api");

// Only difference here to the previous version:
// Transport changed to Provider to align with web3/metamask/next-gen
// (Could also push in the web3 provider once they support the new .send() interface shortly
// PS: Could still use Transport as per normal, it will just show a warning message
const parity = new ParityApi(new ParityApi.Provider.Http('...'));
```

Issue #6372, opened 2017-08-24
Parity warp sync is no longer very warpy.


MicahZoltu:
> I'm running:
>
> - **Parity version**: 1.7.0
> - **Operating system**: Windows
> - **And installed**: via installer

---

I just installed Parity onto a brand new computer with a fresh install of Windows 10 Pro.  The computer has Hyper-V enabled and Docker, but otherwise is a stock Windows 10 Pro computer of reasonable power (Dell XPS 9560).  It is connected via wireless ab/n 5GHz to a nearby router and is able to easily max out my internet connection (100Mb/s down, 12 Mb/s up).

It launched after install about 12 hours ago and at first warp sync was moving quickly.  However, at about 91.46% warp restore (pretty shortly into the initial sync) the restore progress froze at 91.46% and best block started counting up from 0.  About 12 hours later it is only up to best block 3,594,000.

On previous versions of parity launching Parity with no chain data and no extra parameters (other than the default `ui`) would result in a full restore in about an hour on a less powerful computer.

Over the course of this time, it has written 2 TB to disk (presumably mostly overwrites since I only have a 1TB disk) and it has read almost nothing off of disk (42MB).  It has received ~7GB over the network and only sent about 300MB.

It seems there are several issues floating around about Parity's footprint (one of them even by me) and I apologize if this should have been a comment on one of those, but none of them expressed the same symptoms so I wasn't sure.

Some cumulative numbers across the lifetime of the process:
![image](https://user-images.githubusercontent.com/886059/29677620-81b0d59e-88b0-11e7-9e0e-8047afed19e1.png)
![image](https://user-images.githubusercontent.com/886059/29677636-8f5e90fa-88b0-11e7-8629-15647a5a4e4c.png)

MicahZoltu:
Upon further inspection, it appears that the initial launch of Parity post-install is `ui --warp --mode=passive` (this differs from the start menu shortcut that is created which is just `ui`).

MicahZoltu:
As an update, I just restarted my computer (for unrelated reasons) and Parity crashed on startup the first time, then on re-launch it started the Warp Restore over again from 0% and my best block says 3,831,654 (where it was at time of restart).

Perhaps this should be a separate issue (let me know if so) but it would seem that restarting in the middle of a warp restore causes problems.

5chdn:
Please share some logs. It's hard to tell why the warp sync is stuck at a certain state / block.

MicahZoltu:
[...]

PR #6560, opened 2017-09-20
Sync progress and error handling fixes

arkpar:


sync/src/chain.rs
```diff
@@ -1153,7 +1153,7 @@ impl ChainSync {
 		trace!(target: "sync", "== Connected {}: {}", peer, io.peer_info(peer));
 		if let Err(e) = self.send_status(io, peer) {
 			debug!(target:"sync", "Error sending status request: {:?}", e);
-			io.disable_peer(peer);
+			io.disconnect_peer(peer);
```
arkpar: Send errors are caused by dropped connections, not by misbehaviour.

5chdn:
Does this partially address #6372 ?

arkpar:
Yes, but only partially. This will prevent occasional dropping of peers with snapshots.

PR #6574, opened 2017-09-22
Check for nondeterministic operations in WASM contract

fckt:
WasmInterpreter::exec will return an vm::Error error if there is any nondeterministic operations found in to be executed WASM module

### DIFF
--- a/ethcore/wasm/src/lib.rs
+++ b/ethcore/wasm/src/lib.rs
@@ -113,13 +113,15 @@ impl vm::Vm for WasmInterpreter {
 
 		let mut cursor = ::std::io::Cursor::new(&*code);
 
-		let contract_module = wasm_utils::inject_gas_counter(
-			elements::Module::deserialize(
-				&mut cursor
-			).map_err(|err| {
-				vm::Error::Wasm(format!("Error deserializing contract code ({:?})", err))
-			})?
-		);
+		let raw_module = elements::Module::deserialize(&mut cursor).map_err(|err| {
+			vm::Error::Wasm(format!("Error deserializing contract code ({:?})", err))
+		})?;
+
+		if !wasm_utils::is_deterministic(&raw_module) {
+			return Err(vm::Error::Wasm(format!("Contract code contains nondeterministic operations")));
+		}
+
+		let contract_module = wasm_utils::inject_gas_counter(raw_module);
 
 		let d_ptr = runtime.write_descriptor(
 			call_args::CallArgs::new(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3476,7 +3476,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "wasm-utils"
 version = "0.1.0"
-source = "git+https://github.com/paritytech/wasm-utils#c69d8327c805c3f904feb3da425eafbd73e68d9f"
+source = "git+https://github.com/paritytech/wasm-utils#a6b6d75be0680568209813924a5ec0ad89e86697"
 dependencies = [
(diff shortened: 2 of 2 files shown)


ethcore/wasm/src/lib.rs
```diff
@@ -113,13 +113,15 @@ impl vm::Vm for WasmInterpreter {
...
-				&mut cursor
-			).map_err(|err| {
-				vm::Error::Wasm(format!("Error deserializing contract code ({:?})", err))
-			})?
-		);
+		let raw_module = elements::Module::deserialize(&mut cursor).map_err(|err| {
+			vm::Error::Wasm(format!("Error deserializing contract code ({:?})", err))
+		})?;
+
+		if !wasm_utils::is_deterministic(&raw_module) {
```
rphmeier: does this check take long? is it O(n) of the contract code? if so, shouldn't it cost gas?

why not perform the check once when the contract code is created?

NikVolf: 1. we don't have schedule for wasm yet
2. it will be optimized to check while general validation pass
3. you can't trust arbitrary blockchain code to be correct wasm code - for example, evm can create wasm code as well during the transaction

rphmeier: > you can't trust arbitrary blockchain code to be correct wasm code
sure, but if you have a protocol rule to execute all code beginning with a magic number as wasm (with nondeterminism disallowed), you can also check determinism of all create-code beginning with that magic number. since the cost of checking this is cheap, it can easily be factored in to the regular memory copying cost of creating contracts.

NikVolf: right, but then you will have to decode wasm module in this create-code and only then check it for having valid opcodes, which also takes some memory; while we decode and validate wasm before run anyway

arkpar: I'd agree that if we add new consensus rule we might as well require all wasm code (identified by prefix) to be valid on deployment. 
> while we decode and validate wasm (for regular rules) before run anyway

Could this be done on creation as well?

NikVolf: @arkpar 

No, validation is rather complex step required to run code anyway (it produces some internal state used during the execution)

@svyatonik might give some more details

svyatonik: Already had similar discussion with @fckt && @NikVolf :) So just 
1) validation became mandatory step, when we moved to plain opcodes structure - i.e. if previously we had FunctionBody = `IF([a == b], [c = c + 1], [c = c + 2])`, now we have moved to smth like `[a == b, if, begin, c = c + 1, else, c = c + 2, end]`. Interpreter currently uses validation to markup code (i.e. remember positions of matching begin+end -s). There's also some internal cache is created during validation.
2) Interpreter itself is still able to run without validation (i.e. it will return runtime errors, instead of validation-time errors), but this markup should be provided in some other way
3) wasm specification tells that before any module can be instantiated, if must be validated
4) I assume that functions validation can be performed on creation, but there are some steps, which should be performed when runtime environment is ready (it is about everything that is imported by the module - memory, functions, ...)

fckt: This validation is a specific requirement for incoming webassembly code (not module). IMO it shouldn't be considered as a part of standard module validation (which part of module initialization)

NikVolf: @svyatonik those runtime checks in `2.` should be dropped anyway to gain some performance

fckt: Indeed deserialization and code scan for every call could open a vulnerability for DDOS attacks cause call ops have a fixed gas price. What we might do is:
1) move code validation into deserialization
2) cache deserialized code in order to not deserialize it every call

Thanks @rphmeier for pointing out!

PR #6621, opened 2017-10-02
Add Musicoin and MCIP-3 UBI hardfork. 

5chdn:
What:

 - `--chain musicoin` logic for core and wallet
 - ethash logic for [MCIP-3](https://github.com/Musicoin/MCIPs/blob/master/MCIPS/mcip-3.md) transition scheduled for block 1_200_001 

Rationale:

 - eliminates the need for musicoin to distribute their own [fork of parity](https://github.com/immartian/rmc/)


5chdn:
Tested MCIP-3 transition against [GMC 2.0.0](https://github.com/Musicoin/go-musicoin/)

ethcore/src/ethereum/ethash.rs
```diff
@@ -186,24 +205,38 @@ impl Engine<EthereumMachine> for Arc<Ethash> {
...
 
 		let n_uncles = LiveBlock::uncles(&*block).len();
 
-		// Bestow block reward
+		// Bestow block rewards.
 		let result_block_reward = reward + reward.shr(5) * U256::from(n_uncles);
 		let mut uncle_rewards = Vec::with_capacity(n_uncles);
 
-		self.machine.add_balance(block, &author, &result_block_reward)?;
+		if number >= self.ethash_params.mcip3_transition {
```
tomusdrw: Is it possible to write tests for that?

ethcore/src/ethereum/ethash.rs
```diff
@@ -186,24 +205,38 @@ impl Engine<EthereumMachine> for Arc<Ethash> {
...
+			let miner_reward = self.ethash_params.mcip3_miner_reward.unwrap();
+			let ubi_contract = self.ethash_params.mcip3_ubi_contract;
+			let ubi_reward = self.ethash_params.mcip3_ubi_reward.unwrap();
+			let dev_contract = self.ethash_params.mcip3_dev_contract;
+			let dev_reward = self.ethash_params.mcip3_dev_reward.unwrap();
```
rphmeier: we don't use `unwrap()`, favor `expect()` with a proof that it will never panic. That these fields are either all set or all not should be checked at spec parsing time. Grouping them into a `struct mcip3` parsed optionally would make that logic a lot easier.

ethcore/src/ethereum/ethash.rs
5chdn: Yes, would be lucky to do that. Do I just add tests at the bottom of ethash.rs? Can I reuse the `test_spec()` or should I create a new test chain that includes the transition?

ethcore/src/ethereum/ethash.rs
5chdn: This makes sense. I have rewritten the logic, now a malformed mcip3 spec will just behave like a default block reward scheme.

ethcore/src/ethereum/ethash.rs
```diff
@@ -219,7 +252,7 @@ impl Engine<EthereumMachine> for Arc<Ethash> {
 			self.machine.add_balance(block, a, reward)?;
 		}
 
-		// note and trace.
+		// Note and trace.
 		self.machine.note_rewards(block, &[(author, result_block_reward)], &uncle_rewards)
```
rphmeier: without altering this line, block rewards from this change won't be traced. I would recommend using a 
`Cow<&'a [(Address, U256)]>` to store the main block reward initially to avoid an allocation on every block unless it's a musicoin chain.

rphmeier: or a separate call to `machine.note_rewards` within the musicoin if block would serve the same purpose.

rphmeier:
LGTM except for tracing nit

ethcore/src/ethereum/ethash.rs
5chdn: Should I call `machine.note_rewards` for _each_ block reward or only for the miner reward?

5chdn:
Reused `result_block_reward` variable to make sure the correct reward is traced. Should be all now.

ethcore/src/ethereum/ethash.rs
rphmeier: `note_rewards` takes two slices of `&[(Address, U256)]`. The first one is for block rewards and the second for uncle rewards. I don't think it matters if you make multiple calls to the function.

@grbIzl tracing does handle multiple block rewards correctly, right?

grbIzl: @rphmeier it should. It's pretty straightforward.

rphmeier: good. @5chdn then you would say the correct implementation traces only the miner reward and not any of the others? LGTM if so.

PR #6624, opened 2017-10-03
SecretStore: exclusive sessions

svyatonik:
ServersSetChange session can't be started if there are some sessions active.
Any session can't be started if ServersSetChange session is active.

Alternative implementation: wait() on all active sessions until requested session can be started. Haven't implemented this, because:
1) signing/decryption/generation/encryption sessions are started by HTTP request && client could possibly get timeout error before wait() is completed => lost hash=>key entry (in case of generation session) + a lot of duplicate decryption/signing sessions (if he's retries)
2) ServersSetChange can take a lot of time (when there are a lot of keys in SS) && it is better (imho) to fail asap
3) server we are requesting, can be excluded from SS when ServersSetChange session is completed => session will fail anyway


svyatonik:
It is TODO6 from #6605

PR #6630, opened 2017-10-04
SecretStore: ShareRemove of 'isolated' nodes

svyatonik:
It is a fix of TODO4 from #6605

The case is:
1) node is permanently removed because of failure (for example)
2) we decide to remove/add share to the key, this node holds share for
In current implementation every node that have a share of given key, must participate in any admin session with this key. Obviously, it is impossible in this case, because node(s) is already 'isolated' from cluster.

So the idea is to give ability to run admin sessions even if some owners of shares are already isolated:
1) ShareRemove is changed, so that it can be runned without communicating to isolated nodes - i.e. you can 'forget' isolated nodes.
2) it is still impossible to run ShareAdd && ShareMove with 'isolated' participants
3) ServesSetChange session is changed so that for every key 'isolated' nodes are removed in prior to adding/moving key shares

Since in (1) session is started explicitely by administrator && in (3) only isolated nodes which are not in new_nodes_set are removed, we may assume that isolated node is not going to return.
But this case (re-addition of isolated node to the cluster) should be taken into account when working on TODO#3 from #6605.


NikVolf:
need rebase @svyatonik

svyatonik:
Thanks!

PR #6631, opened 2017-10-04
SecretStore: session level timeout

svyatonik:
TODO5.2 from #6605


NikVolf:
need rebase @svyatonik

Issue #6056, opened 2017-07-14
WASM MVP


NikVolf:
Whats' left

- [x] calls (regular, static, call_code) #6132 
- [x] ~~storage gas metering~~ (managed by regular schedule)
- [x] memory allocations gas metering #6638 
- [x] static memory region gas charge #6638 
- [x] cost table for wasm opcodes #6638

NikVolf:
waiting for #6073 for proper schedule management

PR #6638, opened 2017-10-04
WASM gas schedule

NikVolf:
this actually closes the WASM MVP #6056, since proper gas metering is all that left there

WASM itself can be parameterized by much broader range of opcode types, but seems like they all have roughly the same cpu costs


ethcore/vm/src/schedule.rs
```diff
@@ -113,6 +113,49 @@ pub struct Schedule {
...
 	/// Enable EIP-86 rules
 	pub eip86: bool,
+	/// Wasm extra schedule settings
+	pub wasm: WasmCosts,
+}
+
+/// Wasm cost table
+pub struct WasmCosts {
+	/// Arena allocator cost, per byte
+	pub alloc: usize,
```
arkpar: Why `usize`? `usize` should only be used for representing memory area sizes. Hence the name. `u64` should prevent any unexpected issues on 32-bit platforms.

ethcore/wasm/src/runtime.rs
```diff
@@ -102,6 +103,13 @@ pub struct RuntimeContext {
 	pub value: U256,
 }
 
+macro_rules! charge {
```
arkpar: Why not a simple function?

NikVolf: because of double borrowing of self

ethcore/vm/src/schedule.rs
NikVolf: And why all regular `Schedule` members is usize?

arkpar: No idea, but I think it needs to be fixed.

ethcore/wasm/src/runtime.rs
arkpar: ugh, ok

ethcore/vm/src/schedule.rs
NikVolf: https://github.com/paritytech/parity/blob/c7ea25227a01aa9d22964f84489d7f16a224ce5c/ethcore/evm/src/evm.rs#L59

oh this is because of that, gas counter uses different size across platforms

i don't use it, so i might stick to u64

ethcore/wasm/src/runtime.rs
```diff
@@ -369,13 +412,17 @@ impl<'a, 'b> Runtime<'a, 'b> {
 		-> Result<Option<interpreter::RuntimeValue>, InterpreterError>
 	{
 		let amount = context.value_stack.pop_as::<i32>()? as u32;
+
+		charge!(self, self.ext.schedule().wasm.alloc * amount);
```
rphmeier: depending on the alloc cost, this seems like it could easily overflow a u32 when allocating just a few MB.

ethcore/wasm/src/runtime.rs
rphmeier: seems like a `fn charge<F>(&mut self, f: F) where F: FnOnce(&Schedule) -> u64` would work in all cases here

ethcore/wasm/src/runtime.rs
```diff
@@ -211,6 +249,9 @@ impl<'a, 'b> Runtime<'a, 'b> {
 
 		let code = self.memory.get(code_ptr, code_len as usize)?;
 
+		charge!(self, self.ext.schedule().create_gas as u32);
+		charge!(self, (self.ext.schedule().create_data_gas * code.len()) as u32);
```
rphmeier: again, using u32 for gas here seems a little flaky. In the EVM code we use u64 in general.

ethcore/wasm/src/runtime.rs
NikVolf: it's super easy operation with arena allocator, so it should not be priced more than memory opcodes
and memory limit for wasm is just 16mb

i will change to u64 anyway though to be uniform

NikVolf:
@rphmeier @arkpar 
updated!

PR #6639, opened 2017-10-04
Balance queries implemented in WASM runtime

fckt:


ethcore/wasm/src/runtime.rs
```diff
@@ -163,6 +163,19 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
+	/// Fetches balance for address
+	pub fn balance(&mut self, context: InterpreterCallerContext)
+		-> Result<Option<interpreter::RuntimeValue>, InterpreterError>
+	{
+		let mut context = context;
+		let return_ptr = context.value_stack.pop_as::<i32>()? as u32;
+		let address = self.pop_address(&mut context)?;
+		let balance = self.ext.balance(&address).map_err(|_| UserTrap::BalanceQueryError)?;
+		let value: H256 = balance.into();
+		self.memory.set(return_ptr, &*value)?;
```
rphmeier: does there need to be a check that the return pointer is actually valid?

NikVolf: how do you imagine such check?

rphmeier: i'm not sure, just not familiar with the intricacies of WASM. to me it looks like it just writes into whatever pointer it's given. obviously if the pointer is invalid then it should lead to an error in the WASM runtime, but it shouldn't cause a panic in `self.memory.set`. is it sure that it doesn't?

svyatonik: Yes - it doesn't panic - that's why `memory.set` returns Result<>

PR #6641, opened 2017-10-04
Light Client: fetch transactions/receipts by transaction hash

rphmeier:
Fetching a transaction (for RPC) by hash is different than all other requests because it requires information about the block to be included in the RPC result. The same goes for receipts. This means that we have to request more data after receiving the response before we can check its validity.

Implemented now with a `TransactionIndex` request. This returns a `(block number, block hash, index)` tuple. The block is fetched, checked for existence in the local chain, and then searched for the transaction. If we are fetching a receipt, the block receipts are then fetched and the receipt is extracted.

Using the trie roots in the header might be a better approach, but for now repeat requests will mostly draw from cached data and will not have to go to the network for anything heavy.

Also makes light protocol request cost table parsing more lenient, allowing some request types to be  present while others might not be. This allows for easy backwards-compatible extension of the light protocol with new request types.


svyatonik:
@rphmeier There are some compilation errors - have you seen it?

rphmeier:
Build failing due to test log being too long. @paritytech/ci any workaround?

rphmeier:
And now reaching OOM while building.

PR #6646, opened 2017-10-05
move additional_params to machine, fixes registry on non-ethash chains

rphmeier:
most implementations of this function were lost in the recent refactor.

### DIFF
--- a/ethcore/src/engines/mod.rs
+++ b/ethcore/src/engines/mod.rs
@@ -191,9 +191,6 @@ pub trait Engine<M: Machine>: Sync + Send {
 	/// Additional engine-specific information for the user/developer concerning `header`.
 	fn extra_info(&self, _header: &M::Header) -> BTreeMap<String, String> { BTreeMap::new() }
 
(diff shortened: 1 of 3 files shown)


rphmeier:
I think this closes https://github.com/paritytech/parity/issues/6636

ethcore/src/engines/mod.rs
```diff
@@ -396,6 +393,11 @@ pub trait EthEngine: Engine<::machine::EthereumMachine> {
 	fn supports_wasm(&self) -> bool {
 		self.machine().supports_wasm()
 	}
+
+	/// Additional information.
+	fn additional_params(&self) -> HashMap<String, String> {
```
tomusdrw: Could probably be `BTreeMap` as well.

rphmeier: agreed, but it's beyond the scope of this PR.
TBH, it's only used for the registry, so I don't see that much of a reason to make it a stringly-typed map in the first place.

PR #6651, opened 2017-10-05
remove RPC parameter leniency now that mist formats correctly

rphmeier:
Closes https://github.com/paritytech/parity/issues/6648

In particular:
  - accepting decimal-formatted block numbers
  - parsing the empty string as empty bytes.


rphmeier:
(worth backporting?)

debris:
It's better not to backport it. I expect some applications to depend on parity misbehaviour. It's better to wait till `1.9` with it

PR #6664, opened 2017-10-06
Fixed build warnings

arkpar:


dapps/js-glue/src/codegen.rs
```diff
@@ -102,7 +102,7 @@ fn implement_webapp(cx: &ExtCtxt, builder: &aster::AstBuilder, item: &Item, push
 	let files_impl = quote_item!(cx,
 		impl $type_name {
 			fn files() -> ::std::collections::HashMap<&'static str, File> {
-				let mut files = ::std::collections::HashMap::new();
```
tomusdrw: That won't work if `$statements` are not empty. Actually it shouldn't be printed while compiling Parity at all (since we always have some files there).

dapps/js-glue/src/codegen.rs
```diff
@@ -101,6 +101,7 @@ fn implement_webapp(cx: &ExtCtxt, builder: &aster::AstBuilder, item: &Item, push
 
 	let files_impl = quote_item!(cx,
 		impl $type_name {
+			#[warn(unused_mut)]
```
rphmeier: ishould it be `#[allow(unused_mut)]`?

arkpar: Weirdly enough it did not produce any warnings when building like this.

gavofyork:
doesn't build

5chdn:
CI:

> fatal runtime error: allocator memory exhausted

@paritytech/ci ^

PR #6598, opened 2017-09-26
Updated systemd files for linux (Resolves #6592)

AntoniosHadji:
Previous version put $BASE directory in /root directory and allowed unclean exit / abort.
This version clearly explains how to run as root or as specific user and exits cleanly.

* use config.toml for configuration for single configuration format.
* send SIGHUP for clean exit.
* restart on fail.

Tested on Ubuntu 16.04.3 LTS with 4.10.0-33-generic x86_64 kernel

PR #6597, opened 2017-09-26
Fix wallet view

ngotchac:
Closes #6672

The Multisig Wallet View wouldn't load if multiple transactions happened within one block. This PR fixes it.

PR #6675, opened 2017-10-09
[beta] Backporting

arkpar:


5chdn:
for UI:

- #6597 Fix wallet view (Once approved)
- #6602 Disallow pasting recovery phrases on first run

for the release maybe somewhat worth to have in 1.7.3:

- #6598 Updated systemd files for linux
- #6633 Don't expose port 80 for parity anymore

PR #6686, opened 2017-10-10
Fixed potential modexp exp len overflow

arkpar:


rphmeier:
wouldn't the right approach to simply use `saturating_mul` when calculating the cost using the `adjusted_exp_len`?

ethcore/src/builtin.rs
```diff
@@ -107,12 +107,15 @@ impl Pricer for ModexpPricer {
 		}
 
 		let base_len = base_len.low_u64();
-		let exp_len = exp_len.low_u64();
 		let mod_len = mod_len.low_u64();
 		let m = max(mod_len, base_len);
 		if m == 0 {
 			return U256::zero();
 		}
+		if exp_len > max_len {
```
tomusdrw: The conditions looks a bit messy now. What about this:
```rust
// max(mod_len, base_len) == 0
if mod_len.is_zero() && base_len.is_zero() {
  return U256::zero()
}

let max_len = U256::from(u32::max_value() / 2);
if base_len > max_len || mod_len > max_len || exp_len > max_len {
  return U256::max_value();
}

let (base_len, exp_len, mod_len) = (base_len.low_u64(), exp_len.low_u64(), mod_len.low_u64());
...
```

ethcore/src/builtin.rs
```diff
@@ -133,8 +132,7 @@ impl ModexpPricer {
 		let bit_index = if exp_low.is_zero() { 0 } else { (255 - exp_low.leading_zeros()) as u64 };
 		if len <= 32 {
 			bit_index
-		}
-		else {
+		} else {
```
tomusdrw: :+1: Was hurting my eyes a bit :)

rphmeier:
> u32::max_value() upper bound is used for other two params

is that a hard bound specified by the EIP? otherwise, why introduce the arbitrary (and technically wrong) restriction, when really gas should be the only limiting factor? Doing computations on numbers a few gigabytes long is within reach of modern computers even if impractical

arkpar:
This PR does not introduce any additional restrictions. Restriction for 32-bit addressable memory is already in the evm implementation. There are quite a few places where we choose more efficient gas calculation in plain integers. This is one of them. Changing that is out of scope of this PR

ethcore/src/builtin.rs
```diff
@@ -101,18 +101,17 @@ impl Pricer for ModexpPricer {
 		let exp_len = read_len();
```
Vurich: It won't let me add comments on the actual line, but on line 92 above here: Why do we zero-extend the input and `read_exact` instead of just using `read`? The buffer's already filled with zeroes.

ethcore/src/builtin.rs
```diff
@@ -101,18 +101,17 @@ impl Pricer for ModexpPricer {
...
+		let (base_len, exp_len, mod_len) = (base_len.low_u64(), exp_len.low_u64(), mod_len.low_u64());
 
-		let base_len = base_len.low_u64();
-		let exp_len = exp_len.low_u64();
-		let mod_len = mod_len.low_u64();
 		let m = max(mod_len, base_len);
-		if m == 0 {
-			return U256::zero();
-		}
 		// read fist 32-byte word of the exponent.
```
Vurich: Fist

ethcore/src/builtin.rs
```diff
@@ -101,18 +101,17 @@ impl Pricer for ModexpPricer {
 		let exp_len = read_len();
 		let mod_len = read_len();
 
+		if mod_len.is_zero() && base_len.is_zero() {
+			return U256::zero()
+		}
+
 		let max_len = U256::from(u32::max_value() / 2);
-		if base_len > max_len || mod_len > max_len {
+		if base_len > max_len || mod_len > max_len || exp_len > max_len {
```
gavofyork: does this work properly with base = 1, mod = 1, exp = 2^256-1?

gavofyork: @arkpar ^^^

arkpar: Yes, according to the spec it should fail because gas calculation formula would return a huge value.

PR #6689, opened 2017-10-10
Migrate to Futures in SigningQueue

### LINKED ISSUE #4490: Refactor signer to use futures instead of callbacks
Details: https://github.com/paritytech/parity/pull/4485#discussion_r100246661

kirushik:
Closes #4490

### DIFF
--- a/rpc/src/lib.rs
+++ b/rpc/src/lib.rs
@@ -23,6 +23,7 @@
 extern crate ansi_term;
 extern crate cid;
 extern crate crypto as rust_crypto;
+#[macro_use] extern crate futures;
 extern crate futures_cpupool;
 extern crate itertools;
 extern crate multihash;
--- a/rpc/src/lib.rs
+++ b/rpc/src/lib.rs
@@ -20,10 +20,11 @@
 #![cfg_attr(feature="dev", feature(plugin))]
 #![cfg_attr(feature="dev", plugin(clippy))]
 
+#[macro_use] extern crate futures;
+
 extern crate ansi_term;
 extern crate cid;
 extern crate crypto as rust_crypto;
-#[macro_use] extern crate futures;
 extern crate futures_cpupool;
 extern crate itertools;
 extern crate multihash;
--- a/rpc/src/lib.rs
+++ b/rpc/src/lib.rs
@@ -20,7 +20,8 @@
 #![cfg_attr(feature="dev", feature(plugin))]
 #![cfg_attr(feature="dev", plugin(clippy))]
 
-#[macro_use] extern crate futures;
+#[macro_use]
+extern crate futures;
 
 extern crate ansi_term;
 extern crate cid;
--- a/rpc/src/v1/helpers/signing_queue.rs
+++ b/rpc/src/v1/helpers/signing_queue.rs
@@ -23,7 +23,7 @@ use v1::helpers::{ConfirmationRequest, ConfirmationPayload, oneshot, errors};
 use v1::types::{ConfirmationResponse, H160 as RpcH160, Origin, DappId as RpcDappId};
 
 use jsonrpc_core::Error;
-use jsonrpc_core::futures::{Future, Poll};
+use jsonrpc_core::futures::Future;
 
 /// Result that can be returned from JSON RPC.
 pub type RpcResult = Result<ConfirmationResponse, Error>;
@@ -74,7 +74,7 @@ pub const QUEUE_LIMIT: usize = 50;
 pub trait SigningQueue: Send + Sync {
 	/// Add new request to the queue.
 	/// Returns a `Result` wrapping  `ConfirmationReceiver` which is a `Future` awaiting for resolution of the given request.
-	fn add_request(&self, request: ConfirmationPayload, origin: Origin) -> Result<ConfirmationReceiver, QueueAddError>;
+	fn add_request(&self, request: ConfirmationPayload, origin: Origin) -> Result<(U256, ConfirmationReceiver), QueueAddError>;
 
 	/// Removes a request from the queue.
 	/// Notifies possible token holders that request was rejected.
@@ -136,31 +136,7 @@ struct ConfirmationSender {
 
 /// Receiving end of the Confirmation channel; can be used as a `Future` to await for `ConfirmationRequest`
 /// being processed and turned into `ConfirmationOutcome`
-#[must_use = "futures do nothing unless polled"]
-pub struct ConfirmationReceiver {
-	id: U256,
-	receiver: oneshot::Receiver<ConfirmationOutcome>
-}
-
-impl ConfirmationReceiver {
-	fn new(id: U256, receiver: oneshot::Receiver<ConfirmationOutcome>) -> Self {
-		ConfirmationReceiver { id, receiver }
-	}
-
-	/// `U256` id of the request we're waiting to get confirmed.
-	pub fn id(&self) -> U256 {
-		self.id
-	}
-}
-
-impl Future for ConfirmationReceiver {
-	type Item = ConfirmationOutcome;
-	type Error = Error;
-
-	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-		self.receiver.poll()
-	}
-}
+pub type ConfirmationReceiver =  oneshot::Receiver<ConfirmationOutcome>;
 
 /// Queue for all unconfirmed requests.
 #[derive(Default)]
@@ -220,7 +196,7 @@ impl Drop for ConfirmationsQueue {
 }
 
 impl SigningQueue for ConfirmationsQueue {
-	fn add_request(&self, request: ConfirmationPayload, origin: Origin) -> Result<ConfirmationReceiver, QueueAddError> {
+	fn add_request(&self, request: ConfirmationPayload, origin: Origin) -> Result<(U256, ConfirmationReceiver), QueueAddError> {
 		if self.len() > QUEUE_LIMIT {
 			return Err(QueueAddError::LimitReached);
 		}
@@ -247,7 +223,7 @@ impl SigningQueue for ConfirmationsQueue {
 					origin,
 				},
 			});
-			ConfirmationReceiver::new(id, receiver)
+			(id, receiver)
 		};
 		// Notify listeners
 		self.notify(QueueEvent::NewRequest(id));
@@ -319,9 +295,7 @@ mod test {
 		let request = request();
 
 		// when
-		let future = queue.add_request(request, Default::default()).unwrap();
-
-		let id = U256::from(1);
+		let (id, future) = queue.add_request(request, Default::default()).unwrap();
 		queue.request_confirmed(id, Ok(ConfirmationResponse::SendTransaction(1.into())));
 
 		// then
--- a/rpc/src/v1/helpers/signing_queue.rs
+++ b/rpc/src/v1/helpers/signing_queue.rs
@@ -277,8 +277,6 @@ impl SigningQueue for ConfirmationsQueue {
 
 #[cfg(test)]
 mod test {
-	use std::time::Duration;
-	use std::thread;
 	use std::sync::Arc;
 	use bigint::prelude::U256;
 	use util::Address;
@@ -308,7 +306,7 @@ mod test {
         let mut confirmation = ConfirmationReceiver::new(U256::from(1), receiver);
         assert_eq!(confirmation.is_done(), false);
         sender.send(Ok(ConfirmationResult::Rejected));
-		confirmation.poll();
+		confirmation.poll().unwrap();
         assert_eq!(confirmation.is_done(), true);
 	}
 
@@ -341,7 +339,7 @@ mod test {
 		queue.on_event(move |notification| {
 			r.lock().push(notification);
 		});
-		queue.add_request(request, Default::default()).unwrap();
+		let _future = queue.add_request(request, Default::default()).unwrap();
 		queue.finish();
 
 		// then
@@ -358,7 +356,7 @@ mod test {
 		let request = request();
 
 		// when
-		queue.add_request(request.clone(), Default::default()).unwrap();
+		let _future = queue.add_request(request.clone(), Default::default()).unwrap();
 		let all = queue.requests();
 
 		// then
--- a/rpc/src/v1/helpers/signing_queue.rs
+++ b/rpc/src/v1/helpers/signing_queue.rs
@@ -285,8 +285,10 @@ mod test {
 	use util::Address;
 	use parking_lot::Mutex;
 	use jsonrpc_core::futures::Future;
-	use v1::helpers::{SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest,
-					  ConfirmationResult, ConfirmationPayload, ConfirmationReceiver, oneshot};
+	use v1::helpers::{
+		SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest,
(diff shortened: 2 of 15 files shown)


kirushik:
[clabot:check]

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -320,31 +301,30 @@ mod test {
 	}
 
 	#[test]
+	fn should_mark_future_as_done() {
+		let (sender, receiver) = oneshot::oneshot();
```
NikVolf: we use tabs, not spaces see `.editorconfig`

NikVolf:
a lot of `ConfirmationPromise`-es left in the doc comments i believe

rpc/src/v1/helpers/signing_queue.rs
kirushik: Side-effect of jumping from editor to editor, I guess; shouldn't happen anymore in the future.

I'll double-check spacing in all the files I've touched in this PR, thanks for noticing!

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -75,7 +74,7 @@ pub const QUEUE_LIMIT: usize = 50;
 pub trait SigningQueue: Send + Sync {
 	/// Add new request to the queue.
 	/// Returns a `ConfirmationPromise` that can be used to await for resolution of given request.
```
NikVolf: doc comment needs to be updated

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -101,85 +100,63 @@ pub trait SigningQueue: Send + Sync {
...
-	result: Arc<Mutex<ConfirmationResult>>,
-	listeners: Arc<Mutex<Vec<Listener>>>,
+pub struct ConfirmationSender {
+	sender: oneshot::Sender<ConfirmationResult>,
 	request: ConfirmationRequest,
 }
 
-pub struct ConfirmationPromise {
+#[must_use = "futures do nothing unless polled"]
+pub struct ConfirmationReceiver {
```
NikVolf: No doc comment for public structure

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -101,85 +100,63 @@ pub trait SigningQueue: Send + Sync {
...
-		}
-		// Notify listener
-		let listeners = {
-			let mut listeners = self.listeners.lock();
-			mem::replace(&mut *listeners, Vec::new())
-		};
-		for mut listener in listeners {
-			listener(result.clone());
+impl ConfirmationReceiver {
+	pub fn new(id: U256, receiver: oneshot::Receiver<ConfirmationResult>) -> Self {
```
NikVolf: No doc comment for public method

NikVolf:
Looks great, just some style issues

kirushik:
@NikVolf Thanks for the review!
Will fix documentation after the (long-due) lunch break.

rpc/src/lib.rs
```diff
@@ -23,6 +23,7 @@
 extern crate ansi_term;
 extern crate cid;
 extern crate crypto as rust_crypto;
+#[macro_use] extern crate futures;
```
rphmeier: could you put this import in a separate section so it reads cleaner?

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -101,85 +100,63 @@ pub trait SigningQueue: Send + Sync {
...
 		}
 	}
 
-	fn as_promise(&self) -> ConfirmationPromise {
-		ConfirmationPromise {
-			id: self.request.id,
-			result: self.result.clone(),
-			listeners: self.listeners.clone(),
-		}
+	pub fn id(&self) -> U256 {
```
rphmeier: needs docs here too

rpc/src/v1/impls/signing.rs
```diff
@@ -50,45 +49,40 @@ const MAX_PENDING_DURATION_SEC: u32 = 60;
...
+	Future(ConfirmationReceiver),
 	Value(RpcConfirmationResponse),
 }
 
+impl Future for DispatchResult {
+	type Item = RpcConfirmationResponse;
+	type Error = Error;
+
+	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		match self {
```
rphmeier: more idiomatic to `match *self` here

rpc/src/v1/impls/signing.rs
```diff
@@ -180,11 +174,16 @@ impl<D: Dispatcher + 'static> ParitySigning for SigningQueueClient<D> {
 
 	fn check_request(&self, id: RpcU256) -> Result<Option<RpcConfirmationResponse>, Error> {
 		let id: U256 = id.into();
-		match self.pending.lock().get(&id) {
-			Some(ref promise) => match promise.result() {
-				ConfirmationResult::Waiting => Ok(None),
-				ConfirmationResult::Rejected => Err(errors::request_rejected()),
-				ConfirmationResult::Confirmed(rpc_response) => rpc_response.map(Some),
+		match self.pending.lock().get_mut(&id) {
+			Some(ref mut future) => match future.poll() {
```
rphmeier: fairly certain polling will panic if not done within the context of a task

rphmeier: i suppose _technically_ it is part of a task because this function is future-ized by the RPC wrapper. but that's a sketchy invariant that isn't necessarily guaranteed to hold in the future.

rphmeier: @tomusdrw wouldn't proper future-ization of signing RPCs not require `check_request` or other polling at all?

rpc/src/v1/impls/signing.rs
kirushik: Oh, indeed, thanks!

rpc/src/v1/impls/signing.rs
kirushik: @rphmeier as a second thought, I can cache the result of the polling as an `Option<ConfirmationResult>` and provide a method to look there — it will replace this polling here completely, and move the complexity from `check_request` into `ConfirmationReceiver`.

Seems to be a good idea, let my try that instead.

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -296,13 +280,13 @@ impl SigningQueue for ConfirmationsQueue {
...
-	use std::thread;
-	use std::sync::{mpsc, Arc};
+	use std::sync::Arc;
 	use bigint::prelude::U256;
 	use util::Address;
 	use parking_lot::Mutex;
-	use v1::helpers::{SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest, ConfirmationPayload};
+	use jsonrpc_core::futures::Future;
+	use v1::helpers::{SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest,
+					  ConfirmationResult, ConfirmationPayload, ConfirmationReceiver, oneshot};
```
NikVolf: should be
```rust
use v1::helpers::{SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest, ConfirmationResult,
    ConfirmationPayload, ConfirmationReceiver, oneshot};
```

or 

```rust
use v1::helpers::{
    SigningQueue, ConfirmationsQueue, QueueEvent, FilledTransactionRequest, ConfirmationResult,
    ConfirmationPayload, ConfirmationReceiver, oneshot
};
```
we use only single extra tab for the indentation

5chdn:
Just adding an `onice` label to make sure this does not get merged into 1.8 before Monday. This does not mean you shoudl stop working on/reviewing this. ;-)

debris:
@kirushik you are still working on this, aren't you? :)

kirushik:
@NikVolf @rphmeier PR has changed significantly (to address https://github.com/paritytech/parity/pull/6689#discussion_r143759446 in the most future-proof way), so please re-review.

rphmeier:
Taking a look. In the future, you should also set the status label to `pleasereview`.

rpc/src/lib.rs
```diff
@@ -20,6 +20,8 @@
 #![cfg_attr(feature="dev", feature(plugin))]
 #![cfg_attr(feature="dev", plugin(clippy))]
 
+#[macro_use] extern crate futures;
```
tomusdrw: Please split it into two lines.

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -100,86 +99,77 @@ pub trait SigningQueue: Send + Sync {
...
-		let listeners = {
-			let mut listeners = self.listeners.lock();
-			mem::replace(&mut *listeners, Vec::new())
-		};
-		for mut listener in listeners {
-			listener(result.clone());
+impl ConfirmationReceiver {
+	fn new(id: U256, receiver: oneshot::Receiver<ConfirmationOutcome>) -> Self {
+		ConfirmationReceiver {
+			id: id,
```
tomusdrw: `ConfirmationReceiver { id, receiver }`

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -247,24 +240,25 @@ impl SigningQueue for ConfirmationsQueue {
 			trace!(target: "own_tx", "Signer: ({:?}) : {:?}", id, request);
 
 			let mut queue = self.queue.write();
-			queue.insert(id, ConfirmationToken {
-				result: Arc::new(Mutex::new(ConfirmationResult::Waiting)),
-				listeners: Default::default(),
+			let (sender, receiver) = oneshot::oneshot::<ConfirmationOutcome>();
+
+			queue.insert(id, ConfirmationSender {
+				sender: sender,
```
tomusdrw: `sender,` instead of `sender: sender,`

rpc/src/v1/impls/signing.rs
```diff
@@ -126,12 +142,14 @@ impl<D: Dispatcher + 'static> SigningQueueClient<D> {
...
 					Either::B(future::done(
 						signer.add_request(payload, origin)
-							.map(DispatchResult::Promise)
+							.map(DispatchResult::Future)
 							.map_err(|_| errors::request_rejected_limit())
 					))
 				}
 			}))
 	}
+
```
tomusdrw: Reduntant lines

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -100,86 +99,77 @@ pub trait SigningQueue: Send + Sync {
 
 #[derive(Debug, Clone, PartialEq)]
 /// Result of a pending confirmation request.
-pub enum ConfirmationResult {
-	/// The request has not yet been confirmed nor rejected.
-	Waiting,
+// TODO [kirushik] Let's get rid of this completely, and simplify that `Result<Option<Result...` juggling somehow
```
tomusdrw: Let's do it in this PR then, a typedef might help.

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -100,86 +99,77 @@ pub trait SigningQueue: Send + Sync {
...
+struct ConfirmationSender {
+	sender: oneshot::Sender<ConfirmationOutcome>,
 	request: ConfirmationRequest,
 }
 
-pub struct ConfirmationPromise {
+/// Receiving end of the Confirmation channel; can be used as a `Future` to await for `ConfirmationRequest`
+/// being processed and turned into `ConfirmationOutcome`
+#[must_use = "futures do nothing unless polled"]
+pub struct ConfirmationReceiver {
```
tomusdrw: Do we need that structure at all? Maybe just return `(U256, Receiver)` instead of wrapping it.

rpc/src/v1/impls/signing.rs
```diff
@@ -45,61 +46,76 @@ use v1::types::{
...
-		let non_waiting: Vec<_> = map
-			.iter()
-			.filter(|&(_, val)| val.result() != ConfirmationResult::Waiting)
-			.map(|(key, _)| *key)
-			.collect();
-		for k in non_waiting {
-			map.remove(&k);
+	let future = future.then(move |result| {
+		{
+			let mut pending = pending.write();
```
tomusdrw: Seems that this introduces a race condition when the feature is no longer pending but the result is not in confirmations.
Why do we need  two structures at all? Why not `HashMap<U256, Status>`, where `enum Status { Pending, Done(Result..) }`?

kirushik: @tomusdrw Well, my idea was having a LRU-cache (`TransientHashMap`) of successfull confirmations, and a usual Set of UIDs of pending ones. This way we won't have to override the TransientHashMap behavior to collect obsolete confirmations.

My original line of thinking was that get-and-set from a HashMap is not atomic, so the worst case is the same as with two structures; but I think I ignored the fact that we don't need to actualy _read_ from the HashMap to update it's value (we have all the data in the callback, so just overwriting the value will do).

But considering all that, I have a question — the old code had that garbage collection routine (which was actually just touching all the items on the HashMap in the iteration order, sorta invalidationg the point of LRU cache.)

How should I better address it when I put both Pending and Confirmed transactions in the LRU?
I guess I would need a more specialized structure then, which would evict only Confirmed transactions?
Othewise we might end up kicking out some Pending transactions from the queue, and that will make RPC seem unreliable.

rpc/src/v1/helpers/signing_queue.rs
kirushik: I'll try and see if it will lead to the more elegant code — I'm using `id()` method here and there, and I would need to replace it with explicit id passing.

rpc/src/v1/helpers/signing_queue.rs
kirushik: Fair point

rpc/src/v1/impls/signing.rs
tomusdrw: > My original line of thinking was that get-and-set from a HashMap is not atomic,
It's under a `Mutex` anyway.

> the old code had that garbage collection routine
So the idea with `TransientHashMap` is that elements are evicted after some timeout if they are not actively queried for. It's not LRU and I don't think we actually need it here.
In case some pending transactions are kicked out it means that no-one really cares about them any more (since he is not asking for the status using `checkRequest`).

rpc/src/v1/impls/signing.rs
```diff
@@ -45,61 +45,69 @@ use v1::types::{
...
+	confirmations: Arc<Mutex<TransientHashMap<U256, Option<RpcConfirmationResult>>>>,
+	id: U256,
+	future: RpcConfirmationReceiver) {
+	{
+		let mut confirmations = confirmations.lock();
+		confirmations.insert(id.clone(), None);
 	}
+
+	let future = future.then(move |result| {
+		{
```
tomusdrw: No need to have additional braces.

rpc/src/v1/impls/signing.rs
```diff
@@ -45,61 +45,69 @@ use v1::types::{
...
-			.map(|(key, _)| *key)
-			.collect();
-		for k in non_waiting {
-			map.remove(&k);
-		}
+fn schedule(remote: Remote,
+	confirmations: Arc<Mutex<TransientHashMap<U256, Option<RpcConfirmationResult>>>>,
+	id: U256,
+	future: RpcConfirmationReceiver) {
+	{
```
tomusdrw: I think the braces are not strictly required.

PR #6707, opened 2017-10-11
Allow signer signing display of markdown

jacogr:
parity 2017-10-11 13-46-25


gavofyork:
@5chdn @jacogr will need to forward-port to 1.8, too - please track.

5chdn:
this is in 1.8 already, we rather need to back-port it.

PR #6708, opened 2017-10-11
Emulate signer pubsub on public node

### LINKED ISSUE #6667: Signer on public node (wallet.parity.io) broken

maciejhirsz:
Fixes #6667


jacogr:
Looks as good as it can be for a stop-gap.

PR #6710, opened 2017-10-11
Display vouched overlay on dapps

jacogr:
Add overlay of accounts vouching for a specific dapp, along with count of vouchers.

parity 2017-10-11 17-16-45

TODO: Can be optimised - a lot.


js/src/ui/DappVouchFor/store.js
```diff
@@ -0,0 +1,69 @@
...
+import { vouchfor as vouchForAbi } from '~/contracts/abi';
+
+export default class Store {
+  @observable vouchers = [];
+
+  constructor (api, app) {
+    this._api = api;
+    this._app = app;
+
+    Contracts
```
tomusdrw: That won't really work in case of re-connection or dynamic change, but I supposed we don't need that for initial PoC.

js/src/ui/DappVouchFor/store.js
```diff
@@ -0,0 +1,69 @@
...
+      .get().registry
+      .lookupAddress('vouchfor')
+      .then((address) => {
```
tomusdrw: Why not write everything as `async` function and simplify constructor to only do:

```
constructor() {
  this.initialize().catch(...);
} 
async initialize() {
 ...
}

File: js/src/ui/DappVouchFor/store.js [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+      })
+      .catch((error) => {
+        console.error('vouchFor', error);
+
+        return null;
+      });
+  }
+
+  @action addVoucher = (voucher) => {
+    this.vouchers = uniq([].concat(this.vouchers.peek(), [voucher]));
```
tomusdrw: `this.vouchers.peek()`?

js/src/ui/DappVouchFor/dappVouchFor.js
```diff
@@ -0,0 +1,57 @@
...
+export default class DappVouchFor extends Component {
+  static contextTypes = {
+    api: PropTypes.object.isRequired
+  };
+
+  static propTypes = {
+    app: PropTypes.object.isRequired
+  };
+
+  store = new Store(this.context.api, this.props.app);
```
tomusdrw: TBH I'm pretty sure it will lead to a broken behaviour.

The component instance can be re-used by react and just receive new props, should re-initialize `Store` on `componentWillReceiveProps` if we want to do it right.

(side note: it will also do some excessive calls for each dapp, might be worth to only limit it to network dapps?)

js/src/ui/DappVouchFor/store.js
jacogr: MobX idisyncracy - arrays are actually objects, .peek() returns the actual observed array. If no peek, basically it concats an object.

js/src/ui/DappVouchFor/dappVouchFor.js
jacogr: Good point. It is not optimal usage-wise atm, that will make it better for at least the intial version.

js/src/ui/DappVouchFor/store.js
jacogr: Tried it, eslint didn't like that approach at all. (But probably due to the fact that I mixed promises/async as I still do here instead of fully async function). 

Good point, added to the list of cleanups.

js/src/ui/DappVouchFor/dappVouchFor.js
jacogr: Edit: added only for apps where we have an actual contentHash

js/src/ui/DappVouchFor/store.js
jacogr: 100%, added to list of next PR cleanups.

Issue #6485, opened 2017-09-08
Graceful exit when invalid CLI flags are passed


ngotchac:
Parity should panic and exit with an non-0 exit code when invalid CLI flags are passed.

rphmeier:
panicking seems a little drastic. a helpful message and exit should be fine

5chdn:
I'm not sure what this ticket is about.

    0 ✓ user@alcor ~ $ parity --please-exit-gracefully
    Unknown flag: '--please-exit-gracefully'

    Usage:
      parity [options]
      parity ui [options]
      parity dapp <path> [options]
      parity daemon <pid-file> [options]
      parity account (new | list ) [options]
      parity account import <path>... [options]
      parity wallet import <path> --password FILE [options]
      parity import [ <file> ] [options]
      parity export (blocks | state) [ <file> ] [options]
      parity signer new-token [options]
      parity signer list [options]
      parity signer sign [ <id> ] [ --password FILE ] [options]
      parity signer reject <id> [options]
      parity snapshot <file> [options]
      parity restore [ <file> ] [options]
      parity tools hash <file>
      parity db kill [options]
    1 ✗ user@alcor ~ $ parity --tracing noway
    Loading config file from /home/user/.local/share/io.parity.ethereum/config.toml
    Invalid switch value: noway
    1 ✗ user@alcor ~ $ 

It always exits with a non-`0` exit code (`1` in both cases above, note the `1 ✗` in my `PS1`). This is 1.7.2.

ngotchac:
It used to work this way, but doesn't actually on `master`.

5chdn:
True.

axelchalon:
So, this is caused by
```rust
.global_setting(AppSettings::AllowLeadingHyphen) // allow for example --allow-ips -10.0.0.0/8
```
https://github.com/paritytech/parity/blob/master/parity/cli/usage.rs#L540

PR #6711, opened 2017-10-11
Graceful exit when invalid CLI flags are passed (#6485)

axelchalon:
Fixes https://github.com/paritytech/parity/issues/6485 (due to a bug in Clap; will file issue)


axelchalon:
doesn't crash on invalid values (we just saw that it has been this way since the port to clap); I will look into this and do another PR

axelchalon:
I just fixed parity not throwing an error on invalid values; I'll make a PR

axelchalon:
https://github.com/paritytech/parity/pull/6723

PR #6719, opened 2017-10-12
Check vouch status on appId in addition to contentHash

jacogr:
Basically allow lookups on both the contentHash and the appId for vouching. (De-dupe results). Just using the contentHash leads to the situation where the count gets lost when the app is actually updated.

parity 2017-10-12 15-24-58

Screenshot shows dapp with verification from both contentHash and appId.

### DIFF
--- a/js/src/ui/DappVouchFor/store.js
+++ b/js/src/ui/DappVouchFor/store.js
@@ -28,9 +28,7 @@ export default class Store {
   constructor (api, app) {
     this._api = api;
 
-    const { contentHash, id } = app;
-
-    this.findVouchers(contentHash, id);
+    this.findVouchers(app);
   }
 
   async attachContract () {
@@ -45,7 +43,7 @@ export default class Store {
     return contract;
   }
 
-  async findVouchers (contentHash, id) {
+  async findVouchers ({ contentHash, id }) {
     if (!contentHash) {
       return;
     }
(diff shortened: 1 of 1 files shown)


js/src/ui/DappVouchFor/store.js
```diff
@@ -20,55 +20,70 @@ import { uniq } from 'lodash';
...
     this._api = api;
 
-    const { contentHash } = app;
+    const { contentHash, id } = app;
+
+    this.findVouchers(contentHash, id);
+  }
+
+  async attachContract () {
+    const address = await Contracts.get().registry.lookupAddress('vouchfor');
```
jacogr: Don't lookup registry per store instance and only create a single contract that is used for queries. (next couple of lines after this). Optimisation as discussed in the previous PR.

js/src/ui/DappVouchFor/store.js
```diff
@@ -20,55 +20,70 @@ import { uniq } from 'lodash';
...
+    if (!address || /^0x0*$/.test(address)) {
+      return null;
+    }
+
+    const contract = await this._api.newContract(vouchForAbi, address);
+
+    return contract;
+  }
+
+  async findVouchers (contentHash, id) {
```
jacogr: Made everything async/await to simplify as per comments from previous PR.

PR #6723, opened 2017-10-12
CLI: Reject invalid argument values rather than ignore them

axelchalon:
Previously  (since the CLI port to clap), passing invalid values (e.g. providing a string value to an argument that expects a u32) was failing silently and was processed just like if the argument hadn't been entered.

Now, Parity explicitly refuses to launch if Clap cannot parse the value into the expected type.

```
$ parity --cache=asd                            
error: Invalid value: The argument 'asd' isn't a valid value
```

### DIFF
--- a/parity/cli/usage.rs
+++ b/parity/cli/usage.rs
@@ -38,7 +38,10 @@ macro_rules! return_if_parse_error {
 			Err(clap_error @ ClapError { kind: ClapErrorKind::ValueValidation, .. }) => {
 				return Err(clap_error);
 			},
-			_ => $e
+
+			// Otherwise, if $e is ClapErrorKind::ArgumentNotFound or Ok(),
+			// then convert to Option
+			_ => $e.ok()
 		}
 	)
 }
@@ -590,7 +593,7 @@ macro_rules! usage {
 									ELSE { value_t!(matches, stringify!($arg), $($arg_type_tt)+) }
 								)
 							}
-						)).ok();
+						));
 					)*
 				)*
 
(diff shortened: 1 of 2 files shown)


parity/cli/usage.rs
```diff
@@ -32,6 +32,17 @@ macro_rules! otry {
 	)
 }
 
+macro_rules! return_if_parse_error {
+	($e:expr) => (
+		match $e {
+			Err(clap_error @ ClapError { kind: ClapErrorKind::ValueValidation, .. }) => {
+				return Err(clap_error);
+			},
+			_ => $e
```
debris: it would be better to unwrap the value here rather then calling `ok()` later.

```rust
		match $e {
 			Err(clap_error @ ClapError { kind: ClapErrorKind::ValueValidation, .. }) => {
 				return Err(clap_error);
 			},
 			Ok(e) => e
         }
```

arkpar: What's the reason to ignore errors other than `ClapErrorKind::ValueValidation` btw?

axelchalon: @arkpar the clap macro `value_t!` (whose result I feed into the `return_if_parse_error!` macro) yields a `Result` ; the error can be either `ClapErrorKind::ArgumentNotFound` or `ClapErrorKind::ValueValidation`. If it's `ArgumentNotFound` then the value of the argument in the struct should be `None`.

@debris I thought about it but figured it'd be better if the macro didn't convert to Option for separation of concern, but I agree it'd be better for DRY. `$e` can be either `ClapErrorKind::ValueValidation` error, or `ClapErrorKind::ArgumentNotFound` error, or `Ok(...)` ; in the two latter cases this is normal behaviour and we want to convert it to Option; I guess the following would be good?
```
match $e {
 			Err(clap_error @ ClapError { kind: ClapErrorKind::ValueValidation, .. }) => {
 				return Err(clap_error);
 			},
 			_ => $e.ok()
         }
```

debris: yes, just describe it in the comment ;)

PR #6730, opened 2017-10-12
Fix badges not showing up

ngotchac:


ngotchac:
Stable : #6731
Beta : #6732

Issue #6594, opened 2017-09-26
Auto_update flag not working within config.toml file


noswald:
I'm running:
**Parity version**: 1.7.0
**Operating system**:  Linux
**And installed**: via installer or source (can't remember which)

I recently noticed that parity was not auto updating on my machine, even after using the config.toml file, as outlined here: https://github.com/paritytech/parity/wiki/Configuring-Parity#cli-options

The config.toml file has the following lines:

    [parity]
    # Auto-updates to every release.
    auto_update = "all"

Is this a known issue? What is the recommended fix to enable auto_updates?

5chdn:
Can reproduce with `--auto-update all --release-track beta` on 1.6.10

tomusdrw:
Is it a Parity issue or the new release was just not pushed to the contract?

5chdn:
Probably not pushed. Where is the contract?

debris:
we've investigated this issue with @arkpar . It has been fixed in `1.7.7` and in `1.8.0`. Master will be fixed once #6742 is merged

Issue #6360, opened 2017-08-23
Migrate to ethabi 4.0.0


debris:
I think this could be split into 2 separate steps:

- [x] make current generator use `ethabi 4.0.0`
- [x] add generator functionality to `ethabi`
- [x] replace generator with `use_contract!` macro
- [x] `updater` crate should use `ethabi_contract` crate (@axelchalon)

debris:
@axelchalon after #6742 is merged, it might be good to tackle down `updater crate should use ethabi_contract crate`

PR #6742, opened 2017-10-13
updated ethabi to version 4.0

debris:
part of #6360


debris:
don't merge it for now. `ethabi 4.0` has the same issue with strict encoding that `2.0` had

debris:
closes #6594

PR #6795, opened 2017-10-16
remove internal ipc

debris:


svyatonik:
`secret_store/*` looks fine, thanks!

Issue #6693, opened 2017-10-10
Last part of util refactor


debris:
last part of #6418

- [x] `kvdb` should be separated into 2 separate crates: `kvdb` and `kvrocksdb` which implements key-value interface for rocksdb
- [x] `journal` crate should be separated from util. `overlaydb` and `util_error` crate should be merged into it. also `UtilError` should be just renamed to `Error`.
- [x] `cache` should be separated from `util`
- [ ] `util` should be dissolved

debris:
I will take
- `kvdb` should be separated into 2 separate crates: `kvdb` and `kvrocksdb` which implements key-value interface for rocksdb

0x7CFE:
I will take `cache` and then probably `journal` stuff.

0x7CFE:
Now working on `journal`.

PR #6801, opened 2017-10-17
Refactors journaldb as a separate crate

0x7CFE:
Affects #6693.

- Refactors `journaldb` as a separate crate
- Refactors `ethcore` and `parity` to use it instead of `util`

Note: need to check for now redundant dependencies in `.toml`s.


0x7CFE:
As far as I see, remaining issues are in the tests only https://gitlab.parity.io/parity/parity/-/jobs/67430

However, along with issues, introduced by my refactoring I see other stuff like `keccak`, `ethcore_logger`, `kvdb_memorydb` and so on. Looks like some previuos refactoring broke  dependencies as well.

PR #6804, opened 2017-10-17
util/network uses error_chain

debris:
changes:
- util/network uses error_chain
- turned off backtraces in error_chain 
- changed the name of `NetworkError` to `Error`


sync/src/blocks.rs
```diff
@@ -392,15 +392,15 @@ impl BlockCollection {
 						},
 						None => {
 							warn!("Got receipt with no header {}", h);
-							return Err(NetworkError::BadProtocol)
+							return Err(network::ErrorKind::BadProtocol.into())
```
tomusdrw: There is also `bail!` macro you could use:
`bail!(network::ErrorKind::BadProtocol)`

util/network/Cargo.toml
```diff
@@ -34,6 +34,7 @@ ethcore-logger = { path ="../../logger" }
 ipnetwork = "0.12.6"
 hash = { path = "../hash" }
 serde_json = "1.0"
+error-chain = { version = "0.11", default-features = false }
```
tomusdrw: Why we don't want default features? Is it because we print backtrace in other places?

util/network/src/error.rs
```diff
@@ -82,89 +80,83 @@ impl fmt::Display for DisconnectReason {
...
-	AddressParse(::std::net::AddrParseError),
-	/// Error concerning the network address resolution subsystem.
-	AddressResolve(Option<::std::io::Error>),
-	/// Error concerning the Rust standard library's IO subsystem.
-	StdIo(::std::io::Error),
-	/// Packet size is over the protocol limit.
-	OversizedPacket,
-}
+error_chain! {
+	types {
```
tomusdrw: Isn't that optional?

util/network/Cargo.toml
debris: Yes, we don't want backtraces (at least for now). It may change in future and we may add them to the debug builds

sync/src/blocks.rs
debris: I didn't know about it ;) I will use it next time, ok? :)

tomusdrw: Sure. I haven't yet decided which I like better, so not pushing it :)

util/network/src/error.rs
```diff
@@ -85,6 +85,7 @@ error_chain! {
 		SocketIo(IoError) #[doc = "Socket IO error."];
 		Io(io::Error) #[doc = "Error concerning the Rust standard library's IO subsystem."];
 		AddressParse(net::AddrParseError) #[doc = "Error concerning the network address parsing subsystem."];
+		Decomression(snappy::InvalidInput) #[doc = "Decompression error."];
```
rphmeier: decompression

debris:
I will reopen the pr, cause I have a feeling that CI fails because cargo cached something. I can't reproduce the problem on my machine.

PR #6807, opened 2017-10-17
util crates use tempdir crate instead of devtools to create temp path

debris:
part of #6147 . I haven't changed every usage of devtools, cause it's gonna be easier to review and merge in parts


arkpar:
Needs merge.

PR #6818, opened 2017-10-18
Refresh cached tokens based on registry info & random balances

jacogr:


jacogr:
@ngotchac Thank you. As discussed, leaving that one for UI-2 where we need to do a lot of further optimisations.

Issue #6096, opened 2017-07-19
potential concurrency bug with multiple json-rpc threads


barrymac:
Running a naive simple load test with 150 concurrent users submitting transactions to 

`{"method":"personal_sendTransaction",
"params":[
{
	"from":"0x00a329c0648769a73afac7f9381e08fb43dbea72",
	"to":"0x0000000000000000000000000000000000000000",
	"value":"0x1"},
""],
"id":${ctr},
"jsonrpc":"2.0"}
`

With json-rpc threads on default then all is good. With JSON-RPC threads set to 4 then 3 out of 4 return errors as follows: 

` 2017-07-19 15:26:32  Transaction mined (hash 38d6cbe82b5ec1322803f2e7029bb1bbe8fd773eed5b1cf3babee013c490deaa)
2017-07-19 15:26:32  Error importing transaction: Transaction(AlreadyImported)
2017-07-19 15:26:32  Error importing transaction: Transaction(AlreadyImported)
2017-07-19 15:26:32  Error importing transaction: Transaction(AlreadyImported)
2017-07-19 15:26:32  Transaction mined (hash 1f879f8ed9a59a0740dd93ed3cf47440831c1b2c8133dafaa202a9664e20da0e)
`

This may be due to nonces not being taken from a queue once they are being processed.

rphmeier:
Probably a simple race condition:
  1. all 4 threads acquire a read lock on the queue to see the next nonce of the sender
  2. the threads each construct a transaction with that nonce
  3. the thread which acquires a write lock to import first "wins" the race and the others fail to import the same transaction.

A quick fix would be to attempt transaction construction in a loop, doing a kind of CAS with the existing nonce. A higher throughput fix would involve handing out "incomplete"/"under-construction" slots in the queue for transactions to be constructed in-place.

tomusdrw:
Closed via #6834

PR #6834, opened 2017-10-20
Reserve nonces for signing

tomusdrw:
Solves #6096 

@keorn @barrymac Can you guys check if it solves the issue?
I did a simple test with `--jsonrpc-server-threads 4 --jsonrpc-threads 8` and sending ~10k transactions from a single sender on a PoA chain which seemed to work fine.


keorn:
Yep, solves the issue.

PR #6842, opened 2017-10-20
Fix JSON tracing for sub-calls.

tomusdrw:
CC @holiman @cdetrio


arkpar:
Does this change instructions trace to be show pre-execution data, rather than post-execution?

tomusdrw:
The state is still pre-execution (gas used, memory, stack, etc) although the trace should be printed only on successful gas calculation.
Because of that we were printing `CREATE/CALL/DELEGATECALL` instruction after all sub-call traces which looked weird.

Before:
```
..create initialization (depth: 1)
PUSH (depth: 2)
CREATE (depth: 1) // create executed
```

Now:
```
..create initialization (depth: 1)
CREATE (depth: 1) // create executed
PUSH (depth: 2)
```

holiman:
Looks correct to me

PR #6854, opened 2017-10-22
updated ntp to version 0.3

debris:
ntp updates https://github.com/JeffBelgum/ntp/pull/9


dapps/node-health/Cargo.toml
```diff
@@ -9,7 +9,7 @@ authors = ["Parity Technologies <admin@parity.io>"]
 futures = "0.1"
 futures-cpupool = "0.1"
 log = "0.3"
-ntp = "0.2.0"
+ntp = "0.3.0"
```
MysticRyuujin: Should this be 0.3.1?

debris: last 0 doesn't matter. we also use a lockfile

PR #6860, opened 2017-10-23
Iterate over both buffered and unbuffered database entries

### LINKED ISSUE #6048: Support iteration over both buffered and unbuffered database entries
Currently only flushed data is iterated over. I can mentor.

I wrote a little crate that may help: https://crates.io/crates/interleaved-ordered

efyang:
Closes #6048.

### DIFF
--- a/util/kvdb-rocksdb/src/lib.rs
+++ b/util/kvdb-rocksdb/src/lib.rs
@@ -515,13 +515,14 @@ impl Database {
 		match *self.db.read() {
 			Some(DBAndColumns { ref db, ref cfs }) => {
 				let overlay = &self.overlay.read()[Self::to_overlay_column(col)];
-				let overlay_data = overlay.iter()
-					.filter_map(|(k, v)| match v {
-						&KeyState::Insert(ref value) |
-						&KeyState::InsertCompressed(ref value) =>
+				let mut overlay_data = overlay.iter()
+					.filter_map(|(k, v)| match *v {
+						KeyState::Insert(ref value) |
+						KeyState::InsertCompressed(ref value) =>
 							Some((k.clone().into_vec().into_boxed_slice(), value.clone().into_vec().into_boxed_slice())),
-						&KeyState::Delete => None,
+						KeyState::Delete => None,
 					}).collect::<Vec<_>>();
+				overlay_data.sort();
 
 				let iter = col.map_or_else(
 					|| db.iterator_opt(IteratorMode::Start, &self.read_opts),
--- a/util/kvdb-rocksdb/src/lib.rs
+++ b/util/kvdb-rocksdb/src/lib.rs
@@ -21,6 +21,7 @@ extern crate elastic_array;
 extern crate parking_lot;
 extern crate regex;
 extern crate rocksdb;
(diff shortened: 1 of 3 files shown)


efyang:
[clabot:check]

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -510,17 +512,25 @@ impl Database {
 
 	/// Get database iterator for flushed data.
 	pub fn iter(&self, col: Option<u32>) -> Option<DatabaseIterator> {
-		//TODO: iterate over overlay
 		match *self.db.read() {
 			Some(DBAndColumns { ref db, ref cfs }) => {
+				let overlay = &self.overlay.read()[Self::to_overlay_column(col)];
+				let overlay_data = overlay.iter()
+					.filter_map(|(k, v)| match v {
```
tomusdrw: use `match *` instead of prefixing every option with `&`:
```
match *v {
  KeyState::Insert(ref value) => {
}
```

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -510,17 +512,25 @@ impl Database {
...
+						&KeyState::Insert(ref value) |
+						&KeyState::InsertCompressed(ref value) =>
+							Some((k.clone().into_vec().into_boxed_slice(), value.clone().into_vec().into_boxed_slice())),
+						&KeyState::Delete => None,
+					}).collect::<Vec<_>>();
```
tomusdrw: Do we really need to allocate here?

efyang: I'm having trouble finding some other way to convert `elasticarray`'s to boxed slices.

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -510,17 +512,25 @@ impl Database {
 
 	/// Get database iterator for flushed data.
 	pub fn iter(&self, col: Option<u32>) -> Option<DatabaseIterator> {
-		//TODO: iterate over overlay
 		match *self.db.read() {
 			Some(DBAndColumns { ref db, ref cfs }) => {
+				let overlay = &self.overlay.read()[Self::to_overlay_column(col)];
+				let overlay_data = overlay.iter()
```
rphmeier: hashmap iterators aren't ordered, so interleave-ordered behavior probably won't be right there.

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -536,7 +547,7 @@ impl Database {
 						.expect("iterator params are valid; qed"));
 
 				Some(DatabaseIterator {
-					iter: iter,
+					iter: interleave_ordered(Vec::new(), iter),
```
debris: it should use [`std::iter::empty`](https://doc.rust-lang.org/std/iter/fn.empty.html) instead of `Vec::new`

debris: nvm, I see that types would not match

debris:
merged and reopened on top of latest master in #7048

PR #6861, opened 2017-10-23
fix #6228: do not display eth price in cli for etc

### LINKED ISSUE #6228: eth price in cli should not be displayed for private chains
```
2017-08-04 00:19:29   INFO miner  Updated conversion rate to Ξ1 = US$223.41 (0 wei/gas)
```

nicolasochem:
Current behaviour:

When the client is started with defaults, the miner's gas calibrator
periodically queries the price of ether in usd and uses it to adjust
the wei_per_gas variable, displaying an info message in the cli each
time calibration happens.

This behaviour is also present when running on a non-foundation chain
such as ethereum classic.

When the client is started with --min-gas-price N, this does not happen
as GasPricer::Fixed is instanciated instead of GasPricer::Calibrated.

Solution:

For non-foundation chain, hard-code a default gas price (wei_per_gas,
currently 40M wei) in the spec json file.

When the wei_per_gas parameter is found and greater than zero, force the
miner to start with GasPricer::Fixed, so the calibrator does not run at
all.

### DIFF
--- a/ethcore/res/ethereum/classic.json
+++ b/ethcore/res/ethereum/classic.json
@@ -31,7 +31,8 @@
 		"forkCanonHash": "0x94365e3a8c0b35089c1d1195081fe7489b528a84b22199c916180db8b28ade7f",
 		"eip155Transition": 3000000,
 		"eip98Transition": "0x7fffffffffffff",
-		"eip86Transition": "0x7fffffffffffff"
+		"eip86Transition": "0x7fffffffffffff",
+		"weiPerGas": 40000000
 	},
 	"genesis": {
 		"seal": {
--- a/ethcore/res/ethereum/expanse.json
+++ b/ethcore/res/ethereum/expanse.json
@@ -31,7 +31,8 @@
 		"subprotocolName": "exp",
 		"eip98Transition": "0x7fffffffffffff",
 		"eip86Transition": "0x7fffffffffffff",
-		"eip155Transition": "0x927C0"
+		"eip155Transition": "0x927C0",
+		"weiPerGas": 40000000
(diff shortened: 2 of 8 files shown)


nicolasochem:
@tomusdrw is mining on alternate chains a use case for parity ?

When you pass `--min-gas-price`, `GasPricer::Fixed` gets called. If I disable `GasPricer` on non-foundation chain, the `min-gas-price` argument will be ignored on chains other than Foundation.
https://github.com/paritytech/parity/blob/master/parity/configuration.rs#L650-L652

I think there has to be a default wei_per_gas, otherwise, when mining on ethereum classic, the default behaviour is to mine any transaction, even these with no fee. As different chains may have different default wei_per_gas, it's best to put it in the spec file.

tomusdrw:
Hey @nicolasochem, correct me if I'm wrong but it seems that your PR is doing the following:
- If `weiPerGas` is present in the spec file any gas-price related parameters are ignored and the hardcoded value is used as fixed minimal gas price instead.

What I meant is:
- if `chain != foundation` don't use `GasPricerConfig::Calibrated` as default but rather some fixed minimal gas price from CLI options.

The latter approach will still allow you to override the value with CLI or config and since calibrated gas price doesn't make sense anywhere apart from foundation chain I would disable it everywhere but there by default instead of requiring some spec files modifications.

The ideal solution would include chain-specific defaults, which is we discussed in the team some time ago. Also related: #6228

nicolasochem:
Fine, but I just need to take care of the case where `--min-gas-price` is not set. In the opposite case, we already have what we want. I did that in #6877 .

PR #6874, opened 2017-10-24
Get local transactions by hash in the light client

rphmeier:
After https://github.com/paritytech/parity/pull/6868
cc @tomusdrw


bjornwgnr:
@rphmeier looks like #6868 was merged

PR #6875, opened 2017-10-24
warn when blacklisted account present in store

rphmeier:
Closes https://github.com/paritytech/parity/issues/6799

It might not play well with setups where there are lots of accounts, but we should still do this check. Later we can optimize to store an addresses index if the key store is large enough.


tomusdrw:
Note: This will not warn if the blacklisted account is in the address book (will be just silently removed from there)

PR #6877, opened 2017-10-24
fix #6228: do not display eth price in cli for etc

### LINKED ISSUE #6228: eth price in cli should not be displayed for private chains

nicolasochem:
Current behaviour:

When the client is started with defaults, the miner's gas calibrator
periodically queries the price of ether in usd and uses it to adjust
the wei_per_gas variable, displaying an info message in the cli each
time calibration happens. The info message mentions the price of ETH.

When started with the --min-gas-price option, the calibrator is inactive
and the price is not displayed.

Problem:

When running on an alternate chain such as ethereum classic, the info
message mentioning the ETH price is present, unless the --min-gas-price
option is used.

Solution:

if chain != foundation and --min-gas-price is not set,
don't use GasPricerConfig::Calibrated as default but rather fix
the minimum gas price to zero.


parity/configuration.rs
```diff
@@ -649,6 +649,8 @@ impl Configuration {
 			return Ok(GasPricerConfig::Fixed(to_u256(dec)?));
 		} else if let Some(dec) = self.args.arg_min_gas_price {
 			return Ok(GasPricerConfig::Fixed(U256::from(dec)));
+		} else if self.args.arg_chain != "foundation" {
+			return Ok(GasPricerConfig::Fixed(U256::zero()));
```
arkpar: Should be something like `if Spec_type::from_str(self.chain()).map_or(false, |s| s == SpecType::Foundation` because there are other strings that match the foundation chain.

tomusdrw: Might be worth to wait for #6878 (or integrate same changes) where `self.chain()` returns a `SpecType`

nicolasochem:
@arkpar @tomusdrw I have updated PR to use the updated `self.chain()` that returns a `SpecType`, please review.

tomusdrw:
Closes #6228

PR #6917, opened 2017-10-29
improve noMatchPassword grammar

makc:
no idea which branch to target, so aiming for master


makc:
[clabot:check]

jacogr:
The js/i18n/_defaults is actually the extracted text (via script). It needs to be changed in the actual component where FormattedMessage is used. As it stands it won't have an effect on the UI.

PR #6928, opened 2017-10-30
Fix Windows NSIS installer not working

axelchalon:
The Windows NSIS installer currently wouldn't work.

Since https://github.com/paritytech/parity/pull/6356, the `--warp` argument was removed, and the ordering of the arguments changed (global arguments come before subcommands). So Parity as set up by the installer would throw an error on startup.

Not sure how to test this; I don't think that the Windows installer is tested in the CI?


axelchalon:
Woops just noticed that this is covered in https://github.com/paritytech/parity/pull/6921 ; closing as this is a duplicate.

PR #6953, opened 2017-11-01
Fix order of arguments in run_parity()

axelchalon:
`--can-restart` and `--force-direct` need to come as first arguments (before any subcommand). Fixes https://github.com/paritytech/parity/issues/6906.

### DIFF
--- a/parity/main.rs
+++ b/parity/main.rs
@@ -259,7 +259,7 @@ fn run_parity() -> Option<i32> {
 	use ::std::ffi::OsString;
 	let prefix = vec![OsString::from("--can-restart"), OsString::from("--force-direct")];
 	let res = latest_exe_path().and_then(|exe| process::Command::new(exe)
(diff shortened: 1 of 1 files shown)


arkpar:
Would it be possible to restore old behaviour and allow generic flags after commands instead? There a few other things broken by this change, not to mentions user's scripts and setups.

axelchalon:
I agree that the change of the ordering of the arguments isn't intuitive... AFAIK there's no clean way with *clap* to allow global arguments to be allowed everywhere. To allow this, we could add (internally) every global argument to each subcommand but this would be a bit hacky. It might be the better choice for the users, though.

arkpar:
If we can do that with a macro, so that it does not have to be repeated for each flag or command that would be good enough.

axelchalon:
Yes, that's what I meant. OK, I'll get started on this.

axelchalon:
This PR will be made irrelevant once the following PR is merged : https://github.com/paritytech/parity/pull/7004 (allows global arguments to be passed at the end, just like it used to be before we switched to Clap).

PR #6957, opened 2017-11-01
Add hint in ActionParams for splitting code/data

NikVolf:
WASM can't use merged code/data section, it needs some hint when to split out constructor arguments to pass it as a code arguments


fckt:
lgtm

PR #6961, opened 2017-11-01
Assorted improvements for ethstore and ethkey

tomusdrw:
- Brute forcing presale wallet given a password list (ethstore)
- Multi-threaded generation of prefix-matching addresses (ethkey)
- Prefix-matching brain wallets (ethkey)
- Brain wallets recovery phrases lookup (ethkey) - helps to find a correct phrase if you know the address you want to get yet you made a typo backing the phrase up (or forgot a word)


5chdn:
Does this need docs?

tomusdrw:
There are some (very undetailed) docs in the `--help`, I can add more if you want. Do we have a wiki page for these stuff too?

5chdn:
No, only the READMEs afaik

DaleChristopher:
GREAT!!!! but I'd very much appreciate some kind of instructions geared towards helping the end user/ non-programmer attempting to use this. It might seem straightforward to you guys but I have had zero training in coding and have been trying to figure out what i need to even run it for the last hours... >_<
I'm sure I will figure it out eventually but I'm teaching myself a new skill here with no prior experience. Some kind of basic step by step would be really helpful. @tomusdrw great work btw! cant wait to test it out!

ethkey/src/brain_recover.rs
```diff
@@ -0,0 +1,177 @@
...
+	known_phrase: &str,
+	expected_words: usize,
+	filter: Option<(usize, usize)>,
+) -> Result<String, Error> {
+	let known_words = parity_wordlist::WORDS.iter().cloned().collect::<HashSet<_>>();
+	let mut words = known_phrase.split(' ')
+		.map(|word| match known_words.get(word) {
+			None => {
+				info!("Invalid word '{}', looking for potential substitutions.", word);
+				let substitutions = generate_substitutions(word);
```
debris: what if I defined words myself?

ethkey/README.md
```diff
@@ -16,29 +16,29 @@ Ethereum keys generator.
   Copyright 2016, 2017 Parity Technologies (UK) Ltd
 
 Usage:
-    ethkey info <secret> [options]
+    ethkey info <secret-or-phrase> [options]
     ethkey generate random [options]
-    ethkey generate prefix <prefix> <iterations> [options]
-    ethkey generate brain <seed> [options]
```
debris: this makes it backward incompatible. I was using this option a lot for my personal accounts

ethkey/src/brain_recover.rs
tomusdrw: Then we can't help you. This recovery process is targeted at people who used parity-generated phrase but made a mistake while backing it up.
We could make a more generic recovery process (which allows you to provide your own dictionary) as an extra feature.

ethkey/README.md
tomusdrw: Bumped version of ethkey and ethstore and added relnotes llabel.

IMHO it's more consistent now, if you provide `<seed>` we don't really `generate` anything, we just print address info (similar to when you are providing the secret), there is a `secret` derivation in the background obviously, but I think it significantly differs from `generate random` and `generate prefix` commands.

ethstore/cli/src/main.rs
```diff
@@ -243,19 +245,51 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
...
-				},
-				_ if counter % 100 == 0 => print!("."),
-				_ => {},
-			}
+		let passwords = passwords.lines().map(str::to_owned).collect::<VecDeque<_>>();
+		let passwords = sync::Arc::new(sync::Mutex::new(passwords));
+
+		let mut handles: Vec<thread::JoinHandle<_>> = Vec::new();
+		let threads = num_cpus::get();
+		for _ in 0..threads {
```
dkashitsyn: I may be wrong but looks like Rayon may be used here instead of raw threads.

ethkey/cli/src/main.rs
```diff
@@ -214,12 +249,65 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
...
+		handles.push(thread::spawn(move || match t() {
+			Ok(solution) => {
+				tx.send(solution).expect("Receiving end never dropped.");
+				Ok(())
+			},
+			Err(err) => Err(err),
+		}));
 	}
+
+	if let Ok(solution) = rx.recv() {
```
svyatonik: Is it possible that `recv` will return `Err` here? It only fails when sender is destroyed an here are `max+1` senders, of which this last 1 is never destroyed. So it looks like that if all threads have failed with an error, we will hang in this `recv` call. I can be missing something, though :)

ethkey/cli/src/main.rs
```diff
@@ -214,12 +249,65 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
...
+	let mut handles = Vec::new();
+	let max = num_cpus::get();
+
+	let (tx, rx) = sync::mpsc::sync_channel(1);
+	for i in 0..max {
+		let t = task(i, max);
+		let tx = tx.clone();
+		handles.push(thread::spawn(move || match t() {
+			Ok(solution) => {
+				tx.send(solution).expect("Receiving end never dropped.");
```
svyatonik: When first thread founds a solution, it is received via `rx.recv()` and function returns => receiving end is destroyed. Since we do not wait for other threads on successful return (is it intended?), other threads are still running and when next `t()` call will succeed, this would lead to panic. Not sure it is possible to have several solutions in these tasks, though...

ethkey/cli/src/main.rs
tomusdrw: Good points, thanks. Will addresss both by switching to `rayon` as suggested by @0x7CFE

tomusdrw:
Decided to use `threadpool` instead of rayon, since I'm using a regular iterator and also addressed two comments from Slava by that.

PR #6967, opened 2017-11-02
Events in WASM runtime

NikVolf:


debris:
@fckt can you also review it? and merge if it looks good

fckt:
looks good. Tested on dev chain also.

ethcore/wasm/src/runtime.rs
```diff
@@ -749,6 +767,44 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
+					.checked_mul(schedule.log_data_gas as u64)
+					.and_then(|data_gas| data_gas.checked_add(topics_gas))
+			}
+		)?;
+
+		let mut topics: Vec<H256> = Vec::with_capacity(topic_count as usize);
+		topics.resize(topic_count as usize, H256::zero());
+		for i in 0..topic_count {
+			let offset = i.checked_mul(32).ok_or(UserTrap::MemoryAccessViolation)?
+				.checked_add(topic_ptr).ok_or(UserTrap::MemoryAccessViolation)?;
```
rphmeier: checked_mul here seems like overkill given that 3 is the max value of i.

PR #6970, opened 2017-11-02
Adds validate_node_url() and refactors boot node check (#6907)

### LINKED ISSUE #6907: Invalid enode address format on failed name resolution
If name resolution fails during boot node setup, parity exits with the following error message:

```
Invalid node address format given for a boot node: enode://8925b57acc144d035ca91b32b93a4503103b65c25d4164f2a6b56ed6f0093b96f560e3e8a592816376e7c652c6155a7c8924eb703d2c7a15da3b4973e4ae40c3@host1:30303
```

This may confuse and mislead user, because enode format is actually correct. In order to fix that, we should alter error reporting logic, so that message would contain the actual reason (failed name resolution). "Invalid node address format" should be reported only when enode format is really syntactically invalid.

Version info:
```
Parity/v1.8.1-beta-0e4a06d-20171020/x86_64-linux-gnu/rustc1.21.0
Ubuntu Linux 16.04.3
Installed via parity-deploy.
```

0x7CFE:
This PR fixes #6907 by propagating error value instead of a `bool ` flag.

IMO, full solution should return the actual reason of failure including underlying error message. However, there are some issues with it:
- [ ] Impl `Display` for `NetworkError` should handle error details
- [ ] Impl `FromStr` for `Node` and `NodeEndpoint` return `NetworkError::AddressResolve` even on incorrectly formatted addresses, like `foo`, `enode://foo@bar` and even on an empty string (#6972).

### DIFF
--- a/util/network/src/node_table.rs
+++ b/util/network/src/node_table.rs
@@ -393,7 +393,7 @@ mod tests {
 
 	#[test]
 	fn node_parse() {
-		assert!(validate_node_url("enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770").is_some());
+		assert!(validate_node_url("enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770").is_none());
 		let node = Node::from_str("enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770");
 		assert!(node.is_ok());
 		let node = node.unwrap();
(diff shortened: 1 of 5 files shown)


util/network/src/node_table.rs
```diff
@@ -368,6 +368,15 @@ pub fn is_valid_node_url(url: &str) -> bool {
 	Node::from_str(url).is_ok()
 }
 
+/// Same as `is_valid_node_url` but returns detailed `NetworkError`
+pub fn validate_node_url(url: &str) -> Option<NetworkError> {
```
tomusdrw: Why introduce another method then?
Wouldn't it be better to just do `is_valid_node_url().is_ok()` in places that don't care about the error?

0x7CFE: I thought about it and it was my first intention, however, in that case, method name would conflict with it's semantics. 

Also I'm a bit confused why we validate nodes by effectively doing all the work needed to construct `Node` just to throw it away and use the raw string value. Wouldn't it be better to return `Reusult<Vec<Node>, NetworkError>` from `init_reserved_nodes`? Anyway, I thought, that would cause substantial refactoring and is out of the scope.

For now I think it would be better to stick with the new function and refactor existing calls to `is_valid_node_url`.

0x7CFE:
Removed obsolete variant `is_valid_node_url()`.

@tomusdrw, currently `validate_node_url()` returns `Option<NetworkError>`. Maybe it would be better to return `Result<(), NetworkError>`? That would allow check with `?` and other chaining.

Actually, we may go even further, and return `Result<Node, NetworkError>` but that would require re-exporting of the `Node` struct. Don't sure, if it's viable in this case.

debris:
>  Impl Display for NetworkError should handle error details

this one is probably addressed by #6979

PR #6990, opened 2017-11-06
Fixes for parity-extension

tomusdrw:


js/package.json
```diff
@@ -21,7 +21,7 @@
     "Parity"
   ],
   "scripts": {
-    "build": "npm run build:lib && npm run build:app",
+    "build": "npm run build:lib && npm run build:app && npm run build:embed",
```
jacogr: The only issue here is line 27 - fails on Windows.

tomusdrw: Solved by using `cross-env`

PR #6991, opened 2017-11-06
Adds `personal_signTransaction` RPC method

0x7CFE:
New RPC method `personal_signTransaction` is added. Existing `send_transaction`'s impl is refactored to reuse the common logic.

@tomusdrw, I believe I need to write some tests for this. Where should I host them?

### DIFF
--- a/rpc/src/v1/impls/personal.rs
+++ b/rpc/src/v1/impls/personal.rs
@@ -30,7 +30,7 @@ use v1::helpers::errors;
 use v1::helpers::dispatch::{Dispatcher, SignWith};
 use v1::helpers::accounts::unwrap_provider;
 use v1::traits::Personal;
-use v1::types::{H160 as RpcH160, H256 as RpcH256, U128 as RpcU128, TransactionRequest};
+use v1::types::{H160 as RpcH160, H256 as RpcH256, U128 as RpcU128, TransactionRequest, RichRawTransaction as RpcRichRawTransaction};
 use v1::metadata::Metadata;
 
 /// Account management (personal) rpc implementation.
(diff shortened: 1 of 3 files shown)


rpc/src/v1/impls/personal.rs
```diff
@@ -104,37 +133,21 @@ impl<D: Dispatcher + 'static> Personal for PersonalClient<D> {
...
-			None => accounts
-				.dapp_default_address(meta.dapp_id().into())
-				.map_err(|e| errors::account("Cannot find default account.", e)),
-		};
-
-		let default = match default {
-			Ok(default) => default,
-			Err(e) => return Box::new(future::err(e)),
-		};
+	fn sign_transaction(&self, meta: Metadata, request: TransactionRequest, password: String) -> BoxFuture<RpcH256, Error> {
```
tomusdrw: You need to get whole `RichRawTransaction`, just the hash is not useful at all. The whole point of this RPC method is to get signed RLP (raw transaction) that you will be able to propagate later.

tomusdrw:
The tests should go to `tests/mocked/personal.rs`

rpc/src/v1/tests/mocked/personal.rs
```diff
@@ -96,29 +96,40 @@ fn new_account() {
...
-	}"#;
+	}"#;	
 
 	let response = r#"{"jsonrpc":"2.0","error":{"code":-32021,"message":"Account password is invalid or account does not exist.","data":"SStore(InvalidPassword)"},"id":1}"#;
 
 	assert_eq!(tester.io.handle_request_sync(request.as_ref()), Some(response.into()));
 }
 
 #[test]
+fn sign_transaction_with_invalid_password() {
```
tomusdrw: Would be nice to have a positive test case as well.

0x7CFE: Don't sure how to check if the response is actually valid.

Is there a convenient way to validate the whole JSON response, knowing that it should describe a transaction? I.e. all fields should exist and types should be correct? Maybe I could try to deserialize it?

tomusdrw: Yeah, deserializing it makes sense.

PR #6994, opened 2017-11-06
Transaction Pool re-implementation

tomusdrw:
Initial version of new transaction pool.
Currently implement as a separate crate (and abstract over Transaction).

In subsequent PRs I plan to:
- [ ] integrate with Parity (implement `Ready`, `Scoring` and `Verifier`, penalties, own transactions, etc)
- [ ] Run some benchmarks (and compare with old queue)
- [ ] Migrate old tests to make sure the behaviour is consistent.
- [ ] Cache pending set (and then update it in-place)

The details are explained in docs, but the big picture is this:
- The pool maintains a list of transactions *per sender* ordered by `Score` (think: nonce). (`HashMap<Sender, Vec<Transaction>>`)
- We also know best and worst transaction from every sender and keep a queue of best-first transactions and worst-last transactions.
- Inserting to the queue is cheap (`O(no-of-tx-per-sender + log(no-of-senders))` (worst case), usually `O(log(no-of-tx-per-sender) + log(no-of-senders))`)
- Constructing pending set is relatively cheap (it's not `O(1)` as it used to be, but we are within `O(nlog(n))` (all transactions from a single sender), usually `O(n)`)
- Pluggable scoring allows us to do some interesting prioritizations (like high priority for a chain of small transactions not just looking at a single one = bump priority by sending a very expensive transaction later)
- Decoupling from nonce/gasPrice will allow to easily integrate account abstraction transactions
- Readiness abstraction allows to easily implement scheduled transactions


transaction-pool/src/scoring.rs
```diff
@@ -0,0 +1,144 @@
...
+	type Score: cmp::Ord + Clone + Default + fmt::Debug;
+
+	/// Decides on ordering of `T`s from a particular sender.
+	fn compare(&self, old: &T, other: &T) -> cmp::Ordering;
+
+	/// Decides how to deal with two transactions from a sender that seem to occupy the same slot in the queue.
+	fn choose(&self, old: &T, new: &T) -> Choice;
+
+	/// Updates the transaction scores given a list of transactions and a change to previous scoring.
+	/// NOTE: `txs.len() === scores.len()`
```
0x7CFE: Better to write explicitly

transaction-pool/src/transactions.rs
```diff
@@ -0,0 +1,213 @@
...
+		let index = match self.transactions.binary_search_by(|old| scoring.compare(old, &tx)) {
+			Ok(index) => index,
+			Err(index) => index,
+		};
+
+		// Insert at the end.
+		let len = self.transactions.len();
+		if index == len {
+			// trying to insert least worth transaction
+			return if len == max_count {
```
0x7CFE: This statement is part of a rather long method so all returns should be easily noticeable. When skimming through the code, one may easily misinterpret this statement as return from the if part only.

I believe this `return` statement should be refactored in favour of two explicit `returns` in each condition branch. Another option is to introduce temporary value, like `result` and add explicit `return result` after condition block.

P.S.: Maybe consider splitting this method into smaller parts?

transaction-pool/src/transactions.rs
```diff
@@ -0,0 +1,213 @@
...
+				self.transactions.push(shared.clone());
+				self.scores.push(Default::default());
+				scoring.update_scores(&self.transactions, &mut self.scores, scoring::Change::InsertedAt(index));
+
+				AddResult::Ok(shared)
+			}
+		}
+
+		// Decide if the transaction should be replaced
+		match scoring.choose(&self.transactions[index], &tx) {
```
0x7CFE: If I were you, I'd introduce `result` value here to put an emphasis on a fact that we return a value from that large `match`.

transaction-pool/src/tests/helpers.rs
```diff
@@ -0,0 +1,82 @@
...
+			if new.gas_price > old.gas_price {
+				scoring::Choice::ReplaceOld
+			} else {
+				scoring::Choice::RejectNew
+			}
+		} else {
+			scoring::Choice::InsertNew
+		};
+
+		decision
```
0x7CFE: Since this method is small and intentions are crystal clear, I may suggest omitting the `decision` value. Two nested `if`s would be OK, in my opinion.

transaction-pool/src/status.rs
```diff
@@ -0,0 +1,40 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+/// Light pool status.
+/// This status is cheap to compute and can be called frequently.
+#[derive(Debug, PartialEq, Eq)]
+pub struct LightStatus {
+	/// Memory usage in bytes.
+	pub mem_usage: usize,
+	/// Total number of transactions in the pool.
+	pub count: usize,
```
0x7CFE: Better to `s/count/transaction_count/`. More status fields may be added in future which may cause confusion.

transaction-pool/src/pool.rs
```diff
@@ -0,0 +1,443 @@
...
+		}
+
+		removed
+	}
+
+	/// Returns an iterator of pending (ready) transactions.
+	pub fn pending<R: Ready<T>>(&self, ready: R) -> PendingIterator<T, R, S, L> {
+		PendingIterator {
+			ready,
+			best_transactions: self.best_transactions.clone(),
```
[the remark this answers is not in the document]
0x7CFE: Probably not

transaction-pool/src/transactions.rs
```diff
@@ -0,0 +1,219 @@
...
+		};
+
+		self.transactions.remove(index);
+		self.scores.remove(index);
+		// Update scoring
+		scoring.update_scores(&self.transactions, &mut self.scores, scoring::Change::RemovedAt(index));
+		return true;
+	}
+
+	pub fn cull<R: Ready<T>>(&mut self, ready: &mut R, scoring: &S) -> SmallVec<[Arc<T>; PER_SENDER]> {
```
dkashitsyn: Something makes me think, that this method may be refactored using iterators instead of `reverse`s.

tomusdrw: Since we use `SmallVec` it doesn't have methods that allow easily to consume the content.
Current method doesn't require allocation, but is indeed `O(n)`, since number of transactions per sender is limited I don't expect it to be a problem.

With `Vec` we could use `split_off`, although instead of `Vec` it would be more wise to use `VecDeque`, but that would require more allocations - might be worth to consider though after some initial tests.

transaction-pool/src/transactions.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+/// Represents all transactions from a particular sender ordered by nonce.
+const PER_SENDER: usize = 8;
+#[derive(Debug)]
+pub struct Transactions<T, S: Scoring<T>> {
+	// TODO [ToDr] Consider using something that doesn't require shifting all records.
+	transactions: SmallVec<[Arc<T>; PER_SENDER]>,
+	scores: SmallVec<[S::Score; PER_SENDER]>,
+	_score: PhantomData<S>,
```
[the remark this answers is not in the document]
tomusdrw: Actually we don't, thanks for noticing :)

tomusdrw:
Thank you for the review @0x7CFE. Addressed grumbles and added some more detailed docs about the algo.

PR #7006, opened 2017-11-08
Disable uncles by default

keorn:


rstormsf:
:+1: merge!

5chdn:
This affects Kovan and all PoA, does it?

keorn:
Kovan will stay the same, it just works for new chains. If one wants to keep running a chain with uncles they have to add `maximumUncleCount: 2` to the chain spec.

debris:
How exactly is this value used? I can't see

rstormsf:
#7065

keorn:
@debris It determines the maximum number of uncles included in a block. Uncles are a fairly Ethash specific thing, so they should not be enabled by default.

PR #7011, opened 2017-11-09
test framework and basic test for whisper

rphmeier:


rphmeier:
seems that messages propagate, at least one hop. although the same logic is used for propagating local messages and received messages, so I don't see why the messages wouldn't propagate further if necessary. despite that, some users are reporting whisper nodes as unable to relay messages.

PR #7014, opened 2017-11-10
[beta] Disallow builtin multisig deploy (only watch)

jacogr:
Removes the ability to deploy multisig from the "+ Wallet" button on accounts. The only available option is to watch existing wallets.

Updates for master done as part of https://github.com/paritytech/parity/pull/6935


5chdn:
will this also be ported to stable?

PR #7016, opened 2017-11-10
Fix windows build (with ui rebuild)

tomusdrw:
With some update of rustc running external `npm` command on windows stopped working.


debris:
@tomusdrw does it need to be backported?

PR #7018, opened 2017-11-10
Wasm tests update

pepyakin:
- We no longer use legalization of JS FFI. Thus we can use i64 types in interfaces directly and we no longer need to use `setTempRet0`. Also, LTO have been enabled for the tests, so gas consumption improved a bit.
- All libc related functions are now with prefix `ext_`, e.g. `_malloc` became `_ext_malloc`. This is needed to workaround some emscripten specific behavior and might change in the future.
- New functions `_ext_memmove` and `_ext_memset` added. 
- `mem_copy` fixes. The arguments poped in a wrong order. Also make it really return a value, as LLVM really could use it sometimes. 
- Small improvement of a test ergonomics: gas checks are pushed to the end of the test functions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2253,7 +2253,7 @@ dependencies = [
 
 [[package]]
 name = "parity-wasm"
-version = "0.15.1"
+version = "0.15.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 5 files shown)


NikVolf:
Already discussed and partially reviewed, so quick lgtm :)

PR #7025, opened 2017-11-12
Fix nonce reservation

tomusdrw:
Kudos to @ngotchac


5chdn:
Does this require a patch?

tomusdrw:
https://github.com/paritytech/parity/pull/6834 wasn't backported, so no.

rpc/src/v1/helpers/nonce.rs
```diff
@@ -15,35 +15,84 @@
...
+			pool,
+		}
+	}
+
+	/// Reserve a nonce for particular address.
+	///
+	/// The reserved nonce cannot be smaller than the minimal nonce.
+	pub fn reserve(&mut self, sender: Address, minimal: U256) -> Reserved {
+		if self.nonces.len() + 1 > Self::CLEAN_AT {
+			let to_remove = self.nonces.iter().filter(|&(_, v)| v.is_empty()).map(|(k, _)| *k).collect::<Vec<_>>();
```
rphmeier: HashMap::retain

Issue #6136, opened 2017-07-24
decouple `Client` from `Importer`


rphmeier:
`Importer` handles verifying, enacting, and writing incoming blocks.
`Client` handles reading state and blockchain data from previously-enacted blocks.

I would estimate that currently each of these roles takes up about half of the code of `Client`, so the refactor should be a relatively even split.

0x7CFE:
What's the status for this issue?

PR #7038, opened 2017-11-13
`Client` refactoring

0x7CFE:
Affects #6136.

The main idea is to decouple `Importer` from the `Client`. Also minor tweaks are to be performed to enhance code readability.

This PR is very much in progress.

---
This change is [Reviewable](https://reviewable.io/reviews/paritytech/parity/7038)


ethcore/src/blockchain/best_block.rs
```diff
@@ -19,7 +19,8 @@ use bigint::hash::H256;
 use bytes::Bytes;
 use header::BlockNumber;
 
-/// Best block info.
+/// Best block info. Best block is the block with highest combined difficulty
+/// (usually the block with the highest block number). Sometimes refered as latest block.
```
rphmeier: best block heuristic is really dependent on the consensus engine. this new description is only applicable to the GHOST fork-choice rule

ethcore/src/client/client.rs
```diff
@@ -144,32 +144,70 @@ impl SleepState {
 /// Blockchain database client backed by a persistent database. Owns and manages a blockchain and a block queue.
 /// Call `import_block()` to import a block asynchronously; `flush_queue()` flushes the queue.
 pub struct Client {
+	/// Flag used to disable the client forever. Not to be confused with `liveness`.
+	///
+	/// For example, auto-updater will disable client forever if there is a
+	/// hard fork registered on-chain that we don't have capability for.
+	/// When hard fork block rolls around, the client (if `update` is false)
+	/// knows it can't proceed further.
```
rphmeier: the fields here are not public. comments are great, but they should be `//`

ethcore/src/trace/db.rs
```diff
@@ -102,7 +102,10 @@ enum CacheId {
 	Bloom(TraceGroupPosition),
 }
 
-/// Trace database.
+/// Whenever a transaction is executed by evm it's execution trace is stored
+/// in trace database. Each trace has information, which contracts have been
+/// touched, which have been created during the execution of transaction, and
+/// which calls failed.
```
rphmeier: multi-line docs should have a single line first, then a gap, and then the rest. it formats much more nicely on the generated docs

ethcore/src/client/client.rs
debris: I see no reason to enforce using `//`. In multiple places in our codebase we use `//` for private properties and structs.

ethcore/src/blockchain/best_block.rs
0x7CFE: Should I rollback the comment entirely, or provide some extra text like this: 
```Rust
/// Contains information on a best block that is specific to the consensus engine. 
///
/// For GHOST fork-choice rule it would typically describe the block with highest 
/// combined difficulty (usually the block with the highest block number).
/// 
/// Sometimes refered as 'latest block'.
```
?

ethcore/src/trace/db.rs
0x7CFE: I've tried, but failed to describe the whole thing in one compact line for a header. Or should I reformat the very same text but with an extra blank line?

On the other hand it wouldn't be great to write meaningless first line that Captain Obvious would be proud of.

ethcore/src/blockchain/best_block.rs
rphmeier: text like that is fine. a couple nits, though: `combined` -> `cumulative` and the last line about sometimes being referred to as the latest block is correct, but those would be misuses of the term, so I'd recommend omitting that statement.

ethcore/src/trace/db.rs
rphmeier: an obvious line like "Block and transaction execution trace database" gives more information than the struct name while allowing for more information to be included later.

ethcore/src/client/client.rs
0x7CFE: RLS provides hover hints for fields commented with `///` only.

0x7CFE:
OK, at this point most of import related data and some of the logic were moved to the `Importer` struct (that would be extracted to a separate source file somewhat later).

Right now I'm trying to move `import_verified_blocks` and I've faced the following problem: the `import_verified_blocks` method invokes `Miner::chain_new_blocks` method that accepts `&MiningBlockChainClient` as a parameter. Current implementor of the `MiningBlockChainClient` trait is the `Client` struct.

Unfortunately, I do not see an easy way to resolve the issue. Either we need to rework the whole trait hierarchy or to redesign the `import_verified_blocks`.

@debris, @rphmeier, do you have something in mind on this issue?

debris:
I opt for reworking the trait hierarchy. Especially `MiningBlockChainClient`. We can start with  `MiningBlockChainClient` being passed to `Miner::chain_new_blocks`. Instead of it, we could pass only a structure which contains all data needed by a miner. I don't know how much work would it be, but to me it seems to be the cleanest solution. It could also positively impact our performance, cause currently we request the same data multiple times.

tomusdrw:
@debris You don't know what data will be needed by Miner beforehand - you need to recover the signatures of transactions to figure out what account balances you need for instance.
That would be possible but would require that a lot of logic currently hapening in miner would need to be placed somewhere else.

Can you elaborate more on what's wrong with `MiningBlockChainClient` trait exactly? Is it just too fat?

debris:
@tomusdrw you're right. It's not trivial. 

> Can you elaborate more on what's wrong with MiningBlockChainClient trait exactly? Is it just too fat?

Yes, I think that's the main issue now.

0x7CFE:
> Can you elaborate more on what's wrong with MiningBlockChainClient trait exactly? Is it just too fat?

Well, since `MiningBlockChainClient: BlockChainClient`, implementor must provide all methods from both traits, so yeah, it's quite fat.

0x7CFE:
Here's the complete list of client methods that are used by `Miner::chain_new_blocks`

From `BlockChainClient`:
```Rust
fn best_block_header(&self) -> encoded::Header;
fn block(&self, id: BlockId) -> Option<encoded::Block>;
fn latest_nonce(&self, address: &Address) -> U256;
fn latest_balance(&self, address: &Address) -> U256;
fn chain_info(&self) -> BlockChainInfo;
```

From `MiningBlockChainClient`
```Rust
fn reopen_block(&self, block: ClosedBlock) -> OpenBlock;
fn prepare_open_block(&self, author: Address, gas_range_target: (U256, U256), extra_data: Bytes) -> OpenBlock;
```

tomusdrw:
Fair enough, I though that the trait was separate.

IMHO we should duplicate methods in the `MiningBlockChainClient` (`Importer`) or extract a common method stub from both `BlockchainClient` and `MiningBlockChainClient`.

A blanket implementation might also work, but that needs to be checked in code.

0x7CFE:
OK, I've digged a little bit into trait hierarchy and may say the following.

Initially, when I tried to refactor `Client` logic I was forced to depend on the methods I do not use. Looks like current trait hierarchy is violating the [single responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) as well as the [interface segregation](https://en.wikipedia.org/wiki/Interface_segregation_principle) principles.

Typically it all happens in C++ code as it forces interface hierarchies due to limitations of it's dynamic dispatch and lack of concept/trait system. Luckily we use Rust, so this should not be a problem ;)

In order to resolve the issue I propose the following:
- Realize what interfaces do we have, i.e. methods are logically related and typically used together;
- Group them in "feature traits" with speaking names, even if trait would have a single method in it;
- Define those traits as supertraits for `BlockChainClient`, `MiningBlockChainClient` and others;
- Rework `Miner` dependencies, so that each method will depend only on those feature traits and those methods, that are really needed;
- Finally, implement required feature traits in `Importer` and move `Client::import_verified_blocks`.

Using this approach we may refactor one method at a time without global changes or code breakage.

Right now, I'm trying to group methods from all affected traits by their roles in order to figure out, what would be the feature traits. I would appreciate any suggestions and/or help.

0x7CFE:
Well, that was harder than I expected. Anyway, I've just committed proof-of-concept implementation of the refactored `Miner` that is now based on aforementioned "feature traits" idea. However, during  the implementation I've faced several fundamental limitations of modern Rust :)

Unfortunately, modern Rust does not support neither structural inheritance nor subtyping [other than for lifetimes](https://doc.rust-lang.org/nomicon/subtyping.html).

Long story short, you could neither cast nor coerce trait object references. For example, this would **NOT** work:
```Rust
// Some struct
struct A;

// Some trait hierarchy
trait Foo {};
trait Bar: Foo {};

// Note, that struct implements both traits
impl Foo for A {};
impl Bar for A {};

// Function accepting base trait object
fn accepts_foo(foo: &Foo) {}

// Function trying to coerce trait objects 
fn accepts_bar(bar: &Bar) {
    accepts_foo(bar); // Nope, incompatible types
    accepts_foo(bar as &Foo); // This would not work either
}
```

This happens because `bar` is a trait object and it could not be coerced to `&Foo`, even though `A` implements `Foo`. Also, current Rust implementation does not allow you to write `bar: &(Foo + Bar)`, since both traits are not automatic.

And that's exactly what happens when you try to split `MiningBlockChainClient` into smaller pieces and use them independently.

For example, `ServiceTransactionChecker::update_from_chain_client` use client only to call `registry_address` method. You may want to do the following:
```Rust
/// Provides information on a blockchain service and it's registry
pub trait RegistryInfo {
	/// Get the address of a particular blockchain service, if available.
	fn registry_address(&self, name: String) -> Option<Address>;
}

/// Blockchain database client. Owns and manages a blockchain and a block queue.
pub trait BlockChainClient : Sync + Send + RegistryInfo { ... }

impl RegistryInfo for Client { ... }

impl ServiceTransactionChecker {
	/// Try to create instance, reading contract address from given chain client.
	pub fn update_from_chain_client(&self, client: &RegistryInfo) { 
		let mut contract = self.contract.lock();
		if contract.is_none() {
			*contract = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned())
				.and_then(|contract_addr| {
					trace!(target: "txqueue", "Configuring for service transaction checker contract from {}", contract_addr);

					Some(Contract::new(contract_addr))
				})
		}
	}
}
```

But when you try to pass `&MiningBlockChainClient` to `update_from_chain_client`, you'll have compile time error. Moreover, you could not cast `&MiningBlockChainClient` to `&RegistryInfo` even taking into account that trait object actually implements `RegistryInfo`.

The main problem is that `&MiningBlockChainClient` is a trait object and trait objects are invariant. Since types were already erased when passing client reference to `Miner`, compiler could not typecheck. Double dynamic dispatch simply doesn't work that way.

Partial solution may be to provide casting methods in the trait hierarchy:
```Rust
trait Foo {};
trait Bar: Foo {
    fn as_foo(&self) -> &Foo;
};
```

But that approach leads to messy trait hierarchies with lots and lots of aggregate traits (just to stick it in place) and basically prevents you from writing compact interfaces and implementations, effectively forcing you into god object antipattern.

Thats why I tried to rethink the `Miner` hierarchy altogether and eventually migrate it to static dispatch. Since we already wanted to dissolve `Miner` sometimes, I thought that it would not be a major issue.

New implementation uses static dispatch where possible and having non-erased types allows you to coerce references automatically.

Please note, that even in new implementation, transaction queue logic remains dynamic, since `&TransactionQueueDetailsProvider` is dynamic. Basically that allows `Miner` to handle several `Client` implementations concurrently. In the end, dynamic dispatch is used only at the "last mile" where it is really needed.

The major drawback of this implementation is that `Miner` could not be trait object anymore, since it now has template methods. It was used as `Arc<MinerService>` in tests, but the change was trivial. In other places in code, both `Client` and `Miner` are used as types.

@todr, @debris, @rphmeier, could you please review this PoC carefully and give me some feedback on whether this approach is viable?

arkpar:
That's a nice summary. +1 for converting to static dispatch.

tomusdrw:
IMHO the `Client` should be split into these separate structs in the future. The traits will give us a nice abstraction for refactoring:
- `BlockChainClient` - able to produce latest and historic state objects
- `Miner` - able to produce pending state objects
- `Importer` - validating and writing blocks/states
- `StateReader` - given state can retrieve data from it or transact

0x7CFE:
@todr 

> This one is weird, does it need to be a trait?

```Rust
/// Provides information on a blockchain service and it's registry
pub trait RegistryInfo {
	/// Get the address of a particular blockchain service, if available.
	fn registry_address(&self, name: String) -> Option<Address>;
}
```

It's used here:

```Rust
impl ServiceTransactionChecker {
	/// Try to create instance, reading contract address from given chain client.
	pub fn update_from_chain_client<C: RegistryInfo>(&self, client: &C)
	{
		let mut contract = self.contract.lock();
		if contract.is_none() {
			*contract = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned())
				.and_then(|contract_addr| {
					trace!(target: "txqueue", "Configuring for service transaction checker contract from {}", contract_addr);

					Some(Contract::new(contract_addr))
				})
		}
	}
}
```

> This one imho indicates a leaking abstraction. Can we analyze what exactly we need from the schedule? Maybe we can extract smaller part of it.

Well, currently `ScheduleInfo` is defined as following:

```Rust
/// Provides `latest_schedule` method
pub trait ScheduleInfo {
	/// Returns latest schedule.
	fn latest_schedule(&self) -> Schedule;
}
```

Honestly, I don't know how to make it even smaller :) It's used by the `Miner` logic in a single place:

```Rust
impl<'a, C> TransactionQueueDetailsProvider for TransactionDetailsProvider<'a, C>
	where C: ScheduleInfo
{
	fn estimate_gas_required(&self, tx: &SignedTransaction) -> U256 {
		tx.gas_required(&self.client.latest_schedule()).into()
	}
}
```

I've extracted only the logic recursively needed by `Miner::chain_new_blocks`. Other parts of `Miner` were left untouched, where possible.

0x7CFE:
Finally, I was able to decouple `Client` and `Importer` logic… sort of. 

Almost all import-related methods were moved to the `Importer` struct. Remaining methods were extracted to separate trait `ImportBlock` that is imlemented by `Client`.

So, current `impl Client` contains only methods that are specific to the client itself, such as:
- Lifecycle management routines: constructor, enable/disable stuff, snapshots, GC
- Various accessors to fields and chain state
- Helper methods such as `last_hashes` stuff
- Event handling routines

The only exceptions left are `import_queued_transactions` and `import_verified_blocks` methods that are used by `ClientService`. These were left because of a reference to a `Client` (see below).

Now the tricky part. 

Unfortunately, it's nearly impossible to decouple `Importer` from `Client` completely. The main reason is that `Importer` needs `Client` to do transaction verification and other stuff. This is a problem, because such references break component hierarchy and prevents further interface segregation.

All such back references are now expressed as `&Client` parameter in `Importer`'s methods. By navigating through use points within such methods, one may easily see what routines are actually needed.

In order to solve the issues we need to:

- Inject concrete trait dependencies instead of the whole `&Client`, so instead of "Just give me a client reference, I know what to do with it" we'd have "I need something that allows me to call contracts, something to acquire write access to a DB, etc"
- Ideally, we should have `Option<Importer>` field in `Client` which would be set to `None` if we only need to query the DB.
- Split `Client` into smaller parts and rework "who owns what", maybe extract all DBs into `DatabaseManager` that would own them and provide statically checked read or write access.

Also we need to move `Importer` to a separate file.

0x7CFE:
@rphmeier suggested to host `Importer` in the `ClientService`, which seems a good idea. Looks like it would allow `import_queued_transactions` and `import_verified_blocks` to be moved to `Importer`.

rphmeier:
> import_queued_transactions

Ideally we would also have `Miner` separate from `Client` as well and this would be used only there.

0x7CFE:
In order to decouple `Client` and `Miner` we need to solve the `BlockId::Pending` dichotomy. Almost all `Client`'s accessors either  query the chain state or, in case of `BlockId::Pending`, delegate the request to  `Miner`. 

Some random methods, just to illustrate the logic:
```Rust
fn block_body(&self, id: BlockId) -> Option<encoded::Body> {
	let chain = self.chain.read();

	if let BlockId::Pending = id {
		if let Some(block) = self.miner.pending_block(chain.best_block_number()) {
			return Some(encoded::Body::new(BlockChain::block_to_body(&block.rlp_bytes(Seal::Without))));
		}
		return self.block_body(BlockId::Latest);
	}

	// We know that `id` is not `Pending`, but we still must pass `&Miner`
	Self::block_hash(&chain, &self.miner, id).and_then(|hash| chain.block_body(&hash))
}

fn block_status(&self, id: BlockId) -> BlockStatus {
	if let BlockId::Pending = id {
		return BlockStatus::Pending;
	}

	// We already covered the `Pending` case, but...
	let chain = self.chain.read();
	match Self::block_hash(&chain, &self.miner, id) {
		Some(ref hash) if chain.is_known(hash) => BlockStatus::InChain,
		Some(hash) => self.importer.block_queue.status(&hash).into(),
		None => BlockStatus::Unknown
	}
}

fn block_hash(chain: &BlockChain, miner: &Miner, id: BlockId) -> Option<H256> {
	match id {
		BlockId::Hash(hash) => Some(hash),
		BlockId::Number(number) => chain.block_hash(number),
		BlockId::Earliest => chain.block_hash(0),
		BlockId::Latest => Some(chain.best_block_hash()),

		// `&Miner` is needed for `Pending` only...
		BlockId::Pending => miner.pending_block_header(chain.best_block_number()).map(|header| header.hash())
	}
}
```

We may never want to check the pending state, but current API forces us to have `Miner`. This ties `Miner` to `Client` and prevents further refactoring.

One way is to rework the accessors contract, so that `Client` would always answer `None` to `BlockId::Pending` queries, meaning that one should ask `Miner` instead. We may even change the return type from `Option` to something more descriptive for such scenario.

However this change is quite serious and may affect logic in a surprising ways, so it should be done carefully.

@rphmeier, any thoughts on that?

tomusdrw:
Imho we should get rid of `BlockId:Pending`, pending can be decided on RPC level most of the time, and RPC can choose between Client and Miner.

lght:
On the road to removing `BlockId::Pending`, could we just remove the conditional in `block_status()`?

Maybe, instead of returning an `Option`, provide a descriptive error to use `Miner::BlockId::Pending` in `block_hash()`.

Similiarly in `block_body()`, could we retrieve latest block from the new decoupled `Miner`?

tomusdrw:
@lght What I mean is not that we shouldn't hadnle `Pending` (i.e. return error), but remove this variant completely, so it's never possible to compile a code like this:
```
client.block_hash(BlockID::Pending)
```

0x7CFE:
> What I mean is not that we shouldn't hadnle Pending (i.e. return error), but remove this variant completely

This was my initial intention, but that time I thought, it would be way too hard and/or inconvenient. If we really could achieve this, it would be a better way, IMO.

tomusdrw:
I don't think it would be inconvenient, it requires some changes in RPC though. Parity is not even producing pending blocks for 90% of the users, so if there is any occurence outside of RPC (where the call explicitly states pending block) it can be safely replaced with `Latest`.

lght:
> What I mean is not that we shouldn't hadnle Pending (i.e. return error), but remove this variant completely

Understand now, does Rust have the ability to throw custom compilation errors? What I'm thinking is a nice error there for devs, pointing to the new API.

0x7CFE:
If we remove `Pending` completely, Rust will mark all use points as invalid, forcing us to fix those parts of the logic. I do not see any problem here.

0x7CFE:
While light client was quite easy to refactor, normal one have some serious issues.

To eliminate the need for `BlockId::Pending` we need to refactor RPC logic in a way that would select `client` or `miner` depending on the `BlockNumber` value.

However, that would require massive refactoring in the `Client` logic. 
For example, current implementation of RPC `balance` calls `client.balance` that internally invokes `miner.pending_state`, which in turn, depends on the `chain` that's owned by `Client`.

In order to break the dependency loop, the following workarounds were proposed:
```Rust
fn balance<B: Backend, T: Into<State<B>>>(block: T, address: &Address) -> Option<U256> {
	block.into().nonce(address).ok()
}
```

```Rust
enum StateOrBlock {
	State(..), // Pending state from the Miner
	Block(BlockId), // Sealed block from DB
}

fn balance<T: Into<StateOrBlock>(block: T, ...) {}
```

Unfortunately, that would require adding generic methods to a trait hierarchy, and that will prevent `Client` from being a trait object. There are a lot of places where `Client` is used as a trait object and it would take a lot of effort to refactor them to a static dispatch (in the same way as `Miner`).

Right now I'm trying to investigate alternative approaches.

0x7CFE:
OK, as a quick and dirty solution we may implement RPC handlers as following:

```Rust
fn balance(&self, address: RpcH160, num: Trailing<BlockNumber>) -> BoxFuture<RpcU256> {
	let address = address.into();

	let result = match num.unwrap_or_default() {
		BlockNumber::Pending => {
			let info = self.client.chain_info();

			if let Some(state) = self.miner.pending_state(info.best_block_number) {
				match state.nonce(&address).ok() {
					Some(balance) => Ok(balance.into()),
					None => Err(errors::state_pruned()) // FIXME
				}
			} else {
				Err(errors::state_pruned()) // FIXME
			}
		},

		num => {
			try_bf!(check_known(&*self.client, num.clone()));

			let num_to_id = |num| match num {
				BlockNumber::Num(num) => BlockId::Number(num),
				BlockNumber::Earliest => BlockId::Earliest,
				BlockNumber::Latest => BlockId::Latest,
				_ => unreachable!("`BlockNumber::Pending` must be handled manually")
			};

			match self.client.balance(&address, num_to_id(num)) {
				Some(balance) => Ok(balance.into()),
				None => Err(errors::state_pruned()),
			}
		}
	};

	Box::new(future::done(result))
}
```

Don't sure if it would be suitable for all methods, though.

0x7CFE:
Finally, thanks to @tomusdrw, I was able to realize the type hierarchy that would probably fit RPC needs. Hopefully, that would be enough to refactor remaining RPC methods and get rid of `BlockId::Pending` completely as well as decouple client and miner stuff.

`Pending` part of refactoring is quite special, because it highly affects not only the trait hierarchy, but also how types are accessed within the generic code.

The main limitation of the code base is the dynamic `&BlockChainClient` legacy, that still prevents us from having static dispatch where needed. In particular, it does not allow us to refactor existing `Client` methods to be generic.

For example, this implementation is desired, but is not currently possible, since it would be object unsafe:
```Rust
fn balance<T: Into<StateOrBlock>>(&self, address: &Address, state: S) -> Option<U256> {
	match state.into() {
		StateOrBlock::State(s) => s.balance(address).ok(),
		StateOrBlock::Block(id) => self.state_at(id).and_then(|s| s.balance(address).ok())
	}
}
```

So, we ended up having non-generic client methods that accept `state: StateOrBlock`, forcing users to do `.into()` all over the codebase. It's dirty, but still manageable.

The other issue was the miner and client decoupling. Previously client used miner internally to access pending state in case `BlockId::Pending` was passed. That was considered poor practice, since it ties modules together and prevents refactoring.

We decided to eliminate `BlockId::Pending` and instead pass `StateOrBlock` enum that would contain either block id from the sealed chain, or a pending state to use. In the latter case caller should obtain state manually.

And that's what actually happens on the RPC side. Luckily, RPC handlers have access to both `Client` and `Miner` and may delegate call to either of them. So, in case pending state is required, RPC handler calls newly exported `MinerService::pending_state` method and wraps it in `StateOrBlock::State`. And here where the problems arise, actually.

The problem is that `State` is generic over backend. For example, normal `Client` uses `state_db::StateDB` as a backend, but other implementations may be possible. Also, we have an abstraction layer over `Miner` and `Client`: `MinerService` and `BlockChainClient` respectively. And we definitely don't want to expose implementation details there. We could not leave state generic either, as it would break object safety.

Current solution uses associated types within `MinerService` and newly added `StateClient` (impld by `Client`) traits:

```Rust
/// Miner client API
pub trait MinerService : Send + Sync {
	type State: StateInfo + 'static;
// ...
	fn pending_state(&self, latest_block_number: BlockNumber) -> Option<Self::State>;
}

/// Provides methods to access chain state
pub trait StateClient {
	type State: StateInfo + 'static;

	/// Get a copy of the best block's state.
	fn latest_state(&self) -> Self::State;
}
```

`StateInfo` is a trait that is implemented by `State` and provides read-only access to the state data:
```Rust
/// Provides subset of `State` methods to query state information
pub trait StateInfo {
	/// Get the nonce of account `a`.
	fn nonce(&self, a: &Address) -> trie::Result<U256>;

	/// Get the balance of account `a`.
	fn balance(&self, a: &Address) -> trie::Result<U256>;

	/// Mutate storage of account `address` so that it is `value` for `key`.
	fn storage_at(&self, address: &Address, key: &H256) -> trie::Result<H256>;

	/// Get accounts' code.
	fn code(&self, a: &Address) -> trie::Result<Option<Arc<Bytes>>>;
}

// Blanket implementation for all backends
impl<B: Backend> StateInfo for State<B> {
	fn nonce(&self, a: &Address) -> trie::Result<U256> { State::nonce(self, a) }
	fn balance(&self, a: &Address) -> trie::Result<U256> { State::balance(self, a) }
	fn storage_at(&self, address: &Address, key: &H256) -> trie::Result<H256> { State::storage_at(self, address, key) }
	fn code(&self, address: &Address) -> trie::Result<Option<Arc<Bytes>>> { State::code(self, address) }
}
```

RPC handlers use `StateInfo` to query state information without knowing the actual type. So for most of handlers the following helper method works OK:

```Rust
fn get_state(&self, number: BlockNumber) -> StateOrBlock {
	match number {
		BlockNumber::Num(num) => BlockId::Number(num).into(),
		BlockNumber::Earliest => BlockId::Earliest.into(),
		BlockNumber::Latest => BlockId::Latest.into(),

		BlockNumber::Pending => {
			let info = self.client.chain_info();

			self.miner
				.pending_state(info.best_block_number)
				.map(|s| Box::new(s) as Box<StateInfo>)
				.unwrap_or(Box::new(self.client.latest_state()) as Box<StateInfo>)
				.into()
		}
	}
}
```

However, in some places we need to access the state mutably. For example, `Client::call` and `call_many` need to mutate the state to do their job. Unfortunately, looks like it's impossible to export proper mutable API from `StateInfo` to be used dynamically, since there are a lot of generic types in the original `State` API.

Luckily, latest implementation of `Client::latest_state` and `Miner::pending_state` export real types instead of type erased `Box<StateInfo>`. Since RPC logic (`EthClient`) is generic over `M: MinerService` and `C: MiningBlockChainClient + StateClient`, it would have access to a real state types to be passed to `call` routines.

In order to do so, we need to refactor the `call` and `call_many` methods, so that they would accept mutable state and block header as parameters, instead of obtaining them internally. I hope, it would not cause any more issues.

ethcore/src/client/client.rs
```diff
@@ -141,283 +143,288 @@ impl SleepState {
...
+						proposed_blocks.push(block.bytes);
+					} else {
+						imported_blocks.push(header.hash());
 
-	/// The env info as of the best block.
-	pub fn latest_env_info(&self) -> EnvInfo {
-		self.env_info(BlockId::Latest).expect("Best block header always stored; qed")
-	}
+						let route = self.commit_block(closed_block, &header, &block.bytes, client);
+						import_results.push(route);
```
debris: all import routes can be flattened to single one

ethcore/src/client/client.rs
```diff
@@ -451,24 +458,24 @@ impl Client {
...
+			client.factories.clone(),
 			is_epoch_begin,
 		);
 		let mut locked_block = enact_result.map_err(|e| {
 			warn!(target: "client", "Block import failed for #{} ({})\nError: {:?}", header.number(), header.hash(), e);
 		})?;
 
-		if header.number() < self.engine().params().validate_receipts_transition && header.receipts_root() != locked_block.block().header().receipts_root() {
+		if header.number() < engine.params().validate_receipts_transition && header.receipts_root() != locked_block.block().header().receipts_root() {
 			locked_block = locked_block.strip_receipts();
```
debris: this is interesting. 

`import_verified_blocks` calls this function... and this function replaces receipt_root in block_header? Can it possibly lead to invalid block being imported?

ethcore/src/client/client.rs
```diff
@@ -1457,23 +1596,28 @@ impl BlockChainClient for Client {
 
 	fn block_hash(&self, id: BlockId) -> Option<H256> {
 		let chain = self.chain.read();
-		Self::block_hash(&chain, &self.miner, id)
+		Self::block_hash(&chain, id)
 	}
 
-	fn code(&self, address: &Address, id: BlockId) -> Option<Option<Bytes>> {
-		self.state_at(id).and_then(|s| s.code(address).ok()).map(|c| c.map(|c| (&*c).clone()))
+	fn code(&self, address: &Address, state: StateOrBlock) -> Option<Option<Bytes>> {
```
debris: I think it was already mention, but we should replace `Option<Option<...>>` with something more descriptive. Currently it's quite confusing

ethcore/src/client/client.rs
```diff
@@ -1794,56 +1894,60 @@ impl BlockChainClient for Client {
...
+				.unwrap_or_else(Vec::new);
+
+			for h in uncles {
+				if !block.uncles().iter().any(|header| header.hash() == h) {
+					let uncle = chain.block_header(&h).expect("find_uncle_hashes only returns hashes for existing headers; qed");
+					block.push_uncle(uncle).expect("pushing up to maximum_uncle_count;
+												push_uncle is not ok only if more than maximum_uncle_count is pushed;
+												so all push_uncle are Ok;
+												qed");
+					if block.uncles().len() >= max_uncles { break }
```
debris: what if the block had already maximum of uncles before reopening?

ethcore/src/client/client.rs
0x7CFE: Yep, we have discussed it in https://github.com/paritytech/parity/issues/7098. Still I think, we need to refactor one thing at a time. Will probably return to it later, when dust settles.

ethcore/src/client/client.rs
0x7CFE: This PR didn't introduced new logic here, code was moved from an existing `Client` impl. Still it's a good question.

0x7CFE:
The more I dig in, the more complicated it gets.

Last refactoring issue was related to a `State` struct that is generic over `B: Backend` and is used all over the client API. In some cases it's used immutably to access info. In that cases we may pass a reference to a boxed trait object to invoke accessors.

Unfortunately, as it was described in the previous message, that is not enough. Methods like `call` and `call_many` require mutable access to the state. However, knowing generic nature of the `State` it would not be wise to leak the underlying implementation into the public API. Still, we need to access state as a type when dealing with mutable access, since mutating methods are quite complex and, moreover, are generic and not object safe. Therefore we need to pass `State` as a type in that cases.

Our last hope was that if we refactor `call` and `call_many` to accept state instead of obtaining it inside, thus we may decouple `Miner` logic from the `Client`.

One way to solve both issues (provide generic API and still have full type access) is to use associated types. The following traits was exported:

```Rust
/// Provides methods to access chain state
pub trait StateClient {
	type State: StateInfo + 'static;

	/// Get a copy of the best block's state.
	fn latest_state(&self) -> Self::State;

	/// Attempt to get a copy of a specific block's final state.
	///
	/// This will not fail if given BlockId::Latest.
	/// Otherwise, this can fail (but may not) if the DB prunes state or the block
	/// is unknown.
	fn state_at(&self, id: BlockId) -> Option<Self::State>;
}

/// Provides `call` and `call_many` methods
pub trait Call {
	/// State type accepted by methods
	type State: StateInfo;

	/// Makes a non-persistent transaction call.
	fn call(&self, tx: &SignedTransaction, analytics: CallAnalytics, state: &mut Self::State, header: &Header) -> Result<Executed, CallError>;

	/// Makes multiple non-persistent but dependent transaction calls.
	/// Returns a vector of successes or a failure if any of the transaction fails.
	fn call_many(&self, txs: &[(SignedTransaction, CallAnalytics)], state: &mut Self::State, header: &Header) -> Result<Vec<Executed>, CallError>;
}
```

The idea was that `Client` would implement those traits providing correct associate type (`State<state_db::StateDB>`) and RPC code would have access to the real types in handlers that require `call` and `call_many`. RPC would provide state either from client or from miner depending on the `BlockNumber` something like this:

```Rust
fn call(&self, meta: Self::Metadata, request: CallRequest, num: Trailing<BlockNumber>) -> BoxFuture<Bytes> {
	let request = CallRequest::into(request);
	let signed = try_bf!(fake_sign::sign_call(request, meta.is_dapp()));

	let num = num.unwrap_or_default();

	let result = if num == BlockNumber::Pending {
		 // TODO call using miner's state and pending block header
		unimplemented!()
	} else {
		let id = match num {
			BlockNumber::Num(num) => BlockId::Number(num),
			BlockNumber::Earliest => BlockId::Earliest,
			BlockNumber::Latest => BlockId::Latest,
		};

		let state = self.client.state_at(id).unwrap(); // TODO
		let header = self.client.block_header(id).unwrap(); // TODO

		self.client.call(&signed, Default::default(), &mut state, &header)
	};

	Box::new(future::done(result
		.map(|b| b.output.into())
		.map_err(errors::call)
	))
}
```

Unfortunately, we could not easily mix associated types of `StateClient::State` and `Call::State` in one expression, since Rust could not guarantee that these types would always be the same.

Right now we have the following options:
- merge `StateClient` and `Call` traits to a single one (to have only one associated type)
- provide mutable `State` API that would be object safe (and hide the details behind trait objects)
- eliminate dynamic dispatch completely and use convenient generics

I'm afraid, that the first option, being the simplest, still may not allow us to solve all of the issues.

0x7CFE:
> Unfortunately, we could not easily mix associated types of StateClient::State and Call::State in one expression, since Rust could not guarantee that these types would always be the same.

Looks like this issue may be solved by introducing trait bound that would explicitly state that associated types must be the same: 

```Rust
impl<..., C, T: StateInfo> Eth for EthClient<...> where
	C: StateClient<State=T> + Call<State=T>, ...
```

0x7CFE:
OK, at this point I've managed to refactor `call` and `call_many` methods and it's use points in the RPC.

The only method remaining is `call_contract` that's used in a bit different pattern. In order to refactor it as well, I need to know, whether it's use context implies only read-only access to an existing state or pending state is also possible. 

The issue is, if we alter `call_contract`'s parameters to accept `State` and `Header` as in `call`, we would need to compute such parameters at all use points. In some of them `Miner` may not be available, so it would be additional pain to pass it around.

Currently, `call_contract` is used from the following places:
- `Client::registry_address`, clearly R/O access to an existing state (no miner needed)
- `ValidatorSafeContract::default_caller`, looks like it's also miner independent logic
- `ServiceTransactionChecker::check`, by design, it's part of the `Miner` logic
- `TransactionFilter::transaction_allowed`, ?

I would be glad if someone would shed a light on the actual purpose and implications of aforementioned use places, probably excluding `Miner`'s stuff, which I already understand well.

ethcore/src/client/traits.rs
```diff
@@ -48,12 +50,135 @@ use types::block_status::BlockStatus;
...
+pub trait Call {
+	/// State type accepted by methods
+	type State: StateInfo;
+
+	/// Makes a non-persistent transaction call.
+	fn call(&self, tx: &SignedTransaction, analytics: CallAnalytics, state: &mut Self::State, header: &Header) -> Result<Executed, CallError>;
+
+	/// Makes multiple non-persistent but dependent transaction calls.
+	/// Returns a vector of successes or a failure if any of the transaction fails.
+	fn call_many(&self, txs: &[(SignedTransaction, CallAnalytics)], state: &mut Self::State, header: &Header) -> Result<Vec<Executed>, CallError>;
```
debris: why is `&mut Self::State` needed and not just `&Self::State`?

0x7CFE: Because it's accessed mutably in the client internals.

0x7CFE: This is the main point that requires mutable access: https://github.com/paritytech/parity/blob/client_split/ethcore/src/client/client.rs#L1216

debris: Isn't it contradictory with the method description (`non-persistent` calls)? Maybe we could add a comment that state changes should not be applied ?

tomusdrw: `non-persistent` = state is not applied
`dependent` = call `n+1` uses the state after `n` was applied

I'm fine with additional comment explaining that.

0x7CFE: Well, I don't see any problem, actuall. If you have a state that may be used for a call, you pass it "as is", but method prototype explicitly states, that passed state _may_ and probably _would_ be modified. If you're not comfortable with it, make a clone and pass it instead.

0x7CFE: > Isn't it contradictory with the method description (non-persistent calls)? 

I believe that's what `call_many` is for in the first place. It's ability to call several contracts on the same mutable state, instead of calling them one by one without having state shared between the calls.

Of course, we may think about more suitable API. Currently, it's just a way to solve the initial issue of client and miner decoupling.

tomusdrw: Probably `call` and `call_many` in trait definition could be unifined into `call(Vec<>)`

0x7CFE:
Finally, `BlockId::Pending` was completely eliminated from the code and all places were `BlockNumber::Pending` is used (mainly in RPC logic) were refactored to handle it explicitly.

I would be really glad, if someone familiar with client logic would check control paths that affect `Pending` and ensure, that I correctly interpreted and refactored the logic.

Most controversial, from my point of view, is the following pattern:
```Rust
let info = self.client.chain_info();
let pending_block = self.miner.pending_block(info.best_block_number);
```

Note, that actually we're querying `miner` but best block is taken from the `client`. Maybe I don't quite understand the underlying implications, but it looks very suspicious to me. Maybe it worked before, when miner was part of the client and current refactoring may render this assumption invalid.

So, the following tasks are on horizon:
- Update the tests
- Solve the merge conflicts and update PR according to changes in `master`
- Recent changes of trait hierarchy and associated types make dynamic `State` dispatch redundant and it may be factored out
- Eliminate the need to pass `&Client` to `Importer` methods, make this dependency more transparent
- Think about `DatabaseManager` and type checked access patterns
- Adopt something like [capability based APIs](http://zsck.co/writing/capability-based-apis.html)
- Move `Importer` to a separate file

P.S.: Tests were not refactored yet and probably would not compile. Making them work is another story.

tomusdrw:
```
let info = self.client.chain_info();
let pending_block = self.miner.pending_block(info.best_block_number);
```

It's required because miner is also storing oldish blocks (in case a delayed seal for such blocks is received in the future - they still can be included as uncles), so when we are attempting to get pending state from miner we need to be sure that it's no older than the latest state.

In principle it shouldn't be required, since right after new block is imported new pending block should be created, but since this happens concurrently we could have a race condition and this check prevents it.

rpc/src/v1/impls/eth.rs
```diff
@@ -143,9 +176,46 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM> EthClient<C, SN, S, M, EM> where
...
+				let extra = pending_block.as_ref().map(|b| self.client.engine().extra_info(&b.header));
+
+				(pending_block.map(|b| encoded::Block::new(b.rlp_bytes(Seal::Without))), Some(difficulty), extra)
+			},
+
+			BlockNumberOrId::Number(num) => {
+				let id = match num {
+					BlockNumber::Latest => BlockId::Latest,
+					BlockNumber::Earliest => BlockId::Earliest,
+					BlockNumber::Num(n) => BlockId::Number(n),
```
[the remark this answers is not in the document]
0x7CFE: > In principle `BlockNumberOrId::Id(hash)` and `BlockNumberOrId::Number(BlockNumber::Num(num))` can refer to pending block as well.

Hm… I thought I've ported this part from the client code as is. But I agree, it seems reasonable. 

Also, I think, it would be better to abstract all those matches somehow to conform DRY principle. But the problem is that `Pending` handling is highly context dependent. I've tried to write a function that would match but that time I thought, only `Pending` should be handled by `Miner`. And it messed the logic in a way that verbose matches seemed the best option.

Anyway, I will look again for a way to handle it properly.

0x7CFE:
At this point I've managed to make tests compile. Unsurprisingly, some of them fail. Failed tests related to recent modifications of call methods and added mock impls that currently do nothing.

In order to bring tests back to life we need to design correct mock impls for several traits like `StateClient`.

0x7CFE:
Tests were updated according to the refactoring changes and all pass now.

`StateClient` and `Call` traits were implemented for `TestBlockChainClient` and `TestMinerService` with  used `()` as associated `State` type. That way it prevents unneeded dependencies on a real `State<StateDB>` and simplifies logic in mock code, since `state` parameter is not used by mocked methods anyway.

I would appreciate review of the test code to check for false positives.

0x7CFE:
Finally, trait hierarchy was modified according to @tomusdrw comment:  https://github.com/paritytech/parity/pull/7038#pullrequestreview-79464201

Currently, `Miner` API  and implementation uses only aggregate traits like `BlockChain` (which is actually `ChainInfo + BlockInfo + TransactionInfo`) or `SealedBlockImporter` ( which is  `ImportSealedBlock + BroadcastProposalBlock`) insteaad of "elementary" feature traits. That way at a later date we may move methods from elementary traits to aggregate ones without breaking code.

Now it looks something like this:
```Rust
/// New chain head event. Restart mining operation.
fn update_sealing<C>(&self, chain: &C)
    where C: AccountData + BlockChain + RegistryInfo 
          + CallContract + BlockProducer + SealedBlockImporter;

/// Submit `seal` as a valid solution for the header of `pow_hash`.
/// Will check the seal, but not actually insert the block into the chain.
fn submit_seal<C: SealedBlockImporter>(&self, chain: &C, pow_hash: H256, seal: Vec<Bytes>) -> Result<(), Error>;

/// Get the sealing work package and if `Some`, apply some transform.
fn map_sealing_work<C, F, T>(&self, client: &C, f: F) -> Option<T>
    where C: AccountData + BlockChain + BlockProducer + CallContract,
          F: FnOnce(&ClosedBlock) -> T,
          Self: Sized;
```

In my opinion it's good that just by looking at a method declaration you already may have an idea what it does and what permissions it require.

However, there are some issues remaining. And most of them are due to wide use of trait objects in the codebase:

- We could not merge `Call` and `CallContract` traits, since former has an associated type `State` whereas latter is used a a bound of `BlockChainClient` trait which, in turn, is used in a trait object pattern.

- There are two traits `ImportBlock` and `SealedBlockImporter` that could not be merged for the same reason. But we would probably not want to do that, since implications of these traits are different.

- `ImportBlock` is used as a bound for `BlockChainClient` and not `MiningBlockChainClient` which is kinda strange. And it's not a trivial task to move `import_block` to `MiningBlockChainClient` because of existing `&BlockChainClient` references across the codebase. This may be done eventually, but in a separate PR.

-  Some traits like `RegistryInfo` and `ScheduleInfo` indicate a leaking abstraction somewhere and it need to be reworked again. Looks like such methods do not need to be in traits in the first place.

Aside of that PR looks ready for me. All tests pass.

0x7CFE:
Even if review would be successful I do _not_ recommend merging it, before I run some tests on Aura and other chains.

0x7CFE:
OK, I performed tests using Aura as well as the Foundation chains. Looks like all is working as expected, so we may probably consider this branch safe to merge when review would be completed.

ethcore/src/client/client.rs
```diff
@@ -771,56 +624,290 @@ impl Client {
...
-						};
+							res.map(|(output, proof)| (output, proof.into_iter().map(|x| x.into_vec()).collect()))
+						};
+
+						match with_state.generate_proof(&call) {
+							Ok(proof) => proof,
+							Err(e) => {
+								warn!(target: "client", "Failed to generate transition proof for block {}: {}", hash, e);
+								warn!(target: "client", "Snapshots produced by this client may be incomplete");
+								Vec::new()
```
debris: if `client` failed to generate `proof`, why do we insert an invalid proof (`Vec::new()`) into chain? line 644

debris: nvm. I see that this code hasn't been changed in this pr

ethcore/src/client/client.rs
```diff
@@ -1955,13 +2064,37 @@ impl MiningBlockChainClient for Client {
...
+impl BroadcastProposalBlock for Client {
+	fn broadcast_proposal_block(&self, block: SealedBlock) {
+		self.notify(|notify| {
+			notify.new_blocks(
+				vec![],
+				vec![],
+				vec![],
+				vec![],
+				vec![],
+				vec![block.rlp_bytes()],
```
debris: should these values be empty vectors?

ethcore/src/client/client.rs
```diff
@@ -1955,13 +2064,37 @@ impl MiningBlockChainClient for Client {
...
+				0,
+			);
+		});
+	}
+}
+
+impl SealedBlockImporter for Client {}
```
debris: why is it empty?

ethcore/src/client/client.rs
```diff
@@ -1165,79 +1251,235 @@ impl Client {
...
+	fn transaction_block(&self, id: TransactionId) -> Option<H256> {
+		self.transaction_address(id).map(|addr| addr.block_hash)
+	}
+}
+
+impl BlockChainTrait for Client {}
+
+impl RegistryInfo for Client {
+	fn registry_address(&self, name: String, block: BlockId) -> Option<Address> {
+		let address = match self.registrar_address {
```
debris: latest version of rust supports `?` for options, so you can write:

```rust
let address = self.registrar_address?;
```

ethcore/src/client/client.rs
debris: nvm, I see that it hasn't been changed in this pr

ethcore/src/client/client.rs
```diff
@@ -141,283 +143,288 @@ impl SleepState {
 	}
 }
 
+struct Importer {
+	/// Lock used during block import
+	pub import_lock: Mutex<()>, // FIXME Maybe wrap the whole `Importer` instead?
```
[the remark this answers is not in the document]
rphmeier: agree completely

5chdn:
@debris could you label it accordingly? Is this a "looksgood" or "mustngrumble"? Or does this require additional work?

debris:
looksgood and it requires additional review

ethcore/src/client/client.rs
0x7CFE: Yeah, good point.

ethcore/src/client/traits.rs
```diff
@@ -46,80 +49,198 @@ use types::block_status::BlockStatus;
...
+	/// Attempt to get a copy of a specific block's final state.
+	///
+	/// This will not fail if given BlockId::Latest.
+	/// Otherwise, this can fail (but may not) if the DB prunes state or the block
+	/// is unknown.
+	fn state_at(&self, id: BlockId) -> Option<Self::State>;
+}
+
+/// Provides various blockchain information, like block header, chain state etc.
+pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo /* + StateClient */ {}
```
rphmeier: Is the `StateClient` bound desired? Why is it commented?

ethcore/src/client/traits.rs
rphmeier: yeah, the `call_many` functionality seems like a superset of `call`

ethcore/src/client/client.rs
rphmeier: to answer the original question, I was a bit ambivalent about what to do. It indicates that either the validator contract or the local database are broken so whatever we do the chain is going to break at some point.

ethcore/src/client/client.rs
```diff
@@ -894,20 +981,31 @@ impl Client {
...
 	}
 
+	/// Get a copy of the best block's state.
+	pub fn latest_state(&self) -> State<StateDB> {
+		let header = self.best_block_header();
+		State::from_existing(
+			self.state_db.read().boxed_clone_canon(&header.hash()),
+			header.state_root(),
+			self.engine.account_start_nonce(header.number()),
+			self.factories.clone())
```
rphmeier: style: close paren should go on following line, with one level less indentation

ethcore/src/client/traits.rs
0x7CFE: Nope, it's an artifact. Previously I thought, that `BlockChain` should bound on `StateClient`  but it wasn't the case.

PR #7040, opened 2017-11-13
squashed ethcore-network changes which introduce error-chain

debris:
squashed #6979


debris:
@arkpar can you look at this pr? especially last commit. Without static linking it was simply not working :/

debris:
Can we merge it as it is right now? Or is there any reason why snappy shouldn't be statically linked?

PR #7044, opened 2017-11-13
Bumb stable to v1.7.9

5chdn:


5chdn:
Need to retag after the merge.

debris:
by replacing everything with sed you have modified lock file checksums :D 

There are actually only 3 file, which need to be updated
- `Cargo.toml`
- `util/Cargo.toml`
- `ethcore/Cargo-toml`
-
 https://github.com/paritytech/parity/blob/48c7c1dbb7bb1746deda371318f10a7eef760ec1/nsis/installer.nsi#L13

After that run:

```
cargo update -p ethcore-util -p ethcore
```

and lock files will be updated accordingly ;)

5chdn:
Yeah just noticed, lol.

5chdn:
ethcore/Cargo.toml is at 1.7.0 - should I leave it as is?

debris:
actually, yes

updating `ethcore/Cargo.toml` and `util/Cargo.toml` is optional

PR #7048, opened 2017-11-13
reopened 6860 - iterate over both buffered and unbuffered database entries

debris:
original pr: #6860


rphmeier:
can you `cargo update -p interleaved-ordered`? there was a bug fixed in 0.1.1.

5chdn:
Thanks @efyang

Issue #6973, opened 2017-11-02
Recommend decreasing usd_per_tx


MysticRyuujin:
Currently the default for `usd_per_tx` is `0.0025` however, `.1 GWei` and even smaller transactions are fairly common.

A standard transaction with 21000 gas @ .1 Gwei is only about $0.0006 right now... that means that a significant chunk of transactions fall below the minimum gas price as calculated by `usd_per_tx`

5chdn:
There is nothing wrong with paying a quarter-cent for a transaction by default. 

If you really desire to lower your transaction fees significantly, you can fully control this behavior with `--usd-per-tx 0.0006`

Office-Julia:
Actually we can accept, but only a well thought-through PR.
You should be sure to provide statistics that back up and proves requirement of any alteration you would like to make. 

Regards,
Julia.

MysticRyuujin:
@5chdn while I agree there is nothing wrong with paying a quarter-cent for a transaction, I guess I don't understand why Parity is defining a minimum at all, let alone one that is well below the current "Safe Low" defined by https://ethgasstation.info/ by an order of 4x

PR #7058, opened 2017-11-14
Reduce usd_per_tx

MysticRyuujin:
Currently the default minimum transaction fee is .0025 however, a huge amount of transactions are falling below this default minimum. I don't feel like it's Parity's duty to set a default minimum transaction fee, especially when that default is about 4x higher than what [ETH Gas Station](https://ethgasstation.info/) calls the "Safe Low" for a transaction.

While any number of arguments can be made for what a reasonable transaction fee should be, I don't think the place to decide that is in the default configuration...

I've submitted this pull request as a place to have that discussion 😄

---
This change is [Reviewable](https://reviewable.io/reviews/paritytech/parity/7058)


MysticRyuujin:
[clabot:check]

5chdn:
I have no objections, really. Just want to say that 0.25 US cents are already cheap. 0.01 US cents is good too. Potential blockchain spammers will use a custom gas price anyways.

folsen:
The only thing I'd be worried about with setting it too low is that in times of high pressure, a lot of transactions will get stuck and we'll get a huge influx of support requests saying "Your application isn't working and my transactions aren't going through."

25c may be too high, 1c is pretty damn low though. Do we have _any_ data at all to say "95% of the time, X would get a transaction through reliably"?

I think the default should be data-driven to as large extent as possible.

tomusdrw:
@folsen Note this setting only affect miners, not regular users of Parity Wallet. The `eth_gasPrice` RPC method is calculating a median gas price from last 100 blocks anyway.

5chdn:
"Closes" #6973 ;)

5chdn:
> 25c may be too high, 1c is pretty damn low though. 

Note, it's 0.25c and 0.01c.

> Do we have any data at all to say "95% of the time, X would get a transaction through reliably"?

https://ethgasstation.info/ says;

- `Std Cost for Transfer $0.01` (1.00c)
- `SafeLow Cost for Transfer $0.001` (0.10c)

I don't see why we should default to something 10x lower than the statistical `SafeLow Cost`. Paying a quarter cent for a transaction is perfectly reasonable.

And if any miner disagrees, they can still override it with a custom configuration.

MysticRyuujin:
What? Let's stop changing the decimal places for clarity here...

Currently the default minimum transaction fee is $0.0025
SafeLow Cost for Transfer $0.001

$0.001 < $0.0025

So I don't see why you should default to rejecting transactions that are within the safe low cost by a factor of 2.5x either...

5chdn:
Reopening for 2nd opinion.

tomusdrw:
The setting will affect network propagation (relaying) of low gas transactions which seems reasonable for me.
The miners are also affected, but they should run with custom settings anyway and set the price according to their own analysis.

That said it should be mentioned in release notes. @MysticRyuujin Would you mind fixing the tests? (don't mind the RpcList, they are not related, just the price needs to be updated).

MysticRyuujin:
@tomusdrw can you point me at which test file(s) need to be updated?

5chdn:
Now the safe low is 30c. :-p

tomusdrw:
@MysticRyuujin values here needs to be updated:
https://github.com/paritytech/parity/blob/de8d20cea9eab9ab563a7e6314a40bd5432ed26a/parity/params.rs#L229
```
to
```
Calibrated { initial_minimum: 476190464, usd_per_tx: 0.0001, }
```

**Comment by 5chdn** [comment]
Can you fix tests and conflicts?

**Comment by MysticRyuujin** [comment]
So it currently says:
```
initial_minimum: 11904761856u64.into(),
usd_per_tx: 0.0025f32,
```

But I don't know how to convert `Calibrated { initial_minimum: 476190464, usd_per_tx: 0.0001, }` into the proper values to replace the above...

rphmeier:
@MysticRyuujin yes, that would fix it

5chdn:
Fixed the build but the tests still fail

MysticRyuujin:
Should `476190474u64` be `476190464u64` ?

5chdn:
Yes sorry

Issue #7074, opened 2017-11-17
Taking a long time for executing estimate_gas of tx with infinite loop.


miyao-gmo:
When estimate_gas of tx with infinite loop was executed, RPC server returns no response for a long time.
In a following environment, it takes about 3 minutes for estimate_gas.
```
parity: v1.8.2
OS: ubuntu 16.04.1
Memory: 3G
```
```
contract A {
  function loop() {
    while (true) {}
  }
}
```
```
> var calcInfiniteLoopTime = function() {
... var start = new Date().getTime();
... A.loop.estimateGas(function(err, gas) {
..... var end = new Date().getTime();
..... console.log(end - start);
..... });
... }
> calcInfiniteLoopTime();
> 170040
```

Tx with 1,000,000,000,000 gas was executed (https://github.com/paritytech/parity/blob/master/ethcore/src/client/client.rs#L1234), so it takes a long time.
A parity node owner should decide whether estimate_gas upper limit is a block gas limit or UPPER_CEILING(1,000,000,000,000) by config.
If I skipped estimate_gas execution with 1,000,000,000,000 gas, it takes about 80 msec.

PR #7075, opened 2017-11-17
escape inifinite loop in estimte_gas

miyao-gmo:
#7074

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1231,12 +1231,12 @@ impl BlockChainClient for Client {
 	}
 
 	fn estimate_gas(&self, t: &SignedTransaction, block: BlockId) -> Result<U256, CallError> {
-		const UPPER_CEILING: u64 = 1_000_000_000_000u64;
-		let (mut upper, env_info)  = {
+		let (mut upper, max_upper, env_info)  = {
 			let mut env_info = self.env_info(block).ok_or(CallError::StatePruned)?;
-			let initial_upper = env_info.gas_limit;
-			env_info.gas_limit = UPPER_CEILING.into();
-			(initial_upper, env_info)
+			let init = env_info.gas_limit;
+			let max = init * U256::from(10);
+			env_info.gas_limit = max;
+			(init, max, env_info)
 		};
 
 		// that's just a copy of the state.
@@ -1257,9 +1257,7 @@ impl BlockChainClient for Client {
 		};
 
 		if !cond(upper)? {
-			// impossible at block gas limit - try `UPPER_CEILING` instead.
-			// TODO: consider raising limit by powers of two.
-			upper = UPPER_CEILING.into();
+			upper = max_upper;
 			if !cond(upper)? {
 				trace!(target: "estimate_gas", "estimate_gas failed with {}", upper);
 				let err = ExecutionError::Internal(format!("Requires higher than upper limit of {}", upper));


miyao-gmo:
[clabot:check]

rphmeier:
I don't think this configuration belongs in the chain spec. It's also not really an infinite loop, just one bounded at an arbitrary 1 Trillion gas. We could reasonably set the upper limit to `block_gas_limit * 100` or something like that without requiring additional configuration.

miyao-gmo:
I see.
I agree that upper limit is changed from 1 Trillion gas.
However, I think `block_gas_limit * 100` is too large value, so `block_gas_limit * 2` is enough.

rphmeier:
I think it'd be better if we had the upper limit still significantly greater than the block gas limit.

miyao-gmo:
OK.
I changed PR as you say.

ethcore/src/client/client.rs
```diff
@@ -1231,12 +1231,11 @@ impl BlockChainClient for Client {
...
 
 	fn estimate_gas(&self, t: &SignedTransaction, block: BlockId) -> Result<U256, CallError> {
-		const UPPER_CEILING: u64 = 1_000_000_000_000u64;
-		let (mut upper, env_info)  = {
+		let (upper, env_info)  = {
 			let mut env_info = self.env_info(block).ok_or(CallError::StatePruned)?;
-			let initial_upper = env_info.gas_limit;
-			env_info.gas_limit = UPPER_CEILING.into();
-			(initial_upper, env_info)
+			let upper = env_info.gas_limit * U256::from(100); // enough gas for estimate_gas
```
tomusdrw: The loggic is quite different now.

Previously:
1. Start with `block_gas_limit`
2. If that's not enough proceed with `UPPER_CEILING`.

Currently:
1. Start with `block_gas_limit * 100`

I think we should keep the starting point at `block_gas_limit`, since it will converge much faster for valid transactions and fallback to `100 * limit` only if the first fails.

miyao-gmo: I think a execution time in estimate_gas of a valid transaction with `block_gas_limit` is a same with `block_gas_limit * 100`, therefore, I changed to one phase estimate_gas from two phases.

I compared a execution time with `block_gas_limit` and `block_gas_limit * 100`.
```
> var calcValid = function() {
... var start = new Date().getTime();
... sample.validTx.estimateGas(function (err, gas) {
..... var end = new Date().getTime();
..... console.log(end - start);
..... });
... }
```

The following result is a execution time with `block_gas_limit`.
```
> calcValid()
undefined
> 75
```

The following is with `block_gas_limit * 100`.
```
> calcValid()
undefined
> 80
```
These result shows a execution time in estimate_gas of a valid transaction with `block_gas_limit` is a same with `block_gas_limit * 100`.

tomusdrw: This time will depend on complexity of the transaction (CPU, IO). You can see that there is even a special case for transfer transactions (they will return immediately, by checking the lower bound first). I'm pretty positive that `block_gas_limit` should converge ~6 times faster.

miyao-gmo: I understood.
I will change PR as you say.

By the way, I think `block_gas_limit * 100` is so large, but what do you think?

tomusdrw: It's pretty big, but afair users requested that estimate gas should work way over block gas limit. It's a significant improvement (in terms of time to fail) over what we have now, so I'm in favour of including this PR (or `block_gas_limit * 10`) for 1.9 and probably re-visiting it if we see that it still takes too much time for users.

tomusdrw:
Adding patch label, since we should also "fix" this in stable and beta.

ethcore/src/client/client.rs
miyao-gmo: OK, thank you.
I fixed.

tomusdrw:
@miyao-gmo Thank you!

PR #7082, opened 2017-11-17
Remove unused and duplicated files in js-old

jacogr:
Speed up build, size of precompiled repo & final binary size, only wallet is actually built with no duplicated packages & imports.

```
67424436 Nov 17 13:02 target/release/parity
```


debris:
can't wait to have this merged 🎉

PR #7083, opened 2017-11-17
Optimize & group dapp requests

amaurymartiny:
1. Optimized the code where a dapp receives a request to call a method:
    - Refactored to remove external variables, n^2 code etc.
    - Used better terminology (hopefully)
2. UI simplication: Group similar methods (eg `parity_newAccountFromPhrase` & `parity_newAccountFromSecret`) into the same methodGroup (eg `accountsCreate`). Approving a methodGroup approves all requested methods inside that group.
3. Add a `shell_getMethodGroups` api.


js/src/DappRequests/store.js
```diff
@@ -41,131 +38,137 @@ export default class Store {
...
+        result: null,
+        to: from,
+        token
+      },
+      '*'
+    );
+  };
+
+  @action addAppPermission = (method, appId) => {
+    this.permissions = Object.assign({}, this.permissions, {
```
tomusdrw:
```js
this.permissions = {
  ...this.permisions,
  [..]: true,
};
```
?

Is it required to copy over the permissions anyway? It might be called in a loop currently (approveGroup -> approveRequest -> addAppPermission) and it seems like a lot of overhead.

js/src/DappRequests/store.js
```diff
@@ -276,12 +268,20 @@ export default class Store {
       this.executePubsubCall(data, source);
     } else if (subId) {
       subId === '*'
-        ? this.provider.unsubscribeAll().then(v => this._methodCallbackPost(id, from, source, token)(null, v))
-        : this.provider.unsubscribe(subId).then(v => this._methodCallbackPost(id, from, source, token)(null, v));
+        ? this.provider
+          .unsubscribeAll()
+          .then(v =>
+            this._methodCallbackPost(id, from, source, token)(null, v)
```
tomusdrw: Could be extracted outside of the ternary.

js/src/DappRequests/store.js
```diff
@@ -41,131 +38,137 @@ export default class Store {
...
-  @action addTokenPermission = (method, token) => {
-    const id = `${method}:${this.tokens[token]}`;
-
-    this.permissions = Object.assign({}, this.permissions, {
-      [id]: true
-    });
-    this.savePermissions();
-  }
+    // Add a new request in this.requests
+    this.requests = {
```
tomusdrw: Why copy `this.requests` over to a new object?

tomusdrw: Oh it's because of `mobx`, right? We need to assign a new object to refresh the store? Makes sense.

js/src/DappRequests/store.js
```diff
@@ -41,131 +38,137 @@ export default class Store {
...
-  @action approveRequest = (queueId, approveAll) => {
-    const queued = this.findRequest(queueId);
+  @action approveRequestGroup = (groupId, appId) => {
+    // Get methods of this requestGroup
+    const { methods } = methodGroups[groupId];
 
-    if (approveAll) {
-      const { request: { data: { method, token, params } } } = queued;
+    methods.forEach(method => {
+      if (!this.requests[`${method}:${appId}`]) {
```
tomusdrw: `methods.filter`?

js/src/DappRequests/store.js
```diff
@@ -41,131 +38,137 @@ export default class Store {
...
-  }
+      this.approveRequest(method, appId);
+    });
+  };
 
-  @action rejectRequest = (queueId) => {
-    const { request: { data } } = this.findRequest(queueId);
-    const source = this.sources[queueId];
+  @action rejectRequest = (method, appId) => {
+    const { source, data } = this.requests[`${method}:${appId}`];
```
tomusdrw: `${method}:${appId}` is scattered around in many places, maybe put it in a helper method?

js/src/DappRequests/store.js
amaurymartiny: I don't quite understand this. I'm copying this.permissions for the same reason as this.requests above, for mobx to update the store.

js/src/DappRequests/store.js
amaurymartiny: Exactly. Since we're dynamically adding new properties to the observable object, I think that's the best way to update the store

js/src/DappRequests/store.js
amaurymartiny: Fixed

js/src/DappRequests/store.js
amaurymartiny: Fixed

js/src/DappRequests/store.js
amaurymartiny: Fixed

js/src/DappRequests/store.js
tomusdrw: Yup, sorry. I wrote the comments in opposite order :) Still replacing `Object.assign` with more idiomatic `...` pattern would look nice.

amaurymartiny: OK, that's fixed in the last commit

PR #7097, opened 2017-11-20
WASM parse payload from panics

pepyakin:
This change implements a new format of a panic payload (see https://github.com/paritytech/pwasm-std/pull/42).
So if a contract has been compiled with a `pwasm-std/panic_with_msg` feature then it's panic messages will be printed (by default they are stripped).

Example of a panic message:
```
arithmetic operation overflow, /Users/pepyakin/.cargo/registry/src/github.com-1ecc6299db9ec823/bigint-4.2.0/src/uint.rs:1227:0
```

### DIFF
--- a/ethcore/wasm/src/runtime.rs
+++ b/ethcore/wasm/src/runtime.rs
@@ -21,6 +21,7 @@ use std::sync::Arc;
 use byteorder::{LittleEndian, ByteOrder};
 
 use vm;
+use panic_payload;
 use parity_wasm::interpreter;
 use wasm_utils::rules;
 use bigint::prelude::U256;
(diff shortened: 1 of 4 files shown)


ethcore/wasm/src/runtime.rs
```diff
@@ -626,16 +627,30 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
-		let msg_ptr = context.value_stack.pop_as::<i32>()? as u32;
-
-		let msg = String::from_utf8(self.memory.get(msg_ptr, msg_len as usize)?)
-			.map_err(|_| UserTrap::BadUtf8)?;
-
-		trace!(target: "wasm", "Contract custom panic message: {}", msg);
-
-		Err(UserTrap::Panic(msg).into())
-	}
+        let payload_len = context.value_stack.pop_as::<i32>()? as u32;
```
debris: use tabs instead of spaces

pepyakin: Thanks! It happened after I copied this chunk after rustfmt : (

PR #7100, opened 2017-11-20
Local dapp development URL

jacogr:


amaurymartiny:
Working for me 👍  managed to run a dapp served by it's own dev server.

tomusdrw:
@paritytech/ci Build status is broken on this one, seems that the rust compilation/tests did not run at all on this PR. The modified files detection didn't work all that well, it should take all changes on the branch not only last commit. Created a separate issue: https://github.com/paritytech/parity/issues/7130

PR #7101, opened 2017-11-20
SecretStore: Kovan integration initial version

svyatonik:
Based on https://github.com/paritytech/contracts/pull/91

What is in this PR:
1) I needed ability to submit transactions to contract && since I already had a native_generator-generated code for this, contract, I've also added ability to encode method arguments to generated code
2) there are now 2 possible API endpoints for KeyServer: http && service contract. Both are optional (configuration methods are different, though)
3) the only API method available via service contract is currently 'GenerateServerKey'. It generates distributed KeyPair, which can be used to sign messages later. It **is not** intended for encryption (encryption-related API methods are currently questionable && not a part of this PR)
4) service contract listener:
4.1) listens for `ServerKeyRequested` event && checks if this request should be processed by this KeyServer. It is because of SS specific - there must be a 'master node' in each session (including generation session) => we can't just start session on all nodes at once. Basically, all possible `H256::max_value()` server_key_id values are evenly distributed among all valid KeyServers. Note: it doesn't matter who actually is a master node - it is just to beat SS protocol caveats.
4.2) when appropriate key is requested, node starts key generation. If everything is OK, key pair is generated && all nodes know a Public of this KeyPair. Every KeyServer signs this Public with its own key && sends tx back to the contract. When enough (threshold + 1) conformations are received, `ServerKeyGenerated` event is fired by the contract.
4.3) if generation fails, there's no much we can do. Possible options I've considered:
4.3.1) admit that we can't generate the key && send fee back to the requester;
4.3.2) admit that we can't generate the key, but do not return fee (we actually have made some actions);
4.3.3) retry until success (optimistically)
Of these options I've selected the last one, because there's actually no error can occur because of requester - either there are some problems with SS configuration/nodes, or we're trying to start session during maintenance interval (like when `ServersSetChange` session is active).
4.4) retry occurs every N (30) blocks && only M (1) 2-nd time failures are allowed during this retry. If M failures has occured => we stop retrying until next 30 blocks are mined.
5) if key K1 is already generated && someone asks to generate key with the same K1 id, SS should just publish existing key 

What is not in this PR (but still about SS+Kovan integration):
1) other SS APIs (need to discuss if they're suitable for usage in public blockchain at all)
2) when validators set changes, `ServersSetChange` session must be automatically started (next step)
3) integration with validators set contract - currently validators set is fixed in `service_validators_fixed.sol` contract. Have failed to configure PoA, based on validators set contract. This shouldn't affect Rust code, though (as I see it now)
4) proper testing - currently having issues with UI on master => have tested it on my local 1.8 branch, then cherry-picked back to master


secret_store/src/key_server_cluster/cluster_sessions.rs
```diff
@@ -342,12 +294,51 @@ impl<S, SC, D> ClusterSessionsContainer<S, SC, D> where S: ClusterSession, SC: C
 			queue: VecDeque::new(),
 		};
 		sessions.insert(session_id, queued_session);
+
+		// notify listeners
+		let mut listeners = self.listeners.lock();
+		let mut listener_index = 0;
+		loop {
+			if listener_index >= listeners.len() {
```
tomusdrw: Maybe convert to `while listener_index < listeners.len() { }`

secret_store/src/key_server_cluster/cluster_sessions.rs
```diff
@@ -342,12 +294,51 @@ impl<S, SC, D> ClusterSessionsContainer<S, SC, D> where S: ClusterSession, SC: C
...
 		Ok(session)
 	}
 
 	pub fn remove(&self, session_id: &S::Id) {
-		if self.sessions.write().remove(session_id).is_some() {
+		if let Some(session) = self.sessions.write().remove(session_id) {
 			self.container_state.lock().on_session_completed();
+
+			// notify listeners
+			let mut listeners = self.listeners.lock();
```
tomusdrw: Could be abstracted as:
```
self.notify_listeners(|listener| listener.on_session_removed(session.session.clone()));
```

secret_store/src/key_server_set.rs
```diff
@@ -110,7 +114,12 @@ impl CachedContract {
 	}
 
 	pub fn update(&mut self, enacted: Vec<H256>, retracted: Vec<H256>) {
-		if let Some(client) = self.client.upgrade() {
+		if let (Some(client), Some(sync)) = (self.client.upgrade(), self.sync.upgrade()) {
+			// do not update initial server set until fully synchronized
+			if sync.status().is_syncing(client.queue_info()) {
```
tomusdrw: Same as perviously, might be good to check full security.

secret_store/src/lib.rs
```diff
@@ -53,33 +54,51 @@ mod types;
...
-
+pub fn start(client: Arc<Client>, sync: Arc<SyncProvider>, self_key_pair: Arc<NodeKeyPair>, config: ServiceConfiguration) -> Result<Box<KeyServer>, Error> {
 	let acl_storage: Arc<acl_storage::AclStorage> = if config.acl_check_enabled {
-			acl_storage::OnChainAclStorage::new(&client)
+			acl_storage::OnChainAclStorage::new(&client, &sync)
 		} else {
 			Arc::new(acl_storage::DummyAclStorage::default())
 		};
-	let key_server_set = key_server_set::OnChainKeyServerSet::new(&client, config.cluster_config.nodes.clone())?;
+	let key_server_set = key_server_set::OnChainKeyServerSet::new(&client, &sync, config.cluster_config.nodes.clone())?; // TODO: return empty set until fully synced
```
tomusdrw: Is the TODO still valid? I though we are checking sync status before attempting any actions?

secret_store/src/listener/http_listener.rs
```diff
@@ -68,83 +68,44 @@ enum Request {
...
-		let http_server = listener_address
-			.map(|listener_address| format!("{}:{}", listener_address.address, listener_address.port))
-			.map(|listener_address| HttpServer::http(&listener_address).expect("cannot start HttpServer"))
-			.map(|http_server| http_server.handle(KeyServerHttpHandler {
+		let listener_address = format!("{}:{}", listener_address.address, listener_address.port);
+		let http_server = HttpServer::http(&listener_address).expect("cannot start HttpServer");
+		let http_server = http_server.handle(KeyServerHttpHandler {
 				handler: shared_handler.clone(),
-			}).expect("cannot start HttpServer"));
+			}).expect("cannot start HttpServer");
```
tomusdrw: Indentation.

secret_store/src/listener/http_listener.rs
```diff
@@ -68,83 +68,44 @@ enum Request {
...
 		let shared_handler = Arc::new(KeyServerSharedHttpHandler {
 			key_server: key_server,
 		});
 
-		let http_server = listener_address
-			.map(|listener_address| format!("{}:{}", listener_address.address, listener_address.port))
-			.map(|listener_address| HttpServer::http(&listener_address).expect("cannot start HttpServer"))
-			.map(|http_server| http_server.handle(KeyServerHttpHandler {
+		let listener_address = format!("{}:{}", listener_address.address, listener_address.port);
+		let http_server = HttpServer::http(&listener_address).expect("cannot start HttpServer");
```
tomusdrw: Shouldn't this be converted to an error?

secret_store/src/listener/service_contract.rs
```diff
@@ -0,0 +1,314 @@
...
+use bigint::hash::H256;
+use bigint::prelude::U256;
+use listener::service_contract_listener::ServiceTask;
+use {ServerKeyId, NodeKeyPair, ContractAddress};
+
+/// Name of the SecretStore contract in the registry.
+const SERVICE_CONTRACT_REGISTRY_NAME: &'static str = "secretstore_service";
+
+/// Key server has been added to the set.
+const SERVER_KEY_REQUESTED_EVENT_NAME: &'static [u8] = &*b"ServerKeyRequested(bytes32,uint256)";
```
tomusdrw: For future: it should be in generated code, might be good to migrate to `ethabi_derive` some time soon.

secret_store/src/listener/service_contract.rs
```diff
@@ -0,0 +1,314 @@
...
+	/// Current request index.
+	index: U256,
+	/// Requests length.
+	length: U256,
+}
+
+impl OnChainServiceContract {
+	/// Create new on-chain service contract.
+	pub fn new(client: &Arc<Client>, sync: &Arc<SyncProvider>, address: ContractAddress, self_key_pair: Arc<NodeKeyPair>) -> Self {
+		let contract_addr = match &address {
```
tomusdrw: No need to prepend with `&` if you use `ref` in Address it won't be moved here.

secret_store/src/listener/service_contract.rs
```diff
@@ -0,0 +1,314 @@
...
+		let signed_server_key: Signature = signed_server_key.into_electrum().into();
+		let transaction_data = contract.encode_server_key_generated_input(server_key_id.clone(),
+			server_key.to_vec(),
+			signed_server_key.v(),
+			signed_server_key.r().into(),
+			signed_server_key.s().into()
+		)?;
+
+		// send transaction
+		if contract.address != Default::default() {
```
tomusdrw: Already tested in L204, and cannot change since we still hold the lock.

secret_store/src/listener/service_contract.rs
```diff
@@ -0,0 +1,314 @@
...
+	fn next(&mut self) -> Option<(bool, ServiceTask)> {
+		if self.index >= self.length {
+			return None;
+		}
+
+		let index = self.index.clone();
+		self.index = self.index + 1.into();
+
+		let self_address = public_to_address(self.self_key_pair.public());
+		let do_call = |a, d| future::done(self.client.call_contract(BlockId::Latest, a, d));
```
tomusdrw: Note that you are using `Latest` here and the state may change in the meantime. Are we sure that the contract content does not change (including `length`)?

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+					}
+				},
+			}
+		}
+
+		Ok(processed_requests)
+	}
+
+	/// Generate server key.
+	fn generate_server_key(data: &Arc<ServiceContractListenerData>, server_key_id: &ServerKeyId, threshold: &H256) -> Result<Public, String> {
```
tomusdrw: why `H256` for threshold? You could use `usize`, no?

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+		let threshold_num = threshold.low_u64();
+		if threshold != &threshold_num.into() || threshold_num >= ::std::usize::MAX as u64 {
+			return Err(format!("invalid threshold {:?}", threshold));
+		}
+
+		// key server expects signed server_key_id in server_key_generation procedure
+		// only signer could store document key for this server key later
+		// => this API (server key generation) is not suitable for usage in encryption via contract endpoint
+		let author_key = Random.generate().map_err(|e| format!("{}", e))?;
```
tomusdrw: `Random.generate()` is infallible.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+
+/// Retry-related data.
+#[derive(Default)]
+struct ServiceContractRetryData {
+	/// Server keys, which we have generated (or tried to generate) since last retry moment.
+	pub generated_keys: HashSet<ServerKeyId>,
+}
+
+/// Service tasks queue.
+struct TasksQueue {
```
tomusdrw: Why not `std::sync::mpsc::channel()` or `sync_channel(1024)` to prevent creating unbounded queue?

tomusdrw: Ok, I see that you are pushing multiple tasks at once, this structure is probably better since it won't be doing unecessary allocations.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+			if is_confirmed {
+				continue;
+			}
+
+			let request_result = match task {
+				ServiceTask::GenerateServerKey(server_key_id, threshold) => {
+					// only process request, which haven't been processed recently
+					// there could be a lag when we've just generated server key && retrying on the same block
+					// (or before our tx is mined) - state is not updated yet
+					if retry_data.generated_keys.contains(&server_key_id){
```
tomusdrw: missing space between `)` and `{`

secret_store/src/listener/service_contract.rs
tomusdrw: After reviewing the rest of the code I think we should wait for some confirmations - the latest block can end up on a fork (even if we trust authorities) and the transaction may be included a bit later, but the state can be different.

I think for security we should fetch `latest - 3` and refer to it using hash, not number.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+		}
+
+		let reason = "enacted.len() != 0; qed";
+		self.process_service_contract_events(
+			enacted.first().expect(reason).clone(),
+			enacted.last().expect(reason).clone());
+
+		// schedule retry if received enough blocks since last retry
+		// it maybe inaccurate when switching syncing/synced states, but that's ok
+		if self.data.last_retry.fetch_add(enacted_len, Ordering::Relaxed) >= RETRY_INTEVAL_BLOCKS {
```
tomusdrw: some blocks might be retracted at the same time, so the progression might be a bit smaller than assumed here. That said it probably doesn't matter that much here.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+use key_server_cluster::generation_session::SessionImpl as GenerationSession;
+use key_storage::KeyStorage;
+use listener::service_contract::ServiceContract;
+use {ServerKeyId, NodeKeyPair, KeyServer};
+
+/// Retry interval (in blocks). Every RETRY_INTEVAL_BLOCKS blocks each KeyServer reads pending requests from
+/// service contract && tries to re-execute. The reason to have this mechanism is primarily because keys
+/// servers set change takes a lot of time + there could be some races, when blocks are coming to different
+/// KS at different times. This isn't intended to fix && respond to general session errors!
+const RETRY_INTEVAL_BLOCKS: usize = 30;
```
tomusdrw: typo

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+	}
+}
+
+impl ClusterSessionsListener<GenerationSession> for ServiceContractListener {
+	fn on_session_removed(&self, session: Arc<GenerationSession>) {
+		// only publish when the session is started by another node
+		// when it is started by this node, it is published from process_service_task
+		if !is_processed_by_this_key_server(&*self.data.key_server_set, &*self.data.self_key_pair, &session.id()) {
+			// by this time sesion must already be completed - either successfully, or not
+			debug_assert!(session.is_finished());
```
tomusdrw: I would turn it into `assert`, since we rarely run node in debug mode.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+
+			// ignore result - the only thing that we can do is to log the error
+			let _ = session.wait(Some(Default::default()))
```
tomusdrw: Might be clearer to do:
```
match session.wait().map_err(..).and_then(..) {
   Ok(_) => trace!
   Err(err) => warn!
```
then.

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+				.map_err(|error| {
+					warn!(target: "secretstore", "{}: failed to process GenerateServerKey({}) request with: {}",
+						self.data.self_key_pair.public(), session.id(), error);
+					error
+				});
+		}
+	}
+}
+
+impl TasksQueue {
```
tomusdrw: That's a pretty elegant structure, consider putting it to a separate module and making it generic?

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -0,0 +1,699 @@
...
+		return false;
+	}
+	let this_server_index = match servers.keys().enumerate().find(|&(_, s)| s == self_key_pair.public()) {
+		Some((index, _)) => index,
+		None => return false,
+	};
+
+	let server_key_id_value: U256 = server_key_id.into();
+	let range_interval = U256::max_value() / total_servers_count.into();
+	let range_begin = (range_interval + 1.into()) * this_server_index.into();
```
tomusdrw: Might panic in case there is only one server (overflow).

secret_store/src/key_server_cluster/admin_sessions/share_add_session.rs
```diff
@@ -430,7 +427,9 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 		}
 
 		// we only expect this message once
-		if data.key_share_threshold.is_some() || data.key_share_author.is_some() || data.key_share_common_point.is_some() || data.key_share_encrypted_point.is_some() {
+		if data.key_share_threshold.is_some() || data.key_share_author.is_some() ||
```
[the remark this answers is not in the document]
svyatonik: Since all these fields are set at once, I've extracted these to the separate structure. So it's now a single check only

secret_store/src/acl_storage.rs
```diff
@@ -105,19 +109,26 @@ impl CachedContract {
...
-						.upgrade()
-						.ok_or("Calling contract without client".into())
-						.and_then(|c| c.call_contract(BlockId::Latest, a, d)));
-				contract.check_permissions(do_call, address, document.clone())
-					.map_err(|err| Error::Internal(err))
-					.wait()
+		match (self.client.upgrade(), self.sync.upgrade()) {
+			(Some(client), Some(sync)) => {
+				// we can not tell if access to document is allowed until fully synchronized
+				if sync.status().is_syncing(client.queue_info()) {
```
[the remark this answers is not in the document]
svyatonik: Found it - `security_level().is_full()`. Thanks - didn't knew about this!

secret_store/src/key_server_cluster/client_sessions/decryption_session.rs
```diff
@@ -206,6 +200,18 @@ impl SessionImpl {
...
 	}
 
+	/// Wait for session completion.
+	pub fn wait(&self) -> Result<EncryptedDocumentKeyShadow, Error> {
+		let mut data = self.data.lock();
+		if !data.result.is_some() {
+			self.core.completed.wait(&mut data);
+		}
+
+		data.result.as_ref()
```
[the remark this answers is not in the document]
svyatonik: :) That's actually a toughest one :) I guess I can make it generic but extracting this single method will result in even more LOCs than current implementation :( There's need for something bigger - like adding something like generic 'meta session'.

I'm trying to abstract some entities from time to time, but this activity is often suppressed by add-new-feature-activities-asap (I guess it's me who's adding asap suffix here, but anyway...). So I kindly ask you to ignore this repetitions - I know there are problems && I'm going to fix these once everything stabilizes (at least all features are implemented).

Please respond here, if you insist.

secret_store/src/lib.rs
svyatonik: Actually I've already removed this in next PR (which is on top of this one) :) But nvm - also removed here

secret_store/src/listener/http_listener.rs
svyatonik: Actually I've got this pattern from other parts of our code - there's a lot of expects like that when something crucial is **failed to initialize**. Like this:
https://github.com/paritytech/parity/blob/master/util/network/src/discovery.rs#L112
https://github.com/paritytech/parity/blob/master/util/reactor/src/lib.rs#L43
...

I do not know - what is better, though :) If you say - `Err` is better, please respond - I'll add the handling

secret_store/src/listener/service_contract.rs
svyatonik: Right. Initial idea behind this listener was to react as soon as possible. The only bad thing that could happen is that SS will generate the key for free (without fee paid by requester && without publishing it on the contract). So I considered this as a possible annoyance, not an issue.

This is completely different from next PR (auto-migration), where reaction to 'forked' changes could potentially lead to data loss. That's why there were no any confirmations requirements && in next PR they're stronger than 3 confirmations.

But I guess you're right here and it is an actual issue. So I've updated PR:
1) added required confirmations to `PendingRequestsIterator`
2) since we wait for several blocks anyway, I've completely removed reaction to contract' events - I suppose adding support for this will increase complexity of the code significantly without any actual output (we won't react to events immediately anyway)

secret_store/src/listener/service_contract_listener.rs
svyatonik: Actually the main reason behind this queue is an ability to shutdown processing thread right before queued tasks are processed. I mean - if there are 1000 queued tasks, then we could not push `Shutdown` task to the front of `mpsc::queue` && shutdown right after current task execution will be completed. Am I right that it is impossible with `mpsc::queue` (I could only think of dropping `transmitter` => this could lead to an error in next `receiver.wait()`, but this seems like a hack to me)?

secret_store/src/listener/http_listener.rs
tomusdrw: The second example is only spawning an event loop (afair it fails only in some extremely rare, low-level cases).
The first one seems similar, I'm opting for an error, since for instance the port can be taken already and that imho shouldn't lead to a crash but rather a more friendly message.

secret_store/src/listener/service_contract_listener.rs
svyatonik: That's how `usize`/`u64` is read from the contract. I'm converting this to required type later (with appropriate error message if conversion has failed).

secret_store/src/listener/service_contract_listener.rs
svyatonik: Right - there's no need to calc exact number of rectracted/enacted blocks here - just a rough estimate

secret_store/src/key_server_cluster/client_sessions/decryption_session.rs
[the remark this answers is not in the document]
tomusdrw: All good, not insisting :) But maybe some kind of a helper function would help here?
```
wait(&mut self.core.completed, self.data.lock())
```

secret_store/src/listener/service_contract_listener.rs
tomusdrw: Dropping the transmitter is the only way, but afair `receiver.wait()` cannot fail in any other case, so I think this is the way it was intended to be used.

secret_store/src/listener/service_contract_listener.rs
svyatonik: Do you mean moving this to the local module of the SecretStore? Or somewhere in utils? I'm for the first scenario, if you confirm

secret_store/src/listener/service_contract_listener.rs
svyatonik: Thanks, good catch!!!

svyatonik:
Thanks for such a detailed review, @tomusdrw !!! That's actually what I'm missing when working on SecretStore!
Tried to address all the issues && responded where appropriate.

svyatonik:
I see some responses came, while I was adding comments - taking a time to fix these

secret_store/src/listener/service_contract_listener.rs
tomusdrw: A local module is fine for now, it will clean up this one and it will be easier to re-use in the future.

secret_store/src/listener/service_contract.rs
svyatonik: Me-of-today disagree with me-of-yesterday - think that both retry && reactions to contract events are important. And that is why:
1) retry was designed to deal with stuck requests - e.g. which have been received when SS wasn't fully-connected. That's why it executes rarely && stops processing stuck requests just after first `MAX_FAILED_RETRY_REQUESTS` fail (currently 1)
2) reacting to contract events was designed to respond to requests as soon as possible (in my setup key was generated within <1s && published in the next block). Now this guarantee has shifted (we wait for 3 blocks before processing request), but we still want to process:
2.1) all incoming requests
2.2) requests that have exact age of 3 blocks (omitting reorg case here) - i.e. as soon as possible, given 3 confirmations required

So I've reverted removal and instead of reacting to passed `enacted` blocks (in `read_logs`) I keep track of last-block-from-which-events-were-read-from && use the `Client::tree_route` to build route from this block to the latest block with 3 confirmations.

secret_store/src/listener/service_contract_listener.rs
svyatonik: Anyway - I've left the `TasksQueue` class, because it is useful in tests.

secret_store/src/listener/service_contract_listener.rs
svyatonik: Yes, but it returns a `Result` :) Let's fix the origin of the problem - i.e. `Random.generate()`. Not in this PR, of course. I'll add this to my local txt-file based issue tracker.

secret_store/src/listener/service_contract.rs
```diff
@@ -233,6 +298,13 @@ impl Iterator for PendingRequestsIterator {
 	}
 }
 
+/// Get hash of the last block with at least n confirmations.
+fn get_confirmed_block_hash(client: &Client, confirmations: u64) -> Option<H256> {
+	client.block_number(BlockId::Latest)
+		.and_then(|b| b.checked_sub(confirmations))
```
tomusdrw: It's an edge case, but I would say that for `b < confirmations` we should just return `0`

secret_store/src/listener/service_contract_listener.rs
```diff
@@ -301,16 +319,21 @@ impl Drop for ServiceContractListener {
 
 impl ChainNotify for ServiceContractListener {
 	fn new_blocks(&self, _imported: Vec<H256>, _invalid: Vec<H256>, enacted: Vec<H256>, _retracted: Vec<H256>, _sealed: Vec<H256>, _proposed: Vec<Bytes>, _duration: u64) {
-		self.data.contract.update();
-		if !self.data.contract.is_actual() {
+		let enacted_len = enacted.len();
+		if enacted_len == 0 {
```
tomusdrw: `enacted.is_empty()`?
EDIT: just noticed you are using `enacted_len` later on, so I guess it's fine.

secret_store/src/listener/service_contract.rs
svyatonik: Fixed. Just curious - the only reason is to process requests from block#0 asap (instead of waiting for 3 blocks)? Or is there some other problem with previous pattern?

tomusdrw: Yeah. I think that getting `None` from this method would indicate some errornous state (non existing historic block?) and it wouldn't be exactyl the same semantics for block `0`.

PR #7125, opened 2017-11-22
Update packages, pull in compiled-only repos

jacogr:
... as a bonus, re-worked dapp methods permissions manager is available

cc @amaurymartiny


js/package.json
```diff
@@ -44,6 +44,19 @@
     "test:coverage": "cross-env NODE_ENV=test istanbul cover _mocha -- --compilers ejs:ejsify 'src/**/*.spec.js'"
   },
   "devDependencies": {
+    "@parity/dapp-console": "paritytech/dapp-console",
```
tomusdrw: Was this change intended? Why is it oly devDependency?

jacogr: This is to make the distinction between -

(a) packages referenced in the actual code via `import ... from '...'` (dependency)
(b) packages only referenced in the webpack build (devDependency) - the dapps (mostly) fall into this category

Issue #7112, opened 2017-11-21
Improve usability of Connection Dialogue


bnvk:
Currently, the `Connection Dialogue` is a bit technical and not as user friendly as it could be. For instance when no websockets token exists, the message reads:

```
Unable to make a connection to the Parity Secure API...
```

A more user friendly message would be something like:

```
Unable to connect to the Parity app...
```

There are multiple Connection Dialogues throughout the application, so I will send more detailed specific suggestions / fixes in individual pull-requests.

5chdn:
Thanks, pull requests are much appreciated :)

Tbaut:
Moving to a new repo, we will not migrate low prio issues. Please reopen it in the new repo if you believe it should be addressed: https://github.com/parity-js/shell/issues/new

PR #7134, opened 2017-11-24
usability improvements to security token Dialog #7112

bnvk:
**Old Dialog**

screenshot-2017-11-21 parity

**New Dialog**

screenshot-2017-11-24 parity

The changes I've made include

- Simpler, more human friendly text
- Cleaner handling of text, inputs, instructions, etc...
- Changed icon to better reflect context

### DIFF
--- a/js/src/Connection/connection.css
+++ b/js/src/Connection/connection.css
@@ -53,9 +53,9 @@
 }
 
 .form {
-  margin-top: 0.75em;
+  margin-top: 1em;
   padding: 0 4em;
-  text-align: left;
+  text-align: center;
 }
 
 .timestamp {
@@ -71,14 +71,15 @@
 }
 
 .icons {
-  padding-top: 1.5em;
+  margin-top: 2em;
+  margin-bottom: 0px;
 }
 
 .icon,
 .iconSmall {
   display: inline-block;
-  padding: 1em;
-  padding-bottom: 1em;
+  margin-top: 2em;
+  padding: 0 1.5em;
   vertical-align: middle;
 }
 
@@ -92,10 +93,13 @@
 }
 
 .console {
+  width: 80%;
   font-family: 'Roboto Mono', monospace;
-  background: rgba(0, 0, 0, 0.25);
+  background: rgba(0, 0, 0, 0.65);
   font-size: 16px;
-  padding: 0 0.25em;
+  padding: 0.5em 0.25em;
+  margin: 1em auto;
+  border-radius: 3px;
 }
 
 @keyframes pulse {
@@ -123,10 +127,11 @@
   color: rgb(208, 208, 208);
 }
 
-.formInput input {
+.formInput input[type=text] {
   background: rgba(0, 0, 0, 0.25) !important;
+  text-align: center;
 }
 
-.formInput input:focus {
+.formInput input[type=text]:focus {
   background: rgba(0, 0, 0, 0.25) !important;
 }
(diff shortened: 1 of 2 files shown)


tomusdrw:
Could we add an information that if you are connecting to a remote host the time needs to be synchronized? I remember that many issues where actually caused by non-synchronized time.

I would add this additional info when user provides the token, but we still can't connect.

jacogr:
@tomusdrw We have the "big scary red box" on top of the screen screen now when the time is out - cannot be closed. Do we still need it one here?

jacogr:
@5chdn Not back-portable without re-doing, structure is completely different between old and new. Should stick to bugs in 1.8.

tomusdrw:
It's different though, this is displayed when node time is off (which means that it won't synchronize with the network).
The sync i'm talking about is between the computer running the browser and the machine running a node (in case you are connecting to a remote host) - if the time is off between them you won't even be able to connect (the token will be constantly rejected even if valid).

bnvk:
> Could we add an information that if you are connecting to a remote host the time needs to be synchronized? I remember that many issues where actually caused by non-synchronized time.

It is my understanding that this **remote host** scenario @tomusdrw describes is a minority of use cases, and the **same machine / localhost** configuration *is* the vast majority of users. The previous message about this was EXTREMELY confusing to an end user who would be the later (majority) type of install. 

I suggest use some lower level config value (or host detection) that gets exposed to the interface and ONLY shows such a time clock message for **remote host** instances.

tomusdrw:
@bnvk Completely agree that the dialog should be improved, not suggesting to put more text on the current dialog at all.
I also agree that remote wallet configuration is not main use case for this (although regular users should never see this dialog anyway, cause the tray applications are always supplying the token).
Unfortunatelly we cannot easily detect the remote wallet configuration - people may use ssh tunneling and in such cases they are still connecting to `localhost`.

What I suggest is that when user enters the token, but we still cannot connect instead of just displaying a misleading "Invalid token" error message we should also give them a hint that the token might be valid and if they are running a remote wallet scenario they should also check time synchronization.

bnvk:
> people may use ssh tunneling and in such cases they are still connecting to localhost.

That's a pretty edge case user, no?  IMHO, having info that pertains to such a small amount of users  is best left to documentation or a wiki. 

>  when user enters the token, but we still cannot connect instead

Ok, but as @jacogr points out "We have the "big scary red box" on top of the screen screen now " is this the same case? Do you have any read on how to best determine this in the codebase?

tomusdrw:
> That's a pretty edge case user, no? 
That's the recommended way of doing remote wallet, I don't know exactly how many people does that though, just judging from the amount of requests we got for this.

> is this the same case?
It's a different case, when the time of the node is not synchronized.

As mentioned earlier, I don't want to significantly re-design the dialog, I'm really happy that you changed it.
Just asking to re-phrase the error message here:
https://github.com/paritytech/parity/blob/d105bc272eb8d91b67b01b0fba8e8244b08d0f12/js/src/Connection/connection.js#L116

to something like "The token is invalid or the time between browser and remote node is not synchronized"

bnvk:
IIUC, the line you highlighted `defaultMessage='invalid signer token' ` is an error that gets kicked back when clocks are out of sync. Making that show a hidden message would make sense. Does that sound ok to you @tomusdrw

tomusdrw:
Yeah, the message is rendered when you can't connect to the node after typing the token and it may indicate two possibilities:
1. The token you entered is invalid
2. The clocks are not in sync between client/host machine

PR #7142, opened 2017-11-25
WASM test runner utility

NikVolf:
Allows to run json-declarative tests (see sample json) using parity wasm runtime


ethcore/wasm/run/src/runner.rs
```diff
@@ -0,0 +1,204 @@
...
+				let used_gas = fixture.gas_limit.unwrap_or(0) - gas_left.low_u64();
+				if gas != used_gas {
+					fails.push(Fail::UsedGas { expected: gas, actual: used_gas });
+				}
+			},
+			Assert::HasCall(ref locator) => {
+				let mut found = false;
+
+				for fake_call in ext.calls.iter() {
+					let mut mat4 = true;
```
pepyakin: What is mat4?

NikVolf: `match` is a reserved word.. :)

pepyakin: Ah, I see, thanks! 

Maybe it will be a little more legible to use `match_found`? 
Also, AFAIK, in Rust there is a common idiom to use underscore suffix in such cases, i.e. `match_`

Issue #7061, opened 2017-11-15
Parity no longer compatible with expanse


wariner84:
Parity (1.8.2) is no longer compatible with expanse, since it is working on the wrong chain. Expanse forked at block 800’000.

5chdn:
Is expanse no longer maintaining parity compatibility?

We should consider removing it then.

rphmeier:
@5chdn is right. we only maintain specs for most chains up to the point that the developers for those chains are willing to maintain them

5chdn:
Found the changes in here: https://github.com/expanse-org/go-expanse/releases/tag/v1.7.2

> # Expanse Changes 1.7.2
> 
> - Upgrade Objectives:
> - Upgrade EXP to the latest ETH:Byzantium protocol.
> - Make EXP transactions faster
> - Decrease uncle inflation
> - Maintain Block Reward 8/min
> 
> # Byzantium Objectives:
> 
> - Addition of ‘REVERT’ opcode, which permits error handling without consuming all gas (EIP 140)
> - Transaction receipts now include a status field to indicate success or failure EIP 658)
> - Elliptic curve addition and scalar multiplication on alt_bn128 (EIP 196) and pairing checks (EIP 197), permitting
> - ZK-Snarks and other cryptographic mathemagic™
> - Support for big integer modular exponentiation (EIP 198), enabling RSA signature verification and other cryptographic applications
> - Support for variable length return values (EIP 211)
> - Addition of the ‘STATICCALL’ opcode, permitting non-state-changing calls to other contracts (EIP 214)
> - Changes to the difficulty adjustment formula to take uncles into account (EIP 100)
> 
> # Important Information:
> 
> - Fork Block Number: 800,000
> - New Block Target: 30 Seconds
> - New Block Reward: 4 EXP

Friendly ping to EXP dev @chrisfranko - could you comment on this? Is EXP support in Parity still desired? :)

chrisfranko:
Yes we would still love parity support. When making the fork I didnt see where to make the corresponding changes to parity to submit the pr. If you could kindly point me to where i can make the Byzantium changes that would be awesome!

5chdn:
I think most of the changes can be applied by adjusting the chain spec in

https://github.com/paritytech/parity/blob/master/ethcore/res/ethereum/expanse.json

I'm not quite understanding the complexity of the changes involved. Did you just enable Byzantium? Is the block target time and reward related to EIP649?

[...]

PR #7146, opened 2017-11-27
Byzantium Update for Expanse

chrisfranko:
Here the changes go. Hope I didnt miss anything.

### DIFF
--- a/ethcore/src/ethereum/ethash.rs
+++ b/ethcore/src/ethereum/ethash.rs
@@ -90,6 +90,10 @@ pub struct EthashParams {
 	pub eip649_delay: u64,
 	/// EIP-649 base reward.
 	pub eip649_reward: Option<U256>,
+	/// EXPIP-2 duration limit
+	pub expip2_duration_limit: u64,
+	/// EXPIP-2 block height
+	pub expip2_transition: u64,
 }
 
 impl From<ethjson::spec::EthashParams> for EthashParams {
@@ -355,7 +359,14 @@ impl Ethash {
 			self.ethash_params.difficulty_bound_divisor
 		};
 
-		let duration_limit = self.ethash_params.duration_limit;
+		let expip2_hardfork = header.number() > self.ethash_params.expip2_transition;
+
+		let duration_limit = if expip2_hardfork {
+			self.ethash_params.expip2_duration_limit
+		}else{
+			self.ethash_params.duration_limit
+		}
+
 		let frontier_limit = self.ethash_params.homestead_transition;
 
 		let mut target = if header.number() < frontier_limit {
@@ -520,7 +531,7 @@ mod tests {
 		let (eras, reward) = ecip1017_eras_block_reward(eras_rounds, start_reward, block_number);
 		assert_eq!(15, eras);
 		assert_eq!(U256::from_str("271000000000000").unwrap(), reward);
-		
+
 		let block_number = 250000000;
 		let (eras, reward) = ecip1017_eras_block_reward(eras_rounds, start_reward, block_number);
 		assert_eq!(49, eras);
--- a/ethcore/res/ethereum/expanse.json
+++ b/ethcore/res/ethereum/expanse.json
@@ -19,7 +19,9 @@
 				"eip161dTransition": "0x927C0",
 				"eip649Reward": "0x3782DACE9D900000",
 				"eip100bTransition": 800000,
-				"eip649Transition": 800000
+				"eip649Transition": 800000,
+				"expip2DurationLimit": "0x1E",
+				"expip2Transition": 800000
 			}
 		}
 	},
--- a/ethcore/res/ethereum/expanse.json
+++ b/ethcore/res/ethereum/expanse.json
@@ -16,7 +16,10 @@
 				"eip150Transition": "0x927C0",
 				"eip160Transition": "0x927C0",
 				"eip161abcTransition": "0x927C0",
-				"eip161dTransition": "0x927C0"
+				"eip161dTransition": "0x927C0",
+				"eip649Reward": "0x3782DACE9D900000",
+				"eip100bTransition": 800000,
+				"eip649Transition": 800000
 			}
 		}
 	},
@@ -31,7 +34,11 @@
 		"subprotocolName": "exp",
 		"eip98Transition": "0x7fffffffffffff",
 		"eip86Transition": "0x7fffffffffffff",
-		"eip155Transition": "0x927C0"
+		"eip155Transition": "0x927C0",
+		"eip140Transition": 800000,
+		"eip211Transition": 800000,
+		"eip214Transition": 800000,
+		"eip658Transition": 800000
 	},
 	"genesis": {
 		"seal": {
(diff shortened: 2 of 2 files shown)


debris:
I believe these are indeed all changes. @5chdn can you confirm?

also, @chrisfranko do you have any document specifying these changes (and block transition number)? It would be nice to keep a link to a doc like that in the pr.

debris:
@closes #7061

chrisfranko:
we removed the difficulty bomb all together early on.

chrisfranko:
the zsnarks stuff that was in the gexp changelog so i just rolled with it

chrisfranko:
https://github.com/expanse-org/go-expanse/blob/master/consensus/ethash/consensus.go#L318

5chdn:
It seems you don't implement EIP 649 at all. Diff to [go-ethereum](https://github.com/ethereum/go-ethereum/blob/master/consensus/ethash/consensus.go#L316):

screenshot at 2017-11-27 14-51-17

chrisfranko:
Right because we completely removed the bomb before homestead. From my recollection parity had a bomb defuse param in the config file.

So if !bombDefusal do eip649 stuff, else dont.

chrisfranko:
Also what software is that, because that is super helpful.

5chdn:
I'm using http://meldmerge.org/ for quick diffs and merging.

So you are not using EIP-649, what else changes the difficulty in Byzantium?

chrisfranko:
The uncle calculations are pretty much the only thing extra in Byzantium. And the block reward being 4 instead of 8 and blocktime being 30 seconds instead of 60.

chrisfranko:
I think we found what I forgot to change. Before byzantium blocktime and after.

5chdn:
Requires some logic in the code. Does this fork have a name yet? Would certainly help.

Maybe looking into #6621 can help you to understand how we implement custom chain parameters.

chrisfranko:
In GO
ETH - x.Div(x, big9)
EXP - x.Div(x, big30)

In Parity i'm assuming this is duration_limit

https://github.com/ethereumproject/parity/blob/fe84144f56d97a82f54619f8f96d3d83c50108ee/ethcore/src/ethereum/ethash.rs#L86

This changes in eth from 10, to 9, but i dont cant seem to find where. Im sorry if im over looking it, i'm a complete rust noob.

5chdn:
@chrisfranko do you have any kind of formalization what exactly changes at 800_000? I personally want to see Expanse support in Parity, but I have issues to follow what this hard-fork exactly does. Do you have any Expanse Improvement Proposals or any similar process? Anything that allows developers to understand which consensus rules were changed and how?

chrisfranko:
The EXP fork at 800k is exactly what ETH did. Nothing different other than
we lowered our blocktime to 30 seconds and halved our block reward to 4.
ETH also lowerd their blocktime and reward. Everything else is the same.

5chdn:
But _how_ did you reduce it to 30 seconds? Simply doing a Byzantium transition on EXP does not do the trick. Did you change any code, could you point me to a commit maybe?

chrisfranko:
**Lowering block target**
https://github.com/expanse-org/go-expanse/blob/master/consensus/ethash/consensus.go#L334

chrisfranko:
Blocked? And removed? I was on vacation. We did everything we were supposed to do to maintain this.

5chdn:
Not sure why this card glitched into the Documentation board. Removed it.

chrisfranko:
im so confused

5chdn:
Let me try to explain. Your chain spec is fine. But it's not enough, you have some custom changes in your difficulty calculation that are not in line with those we implemented for our Ethereum-based engines. That's totally fine but requires some logic on core code-level to describe the difficulty correctly in case of the Expanse fork. 

First, we need a name, let's call it EXP-800. We then need to introduce some `exp800transition` in the chain spec and also in the Ethash module. Compare with other available transitions, e.g., MCIP-3, ECIP-1010 or EIP-649.

See https://github.com/paritytech/parity/blob/master/ethcore/src/ethereum/ethash.rs#L340-L411 for example. This is the parity implementation of calculate_difficulty we discussed earlier. This needs a special case for EXP-800. If you read through this function, you will see a lot of checks and special cases for other transitions, like EIP-100B, EIP-649, ECIP-1010. 

Now, all we need is a working implementation of the difficulty for EXP-800. You could template it on top of EIP-649 and modify it to match your custom Go-implementation. I'm sorry if this was not clear in my earlier comments.

Just to highlight two things: (1) A chain spec is only able to modify the specification within the existing boundaries and EXP-800 is beyond that scope. (2) Parity does not run one chain but many different ones; this abstraction allows for so many opportunities configuring different blockchain environments, but once you want custom consensus rules, it must be carefully specified somewhere, not only for one chain but for all supported chains. :)

chrisfranko:
Yes that makes alot more sense to me now, ill work on it tonight and
tomorrow.

5chdn:
I don't think this is enough consensus-wise. It does not compile.

I'll check if it syncs later tonight. WIP: https://github.com/paritytech/parity/commits/chrisfranko-patch-2

ethcore/src/ethereum/ethash.rs
```diff
@@ -355,7 +359,14 @@ impl Ethash {
...
 		};
 
-		let duration_limit = self.ethash_params.duration_limit;
+		let expip2_hardfork = header.number() > self.ethash_params.expip2_transition;
+
+		let duration_limit = if expip2_hardfork {
+			self.ethash_params.expip2_duration_limit
+		}else{
+			self.ethash_params.duration_limit
+		}
```
5chdn: `};`

5chdn:
2018-01-04 20:08:50  Syncing  #790509 01bc…da92   288 blk/s 2023 tx/s  44 Mgas/s   7721+  121 Qed   #798364    6/25 peers   7 MiB chain 98 MiB db 25 MiB queue 12 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
	2018-01-04 20:09:00  Syncing  #794463 451f…aa6b   401 blk/s 2613 tx/s  57 Mgas/s   7938+   50 Qed   #802474    3/25 peers   7 MiB chain 98 MiB db 23 MiB queue 8 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
	2018-01-04 20:09:10  Syncing  #798298 bc4c…466b   381 blk/s 2524 tx/s  55 Mgas/s   5116+   38 Qed   #803446    3/25 peers   7 MiB chain 98 MiB db 13 MiB queue 8 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
	2018-01-04 20:09:14  Stage 3 block verification failed for #800000 (68b8…a7ed)
	Error: Block(InvalidDifficulty(Mismatch { expected: 23024933077909, found: 23069903650326 }))
	2018-01-04 20:09:20  Syncing  #799999 6b8a…bcb2   174 blk/s  821 tx/s  20 Mgas/s    284+    0 Qed   #804594    5/25 peers   4 MiB chain 98 MiB db 672 KiB queue 8 MiB sync  RPC:  0 conn,  0 req/s,   0 µs

Hm ...

5chdn:
`metropolisDifficultyIncrementDivisor` was missing. #7463 should do the trick.

5chdn:
That was a duplicate entry.

PR #7159, opened 2017-11-28
fixes typo in user config path

r1b:
References the correct [parity_base](https://github.com/paritytech/parity/blob/bb1bbebfd63b5870478ff4bcd60d4a1ac1e00d36/ethstore/src/dir/paths.rs#L65) in the systemd service.


r1b:
[clabot:check]

5chdn:
Thanks.

PR #7191, opened 2017-12-04
Update mocha import stubs

jacogr:
Update ignores of unknown files in mocha tests, e.g. `import html from './page.ejs'`

Imports for mocha tests:
- Introduce generic filetype skip handler
- Remove ejsify (security advisory, replaced by generic)
- Remove ignore-style (replaced by generic)

Cleanups for tests:
- Remove misplaced .only in js-old, fixup changed tests

### DIFF
--- a/js/test/mocha.opts
+++ b/js/test/mocha.opts
@@ -1,3 +1,3 @@
 -r ./test/babel
 -r ./test/mocha.config
--r ignore-styles
+-r ./test/ignoreImports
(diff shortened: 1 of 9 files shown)


js/test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 -r ./test/babel
 -r ./test/mocha.config
--r ignore-styles
+-r ./test/ignoreImports
```
amaurymartiny: So ignoreImports is ignore-styles + ignore ejs, correct?

jacogr: Yes. We also do funny things with markdown (spotted it in my other PR, so adding that as well - all the WebPack loader stuff)

PR #7195, opened 2017-12-04
Update FirstRun for UI-2

jacogr:
Closes https://github.com/paritytech/parity/issues/7193


js/src/FirstRun/TnC/tnc.js
```diff
@@ -34,6 +34,7 @@ export default function TnC ({ hasAccepted, onAccept }) {
     <div className={ styles.tnc }>
       <ReactMarkdown
         className={ styles.markdown }
+        escapeHtml={ false }
```
jacogr: Issue where react-markdown currently doesn't render properly. From their page -

"Inline HTML is currently broken for any tags that include attributes. A vague idea of how to fix this has been planned, but if you're feeling up to the task, create an issue and let us know!"

js/src/FirstRun/firstRun.js
```diff
@@ -16,353 +16,61 @@
...
-                id='firstRun.button.close'
-                defaultMessage='Close'
-              />
-            }
-            onClick={ this.onClose }
+        hideClose
+        title={
+          <FormattedMessage
+            id='firstRun.title.termsOnly'
+            defaultMessage='Terms & Conditions'
```
amaurymartiny: VSCode shows this & symbol as red, i.e. should be replaced be &amp;. Not sure though, we can leave it like this.

jacogr: Easier to update and be 100% sure.

Issue #7080, opened 2017-11-17
Kovan Maintenance Checklist


5chdn:
**This is high-priority!**

## Priority List
Add/Remove items at will. **Bold items require a HF.**

Priority 1:
- [x] **enable Byzantium on Kovan**, ref #4833 
- [x] **disable uncles on Kovan** - https://github.com/paritytech/parity/pull/7006
- [x] Fix aura difficulty race - https://github.com/paritytech/parity/pull/7198
- [x] Wait for future blocks in AuRa https://github.com/paritytech/parity/pull/7368
- [ ] update parity branding on kovan-related sites

Priority 2:
- [ ] **switch to contract-based validator set** - https://wiki.parity.io/Validator-Set.html
- [ ] deploy admin contract - https://github.com/paritytech/contracts/pull/92
- [ ] **allow administration contracts to enact changes on the state** (soft fork)

Priority 3:
- [ ] upload Kovan image to MS Azure
- [x] enable secret-store servers on authority nodes - https://github.com/paritytech/parity/pull/7101
- [ ] (re)deploy Foundation-Kovan Bridge for contract-based validator set - https://github.com/paritytech/parity-bridge/pull/48

### Hard-Fork 1
- [x] decide on fork block number or date `5067000` #7259
- [x] publish releases: Parity 1.8.4-beta & Parity 1.7.10-stable
   - [x] **enable Byzantium on Kovan**, ref #4833 
   - [x] **disable uncles on Kovan** - https://github.com/paritytech/parity/pull/7006
   - [x] Fix aura difficulty race - https://github.com/paritytech/parity/pull/7198

### Hard-Fork 2
- [ ] decide on fork block number or date
- [ ] publish releases: Parity 1.10.X-beta & Parity 1.9.X-stable
   - [x] enable WASM VM - https://github.com/paritytech/parity/issues/7292
   - [ ] **switch to contract-based validator set** - https://wiki.parity.io/Validator-Set.html
   - [ ] **allow administration contracts to enact changes on the state** (soft fork)
- [ ] deploy admin contract - https://github.com/paritytech/contracts/pull/92
- [ ] upload Kovan image to MS Azure
- [x] enable secret-store servers on authority nodes - https://github.com/paritytech/parity/pull/7101
- [ ] (re)deploy Foundation-Kovan Bridge for contract-based validator set - https://github.com/paritytech/parity-bridge/pull/48

folsen:
[...]

PR #7196, opened 2017-12-04
Maximum uncle count transition

tomusdrw:
Related #7006 #7080

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -65,6 +65,8 @@ pub struct AuthorityRoundParams {
 	pub immediate_transitions: bool,
 	/// Block reward in base units.
 	pub block_reward: U256,
+	/// Number of accepted uncles transition block.
+	pub maximum_uncle_count_transition: u64,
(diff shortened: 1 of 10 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -79,6 +81,7 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 			validate_step_transition: p.validate_step_transition.map_or(0, Into::into),
 			immediate_transitions: p.immediate_transitions.unwrap_or(false),
 			block_reward: p.block_reward.map_or_else(Default::default, Into::into),
+			maximum_uncle_count_transition: p.maximum_uncle_count_transition.map_or(0, Into::into),
```
arkpar: This will silently break existing networks for Aura users. Consider setting to `u64::max_value()` instead.

tomusdrw: It's keeping the behavour introduced in #7006, I think we don't really want anyone running AuRa with uncles. @keorn ?

Issue #6761, opened 2017-10-14
UnknownUncleParent while syncing


phahulin:
Hello

> I'm running:
>
> - **Parity version**: Parity/v1.7.0-beta-5f2cabd-20170727/x86_64-linux-gnu/rustc1.18.0
> - **Operating system**: Ubuntu 16.04.1 LTS
> - **And installed**: via binaries

We have a private PoA network with validation via a contract. Yesterday we had a network issue and some nodes ran out of sync for a period of time. Later that day they all synced and network seemed to be working fine. However, today clients can't sync with it, the following error appears in logs:

```
Stage 3 block verification failed for #880003 (72cd…3db3)
Error: Block(UnknownUncleParent(6a1cd18267a8a60f7c86dac2b19ded1e2ad9284efefeace0ae8a376d6f961d69))
```
[More logs here](https://github.com/phahulin/-misc-temp/raw/master/log.txt.gz)

Could someone help figure out what might be the origin of this error? Using web3 I can find block 880003, its uncles, their parent blocks... so why are they _unknown_? what does it mean?

1. Block 880003:

<details>
<pre>
> web3.eth.getBlock(880003)
{ author: '0xdd0bb0e2a1594240fed0c2f2c17c1e9ab4f87126',
  difficulty: { [String: '3.40282366920938463463374607431768211453e+38'] s: 1, e: 38, c: [ 34028236692, 9384634633746, 7431768211453 ] },
  extraData: '0xd5830107008650617269747986312e31382e30826c69',
  gasLimit: 4700000,
  gasUsed: 0,
  hash: '0x72cd3a741f6d69f98fc14d32bb45d5ff022f0b35feea4be27b04a2c2c8a63db3',
  logsBloom: '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
  miner: '0xdd0bb0e2a1594240fed0c2f2c17c1e9ab4f87126',
  number: 880003,
  parentHash: '0xe6628345aa8b67a828586946ef2ca69073de7a9cd2bf8e444440693576c84c94',
[...]

PR #7198, opened 2017-12-04
Fix aura difficulty race

rphmeier:
Previously this sequence of events was possible:
- update_sealing called at step S when best block (from step S) imported
- populate_from_parent creates a header with invalid difficulty (2^128 + S - S)
- step changes to step S + 1
- generate_seal seals the block with invalid difficulty at step S + 1
- block is shortcutted into local chain while rejected by network
- potentially fork

This might cause #6761 but will need further investigation.


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -450,20 +450,29 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
 		// first check to avoid generating signature most of the time
 		// (but there's still a race to the `compare_and_swap`)
 		if !self.can_propose.load(AtomicOrdering::SeqCst) { return Seal::None; }
 
 		let header = block.header();
+		let parent_step: U256 = header_step(parent)
+			.expect("Header has been verified; qed").into();
+
 		let step = self.step.load();
+		let expected_diff = U256::from(U128::max_value()) + parent_step - step.into();
```
tomusdrw: Looks like the difficulty calcution is already copy pasted in couple of places, maybe we should extract it into a helper function?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -450,20 +450,29 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
+
+		if header.difficulty() != &expected_diff {
+			return Seal::None;
```
arkpar: Would it make sense to seal and release a block with a previous step instead?

rphmeier: This case can also only happen when `update_sealing` is called upon import of new best block from the prior step, but `update_sealing` also gets called when the step changes. So triggering sealing again for the correct step is unnecessary.

PR #7206, opened 2017-12-05
Rework dapps list

jacogr:
A start, there is more that can be done, provides a springboard for more cleanups.

parity 2017-12-05 16-22-00

### DIFF
--- a/js/src/Dapps/dapps.js
+++ b/js/src/Dapps/dapps.js
@@ -58,7 +58,7 @@ class Dapps extends Component {
                 app={ app }
                 availability={ availability }
                 className={ styles.dapp }
-                key={ app.id }
+                key={ `${index}_${app.id}` }
               />
             ))
           }
--- a/js/src/Dapps/dapps.js
+++ b/js/src/Dapps/dapps.js
@@ -14,19 +14,18 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import omitBy from 'lodash.omitby';
 import { observer } from 'mobx-react';
 import React, { Component } from 'react';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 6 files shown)


js/src/Dapps/dapps.js
```diff
@@ -76,72 +75,36 @@ class Dapps extends Component {
...
-      />
+      <Page>
+        <div className={ styles.dapps }>
+          {
+            applications.map((app, index) => (
+              <DappCard
+                app={ app }
+                availability={ availability }
+                className={ styles.dapp }
+                key={ `${index}_${app.id}` }
```
amaurymartiny: Wouldn't app.id be enough?

js/src/Dapps/dapps.js
```diff
@@ -48,6 +46,7 @@ class Dapps extends Component {
   }
 
   render () {
+    const { availability } = this.props;
     let externalOverlay = null;
 
     if (this.store.externalOverlayVisible) {
```
amaurymartiny: We could use a ? : operator here.

js/src/Dapps/dapps.js
jacogr: Yes, overkill (as I typed it that thought came to mind)

js/src/Dapps/dapps.js
jacogr: Sure.

js/src/Dapps/dapps.js
jacogr: Actually no. Just "fixed", the "reverted" as I remembered why I took this longer route.

When I have the app as both a local app (linked from dapps) as well as an app of the network, the ids are the same. (It was not an issue previously since they were in separate lists)

amaurymartiny: OK makes sense.

PR #7235, opened 2017-12-07
[stable] Backports + Kovan HF

tomusdrw:


arkpar:
Version bump also required here:
https://github.com/paritytech/parity/blob/stable/nsis/installer.nsi#L13
https://github.com/paritytech/parity/blob/stable/mac/Parity.pkgproj#L465

arkpar:
Byzantium built-ins are missing in the spec

PR #7236, opened 2017-12-07
UI redirect to 127.0.0.1 when localhost requested

jacogr:
1. Redirect requests from `localhost<port>` to `127.0.0.1:<port>`, passing token as required
2. Update `--ui-no-validation` to use `127.0.0.1:3000` instead of `localhost:3000`

### DIFF
--- a/js/src/index.parity.js
+++ b/js/src/index.parity.js
@@ -19,9 +19,10 @@ import 'whatwg-fetch';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
+import qs from 'querystring';
 import injectTapEventPlugin from 'react-tap-event-plugin';
 import { IndexRoute, Redirect, Route, Router, hashHistory } from 'react-router';
-import qs from 'querystring';
+import localStore from 'store';
 
 import ContractInstances from '@parity/shared/lib/contracts';
 import { initStore } from '@parity/shared/lib/redux';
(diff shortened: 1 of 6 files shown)


js/src/index.parity.js
```diff
@@ -52,12 +53,34 @@ if (process.env.NODE_ENV === 'development') {
...
+const searchIndex = window.location.search
+  ? window.location.search.indexOf(TOKEN_QS)
+  : -1;
 let token = null;
 
-if (window.location.hash && window.location.hash.indexOf(AUTH_HASH) === 0) {
-  token = qs.parse(window.location.hash.substr(AUTH_HASH.length)).token;
+if (hashIndex !== -1) {
+  // extract from hash (e.g. http://127.0.0.1:8180/#/auth?token=...)
+  token = qs.parse(window.location.hash.substr(hashIndex + 1)).token;
```
tomusdrw: Should we do `try-catch` here in case `qs.parse` fails?

js/src/index.parity.js
```diff
@@ -52,12 +53,34 @@ if (process.env.NODE_ENV === 'development') {
...
+if (window.location.hostname === 'localhost') {
+  const { hash, port } = window.location;
+
+  if (!token) {
+    // we don't have a token, attempt from localStorage
+    token = localStore.get('sysuiToken').match(/.{1,4}/g).join('-');
+  }
+
+  // reconstruct location, redirecting to 127.0.0.1
+  window.location = `http://127.0.0.1:${port}/${hashIndex !== -1 || !hash ? '#/' : hash}${token ? '?token=' : ''}${token || ''}`;
```
tomusdrw: 1. Could you use `window.location.protocol` here instead of `http`? (Future-proofing :))
2. If `?token` is part of `hash` it will be passed twice afaict, will `?token=...?token=...` be parsed correctly?

js/src/index.parity.js
```diff
@@ -52,12 +53,34 @@ if (process.env.NODE_ENV === 'development') {
 }
 */
 
-const AUTH_HASH = '#/auth?';
-
+const TOKEN_QS = '?token=';
+const hashIndex = window.location.hash
```
tomusdrw: Perhaps it's worth to extract this into a utility function so that we can test it?
```
function redirect(protocol, hostname, hash, search, localStore) {
 ... 
}
```

js/src/index.parity.js
jacogr: Agreed.

js/src/index.parity.js
jacogr: Currently (looked at the code itself), doesn't have exceptions, but rather safe than sorry.

js/src/index.parity.js
jacogr: Makes sense.

PR #7239, opened 2017-12-07
improve building from source

jdetychey:
Lots of people are having trouble building from source and keep up to date. The main reason seems to be that cargo/rustup is hard to work with. The 'abort' panic strategy error is very frustrating and seems to be solved 100% of time with a clean.
 Adding those lines:
 When compiling a crate if you receive this error:
```
error: the crate is compiled with the panic strategy `abort` which is incompatible with this crate's strategy of `unwind`
```
Cleaning the repository will most likely solve the issue, try:
```
cargo clean
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,12 +94,19 @@ $ cargo build --release
 ```
 
 This will produce an executable in the `./target/release` subdirectory.
-Note: if cargo fails to parse manifest try:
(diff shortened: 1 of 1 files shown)


jdetychey:
You're welcome parity-cla-bot, I did it years ago (-: :robot:

PR #7240, opened 2017-12-07
Reduce max block timestamp drift to 15 seconds

andresilva:
Reduce the current max block timestamp drift from 30 seconds to 15 seconds.


andresilva:
[clabot:check]

PR #7246, opened 2017-12-08
Statically link the C runtime on Windows

### LINKED ISSUE #6643: Error (0xc000007b) at starting application after installation

> I'm running:
>
> - **Parity version**: 1.7.2
> - **Operating system**: Windows /
> - **And installed**: via installer

i install parity via installer and after installation when i tried to start it gives me error unable to start correctly (0xc000007b)

---

tomaka:
Embed the C runtime inside Parity instead of linking dynamically.
Without this option, the user has to install the MSVC redistributable packages.

Untested, but I'm confident that it should fix #6643

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,3 @@
+[target.x86_64-pc-windows-msvc]
+# Link the C runtime statically ; https://github.com/paritytech/parity/issues/6643
+rustflags = ["-Ctarget-feature=+crt-static"]
--- a/nsis/installer.nsi
+++ b/nsis/installer.nsi
@@ -92,10 +92,6 @@ section "install"
 	file /oname=ptray.exe ..\windows\ptray\x64\Release\ptray.exe
 
 	file "logo.ico"
-	file vc_redist.x64.exe
-
-	ExecWait '"$INSTDIR\vc_redist.x64.exe"  /passive /norestart'
-	delete $INSTDIR\vc_redist.x64.exe
 	# Add any other files for the install directory (license files, app data, etc) here
 
 	# Uninstaller - See function un.onInit and section "uninstall" for configuration


arkpar:
Parity installer includes and installs VC runtime redistributable. #6643 Does not seem to be related. Usually there's a different error when a runtime library is missing.

arkpar:
I don't have a strong opinion on this. Dynamic CRT allows Microsoft to patch any security vulnerabilities that are found in the runtime. Static on the other hand reduces our support burden and simplifies installation and deployment.

debris:
I'd like to merge it. @tomusdrw what do you think?

tomusdrw:
Fine with me IF someone actually tests it on Windows box.

debris:
cc @paritytech/ci

kirushik:
Will/should we be able to drop https://github.com/paritytech/parity/blob/master/nsis/installer.nsi#L95-L99 together with this change?

svyatonik:
Just in case you guys are waiting for windows build, here are my results:
1) I was unable to build parity from this branch, because of error `error: unknown codegen option: ' target feature'`. This can be fixed by removing space between `-C` and `target-feature=+crt-static`
2) could confirm that after adding this option exe-file doesn't contains any CRT-dependencies (checked this using `depends.exe`)
3) could confirm that this exe file runs on the build-machine (i.e. with CRT installed). Can't test on plain windows, as I have no such PC/image :( But I guess everything should work (no CRT dependencies found)

So I suggest to fix the 'space issue' && merge this PR, if everyone are ok with that

tomaka:
Space issue fixed!

arkpar:
CRT should be removed from the installer as per @kirushik's comment

PR #7260, opened 2017-12-11
Inject parity script to all dapps // Expand dapps to any ZIP file

ngotchac:
This PR injects too all HTML files served by the dapps server the `parity.js` script that instantiate a new `window.ethereumProvider` object.

It also adds the possibility to now link a dapp to a ZIP file (like a Github Release), instead of only the `repo` + `commit` tupple
To do that, one must set as the `commit` field in the GithubHint contract entry the value `0x00..001`. This is to prevent the dapp server to extract zip files that should remain zip files, and not dapps.


dapps/src/tests/fetch.rs
```diff
@@ -167,13 +167,15 @@ fn should_return_fetched_dapp_content() {
 	response1.assert_status("HTTP/1.1 200 OK");
 	assert_security_headers_for_embed(&response1.headers);
 	assert_eq!(
```
tomusdrw: could use `assert!(condition, format, params..)` instead of `assert_eq!(expression, true, format, params..)`

hash-fetch/src/urlhint.rs
```diff
@@ -85,14 +89,16 @@ pub struct Content {
 	/// MIME type of the content
 	pub mime: Mime,
 	/// Content owner address
-	pub owner: Address,
+	pub owner: Address
```
tomusdrw: In rust we try to have trailing commas after each line.

hash-fetch/src/urlhint.rs
```diff
@@ -122,6 +128,15 @@ impl URLHintContract {
 	}
 }
 
+fn get_urlhint_content(account_slash_repo: String, owner: Address) -> Content {
+	let mime = guess_mime_type(&account_slash_repo).unwrap_or(mime::APPLICATION_JSON);
+	Content {
+		url: account_slash_repo,
+		mime: mime,
+		owner: owner
```
tomusdrw: Could be simplified to:
```
Content {
  url: account_slash_repo,
  mime,
  owner,
}
```

hash-fetch/src/urlhint.rs
ngotchac: Can't there be a linter for this?

dapps/src/tests/fetch.rs
ngotchac: :+1:

hash-fetch/src/urlhint.rs
tomusdrw: I'd love to have one, we tried using `rustfmt` some time ago, but there were many places were the formatting was a bit weird. We should give it a go again soon though!

PR #7264, opened 2017-12-11
Simplify status + content display overlaps/page fixing

jacogr:
1. Get rid of the display + fixed approach, rather have the status + content in 2 separate section, the lower content layer being scrollable.
2. Apply consistently to both dapps and dapp views.
3. Firefox now renders consistently again (although in that case the apps take up too much space, to be fixed in https://github.com/paritytech/parity/issues/7224)

Closes https://github.com/paritytech/parity/issues/7242 - error bar is still always visible, however doesn't overlap

### DIFF
--- a/js/src/Dapps/dapps.js
+++ b/js/src/Dapps/dapps.js
@@ -50,19 +50,17 @@ class Dapps extends Component {
     const applications = [].concat(this.store.visibleLocal, this.store.visibleViews, this.store.visibleBuiltin, this.store.visibleNetwork);
 
     return (
-      <Page>
-        <div className={ styles.dapps }>
-          {
-            applications.map((app, index) => (
-              <DappCard
-                app={ app }
-                availability={ availability }
-                className={ styles.dapp }
-                key={ `${index}_${app.id}` }
-              />
-            ))
-          }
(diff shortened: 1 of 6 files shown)


js/src/Dapps/dapps.js
```diff
@@ -50,19 +50,17 @@ class Dapps extends Component {
...
-              <DappCard
-                app={ app }
-                availability={ availability }
-                className={ styles.dapp }
-                key={ `${index}_${app.id}` }
-              />
-            ))
-          }
-        </div>
+      <Page className={ styles.dapps }>
```
amaurymartiny: I'm modifying this heavily in #7266, I think you can remove it from this PR. It would save some time for me or you resolving conflicts

jacogr: Yup, just made a comment - think we need this one in first and then make sure that your still works 100% as expected. I'm worried since the top layout is now not a fixed bar anymore, rather a proper layout. (Needed to get done since we have had complaints about overlaps)

amaurymartiny: Agreed, i'll rebase mine on yours.

PR #7282, opened 2017-12-14
Problem: AuRa's unsafeties around step duration

yrashk:
Firstly, `Step.duration_remaining` casts it to u32, unnecesarily
limiting it to 2^32. While theoretically this is "good enough" (at 3
seconds steps it provides room for a little over 400 years), it is
still a lossy way to calculate the remaining time until the next step.

Secondly, step duration might be zero, triggering division by zero
in `Step.calibrate`

Solution: rework the code around the fact that duration is
typically in single digits and never grows, hence, it can be represented
by a much narrower range (u8) and this highlights the fact that
multiplying u64 by u8 will only result in an overflow in even further
future, at which point we should panic informatively (if anybody's
still around)

Similarly, panic when it is detected that incrementing the step
counter wrapped around on the overflow of usize.

As for the division by zero, prevent it by making zero an invalid
value for step duration. This will make AuRa log the constraint
mismatch and panic (after all, what purpose would zero step duration
serve? it makes no sense within the definition of the protocol,
as finality can only be achieved as per the specification
if messages are received within the step duration, which would violate
the speed of light and other physical laws in this case).

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -51,8 +51,12 @@ mod finality;
 
 /// `AuthorityRound` params.
 pub struct AuthorityRoundParams {
-	/// Time to wait before next block or authority switching.
-	pub step_duration: Duration,
+	/// Time to wait before next block or authority switching,
+	/// in seconds.
+	///
+	/// Deliberately typed as u16 as too high of a value leads
+	/// to slow block issuance.
+	pub step_duration: u16,
 	/// Starting step,
 	pub start_step: Option<u64>,
 	/// Valid validators.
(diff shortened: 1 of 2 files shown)


yrashk:
The test failure seems to be unrelated to the patch:

```
$ if [ $RUST_FILES_MODIFIED -eq 0 ]; then echo "Skipping Rust tests since no Rust files modified."; else ./test.sh $CARGOFLAGS; fi
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating git repository `https://github.com/paritytech/rust-secp256k1`
    Updating git repository `https://github.com/paritytech/bn`
    Updating git repository `https://github.com/paritytech/hyper`
    Updating git repository `https://github.com/paritytech/rust-snappy`
    Updating git repository `https://github.com/paritytech/jsonrpc.git`
    Updating git repository `https://github.com/paritytech/hidapi-rs`
    Updating git repository `https://github.com/paritytech/libusb-rs`
    Updating git repository `https://github.com/paritytech/trezor-sys`
    Updating git repository `https://github.com/paritytech/rust-rocksdb`
ERROR: Job failed: execution took longer than 3h10m0s seconds
```

yrashk:
[clabot:check]

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1073,4 +1109,50 @@ mod tests {
 		assert_eq!(aura.maximum_uncle_count(1), 0);
 		assert_eq!(aura.maximum_uncle_count(100), 0);
 	}
+
+    #[test]
```
rphmeier: indentation issue here

rphmeier:
LGTM except for indentation in one place

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -92,26 +103,47 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
 impl Step {
 	fn load(&self) -> usize { self.inner.load(AtomicOrdering::SeqCst) }
 	fn duration_remaining(&self) -> Duration {
 		let now = unix_now();
-		let step_end = self.duration * (self.load() as u32 + 1);
-		if step_end > now {
-			step_end - now
-		} else {
-			Duration::from_secs(0)
+		let expected_seconds = (self.load() as u64)
```
tomusdrw:
```rust
let expected_seconds = (self.load() as u64)
			.checked_add(1)
			.and_then(|ctr| ctr.checked_mul(self.duration as u64))
            .map(Duration::from_secs);
match expected_seconds {
  Some(step_end) if step_end > now => {
    step_end - now
  },
  Some(_) => {
     Duration::from_secs(0)
  },
  None => {
     ...
  }
}
```

PR #7285, opened 2017-12-14
CSS fixes for v1

jacogr:
Fixes https://github.com/paritytech/parity/issues/7284

(Thanks to @ngotchac for coming up with a solution)


amaurymartiny:
Yes, looks nice now!

PR #7299, opened 2017-12-15
Fix inject.js & Signer store duplication

jacogr:


amaurymartiny:
Yes, fixes both issues!

Just one small comment, I have a "The prop `date` is marked as required in `Request`, but its value is `undefined`"

amaurymartiny:
Will fix in https://github.com/paritytech/js-ui/issues/28

PR #7318, opened 2017-12-18
UI package bump

jacogr:
+ small alignment fixes to statusbar plugins (non-mexican wave now)

Closes https://github.com/paritytech/parity/issues/7319


amaurymartiny:
Couldn't repro the mexican wave, but looks good on my browsers.

PR #7323, opened 2017-12-18
SecretStore: tx retry pattern

svyatonik:
on top of #7271 

**It is PR for a single commit only**: https://github.com/paritytech/parity/commit/51cb9ff45b235252a90d826221ca14371ef40b65

This is a fix of TODO1 from #7271 . Wanted to make this a separate PR, because I'm not sure if I've selected a correct pattern. The problem is that we need to submit transaction sometimes (to start/conform migration). And I wanted to make sure that transaction will be mined eventually.

SS auto-migration has its own specific - transactions are (going to be) submitted by authorities in PoA/Kovan (initially I've just copied the pattern from `report_malicious`/`report_benign`). So most probably (not 100% sure, though) they will be mined in next block. But I want to be sure that the same scheme will work for other cases (running on non-mining node, not in PoA, ...).

The idea is simple: if transaction is not mined within N blocks, resubmit it. Same nonce is used (as I understood from `transact_contract` code). So imho this will help with several cases:
- case when submitter account had no enough eth for this tx
- case when tx has never reached miner node (dropped from queue? no peers?)

This won't help, however, with cases when gas price is too low, but since the price is calculated from the tx queue, it shouldn't be a problem (eventually we will calculate the correct price).


5chdn:
Please resolve conflicts :)

PR #7331, opened 2017-12-19
SecretStore: return error 404 when there's no key shares for given key on all nodes

svyatonik:
The same behavior was there before `KeyVersionNegotiationSession` was introduced: return HTTP 404 error when trying to restore the key before it was generated.
By @grbIzl requrest

### DIFF
--- a/secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
+++ b/secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
@@ -227,7 +227,7 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 		// try to complete session
 		Self::try_complete(&self.core, &mut *data);
 		if no_confirmations_required && data.state != SessionState::Finished {
-			return Err(Error::ConsensusUnreachable);
+			return Err(Error::MissingKeyShare);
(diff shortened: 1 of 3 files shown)


secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -227,7 +227,7 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 		// try to complete session
 		Self::try_complete(&self.core, &mut *data);
 		if no_confirmations_required && data.state != SessionState::Finished {
-			return Err(Error::ConsensusUnreachable);
```
tomusdrw: What's the reason for this change? Can you elaborate?

svyatonik: Do you mean this particular line? It's `initialize` method which is called on the 'session master' node. `no_confirmations_required` means that there are no other nodes in the SecretStore (i.e. SS consists of the single KeyServer && we are on this KeyServer). Since we haven't found key share for given key on this key server (`state != SessionState::Finished`) => this means that key haven't been generated at all => we need to return `MissingKeyShare`.

`MissingKeyShare` in turn is converted to the 404 HTTP error later

PR #7336, opened 2017-12-19
SecretStore: secretstore_signRawHash method

### LINKED ISSUE #7248: RPC method to generate signatures for Secretstore

> I'm running:
>
> - **Which Parity version?**: 1.9.0

I am looking into secretstore. Discussed with Slava over the Riot EWF channel, that eth_sign is not appropriate to generate the signature for the secretID in KeyServer::generate_document_key().

It would be good to have an RPC method to replicate the behaviour of this script:
https://gist.github.com/svyatonik/cf36dec2623685eab0cbcd87fca0a244

Maybe also output the address?

svyatonik:
closes #7248 

Overview:
1) added `secretstore_signRawHash` method
2) renamed `secretstore_signServersSet` (semantics is changed also)
3) marked `secretstore` API set as unsafe, because it now allows to sign any passed raw hash (without prepending with the prefix && without applying hash function)

`secretstore_signRawHash` can now be used to sign the server key id, which was initially intended to be a `keccak256(document_contents)`. It also should be used to sign the servers set hash after calling the `secretstore_serversSetHash` method.

### DIFF
--- a/parity/rpc_apis.rs
+++ b/parity/rpc_apis.rs
@@ -66,7 +66,7 @@ pub enum Api {
 	Traces,
 	/// Rpc (Safe)
 	Rpc,
-	/// SecretStore (Safe)
+	/// SecretStore (UNSAFE: arbitrary hash signing)
 	SecretStore,
 	/// Whisper (Safe)
 	// TODO: _if_ someone guesses someone else's key or filter IDs they can remove
@@ -602,7 +602,6 @@ impl ApiSet {
 			Api::EthPubSub,
 			Api::Parity,
 			Api::Rpc,
-			Api::SecretStore,
 			Api::Whisper,
 			Api::WhisperPubSub,
 		].into_iter().cloned().collect();
@@ -627,6 +626,7 @@ impl ApiSet {
 				public_list.insert(Api::ParityAccounts);
 				public_list.insert(Api::ParitySet);
 				public_list.insert(Api::Signer);
+				public_list.insert(Api::SecretStore);
 				public_list
 			},
 			ApiSet::All => {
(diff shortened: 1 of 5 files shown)


rpc/src/v1/tests/mocked/secretstore.rs
```diff
@@ -100,7 +101,29 @@ fn rpc_secretstore_shadow_decrypt() {
 }
 
 #[test]
-fn rpc_secretstore_sign_servers_set() {
+fn rpc_secretstore_servers_set_hash() {
+	let deps = Dependencies::new();
```
tomusdrw: You can consider using new `jsonrpc_test`:
https://github.com/paritytech/jsonrpc/blob/master/test/src/lib.rs#L28

parity/rpc_apis.rs
```diff
@@ -627,6 +626,7 @@ impl ApiSet {
 				public_list.insert(Api::ParityAccounts);
 				public_list.insert(Api::ParitySet);
 				public_list.insert(Api::Signer);
+				public_list.insert(Api::SecretStore);
```
debris: just a question. Above you've added a comment `SecretStore (UNSAFE: arbitrary hash signing)`, but here `SecretStore` is added to a `SafeContext`. Isn't it contradictory?

svyatonik: I was also confused a bit :) If I understood correctly, there are:
1) safe/unsafe APIs
2) safe/unsafe API Contexts
So: SecretStore API is unsafe => it is only enabled when API context is safe. Here's couple of proofs I've found:
1) https://github.com/paritytech/parity/blob/master/parity/rpc_apis.rs#L148 (reveres proof)
2) ParityAccounts API, which is also unsafe, is included to the `SafeContext` API set

tomusdrw: Yes, indeed that the reasoning, although It's extremely confusing (sorry about that:() Feel free to rename it.

PR #7338, opened 2017-12-19
Fix dappIcon & Fix Signer Pending

### LINKED ISSUE #7219: DappIcon in UI does not load in all cases, i.e. when dappUrl is set

amaurymartiny:
- DappIcons all display correctly. Fix #7219
- Put signerPending from js-ui into this repo.

PS: @tomusdrw I finally decided not to change to the notification UI in the extension for now: it was way too much headache (popup placed randomly, icons not loading...) for not much result. We can do that when we decide to clean the extension, cf your issue.

### DIFF
--- a/js/test/babel.js
+++ b/js/test/babel.js
@@ -14,13 +14,5 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-require('babel-register')({
-  ignore: function (filename) {
-    if (filename.indexOf('node_modules') !== -1) {
-      return filename.indexOf('@parity') === -1;
-    }
-
-    return false;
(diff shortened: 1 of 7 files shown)


js/test/babel.js
```diff
@@ -14,13 +14,5 @@
...
-require('babel-register')({
-  ignore: function (filename) {
-    if (filename.indexOf('node_modules') !== -1) {
-      return filename.indexOf('@parity') === -1;
-    }
-
-    return false;
-  }
-});
+require('babel-register')();
```
amaurymartiny: Explanation: All `@parity/*` export transpiled JS, so no need to babelify them again as done before.

jacogr: Good catch.

PR #7346, opened 2017-12-20
Problem: Cargo.toml file contains [project] key

yrashk:
This key is not recognized by some software
(like carnix).

It is also not documented in http://doc.crates.io/manifest.html

Solution: rename this key to [package]


tomusdrw:
Thanks @yrashk !

PR #7348, opened 2017-12-20
Upgrade to RocksDB 5.8.8 and tune settings to reduce space amplification

### LINKED ISSUE #6280: Initial Sync disk IO / write amplification / disk usage
> - **Parity version**: 1.7.0
> - **Operating system**: Ubuntu 16.04 LTS
> - **And installed**: via deb Package from https://parity.io/parity.html
> - **cmd line**: parity daemon /foo/bar/parity.pid
> - **config.toml**
```
[ui]
disable = true

[network]
nat = "any"
discovery = true
no_warp = true
allow_ips = "public"

[rpc]
disable = true

[websockets]
disable = true

[ipc]
disable = true

[dapps]
disable = true

[footprint]
tracing = "off"
db_compaction = "ssd"
pruning = "archive"
cache_size = 55000

[snapshots]
disable_periodic = true

[misc]
log_file = "/foo/bar/parity.txt"
```

I'm trying to setup a full node with complete history, thus `pruning=archive`.

Disk IO looks like this on Amazon EC2 instance type *c4.8xlarge*:

```
Device:            tps    MB_read/s    MB_wrtn/s    MB_read    MB_wrtn
xvdf           4252.00         0.36       519.55          0        519
```

## parity is *constantly* writing to disk with ~300-500 MByte/s and peaks reach ~5000 IOPS.

What really bothers me is that parity is wasting so much disk space. There are times where i can see that /home grows by 1 GB/s just watching `df -h /home`.

Some time after 2 Millions blocks were passed disk usage on /home was ~80 GB of data from parity alone. When stopping parity, that 80 GB of disk usage magically shrinks to ~37 GB of disk usage just to grow with 1 GB/s again after restarting parity.

Parity even ran out of disk space after filling up a 100GB EBS volume on Amazon AWS and at that time it had only downloaded about 50% of blocks.

## My questions are:

* why is parity writing at such a high volume to disk even though I've set cache size to 55GB?
  * is this some sort of write amplification? it isn't logical for me that an application that downloads less than 1 MB/s from the internet is writing 500 MB/s to disk at all times.
* why is parity using so much disk space temporarily? Even without restarting parity, disk usage sometimes goes down 20 GB or more from one second to the next.

andresilva:
This PR is an attempt to fix the space amplification caused by RocksDB. The main changes are:

* Switch to level-style compaction
* Increase default block size (16K), and use bigger blocks for HDDs (64K)
* Increase default file size base (64MB SSDs, 256MB HDDs)
* Create a single block cache shared across all column families
* Tune compaction settings using RocksDB helper functions, taking into account
  memory budget spread across all columns
* Use a global write buffer limit proportional to the memory budget to force earlier flushing
* Configure backgrounds jobs based on the number of CPUs
* Set some default recommended settings

That is based on the recommendations from the RocksDB wiki and guided by benchmarks I ran (importing 2M blocks, restoring a snapshot).

I also started running an archive node which is currently synced at around block 3.5M. This is how the DB size has been over time: db size

A user on #6280 submitted this graph of db size when syncing: https://i.imgur.com/1IpiTey.png.

The `cache-size-db` is now used as a memory budget for setting up RocksDB settings, instead of referring only to the actual block cache. This better reflects the overall memory usage for RocksDB and allows us to achieve different trade-offs (giving it more memory could reduce IO write amplification and increase performance, at the cost of potentially increasing space amplification). Should we rename this option to `memory-budget-db` or something like that? Also, since this no longer refers to only the size of the block cache I increased its default value to 128MB, memory usage should be below this since, in this case, we would have a global flush limit of 64MB and we would use 42MB for the block cache (one third of the memory budget).

These new settings require a re-sync since some options are not compatible with the previous ones (I'm not sure which ones, I haven't tested them individually). Even if all the options were compatible, since we're changing the compaction algorithm it is better to start from scratch to make sure that the LSM has the expected "shape". I'm not familiar with the migration module, but we could probably rebuild the database with the new settings from the existing one. I can look into it if you think this is necessary.

Depends on paritytech/rust-rocksdb#10.
Should fix #6280 but it would be cool to have more people test this on different environments.


tomusdrw:
We should consider backporting this to `beta`.

kirushik:
Wow, such a wonderful work, @andresilva .

I would support the idea of migrating old database into the old one. While it's a really reasonable to want a fresh database storage, not having our users to re-sync from scratch will be a huge improvement (warp sync is still not very warpy, by the way, #6372). It hurts to think about our users enabling autoupdates and then us rendering their nodes unusable for days while resyncing!

Bonus _brownie points_ if we won't require 2× disk space while migrating, but that's something we can live with, I guess.

tomusdrw:
@kirushik @andresilva +1 for the migration, I think our migration framework needs a bit of work to support such migration, but it is definitely required.

andresilva:
@kirushik @tomusdrw I agree, I will look into the migrations module. I'm not sure it's worth it to avoid the 2x space overhead but I'll check how feasible it is to implement (I'm also not sure how much space you'd actually save while deleting data from the old database, since compactions are asynchronous).

The need for a re-sync actually got me thinking about our auto-update system, e.g. we could take down a network by having all nodes auto-update and start a re-sync right away (the same thing could happen with random bugs / byzantine behavior that is introduced in an auto-update). There should be a way to do a phased roll out `</offtopic>`.

tomusdrw:
I think 2x space overhead, is a minor goal. We can recommend a block export/import instead of migration for some users.
Alternatively would it be possible to introduce a backward compatibility flag (i.e. use old compaction settings)?

Re auto-update: I suppose it would be cool to expose `--autoupdate-delay` flag for users, so they can decide on the "certainty" (mettered either in blocks or in time) of the update themselves (by default it should probably choose some random value between 0 and 100 blocks).

andresilva:
@tomusdrw Yes, I think it's possible to add a backward compatibility flag. Alternatively, we should be able to auto-detect if it's a database with old settings. RocksDB stores the used options in a file, so in case we can't open the database we can just fallback to whatever settings were being used. I think we may not need a migration if this works properly, we can just advise users that are having issues to do a `db kill` or an import/export.

tomusdrw:
Great! That sound like the best option.

andresilva:
I was wrong, it is indeed possible to open an existing database with these new settings. I had tested this locally with a `kovan` chain and it started doing a snapshot restore which led me to believe that we weren't able to open the existing database (which is not the case, we simply always prefer restoring from snapshots if a new one is available).

Unfortunately I can't implement the behavior of reading the RocksDB options from the file (https://github.com/facebook/rocksdb/wiki/RocksDB-Options-File) since that functionality isn't exposed in the C bindings, and so we have no way to use it in our Rust bindings. We don't change any of the settings in [`CheckOptionsCompatibility`](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/utilities/options_util.h#L81) and the RocksDB FAQ has this to say about changing compaction:

```
Q: What will happen if I open RocksDB with a different compaction style?

A: When opening a rocksdb database with a different compaction style or 
compaction settings, one of the following scenarios will happen:

    * The database will refuse to open if the new configuration is incompatible 
      with the current LSM layout.
    * If the new configuration is compatible with the current LSM layout, then 
      rocksdb will continue and open the database. However, in order to make
      the new options take full effect, it might require a full compaction.
```

So I think we should be safe as it is. Is anyone willing to test this locally and make sure that they don't lose their database? The update should be visible in database's OPTIONS file:

```
❯ rg rocksdb_version ~/Library/Application\ Support/io.parity.ethereum/chains/ethereum/
/Users/andre/Library/Application Support/io.parity.ethereum/chains/ethereum/db/906a34e69aec8c0d/overlayrecent/db/OPTIONS-351370
8:  rocksdb_version=5.8.8

/Users/andre/Library/Application Support/io.parity.ethereum/chains/ethereum/db/906a34e69aec8c0d/overlayrecent/db/OPTIONS-349830
8:  rocksdb_version=4.11.2
```

tomusdrw:
@andresilva Successfuly run this version of old Kovan database. Still syncing, will try mainnet next.

andresilva:
@tomusdrw Cool! I also tested switching back to the old settings on a new database and it worked properly. In the meantime my archive node has finished syncing, [here's](https://i.imgur.com/qfWeUSx.png) the final graph of DB size over time. Took around 48 hours to do the full sync, and it seems there are no more temporary spikes in db size by a factor of 2.

tomusdrw:
Since the change is non-breaking and it improves the archive nodes greately, we should consider backporting it to 1.8 stream. @arkpar @debris thoughts?

5chdn:
> Took around 48 hours to do the full sync

Just came across this PR now, wow! :heart: 

If this does not break too much, we should really backport it to beta and push it asap next year. :+1:

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -265,31 +266,35 @@ impl Database {
...
-		opts.set_max_open_files(config.max_open_files);
-		opts.create_if_missing(true);
 		opts.set_use_fsync(false);
-		opts.set_db_write_buffer_size(DB_WRITE_BUFFER_SIZE);
+		opts.create_if_missing(true);
+		opts.set_max_open_files(config.max_open_files);
+		opts.set_parsed_options("keep_log_file_num=1")?;
+		opts.set_parsed_options("bytes_per_sync=1048576")?;
+		opts.set_db_write_buffer_size(config.memory_budget_per_col() / 2);
+		opts.increase_parallelism(cmp::max(1, ::num_cpus::get() as i32 / 2));
```
arkpar: I'd recommend setting it to `num_cpus - 1`  if there is more than one CPU. One CPU should be always available for block import job to minimize import latency.

tomusdrw: We set it to `num_cpus / 2` currently, so only using half of the cores (which makes sense to me)

jo-tud:
yipeeh!

debris:
🎉

roninkaizen:
using the rocksdb5 Version published some time in december brought back stability even to me-

still one of the best "upcatching" node-version-
and yes-
some day i tested non accepting "random" peers and setting up some
of the mentioned nodes from the bootnodes section,
violà- constantly connected to 8 peers-
who needs more peers as they do not "behave/sync" as intended?
the disc-space usage grew a little, but not so enormous as described from the others.

PR #7355, opened 2017-12-21
Get rid of clippy remainings.

tomusdrw:
We don't really use this feature any more.

I tried running clippy as cargo plugin, but unfortunatelly it failed to compile Parity on latest nightly.

### DIFF
--- a/ethcore/src/lib.rs
+++ b/ethcore/src/lib.rs
@@ -16,23 +16,6 @@
 
 #![warn(missing_docs)]
 #![cfg_attr(feature="benches", feature(test))]
-#![cfg_attr(feature="dev", feature(plugin))]
-#![cfg_attr(feature="dev", plugin(clippy))]
-
-// Clippy settings
-// Most of the time much more readable
-#![cfg_attr(feature="dev", allow(needless_range_loop))]
-// Shorter than if-else
-#![cfg_attr(feature="dev", allow(match_bool))]
-// Keeps consistency (all lines with `.clone()`).
-#![cfg_attr(feature="dev", allow(clone_on_copy))]
-// Complains on Box<E> when implementing From<Box<E>>
-#![cfg_attr(feature="dev", allow(boxed_local))]
-// Complains about nested modules with same name as parent
-#![cfg_attr(feature="dev", allow(module_inception))]
-// TODO [todr] a lot of warnings to be fixed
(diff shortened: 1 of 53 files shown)


svyatonik:
Does this also counts as 'clippy remainings'? :)
https://github.com/paritytech/parity/blob/master/parity/helpers.rs#L220
https://github.com/paritytech/parity/blob/master/parity/helpers.rs#L456
https://github.com/paritytech/parity/blob/master/util/patricia_trie/src/triedbmut.rs#L439

Maybe worth adding to this PR as well. Otherwise - lgtm.

tomusdrw:
Good catch. Forgot to update `sed` params for those :)

util/src/lib.rs
```diff
@@ -15,23 +15,6 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 #![warn(missing_docs)]
-#![cfg_attr(feature="dev", feature(plugin))]
-#![cfg_attr(feature="dev", plugin(clippy))]
-
-// Clippy settings
-// Most of the time much more readable
-#![cfg_attr(feature="dev", allow(needless_range_loop))]
```
debris: if we want to run clippy as a plugin, we should not remove those, but only rename to:

```
#[cfg_attr(feature="cargo-clippy", allow(needless_range_loop))]
```

ethcore/src/lib.rs
```diff
@@ -16,23 +16,6 @@
 
 #![warn(missing_docs)]
 #![cfg_attr(feature="benches", feature(test))]
-#![cfg_attr(feature="dev", feature(plugin))]
-#![cfg_attr(feature="dev", plugin(clippy))]
-
-// Clippy settings
-// Most of the time much more readable
-#![cfg_attr(feature="dev", allow(needless_range_loop))]
```
debris: if we want to run clippy as a plugin, we should not remove those, but only rename to:

```
#[cfg_attr(feature="cargo-clippy", allow(needless_range_loop))]
```

tomusdrw: I tried running `cargo-clipp` as a plugin but it didn't work (I think parity failed to compile on nightly). Let's look at this from a fresh perspective when it works, some of the things here might not be valid any more, I would prefer to fix the code at some point rather than disable lints.
What do you think?

debris: fair enough ;)

PR #7356, opened 2017-12-21
Wasm runtime update

NikVolf:
- call now obliged to specify gas limit
- overall update of wasm-tests and corresponding gas costs
- some environment for calls was fixed to be passed correctly

### DIFF
--- a/ethcore/wasm/src/env.rs
+++ b/ethcore/wasm/src/env.rs
@@ -69,7 +69,7 @@ pub const SIGNATURES: &'static [UserFunctionDescriptor] = &[
 	),
 	Static(
 		"_ccall",
-		&[I64, I32, I32, I32, I32, I32],
+		&[I64, I32, I32, I32, I32, I32, I32],
 		Some(I32),
 	),
 	Static(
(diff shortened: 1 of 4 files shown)


ethcore/wasm/src/env.rs
```diff
@@ -69,17 +69,17 @@ pub const SIGNATURES: &'static [UserFunctionDescriptor] = &[
 	),
 	Static(
 		"_ccall",
-		&[I32; 6],
+		&[I64, I32, I32, I32, I32, I32],
```
pepyakin: Is it really a correct signature? In `pwasm-std/pwasm-ethereum` it is defined like

```rust
		pub fn ccall(
			gas: i64,
			address: *const u8,
			val_ptr: *const u8,
			input_ptr: *const u8,
			input_len: u32,
			result_ptr: *mut u8,
			result_len: u32,
		) → i32;
```

which have 7 arguments, but in code above it is only 6.

NikVolf: Nice catch, regular call lacks dedicated test also

pepyakin:
Also, could you please update method signatures, e.g.

https://github.com/paritytech/parity/blob/d5b81ead71a8c4b3eca5cc3dd18e575534cf73a6/ethcore/wasm/src/runtime.rs#L305-L314

Issue #7350, opened 2017-12-20
Transaction permission contract is not used by validators


VladLupashevskyi:
> I'm running:
>
> - **Which Parity version?**: 1.8.4 beta + compiled unstable
> - **Which operating system?**: MacOS
> - **How installed?**: via homebrew and from source
> - **Are you fully synchronized?**: yes
> - **Did you try to restart the node?**: yes

Hi people,

I was testing transaction permission contract on my Aura PoA chain since we need to implement some permission logic for our users based on tx types.

I set up basic transaction permission contract to return always 0 for everybody, so no transactions were allowed.

Everything worked fine when I was using 1.8.4 beta version of Parity for sealers and non-sealers.

Then I decided to change source code and recompile parity for non-sealers.

So I changed return value in function [transaction_allowed](https://github.com/paritytech/parity/blob/master/ethcore/src/tx_filter.rs#L105) to return always true. 

I run this tweaked version for common account (my sealers were still running beta version of parity) and tried to send some ether. Finally transaction was approved by the sealer, although it shouldn't since I denied all txes in transaction permission contract.

Summarizing the above I conclude that tx types are not checking by validators and this issue should be somehow fixed.

Best regards,
Vlad

PR #7359, opened 2017-12-21
Added checking tx-type using transactions permission contract for miners

VladLupashevskyi:
In this commit you can find solution to the issue [#7350](https://github.com/paritytech/parity/issues/7350) which I opened recently.

Now block miners are checking whether the tx type is allowed for sender using transaction permission contract.


tomusdrw:
I think the check should actually be done **before** importing transactions to the queue as well. Otherwise we might be storing and propagating a lot of invalid transactions.

Should be done here:
https://github.com/paritytech/parity/blob/master/ethcore/src/miner/miner.rs#L675

Other than that: great work, thank you @VladLupashevskyi

ethcore/src/miner/miner.rs
```diff
@@ -419,7 +420,15 @@ impl Miner {
 		for tx in transactions {
 			let hash = tx.hash();
 			let start = Instant::now();
-			let result = open_block.push_transaction(tx, None);
+			// Check whether transaction type is allowed for sender
+			let result = match self.engine.machine().verify_transaction(&tx, open_block.header(), chain.as_block_chain_client()) {
+				Err(Error::Transaction(TransactionError::NotAllowed)) => {
+					Err(From::from(TransactionError::NotAllowed))
```
rphmeier: favor `.into()` instead of `From::from`

ethcore/src/miner/miner.rs
```diff
@@ -419,7 +420,15 @@ impl Miner {
 		for tx in transactions {
 			let hash = tx.hash();
 			let start = Instant::now();
-			let result = open_block.push_transaction(tx, None);
+			// Check whether transaction type is allowed for sender
+			let result = match self.engine.machine().verify_transaction(&tx, open_block.header(), chain.as_block_chain_client()) {
```
rphmeier: I guess `open_block`'s parent state always has to be available in the client, so this is fine.

arkpar:
Agree with @tomusdrw. It is assumed that transaction queue contains valid transactions only.

tomusdrw:
@arkpar although the state of the transaction might change in the meantime, with current queue we have no way of invalidating such transactions, so I'm ok with check in both places.

arkpar:
Makes sense

ethcore/src/miner/miner.rs
```diff
@@ -680,6 +698,12 @@ impl Miner {
 						Err(e)
 					},
 					Ok(transaction) => {
+						// This check goes here because verify_transaction takes SignedTransaction parameter
+						if self.engine.machine().verify_transaction(&transaction, &best_block_header, client.as_block_chain_client()).is_err() {
```
arkpar: The error here should not be interpreted as permission failure. Other checks might be added in the future. Best would just to pass it to the caller:
```rust
self.engine.machine().verify_transaction(&transaction, &best_block_header, client.as_block_chain_client())?;
```

VladLupashevskyi:
@arkpar Agreed. And thank you, my Rust knowledge is much better now :) I didn't know about this amazing "?" operator before.

PR #7366, opened 2017-12-22
added a new RPC call trace_replayBlockTransactions

tzapu:
in relation to https://github.com/paritytech/parity/issues/7365

as i mentioned this is my first go at rust so I could do with some pointers about error checking or tests, or whatever else is needed to make this mergeable

thank you guys for everything


tzapu:
[clabot:check]

ethcore/src/client/client.rs
```diff
@@ -1318,6 +1318,25 @@ impl BlockChainClient for Client {
...
+		let mut env_info = self.env_info(block).ok_or(CallError::StatePruned)?;
+		let body = self.block_body(block).ok_or(CallError::StatePruned)?;
+		let mut state = self.state_at_beginning(block).ok_or(CallError::StatePruned)?;
+		let txs = body.transactions();
+
+		const PROOF: &'static str = "Transactions fetched from blockchain; blockchain transactions are valid; qed";
+		const EXECUTE_PROOF: &'static str = "Transaction replayed; qed";
+		for t in txs {
+			let t = SignedTransaction::new(t).expect(PROOF);
+			let x = self.do_virtual_call(&env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
```
tomusdrw: You can just do `self.do_virtual_call(..)?` here. But you are right that the call should never error here, so I'm fine with expect as well.

ethcore/src/client/test_client.rs
```diff
@@ -429,6 +429,12 @@ impl BlockChainClient for TestBlockChainClient {
 		self.execution_result.read().clone().unwrap()
 	}
 
+	// TODO add test
+	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Vec<Executed>, CallError> {
+		let res = Vec::with_capacity(0);
```
tomusdrw: Leave this as `unimplemented!()`

ethcore/src/client/client.rs
```diff
@@ -1318,6 +1318,25 @@ impl BlockChainClient for Client {
 		self.do_virtual_call(&env_info, &mut state, &t, analytics)
 	}
 
+	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Vec<Executed>, CallError> {
```
tomusdrw: To share the code with `replay()` you could return a `Result<Box<Iterator<Item = Executed>>, CallError>` here and then `replay()` just becomes:
```
let index = <get-tx-index>;
Ok(self.replay_block(...)?.nth(index))
```

tzapu:
@tomusdrw thank you for taking at look at this

i tried implementing the suggested changes, but my non existent rust knowledge got the best of me
i have no idea how i would ofrmulate this to work
```
	fn replay_block_transactions(&self, block_number: BlockNumber, flags: TraceOptions) -> Result<Box<Iterator<Item = TraceResults>>> {
		self.client.replay_block_transactions(block_number.into(), to_call_analytics(flags))
			.map(|results| results.into_iter().map(TraceResults::from).collect())
			.map_err(errors::call)
	}
```
with the neew result, or if the rest of the changes are ok
thank you again, maybe you can give me some tips

tzapu:
i think i may have sorted it
while testing, if i make the request for a not synced yet block i get `This request is not supported because your node is running with state pruning. Run with --pruning=archive.` which is not quite accurate.
How would i go changing that to something more appropriate?

ethcore/src/client/client.rs
```diff
@@ -1293,7 +1293,7 @@ impl BlockChainClient for Client {
 		trace!(target: "estimate_gas", "estimate_gas chopping {} .. {}", lower, upper);
 		binary_chop(lower, upper, cond)
 	}
-
+/*
```
tomusdrw: Please remove commented out code.

ethcore/src/client/client.rs
```diff
@@ -1318,6 +1318,33 @@ impl BlockChainClient for Client {
 		self.do_virtual_call(&env_info, &mut state, &t, analytics)
 	}
 
+*/
+	fn replay(&self, id: TransactionId, analytics: CallAnalytics) -> Result<Executed, CallError> {
+		let address = self.transaction_address(id).ok_or(CallError::TransactionNotFound)?;
+		let block = BlockId::Hash(address.block_hash);
+
+		Ok(self.replay_block_transactions(block, analytics)?.nth(address.index).unwrap())
```
tomusdrw: Instead of `unwrap` use `expect(<proof>)`. Here the proof will be something like: "The transaction address contains a valid index within block; qed".

ethcore/src/client/client.rs
```diff
@@ -1318,6 +1318,33 @@ impl BlockChainClient for Client {
...
+		const PROOF: &'static str = "Transactions fetched from blockchain; blockchain transactions are valid; qed";
+		const EXECUTE_PROOF: &'static str = "Transaction replayed; qed";
+		for t in txs {
+			let t = SignedTransaction::new(t).expect(PROOF);
+			let x = self.do_virtual_call(&env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
+			env_info.gas_used = env_info.gas_used + x.gas_used;
+			replays.push(x);
+		}
+
+		Ok(Box::new(replays.into_iter()))
```
tomusdrw: It's almost fine - right now you are executing all transactions and putting the results into a vector, but `Iterator` allows you to do it lazily (which will make `replay` work faster).
It should look like this (instead of a loop you should use `map`):
```
let txs = block.transactions();
Ok(Box::new(txs.into_iter()
  .map(|tx| {
     ...
     self.do_virtual_call(...).expect(EXECUTE_PROOF)
   })))
```

ethcore/src/client/test_client.rs
```diff
@@ -429,6 +429,11 @@ impl BlockChainClient for TestBlockChainClient {
 		self.execution_result.read().clone().unwrap()
 	}
 
+	// TODO add test
```
tomusdrw: Please remove.

tomusdrw:
Could you also prepare a mocked RPC test?

tzapu:
again, i would like to thank you for all the hand holding
i have added a naive rpc test, it looks like all the test pass

tzapu:
again, i would like to thank you for all the hand holding
i have added a naive rpc test, it looks like the tests pass

rpc/src/v1/traits/traces.rs
```diff
@@ -56,5 +56,9 @@ build_rpc_trait! {
 		/// Executes the transaction with the given hash and returns a number of possible traces for it.
 		#[rpc(name = "trace_replayTransaction")]
 		fn replay_transaction(&self, H256, TraceOptions) -> Result<TraceResults>;
+
+		/// Executes the transaction with the given hash and returns a number of possible traces for it.
```
andresilva: `Executes all the transactions at the given block and returns a number of possible traces for each transaction.`

ethcore/src/client/traits.rs
```diff
@@ -196,6 +196,9 @@ pub trait BlockChainClient : Sync + Send {
 	/// Replays a given transaction for inspection.
 	fn replay(&self, t: TransactionId, analytics: CallAnalytics) -> Result<Executed, CallError>;
 
+	/// Replays a given transaction for inspection.
```
andresilva: `Replays all the transactions in a given block for inspection.`

tzapu:
thanks guys, much appreciated, happy new year
updated

tzapu:
thanks guys, for everything

ankitchiplunkar:
Would it be possible to also add uncle rewards and miner rewards with this JSON RPC?

PR #7387, opened 2017-12-27
Fix default CORS settings.

tomusdrw:
- Expect a list in the config file.


parity/cli/mod.rs
```diff
@@ -466,7 +466,7 @@ usage! {
 			"--jsonrpc-threads=[THREADS]",
 			"Turn on additional processing threads in all RPC servers. Setting this to non-zero value allows parallel cpu-heavy queries execution.",
 
-			ARG arg_jsonrpc_cors: (Option<String>) = None, or |c: &Config| otry!(c.rpc).cors.clone(),
+			ARG arg_jsonrpc_cors: (String) = "none", or |c: &Config| otry!(c.rpc).cors.as_ref().map(|vec| vec.join(",")),
 			"--jsonrpc-cors=[URL]",
 			"Specify CORS header for JSON-RPC API responses.",
```
andresilva: Maybe add: "Special options: \"all\", \"none\"."

parity/cli/mod.rs
```diff
@@ -538,7 +538,7 @@ usage! {
 			"--ipfs-api-hosts=[HOSTS]",
 			"List of allowed Host header values. This option will validate the Host header sent by the browser, it is additional security against some attack vectors. Special options: \"all\", \"none\".",
 
-			ARG arg_ipfs_api_cors: (Option<String>) = None, or |c: &Config| otry!(c.ipfs).cors.clone(),
+			ARG arg_ipfs_api_cors: (String) = "none", or |c: &Config| otry!(c.ipfs).cors.as_ref().map(|vec| vec.join(",")),
 			"--ipfs-api-cors=[URL]",
 			"Specify CORS header for IPFS API responses.",
```
andresilva: Same as above regarding special options.

parity/cli/mod.rs
tomusdrw: Good point, thanks! Done.

PR #7397, opened 2017-12-28
SecretStore: do not send initialization request if session has finished/failed instantly

svyatonik:
Previously, during decryption session (also signing session):
1) initialization request has been sent to all other nodes;
2) self-response has been processed => session could finish/fail instantly (if single nodes is enough - i.e. threshold = 0)
3) session on master node is removed
4) 'slave' nodes are receiving init request && session is initialized on these nodes
5) after ~60s session on slave nodes is timeouted

After this PR, before sending init request to other nodes, we first try to complete the session on single node.


svyatonik:
reported by @grbIzl

Issue #7144, opened 2017-11-26
Whisper problem: PoW too low to compete with other messages (and potential source of error)


zot:
I'm running parity 1.8.3 on Linux, testing with both an installed release and a release built from source on a dev network. My Dapp is using web3.js 1.0.0-beta.18.

When I post a whisper message, I always get the error: "PoW too low to compete with other messages". I'm using using priority 10000 and ttl 100 (I've tested with many variations).

I added some debug logs and got an inconsistent result which I think may be linked to the problem.

`Whisper::post()` will return `Err(whisper_error("PoW too low to compete with other messages"))` if `relay()` returns `false`.  `Relay` is bound to `PoolHandle::relay()` in [whisper.rs, line 51](https://github.com/paritytech/parity/blob/master/parity/whisper.rs#L51) but relay seems to be behaving oddly, based on some debug logs I put in.  Here is the change I made:

```rust
fn relay(&self, message: Message) -> bool {
	let mut res = false;
	let mut message = Some(message);
	debug!(target: "whisper", "RELAYING...");
	self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut move |ctx| {
		if let Some(message) = message.take() {
			res = self.handle.post_message(message, ctx);
			debug!(target: "whisper", "POSTED MESSAGE, RES: {}", res);
		} else {
			debug!(target: "whisper", "NO MESSAGE POSTED");
        	}
	});
	debug!(target: "whisper", "FINISHED RELAYING, RES: {}", res);
	res
}
```
This produces the following log messages:
> 2017-11-26 13:53:01   DEBUG whisper  RELAYING...
> 2017-11-26 13:53:01   DEBUG whisper  POSTED MESSAGE, RES: true
> 2017-11-26 13:53:01   DEBUG whisper  FINISHED RELAYING, RES: false

The `with_proto_context block()` sets `res` to `true` but outside of the block, `res` is still `false`. This returns `false` to `Whisper::post()` and, I believe, results in the (wrongly reported) error.

yrashk:
@zot I think your assessment is spot on. The reason `relay` compiles is because `bool` implements Copy and that's what happens within the move semantics. The value just gets copied.

rphmeier:
Technically it's just an RPC error if the messages actually did get posted to the network. But nice catch and fix regardless!

PR #7421, opened 2018-01-02
Problem: sending any Whisper message fails

yrashk:
The error is "PoW too low to compete with other messages"

This has been previously reported in #7144

Solution: prevent the move semantics

The source of the error is in PoolHandle.relay
implementation for NetPoolHandle.

Because of the move semantics, `res` variable is in fact
copied (as it implements Copy) into the closure and for
that reason, the returned result is always `false.

---

I am not sure what's the best way to cover this with a test. Any suggestions?

### DIFF
--- a/parity/whisper.rs
+++ b/parity/whisper.rs
@@ -51,7 +51,7 @@ impl PoolHandle for NetPoolHandle {
 	fn relay(&self, message: Message) -> bool {
 		let mut res = false;
 		let mut message = Some(message);
-		self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut move |ctx| {
+		self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut |ctx| {
 			if let Some(message) = message.take() {
 				res = self.handle.post_message(message, ctx);
 			}


debris:
cc @rphmeier

parity/whisper.rs
```diff
@@ -51,7 +51,7 @@ impl PoolHandle for NetPoolHandle {
 	fn relay(&self, message: Message) -> bool {
 		let mut res = false;
 		let mut message = Some(message);
-		self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut move |ctx| {
+		self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut |ctx| {
```
rphmeier: that's a tricky one. no wonder the tests in the whisper module didn't catch it.

rphmeier: I am a little disappointed that rustc doesn't give you a warning for unused_mut when the variable is only copied into a closure.

Minimal playpen: https://play.rust-lang.org/?gist=c66bc6098aeacddb71550878c0d887f9&version=stable

rphmeier: https://github.com/rust-lang/rust/issues/47128

rstormsf:
Go https://POA.network Great work

ensrationis:
@rstormsf Thanks for alert :) We will try.

rstormsf:
@ensrationis actually we have found more issues with Whisper in parity. False alarm :-(

PR #7422, opened 2018-01-02
Fix whisper response.

### LINKED ISSUE #7144: Whisper problem: PoW too low to compete with other messages (and potential source of error)

tomusdrw:
Resolves #7144

### DIFF
--- a/parity/whisper.rs
+++ b/parity/whisper.rs
@@ -51,11 +51,14 @@ impl PoolHandle for NetPoolHandle {
 	fn relay(&self, message: Message) -> bool {
 		let mut res = false;
 		let mut message = Some(message);
-		self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut move |ctx| {
-			if let Some(message) = message.take() {
-				res = self.handle.post_message(message, ctx);
-			}
-		});
+		{
+			let res = &mut res;
+			self.net.with_proto_context(whisper_net::PROTOCOL_ID, &mut move |ctx| {
+				if let Some(message) = message.take() {
+					*res = self.handle.post_message(message, ctx);
+				}
+			});
+		}
 		res
 	}
 


tomusdrw:
Just noticed #7421.

PR #7425, opened 2018-01-02
Use RwLock for state DB

tomusdrw:
To allow parallel reads.

Had to introduce `Mutex` in state_db_cache though (since `Account` has `Cell/RefCell` inside which makes it `!Sync`). However the `Mutex` is guaranteed to be accessed from single thread (withing `sync_cache` method) and actually `Account` is never modified (not even read from) in that context, it should be safe to just wrap it with something that provides `unsafe Sync`, but IMHO better be safe with this.


ethcore/src/state_db.rs
```diff
@@ -58,7 +58,7 @@ struct CacheQueueItem {
 	/// Account address.
 	address: Address,
 	/// Acccount data or `None` if account does not exist.
-	account: Option<Account>,
+	account: Mutex<Option<Account>>,
```
arkpar: Why is there a mutex required here? Not sure that creating thousands of mutexes is a good idea.

arkpar:
Parallel reads were actually possible before this change. The mutex was never held for the actual read but only to make a clone.

tomusdrw:
@arkpar Let me know if this is any better: https://github.com/paritytech/parity/pull/7456
if you are not happy with it we can revert the PR.

> The mutex was never held for the actual read but only to make a clone.
Indeed that was the case although IMHO it's still worth to relax the requirements by using RwLock.

PR #7426, opened 2018-01-02
Fix final feature.

tomusdrw:


debris:
it's a first time I see this. Isn't `final` and `unstable` providing unnecessary confusion? We already have `stable/beta/nightly`

tomusdrw:
We want to distinguish between versions prepared by us and self-built. Every self-built version is `unstable`, otherwise we use `final` stuff.

Issue #5689, opened 2017-05-24
Stage 1 block verification failed for 5c00…6234: Block(DifficultyOutOfBounds(OutOfBounds { min: None, max: Some(340282366920938463463374607431768211455), found: 340282366920938463463374607431768211456 }))


Jacob-H5:
Hi, I have a private net with  2 sealer and 4 non-sealing nodes  (6 nodes). The non-sealing nodes erorred with the following:

```
Stage 1 block verification failed for 5c00…6234: Block(DifficultyOutOfBounds(OutOfBounds { min: None, max: Some(340282366920938463463374607431768211455), found: 340282366920938463463374607431768211456 }))
```

and lost connection two minutes later. After around 15 minutes they reconnected but didn't import any blocks. Deleted the chainfolder on one of the nodes and tried to resync, but the same error occured.

here is my spec:

```
{
  "name": "POA",
  "engine": {
    "authorityRound": {
      "params": {
        "gasLimitBoundDivisor": "0x400",
        "maxCodeSize": 98304,
        "stepDuration": "1",
        "validators" : {
          "list": [
            "0x0063608C83BAEA08A4b3d33a56C82a63e3e39D6D",
            "0x23ce2f32d13b0e7ef15f93367eb22aa3d7d529e0"
          ]
        }
      }
    }
  },
  "params": {
    "maximumExtraDataSize": "0x20",
    "minGasLimit": "0x1999999999999998",
    "networkID" : "0x76C"
  },
  "genesis": {
    "seal": {
      "authorityRound": {
        "step": "0x0",
        "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      }
    },
    "difficulty": "0x20000",
    "gasLimit": "0x1999999999999998"
  },
  "accounts": {
    "0x0000000000000000000000000000000000000001": { "balance": "1", "builtin": { "name": "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
    "0x0000000000000000000000000000000000000002": { "balance": "1", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
    "0x0000000000000000000000000000000000000003": { "balance": "1", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
[...]

PR #7451, opened 2018-01-04
Advance AuRa step as far as we can and prevent invalid blocks.

### LINKED ISSUE #7441: Parity PoA authorities are not syncing
_**Authority sync not work correct**._

I'm running a private parity poa network with verion 1.7.7 stable.
The following issue I get:

> Stage 1 block verification failed for b4e8…ee34: Block(DifficultyOutOfBounds(OutOfBounds { min: None, max: Some(340282366920938463463374607431768211455), found: 340282366920938463463374607431768211456 }))

and after that the two atuhorities are not able to sync again correct.
In the following screenshot you can see that on of the authorities imports higher blocknumbers and than back again lower blocknumber (the lower numbers are the same like at the other authority and the hash values are correct too). But it never stops to do add first the wrong blocknumber and then the correct one. So my app doesn't work correct!
"wrong authority"
image

"working authority"
image

All other none authority nodes got the same blocknumber as in the second pic.

Thank you in advance for some help!

Best

Fatih

tomusdrw:
Currently the step might be lagging behind since we are relying on perfect timing of `timeout` invocation. In case event loop is busy doing some other stuff this might not happen and the step will off by one.

Since we are accepting a future-step blocks (window of 1), sometimes this may trigger sealing a new block with `step < parent_step` especially if timeout did not run.

The change introduces a check for that in `generate_seal` and also attempts to progress the `step` as far as possible by looping inside `timeout` callback.

Also the timeout is now triggered more often, but I'm not super convinced it improves anything - let me know.

Resolves #7441 #5689


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -474,9 +474,15 @@ impl IoHandler<()> for TransitionHandler {
...
-				let remaining = engine.step.duration_remaining();
-				io.register_timer_once(ENGINE_TIMEOUT_TOKEN, remaining.as_millis())
+				// NOTE we might be lagging by couple of steps in case the timeout
+				// has not been called fast enough.
+				// Make sure to advance up to the actual step.
+				while engine.step.duration_remaining().as_millis() == 0 {
+					engine.step();
+				}
+
+				let next_run_at = engine.step.duration_remaining().as_millis() >> 2;
```
andresilva: Not sure if this is too much.

tomusdrw:
@ddorgan did some tests with this PR and it seems the warnings are still produces, so there are still some circumstances under which invalid block is produced, but prevented by the check in this PR.

rphmeier:
Wouldn't it be better to just delay blocks offset by one step as well?

tomusdrw:
@rphmeier Fixed. I removed the acceptance threshold for future blocks.

PR #7453, opened 2018-01-04
Add `personal_sign` and `personal_ecRecover` RPC methods

### LINKED ISSUE #4460: Add `personal_sign` and `personal_recover`
Details:
https://github.com/ethereum/go-ethereum/pull/2940

andresilva:
Fixes #4460.

### DIFF
--- a/rpc/src/v1/impls/personal.rs
+++ b/rpc/src/v1/impls/personal.rs
@@ -62,13 +62,13 @@ impl<D: Dispatcher> PersonalClient<D> {
 }
 
 impl<D: Dispatcher + 'static> PersonalClient<D> {
-	fn do_sign(&self, address: RpcH160, data: RpcBytes, password: String) -> BoxFuture<RpcH520> {
+	fn do_sign(&self, data: RpcBytes, account: RpcH160, password: String) -> BoxFuture<RpcH520> {
 		let dispatcher = self.dispatcher.clone();
 		let accounts = try_bf!(self.account_provider());
 
-		let payload = RpcConfirmationPayload::EthSignMessage((address.clone(), data).into());
+		let payload = RpcConfirmationPayload::EthSignMessage((account.clone(), data).into());
 
-		Box::new(dispatch::from_rpc(payload, address.into(), &dispatcher)
+		Box::new(dispatch::from_rpc(payload, account.into(), &dispatcher)
 				 .and_then(|payload| {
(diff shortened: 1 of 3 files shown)


rpc/src/v1/impls/personal.rs
```diff
@@ -133,6 +174,14 @@ impl<D: Dispatcher + 'static> Personal for PersonalClient<D> {
 		}
 	}
 
+	fn sign(&self, data: RpcBytes, account: RpcH160, password: String) -> BoxFuture<RpcH520> {
+		self.do_sign(data, account, password)
+	}
+
+	fn ec_recover(&self, data: RpcBytes, signature: RpcH520) -> BoxFuture<RpcH160> {
+		self.do_ec_recover(data, signature)
```
tomusdrw: Why is it not implemented inline?

andresilva:
I tried to use cryptokitties with this and it made a call to `personal_sign` with only two arguments (missing password). According to https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_sign I think the password should be mandatory, but it shows the password as optional for the console API. Should I implement the same behavior, i.e. if no password is provided the account should already be unlocked (same as `eth_sign`).

rpc/src/v1/impls/personal.rs
andresilva: Inlined both.

andresilva:
I've confirmed that geth requires the 3 parameters to be present using the RPC API, so we should leave it as is. I'm not sure why metamask has implemented it with only 2 arguments though.

andresilva:
Can I rebase this to avoid the merges?

5chdn:
Yes, rebasing should work.

5chdn:
Please resolve conflicts :)

andresilva:
Rebased.

wliment:
@andresilva  i build from master branch source but `personal_ecRecover` method not found

```
curl --data '{"method":"personal_ecRecover","params":[ "Hello","0x190dc01f46fe8ad8ab57eeca71999329773d34d9eff000b91f6e4f947722868718e0bb85d396ea09f066ba0a46d4917ffa26f421833116abc4b62c8e91c5cd2e1b"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}
```

5chdn:
@wliment did you enable the personal jsonrpc module?

PR #7456, opened 2018-01-04
Avoid creating thousands of `Mutexes` in state db.

tomusdrw:
Addresses: https://github.com/paritytech/parity/pull/7425#discussion_r159543126


tomusdrw:
Yeah, although it doesn't really require `Mutex` at all and it could introduce some performance hit. The `Mutex` is only required because of how `Account` caching is implemented. Long term I think we could do something similar with Account as we did with transactions - so everytime something is cached we have a different type to denote that.

PR #7459, opened 2018-01-04
Use `parity-react-scripts` for Wallet v1

ngotchac:
Use the `parity-react-script`package for `js-old` folder, ie. the v1 wallet.
The next steps would be to convert all the other dapps to use it as well, and change the way the UI gets build (would be simpler to just download the releases ZIP files, and put it in the build). The v1 wallet should IMO also be a dapp in a separate repo, not sure why it is still there.

[it actually deletes +/- 1,500 lines and adds +/- 309 lines ; numbers are big because of the lock files]

---
This change is [Reviewable](https://reviewable.io/reviews/paritytech/parity/7459)


jacogr:
100% - v1 shouldn't really be in this repo anymore. (Remember it started as a "stopgap".) However, at this point in the 1.9 release cycle, I don't want to split it out.

ngotchac:
Tests are failing because of linting issues : updated linting dependencies so some rules changed.
There is a separate PR to fix linting issues here: https://github.com/paritytech/parity/pull/7468

jacogr:
Looks good. Let's just on-ice this until 1.9 is out. As discussed, I'm a bit cautious about making major changes this close to release. (Although technically, this should not affect anything, just want to keep the non-bug-fix changes to an absolute minimum)

5chdn:
Please resolve conflicts :)

jacogr:
Can we re-open on https://github.com/Parity-JS/dapp-wallet

PR #7468, opened 2018-01-05
Fix JS linting issues in `js-old`

ngotchac:


jacogr:
Shouldn't this be part of the react-script PR?

jacogr:
Looks correct. Only issues -

1. Cannot minify `ethereumjs-util` (see CI)
2. Would rather have it in the react-scripts PR instead of merging, having master not work and then merging this one.

ngotchac:
1. Yep, fixing this issue right now.
2. This PR is against the other PR's branch. I just wanted not to mix up automatic linting fixes and code changes for the PR review.

PR #7460, opened 2018-01-04
dissolve util

debris:
This is the last part of major cleanup of util. closes #6693
It replace old bigint crate with [ethereum-types](https://github.com/paritytech/primitives). Standalone crate which can be used by all ethereum-like projects

Issue #6249, opened 2017-08-07
Unify bloom type


debris:
todo: 

- [x] update `bloomchain` crate from `0.1` to `0.2`
  - function `contains` has been renamed to `contains_bloom`
- [x] build blooms using `ethbloom 0.2`
- [x] use `ethbloom::Bloom` instead of `Bloom`/`LogBloom`/whatever else we use now
- [x] remove `util/bloomable`

benchmarks:
- `bloomchain 0.2` is 25% faster than `0.1`
- `ethbloom` benches:
  ```
  test accrue_hash                  ... bench:          19 ns/iter (+/- 3)
  test contains_hash                ... bench:         345 ns/iter (+/- 20)
  test does_not_contain_hash        ... bench:         145 ns/iter (+/- 38)
  test does_not_contain_random_hash ... bench:      22,577 ns/iter (+/- 6,791)
  ```

- `parity`, benches - branch [bloomable_benches](https://github.com/paritytech/parity/compare/bloomable_benches?expand=1)
  
  ```
  test accrue_hash                  ... bench:         368 ns/iter (+/- 31)
  test contains_hash                ... bench:         371 ns/iter (+/- 81)
  test does_not_contain_hash        ... bench:         373 ns/iter (+/- 79)
  test does_not_contain_random_hash ... bench:      48,444 ns/iter (+/- 13,557)
  ```

`accrue_hash` - accrues 2 hashes
`contains_hash` - checks if bloom contains 2 hashes (positive)
`test_does_not_contain_hash` - checks if bloom contains 2 hashes (negative)
`test does_not_contain_random_hash` checks if bloom contains 256 hashes (negative)

links:
- [ethbloom](https://crates.io/crates/ethbloom)
- [bloomchain](https://crates.io/crates/bloomchain)

stankec:
Hey! I've started working on this but got a bit confused.

I've updated bloomchain and added ethbloom and started switching out Bloom and LogBloom for ethbloom::Bloom. Though I'm not sure if that should be done to the Bloom type from `ethcore/src/basic_types.rs` (that uses the definition from `ethcore/types/src/log_entry.rs`)

My changes so far - https://github.com/stankec/parity/commits/feature/unify_bloom_type

debris:
[...]

PR #7475, opened 2018-01-05
bloom refactor

debris:
on top of #7460 (only to resolving tens of conflicts later)

related pr to primitives: https://github.com/paritytech/primitives/pull/9
- moves https://github.com/debris/ethbloom to the primitives repo

this pr:
- moves https://github.com/paritytech/bloomchain to parity repo
- fixes all points of #6249
  
it should speed up syncing a bit (cause building blooms is 20x faster)


util/bloomchain/src/position/manager.rs
```diff
@@ -0,0 +1,142 @@
...
+/// Simplifies working with bloom indexes.
+pub struct Manager {
+	index_size: usize,
+	level_sizes: Vec<usize>,
+}
+
+impl Manager {
+	/// Creates new indexer.
+	pub fn new(index_size: usize, levels: usize) -> Self {
+		if levels == 0 {
```
svyatonik: maybe `assert!(levels != 0, "Manager requires at least 1 level.");`

util/bloomchain/src/group/position/manager.rs
```diff
@@ -0,0 +1,28 @@
...
+	pub fn new(index_size: usize) -> Self {
+		Manager {
+			index_size: index_size
+		}
+	}
+
+	pub fn group_position(&self, pos: &BloomPosition) -> GroupPosition {
+		GroupPosition {
+			level: pos.level,
+			index: pos.index / self.index_size,
```
svyatonik: Still trying to get the whole picture - is there some check if `index_size` is not a zero? Does it worth adding a check/assert to this library? Or should it be managed outside?

debris: so it's exactly the same library and logic we use today and we've been using for last 2 years. `index_size` is a constant set in ethcore and is always 16. But you are right, there should be an assertion here :/

util/bloomchain/src/chain.rs
```diff
@@ -0,0 +1,158 @@
...
+	fn blocks(&self, range: &Range<Number>, bloom: &Bloom, level: usize, offset: usize) -> Option<Vec<usize>> {
+		let index = self.positioner.position(offset, level);
+
+		match self.db.bloom_at(&index) {
+			None => return None,
+			Some(level_bloom) => match level {
+				// if we are on the lowest level
+				0 if level_bloom.contains_bloom(bloom) => return Some(vec![offset]),
+				// return None if current level doesnt contain given bloom
+				_ if !level_bloom.contains_bloom(bloom) => return None,
```
svyatonik: Maybe swap with first arm (`0 if ...`) and leave single `contains_bloom` call? Like:
`
match level {
 _ if !level_bloom.contains(bloom) => ...
 0 => ...
 _ => ...
}
`

util/bloomchain/src/group/position/manager.rs
svyatonik: Don't bother with this then - I'm just checking for possible issues. If that is not an issue - just ignore, pelase

PR #7497, opened 2018-01-08
Expose default gas price percentile configuration in CLI

tomusdrw:


parity/cli/mod.rs
```diff
@@ -1546,6 +1552,7 @@ mod tests {
 			arg_tx_time_limit: Some(100u64),
 			arg_relay_set: "cheap".into(),
 			arg_min_gas_price: Some(0u64),
+			arg_gas_price_percentile: 50usize,
```
andresilva: Maybe override the default for the test in `tests/config.full.toml`.

5chdn:
Are we sticking with 50 as default for now?

util/stats/src/lib.rs
```diff
@@ -121,7 +133,19 @@ impl<T: Ord + Copy + ::std::fmt::Display> Histogram<T>
...
+	use super::*;
+
+	#[test]
+	fn check_corpus() {
+		let corpus = Corpus::from(vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
+		assert_eq!(corpus.percentile(0), None);
+		assert_eq!(corpus.percentile(1), None);
+		assert_eq!(corpus.percentile(101), Some(&10));
+		assert_eq!(corpus.percentile(100), Some(&10));
+		assert_eq!(corpus.percentile(50), Some(&5));
```
andresilva: I think for a bigger data set this should be the same as median.

tomusdrw: We have even number of elements here.

PR #7503, opened 2018-01-09
Use https connection

computerality:
Use https when connecting to etherscan.io API for price-info


computerality:
[clabot:check]

PR #7507, opened 2018-01-09
bump to 1.10.0

debris:
- bumped only absolutely necessary files to `1.10.0`
- also moved current `master` to `staging` branch ([link](https://github.com/paritytech/parity/tree/staging))

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1903,7 +1903,7 @@ dependencies = [
 
 [[package]]
 name = "parity"
-version = "1.9.0"
+version = "1.10.0"
 dependencies = [
  "ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "app_dirs 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


debris:
> There used to be this in earlier releases:
>
> - [ ] util/Cargo.toml (version key)
> - [ ] cargo update -p ethcore-util
> 
> Is this parity-version now?

yes, but it's irrelevant. It's only about compiler version. Cargo.tomls' versions other than the one in the main `Cargo.toml` are never used.

5chdn:
> yes, but it's irrelevant. It's only about compiler version. Cargo.tomls' versions other than the one in the main Cargo.toml are never used.

`util/version/Cargo.toml` reads:

    # NOTE: this value is used for Parity version string.
    version = "1.9.0"

This should be bumped. I just don't know how since we changed it. cc @tomusdrw

debris:
@5chdn you are right, I've updated this version and comment. Link to the [docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

5chdn:
"looks good" but it's not working yet.

Issue #7502, opened 2018-01-08
current master does not build on windows


Trolldemorted:
> I'm running:
>
> - **Which Parity version?**: bdc3d6fee049111e54802cc132488270e2458036
> - **Which operating system?**: W10 Pro N 64bit
> - **Which rustc version?**: rustc 1.21.0 (3b72af97e 2017-10-09)

I just cloned the repo, executed cargo build --release and the linker fails because it cannot find the symbol `rocksdb_options_set_disable_data_sync`.

5chdn:
Ouch. On it.

Trolldemorted:
While we are at it: you might want to setup continuous ingetration builds for all platforms - e.g. appveyor should be free for opensource projects.

5chdn:
We are using gitlab and it's running builds for all platforms.

https://gitlab.parity.io/parity/parity/pipelines/17021

5chdn:
https://github.com/paritytech/rust-rocksdb/pull/13

5chdn:
https://github.com/spacejam/rust-rocksdb/pull/139

5chdn:
confirmed working. can be closed once this is forward-ported to staging and master.

PR #7508, opened 2018-01-09
[Beta] RocksDB fix

andresilva:


5chdn:
we need this also for master and staging now

closes #7502

PR #7533, opened 2018-01-11
Make 1.9 beta

5chdn:
cc @jacogr - how to handle the UI stuff for 1.9 beta vs. 1.10 nightly?


jacogr:
@5chdn It _should_ just magically work once the branch has moved. (The required repos is created and waiting on `github.com/js-dist-paritytech`, however we would just need to trigger a JS build once the branch replaces beta)

PR #7539, opened 2018-01-11
[WASM] mem_cmp added to the Wasm runtime

fckt:
In order to support wasm-unknown-unknown target runtime has to provide `memcmp` which previously was provided by emscripten


ethcore/wasm/src/runtime.rs
```diff
@@ -567,6 +568,28 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
+		//
+		// method signature:
+		//   fn memcmp(cx: *const c_void, ct: *const c_void, n: usize) -> i32;
+		//
+
+		let len = context.value_stack.pop_as::<i32>()? as u32;
+		let cx = context.value_stack.pop_as::<i32>()? as u32;
+		let ct = context.value_stack.pop_as::<i32>()? as u32;
+
+		self.charge(|schedule| schedule.wasm.mem_copy as u64 * len as u64)?; // TODO: charge cmp
```
NikVolf: todo should be addressed

ethcore/wasm/src/runtime.rs
```diff
@@ -567,6 +568,28 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
+		let len = context.value_stack.pop_as::<i32>()? as u32;
+		let cx = context.value_stack.pop_as::<i32>()? as u32;
+		let ct = context.value_stack.pop_as::<i32>()? as u32;
+
+		self.charge(|schedule| schedule.wasm.mem_cmp as u64 * len as u64)?;
+
+		let cx = self.memory.get(cx, len as usize)?;
+		let ct = self.memory.get(ct, len as usize)?;
+		let result = unsafe {
+			memcmp(cx.as_ptr() as *const c_void, ct.as_ptr() as *const c_void, len as usize)
```
NikVolf: memcmp should fail when invoked on overlapping region, because otherwise it's undefined behaviour?

NikVolf: oh, on the second thought, it shouldn't be

ethcore/wasm/src/runtime.rs
```diff
@@ -567,6 +568,28 @@ impl<'a, 'b> Runtime<'a, 'b> {
 		&*self.memory
 	}
 
+	fn mem_cmp(&mut self, context: InterpreterCallerContext)
+		-> Result<Option<interpreter::RuntimeValue>, InterpreterError>
+	{
+		//
+		// method signature:
+		//   fn memcmp(cx: *const c_void, ct: *const c_void, n: usize) -> i32;
```
NikVolf: https://github.com/paritytech/pwasm-std/blob/master/libc/src/lib.rs#L8

should match here

PR #7548, opened 2018-01-13
Replace `unwrap()` in UntrustedRlp Display impl

lght:
Use `match { ... }` in place of `unwrap()`.

### DIFF
--- a/util/rlp/src/untrusted_rlp.rs
+++ b/util/rlp/src/untrusted_rlp.rs
@@ -114,18 +114,29 @@ impl<'a> Clone for UntrustedRlp<'a> {
 	}
 }
 
-impl<'a> fmt::Display for UntrustedRlp<'a> {
-	fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-		match self.prototype() {
-			Ok(Prototype::Null) => write!(f, "null"),
-			Ok(Prototype::Data(_)) => write!(f, "\"0x{}\"", self.data().unwrap().to_hex()),
-			Ok(Prototype::List(len)) => {
-				write!(f, "[")?;
-				for i in 0..len-1 {
-					write!(f, "{}, ", self.at(i).unwrap())?;
-				}
-				write!(f, "{}", self.at(len - 1).unwrap())?;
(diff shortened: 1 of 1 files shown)


util/rlp/src/untrusted_rlp.rs
```diff
@@ -114,18 +114,29 @@ impl<'a> Clone for UntrustedRlp<'a> {
...
+			Ok(Prototype::Data(_)) => {                                                                                         
+				match self.data() {                                                                                                         
+					Ok(rlp) => write!(f, "\"0x{}\"", rlp.to_hex()),                                                                         
+					Err(err) => write!(f, "{:?}", err)                                                                                      
+				}                                                                                                                           
+			},                                                                                                                              
+			Ok(Prototype::List(len)) => {                                                                                       
+				write!(f, "[")?;                                                                                            
+				for i in 0..len-1 {                                                                                         
+					let _ = match self.at(i) {
```
rphmeier: disregarding errors while writing is bad. why not simply use `expect` instead of rewriting? Or even rewrite using `for (i, item) in self.iter().enumerate()` in the list case?

lght: > why not simply use expect instead of rewriting?

The point of these fixes was to remove panics from UntrustedRlp handling, using `expect` would just reintroduce the same problem.

I was following the pattern from other error conditions in this section of code. Definitely open to  suggestions for how to handle the errors without panicking.

rphmeier: sure, then I would recommend using an iterator as it will be cleaner:

```rust
write!(f, "[")?;
for (i, item) in self.iter().enumerate() {
    write!(f, "{}", item)?;
}
write!(f, "]")?;
```

lght: I do like the conciseness of your solution. It does have the drawback of making the source of the error more obscure (now the `.at()` source is hidden behind the `.iter()` implementation).

Should there be a more specific error raised by `untrusted_rlp.at()` to make the error's origin more clear? Currently the error resolves to `RlpExpectedToBeData`, which doesn't point to the `.at()` implementation at all.

PR #7559, opened 2018-01-15
fix windows and centos builds

General-Beck:
windows installer `parity_$VER_installer.exe`
rpm package `parity_$VER_x86_64.rpm`
add timeout for coverage test's 10 min


5chdn:
Thanks a lot!

PR #7563, opened 2018-01-15
bump some of our core dependencies

debris:


debris:
@paritytech/ci there is something seriously broken with the CI, cause 5146bd01135f8ef0a6a2ba06fa429430cd3777fd didn't trigger build to start

kirushik:
Should be fixed by #7564, we've got our `grep` arguments wrong.
Please update to the latest master after that one is merged.

util/version/build.rs
```diff
@@ -33,5 +33,5 @@ fn main() {
 		pub fn rustc_version() -> &'static str {{
 			\"{}\"
 		}}
-	", rustc_version::version()).as_bytes()).unwrap();
+	", rustc_version::version().unwrap()).as_bytes()).unwrap();
```
tomusdrw: Why is it safe?

debris: it's not. neither of the `unwraps` here is safe. I will fix it

PR #7564, opened 2018-01-15
Rust change detections should do inverse grep

kirushik:


kirushik:
@paritytech/ci review anyone, please?
It's sorta-urgent, our testing pipeline is completely useless without this fix.

PR #7586, opened 2018-01-16
separated ethcore_migrations from ethcore

debris:
first of many pull requests which reduce the size of ethcore crate


ethcore/migrations/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "ethcore-migrations"
+version = "0.1.0"
+authors = ["debris <marek.kotewicz@gmail.com>"]
```
kirushik: Don't we use `authors = ["Parity Technologies <admin@parity.io>"]` everywhere?

debris: haha, true ;)

ethcore/src/lib.rs
```diff
@@ -154,7 +153,7 @@ mod cache_manager;
 mod blooms;
 mod basic_types;
 mod pod_account;
-mod state_db;
+pub mod state_db;
```
svyatonik: Probably worth moving to `pub mod` section (a few lines above)?

parity/main.rs
```diff
@@ -52,6 +52,7 @@ extern crate ethcore_devtools as devtools;
 extern crate ethcore_io as io;
 extern crate ethcore_light as light;
 extern crate ethcore_logger;
+extern crate ethcore_migrations;
```
tomusdrw: Maybe `extern crate ethcore_migrations as migrations;` to keep it consitent?

PR #7589, opened 2018-01-16
add ethkey & parity-evm to darwin pkg

General-Beck:
+new status icons


General-Beck:
https://gitlab.parity.io/parity/parity/-/jobs/75778

PR #7590, opened 2018-01-16
[Wasm] memcmp fix and test added

fckt:


ethcore/wasm/src/tests.rs
```diff
@@ -491,6 +491,59 @@ fn keccak() {
 	assert_eq!(gas_left, U256::from(81_067));
 }
 
+// memcmp test.
+#[test]
+fn memcmp() {
```
NikVolf: `reqrep_test!` macro is more suitable for this kind of tests

Issue #5893, opened 2017-06-21
libssl1.0.0 is not available anymore


jpritikin:
Do you plan to migrate to the [libssl 1.1.0f-3](https://packages.debian.org/stretch/libssl-dev) available in newly release debian stable?

5chdn:
ssl 1.1.0 has been around for a while, if dinosaurs like debian make the switch, we are probably good to do that, too :+1:

jpritikin:
Is this already fixed by PR #4260 ?

5chdn:
No, PR #4260 introduced this issue.

calvintam236:
Actually, can you [build and release static binary](https://chr4.org/blog/2017/03/15/cross-compile-and-link-a-static-binary-on-macos-for-linux-with-cargo-and-rust/) of parity?

iFA88:
If you wish to use parity on Debian 9 ( stretch ) you need download the package from https://packages.debian.org/jessie/libssl1.0.0 and install like `dpkg --install libssl1.0.0_1.0.1t-+deb8u6_amd64.deb` as root. For me works.

5chdn:
Or just build parity from source.

calvintam236:
@5chdm the problem is that stretch does not have this old version in package manager. how can you build one without dependency?

@iFA88 not ideal for production environment

5chdn:
@calvintam236 building from source just links the binary against any available libssl version.

This issue is that the precompiled binaries link against a certain version. But Ubuntu didn't make the switch to 1.1.x yet.... It's really hard to understand why. The only things I can recommend here is either stop using Debian-based systems or compile from source :)

calvintam236:
@5chdn that's why I suggest to release a static build binary - all dependencies are included the binary, so it can run on any Linux.

l0rb:
@calvintam236 We are running parity in a docker container for production, that solves all the issues with dependencies.

5chdn:
Could anyone check if our debian binaries/packages work?


- debian | x86_64 | [parity 1.7.10](http://d1h4xl4cr1h0mo.cloudfront.net/stable/x86_64-unknown-debian-gnu/parity) | 6019681788349877c1fbb591d62e7d8e
[...]

PR #7596, opened 2018-01-17
Fix libssl1.1 package name for debian

5chdn:
package is actually called `libssl1.1`

    openssl is already the newest version (1.1.0f-3+deb9u1).
    The following packages have unmet dependencies:
     parity : Depends: libssl1.1.0 (>= 1.1.0) but it is not installable

---

     $ apt search libssl1.1
    Sorting... Done
    Full Text Search... Done
    libssl1.1/stable,stable,now 1.1.0f-3+deb9u1 amd64 [installed]
      Secure Sockets Layer toolkit - shared libraries

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,7 +41,7 @@ linux-stable-debian:
     - stable
     - triggers
   script:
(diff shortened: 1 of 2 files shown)


5chdn:
Once and for all fixes #5893

PR #7603, opened 2018-01-17
Light client RPCs

tomusdrw:
Implement `eth_getCode` and `parity_registry` for light client.

Also resolve UI providers earlier when running in light client mode.


jacogr:
JS LGTM. (Just take care of the minor style issues - https://gitlab.parity.io/parity/parity/-/jobs/75937)

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -432,6 +413,23 @@ impl LightFetch {
 			Either::B(extract_transaction)
 		}))
 	}
+
+	fn send_requests<T, F>(&self, reqs: Vec<OnDemandRequest>, parse_response: F) -> BoxFuture<T> where
```
rphmeier: you might be able to make this more generic by accepting a `R: on_demand::RequestAdapter` and `F: FnOnce(R::Out) -> T`.

Note that this trait is implemented for `Vec<OnDemandRequest>`
A high level API would help you construct tuples with consistent back-references and then we actually save a few allocations on each request.

PR #7605, opened 2018-01-17
Aura: Broadcast empty step messages instead of creating empty blocks

andresilva:
Instead of creating empty blocks authorities broadcast a signed `EmptyStep(step, parent_hash)` message. Each authority accumulates the empty messages it receives and these are added to the seal when creating a block with transactions, or by creating an empty block if there are more than `n` accumulated messages. The accumulated messages are included in the seal as proof of the skipped steps. The `parent_hash` is used for validating the message on receipt (we want to make sure we're receiving the message from the expected proposer) and also to filter empty step messages when adding them to seal.

The chain score was updated to take into account the number of empty step messages, and all of the signers of empty messages are taken into account when checking for finality since they signal approval by the signer of the `parent_hash`.

Reward attribution was updated so that the block author and all of the signers of empty step messages are rewarded, this will lead to the same behavior as if those authorities had created empty blocks.

A `emptyStepsTransition` option was added that enables empty steps messages at the given block, the number of maximum empty steps is configured with `maximumEmptySteps`.

When creating a new block, as it flows through the engine pipeline we "guess" which `empty_steps` will be included in the seal. This is a bit racy but is guarded against by the block difficulty (if we receive an empty step message in the meantime the block will have a different score and we'll detect it in `generate_seal`). This can be improved by introducing a context associated with the block as it flows through the engine (#7840).

Demo: https://asciinema.org/a/tyZSXmDSuZtL4hPTt4OmZoKOf
_The demo says the block time should be 10 seconds when empty steps are enabled but it should actually be 12 seconds (5 empty steps + 1 empty block)._

_Edit: updated description_


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -597,12 +755,32 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
 				// only issue the seal if we were the first to reach the compare_and_swap.
 				if self.can_propose.compare_and_swap(true, false, AtomicOrdering::SeqCst) {
-					return Seal::Regular(vec![encode(&step).into_vec(), encode(&(&H520::from(signature) as &[u8])).into_vec()]);
+
+					// clear old `empty_steps` messages
+					self.empty_steps.lock().retain(|e| U256::from(e.step) > parent_step);
+					let fields = vec![
+						encode(&step).into_vec(),
+						encode(&(&H520::from(signature) as &[u8])).into_vec(),
+						::rlp::encode_list(&empty_steps).into_vec(),
```
andresilva: I'm having some trouble with this, for some reason when the other nodes sync the block they can't deserialize the seal, it actually fails validation as they see a larger number of fields in the seal than expected (3).

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -265,6 +267,67 @@ impl EpochManager {
...
+		let empty_step_rlp = rlp.at(1)?;
+
+		let step = empty_step_rlp.val_at(0)?;
+		let parent_hash = empty_step_rlp.val_at(1)?;
+
+		Ok(EmptyStep { signature, step, parent_hash })
+	}
+}
+
+pub fn empty_step_full_rlp(signature: &H520, empty_step_rlp: &Bytes) -> Bytes {
```
rphmeier: prefer `Vec<u8>` to bytes, and `&Bytes` == `&[u8]`

ethcore/src/engines/authority_round/mod.rs
rphmeier: check the `Engine::seal_fields` function is returning the right count

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -800,11 +982,17 @@ impl Engine<EthereumMachine> for AuthorityRound {
 				chain(hash).and_then(|header| {
 					if header.number() == 0 { return None }
 
-					let res = (hash, header.author().clone());
-					trace!(target: "finality", "Ancestry iteration: yielding {:?}", res);
+					if let Ok(signers) = header_signers(&header) {
+						let res = (hash, signers);
```
rphmeier: this will import into `RollingFinality` as though the empty step signers have signed on the hash of the current block, but it's actually wrong: the empty step signers have only signaled their approval of the parent header.

rphmeier: it will probably be easier to alter the iterator logic than to alter the rolling finality checker

ethcore/src/engines/authority_round/mod.rs
andresilva: I updated the `seal_fields` function to return the correct number of fields. When deserializing the other nodes see a variable number of fields which I think varies with the size of the `empty_steps` list.

> 2018-01-15 17:52:23 UTC IO Worker #3 WARN client  Stage 1 block verification failed for 50e0…c519: Block(InvalidSealArity(Mismatch { expected: 3, found: 11 }))

ethcore/src/engines/authority_round/mod.rs
andresilva: Yes that's true, the empty step signers only signed on the parent block. The reason I altered the rolling finality checker was to make sure it didn't return duplicate hashes.

rphmeier: I think the current alterations are good actually, just that we need to figure out the correct way to input. I think that's one of the pieces of code that's most important to have tests for, especially if we have entry merging logic

phahulin:
Hi guys, while this is still WIP, could you clarify what will this change bring to AuRa? How does it relate to `--force-sealing`?

rphmeier:
@phahulin For security, aura currently requires validators to broadcast empty blocks (i.e. `--force-sealing`) when it is their turn to produce. This change allows validators to instead broadcast a special signed message stating that they have skipped their turn, but indicate the best block known to them. `--force-sealing` will still be necessary to trigger the "empty step statement" creation process, but the chain will grow much more slowly when there are no transactions.

phahulin:
@rphmeier thanks.. Will this feature be optional or will it replace current force-sealing mechanism? I'm asking in part because at the present moment validators receive block rewards even when they seal empty blocks, will this change?

5chdn:
FYI merge master to get CI tests

rphmeier:
@General-Beck why push to this branch?

rphmeier:
So far LGTM. I think it's fine to leave it in Aura for now.
I just realized another thing we have to watch out for is the devp2p max packet size of ~16MB. If the block size passes that threshold then it won't be possible to relay.

As such this is probably only something to include once we have libp2p support instead.

phahulin:
@rphmeier @andresilva Could you explain in a bit more details how will this change current sealing of empty blocks? Will this be optional or replace force-sealing? Will validators still get reward if there are no transactions to sign or block sealing stops?

andresilva:
@phahulin This is optional, I've added a transition to enable empty steps support. The maximum number of empty steps that are allowed is configurable and when that threshold is reached the authorities will seal an empty block. The idea would be to use this with force-sealing, this way we would avoid the bloat of generating a lot of empty blocks while still maintaining the same security guarantees. Yes, it does affect rewards since we won't be producing as many empty blocks (assuming a comparison with force sealing).

rphmeier:
as far as rewards go we could also change the logic s.t. block rewards are also given to anyone whose empty step message is included in the block.

andresilva:
@phahulin I've implemented @rphmeier's suggestion for block rewards, so it will now behave similarly wrt reward attribution as it does currently.

phahulin:
@andresilva thank you for the video!

andresilva:
@rphmeier I think this is ready to be reviewed.

ethcore/src/engines/authority_round/finality.rs
```diff
@@ -52,28 +52,31 @@ impl RollingFinality {
...
 	/// Fails if any provided signature isn't part of the signers set.
 	pub fn build_ancestry_subchain<I>(&mut self, iterable: I) -> Result<(), UnknownValidator>
-		where I: IntoIterator<Item=(H256, Address)>
+		where I: IntoIterator<Item=(H256, Vec<Address>)>
 	{
 		self.clear();
-		for (hash, signer) in iterable {
-			if !self.signers.contains(&signer) { return Err(UnknownValidator) }
+		for (hash, signers) in iterable {
+			if !signers.iter().all(|s| self.signers.contains(s)) { return Err(UnknownValidator) }
```
rphmeier: using `any` and putting the negation inside the closure will short-circuit as soon as the first unknown is detected

ethcore/src/engines/mod.rs
```diff
@@ -409,27 +409,31 @@ pub mod common {
...
+				.map_err(::error::Error::from)
+				.and_then(|_| fields.state.commit());
+
+			fields.traces.as_mut().map(move |traces| {
+  				let mut tracer = ExecutiveTracer::default();
+  				tracer.trace_reward(*receiver, reward, RewardType::Block);
+  				traces.push(tracer.drain())
+			});
+
+			// Commit state so that we can actually figure out the state root.
```
rphmeier: this comment seems outdated. the state commit is done after the call to `engine.on_close_block` now

ethcore/src/engines/mod.rs
```diff
@@ -409,27 +409,31 @@ pub mod common {
...
-  			let mut tracer = ExecutiveTracer::default();
-  			tracer.trace_reward(block_author, reward, RewardType::Block);
-  			traces.push(tracer.drain())
-		});
-
-		// Commit state so that we can actually figure out the state root.
-		if let Err(ref e) = res {
-			warn!("Encountered error on bestowing reward: {}", e);
+
+		for receiver in receivers {
```
rphmeier: couldn't we just call the `bestow_block_reward` function in a loop from `authority_round`?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -91,6 +98,8 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 			block_reward: p.block_reward.map_or_else(Default::default, Into::into),
 			maximum_uncle_count_transition: p.maximum_uncle_count_transition.map_or(0, Into::into),
 			maximum_uncle_count: p.maximum_uncle_count.map_or(0, Into::into),
+			empty_steps_transition: p.empty_steps_transition.map_or(u64::max_value(), |n| ::std::cmp::max(n.into(), 1)),
```
rphmeier: why is it lower bounded at 1?

rphmeier:
A couple grumbles and questions, in general looks fine. The authority_round module is getting fairly cluttered. It would be nice to refactor into two separate modules, one for pre-transition logic and one for post, with the `AuthorityRound` struct simply delegating into either one as necessary.

ethcore/src/engines/authority_round/mod.rs
andresilva: The genesis block seal doesn't include the empty steps and this way we don't have to deal with it.

andresilva:
Fixed grumbles. I agree that we should factor out the distinct behavior, created #7875.

PR #7610, opened 2018-01-18
Problem: AttachedProtocols don't get registered

### LINKED ISSUE #7566: AttachedProtocol registration is a no-op
> I'm running:
>
> - **Which Parity version?**: master
> - **Which network are you connected to?**: other

I was investigating issues I am having with Whisper support. I've enabled Whisper on a custom test network and inserted traces into Whisper handler implementation (`Network<T>` and `NetworkProtocolHandler for Network<T>`) and I noticed that the handler was never invoked. 

After further research on this matter, I found out that `AttachedProtocol`'s `register` function does nothing: https://github.com/paritytech/parity/blob/master/sync/src/api.rs#L172 but there was an implementation originally: https://github.com/paritytech/parity/commit/99075ad22a8c629d33520aae3ce90c91d9418a1b#diff-5212acb6bcea60e9804ba7b50f6fe6ec and it did the actual expected logic of registering the protocol in the `NetworkService`. 

However, as of https://github.com/paritytech/parity/commit/16d84f8ecfe223918bd9a9ed9d0ddc4ee63317ab#diff-5212acb6bcea60e9804ba7b50f6fe6ec ("finished removing ipc") this implementation is gone and only the no-op function is left.

Which leads me to a conclusion that in fact Whisper's handler never gets registered in the service and therefore two nodes won't communicate using it.

Is there something I am missing and this is how it is supposed to be or was that implementation of `AttachedProtocol.register` removed in error? Is there something else I should look at to have a complete picture?

yrashk:
I was investigating issues I am having with Whisper support. I've
enabled Whisper on a custom test network and inserted traces into
Whisper handler implementation (Network<T> and NetworkProtocolHandler
for Network<T>) and I noticed that the handler was never invoked.

After further research on this matter, I found out that
AttachedProtocol's register function does nothing:
https://github.com/paritytech/parity/blob/master/sync/src/api.rs#L172
but there was an implementation originally:
99075ad#diff-5212acb6bcea60e9804ba7b50f6fe6ec and it did the actual
expected logic of registering the protocol in the NetworkService.

However, as of 16d84f8#diff-5212acb6bcea60e9804ba7b50f6fe6ec ("finished
removing ipc") this implementation is gone and only the no-op function
is left.

Which leads me to a conclusion that in fact Whisper's handler never gets
registered in the service and therefore two nodes won't communicate
using it.

Solution: Resurrect original non-empty `AttachedProtocols.register`
implementation

Resolves #7566

### DIFF
--- a/sync/src/api.rs
+++ b/sync/src/api.rs
@@ -169,7 +169,18 @@ pub struct AttachedProtocol {
 }
 
 impl AttachedProtocol {
-	fn register(&self, _network: &NetworkService) {}
+	fn register(&self, network: &NetworkService) {
+		let res = network.register_protocol(
+			self.handler.clone(),
+			self.protocol_id,
+			self.packet_count,
+			self.versions
+		);
+
+		if let Err(e) = res {
+			warn!(target: "sync", "Error attaching protocol {:?}: {:?}", self.protocol_id, e);
+		}
+	}
 }
 
 /// EthSync initialization parameters.


debris:
thanks! nice catch!

PR #7613, opened 2018-01-18
Fix Temporarily Invalid blocks handling

### LINKED ISSUE #7527: AuRa: TemporarilyInvalid - block too early
> I'm running:
>
> - **Which Parity version?**: 1.8.6
> - **Are you fully synchronized?**: -
> - **Which network are you connected to?**: local

I have an AuRa network consisting of a single validator and a few ordinary nodes, they were all running on 1.8.4. Today I upgraded some of the ordinary nodes to 1.8.6 (replaced `parity` binary with the new version) and they stop syncing with this errors like this:
```
2018-01-10 21:16:09 UTC Verifier #0 TRACE engine  verify_block_external: block too early
2018-01-10 21:16:09 UTC Verifier #0 WARN client  Stage 4 block verification failed for #226896 (cf36…c81c)
Error: Block(TemporarilyInvalid(OutOfBounds { min: None, max: Some(1515618965), found: 1515618970 }))
```
If I downgrade them to 1.8.5 or 1.8.4 it doesn't reproduce. Time is in sync on all nodes (`chrony` is used). Nodes that were not upgraded are running ok.

tomusdrw:
In AuRa

Closes #7527


andresilva:
Fixes #7527.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -653,26 +645,38 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
 				OutOfBounds { min: None, max: Some(U256::from(U128::max_value())), found: *header.difficulty() }
-			)))
-		} else {
-			Ok(())
+			)));
+		}
+
+		// TODO [ToDr] Should this go from epoch manager?
+		// If yes then probably benign reporting needs to be moved further in the verification.
+		let set_number = header.number();
```
andresilva: We should still get a comment on whether this is correct or if we need to use the epoch transition number.

PR #7614, opened 2018-01-18
removed old, unused build.rs

debris:
`const MIN_RUSTC_VERSION: &'static str = "1.15.1"` was not updated in a long time, which indicates that this value is simply unused


tomusdrw:
IMHO we could just move it to a separate crate (with only this build script) and perhaps even parse `README.md` for latest version or sth.

I think it's worth to keep an maintain that (and maybe actually compiling on 1.15 still works:))

debris:
The main problem with this script (even if we remember to update it) is that there is no guarantee regarding the order of compilation. So even though we have the script, the compilation of other crate / library can fail.

There is a number of rfcs to rust/cargo to add minimal version of rustc required. I think, for now we could just remove our code and add the minimal version when the rfc is standarized.

PR #7562, opened 2018-01-15
SecretStore: threshold decrease PoC

svyatonik:
Test to prove that "Unresolved: 2.2" from #7271 description is possible

---
This change is [Reviewable](https://reviewable.io/reviews/paritytech/parity/7562)

PR #7615, opened 2018-01-18
SecretStore: threshold ECDSA PoC

svyatonik:
It is a test of crypto-math behind possible implementation of ECDSA-compatible SS session.

The only known limitation is that it requires 2 * t <= N. I.e. it isn't possible to make ECDSA signature when key was generated, for example 4-of-6 session.
Workaround is to temporarily (for the duration of this session) decrease key threshold (see #7562 for PoC). But we need to agree on this (it is a part of #7271 discussion, actually).

So if this PR will be merged before #7271 is reviewed, I'll start adding ECDSA session without automatic threshold decreasing && wait until this solution will be reviewed.


ethkey/src/secret.rs
```diff
@@ -48,6 +48,11 @@ impl Secret {
 		Secret { inner: h }
 	}
 
+	/// Creates zero key, which is invalid for crypto operations, but valid for math operation.
+	pub fn zero() -> Self {
+		Secret { inner: Default::default() }
```
svyatonik: Didn't wanted to implement `Default` for `Secret`, since it actually produces invalid `Secret`.

ethkey/src/secret.rs
```diff
@@ -61,51 +66,91 @@ impl Secret {
 
 	/// Inplace add one secret key to another (scalar + scalar)
 	pub fn add(&mut self, other: &Secret) -> Result<(), Error> {
-		let mut key_secret = self.to_secp256k1_secret()?;
-		let other_secret = other.to_secp256k1_secret()?;
-		key_secret.add_assign(&SECP256K1, &other_secret)?;
+		match (self.is_zero(), other.is_zero()) {
```
svyatonik: The reason behind these changes is that starting with this PR, I need to handle 0 EC scalars.
0 is the invalid key in secp256k1 && library will return an error when trying to work with it.
Long story short: it wasn't a best idea to use `Secret` for EC arithmetic (the main assumption was that it will minimize `U256` <-> `Secret` convertations).
And the general solution is either to find&use/create separate EC-arithmetic library.
If you're strong against these changes - please comment && I'll close this PR && will switch to separate type for EC calculations first.

secret_store/src/key_server_cluster/math.rs
```diff
@@ -528,6 +563,65 @@ pub mod tests {
...
+			mul_share
+		}).collect();
+
+		mul_shares
+	}
+
+	fn run_reciprocal_protocol(t: usize, artifacts: &KeyGenerationArtifacts) -> Vec<Secret> {
+		// === Given a secret x mod r which is shared among n players, it is
+		// === required to generate shares of 1/x mod r with out revealing
+		// === any information about x or 1/x.
```
rphmeier: nit: can we frame that as a multiplicative inverse instead of `1/x`?

secret_store/src/key_server_cluster/math.rs
```diff
@@ -434,7 +434,27 @@ pub mod tests {
...
+		secret_shares: Vec<Secret>,
+	}
+
+	fn prepare_polynoms1(t: usize, n: usize, secret_required: Option<Secret>) -> Vec<Vec<Secret>> {
+		let mut polynoms1: Vec<_> = (0..n).map(|_| generate_random_polynom(t).unwrap()).collect();
+		// if we need specific secret to be shared, update polynoms so that sum of their free terms = required secret
+		if let Some(mut secret_required) = secret_required {
+			for polynom1 in polynoms1.iter_mut().take(n - 1) {
+				let secret_coeff1 = generate_random_scalar().unwrap();
+				secret_required.sub(&secret_coeff1).unwrap();
```
rphmeier: is that tested to wrap around?

svyatonik: Thanks for reviewing! Yes - it will wrap around in case of underflow. Please see `run_zero_key_generation` - 0 is passed here as `secret_required` => any non-zero value (which is generated by `generate_random_scalar`) will cause underflow => yes it is tested.

secret_store/src/key_server_cluster/math.rs
svyatonik: Sure :) It is just easier to write `1/x` instead of `multiplicative inverse of x`. Did `inv(x)` sounds good, or it is better to write its 'full name'?

rphmeier: `inv(x)` is fine for me

rphmeier:
Did an initial review of the code and it looks fine. I'll dive into the math tomorrow.
Just wondering if you've seen this paper: https://eprint.iacr.org/2016/013.pdf as it's more recent than the one linked in the PR and seems to get rid of the `n >= 2t+1` requirement.

svyatonik:
@rphmeier Thanks for the link!!! I'll read this paper on the weekend, but iirc last time I was searching for `threshold-optimal ECDSA` (this was ~2 months ago), every paper had its own 'issue' - like one was strictly about 2-of-2 scheme, another required 2 * t nodes to recover the key, another was about using different key generation scheme (secret = multiplication of shares, instead of addition [that's what we currently use]), another was about 'combinatorics threshold' (i.e. generate t-of-n shares for each possible combination of nodes). I'll mark PR as `gotissues` until I'm done with reading.

svyatonik:
So the key generation scheme in this paper differs from what we currently use (at first glance). Since the most of SS functionality is based on current KG scheme (like Enc/Dec + migration), I'll opt for using proposed (in this PR) signature generation prototype now (there are) + detailed review (and possibly including it to the SS [TODO for myself: research if shares migration is possible in this scheme]) of your paper later.

svyatonik:
Found some issue with `serialize_ecdsa_signature`

svyatonik:
Turned out to be issue with `nonce_public` calculation - fixed it. I've also:
- extracted some parts of tests into separate functions, because I'll need these for ECDSA session implementation;
- added tests for different key sharing schemes;
- added random message hash generation to the test (previously it always was less than EC order).

debris:
@rphmeier could you review this pr again? I don't feel qualified to review it without diving into math

ethkey/src/secret.rs
rphmeier: not strongly against these changes, but down the road it could be nice to have a separate type for secp256k1 scalars which has `std::ops` traits implemented for it.

secret_store/src/key_server_cluster/math.rs
```diff
@@ -28,12 +29,27 @@ pub struct EncryptedSecret {
...
+pub fn zero_scalar() -> Secret {
+	Secret::zero()
+}
+
+/// Convert hash to EC scalar (modulo curve order).
+pub fn to_scalar(hash: H256) -> Result<Secret, Error> {
+	let scalar: U256 = hash.into();
+	let scalar: H256 = (scalar % math::curve_order()).into();
+	let scalar = Secret::from_slice(&*scalar);
+	scalar.check_validity()?;
```
rphmeier: any secret in  `0..curve_order` is valid though, right?

secret_store/src/key_server_cluster/math.rs
```diff
@@ -418,10 +429,108 @@ pub fn verify_signature(public: &Public, signature: &(Secret, Secret), message_h
 	Ok(combined_hash == signature.0)
 }
 
+
+/// Compute R part of ECDSA signature.
+#[cfg(test)]
+pub fn compute_ecdsa_r(nonce_public: &Public) -> Result<Secret, Error> {
```
rphmeier: I think this is better explained as `x_coordinate(point)` than as "r", which requires diving a bit more into implementations of ecdsa

secret_store/src/key_server_cluster/math.rs
```diff
@@ -418,10 +429,108 @@ pub fn verify_signature(public: &Public, signature: &(Secret, Secret), message_h
...
+}
+
+/// Serialize ECDSA signature to [r][s]v form.
+#[cfg(test)]
+pub fn serialize_ecdsa_signature(nonce_public: &Public, signature_r: Secret, mut signature_s: Secret) -> Signature {
+	// compute recvery param
+	let mut signature_v = {
+		let nonce_public_x = &nonce_public[0..32];
+		let nonce_public_x: U256 = nonce_public_x.into();
+		let nonce_public_x: H256 = nonce_public_x.into();
```
rphmeier: `let nonce_public_x = H256::from_slice(&nonce_public[0..32])`

secret_store/src/key_server_cluster/math.rs
```diff
@@ -418,10 +429,108 @@ pub fn verify_signature(public: &Public, signature: &(Secret, Secret), message_h
...
+}
+
+/// Compute S part of ECDSA signature from shares.
+#[cfg(test)]
+pub fn compute_ecdsa_s(t: usize, signature_s_shares: &[Secret], id_numbers: &[Secret]) -> Result<Secret, Error> {
+	let double_t = t * 2;
+	debug_assert!(id_numbers.len() >= double_t + 1);
+	debug_assert_eq!(signature_s_shares.len(), id_numbers.len());
+
+	compute_joint_secret_from_shares(double_t,
```
rphmeier: This interpolation doesn't seem to be possible with only `t` participants (as it's polluted by the shares of the `z` blinding factor).

secret_store/src/key_server_cluster/math.rs
```diff
@@ -681,12 +875,63 @@ pub mod tests {
...
+			let signature_r = compute_ecdsa_r(&nonce_public).unwrap();
+
+			// compute shares of inv(nonce) so that both nonce && inv(nonce) are still unknown to all nodes
+			let nonce_inv_shares = run_reciprocal_protocol(t, &nonce_artifacts);
+
+			// compute multiplication of secret-shares * inv-nonce-shares
+			let mul_shares = run_multiplication_protocol(t, &artifacts.secret_shares, &nonce_inv_shares);
+
+			// compute shares for s portion of signature: nonce_inv * (message_hash + secret * signature_r)
+			// every node broadcasts this share
```
rphmeier: if each node's participation (or at least 2t + 1) is required to create a share here, then can it really be considered a t-of-n scheme?

secret_store/src/key_server_cluster/math.rs
svyatonik: yes. except for 0 itself

secret_store/src/key_server_cluster/math.rs
svyatonik: Replaced with `2*t+1` (it is actually `compute_ecdsa_s` who take care of this, but it might be confusing to see n in this test)

secret_store/src/key_server_cluster/math.rs
svyatonik: I've added `public_x` and `public_y` (and used `public_x` in this function), but left `compute_ecdsa_r` - it is easier to think of `R` and `S` parts in ECDSA session (next PR). I.e. leave the fact that R is actually a `point.x` behind 'math-wall'.

secret_store/src/key_server_cluster/math.rs
rphmeier: yes, but if running `compute_ecdsa_s` is part of a t-of-n scheme then I don't understand how it could ever complete (relevant to my other comment)

rphmeier: or have I got it backwards somehow, and it's actually a 2t-of-n scheme because we have to interpolate such polynomials?

svyatonik: `signature_s_shares` are the shares of S in `(2*t+1)-of-N` scheme. So what `compute_ecdsa_s` does is exactly interpolating 2*t-degree polynomial.

Issue #7609, opened 2018-01-18
--ui-no-validation, --no-hardware-wallets, other flags are ignored, token creation dapp inaccessible


trtg:
I'm running:
Which Parity version?**: master
> - **Which operating system?**:  Linux
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: no (running private chain)
> - **Which network are you connected to?**: private test
> - **Did you try to restart the node?**: yes, but not relevant

Locally trying to use the token creation dapp, the UI will not load if accessed via 127.0.0.1:8180 with this error:
`
Refused to display 'http://127.0.0.1:8545/ui/tokendeploy.html' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 0.0.0.0:8180 *.web3.site:8180 home.web3.site *:8180".
`
this occurs even if --ws-origins, --ws-hosts, --jsonrpc-hosts, are set to all and --ui-no-validation are set. In fact, even when combined --with --unsafe-expose it's impossible to access the token creation dapp by clicking through the UI, while using 127.0.0.1:8180

Trying to access via web3.site also fails but for a different reason- it brings up this dialog:
unable to make a connection
following the instructions provided in that dialog does nothing and the page remains inaccessible.

For context, this is an issue when using the docker-compose setup for the 3 node PoA setup. When using parity directly the token creation dapp page is at least accessible. Also the token registry dapp does not appear to have the origin problem seen by the token creation dapp.

Also the --no-hardware-wallets flag appears to be ignored as well- in the token creation dapp, I can see a slew of trezor related errors due to being inaccessible even when --no-hardware-wallets was specified.

tomusdrw:
Seems like an issue with Chrome handling of `*:8180` CSP, works fine for me on Firefox.
Preparing a PR with a workaround for Chrome.

tomusdrw:
`--no-hardware-wallets` is handled on the node side, UI is probably sending some requests but they all fail, which is correct. Unless something is not working I think it's a non-issue (perhaps a dev-annoyance seeing failures in the console).

PR #7621, opened 2018-01-18
CSP hardening

tomusdrw:
+ a fix for #7609 (chrome not intepreting `*:8180` correctly)
+ removed `web3` instance from `inject.js` (since it uses `unsafe-eval`)


5chdn:
Please make sure all "fixes" receive a patch label.

jacogr:
LGTM, also tested it locally earlier today.

PR #7624, opened 2018-01-18
Replace RandomTempDir with tempdir::Tempdir

debris:


ethcore/src/snapshot/tests/proof_of_authority.rs
```diff
@@ -236,14 +236,14 @@ fn fixed_to_contract_only() {
 	// 6, 7, 8 prove finality for transition at 6.
 	// 3 beyond gets us to 11.
 	assert_eq!(client.chain_info().best_block_number, 11);
-	let reader = snapshot_helpers::snap(&*client);
+	let (reader, _tempdir) = snapshot_helpers::snap(&*client);
```
kirushik: Wouldn't this trick cause problems when Non-lexical lifetimes land in the compiler? `_tempdir` would be dropped immediately then.
Maybe it's better to transfer ownership of the folder into the `SnapshotReader`? It would feel natural.

debris: > Wouldn't this trick cause problems when Non-Lexical Lifetimes land in the compiler? _tempdir would be dropped immediately then.

I don't think it will ever be a problem. It would be a huuuuge issue for multiple projects as all guards would stop working as intended

And even if someone would be crazy enough to introduce this change, we'd just modify the code, so the tuple is not destructured with `let (..,..)` statement.

> Maybe it's better to transfer ownership of the folder into the `SnapshotReader`? It would feel natural.

It would change nothing in this case, cause temporary directory is needed only for test. In production code `SnapshotReader` would never remove a directory in a `drop` function.

kirushik: OK, sounds reasonable.

PR #7669, opened 2018-01-23
fix snapcraft build for stable

General-Beck:


5chdn:
This is still failing: https://gitlab.parity.io/parity/parity/-/jobs/77503

PR #7685, opened 2018-01-24
Move js & js-old code to github.com/parity-js

jacogr:
So long and thanks for all the fish.

- Move js code to [github.com/parity-js](https://github.com/Parity-JS) into separate repos
- Updated shell & wallet references with `cargo update -p parity-ui`
- Tested operation of UI with `cargo build --release --no-default-features --features ui`
- Update CI scripts to remove `js-build` & `js-release`

cc @amaurymartiny @ngotchac @tomusdrw


Cargo.lock
```diff
@@ -2214,13 +2214,15 @@ dependencies = [
 [[package]]
 name = "parity-ui-dev"
 version = "1.9.0"
+source = "git+https://github.com/parity-js/shell.git?rev=c4ad493b86fd8dda0364881d9cff39c95e2ee73b#c4ad493b86fd8dda0364881d9cff39c95e2ee73b"
```
jacogr: This auto-gen line _looks_ funny.

tomusdrw: It's because you have set explicit `rev` in `Cargo.toml` file. Instead you could just set up github repo and let `Cargo.lock` maintain the revision.
But it's fine, might be a bit more pain in the ass to update though:
```
1. edit Cargo.toml
2. Run $ cargo update -p parity-ui-dev
```
vs
```
$ cargo update -p parity-ui-dev
```

jacogr:
Updating will have to be done manually, i.e. like we do for other packages anyway.

We can still push from the Travis build for the shell & wallet to update the refs, but since we want to not have Parity (1.10) build the UI in, I would rather just update manually when we need to since it is not long-term. (But if need be, can just push from CI again)

Cargo.lock
jacogr: 100%, bit more of a pain with the revs - this does feel safer since the shell will morph as we get closer to having it dapp-able.

debris:
I have no idea why 'js' report is `requried`. @5chdn is it because of our github repo settings?

jacogr:
@debris it was changed on the repo settings by @5chdn - should be fine now.

PR #7688, opened 2018-01-24
Distinguish debian and centos builds from default linux.

5chdn:
Problem: Debian and Ubuntu .deb have same file name.This caused issues with the One-Line-Installer.

I remove the $ARC and set it to debian to distinguish the difference. We only build debian for x86_64 anyways :)

Same for centos for convenience.


General-Beck:
RPM filename
_<name>-<version>-<release>.<architecture>.rpm_ where architecture **x86_64**
DEB filename 
_<name>-<version>-<release>.<architecture>.deb_ where architecture **amd64**
it is the standard

Issue #7697, opened 2018-01-25
Parity connecting to very few peers


ddorgan:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Parity version?**: 1.8.x - happens on version from 1.8.2 to 1.8.6 
> - **Which operating system?**: Linux
> - **How installed?**: deb package
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum 
> - **Did you try to restart the node?**: yes

Sometimes parity is only connected to a small number of peers, maybe 1 or 2. Sometimes it has no connections at all. I tried upgrading parity to the latest (1.8.7) but the issue continues. 

Once I remove the file chains/ethereum/network/nodes.json then the issue seems to go away. The nodes.json also has quite a lot of entries (+175k). I can provide an example if needed.

Once the file is removed the node will sync as expected, reaching 25 peers.

Expected behaviour is that a normal sync can happen without having to remove this file.

roninkaizen:
would plan a job,
stopping parity every 6 hours, 
wait until done,
deleting that the nodes.json file,
restarting parity
end

nueverest:
I had the same issue on upgrading to 1.9.

Deleting nodes.json fixed it for me.

5chdn:
#7716 should have addressed this (hopefully).

PR #7716, opened 2018-01-29
Filter-out nodes.json

tomusdrw:
If we reach more than 1024 nodes in the table, filter out the ones that we never attempted to connect to.

(untested yet, need to get better internet connection to test it properly)

### DIFF
--- a/util/network/src/lib.rs
+++ b/util/network/src/lib.rs
@@ -80,17 +80,20 @@ extern crate path;
 extern crate ethcore_logger;
 extern crate ipnetwork;
 extern crate keccak_hash as hash;
+extern crate serde;
 extern crate serde_json;
 extern crate snappy;
 
 #[macro_use]
 extern crate error_chain;
-
 #[macro_use]
 extern crate log;
+#[macro_use]
+extern crate serde_derive;
 
 #[cfg(test)]
 extern crate tempdir;
+
 
 mod host;
 mod connection;
--- a/util/network/src/lib.rs
+++ b/util/network/src/lib.rs
@@ -94,7 +94,6 @@ extern crate serde_derive;
 #[cfg(test)]
 extern crate tempdir;
 
-
 mod host;
 mod connection;
 mod handshake;
@@ -210,4 +209,3 @@ pub enum AllowIP {
     /// Block all addresses
     None,
 }
-
--- a/util/network/src/node_table.rs
+++ b/util/network/src/node_table.rs
@@ -250,9 +250,15 @@ impl NodeTable {
 			.filter(|n| n.endpoint.is_allowed(&filter))
 			.collect();
 		refs.sort_by(|a, b| {
-			let ord = a.failure_percentage().cmp(&b.failure_percentage());
-			if ord == Ordering::Equal { a.failures.cmp(&b.failures) }
-			else { ord }
+			let mut ord = a.failure_percentage().cmp(&b.failure_percentage());
+			if ord == Ordering::Equal {
+				ord = a.failures.cmp(&b.failures);
+				if ord == Ordering::Equal {
+					// we use reverse ordering for number of attempts
+					ord = b.attempts.cmp(&a.attempts);
+				}
+			}
+			ord
 		});
 		refs.into_iter().map(|n| n.id).collect()
 	}
--- a/util/network/src/node_table.rs
+++ b/util/network/src/node_table.rs
@@ -448,26 +448,42 @@ mod tests {
 	}
 
 	#[test]
-	fn table_failure_order() {
+	fn table_failure_ratio_order() {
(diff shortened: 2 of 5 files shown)


util/network/src/node_table.rs
```diff
@@ -282,77 +286,63 @@ impl NodeTable {
...
+			warn!("Error creating node table directory: {:?}", e);
+			return;
+		}
+		path.push(NODES_FILE);
+		let node_ids = self.nodes(IpFilter::default());
+		let len = node_ids.len();
+		let nodes = node_ids.into_iter()
+			.map(|id| self.nodes.get(&id).expect("self.nodes() only returns node IDs from self.nodes"))
+			.map(|node| node.clone())
+			.filter(|node| if len > MAX_NODES { node.last_attempted.is_some() } else { true })
```
andresilva: Maybe we should change the nodes sorting so that nodes that we never connected to have a lower priority and then we `.take(MAX_NODES)` on the sorted list. This way we always get a node table with a predictable size.

tomusdrw: Sounds like a good idea, we might need to tune the comparator a bit though. Might be difficult to decide what nodes are more valuable: never connected or connected once, but with a single failure.
In general I need to double check if we reset `failures` counter in case we successfuly connect to the node, otherwise valuable peers might be dropped to often (with time they would accumulate a lot of failures, so the preference would be to completely new nodes).

andresilva:
Here's what my `nodes.json` looks like (https://gist.github.com/andresilva/78e8caeb021378206cd43863840af519). After I restart the node it seems to pick up peers a bit faster in the beginning but it may just be my impression. 🙃

```
❯ ./target/release/parity --min-peers 100
2018-01-30 01:22:51 UTC Starting Parity/v1.10.0-unstable-6dd69196d-20180129/x86_64-macos/rustc1.24.0-beta.8
...
2018-01-30 01:23:04 UTC Imported #4997033 1888…31df (179 txs, 7.49 Mgas, 636.76 ms, 27.11 KiB) + another 1 block(s) containing 76 tx(s)
2018-01-30 01:23:20 UTC Imported #4997033 d609…65b8 (166 txs, 6.66 Mgas, 363.01 ms, 25.29 KiB)
2018-01-30 01:23:28 UTC   36/100 peers   219 KiB chain 100 MiB db 0 bytes queue 31 KiB sync  RPC:  0 conn,  0 req/s,   0 µs
```

util/network/src/lib.rs
```diff
@@ -80,18 +80,21 @@ extern crate path;
...
-
 #[macro_use]
 extern crate log;
+#[macro_use]
+extern crate serde_derive;
 
 #[cfg(test)]
 extern crate tempdir;
 
+
```
tomusdrw: Superflous.

util/network/src/node_table.rs
```diff
@@ -143,18 +139,30 @@ pub struct Node {
...
+		}
+	}
+
+	/// Returns the node's failure percentage (0..100) in buckets of 5%. If there are 0 connection attempts for this
+	/// node the default failure percentage is returned (50%).
+	pub fn failure_percentage(&self) -> usize {
+		if self.attempts == 0 {
+			DEFAULT_FAILURE_PERCENTAGE
+		} else {
+			((self.failures as f64 / self.attempts as f64 * 100.0 / 5.0).round() * 5.0) as usize
```
tomusdrw: Do we really need to fallback to floats? What about:
```
self.failures * 10_000 / self.attempts
```
(then we get a number within `0 .. 10_000` representing percents with two decimal places)

util/network/src/node_table.rs
```diff
@@ -221,23 +232,43 @@ impl NodeTable {
...
-		let mut refs: Vec<&Node> = self.nodes.values().filter(|n| !self.useless_nodes.contains(&n.id) && n.endpoint.is_allowed(&filter)).collect();
-		refs.sort_by(|a, b| a.failures.cmp(&b.failures));
-		refs.iter().map(|n| n.id.clone()).collect()
+		let mut refs: Vec<&Node> = self.nodes.values()
+			.filter(|n| !self.useless_nodes.contains(&n.id))
+			.filter(|n| n.endpoint.is_allowed(&filter))
+			.collect();
+		refs.sort_by(|a, b| {
+			let mut ord = a.failure_percentage().cmp(&b.failure_percentage());
+			if ord == Ordering::Equal {
```
tomusdrw:
```rust
a.failure_percentage().cmp(&b.failure_percentage())
  .then_with(|| a.failures.cmp(&b.failures))
  .then_with(|| b.attempts.cmp(&a.attempts)
```

util/network/src/node_table.rs
```diff
@@ -143,18 +139,30 @@ pub struct Node {
 	pub id: NodeId,
 	pub endpoint: NodeEndpoint,
 	pub peer_type: PeerType,
+	pub attempts: u32,
 	pub failures: u32,
-	pub last_attempted: Option<Tm>,
```
tomusdrw: Interesting, I didn't notice it's not used at all.

andresilva:
Does this completely fix #7697 (and other duplicate issues)?

arkpar:
LGTM

nueverest:
Deleting nodes.json fixed the issue for me.

PR #7723, opened 2018-01-29
ethabi version 5

### LINKED ISSUE #6360: Migrate to ethabi 4.0.0

### LINKED ISSUE #7720: Retain cycle in updater
https://github.com/paritytech/parity/blob/bc8216a82c7a5db80710bb6863d6be7af7e26151/updater/src/updater.rs#L134

`r` is `self`. it owns `fetcher` which owns strong reference to `self`

debris:
replaces #6896 

the real number of lines changed is: **+1,053 -2,097**. *560* added lines belong to a single json file which was inlined before

changes:
- bumped ethabi to version `5.x`
- fixed #7720 - retain cycle in updater
- closes #6360 
- removed `native_contracts` crate. we use `ethabi_contract` and `ethabi_derive` instead. Now creating contract interface has no runtime cost
- fixes updates when updater is configured to track minor version
- bumped `ethereum-types` to version `0.2`. fixed cases when `x.hex()` was used instead of `format!("{:x}", x)`
- updated several libraries to their latest version to resolve dependencies conflicts


updater/src/updater.rs
```diff
@@ -161,21 +176,34 @@ impl Updater {
 	}
 
 	fn collect_latest(&self) -> Result<OperationsInfo, String> {
-		if let Some(ref operations) = *self.operations.lock() {
+		if let &Some(ref do_call) = &*self.do_call.lock() {
```
tomusdrw: Why is `&` needed?

updater/src/updater.rs
```diff
@@ -118,20 +120,23 @@ fn platform() -> String {
...
 			sync: sync.clone(),
-			fetcher: Mutex::new(None),
-			operations: Mutex::new(None),
+			fetcher,
+			operations_contract: operations_contract::Operations::default(),
 			exit_handler: Mutex::new(None),
 			this: VersionInfo::this(),
 			state: Mutex::new(Default::default()),
 		});
-		*r.fetcher.lock() = Some(fetch::Client::with_fetch(r.clone(), fetch, remote));
```
debris: removed retain cycle

updater/src/updater.rs
```diff
@@ -161,41 +184,68 @@ impl Updater {
...
+			.map_err(|e| format!("{:?}", e))?;
+
+
+		// get the release info for the latest version in track
+		let in_track = self.collect_release_info(latest_in_track)?;
+		let mut in_minor = Some(in_track.clone());
+		const PROOF: &'static str = "in_minor initialised and assigned with Some; loop breaks if None assigned; qed";
+
+		// if the minor version has changed, let's check the minor version on a different track
+		while in_minor.as_ref().expect(PROOF).version.version.minor != self.this.version.minor {
```
debris: fixed tracking minor version by updater

secret_store/src/key_server_set.rs
```diff
@@ -168,6 +168,77 @@ impl ChainNotify for OnChainKeyServerSet {
 	}
 }
 
+trait KeyServerSubset<F: Fn(Vec<u8>) -> Result<Vec<u8>, String>> {
```
debris: @svyatonik please review changes in secret store. I made them, cause it was the easiest way to replace old contracts interface with the new one

updater/src/updater.rs
```diff
@@ -246,17 +297,6 @@ impl Updater {
 			return;
 		}
 
-		if self.operations.lock().is_none() {
-			if let Some(ops_addr) = self.client.upgrade().and_then(|c| c.registry_address("operations".into())) {
-				trace!(target: "updater", "Found operations at {}", ops_addr);
-				let client = self.client.clone();
-				*self.operations.lock() = Some(Operations::new(ops_addr, move |a, d| client.upgrade().ok_or("No client!".into()).and_then(|c| c.call_contract(BlockId::Latest, a, d))));
```
debris: I haven't investigated that, but this lazy evaluation could lead to another retain cycle.

`updater` owns `operations`. `operations` owns `client`

but it was not clearly visible, since `updater` owns a weak reference to a `client`

debris:
updater needs much more work, but I for now, I've decided to not modify it's core logic in this pr.

I'll create a separate pr with tests and later I'll refactor it

secret_store/src/key_server_set.rs
svyatonik: Thanks! Code looks perfect! Right now I'm updating SS service contract && will do real-world testing of these changes (all three contracts) before making a PR.

updater/src/types/release_track.rs
```diff
@@ -69,9 +70,10 @@ impl From<u8> for ReleaseTrack {
 	}
 }
 
-impl Into<u8> for ReleaseTrack {
-	fn into(self) -> u8 {
-		match self {
+
+impl From<ReleaseTrack> for u8 {
+	fn from(rt: ReleaseTrack) -> Self {
+		match rt {
```
svyatonik: Maybe just `rt as u8` instead of that `match`?

ethcore/src/snapshot/tests/proof_of_authority.rs
```diff
@@ -136,8 +136,8 @@ fn make_chain(accounts: Arc<AccountProvider>, blocks_beyond: usize, transitions:
 			vec![transaction]
 		};
 
-		let contract_1 = ValidatorSet::new(*CONTRACT_ADDR_1);
-		let contract_2 = ValidatorSet::new(*CONTRACT_ADDR_2);
+		let contract_1 = test_validator_set::ValidatorSet::default();
```
svyatonik: I can be wrong, but looks like both `contract_1` and `contract_2` can now be replaced with single `contract` variable.

debris: you are absolutely right!

ethcore/light/src/client/header_chain.rs
```diff
@@ -788,7 +789,7 @@ mod tests {
 			header.set_parent_hash(parent_hash);
 			header.set_number(i);
 			header.set_timestamp(rolling_timestamp);
-			header.set_difficulty(*genesis_header.difficulty() * i.into());
+			header.set_difficulty(*genesis_header.difficulty() * i as u32);
```
tomusdrw: perhaps change the loop range to be `u32` instead?

ethcore/node_filter/src/lib.rs
```diff
@@ -20,9 +20,14 @@ extern crate ethcore;
 extern crate ethcore_bytes as bytes;
 extern crate ethcore_network as network;
 extern crate ethereum_types;
-extern crate native_contracts;
-extern crate futures;
 extern crate parking_lot;
+extern crate lru_cache;
+
+extern crate ethabi;
```
tomusdrw: This should be in previous section.

ethcore/node_filter/src/lib.rs
```diff
@@ -20,9 +20,14 @@ extern crate ethcore;
...
+#[macro_use]
+extern crate ethabi_derive;
+#[macro_use]
+extern crate ethabi_contract;
```
tomusdrw: Redundant.

ethcore/node_filter/src/lib.rs
```diff
@@ -33,33 +38,34 @@ extern crate kvdb_memorydb;
 extern crate ethcore_io as io;
 
 use std::sync::Weak;
-use std::collections::HashMap;
-use native_contracts::PeerSet as Contract;
-use network::{NodeId, ConnectionFilter, ConnectionDirection};
+use lru_cache::LruCache;
+use parking_lot::Mutex;
```
tomusdrw: Why these two are in the same section as `std`? IMHO it should be 
```
std-imports

external-imports

internal-imports
```
all sorted.

ethcore/src/engines/validator_set/contract.rs
```diff
@@ -22,50 +22,48 @@ use ethereum_types::{H256, Address};
...
 
 impl ValidatorContract {
-	// could be `impl Trait`.
-	// note: dispatches transactions to network as well as execute.
-	// TODO [keorn]: Make more general.
-	fn transact(&self) -> Box<Fn(Address, Bytes) -> Result<Bytes, String>> {
-		let client = self.client.read().clone();
-		Box::new(move |a, d| client.as_ref()
+	fn transact(&self, data: Bytes) -> Result<(), String> {
+		let client = self.client.read().clone().as_ref()
```
tomusdrw: You sure, you need both `clone()` and `.as_ref()`?

ethcore/src/engines/validator_set/contract.rs
```diff
@@ -22,50 +22,48 @@ use ethereum_types::{H256, Address};
...
 
+use_contract!(validator_report, "ValidatorReport", "res/contracts/validator_report.json");
+
 /// A validator contract with reporting.
 pub struct ValidatorContract {
+	contract_address: Address,
 	validators: ValidatorSafeContract,
-	provider: Provider,
+	provider: validator_report::ValidatorReport,
 	client: RwLock<Option<Weak<EngineClient>>>, // TODO [keorn]: remove
```
tomusdrw: Actually `Option<Weak<>>` doesn't make any sense, since you can always do `Weak::new()` (will just return `None`) when you try to upgrade it.

evmbin/src/display/mod.rs
```diff
@@ -39,6 +39,6 @@ pub fn u256_as_str(v: &U256) -> String {
 	if v.is_zero() {
 		"\"0x0\"".into()
 	} else {
-		format!("\"{:x}\"", v)
+		format!("\"0x{:x}\"", v)
```
tomusdrw: The whole method should not be needed any more see the TODO.

ethcore/src/tx_filter.rs
```diff
@@ -79,23 +79,15 @@ impl TransactionFilter {
...
-							tx_permissions::NONE
-						})
-					}
-					_ => tx_permissions::NONE,
-				};
+				let contract_address = self.contract_address;
+				let permissions = self.contract.functions()
+					.allowed_tx_types()
+					.call(sender, &|data| client.call_contract(BlockId::Hash(*parent_hash), contract_address, data))
+					.map(|p| p.low_u32())
```
tomusdrw: Where does this come from?

hash-fetch/src/lib.rs
```diff
@@ -27,17 +27,20 @@ extern crate futures;
...
 extern crate mime;
 extern crate mime_guess;
-extern crate native_contracts;
 extern crate parity_reactor;
 extern crate rand;
 extern crate rustc_hex;
 
 pub extern crate fetch;
 
+extern crate ethabi;
```
tomusdrw: move `ethabi` to other regular imports

hash-fetch/src/urlhint.rs
```diff
@@ -173,36 +171,30 @@ fn decode_urlhint_output(output: (String, H160, Address)) -> Option<URLHintResul
...
-						let do_call = move |_, data| client.call(address, data);
-						Either::B(urlhint.entries(do_call, H256(fixed_id)).map(decode_urlhint_output))
-					}
-				}
-			}))
+		let get_address = self.registrar.functions().get_address();
+		let entries = self.urlhint.functions().entries();
+		let data = get_address.input(keccak("githubhint"), "A");
+
+		let future = client.call(address, data)
```
tomusdrw: Why is it not using `get_address.call`?

miner/Cargo.toml
```diff
@@ -13,15 +13,18 @@ hyper = { git = "https://github.com/paritytech/hyper", default-features = false
...
 keccak-hash = { path = "../util/hash" }
 linked-hash-map = "0.5"
 log = "0.3"
-native-contracts = { path = "../ethcore/native_contracts" }
 parking_lot = "0.5"
 rustc-hex = "1.0"
 table = { path = "../util/table" }
 transient-hashmap = "0.4"
+
+ethabi = "5.1"
```
tomusdrw: Should be sorted together with other imports.

miner/src/lib.rs
```diff
@@ -39,6 +38,12 @@ extern crate rustc_hex;
 #[cfg(test)]
 extern crate ethkey;
 
+extern crate ethabi;
```
tomusdrw: Move `ethabi` to other imports and `ethabi_derive/ethabi_contract` to `macro-use` imports

miner/src/service_transaction_checker.rs
```diff
@@ -16,13 +16,13 @@
 
 //! A service transactions contract checker.
 
-use futures::{future, Future};
-use native_contracts::ServiceTransactionChecker as Contract;
-use ethereum_types::{U256, Address};
 use parking_lot::Mutex;
+use ethereum_types::Address;
```
tomusdrw: Sort

miner/src/service_transaction_checker.rs
```diff
@@ -37,34 +37,34 @@ pub trait ContractCaller {
...
 	pub fn update_from_chain_client(&self, client: &ContractCaller) {
-		let mut contract = self.contract.lock();
-		if contract.is_none() {
-			*contract = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME)
-				.and_then(|contract_addr| {
-					trace!(target: "txqueue", "Configuring for service transaction checker contract from {}", contract_addr);
-
-					Some(Contract::new(contract_addr))
-				})
+		let mut contract_address = self.contract_address.lock();
```
tomusdrw: The whole method is useless imho. just call registrar on `check` every time (since the address of the contract may change in the meantime) or call once at the start and memoize if we don't care that it might change.

With this approach we only support a case where the contract address is set to the registry, but we started earlier. IMHO it's a weird / too specific.

miner/src/service_transaction_checker.rs
```diff
@@ -37,34 +37,34 @@ pub trait ContractCaller {
...
+				trace!(target: "txqueue", "Configuring for service transaction checker contract from {}", address);
+				*contract_address = Some(address);
+			}
 		}
 	}
 
 	/// Checks if service transaction can be appended to the transaction queue.
 	pub fn check(&self, client: &ContractCaller, tx: &SignedTransaction) -> Result<bool, String> {
-		debug_assert_eq!(tx.gas_price, U256::zero());
+		debug_assert!(tx.gas_price.is_zero());
```
tomusdrw: `debug asserts` are useless, just use regular `assert` or remove

secret_store/src/key_server_set.rs
```diff
@@ -299,14 +357,13 @@ impl CachedContract {
 	}
 
 	fn read_from_registry(&mut self, client: &Client, new_contract_address: Option<Address>) {
-		self.contract = new_contract_address.map(|contract_addr| {
+		if let Some(contract_addr) = new_contract_address {
```
tomusdrw: This is a change in the logic. It keeps the old address even if you pass `None` as `new_contract_address`

secret_store/src/lib.rs
```diff
@@ -43,11 +43,16 @@ extern crate ethereum_types;
 extern crate ethcrypto;
 extern crate ethkey;
 extern crate ethsync;
-extern crate native_contracts;
 extern crate keccak_hash as hash;
 extern crate kvdb;
 extern crate kvdb_rocksdb;
 
+extern crate ethabi;
```
tomusdrw: Should be joined with other imports imho (and sorted)

updater/src/lib.rs
```diff
@@ -16,23 +16,26 @@
 
 //! Updater for Parity executables
 
-#[macro_use] extern crate log;
+#[macro_use]
```
tomusdrw: We usually have a convention of having separate section for `macro_use` extern crates.

ethcore/light/src/client/header_chain.rs
debris: there is also `header.set_number(i)` which uses `i`. So one of them have to be cased

ethcore/node_filter/src/lib.rs
debris: it's not. it exposes `use_contract!` macro which is used

ethcore/src/engines/validator_set/contract.rs
debris: good catch!

ethcore/src/engines/validator_set/contract.rs
debris: you are right ;)

debris: definitely not, that's just sloppy refactoring :p

ethcore/src/tx_filter.rs
debris: native_contracts_generator was implicitly doing these conversion, when type was `u32`. `ethabi_derive` does not do that

evmbin/src/display/mod.rs
debris: ok, I'll try

hash-fetch/src/urlhint.rs
debris: cause `client.call` returns a future and async calls and not supported in the current version of `ethabi_derive`. afaik, that's the only place where async calls where actually used in parity

secret_store/src/key_server_set.rs
debris: great catch!

ethcore/node_filter/src/lib.rs
tomusdrw: (I mean the new line :))

debris: kk :)

ethcore/src/engines/validator_set/contract.rs
debris: actually, I can'y do that, cause `EngineClient` is a trait.

```
cargo check -p ethcore
   Compiling ethcore v1.9.0 (file:///Users/marek/projects/ethcore/parity/ethcore)
error[E0277]: the trait bound `client::traits::EngineClient: std::marker::Sized` is not satisfied
  --> ethcore/src/engines/validator_set/contract.rs:48:24
   |
48 |    client: RwLock::new(Weak::new()),
   |                        ^^^^^^^^^ `client::traits::EngineClient` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `client::traits::EngineClient`
   = note: required by `<std::sync::Weak<T>>::new`
```

debris:
@tomusdrw thanks for review! Can you take a look at the pr again? 😉

updater/src/updater.rs
```diff
@@ -161,41 +184,68 @@ impl Updater {
...
 
-			Ok(OperationsInfo {
-				fork: operations.latest_fork()? as u64,
-				this_fork: this_fork,
-				track: in_track,
-				minor: in_minor,
-			})
-		} else {
-			Err("Operations not available".into())
+			in_minor = Some(self.collect_release_info(latest_in_track)?);
```
tomusdrw: Imho would be cleaner to use `loop` instead of `while` (and enclose the whole thing in a function).
```rust
let (in_track, in_minor) = self.find_versions(self.track());

// impl
fn find_versions(&self, current_track: _) -> (_, Option<_>) {
  let mut version_in_track = None;
  let mut track = current_track;
  loop {
    let latest_in_track = ...;
    let in_minor = self.collect_release_info(latest_in_track)?;
    // update initial version_in_track
    version_in_track = version_in_track.or_else(|| Some(in_minor.clone()));
    // check version
    if in_minor.version.version.minor == self.this.version.minor {
       return (version_in_track.unwrap(), Some(in_minor));
    }
    // check next track
    track = match in_minor.track {
      ReleaseTrack::Beta => ReleaseTrack::Stable,
      ReleaseTrack::Nightly => ReleaseTrack::Beta,
      _ => return (version_in_track.unwrap(), None),
    };
  }
}

tomusdrw: One more idea:
```rust

let in_track = self.find_version(self.track());
let in_minor = self.find_minor_version(in_track);


// impl
fn find_minor_version(&self, start_track: _) -> Option<_> {
   let mut track = start_track;
   while track.version.version.minor != self.this.version.minor {
     let next_track = match track.track {
      ReleaseTrack::Beta => ReleaseTrack::Stable,
      ReleaseTrack::Nightly => ReleaseTrack::Beta,
      _ => return None,
     };
     track = self.find_version(next_track);
   }
   return Some(track);
}

fn find_version(track: _) -> _ {
    let release_info = ...;
    self.collect_release_info(release_info)?;
}

Issue #7731, opened 2018-01-30
Pub/Sub newHeads is not compatible with geth


tjade273:
> I'm running:
>
> - **Which Parity version?**: 1.8.6
> - **Which operating system?**: Linux
> - **How installed?**: via installer 
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: all
> - **Did you try to restart the node?**: yes

The Geth ethclient.go RPC client sends an empty dictionary as the second argument to `eth_subscribe` with `newHeads`.

https://github.com/ethereum/go-ethereum/blob/762f3a48a00da02fe58063cb6ce8dc2d08821f15/ethclient/ethclient.go#L298

The Parity server expects no second argument, leading to an error:
> "Expected no parameters."

https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/rpc/src/v1/impls/eth_pubsub.rs#L263

The documentation includes examples with and without the empty dictionary.  

https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB

debris:
Closing, cause both clients are now compatible

PR #7732, opened 2018-01-30
Loosened restrictions on parameters to newHead subscription

tjade273:
Addresses #7731 

Currently the patch simply allows any parameter that is parseable as a log, then ignores it.

We could instead restrict it to only the empty log, but this is more general.

### DIFF
--- a/rpc/src/v1/impls/eth_pubsub.rs
+++ b/rpc/src/v1/impls/eth_pubsub.rs
@@ -251,17 +251,15 @@ impl<C: Send + Sync + 'static> EthPubSub for EthPubSubClient<C> {
 		kind: pubsub::Kind,
 		params: Trailing<pubsub::Params>,
 	) {
+
 		let error = match (kind, params.into()) {
-			(pubsub::Kind::NewHeads, None) => {
+			(pubsub::Kind::NewHeads, _) => {
 				self.heads_subscribers.write().push(subscriber);
 				return;
 			},
 			(pubsub::Kind::Logs, Some(pubsub::Params::Logs(filter))) => {
 				self.logs_subscribers.write().push(subscriber, filter.into());
 				return;
-			},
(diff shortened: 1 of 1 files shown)


tjade273:
[clabot:check].

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -251,18 +251,16 @@ impl<C: Send + Sync + 'static> EthPubSub for EthPubSubClient<C> {
 		kind: pubsub::Kind,
 		params: Trailing<pubsub::Params>,
 	) {
+
 		let error = match (kind, params.into()) {
-			(pubsub::Kind::NewHeads, None) => {
+			(pubsub::Kind::NewHeads, _) => {
```
tomusdrw: I don't like loosening it completely. I think the RPC and the spec should be as specific as possible, we'll run into plenty of issues if we allow leniency of parameters (like developers wondering why a parameter doesn't work if they make a typo or the parameter is not supported).

IMHO The spec just not got properly updated after `newBlocks` has been renamed to `newHeads`, `includeTransactions` paramteter in examples doesn't make much sense in case of headers-only subscription.

Having said that, I'm ok with allowing:
`eth_subscribe('newHeads')` and `eth_subscribe('newHeads', {})` call (optional options argument that we might add in the future), but not `eth_subscribe('newHeads', {'anything': true})`.

@tjade273 what do you think?

tomusdrw:
BTW: would be good to add tests for both behaviours.

rpc/src/v1/impls/eth_pubsub.rs
tjade273: Makes sense, I'll tweak it

5chdn:
Let me know when you are ready, @tjade273 :)

PR #7741, opened 2018-01-30
Transaction-pool fixes

tomusdrw:
Exposing some missing stuff so that it can be published at crates.io and used in polkadot.


transaction-pool/src/pool.rs
```diff
@@ -109,6 +109,8 @@ impl<T, S, L> Pool<T, S, L> where
 
 		ensure!(!self.by_hash.contains_key(transaction.hash()), error::ErrorKind::AlreadyImported(*transaction.hash()));
 
+		// TODO [ToDr] Most likely move this after the transsaction is inserted.
```
svyatonik: typo

PR #7742, opened 2018-01-30
[beta] Token filter balances (throttle)

jacogr:
- Split token retrievals into 128-item sized chunks.
- Align `parity-js/shared` for shell
- Tested with `cargo build --release --=no-default-features --features ui`

parity 2018-01-31 10-22-00


5chdn:
Can we also have this fix in stable?

PR #7796, opened 2018-02-02
Updated WASM Runtime & new interpreter (wasmi)

NikVolf:
cc @pepyakin 

- mem* removed from the runtime, we'll add it back if optimization needed
- wasmi replaces deprecated parity-wasm interpreter
- wasm32-unknown-unknown import/export name conventions
- dropped emscripten runtime glue completely (STACKTOP, llvmtrap, memory base, etc..)
- added wasm opcodes scaling factor of 3/8


ethcore/wasm/src/runtime.rs
```diff
@@ -59,322 +50,266 @@ pub enum UserTrap {
...
+		self.result = self.memory.get(ptr, len as usize)?;
+
+		Ok(())
+	}
+
+	pub fn result(&self) -> &[u8] {
+		&self.result
+	}
+
+	pub fn dissolve(self) -> Vec<u8> {
```
pepyakin: How about `into_result`?

ethcore/wasm/src/lib.rs
```diff
@@ -61,139 +58,97 @@ impl From<Error> for vm::Error {
...
+			&loaded_module,
+			&wasmi::ImportsBuilder::new().with_resolver("env", &instantiation_resolover)
+		).map_err(Error)?;
 
 		if params.gas > ::std::u64::MAX.into() {
 			return Err(vm::Error::Wasm("Wasm interpreter cannot run contracts with gas >= 2^64".to_owned()));
 		}
 
+		let initial_memory = instantiation_resolover.memory_size().map_err(Error)?
+			.saturating_sub(MEMORY_STIPEND);
```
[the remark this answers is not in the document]
NikVolf: I moved to the schedule

NikVolf:
@pepyakin addressed

ethcore/wasm/src/env.rs
```diff
@@ -14,183 +14,276 @@
...
+	max_memory: u32,
+	memory: RefCell<Option<MemoryRef>>,
+}
+
+impl ImportResolver {
+	/// New import resolver with specifed maximum amount of inital memory.
+	pub fn with_limit(max_memory: u32) -> ImportResolver {
+		ImportResolver {
+			max_memory: max_memory,
+			.. Default::default()
```
pepyakin: Is this supposed to be memory `(0, 0)`?

pepyakin: Ah, I see. 
Maybe it would be better to explicitly initialize `memory` 
like `RefCell::new(None)` ?

NikVolf: changed!

ethcore/wasm/src/env.rs
```diff
@@ -14,183 +14,276 @@
...
+	}
+
+	/// Returns memory that was instantiated during the contract module
+	/// start. If contract does not use memory at all, the dummy memory of length (0, 0)
+	/// will be created instead. So this method always returns memory instance
+	/// unless errored.
+	pub fn memory_ref(&self) -> Result<MemoryRef, Error> {
+		{
+			let mut mem_ref = self.memory.borrow_mut();
+			if mem_ref.is_none() { *mem_ref = Some(MemoryInstance::alloc(0, Some(0))?); }
```
pepyakin: This invocation actually can't return error since params `(0, Some(0))` are perfectly valid

ethcore/wasm/src/env.rs
```diff
@@ -14,183 +14,276 @@
...
+/// Also manages initial memory request from the runtime.
+#[derive(Default)]
+pub struct ImportResolver {
+	max_memory: u32,
+	memory: RefCell<Option<MemoryRef>>,
+}
+
+impl ImportResolver {
+	/// New import resolver with specifed maximum amount of inital memory.
+	pub fn with_limit(max_memory: u32) -> ImportResolver {
```
pepyakin: I think it is not clear that `max_memory` is denoted in pages. Maybe it would be better if we leave a comments on this? Or maybe change the name to something like `max_memory_pages`

ethcore/wasm/src/env.rs
```diff
@@ -14,183 +14,276 @@
...
+	}
+
+	fn resolve_memory(
+		&self,
+		field_name: &str,
+		descriptor: &MemoryDescriptor,
+	) -> Result<MemoryRef, Error> {
+		if field_name == "memory" {
+			if descriptor.initial() >= self.max_memory ||
+				(descriptor.maximum().map_or(false, |m| m < descriptor.initial() || m > self.max_memory))
```
pepyakin: This is an interesting check.

If I understand the intent correctly, it checks whether `initial` or `maximum` number of memory pages requested by the module is greater than some limit. In case if `maximum` isn't provided then this check is passed and runtime proceeds to execution.

If that's correct, then I don't see the point. What stops the contract writer to not set `maximum` limit on memory and then allocate amount of memory larger than the limit (don't we have the control on `grow_memory` op, do we?).

And as I remember, don't we agreed that we don't want to restrict the upper memory limit by any means except the gas price?

ethcore/wasm/src/env.rs
```diff
@@ -14,183 +14,270 @@
...
+use std::cell::RefCell;
+use wasmi::{
+	self, Signature, Error, FuncRef, FuncInstance, MemoryDescriptor,
+	MemoryRef, MemoryInstance,
+};
+
+/// Internal ids all functions runtime supports. This is just a glue for wasmi interpreter
+/// that lacks high-level api and later will be factored out
+pub mod ids {
+	pub const STORAGE_WRITE_FUNC: usize = 0;
```
pepyakin: I don't see the point of this 10-step numbering.
Also, maybe we can use some macro which will generate C-like enum, so we don't have to number this constants by hand?

ethcore/wasm/src/runtime.rs
```diff
@@ -59,334 +50,317 @@ pub enum UserTrap {
...
 		if !self.charge_gas(amount as u64) {
-			Err(UserTrap::GasLimit.into())
+			Err(Error::GasLimit)
 		} else {
 			Ok(())
 		}
 	}
 
-	pub fn overflow_charge<F>(&mut self, f: F) -> Result<(), InterpreterError>
+	/// Adjusted charge applies wasm divisor coeff as
```
pepyakin: Im not sure what you meant here...

ethcore/wasm/src/env.rs
NikVolf: But i don't want to depend on this invariant from other lib

ethcore/wasm/src/env.rs
NikVolf: It's implementation detail, will refactored once more high-level api for wasmi is available

pepyakin: Ok, I think I'm fine with it.

ethcore/wasm/src/runtime.rs
```diff
@@ -59,334 +50,317 @@ pub enum UserTrap {
...
+		Ok(())
+	}
+
+	/// Return currently used schedule.
+	pub fn schedule(&self) -> &vm::Schedule {
+		self.ext.schedule()
+	}
+
+	/// Return. Syscall takes 2 arguments - pointer in sandboxed memory where result is and
+	/// the lendth of the result.
```
pepyakin: typo: lendth → length

ethcore/wasm/src/env.rs
pepyakin: But otherwise this code embeds this invariant into the consensus

ethcore/wasm/src/runtime.rs
```diff
@@ -59,334 +50,317 @@ pub enum UserTrap {
...
 	{
+		let gas: u64 = args.nth(0)?;
+		let address = self.address_at(args.nth(1)?)?;
 
-		trace!(target: "wasm", "runtime: call code");
-		let mut context = context;
-		let result_alloc_len = context.value_stack.pop_as::<i32>()? as u32;
-		trace!(target: "wasm", "    result_len: {:?}", result_alloc_len);
+		let vofs = if use_val { 1 } else { 0 };
+		let val = if use_val { Some(self.u256_at(args.nth(2)?)?) } else { None };
```
pepyakin: I think such sorta dynamic signatures might lead to a disaster.
Can we fix this somehow? For example, change `pwasm-std` in a such way that we always pass some dummy `val` parameter?
I think even duplicating parameter decoding in methods `*call` below would be a better solution.

NikVolf: It won't lead to disaster if there is test that shows that all arguments are mapped correctly, which is the case here

pepyakin: Can we do this multiplexing on the pwasm-std side? This way we'll get rid of all this `dcall`/`scall` and etc exports.

ethcore/wasm/src/runtime.rs
```diff
@@ -402,449 +376,208 @@ impl<'a, 'b> Runtime<'a, 'b> {
...
 		//
+		trace!(target: "wasm", "runtime: CREATE");
+		let endowment = self.u256_at(args.nth(0)?)?;
+		trace!(target: "wasm", "       val: {:?}", endowment);
+		let code_ptr: u32 = args.nth(1)?;
+		trace!(target: "wasm", "  code_ptr: {:?}", code_ptr);
+		let code_len: u32 = args.nth(2)?;
+		trace!(target: "wasm", "  code_len: {:?}", code_len);
+		let result_ptr: u32 = args.nth(3)?;
+		trace!(target: "wasm", "result_ptr: {:?}", result_ptr);
```
pepyakin: I think this code will more legible if we won't interleave argument decoding code with displaying it, as we do in `do_call`.
At least, I'm wondering, should we synchronize them?

ethcore/wasm/src/runtime.rs
NikVolf: I don't think so, these are mostly different operations, and might be handled completely differently lately, so it might be good idea to differentiate them at link time

also, EVM has different opcodes for those, so it's good to have 1-1 mapping `opcode<->wasm external`

ethcore/wasm/src/runtime.rs
NikVolf: why?

ethcore/wasm/src/lib.rs
```diff
@@ -61,139 +56,111 @@ impl From<Error> for vm::Error {
...
-		let static_segment_cost = data_section_length * runtime.ext().schedule().wasm.static_region as u64;
-		runtime.charge(|_| static_segment_cost).map_err(Error)?;
-
-		let d_ptr = {
-			match params.params_type {
-				vm::ParamsType::Embedded => {
-					runtime.write_descriptor(
-						if data_position < code.len() { &code[data_position..] } else { &[] }
-					).map_err(Error)?
+		// We count all memory except free (MEMORY_STIPEND)
```
pepyakin: What's the MEMORY_STIPEND?

NikVolf:
@pepyakin all addressed

PR #7807, opened 2018-02-05
Fix build

NikVolf:


NikVolf:
it's broken, because Cargo.lock is also not up-to-date

if you checkout master and do `./test.sh`, it will rebuild Cargo.lock and fail to compile

debris:
ahh, you are right! I thought you changed `0.1 -> 0.2`, but it is `0.2 -> 0.1`

PR #7808, opened 2018-02-05
Update references to UI shell & wallet

jacogr:
- Update references on `Cargo.toml` with `cargo update -p parity-ui`
- Updates references to UI with visible token balances


5chdn:
What is this used for?

jacogr:
As it stands, Parity uses external UI crates when building the binary to serve over :8180. (The old js/js-old folders now in separate repos.) As soon as the stand-alone shell is available, will be dropped, Rome was not built in a day, so still need to make sure the UI as it stands works.

PR #7812, opened 2018-02-05
Adjust storage update evm-style

NikVolf:
This will charge for setting already initialized storage entries less, also will track refunded (storage set back to zero from non-zero)


ethcore/wasm/src/runtime.rs
```diff
@@ -218,11 +218,21 @@ impl<'a> Runtime<'a> {
...
 
-		self.adjusted_charge(|schedule| schedule.sstore_set_gas as u64)?;
-
 		let val = self.h256_at(val_ptr)?;
+		let former_val = self.ext.storage_at(&key).map_err(|_| Error::StorageUpdateError)?;
+
+		if former_val == H256::zero() && val != H256::zero() {
+			self.adjusted_charge(|schedule| schedule.sstore_set_gas as u64)?;
+		} else {
+			self.adjusted_charge(|schedule| schedule.sstore_reset_gas as u64)?;
```
pepyakin: So the decision table for this check looks like this:

case |  rate     |
------------ | -------------
0 → 0  |     reset |
0 → ≠0 |     set
≠0 → 0 |     reset
≠0 → ≠0   | reset

Is that intended?

NikVolf: Yes

NikVolf: != 0 -> 0

also leads to `storage_clears ++` which is later handled by inter-vm rules

pepyakin: Ah, I see, thanks.

PR #7832, opened 2018-02-07
Store updater metadata in a single place

tomusdrw:
Move all version related stuff to `util/version/Cargo.toml`.
Updater service will be able to read the toml file and extract all required info.

Grouping the data together will also increase the chance that the stuff is actually modified. We had `FORK_SUPPORTED_KOVAN: u64 = 0` since probably no one realised where it is and that it's required to update that.

Bonus: removes completely unused `SemanticVersion` stuff.


util/version/Cargo.toml
```diff
@@ -1,10 +1,23 @@
...
 authors = ["Parity Technologies <admin@parity.io>"]
 build = "build.rs"
 
+[package.metadata]
+# This versions track. Should be changed to `stable` or `beta` when on respective branches.
+# Used by auto-updater and for Parity version string.
+track = "nightly"
+
+# Latest supported fork blocks for various networks. Used ONLY by auto-updater.
+[package.metadata.forks]
```
andresilva: Where are these values used?

debris: I guess nowhere in parity... `Used ONLY by auto-updater` so https://github.com/paritytech/push-release . am I right @tomusdrw ?

PR #7842, opened 2018-02-08
WASM: Disable internal memory

NikVolf:
It's useless for meaningful contracts, while having it opens network for dos attack

### DIFF
--- a/ethcore/wasm/src/parser.rs
+++ b/ethcore/wasm/src/parser.rs
@@ -56,12 +56,20 @@ pub fn payload<'a>(params: &'a vm::ActionParams, schedule: &vm::Schedule)
 		},
 	};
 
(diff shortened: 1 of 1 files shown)


rphmeier:
if it's metered the same way as the regular (?) memory is it really a DoS attack?

NikVolf:
@rphmeier sure, could be, but webassembly spec does not allow to judge about module internal memory (as long as it is not exported) during execution

while we of course can hack around and meter internal memoery it by the module composition before instantiation, I would prefer to avoid it at all so other implementations that use different engines and api-s (which might not allow this hack) could match rules

So I can still add metering, but it's really hard to imagine contract that does not use shared memory with runtime, it can do almost nothing.

NikVolf:
@rphmeier 
tl;dr; We gain almost nothing while impede compatibility with other implementations

PR #7846, opened 2018-02-09
Fix checksums and auto-update push

tomusdrw:
Should go after https://github.com/paritytech/push-release/pull/9 is deployed.

- Common way to compute md5/sha256/sha3 sums
- Safe curl wrapper, which fails the build on non-2xx exit codes.
- Use more specific `BUILD_PLATFORM` instead of generic `PLATFORM` when pushing build to updater.


General-Beck:
it is necessary to check in ```nightly```

5chdn:
Thanks so much, this was actually on my agenda. Best way to test this is merge and bump nightly, so I agree with @General-Beck here.

tomusdrw:
Yup, sure. Just wait for the new updater to be deployed.

5chdn:
Or actually, let me tag this PR once it's deployed. Any ETA?

tomusdrw:
@5chdn only `nightly` or `^v\d+.\d+.\d$` tags will now be accepted by the updater. So I think we should do both:
1. Tag master (should go fine, but should be rejected by the updater)
2. Push to `nightly` - this should push a new release on kovan.
3. By tagging before the updater is released we can also take a path where the build fails, because updater crashes.

5chdn:
cc  @fevo1971

5chdn:
This fail is probably related: https://gitlab.parity.io/parity/parity/-/jobs/78627
    push release
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
    
      0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
    100   228  100    76  100   152   1374   2748 --:--:-- --:--:-- --:--:--  2763
    child "sha3" fails because ["sha3" must only contain hexadecimal characters]\n
    Unable to push info to updater service.

5chdn:
This probably too: https://gitlab.parity.io/parity/parity/-/jobs/78631

    Checksum calculation:
    Parity file SHA3: 39746347b20456e15081b540a06188eb41985a1cb56764350a52e53e7ba4db8d
    {:timestamp=>"2018-02-12T20:57:29.827336+0000", :message=>"File already exists, refusing to continue: parity-1.10.0-1.x86_64.rpm", :level=>:fatal}

PR #7848, opened 2018-02-09
Add a timeout for light client sync requests

### LINKED ISSUE #6319: Parity Light Client Sync Stuck

> I'm running:
>
> - **Parity version**: Parity/v1.8.0-unstable-b9c1d0bd1-20170814/x86_64-macos/rustc1.18.0
> - **And installed**: from source (`rustc 1.20.0-nightly (6d9d82d3d 2017-07-14)`)

---
I tried running `parity --light --scale-verifiers --num-verifiers 4 --fast-and-loose` (and `parity --light`) but both times I get stuck at:

    2017-08-16 21:42:48  Syncing #1925655 ac02…fc53     0 hdr/s      0+    0 Qed  #1925655    9/50 peers   2 MiB cache 0 bytes queue  RPC:  1 conn,  4 req/s, 139 µs

I would expect it to sync all the way to the current tip, which is currently 4168389.

tomaka:
Fix #6319 

This PR adds a 7 seconds timeout when we request a header range from a remote. If the timeout expires, the request will be re-assigned, most likely to a different node.

My machine is still syncing with this patch right now, but I've successfully synced more than 2 million blocks without dropping to 0 hdr/s at any point.

EDIT: Briefly dropped to 0 hdr/s around the 3 millionth block, but I think it's because my Internet connection briefly dropped. Sync resumed 10 seconds later.

EDIT2: Fully sync'ed from the beginning without any interruption except the one of my first edit.

### DIFF
--- a/sync/src/light_sync/mod.rs
+++ b/sync/src/light_sync/mod.rs
@@ -58,7 +58,12 @@ mod sync_round;
 #[cfg(test)]
 mod tests;
 
-const REQ_TIMEOUT_MILLISECS: u64 = 7000;
+// Base number of milliseconds for the header request timeout.
+const REQ_TIMEOUT_MILLISECS_BASE: u64 = 7000;
+// Additional number of milliseconds for each requested header.
+// If we request N headers, then the timeout will be:
+//  REQ_TIMEOUT_MILLISECS_BASE + N * REQ_TIMEOUT_MILLISECS_PER_HEADER
+const REQ_TIMEOUT_MILLISECS_PER_HEADER: u64 = 10;
 
 /// Peer chain info.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


tomaka:
The CI failure is in secret store and looks unrelated to my change.

tomaka:
I tried two other full synchronizations over the week end, and unfortunately the second one has gotten stuck, meaning that the issue isn't entirely solved :-/

rphmeier:
The network service already handles timeouts here: https://github.com/paritytech/parity/blob/master/ethcore/light/src/net/mod.rs#L88

although for groups of 256 headers I think it will end up timing out after a minute which is far too long. It should really be mapped onto something like a linear function something like `7 + 0.01x` where x is the number of headers requested.

5chdn:
Just tested, this does indeed resolve #6319 on first shot.

PR #7849, opened 2018-02-09
Kovan WASM fork code

NikVolf:
in chainspec json files, there is a breaking change with wasm flag, one should replace

`wasm: true` with `wasmActivationTransition: "0x0"`

### DIFF
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -70,9 +70,6 @@ unexpected = { path = "../util/unexpected" }
 journaldb = { path = "../util/journaldb" }
 tempdir = "0.3"
 
-[dev-dependencies]
-wabt = "0.1"
-
 [features]
 jit = ["evm/jit"]
 evm-debug = ["slow-blocks"]
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -70,6 +70,9 @@ unexpected = { path = "../util/unexpected" }
 journaldb = { path = "../util/journaldb" }
 tempdir = "0.3"
 
+[dev-dependencies]
+wabt = "0.1"
+
 [features]
 jit = ["evm/jit"]
 evm-debug = ["slow-blocks"]
--- a/ethcore/res/ethereum/kovan_wasm_test.json
+++ b/ethcore/res/ethereum/kovan_wasm_test.json
@@ -1,0 +1,74 @@
+{
+	"name": "Kovan",
+	"dataDir": "kovan",
+	"engine": {
+		"authorityRound": {
+			"params": {
+			"stepDuration": "4",
+			"blockReward": "0x4563918244F40000",
+				"validators" : {
+					"list": [
+						"0x00D6Cc1BA9cf89BD2e58009741f4F7325BAdc0ED",
+						"0x00427feae2419c15b89d1c21af10d1b6650a4d3d",
+						"0x4Ed9B08e6354C70fE6F8CB0411b0d3246b424d6c",
+						"0x0020ee4Be0e2027d76603cB751eE069519bA81A1",
+						"0x0010f94b296a852aaac52ea6c5ac72e03afd032d",
+						"0x007733a1FE69CF3f2CF989F81C7b4cAc1693387A",
+						"0x00E6d2b931F55a3f1701c7389d592a7778897879",
+						"0x00e4a10650e5a6D6001C38ff8E64F97016a1645c",
+						"0x00a0a24b9f0e5ec7aa4c7389b8302fd0123194de"
+					]
+				},
+				"validateScoreTransition": 1000000,
+				"validateStepTransition": 1500000,
+				"maximumUncleCountTransition": 5067000,
+				"maximumUncleCount": 0
+			}
+		}
+	},
+	"params": {
+		"gasLimitBoundDivisor": "0x400",
+		"registrar" : "0xfAb104398BBefbd47752E7702D9fE23047E1Bca3",
+		"maximumExtraDataSize": "0x20",
+		"minGasLimit": "0x1388",
+		"networkID" : "0x2A",
+		"forkBlock": 4297256,
+		"forkCanonHash": "0x0a66d93c2f727dca618fabaf70c39b37018c73d78b939d8b11efbbd09034778f",
+		"validateReceiptsTransition" : 1000000,
+		"eip155Transition": 1000000,
+		"validateChainIdTransition": 1000000,
+		"eip140Transition": 5067000,
+		"eip211Transition": 5067000,
+		"eip214Transition": 5067000,
+		"eip658Transition": 5067000,
+		"wasmActivation": 10
+	},
+	"genesis": {
+		"seal": {
+			"authorityRound": {
+				"step": "0x0",
+				"signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+			}
+		},
+		"difficulty": "0x20000",
+		"gasLimit": "0x5B8D80"
+	},
+	"accounts": {
+		"0x0000000000000000000000000000000000000001": { "balance": "1", "builtin": { "name": "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
+		"0x0000000000000000000000000000000000000002": { "balance": "1", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
+		"0x0000000000000000000000000000000000000003": { "balance": "1", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
+		"0x0000000000000000000000000000000000000004": { "balance": "1", "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
+		"0x0000000000000000000000000000000000000005": { "builtin": { "name": "modexp", "activate_at": 5067000, "pricing": { "modexp": { "divisor": 20 } } } },
+		"0x0000000000000000000000000000000000000006": { "builtin": { "name": "alt_bn128_add", "activate_at": 5067000, "pricing": { "linear": { "base": 500, "word": 0 } } } },
+		"0x0000000000000000000000000000000000000007": { "builtin": { "name": "alt_bn128_mul", "activate_at": 5067000, "pricing": { "linear": { "base": 40000, "word": 0 } } } },
+		"0x0000000000000000000000000000000000000008": { "builtin": { "name": "alt_bn128_pairing", "activate_at": 5067000, "pricing": { "alt_bn128_pairing": { "base": 100000, "pair": 80000 } } } },
+		"0x00521965e7bd230323c423d96c657db5b79d099f": { "balance": "1606938044258990275541962092341162602522202993782792835301376" }
+	},
+	"nodes": [
+		"enode://56abaf065581a5985b8c5f4f88bd202526482761ba10be9bfdcd14846dd01f652ec33fde0f8c0fd1db19b59a4c04465681fcef50e11380ca88d25996191c52de@40.71.221.215:30303",
+		"enode://d07827483dc47b368eaf88454fb04b41b7452cf454e194e2bd4c14f98a3278fed5d819dbecd0d010407fc7688d941ee1e58d4f9c6354d3da3be92f55c17d7ce3@52.166.117.77:30303",
+		"enode://8fa162563a8e5a05eef3e1cd5abc5828c71344f7277bb788a395cce4a0e30baf2b34b92fe0b2dbbba2313ee40236bae2aab3c9811941b9f5a7e8e90aaa27ecba@52.165.239.18:30303",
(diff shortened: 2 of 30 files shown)


ethcore/src/executive.rs
```diff
@@ -152,8 +153,13 @@ impl TransactOptions<trace::NoopTracer, trace::NoopVMTracer> {
...
 
-pub fn executor(machine: &Machine, vm_factory: &Factory, params: &ActionParams) -> Box<vm::Vm> {
-	if machine.supports_wasm() && params.code.as_ref().map_or(false, |code| code.len() > 4 && &code[0..4] == WASM_MAGIC_NUMBER) {
+pub fn executor(
+	machine: &Machine,
+	vm_factory: &Factory,
+	params: &ActionParams,
+	blocknumber: BlockNumber,
+) -> Box<vm::Vm> {
+	if machine.supports_wasm(blocknumber) && params.code.as_ref().map_or(false, |code| code.len() > 4 && &code[0..4] == WASM_MAGIC_NUMBER) {
```
debris: I don't like this change at all. The `if` statement that we have right now is terrible and this change makes it even worse. This function/interface needs to be completely reworked

ethcore/Cargo.toml
```diff
@@ -70,6 +70,9 @@ unexpected = { path = "../util/unexpected" }
 journaldb = { path = "../util/journaldb" }
 tempdir = "0.3"
 
+[dev-dependencies]
+wabt = "0.1"
```
debris: I refuse pulling a huge c++ dependency to ethcore as a mandatory dependency. It needs to be do differently.

ethcore/res/ethereum/kovan_wasm_test.json
```diff
@@ -0,0 +1,74 @@
...
+		"forkBlock": 4297256,
+		"forkCanonHash": "0x0a66d93c2f727dca618fabaf70c39b37018c73d78b939d8b11efbbd09034778f",
+		"validateReceiptsTransition" : 1000000,
+		"eip155Transition": 1000000,
+		"validateChainIdTransition": 1000000,
+		"eip140Transition": 5067000,
+		"eip211Transition": 5067000,
+		"eip214Transition": 5067000,
+		"eip658Transition": 5067000,
+		"wasmActivation": 10
```
tomusdrw: The convention seems to be `*Transition`, would keep it.

ethcore/src/machine.rs
```diff
@@ -377,8 +377,8 @@ impl EthereumMachine {
 	}
 
 	/// If this machine supports wasm.
-	pub fn supports_wasm(&self) -> bool {
-		self.params().wasm
+	pub fn supports_wasm(&self, blocknumber: BlockNumber) -> bool {
```
tomusdrw: Do we need a separate public method for that? perhaps use `Schedule` directly on the call site.

ethcore/src/spec/spec.rs
```diff
@@ -111,6 +111,8 @@ pub struct CommonParams {
 	pub remove_dust_contracts: bool,
 	/// Wasm support
 	pub wasm: bool,
```
tomusdrw: Why is that needed? Seems that it's just `wasmTransition: 0`. We already did breaking changes to spec files in the past, we should just highlight that in the release notes.

ethcore/vm/src/action_params.rs
```diff
@@ -112,6 +112,14 @@ impl Default for ActionParams {
 	}
 }
 
+impl ActionParams {
+	pub fn from_origin(mut self, address: Address) -> Self {
```
tomusdrw: Seems to be used only for tests, might be misleading to include it everwhere in the code.

ethcore/vm/src/schedule.rs
```diff
@@ -113,6 +113,8 @@ pub struct Schedule {
 	pub kill_dust: CleanDustMode,
 	/// Enable EIP-86 rules
 	pub eip86: bool,
+	/// Is wasm active
+	pub wasm_activated: bool,
```
tomusdrw: Maybe `wasm: Option<WasmCosts>` will be more future proof?

ethcore/src/executive.rs
tomusdrw: Yeah, perhaps it should just go to `vm_factory`? This function seems to be another factory that delegates to `vm_factory` in case it's not wasm.

```rust
vm_factory.create(params.gas, params.code, blocknumber)
```

ethcore/Cargo.toml
NikVolf: it's not mandatory, it's only for running tests

ethcore/src/executive.rs
NikVolf: I wanted to go this way, but last time I checked it was just not possible for some reason I can't recall. perhaps I should try again and find this reason again

ethcore/res/ethereum/kovan_wasm_test.json
NikVolf: "wasmTransition" might imply that evm no longer works, i think it's not the best convention in this case

tomusdrw: Maybe `enableWasmTransition` or `supportWasmTransition` then?

NikVolf: both `enableWasmTransition` and `supportWasmTransition` implies the possibilty of such transition at all, not some blocknumber where it is activated

it is not really about transition as eip-s

tomusdrw: `supportWasmTransition` and `enableWasmTransition` implies for me that `wasm is supported` or `wasm is enabled` starting at some block. Similar to how `eip658Transition` implies that `eip658` is active at some point (it doesn't disable any previous eips).
That said my main reason is to complain is to keep the naming consistent here, if in the future we support a fork to wasm-only network we can either have: `wasmOnlyTransition` or `disableEvmTransition` there. Introducing another convention will make the file even messier imho.

ethcore/Cargo.toml
debris: It's mandatory for everyone working on daily basis on the project. It adds `cmake` with all of it's burden to our toolchain.

NikVolf: np, but it will be just bytecode dump as a test

we already have c++ build tools prerequisites by the way (rocksdb will be first notable example), they just don't use cmake crate directly and instead use some obscure build scripts, so this cmake reference does not add anything to the list instead of one-time compilation of wabt

ethcore/vm/src/schedule.rs
NikVolf: how do you imagine this line then:

```
		if self.wasm || block_number >= self.wasm_activation {
			schedule.wasm_activated = true;
		}

```

in `CoimmonParams::update_schedule`

`schedule.wasm = Some(...<WHAT?>)` ?

where the prices will be held until they are suddenly active?

NikVolf: On the second thought, no need to keep them, they can be just default :)

NikVolf:
@debris , @tomusdrw 

all addressed!

ethcore/vm/src/schedule.rs
tomusdrw: Yeah, they can be default for now, but later probably come from spec (that's why I suggested merging them with the flag).

ethcore/wasm/src/lib.rs
```diff
@@ -79,8 +79,8 @@ impl vm::Vm for WasmInterpreter {
 			&wasmi::ImportsBuilder::new().with_resolver("env", &instantiation_resolover)
 		).map_err(Error)?;
 
-		let adjusted_gas = params.gas * U256::from(ext.schedule().wasm.opcodes_div) /
-			U256::from(ext.schedule().wasm.opcodes_mul);
+		let adjusted_gas = params.gas * U256::from(ext.schedule().wasm().opcodes_div) /
```
tomusdrw: Each call to `ext.schedule()` is virtual, can we cache it in a variable somewhere at the top?

ethcore/wasm/src/parser.rs
```diff
@@ -24,12 +24,12 @@ use parity_wasm::peek_size;
 fn gas_rules(schedule: &vm::Schedule) -> rules::Set {
 	rules::Set::new({
 		let mut vals = ::std::collections::HashMap::with_capacity(4);
-		vals.insert(rules::InstructionType::Load, schedule.wasm.mem as u32);
-		vals.insert(rules::InstructionType::Store, schedule.wasm.mem as u32);
-		vals.insert(rules::InstructionType::Div, schedule.wasm.div as u32);
-		vals.insert(rules::InstructionType::Mul, schedule.wasm.mul as u32);
+		vals.insert(rules::InstructionType::Load, schedule.wasm().mem as u32);
```
tomusdrw: Just wondering, maybe all these methods could just do with `WasmCost` instead of entire schedule?

ethcore/wasm/src/lib.rs
NikVolf: I don't think so, it's mutably borrows self all the way down the end of the function few times

I think it will be also optimized / devirtualized anyway

rphmeier:
Prefer to put this in after #7605

ethcore/res/ethereum/kovan_wasm_test.json
```diff
@@ -0,0 +1,74 @@
+{
+	"name": "Kovan",
+	"dataDir": "kovan",
```
rphmeier: these should reflect the test-ness of the spec

ethcore/res/ethereum/kovan_wasm_test.json
```diff
@@ -0,0 +1,74 @@
...
+		"0x0000000000000000000000000000000000000005": { "builtin": { "name": "modexp", "activate_at": 5067000, "pricing": { "modexp": { "divisor": 20 } } } },
+		"0x0000000000000000000000000000000000000006": { "builtin": { "name": "alt_bn128_add", "activate_at": 5067000, "pricing": { "linear": { "base": 500, "word": 0 } } } },
+		"0x0000000000000000000000000000000000000007": { "builtin": { "name": "alt_bn128_mul", "activate_at": 5067000, "pricing": { "linear": { "base": 40000, "word": 0 } } } },
+		"0x0000000000000000000000000000000000000008": { "builtin": { "name": "alt_bn128_pairing", "activate_at": 5067000, "pricing": { "alt_bn128_pairing": { "base": 100000, "pair": 80000 } } } },
+		"0x00521965e7bd230323c423d96c657db5b79d099f": { "balance": "1606938044258990275541962092341162602522202993782792835301376" }
+	},
+	"nodes": [
+		"enode://56abaf065581a5985b8c5f4f88bd202526482761ba10be9bfdcd14846dd01f652ec33fde0f8c0fd1db19b59a4c04465681fcef50e11380ca88d25996191c52de@40.71.221.215:30303",
+		"enode://d07827483dc47b368eaf88454fb04b41b7452cf454e194e2bd4c14f98a3278fed5d819dbecd0d010407fc7688d941ee1e58d4f9c6354d3da3be92f55c17d7ce3@52.166.117.77:30303",
```
rphmeier: what are these bootnodes?

ethcore/vm/src/tests.rs
```diff
@@ -85,6 +85,11 @@ impl FakeExt {
 		ext.schedule = Schedule::new_byzantium();
 		ext
 	}
+
+	pub fn with_wasm(mut self) -> Self {
```
rphmeier: doc-string

ethcore/res/ethereum/kovan_wasm_test.json
NikVolf: it's just copy from `kovan.json`

NikVolf:
@rphmeier it's easy to check (just merge 2 branches in 1), there is no conflicts with #7605 for now

json/src/spec/params.rs
```diff
@@ -117,6 +115,11 @@ pub struct Params {
 	/// Transaction permission contract address.
 	#[serde(rename="transactionPermissionContract")]
 	pub transaction_permission_contract: Option<Address>,
+	/// Wasm support flag
+	pub wasm: Option<bool>,
```
tomusdrw: Is it still needed? What if it's set to `wasm: true && wasm_activiation_transition: 500` is it activated right away or at block `500`?

IMHO we should keep spec files clean (even if it requires a migration of existing spec files), especially given `wasm` was always experimental to date.

ethcore/src/factory.rs
```diff
@@ -15,14 +15,44 @@
...
+
+/// Virtual machine factory
+#[derive(Default, Clone)]
+pub struct VmFactory {
+	evm: EvmFactory,
+}
+
+impl VmFactory {
+	pub fn create(&self, params: &ActionParams, schedule: &Schedule) -> Box<Vm> {
+		if schedule.wasm.is_some() && params.code.as_ref().map_or(false, |code| code.len() > 4 && &code[0..4] == WASM_MAGIC_NUMBER) {
```
tomusdrw: Does WASM work with `gas > u64::max_value()`? I think I saw couple of `low_u64` calls below.

NikVolf: It will fail later in the vm itself:

https://github.com/paritytech/parity/blob/master/ethcore/wasm/src/lib.rs#L85

json/src/spec/params.rs
NikVolf: no, not needed, probably should mention that it requires to change in json spec to replace

`wasm: true` with `wasmActivationTransition: "0x0"`

PR #7855, opened 2018-02-10
Update gitlab-build.sh

General-Beck:
fix build ```version``` after https://github.com/paritytech/parity/pull/7723


General-Beck:
@debris before the script looked for the version by other rules

debris:
I believe you meant #7832 . Actually that pr is tagged as `patch`, so this one should be as well

5chdn:
yes, just figured we need this on beta now

Issue #7793, opened 2018-02-02
investigate if thread::sleep is really needed in hw module


debris:
it may potentially fix: #7516

suspicious code snippets:

https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/hw/src/lib.rs#L146-L158

https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/hw/src/lib.rs#L190-L204

https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/hw/src/ledger.rs#L239-L245

https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/hw/src/trezor.rs#L206-L212

and check out these changes:
https://github.com/dcuddeback/libusb-rs/compare/master...paritytech:master

debris:
cc @niklasad1

niklasad1:
@debris 
I think you can close this issue but just the clarify why the delay is necessary (but these are moved into separate threads now and should not block ```the main thread```)

1. At least Trezor need some time boot-up ~250 ms 
2. The loop iterations are unnecessary if a long enough delay is introduced
3. There are two cases:
    * The device is connected when parity starts (the delay could probably be skipped but that seems a little bit risky because we don't know how long the device have been connected)
    * A device connects/disconnects when parity is running

All this should be fixed in [#7860](https://github.com/paritytech/parity/pull/7860) but it is at least ```investigated```

Issue #7516, opened 2018-01-10
Parity crashes when usb wireless dongle for the keyboard/mouse is removed


zabirauf:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Parity version?v1.7.8-stable-d5fcf3b-20171025
> - **Which operating system?**: Windows
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: no
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**:  yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

I have a wireless keyboard and mouse connected (specifically Microsoft Sculp Ergonomic Keyboard). I have parity running with the following command `parity.exe ui --base-path PATH_TO_BASE_DIRECTORY`. When I take out my wireless dongle for the keyboard mouse parity crashes. Not sure how it is linked but I'm able to consistently reproduce the issue.

![image](https://user-images.githubusercontent.com/1104560/34765640-e3f4d7c4-f5a6-11e7-8b1c-e8de1feb213b.png)

I see the following in the event logs

```
Fault bucket 1723688955156595548, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: parity.exe
P2: 0.0.0.0
P3: 59f07900
P4: parity.exe
P5: 0.0.0.0
P6: 59f07900
P7: c0000005
P8: 0000000000e52ebe
P9: 
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER5C5.tmp.mdmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER673.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER66F.tmp.csv
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER69F.tmp.txt

These files may be available here:
C:\ProgramData\Microsoft\Windows\WER\ReportArchive\AppCrash_parity.exe_ae225145735dc7c4eb453db6f25b6ba7036cf_c98f36eb_2b3409cb

Analysis symbol: 
Rechecking for solution: 0
Report Id: bef8afb9-a58d-4ba0-91c5-c88f78de6812
Report Status: 268435456
Hashed bucket: 6b717adcf6f028a387ebc5f8e35f835c
```

```
[...]

PR #7860, opened 2018-02-11
Hardware-wallet/usb-subscribe-refactor

niklasad1:
Hi,

This PR changes how events from ```libusb/Hotplug``` are subscribed in more fine-grained manner i.e, by manufacturer_id, product_id and USB class. 

By doing this it fixes [#7516](https://github.com/paritytech/parity/issues/7516) & [#7793](https://github.com/paritytech/parity/issues/7793) it seems to be Windows only issue because we tested it earlier on both Linux/Ubuntu and OSX

### DIFF
--- a/hw/src/ledger.rs
+++ b/hw/src/ledger.rs
@@ -20,18 +20,21 @@
 use std::cmp::min;
 use std::fmt;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::time::Duration;
+use std::thread;
 
 use ethereum_types::{H256, Address};
 use ethkey::Signature;
 use hidapi;
+use libusb;
 use parking_lot::{Mutex, RwLock};
 
 use super::{WalletInfo, KeyPath};
 
-const LEDGER_VID: u16 = 0x2c97;
-const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001]; // Nano S and Blue
+// Expose these to the callback configuration
+pub const LEDGER_VID: u16 = 0x2c97;
+pub const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001]; // Nano S and Blue
 const ETH_DERIVATION_PATH_BE: [u8; 17] = [4, 0x80, 0, 0, 44, 0x80, 0, 0, 60, 0x80, 0, 0, 0, 0, 0, 0, 0]; // 44'/60'/0'/0
 const ETC_DERIVATION_PATH_BE: [u8; 21] = [5, 0x80, 0, 0, 44, 0x80, 0, 0, 60, 0x80, 0x02, 0x73, 0xd0, 0x80, 0, 0, 0, 0, 0, 0, 0]; // 44'/60'/160720'/0'/0
 
@@ -234,16 +237,10 @@ impl Manager {
 	fn open_path<R, F>(&self, f: F) -> Result<R, Error>
 		where F: Fn() -> Result<R, &'static str>
 	{
-		let mut err = Error::KeyNotFound;
-		// Try to open device a few times.
-		for _ in 0..10 {
-			match f() {
-				Ok(handle) => return Ok(handle),
-				Err(e) => err = From::from(e),
-			}
-			::std::thread::sleep(Duration::from_millis(200));
+		match f() {
+			Ok(handle) => Ok(handle),
+			Err(e) => Err(From::from(e)),
 		}
-		Err(err)
 	}
 
 	fn send_apdu(handle: &hidapi::HidDevice, command: u8, p1: u8, p2: u8, data: &[u8]) -> Result<Vec<u8>, Error> {
@@ -332,6 +329,40 @@ impl Manager {
 		let new_len = message.len() - 2;
 		message.truncate(new_len);
 		Ok(message)
+	}
+}
+
+pub struct EventHandler {
+	ledger: Weak<Manager>,
+}
+
+impl EventHandler {
+	pub fn new(ledger: Weak<Manager>) -> Self {
+		Self { ledger: ledger }
+	}
+}
+
+
+impl libusb::Hotplug for EventHandler {
+	fn device_arrived(&mut self, _device: libusb::Device) {
+		println!("ledger arrived");
+		Duration::from_millis(1000);
+		if let Some(t) = self.ledger.upgrade() {
+			// Wait for the device to bootup
+			thread::sleep(Duration::from_millis(1000));
+			if let Err(e) = t.update_devices() {
+				debug!("Ledger Connect Error: {:?}", e);
+			}
+		}
+	}
+
+	fn device_left(&mut self, _device: libusb::Device) {
+		println!("ledger left");
+		if let Some(l) = self.ledger.upgrade() {
+			if let Err(e) = l.update_devices() {
+				debug!("Ledger Disconnect Error: {:?}", e);
+			}
+		}
 	}
 }
 
--- a/hw/src/ledger.rs
+++ b/hw/src/ledger.rs
@@ -57,10 +57,14 @@ pub enum Error {
 	Protocol(&'static str),
 	/// Hidapi error.
 	Usb(hidapi::HidError),
+	/// Libusb error
+	LibUsb(libusb::Error),
 	/// Device with request key is not available.
 	KeyNotFound,
 	/// Signing has been cancelled by user.
 	UserCancel,
+	/// Invalid Product
+	InvalidProduct,
 }
 
 impl fmt::Display for Error {
@@ -68,8 +72,10 @@ impl fmt::Display for Error {
 		match *self {
 			Error::Protocol(ref s) => write!(f, "Ledger protocol error: {}", s),
 			Error::Usb(ref e) => write!(f, "USB communication error: {}", e),
+			Error::LibUsb(ref e) => write!(f, "LibUSB communication error: {}", e),
 			Error::KeyNotFound => write!(f, "Key not found"),
 			Error::UserCancel => write!(f, "Operation has been cancelled"),
+			Error::InvalidProduct=> write!(f, "Unsupported product was entered"),
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


niklasad1:
[clabot:check].

hw/src/ledger.rs
```diff
@@ -20,18 +20,21 @@
...
+use libusb;
 use parking_lot::{Mutex, RwLock};
 
 use super::{WalletInfo, KeyPath};
 
-const LEDGER_VID: u16 = 0x2c97;
-const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001]; // Nano S and Blue
+// Expose these to the callback configuration
+pub const LEDGER_VID: u16 = 0x2c97;
+pub const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001]; // Nano S and Blue
```
rphmeier: needs documentation for public items

hw/src/ledger.rs
```diff
@@ -234,16 +249,10 @@ impl Manager {
...
-		// Try to open device a few times.
-		for _ in 0..10 {
-			match f() {
-				Ok(handle) => return Ok(handle),
-				Err(e) => err = From::from(e),
-			}
-			::std::thread::sleep(Duration::from_millis(200));
+		match f() {
+			Ok(handle) => Ok(handle),
+			Err(e) => Err(From::from(e)),
```
rphmeier: `f().map_err(Into::into)`

hw/src/ledger.rs
```diff
@@ -335,6 +344,53 @@ impl Manager {
 	}
 }
 
+pub struct EventHandler {
```
rphmeier: needs docs

hw/src/ledger.rs
```diff
@@ -335,6 +344,53 @@ impl Manager {
...
+	ledger: Weak<Manager>,
+}
+
+impl EventHandler {
+	pub fn new(ledger: Weak<Manager>) -> Self {
```
rphmeier: and here

hw/src/lib.rs
```diff
@@ -33,13 +33,16 @@ use ethkey::{Address, Signature};
...
+use std::sync::Arc;
 use std::sync::atomic;
 use std::sync::atomic::AtomicBool;
 use std::thread;
 use std::time::Duration;
 use ethereum_types::U256;
 
+const USB_DEVICE_CLASS_DEVICE: u8 = 0;
+
+
```
rphmeier: redundant whitespace

hw/src/lib.rs
```diff
@@ -128,84 +131,81 @@ impl From<libusb::Error> for Error {
...
-		}
-	}
-}
 
 impl HardwareWalletManager {
 	pub fn new() -> Result<HardwareWalletManager, Error> {
-		let usb_context = Arc::new(libusb::Context::new()?);
+		let usb_context_trezor = Arc::new(libusb::Context::new()?);
+		let usb_context_ledger = Arc::new(libusb::Context::new()?);
+		let mut active_threads: Vec<Option<thread::JoinHandle<()>>> = Vec::with_capacity(10);
```
rphmeier: why bother setting the capacity to 10? (vec internally uses initial of 4 AFAIK)

hw/src/lib.rs
```diff
@@ -128,84 +131,81 @@ impl From<libusb::Error> for Error {
...
-		let thread_exiting = exiting.clone();
+		let thread_exiting_ledger = exiting.clone();
+		let thread_exiting_trezor = exiting.clone();
 		let l = ledger.clone();
 		let t = trezor.clone();
-		let thread = thread::Builder::new()
-			.name("hw_wallet".to_string())
+
+		// Ledger event thread
+		// FIXME: check if we can move it to ledger.rs (lifetime issue)
```
rphmeier: elaborate?

hw/src/trezor.rs
```diff
@@ -407,6 +404,41 @@ impl Manager {
 	}
 }
 
+pub struct EventHandler {
```
rphmeier: docs

hw/src/trezor.rs
```diff
@@ -407,6 +404,41 @@ impl Manager {
...
+	fn device_arrived(&mut self, _device: libusb::Device) {
+		println!("trezor arrived");
+		if let Some(t) = self.trezor.upgrade() {
+			// Wait for the device to boot-up
+			thread::sleep(Duration::from_millis(1000));
+			if let Err(e) = t.update_devices() {
+				debug!("TREZOR Connect Error: {:?}", e);
+			}
+			println!("unlocked devices: {:?}", t.list_devices());
+			println!("locked devices: {:?}", t.list_locked_devices());
```
rphmeier: stray printlns in this file

hw/src/trezor.rs
```diff
@@ -407,6 +404,41 @@ impl Manager {
...
+		Self { trezor: trezor }
+	}
+}
+
+impl libusb::Hotplug for EventHandler {
+	fn device_arrived(&mut self, _device: libusb::Device) {
+		println!("trezor arrived");
+		if let Some(t) = self.trezor.upgrade() {
+			// Wait for the device to boot-up
+			thread::sleep(Duration::from_millis(1000));
```
rphmeier: this seems arbitrary and prone to race conditions. is there a better way?

hw/src/ledger.rs
```diff
@@ -335,6 +344,53 @@ impl Manager {
...
+			Err(Error::InvalidDevice)
+		}
+	}
+}
+
+
+impl libusb::Hotplug for EventHandler {
+	fn device_arrived(&mut self, device: libusb::Device) {
+		println!("ledger arrived");
+		Duration::from_millis(1000);
```
rphmeier: this line does nothing, stray println above

hw/src/lib.rs
```diff
@@ -128,84 +131,81 @@ impl From<libusb::Error> for Error {
 
 /// Hardware wallet management interface.
 pub struct HardwareWalletManager {
-	update_thread: Option<thread::JoinHandle<()>>,
+	active_threads: Vec<Option<thread::JoinHandle<()>>>,
```
rphmeier: i'm curious about the motivation to have a separate thread for each HW wallet -- I am mildly worried about the overhead of waking these threads up fairly often to poll for exit and incurring context switches during regular import. 

The other downside to joining the threads manually is that bugs in libusb or drivers can cause Parity's exit to completely hang -- this is something we want to avoid because the `HardwareWalletManager` cleanup might be blocking e.g. rocksdb cleanup and the inevitable forced shutdown could break things.

This is a separate issue, but we have a `StopGuard` utility that currently wraps the `Arc<AtomicBool>` logic, with the idea that there is a single one shared by all the utilities in the program (with only the main thread capable of writing to it). I don't think we do this quite yet, but it would be nice to have the main thread signal stop and wait for a few seconds or for the ref-count to go to zero (whichever comes first).

rphmeier:
Approach generally looks good, major point of question is the shutdown logic.

hw/src/lib.rs
niklasad1: @rphmeier 
Great observations, the main motivation is to check whether `ÙSB Hotplug callbacks` are the reason of the crashes along with removing dummy for loops. By splitting it to two threads makes it possible the control callbacks in fine-grained manner instead of subscribing to all devices and the code becomes is to understand IMO. 

It's possible to do it with one thread and some additional checks though, but let's wait for the test on the Windows machine take it from there (I would be very surprised if it still fails... 🔨 )

Ok, didn't know that I just done similar to the previous implementation

hw/src/trezor.rs
niklasad1: Hrm, only a physical device TREZOR device can generate this and the ```update_devices()``` iterates through the devices and read/write to them after acquiring a RwLock. Thus, the risk is pretty low that somebody plugs in two TREZOR keys at the same time

But still ```Weak::upgrade() -> Rc``` is not atomic, so it that what you refer to? That could probably be solved by another lock!

Thoughts/suggestions?

hw/src/ledger.rs
niklasad1: Oops, should be ```thread::sleep(Duration::from_millis(1000))``` ofc

niklasad1:
> provide a target for log events

The logs looks as follows:
```bash 
2018-02-13 16:00:25  hw_wallet_trezor DEBUG hardware_wallet  Trezor startup
2018-02-13 16:00:29  hw_wallet_trezor DEBUG hardware_wallet::trezor  Trezor V1 left
2018-02-13 16:00:35  hw_wallet_trezor DEBUG hardware_wallet::trezor  Trezor V1 arrived
2018-02-13 16:00:40  hw_wallet_trezor DEBUG hardware_wallet::trezor  Trezor V1 left
```
But I don't really understand your comment, can you clarify what you mean?

niklasad1:
Update: It's now confirmed that this fixes [#7516](https://github.com/paritytech/parity/issues/7516) and is it sufficient change the shutdown logic or move back to one thread to get this merged?

ping @rphmeier @debris

hw/src/lib.rs
```diff
@@ -260,9 +259,11 @@ impl HardwareWalletManager {
 impl Drop for HardwareWalletManager {
 	fn drop(&mut self) {
 		self.exiting.store(true, atomic::Ordering::Release);
-		if let Some(thread) = self.update_thread.take() {
-			thread.thread().unpark();
-			thread.join().ok();
+		for thread in self.active_threads.iter_mut() {
```
rphmeier: better to kill the source of a potential hang for now. what are the repercussions if the process exits _without_ joining the threads?

niklasad1: omg. yes now I get your point the threads will never terminate and consequently this will hang. 

If the ```main thread``` just terminates the two other threads will run and listen for new events and try to connect/disconnect to inserted trezors and ledgers. But I think that is ok since these threads as suppose to live as long as parity does!

Otherwise I think a message passing channel would be appropriate here and signal the threads because rust doesn't provide ```thread cancel``` as pthread does.

Thoughts?

rphmeier: I think signalling that Parity has gone down with an atomic bool or similar is fine. Just that I don't think blocking while waiting for them to join is necessarily a good idea.

niklasad1:
FYI, I think the next step should be to remove HardwareWalletManager struct and replace it with a trait that all Wallets implement instead

rphmeier:
The `thread:sleep` stuff on startup isn't great, would prefer it to be polled, but I guess it works for now.

PR #7867, opened 2018-02-12
Fix CSP for dapps that require eval.

tomusdrw:
Allows dapps to opt-in for `unsafe-eval` by specifying `allowJsEval` in the manifest.

PRs to dapps:
https://github.com/Parity-JS/dapp-console/pull/2

### DIFF
--- a/dapps/src/apps/app.rs
+++ b/dapps/src/apps/app.rs
@@ -14,12 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-use endpoint::EndpointInfo;
-
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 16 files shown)


dapps/src/apps/app.rs
```diff
@@ -14,12 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-use endpoint::EndpointInfo;
-
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct App {
-	pub id: String,
+	pub id: Option<String>,
```
andresilva: Does this have to be unique?

tomusdrw: Yes, it's used as dapp address (:8545/<id>), but for network dapps we override that with a hash anyway. It's only used as-is for local dapps (you can go to :8545/testdapp)

PR #7868, opened 2018-02-12
Flush keyfiles. Resolves #7632

### LINKED ISSUE #7632: Contracts and accounts are disappeared

> I'm running:
>
> - **Which Parity version?**: 1.8.6 beta-2d051e4-20180109
> - **Which operating system?**: Linux Debian 7
> - **How installed?**: via deb package
> - **Which network are you connected to?**: ropsten

Contracts and Accounts are disappeared after restart the node. In web3 there is all accounts (web3.eth.accounts = [{...},...]).  
I've use ssh port forwarding: ssh -L 8545:127.0.0.1:8545 -L 8546:127.0.0.1:8546 -L 8180:127.0.0.1:8180 user@host
I've specified command line arguments: parity daemon ./parity.pid --db-compaction ssd --cache-size 1024 --force-ui --jsonrpc-cors all --identity myid --allow-ips public --log-file parity.log --nat extip:1.2.3.4 --chain ropsten --jsonrpc-apis all --pruning=archive --reserved-peers ropstenpeers.txt

Also, I've checked the directory with keys - all keys still there. It looks like some UI bug.

tomusdrw:


ethstore/src/accounts_dir/disk.rs
```diff
@@ -166,17 +167,17 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
...
 			// save the file
 			let mut file = fs::File::create(&keyfile_path)?;
-			if let Err(err) = self.key_manager.write(original_account, &mut file).map_err(|e| Error::Custom(format!("{:?}", e))) {
-				drop(file);
-				fs::remove_file(keyfile_path).expect("Expected to remove recently created file");
-				return Err(err);
-			}
+
+			// Write key content
+			self.key_manager.write(original_account, &mut file).map_err(|e| Error::Custom(format!("{:?}", e)))?;
```
0x7CFE: Wouldn't the simple `to_string()` do the trick here?

tomusdrw: It's a custom `Error` declared as enum. It only has `Debug` and `Display` implementations.

tomusdrw:
@andresilva Correct. I'm also very unsure if dropping a file actually flushes it's content to disk.

andresilva:
@tomusdrw I think it will only close its underlying file descriptor, which does not guarantee that data is synced. So we should leave the explicit `sync_data` call.

PR #7869, opened 2018-02-12
Clean up docs formatting for Wasm runtime

fckt:


ethcore/wasm/src/runtime.rs
```diff
@@ -137,7 +137,9 @@ impl<'a> Runtime<'a> {
 		Ok(U256::from_big_endian(&buf[..]))
 	}
 
-	/// Charge specified amount of gas, returning false is gas limit exceeded, true if not.
+	/// Charge specified amount of gas
+	///
+	/// Returns false is gas limit exceeded and true if not.
```
0x7CFE: `s/is/if/`

PR #7870, opened 2018-02-12
Auto publish beta-release and stable-release

5chdn:
beta release and stable release should be pushed automatically to vanity after tagging and building

cc @fevo1971 @General-Beck


5chdn:
partially addresses https://github.com/paritytech/scripts/issues/20

arkpar:
The original idea for these tags was just to point to latest beta or stable release. So that CI or website, or any other service could tell that latests beta is e.g. `v1.9.2` They were not supposed to be built at all. There's no need to produce binaries for `beta-release` since it's a moving target. Only binaries for a specific version make sense. If there's a need to have latest binaries elsewhere they can be just copied.

5chdn:
That's right. I will just sync them with aws cli.

PR #7873, opened 2018-02-13
Fix wallet import

andresilva:
This PR updates the wallet import to always generate a new account id and handle duplicate key filenames.


ethstore/src/accounts_dir/disk.rs
```diff
@@ -153,8 +153,42 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
 		)
 	}
 
-	/// insert account with given file name
-	pub fn insert_with_filename(&self, account: SafeAccount, filename: String) -> Result<SafeAccount, Error> {
+	/// insert account with given file name. if the filename is a duplicate of any stored account, a number is appended
+	/// to the filename.
+	pub fn insert_with_filename(&self, account: SafeAccount, mut filename: String) -> Result<SafeAccount, Error> {
```
svyatonik: Could you please check that this comment in `update` (https://github.com/paritytech/parity/blob/006b2f35640d2b729b77b4c5f93b2e9b6a10c5d4/ethstore/src/accounts_dir/disk.rs#L202) is still valid? I mean - if existing account is updated (the file with the same name already exists), file won't be duplicated (with suffix append).

ethstore/src/accounts_dir/disk.rs
```diff
@@ -153,8 +153,42 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
...
+		// check for duplicate filenames and append/increment counter
+		let dups: Vec<_> = self.files()?.into_iter().filter_map(|f| {
```
svyatonik: Also - maybe just check if file exists (`fs::metadata(_)is_ok()`) and then just add a random suffix, instead of reading all the files list (`self.files()`) and trying to maintain an order (do not know if an order is important here)? This should increase performance of insert/update - iirc there are cases when there are a lot (>>10000) key files in single dir.
Do not have a strong opinion on this, though

andresilva:
@svyatonik I simplified the deduplication according to your suggestion and fixed the `update` .

ethstore/src/accounts_dir/disk.rs
```diff
@@ -153,22 +153,31 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
 		)
 	}
 
-	/// insert account with given file name
-	pub fn insert_with_filename(&self, account: SafeAccount, filename: String) -> Result<SafeAccount, Error> {
+
+	/// insert account with given filename. if the filename is a duplicate of any stored account and dedup is set to
+	/// true, a random suffix is appended to the filename.
+	pub fn insert_with_filename(&self, account: SafeAccount, mut filename: String, dedup: bool) -> Result<SafeAccount, Error> {
```
debris: why is `dedup` optional and not a default behaviour?

tomusdrw: Afair this function is also used to update the key file (for instance after a name change or meta change).

ethstore/src/accounts_dir/disk.rs
```diff
@@ -286,6 +291,14 @@ impl KeyFileManager for DiskKeyFileManager {
 	}
 }
 
+fn account_filename(account: &SafeAccount) -> String {
+	// build file path
+	account.filename.as_ref().cloned().unwrap_or_else(|| {
```
tomusdrw: `.as_ref().cloned() -> .clone()`

ethstore/src/accounts_dir/disk.rs
```diff
@@ -153,22 +153,31 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
...
 	}
 
-	/// insert account with given file name
-	pub fn insert_with_filename(&self, account: SafeAccount, filename: String) -> Result<SafeAccount, Error> {
+
+	/// insert account with given filename. if the filename is a duplicate of any stored account and dedup is set to
+	/// true, a random suffix is appended to the filename.
+	pub fn insert_with_filename(&self, account: SafeAccount, mut filename: String, dedup: bool) -> Result<SafeAccount, Error> {
+		// path to keyfile
+		let mut keyfile_path = self.path.clone();
```
tomusdrw: `let keyfile_path = self.path.join(&filename)`

andresilva:
Fixed grumbles.

PR #7883, opened 2018-02-13
Update gitlab-test.sh

General-Beck:
https://github.com/paritytech/parity/issues/7871

### DIFF
--- a/scripts/gitlab-test.sh
+++ b/scripts/gitlab-test.sh
@@ -16,10 +16,10 @@ rust_test () {
   rustup show
   if [[ "${RUST_FILES_MODIFIED}" == "0" ]];
     then echo "Skipping Rust tests since no Rust files modified.";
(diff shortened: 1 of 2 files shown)


scripts/gitlab-test.sh
```diff
@@ -16,10 +16,10 @@ rust_test () {
   rustup show
   if [[ "${RUST_FILES_MODIFIED}" == "0" ]];
     then echo "Skipping Rust tests since no Rust files modified.";
-    else ./test.sh;
+    else ./test.sh || exit $?;
   fi
   if [[ "$CI_COMMIT_REF_NAME" == "nightly" ]];
-    then sh scripts/aura-test.sh;
+    then sh scripts/aura-test.sh || exit $?;
```
tomusdrw: I think `scritps/aura-test.sh` should also have `set -eu` on top.

PR #7884, opened 2018-02-13
Fix snapcraft nightly

5chdn:


5chdn:
This is blocked by https://github.com/paritytech/parity-import-tests/issues/3

5chdn:
Nightly is all green now: https://gitlab.parity.io/parity/parity/pipelines/17901

:tada:

.gitlab-ci.yml
```diff
@@ -142,7 +142,7 @@ linux-snap:
     - rust-stable
   artifacts:
     paths:
-    - scripts/parity_*_amd64.snap
+    - parity.zip
```
tomusdrw: Should we add something that indicates that it's a snap? (to don't confuse it with packed binary or sth)

General-Beck: it only in artifacts on gitlab. for tests, we are not push in release

PR #7887, opened 2018-02-14
SecretStore: store key author address in db

svyatonik:
on top of #7886 
this PR is only for commit https://github.com/paritytech/parity/commit/c7cae5c1104fccd0f46eec361d716e92c290f28d

Since we only know `msg.sender` when server key generation session is started via service contract => store address instead of full public in key entry.


folsen:
Isn't there any case where we might want the public key instead of just an address? Or is any identifier in this case just as good as any other? I'm just thinking we're potentially throwing data away if it's _not_ used via service contract, and so if it would be better to have a more generic structure that allows an address _or_ a public key.

svyatonik:
I can only imagine one potential case - we need to encrypt something with author' public - but the only thing SS hold are keys and author is not guaranteed to have a read access to its keys => author became requester when querying for keys/signatures => public is not required.

In all other cases, I guess, it is not required at all. Moreover, author always was a 'technical field' && it is only used to tie two requests together (second request can only be produced by the 'author' of the first one). Hope some day we'll get rid of this field at all (just after someone/me will find a way to securely generate document key in Secret Store, without requiring third party (author)).

PR #7903, opened 2018-02-15
Beta: Fix CSP for dapps that require eval. (#7867)

5chdn:
* Add allowJsEval to manifest.

* Enable 'unsafe-eval' if requested in manifest.


5chdn:
I don't see why the js-test suddenly fails, it's totally unrelated.

@General-Beck

PR #7844, opened 2018-02-08
Light client: randomize the peer we dispatch requests to

tomaka:
cc #6319 

The issue seems to be caused by non-responsive peers clogging the pipeline.
While this PR doesn't fix the issue entirely, it greatly mitigates it by randomly choosing a peer to query every time we do a request.

Even if we ignore #6319, I think this is a desirable change.

PR #7917, opened 2018-02-16
Fix a division by zero in light client RPC handler

tomaka:
Accidentally introduced in #7844

### DIFF
--- a/ethcore/light/src/on_demand/mod.rs
+++ b/ethcore/light/src/on_demand/mod.rs
@@ -18,6 +18,7 @@
 //! The request service is implemented using Futures. Higher level request handlers
 //! will take the raw data received here and extract meaningful results from it.
 
+use std::cmp;
 use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


ethcore/light/src/on_demand/mod.rs
```diff
@@ -366,7 +367,7 @@ impl OnDemand {
 			.filter_map(|pending| {
 				// the peer we dispatch to is chosen randomly
 				let num_peers = peers.len();
-				let rng = rand::random::<usize>() % num_peers;
+				let rng = rand::random::<usize>() % cmp::max(num_peers, 1);
```
tomusdrw: Maybe it should be a question to the previous PR, but why are we always skipping the same peers from the front of `peers`?

Wouldn't it be better to randomly partition the collection instead?

tomaka: I'm not sure I get it.
The code only chooses one node to dispatch the request to (not a range of nodes). And since we go through the list of nodes twice, each node has an equal chance to be chosen if I'm not mistaken.

tomusdrw: I see, misread the code below. Looks good.

5chdn:
@tomaka could you backport #7844 and #7917 to beta? cheers :beers:

PR #7920, opened 2018-02-16
triehash optimisations

debris:
significantly reduce number of heap allocations during block verification


debris:
it's Friday evening, will do it tomorrow 😄

PR #7921, opened 2018-02-16
gitlab cache

General-Beck:
it is necessary to test


5chdn:
lgtm, let's test it here: https://gitlab.parity.io/parity/parity/pipelines/18035

PR #7926, opened 2018-02-17
bring back trie and triehash benches

debris:


debris:
the difference is not visible in benches, cause all allocations were happening outside of `triehash` module

PR #7928, opened 2018-02-17
removed redundant PodAccount::new method

debris:
it's also more readable

### DIFF
--- a/ethcore/src/pod_state.rs
+++ b/ethcore/src/pod_state.rs
@@ -89,7 +89,12 @@ mod test {
 
 	#[test]
 	fn create_delete() {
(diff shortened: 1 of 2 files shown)


ethcore/src/pod_state.rs
```diff
@@ -89,7 +89,12 @@ mod test {
 
 	#[test]
 	fn create_delete() {
-		let a = PodState::from(map![ 1.into() => PodAccount::new(69.into(), 0.into(), vec![], map![]) ]);
+		let a = PodState::from(map![ 1.into() => PodAccount {
```
tomusdrw: Could be even simpler if there was `Default` for `PodAccount` (possibly with `#[cfg(test)]`)

PR #7931, opened 2018-02-18
 added rlp::encode_short and RlpShortStream

debris:
includes #7926, because I also added one benchmark to show performance impact on triehash's `ordered_trie_root` function

Currently each encoding of rlp allocates 1024 (!!!) bytes on stack. And very often we use it to encode only a single byte! Let's make it configurable.

before:

```
test triehash_ordered_insertions_random_mid ... bench:   1,821,843 ns/iter (+/- 335,640)
```

after:

```
test triehash_ordered_insertions_random_mid ... bench:   1,140,144 ns/iter (+/- 140,133)
```

### DIFF
--- a/ethcore/light/src/client/header_chain.rs
+++ b/ethcore/light/src/client/header_chain.rs
@@ -41,7 +41,7 @@ use ethcore::engines::epoch::{
 	PendingTransition as PendingEpochTransition
 };
 
-use rlp::{Encodable, Decodable, DecoderError, RlpStream, Rlp, UntrustedRlp};
+use rlp::{Encodable, Decodable, DecoderError, RlpStream, Rlp, UntrustedRlp, RlpConfigurableStream, RlpBuffer};
 use heapsize::HeapSizeOf;
 use ethereum_types::{H256, H264, U256};
 use plain_hasher::H256FastMap;
@@ -96,7 +96,7 @@ impl HeapSizeOf for Entry {
 }
 
 impl Encodable for Entry {
-	fn rlp_append(&self, s: &mut RlpStream) {
+	fn rlp_append<E: RlpBuffer>(&self, s: &mut RlpConfigurableStream<E>) {
 		s.begin_list(self.candidates.len());
 
 		for candidate in &self.candidates {
--- a/util/rlp/src/lib.rs
+++ b/util/rlp/src/lib.rs
@@ -56,13 +56,13 @@ mod common;
 mod impls;
 
 use std::borrow::Borrow;
-use elastic_array::ElasticArray1024;
+use elastic_array::{ElasticArray1024, ElasticArray8};
 
 pub use error::DecoderError;
 pub use traits::{Decodable, Encodable, Compressible};
 pub use untrusted_rlp::{UntrustedRlp, UntrustedRlpIterator, PayloadInfo, Prototype};
 pub use rlpin::{Rlp, RlpIterator};
-pub use stream::RlpStream;
+pub use stream::{RlpShortStream, RlpStream, RlpBuffer, RlpConfigurableStream};
 pub use compression::RlpType;
 
 /// The RLP encoded empty data (used to mean "null value").
@@ -104,6 +104,12 @@ pub fn decode_list<T>(bytes: &[u8]) -> Vec<T> where T: Decodable {
 /// ```
 pub fn encode<E>(object: &E) -> ElasticArray1024<u8> where E: Encodable {
 	let mut stream = RlpStream::new();
+	stream.append(object);
+	stream.drain()
+}
+
+pub fn encode_short<E>(object: &E) -> ElasticArray8<u8> where E: Encodable {
+	let mut stream = RlpShortStream::new();
 	stream.append(object);
 	stream.drain()
 }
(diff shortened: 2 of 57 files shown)


util/rlp/src/lib.rs
```diff
@@ -108,6 +108,12 @@ pub fn encode<E>(object: &E) -> ElasticArray1024<u8> where E: Encodable {
 	stream.drain()
 }
 
+pub fn encode_short<E>(object: &E) -> ElasticArray8<u8> where E: Encodable {
```
rphmeier: would really like it to be something like `rlp::encode_into<E: Encodable, U: Growable + DerefMut<[u8]>>` where that trait is implemented for vecs, all elastic arrays, etc.

Extend `Encodable` trait with a `type Encoded: Growable + DerefMut<[u8]>>` and `rlp::encode<E: Encodable>(e: &E) -> E::Encoded` and we can easily encode a type into the exact amount of space it needs without thinking about it.

debris: great idea! but it requires much more modifications than changing signature of a method. I'll create an issue to not forget about it

ethcore/light/src/client/header_chain.rs
```diff
@@ -96,7 +96,7 @@ impl HeapSizeOf for Entry {
 }
 
 impl Encodable for Entry {
-	fn rlp_append(&self, s: &mut RlpStream) {
+	fn rlp_append<E: RlpBuffer>(&self, s: &mut RlpConfigurableStream<E>) {
```
tomusdrw: Why is the generic parameter needed? Can't you do:

```
impl<B: RlpBuffer, S: RlpConfigurableStream<B>> RlpStream for S {

}
```
?

util/rlp/src/lib.rs
tomusdrw: Why not start with `ElasticArray8<u8>` or even `ElasticArray1<u8>` and then switch to `ElasticArray1024` if it's not enough? The return type could probably be an enum with required conversions / Derefs.

util/rlp/src/stream.rs
```diff
@@ -28,32 +29,75 @@ impl ListInfo {
...
 
-impl RlpStream {
+impl_rlp_buffer_for_elastic_array!(ElasticArray8);
+impl_rlp_buffer_for_elastic_array!(ElasticArray1024);
+
+/// Should be used to encode should stream of rlp items.
+pub type RlpShortStream  = RlpConfigurableStream<ElasticArray8<u8>>;
+
+/// Should be used to encode stream of rlp items.
+pub type RlpStream = RlpConfigurableStream<ElasticArray1024<u8>>;
```
tomusdrw: As suggested in other comment, would extract `RlpStream` to a trait and then implement that for every kind of `RlpConfigurableStream` to simplify method signatures.

tomusdrw:
IMHO the signatures of methods appending to a stream got complcated unnecessarily. I think we can leave with one trait for `RlpStream`, cause you don't really care about the internal buffer there.

util/rlp/src/lib.rs
rphmeier: if it's an enum it will take even more stack space than the largest variant. although since it can leave the end uninitialized for smaller variants it might still be faster than using 1024 for everything.

debris: exactly!

debris:
@tomusdrw 

> IMHO the signatures of methods appending to a stream got complcated unnecessarily. I think we can leave with one trait for RlpStream, cause you don't really care about the internal buffer there.

I believe it's also inevitable in @rphmeier solution. The code would look like this: 

```rust
trait Encodable {
  // described a type of buffer which should be used to encode this structure.
  type Encoded: Growable + DerefMut<[u8]>>;

  // E != Encoded. That's because this object might be a property of the other type with different encoding buffer.
  fn rlp_append<E: Growable + DerefMut<[u8]>>(&self, s: &mut RlpConfigurableStream<E>);
}

impl Encodable for Example {
  type Encoded = ElasticArray1024<u8>;

  fn rlp_append<E: Growable + DerefMut<[u8]>>(&self, s: &mut RlpConfigurableStream<E>) { 
    // ... 
  }
}

```

util/rlp/src/lib.rs
rphmeier: I don't think so. I'm pretty sure the `ElasticArray1024` has to zero all of its internal buffer before use.

tomusdrw:
> I believe it's also inevitable in @rphmeier solution.

I'm not talking about rob's solution. You can simplify your current solution, since you never care about `E: RlpBuffer` in the serialization methods at all, it's an implementation detail and does not need to be exposed.

What you care is the consistent API provided with `RlpStream`-like interface and then on the very top level you care about the output type (short(Array8)/long(Array1024)) and those relationships can be encoded.

That said, for rob's solution:
```
impl<E: Growable + DerefMut<[u8]>> RlpStream for RlpConfigurableStream<E> {
  // all methods used in all `rlp_append(&self, s: &mut RlpStream)` or `rlp_append<S: RlpStream>(&self, s: &mut S)` methods
}

debris:
@tomusdrw I don't understand, at all :p

debris:
I'll close it for now. Let's revisit this pr/problem in some time

PR #7882, opened 2018-02-13
Validate gasLimitBoundDivisor != 0

asymmetric:
This PR adds a few things:

* validation against the `gasLimitBoundDivisor` being 0.
* a `spec::Error` type that wraps both parsing errors and semantic ones. 
* a couple of tests, one for the above case, another one for a missing field in the Spec.

Closes #7482.

PR #7933, opened 2018-02-18
spec: Validate required divisor fields are not 0

### LINKED ISSUE #7482: Parity crashes when attempting to start a private network
> I'm running:
>
> - **Which Parity version?**: 1.8.5-beta
> - **Which operating system?**: Windows / MacOS / __Linux__
> - **How installed?**: via installer / homebrew / __binaries__ / from source
> - **Are you fully synchronized?**: no / __yes___
> - **Did you try to restart the node?**: no / __yes__

I'm trying to run a private network using parity. I configured my genesis file to look like this:
```
{
    "name": "Bowery",
    "engine": {
        "authorityRound": {
            "params": {
                "stepDuration": "12",
                "validators": {
                    "list": [
                        "0x50b46ac692821cf0605192e452e28c8389e138ca",
                        "0x289386c7ebb6c5bc062cfc957e124d54a9aa1609"
                    ]
                },
                "blockReward": "8080808080808080808"
            }
        }
    },
    "genesis": {
        "seal": {
            "authorityRound": {
                "step": "0x0",
                "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
            }
        },
        "difficulty": "0x0",
        "gasLimit": "0x2fefd8"
    },
    "params": {
        "networkID": "0x67",
        "minGasLimit": "0x1388",
        "gasLimitBoundDivisor": "0x0"
    },
    "accounts": {
        "0x0000000000000000000000000000000000000001": { "balance": "1", "builtin": { "name":   "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
        "0x0000000000000000000000000000000000000002": { "balance": "1", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
        "0x0000000000000000000000000000000000000003": { "balance": "1", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
        "0x0000000000000000000000000000000000000004": { "balance": "1", "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
        "0x50b46ac692821cf0605192e452e28c8389e138ca": {
            "balance": "1000000000000000000"
        },
        "0x289386c7ebb6c5bc062cfc957e124d54a9aa1609": {
            "balance": "1000000000000000000"
        }
    }
}
```

I generated these validator accounts by fist inputting null addresses and running `parity --chain bowery_gen.json account new`. `bower_gen.json` is the relative path to the json file I pasted above. When starting the parity node with the command `parity --chain bower_gen.json` I receive the following message: 

```

stack backtrace:
   0:     0x560a2676ba1c - <no info>
Thread 'IO Worker #3' panicked at 'assertion failed: your_bits != 0', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bigint-4.2.0/src/uint.rs:1227
[1]    581 abort (core dumped)  parity --chain bowery_gen.json

```

I was following the documentation [here](https://paritytech.github.io/wiki/Chain-specification). The docs were a bit sparse so I may be doing something wrong here. Any guidance is appreciated!

asymmetric:
This is a replacement for https://github.com/paritytech/parity/pull/7882, implementing the check at the the deserialization level with the addition of a `#[serde(deserialize_with)` attribute on the required `Uint` divisor fields.

~`Optional<Uint>` divisor fields are not checked at the moment, since I was having trouble getting that to work.~

/cc @andreasilva @rphmeier

Closes #7482.
Closes #7882.

### DIFF
--- a/json/src/spec/params.rs
+++ b/json/src/spec/params.rs
@@ -16,7 +16,7 @@
 
 //! Spec params deserialization.
 
-use uint::Uint;
+use uint::{self, Uint};
 use hash::{H256, Address};
 use bytes::Bytes;
 
@@ -102,6 +102,7 @@ pub struct Params {
 	pub wasm: Option<bool>,
 	/// See `CommonParams` docs.
 	#[serde(rename="gasLimitBoundDivisor")]
+	#[serde(deserialize_with="uint::validate_non_zero")]
 	pub gas_limit_bound_divisor: Uint,
 	/// See `CommonParams` docs.
 	pub registrar: Option<Address>,
@@ -148,5 +149,22 @@ mod tests {
 		assert_eq!(deserialized.account_start_nonce, Some(Uint(U256::from(0x01))));
 		assert_eq!(deserialized.gas_limit_bound_divisor, Uint(U256::from(0x20)));
 		assert_eq!(deserialized.max_code_size, Some(Uint(U256::from(0x1000))));
+	}
+
+	#[test]
+	#[should_panic(expected = "a non-zero value")]
+	fn test_zero_value_divisor() {
+		let s = r#"{
+			"maximumExtraDataSize": "0x20",
+			"networkID" : "0x1",
+			"chainID" : "0x15",
+			"subprotocolName" : "exp",
+			"minGasLimit": "0x1388",
+			"accountStartNonce": "0x01",
+			"gasLimitBoundDivisor": "0x0",
+			"maxCodeSize": "0x1000"
+		}"#;
+
+		let _deserialized: Params = serde_json::from_str(s).unwrap();
 	}
 }
--- a/json/src/spec/ethash.rs
+++ b/json/src/spec/ethash.rs
@@ -31,11 +31,11 @@ pub struct EthashParams {
 	pub difficulty_bound_divisor: Uint,
 	/// See main EthashParams docs.
 	#[serde(rename="difficultyIncrementDivisor")]
-	#[serde(deserialize_with="uint::validate_optional_non_zero")]
+	#[serde(default, deserialize_with="uint::validate_optional_non_zero")]
 	pub difficulty_increment_divisor: Option<Uint>,
 	/// See main EthashParams docs.
 	#[serde(rename="metropolisDifficultyIncrementDivisor")]
-	#[serde(deserialize_with="uint::validate_optional_non_zero")]
+	#[serde(default, deserialize_with="uint::validate_optional_non_zero")]
 	pub metropolis_difficulty_increment_divisor: Option<Uint>,
 	/// See main EthashParams docs.
 	#[serde(rename="durationLimit")]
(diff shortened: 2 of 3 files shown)


json/src/spec/params.rs
```diff
@@ -149,4 +150,21 @@ mod tests {
...
+			"networkID" : "0x1",
+			"chainID" : "0x15",
+			"subprotocolName" : "exp",
+			"minGasLimit": "0x1388",
+			"accountStartNonce": "0x01",
+			"gasLimitBoundDivisor": "0x0",
+			"maxCodeSize": "0x1000"
+		}"#;
+
+		let _deserialized: Params = serde_json::from_str(s).unwrap();
```
debris: how does printed error look like? Does it say describe which field is expected to have `a non-zero value`?

Maybe instead of using `should_panic` you could do:

```rust
assert_eq!(serde_json::from_str(s).unwrap_err().to_string(), "error message");
```

debris:
@paritytech/ci @General-Beck commit 555e0cbb4ec7b40a57561f435da971c9aadc8971 is not run by CI. Could you check why?

andresilva:
@asymmetric I think the following will work for `Option<Uint>`:

```rust
pub fn validate_opt_non_zero<'de, D>(d: D) -> Result<Option<Uint>, D::Error> where D: Deserializer<'de> {
	let value: Option<Uint> = Option::deserialize(d)?;

	if let Some(value) = value {
		if value == Uint(U256::from(0)) {
			return Err(Error::invalid_value(Unexpected::Unsigned(value.into()), &"a non-zero value"))
		}
	}

	Ok(value)
}
```

You then need to enable default when declaring the `deserialize_with`:

```rust
#[serde(rename="difficultyIncrementDivisor")]
#[serde(default, deserialize_with="uint::validate_opt_non_zero")]
pub difficulty_increment_divisor: Option<Uint>,
```

json/src/spec/params.rs
andresilva: It doesn't specify which field is failing, only it's position. The error message looks like this:
```"invalid value: integer `0`, expected a non-zero value at line 3 column 39"```. I agree the test can be converted not to use `should_panic`. It would be nice to show the field name but this is the same behavior we already have for other fields.

**Comment by asymmetric** [comment]
@andresilva Thanks, that seems to have done the trick!

I was declaring the validation function as

```rust
pub fn validate_optional_non_zero<'de, D>(d: Option<D>) -> Result<Option<Uint>, D::Error> where D: Deserializer<'de>
```

(i.e. the argument was an `Option<D>`), which caused:

```shell
error[E0308]: mismatched types
  --> json/src/spec/ethash.rs:23:35
   |
23 | #[derive(Clone, Debug, PartialEq, Deserialize)]
   |                                   ^^^^^^^^^^^
   |                                   |
   |                                   expected enum `std::option::Option`, found type parameter
   |                                   help: try using a variant of the expected type: `serde::export::Some(__deserializer)`
   |
   = note: expected type `std::option::Option<_>`
              found type `__D`
```

By the way, it compiles even without `#[serde(default)]`. Should I still add it?

json/src/spec/params.rs
asymmetric: Why is it better to avoid `#[should_panic(expect)]`? I thought that using an attribute would make more sense, since it's available.

5chdn:
Tests fail :)

andresilva:
@asymmetric I think when you use `deserialize_with` it no longer sets default to true, which means missing fields are not deserialized as `None` and instead you get a missing field error (this is way the tests are failing).

PR #7934, opened 2018-02-18
fixed broken logs

### LINKED ISSUE #6637: Missing entries in eth_getLogs response
I'm running:
**Parity version**: 1.6.10
**Operating system**: Linux
**And installed**: from source
**Starting params**: --jsonrpc-interface all --jsonrpc-hosts=“all” --pruning archive --max-peers 100 --base-path /var/parity --geth --force-direct -l own_tx=trace --mode active --no-download --tx-queue-size 65536 --tx-queue-gas off --force-sealing --reseal-on-txs all --reseal-min-period 1000 --work-queue-size 1 --remove-solved --no-warp --tracing off --db-compaction ssd

We have 2 parity nodes, built and ran in the same way in archive mode. After initial sync is done, they start to miss events on the **eth_getLogs** requests(all the blocks prior to end of initial sync seem to be consistent). Interesting thing is that they are missing **different** logs. Previously we experienced the same behavior on 1.5.12.
Here is the query to receive events count from `0x60bf91ac87fee5a78c28f7b67701fbcfa79c18ec` contract on block span 4309200-4310200.
```
curl 'http://localhost:8545' -H 'Content-Type: application/json' -H 'Accept: application/json' --data-binary '{"id":1,"jsonrpc":"2.0","params":[{"fromBlock":"0x41c0d0","toBlock":"0x41c4b8","address":"0x60bf91ac87fee5a78c28f7b67701fbcfa79c18ec"}],"method":"eth_getLogs"}' --compressed | grep -o logIndex | wc -l
```

On our nodes we receive 926 and 934 entries.
Etherscan API returns: 937 for the same request: https://api.etherscan.io/api?module=logs&action=getLogs&fromBlock=4309200&toBlock=4310200&address=0x60bf91ac87fee5a78c28f7b67701fbcfa79c18ec
If anyone has nodes in archive mode, I'd like to ask to run the same query and report the results.

Unfortunately this issue makes the filters feature in parity nodes absolutely unreliable for live environments.

debris:
fixes #6637

what caused the bug?

1. The only place where blockchain's `blocks_blooms` is updated is in a function `blooms_at`, line 173. This function updates bloom fetched from db.

https://github.com/paritytech/parity/blob/ff0c44c060e111c6234a47cd1aa8f6411533a202/ethcore/src/blockchain/blockchain.rs#L170-L177

2. That means, that if there are two consecutive blocks inserted into blockchain, but there was no database transaction flush in the meantime, blooms for new blocks will be created incorrectly. This is caused, by the fact, that our blooms are stored in multiple layers and each layer contain information about blooms from 16 consecutive blocks

3. Initially I suspected unordered insert to cause invalid log problems, that's why test function is using `insert_unordered_block`. However, unordered insert is unrelated and log problems occur always when there is no database transaction flush between two consecutive insertions (no matter if ordered or not).

# update

I decided to completely remove bloom groups from blockchain database. Block blooms are almost full and merging them is completely redundant.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -16,11 +16,10 @@
 
 //! Blockchain database.
 
-use std::collections::{HashMap, HashSet, hash_map};
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::mem;
 use itertools::Itertools;
-use bloomchain as bc;
 use heapsize::HeapSizeOf;
 use ethereum_types::{H256, Bloom, U256};
 use parking_lot::{Mutex, RwLock};
@@ -31,7 +30,6 @@ use transaction::*;
 use views::*;
 use log_entry::{LogEntry, LocalizedLogEntry};
 use receipt::Receipt;
-use blooms::{BloomGroup, GroupPosition};
 use blockchain::best_block::{BestBlock, BestAncientBlock};
 use blockchain::block_info::{BlockInfo, BlockLocation, BranchBecomingCanonChainData};
 use blockchain::extras::{BlockReceipts, BlockDetails, TransactionAddress, EPOCH_KEY_PREFIX, EpochTransitions};
@@ -46,9 +44,6 @@ use engines::epoch::{Transition as EpochTransition, PendingTransition as Pending
 use rayon::prelude::*;
 use ansi_term::Colour;
 use kvdb::{DBTransaction, KeyValueDB};
-
-const LOG_BLOOMS_LEVELS: usize = 3;
-const LOG_BLOOMS_ELEMENTS_PER_INDEX: usize = 16;
 
 /// Interface for querying blocks by hash and by number.
 pub trait BlockProvider {
@@ -163,17 +158,7 @@ enum CacheId {
 	BlockDetails(H256),
 	BlockHashes(BlockNumber),
 	TransactionAddresses(H256),
-	BlocksBlooms(GroupPosition),
 	BlockReceipts(H256),
-}
-
-impl bc::group::BloomGroupDatabase for BlockChain {
-	fn blooms_at(&self, position: &bc::group::GroupPosition) -> Option<bc::group::BloomGroup> {
-		let position = GroupPosition::from(position.clone());
-		let result = self.db.read_with_cache(db::COL_EXTRA, &self.blocks_blooms, &position).map(Into::into);
-		self.cache_man.lock().note_used(CacheId::BlocksBlooms(position));
-		result
-	}
 }
 
 /// Structure providing fast access to blockchain data.
@@ -181,8 +166,6 @@ impl bc::group::BloomGroupDatabase for BlockChain {
 /// **Does not do input data verification.**
 pub struct BlockChain {
 	// All locks must be captured in the order declared here.
-	blooms_config: bc::Config,
-
 	best_block: RwLock<BestBlock>,
 	// Stores best block of the first uninterrupted sequence of blocks. `None` if there are no gaps.
 	// Only updated with `insert_unordered_block`.
@@ -199,7 +182,6 @@ pub struct BlockChain {
 	block_details: RwLock<HashMap<H256, BlockDetails>>,
 	block_hashes: RwLock<HashMap<BlockNumber, H256>>,
 	transaction_addresses: RwLock<HashMap<H256, TransactionAddress>>,
-	blocks_blooms: RwLock<HashMap<GroupPosition, BloomGroup>>,
 	block_receipts: RwLock<HashMap<H256, BlockReceipts>>,
 
 	db: Arc<KeyValueDB>,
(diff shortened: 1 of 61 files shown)


ethcore/src/blockchain/blockchain.rs
```diff
@@ -1005,29 +1006,50 @@ impl BlockChain {
...
-		{
-			let mut write_blocks_blooms = self.blocks_blooms.write();
-			batch.extend_with_cache(db::COL_EXTRA, &mut *write_blocks_blooms, update.blocks_blooms, CacheUpdatePolicy::Remove);
-		}
-
 		// These cached values must be updated last with all four locks taken to avoid
 		// cache decoherence
 		{
 			let mut best_block = self.pending_best_block.write();
+			let mut write_blocks_blooms = self.blocks_blooms.write();
```
tomusdrw: Lock order is changed and does not match the order of fields initialization.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1005,29 +1006,50 @@ impl BlockChain {
...
-						hash: update.info.hash,
-						number: update.info.number,
-						total_difficulty: update.info.total_difficulty,
-						timestamp: update.timestamp,
-						block: update.block.to_vec(),
-					});
+				BlockLocation::BranchBecomingCanonChain(_) => {
+					// clear all existing blooms, cause they may be created for block
+					// number higher than current best block
+					*write_blocks_blooms = update.blocks_blooms;
```
tomusdrw: So we override `write_blocks_blooms`, but we don't really clear anything from the `batch`. So we may end up writing some additional data to the database, right? I guess it's not harmful?

tomusdrw: On a second thought:

Why do we actually clear everything from `write_blocks_blooms`? It's actually a cache of whatever is in the database, no?

So do I understand correctly that previously:
1. The cache was reflecting what was in the db (since we always did `extend_with_cache`)
2. The problem was that some keys were overwritten instead of `accrued` (that's what the `CanonChain` branch is doing)
3. Currently we may end up clearing the cache, but at least the database should contain correct values?

Could you also elaborate on what is actually stored in `update.blocks_blooms` (what is used as a `key` and `value`)?

ethcore/src/blockchain/blockchain.rs
debris: like in many other places in this file (although I check and locks never interfere) :) But you are right, I will bring back the old lock order, cause this change is not justified.

ethcore/src/blockchain/blockchain.rs
debris: > So we override `write_blocks_blooms`, but we don't really clear anything from the `batch`. So we may end up writing some additional data to the database, right? I guess it's not harmful?

There is nothing in the batch to be cleared.

> Why do we actually clear everything from write_blocks_blooms? It's actually a cache of whatever is in the database, no?

cause new `BlockLocation` is `BranchBecomingCanonChain`. Database entries need to be updated and so is cache.

> 1. The cache was reflecting what was in the db (since we always did extend_with_cache)

No, it was always empty after executing this function, cause `UpdatePolicy` was `Remove`.

https://github.com/paritytech/parity/blob/5b4abec2dbc3f5653b2caf2cc481411a29c7d00b/ethcore/src/db.rs#L127-L131

2. The problem was that some keys were overwritten instead of `accrued` (that's what the `CanonChain` branch is doing)

No. The problem was that there were never in cache and not yet in database, cause noone has written transaction batch to it.

3. Currently we may end up clearing the cache, but at least the database should contain correct values?

We do not care about database, cause everything is in the cache (as it should be).

andresilva:
This only fixes the issue going forward, so I think people that already have a borked database will still have this issue. We should remember this PR if we see issues related to missing log data in the future.

debris:
> This only fixes the issue going forward, so I think people that already have a borked database will still have this issue. We should remember this PR if we see issues related to missing log data in the future.

Borked database will no longer be a problem

ethcore/src/blockchain/blockchain.rs
```diff
@@ -351,11 +333,13 @@ impl BlockProvider for BlockChain {
...
 	/// Returns numbers of blocks containing given bloom.
 	fn blocks_with_bloom(&self, bloom: &Bloom, from_block: BlockNumber, to_block: BlockNumber) -> Vec<BlockNumber> {
-		let range = from_block as bc::Number..to_block as bc::Number;
-		let chain = bc::group::BloomGroupChain::new(self.blooms_config, self);
-		chain.with_bloom(&range, bloom)
+		// +1, cause it's inclusive range
+		(from_block..to_block + 1)
 			.into_iter()
-			.map(|b| b as BlockNumber)
+			.filter_map(|number| self.block_hash(number).map(|hash| (number, hash)))
```
tomusdrw: Could possibly be done concurrently, no?

ethcore/src/blockchain/blockchain.rs
```diff
@@ -351,11 +333,13 @@ impl BlockProvider for BlockChain {
 
 	/// Returns numbers of blocks containing given bloom.
 	fn blocks_with_bloom(&self, bloom: &Bloom, from_block: BlockNumber, to_block: BlockNumber) -> Vec<BlockNumber> {
```
tomusdrw: Imho would be best to return hashes here, we convert to numbers, then in `client` we pass `numbers` to `blockchain::logs` method that converts to hashes again. Seems pretty wasteful

tomusdrw:
On a second thought. Should we have a (optional?) migration that removes block groups from DB? Or maybe even a separate binary that you can run to clean up the db?

ethcore/src/blockchain/blockchain.rs
debris: yes!

debris:
@tomusdrw 

> On a second thought. Should we have a (optional?) migration that removes block groups from DB? Or maybe even a separate binary that you can run to clean up the db?

Yes, let's clean it up!

@andresilva 

> LGTM. We're also using BloomGroup in tracedb so we should probably make sure that everything's correct there. I agree with @tomusdrw that it would be nice to have a migration clean up the database (not sure how hard that is).

These are different blooms, but they have the same problem. I'll take care of them after fixing these blooms :)

PR #7936, opened 2018-02-18
simplify Client::filter_traces method

debris:


ethcore/src/client/client.rs
```diff
@@ -1692,30 +1692,22 @@ impl BlockChainClient for Client {
...
-					None => traces_iter.collect(),
-				})
-			},
-			_ => None,
-		}
+		let traces = self.tracedb.read()
+			.filter(&db_filter)
+			.into_iter()
+			.skip(filter.after.unwrap_or(0))
+			.take(filter.count.unwrap_or(usize::max_value()))
```
rphmeier: the micro-optimizer in me thinks this will occur an additional cost per `next()` call even when `count` is `None` that the old version wouldn't have.

debris: haha 😂

PR #7942, opened 2018-02-19
removed not working blocks compression

debris:
there're several inconsistencies in blockchain block insertion. blocks are inserted 

a) without compression:

https://github.com/paritytech/parity/blob/b77771171d8f5d185bc35b60b8bc9e47d4dfcae3/ethcore/src/blockchain/blockchain.rs#L532-L533

b) with manual compression:

https://github.com/paritytech/parity/blob/b77771171d8f5d185bc35b60b8bc9e47d4dfcae3/ethcore/src/blockchain/blockchain.rs#L731-L736

c) with compression happening elsewhere:

https://github.com/paritytech/parity/blob/b77771171d8f5d185bc35b60b8bc9e47d4dfcae3/ethcore/src/blockchain/blockchain.rs#L932-L933

Which made me come to conclusion that block insertion is either not working or we have a critical bug. It's probably the first one, cause blocks compression was introduced for full blocks and now we store headers and bodies separately:

https://github.com/paritytech/parity/blob/c220631842bf47cded1550d5fe26e9cba8531c72/ethcore/src/blockchain/blockchain.rs#L513-L517

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -233,17 +233,14 @@ impl BlockProvider for BlockChain {
 
 	/// Get raw block data
 	fn block(&self, hash: &H256) -> Option<encoded::Block> {
-		match (self.block_header_data(hash), self.block_body(hash)) {
-			(Some(header), Some(body)) => {
-				let mut block = RlpStream::new_list(3);
-				let body_rlp = body.rlp();
-				block.append_raw(header.rlp().as_raw(), 1);
-				block.append_raw(body_rlp.at(0).as_raw(), 1);
-				block.append_raw(body_rlp.at(1).as_raw(), 1);
-				Some(encoded::Block::new(block.out()))
-			},
-			_ => None,
-		}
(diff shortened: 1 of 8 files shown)


debris:
> This breaks the database format, doesn't it?

No it doesn't. broke database format long ago. Compress is useless for headers and bodies separately. It works only for full blocks.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -307,8 +303,8 @@ impl BlockProvider for BlockChain {
 			.expect("Low level database error. Some issue with disk?");
 
 		let result = match opt {
-			Some(b) => {
-				let bytes: Bytes = UntrustedRlp::new(&b).decompress(RlpType::Blocks).into_vec();
```
debris: decompress removed here

ethcore/src/blockchain/blockchain.rs
```diff
@@ -271,8 +267,8 @@ impl BlockProvider for BlockChain {
...
```
debris: and here

PR #7950, opened 2018-02-19
kvdb-rocksdb: remove buffered operations when committing transaction

andresilva:
When committing a transaction any keys associated with the transaction are dropped from the overlay.
The test shows the expected semantics for buffered operations.

### DIFF
--- a/util/kvdb-rocksdb/src/lib.rs
+++ b/util/kvdb-rocksdb/src/lib.rs
@@ -491,6 +491,9 @@ impl Database {
 				let batch = WriteBatch::new();
 				let ops = tr.ops;
 				for op in ops {
+					// remove any buffered operation for this key
+					self.overlay.write()[Self::to_overlay_column(op.col())].remove(op.key());
+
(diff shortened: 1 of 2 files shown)


util/kvdb-rocksdb/src/lib.rs
```diff
@@ -491,6 +491,9 @@ impl Database {
 				let batch = WriteBatch::new();
 				let ops = tr.ops;
 				for op in ops {
+					// remove any buffered operation for this key
+					self.overlay.write()[Self::to_overlay_column(op.col())].remove(op.key());
```
andresilva: Not sure if I should take a read lock first to check if there's any entry, instead of always locking for write.

kirushik: No sure either; is there any reliable way to just benchmark this?

debris: I believe, we always lock for write

PR #7957, opened 2018-02-20
simplify compression and move it out of rlp crate

debris:
replaces #7942

changes:
- removed `DBTransaction::put_compressed` and all associated types
- cleaned up rlp compression module
- moved it to a separate crate - `rlp_compression`
- removed `deep_decompress` functionality, cause there was no `deep_compress`
- unified block compression in `blockchain` module

possible improvements:
- move `rlp_compress/common` to `ethcore` so the crate is not spoiled with ethereum specific compressions


Cargo.toml
```diff
@@ -131,4 +131,5 @@ members = [
 	"miner",
 	"transaction-pool",
 	"whisper",
+	"util/rlp_compress"
```
andresilva: Is this necessary?

debris: no, that's leftover from before this crate was used by ethcore ;)

PR #7965, opened 2018-02-21
Increase max download limit to 64MB

andresilva:
On macOS the `parity` binary is larger than 50MB and the auto-update failed because of this.

`2018-02-21 10:16:50 UTC Unable to fetch update (1.9.3-beta-804d…b95b): IO(Error { repr: Custom(Custom { kind: PermissionDenied, error: StringError("Size limit reached.") }) })`

### DIFF
--- a/util/fetch/src/client.rs
+++ b/util/fetch/src/client.rs
@@ -150,8 +150,8 @@ impl Fetch for Client {
 	type Result = CpuFuture<Response, Error>;
 
 	fn new() -> Result<Self, Error> {
-		// Max 50MB will be downloaded.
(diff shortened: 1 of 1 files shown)


andresilva:
@tomusdrw Done. Should we decrease the original limit or leave it at 50MB as it was?

tomusdrw:
@andresilva IMHO we should have 32MB (or 64MB) limit for dapps/content and then perhaps 128MB for the updater (to be future-proof).

5chdn:
Does not compile

andresilva:
Oops, fixed and increased the updater size limit to 128MB.

PR #7968, opened 2018-02-21
CI: Fix cargo cache

General-Beck:
Blocking waiting for file lock on the registry index

### DIFF
--- a/scripts/gitlab-build.sh
+++ b/scripts/gitlab-build.sh
@@ -48,8 +48,12 @@ set_env_win () {
   echo "@ signtool sign /f "\%"1 /p "\%"2 /tr http://timestamp.comodoca.com /du https://parity.io "\%"3" > sign.cmd
 }
 build () {
-  echo "Remove index"
-  rm -rf cargo/registry/index/*.
+  if [[ "windows" = $IDENT ]]
+  then
+    # This is a nasty hack till we figure out the proper cargo caching strategy
+    echo "Remove index"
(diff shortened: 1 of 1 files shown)


scripts/gitlab-build.sh
```diff
@@ -48,6 +48,8 @@ set_env_win () {
   echo "@ signtool sign /f "\%"1 /p "\%"2 /tr http://timestamp.comodoca.com /du https://parity.io "\%"3" > sign.cmd
 }
 build () {
+  echo "Remove index"
+  rm -rf cargo/registry/index/*.
```
debris: I guess you meant, `.cargo/registry/index`? That's a terrible idea! Cargo will have to redownload it's crates index and cargo builds may interfere with each.

5chdn: Cargo cache is in `cargo/` #7921: https://github.com/paritytech/parity/pull/7921/files

The path "is correct", but I can't evaluate if this is the correct way to address this. Problem is that windows builds fail due to a build lock even though there is no other build running.

https://gitlab.parity.io/parity/parity/-/jobs/80511

Please advise :)

kirushik: @General-Beck Let's wrap this in the conditional `if [[ "windows" = $IDENT ]]` for now?
I guess the actual root cause is our cache key selection, but I would do that as a part of enhancing the #7988

debris: @5chdn I believe this is not a correct way of addressing the problem

PR #7969, opened 2018-02-21
Remove Option from database the column numbers

tomaka:
General cleanup refactoring.

This PR turns the lines like:

```rust
const COL_FOO: Option<u32> = Some(...);
```

Into:

```rust
const COL_FOO: u32 = ...;
```

And adjusts the rest of the code for these changes.
This notably removes some `COL_FOO.unwrap_or(0)` or `COL_FOO.expect(...)` which didn't really make sense.

Note that an ever better change would be to force using columns everywhere in the API and totally remove the `Option`s. However the secret_store uses the `None` column, and turning that into an actual column would require a migration which I'm not sure how to write.

### DIFF
--- a/ethcore/light/src/client/header_chain.rs
+++ b/ethcore/light/src/client/header_chain.rs
@@ -187,7 +187,7 @@ pub struct HeaderChain {
 	best_block: RwLock<BlockDescriptor>,
 	live_epoch_proofs: RwLock<H256FastMap<EpochTransition>>,
 	db: Arc<KeyValueDB>,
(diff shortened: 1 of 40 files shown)


tomaka:
Actually I think the API can be improved by splitting the methods that operate on a `Some` column from methods that operate on a `None` column. I'm going to do that and push on this PR.

tomaka:
Giving up with this PR. The changes to make it proper are actually tremendous and not worth spending several days on.

PR #7974, opened 2018-02-22
removed old migrations

debris:
- removed all migrations from 2016
- added migration to remove bloom group from database


util/migration/src/lib.rs
```diff
@@ -138,7 +138,7 @@ pub trait SimpleMigration: 'static {
 	fn version(&self) -> u32;
 	/// Should migrate existing object to new database.
 	/// Returns `None` if the object does not exist in new version of database.
-	fn simple_migrate(&mut self, key: Vec<u8>, value: Vec<u8>) -> Option<(Vec<u8>, Vec<u8>)>;
+	fn simple_migrate(&mut self, key: Vec<u8>, value: Vec<u8>, col: Option<u32>) -> Option<(Vec<u8>, Vec<u8>)>;
```
debris: This is the only new thing in this pr. I added it, cause it's good to know what column we are migrating

5chdn:
Does not compile

util/migration/src/lib.rs
```diff
@@ -130,12 +130,14 @@ pub trait Migration: 'static {
...
 
-/// A simple migration over key-value pairs.
+/// A simple migration over key-value pairs of a single column.
 pub trait SimpleMigration: 'static {
 	/// Number of columns in database after the migration.
 	fn columns(&self) -> Option<u32>;
 	/// Version of database after the migration.
 	fn version(&self) -> u32;
+	/// Index of column which should be migrated.
+	fn migrated_column_index(&self) -> Option<u32>;
```
debris: without this function, the trait is completely useless, cause we don't know which column we are currently migrating

debris:
I think that migrations should be completely reworked. This pr makes a minimal effort take the most out of what we currently have.

debris:
shoud be backported together with #7934

util/migration/src/lib.rs
```diff
@@ -157,8 +160,12 @@ impl<T: SimpleMigration> Migration for T {
...
 
 		for (key, value) in iter {
-			if let Some((key, value)) = self.simple_migrate(key.into_vec(), value.into_vec()) {
-				batch.insert(key, value, dest)?;
+			if migration_needed {
+				if let Some((key, value)) = self.simple_migrate(key.into_vec(), value.into_vec()) {
+					batch.insert(key, value, dest)?;
+				}
+			} else {
+				batch.insert(key.into_vec(), value.into_vec(), dest)?;
```
tomusdrw: By default we are just altering the database, do we really need to overwrite the data?

debris: Unfortunately, we are not altering, but swapping :( 

https://github.com/paritytech/parity/blob/25b19835e316f53366df67b66244a8eee95bc865/util/migration/src/lib.rs#L291-L304

rphmeier: yes, since it can fail we need to keep a backup until it completes

Issue #7821, opened 2018-02-07
remove registrar logic from the Client


debris:
This needs to be done after #7723 is merged. 

Currently registrar contract is imported and used from 2 places:

```
ethcore/src/client/client.rs:83:use_contract!(registry, "Registry", "res/contracts/registrar.json");
hash-fetch/src/urlhint.rs:30:use_contract!(registry, "Registry", "res/registrar.json");
```

Let's remove it from the client.

niklasad1:
Only partly, because it requires some major refactoring in instantiation of `ethcore/client` in order to import the registrar contact from `registrar`. So, for now the contract `registrar.json`is duplicated in `registrar` and in `ethcore`!

tobz1000:
I'm interested in giving this a go. What is the goal of the deduplication architecture-wise?

If the idea is to remove registrar code from the client, it sounds like the plan would be:

- Remove `RegistryInfo` trait
- impl `RegistrarClient` for `ethcore::client::Client` (it looks like this impl is already more-or-less done in `parity::run::execute_impl`)
- Replace uses of `client.registry_address` with `registrar.get_address`, where the registrar's backing client is `ethcore::client::Client`.

However, this would introduce unnecessary boxing, and would produce a future which isn't really async. I can't see `get_address` actually being used asynchronously - but I might be missing something.

If I haven't quite got the right idea, let me know.

dvdplm:
@tobz1000 cool! And thanks! :)

So the goal is essentially to simplify and clarify the code (and avoid potential bugs stemming from duplicates ofc). Yes, you are correct that using `registrar.get_address()` is what we want. As for the async-ness of the code I'd say it's prooooobably an artifact of intention rather than actual usage. It does indeed seem like the code using it today is all synchronous and if it makes things simpler for you it's ok to refactor in that sense (if you prefer sticking with the current API, returning `impl Future` might be a way to avoid the `Box`).

tobz1000:
Yeah, I think it's better to remove the async-ness from the trait - otherwise, as it stands, a user might expect it not to block when it will.

If the client is implemented asynchronously in future, we could introduce a `RegistrarClientAsync` or similar, if desired. `RegistryClient` could then have a blanket-impl over `RegistrarClientAsync` by blocking on its execution.

PR #7978, opened 2018-02-22
[contract-client] refactor

niklasad1:
Hi,
This is a step toward closing [#7821](https://github.com/paritytech/parity/issues/7821) but after realizing that it requires major refactoring in ```ethcore``` we decided to skip that because it also dependent on [#7038](https://github.com/paritytech/parity/pull/7038).

Thus, the PR provides the following:
* A new crate ```contract-client``` that is meant handle interactions to the registrar contract instead of having multiple interfaces and moves this functionality from ```hash-fetch```


contract_client/src/contract_client.rs
```diff
@@ -0,0 +1,64 @@
+use futures::{Future, future};
+use ethabi::{Address, Bytes};
+use std::sync::Arc;
+use keccak_hash::keccak;
+
+use_contract!(registry, "Registry", "res/registrar.json");
+
+// This need to be documented why it is concatenated with the hash
+const MAGIC_CONST: &'static str = "A";
```
rphmeier: I think it's meant to be an "A" record as in DNS

rphmeier:
> 1. synchronous (to be used by LightClient)
> 2. asyncronous (to be used by everyhing else)

This seems backwards. The light client by nature serves contract call requests asynchronously (as it has to go to the network), while full nodes don't even have a futures-based API for doing so.

contract_client/src/contract_client.rs
```diff
@@ -0,0 +1,64 @@
...
+		}
+		.map_err(|e| e.to_string()));
+
+		Box::new(future)
+	}
+}
+
+/// Syncronous contract interface.
+/// Should execute transaction using current blockchain state.
+pub trait ContractClient: Send + Sync {
```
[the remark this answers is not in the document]
rphmeier: and is this trait actually used anywhere?

rphmeier:
`contract-client` as a crate name is way too general. Maybe `registrar-wrapper` or something which describes its intended use.

contract_client/src/contract_client.rs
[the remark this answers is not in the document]
niklasad1: No, it is not used anywhere but ideally this should be used ```ethcore``` which is synchronous if I'm not mistaken!

niklasad1:
> This seems backwards. The light client by nature serves contract call requests asynchronously (as it has to go to the network), while full nodes don't even have a futures-based API for doing so.

Ok, maybe I misunderstood Marek then but at least the ```API``` that ```ethcore/miner``` is using is synchronous by only looking at the [method signatures](https://github.com/paritytech/parity/blob/master/ethcore/src/client/traits.rs#L274-#L283). Maybe that requires another trait for later but that might change depending on the other PR if I understand correct.

niklasad1:
> contract-client as a crate name is way too general and not a good description of its contents. Maybe registrar-wrapper or something which describes its intended use.

I agree and disagree :smile: as long as trait contains the ```ContractClient``` I want something in the name to indicate that. However if it's possible to remove ```ContractClient``` I'd be happy to rename it registrar-wrapper / registrar-client or something similar.

0x7CFE:
@niklasad1, #7038 is almost complete and needs final review before merge. I hope I would not block you for a long time.

rphmeier:
@niklasad1 or even rename the `ContractClient` trait to something else. It's a terrible name for the trait and the crate.

niklasad1:
@rphmeier 
What do you think about re-naming the crate to ```registrar``` and re-naming the ```trait RegistrarClient```?

Because the crate should encapsulate the registrar with a uniform interface for client and it makes sense? Yepp, naming is one of the hardest problems in CS :zap:

rphmeier:
@niklasad1 sounds fine to me

niklasad1:
@tomusdrw 

> Synchronous variant is not really used anywhere Is that expected?

Yes, that is correct. The intention was to use `Synchronous` in `èthcore/miner` because it has its own interface to the `registrar` with different types. But I will look into it now when [#7038](https://github.com/paritytech/parity/pull/7038) is merged.

> Also in some places we could use more concrete types than Box<Future...>, but that's a minor thing, probably an improvement we can do in future.

Ok, do you want me create an issue for that?

tomusdrw:
@niklasad1 don't bother, it's not a hot code anyway, just good that it will be possible in future.

registrar/src/registrar.rs
```diff
@@ -0,0 +1,61 @@
+use futures::{Future, future, IntoFuture};
```
andresilva: Missing license header.

registrar/src/registrar.rs
```diff
@@ -0,0 +1,61 @@
...
+use ethabi::{Address, Bytes};
+use std::sync::Arc;
+use keccak_hash::keccak;
+
+use_contract!(registry, "Registry", "res/registrar.json");
+
+// Maps a domain name to IPv4 address
```
andresilva: `Maps a domain name to an Ethereum address`

PR #7983, opened 2018-02-23
clean up account fmt::Debug

debris:


ethcore/src/pod_state.rs
```diff
@@ -76,7 +76,12 @@ impl fmt::Display for PodState {
 
 /// Calculate and return diff between `pre` state and `post` state.
 pub fn diff_pod(pre: &PodState, post: &PodState) -> StateDiff {
-	StateDiff { raw: pre.get().keys().merge(post.get().keys()).filter_map(|acc| pod_account::diff_pod(pre.get().get(acc), post.get().get(acc)).map(|d|(acc.clone(), d))).collect() }
+	StateDiff {
+		raw: pre.get().keys()
+			.merge(post.get().keys())
+			.filter_map(|acc| pod_account::diff_pod(pre.get().get(acc), post.get().get(acc)).map(|d| (acc.clone(), d)))
+			.collect()
+	}
```
debris: no change here, I just find this one-liner completely unreadable

ethcore/src/state/account.rs
```diff
@@ -479,7 +479,12 @@ impl Account {
 
 impl fmt::Debug for Account {
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-		write!(f, "{:?}", PodAccount::from_account(self))
+		f.debug_struct("Account")
+			.field("balance", &self.balance)
+			.field("nonce", &self.nonce)
+			.field("code", &self.code())
+			.field("storage", &self.storage_changes.iter().collect::<BTreeMap<_, _>>())
+			.finish()
```
debris: removes redundant code and storage allocation on debug

rphmeier: but the allocation is still present one line above

debris: you are right! it's just way smaller, cause the items are not cloned :p updated the description :)

PR #7985, opened 2018-02-23
moved PerfTimer to a separate crate - "trace-time"

debris:
utilities should not be a part of ethcore

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -24,7 +24,6 @@ use itertools::Itertools;
 
 // util
 use hash::keccak;
-use timer::PerfTimer;
 use bytes::Bytes;
 use journaldb;
 use util_error::UtilError;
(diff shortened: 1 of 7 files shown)


ethcore/src/client/client.rs
```diff
@@ -888,7 +887,7 @@ impl Client {
 	/// Import transactions from the IO queue
 	pub fn import_queued_transactions(&self, transactions: &[Bytes], peer_id: usize) -> usize {
 		trace!(target: "external_tx", "Importing queued");
-		let _timer = PerfTimer::new("import_queued_transactions");
+		trace_time!("import_queued_transactions");
```
rphmeier: are we 100% sure that this actually creates a binding at this scope and it isn't dropped immediately?

rphmeier: https://play.rust-lang.org/?gist=050a982e9374596fbf6b45375b8a1ada&version=stable

seems to indicate that it does, and it is hygienic.

PR #7988, opened 2018-02-23
gitlab-ci.yml refactoring v1

General-Beck:
add i386,arm64,armv7 snap packages & snap release
added a protect variable for protected branches and tags
rustfmt&clippy tests (they are necessary to us?)
changes in a caching
retry job for centos #if version `GLIBC_2.18' not found
anchors in gitlab-ci.yml


.gitlab-ci.yml
```diff
@@ -1,197 +1,152 @@
 stages:
   - test
-  - push-release
```
tomusdrw: I think this should stay. It notifies upater about new version (not specific build)

.gitlab-ci.yml
```diff
@@ -1,197 +1,152 @@
...
     - cargo/
   untracked: true
-linux-stable:
+.list_of_refs: &list_of_refs #list of git refs for build and/or push, deploy
   stage: build
-  image: parity/rust:gitlab-ci
   only:
+    - stable
     - beta
+    - gitlab
```
tomusdrw: That's a testing branch, right?

.gitlab-ci.yml
```diff
@@ -1,197 +1,152 @@
...
     - tags
-    - stable
     - triggers
+.artifacts_rules: &artifacts_rules
+  artifacts:
+    when: on_success
+    expire_in: 1 mos
```
tomusdrw: Actually might be better to just keep fixed number of artifacts (i.e. might be good to download old stable binaries even after a month).

.gitlab-ci.yml
```diff
@@ -1,197 +1,152 @@
...
-    - triggers
   script:
     - scripts/gitlab-build.sh armv7-unknown-linux-gnueabihf armv7-unknown-linux-gnueabihf armhf arm-linux-gnueabihf-gcc arm-linux-gnueabihf-g++ ubuntu
   tags:
     - rust-arm
   artifacts:
     paths:
-    - parity.zip
+    - artifacts/
+    expire_in: 1 mos
```
tomusdrw: same here, I think would be better to keep fixed number of artifacts instead.

scripts/gitlab-build.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
-
 set -e # fail on any error
-set -u # treat unset variables as error
```
tomusdrw: Imho was pretty useful. Maybe set `$protected` to some default value in an if statement?

scripts/gitlab-build.sh
```diff
@@ -150,11 +148,13 @@ make_pkg () {
...
 }
 sign_exe () {
+  if [[ -z "$protect" ]]; then echo "__________Skipping sign binaries__________"&&return; fi
   ./sign.cmd $keyfile $certpass "target/$PLATFORM/release/parity.exe"
+  ./sign.cmd $keyfile $certpass windows/ptray/x64/release/ptray.exe
 }
 make_exe () {
   ./msbuild.cmd
-  ./sign.cmd $keyfile $certpass windows/ptray/x64/release/ptray.exe
+  sign_exe
```
tomusdrw: Afair `sign_exe` signs `parity.exe`, this here signs `ptray.exe`

scripts/gitlab-build.sh
```diff
@@ -192,47 +255,54 @@ push_binaries () {
   aws s3api put-object --bucket $S3_BUCKET --key $CI_BUILD_REF_NAME/$BUILD_PLATFORM/"parity_"$VER"_"$IDENT"_"$ARC"."$EXT".md5" --body "parity_"$VER"_"$IDENT"_"$ARC"."$EXT".md5"
   aws s3api put-object --bucket $S3_BUCKET --key $CI_BUILD_REF_NAME/$BUILD_PLATFORM/"parity_"$VER"_"$IDENT"_"$ARC"."$EXT".sha256" --body "parity_"$VER"_"$IDENT"_"$ARC"."$EXT".sha256"
 }
-make_archive () {
-  echo "add artifacts to archive"
-  rm -rf parity.zip
-  zip -r parity.zip target/$PLATFORM/release/parity$S3WIN target/$PLATFORM/release/parity-evm$S3WIN target/$PLATFORM/release/ethstore$S3WIN target/$PLATFORM/release/ethkey$S3WIN parity$S3WIN.md5 parity-evm$S3WIN.md5 ethstore$S3WIN.md5 ethkey$S3WIN.md5 parity$S3WIN.sha256 parity-evm$S3WIN.sha256 ethstore$S3WIN.sha256 ethkey$S3WIN.sha256
+prepare_artifacts () {
+  echo "__________Copy all artifacts to one place__________"
+  rm -rf artifacts
```
tomusdrw: I think this will fail if the folder does not exist. Should be `rm -rf artifacts || true` maybe?

scripts/gitlab-build.sh
```diff
@@ -192,47 +255,54 @@ push_binaries () {
...
     LIBSSL="libssl1.1 (>=1.1.0)"
-    echo "Use libssl1.1 (>=1.1.0) for Debian builds"
-    build
-    strip_binaries
-    calculate_checksums
-    make_deb
-    make_archive
-    push_binaries
+    echo "__________Use libssl1.1 (>=1.1.0) for Debian builds__________"
+    build_and_push
```
tomusdrw: This changes the behaviour since we are pushing to auto-updater as well. Right now auto-updater is smart enough to reject such request (if the PLATFORM is set correctly). Although I think it would be good to avoid pushing to auto-updater if we know it's gonna be rejected (i.e. we only care about darwin, linux and windows packages).

scripts/gitlab-build.sh
```diff
@@ -241,52 +311,33 @@ case $BUILD_PLATFORM in
     strip_binaries
     calculate_checksums
     make_rpm
-    make_archive
+    prepare_artifacts
     push_binaries
+    updater_push_release
```
tomusdrw: Same, we don't care about centos for auto-updating.

scripts/gitlab-build.sh
```diff
@@ -296,40 +347,22 @@ case $BUILD_PLATFORM in
...
-    snapcraft logout
-    $MD5_BIN "parity_"$VER"_amd64.snap" > "parity_"$VER"_amd64.snap.md5"
-    $SHA256_BIN "parity_"$VER"_amd64.snap" > "parity_"$VER"_amd64.snap.sha256"
-    echo "add artifacts to archive"
-    rm -rf parity.zip
-    zip -r parity.zip "parity_"$VER"_amd64.snap" "parity_"$VER"_amd64.snap.md5" "parity_"$VER"_amd64.snap.sha256"
+    make_snap
+    ;;
+  i686-unknown-snap-gnu)
+    set_env
```
tomusdrw: We do we need `set_env` here but not for other snaps? Do we actually care about that at all?

scripts/gitlab-test.sh
```diff
@@ -44,4 +44,12 @@ case $TEST_SWITCH in
   test-coverage)
     coverage_test
     ;;
+  rustfmt)
+    cargo install rustfmt-nightly
```
tomusdrw: we don't really have a good config for rustfmt yet, don't think it's needed.

scripts/gitlab-test.sh
```diff
@@ -44,4 +44,12 @@ case $TEST_SWITCH in
...
+    cargo fmt -- --write-mode=diff
+    ;;
+  clippy)
+    cargo install clippy
+    cargo clippy -- -D warnings
```
tomusdrw: same here, right now it will fail with millions of errors, since we don't use it. I think it would be better to first fix the codebase and then introduce this.

.gitlab-ci.yml
General-Beck: yes

.gitlab-ci.yml
General-Beck: we have them in S3 and vanity

.gitlab-ci.yml
General-Beck: no, we have a difference in quantity
https://gitlab.parity.io/parity/parity/-/jobs/80231/artifacts/browse/artifacts/ arm64 snap
https://gitlab.parity.io/parity/parity/-/jobs/80228/artifacts/browse/artifacts/ arm64 deb

scripts/gitlab-build.sh
General-Beck: temporary solution

scripts/gitlab-build.sh
General-Beck: in this case it is necessary

scripts/gitlab-test.sh
General-Beck: it is necessary to add in TODO

scripts/gitlab-build.sh
rphmeier: to what? can you elaborate?

scripts/gitlab-build.sh
rphmeier: why? can you elaborate?

debris:
@General-Beck please address unanswered questions

General-Beck:
@debris all work now here. It is a lot of changes, but I will finish soon. https://gitlab.parity.io/parity/parity/tree/gitlab

debris:
please reopen when it's updated

PR #7991, opened 2018-02-23
ExecutedBlock cleanup

debris:
changes:
- removed redundant `BlockRef` and `BlockRefMut` structure
- added new structure `Tracing` which represents underlaying data better than `Option<Vec<Vec<FlatTrace>>>`
- removed redundant `bestow_block_reward` function which, I believe, contained bugs. Use machine instead

I next prs ~I will get rid of `BlocksRef` structure and~ I'll try to reduce number of clones we make during block verification


ethcore/src/engines/mod.rs
```diff
@@ -393,35 +393,3 @@ pub trait EthEngine: Engine<::machine::EthereumMachine> {
...
-	use ethereum_types::{Address, U256};
-
-	/// Give reward and trace.
-	pub fn bestow_block_reward(block: &mut ExecutedBlock, reward: U256, receiver: Address) -> Result<(), Error> {
-		let fields = block.fields_mut();
-
-		// Bestow block reward
-		let res = fields.state.add_balance(&receiver, &reward, CleanupMode::NoEmpty)
-			.map_err(::error::Error::from)
-			.and_then(|_| fields.state.commit());
```
debris: issue 1 -> `bestow_block_reward` is called from `on_close_block` and it does `state.commit()`. It shouldn't. What's more this function is called in a loop, which so it's possible that there were multiple commits made in a single `on_close_block`.

ethcore/src/engines/mod.rs
```diff
@@ -393,35 +393,3 @@ pub trait EthEngine: Engine<::machine::EthereumMachine> {
...
-
-		fields.traces.as_mut().map(move |traces| {
-  			let mut tracer = ExecutiveTracer::default();
-  			tracer.trace_reward(receiver, reward, RewardType::Block);
-  			traces.push(tracer.drain())
-		});
-
-		if let Err(ref e) = res {
-			warn!("Encountered error on bestowing reward: {}", e);
```
debris: issue 2 -> error is silently ignored

ethcore/src/block.rs
```diff
@@ -84,6 +84,38 @@ impl Decodable for Block {
 	}
 }
 
+/// Container for block traces.
+#[derive(Clone)]
+pub enum Tracing {
```
tomusdrw: Don't feel like it's a right place to store this structure.

ethcore/src/client/client.rs
```diff
@@ -652,10 +651,7 @@ impl Client {
 
 		// Commit results
 		let receipts = block.receipts().to_owned();
-		let traces = block.traces().clone().unwrap_or_else(Vec::new);
-		let traces: Vec<FlatTransactionTraces> = traces.into_iter()
-			.map(Into::into)
-			.collect();
+		let traces = block.traces().clone().drain();
```
tomusdrw: Seems like unnecessary clone, since `block` is discarded anyway.

debris: totally agree!

But for now, we can't omit `clone()`, cause `block` is generic over `IsBlock` trait, and `IsBlock` has only `fn traces(&self)` method.

debris:
> lgtm, although would first like to merge large pending PRs that probably affect similar parts of the code (instead of causing more conflicts)

If you are talking about #7038, it doesn't modify `block.rs` and conflicts should not exist / or be insignificant

ethcore/src/client/client.rs
debris: that's why I wrote, that in next pr 'I'll try to reduce number of clones we make during block verification' ;)

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -991,12 +993,15 @@ impl Engine<EthereumMachine> for AuthorityRound {
 				};
 
 			for empty_step in empty_steps {
-				::engines::common::bestow_block_reward(block, self.block_reward, empty_step.author()?)?;
+				let author = empty_step.author()?;
+				self.machine.add_balance(block, &author, &self.block_reward)?;
+				self.machine.note_rewards(block, &[(author, self.block_reward)], &[])?;
```
rphmeier: I think this would be more efficient if we had only one call to `note_rewards`.

PR #8018, opened 2018-02-28
[hardware wallet] sleeping -> pollling

niklasad1:
/cc @rphmeier 

This PR replaces the ```thread::sleep``` with polling instead.
Note, that the polling has a timeout configured to 500ms and I will very surprised if more bootup time is needed by the devices. However, if anyone has specs on the devices that would be very useful

I tested it successfully on Linux/Ubuntu with ```TREZOR```


hw/src/ledger.rs
```diff
@@ -354,38 +353,50 @@ impl Manager {
 		}
 	}
 
+	// Try to connect to the device using polling at most 1 second
+	// Failure will be logged via the debug log
+	fn try_connect_polling(ledger: Arc<Manager>, duration: Duration) -> bool {
```
rphmeier: should be a free function, not in the `Manager` impl

PR #8024, opened 2018-03-01
Remove all redux dependencies from the signers

amaurymartiny:
See https://github.com/Parity-JS/shell/pull/9. Updating refs here to latest shell.


jacogr:
Checked specific references against repos, up to date.

PR #8025, opened 2018-03-01
prelude to the block module cleanup

debris:


ethcore/src/block.rs
```diff
@@ -425,7 +398,7 @@ impl<'x> OpenBlock<'x> {
 		if s.block.header.transactions_root().is_zero() || s.block.header.transactions_root() == &KECCAK_NULL_RLP {
 			s.block.header.set_transactions_root(ordered_trie_root(s.block.transactions.iter().map(|e| e.rlp_bytes())));
 		}
-		let uncle_bytes = s.block.uncles.iter().fold(RlpStream::new_list(s.block.uncles.len()), |mut s, u| {s.append_raw(&u.rlp(Seal::With), 1); s} ).out();
+		let uncle_bytes = encode_list(&s.block.uncles).into_vec();
```
rphmeier: `Header::encode` does the seal as well?

rphmeier:
LGTM other than grumble

ethcore/src/block.rs
debris: yes

PR #8030, opened 2018-03-01
Drop built-in dapps for stable (binary size)

jacogr:
Tests locally -

```
Pre : 89474728
Post: 83532768
```

Be aware, numbers above is with `--no-default-features` flags, so decrease is important, however overall base size in this config is much larger than done with defaults. Final sizes only checked once there is a pre-compiled build, i.e. merge


js/src/Dapps/blacklist.js
```diff
@@ -0,0 +1,14 @@
+module.exports = [
+  // tokendeploy
+  '0xf9f2d620c2e08f83e45555247146c62185e4ab7cf82a4b9002a265a0d020348f',
```
debris: where are these hashes coming from?

jacogr: They are the actual ids are registered in the registries. (Initially created using the sha3 of the name, same across all chains)

kirushik:
Are those measurements in bytes?
Any expectation on how that would translate in the actual binary size -- since obviously we have some layers of compression before embedding resources in the final released binary.

jacogr:
In bytes yes, on Mac OS X.

I need to build with `--no-default-features --features ui` to pull in the changes locally since there is no pre-compiled (yet), would love to be more accurate. :(

However, the JS itself probably won't compress, it is run through Uglify already.

PR #8031, opened 2018-03-01
fixed parsing ethash seals and verify_block_undordered

debris:
`rlp::decode` in unsafe, therefore `Header::mix_hash` and `Header::nonce` are unsafe. Added a test for `verify_block_unordered` which panicked when parsing seal with correct len, but invalid seal fields


ethcore/src/ethereum/ethash.rs
```diff
@@ -439,18 +466,6 @@ impl Ethash {
 	}
 }
 
-impl Header {
```
rphmeier: glad we can kill these impls

PR #8042, opened 2018-03-02
Echo back the message hash of a ping in the pong request

### LINKED ISSUE #8038: devp2p: node discovery ping echo hash bug
Attributes to: current ``master``, os-and sync-independent

### Intro

Hi, this is Holger from the ``ethereumjs`` team. We had an [issue](https://github.com/ethereumjs/ethereumjs-devp2p/issues/17) on our devp2p implementation (``ethereumjs-devp2p``) not being able to connect to Parity clients, which in the end turned out to be a bug on the Parity side (I now fixed with a hotfix for the moment).

### Bug Description

When sending a ``DPT`` ``ping`` message to Parity the echoed hash is not composed correctly. In the ``on_ping()`` method in ``parity/util/network/src/discovery.rs`` only the RLP-enocded message content [is taken](https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/util/network/src/discovery.rs#L421) for creating the hash:

```rust
fn on_ping(&mut self, rlp: &UntrustedRlp, node: &NodeId, from: &SocketAddr) -> Result<Option<TableUpdates>, Error> {
   // ...
   let hash = keccak(rlp.as_raw());
   let mut response = RlpStream::new_list(2);
   dest.to_rlp_list(&mut response);
   response.append(&hash);
   self.send_packet(PACKET_PONG, from, &response.drain());
}
```

This differs from the specification (see ["Node Discovery"](https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery)) where the echo hash is created by also hashing ``signature`` and ``packet-type``:

```
hash: sha3(signature || packet-type || packet-data)

Pong packet-type: 0x02
struct Pong
{
	Endpoint to;
	h256 echo;
	uint32_t timestamp;
};
```

### Consequences/Severity

The bug is **potentially severe**, totally preventing connections (at least uni-directional) to other client implementations, depending if the hash is checked/used on the specific client implementation side.

### Possible Solutions

Parity is currently [not validating](https://github.com/paritytech/parity/blob/e95b09348386d01b71901365785c5fa3aa2f7a6d/util/network/src/discovery.rs#L432) the echoed hash in ``on_pong()`` method. So it should be possible to update the echo hash to the correct specification without introducing connectivity problems with older-version Parity clients. This is not tested though.

folsen:
Fixes #8038

Previously, the pong sent back would in the `echo` field contain the hash of just the data of the ping. However, it seems that what is intended here is to echo back the hash that includes signature as well.

Looking at the [go implementation](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/udp.go#L588), it seems that this is what they are doing.


NikVolf:
cc @arkpar

andresilva:
I think the rebase has gone wrong, the commits from master are duplicated with different hashes.

debris:
merging since it does not matter if we squash the commit ;)

PR #8046, opened 2018-03-03
removed redundant Box from patricia-trie Result

debris:
`patricia_trie::Error` has always 32 bytes. There is no point in boxing it, cause `patricia_trie::Result`  value is almost always bigger


util/patricia_trie/src/triedbmut.rs
```diff
@@ -342,7 +342,7 @@ impl<'a> TrieDBMut<'a> {
 
 	// cache a node by hash
 	fn cache(&mut self, hash: H256) -> super::Result<StorageHandle> {
-		let node_rlp = self.db.get(&hash).ok_or_else(|| Box::new(TrieError::IncompleteDatabase(hash)))?;
+		let node_rlp = self.db.get(&hash).ok_or_else(|| TrieError::IncompleteDatabase(hash))?;
```
debris: it's not bigger here, but this function is not performance critical

debris: actually it is

Issue #8033, opened 2018-03-01
rlp module refactor


debris:
should fix issues like #8031 or #8032

- `rlp::decode` should return `Result<_, DecoderError>`
- get rid of `Rlp`
- `UntrustedRlp` should be renamed to `Rlp`

debris:
issues is marked as `easy` and `epic`, cause it's easy, but time consuming

ascjones:
I'm looking at this

ascjones:
This is being split across several PRs, since lots of changes

Phases:

- [ ] Replace 'trusted' usages of Rlp with UntrustedRlp and handle Results, unsafely unwrapping where appropriate
- [ ] Replace usages of Rlp where data received from other nodes, and propogate Results
- [ ] Delete original Rlp and rename UntrustedRlp to Rlp

ascjones:
Just the last bit needs doing now, changing decode to return `Result<_, DecoderError>`. I'll leave that here for now, I've unassigned myself for now so anyone else can pick it up or I'll come back to it later.

dvdplm:
I'll have a stab at the `decode` changes. :)

ascjones:
All done!

PR #8047, opened 2018-03-04
structured rlp encoding in journaldb

debris:
As a part of #8033, I'm trying to get rid of all usages of `Rlp`. This pr get rids of unsafe usages of `Rlp` in `journaldb`. To reduce the number of calls to `expect()` I introduced structured encoding in the module.


andresilva:
Guess I was too late but I had a pending review for this. Just minor stuff though.

ascjones:
Too quick on the draw - I can make those changes if okay with @debris

Issue #7272, opened 2017-12-12
Auto-updater improvements


andresilva:
- [x] A random delay should be added before starting an auto-update. This is necessary to make sure that all existing nodes don't trigger an auto-update and restart at the same time.
- [x] When there is a failure in downloading the binary (hash mismatch, eg) the time between next attempt should increase exponentially.

andresilva:
https://github.com/paritytech/parity/pull/7348#issuecomment-353059939

rphmeier:
wouldn't this random delay be encapsulated by the differences in time to receive the block that triggers the update?

tomusdrw:
It's too little. For Kovan block time is 3seconds, and if every authority runs with auto-updated we are basically killing the entire network.
The second issue is potential database migrations that might happen right after upgrade (they can take up to 30 minutes). See the linked comment for proposed solution.

lght:
> For Kovan block time is 3seconds, and if every authority runs with auto-updated we are basically killing the entire network.

Maybe if we have default threshold based on time *and* block height.

For example, set a time minimum randomly selected from a range (1-12hrs?), and compare to a minimum block height delta (500-1K blocks?), using whichever is longer? Potentially setting different defaults for setups that would require different recovery times (e.g.archive vs light nodes).

Different default thresholds would help with archive nodes taking 48hrs to sync (#7438 proposed solution)[https://github.com/paritytech/parity/pull/7348#issuecomment-353059939]. Their default time delay could be a randomized value in 48-96hr range.

PR #8059, opened 2018-03-05
updater: apply exponential backoff after download failure

andresilva:
Part of #7272.


andresilva:
No, we don't have any tests for the updater (or at least not any meaningful ones) and I don't think it is very testable right now. I want to refactor it as part of #7272.

PR #8060, opened 2018-03-05
Limit incoming connections. 

arkpar:
This limits incoming connections to `max(max_peers - min_peers, min_peers / 2)`
effectively reserving at least `min_peers / 2` connections to be outbound.

Needs careful review and testing on a machine that can accept connections.

### DIFF
--- a/util/network-devp2p/src/host.rs
+++ b/util/network-devp2p/src/host.rs
@@ -507,11 +507,6 @@ impl Host {
 		self.sessions.read().iter().any(|e| e.lock().id() == Some(id))
 	}
 
-	fn handshake_count(&self) -> usize {
-		let (handshakes, ..) = self.session_count();
-		handshakes
-	}
-
 	fn keep_alive(&self, io: &IoContext<NetworkIoMessage>) {
 		let mut to_kill = Vec::new();
 		for e in self.sessions.read().iter() {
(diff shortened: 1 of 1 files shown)


util/network-devp2p/src/host.rs
```diff
@@ -566,7 +574,7 @@ impl Host {
 			self.connect_peer(&id, io);
 			started += 1;
 		}
-		debug!(target: "network", "Connecting peers: {} sessions, {} pending, {} started", self.session_count(), self.handshake_count(), started);
+		debug!(target: "network", "Connecting peers: {} sessions, {} pending, {} started", egress_count + ingress_count, self.handshake_count(), started);
```
kirushik: Why not `handshake_count`? Calling a method initiates another iteration of `session_count()`.

kirushik:
Please don't forget to battle-test this on nightly before backporting to other release channels (which, I think, we should do eventually).

util/network-devp2p/src/host.rs
```diff
@@ -707,12 +710,17 @@ impl Host {
 								(info.config.min_peers as usize, max_peers as usize, info.config.non_reserved_mode == NonReservedPeerMode::Deny, info.id().clone())
 							};
 
+							max_peers = max(max_peers, min_peers);
```
andresilva: Why is this needed? (assuming we validate in the config that `max_peers >= min_peers`)

andresilva:
I'll test this tomorrow once I have a proper internet connection.

util/network-devp2p/src/host.rs
arkpar: Just another sanity check, so that this code does not rely on config.

andresilva:
This is working properly for me.

ddorgan:
@arkpar @andresilva I installed master on a boot node and only had about 11 peers after 30 minutes. When I reverted to 1.8.4 it got 18 peers straight away.

Is there logs or debug info I can provide to you on this?

BTW the node is running with these options:
```
parity --chain kovan --log-file parity.log --auto-update=all --no-dapps --no-ui --max-peers 500 --snapshot-peers 1200 --pruning-history 1200   --jsonrpc-apis web3,eth,parity,parity_set,net,traces,rpc --no-warp --allow-ips public
```

PR #8061, opened 2018-03-05
network: init discovery using healthy nodes

andresilva:


andresilva:
Not entirely sure about this as I'm not very familiar with the discovery code.

util/network/src/node_table.rs
```diff
@@ -241,24 +241,32 @@ impl NodeTable {
...
 
-	/// Unordered list of all entries
-	pub fn unordered_entries(&self) -> Vec<NodeEntry> {
-		self.nodes.values().map(|n| NodeEntry {
+	/// Returns node ids sorted by failure percentage, for nodes with the same failure percentage the absolute number of
+	/// failures is considered.
+	pub fn nodes(&self, filter: IpFilter) -> Vec<NodeId> {
+		self.ordered_entries().iter()
+			.filter(|n| n.endpoint.is_allowed(&filter))
+			.map(|n| n.id).collect()
```
rphmeier: put the collect on the following line

PR #8062, opened 2018-03-05
Added support for evm_increaseTime

ChrisPaprocki:
This is the proposal to add evm_increaseTime known from testrpc to Parity.

Probably most controversial change is in miner where is_time_increasable() is invoked.

Also, this changes are tightly coupled with PR to Parity-JS/api. Pls take a look here:

https://github.com/Parity-JS/api/pull/11


ChrisPaprocki:
[clabot:check]

parity/rpc_apis.rs
```diff
@@ -49,6 +49,8 @@ pub enum Api {
 	Net,
 	/// Eth (Safe)
 	Eth,
+	/// Evm (Safe)
```
tomusdrw: This API set is definitely unsafe, I don't think we should allow dapps to have access to those methods by default.
At best it should be enabled by default only when running with `--chain=dev`, but not for other chains.

ChrisPaprocki:
No, it is not speced, at least not outside the code. 
Could you suggest where you would see it specified?

Yes, it does affect all future blocks.
is_time_increasable() has been added in the way that time can be increased only when (sealing) engine permits it. So this solution is considered to be not strong enough?

ethcore/src/miner/miner.rs
```diff
@@ -266,6 +267,7 @@ pub struct Miner {
 	notifiers: RwLock<Vec<Box<NotifyWork>>>,
 	gas_pricer: Mutex<GasPricer>,
 	service_transaction_action: ServiceTransactionAction,
+	increase_time: RwLock<u64>,
```
tomusdrw: Imho because it's so engine-specific it should be managed by the engine not miner, since engine is populating the header anyway.

ChrisPaprocki:
That's sensible. I'll make it disabled if not on dev chain. 
thanks for the review so far.

debris:
it's been 3 weeks without an update. please reopen when it's ready :)

PR #8057, opened 2018-03-05
Add Kovan WASM activation blocknumber

NikVolf:
closes #7292 

*** REQUIRES 7970 ***

PR #8067, opened 2018-03-07
Max code size on Kovan

tomusdrw:
Should be set to the same block number as #8057


NikVolf:
So what is max code size will be after 6,5kk?
Will it be the same as ethereum main net?

PR #8071, opened 2018-03-07
Hardware Wallet trait

niklasad1:
### Abstract

Currently, we have two very similar implementations of two different vendors and it would be good have a `trait/interface` that hardware wallets should provide and simplify to add new devices. 

This is PR is an attempt to provide such a `interface/trait` and moves the thread management to each Wallet implementation. 

### Todo/Future work
I tried to get rid of the Hardware Wallet because it more or less only a wrapper to provide an interface to `ethcore/account_provider`. The intention was to use dynamic dispatch by introducing a vector trait objects of the new `Wallet` trait but the following issues made me give up for now:

1.  HidApi needs to wrapped in an `Arc` because two objects can endup in a runtime error. Thus, it can't be created in `Ledger/Trezor` then it must be created in `ethcore`  (I prefer not to have dependencies to hidapi there).
So HardwareWallet it is for now :cry: 
2. Trait objects with associated types doesn't work without introducing more traits or changing the trait. If anyone have suggestions how to do it I'd be happy to change this


hw/src/ledger.rs
```diff
@@ -20,21 +20,22 @@
...
 
-use super::{WalletInfo, KeyPath};
+use super::{WalletInfo, KeyPath, Device, Wallet, USB_DEVICE_CLASS_DEVICE};
 
-/// Ledger vendor ID
-pub const LEDGER_VID: u16 = 0x2c97;
-/// Legder product IDs: [Nano S and Blue]
-pub const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001];
+const LEDGER_VID: u16 = 0x2c97;
+const LEDGER_PIDS: [u16; 2] = [0x0000, 0x0001];
```
folsen: Why remove the comments?

hw/src/ledger.rs
```diff
@@ -365,18 +255,167 @@ fn try_connect_polling(ledger: Arc<Manager>, timeout: Duration) -> bool {
 	false
 }
 
+impl <'a>Wallet<'a> for Manager {
+	type Error = Error;
+	type Transaction = &'a [u8];
+
+	/// This can be a problem :)
```
folsen: Why? What's the problem?

hw/src/ledger.rs
```diff
@@ -365,18 +255,167 @@ fn try_connect_polling(ledger: Arc<Manager>, timeout: Duration) -> bool {
...
+					info: WalletInfo {
+						name: name,
+						manufacturer: manufacturer,
+						serial: serial,
+						address: addr,
+					},
+				})
+			}
+			// This variant is not possible, but the trait forces this return type
+			Ok(None) => Err(Error::Placeholder),
```
folsen: If it's truly not possible, then I'd suggest calling it something more self-explanatory like `Error::ImpossibleError` (and then we can have a laugh when we see it in production somewhere in a year ;P)

hw/src/ledger.rs
```diff
@@ -365,18 +255,167 @@ fn try_connect_polling(ledger: Arc<Manager>, timeout: Duration) -> bool {
...
+	fn open_path<R, F>(&self, f: F) -> Result<R, Self::Error>
+		where F: Fn() -> Result<R, &'static str>
+	{
+		f().map_err(Into::into)
+	}
+}
+
 /// Ledger event handler
-/// A seperate thread is hanedling incoming events
+/// A separate thread is hanedling incoming events
```
folsen: You fixed one spelling error but missed another ;P: handling

hw/src/trezor.rs
```diff
@@ -36,10 +39,8 @@ use protobuf::{Message, ProtobufEnum};
 
 use trezor_sys::messages::{EthereumAddress, PinMatrixAck, MessageType, EthereumTxRequest, EthereumSignTx, EthereumGetAddress, EthereumTxAck, ButtonAck};
 
-/// Trezor v1 vendor ID
-pub const TREZOR_VID: u16 = 0x534c;
-/// Trezor product IDs
-pub const TREZOR_PIDS: [u16; 1] = [0x0001];
+const TREZOR_VID: u16 = 0x534c;
+const TREZOR_PIDS: [u16; 1] = [0x0001];
```
folsen: I'd say keep the comments here again.

niklasad1:
Sorry, for updates here but I'm trying to understand why the CI fails ^^

5chdn:
Restarted CI

niklasad1:
I don't really understand why ```libusb::Context::new()``` fails it works fine locally on my machine :(
I will check this out tomorrow!

5chdn:
test `ledger::smoke` failed. need help, @niklasad1?

niklasad1:
@5chdn I don't understand why it fails TBH, I have tested it locally on my Linux machine and on a remote Windows machine and it works but not on the CI :cry:

folsen:
@niklasad1 if it's libusb that fails getting a context, maybe it's because it's running under some different hardware constraints like there not existing any USB interface at all, or it's because CI is trying to run a bunch of tests in parallel, and while libusb itself is supposed to be safe against that, maybe there's some race condition in the rust bindings or something weird like that happening? Or because it's running multiple builds on the same machine or something. I'm not actually sure what kind of machine it is that's running these tests.

niklasad1:
@folsen @5chdn @General-Beck 

Haha omg, it's executed in a `docker container` without access to `USB bus` but somehow the `HidApi` doesn't need access to it and the old tests work because it actually never uses `libusb::Context` only the `HidApi`. But, still it is weird to have an optional test that passes even if no `actual ledger` is connected!

So, is it okay to ignore the `ledger test` as we do for `trezor`? 

Alternatively, run the `docker container` with access to the `USB bus` something like:
```bash
$ sudo docker run -t -i --privileged -v /dev/bus/usb:/dev/bus/usb gitlab-ci bash
```

folsen:
We should put in some more actual like serialization tests etc / unit tests for the data exchange and do some mocking. But for now, lets just ignore the ledger test.

Issue #7027, opened 2017-11-12
regenerate protobuf files automatically


debris:
use snippet like this one https://github.com/paritytech/libp2p-rs/pull/33/files#r150420370

debris:
Closing as per https://github.com/paritytech/parity/pull/8073#issuecomment-374543166

PR #8073, opened 2018-03-08
Protobus generator script

niklasad1:
Attempt to close [#7027](https://github.com/paritytech/parity/issues/7027) it's more or less a copy-paste from [rust-libp2p](https://github.com/libp2p/rust-libp2p)

I couldn't find ```key.proto``` and ```structs.proto``` so I copied them too!

Note, that the rust files are generated in ```scripts/protobus/src``` and should probably be generated at some other place but where I have no idea!

### DIFF
--- a/scripts/protobus/keys.proto
+++ b/scripts/protobus/keys.proto
@@ -1,0 +1,15 @@
+enum KeyType {
+  RSA = 0;
+  Ed25519 = 1;
+  Secp256k1 = 2;
+}
+
+message PublicKey {
+  required KeyType Type = 1;
+  required bytes Data = 2;
+}
+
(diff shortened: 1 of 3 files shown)


andresilva:
I think what @debris meant was that we should use the same approach for our own protobuf usage. Those messages you added are specific to libp2p's secio. As far as I can see we don't use protobuf directly in parity, only in [paritytech/trezor-sys](https://github.com/paritytech/trezor-sys/).

niklasad1:
@andresilva Thanks, then this PR doesn't make sense. I'll close this then

@debris Can we close [#7027](https://github.com/paritytech/parity/issues/7027) if it's directly parity related?

PR #8074, opened 2018-03-08
New Transaction Queue implementation

tomusdrw:
Finally finished a small miner refactor and integration with `transaction-pool` library.
Most of the tests for the old queue has been migrated (just removed the ones that didn't make sense with new approach).

We should be able to handle much larger pools now, although only the default for mem limit has been rised (previously it was the limiting factor, not the 8192 max count).

Differences:
- [x] Previous queue had `O(1)` time to get pending set, but `O(n^2)` insertion time. And obviously insertion/removal happens much more often than retrieving the pending set (only for propagation and pending block building) Currently we have `O(n * log(senders))` pending set time (with cache) and `O(tx_per_sender)` (usually within `log(tx_per_sender)`) insertion time.
- [x] Verification is done in parallel
- [x] `Scoring` and `Readiness` are separated from the pool, so it's easier to customize them or introduce different definitions (for instance for https://github.com/ethereum/EIPs/issues/859 or private transactions, etc)
- [x] Banning removed (I don't think it was used at all, feel free to request it again if it's needed), soft-penalization introduced instead (if transaction exceeds the limit other transactions from that sender get lower priority)
- [x] There is no explicit distinction between current and future transactions in the pool - `Readiness` determines that. Because of this we additionally remove `future` transactions that occupy the pool for long time (see `Ready` implementation for details).

What's missing:
- [x] Clean up todos :)
- [x] Performance tests (testing together with @0x7CFE)
- [x] Long-lived tests on existing networks
- [x] Some profiling to identify hot code (already did some work on that, seems that it makes sense cache state nonces for instance)
- [ ] Alternative prioritization strategies (will be introduced in subsequent PRs), ideas:
      - [ ] `GasPriceChain` - Next transaction affects priority of previous transaction (so you can send a new transaction with high gas price to speed up inclusion of a previous one that is stuck).
      - [ ] `GasOnly/GasFactor` - for backward compatibility, we might skip that cause it seems that no one is really using those.
      - [ ] `NoSpam` - De-prioritize transactions from senders that send a lot of transactions or a lot of replacement transactions.


ethcore/light/src/provider.rs
```diff
@@ -281,7 +281,8 @@ impl<T: ProvingBlockChainClient + ?Sized> Provider for T {
 	}
 
 	fn ready_transactions(&self) -> Vec<PendingTransaction> {
-		BlockChainClient::ready_transactions(self)
+		unimplemented!()
+		// BlockChainClient::ready_transactions(self)
```
rphmeier: needs to be reimplemented

ethcore/src/miner/miner.rs
```diff
@@ -117,267 +110,185 @@ pub struct MinerOptions {
...
-	transaction_listener: RwLock<Vec<Box<Fn(&[H256]) + Send + Sync>>>,
-	sealing_work: Mutex<SealingWork>,
-	next_allowed_reseal: Mutex<Instant>,
-	next_mandatory_reseal: RwLock<Instant>,
-	sealing_block_last_request: Mutex<u64>,
-	// for sealing...
+	sealing: Mutex<SealingWork>,
+	params: RwLock<AuthoringParams>,
+	listeners: RwLock<Vec<Box<NotifyWork>>>,
+	gas_pricer: Mutex<GasPricer>,
```
0x7CFE: Looks like this part may subject to refactoring using the idea mentioned in https://github.com/paritytech/parity/issues/7079

ethcore/src/miner/miner.rs
```diff
@@ -117,267 +110,185 @@ pub struct MinerOptions {
...
-	/// Creates new instance of miner without accounts, but with given spec.
-	pub fn with_spec(spec: &Spec) -> Miner {
-		Miner::new_raw(Default::default(), GasPricer::new_fixed(20_000_000_000u64.into()), spec, None)
-	}
-
-	fn forced_sealing(&self) -> bool {
-		self.options.force_sealing || !self.notifiers.read().is_empty()
+	/// Creates new instance of miner with given spec and accounts.
+	///
+	/// NOTE This should be only used for tests.
```
0x7CFE: Why not mark it as `cfg(test)` then?

ethcore/src/miner/miner.rs
```diff
@@ -488,338 +454,262 @@ impl Miner {
...
 
-	/// Asynchronously updates minimal gas price for transaction queue
-	pub fn recalibrate_minimal_gas_price(&self) {
-		debug!(target: "miner", "minimal_gas_price: recalibrating...");
-		let txq = self.transaction_queue.clone();
-		self.gas_pricer.lock().recalibrate(move |price| {
-			debug!(target: "miner", "minimal_gas_price: Got gas price! {}", price);
-			txq.write().set_minimal_gas_price(price);
-		});
+	fn forced_sealing(&self) -> bool {
```
0x7CFE: It would be best to add comment to this method as well

ethcore/src/miner/mod.rs
```diff
@@ -17,191 +17,170 @@
...
+	/// Is it currently sealing?
+	fn is_currently_sealing(&self) -> bool;
 
-	/// Get the lower bound of the gas limit we wish to target when sealing a new block.
-	fn gas_floor_target(&self) -> U256;
+	/// Get the sealing work package preparing it if doesn't exist yet.
+	///
+	/// Returns `None` if engine seals internally.
+	fn work_package<C>(&self, chain: &C) -> Option<(H256, BlockNumber, u64, U256)>
+		where C: BlockChain + CallContract + BlockProducer + SealedBlockImporter + Nonce + Sync;
```
0x7CFE: I thought, we wanted to get rid of too granular traits like `Nonce`, so it would probably be better to bound on aggregate traits like `AccountData` here.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -555,7 +555,7 @@ fn verify_external(header: &Header, validators: &ValidatorSet, empty_steps_trans
 		};
 
 		let header_seal_hash = header_seal_hash(header, empty_steps_rlp);
-		!verify_address(&correct_proposer, &proposer_signature, &header_seal_hash)?
+		!ethkey::verify_address(&correct_proposer, &proposer_signature, &header_seal_hash)?
```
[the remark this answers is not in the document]
rphmeier: really beyond the scope of this PR. see the documentation on Aura in general and the engine module.

ethcore/src/miner/miner.rs
rphmeier: could be extracted, but the reason is probably that not all relevant tests reside in the same crate and `cfg(test)` isn't replicated when compiling dependencies.

ethcore/light/src/transaction_queue.rs
```diff
@@ -120,6 +120,18 @@ impl AccountTransactions {
 	}
 }
 
+/// Transaction import result.
+pub enum ImportResult {
```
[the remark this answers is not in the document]
rphmeier: agree, but worth noting that this is a type that existed before, only moved.

ethcore/src/engines/validator_set/contract.rs
```diff
@@ -169,8 +169,8 @@ mod tests {
 		let validator_contract = "0000000000000000000000000000000000000005".parse::<Address>().unwrap();
 
 		// Make sure reporting can be done.
```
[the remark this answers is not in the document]
rphmeier: the code below makes sure that reporting can be done. this is not a TODO

ethcore/src/client/client.rs
```diff
@@ -969,19 +977,26 @@ impl Client {
...
 		self.queue_transactions.fetch_sub(transactions.len(), AtomicOrdering::SeqCst);
-		let txs: Vec<UnverifiedTransaction> = transactions.iter().filter_map(|bytes| UntrustedRlp::new(bytes).as_val().ok()).collect();
+
+		let txs: Vec<UnverifiedTransaction> = transactions
+			.iter()
+			.filter_map(|bytes| UntrustedRlp::new(bytes).as_val().ok())
+			.collect();
+
+		// Notify sync that the transactions were received from given peer
 		let hashes: Vec<_> = txs.iter().map(|tx| tx.hash()).collect();
```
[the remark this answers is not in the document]
rphmeier: sure, although I would classify it as beyond the scope of the PR. this PR only changes behavior of the underlying queue and only changes code around it to the extent that is needed. I think it is large enough already and I've filed an issue at #8082

ethcore/src/client/traits.rs
```diff
@@ -367,7 +369,7 @@ pub trait BlockChainClient : Sync + Send + AccountData + BlockChain + CallContra
 	fn pruning_info(&self) -> PruningInfo;
 
 	/// Import a transaction: used for misbehaviour reporting.
-	fn transact_contract(&self, address: Address, data: Bytes) -> Result<TransactionImportResult, EthcoreError>;
+	fn transact_contract(&self, address: Address, data: Bytes) -> Result<(), transaction::Error>;
```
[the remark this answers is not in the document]
rphmeier: agreed. this is a part of the more general engine refactoring

ethcore/src/engines/mod.rs
```diff
@@ -304,7 +304,7 @@ pub trait Engine<M: Machine>: Sync + Send {
 	fn set_signer(&self, _account_provider: Arc<AccountProvider>, _address: Address, _password: String) {}
 
 	/// Sign using the EngineSigner, to be used for consensus tx signing.
-	fn sign(&self, _hash: H256) -> Result<Signature, Error> { unimplemented!() }
+	fn sign(&self, _hash: H256) -> Result<Signature, SignError> { unimplemented!() }
```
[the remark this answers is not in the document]
rphmeier: should be removed later on

miner/src/pool/queue.rs
```diff
@@ -0,0 +1,433 @@
...
+			.map(|result| match result {
+				Ok(verified) => match self.pool.write().import(verified) {
+					Ok(_imported) => Ok(()),
+					Err(err) => Err(convert_error(err)),
+				},
+				Err(err) => Err(err),
+			})
+			.collect::<Vec<_>>();
+
+		if results.iter().any(|r| r.is_ok()) {
```
0x7CFE: Didn't get the idea, why `any` is used here instead off `all`?

Also, if `all` was actually intended, all of this could be simplified by using the following `collect` transformation: `Vec<Result<T, E>>` → `Result<Vec<T>, E>`

0x7CFE: Ah, I probably missed the whole method's point. If I understood correctly, we need to return vector of transaction import results, since transactions are usually independent. And we don't want to stop processing the import if some of them failed.

miner/src/pool/verifier.rs
```diff
@@ -0,0 +1,273 @@
...
+				minimal_gas,
+			);
+
+			bail!(transaction::Error::InsufficientGas {
+				minimal: minimal_gas,
+				got: *tx.gas(),
+			})
+		}
+
+		let is_own = if let Transaction::Local(..) = tx { true } else { false };
```
0x7CFE: I believe, in this case `match` would be more compact and eye pleasing :)

```Rust
let is_own = match tx {
	Transaction::Local(..) => true,
	_ => false,
};
```

ethcore/src/miner/miner.rs
```diff
@@ -291,31 +291,13 @@ impl Miner {
...
-				}
-			}
-		)
+		self.sealing.lock().queue
+			.peek_last_ref()
+			.and_then(|b| if b.block().header().number() > latest_block_number {
+				Some(f(b))
+			} else {
+				None
+			})
```
0x7CFE: Just to note, at some point in future this would be possible to write using `Option::filter` that's now [implemented, but unstable](https://github.com/rust-lang/rust/issues/45860).

ethcore/src/client/client.rs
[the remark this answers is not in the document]
0x7CFE: Unfortunately, I was wrong that the change is trivial. Both `txs` and `hashes` are actually used in the code below. 

However, at least, we may eliminate `hashes.clone()` by changing `transactions_received` callback to accept  `&[H256]`, `Cow` or `Rc`, instead of bare `Vec<H256>`. Not every handler would like to own that vector. But current implementation is forced to clone `hashes` to pass it to handler even it has nothing to do with it, like default one.

ethcore/src/miner/blockchain_client.rs
```diff
@@ -138,45 +141,90 @@ impl<'a, C: 'a> pool::client::Client for BlockChainClient<'a, C> where
...
+
+/// A `Sync` wrapper for `Header`.
+#[derive(Clone)]
+pub struct SyncHeader(Header);
+// Since hashes are pre-computed
+// we can safely implement Sync here
+//
+// TODO [ToDr] Remove this wrapper when `Header`
+// does not use `RefCells` any more.
+unsafe impl Sync for SyncHeader {}
```
0x7CFE: I'm afraid such approach is dangerous and may introduce undefined behaviour. Even that data is read only, data visibility is not enforced by memory barriers.

Suppose the following situation: 
1. Thread 1 created `SyncHeader` by calling `new`
2. Very likely, `SyncHeader`'s data is now resides in the data cache of Core 1 that executed Thread 1
3. Immediately afterwards `SyncHeader` is accessed by Thread 2 that is being executed on Core 2
4. Since cache coherence was not enforced by memory barriers, Core 2 may have outdated view on that part of the memory, hence resulting in an UB.

rphmeier: how can it have an outdated view of the header if it's read-only? is it if the compiler optimizes away the copy when moving in the header to the `SyncHeader` wrapper? wouldn't forcing an `mfence` before returning the `SyncHeader` be enough to combat that?

0x7CFE: > how can it have an outdated view of the header if it's read-only?

Short answer is: it could have outdated view on the memory occupied by header, not the header itself. `Sync` is `unsafe` for a reason, so we definitely need to understand what we're doing before accepting such solution.

Long answer that I wrote for @tomusdrw (please correct me if I'm wrong somewhere) is below:

…So, speaking about `unsafe Sync` I believe the problem may be explained easily if we'd abstract of language entities and would define it in terms of raw memory. I tried to write as detailed as possible, since it's quite complicated topic. Please skip some parts of the explanation if you feel that they're trivial.

Modern CPUs are much faster than RAM, so several layers of cahces are introduced to exploit the data locality principle.

CPU tries really hard not to use RAM until it's really needed. So, data that was read from memory is stored into data cache in order to speedup futher reads from the same memory region.

Likewise, writes to memory regions are delayed as much as possible and usually stored in dirty cache lines as well as write combine buffer, that tries to aggregate several writes into one line and then perform only a single transfer to RAM that combines all writes performed since last sync.

Because all such operations are performed by every core in the system, almost independently of other cores, each one may have slightly different understanding of what memory state should be considered actual and up to date. And that's where the problems arise, typically.

When we speak about read-only object in memory, we usually mean, that our business logic would not attempt to modify that region after it was initialized. But CPU doesn't know that, it treats all heap memory as R/W (unless pages are protected, but that's another story).

Again, CPU knows only its own cache that may or may not be coherent with the memory's contents.

Now back to our situation.

For the sake of simplicity, let's assume that we have system with two cores and two threads, each one tied to its respective core. Suppose also, that we're speaking about a memory region at address `0xDEADBEEF` of length 64 bytes that is mapped to a single cache line.

Prior to our case, cores were very busy writing `FF`s to that memory region, so after some time memory `M`, and caches `C1` and `C2` all agreed that cache line with address `0xDEADBEEF` contains 64 `FF` bytes.

Now, let's say that `SyncHeader::new` is invoked by a thread that is running on Core 1. At the very same time, Core 2 was doing nothing, so its cache `C2` still holds mapping `0xDEADBEEF -> FFFF...FF`.

From CPUs pespective `SyncHeader::new` is just a bunch of writes to some memory location that was allocated on a heap. And, surprise surprise, it all happened that this address was `0xDEADBEEF`. Since CPU1 really does not want to dump its cache contents to the main memory, for some time all that data resides in cache `C1`, so it contents are something like: `0xDEADBEEF -> <SyncHeader struct>..FFFF`.

After first write barrier, CPU1 would be forced to dump this cache line `C1` to the main memory `M` as well as send a special commaid using cache coherence protocol, that would invalidate cache line in `C2`, so that CPU2 would know that this memory region was altered and its cache contents are outdated. Depending on implementation and coherence protocol, CPU may either pull new contents from the main memory or update its contents directly from CPU1's cache line.

But prior to memory barrier, nothing could tell CPU2 that memory was altered. And that's exactly what happens in case of our implementation that does not perform any synchronization primitives.

So, in order to recap, there may be a small time interval where second core has outdated entry in its cache line and it knows nothing that this line was modified by other core.


> wouldn't forcing an mfence before returning the SyncHeader be enough to combat that?

Yes, that would solve the issue. Prior to that aforementioned scenario is possible.

debris:
closes #1966

0x7CFE:
@tomusdrw, miner source code still contains the following banners in several places:

```
// --------------------------------------------------------------------------
// | NOTE Code below requires transaction_queue and sealing locks.          |
// | Make sure to release the locks before calling that method.             |
// --------------------------------------------------------------------------
```

Since locks domains were refactored we need to update or remove this text.

tomusdrw:
@0x7CFE `unsafe Sync` removed, will prepare a separate PR to make `Header` `Sync` and then revisit current approach. Also fixed the comments in miner.

Together with @0x7CFE we've also run some benchmarks and compared current queue implementation (with additional miner improvements) to the previous one.

Results below.

### Benchmark A
1. Corpus: 50k transactions = 10k senders, 5 transactions each
2. Pool import time ~ 3seconds
3. Creating block with all 50k transactions ~ 9seconds
4. Clearing the pool ~ 1second

### Benchmark B

(should be a bit faster = less db writes and reads for nonces)
1. Corpus: 50k transactions = 1 senders
2. Pool import time ~ ?
3. Creating block with all 50k transactions ~ ?
4. Clearing the pool ~ ?
^^ @0x7CFE let's re-run this benchmark again

For previous queue, the result for Benchmark B was around 36 seconds to import and mine all transactions, so it seems we have almost *3x improvement now*. :tada:

debris:
> For previous queue, the result for Benchmark B was around 36 seconds to import and mine all transactions, so it seems we have almost 3x improvement now

What about Benchmark A? What is the difference?

miner/src/pool/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+use transaction;
+use txpool;
+
+pub mod client;
+pub mod listener;
+pub mod local_transactions;
+pub mod queue;
+pub mod ready;
+pub mod scoring;
+pub mod verifier;
```
debris: do we need to have all of these modules public?

miner/src/pool/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[cfg(test)]
+pub mod tests;
```
debris: probably does not need to be public

miner/src/pool/mod.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
debris: 2015-2018 :)

miner/src/pool/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Transaction Pool
+
+use ethereum_types::{H256, H160 as Address};
```
debris: you can just write 

```rust
use ethereum_types::{H256, Address};
```

miner/src/pool/queue.rs
```diff
@@ -0,0 +1,440 @@
...
+	}
+}
+
+#[derive(Debug)]
+struct CachedPending {
+	block_number: u64,
+	current_timestamp: u64,
+	nonce_cap: Option<U256>,
+	has_local_pending: bool,
+	pending: Option<Vec<Arc<pool::VerifiedTransaction>>>,
```
debris: why is it `Option<Vec<_>>` ? It could be only `Vec<_>`

debris: is `none()` being created often?

miner/src/pool/queue.rs
```diff
@@ -0,0 +1,440 @@
...
+	) -> Vec<Result<(), transaction::Error>> {
+		// Run verification
+		let _timer = ::trace_time::PerfTimer::new("queue::verifyAndImport");
+		let options = self.options.read().clone();
+
+		let verifier = verifier::Verifier::new(client, options, self.insertion_id.clone());
+		let results = transactions
+			.into_par_iter()
+			.map(|transaction| verifier.verify_transaction(transaction))
+			.map(|result| match result {
```
debris: this `match` should be a match

miner/src/pool/queue.rs
```diff
@@ -0,0 +1,440 @@
...
+
+	/// Add a callback to be notified about all transactions entering the pool.
+	pub fn add_listener(&self, f: Box<Fn(&[H256]) + Send + Sync>) {
+		let mut pool = self.pool.write();
+		(pool.listener_mut().1).0.add(f);
+	}
+}
+
+
+fn convert_error(err: txpool::Error) -> transaction::Error {
```
debris: why did you implement `fn convert_error` instead of `into`?

miner/src/pool/queue.rs
tomusdrw: Yeah, `Option` does distinguish between: no-cached pending set and cached empty pending set. We could try to construct pending set every time we see an empty vector, since it should be cheap anyway. Is that what you mean?

miner/src/pool/queue.rs
tomusdrw: Fixed.

miner/src/pool/queue.rs
tomusdrw: neither `txpool::Error` nor `transaction::Error` is defined in that crate.

tomusdrw:
Yup, onice until #8150 is merged.

ethcore/src/miner/miner.rs
bjornwgnr: Should these kind of notes be added as "to-do" comments within the code?

5chdn:
Has "some" conflicts :)

tomusdrw:
@5chdn yeah, yeah. I'm on it :)

ethcore/src/miner/miner.rs
0x7CFE: It's not _that_ important IMO, but maybe we should.

transaction-pool/src/tests/mod.rs
```diff
@@ -125,7 +125,7 @@ fn should_reject_if_above_count() {
 	let tx2 = b.tx().nonce(1).new();
 	let hash = *tx2.hash();
 	txq.import(tx1).unwrap();
-	assert_eq!(txq.import(tx2).unwrap_err().kind(), &error::ErrorKind::TooCheapToEnter(hash, "0".into()));
+	assert_eq!(txq.import(tx2).unwrap_err().kind(), &error::ErrorKind::TooCheapToEnter(hash, "0x0".into()));
```
0x7CFE: In other places I saw "0x" variant. How is this different? I think we should use the same style everywhere.

tomusdrw: `0x` is used to encode empty set of bytes, with numbers we never skip leading zeros, but representation of `0` is an exception (`0x0`). The string encoding is the same as in json-rpc, this particular change was required after bumping `ethereum-types`, where we changed `Display` and `Debug` implementations of uints to be more consistent (i.e. prefix with `0x` every time we display hex).

ethcore/src/miner/service_transaction_checker.rs
```diff
@@ -30,19 +29,32 @@ pub struct ServiceTransactionChecker {
 	contract: service_transaction::ServiceTransaction,
 }
 
+// TODO [ToDr] https://github.com/paritytech/ethabi/pull/84
```
andresilva: I think there's already a release with this in.

miner/src/pool/ready.rs
```diff
@@ -0,0 +1,212 @@
...
+//!
+//! Let's define `S = state nonce`. Transactions are processed
+//! in order, so we first include transaction with nonce `S`,
+//! but then we are able to include the one with `S + 1` nonce.
+//! So bear in mind that transactions can be included in chains
+//! and their readiness is dependent on previous transactions from
+//! the same sender.
+//!
+//! There are three possible outcomes:
+//! - The transaction is old (stalled; state nonce < transaction nonce)
```
andresilva: Shouldn't this be `state nonce > transaction nonce`?

miner/src/pool/ready.rs
```diff
@@ -0,0 +1,212 @@
...
+//! - The transaction is ready (current; state nonce == transaction nonce)
+//! - The transaction is not ready yet (future; state nonce > transaction nonce)
```
andresilva: Shouldn't this be `state nonce < transaction nonce`?

tomusdrw:
@andresilva Good catches, addressed (except for `so avoid` - I think that one is correct; it's a hint for the implementer to avoid expensive checks in that function)

PR #8075, opened 2018-03-08
Implement hardcoded sync in the light client

tomaka:
Adds a `hardcodedSync` field to the chain specifications.
When set, contains the RLP of a header, a total difficulty, and a list of CHT root hashes. If there is no light client database, the light client will immediately jump to the given block.

After this change, the light client synchronizes in 10 to 15 seconds even at the first launch on the main network.

For now only the specs of the main network have been updated.


ethcore/light/src/client/header_chain.rs
```diff
@@ -270,15 +271,47 @@ impl HeaderChain {
...
+				}
+
+				let decoded_header = hardcoded_sync.header.decode();
+
+				// check that we have enough hardcoded CHT roots. avoids panicking later.
+				{
+					let cht_num = cht::block_to_cht_number(decoded_header.number() - 1)
+						.expect("specs provided a hardcoded block with height 0");
+					if cht_num >= hardcoded_sync.chts.len() as u64 {
+						panic!("specs didn't provide enough CHT roots for its hardcoded block");
```
rphmeier: rule for panics of any kind is "prove or remove" -- We may rather want to print a warning and fall back to hardcoded-less sync. Or move these checks to the deserialization phase where we can exit with an error more gracefully.

ethcore/light/src/client/header_chain.rs
```diff
@@ -299,11 +332,36 @@ impl HeaderChain {
 	///
 	/// If the block is an epoch transition, provide the transition along with
 	/// the header.
+	#[inline]
```
rphmeier: does that actually help? generally I figure LLVM will figure it out.

ethcore/light/src/client/header_chain.rs
```diff
@@ -479,6 +542,49 @@ impl HeaderChain {
...
+
+		for cht_num in 0.. {
+			let cht = match self.cht_root(cht_num) {
+				Some(cht) => cht,
+				_ if cht_num != 0 => {
+					let h_num = 1 + cht_num as u64 * cht::SIZE;
+					let header = if let Some(header) = self.block_header(BlockId::Number(h_num)) {
+						header
+					} else {
+						panic!("Header of block #{} not found in DB", h_num);
```
rphmeier: same: do not panic.

ethcore/light/src/client/header_chain.rs
```diff
@@ -479,6 +542,49 @@ impl HeaderChain {
...
+					};
+
+					let decoded = header.decode();
+
+					let entry: Entry = ::rlp::decode(&self.db.get(self.col, era_key(h_num).as_bytes()).unwrap().unwrap());
```
rphmeier: and we really never use unwrap.

ethcore/light/src/client/header_chain.rs
```diff
@@ -479,6 +542,49 @@ impl HeaderChain {
...
+					let total_difficulty = entry.candidates.iter()
+						.find(|c| c.hash == decoded.hash())
+						.expect("no candidate matching block found in DB")
```
rphmeier: should rather prove based on the fact that an inconsistent but correctly-loaded database shouldn't be possible.

ethcore/light/src/client/header_chain.rs
```diff
@@ -479,6 +542,49 @@ impl HeaderChain {
...
+				},
+				_ => {
+					return None;
+				},
+			};
+
+			chts.push(cht);
+		}
+
+		unreachable!("we are after an infinite loop")
```
rphmeier: could be a `loop` with a break-value

parity/run.rs
```diff
@@ -258,6 +258,11 @@ fn execute_light_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger
 	let service = light_client::Service::start(config, &spec, fetch, db, cache.clone())
 		.map_err(|e| format!("Error starting light client: {}", e))?;
 	let client = service.client();
+	// Note: uncomment this code if you want to regenerate the JSON of the hardcoded sync feature
+	// in the chain specifications.
+	/*if let Some(hs) = client.read_hardcoded_sync() {
```
rphmeier: we could make this a subcommand of some kind

rphmeier:
needs an opt-out flag also (off by default is fine, just that it should be there).

parity/run.rs
rphmeier: generally not a fan of the commented-out code. the CLI is a PITA to work with though :\

rphmeier:
looks good other than grumbles

tomaka:
Should I make `read_hardcoded_sync()` return `None` in case of error? Or rather `Result<Option<...>, ...>`? I admit that I didn't really know what to do with it, it's an internal-ish method.

rphmeier:
`Result<Option<_>>` would be better IMO

tomaka:
Sorry for the bad PR quality. I think it should be good now!

rphmeier:
needs merge

tomaka:
@rphmeier Conflicts solved!

parity/export_hardcoded_sync.rs
```diff
@@ -0,0 +1,125 @@
...
+		no_hardcoded_sync: true,
+	};
+
+	config.queue.max_mem_use = cmd.cache_config.queue() as usize * 1024 * 1024;
+
+	// start on_demand service.
+	let on_demand = Arc::new(::light::on_demand::OnDemand::new(cache.clone()));
+
+	let sync_handle = Arc::new(RwLock::new(Weak::new()));
+	let fetch = ::light_helpers::EpochFetch {
```
rphmeier: on_demand, sync, network etc. are not necessary to start here -- a `NoOpEpochFetch` (I think it already exists for tests) will be enough.

rphmeier:
LGTM except for grumble

tomaka:
r?

ethcore/res/ethereum/foundation.json
```diff
@@ -173,6 +173,2593 @@
...
+		"totalDifficulty": "3144406426008470895785",
+		"CHTs": [
+			"0x0eb474b7721727204978e92e27d31cddff56471911e424a4c8271c35f9c982cc",
```
5chdn: Is this anyhow possible without adding 2500 lines to the chain spec?

tomaka: Hmmm, what would the alternative be? We're hardcoding headers in the client after all, we have to put them *somewhere*.

rphmeier: the alternative is to allow the client to be deployed w/o history or CHT roots and have the history fetched afterwards

Issue #8003, opened 2018-02-26
Thread '<unnamed>' panicked at 'overflow converting duration to nanoseconds'


jennazenk:
> - **Which Parity version?**: v1.9.0-beta-53ec114-20180125/x86_64-macos/rustc1.23.0
> - **Which operating system?**: MacOS
> - **How installed?**: via installer
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

When running `parity --chain foundation`, I got the following: 

```
stack backtrace:

Thread '<unnamed>' panicked at 'overflow converting duration to nanoseconds', src/libcore/option.rs:874
```

Restarting the node fixed it, just letting you guys know!

debris:
It's probably quite serious. There are multiple places in our codebase when we create `Duration::from_secs` and add it to `Instant` or `SystemTime`. rust std executes the addition by converting `Duration` to nanoseconds which may cause unhandled overflows.

azulmarino:
Hit on Parity v1.9.4-20180228 beta

folsen:
The issue was identified to probably come from expontential backoff time additions in auto-updater. There may however be other cases where it's from adding times with different resolutions, see rust-lang/rust#48980

PR #8078, opened 2018-03-09
Auto-updater improvements

### LINKED ISSUE #7272: Auto-updater improvements

andresilva:
* Refactored auto-updater logic into state machine
* Refactored to make it testable
* Added `--auto-update-delay` to randomly delay updates by `n` blocks. This takes into account the number of the block of the update release (old updates aren't delayed).
* Added `--auto-update-check-frequency` to define the periodicity of auto-update checks in number of blocks.

Fixes #7272.


andresilva:
I still need to write some tests for this. I need help in testing this manually, by deploying the Operations contract into the Dev chain and triggering an update there.

andresilva:
I tested this locally by deploying the `Operations` and `GitHubHint` contracts and creating a new release, it seems to be working properly. I still want to write unit tests for this although I'm not sure how yet. 😛

updater/src/updater.rs
```diff
@@ -245,6 +294,82 @@ impl Updater {
 		})
 	}
 
+	fn release_block_number(&self, from: BlockNumber, release: &ReleaseInfo) -> Option<BlockNumber> {
```
folsen: Pretty big piece of code, can I get a comment about what it does and when to use it?

updater/src/updater.rs
```diff
@@ -255,47 +380,151 @@ impl Updater {
...
+									Ok(path) => {
+										let setup = |path: &Path| -> Result<(), String> {
+											let dest = this.updates_path(&Self::update_file_name(&release.version));
+											if !dest.exists() {
+												info!(target: "updater", "Fetched latest version ({}) OK to {}", release.version, path.display());
+												fs::create_dir_all(dest.parent().expect("at least one thing pushed; qed")).map_err(|e| format!("Unable to create updates path: {:?}", e))?;
+												fs::copy(path, &dest).map_err(|e| format!("Unable to copy update: {:?}", e))?;
+												restrict_permissions_owner(&dest, false, true).map_err(|e| format!("Unable to update permissions: {}", e))?;
+												info!(target: "updater", "Copied updated binary to {}", dest.display());
+											}
```
folsen: > The answer to that is that if you need more than 3 levels of indentation, you're screwed anyway, and should fix your program.

– Linus Torvalds

No but seriously though, is there a way to break this up somehow?

updater/src/updater.rs
```diff
@@ -255,47 +380,151 @@ impl Updater {
...
+											state.status = UpdaterStatus::Ready { release: release.clone() };
+											// will lock self.state
+											drop(state);
+											this.updater_step();
+										}
+									},
+									// There was an error fetching the update, apply a backoff delay before retrying
+									Err(err) => {
+										let delay = 2usize.pow(retries) as u64;
+										let backoff = (retries, Instant::now() + Duration::from_secs(delay));
```
folsen: I would suggest putting a maximum here, otherwise it goes from 6 days to 12 days to a month etc around 20 retries, meaning we have like ~2 weeks to fix any issue before it takes a month before any node updates. Having a maximum wait of a week seems pretty reasonable?

updater/src/updater.rs
```diff
@@ -255,47 +380,151 @@ impl Updater {
...
+					state.status = UpdaterStatus::Fetching { release: release.clone(), binary, retries: backoff.0 + 1 };
+					// will lock self.state
+					drop(state);
+					fetch(binary);
+				},
+				// the update is ready to be installed
+				UpdaterStatus::Ready { ref release } if *release == latest.track => {
+					let auto = match self.update_policy.filter {
+						UpdateFilter::All => true,
+						UpdateFilter::Critical if release.is_critical /* TODO: or is on a bad fork */ => true,
```
folsen: Intentionally or unintentionally left TODO?

updater/src/updater.rs
debris: It's also probably a cause of #8003 crash

updater/src/updater.rs
```diff
@@ -68,34 +78,69 @@ impl Default for UpdatePolicy {
...
+	status: UpdaterStatus,
 }
 
 /// Service for checking for updates and determining whether we can achieve consensus.
-pub struct Updater {
+pub struct Updater<O = OperationsContractClient, F = fetch::Client> {
 	// Useful environmental stuff.
 	update_policy: UpdatePolicy,
-	weak_self: Mutex<Weak<Updater>>,
+	weak_self: Mutex<Weak<Updater<O, F>>>,
```
debris: can we somehow get rid of this? I know that this will require moving some logic to a separate structure, but `weak_self ` even sounds terribly :p

updater/src/updater.rs
debris: please exit fast on if let statements, cause it's difficult to read it :)

updater/src/updater.rs
andresilva: The current updater logic will drop the backoff as soon as we a new release is pushed, so we won't get stuck in any wait case. I will add a test for that 😉.

updater/src/updater.rs
andresilva: Agreed, I'll clean it up.

updater/src/updater.rs
andresilva: It has a comment up there in the trait definition:

> /// Fetches the block number when the given release was added, checking the interval [from; latest_block].

I can make it more descriptive if you feel it's necessary.

updater/src/updater.rs
andresilva: This was copied from the existing code. I'm not sure how to fix it.

roninkaizen:
parity-andre-updater-improvements, having downloaded and builded it yesterday,
and tested it on 3 different system today i feel fine with the build level, while compiling from "base-code" it did not even throw a warning-

andresilva:
I've added a bunch of tests and made some changes to increase readability. Still need to fix the `weak_self`.

updater/src/updater.rs
folsen: Ah sorry, I wasn't looking back at the trait to compare.

andresilva:
I tried to remove the `weak_self` but the only way to remove it is if the updater is driven externally (by something other than the `ChainNotify`), i.e. the fetch callback cannot drive the updater (only signal success somewhere). I think this would complicate things a bit and I'm not sure it's worth it. I'm also not particularly fond of the `weak_self`, although this is a pattern we also use in some other places, so let me know if I should leave it like this or if you have any suggestions for removing it.

updater/src/updater.rs
```diff
@@ -110,60 +158,80 @@ pub struct Updater {
...
-	} else {
-		version::platform()
-	}
+lazy_static! {
+	static ref PLATFORM: String = {
+		if cfg!(target_os = "macos") {
+			"x86_64-apple-darwin".into()
+		} else if cfg!(windows) {
+			"x86_64-pc-windows-msvc".into()
+		} else if cfg!(target_os = "linux") {
```
kirushik: If this is going to be backported, we might encounter problems with different openssl versions on different distros. See three different version of linux packages [here](https://github.com/paritytech/parity/releases/tag/v1.9.4) for example.

How do we handle that in out current updater? We might've been breaking clients' installations on Debian and CentOS in the worst case.

andresilva: I don't think we handle it at all, AFAIK only one Linux release is published in the Operations contract for a given platform. Maybe we can start detecting the distro and treat those as different platforms: "x64-ubuntu-linux-gnu"/"x64-centos-linux-gnu" etc. I think this will require changes in https://github.com/paritytech/push-release and in our CI pipeline.

jamesray1:
I haven't looked into how auto-update works but I think users should be prompted with a list of changes for each update (particularly for forks and especially for contentious changes) so that they can choose whether to look into the changes before updating. I expect most users will just update like they would with checking yes without reading terms and conditions, but having the choice to look further into it is important.

andresilva:
@jamesray1 I appreciate your concerns about informing users on the list of changes for an update, although that is outside the scope of this PR (which is just refactoring the current behavior of the updater). For now, users can disable the installation of updates (`--auto-update none`). They are still notified about new updates and they can then decide for themselves whether they want to update or not. I feel like what you are suggesting would be more suitable to a GUI app which could be built using the existing RPC methods (https://wiki.parity.io/JSONRPC-parity_set-module#parity_upgradeready https://wiki.parity.io/JSONRPC-parity_set-module#parity_executeupgrade).

jamesray1:
@andresilva Yes, a GUI app for this kind of update notification would be most appropriate.

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+						// will lock self.state
+						drop(state);
+						self.updater_step();
+					}
+				},
+				// There was an error fetching the update, apply a backoff delay before retrying
+				Err(err) => {
+					let delay = 2usize.pow(retries) as u64;
+					// cap maximum backoff to 1 month
+					let delay = cmp::min(delay, 30 * 24 * 60 * 60);
```
tomusdrw: I'd say a day/week would be enough. To reach 2^21 (month) we already need to spend 48 days waiting for previous delays.

With week (2^19) it's 12 days waiting before reaching the cap
and with a day (2^16) it's 2 days waiting.

(Given my math is not off :weary: )

And I think it shouldn't be an issue to handle one request a week/day from every single node in case something goes wrong.

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+			}
+		}
+	}
+
+	fn updater_step(&self) {
+		let current_block_number = self.client.upgrade().map_or(0, |c| c.block_number(BlockId::Latest).unwrap_or(0));
+
+		let mut state = self.state.lock();
+
+		if let Some(latest) = state.latest.clone() {
```
tomusdrw: `if let Some(ref latest) = state.latest.clone()` should be enough, you are cloning it again anyway.

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+				UpdaterStatus::FetchBackoff { ref release, backoff, .. } if *release == latest.track && self.time_provider.now() < backoff.1 => {},
+				// we're delaying the update until the given block number
+				UpdaterStatus::Waiting { ref release, block_number, .. } if *release == latest.track && current_block_number < block_number => {},
+				// we're at (or past) the block that triggers the update, let's fetch the binary
+				UpdaterStatus::Waiting { ref release, block_number, binary } if *release == latest.track && current_block_number >= block_number => {
+					info!(target: "updater", "Update for binary {} triggered", binary);
+
+					state.status = UpdaterStatus::Fetching { release: release.clone(), binary, retries: 1 };
+					// will lock self.state
+					drop(state);
```
tomusdrw: Isn't it possible to isolate the locks better? Or maybe pass the lock guard to `fetch` instead?

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
-					_ => false,
+					// There was a fatal error setting up the update, disable the updater
+					if let Err(err) = setup(&path) {
+						state.status = UpdaterStatus::Disabled;
+						warn!("{}", err);
+					} else {
+						state.status = UpdaterStatus::Ready { release: release.clone() };
+						// will lock self.state
+						drop(state);
+						self.updater_step();
```
tomusdrw: Perhaps pass the lock guard to the method instead?

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+					info!(target: "updater", "Update for binary {} triggered", binary);
+
+					state.status = UpdaterStatus::Fetching { release: release.clone(), binary, retries: 1 };
+					// will lock self.state
+					drop(state);
+					fetch(binary);
+				},
+				// we're ready to retry the fetch after we applied a backoff for the previous failure
+				UpdaterStatus::FetchBackoff { ref release, backoff, binary } if *release == latest.track && self.time_provider.now() >= backoff.1 => {
+					state.status = UpdaterStatus::Fetching { release: release.clone(), binary, retries: backoff.0 + 1 };
```
tomusdrw: Should we increase the backoff even if `latest.track` is now different?

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+						if path.exists() {
+							info!(target: "updater", "Already fetched binary.");
+							state.status = UpdaterStatus::Ready { release: latest.track.clone() };
+							// will lock self.state
+							drop(state);
+							self.updater_step();
+
+						} else if self.update_policy.enable_downloading {
+							let update_block_number = {
+								let from = current_block_number.saturating_sub(self.update_policy.max_delay);
```
tomusdrw: Should the delay be capped with upcoming fork block number?

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
-								let f = move |r: Result<PathBuf, fetch::Error>| if let Some(this) = weak_self.upgrade() { this.fetch_done(r) };
-								let a = fetch::Abort::default().with_max_size(self.update_policy.max_size);
-								self.fetcher.fetch(b, a, Box::new(f));
-							}
-						}
-					}
-				}
+		// Only check for updates every n blocks
+		let current_block_number = self.client.upgrade().map_or(0, |c| c.block_number(BlockId::Latest).unwrap_or(0));
+		if current_block_number % cmp::max(self.update_policy.frequency, 1) != 0 {
```
tomusdrw: How often do we `poll`? What's the guarantee that it will ever be called?

updater/src/updater.rs
```diff
@@ -248,189 +301,406 @@ impl Updater {
...
+	fn release_block_number(&self, from: BlockNumber, release: &ReleaseInfo) -> Option<BlockNumber> {
+		let client = self.client.upgrade()?;
+		let address = client.registry_address("operations".into(), BlockId::Latest)?;
+
+		let filter = Filter {
+			from_block: BlockId::Number(from),
+			to_block: BlockId::Latest,
+			address: Some(vec![address]),
+			topics: vec![
+				Some(vec![*RELEASE_ADDED_EVENT_NAME_HASH]),
```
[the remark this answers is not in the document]
andresilva: Yes I think `ethabi` supports this, I'll fix it.

updater/src/updater.rs
andresilva: If `latest.track` changes then we don't enter this branch and instead start downloading the new release (I think there's a test for this).

updater/src/updater.rs
andresilva: This is currently driven by `ChainNotify` and we only `poll` after we're done syncing.

tomusdrw: Ok! I was worried that if it's run by a timer instead it might go into some weird stalled situations where it's aligned with multiplication of block time and the frequency would work incorrectly.

updater/src/updater.rs
folsen: #hedidthemath

updater/src/updater.rs
andresilva: Actually `fetch` doesn't lock on `self.state` so these are unnecessary. I updated `updater_step` and `execute_upgrade` to take the `MutexGuard`, all explicit drops have been removed.

updater/src/updater.rs
andresilva: We can't borrow since we mutate on `state`. But I've updated it to avoid cloning `latest` again when triggering the fetch. I also removed a similar `latest` clone in `poll`.

andresilva:
@tomusdrw I think I've addressed all your comments please re-review.

Issue #7576, opened 2018-01-16
--max-peers not honored if --min-peers is not specified


GoodMirek:
> I'm running:
>
> - **Which Parity version?**:
```
parity --version
Parity
  version Parity/v1.10.0-unstable-ad14e656f-20180115/x86_64-linux-gnu/rustc1.22.1
```
> - **Which operating system?**: Linux
```
uname -a
Linux hostx 4.14.13-300.fc27.x86_64 #1 SMP Thu Jan 11 04:00:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
> - **How installed?**: from source
> - **Are you fully synchronized?**: no
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

Running command:
```
parity --max-peers 10
```
does not limit the maximum number of peers, unless `--min-peers` is specified too.
It seems that the default value of min-peers prevails CLI option of `--max-peers`. When `--max-peers` is specified to be lower than the min-peers default value, then it is not honored.

Expected behavior is that explicitly expressed options always prevails defaults, so `--max-peers` should have been honored.

File `~/.local/share/io.parity.ethereum/config.toml` does not exist on the affected system.

5chdn:
That's intended, you can not have less than minimum peers :)

GoodMirek:
I would think that explicitly specified options should prevail the defaults.

In this case, I would expect that specifying `--max-peers` without providing `--min-peers` should automatically lower min-peers to be the same or lower than the max-peers. In order not to constantly connect/disconnect peers, which behavior is probably costly (is it?), I can imagine that it can work in e.g. following way:
If `--min-peers` is omitted and only `--max-peers` is specified then min-peers could be set to a half of max-peers, but not less than e.g. 10 and never more than max-peers.

niklasad1:
Solved in #8087

PR #8087, opened 2018-03-12
Honor --max-peers if --min-peers is not specified

niklasad1:
Attempt to solve [#7576](https://github.com/paritytech/parity/issues/7576)

This PR changes the following:
- min and max in `cli/mod.rs` to `Option<u16>`
- none provided              => use defaults
- min-peers only              => max-peers = max(x, DEFAULT_MAX_PEERS)
- max-peers only             => min-peers = min(x, DEFAULT_MIN_PEERS)
- min-peers > max-peers => ERROR
- Added tests for this

I'm not sure if this is the right approach but seemed most straight-forward to me

### DIFF
--- a/parity/cli/usage.rs
+++ b/parity/cli/usage.rs
@@ -161,6 +161,7 @@ macro_rules! usage {
 			Clap(ClapError),
 			Decode(toml::de::Error),
 			Config(String, io::Error),
+			PeerConfiguration,
 		}
 
 		impl ArgsError {
@@ -177,6 +178,10 @@ macro_rules! usage {
 						println_stderr!("{}", e);
 						process::exit(2)
 					},
+					ArgsError::PeerConfiguration => {
+						println_stderr!("You have supplied `min_peers` > `max_peers`");
+						process::exit(2)
+					}
 				}
 			}
 		}
@@ -312,9 +317,16 @@ macro_rules! usage {
 			pub fn parse<S: AsRef<str>>(command: &[S]) -> Result<Self, ArgsError> {
 				let raw_args = RawArgs::parse(command)?;
 
+				// Invalid configuration pattern `mix_peers` > `max_peers`
+				if let (Some(max_peers), Some(min_peers)) = (raw_args.arg_max_peers, raw_args.arg_min_peers) {
+					if min_peers > max_peers {
+						return Err(ArgsError::PeerConfiguration)
+					}
+				} 
+
 				// Skip loading config file if no_config flag is specified
-				if raw_args.flag_no_config {
-					return Ok(raw_args.into_args(Config::default()));
+				else if raw_args.flag_no_config {
+					return Ok(raw_args.into_args(Config::default()))
 				}
 
 				let config_file = raw_args.arg_config.clone().unwrap_or_else(|| raw_args.clone().into_args(Config::default()).arg_config);
--- a/parity/cli/usage.rs
+++ b/parity/cli/usage.rs
@@ -317,16 +317,16 @@ macro_rules! usage {
 			pub fn parse<S: AsRef<str>>(command: &[S]) -> Result<Self, ArgsError> {
 				let raw_args = RawArgs::parse(command)?;
 
-				// Invalid configuration pattern `mix_peers` > `max_peers`
 				if let (Some(max_peers), Some(min_peers)) = (raw_args.arg_max_peers, raw_args.arg_min_peers) {
+					// Invalid configuration pattern `mix_peers` > `max_peers`
 					if min_peers > max_peers {
-						return Err(ArgsError::PeerConfiguration)
+						return Err(ArgsError::PeerConfiguration);
 					}
-				} 
+				}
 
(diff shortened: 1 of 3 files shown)


twittner:
Sorry for adding a reviewer, mouse click error.

parity/configuration.rs
```diff
@@ -467,7 +474,13 @@ impl Configuration {
 	}
 
 	fn min_peers(&self) -> u32 {
-		self.args.arg_peers.unwrap_or(self.args.arg_min_peers) as u32
+		match (self.args.arg_max_peers, self.args.arg_min_peers) {
```
tomusdrw: why not:
```
self.args.arg_min_peers
   .or(self.args.arg_max_peers)
   .unwrap_or(DEFAULT_MIN_PEERS) as u32
```
?

parity/configuration.rs
```diff
@@ -455,8 +457,13 @@ impl Configuration {
 	}
 
 	fn max_peers(&self) -> u32 {
-		let peers = self.args.arg_max_peers as u32;
-		max(self.min_peers(), peers)
+		match (self.args.arg_max_peers, self.args.arg_min_peers) {
```
tomusdrw:
```rs
self.args.arg_max_peers
   .or(self.args.arg_min_peers)
   .unwrap_or(DEFAULT_MAX_PEERS) as u32
```

parity/configuration.rs
niklasad1: yepp, that is way cleaner

parity/cli/usage.rs
```diff
@@ -312,9 +317,16 @@ macro_rules! usage {
...
+				if let (Some(max_peers), Some(min_peers)) = (raw_args.arg_max_peers, raw_args.arg_min_peers) {
+					if min_peers > max_peers {
+						return Err(ArgsError::PeerConfiguration)
+					}
+				} 
+
 				// Skip loading config file if no_config flag is specified
-				if raw_args.flag_no_config {
-					return Ok(raw_args.into_args(Config::default()));
+				else if raw_args.flag_no_config {
```
tomusdrw: This `else` is fishy, why do we ignore `no-config` flag if max_peers or min_peers are not provided?

andresilva: Yes, I also think this `else` is a mistake.

parity/configuration.rs
```diff
@@ -455,8 +457,9 @@ impl Configuration {
 	}
 
 	fn max_peers(&self) -> u32 {
-		let peers = self.args.arg_max_peers as u32;
-		max(self.min_peers(), peers)
+		self.args.arg_max_peers
+			.or(self.args.arg_min_peers)
+			.unwrap_or(DEFAULT_MAX_PEERS) as u32
```
andresilva: In case `max_peers` isn't defined can we set it to `max(min_peers, DEFAULT_MAX_PEERS)`? With the current code if you start parity with just `--min-peers 5`, max peers will also be set to 5 (I'd expect it to use the default value).

parity/configuration.rs
```diff
@@ -467,7 +470,9 @@ impl Configuration {
 	}
 
 	fn min_peers(&self) -> u32 {
-		self.args.arg_peers.unwrap_or(self.args.arg_min_peers) as u32
+		self.args.arg_min_peers
```
andresilva: Same as above but with `min(max_peers, DEFAULT_MIN_PEERS)`.

parity/cli/usage.rs
niklasad1: Yepp, you’re right it should be s/else if/if.

I made a faulty assumption that first if statement catches only errors!

parity/configuration.rs
niklasad1: Yepp, I'll buy that when running ```parity --min-peers 5``` to set set max_value to default

parity/configuration.rs
```diff
@@ -468,8 +471,9 @@ impl Configuration {
 	}
 
 	fn max_peers(&self) -> u32 {
-		let peers = self.args.arg_max_peers as u32;
-		max(self.min_peers(), peers)
+		self.args.arg_max_peers
+			.or(max(self.args.arg_min_peers, Some(DEFAULT_MAX_PEERS)))
```
niklasad1: This will always return `T` as long at least one of the arguments is `Some` but I kept `unwrap_or` if somebody reacts if we use `raw unwrapping`

parity/configuration.rs
```diff
@@ -480,7 +484,9 @@ impl Configuration {
 	}
 
 	fn min_peers(&self) -> u32 {
-		self.args.arg_peers.unwrap_or(self.args.arg_min_peers) as u32
+		self.args.arg_min_peers
+			.or(min(self.args.arg_max_peers, Some(DEFAULT_MIN_PEERS)))
+			.unwrap_or(DEFAULT_MIN_PEERS) as u32
 	}
```
niklasad1: same here!

PR #8093, opened 2018-03-12
[stable] Optimize JS build

tomusdrw:
Introduces common chunk between `bundle.js` and `embed.js` (should save ~3MB)

### DIFF
--- a/js/webpack/app.js
+++ b/js/webpack/app.js
@@ -167,7 +167,7 @@ module.exports = {
 
   plugins: (function () {
     let plugins = []
-      .concat(Shared.getPlugins(isProd, true))
+      .concat(Shared.getPlugins())
       .concat(
         new WebpackErrorNotificationPlugin(),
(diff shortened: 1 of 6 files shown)


js/webpack/app.js
```diff
@@ -164,101 +165,96 @@ module.exports = {
   },
 
   plugins: (function () {
-    let plugins = Shared.getPlugins().concat(
+    let plugins = Shared.getPlugins(isProd, true).concat(
```
amaurymartiny: `Shared.getPlugins` only takes 1 argument

js/webpack/shared.js
```diff
@@ -38,7 +37,6 @@ function getPlugins (_isProd = isProd) {
   const plugins = [
```
amaurymartiny: getPlugins only takes 1 argument here, but has 2 when it's called.

PR #8099, opened 2018-03-13
Replace reqwest with hyper

### LINKED ISSUE #7829: Replace `reqwest` with `hyper+rustls`

twittner:
The approach taken is to spawn a background thread running a tokio `Core` and hyper `Client`, processing requests sent over a `mpsc` channel and send back responses over `oneshot` channels. Hyper client itself does not implement `Send` so this way we can keep the `Send` bound of the `Fetch` trait itself and make this a local change.

To accommodate existing synchronous uses of fetch responses an adapter `BodyReader` is provided which waits for response body `Chunk`s to arrive. 

Resolves #7829.


twittner:
[clabot:check]

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+				let k = cmp::min(self.chunk.len() - self.offset, buf.len() - m);
+				let c = &self.chunk[self.offset .. self.offset + k];
+				(&mut buf[m .. m + k]).copy_from_slice(c);
+				self.offset += k;
+				m += k;
+				if m == buf.len() {
+					break
+				}
+			} else {
+				let body = self.body.take().expect("within this loop `self.body` is always defined");
```
folsen: Tiny nitpick but we have a standardized format for these texts: https://wiki.parity.io/Coding-guide

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
-		}
+#[cfg(test)]
+mod test {
+	use super::*;
+
+	#[test]
+	fn it_should_fetch() {
+		let client = Client::new().unwrap();
+		let abort = Abort::default().with_max_duration(Duration::from_secs(10));
+		let future = client.fetch("https://httpbin.org/drip?numbytes=3&duration=3&delay=1&code=200", abort);
```
folsen: Is there no library to mock responses like this locally so we don't have to have an internet connection to run tests? (And potentially fail tests because `httpbin.org` is down or blocking us or something)

twittner: I know of [mockito](https://docs.rs/mockito/0.9.1/mockito/) which can replace some of these tests. However it requires `RUST_TEST_THREADS=1` and certain test cases like timeouts are curently not supported AFAICT.

dapps/src/tests/helpers/fetch.rs
```diff
@@ -96,11 +98,8 @@ impl FakeFetch {
 impl Fetch for FakeFetch {
 	type Result = Box<Future<Item = fetch::Response, Error = fetch::Error> + Send>;
 
-	fn new() -> Result<Self, fetch::Error> where Self: Sized {
-		Ok(FakeFetch::default())
-	}
-
-	fn fetch_with_abort(&self, url: &str, _abort: fetch::Abort) -> Self::Result {
+	fn fetch(&self, url: &str, abort: fetch::Abort) -> Self::Result {
+		let u = Url::parse(url).unwrap();
```
tomusdrw: Perhaps `Url` crate should be re-exported from `fetch` if it's required to instantiate some other public types (`Response`)?

dapps/src/handlers/fetch.rs
```diff
@@ -303,7 +307,7 @@ impl ContentFetcherHandler {
 					Ok(ValidatorResponse::Streaming(stream)) => {
 						trace!(target: "dapps", "Validation OK. Streaming response.");
 						let (reading, response) = stream.into_response();
-						fetch2.process_and_forget(reading);
+						pool.spawn(reading).forget();
```
tomusdrw: That could probably be pass-through (Since we use `BodyParser` to convert async->sync and later we use `into_response()` to convert from sync->async). To be investigated in subsequent PRs.

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+			let _ = self.core.clone().send(None).wait();
+		}
 	}
+}
+
+impl Client {
+	/// Create a new fetch client.
+	pub fn new() -> Result<Self, Error> {
+		let startup_done = Arc::new((Mutex::new(Ok(())), Condvar::new()));
+		let (tx_proto, rx_proto) = mpsc::channel(64);
```
tomusdrw: Might be a bit too big for our use case, since every sender get's an additional slot anyway. Probably 16 would do.

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+			// ignore send error as it means the background thread is gone already
+			let _ = self.core.clone().send(None).wait();
+		}
 	}
+}
+
+impl Client {
+	/// Create a new fetch client.
+	pub fn new() -> Result<Self, Error> {
+		let startup_done = Arc::new((Mutex::new(Ok(())), Condvar::new()));
```
tomusdrw: I suppose `std::sync::mpsc::channel/sync_channel()` would be a bit simpler here, and also supports timeouts.

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+					client.request(get(&url))
+						.map(move |resp| Response::new(url2, resp, abort2))
+						.from_err()
+						.and_then(move |resp| {
+							if abort.is_aborted() {
+								debug!(target: "fetch", "fetch of {} aborted", url);
+								return Err(Error::Aborted)
+							}
+							if let Some(next_url) = redirect_location(url, &resp) {
+								if redirects >= abort.max_redirects() {
```
tomusdrw: Is there a way to flatten the indentation here?

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+									return Err(Error::TooManyRedirects)
+								}
+								Ok(Loop::Continue((client, next_url, abort, redirects + 1)))
+							} else {
+								let content_len = resp.headers.get::<ContentLength>().cloned();
```
tomusdrw: What about chunked responses? Seems that we should handle them in `BodyReader`?

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
 		}
+		_ => None
+	}
+}
+
+// Build a simple GET request for the given Url.
+fn get(u: &Url) -> hyper::Request {
+	let uri = u.as_ref().parse().expect("Every valid URL is aso a URI.");
+	let mut rq = Request::new(Method::Get, uri);
+	rq.headers_mut().set(UserAgent::new("Parity Fetch Neo"));
```
tomusdrw: :+1: :)

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
-	fn poll(&mut self) -> futures::Poll<Self::Item, Self::Error> {
+	fn poll(&mut self) -> futures::Poll<Option<Self::Item>, Self::Error> {
 		if self.abort.is_aborted() {
-			trace!(target: "fetch", "Fetch of {:?} aborted.", self.url);
-			return Err(Error::Aborted);
+			debug!(target: "fetch", "fetch of {} aborted", self.url);
+			return Err(Error::Aborted)
 		}
+		match try_ready!(self.body.poll()) {
+			None    => Ok(Async::Ready(None)),
```
tomusdrw: Mixed tabs and spaces

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
 }
 
-/// Fetch Error
+impl io::Read for BodyReader {
+	fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+		let mut m = 0;
+		while self.body.is_some() {
+			// Can we still read from the current chunk?
+			if self.offset < self.chunk.len() {
+				let k = cmp::min(self.chunk.len() - self.offset, buf.len() - m);
```
tomusdrw: Might be good to use a bit more descriptive names for `m` and `k`

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+				(&mut buf[m .. m + k]).copy_from_slice(c);
+				self.offset += k;
+				m += k;
+				if m == buf.len() {
+					break
+				}
+			} else {
+				let body = self.body.take().expect("within this loop `self.body` is always defined");
+				match body.into_future().wait() { // wait for next chunk
+					Err((e, _))   => {
```
tomusdrw: A bunch of more tabs and space mixture for identation.

util/fetch/src/client.rs
```diff
@@ -14,347 +14,554 @@
...
+		while self.body.is_some() {
+			// Can we still read from the current chunk?
+			if self.offset < self.chunk.len() {
+				let k = cmp::min(self.chunk.len() - self.offset, buf.len() - m);
+				let c = &self.chunk[self.offset .. self.offset + k];
+				(&mut buf[m .. m + k]).copy_from_slice(c);
+				self.offset += k;
+				m += k;
+				if m == buf.len() {
+					break
```
tomusdrw: I would put `return Ok(m)` here and instead of the other `break` - thanks to this we don't need the `Ok(m)` at the bottom

util/fetch/src/client.rs
tomusdrw: Perhaps just spawn an http server in tests? Or at least bind a TcpSocket to a random port and respond with some predefined data.

dapps/src/tests/helpers/fetch.rs
twittner: Yes, this may be convenient.

util/fetch/src/client.rs
twittner: I updated the tests to run a local test server instead.

util/fetch/src/client.rs
twittner: Right, I updated `BodyReader` to track the number of bytes read and error out if the maximum is exceeded. While this logic was part of the `Stream` impl of `Response` it wasn't in `BodyReader`.

util/fetch/src/client.rs
twittner: I think `Ok(m)` is still needed there, as the `while` loop might theoretically not be entered at all.

util/fetch/src/client.rs
twittner: Thanks, I updated the message.

util/fetch/src/client.rs
tomusdrw: Oh, that's true. We could use `loop` and `if let Some(body) = self.body.take()` inside to avoid `expect`?

kirushik:
:tada:

Issue #7915, opened 2018-02-16
Get the light client to compile for wasm


tomaka:
cc @rphmeier 

In order to be able to run the light client inside of a browser extension.

Problematic libraries include `mio` (doesn't support emscripten), `ring` (uses raw assembly), `rocksdb` and `openssl` (both will be hard to compile).

kirushik:
`openssl` will likely be replaced with `rustls` as a part of #7829, so it's eventually will be the `ring` issue again.

tomaka:
A good first step for this issue is either #7954 or #7955.

rphmeier:
I would break up the tasks for this like so:
  - [ ] Break out any explicit dependency on `ethcore-network` from `ethcore-light`, instead doing everything through traits
  - [ ] Break out any explicit dependency on `rocksdb` (this may already not exist as everything is done through the `KeyValueDb` trait.
  - [ ] Compile `ethcore-light` and `ethcore-rpc` for WASM.
  - [ ] Write a Rust wrapper which expects extern functions that can provide network and DB context somehow and also exports a function for making RPC calls.
  - [ ] Write JS code that uses libp2p and local storage to fulfill the requirements of the Rust wrapper and can use its RPC function to answer its queries

tomaka:
Should the network code be hard-coded to libp2p/devp2p (eg. pass an instance of `libp2p::Transport`), or should it be more generic?

I'd go for the first option, because Parity may depend on some specific networking behaviours. If we go for the second option, every single corner case would need to be documented in the API.

rphmeier:
Keeping it generic seems the best way to make it testable. Right now the light client network code is fairly network-independent and exposes a few hooks that can be called by devp2p or libp2p or an in-memory test network.

tomaka:
> In general, how are you supposed to unit test libp2p-based code?

You can fairly easily create an implementation of [the `Transport` trait](https://github.com/libp2p/rust-libp2p/blob/3b859b683357abbb13bb9f48c97643363446beb1/libp2p-swarm/src/transport.rs#L73-L103) that reads/write to/from memory, or that calls a custom handler.

tomaka:
> Take ethcore-io stuff out of ethcore to cut out the mio dependency.

This is really difficult, so instead I suggest adding an optional `mio` feature to `ethcore-io`.
Only `network-devp2p` would require that `mio` feature.

tomaka:
Notes for myself for crypto libraries alternatives:
[...]

PR #8100, opened 2018-03-13
Remove the time dependency where possible

tomaka:
cc #7915 

The `time` crate is no longer maintained. This PR replaces it with `std::time` where possible. Wherever `strftime` is used, the `time` dependency has been kept because there is no straight-forward replacement yet.

In addition to being IMO a positive change, this should also help compile for wasm.

The logic of the code has been preserved, except in a few places where initializing a `u64` to `0` has been replaced with initializing an `Instant` to `Instant::now()`. Since the `0` was mostly a place-holder (similar to `None`), the behaviour should be exactly the same in the end.


tomaka:
Note that in the future when `const fns` are stable, we should replace the various `u64` constants with `Duration` constants.

tomaka:
After this PR, `time` is used in the following places:

- In `hyper` and related crates ; hyper will probably be stripped down for the client light, because we don't need anything HTTP-related.
- `ethstore` because it requires `strftime` to generale file names. It should be possible to tweak something for wasm here.
- `zip` (and `msdos_time`) ; used only by the `dapps` crate which we don't need for the light-client-in-the-browser objective.
- `node_health` (and its dependency `ntp`) and `logger` because they require `strftime`. This is probably the hardest one to fix.
- `rust-crypto` ; I should be able to submit a PR there.
- `transient-hashmap` ; I should be able to submit a PR there.
- `vergen` ; but it's only used as a build-dependency so it's not a problem.

tomaka:
I just realized that `std::time::Instant` is *not* steady, so replacing `SteadyTime` with `Instant` could be wrong. However considering that it's only about caching and network timeouts, I don't think that we actually need steady clocks.

niklasad1:
@tomaka 
What about [std::time::SystemTime](https://doc.rust-lang.org/std/time/struct.SystemTime.html) it is not monotonically increasing but steady I think.

PR #8101, opened 2018-03-13
Fix comment for gas extern in Wasm runtime

fckt:


NikVolf:
New comment is wrong, this external is not only about block of execution, it is general gas charge; so please revert to previous or update to something with complete meaning

fckt:
@NikVolf 
Isn't `gas_charge` is about general gas charge?
Are you sure we better to revert the previous which is:
`Report gas cost with the params passed in wasm stack`

NikVolf:
`gas_charge` is not an extern function

Previous comment was at least correct. For example, grow_memory also is charged via `gas` extern, it's not  only about blocks.

fckt:
let’s just make a pull request to extend/fix the comment. The problem with the previous version was that it confuses user and not clarifies the actual purpose of the extern

PR #8102, opened 2018-03-13
Fix Cargo.lock not parsable

tomaka:
`master` doesn't build for me because `Cargo.lock` isn't parsable.


tomaka:
```
[pierre@pierre-thinkpad parity]$ git fetch upstream
[pierre@pierre-thinkpad parity]$ git checkout upstream/master
HEAD is now at 102bc7809... Additional data in EVMTestClient (#7964)
[pierre@pierre-thinkpad parity]$ cargo b --verbose
error: failed to parse lock file at: /home/pierre/Projets/parity/Cargo.lock

Caused by:
  package `ethabi-derive 5.0.5 (registry+https://github.com/rust-lang/crates.io-index)` is specified as a dependency, but is missing from the package list
```

PR #8103, opened 2018-03-13
Replace std::env::temp_dir with tempdir in tests

niklasad1:
Attempt to close [#6147](https://github.com/paritytech/parity/issues/6147)

The remaining occurrences of `std::env::temp_dir()` are the following: 
```bash
$ git grep -n "::std::env::temp_dir"
ethcore/src/ethereum/mod.rs:36:         None => Spec::load(&::std::env::temp_dir(), b)
ethcore/src/ethereum/mod.rs:148:                let spec = new_morden(&::std::env::temp_dir());
ethcore/src/ethereum/mod.rs:163:                let morden = new_morden(&::std::env::temp_dir());
ethcore/src/ethereum/mod.rs:174:                let frontier = new_foundation(&::std::env::temp_dir());
ethcore/src/spec/spec.rs:416:                   &::std::env::temp_dir(),
evmbin/src/main.rs:243:                         spec::Spec::load(&::std::env::temp_dir(), file)?
evmbin/src/main.rs:246:                         ethcore::ethereum::new_foundation(&::std::env::temp_dir())
parity/account.rs:74:   let spec = spec.spec(&::std::env::temp_dir())?;
```

### DIFF
--- a/ethash/src/compute.rs
+++ b/ethash/src/compute.rs
@@ -315,6 +315,7 @@ fn calculate_dag_item(node_index: u32, cache: &[Node]) -> Node {
 mod test {
 	use super::*;
 	use std::fs;
+	use tempdir::TempDir;
 
 	#[test]
 	fn test_get_cache_size() {
@@ -387,7 +388,7 @@ mod test {
 		];
 		let nonce = 0xd7b3ac70a301a249;
 		// difficulty = 0x085657254bd9u64;
-		let light = NodeCacheBuilder::new(None).light(&::std::env::temp_dir(), 486382);
+		let light = NodeCacheBuilder::new(None).light(TempDir::new("").unwrap().path(), 486382);
 		let result = light_compute(&light, &hash, nonce);
 		assert_eq!(result.mix_hash[..], mix_hash[..]);
 		assert_eq!(result.value[..], boundary[..]);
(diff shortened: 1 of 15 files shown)


ethash/src/lib.rs
```diff
@@ -135,7 +138,9 @@ impl EthashManager {
 
 #[test]
 fn test_lru() {
-	let ethash = EthashManager::new(&::std::env::temp_dir(), None);
+	use tempdir::TempDir;
+
+	let ethash = EthashManager::new(TempDir::new("").unwrap().path(), None);
```
debris: It won't work as expected, cause `TempDir` is dropped before `EthashManager`, so the directory won't be cleaned up

ethcore/node_filter/src/lib.rs
```diff
@@ -126,13 +128,14 @@ mod test {
...
 	use super::NodeFilter;
+	use tempdir::TempDir;
 
 	/// Contract code: https://gist.github.com/arkpar/467dbcc73cbb85b0997a7a10ffa0695f
 	#[test]
 	fn node_filter() {
 		let contract_addr = Address::from_str("0000000000000000000000000000000000000005").unwrap();
 		let data = include_bytes!("../res/node_filter.json");
-		let spec = Spec::load(&::std::env::temp_dir(), &data[..]).unwrap();
+		let spec = Spec::load(&TempDir::new("").unwrap().path(), &data[..]).unwrap();
```
debris: same here, `TempDir` will be dropped to early

ethash/src/compute.rs
```diff
@@ -387,26 +388,26 @@ mod test {
 		];
 		let nonce = 0xd7b3ac70a301a249;
 		// difficulty = 0x085657254bd9u64;
-		let light = NodeCacheBuilder::new(None).light(&::std::env::temp_dir(), 486382);
+		let light = NodeCacheBuilder::new(None).light(TempDir::new("").unwrap().path(), 486382);
```
debris: and here ;)

niklasad1:
Well, technically I think the first commit is fine because the `argument<path>` in this case must have a bigger lifetime than the function itself and therefore the `tmpdir` will cleaned up.

Am I wrong?

debris:
> Well, technically I think the first commit is fine because the argument<path> in this case must have a bigger lifetime than the function itself and therefore the tmpdir will be cleaned up.
>
> Am I wrong?

You are right when it comes to function calls, but in many places the path was used to initialize a structure (e.g. `EthashManager`), and because `TempDir` was dropped before `EthashManager`, the directory was never cleaned ;)

PR #8109, opened 2018-03-13
optional dependency on secp256k1 for ethcrypto

rphmeier:
a lot of this crypto code is more generally useful even without an secp256k1 dependency


folsen:
LGTM, but could we also add a readme in this folder for what's in this crate and why you would want to use what feature? I.E. "By default you can do this, if you enable this feature you can also do that"

PR #8114, opened 2018-03-14
Keep existing blocks when restoring from a snapshot

Vurich:
This is currently blocked on testing. I don't know how to test this code, especially since it's heavily tied to the file system.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -18,7 +18,7 @@
 
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
-use std::mem;
+use std::{mem, cmp};
 use itertools::Itertools;
 use heapsize::HeapSizeOf;
 use ethereum_types::{H256, Bloom, U256};
@@ -173,7 +173,7 @@ pub struct BlockChain {
 	best_ancient_block: RwLock<Option<BestAncientBlock>>,
 	// Stores the last block of the last sequence of blocks. `None` if there are no gaps.
 	// This is calculated on start and does not get updated.
-	first_block: Option<H256>,
+	first_block: RwLock<Option<H256>>,
 
 	// block cache
 	block_headers: RwLock<HashMap<H256, Bytes>>,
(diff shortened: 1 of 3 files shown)


ethcore/src/blockchain/blockchain.rs
```diff
@@ -195,6 +195,77 @@ pub struct BlockChain {
 	pending_transaction_addresses: RwLock<HashMap<H256, Option<TransactionAddress>>>,
 }
 
+// TODO: Should this and `BestAncientBlock` be the same struct?
+#[derive(Clone, PartialEq, Eq)]
+struct Block {
```
rphmeier: having this be called `Block` in the `blockchain` module is pretty confusing

ethcore/src/snapshot/service.rs
```diff
@@ -481,15 +481,26 @@ impl Service {
...
+			let recovery = self.temp_recovery_dir();
 			let snapshot_dir = self.snapshot_dir();
 
 			if snapshot_dir.exists() {
+				for file in fs::read_dir(&snapshot_dir)? {
+					let file = file?;
+
+					let new_path = recovery.with_file_name(file.file_name());
+
+					fs::copy(file.path(), &new_path)?;
```
rphmeier: copying the whole current snapshot every time we attempt to overwrite the old? also doesn't account for subdirectories.

I think it's fine if recovery failure ends up with the local snapshot also disappearing. We won't use nearly as much space on disk and it's such an edge case that it's not worth the extra cost.

rphmeier: actually I think this code is completely broken. We only want the files in `temp_recovery_dir`, not the files in `snapshot_dir`. Copying them over will cause us to keep around all the snapshots we've ever downloaded (as opposed to the single most recent one).

and only the oldest of them will be accessible: one of the files in the current snapshot dir is a `MANIFEST`, which would then overwrite the `MANIFEST` in the temp recovery dir.

rphmeier:
`migrate_from` should be called around here: https://github.com/paritytech/parity/pull/8114/files#diff-09408288a391322d847f21fbec62b925L191

Testing the service itself is fairly difficult, but testing `migrate_from` doesn't require any file system access at all. As long as `migrate_from` works, I don't think we need to worry about testing that a specific call to it works in the service.

ethcore/src/snapshot/service.rs
Vurich: Ok, we need to chat about this so I can get it right. I have no idea what the structure of this on the file system is, I was just trying to infer it from the code.

ethcore/src/blockchain/blockchain.rs
Vurich: Agreed. Maybe `MinimalBlockMetadata` would be more accurate (if kinda verbose)

ethcore/src/snapshot/service.rs
rphmeier: once we've restored from a snapshot, we keep it on disk in the `snapshot_dir` and repropagate. but until the restoration is finished we keep the snapshot in a temporary directory. If there's an old snapshot, it's fine to overwrite because we would have only restored if the new one is more recent. The changes you included make it so we keep every snapshot we've ever downloaded, but only repropagate the oldest.

5chdn:
Please reopen once you continue working on this.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1299,6 +1389,115 @@ impl BlockChain {
...
+					hash: low.hash,
+					number: low.number,
+				}
+			);
+
+			let mut transaction = self.db.transaction();
+			transaction.put(db::COL_EXTRA, b"first", &high.hash);
+			transaction.put(db::COL_EXTRA, b"ancient", &low.hash);
+
+			self.db.write(transaction).map_err(|_| ())?;
```
ngotchac: There might be an issue here in this specific scenario (`N` is the number of blocks included in a snapshot):

You got a snapshot until block `B1`, and synced ancient blocks from 0 to `AB1`, new block up to `B1 + N`. Some time later, you get a snapshot at `B2 >> B1 + N`. So after the DB migration, you basically have `0` to `AB1`, `B1` to B1 + N`, `B2` to `B2 + N`.

The `new_gap` is then `[AB1 ; B1]`, so `first` would be `B1` which is wrong since you actually miss `[B1 + N ; B2]`.
Even if the gap would be `[AB1 ; B2]`, by writing to DB the blocks `[B1; B1+N]` I think you will run into situations where the old block downloader will stop at `B1` since it hit a known block (cf `insert_unordered_block` method).

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1299,6 +1389,115 @@ impl BlockChain {
...
+
+			into.write(transaction)
+		}
+
+		migrate_hashmap(&mut *self.block_headers.write(), &*other.block_headers.read());
+		migrate_hashmap(&mut *self.block_bodies.write(), &*other.block_bodies.read());
+		migrate_hashmap(&mut *self.block_details.write(), &*other.block_details.read());
+		migrate_hashmap(&mut *self.block_hashes.write(), &*other.block_hashes.read());
+		migrate_hashmap(&mut *self.transaction_addresses.write(), &*other.transaction_addresses.read());
+		migrate_hashmap(&mut *self.block_receipts.write(), &*other.block_receipts.read());
```
ngotchac: I'm not sure these are necessary, since they are caches.

PR #8125, opened 2018-03-15
secret_store: use `ecies::encrypt`/`ecies::decrypt`.

twittner:


secret_store/src/key_server_cluster/io/message.rs
```diff
@@ -241,15 +241,15 @@ pub fn deserialize_message(header: &MessageHeader, payload: Vec<u8>) -> Result<M
 pub fn encrypt_message(key: &KeyPair, message: SerializedMessage) -> Result<SerializedMessage, Error> {
 	let mut header: Vec<_> = message.into();
 	let payload = header.split_off(MESSAGE_HEADER_SIZE);
-	let encrypted_payload = encrypt_single_message(key.public(), &payload)?;
```
svyatonik: I don't mind using the `&[]`, but there are already a couple of places in SS where `&ethcrypto::DEFAULT_MAC` is passed instead. So maybe better to use the same value everywhere. But it is up to you.

PR #8130, opened 2018-03-16
explicitly mention pruning_history uses RAM, Fixes #8129

### LINKED ISSUE #8129: Pruning History keeps states in memory - not explicit
> I'm running:
>
> - **Which Parity version?**: 1.9.4 & 1.8.11
> - **How installed?**: .deb package

In an attempt to get a "semi-pruned-archive" node going (basically turn on archive from block 4_400_000) I had set pruning-history to 1_100_000. The config generator complained but parity didn't. **I presumed this was stored on disk in the same way a `pruning=archive` node would.**

Suggestion is to make it *explicit* that `pruning_history` is stored in RAM.

Ran into an issue where nodes would eat 29 GB of memory (this was 100% of available ram) then crash _without an error and with exit code 0_ 

parity would climb to 20-25 GB of ram using 1 cpu core only, oscillate in the range for a bit (2-3 min), spike up to 29GB and restart.

I started with 1.9.4 then tried migrating down to 1.8.11 with the same db / config, then removed `purning_history`, and still had the same issue. Migrated back up to 1.9.4 and the issue persisted. Then deleted `./chains` and parity started syncing again.

One of the funny thing was `./chains` was only 12GB when this was happening - so not sure why parity was eating 2.5x that in RAM. (Had sync'd up to about block 1_300_000)

XertroV:
Fixes #8129

### DIFF
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -768,7 +768,7 @@ usage! {
 
 			ARG arg_pruning_history: (u64) = 64u64, or |c: &Config| c.footprint.as_ref()?.pruning_history.clone(),
 			"--pruning-history=[NUM]",
-			"Set a minimum number of recent states to keep when pruning is active.",
+			"Set a minimum number of recent states to keep in memory when pruning is active.",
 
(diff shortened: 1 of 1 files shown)


XertroV:
[clabot:check]

PR #8132, opened 2018-03-16
Backport #8099 to beta

debris:
I created branch during cherry-pick, so the commit message is malformed, but it's only a backport


tomusdrw:
Please check binary size before merging, if it increases the size we should not backport it to `beta`.

andresilva:
I tested the size by merging this PR into #8136. On darwin, after stripping binaries:

Before: `-rwxr-xr-x  1 andre  staff  51827304 Mar 19 18:55 target/release/parity`
After: `-rwxr-xr-x  1 andre  staff  52282412 Mar 19 19:00 target/release/parity`

It's still under the 50MB mark but very close.

tomusdrw:
@andresilva @debris Let's realease without it for now, and maybe include it in next patch release. Hopefuly by then a lot of people upgrade to a non 50MB limited version so we will have more breathing space with the size.

PR #8134, opened 2018-03-16
Check one step deeper if we're on release track branches

kirushik:


General-Beck:
it is necessary to test in a master and beta branch

PR #8137, opened 2018-03-16
ethcore: postpone Kovan hard fork

andresilva:
```
❯ date -d "+$(( (6600000 - 6389810) * 4 )) seconds"
Mon Mar 26 10:45:53 WEST 2018
```

### DIFF
--- a/ethcore/res/ethereum/kovan.json
+++ b/ethcore/res/ethereum/kovan.json
@@ -36,14 +36,14 @@
 		"forkCanonHash": "0x0a66d93c2f727dca618fabaf70c39b37018c73d78b939d8b11efbbd09034778f",
 		"eip155Transition": 1000000,
 		"maxCodeSize": 24576,
-		"maxCodeSizeTransition": 6500000,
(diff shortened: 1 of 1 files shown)


tomusdrw:
Can we also update fork number in `util/version/Cargo.toml` and mark the release as critical?

util/version/Cargo.toml
```diff
@@ -13,13 +13,13 @@ build = "build.rs"
 track = "nightly"
 
 # Indicates a critical release in this track (i.e. consensus issue)
-critical = false
+critical = true
```
andresilva: When do we switch this back to `false` afterwards?

tomusdrw: That's a bit unclear for me as well :)
I suppose we should flip it back when bumping new version, but clients updating from say 1.x.N to 1.x.(N+2) should still consider that critical, even though `N+2` is not. I don't think that logic is implemented in the updater currently, but don't consider that a blocker.

PR #8144, opened 2018-03-17
[stable] Fix scripts. Force JS rebuild.

tomusdrw:


General-Beck:
bulletproof logic!

tomusdrw:
@General-Beck We don't care for that optimization for stable/beta (it's really only useful for PRs and master). Everytime we merge pull request we want to the build to proceed. Right now it doesn't work at all.

General-Beck:
@tomusdrw I agree. optimization from 1.10 +

PR #8150, opened 2018-03-19
private transactions pr improvements

debris:
changes:
- [x] Client does not contain `private_notify`
- [x] using io service thread pool for background processing is less complicated:
  ```diff
  - private_transactions -> channel -> client_service -> client -> private_transactions
  + private_transactions -> channel -> client_service -> private_transactions
  ```
- [x] ethsync does not depend on private_transactios module
- [x] simplified ethsync tests
- [x] implement `PrivateTxHandler` for `Provider` from `private_transactions` module
- [x] rename `private_transactions` to `ethcore-private-tx` and move it to `ethcore` directory


ethcore/private-tx/src/lib.rs
```diff
@@ -146,13 +148,16 @@ pub struct Receipt {
 
 /// Manager of private transactions
 pub struct Provider {
-	encryptor: Arc<Encryptor>,
-	config: ProviderConfig,
+	encryptor: Box<Encryptor>,
```
debris: encryptor is never shared, so there is no reason to make wrap it `Arc`

ethcore/private-tx/src/lib.rs
```diff
@@ -146,13 +148,16 @@ pub struct Receipt {
...
+	validator_accounts: HashSet<Address>,
```
debris: `validator_accounts` is only accessed with `contains` method, so there it should be `HashSet`, not `Vec`

ethcore/private-tx/src/lib.rs
```diff
@@ -146,13 +148,16 @@ pub struct Receipt {
...
+	signer_account: Option<Address>,
+	passwords: Vec<String>,
 	notify: RwLock<Vec<Weak<ChainNotify>>>,
 	transactions_for_signing: Mutex<SigningStore>,
 	transactions_for_verification: Mutex<VerificationStore>,
 	client: Arc<Client>,
 	accounts: Arc<AccountProvider>,
+	channel: IoChannel<ClientIoMessage>,
```
debris: `Provider` is initiated with channel instead of fetching it from the client

ethcore/private-tx/src/lib.rs
```diff
@@ -259,19 +272,16 @@ impl Provider where {
 
 	/// Process received private transaction
 	pub fn import_private_transaction(&self, rlp: &[u8]) -> Result<(), Error> {
-		let validator_accounts = self.config.validator_accounts.clone();
 		trace!("Private transaction received");
-		if validator_accounts.is_empty() {
-			self.broadcast_private_transaction(rlp.into());
-			return Ok(());
-		}
```
debris: imo this was a logic error, cause we could broadcast was done before checking the validity of private transaction rlp

ethcore/private-tx/src/encryptor.rs
```diff
@@ -86,12 +86,11 @@ pub struct SecretStoreEncryptor {
 
 impl SecretStoreEncryptor {
 	/// Create new encryptor
-	pub fn new(config: EncryptorConfig) -> Result<Self, Error> {
+	pub fn new(config: EncryptorConfig, client: FetchClient) -> Result<Self, Error> {
 		Ok(SecretStoreEncryptor {
-			config: config,
-			client: FetchClient::new()
-				.map_err(|e| ErrorKind::Encrypt(format!("{}", e)))?,
```
debris: We should reuse existing `FetchClient`, cause each new instance of it creates new `CpuPool`.

parity/run.rs
```diff
@@ -697,8 +692,8 @@ pub fn execute_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>)
 	if let Some(filter) = connection_filter {
 		service.add_notify(filter);
 	}
-	private_provider.add_notify(chain_notify.clone());
-	service.add_notify(private_provider.clone());
+	private_tx_provider.add_notify(chain_notify.clone());
+	service.add_notify(private_tx_provider.clone());
```
debris: this is utterly bad... Currently there is no way to prove in a test that this notifications has been set. Tbh I don't even know how to change this without significant changes to our api

sync/src/tests/private_transactions.rs
```diff
@@ -1,150 +0,0 @@
...
-	let signer_config = ProviderConfig{
-		validator_accounts: Vec::new(),
-		signer_account: Some(s0.address()),
-		passwords: vec!["".into()],
-	};
-
-	let mut net = TestNet::with_spec_and_accounts(2, SyncConfig::default(), seal_spec, Some(ap.clone()), Some(vec![signer_config, validator_config]));
-	let client0 = net.peer(0).chain.clone();
-	let client1 = net.peer(1).chain.clone();
-	let provider0 = net.peer(0).private_provider.clone().unwrap().clone();
```
debris: imo, this is a bad tests. It does not prove that private_transactions are working. it's partly caused by the fact how our api looks like currently

This test mutates several objects (client, miner, engine, net) and assumes that real code will mutate these objects in the same way before executing private transaction

ethcore/private-tx/src/lib.rs
```diff
@@ -240,9 +251,8 @@ impl Provider where {
 
 	/// Try to unlock account using stored passwords
 	fn unlock_account(&self, account: &Address) -> bool {
-		let passwords = self.config.passwords.clone();
-		for password in passwords {
-			if let Ok(()) = self.accounts.unlock_account_temporarily(account.clone(), password) {
+		for password in &self.passwords {
+			if let Ok(()) = self.accounts.unlock_account_temporarily(account.clone(), password.clone()) {
```
debris: this is a huge security flaw,

signing operation is not atomic: unlock_account unlocks account for one usage. there is no guarantee that one or more other service will not use unlocked accounts in the meantime

ethcore/private-tx/src/encryptor.rs
```diff
@@ -187,9 +186,8 @@ impl SecretStoreEncryptor {
 
 	/// Try to unlock account using stored passwords
 	fn unlock_account(&self, account: &Address, accounts: Arc<AccountProvider>) -> bool {
-		let passwords = self.config.passwords.clone();
-		for password in passwords {
-			if let Ok(()) = accounts.unlock_account_temporarily(account.clone(), password) {
+		for password in &self.config.passwords {
+			if let Ok(()) = accounts.unlock_account_temporarily(account.clone(), password.clone()) {
```
debris: another security flaw,

also this method is duplicated in 2 different implementations

ethcore/private-tx/src/lib.rs
```diff
@@ -530,15 +528,15 @@ impl Provider where {
 
 		let mut state = self.client.state_at(block).ok_or(ErrorKind::StatePruned)?;
 		// TODO: in case of BlockId::Latest these need to operate on the same state
```
debris: I don't know if this is still relevant

sync/src/tests/private_transactions.rs
grbIzl: It's not the test for private transactions as a whole functionality. It's mostly for its sync part (that's why it's placed here) to test, that messages communicated properly between clients.

ethcore/private-tx/src/lib.rs
grbIzl: It was before :-(

ethcore/private-tx/src/encryptor.rs
tomusdrw: Can we do anything about this?

ethcore/private-tx/src/lib.rs
```diff
@@ -146,33 +145,45 @@ pub struct Receipt {
...
 
 impl Provider where {
 	/// Create a new provider.
-	pub fn new(client: Arc<Client>, accounts: Arc<AccountProvider>, encryptor: Arc<Encryptor>, config: ProviderConfig) -> Result<Self, Error> {
+	pub fn new(
+		client: Arc<Client>,
+		accounts: Arc<AccountProvider>,
+		encryptor: Box<Encryptor>,
+		config: ProviderConfig,
+		channel: IoChannel<ClientIoMessage>
```
tomusdrw: comma missing

ethcore/private-tx/src/encryptor.rs
```diff
@@ -145,7 +144,7 @@ impl SecretStoreEncryptor {
 		// response is JSON string (which is, in turn, hex-encoded, encrypted Public)
 		let encrypted_bytes: ethjson::bytes::Bytes = result.trim_matches('\"').parse().map_err(|e| ErrorKind::Encrypt(e))?;
 		if !self.unlock_account(&requester, accounts.clone()) {
```
tomusdrw: Do we really need to `clone()` the Arc here seems that `&*accounts` would suffice. Anyhow, why do we need unlocking at all? Just attempt to decrypt and let it fail.

ethcore/private-tx/src/encryptor.rs
```diff
@@ -187,9 +186,8 @@ impl SecretStoreEncryptor {
 
 	/// Try to unlock account using stored passwords
 	fn unlock_account(&self, account: &Address, accounts: Arc<AccountProvider>) -> bool {
-		let passwords = self.config.passwords.clone();
-		for password in passwords {
-			if let Ok(()) = accounts.unlock_account_temporarily(account.clone(), password) {
+		for password in &self.config.passwords {
```
tomusdrw: I don't get, why do we have one `account`, but multiple passwords? Also why not just pass password as additional argument to required method?

ethcore/private-tx/src/encryptor.rs
```diff
@@ -204,7 +202,7 @@ impl SecretStoreEncryptor {
 			Ok(accounts.sign(key_server_account.clone(), None, H256::from_slice(&contract_address_extended))?)
```
tomusdrw: Doesn't make sense to unlock then sign, just pass the password as second argument or if we need multiple passwords for whatever reason we should just try to sign in a loop.

ethcore/service/src/service.rs
```diff
@@ -25,19 +25,51 @@ use kvdb::KeyValueDB;
...
+	}
+
+	/// Returns underlying provider.
+	pub fn provider(&self) -> Arc<ethcore_private_tx::Provider> {
+		self.provider.clone()
+	}
+}
+
+impl PrivateTxHandler for PrivateTxService {
+	fn import_private_transaction(&self, rlp: &[u8]) -> Result<(), String> {
```
tomusdrw: Do we really want to have stringly typed errors everywhere?

sync/src/private_tx.rs
```diff
@@ -0,0 +1,60 @@
...
+	fn import_private_transaction(&self, _rlp: &[u8]) -> Result<(), String> {
+		Ok(())
+	}
+
+	fn import_signed_private_transaction(&self, _rlp: &[u8]) -> Result<(), String> {
+		Ok(())
+	}
+}
+
+/// Simple private transaction handler. Used for tests.
```
tomusdrw: mark it as `#[cfg(test)]` then

sync/src/private_tx.rs
```diff
@@ -0,0 +1,60 @@
...
+pub trait PrivateTxHandler: Send + Sync + 'static {
+	/// Function called on new private transaction received.
+	fn import_private_transaction(&self, rlp: &[u8]) -> Result<(), String>;
+
+	/// Function called on new signed private transaction received.
+	fn import_signed_private_transaction(&self, rlp: &[u8]) -> Result<(), String>;
+}
+
+/// Nonoperative private transaction handler. Used for tests.
+pub struct NoopPrivateTxHandler;
```
tomusdrw: `#[cfg(test)]`

ethcore/src/client/client.rs
```diff
@@ -752,7 +751,7 @@ impl Client {
 			db: RwLock::new(db),
 			state_db: RwLock::new(state_db),
 			report: RwLock::new(Default::default()),
-			private_notify: RwLock::new(None),
+			//private_notify: RwLock::new(None),
```
tomusdrw: Remove

ethcore/src/error.rs
```diff
@@ -272,6 +272,13 @@ impl fmt::Display for Error {
 	}
 }
 
+impl error::Error for Error {
+	fn description(&self) -> &str {
+		// improve description
```
tomusdrw: indeed

sync/src/tests/helpers.rs
```diff
@@ -551,8 +554,8 @@ impl IoHandler<ClientIoMessage> for TestIoHandler {
 			ClientIoMessage::NewMessage(ref message) => if let Err(e) = self.client.engine().handle_message(message) {
 				panic!("Invalid message received: {}", e);
 			},
-			ClientIoMessage::NewPrivateTransaction => if let Err(e) = self.client.handle_private_message() {
-				warn!("Failed to handle private transaction {:?}", e);
+			ClientIoMessage::NewPrivateTransaction => {
+				*self.private_tx_queued.lock() += 1;
```
tomusdrw: `AtomicUsize`?

sync/src/tests/private_transactions.rs
tomusdrw: IMHO even partial test is better than none, so removing it without providing alternative is not so good.

parity/run.rs
tomusdrw: IMHO it's fine, you cannot prove many things with unit tests, that's why sometimes you write integration ones. That said,  the whole plumbing in `parity/run.rs` (especially since it's duplicated for full and light client) is currently non testable, and we should at least try to make it testable or have a framework for even more high-level (hive-like) tests (run couple of instances of parity).

ethcore/service/src/service.rs
```diff
@@ -25,19 +25,51 @@ use kvdb::KeyValueDB;
...
+
+impl PrivateTxService {
+	fn new(provider: Arc<ethcore_private_tx::Provider>) -> Self {
+		PrivateTxService {
+			provider,
+		}
+	}
+
+	/// Returns underlying provider.
+	pub fn provider(&self) -> Arc<ethcore_private_tx::Provider> {
```
tomusdrw: Why some code relies on the internal provider? Not sure I follow the abstractions here I suppose this struct is created to be able to implement the trait from `ethsync`, but I don't see why we make that one public and also the internals public.
Perhaps implement the trait for the entire `Service` instead? This looks like multiplying entities for no good reason.

ethcore/private-tx/src/encryptor.rs
grbIzl: Reworked

ethcore/private-tx/src/encryptor.rs
grbIzl: User can use several different accounts: for signing transaction, for verification, for signing messages. Usage of password reworked

ethcore/private-tx/src/lib.rs
grbIzl: Reworked

ethcore/private-tx/src/encryptor.rs
grbIzl: Reworked

debris:
merged with master and them into `PrivateTransactionsNetwork` pr. Will address the remaining issues over there

PR #8152, opened 2018-03-19
[stable] Postpone Kovan hard fork (#8137)

tomusdrw:
* ethcore: postpone Kovan hard fork

* util: update version fork metadata


tomusdrw:
@andresilva yeah, it seems that all the changes are already in `stable`, so I allowed an empty commit to just reference that PR. Feel free to double check that though.

andresilva:
Yep, everything is in there (https://github.com/paritytech/parity/pull/8133/commits/e680361274b7b676f598dbe15692ae4be5a6072d).

PR #8154, opened 2018-03-19
Revert Andre fix coverage build

General-Beck:
https://github.com/paritytech/parity/pull/8118#event-1522171535


andresilva:
Replaced with #8287.

PR #8163, opened 2018-03-20
Some tiny modifications.

EighteenZi:
1. fix some typo in the comment.
2. sort the order of methods in 'impl state::Backend for StateDB`

[clabot:check]


EighteenZi:
[clabot:check]

PR #8164, opened 2018-03-20
Change name Wallet -> UI

amaurymartiny:
After some discussing, Parity UI seemed to be the most popular name for standalone ui app (instead of Parity Wallet).

To be backported to 1.10

### DIFF
--- a/parity/run.rs
+++ b/parity/run.rs
@@ -893,8 +893,8 @@ pub fn execute_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>)
 
 pub fn execute(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>) -> Result<(bool, Option<String>), String> {
 	if cmd.ui_conf.enabled {
-		warn!("Parity browser interface is deprecated. It's going to be removed in the next version, use standalone Parity Wallet instead.");
-		warn!("Standalone Parity Wallet: https://github.com/Parity-JS/shell/releases");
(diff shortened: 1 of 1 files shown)


parity/run.rs
```diff
@@ -893,8 +893,8 @@ pub fn execute_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>)
 
 pub fn execute(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>) -> Result<(bool, Option<String>), String> {
 	if cmd.ui_conf.enabled {
-		warn!("Parity browser interface is deprecated. It's going to be removed in the next version, use standalone Parity Wallet instead.");
-		warn!("Standalone Parity Wallet: https://github.com/Parity-JS/shell/releases");
+		warn!("Parity browser interface is deprecated. It's going to be removed in the next version, use standalone Parity UI instead.");
```
tomusdrw: Looks good, but noticed that it's barely visible, so maybe we could make it bold?

5chdn:
I remember trying very hard last year to do exactly the opposite. UI was our internal reference for a product initially called "wallet".

I would advise against communicating "UI" externally unless you really insist on this terminology for the broader audience. Also, references would have to be updated everywhere in the documentation. https://wiki.parity.io/Parity-Wallet

jacogr:
The "Wallet" is now just a dapp. We don't have a "Wallet"-only anymore, we have a "dapp browser". (This all changed in 1.9)

PR #8166, opened 2018-03-20
Handle rlp decoding Result in patricia trie

ascjones:
Part of #8033, replacing existing usage of `Rlp` with `UntrustedRlp`. 

Unwraps the result of decoding a node's rlp, assuming valid rlp since it is a local read from the db.

Next step might be to introduce another case to TrieError for a DecoderError to avoid the unwraps.

### DIFF
--- a/util/patricia_trie/src/node.rs
+++ b/util/patricia_trie/src/node.rs
@@ -40,9 +40,31 @@ pub enum Node<'a> {
 
 impl<'a> Node<'a> {
 	/// Decode the `node_rlp` and return the Node.
-	pub fn decoded(node_rlp: &'a [u8]) -> Self {
+	pub fn decoded(node_rlp: &'a [u8]) -> Result<Self, DecoderError> {
 		let r = UntrustedRlp::new(node_rlp);
-		Node::decode_rlp(&r).expect("Self encoded Rlp should be valid; qed")
+		match r.prototype()? {
+			// either leaf or extension - decode first item with NibbleSlice::???
+			// and use is_leaf return to figure out which.
+			// if leaf, second item is a value (is_data())
+			// if extension, second item is a node (either SHA3 to be looked up and
+			// fed back into this function or inline RLP which can be fed back into this function).
+			Prototype::List(2) => match NibbleSlice::from_encoded(r.at(0)?.data()?) {
+				(slice, true) => Ok(Node::Leaf(slice, r.at(1)?.data()?)),
+				(slice, false) => Ok(Node::Extension(slice, r.at(1)?.as_raw())),
+			},
+			// branch - first 16 are nodes, 17th is a value (or empty).
+			Prototype::List(17) => {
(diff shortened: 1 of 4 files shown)


util/patricia_trie/src/node.rs
```diff
@@ -39,30 +40,30 @@ pub enum Node<'a> {
...
 				}
-				Node::Branch(nodes, if r.at(16).is_empty() { None } else { Some(r.at(16).data()) })
+				Ok(Node::Branch(nodes, if r.at(16)?.is_empty() { None } else { Some(r.at(16)?.data()?) }))
 			},
 			// an empty branch index.
-			Prototype::Data(0) => Node::Empty,
+			Prototype::Data(0) => Ok(Node::Empty),
 			// something went wrong.
-			_ => panic!("Rlp is not valid.")
+			_ => Err(DecoderError::Custom("Rlp is not valid."))
```
debris: maybe use `bail!` instead [docs](https://docs.rs/error-chain/0.11.0/error_chain/macro.bail.html)

util/patricia_trie/src/triedb.rs
```diff
@@ -97,7 +96,10 @@ impl<'db> TrieDB<'db> {
 			Node::Extension(ref slice, ref item) => {
 				write!(f, "'{:?} ", slice)?;
 				if let Ok(node) = self.get_raw_or_lookup(&*item) {
-					self.fmt_all(Node::decoded(&node), f, deepness)?;
+					match Node::decoded(&node) {
+						Ok(n) => self.fmt_all(n, f, deepness)?,
+						Err(err) => writeln!(f, "ERROR decoding node extension Rlp: {}", err)?,
```
debris: unrelated:
instead of `writeln`, we should use [debug_struct](https://doc.rust-lang.org/1.5.0/std/fmt/struct.Formatter.html#method.debug_struct) which takes into an account an [alternate](https://doc.rust-lang.org/1.5.0/std/fmt/struct.Formatter.html#method.alternate) flag of the formatter which allows [pretty-printing](https://doc.rust-lang.org/std/fmt/trait.Debug.html#examples)

util/patricia_trie/src/node.rs
ascjones: Done.

util/patricia_trie/src/node.rs
```diff
@@ -63,7 +63,7 @@ impl<'a> Node<'a> {
 			// an empty branch index.
 			Prototype::Data(0) => Ok(Node::Empty),
 			// something went wrong.
-			_ => Err(DecoderError::Custom("Rlp is not valid."))
+			_ => bail!("Rlp is not valid.")
```
debris: ah, sorry. I thought that this Error is already an error from error_chain. Wouldn't ask for this change if I knew it was not.

debris:
I reverted this last commit. Sorry, my suggestion was wrong. I'll merge the pr once it's green

PR #8173, opened 2018-03-21
SecretStore: get rid of engine.signer dependency

svyatonik:
Previously it was assumed that in PoA networks KeyServers will only be running on authority nodes && both `engine.signer` and `secretstore.self_secret` will be the same. This is PR to get rid of this requirement && that's how KS will be initially deployed on Kovan (not on authority nodes).


svyatonik:
Also fixed 2 small issues, found while testing:
1) 1-of-1 generation session was completed only after timeout
2) illegal (obsolete after new version of contracts) condition in `Retry` task processing

secret_store/src/trusted_client.rs
```diff
@@ -54,4 +65,25 @@ impl TrustedClient {
 	pub fn get_untrusted(&self) -> Option<Arc<Client>> {
 		self.client.upgrade()
 	}
+
+	/// Transact contract.
+	pub fn transact_contract(&self, contract: Address, tx_data: Bytes) -> Result<(), Error> {
+		let client = self.client.upgrade().ok_or(Error::Internal("cannot submit tx when client is offline".into()))?;
+		let miner = self.miner.upgrade().ok_or(Error::Internal("cannot submit tx when miner is offline".into()))?;
```
rphmeier: `ok_or_else` would prevent unconditional allocations of the errors messages.

PR #8176, opened 2018-03-21
Avoid allocations when computing triehash.

tomusdrw:
Analyzed size of the vectors before doing the change. `as_nibbles` is used for indices and the sizes are only 2 (mostly),4,6. `hex` is always half of it.


debris:
have you benched that? :D

Issue #6507, opened 2017-09-12
Cannot run Rust tests on Parity Gitlab in the specific scenario


grbIzl:
Please look through https://github.com/paritytech/parity/pull/6312
1) I've made a commit with Rust and Js code b316852. The checks failed because of syntax error in Js code
2) I've fixed syntax error in js and made new commit b18d555. This time it contained only Js code.
3) I see in logs of Gitlab jobs for the last commit "Skipping Rust tests since no Rust files modified." message. And all checks are green now.

Actual results:
Rust tests were not run after the first commit and they were not run after the second one. All checks are green. 

Expected results:
I need to run Rust tests for the first commit (of for changes made in terms of the first commit) somehow.

rphmeier:
cc @paritytech/ci

PR #8180, opened 2018-03-21
Updated jsonrpc to point to the 1.11 branch

kirushik:


debris:

PR #8181, opened 2018-03-21
Updated jsonrpc to include latest backports (beta)

kirushik:


debris:

twittner:
@tomusdrw: I meant to make another push to update hyper to 0.11.24 (local compilation always takes a long time on my machine). This way the same hyper version would be used on all three branches. Do you want me to open another PR for that?

PR #8182, opened 2018-03-21
Updated jsonrpc to include latest backports (1.9)

kirushik:


debris:

niklasad1:
But this error looks actually like there are multiple of versions of `jsonrpc`:

```bash
Compiling jsonrpc-http-server v8.0.0 (https://github.com/paritytech/jsonrpc.git?branch=parity-1.9#e23a96cb)
```

Have you checked that all references to ```jsonrpc``` are updated correctly?

twittner:
@niklasad1: If I am not mistaken, there are actually some dependency updates missing, e.g. hyper version >= 0.11.20. I am in the process of getting those updates in as well. Cf. #8180, #8181.

PR #8200, opened 2018-03-23
fixed ethcore tx_filter

debris:
- `TransactionFilter` was never added as `ChainNotify`
- therefore cache was never cleared
- ~this could lead to incorrect permissions being granted~

a simplest solution is to ~remove a cache at all, cause it would be cleared on every block anyway~ use `lru_cache` instead of manually cleaning the cache

### DIFF
--- a/ethcore/src/tx_filter.rs
+++ b/ethcore/src/tx_filter.rs
@@ -16,11 +16,10 @@
 
 //! Smart contract based transaction filter.
 
-use std::collections::HashMap;
-use std::collections::hash_map::Entry;
 use ethereum_types::{H256, Address};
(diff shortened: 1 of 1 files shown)


debris:
> This is pretty hot code during transaction import (note we are checking EVERY transaction), imho cache would make sense here

You are right. I looked at the code again and actually not cleaning the cache does not lead to incorrect behaviour. Cache just becomes useless.

>  Why not add it to ChainNotify?

I want to get rid of `ChainNotify`

I updated the pr. Replaced `HashMap` with `LruCache`

debris:
cc @andresilva @tomusdrw

Issue #6138, opened 2017-07-24
Whisper CLI


rphmeier:
Whisper-only binary. Should be in `whisper/cli`.
  - DevP2P configuration 
  - RPC server kinds 
  - ports 
  - whisper resident pool size
  - should have its own `Metadata` type for pub-sub.
  - Should only serve whisper RPCs.

rphmeier:
I would say this is `Easy` in difficulty, but `Substantial` in the amount of work. I can mentor

5chdn:
:-)

axelchalon:
I'd be interested in working on this!

niklasad1:
@rphmeier @axelchalon 
What is the status on this? 
I can take over this if you want

axelchalon:
I had started some work on this a while back (https://github.com/axelchalon/parity/tree/whisper-v6-cli for what it's worth). I think you can take it over.

PR #8201, opened 2018-03-23
Whisper cli

niklasad1:
Hey,

This is an attempt to close [#6138](https://github.com/paritytech/parity/issues/6138)

EDIT:
I removed the tests for because it more or less requires duplication of the existing tests in ```rpc``` and ```whisper``` because these are private I can't just use them directly. 

It's alot of work re-implement (copy-paste) them so I don't think it is worth the effort. But maybe a couple of CLI arguments tests and some `rough` test that jsonrpc at least works!

Thoughts?


5chdn:
Nice. /cc @gballet FYI

Needs docs once merged.

whisper/cli/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+Whisper CLI.
+	Copyright 2017 Parity Technologies (UK) Ltd
+
+Usage:
+	whisper [options]
+	whisper [-h | --help]
+
+Options:
+	--whisper-pool-size SIZE       Specify Whisper pool size [default: 10].
+	-p, --port PORT                Specify which port to use [default: 8545].
```
rphmeier: maybe clarify that this is the RPC port

rphmeier:
except for the small grumble, LGTM

rphmeier:
I think we should use proper error handling; this failure is likely due to other bugs in the whisper, network, or RPC implementation and we should fix those rather than working around them.

niklasad1:
@rphmeier 

I think it is good to go now, but I had to introduce an additional variable to `cancel/join` the `Decryption Worker Thread` because the `mspc` sends `Fn()` closure which I can't determine whether it is an actual message or to cancel the thread.

Alternately we could just skip joining threads and terminate!

rphmeier:
> Alternately we could just skip joining threads and terminate!

Yep, also an option. The way it was before it would deadlock.

PR #8204, opened 2018-03-23
Allow unsafe js eval for Parity Wallet (v1)

tomusdrw:


amaurymartiny:
Dapp-develop is still taking some time, so I'll merge this for the 1.10.1 patch release.

PR #8209, opened 2018-03-24
Bump wasmi and parity-wasm versions

NikVolf:
they contain minor correctness fixes for degenerate cases found by fuzzer


ethcore/wasm/run/src/runner.rs
```diff
@@ -130,7 +130,7 @@ pub fn run_fixture(fixture: &Fixture) -> Vec<Fail> {
 		Err(e) => { return Fail::load(e); },
 	};
 
-	let mut ext = FakeExt::new();
+	let mut ext = FakeExt::new().with_wasm();
```
NikVolf: it's just a tool (separate binary) update to use default schedule

PR #8231, opened 2018-03-27
node-filter does not use ChainNotify

debris:


tomusdrw:
@andresilva look for `LruCache::new(4096)` in the constructor. It can include stalled entries, but it's not really that much of an issue IMHO (just potentially wasted memory).

andresilva:
Derp, I should have realised it with the name `LruCache`. There's no need to care about stale entries if it's bounded. @debris there's a test that needs fixing (removing a call to `clear_cache`).

debris:
fixed the test and unfortunately removed 1 assertion from it as it is quite complicated to make it work without writing a lot of boilerplate code

created an issue for that

PR #8233, opened 2018-03-27
Replace all Rlp usages with UntrustedRlp except for ethcore views

ascjones:
Part of #8033, replacing Rlp with UntrustedRlp.

Now everywhere (except ethcore views) are using `UntrustedRlp` and handling the `Result<_,DecoderError>`. 

This paves the way for the next phase, which would be to replace Rlp with UntrustedRlp in the ethcore views. Once that is done we can remove the legacy Rlp implementation.


ethcore/light/src/client/header_chain.rs
```diff
@@ -74,6 +74,38 @@ pub struct BlockDescriptor {
 	pub total_difficulty: U256,
 }
 
+// best block data
+struct BestAndLatest {
```
debris: handwritten implementations of `Encodable` and `Decodable` can be replaced with 

```
#[derive(RlpEncodable, RlpDecodable)]
struct {
}
```

ethcore/src/client/test_client.rs
```diff
@@ -469,7 +469,7 @@ impl ChainInfo for TestBlockChainClient {
 impl BlockInfo for TestBlockChainClient {
 	fn block_header(&self, id: BlockId) -> Option<encoded::Header> {
 		self.block_hash(id)
-			.and_then(|hash| self.blocks.read().get(&hash).map(|r| Rlp::new(r).at(0).as_raw().to_vec()))
+			.and_then(|hash| self.blocks.read().get(&hash).map(|r| UntrustedRlp::new(r).at(0).unwrap().as_raw().to_vec()))
```
tomusdrw: We fought hard to get rid of `unwraps()` in the code base and this PR intorudces it back, can we please replace them with `expects()`?

ethcore/light/src/client/header_chain.rs
ascjones: Of course! Fixed.

ethcore/src/client/test_client.rs
ascjones: Yes sorry that unwrap snuck in. Removed.

tomusdrw: There seems to be way more than that one. It's fine in tests code, but would avoid it in regular code.

ascjones: I've removed the unwraps I added in this PR. There are plenty in there still from before - like you say probably okay for tests. Which is why I was lazy in this case and just added them. Broken windows and all!

sync/src/blocks.rs
```diff
@@ -293,9 +294,9 @@ impl BlockCollection {
 				let mut block_rlp = RlpStream::new_list(3);
 				block_rlp.append_raw(&block.header, 1);
```
tomusdrw: `new_from_header_and_body()`?

ascjones: Good idea. See latest commit.

PR #8236, opened 2018-03-27
Wasm runner fix

fckt:
There was a runtime error
thread 'main' panicked at 'Wasm schedule expected to exist while checking wasm contract. Misconfigured client?', src/libcore/option.rs:891:5

### DIFF
--- a/ethcore/wasm/run/src/fixture.rs
+++ b/ethcore/wasm/run/src/fixture.rs
@@ -41,8 +41,8 @@ pub struct Fixture {
 
 #[derive(Deserialize, Debug)]
 pub struct StorageEntry {
-	pub key: Uint,
-	pub value: Uint,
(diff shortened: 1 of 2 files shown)


ethcore/wasm/run/src/fixture.rs
```diff
@@ -41,8 +41,8 @@ pub struct Fixture {
 
 #[derive(Deserialize, Debug)]
 pub struct StorageEntry {
-	pub key: Uint,
```
NikVolf: why? uint provide more flexibility, you don't need to write `0x0000000000000000..00000000` for `0x`

fckt: but with explicit H256 you’ll never mess up with endianess on converting to bytes

PR #8242, opened 2018-03-28
Update musicoin spec in line with gmc v2.6.2

Varunram:
ref. @5chdn


Varunram:
[clabot:check]

5chdn:
what exactly is GMC 2.6.2? can't sync beyond the hf with either of the specs.

PR #8249, opened 2018-03-28
Addressing grumbles in PrivateTransactions

tomusdrw:
Contains a bunch of TODOs that could be extracted as separate issues that we can work on in future.

### DIFF
--- a/ethcore/private-tx/src/messages.rs
+++ b/ethcore/private-tx/src/messages.rs
@@ -57,6 +57,8 @@ impl SignedPrivateTransaction {
 			private_transaction_hash: private_transaction_hash,
 			r: sig.r().into(),
 			s: sig.s().into(),
+			// TODO [ToDr] Somehow share more code with ethcore_transaction
+			// especially on signing stuff.
 			v: sig.v() as u64 + if let Some(n) = chain_id { 35 + n * 2 } else { 27 },
 		}
 	}
--- a/rpc/src/v1/impls/private.rs
+++ b/rpc/src/v1/impls/private.rs
@@ -45,9 +45,9 @@ impl PrivateClient {
 		}
 	}
 
-	fn unwrap_manager(&self) -> Result<Arc<PrivateTransactionManager>, Error> {
+	fn unwrap_manager(&self) -> Result<&PrivateTransactionManager, Error> {
 		match self.private {
-			Some(ref arc) => Ok(arc.clone()),
+			Some(ref arc) => Ok(&**arc),
 			None => Err(errors::public_unsupported(None)),
 		}
 	}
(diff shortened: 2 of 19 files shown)


rpc/src/v1/impls/private.rs
```diff
@@ -90,7 +90,12 @@ impl Private for PrivateClient {
 			data: Some(transaction.data.into()),
 			condition: None,
 		};
+		// TODO [ToDr] This is invalid?
```
grbIzl: Addressed in https://github.com/paritytech/parity/pull/6422

rpc/src/v1/impls/private.rs
```diff
@@ -90,7 +90,12 @@ impl Private for PrivateClient {
...
+		// private-tx/src/lib.rs#L532 uses engine to figure out the scheme.
 		let (contract_address, _) = contract_address(CreateContractAddress::FromSenderAndNonce, &signed_transaction.sender().clone(), &transaction.nonce.into(), &[]);
+		// TODO [ToDr] Figure out this:
+		// Why this type contains the entire transaction (request)?
+		// Why Request and not entire SignedTransaction?
```
grbIzl: I assume, that the user will sign the transaction by himself

ethcore/private-tx/src/messages.rs
```diff
@@ -57,6 +57,8 @@ impl SignedPrivateTransaction {
 			private_transaction_hash: private_transaction_hash,
 			r: sig.r().into(),
 			s: sig.s().into(),
+			// TODO [ToDr] Somehow share more code with ethcore_transaction
+			// especially on signing stuff.
```
grbIzl: Addressed in https://github.com/paritytech/parity/pull/6422

rpc/src/v1/impls/private.rs
tomusdrw: But that's the output type, right? We already had signed transaction after `client.public_creation_transaction` and the we disregard the signature and return just the request? So why sign it in the first place?

grbIzl: So may be I don't understand your question. Let me show this process, basing on our test (see more comments and explanations inline as additional comments):

trace!("Creating private contract");
// Create an input for deployment transaction. The input basically contains private contract in data
let private_contract_test = "606060...79d2080029".from_hex().unwrap();
let mut private_create_tx = Transaction::default();
private_create_tx.action = Action::Create;
private_create_tx.data = private_contract_test;
let private_create_tx_signed = private_create_tx.sign(&key1.secret(), None);
// The input is created. Pass it and additional parameters to API in order to create a deployment transaction for the contract from the input
let validators = vec![key3.address(), key4.address()];
let public_tx = pm.public_creation_transaction(BlockId::Latest, &private_create_tx_signed, &validators, 0.into()).unwrap();
// Deployment transaction returned, in order to deploy it in the chain we need to sign it as well
let public_tx = public_tx.sign(&key1.secret(), chain_id);
trace!("Transaction created. Pushing block");
push_block_with_transactions(&client, &[public_tx]);

5chdn:
How to proceed? Some conflicts in ethcore and rpc.

PR #8251, opened 2018-03-28
Upgrader `remove raw unwrap` and bump semver

niklasad1:
If this is ok then we need to bump ```rustc_version crate ``` to 0.2.2 it depends on semver 0.9.0 instead of 0.6.0

```bash
│           │   [build-dependencies]
│           │   └── rustc_version v0.2.1
│           │       [dependencies]
│           │       └── semver v0.6.0
```


parity/upgrade.rs
```diff
@@ -32,6 +32,13 @@ pub enum Error {
 	CannotCreateConfigPath,
 	CannotWriteVersionFile,
 	CannotUpdateVersionFile,
+	Semver(SemVerError),
```
andresilva: `SemVer`?

debris:
neess merge with master ;)

PR #8252, opened 2018-03-28
Add missing license header for runtime.rs

pepyakin:


ethcore/wasm/src/runtime.rs
```diff
@@ -1,3 +1,19 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
5chdn: 2018? :-)

PR #8258, opened 2018-03-29
Bump ethabi & ethereum-types.

tomusdrw:
Related https://github.com/paritytech/ethabi/pull/85#issuecomment-377168526


util/network-devp2p/src/host.rs
```diff
@@ -1099,7 +1099,7 @@ fn save_key(path: &Path, key: &Secret) {
 	if let Err(e) = restrict_permissions_owner(path, true, false) {
 		warn!(target: "network", "Failed to modify permissions of the file ({})", e);
 	}
-	if let Err(e) = file.write(&key.hex().into_bytes()) {
+	if let Err(e) = file.write(&key.hex().into_bytes()[2..]) {
```
andresilva: Was this needed or was changed just for "style"? I.e. will `Secret::from_str` work with and without the `0x` prefix?

tomusdrw: That was required, test caught it.

PR #8261, opened 2018-03-29
Fix warning about --no-hardcoded-sync being printed when it shouldn't

tomaka:
cc @tomusdrw

### DIFF
--- a/parity/run.rs
+++ b/parity/run.rs
@@ -575,7 +575,7 @@ pub fn execute_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>)
 	}
 
 	// display warning if using --no-hardcoded-sync
-	if !cmd.no_hardcoded_sync {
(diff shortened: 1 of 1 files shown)


parity/run.rs
```diff
@@ -575,7 +575,7 @@ pub fn execute_impl(cmd: RunCmd, can_restart: bool, logger: Arc<RotatingLogger>)
 	}
 
 	// display warning if using --no-hardcoded-sync
-	if !cmd.no_hardcoded_sync {
+	if cmd.no_hardcoded_sync {
```
5chdn: does this show this warning if you actually run --light?

tomusdrw: Yeah, this is in a function that sets up the full client only.

5chdn: Cool

tomaka: No, the `execute_impl` function is run only when we start a full node. This `if` is never reached on the light client.

PR #8262, opened 2018-03-29
Enable UI by default, but only display deprecation notice

tomusdrw:
Looks a bit shit, would be good to present some more info there. Feel free to edit the HTML file.

screenshot from 2018-03-29 14-30-18


Tbaut:
Thanks a lot @tomusdrw 
Does this page get shown only for 127.0.0.1:8180/#/v1 or also for 127.0.0.1:8180/ ?
I would also generally talk about the "Browser-based User Interface" rather than "Parity UI" because this is the name of the new app. 
I'd make the change depending on everyone's feedback.

dapps/ui-deprecation/build/index.html
```diff
@@ -107,7 +107,7 @@
 		<div class="parity-navbar">
 		</div>
 		<div class="parity-box">
-			<h1>Parity Wallet is deprecated.</h1>
+			<h1>Parity browser UI is deprecated.</h1>
```
tomusdrw: I thought the product name is actually `Parity Wallet`, no?

5chdn: We are moving in circles here :rofl: 

https://github.com/paritytech/parity/pull/8164#issuecomment-374622582

tomusdrw: Yeah, afaiu we are moving away from "Parity Wallet" (browser wallet) in favour of standalone "Parity UI" (electron app).

tomusdrw: It is, Parity will not support any dapps, UI, wallets whatsoever. The distinction between "Parity Wallet" being a dapp was not really communicated I don't recall changing the names when UI2 has been rolled out. That said I'm not really requesting a change here, just discussing if that's a correct "story" we are telling here.

Anyway, looks good to me :+1: :)

folsen:
@tomusdrw @5chdn @Tbaut this text is much clearer IMO, see my commit

5chdn:
Thanks.

PR #8276, opened 2018-03-30
Cleanup Ellaism bootnodes

ellaismer:
Removed several bootnodes known not to be running anymore. Added two new ones.


ellaismer:
@5chdn That's weird. Most of those bootnodes are actual peers, and I'm able to use them to successfully sync the network. It might be that some peers are full so they reject new peers. And how did you test whether they're working?

5chdn:
```bash
$ parity --chain ./ellaism.json --reserved-peers ./ellaism.peer --reserved-only                                                                                                                               
```

with [ellaism.json](https://github.com/ellaismer/parity/blob/bdc159bd23737493b674ad830782f0fb3fe20bde/ethcore/res/ethereum/ellaism.json) and where `ellaism.peer` contains the bootnodes

5chdn:
maybe you should increase the max-peers of the bootnodes?

ellaismer:
I checked bootnodes that I have control (the first three). None of them reached max-peers.

There're some nodes there I need to run an older Parity version (v1.6.10), would that be an issue?

5chdn:
It's working well if I drop the reserved-only option, so I'm fine with merging this.

PR #8278, opened 2018-03-30
use constant durations

### LINKED ISSUE #8260: Replace the various constants indicating durations with `Duration` objects
Parity's codebase contains this kind of code at several places:

```rust
// at the top of the file
const FOO_MILLISECONDS: u64 = 100;

// and then in some function:
blablabla(Duration::from_millis(FOO_MILLISECONDS));
```

But Rust 1.25 that was released yesterday allows creating constant `Duration`s, which means that we can do this instead:

```rust
// at the top of the file
const FOO: Duration = Duration::from_millis(100);

// and then in some function:
blablabla(FOO);
```

This is arguably better, for both the small performance gain and for strong typing.

rleungx:
Closes #8260.

### DIFF
--- a/secret_store/src/key_server_cluster/cluster.rs
+++ b/secret_store/src/key_server_cluster/cluster.rs
@@ -1104,6 +1104,8 @@ pub mod tests {
 	use key_server_cluster::key_version_negotiation_session::{SessionImpl as KeyVersionNegotiationSession,
 		IsolatedSessionTransport as KeyVersionNegotiationSessionTransport};
 
+	const TIMEOUT: Duration = Duration::from_millis(300);
+
 	#[derive(Default)]
 	pub struct DummyClusterClient;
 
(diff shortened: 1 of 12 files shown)


secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1388,11 +1388,11 @@ pub mod tests {
 		let mut core = Core::new().unwrap();
 		let clusters = make_clusters(&core, 6028, 3);
 		run_clusters(&clusters);
-		loop_until(&mut core, time::Duration::from_millis(300), || clusters.iter().all(all_connections_established));
+		loop_until(&mut core, Duration::from_millis(300), || clusters.iter().all(all_connections_established));
```
tomusdrw: Perhaps those Durations could be moved to constants as well?

rleungx: Any naming suggestion? Something like `TIMEOUT`?

tomaka:
In my opinion the constants should be renamed as well, for example `TIMEOUT_SECS` renamed to `TIMEOUT`, `UPDATE_INTERVAL_MS` to `UPDATE_INTERVAL`, and so on.

PR #8280, opened 2018-03-31
eth_uninstallFilter should return false for non-existent filter

### LINKED ISSUE #7783: eth_uninstallFilter returns true for any valid parameter

> I'm running:
>
> - **Which Parity version?**: Parity/v1.8.7-stable-e322fd8-20180124/x86_64-linux-gnu/rustc1.23.0
> - **How installed?**: via  binaries
> - **Which network are you connected to?**: private chain

The method eth_uninstallFilter returns "true" for any valid parameter, regardless if the filter was successfully uninstalled, which is inconsistent with the documented behavior: 

from [doc](https://paritytech.github.io/wiki/JSONRPC-eth-module.html#eth_uninstallfilter)
>Returns
>Boolean - true if the filter was successfully uninstalled, otherwise false.

Examples:

For existing filter:

```
curl --data '{"method":"eth_uninstallFilter","params":["0x3"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","result":true,"id":1}
```

Re-trying on uninstalled filter, would expect to see `false` return value:

```
curl --data '{"method":"eth_uninstallFilter","params":["0x3"],"id":2,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","result":true,"id":2}
```

On a non-existent filter:

```
curl --data '{"method":"eth_uninstallFilter","params":["0xdeadbeefdeadbeef"],"id":3,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","result":true,"id":3}
```

sorpaas:
Fixes #7783

### DIFF
--- a/rpc/src/v1/helpers/poll_manager.rs
+++ b/rpc/src/v1/helpers/poll_manager.rs
@@ -75,10 +75,7 @@ impl<F, T> PollManager<F, T> where T: Timer {
 
 	/// Removes poll info.
 	pub fn remove_poll(&mut self, id: &PollId) -> bool {
-		match self.polls.remove(id) {
-			Some(_) => true,
-			None => false,
-		}
+		self.polls.remove(id).is_some()
 	}
 }
 
--- a/rpc/src/v1/helpers/poll_manager.rs
+++ b/rpc/src/v1/helpers/poll_manager.rs
@@ -74,8 +74,11 @@ impl<F, T> PollManager<F, T> where T: Timer {
 	}
 
 	/// Removes poll info.
-	pub fn remove_poll(&mut self, id: &PollId) {
-		self.polls.remove(id);
+	pub fn remove_poll(&mut self, id: &PollId) -> bool {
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/poll_manager.rs
```diff
@@ -74,8 +74,11 @@ impl<F, T> PollManager<F, T> where T: Timer {
 	}
 
 	/// Removes poll info.
-	pub fn remove_poll(&mut self, id: &PollId) {
-		self.polls.remove(id);
+	pub fn remove_poll(&mut self, id: &PollId) -> bool {
+		match self.polls.remove(id) {
```
tomusdrw: `self.polls.remove(id).is_some()` should suffice.

PR #8281, opened 2018-03-31
Return null number for pending block in eth_getBlockByNumber

### LINKED ISSUE #8028: JSON-RPC: Wrong specification or implementation?
1.
According to documentation of [eth_getBlockByNumber](https://github.com/paritytech/wiki/blob/master/JSONRPC-eth-module.md#eth_getblockbynumber) when `pending` block requested result should have null in `number` field.

> number: QUANTITY - the block number. null when its pending block.

But in the code of Parity I've found that it always [returns number of next block](https://github.com/paritytech/parity/blob/ec7282d05c0639064e0b5ea8cfbd08a0f6af4375/ethcore/src/block.rs#L261) :

```
impl<'x> OpenBlock<'x> {
	/// Create a new `OpenBlock` ready for transaction pushing.
	...
        let number = parent.number() + 1;
        ...

```

Could you please clarify can `number` has `null` value or not ?

2.

Another question is about `miner` field. Could it has `null` value or not ?

sorpaas:
Fixes #8028

Check whether it's fetching the pending block. If so, write the block number as null in JSONRPC.


rpc/src/v1/impls/eth.rs
```diff
@@ -230,7 +230,10 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
 						state_root: view.state_root().into(),
 						transactions_root: view.transactions_root().into(),
 						receipts_root: view.receipts_root().into(),
-						number: Some(view.number().into()),
```
tomusdrw: According to the same spec `hash` should be null as well.

rpc/src/v1/impls/eth.rs
```diff
@@ -180,9 +180,9 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
 	fn rich_block(&self, id: BlockNumberOrId, include_txs: bool) -> Result<Option<RichBlock>> {
 		let client = &self.client;
 
-		let client_query = |id| (client.block(id), client.block_total_difficulty(id), client.block_extra_info(id));
+		let client_query = |id, is_pending| (client.block(id), client.block_total_difficulty(id), client.block_extra_info(id), is_pending);
```
tomusdrw: Just inline `false` here, no point to pass `is_pending`

sorpaas: Ooops right.

sorpaas:
According to the spec. `hash`, `number`, `nonce` and `logsBloom` should be null for pending block. If I get this right, pending blocks are not sealed so they don't have `nonce` or `mixHash` already. And `hash`, `number` and `logsBloom` are set to null in this PR.

PR #8118, opened 2018-03-14
Disable link-dead-code in code coverage CI build

andresilva:
- https://gitlab.parity.io/parity/parity/-/jobs/81532
- https://gitlab.parity.io/parity/parity/-/jobs/81590

The code coverage build has been failing to compile `futures-timer` and it seems to be related to https://github.com/rust-lang/rust/issues/45629 (https://github.com/codecov/example-rust/pull/10). Let's see if this fixes the build.

PR #8287, opened 2018-04-02
Revert "ci: disable link-dead-code in coverage build (#8118)"

andresilva:
This reverts commit 4d1cb01da0ceff3b025c0f030da7aae86441206a.

https://github.com/paritytech/parity/pull/8118#event-1522171535

This will break the coverage build but at least it won't report an erroneous coverage percentage.

### DIFF
--- a/scripts/cov.sh
+++ b/scripts/cov.sh
@@ -12,7 +12,7 @@
 ### Running coverage
 
 set -x
-cargo test --all --exclude evmjit --no-run || exit $?
+RUSTFLAGS="-C link-dead-code" cargo test --all --exclude evmjit --no-run || exit $?
 KCOV_TARGET="target/cov"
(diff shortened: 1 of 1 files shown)


5chdn:
What are we doing with this now? Hoping it will be fixed in Rust?

andresilva:
There are two crates that break the compiler with `link-dead-code` (only one shows up in the build logs):
- `futures-timer` - I plan on refactoring `parity-reactor` to use the new tokio `Runtime` (https://tokio.rs/blog/2018-03-tokio-runtime/). This will give us more efficient global timers and we can then drop the `futures-timer` dependency.
- `patricia-trie` - I tested and this is fixed with Rust 1.26 (which should be released in a couple of weeks).

PR #8294, opened 2018-04-03
dapps-fetcher: calculate keccak in-flight while reading the response

sorpaas:
This fixes a TODO item in dapps-fetcher.

A new function, called `keccak_buffer_and_write` is added. This allows attaching another BufWrite. When reading the input, it writes it to BufWrite as well. The original function `keccak_buffer` is then just setting the parameter to a Sink.


sorpaas:
Renamed to `keccak_pipe`. The test currently fails. I'm figuring out what might be wrong..

PR #8300, opened 2018-04-04
Remove InvalidParentHash in favor of assert!

sorpaas:
As the TODO item states (and I manually checked), `InvalidParentHash` error if happens, indicate a codebase bug. (Only used in `verify_parent` and parent is always fetched locally.)

So here we remove InvalidParentHash and use an assert! instead.


sorpaas:
I need to look into this. The test indicates that there're cases where InvalidParentHash is probably expected behavior...

sorpaas:
Okay I think this is fixed. I should have asserted a true case, but instead asserted false.

Issue #8296, opened 2018-04-03
Multiple Rayon dependencies


0x7CFE:
Current dependency tree contains two Rayon versions: `v0.8.2` and `v0.9.0`. 
This is suboptimal and should be refactored eventually.

Actual for `master` revision 9f775a7673bb4e3d6197703d248b0c120eee3fa9.

PR #8302, opened 2018-04-04
Ethcore now uses rayon 0.9 as a dependency (#8296)

0x7CFE:
After this change, runtime dependencies would contain only Rayon `0.9`. Affects https://github.com/paritytech/parity/issues/8296.

P.S.: Maybe we should move right to the `1.0` instead?


debris:
> P.S.: Maybe we should move right to the `1.0` instead?

Let's do it! And let's bump `cc` to version `1.0.9`. It also uses `rayon 1.0`

0x7CFE:
> Let's do it! And let's bump cc to version 1.0.9. It also uses rayon 1.0

Done in https://github.com/paritytech/parity/pull/8304

PR #8305, opened 2018-04-04
Decouple timestamp open-block-assignment/verification to Engine

### LINKED ISSUE #7694: Block timestamp drift on dev chain with --reseal-min-period 0

> I'm running:
>
> - **Which Parity version?**: 1.8.5
> - **How installed?**: Docker
> - **Which network are you connected to?**: development chain

Hello,
today we noticed very strange behavior (at least it's really suprising to me) that when we're using `--reseal-min-period 0` (with other small values as well) flag while running our parity node with development chain, block.timestamp is quickly drifting to the future e.g. after running some transactions, it's 15 mins in the future compared to the time on dev machine and container running Parity.

Is that an expected behavior or it's an unexpected side-effect of forcing really small `reseal-min-period` ? 

Regards,
Piotr

sorpaas:
Fixes #7694

The bug happens because when Parity constructs a new open block, the timestamp will be set at least one second ahead of parent timestamp. For most of the time, this works fine, and most Engine rules also require this. However, because dev chain seals a block for every transaction, it becomes a problem if one tries to send more than one transaction per second.

This PR adds two new predefined functions in Engine trait, `open_block_header_timestamp` and `is_timestamp_valid`. All other engines are unchanged except `InstantSeal`, where `open_block_header_timestamp` would not do "plus one", and `is_timestamp_valid` would accept it if parent timestamp equals to current one. Those replaces hard coded rules in `verification` mod and `Header::set_timestamp_now`.

This also makes the function `set_timestamp_now` unused. And I read a comment saying this function "kind of sucks"... So I removed it.

### DIFF
--- a/ethcore/src/block.rs
+++ b/ethcore/src/block.rs
@@ -267,7 +267,7 @@ impl<'x> OpenBlock<'x> {
 		r.block.header.set_parent_hash(parent.hash());
 		r.block.header.set_number(number);
 		r.block.header.set_author(author);
-		r.block.header.set_timestamp_now(parent.timestamp());
+		r.block.header.set_timestamp(engine.open_block_header_timestamp(parent.timestamp()));
 		r.block.header.set_extra_data(extra_data);
 
 		let gas_floor_target = cmp::max(gas_range_target.0, engine.params().min_gas_limit);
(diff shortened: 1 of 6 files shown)


kretz1904BADFISHY:
Who is this person I'm Robert J. Kreczmer owner of Idenity this guy's createing hacks on. I've have been under attack now 11MTHS.

kretz1904BADFISHY:
XXVIA ROBERT KRECZMER
ROBERT KRECZMER AIS
ALL ROBERT KRECZMERS
JJROPPTS

PR #8306, opened 2018-04-04
Remove obsolete comment

kseo:


rphmeier:
why do you think this is obsolete?

kseo:
@rphmeier Oops, I checked only the full client and missed the light client. I will close the PR.

PR #8309, opened 2018-04-04
Read registry_address from block with REQUEST_CONFIRMATIONS_REQUIRED

sorpaas:
Use REQUEST_CONFIRMATIONS_REQUIRED instead of the latest block. Fix one TODO item in code comment.

### DIFF
--- a/secret_store/src/helpers.rs
+++ b/secret_store/src/helpers.rs
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
 // This file is part of Parity.
 
 // Parity is free software: you can redistribute it and/or modify
--- a/secret_store/src/helpers.rs
+++ b/secret_store/src/helpers.rs
@@ -1,0 +1,12 @@
+use ethcore::client::{Client, BlockChainClient, BlockId};
+use ethereum_types::H256;
+
+/// Number of confirmations required before request can be processed.
+pub const REQUEST_CONFIRMATIONS_REQUIRED: u64 = 3;
+
+/// Get hash of the last block with at least n confirmations.
+pub fn get_confirmed_block_hash(client: &Client, confirmations: u64) -> Option<H256> {
+	client.block_number(BlockId::Latest)
+		.map(|b| b.saturating_sub(confirmations))
+		.and_then(|b| client.block_hash(BlockId::Number(b)))
+}
--- a/secret_store/src/helpers.rs
+++ b/secret_store/src/helpers.rs
@@ -17,6 +17,7 @@
 use ethcore::client::{Client, BlockChainClient, BlockId};
 use ethereum_types::H256;
 
+// TODO: Instead of a constant, make this based on consensus finality.
 /// Number of confirmations required before request can be processed.
 pub const REQUEST_CONFIRMATIONS_REQUIRED: u64 = 3;
 
--- a/secret_store/src/helpers.rs
+++ b/secret_store/src/helpers.rs
@@ -1,3 +1,19 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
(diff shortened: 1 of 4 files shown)


secret_store/src/listener/service_contract.rs
```diff
@@ -234,16 +234,16 @@ impl OnChainServiceContract {
...
 			if let Some(client) = self.client.get() {
-				// update contract address from registry
-				let service_contract_addr = client.registry_address(self.name.clone(), BlockId::Latest).unwrap_or_default();
-				if self.data.read().contract_address != service_contract_addr {
-					trace!(target: "secretstore", "{}: installing {} service contract from address {}",
-						self.self_key_pair.public(), self.name, service_contract_addr);
-					self.data.write().contract_address = service_contract_addr;
+				if let Some(block_hash) = get_confirmed_block_hash(&*client, REQUEST_CONFIRMATIONS_REQUIRED) {
+					// update contract address from registry
+					let service_contract_addr = client.registry_address(self.name.clone(), BlockId::Hash(block_hash)).unwrap_or_default();
```
debris: I'm not sure whether it will work with not archive nodes. cc @tomusdrw

debris:
this should be also reviewed by @svyatonik

secret_store/src/listener/service_contract.rs
rphmeier: right, the state might be pruned `get_confirmed_block_hash` could return the oldest ancestor possible in that case, but then the call to `reigistry_address` will race with the pruning. We really need some way to block pruning while we inspect state.

rphmeier:
Using a constant number of confirmations is a little tenuous. Can we add a `TODO` to do this based on consensus finality where possible instead?

sorpaas:
@rphmeier There were originally a `TODO` item and I was trying to fix that. Looks like this becomes complicated if pruning is enabled, and I don't think I have a way to fix this...

I think the original `TODO` comment should be enough as a holder for anyone looks into this. So I'm going to close this PR if that's okay.

svyatonik:
@debris @rphmeier Thanks for pointing to this. This code should work only if we're synchronized (`!sync.status().is_syncing()`) and have a full trust (`chain_info().security_level().is_full()`) => I guess that's kind of a guarantee that we're dealing with the most up-to-date state (at least that's how updater works). Also, please correct me if I'm wrong (that's what I'm not 100% sure about), but pruning leaves some recent states and since we're dealing with block `latest_block - 3`, it's state is also not pruned (I believe that `recent_states.length > 3`). So this means that we have some time to work with this state (until it'll be pruned). So the problem would only arise if we're trying to deal with this state for too long. And in case if state is pruned - trying to read from contract at this state would lead to an error. Am I correct? If so, I wouldn't consider this as a serious problem && I'm for letting this PR merged-in.

@rphmeier I guess number of confirmation is an universal decision (and afair our UI and I guess most of DApps use the similar approach). Thanks for pointing that there's another way, though. @sorpaas could you please add TODO for this.

@sorpaas Thanks for helping with this TODO :) It is a little bigger than just fixing `service_contract.rs`, though :) And while changing the `acl_storage.rs` in the same way is arguable (and I wouldn't recommend to do this), changing the `client.registry_address` call in the `key_server_set.rs` is desirable. Could you please also change this in the similar way? Or just move original TODO to that file, instead or removing it?

sorpaas:
@svyatonik Okay. Will do!

rphmeier:
@svyatonik 

> trying to read from contract at this state would lead to an error

yeah. since it's only 3 blocks behind the head of the chain (minimum pruning we allow _for now_ is 8) this is probably not a problem, but it's an assumption that can be broken well outside of this module. Although if this is triggered while we're importing many blocks and not just at the head of the chain, there is a pretty bad race condition.

The finality stuff isn't useful to most DApps since they're on the mainnet, so their only choice is to rely on confirmations until CasperFFG. But where possible, it's more correct and safer to use finality.

secret_store/src/helpers.rs
```diff
@@ -0,0 +1,12 @@
+use ethcore::client::{Client, BlockChainClient, BlockId};
```
svyatonik: Could you please add license preamble to the beginning of the file?

svyatonik:
Also please do not forget to add TODO about using finality when possible (see my big comment above). Thanks!

secret_store/src/helpers.rs
svyatonik: Or maybe just make this function a member of `TrustedClient`? Both have similar purpose.

svyatonik: Don't think so - guess it will be enough for new file only. Thanks!

sorpaas:
@svyatonik Sorry I missed that!

secret_store/src/helpers.rs
sorpaas: Okay I'll send a separate PR to change license preamble.

sorpaas: Okay. Got it!

PR #8313, opened 2018-04-04
Bump `log` and `env_logger`

niklasad1:
But it introduces dependencies to multiple versions of `log` and `env_logger` because other crates such as `jsonrpc`, `hyper`, `pwasm-utils` et. al is using old versions of `log` or `env_logger`


tomusdrw:
We should bump it in the libraries as well, afair `log=0.3` doesn't produce any results if used with `env_logger=0.5`

ddorgan:
[clabot:check]

niklasad1:
@tomusdrw 

Ok, sounds reasonable to me.
Looks like most the dependencies can be solved just by bumping the version except `our own crates` + `tokio-proto` 

So this is blocked for now!

niklasad1:
This is really on ice, to bump this I need to bump `rustls` which in turn depends on `0.13.0-alpha` that will cause problems for us because we use our own fork of `ring`.

PR #8318, opened 2018-04-05
final private tx pr

debris:
I believe we can merge Private Transactions pr after this code is in


parity/run.rs
```diff
@@ -709,11 +710,15 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
...
 
 	service.add_notify(chain_notify.clone());
-	service.add_notify(private_tx_provider.clone());
 
-	// TODO [ToDr] PrivateTX should use separate notifications
-	// re-using ChainNotify for this is a bit abusive.
-	private_tx_provider.add_notify(chain_notify.clone());
+	// provider not added to a notification center is effectively disabled
+	// TODO [debris] refactor it later on
+	if cmd.private_tx_enabled {
```
debris: I know that it is not the cleanest way to do it, but it was the simplest solution for now.

PR #8326, opened 2018-04-06
Fix the JSONRPC API not running with the light client

tomaka:
Introduced by myself in #8284 unfortunately :-/


amaurymartiny:
Closes https://github.com/parity-js/shell/issues/42.

PR #8330, opened 2018-04-06
[WIP] add ed25519 signature verification precompile

oberstet:
See also:

* [EIP 665](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-665.md)
* [go-ethereum PR](https://github.com/ethereum/go-ethereum/pull/16453)
* [pyethereum PR](https://github.com/ethereum/pyethereum/pull/862)
* [parity PR](https://github.com/paritytech/parity/pull/8330)
* cpp-ethereum PR [TODO]

TODO: fill in test vectors in unit test

### DIFF
--- a/ethcore/src/builtin.rs
+++ b/ethcore/src/builtin.rs
@@ -29,6 +29,8 @@ use bytes::BytesRef;
 use ethkey::{Signature, recover as ec_recover};
 use ethjson;
 
+use crypto::ed25519;
+
 #[derive(Debug)]
 pub struct Error(pub &'static str);
 
@@ -220,6 +222,7 @@ fn ethereum_builtin(name: &str) -> Box<Impl> {
 		"alt_bn128_add" => Box::new(Bn128AddImpl) as Box<Impl>,
 		"alt_bn128_mul" => Box::new(Bn128MulImpl) as Box<Impl>,
 		"alt_bn128_pairing" => Box::new(Bn128PairingImpl) as Box<Impl>,
+		"ed25519_verify" => Box::new(Ed25519VerifyImpl) as Box<Impl>,
 		_ => panic!("invalid builtin name: {}", name),
 	}
 }
@@ -255,6 +258,9 @@ struct Bn128MulImpl;
 
 #[derive(Debug)]
 struct Bn128PairingImpl;
+
+#[derive(Debug)]
+struct Ed25519VerifyImpl;
 
 impl Impl for Identity {
 	fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


oberstet:
@5chdn yeah, definitely! well, I am willing to do more work, but I am stuck on getting some guidance / decision. as described here https://gitter.im/ethereum/AllCoreDevs?at=5ad2e7587c3a01610de35a31, the main options for the ed25519 library we could use are

* libhacl-c
* libsodium

I think the latter would be a robust, conservative and practical choice, and I would rewrite/expand the PRs (not only this here) - if there is support. If not, this is just a waste of time ..

oberstet:
@sorpaas I am a Rust newbie, don't know the community/ecosystem much, so take this with a grain of salt:

* the current PR is based on https://docs.rs/rust-crypto/0.2.36/crypto/ (https://github.com/DaGenix/rust-crypto), which seems widely used, but nevertheless not audited that much (https://internals.rust-lang.org/t/why-is-a-trusted-feature-complete-crypto-library-not-a-top-priority-for-the-rust-community/3125)
* the former implementation already is pure Rust (https://github.com/DaGenix/rust-crypto/blob/master/src/curve25519.rs) - some 5000 LOC ! and that is hard core math code .. mmh.
* HACL is _formally verified_ (also the transpiled C code) => there is no need for audit .. it is fine by definition. (I am trusting the researchers here .. I have looked a bit into HACL, F* etc .. the guys behind those things are cracks. also, HACL is going into Firefox etc)
* libsodium is probably the most widely used Ed25519 implementation, it is written/maintained by experts

The library you link: https://github.com/dalek-cryptography/ed25519-dalek#warnings => is using https://github.com/isislovecruft/curve25519-dalek for curce 25519 math - and that lib is quite new. and while Isis (https://github.com/isislovecruft) is certainly super cool dev, mmmh.

If you ask me right away, what to go for, I would only consider libsodium and HACL. Eg if we fuck this up, and it is because we have chosen some "random code from the Web", users will be angry. Righly so. In that light, I could only continue sleeping well choosing sth that can be defended - the choice (if there is an issue .. which always can happen).

oberstet:
Rgd compile switch to swap the actual implementation being used: could do, question is why? for "regular" use, implementations will behave identical .. the problem is the corner cases: does the elliptic curve math implementation fall apart for certain corner cases? _testing_ this using a handful of test vectors is _not_ enough. and because that is tricky, I'd rather rely on ideally _one_ underlying implementation for _all_ the Eth clients. From how I understood the discussion that happened on Gitter channel, this is also what others hinted at (to address the "consensus failure" catastrophic scenario).

5chdn:
Hi @oberstet - are you still working on this?

5chdn:
Marking this as stale. Happy to reopen this whenever you get back to it. :hugs:

PR #8335, opened 2018-04-09
Use hyper 0.11 in ethcore-miner and improvements in parity-reactor

sorpaas:
This aligns `ethcore-miner`'s `hyper` dependency with the rest of the crates. However, we still have `secret-store` depending on `hyper` 0.10 and I'm planning to fix that in a separate PR.

* `parity-reactor`'s `Remote::spawn_fn` now exposes the raw `Remote`'s handle. This allows the EventLoop to be used in many tokio contexts.
* `parity-miner` is updated to use async hyper request. This, probably until we get tokio runtime, would need to pass the Remote handle around.


miner/src/work_notify.rs
```diff
@@ -91,51 +83,27 @@ impl NotifyWork for WorkPoster {
...
-				}
+		let urls = self.urls.clone();
+
+		self.remote.spawn_fn(move |handle| {
+			let client = Config::default()
+				.keep_alive(true).build(handle);
+
+			let mut future: Box<Future<Item=(), Error=()>> = Box::new(future::ok(()));
+			for url in urls {
+				let mut req = Request::new(Method::Post, url.clone());
```
tomusdrw: Wouldn't it be better to use `fetch` instead of raw `hyper`?

rpc/src/v1/tests/eth.rs
```diff
@@ -35,6 +35,7 @@ use kvdb_memorydb;
 use miner::external::ExternalMiner;
 use miner::transaction_queue::PrioritizationStrategy;
 use parking_lot::Mutex;
+use parity_reactor::Remote;
```
tomusdrw: I'm not sure if `parity_reactor` abstraction is something we want/should keep.

miner/src/work_notify.rs
sorpaas: I'm not sure whether we can use `fetch` to send POST request. The `fetch` crates looks like to use a background thread and is hard coded to only send GET request? https://github.com/paritytech/parity/blob/master/util/fetch/src/client.rs#L216

debris: this is changed in #6422 which will be merged in a second https://github.com/paritytech/parity/pull/6422/files#diff-3eef53ff37b934cea2e67f6f251a2377R131

sorpaas: Got it. I'll try to get a version based on `fetch` once that PR is merged.

Issue #4842, opened 2017-03-09
Parity-as-a-library


rphmeier:
- [ ] APIs for running full and light nodes
- [ ] C interface

rphmeier:
...and then the ability to cross-compile this library to IoS and Android.
@debris in your experience what will be the hardest part of that?

debris:
> ... what will be the hardest part of that?

Probably cross-compiling libraries containing C code... Besides that, writing C interfaces itself, cause it's super time consuming

arkpar:
The C interface can be simple enough. It's mostly configuration, creating, starting and stopping Parity service and polling for status to detect critical errors. In the simplest implementation configuration can be done by passing a text in toml format. So basically just 3 functions.
```
parity_service_handle parity_create(const char* config);
void parity_destroy(parity_service_handle handle);
void parity_poll_status(parity_service_handle handle, parity_status* status);
```
All other communication goes over RPC/IPC.

If we want to expose our rich API to C, these bindings need to be auto-generated from the RPC traits.

rphmeier:
That definitely seems like a good approach.
I wonder about the viability of running an RPC server for a hosted node within an app. We might want to just have a native code RPC endpoint

`char* parity_rpc(parity_service_handle handle, char* request);`
which just wraps a thread pool + event loop + `future.wait()` internally.

adrianbrink:
@debris But wouldn't that be the worry about an implementor to compile libraries that contain C code. From our side we would only need to worry about writing the C interfaces (which will take a lot of time).

adrianbrink:
To expose the C api something like [rusty-binder](https://gitlab.com/rusty-binder/rusty-binder) or [rusty-cheddar](https://github.com/Sean1708/rusty-cheddar) would be good to use. 

Also, let's see if the rust team will support this in the [future](https://github.com/rust-lang/rust/issues/10530).

rphmeier:
@adrianbrink I think @debris was referring to C/C++ libraries we have as dependencies, which may be hard to cross-compile.
To expose a C API we can just use `#[no_mangle] pub extern fn` in Rust code.

adrianbrink:
Oh right. Didn't know parity had C/C++ deps. 
Yeah, rusty-cheddar/binder could be used then to auto-generate the C header files.

tomaka:
Reopening as I'm going to work on this soon-ish.

PR #8337, opened 2018-04-09
Parity as a library

tomaka:
cc #4842 

This extracts most of the content of the `parity` crate into a new crate named `parity-lib`.
The `parity` crate mostly contains the `main` function, plus everything that is related to restarting the binary and applying updates.

The API of `parity-lib` consists in a single `start` function that accepts a list of CLI arguments as strings.
While accepting arguments as strings is not a great thing to do, it is the most straight-forward solution for now and can always be improved in the future.

Also for now `start` just blocks until we Ctrl+C or an update is ready or an RPC request to change the chain happens. This is obviously not great either and will be changed in the future. However I don't want to introduce too many changes at once because of conflicts with other PRs (eg. #8262) and the hell of solving conflicts.

I recommend reviewing commits one by one for an easier review. The first commit simply moves files, and the other ones apply some changes.


tomaka:
Going to restart from scratch, I'm in a clusterfuck of conflicts.

PR #8340, opened 2018-04-09
ethcrypto renamed to ethcore-crypto and moved to ethcore dir

debris:
to unify crate names and locations


secret_store/src/key_server.rs
```diff
@@ -21,7 +21,7 @@ use std::sync::mpsc;
 use futures::{self, Future};
 use parking_lot::Mutex;
 use tokio_core::reactor::Core;
-use ethcrypto;
+use crypto;
```
niklasad1: s/ethcore/crypto in the entire file

For example you forgot: https://github.com/paritytech/parity/blob/7b4955937f244b27d92b948cdedf1b3a2b598551/secret_store/src/key_server.rs#L108

rphmeier:
@niklasad1 historical reasons :) but I agree as well!

PR #8350, opened 2018-04-10
Bump beta to 1.10.1

5chdn:


util/version/Cargo.toml
```diff
@@ -3,7 +3,7 @@
 [package]
 name = "parity-version"
 # NOTE: this value is used for Parity version string (via env CARGO_PKG_VERSION)
-version = "1.10.0"
+version = "1.10.1"
```
tomusdrw: We should actually unflag `critical` a bit below this

5chdn:
Done

util/version/Cargo.toml
tomusdrw: Authorities should be running with `--auto-update` anyway. Marking it as critical will cause all Parity nodes to update (and it's a hassle to not deploy critical update on mainnet, we should actually have this flag separate between networks).

PR #8354, opened 2018-04-10
ethstore renamed to ethcore-store

debris:


debris:
> Readme still contains occurences of ethstore.

will fix it

> Can I also ask what exactly is a point of this rename if the binary is still called `ethstore`? Isn't it misleading?

At first, I'd like to unify the naming convention of our crates. `ethstore` is a binary and tbh I don't know if it's name should be changed

tomusdrw:
Is `ethcore-*` really a good naming convention? What is the rule here? What packages should go into a (already behemoth) `ethcore` subdirectory? (BTW I hope we won't rename `ethash` to `ethcore-hash`)
We also have couple of crates with `parity-*` names - this imho makes more sense.

debris:
> Is `ethcore-*` really a good naming convention? What is the rule here? What packages should go into a (already behemoth) `ethcore` subdirectory? (BTW I hope we won't rename `ethash` to `ethcore-hash`)

I don't think that `ethcore` is a good name either. imo, it should be either `core` or `ethereum`. So we would have either `core-ethash` or `ethereum-ethash` 🙄 

> We also have couple of crates with `parity-*` names - this imho makes more sense.

Parity names make a lot of sense to me as well. I'd just distinguish things that are ethereum specific from things that are parity specific

tomusdrw:
>  I'd just distinguish things that are ethereum specific from things that are parity specific

By ethereum-specific you mean usable wihout Parity (Ethereum Client)? While I agree that `ethcore-store` would be useful to generate keys for say Geth, I don't think this holds for _anything_ else within current `ethcore` crate, how is `ethcore-service`, `ethcore-client`, `ethcore-light` or `ethcore-migrations` ethereum-specific but not parity-specific? Isn't it the other way around?

5chdn:
How much longer shall we freeze this?

5chdn:
Closing as stale.

To contribute to this discussion, we should not determine if anything is ethereum-specific or parity-specific, but just name all of our crates `parity-*`.

PR #8356, opened 2018-04-10
Add `util/mem` to zero out memory on drop.

twittner:


ethkey/src/secret.rs
```diff
@@ -234,6 +246,35 @@ impl Deref for Secret {
...
+	}
+}
+
+impl AsRef<[u8]> for H256Mut {
+	fn as_ref(&self) -> &[u8] {
+		&self.val.0
+	}
+}
+
+impl AsMut<[u8]> for H256Mut {
```
debris: Is lack of this `impl`  the only reason you created `H256Mut`? I see no reason why it shouldn't be added to [paritytech/primitives](https://github.com/paritytech/primitives). Please make a pr with this change to that repo :)

twittner: Yes, will do.

tomusdrw: TBH I don't see if it fits primitives. All primitives we have are immutable, do we want to have both mutable and immutable ones there? IMHO makes the library unnecessary bloated.

twittner: There is already `DerefMut` in primitives though.

util/mem/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::ops::{Deref, DerefMut};
+use std::ptr;
+
+/// Wrapper to zero out memory when dropped.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Memzero<T: AsMut<[u8]>> {
```
tomusdrw: Why not replace the content in place and just require that `T: Default`?

ethkey/src/secret.rs
tomusdrw: Ah, ok. I thought we want to introduce `H256Mut` struct into that repo. If it's only about adding missing `AsMut` trait implementation then it's all good!

util/mem/src/lib.rs
twittner: I don't understand. Could you explain this a more fully please?

tomusdrw: Nvm, actually it's better with `AsMut` if we have all the supported structs.

whisper/src/rpc/crypto.rs
```diff
@@ -106,12 +107,14 @@ impl EncryptionInstance {
...
 					AesEncode::OnTopics(topics) => {
 						let mut buf = Vec::new();
-						let key = H256(key);
+						let key = H256(key.into());
 
 						for topic in topics {
 							buf.extend(&*(topic ^ key));
 						}
 
+						let _ = Memzero::from(key.0);
```
rphmeier: doesn't that just copy the value of `key` into some other place in the stack and then zero that?

twittner: Yes, thanks.

PR #8359, opened 2018-04-10
Use async hyper server in secret_store and upgrade igd

sorpaas:
This completely removes hyper 0.10 dependency.

The secret_store's HTTP listener is changed to use async hyper, and tokio runtime. Currently hyper doesn't implement `Send` for some of the important crates, so when we get an incoming socket, it needs to be processed in the same thread as the socket is received. (`current_thread::spawn`) Looks like this limitation will be removed in hyper 0.12.

When `Runtime` drops, it shuts down itself. So we don't need to implement `Drop` for the listener any more.


twittner:
Looks good to me.

PR #8360, opened 2018-04-10
EIP-758: Smart contract function return data by jsonrpc subscription & filter

adisney:
# EIP-758

## Motivation
Currently, if a smart contract function invocation would return a value, that data does not get made available to the client. As per [EIP-758](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-758.md), this pull request enables users to retrieve smart contract return data via either a jsonrpc websocket subscription (`eth_subscribe`) or the jsonrpc http polling filter (`eth_newReturnDataFilter`).

Augur has a bounty open for this feature to be implemented in either Geth or Parity.

## Usage
Prior to invoking a function on a smart contract, the user must either establish a websocket subscription or create a json polling filter. Once the transaction is mined, the return data will be made available to them.

### Websocket
Once a user has established a websocket connection, they can create a subscription for return data.
```
{"jsonrpc": "2.0", "id": 1, "method": "eth_subscribe", "params": ["returnData"]}
```
Return data will stream back to the user over the websocket.
```
{
  "jsonrpc": "2.0",
  "method": "eth_subscription",
  "params": {
    "result": {
      "transactionHash": "0x00000000000000000000000000000000000000000000000000000000deadbeef",
      "returnData": "0x000000000000000000000000000000000000000000000000000000000000002a",
      "removed": false
    },
    "subscription": "0x00000000000000000000000000000b0b"
  }
}
```

With the websocket subscription, if a chain reorganization takes place, any transaction that was mined on the removed block will again produce return data, however the `removed` field will now have the value `true`.

### Polling Filter
Via the jsonrpc interface, a user can send a new filter request to create the filter.
```
{"jsonrpc": "2.0", "id": 1, "method": "eth_newReturnDataFilter", "params": []}
```
To obtain updates from the filter, the user can invoke the `eth_getFilterChanges` jsonrpc method with the appropriate filter id as a parameter.
```
{"jsonrpc": "2.0", "id": 2, "method": "eth_getFilterChanges", "params": ["0x0"]}
```
## Shortcomings
The spec does not allow for a filter to be added to either the websocket or polling filter, so there is currently no way to limit the data that is returned. Every transaction will result in return data, so the user will have to find their transaction using the transaction hash they received when invoking the function. 
This concern is addressed in the notes below.

### EIP Update - New Polling Filter
This EIP has an [open issue](https://github.com/ethereum/EIPs/pull/985) that is likely to result in an update to the specification. The current method for creating a polling filter is `eth_newFilter`, taking `"returnData"` in the parameters.
```
{"jsonrpc": "2.0", "id": 1, "method": "eth_newFilter", "params": ["returnData"]}
```
This will likely be changed so that the method is named `eth_newReturnDataFilter`, taking no parameters.
```
{"jsonrpc": "2.0", "id": 1, "method": "eth_newReturnDataFilter", "params": []}
```
This suggested modification better adheres to the convention currently laid out for the other polling filters: `eth_newFilter`, `eth_newBlockFilter`, and `eth_newPendingTransactionFilter`. The above modification has support from the originator of the EIP, [tinybike](https://github.com/tinybike), as per our conversations on the Augur Discord.

### EIP Update - Filter By Address
Another desirable change to the specification would be the addition of filters to both the websocket subscription and polling filter creation so that users can limit the data that is returned to include only function invocations originated by specific owner addresses. These updates would be relatively simple to add to my implementation of the EIP.

Even without that filter, this pull request represents a significant modification to the current method of obtaining function return data.


ethcore/src/client/test_client.rs
```diff
@@ -622,7 +623,10 @@ impl BlockChainClient for TestBlockChainClient {
 	}
 
 	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		Ok(Box::new(self.execution_result.read().clone().unwrap().into_iter()))
+		match self.execution_result.read().clone() {
```
tomusdrw: Since it's a test client I think it would be better to require that the execution result is mocked and fail if it's not instead of falling back to empty iterator by default.

rpc/src/v1/impls/light/eth.rs
```diff
@@ -68,8 +69,8 @@ pub struct EthClient<T> {
 impl<T> EthClient<T> {
 	fn num_to_id(num: BlockNumber) -> BlockId {
 		// Note: Here we treat `Pending` as `Latest`.
-		//       Since light clients don't produce pending blocks
-		//       (they don't have state) we can safely fallback to `Latest`.
+		//		 Since light clients don't produce pending blocks
```
tomusdrw: I think one tab indentation should be enough

rpc/src/v1/types/return_data.rs
```diff
@@ -0,0 +1,30 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Tree route info type definition
+
+use ethereum_types::H256;
+use v1::types::Bytes;
+
+/// Represents return data of a contract function call
+#[derive(Debug, Clone, Serialize, Eq, PartialEq)]
+pub struct ReturnData {
```
tomusdrw: Docs for fields missing

rpc/src/v1/types/return_data.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
tomusdrw: 2018

rpc/src/v1/types/pubsub.rs
```diff
@@ -30,6 +30,8 @@ pub enum Result {
 	Log(Log),
 	/// Transaction hash
 	TransactionHash(H256),
+	/// Transaction return data
+	ReturnData(ReturnData)
```
tomusdrw: Missing trailing comma

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -172,6 +178,21 @@ impl<C> ChainNotificationHandler<C> {
...
+	fn notify_return_data<F>(&self, enacted: &[H256], calculate_return_data: F) where
+		F: Fn(&[H256]) -> Vec<ReturnData>
+	{
+		let return_data = calculate_return_data(enacted);
+
+		for subscriber in self.return_data_subscribers.read().values() {
+			let remote = self.remote.clone();
+			let subscriber = subscriber.clone();
+
+			for return_datum in return_data.clone() {
```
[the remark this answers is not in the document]
adisney: You're definitely correct that `remote` and `subscriber` didn't need to be cloned. I believe `return_data` does, however, because the same return data is sent to all subscribers.

adisney:
@tomusdrw, is it your suggestion that I add a cli flag that turns on these features *and* implement the cache, or would adding the cli flag for "unstable" features allow the pull request to go forward without the cache? I'm just trying to prioritize. :slightly_smiling_face: 

As for the `from` and `to` filtering, since those filtering features are not officially a part of the EIP should they be added now? As I mentioned in the initial PR, there is an open issue and PR for the EIP that will likely result in these being added, but they are not yet included in the spec itself.

adisney:
It is worth noting that tinybike has approved the change to the EIP to update the polling filter jsonrpc method name to `eth_newReturnDataFilter`, it just has to be merged, which I anticipate will happen shortly.

Once that is merged, I believe that the implementation of the EIP in this PR will conform with the standard jsonrpc interface. If we agree on that, @tomusdrw, do you believe a cli flag is still necessary?

adisney:
I propose that if caching is a requirement, then it be done not during regular block execution, but only after the first time the whole block is replayed. That way clients that don't need caching don't end up storing a lot of data they don't need.

The location I'm suggesting is in [Client::replay_block_transactions](https://github.com/paritytech/parity/blob/master/ethcore/src/client/client.rs#L1567)

I also think it's worth discussing if a chain re-org will invalidate the caching. My suggested implementation would not have any issues with this since the entire block's transactions get executed all at once and a re-org would not invalidate the transactions executed in any given block, even if that block got removed from the canonical chain. Thus, caching would only have to occur once and would reliably return the correct txn executions going forward. 

This is significant because the EIP requires, for the websocket connection, that an update be sent after a re-org indicating a txn return data has been removed. Caching would still be useful in this case.

rpc/src/v1/impls/light/eth.rs
```diff
@@ -533,6 +534,10 @@ impl<T: LightChainClient + 'static> Filterable for EthClient<T> {
 		self.client.block_hash(id).map(Into::into)
 	}
 
+	fn block_body(&self, _id: BlockId) -> JsonRpcResult<Option<encoded::Body>> {
+		Err(errors::light_unimplemented(None))
```
rphmeier: this can be easily implemented using the `light_fetch` helper

rphmeier: `light_fetch.block(...).map(|block| /* convert into encoded body somehow */`

It _will_ stall forever if the block is unknown, however. we will introduce a timeout in the future.

adisney: Given that it may stall, should this feature be implemented for light clients only after the timeout has been introduced?

rphmeier: It would be better to do it now; it will work for real blocks and only fail when people put bad data in and call it a block hash.

adisney: I've started work to address this concern. I've got an intermediate knowledge of Rust and I've run into an issue with lifetimes upon introducing Futures to my implementation of the polling filter. I'm a bit stuck and would really appreciate it if you could take a look at my issue by clone and compiling my code.

https://github.com/adisney/parity/tree/eip758-light-block-view

If you clone my parity fork and compile the `eip758-light-block-view` branch you'll find an error regarding lifetimes.

Thanks!

rphmeier: left some comments

adisney: @rphmeier Thanks for the help. I still seem to be having the same lifetime error. I pasted the error I'm seeing into a comment on the commit [here](https://github.com/adisney/parity/commit/a0a5433ae93ec6ffbc68797324acc8defcb7f20c).

adisney:
@rphmeier Thanks for the help on my compilation issues! I've added the change you suggested. Working on a cache now.

tomusdrw:
@adisney Sorry for a late response, somehow missed the discussion here.

Cache in `replay_block_transactions` works for me, the best would be to:
1. Detect that there are pending subscriptions
2. And fill cache during block import

Otherwise we are always executing block twice.

That said, we can work on that in a separate PR, and just log an issue for now.

5chdn:
Assuming this is still in progress?

adisney:
@5chdn I'm still moving forward on this, yes. Other things have me a bit busier, so things are going slightly slower. Seems like I'm closing to addressing all of the raised issues, though, so hopefully this is nearly done.

@tomusdrw, so you're looking for basic caching, at the very least, correct?

Also, the EIP-758 spec has been updated since I submitted this PR and my implementation now conforms to the spec, as is. The filtering would be nice to have but would be outside the scope of the spec as written. Knowing that, do you still want any command line flags to turn this feature on? Seems like it might not be needed since it's no longer an experimental feature.

tomusdrw:
@adisney Yup, a basic LRU cache (RPC layer?, cc @debris) of replayed blocks will work for now, when this gets merged I'll create an issue to work on a more sophisticated solution (pre-filling the cache during block processing when there are active subscriptions, or just handling subscriptions directly when executing a block).

adisney:
I'm looking into adding a MemoryLruCache. It seems to me that adding a MemoryLruCache will require changing many (all?) of the functions on the BlockChainClient to take a mut ref to self, which doesn't seems desirable.

I was thinking, as an alternative, of passing an Option<&mut MemoryLruCache> to replay_block_transactions and caching if it's != None.

Thoughts?

tomusdrw:
@adisney just wrap the cache into a `Mutex` or alternatively cache the result of calling `BlockChainClient` on the RPC layer (that might require more changes though).

adisney:
Caching is complete!

It looks like there were some non-trivial structural changes to some of the pubsub and filter code. I'm working on merging those changes in now.

adisney:
Changes have been merged

ethcore/src/executed.rs
```diff
@@ -23,172 +23,181 @@ use vm;
...
-	Internal(String),
-	/// Returned when generic transaction occurs
-	TransactionMalformed(String),
+    /// Returned when there gas paid for transaction execution is
+    /// lower than base gas required.
+    NotEnoughBaseGas {
+        /// Absolute minimum gas required.
+        required: U256,
+        /// Gas provided.
+        got: U256
```
debris: we use tabs instead of spaces. Please make sure that you code editor utilizes `.editorconfig` and revert these unnecessary changes

adisney:
@debris apologies. That has been fixed. I am configuring my editor to use .editorconfig going forward.

ethcore/src/client/client.rs
```diff
@@ -37,17 +37,17 @@ use blockchain::{BlockChain, BlockProvider, TreeRoute, ImportRoute, TransactionA
 use client::ancient_import::AncientVerifier;
 use client::Error as ClientError;
 use client::{
-	Nonce, Balance, ChainInfo, BlockInfo, CallContract, TransactionInfo,
-	RegistryInfo, ReopenBlock, PrepareOpenBlock, ScheduleInfo, ImportSealedBlock,
-	BroadcastProposalBlock, ImportBlock, StateOrBlock, StateInfo, StateClient, Call,
-	AccountData, BlockChain as BlockChainTrait, BlockProducer, SealedBlockImporter,
-	ClientIoMessage
+    Nonce, Balance, ChainInfo, BlockInfo, CallContract, TransactionInfo,
```
debris: I believe that you forgot about this file :)

adisney:
Gotcha. Updated the other files, as well!

adisney:
@debris are further changes needed?

5chdn:
paging @debris

adisney:
There has been some progress on finalizing the EIP spec, so I am proceeding with making speculative changes to anticipate those changes being approved officially.

ethcore/src/client/client.rs
```diff
@@ -1551,23 +1556,35 @@ impl BlockChainClient for Client {
 	}
 
 	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		let mut env_info = self.env_info(block).ok_or(CallError::StatePruned)?;
-		let body = self.block_body(block).ok_or(CallError::StatePruned)?;
-		let mut state = self.state_at_beginning(block).ok_or(CallError::StatePruned)?;
-		let txs = body.transactions();
-		let engine = self.engine.clone();
+		let block_hash = self.block_hash(block).ok_or(CallError::StatePruned)?;
+		let data_cached = {
```
tomusdrw: Instead of acquiring the lock twice, better do this:
```rust
if let Some(cached) = self.replay_cache.lock().get_mut(&block_hash) {
  return Ok(Box::new(cached.clone().into_iter()));
}
// whatever should be in else block
//...
```

ethcore/src/client/client.rs
```diff
@@ -1551,23 +1556,35 @@ impl BlockChainClient for Client {
...
-			})))
+			let result: Vec<Executed> = txs.into_iter()
+				.map(move |t| {
+					let t = SignedTransaction::new(t).expect(PROOF);
+					let machine = engine.machine();
+					let x = Self::do_virtual_call(machine, &env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
+					env_info.gas_used = env_info.gas_used + x.gas_used;
+					x
+				}).collect();
+			self.replay_cache.lock().insert(block_hash, result.clone());
```
tomusdrw: The cache has to take `Analytics` into consideration as well - currently you can get a cached result but for different `Analytics`.
Please try using `LruCache<(H256, Analytics), Vec<Results>>` as the cache key.

ethcore/src/client/test_client.rs
tomusdrw: Still to be addressed.

rpc/src/v1/types/return_data.rs
```diff
@@ -0,0 +1,34 @@
...
+use ethereum_types::H256;
+use v1::types::Bytes;
+
+/// Represents return data of a contract function call
+#[derive(Debug, Clone, Serialize, Eq, PartialEq)]
+pub struct ReturnData {
+    /// Hash of the transactions.
+	#[serde(rename="transactionHash")]
+	pub transaction_hash: H256,
+    /// Return data as bytes.
```
tomusdrw: Indentation is off

rpc/src/v1/types/return_data.rs~master
```diff
@@ -0,0 +1,30 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
tomusdrw: Please remove the whole file.

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -174,6 +180,28 @@ impl<C> ChainNotificationHandler<C> {
...
+		F: Fn(H256, &Ex) -> Option<Vec<ReturnData>>,
+		Ex: Send,
+	{
+		let return_data = enacted
+			.iter()
+			.filter_map(|&(hash, ref ex)| {
+				match calculate_return_data(hash, ex) {
+					Some(return_data) => Some(return_data),
+					None => None,
+				}
```
debris: this `match` is redundant

rpc/src/v1/impls/eth_filter.rs
```diff
@@ -246,11 +270,75 @@ impl<T: Filterable + Send + Sync + 'static> EthFilter for T {
...
+							}
+						})
+					.collect();
+					let return_data: Vec<_> = executed
+						.into_iter()
+						.map(|(block_id, output_bytes)| {
+							self.block_body(block_id)
+								.map(|body| {
+									match body {
+										None => vec![],
```
debris: What if block body does not exist? Is it possible? Later on you do `flat_map` on the result. If it's possible for a block body to not exist, the function will produce incorrect results. If it always exist, please write `.expect` here with a proof

5chdn:
Marking as _stale_. :hatching_chick:

5chdn:
Will reopen once you continue working on it. :+1:

adisney:
I'm hoping to get back to work on this in a week or so!

PR #8362, opened 2018-04-11
Dev chain - increase gasLimit to 8'000'000

Tbaut:
- fix https://github.com/paritytech/parity/issues/8342


tomusdrw:
Note it makes old development chains incompatible

Tbaut:
We'll probably need to make this clear and visible in the version Changelog @5chdn

PR #8366, opened 2018-04-11
parity uses winapi 0.3.4

debris:


parity/main.rs
```diff
@@ -250,8 +249,8 @@ fn global_init() {
 	// When restarting in the same process this reinits windows sockets.
 	unsafe {
 		const WS_VERSION: u16 = 0x202;
-		let mut wsdata: ::winapi::winsock2::WSADATA = ::std::mem::zeroed();
-		::ws2_32::WSAStartup(WS_VERSION, &mut wsdata);
+		let mut wsdata: ::winapi::um::winsock2::WSADATA = ::std::mem::zeroed();
+		::winapi::um::winsock2:::WSAStartup(WS_VERSION, &mut wsdata);
```
andresilva: Extra `:` before `WSAStartup`.

parity/url.rs
```diff
@@ -16,33 +16,20 @@
...
-}
-
 #[cfg(windows)]
 pub fn open(url: &str) {
 	use std::ffi::CString;
 	use std::ptr;
+	use winapi::um::shellapi::ShellExecuteA;
+	use winapi::um::winuser::SW_SHOWNORMAL as Normal;
 
 	unsafe {
```
niklasad1: I think you can remove unsafe here now!

parity/url.rs
```diff
@@ -16,34 +16,19 @@
...
-			CString::new("open").unwrap().as_ptr(),
-			CString::new(url.to_owned().replace("\n", "%0A")).unwrap().as_ptr(),
-			ptr::null(),
-			ptr::null(),
-			shell::Normal);
-	}
+	use winapi::um::shellapi::ShellExecuteA;
+	use winapi::um::winuser::SW_SHOWNORMAL as Normal;
+
+	ShellExecuteA(ptr::null_mut(),
```
niklasad1: Sorry, but I will repeat myself again because my comment was removed in the latest change!

But, actually `ShellExecuteA` can fail. Why doesn't we care about that?

https://msdn.microsoft.com/en-us/library/windows/desktop/bb762153(v=vs.85).aspx

debris: cause we're lazy :p created an issue for that :)

parity/url.rs
andresilva: Can you explain me why this can be removed? `ShellExecuteA` is still marked as unsafe.

debris: ah, for some reason I thought that winapi interface is safe... I should have double checked :/

andresilva: But shouldn't the build have failed?

niklasad1: @andresilva I think CI only runs on Linux right? So, using conditional compilation flags will never compile stuff that isn't `#[cfg(target_os="linux")]`

andresilva: You're right didn't think of that 👍

PR #8371, opened 2018-04-11
Unify and limit rocksdb dependency places

sorpaas:
This clears some of the roadblocks for #7865.

Removed rocksdb dependency in `secretstore`, and limit it to the following places:

* Test dependencies.
* Parity CLI.
* `migration_rocksdb`.

Some crate changes:

* `ethcore_migrations` is moved to CLI's `db::migration`, because it only contains two bare structs.
* Renamed `util/migration` to `util/migration_rocksdb`, and it is assumed that ParityDB doesn't need those migration logic.

Refactoring in Parity CLI.

* It is made so that all the rest of CLI does not depend on `rocksdb` any more. Instead, we have a `db` mod which exposes all DB operations. The DB mod exposes a renamed mod called `rocksdb`, so in the future we can just use feature flags to expose another mod binding with a different database. Once the interface is mature, we can move to use trait objects to dynamically dispatch db opening.

Comments on `ethcore/src/test_helpers.rs`:

* This file needs to be made `pub` because some of the helpers are used by `private-tx`. However, due to probably https://github.com/rust-lang/cargo/issues/2911 we cannot make it test only. I've tried to move `test_helpers` as a separate crate, but that introduces cyclic dependency and while Cargo happily accept it, trait impls won't work. So the temporary fix is to just create `test_helpers_internal` and move test only dependencies there. In the future we may want to consider organizing those helpers better.


ethcore/src/lib.rs
```diff
@@ -94,6 +94,7 @@ extern crate ansi_term;
 extern crate unexpected;
 extern crate kvdb;
 extern crate kvdb_memorydb;
+#[cfg(test)]
```
tomusdrw: Please keep the deps sorted.

parity/db/rocksdb/helpers.rs
```diff
@@ -0,0 +1,22 @@
+use std::path::Path;
```
tomusdrw: Preamble comment missing.

parity/db/rocksdb/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+// This function will only be used when compiling with secretstore feature.
+#[allow(dead_code)]
+pub fn open_secretstore_db(data_path: &str) -> Result<Arc<KeyValueDB>, String> {
+	let mut db_path = PathBuf::from(data_path);
+	db_path.push("db");
+	let db_path = db_path.to_str().ok_or_else(|| "Invalid secretstore path".to_string())?;
+	Ok(Arc::new(Database::open_default(&db_path).map_err(|e| format!("Error opening database: {:?}", e))?))
+}
+
+pub fn open_client_db(client_path: &Path, client_config: &ClientConfig) -> Result<Arc<KeyValueDB>, String> {
```
tomusdrw: Missing docs for public methods? Would be best to add `#![warn(missing_docs)]`

parity/export_hardcoded_sync.rs
```diff
@@ -90,19 +87,8 @@ pub fn execute(cmd: ExportHsyncCmd) -> Result<String, String> {
...
-			wal: cmd.wal,
-			.. DatabaseConfig::with_columns(NUM_COLUMNS)
-		};
-
-		Arc::new(Database::open(
-			&db_config,
-			&db_dirs.client_path(algorithm).to_str().expect("DB path could not be converted to string.")
-		).map_err(|e| format!("Error opening database: {}", e))?)
-	};
+	let db = db::open_db(&db_dirs.client_path(algorithm).to_str().expect("DB path could not be converted to string."),
```
tomusdrw: Maybe have each argument in a separate line?

parity/run.rs
```diff
@@ -250,19 +247,8 @@ fn execute_light_impl(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<Runnin
...
-			wal: cmd.wal,
-			.. DatabaseConfig::with_columns(NUM_COLUMNS)
-		};
-
-		Arc::new(Database::open(
-			&db_config,
-			&db_dirs.client_path(algorithm).to_str().expect("DB path could not be converted to string.")
-		).map_err(|e| format!("Error opening database: {}", e))?)
-	};
+	let db = db::open_db(&db_dirs.client_path(algorithm).to_str().expect("DB path could not be converted to string."),
```
tomusdrw: Same comment regarding formatting.

parity/db/rocksdb/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use cache::CacheConfig;
+
+mod migration;
+mod helpers;
+
+pub use self::migration::migrate;
+
+// This function will only be used when compiling with secretstore feature.
+#[allow(dead_code)]
```
tomusdrw: Can't we have `#[cfg(feature = )]` here instead of dead code?

secret_store/src/key_storage.rs
```diff
@@ -82,17 +82,17 @@ pub trait KeyStorage: Send + Sync {
 
 /// Persistent document encryption keys storage
 pub struct PersistentKeyStorage {
-	db: Database,
+	db: Arc<KeyValueDB>,
```
tomusdrw: Introduces virtual calls, but I guess it's not performance critical.

parity/db/rocksdb/mod.rs
sorpaas: I added docs for all public methods in `db/rocksdb/mod.rs`.

The weird thing is that the CLI already has `#![warn(missing_docs)]` in `main.rs` but I never saw any missing_docs warnings when building it.

sorpaas:
Fixed the grumbles!

parity/db/rocksdb/mod.rs
tomusdrw: It has to be added to every crate, I suppose `db` crate might not have it.

sorpaas: That's still a little bit weird though... `parity/db` is a mod not a separate crate.

sorpaas: Okay I think I found out why. `#![warn(missing_docs)]` will only warn you if it's an exported item. Because none of `db` mod is exported, so there was no warning whether it has docs or not.

PR #8373, opened 2018-04-11
Use tokio::spawn in secret_store listener and fix Uri

sorpaas:
Sorry I broke it... XD

* Turned out that we can use `tokio::spawn` here. :smile: The thing I missed is in the below `Service` definition -- you must mark the boxed Future to be Send (and it indeed can be Send already), otherwise the type information is lost.
* We cannot use `current_thread` there because I misunderstood how it works -- `current_thread` will only try to get the `CurrentThread` executor, regardless whether there's already other executors in this thread. That's why it panicked.
* `Uri`'s absolute_path meaning is changed in hyper 0.11.

### DIFF
--- a/secret_store/src/listener/http_listener.rs
+++ b/secret_store/src/listener/http_listener.rs
@@ -105,7 +105,7 @@ impl KeyServerHttpListener {
 				});
 
 				// TODO: Change this to tokio::spawn once hyper is Send.
-				serve.wait();
+				let _ = serve.wait();
 				future::ok(())
 			});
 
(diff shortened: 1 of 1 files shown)


secret_store/src/listener/http_listener.rs
```diff
@@ -203,11 +201,13 @@ impl KeyServerHttpHandler {
 	}
 }
 
+unsafe impl Send for KeyServerHttpHandler { }
```
sorpaas: Sorry this line is not necessary. Will remove.

PR #8377, opened 2018-04-11
ParityShell::open `Return result`

### LINKED ISSUE #8369: parity::open should return error if it fails
discussion: https://github.com/paritytech/parity/pull/8366#discussion_r180733188

niklasad1:
Attempt to close #8369

### DIFF
--- a/parity/url.rs
+++ b/parity/url.rs
@@ -26,25 +26,25 @@ pub enum Error {
 }
 
 impl From<std::io::Error> for Error {
-    fn from(err: std::io::Error) -> Self {
-        Error::ProcessError(err)
-    }
+	fn from(err: std::io::Error) -> Self {
+		Error::ProcessError(err)
+	}
 }
 
 impl From<std::ffi::NulError> for Error {
-    fn from(err: std::ffi::NulError) -> Self {
-        Error::FfiNull(err)
-    }
+	fn from(err: std::ffi::NulError) -> Self {
+		Error::FfiNull(err)
+	}
 }
 
 impl std::fmt::Display for Error {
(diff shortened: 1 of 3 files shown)


parity/url.rs
```diff
@@ -16,33 +16,70 @@
...
+	use winapi::shared::minwindef::INT;
+
+	const WINDOWS_SHELL_EXECUTE_SUCCESS: i32 = 32;
 
-	unsafe {
+	let h_instance = unsafe {
 		ShellExecuteA(ptr::null_mut(),
-			CString::new("open").unwrap().as_ptr(),
-			CString::new(url.to_owned().replace("\n", "%0A")).unwrap().as_ptr(),
+			CString::new("open")?.as_ptr(),
```
tomaka: Isn't `unwrap` "better" for `CString::new("open")`? It logically shouldn't fail.

parity/url.rs
```diff
@@ -16,33 +16,70 @@
 
 //! Cross-platform open url in default browser
 
+use std;
+
+#[allow(unused)]
+pub enum Error {
+	ProcessError(std::io::Error),
+	FfiNull(std::ffi::NulError),
+	WindowsShellExecute,
```
tomaka: I'd suggest putting the return value of `ShellExecuteA` (casted as a `c_int`) as parameter, for easier debugging if that actually happens.

parity/url.rs
niklasad1: Yeah, you are right it should not fail unless the `&str` contains `0`!

`let ff = CString::new("\u{0}").unwrap();` will fail!

5chdn:
@tomaka can I dismiss your review or do you want to look at the changes again?

PR #8385, opened 2018-04-12
handle queue import errors a bit more gracefully

rphmeier:
Related to #6319


rphmeier:
yes, agreed -- not necessarily a fix but worth doing. Also worth backporting where possible.

Issue #8239, opened 2018-03-28
impl error::Error for ethcore::Error


debris:
Could be also done by moving `ethcore::Error` to `error_chain`

https://github.com/paritytech/parity/pull/8150#discussion-diff-177361219R276

debris:
yes, it can!

PR #8386, opened 2018-04-12
Move ethcore::Error to error_chain

ascjones:
As suggested in #8239.

Moved some but not all `Errors` in `ethcore` to error_chain: Top level `Error` as well as `ImportError` and `BlockImportError`.


ethcore/src/error.rs
```diff
@@ -195,203 +216,121 @@ pub enum TransactionImportError {
...
+		PowHashInvalid {
+			description("PoW hash is invalid or out of date.")
+			display("PoW hash is invalid or out of date.")
+		}
+	
+		#[doc = "The value of the nonce or mishash is invalid."]
+		PowInvalid {
+			description("The value of the nonce or mishash is invalid.")
+			display("The value of the nonce or mishash is invalid.")
+		}
```
niklasad1: Tab/space formatting

ascjones: Doh. Fixed.

PR #8390, opened 2018-04-13
remove From::from.

EighteenZi:
It seems not necessary. The return type of `transact_with_tracer` is Result<Executed<T::Output, V::Output>, ExecutionError>, and  `Err(ExecutionError(xxx))` is the correct error type.

And self.state.nonce() needs `From::from`, as its' return type is 'trie::Result<U256>'.


rphmeier:
some redundant commits and CI is failing.

twittner:
The changes (removing redundant `From::from` calls) look good to me. The CI failure was a glitch. Maybe the commits could be squashed as some intermediate ones seem unrelated?

debris:
@twittner we squash on merge anyway, so it does not matter :)

PR #8400, opened 2018-04-13
Tx permission contract improvement

VladLupashevskyi:
Transaction regulation plays an important role in permissioned blockchains. Current implementation of the tx permission contract provides very limited possibilities to filter tx types (basic transaction, contract call and contract creation) only by transaction sender address.

I would like to introduce an extended version for tx permission contract, which adds more flexibility to transaction regulation.

Transaction `to` address and `value` in ether have been added to arguments of allowedTxTypes function.
`to` address represents tx destination address in case of basic transaction and contract call. When contract creation comes up this parameter equals to zero address since it doesn't make sense in this case.

I believe that this update broadens the range of options for the tx filtering such as opportunity to filter tx for sender and recipient address pairs as well as to limit ether sending from/to specific addresses.

I would really appreciate these changes to be merged into parity repository, however I understand that it might break blockchains that are using current implementation of tx permission contract so I'm open for the backward compatibility discussion!


ddorgan:
[clabot:check]

ethcore/src/tx_filter.rs
```diff
@@ -60,17 +60,21 @@ impl TransactionFilter {
 	/// Check if transaction is allowed at given block.
 	pub fn transaction_allowed<C: BlockInfo + CallContract>(&self, parent_hash: &H256, transaction: &SignedTransaction, client: &C) -> bool {
 		let mut cache = self.permission_cache.lock();
-
+		let mut to: Address = Address::new();
 		let tx_type = match transaction.action {
```
tomusdrw: I'd prefer a tuple to be returned from `match` statement instead of modifying `to`.

grbIzl:
@tomusdrw @debris  Can we add somehow versioning for our built-in contracts? What do you think? Obviously it's not the last such case, when we might want to change their ABI.

tomusdrw:
@grbIzl Yeah, definittely. I think every our contract should have two obligatory public const functions: `name` (string/bytes32) and `version` (uint). We can start with a super simple versioning scheme where we just increment the version number. CC @kirushik

VladLupashevskyi:
In case of this tx permissions contract improvement we use 3 parameters in `allowedTxTypes` function instead of 1, so the ABI is changed. In order to support backward compatibility we need to support both ABIs for old and new contract, but since it's specified in `use_contract!` macros and defined in `TransactionFilter` struct, I doubt that we can change ABI on the fly.
I have an idea to change name of `contract` field in `TransactionFilter` struct to `deprecated_contract` and use `contract` field for the contract with the improved ABI, but I'm not sure it's perfect solution.
Another idea is to use different function name (for example: `allowedTxTypesExtended`) with 3 params and leave old one as it is. So we have an ABI with two functions and by checking contract version we call corresponding func.
Could somebody give me an advise how to deal with this situation?
Thank you in advise!

VladLupashevskyi:
Here I introduced support for the backward compatibility of previous version of tx permission contract. So basically first the function with 3 params is called and if it fails the function with 1 param is called. I also adapted cache to support 3 params, but in case if old contract version is used this cache will be still using 3 params. How do you think should the additional logic be added for the cache, where contract version will be checked or we can leave it as it is?
I also introduced `contractName`, `contractNameHash` and `contractVersion` constant functions in the new contract for future improvements support.
You can find contract code for tests [here](https://gist.github.com/VladLupashevskyi/a157d8162db85de67de522eeb7c8ee85).

ethcore/src/tx_filter.rs
```diff
@@ -60,36 +63,48 @@ impl TransactionFilter {
...
+					(tx_permissions::CALL, address)
+				} else {
+					(tx_permissions::BASIC, address)
+				}
 		};
 
 		let sender = transaction.sender();
-		let key = (*parent_hash, sender);
+		let value = transaction.value;
+		let key = (*parent_hash, sender, to, value);
```
tomusdrw: Just realized that it makes the cache pretty useless and the whole transaction filter checking way slower - every transaction will cause a state call anyway (cause it's quite unlikely all 3 parameters will be the same).

Perhaps we should consider either:
1. removing the cache completely
2. change the contract return type and allow some quantification (like: permission `None`, sender never allowed)

ethcore/src/tx_filter.rs
```diff
@@ -83,24 +83,26 @@ impl TransactionFilter {
 		let contract_address = self.contract_address;
 
 		// Check permissions in smart contracts
-		let permissions = self.contract.functions()
+		let (permissions, filter_only_sender) = self.contract.functions()
```
VladLupashevskyi: I think we should keep caching for performance reasons, so I came up with an idea to have 2nd return value in the contract function, which will define whether filtering depends on to address and value in ether and if not, sender address will be cached which will increase performance.

ethcore/src/tx_filter.rs
```diff
@@ -122,7 +124,7 @@ mod test {
 	use transaction::{Transaction, Action};
 	use tempdir::TempDir;
 
-	/// Contract code: https://gist.github.com/VladLupashevskyi/a157d8162db85de67de522eeb7c8ee85
+	/// Contract code: https://gist.github.com/VladLupashevskyi/84f18eabb1e4afadf572cf92af3e7e7f
```
tomusdrw: > bool - flag whether #'sender depends on #'to' and #'value'. If it's false
     *    the result will be cahed.

`s/cahed/cached`` and also it seems it's the other way around, right? If the result is `true` it will be cached?

"Caching" also seems a bit too harmless for this for me. We should emphasize that if `true` is returned the same permissions will be applied for all transactions from the same sender WITHOUT calling the contract again.

VladLupashevskyi: @tomusdrw I agree the explanation was too confusing.

I updated gist where flag description is more clarified.

ethcore/src/tx_filter.rs
```diff
@@ -60,36 +63,50 @@ impl TransactionFilter {
...
 		let key = (*parent_hash, sender);
 
 		if let Some(permissions) = cache.get_mut(&key) {
 			return *permissions & tx_type != 0;
 		}
 
 		let contract_address = self.contract_address;
-		let permissions = self.contract.functions()
+
+		// Check permissions in smart contracts
```
grbIzl: Can we mention here (or somewhere, where it's appropriate), that ABI of the contract with version=2 is used? Ideally is to add explicit check of the version in the code and force the next person, who will modify this code, to rely on contract's version. But perhaps at least some comment will be also helpful. PS. May be the fallback to the deprecated contract in case, when version is not retrieved, also will be looking more logically

ethcore/src/tx_filter.rs
```diff
@@ -100,16 +117,135 @@ mod test {
...
+		call_tx.action = Action::Call(Address::from("0000000000000000000000000000000000000005"));
+
+		let mut basic_tx_with_ether_and_to_key7 = Transaction::default();
+		basic_tx_with_ether_and_to_key7.action = Action::Call(Address::from("d41c057fd1c78805aac12b0a94a405c0461a6fbb"));
+		basic_tx_with_ether_and_to_key7.value = U256::from(123123);
+		let mut call_tx_with_ether = Transaction::default();
+		call_tx_with_ether.action = Action::Call(Address::from("0000000000000000000000000000000000000005"));
+		call_tx_with_ether.value = U256::from(123123);
+
+
```
grbIzl: Extra line

ethcore/src/tx_filter.rs
```diff
@@ -100,16 +117,135 @@ mod test {
...
 	use ethkey::{Secret, KeyPair};
 	use super::TransactionFilter;
 	use transaction::{Transaction, Action};
 	use tempdir::TempDir;
 
-	/// Contract code: https://gist.github.com/arkpar/38a87cb50165b7e683585eec71acb05a
+	/// Contract code: https://gist.github.com/VladLupashevskyi/84f18eabb1e4afadf572cf92af3e7e7f
 	#[test]
 	fn transaction_filter() {
+		let spec_data = r#"
```
grbIzl: We are trying to move such string data into external resources and use in the code via: let spec_data = include_str!("../res/filename.json");

grbIzl:
Looks good for me. But test compilation failed

5chdn:
@grbIzl please approve :heavy_check_mark: the PR if it looks good for you.

Also we need two reviews on each PR, so I'll add A0 again.

debris:
This pr needs to be rebased / merged with master

S3bb1:
Hi all, nice change! We're looking after such a feature in Partiy. On question regarding the permissions. Will this also block contract internal calls ? 

Like if any accountid can't call contract (0x02) anymore (but i know the ABI of it). So when i create a new contract (0x03) which calls my needed getter of 0x02 will this also be blocked ?

S3bb1:
@VladLupashevskyi thank you for your response! Okay this is a first step into our need, but we must handle the other side (internal contract calls) also. (Off Topic) Would this even be possible without touching the EVM?

VladLupashevskyi:
Rebased onto master

5chdn:
It looks like you rebased in the wrong direction. Check commits / files.

VladLupashevskyi:
@5chdn Oops, sorry for this confusion. I merged master into this branch. Now everything is ok with files.

5chdn:
@tomusdrw could you have a look at this again? :)

Issue #7162, opened 2017-11-29
EIP-1011: Hybrid Casper FFG / Ethash


5chdn:
http://eips.ethereum.org/EIPS/eip-1011

rphmeier:
I can take it. What this will need

  - [ ] Finality information in the blockchain struct (power reversion limit)
  - [ ] Networking code for casting epoch votes
  - [ ] Modification of Ethash engine for epoch voting (can we fit this into the existing `Engine` interface?

A link to casper specs would be great (I have a grasp of CasperFFG roughly but not details)

5chdn:
I'm on it to figure out the specs.

vbuterin:
Would be happy to provide any assistance or support.

We will try to prioritize improving documentation asap. In the mean time, there's the paper and the python code:

http://github.com/karlfloersch/docker-pyeth-dev  
https://github.com/karlfloersch/pyethereum  
https://github.com/karlfloersch/pyethereum/tree/develop/ethereum/hybrid_casper  
http://github.com/ethereum/casper  

I recommend starting with the fork choice rule (priority 1: highest justified epoch #, priority 2: max total diff) and the validator logic.

5chdn:
cc @pipermerriam

5chdn:
http://notes.eth.sg/MYEwhswJwMzAtADgCwEYBM9kAYBGJ4wBTETKdGZdXAVmRvUQDYg=?view

https://arxiv.org/abs/1710.09437

rphmeier:
The paper is pretty good documentation for a start, but I haven't dug into how votes are broadcast/accumulated in practice -- does a smart contract keep track of votes and justified checkpoints? or does a full node have to track all justified checkpoints internally? How are light clients meant to track finality?

Engineering concerns:
  - `BlockChain` currently holds the whole fork-choice decision, based only on total difficulty. Parts of this will need to be deferred to the `Engine`.
  - `Engine` needs to get a persistent data store to keep track of any votes targeting non-finalized checkpoints to avoid violating slashing conditions. This brings complication due to snapshot restoration replacing the DB in-place.

vbuterin:
> does a smart contract keep track of votes and justified checkpoints?

Yes. The Casper contract basically keeps track of everything.

> How are light clients meant to track finality?

Probably via a special-purpose protocol, like randomly selecting 50
validator indices and verifying that valid votes for enough of those
validators exist.

> Engine needs to get a persistent data store to keep track of any votes
[...]

PR #8401, opened 2018-04-13
Fork choice and metadata framework for Engine

sorpaas:
cc #7162

This adds fork choice and metadata framework suitable for implementing Casper. Changes are discussed in sections below.

#### General comments on finalization, total difficulty and metadata

In this PR, we keep block reorg rules implemented in `BlockChain`, so as not to over-extend `Engine`. And we keep the total difficulty semantics unchanged, so as not to add complexity regarding wrap sync and light client. This requires `BlockChain` know about `is_finalized` and `total_difficulty`.

Metadata is implemented as a binary blob. This is because `Engine` currently is a trait object and we cannot specialize metadata types for each engine. The intention is that an engine will parse the metadata blob as RLP view.

#### `BlockDetails`

`BlockDetails` is extended with `is_finalized` and `metadata`, and is implemented in a backward compatible way. If `is_finalized` is false and `metadata` is empty, it is parsed in the old way, otherwise, in the new, extended way.

#### `LiveBlock`

I think it's suitable to consider `metadata` information as some extensions to a block's state. In most cases, we only allow setting metadata information when a block is live/open. This makes it so that we avoid the ambiguity about what will happen if block reorg happens. In `OpenBlock`, we allow setting of metadata and finalization (instant finality). For cases where info must be set on an ancestry block, we collect a list of `AncestryAction` and set it only when a new block is about to be committed. Currently `AncestryAction` only contains marking finality (post finality).

#### `parity_machine` traits

Several new traits based on `Header` and `LiveBlock` are added, allowing querying and setting metadata and finality information. `Machine` traits contain two new types:

* `ExtendedHeader`, this is a normal header (so it's entirely possible that `Machine::Header == Machine::ExtendedHeader`), but extended with metadata information. In `EthereumMachine`, extended header has the total difficulty (total score) info and finality info.
* `AncestryAction` (as explained above)

#### Engine lifecycle extension

* `on_new_block` is extended with `ancestry`. This is an iterator of `ExtendedHeader`. The intention here is that it allows an engine to populate new block's metadata based on parent block. (But it's not needed by Casper).
* `ancestry_actions` is called before the block is about to commit. This allows the block to set ancestry blocks finality info.
* `is_new_best` is called with `new` and `best` as `ExtendedHeader`. The consideration here not to use an `AncestryIterator` is that an engine can easily get those info from `on_new_block` and `ancestry_actions`.

#### Casper

To implement Casper fork choice rules using this framework:

* In `on_close_block`, set the `score` metadata and `vote` metadata information based on `new.casper`.
* In `ancestry_actions`, check whether we are just in a new epoch, if so, check whether the last epoch as well as this epoch has enough votes. If so, emit an `AncestryAction` that sets the last epoch's block as finalized.
* In `is_new_best`, set the fork choice rule to be based on `score` metadata information.

And a note on finalization: in the above I assume we use the "finalized" term as defined in the [Casper FFG paper](https://github.com/ethereum/research/blob/master/papers/casper-basics/casper_basics.pdf). That is, we consider a block to be finalized as long as a) there's a super majority link, and b) the link is about a direct child. The [python implementation](https://github.com/karlfloersch/pyethereum/blob/develop/ethereum/hybrid_casper/chain.py#L201-L249) looks that to be using slightly different conditional branch, but I think there's a way to prove that they're equivalent (or doesn't matter).


ethcore/src/engines/mod.rs
```diff
@@ -338,6 +340,16 @@ pub trait Engine<M: Machine>: Sync + Send {
 	fn is_timestamp_valid(&self, header_timestamp: u64, parent_timestamp: u64) -> bool {
 		header_timestamp > parent_timestamp
 	}
+
+	/// Check whether a given block is the best block.
+	fn is_new_best(&self, bytes: &[u8], best_block_total_difficulty: U256, provider: &BlockProvider) -> bool {
```
rphmeier: can we abstract around `total_difficulty` and `provider` at the `Machine` level?

sorpaas: Will give it a try.

sorpaas: Here's an attempt: I added assoc type `Machine::BlockMetadata` and `Machine::BlockProvider`, and make `Engine::is_new_best` depend on them only. For `EthEngine`, `BlockMetadata` is currently just `U256` which is the total difficulty (and we can extend it to contain `Vec<(Bytes, Bytes)>` in the future).

Client would first call `Engine::generate_metadata` to get the current set of metadata, and then call `Engine::is_new_best` before passing it to blockchain storage.

ethcore/src/engines/mod.rs
```diff
@@ -338,6 +340,27 @@ pub trait Engine<M: Machine>: Sync + Send {
...
+	/// Check whether a given block is the best block.
+	fn is_new_best(&self, bytes: &[u8], block_metadata: &M::BlockMetadata, best_block_metadata: &M::BlockMetadata, provider: &M::BlockProvider) -> bool;
+}
+
+/// Generate metadata for a new block based on the default total difficulty rule.
+pub fn total_difficulty_generate_metadata(bytes: &[u8], provider: &BlockProvider) -> U256 {
+	let block = BlockView::new(bytes);
+	let header = block.header_view();
+	let parent_hash = header.parent_hash();
+	let parent_details = provider.block_details(&parent_hash).unwrap_or_else(|| panic!("Invalid parent hash: {:?}", parent_hash));
```
rphmeier: it would simplify things a lot just to provide the parent metadata

rphmeier: > we now have logic in engine that needs to check a block's timestamp

it probably should be abstracted at some point. we have `verify_block`/`verify_header` functions that can handle it AFAIK.

> InternalSealer, BlockVerifier, TimestampVerifier, ForkChooser

The sub-traits approach would work if we weren't using Engine as a trait object everywhere. That's why there's the `EthEngine` specialization which is implementation for all `Engine<EthereumMachine>`.

I don't see anything in this PR that requires introduction of a dependency of Ethereum types.
In the parity-machine crate we could add an associated type `Metadata` and a 

```rust
trait TotalDifficulty {
    type Value: Ord;

    fn total_difficulty(&self) -> &Value;
}
```

and then implement this trait for `U256` or some wrapper in the `EthereumMachine` implementation. Our engines then simply need the bound that `M::Metadata: TotalDifficulty`.

machine/src/lib.rs
```diff
@@ -81,8 +81,12 @@ pub trait Machine: for<'a> LocalizedMachine<'a> {
 	type LiveBlock: LiveBlock<Header=Self::Header>;
 	/// A handle to a blockchain client for this machine.
 	type EngineClient: ?Sized;
+	/// A handle to the block provider for this machine.
+	type BlockProvider: ?Sized;
```
rphmeier: doesn't `EngineClient` do this already?

ethcore/src/engines/mod.rs
```diff
@@ -338,6 +340,27 @@ pub trait Engine<M: Machine>: Sync + Send {
 	fn is_timestamp_valid(&self, header_timestamp: u64, parent_timestamp: u64) -> bool {
 		header_timestamp > parent_timestamp
 	}
+
+	/// Generate metadata for a new block.
+	fn generate_metadata(&self, bytes: &[u8], provider: &M::BlockProvider) -> M::BlockMetadata;
```
rphmeier: I'm wary of providing the entire blockchain context here. We do it in a specialized way through `Ancestry` in some localized points. In most cases we won't need anything more than the `parent_metadata`.

I would recommend changing the signature to `generate_metadata(header, ancestry)`. Then we can invoke the rolling finality tracker in `authority_round` from there and embed the most recently-finalized block hash in there and use it in the fork choice. That will pose a problem with strongly-typed metadata according to the machine because if we've defined the "machine" metadata to be only a `TotalDifficulty` then that's not enough for that modification `authority_round`'s fork choice.

The other option is to avoid strongly-typed metadata entirely. This is kind of necessary until we stop using trait-object engines.

sorpaas:
I think this PR is ready for review now.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -409,6 +409,27 @@ impl<'a> Iterator for AncestryIter<'a> {
 	}
 }
 
+impl<'a> AncestryIter<'a> {
+	pub fn to_header_iter(self) -> Box<Iterator<Item=ExtendedHeader> + 'a> {
```
rphmeier: can and should use `impl Iterator` for this

rphmeier:
> I assume we use the "finalized" term as defined in the Casper FFG paper.

For casperFFG yeah. Finality is a broader term which just means that the consensus engine has actually agreed on a value. "safe" engines never finalize more than one thing at a given height.

ethcore/src/block.rs
```diff
@@ -256,6 +282,7 @@ impl<'x> OpenBlock<'x> {
 		gas_range_target: (U256, U256),
 		extra_data: Bytes,
 		is_epoch_begin: bool,
+		ancestry: Box<Iterator<Item=ExtendedHeader> + 'a>,
```
rphmeier: can we avoid the `Box<Iterator>` here? perhaps passing an `&mut Iterator`?

ethcore/src/blockchain/blockchain.rs
```diff
@@ -955,24 +988,46 @@ impl BlockChain {
...
 			} else {
 				BlockLocation::Branch
-			}
+			},
 		}
 	}
 
+	/// Mark a block to be considered finalized. Panic if the block hash is not found.
+	pub fn mark_finalized(&self, batch: &mut DBTransaction, block_hash: H256) {
+		let mut block_details = self.block_details(&block_hash).expect("Block hash should exist from caller.");
```
rphmeier: that's an overly eager panic.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1093,6 +1148,20 @@ impl BlockChain {
 		}
 	}
 
+	/// Iterator that lists `first` and then all of `first`'s ancestors, by extended header.
+	pub fn ancestry_header_iter<'a>(&'a self, first: H256) -> Box<Iterator<Item=ExtendedHeader> + 'a> {
```
rphmeier: the `ancestry_iter` method also returns headers. we could call this `ancestry_with_metadata`

ethcore/src/blockchain/extras.rs
```diff
@@ -177,6 +178,57 @@ pub struct BlockDetails {
...
 	pub children: Vec<H256>,
+	/// Whether the block is considered finalized
+	pub is_finalized: bool,
+	/// Additional block metadata
+	pub metadata: Option<Vec<u8>>,
+}
+
+impl rlp::Encodable for BlockDetails {
+	fn rlp_append(&self, stream: &mut rlp::RlpStream) {
+		let use_short_version = !self.is_finalized;
```
rphmeier: seems like metadata is being discarded for any non-finalized blocks

ethcore/src/blockchain/update.rs
```diff
@@ -22,3 +22,13 @@ pub struct ExtrasUpdate<'a> {
 	/// Modified transaction addresses (None signifies removed transactions).
 	pub transactions_addresses: HashMap<H256, Option<TransactionAddress>>,
 }
+
+/// Extra information in block insertion.
+pub struct ExtrasInsert {
+	/// Is the inserted block considered the best block.
+	pub is_new_best: bool,
+	/// Is the inserted block considered finalized.
+	pub is_finalized: bool,
```
rphmeier: how can `is_new_best` be false and `is_finalized` be true? Any finalized block is always going to be in the fork-choice, and since we only insert fork tips an instantly finalized block would automatically be our new "best block".

ethcore/src/client/client.rs
```diff
@@ -520,6 +523,36 @@ impl Importer {
...
+		let ancestry_actions = self.engine.ancestry_actions(block.block(), chain.ancestry_header_iter(*parent));
+
+		let metadata = block.block().metadata().map(Into::into);
+		let is_finalized = block.block().is_finalized();
+
+		let new = ExtendedHeader {
+			header: header.clone(),
+			is_finalized: is_finalized,
+			metadata: metadata,
+			parent_total_difficulty: chain.block_details(&parent).unwrap_or_else(|| panic!("Invalid parent hash: {:?}", parent)).total_difficulty
```
rphmeier: for this and other panics in the PR, please follow the codebase guideline: https://github.com/paritytech/parity/issues/1026

ethcore/src/engines/mod.rs
```diff
@@ -338,6 +342,30 @@ pub trait Engine<M: Machine>: Sync + Send {
 	fn is_timestamp_valid(&self, header_timestamp: u64, parent_timestamp: u64) -> bool {
 		header_timestamp > parent_timestamp
 	}
+
+	/// Gather all ancestry actions. Called at the last stage when a block is committed. The Engine must guarantee that
+	/// the ancestry exists.
+	fn ancestry_actions<'a>(&self, _block: &M::LiveBlock, _ancestry: Box<Iterator<Item=M::ExtendedHeader> + 'a>) -> Vec<AncestryAction> {
```
rphmeier: for `authority_round` we should move the rolling-finality invocation there.

ethcore/src/engines/mod.rs
```diff
@@ -338,6 +342,30 @@ pub trait Engine<M: Machine>: Sync + Send {
...
+		Vec::new()
+	}
+
+	/// Check whether the given new block is the best block.
+	fn is_new_best(&self, new: &M::ExtendedHeader, best: &M::ExtendedHeader) -> bool;
```
rphmeier: should be called `fork_choice`

rphmeier:
The `Ancestry` in `on_new_block` isn't used and doesn't seem necessary. Where would it be useful?

rphmeier:
Where is metadata actually generated? I don't see it being called anywhere. This should happen after `on_close_block` since we will need to inspect the block's post-state.

sorpaas:
> Where is metadata actually generated? I don't see it being called anywhere.

I currently make the metadata being generated at for anything with `&mut LiveBlock`, so that means `on_new_block` and `on_close_block`. There's a function `OpenBlock::set_metadata`.

It looks like states are never changed after `on_close_block`. So we can just postfix that function to generate metadata for `vote` and `score`?

ethcore/src/blockchain/update.rs
rphmeier: "finalized" means "stable in the fork-choice rule" (i.e. if block B is finalized, `fork_choice(B, B') = B for all B'`. How can we have a best block which is not included in the fork-choice? How can we have a block with no children, which is stable in the fork-choice rule, and call some other block the best block?

rphmeier: > So a immediate finalized new block might not be the best block if it's score is less? 

This misses the entire point of casper. There is no purpose to implementing CasperFFG if we just ignore the finality information from it and continue to follow the chain with the highest total difficulty (a useless metric in the presence of finality; the only metric we would care about in that kind of hybrid system is TD-since-finalized-block).

ethcore/src/blockchain/blockchain.rs
sorpaas: impl trait is not yet stabilized? https://github.com/rust-lang/rust/issues/34511

sorpaas: I'll get it another struct implementing `Iterator`. I guess that's clearer.

ethcore/src/blockchain/update.rs
sorpaas: And in the above case, still consider `B` to be finalized already, and we insert `D`.

```
   B - C
  /
A
  \
   D
```

Even in the situation where `D` has `is_best_block` to be true (because for that we don't yet consulted the finalization rules), we'll never reorg from `C` to `D`. The actual best block will still be `C`.

rphmeier: > If we use the finalized definition in the paper, this is theoretically allowed by CasperFFG 

This is not possible except with some supermajority of validators violating a slashing condition. We should ignore that for the time being and allow "de-finalization" and recovery processes for handling malicious validator sets in future work. For the purposes of the initial work here, let's only consider engines with a fault-tolerance assumption that doesn't permit safety failures.

> D has is_best_block to be true
 
`D` can't have `is_best_block` be true because `D` is no longer a member of the fork-choice. It's a contradiction. `D` might have the highest total difficulty when counting back to the genesis but it is not the best block.

> because we consider fork_choice and finalization rules separately

They shouldn't be considered separately because the fork choice depends on finality

sorpaas: I think maybe we rename `is_best_block` to `is_best_block_pre_finalization`?

Check `is_best_block_pre_finalization` and `is_finalized` separately, and then reach a conclusion of the final `fork_choice` fits most of our use cases. The thing I worry if we move the whole finalization check to `Engine::fork_choice`, is that I think with that we also need to many reorg rule checks from `BlockChain` to `Engine`, which will make `Engine` big.

sorpaas: > `D` can't have `is_best_block` be true because `D` is no longer a member of the fork-choice.

I agree. However, I don't think that should be in `Engine`. I would favor to put that logic to `BlockChain` or `Client` in the future once we implement "finalization pruning`, because if we know a block is outside of the finalization branch, we can discard it without checking any `Engine` rules.

rphmeier: > is_best_block_pre_finalization

this is still inaccurate. it's more of a `is_best_block_assuming_nothing_has_ever_been_finalized` and I don't see how that's actually useful.

rphmeier: `Engine::fork_choice(old_head, new_head) -> enum Choice { Left, Right }`.
and the blockchain makes sure no finalized blocks are reverted. Where does `is_best` even come into this?

sorpaas: > the blockchain calls this and then makes sure no finalized blocks are reverted

Yeah that's the current behavior. At least I got that part right. :smile: 

* I still think that what might be confusing there is naming. So I'm going to rename `is_new_best` to `fork_choice`.
* Before we implement finalization pruning, we may still want to process a block even if it is confirmed to be on a branch. This avoids breaking too many code at once, as those blocks might still be needed somewhere. So I'm in favor of using the behavior "the blockchain calls `Engine::fork_choice` and then makes sure no finalized blocks are reverted".
* If I understand this correctly, the new block's finalization status `new.is_finalized` does not participate in the finalization check. Instead, it's the current best block's (and ancestry's) finalization status. We check whether reorg to the new block is blocked by finalization status of current best block (or its ancestry). The scope of blocks we can reorg to is set before the new block, and regardless of the new block's finalization status. So I think it's safe just to assume that it's an `Engine` logic failure (or really huge network attack) if we ever meet the cases discussed in this thread.

rphmeier: > is_new_best to primitive_fork_choice

I don't see the purpose of that at all. What is the definition of a "primitive fork choice" and how do you derive it from a general fork choice function? Because it seems to me as just a way to introduce bug-prone complexity into the fork choice logic. What is the downside of removing it?

> the new block's finalization status new.is_finalized does not participate in the finalization check

it should, why wouldn't it? is it not part of the chain?

sorpaas: > it should, why wouldn't it? is it not part of the chain?

Using the above example:

```
   B - C
  /
A
  \
   B' - C' - D' - E'
```

The reason why we don't reorg to `E'` is only because `B` is finalized. That is, we check `C` (current best block) and `C`'s ancestry finalization status. Yeah usually on `E'` branch there won't be any finalized block, given we don't have huge network attack. But the decisive factor is on `C`'s branch, not `E'`. Is my understanding on this point correct?

> What is the definition of a "primitive fork choice" and how do you derive it from a general fork choice function?

I think it's equivalent we write:

```
BlockChain::import_block(head, |old_head, new_head| MyEngine.fork_choice(old_head, new_head))
```

Or we write:

```
let primitive_fork_choice = MyEngine.fork_choice(old_head, new_head);
BlockChain::import_block(head, primitive_fork_choice);
```

If we combine the functions together, that would potentially mean we allow "finalization" to mean different things for different Engine. I just think that may not be a good idea. For Casper, `primitive_fork_choice` is basically Casper score comparison.

Another thing is, I don't think we can remove `primitive_fork_choice`. `fork_choice == primitive_fork_choice + finalization_rules` (basically as you said, "the blockchain calls this and then makes sure no finalized blocks are reverted"). But if we combine them together, I think we would probably need to move reorg logic to `Engine`, and my suggestion is that we try to break Engine trait before we make it too big..

rphmeier: > The reason why we don't reorg to E' is only because B is finalized. That is, we check C (current best block) and C's ancestry finalization status. Yeah usually on E' branch there won't be any finalized block, given we don't have huge network attack. But the decisive factor is on C's branch, not E'. Is my understanding on this point correct?

This is correct but I don't see what it has to do with handling instant finality correctly. Attempting a general solution to handling of fault-tolerance violations (e.g. conflicting finalizations) is also probably going to be worse than doing nothing at all and as I suggest we should leave that for future work.

> let primitive_fork_choice = MyEngine.fork_choice(old_head, new_head);
It shouldn't be a "primitive" fork choice, it should be _the_ fork choice, or at least defined as a sub-decision after ensuring no finalized blocks would be reverted (which limits the class of fork-choice rules we can implement but not meaningfully until later-stage sharding).

There is a subtle but important difference between checking for reversion of finalized blocks before and after performing the remainder of the fork-choice operation. The former is more correct, but will behave the same as the latter in all cases except for malicious instant finality. Neither requires an `is_new_best` parameter.

rphmeier: anyway, I guess the current solution is workable as long as it's well-documented that it is broken in that case. in the future we probably will need to move the reorg-checking stuff to engine anyway.

rphmeier: As a concrete example of a fork-choice rule which doesn't fall into this category, imagine a situation where `Parity` nodes are running a parachain for Polkadot and they use a `PolkadotEngine` that wraps a polkadot node. The fork choice for `PolkadotEngine` would be to follow the most recently finalized polkadot block and not to follow the blockchain's notion of finality in any way.

sorpaas: Okay got it. Yeah indeed it won't cover that instant finalization case..

Anyway, I'll add more documentations on `Engine::fork_choice` for how that's currently working.

ethcore/src/blockchain/blockchain.rs
rphmeier: my mistake, thought it was stabilized

sorpaas:
I changed `Engine::fork_choice`'s signature to contain `FinalizationInfo`. This should be much clearer. And now, the fork choice made by engine is the final fork choice. We got rid of the concept of "primitive fork choice".

sorpaas:
Hmm looks like the CI failed with an internal error. Needs restarting...

ethcore/src/client/client.rs
```diff
@@ -545,7 +583,17 @@ impl Importer {
 		);
 
 		state.journal_under(&mut batch, number, hash).expect("DB commit failed");
-		let route = chain.insert_block(&mut batch, block_data, receipts.clone());
+
+		for ancestry_action in ancestry_actions {
+			let AncestryAction::MarkFinalized(ancestry) = ancestry_action;
+			chain.mark_finalized(&mut batch, ancestry).expect("Engine's ancestry action must be known blocks; qed");
```
andresilva: This block might be part of a fork (`fork_choice` returns `ForkChoice::Old`) and yet it will be able to finalize blocks that are part of the canon chain.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -934,45 +998,58 @@ impl BlockChain {
...
-						})
+			location: match extras.fork_choice {
+				ForkChoice::New => {
+					// On new best block we need to make sure that all ancestors
+					// are moved to "canon chain"
+					// find the route between old best block and the new one
+					match route.blocks.len() {
+						0 => BlockLocation::CanonChain,
+						_ => {
+							let retracted = route.blocks.iter().take(route.index).cloned().collect::<Vec<_>>().into_iter().collect::<Vec<_>>();
```
andresilva: `assert(retracted.iter().flat_map(|b| self.block_details(b)).all(|b| !b.is_finalized))`

Not sure if we want to add the assert, but this post-condition should hold.

sorpaas: I think the case is what @rphmeier mentioned above. We may not want to assume too much about how finalization works. Yeah Casper and Aura all have that condition to be true, but there may be Engine that wants to just follow the current finalized blocks, regardless of finalization status of other blocks. And it is why we just pass FinalizationInfo to Engine::fork_choice, but not making the decision on BlockChain struct.

sorpaas:
Tests pass on my local machine. Can we restart it?

ethcore/src/client/client.rs
rphmeier: If `B` is finalized there should not be any fork-choice which does not include `B` in its ancestry. And why couldn't it be that `E` finalizes `D`? I don't see why that's a reasonable assumption.

ethcore/src/engines/mod.rs
```diff
@@ -52,15 +52,34 @@ use spec::CommonParams;
...
+	/// Choose the current best block.
+	Old,
+}
+
+/// Finalization related information.
+#[derive(Debug, PartialEq, Eq)]
+pub struct FinalizationInfo {
+	/// Whether the route from current best block to common ancestor is finalized.
+	pub is_old_route_finalized: bool,
+	/// Whether the route from new block's parent to common ancestor is finalized.
```
rphmeier: do these comments refer to the _whole_ route or not? These properties don't seem particularly useful to me.

machine/src/lib.rs
```diff
@@ -115,3 +158,10 @@ pub trait WithBalances: Machine {
 		_indirect: &[(Address, U256)],
 	) -> Result<(), Self::Error> { Ok(()) }
 }
+
+/// Metadata with total difficulty information.
+pub trait TotalDifficulty {
+	type Value: Ord + Add;
```
rphmeier: this seems redundant with `TotalScore`

ethcore/src/engines/mod.rs
sorpaas: So with the following route:

```
   B - C
  /
A
  \
   D - E - (F)
```

If `C` is the current best block, and we insert block `F` then:

* `is_old_route_finalized` check finalization in `B - C`.
* `is_new_parent_route_finalized` check finalization in `D - E`, without `F`.

We only use `is_old_route_finalized` for most cases (like in Casper), and I agree `is_new_parent_route_finalized` is probably not that useful. But I just included it because we have this information, and it nearly hurts nothing to expose that.

I'm both okay whether we include it or not.

machine/src/lib.rs
sorpaas: Ooops looks like that trait is not used anywhere. Removing.

ethcore/src/client/client.rs
andresilva: So using @sorpaas' example, my original issue was that e.g. `E` could finalize `A` while the current best block is `C` and for some reason that fork (canon) hasn't yet finalized `A`.

sorpaas: @andresilva I think that might be okay. If A is finalized, no matter we reorg from `C` to `E` or not doesn't break the stricter version of finalization rules -- even if A is finalized, we don't move pass any finalized blocks in the sequence `C -> B -> (A) -> D -> E`.

rphmeier: can you provide a definition for what you mean "stable" finality info to mean? I am still concerned by the statement
> That means, if an engine decides (for whatever reason) reorg passing B and set the new best block to E

where engines are allowed to reorganize past finalized blocks.

sorpaas: I use the that term "stable" finality only because Engine don't handle reorg, and we need to make proper responsibility separation of that. So we can define "stable" finality:

* Going from genesis block to all current known branch, the blocks' finality shouldn't be affected by which branch we go first. Or in a weaker form, the Engine's fork choice related to blocks' finality shouldn't be affected by which branch we go first.

Yes. In most cases, an Engine (like Casper and Aura) shouldn't reorg pass a finalized block (and clearly, never reorg pass a finalized block is a subset of "stable" finality). However, here because we provided the `FinalizationInfo` to `Engine::fork_choice` to allow it *potentially* reorg pass a finalized block, I just thought it would be good to give a note there about under what cases this reorg-pass-a-finalized-block action is acceptable. For a particular `Engine`, the property should be much specialized and clear, and we would nearly never need the term "stable" finality.

rphmeier: This PR shouldn't be merged if it is using a custom definition of finality which doesn't align with any conventional usage. Allowing engines to reorganize past finalized blocks is beyond the scope of this PR and should be removed.

sorpaas: @rphmeier Then let's change `fork_choice` not to take `FinalizationInfo` then? We just need to reset several commits. But in that case we'll have the term "primitive fork choice" again, because now not-reorg-pass-finalized-block is a hard-set rule.

https://github.com/paritytech/parity/pull/8401#issuecomment-383263561

ethcore/src/engines/mod.rs
rphmeier: `is_old_route_finalized` checks how many blocks back?
`is_new_parent_route_finalized` is mutually exclusive with `is_old_route_finalized`. This should be an enum.

sorpaas: It only checks blocks in the reorg route. So as in the example https://github.com/paritytech/parity/pull/8401#discussion_r185317727, `is_old_route_finalized` only checks `B - C` -- blocks from best to common ancestor. `is_new_parent_route_finalized` checks `D - E` -- inserting block to common ancestor. They're not mutually exclusive.

ethcore/src/client/client.rs
rphmeier: it doesn't hurt instant finality as long as we do the reversion check before calling out to the engine.

sorpaas: Okay I'll change to that version then. But note that this would be equivalent to reverting back to what we discussed in this thread (https://github.com/paritytech/parity/pull/8401#discussion_r182483430), and because we don't pass `FinalizationInfo` to `fork_choice` any more, it handles a strict subset:

```
impl<A: EngineNewDef> EngineCurrentDef for A {
  fn fork_choice(&self, new: ExtendedHeader, best: ExtendedHeader, finalization: FinalizationInfo) -> ForkChoice {
    if finalization.is_old_route_finalized {
      return ForkChoice::Old;
    }
    EngineNewDef::fork_choice(&self, new, best)
  }
}
```

And I don't think the new version would handle the situation described here https://github.com/paritytech/parity/pull/8401#discussion_r182474954 because in that case we don't follow the notion of finalization of current blockchain and finalize some blocks instantly, those instant finalized blocks won't reorged if there's finalized blocks in the best-to-common-ancestor route.

rphmeier: for those further cases we will need to introduce a notion of hierarchical finality, but that's beyond the scope of the PR. The definitions which I've given so far about stability in the fork-choice are always based on some assumptions holding (like a 2f+1 honest majority assumption). In a future PR we will allow declaring hierarchies of finality with each level being based on stronger assumptions than the one before it. This will allow engines to handle reversions of lower levels of finality for those above it. But this simple black-and-white definition is all we need for the moment.

sorpaas: Okay. I changed the behavior to what's described above.

sorpaas:
Just some notes regarding now this ended up making the finalization rule works, in case the above discussion threads are too long:

* The strict version of finalization is used -- No reorg can move pass any finalized blocks.
* The Client will first check whether the finalization rule is held -- if there's finalized blocks preventing the reorg, `Engine::fork_choice` will not be called and the new block will directly become a branch. If no finalized block prevent the reorg, then it will call `Engine::fork_choice` to decide which branch to choose.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -689,7 +734,9 @@ impl BlockChain {
 		Some(TreeRoute {
 			blocks: from_branch,
 			ancestor: current_from,
-			index: index
+			index: index,
+			is_from_route_finalized: is_from_route_finalized,
+			is_to_route_finalized: is_to_route_finalized,
```
rphmeier: still shouldn't have both of these

ethcore/src/blockchain/blockchain.rs
```diff
@@ -769,11 +816,14 @@ impl BlockChain {
 				location: BlockLocation::CanonChain,
 			};
 
+			// TODO [sorpaas] support wrap sync insertion of finalization and metadata.
```
rphmeier: warp*

ethcore/src/blockchain/blockchain.rs
```diff
@@ -898,7 +948,22 @@ impl BlockChain {
 	/// Inserts the block into backing cache database.
 	/// Expects the block to be valid and already verified.
 	/// If the block is already known, does nothing.
-	pub fn insert_block(&self, batch: &mut DBTransaction, bytes: &[u8], receipts: Vec<Receipt>) -> ImportRoute {
+	pub fn insert_block(&self, batch: &mut DBTransaction, bytes: &[u8], receipts: Vec<Receipt>, extras:ExtrasInsert) -> ImportRoute {
```
rphmeier: space between `extras` and type

ethcore/src/blockchain/blockchain.rs
```diff
@@ -942,45 +1007,58 @@ impl BlockChain {
...
-			} else {
-				BlockLocation::Branch
-			}
+				},
+				ForkChoice::Old => BlockLocation::Branch,
+			},
 		}
 	}
 
+	/// Mark a block to be considered finalized. Panic if the block hash is not found.
```
rphmeier: looks like it just returns `None` if it's not found. `Option<()>` for a function which can error is weird though

ethcore/src/client/client.rs
```diff
@@ -467,6 +470,44 @@ impl Importer {
...
+			ExtendedHeader {
+				parent_total_difficulty: details.total_difficulty - *header.difficulty(),
+				is_finalized: details.is_finalized,
+				metadata: details.metadata,
+
+				header: header,
+			}
+		};
+
+		let route = chain.tree_route(best_hash, *parent).expect("blocks being imported always within recent history; qed");
```
rphmeier: this isn't an accurate reason for it not to panic. should refer to the fact that we don't keep forks without keeping a common ancestor

rphmeier: i think I saw another variant of this expectation somewhere else

ethcore/src/blockchain/blockchain.rs
rphmeier: (the only place where this does make sense is if `prospective.finalized = true`). But this information is already available to the caller.

sorpaas: Yeah I think I got what you meant. If we use the finality definition settled down earlier, then `is_to_route_finalized` should always be `false`. I'm going to remove this variable.

andresilva:
Just finished reviewing the latest changes, LGTM. I'll extend AuRa with `ancestry_actions` using `RollingFinality` to finalize ancestry.

PR #1483, opened 2016-06-29
DAO hard-fork

gavofyork:
In accordance with the specification given on https://gist.github.com/gavofyork/856b27cf1a482585692359e28d3ae4cb

PR #8406, opened 2018-04-15
Non-canonical state-change parameter

5chdn:
Add support for an irregular state change in chain spec (CommonParams). Add example implementation for [EIP-999](http://eips.ethereum.org/EIPS/eip-999), including test chain spec. Could be used as groundwork to refactor the DAO fork changes ([EIP-779](http://eips.ethereum.org/EIPS/eip-779)) as well, ref. #1483.


ethcore/src/machine.rs
```diff
@@ -193,6 +193,30 @@ impl EthereumMachine {
 						.and_then(|b| state.transfer_balance(child, beneficiary, &b, CleanupMode::NoEmpty))?;
 				}
 			}
+
+			if let Some(irregular_changes) = self.params.irregular_state_changes.get(&block.header().number()) {
+				let state = block.state_mut();
+				for &(address, ref irregular_account) in irregular_changes {
+					match irregular_account {
+						&IrregularStateChangeAccount::Set {
```
rphmeier: single arm match can be replaced by an irrefutable destructuring.
`let &IrregularStateChangAccount { none, balance, ref code, ref storage } = irregular_account;`

rphmeier: > if the IrregularStateChangeAccount enum changed to have more variants, it cannot detect it and throw type errors.

that's not true. only irrefutable pattern matches can be used in a let-binding. adding another variant would make that pattern match no longer irrefutable.

ethcore/res/ethereum/multisig_test.json
```diff
@@ -0,0 +1,86 @@
+{
```
tomusdrw: The file doesn't seem to be used anywhere. Would be really awesome to get a unit test for this.

ethcore/src/machine.rs
```diff
@@ -193,6 +193,26 @@ impl EthereumMachine {
 						.and_then(|b| state.transfer_balance(child, beneficiary, &b, CleanupMode::NoEmpty))?;
 				}
 			}
+
+			if let Some(irregular_changes) = self.params.irregular_state_changes.get(&block.header().number()) {
```
tomusdrw: This block does not depend on `ethash_params`, does it have to be inside that particular `if`? If yes could you please add a comment explaining that.

ethcore/src/spec/spec.rs
```diff
@@ -123,6 +156,8 @@ pub struct CommonParams {
 	pub max_code_size_transition: BlockNumber,
 	/// Transaction permission managing contract address.
 	pub transaction_permission_contract: Option<Address>,
+	/// Irregular state change list.
```
tomusdrw: Would be good to mention that the irregular state changes are applied at the very beginning of the block to avoid confusion.

sandakersmann:
The community has rejected bailouts and developer corruption.

rphmeier:
@sandakersmann not particularly constructive on the code changes or in general.

ethcore/src/machine.rs
rphmeier: definitely should be moved outside. other engines might have irregular state changes too.

sandakersmann:
@rphmeier I agree that bailouts belong on private blockchains and not on public blockchains. I just wanted to make it clear that you will find yourself on a minority chain if you merge this to the Parity client for the public Ethereum chain.

rphmeier:
The PR discussion should be used for technical comments only and not as a soapbox. Please take discussion on use-cases of this PR to the correct places.

5chdn:
This PR is not changing the foundation chain spec. This has to be discussed somewhere else.

rphmeier:
I think we should implement this anyway, it's a good refactoring of the codebase.

5chdn:
@sorpaas had some great ideas how to implement this in the `accounts` section rather than introducing irregular state changes, i.e.,

```
"accounts": {
  "863DF6BFa4469f3ead0bE8f9F2AAE51c91A907b4": { "code": "<multisig_code>", "activate_at": 6000000 }
}
```

This would allow rewriting the DAO fork in a very smooth way, and both Casper FFG (EIP-1011) and Blockhash refactoring (EIP-210) could benefit from such changes.

rphmeier:
That approach sounds fine too, as long as it's easy to make multiple changes to one account at different points. The current method makes it really easy to see which changes are grouped together to be applied at a specific block.

PR #8409, opened 2018-04-16
Make dependency `tempdir` test-only

niklasad1:


niklasad1:
This doesn't work didn't realize the `helper module` was exported!

PR #8410, opened 2018-04-16
Check finalized block in blockchain::Blockchain

sorpaas:
* Takes BlockDetails to contain `finalized` and `metadatas` in a backward compatible way -- if finalized is false and metadata is empty, it is stored as the old 4 value RLP. Otherwise, it's stored as the extended 6 value RLP.
* Adds `Blockchain::mark_finalized` and `Blockchain::update_metadata` to update finalized value and metadata for any block (but it's not currently used anywhere).
* Changes `Blockchain::block_info` to refuse re-org if it goes pass a finalized block.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -1006,8 +1006,9 @@ impl BlockChain {
 	}
 
 	/// Update metadata detail for an existing block.
-	pub fn update_metadata(&self, batch: &mut DBTransaction, block_hash: H256, metadata: HashMap<Bytes, Bytes>) -> Result<(), MetadataError> {
+	pub fn update_metadata<T: Into<HashMap<Bytes, Bytes>>>(&self, batch: &mut DBTransaction, block_hash: H256, metadata: T) -> Result<(), MetadataError> {
 		let mut block_details = self.block_details(&block_hash).ok_or(MetadataError::UnknownBlock)?;
+		let metadata: HashMap<Bytes, Bytes> = metadata.into();
 		for (key, value) in metadata {
 			block_details.metadata.insert(key, value);
 		}
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -965,16 +965,28 @@ impl BlockChain {
 
 				assert_eq!(number, parent_details.number + 1);
 
-				match route.blocks.len() {
-					0 => BlockLocation::CanonChain,
-					_ => {
-						let retracted = route.blocks.iter().take(route.index).cloned().collect::<Vec<_>>().into_iter().collect::<Vec<_>>();
-						let enacted = route.blocks.into_iter().skip(route.index).collect::<Vec<_>>();
-						BlockLocation::BranchBecomingCanonChain(BranchBecomingCanonChainData {
-							ancestor: route.ancestor,
-							enacted: enacted,
-							retracted: retracted,
-						})
+				// If anything on the tree route is finalized, refuse to reorg. The common ancestor can be finalized,
+				// but nothing in between.
+				let finalized_block = route.blocks.iter()
+					.map(|route_hash| self.block_details(&route_hash).unwrap_or_else(|| panic!("Invalid block hash in tree route: {:?}", route_hash)))
+					.enumerate()
+					.take_while(|&(i, _)| i < route.index)
+					.any(|(_, details)| details.finalized);
+
+				if finalized_block {
+					BlockLocation::Branch
+				} else {
+					match route.blocks.len() {
+						0 => BlockLocation::CanonChain,
+						_ => {
+							let retracted = route.blocks.iter().take(route.index).cloned().collect::<Vec<_>>().into_iter().collect::<Vec<_>>();
+							let enacted = route.blocks.into_iter().skip(route.index).collect::<Vec<_>>();
+							BlockLocation::BranchBecomingCanonChain(BranchBecomingCanonChainData {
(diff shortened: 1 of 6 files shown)


sorpaas:
In the short term we probably need to keep `total_difficulty` in `BlockDetails`. Moving it out to whatever metadata format we decide would work for full node, but breaks wrap sync and light client. This PR also uses `BlockDetails` for per-block metadata storage. Using this or just use the engine's new global metadata storage would both work, but I just wary whether there's any pruning operations that might remove block extra information. Need to check on this.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -961,16 +965,28 @@ impl BlockChain {
...
-						let enacted = route.blocks.into_iter().skip(route.index).collect::<Vec<_>>();
-						BlockLocation::BranchBecomingCanonChain(BranchBecomingCanonChainData {
-							ancestor: route.ancestor,
-							enacted: enacted,
-							retracted: retracted,
-						})
+				// If anything on the tree route is finalized, refuse to reorg. The common ancestor can be finalized,
+				// but nothing in between.
+				let finalized_block = route.blocks.iter()
+					.map(|route_hash| self.block_details(&route_hash).unwrap_or_else(|| panic!("Invalid block hash in tree route: {:?}", route_hash)))
```
tomusdrw: Wouldn't it be better to include `finalized` already in `tree_route`? Seems that we query block details there anyway.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -979,6 +995,34 @@ impl BlockChain {
 		}
 	}
 
+	/// Mark a block to be considered finalized. Panic if the hash does not exist.
+	pub fn mark_finalized(&self, batch: &mut DBTransaction, block_hash: H256) {
+		let mut block_details = self.block_details(&block_hash).unwrap_or_else(|| panic!("Invalid block hash: {:?}", block_hash));
```
tomusdrw: Maybe better to return `Result` instead of panicking?

ethcore/src/blockchain/blockchain.rs
```diff
@@ -979,6 +995,34 @@ impl BlockChain {
...
+		block_details.finalized = true;
+
+		self.prepare_details_update(batch, block_hash, block_details);
+	}
+
+	/// Update metadata detail for an existing block.
+	pub fn update_metadatas(&self, batch: &mut DBTransaction, block_hash: H256, metadatas: HashMap<Bytes, Bytes>) {
```
tomusdrw: Wouldn't it be better to have metadata in some more structured way? Seems that `HashMap<Bytes, Bytes>` is even worse than `HashMap<String, String>` (the latter is at least human readable)

ethcore/src/blockchain/blockchain.rs
```diff
@@ -979,6 +995,34 @@ impl BlockChain {
...
+		let mut block_details = self.block_details(&block_hash).unwrap_or_else(|| panic!("Invalid block hash: {:?}", block_hash));
+		for (key, value) in metadatas {
+			block_details.metadatas.insert(key, value);
+		}
+
+		self.prepare_details_update(batch, block_hash, block_details);
+	}
+
+	/// Prepares extras block detail update.
+	fn prepare_details_update(&self, batch: &mut DBTransaction, block_hash: H256, block_details: BlockDetails) {
```
tomusdrw: Does it just prepare the update or write it to cache? Seems there is already `prepare_block_details_update` method in this struct and it returns the `HashMap`, might be a bit misleading to have that one here as well.

ethcore/src/blockchain/blockchain.rs
sorpaas: Maybe `HashMap<String, Bytes>`? For many cases (for example, to store Casper score info `U256`), the value itself is readable anyway.

ethcore/src/blockchain/blockchain.rs
sorpaas: That was meant to be a stripped version of `prepare_update`, but yeah looks confusing. I'll change the name.

ethcore/src/blockchain/blockchain.rs
rphmeier: agree, or at least prove it won't happen

ethcore/src/blockchain/blockchain.rs
rphmeier: also: `metadata` is already plural

tomusdrw: What I mean is maybe to have `T: TryFrom<HashMap<Bytes, Bytes>>` or sth (if we want to be super generic), so that the caller can decide on the expected type of metadata, otherwise we might be passing a lot of weakly-typed structures in the codebase and imho would be best to avoid that.

Alternatively maybe an `enum` with possible types of stored metadata if we are ok with having less generic solution and a single place where all kinds of metadata is defined.

sorpaas:
I'm going to move this to #8401 because some of the changes are related and I might create unnecessary merge conflicts. I think I got a design now, and will explain it soon.

ethcore/src/blockchain/blockchain.rs
sorpaas: The rest of `BlockChain` struct is also using panic for pub fns when a block hash is not found. I think for our case is that we're going to use `mark_finalized` here for results returned from Engine, and it will only call this function for already known block hashes. Otherwise, it indicates a logic error in our code. So I think it would be fine to use panic given we indicated the panic condition in the documentation.

PR #8412, opened 2018-04-16
Parity as a library

### LINKED ISSUE #4842: Parity-as-a-library

tomaka:
Close #4842 

Splits `main.rs` in two, `lib.rs` and `main.rs`, and reorganizes things around so that `lib.rs` contains what is necessary to use Parity as a library, and `main.rs` contains the code of the Parity binary.

The way `libparity` is to be used is that you build a `Configuration` object from the CLI arguments, then pass it to the `start` function. In follow-up PRs we can add other ways to build a `Configuration` if necessary, such as from a TOML string. The `start` function then returns an object that represents the client currently running in background threads.

This PR also adds the `parity-clib` crate, which exports symbols usable from C (or any other programming language that can bind to C). A `parity.h` file is included. When using `parity-clib`, the auto-updater is always automatically disabled at runtime.

I didn't provide any full example C code (there are some snippets in the `parity.h`), but it should be pretty easy to write:

```c
#include <cstddef>
#include <parity.h>

void on_restart(void*, const char*, size_t) {}

int main() {
    ParityParams cfg = { 0 };
    cfg.on_client_restart_cb = on_restart;

    const char* args[] = {"--light"};
    size_t str_lens[] = {7};
    if (parity_config_from_cli(args, str_lens, 1, &cfg.configuration) != 0) {
        return 1;
    }

    void* parity;
    if (parity_start(&cfg, &parity) != 0) {
        return 1;
    }

    parity_destroy(parity);
    return 0;
}
```

There are still elements to be improved after this PR:
- Users may not always want to enable logging, or may want to obtain the logs emitted by Parity. This is currently not possible, as everything is directly printed to stdout/stderr.
- Users may want to pass a custom database backend or network backend (for WASM+browser stuff).
- More precise success/error codes?
- Some elements, such as loading the UI, could be moved to `main.rs` instead of `lib.rs`.
- We could add a way to directly send JSON-RPC requests without having to go through a socket.

Appropriate issues should be opened after this PR is merged.

This PR also removes the `parity/stratum.rs` file, and everything behind `#[cfg(feature = "stratum")]`, which was unused.


parity-clib/parity.h
```diff
@@ -0,0 +1,89 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef _PARITY_H_INCLUDED_
+#define _PARITY_H_INCLUDED_
+
+#include <cstddef>
```
niklasad1: This is `cpp` version the library why not `#include <stddef.h>`?

parity-clib/parity.h
```diff
@@ -0,0 +1,89 @@
...
+	/// Note that calling `parity_start` will destroy the configuration object (even on failure).
+	void *configuration;
+
+	/// Callback function to call when the client receives an RPC request to change its chain spec.
+	///
+	/// Will only be called if you enable the `--can-restart` flag.
+	///
+	/// The first parameter of the callback is the value of `on_client_restart_cb_custom`.
+	/// The second and third parameters of the callback are the string pointer and length.
+	void (*on_client_restart_cb)(void *, const char *, size_t);
```
niklasad1: Name the parameters for better readability!

parity-clib/parity.h
```diff
@@ -0,0 +1,89 @@
...
+/// to `len`. The strings don't need to be zero-terminated.
+///
+/// Returns 0 on success, and non-zero on error.
+///
+/// # Example
+///
+/// ```no_run
+/// void* cfg;
+/// const char *args[] = {"--light"};
+/// size_t str_lens[] = {7};
```
niklasad1: Consider to add two `args` in the example because it might not be obvious what `str_lens` is used for i.e, to length of each `arg`

parity-clib/parity.h
```diff
@@ -0,0 +1,89 @@
...
+/// }
+/// ```
+///
+int parity_config_from_cli(char const *const *args, size_t const *arg_lens, size_t len, void **out);
+
+/// Destroys a configuration object created earlier.
+///
+/// **Important**: You probably don't need to call this function. Calling `parity_start` destroys
+/// 				the configuration object as well (even on failure).
+void parity_config_destroy(void *);
```
niklasad1: Docs for `void *`

parity-clib/parity.h
```diff
@@ -0,0 +1,89 @@
...
+
+/// Starts the parity client in background threads. Returns a pointer to a struct that represents
+/// the running client. Can also return NULL if the execution completes instantly.
+///
+/// **Important**: The configuration object passed inside `cfg` is destroyed when you
+/// 				call `parity_start` (even on failure).
+///
+/// Returns 0 on success, and non-zero on error.
+int parity_start(const ParityParams *, void **);
```
niklasad1: Docs for `void **`

tomaka:
> Generally looks good, but is it any particular reason why no null-checks are performed on the arguments passed to parity_xx_yy?

The idea is that we want to avoid defensive programming. If a user accidentally passes a null pointer and everything works fine, then they will not know that there is a bug in their code.

> Also, since you have removed stratum you need to remove the related cli commands and their documentation in https://github.com/paritytech/parity/blob/master/parity/cli/mod.rs

I didn't remove stratum, I only removed some unused code.
My guess is that stratum got refactored at some point, and some old code was left behind. Since this PR touches this code, I realized that it was unused and removed it.

tomaka:
@niklasad1 I cannot panic since we're in C world.

parity-clib/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[repr(C)]
+pub struct ParityParams {
+	pub configuration: *mut c_void,
+	pub on_client_restart_cb: Option<extern "C" fn(*mut c_void, *const c_char, usize)>,
+	pub on_client_restart_cb_custom: *mut c_void,
+}
+
+#[no_mangle]
+pub extern fn parity_config_from_cli(args: *const *const c_char, args_lens: *const usize, len: usize, output: *mut *mut c_void) -> c_int {
```
NikVolf: isn't `pub unsafe extern "C" fn ...` signature recommended for c api?

NikVolf:
shouldn't we add a smoky test (maybe in form of simple shell script)?

parity-clib/src/lib.rs
tomaka: According to the official docs, no: https://doc.rust-lang.org/nomicon/ffi.html#calling-rust-code-from-c
I agree that this is really debatable, though.

NikVolf: Right, i remember i was given this recommendation by Rust guys though
nvm

NikVolf:
yeah, C example which is compiled and run on the CI would be perfect
compilation and running might be simply added to `test.sh`

tomaka:
Added an example. CI fails for unknown reason most likely unrelated to this PR.

test.sh
```diff
@@ -40,8 +40,17 @@ echo "________Validate chainspecs________"
 fi
 
 
-# Running test's
+# Running tests
 echo "________Running Parity Full Test Suite________"
-
 cargo test -j 8 $OPTIONS --features "$FEATURES" --all $1
 
+# Running the C example
```
NikVolf: maybe move to dedicated shell script to keep `test.sh` clean?

tomaka: I think that having two shell scripts makes it work than just one (especially since the script is only 8 lines). But if you have a strong opinion on this I will do it.

tomaka:
For what it's worth, `test.sh` passes locally. It looks like CI has a problem with several pull requests :confused:

5chdn:
Clearly a CI issue. I restarted it _again_. @General-Beck

5chdn:
./test.sh: 51: ./test.sh: cmake: not found

@General-Beck

General-Beck:
`cmake` added to main build image

PR #8413, opened 2018-04-16
Fix an obsolete TODO where `miner` is alias of `author`

sorpaas:
Fix an obsolete TODO item, and document `author` is alias of `miner`.


tomusdrw:
`miner` is strictly related to PoW, `author` is strictly more generic and has been supported by Parity since the very beginning. IMHO we should rather move away from `miner` not `author`. I'd say that just spec needs to be updated.

sorpaas:
I submitted a PR to wiki. There's still an obsolete TODO comment in the code that needs to be removed.

PR #8419, opened 2018-04-17
Block reward contract

### LINKED ISSUE #7879: Aura: support rewarding contract defined in spec
Give block reward to a contract listed in the spec file instead of only to the block author (https://github.com/paritytech/parity/pull/7605#issuecomment-363503197).

andresilva:
Add support for having a contract calculate the block rewards. The engine passes a list of benefactors and reward types to this contract which then returns a list of addresses and respective rewards. This is the contract solidity interface:

```solidity
interface BlockReward {
    // produce rewards for the given benefactors, with corresponding reward codes.
    // only callable by `SYSTEM_ADDRESS`
    function reward(address[] benefactors, uint16[] kind)
		external
		returns (address[], uint256[]);
}
```

Currently I only added support for this in the `AuthorityRound` engine.
Closes #7879.


ethcore/res/authority_round_block_reward_contract.json
```diff
@@ -0,0 +1,61 @@
...
...1803590602001919091929080359060200190820180359060200191909192905050610125565b604051808060200180602001838103835285818151815260200191508051906020019060200280838360005b838110156100cd5780820151818401526020810190506100b2565b50505050905001838103825284818151815260200191508051906020019060200280838360005b8381101561010f5780820151818401526020810190506100f4565b5050505090500194505050505060405180910390f35b61012d610264565b610135610278565b61013d610278565b600073fffffffffffffffffffffffffffffffffffffffe73ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff1614151561018d57600080fd5b85859050888890501415156101a157600080fd5b878790506040518059106101b25750595b90808252806020026020018201604052509150600090505b815181101561021d5785858281811015156101e157fe5b9050602002013561ffff166103e80161ffff16828281518110151561020257fe5b906020019060200201818152505080806001019150506101ca565b878783828280806020026020016040519081016040528093929190818152602001838360200280828437820191505050505050915090915093509350505094509492505050565b602060405190810160405280600081525090565b6020604051908101604052806000815250905600a165627a7a723058203ec48a056fdc2e3567414ab18ad3da7a272e23cc2ea7c5f9c8f38cbcf569fdaf0029"
```
rphmeier: I guess it's not possible to add a comment here linking to some source? IMO we should keep test specs as localized as possible instead of putting them all in the `res` folder. Then we can annotate the loading code with a link to the contract source.

ethcore/src/engines/block_reward.rs
```diff
@@ -0,0 +1,170 @@
...
+
+use ethabi;
+use ethabi::ParamType;
+use ethereum_types::{H160, Address, U256};
+
+use block::ExecutedBlock;
+use error::Error;
+use machine::EthereumMachine;
+
+pub type SystemCall<'a> = FnMut(Address, Vec<u8>) -> Result<Vec<u8>, String> + 'a;
```
rphmeier: grumble: missing docs on this and other members of the module

ethcore/src/engines/block_reward.rs
```diff
@@ -0,0 +1,170 @@
...
+		let input = reward.input(
+			benefactors.iter().map(|&(address, _)| H160::from(address)),
+			benefactors.iter().map(|&(_, ref reward_kind)| u16::from(*reward_kind)),
+		);
+
+		let output = caller(self.address, input)
+			.map_err(Into::into)
+			.map_err(::engines::EngineError::FailedSystemCall)?;
+
+		// since this is a non-constant call we can't use ethabi's function output
```
rphmeier: is there an issue filed about that?

rphmeier:
We should absolutely link to the source of the contract used as close as possible to the tests themselves.

minor grumble: a transition or `activate_at` parameter for the reward contract.

ethcore/src/engines/block_reward.rs
andresilva: I don't think so, there's also no general way to do it since it involves creating a transaction and the return data is not included in the receipt (we could trace the transaction).

andresilva:
I've added the missing docs and added a transition block parameter to enable the reward contract. The example block reward contract that's used in the tests is here: https://github.com/parity-contracts/block-reward/blob/master/contracts/ExampleBlockReward.sol (I added the link in a comment where the spec is loaded). Currently the test spec is shared for the block reward contract client and also for an aura test. 

I agree that it would be better to not have all the test specs in a single folder, maybe we can split this up: https://github.com/paritytech/parity/blob/master/ethcore/src/spec/spec.rs#L809 (also it seems like they're being included in the binary 😟). Should we deal with that in another PR?

rphmeier:
yeah, that's beyond the scope of this one but definitely worth a cleanup.

ethcore/src/engines/block_reward.rs
```diff
@@ -0,0 +1,184 @@
...
+use super::SystemCall;
+
+use_contract!(block_reward_contract, "BlockReward", "res/contracts/block_reward.json");
+
+/// The kind of block reward.
+/// Depending on the consensus engine the allocated block reward might have
+/// different semantics which could lead e.g. to different reward values.
+#[repr(u8)]
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub enum RewardKind {
```
andresilva: We already have a `RewardType` which is used for tracing but it only distinguishes between `Block` and `Uncle` rewards.

andresilva:
Tracked in #8424.

ethcore/src/engines/block_reward.rs
andresilva: The reward contract might have internal state that needs to be updated.

rphmeier: which can be used to track total supply if you have a reward contract from the beginning. Or even just to maintain a record of recent rewards if you fork to include one.

andresilva:
Yes it's not possible for the contract to emit events, this is a limitation of how we currently deal with "system" transactions (hopefully this will be fixed in the future).

5chdn:
Needs docs :p

PR #8421, opened 2018-04-17
More changes for Android

### LINKED ISSUE #8317: Android binary depends on libc++_shared.so
After #8036 Parity can compile for Android using the supplied Dockerfile (in `docker/android`)

However the final binary depends on `libc++_shared.so`, which regular Android devices don't have by default. We need to find out where that dependency comes from, and remove it.

tomaka:
This updates `cc-rs` to 1.10 in order to integrate https://github.com/alexcrichton/cc-rs/pull/311
Fix #8317 

Also adds an entry in the gitlab-ci file.
The TODO in the gitlab-ci file about `c++_shared` is here because the change to the Dockerfile of this PR first needs to be published on DockerHub before the check is in place.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -177,6 +177,23 @@ windows:
     paths:
     - parity.zip
     name: "x86_64-pc-windows-msvc_parity"
+android-armv7:
+  stage: build
+  image: parity/parity-android:latest
+  only:
+    - beta
+    - tags
+    - stable
+    - triggers
+  script:
+    - cargo build --target=armv7-linux-androideabi
+    # TODO: check that `arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/release/parity | grep c++_shared` is empty
+  tags:
+    - rust-arm
+  artifacts:
+    paths:
+    - parity.zip
+    name: "armv7-linux-androideabi_parity"
 docker-build:
   stage: build
   only:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,7 +65,7 @@ name = "backtrace-sys"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cc 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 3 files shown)


tomaka:
Is there a way to trigger the newly-created gitlab-ci job for this PR (to see if it works)? I don't see any option on the gitlab interface.

5chdn:
> Is there a way to trigger the newly-created gitlab-ci job for this PR (to see if it works)? I don't see any option on the gitlab interface.

I'm unsure what this means, jobs are triggered for all PRs https://gitlab.parity.io/parity/parity/-/jobs/85253

tomaka:
@5chdn I think that what you linked is just the main job. The gitlab file contains several others, including mine, that are triggered once at releases.

5chdn:
Oh yes, of course. I'll create jobs for you.

5chdn:
Here: https://gitlab.parity.io/parity/parity/pipelines/19804

5chdn:
LGTM from a CI perspective.

.gitlab-ci.yml
```diff
@@ -180,6 +180,23 @@ windows:
...
+android-armv7:
+  stage: build
+  image: parity/parity-android:latest
+  only:
+    - beta
+    - tags
+    - stable
+    - triggers
+  script:
+    - cargo build --target=armv7-linux-androideabi
```
5chdn: Hey, @tomaka, does this even make sense? This just builds it. Wouldn't it be nice to have the artifacts published via vanity/releases?

5chdn: Shouldn't we try to create a complete build job like this?

```
     - scripts/gitlab-build.sh armv7-linux-androideabi armv7-linux-androideabi armhf arm-linux-androideabi-gcc arm-linux-androideabi-g++ android
```

tomaka: @5chdn In this PR, I just copy-pasted the code of ARM-Linux and didn't think much about it.
However I don't think it's useful to have artifacts. We want to make sure that Parity continues to build on Android, but I think the main usage of Parity on Android will be as a library.

5chdn: Ok perfect!

PR #8432, opened 2018-04-18
 Consolidate crypto functionality in `ethcore-crypto`.

twittner:
This PR attempts to consolidate usage of cryptographic functionality to go through `ethcore-crypto` instead of using libraries such as `rust-crypto` or `ring` directly. `ethcore-crypto` itself uses `ring` to implement most of its functionality (except for AES-128 CTR mode, scrypt and RIPEMD160), thus this PR effectively replaces some uses of `rust-crypto` with `ring`, the other major exception being `network-devp2p`. This should also make it easier to switch to different crypto libraries if necessary. Also the previously used libary `subtle` for constant-time equality tests has been replaced with corresponding functionality from `ring`.

Please note that in order to prevent a circular dependency between `ethkey` and `ethcore-crypto`, the key agreement related functionality of `ethcore-crypto` (i.e. ECDH, ECIES) has been moved to `ethkey`.


ethcore/crypto/src/aes.rs
```diff
@@ -0,0 +1,55 @@
...
+use rcrypto::blockmodes::{CtrMode, CbcDecryptor, PkcsPadding};
+use rcrypto::aessafe::{AesSafe128Encryptor, AesSafe128Decryptor};
+use rcrypto::symmetriccipher::{Encryptor, Decryptor};
+use rcrypto::buffer::{RefReadBuffer, RefWriteBuffer, WriteBuffer};
+
+/// Encrypt a message (CTR mode).
+///
+/// Key (`k`) length and initialisation vector (`iv`) length have to be 16 bytes each.
+/// An error is returned if the input lengths are invalid.
+pub fn encrypt(k: &[u8], iv: &[u8], plain: &[u8], dest: &mut [u8]) -> Result<(), SymmError> {
```
niklasad1: I would like re-name this to something like `encrypt_ctr_128`

ethcore/crypto/src/aes.rs
```diff
@@ -0,0 +1,55 @@
...
+	let mut encryptor = CtrMode::new(AesSafe128Encryptor::new(k), iv.to_vec());
+	encryptor.encrypt(&mut RefReadBuffer::new(plain), &mut RefWriteBuffer::new(dest), true)?;
+	Ok(())
+}
+
+/// Decrypt a message (CTR mode).
+///
+/// Key (`k`) length and initialisation vector (`iv`) length have to be 16 bytes each.
+/// An error is returned if the input lengths are invalid.
+pub fn decrypt(k: &[u8], iv: &[u8], encrypted: &[u8], dest: &mut [u8]) -> Result<(), SymmError> {
```
niklasad1: I would like re-name this to something like `decrypt_ctr_128`

ethcore/crypto/src/aes.rs
```diff
@@ -0,0 +1,55 @@
...
+	encryptor.decrypt(&mut RefReadBuffer::new(encrypted), &mut RefWriteBuffer::new(dest), true)?;
+	Ok(())
+}
+
+
+/// Decrypt a message (CBC mode).
+///
+/// Key (`k`) length and initialisation vector (`iv`) length have to be 16 bytes each.
+/// An error is returned if the input lengths are invalid.
+pub fn decrypt_cbc(k: &[u8], iv: &[u8], encrypted: &[u8], dest: &mut [u8]) -> Result<usize, SymmError> {
```
niklasad1: I would like re-name this to something like `decrypt_cbc_128`

PR #8435, opened 2018-04-18
SecretStore: service pack 1

### LINKED ISSUE #7956: Add Secret Store configuration variable secretstore_acl_checker
It will be nice to allow reference a smart contract , which can manage ACL settings for Secret Store. And put this into configuration variable. # # @svyatonik

svyatonik:
on top of #8357 
closes #7956 

This is a PR for cumulative fixes (didn't wanted to spam with many SS PRs), made during SS initial nodes deployment on Kovan. Details:
- registry entry 'secretstore_acl_checker' is already occupied on Kovan => had to unify configuration options for all 3 contracts, used by SS - now their addresses can be: "none" (or option absence) means do not use contract (ACL = no permissions check, KeyServerSet = nodes list is read from configuration, Service = we do not listen to any service contracts), "registry" means read contract address from registry and address means configure for contract from this address;
- fixed: key servers listens to service contract even if it is isolated from SS - now contract events are ignored in this case;
- fixed: when there's not enough nodes to restore document key, request 'hangs' in service contract - now master node broadcast key version negotiation error + every other node checks if this is actual error && publishes doc key retrieval error;
- fixed: if key server is the only holder of key share, key version negotiation sessions completes without returning key threshold;
- fixed: servers set change session timeouts when completes instantly (only isolated nodes are removed);
- fixed: share add session fails when node known as version holder doesn't report ownership (when joining node, that has lost its SS database);
- fixed: node can miss KeyServerSet contract state changes sometimes;
- fixed: reading pending personal doc shadow data retrieval task from service contract;
- fixed: internal error during ACL check (e.g. syncing is in progress) is considered fatal;
- fixed/added useful traces where required.


5chdn:
Test repeatedly failed:

    tests::send_message_to_handler

niklasad1:
@svyatonik @5chdn I think that error is `mio` related and rebase/merge to master should fix it!

5chdn:
Ah, gotcha. @niklasad1 wanna try to give this PR the first review?

parity/configuration.rs
```diff
@@ -1234,11 +1239,11 @@ struct UiEnabled {
 	pub info_page_only: bool,
 }
 
-fn into_secretstore_service_contract_address(s: &str) -> Result<Option<SecretStoreContractAddress>, String> {
-	match s {
-		"none" => Ok(None),
-		"registry" => Ok(Some(SecretStoreContractAddress::Registry)),
-		a => Ok(Some(SecretStoreContractAddress::Address(a.parse().map_err(|e| format!("{}", e))?))),
+fn into_secretstore_service_contract_address(s: Option<&String>) -> Result<Option<SecretStoreContractAddress>, String> {
+	match s.map(|x| &**x) {
```
niklasad1: I think this should be simplified to `s.map(String::as_str)` for readability purposes!

secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -270,6 +290,14 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 			&KeyVersionNegotiationMessage::KeyVersions(ref message) =>
 				self.on_key_versions(sender, message),
 			&KeyVersionNegotiationMessage::KeyVersionsError(ref message) => {
+				// remember failed continue action
+				match message.continue_with {
+					Some(FailedKeyVersionContinueAction::Decrypt(Some(ref origin), ref requester)) =>
+						self.data.lock().failed_continue_with =
+							Some(FailedContinueAction::Decrypt(Some(origin.clone().into()), requester.clone().into())),
+					_ => (),
```
niklasad1: use `if let` instead because you only care about one pattern!

secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -361,8 +391,47 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 		let confirmations = data.confirmations.as_ref().expect(reason);
 		let versions = data.versions.as_ref().expect(reason);
 		if let Some(result) = core.result_computer.compute_result(data.threshold.clone(), confirmations, versions) {
+			// when master node processing decryption service request, it starts with a key versions negotiation session
```
niklasad1: `when the master node`

`a key version negotiation`

secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -361,8 +391,47 @@ impl<T> SessionImpl<T> where T: SessionTransport {
...
+			// if negotiation fails, only master node knows about it
```
niklasad1: `if the negotiation fails`

secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -361,8 +391,47 @@ impl<T> SessionImpl<T> where T: SessionTransport {
...
+			// => if error is fatal, only master will know about it and report to the contract && request will never be rejected
```
niklasad1: `if the error is fatal, only the master ...`

´report it`

`the request will ...`

secret_store/src/key_server_cluster/admin_sessions/key_version_negotiation_session.rs
```diff
@@ -361,8 +391,47 @@ impl<T> SessionImpl<T> where T: SessionTransport {
 		let confirmations = data.confirmations.as_ref().expect(reason);
```
niklasad1: Out-of-scope for this PR, would be nice if `fn try_complete` would return Result instead!

5chdn:
Needs a 2nd review :)

5chdn:
Long shot, @folsen do you want to review this? :hear_no_evil:

secret_store/src/key_server_cluster/admin_sessions/servers_set_change_session.rs
```diff
@@ -496,22 +503,20 @@ impl SessionImpl {
...
-							&master_node_id,
-							&key_share_owners,
-							new_nodes_set)?;
-
-						if local_plan.new_nodes_map.keys().collect::<BTreeSet<_>>() != master_plan.new_nodes_map.keys().collect::<BTreeSet<_>>() {
-							return Err(Error::InvalidMessage);
-						}
+					let key_share_owners = message.version_holders.iter().cloned().map(Into::into).collect();
+					let new_nodes_set = data.new_nodes_set.as_ref()
+						.expect("new_nodes_set is filled during consensus establishing; change sessions are running after this; qed");
```
niklasad1: Can you explain why `expect()` here instead of returning an error?

svyatonik: 1) because I'm sure that the error won't ever happen. And I'm sure because...
2) ...the `ServerSetChange` session consists of two (omitting details) phases. The first one (`EstablishingConsensus`) is when master node prepares session structure (i.e. asking all other nodes - if they want to participate in the session, what shares do they have && what to do with these shares - either leave as is or add/remove). The second phase starts (`RunningShareChangeSessions`) after first phase is finished and it is the phase where shares are actually altered (new shares are added/old shares are removed). `on_initialize_share_change_session` is called during a second phase - it is checked at the beginning of the method. And `new_nodes_set` is initialized during first phase - on master node during session `initialize` call and on other nodes - when `UnknownSessionsRequest` message is received (that's the transition from phase to phase2).

secret_store/src/key_server_cluster/cluster_sessions.rs
```diff
@@ -571,15 +576,16 @@ mod tests {
...
+		ClusterSessionsContainerState, SESSION_TIMEOUT_INTERVAL};
 
 	pub fn make_cluster_sessions() -> ClusterSessions {
 		let key_pair = Random.generate().unwrap();
 		let config = ClusterConfiguration {
 			threads: 1,
 			self_key_pair: Arc::new(PlainNodeKeyPair::new(key_pair.clone())),
 			listen_address: ("127.0.0.1".to_owned(), 100_u16),
-			key_server_set: Arc::new(MapKeyServerSet::new(vec![(key_pair.public().clone(), format!("127.0.0.1:{}", 100).parse().unwrap())].into_iter().collect())),
+			key_server_set: Arc::new(MapKeyServerSet::new(false, vec![(key_pair.public().clone(), format!("127.0.0.1:{}", 100).parse().unwrap())].into_iter().collect())),
```
niklasad1: Replace `format!` with "127.0.0.1:100".parse().unwrap() to get rid of a String allocation!

5chdn:
I think this is good to go once you resolve conflicts

PR #8438, opened 2018-04-19
Use forked app_dirs crate for reverted Windows dir behavior 

### LINKED ISSUE #8418: Parity default paths changed. (app-dirs)

> I'm running:
>
> - **Which Parity version?**: 1.9.6 and 1.10.1
> - **Are you fully synchronized?**:  yes
> - **Did you try to restart the node?**:  yes

I updated parity first to 1.9.6 after running 1.9.5. For some reason it doesn't load up the config.toml file. Even when I use -c options it tells me the config is not a preset. I need to run parity with some settings, including non-standard location for the DB. Tried 1.10.1 and it's the same.

sorpaas:
Fixes #8418

This uses the forked version for reverted Windows behavior. It also removes the dep unused in CLI.


5chdn:
Do you mind backporting this directly to beta and stable?

PR #8441, opened 2018-04-19
Fix unintentional source code permission 755 -> 644

sorpaas:
Normal source code files should have permission 644, otherwise it's marked executable. This fixes all `*.rs` file permissions.

One thing I didn't touch is files in the `mac` folder. Looks like we have many files there have executable permissions but don't look like executables. I'm not sure about mac permissions so this PR doesn't include them.


niklasad1:
@sorpaas Does these file permission changes have any implications on Windows? The access control is a little different compared to `*UNIX-like*` systems (Linux/OSX)?

sorpaas:
@niklasad1 I think at least not for this change -- most of our other `*.rs` files are already permission 644, and none of our Windows build has trouble. I don't know about how git handles permission on Windows, but if it can change permission there from 644 to 755, I think there won't be any problem for git to change it back.

niklasad1:
Ok, sounds good to me then!

PR #8443, opened 2018-04-20
change files' mode

EighteenZi:
remove rust files' executable mode.


niklasad1:
@EighteenZi 

Thanks, this is good but #8441 provides the same `fix` and I think we should `first come first serve` principle

PR #8446, opened 2018-04-20
Small fixes

ayrat555:
fix typos in comment

### DIFF
--- a/json/src/vm/vm.rs
+++ b/json/src/vm/vm.rs
@@ -22,7 +22,7 @@ use hash::H256;
 use blockchain::State;
 use vm::{Transaction, Call, Env};
 
-/// Reporesents vm execution environment before and after exeuction of transaction.
+/// Represents vm execution environment before and after execution of transaction.
 #[derive(Debug, PartialEq, Deserialize)]
 pub struct Vm {
 	/// Contract calls made internaly by executed transaction.


ayrat555:
[clabot:check]

ayrat555:
[clabot:check]

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -132,7 +131,7 @@ impl<Cost: CostType> vm::Vm for Interpreter<Cost> {
 			reader.position += 1;
 
 			// TODO: make compile-time removable if too much of a performance hit.
-			do_trace = do_trace && ext.trace_next_instruction(
+			let do_trace = ext.trace_next_instruction(
```
tomusdrw: Actually it was used before. The point is that we stop executing `trace_next_instruction` if we receive first `false`.
Removing this condition creates unnecessary overhead of invoking tracer for every executed instruction even if tracing is disabled.

ayrat555: But I guess this variable is set manually?

tomusdrw: It's set manually only for the first iteration, then it's being changed with every iteration of the loop. So if `trace_next_instruction` returns `false` for the first time, next iteration of the loop will not invoke `trace_next_instruction` again (due to `do_trace &&` resolving to false immediately).

niklasad1:
@tomusdrw You're ok with this now right?

PR #8439, opened 2018-04-19
Improve VM executor stack size estimation rules

sorpaas:
This improves the stack size estimation, taking advantage of `scope.builder().stack_size(...)`, so instead of resetting stack once a while by creating new threads, we only need at most one additional thread for a VM execution.

The stack size per depth is now changed to take account whether we are in debug build or release build, so it also accidentally fixes #8376.

Regarding the additional `expect()` in `builder()...spawn()` -- this is because the original bare `spawn` function (https://docs.rs/crossbeam/0.3.2/src/crossbeam/scoped.rs.html#230-234) simply uses an `unwrap()`.

PR #8367, opened 2018-04-11
remove Tendermint extra_info due to seal inconsistencies

ordian:
fixes #7727.

see [my gdb history](https://github.com/ordian/parity-issue8157/blob/master/gdb/thoughts.md)

PR #8449, opened 2018-04-20
[stable] Backports

sorpaas:
* #8438 `Use forked app_dirs crate for reverted Windows dir behavior`


5chdn:
Do you mind backporting these to stable, too?

- #8367
- #8439

sorpaas:
We need to update crossbeam to 0.3 if we want to apply #8439. Looks like that's harmless.

Issue #8427, opened 2018-04-18
Constantinople Release


5chdn:
Tracking issue for Constantinople EIPs implementation progress:

- [x] [EIP 145](http://eips.ethereum.org/EIPS/eip-145) Bitwise shifting instructions in EVM #8451
- [ ] ~~[EIP 210](http://eips.ethereum.org/EIPS/eip-210) Blockhash refactoring #5505 (needs to check current version compliance)~~
- [x] [EIP 1014](http://eips.ethereum.org/EIPS/eip-1014) Skinny CREATE2 #9268
- [x] [EIP 1052](http://eips.ethereum.org/EIPS/eip-1052) EXTCODEHASH Opcode #9234
- [x] [EIP 1234](http://eips.ethereum.org/EIPS/eip-1234) Delay diff bomb #9480
- [x] [EIP 1283](http://eips.ethereum.org/EIPS/eip-1283) Net gas metering for SSTORE (replaces 1087) #9319

Hardfork Meta EIP:

- [EIP 1013](http://eips.ethereum.org/EIPS/eip-1013) Constantinople Fork

Timeline:

> **Proposed Constantinople Timeline**
> Finalize EIPs that are being implemented: July 13th
> Client implementation: July 16th - August 13th
> Testing: August 13th - September 10th
> Testnet: September 10th - October 1st
> Launch: October 8th

folsen:
Geth implementation of 1087 https://github.com/ethereum/go-ethereum/pull/17208

sorpaas:
I'm having some issues getting EIP-1087 implemented in our codebase. With our current state cache method, we can do it, but with (probably noticeable) performance penalty. The issue is that we optimize state commitment to only happen at the end of a block, but EIP-1087 makes the assumption that state commitment happens at the end of every transaction.

So to implement current spec, it seems that we cannot use info from current cache (because storage caches, including checkpoints, also contain changes from previous transactions), but instead, we need to maintain a separate dirty map just to keep track of the gas cost. That dirty map will probably also need to contain the original storage value. So adding up, in this way, it basically doubles the memory size of storage cache.

5chdn:
All done. Can start testing.

PR #8451, opened 2018-04-20
EIP 145: Bitwise shifting instructions in EVM

sorpaas:
cc #8427 

This implements EIP145 with three new opcodes `SHL`, `SHR` and `SAR`, and adds `eip145transition` config parameter. Tests specified in the EIP are included.

### DIFF
--- a/ethcore/src/spec/spec.rs
+++ b/ethcore/src/spec/spec.rs
@@ -197,16 +197,16 @@ impl From<ethjson::spec::Params> for CommonParams {
 			eip155_transition: p.eip155_transition.map_or(0, Into::into),
 			validate_receipts_transition: p.validate_receipts_transition.map_or(0, Into::into),
 			validate_chain_id_transition: p.validate_chain_id_transition.map_or(0, Into::into),
-			eip86_transition: p.eip86_transition.map_or(
-				BlockNumber::max_value(),
+			eip86_transition: p.eip86_transition.map_or_else(
+				BlockNumber::max_value,
 				Into::into,
 			),
-			eip140_transition: p.eip140_transition.map_or(
-				BlockNumber::max_value(),
+			eip140_transition: p.eip140_transition.map_or_else(
+				BlockNumber::max_value,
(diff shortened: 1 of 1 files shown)


ethcore/src/spec/spec.rs
```diff
@@ -220,6 +223,10 @@ impl From<ethjson::spec::Params> for CommonParams {
 				BlockNumber::max_value(),
 				Into::into,
 			),
+			eip145_transition: p.eip145_transition.map_or(
+				BlockNumber::max_value(),
+				Into::into,
+			),
```
debris: not an issue, but wouldn't it may be better to write:

```rust
			eip145_transition: p.eip145_transition.map_or_else(
				BlockNumber::max_value,
				Into::into,
			),
```

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -871,6 +872,58 @@ impl<Cost: CostType> Interpreter<Cost> {
...
+					if sign {
+						U256::max_value()
+					} else {
+						U256::zero()
+					}
+				} else {
+					let shift = shift.as_u32() as usize;
+					let mut shifted = value >> shift;
+					if sign {
+						shifted = shifted | (U256::max_value() << (256 - shift));
```
debris: can `256 - shift` underflow?

ethcore/src/spec/spec.rs
sorpaas: I changed all instance of this in `spec.rs` to use `map_or_else`.

niklasad1: @debris what is the reason for preferring `map_or_else` over `map_or`? Is it because the arguments to `map_or` are eagerly evaluated?

tomusdrw: I think so, yeah. Although `BlockNumber` is just `u64`, so `max_value()` most likely inlined by the compiler anyway.
I like the change though, because of consistency across the codebase.

PR #8453, opened 2018-04-20
Update rust version

AyushyaChitransh:
The earlier version has many deprecated things and should be updated


niklasad1:
Hey, that is not the actual `rustc version` it's a crate for querying rustc.

0.2.2 is the lastest which is used by `parity`, you can find the crate here: https://crates.io/crates/rustc_version

PR #8457, opened 2018-04-20
`duration_ns: u64 -> duration: Duration`

niklasad1:
Just to simplify the code and it is not benefiting from `const` Duration.

Also it is more inefficient in terms of memory usage because a Duration requires `16 bytes` and u64 `8 bytes`

### DIFF
--- a/parity/informant.rs
+++ b/parity/informant.rs
@@ -373,8 +373,7 @@ impl ChainNotify for Informant<FullNodeInformantData> {
 					Colour::White.bold().paint(format!("{}", header_view.hash())),
 					Colour::Yellow.bold().paint(format!("{}", block.transactions_count())),
 					Colour::Yellow.bold().paint(format!("{:.2}", header_view.gas_used().low_u64() as f32 / 1000000f32)),
-					// this will not provide floating point precision because `as_milliseconds` returns u64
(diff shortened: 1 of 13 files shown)


parity/informant.rs
```diff
@@ -373,7 +373,8 @@ impl ChainNotify for Informant<FullNodeInformantData> {
 					Colour::White.bold().paint(format!("{}", header_view.hash())),
 					Colour::Yellow.bold().paint(format!("{}", block.transactions_count())),
 					Colour::Yellow.bold().paint(format!("{:.2}", header_view.gas_used().low_u64() as f32 / 1000000f32)),
-					Colour::Purple.bold().paint(format!("{:.2}", duration as f32 / 1000000f32)),
+					// this will not provide floating point precision because `as_milliseconds` returns u64
+					Colour::Purple.bold().paint(format!("{:.2}", duration.as_milliseconds())),
```
niklasad1: This can be changed provide the same behaviour as before but since we already had a method for specifically implemented for `Duration` I used that one instead!

andresilva: Since it will always be `.00` I think we should remove the `{:.2}` from the format string. I don't think we care about nano precision wrt time it takes to import a block so I guess it's OK to leave it as millis.

niklasad1: Yepp, I agree

Issue #1425, opened 2016-06-24
FreeBSD build


arkpar:

tymat:
Looks like updating the `nix` crate fixes the compilation error related to that issue.

One other remaining compile issue is with the version of `ring` crate which was recently fixed here https://github.com/briansmith/ring/pull/313

DominiLux:
I'm going to volunteer some of my time this week to this challenge.  I love a good port plus I'm determined to get the latest version of parity running on BSD from the master branch.  I'll document my steps and prepare a pure bash script to automate the system configurations.  I'm going to do it from a fresh install and the use the ports tree to compile whatever is needed.  Implementation should be simple enough.  But before I do that I need to spend about 3 or 4 days studying dependencies and possibly creating a custom rust github as I dont think that's available on BSD yet.  I'll let you gentleman know my progress.  I will begin research today.

paukstis:
Hi, I'm following the process with a patch, master and v1.4 branches [in this repo](https://github.com/paukstis/freebsd_parity)

DominiLux:
[...]

PR #8464, opened 2018-04-23
Disable hardware-wallets on platforms that don't support `libusb`

niklasad1:
Hey, 

This PR tries to add functionality to build parity on platforms that don't support `libusb or is stable` by using conditional compilation i.e,

* libusb support -> `#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]`
* not libusb support -> `#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]`

This is really `brute-force` and the syntax is horrible which may be `messy` to read! 

However, I want to know if this is something we want before putting more time on this. Moreover, this makes the `--no-hardware-wallets` useless on these platform because it is already disabled   
arm-linux-androideabi

I have tested it by running (without errors):
```
$ cargo check --target x86_64-unknown-freebsd
```

If any `FreeBSD` user can test this I would appreciate it!

/cc #1425


folsen:
That's a lot of extra noise for all that stuff. I think it's desirable for sure, but I'm not sure about where the tradeoff makes sense in terms of code ugliness/maintainability vs being able to disable it. I'm not sure what's possible theoretically in making this prettier?

Ultimately going the way of geth and having a completely separate account mgmt binary (including hw wallets) might be a good idea.

niklasad1:
Right, I think it should be possible simply `#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "android"))]` to something like `#[cfg(hw_wallet)]` is still noisy but a little bit better!

By using features in the crates that is using based on the target:
```rustc
[target.'cfg(target_os = "linux")'.dependencies]
rpc = { path = "...", features = ["hw_wallet"] }

[target.'cfg(not(target_os = "linux"))'.dependencies]
rpc = { path = "...", features = [] }
```

ethcore/src/account_provider/mod.rs
```diff
@@ -31,14 +31,32 @@ use ethstore::{
...
+	pub use hardware_wallet::{Error as HardwareError, HardwareWalletManager, KeyPath, TransactionInfo};
+	pub use ::transaction::{Action, Transaction};
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "android")))] 
+mod hw {
+	use super::fmt;
+
+	#[derive(Debug)]
+	pub enum HardwareError {}
```
niklasad1: need docs

ethcore/src/account_provider/mod.rs
```diff
@@ -31,14 +31,32 @@ use ethstore::{
...
+	pub struct HardwareWalletManager;
```
niklasad1: need docs

5chdn:
Needs reviews.

ethcore/src/account_provider/no_hw.rs
```diff
@@ -0,0 +1,42 @@
+//! Dummy module for platforms that does not provide support for hardware wallets (libusb)
+
+use super::{fmt, Address};
+
+pub struct WalletInfo {
```
niklasad1: copy-paste of `hardware_wallet::WalletInfo`

rpc/src/lib.rs
```diff
@@ -67,7 +67,10 @@ extern crate parity_version as version;
 extern crate rlp;
 extern crate stats;
 extern crate keccak_hash as hash;
+
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
```
niklasad1: Removed android for now because I am not sure whether android compiles or not. 

I need to grab the `android NDK` in order to test it!

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -713,6 +717,7 @@ fn signature(accounts: &AccountProvider, address: Address, hash: H256, password:
 }
 
 // obtain a hardware signature from the given account.
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
```
niklasad1: I skipped this because it will require to implement alot of more dummy types!

rpc/src/lib.rs
tomaka: The easiest solution is docker.
Go to `parity/docker/android` and create the docker image: `docker build -t parity-android .`
You can then compile Parity with it: `docker run --rm -it -v `pwd`:/usr/code -w /usr/code parity-android` and `cargo build`.

niklasad1: Thanks Pierre!

jtomanik:
Hey @niklasad1 it would be good to add `target_os = "ios"` in places where you already have `target_os = "android"`. This would nicely work with https://github.com/paritytech/parity/issues/7955

niklasad1:
Hey @jtomanik thanks for the heads-up but `iOS` is not supported by libusb according to https://github.com/libusb/libusb/wiki# that's why I only have enabled `hardware_wallet` for Windows, OSX, Linux and Android

jtomanik:
@niklasad1 sorry misread your code, still getting used to Rust syntax

ethcore/src/account_provider/mod.rs
```diff
@@ -280,16 +296,16 @@ impl AccountProvider {
...
-		   .collect()
+			.into_iter()
+			.map(|a| a.address)
+			.filter(|address| !self.blacklisted_accounts.contains(address))
+			.collect()
 		)
 	}
 
 	/// Returns addresses of hardware accounts.
 	pub fn hardware_accounts(&self) -> Result<Vec<Address>, Error> {
```
dvdplm: How about making this private and return `Error`? Seems like the only caller is `hardware_accounts_info()` and it a little misleading as `hardware_accounts()` doesn't actually ever fail (due to the `map_or_else`)?

ethcore/src/account_provider/no_hw.rs
dvdplm: Couldn't this be a newtype of that?

ethcore/src/account_provider/mod.rs
niklasad1: Yeah, that is a great suggestion i.e, to return an `Error` but it can't be private because `rpc` actually calls this!

5chdn:
Needs a 2nd review.

PR #8470, opened 2018-04-24
Transaction Pool improvements

tomusdrw:
- [x] Pool generic over Hash and Sender
- [x] Required unique insertion id to resolve scoring conflicts managed by the pool itself.
- [x] Typo fix

### DIFF
--- a/miner/src/pool/mod.rs
+++ b/miner/src/pool/mod.rs
@@ -105,6 +105,11 @@ impl VerifiedTransaction {
 		self.priority
 	}
 
+	/// Gets transaction insertion id.
+	pub(crate) fn insertion_id(&self) -> usize {
+		self.insertion_id
+	}
+
 	/// Gets wrapped `SignedTransaction`
 	pub fn signed(&self) -> &transaction::SignedTransaction {
 		&self.transaction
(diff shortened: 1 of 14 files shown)


miner/src/pool/mod.rs
```diff
@@ -105,6 +105,11 @@ impl VerifiedTransaction {
 		self.priority
 	}
 
+	/// Gets transaction insertion id.
+	pub(crate) fn insertion_id(&self) -> usize {
+		self.insertion_id
```
andresilva: Where is this `insertion_id` used? It seems we wrap this type in `txpool::Transaction` and use the outer `insertion_id` instead.

tomusdrw: So `pool` manages it's internal `insertion_id` (it's required for uniqueness).

This `insertion_id` here is used for detecting`Future` transactions that occupy the pool for a long time, and never get included: https://github.com/paritytech/parity/blob/4900e4da6b7540ba2f172874e723e926adbc8c77/miner/src/pool/ready.rs#L93
We wait for half of the pool to be replaced, but never less than 100 transactions:
https://github.com/paritytech/parity/blob/4900e4da6b7540ba2f172874e723e926adbc8c77/miner/src/pool/queue.rs#L289

We can have any other retention mechanism that is not tied to ids, but I didn't want to change this in that particular PR (time based or block number based for instance).

Issue #8480, opened 2018-04-25
Node stops responding to RPC calls


ppratscher:
> I'm running:
>
> - **Which Parity version?**: v1.10.2
> - **Which operating system?**: Linux
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: N/A

Since upgrading our parity nodes to 1.9.6 / 1.10.1 / 1.10.2 we see cases where the nodes stop responding to any kind of RPC call. When querying such a node in this state via curl we just get a connection refused error back.

It happens more frequently on nodes that have a high volume of RPC calls. We were able to somewhat reliable reproduce the issue using the following stress test script (the RPC errors usually start to occur after ~1-2 hours):

```javascript
const Web3 = require("web3");
let web3 = new Web3("http://localhost:8545");

let i = 0;
function stress() {
  web3.eth.isSyncing((err, result) => {
    if (err) {
      console.log(err);
    }
    i++
    if (i % 1000 === 0) {
      console.log(i);
    }
  });
}

setInterval(stress, 10);
```
I suspect that the issue might be related to upgrading the jsonrpc module in https://github.com/paritytech/parity/pull/8181

ppratscher:
Small update, I am unable to replicate the issue using parity v1.9.5 & the script posted above.

tomusdrw:
@ppratscher With 1.9.5 do you see some of the requests fail?

ppratscher:
No, not a single request failed during the 3+ hours the test script was running with v1.9.5

ppratscher:
Edit: With a more aggressive version of the test script I was able to get the request to fail also with v1.9.5:

```javascript
const Web3 = require("web3");
let web3 = new Web3("http://localhost:8545");

let i = 0;


function stress() {
  web3.eth.isSyncing((err, result) => {
    if (err) {
      console.log(err);
    }
    i++
    if (i % 1000 === 0) {
      console.log(i);
    }
  });
}

setInterval(stress, 1);
setInterval(stress, 1);
```

tomusdrw:
I'm trying to reproduce it now. Would you mind testing if it also happens for you with `--mode=offline`?

ppratscher:
Yes, it also happens when running the node with mode=offline

tomusdrw:
@ppratscher It's mostly `web3` issue not us. Let me explain:
[...]

PR #8481, opened 2018-04-25
Don't require write lock when fetching status.

tomusdrw:
Possibly related #8480


niklasad1:
BTW, would be interesting to perform some benchmarks on this w.r.t rpc throughput!

PR #8486, opened 2018-04-25
Trace precompiled contracts when the transfer value is not zero

### LINKED ISSUE #7765: VM trace bug(?). Ether disappeared from address.
I run `Parity/v1.8.6-beta-2d051e4-20180109/x86_64-linux-gnu/rustc1.22.1`, but on Etherscan I got the same result.
```
eth.getBalance('0xb57b9206d75c1bb02cb64f97fb5176eae731a62d',3852625)
21811236187340800
eth.getBalance('0xb57b9206d75c1bb02cb64f97fb5176eae731a62d',3852626)
0
```
There is one transaction in the block nr 3852626 who calls this contract.
https://etherscan.io/tx/0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29
Here is the trace:
```
trace.transaction('0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29')
[
    {
        "blockHash": "0x81d43388efb220c385d5d7bffb3c38e7ff3fdd8d0ad05ae8acc7607563d7a67a",
        "transactionHash": "0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29",
        "blockNumber": 3852626,
        "traceAddress": [],
        "subtraces": 3,
        "action": {
            "callType": "call",
            "from": "0x26588a9301b0428d95e6fc3a5024fce8bec12d51",
            "gas": "0x54468",
            "value": "0x0",
            "to": "0xb57b9206d75c1bb02cb64f97fb5176eae731a62d",
            "input": "0x38bbfa503a959d8def976bd51348262afb4586efddb66df187166d76ad34dcd9dfd273c1000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000af1d8d967a0ddd98df1680000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001c04c50010498c36fc999f7a37ffffe7e4207142d9f708b96f190cfaed75133dba388ac694c2f6db039f926344c918a1d96142a95c332a5403b21e94953a9c75bc27d4290a33044022003bd07a86cb31b9beabbeefa55f9343fdff77afd840da1a43a714948c32e70d1022004f86513ff6689c5553599ccc60da62e063411e484e250545f844a0ba3d1e57ffd94fa71bc0ba10d39d464d0d8f465efeef0a2764e3887fcc9df41ded20f505c19bf4b536960336aac1eff008d2f3efdda1edf9a55a01e6afb22a932a4dd1dd600000000000000000ab266f9f4620f7774ec3dd4554872e9f9cef6174ebb79e034ae98b26720a1c08731440220190305d80a98c143c59582cb0139781c52fadeda89f50c8c29350f8cdf157c9102203406245712f238b2dfd71ff10ce99cb0072e5f1b5845c663d194937e5627f497041a098edd9f3f8a25940bec7a6a28c23e234000c9e6ca961a2fe739aa8b660cc380b2be58a4050eceb5c1d15f7d849a162e3b542076ba9494beec1857a68e77e830440220719c29fe7e009c3fd7cb68f635de422726801f42788bb4cfbc73ef9b0c66ee5902201cfa3f98b4ad9e2996bc57215488405b46d3e15fd297adce289b923d249d016b"
        },
        "transactionPosition": 78,
        "type": "call",
        "result": {
            "output": "0x",
            "gasUsed": "0x384e4"
        }
    },
    {
        "blockHash": "0x81d43388efb220c385d5d7bffb3c38e7ff3fdd8d0ad05ae8acc7607563d7a67a",
        "transactionHash": "0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29",
        "blockNumber": 3852626,
        "traceAddress": [
            0
        ],
        "subtraces": 0,
        "action": {
            "callType": "call",
            "from": "0xb57b9206d75c1bb02cb64f97fb5176eae731a62d",
            "gas": "0x3ce30",
            "value": "0x0",
            "to": "0x1d3b2638a7cc9f2cb3d298a3da7a90b67e5506ed",
            "input": "0x38cc4831"
        },
        "transactionPosition": 78,
        "type": "call",
        "result": {
            "output": "0x0000000000000000000000006f28b146804dba2d6f944c03528a8fdbc673df2c",
            "gasUsed": "0x1a2"
        }
    },
    {
        "blockHash": "0x81d43388efb220c385d5d7bffb3c38e7ff3fdd8d0ad05ae8acc7607563d7a67a",
        "transactionHash": "0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29",
        "blockNumber": 3852626,
        "traceAddress": [
            1
        ],
        "subtraces": 0,
        "action": {
            "callType": "call",
            "from": "0xb57b9206d75c1bb02cb64f97fb5176eae731a62d",
            "gas": "0x3b1f7",
            "value": "0x0",
            "to": "0x6f28b146804dba2d6f944c03528a8fdbc673df2c",
            "input": "0xc281d19e"
        },
        "transactionPosition": 78,
        "type": "call",
        "result": {
            "output": "0x00000000000000000000000026588a9301b0428d95e6fc3a5024fce8bec12d51",
            "gasUsed": "0x57b"
        }
    },
    {
        "blockHash": "0x81d43388efb220c385d5d7bffb3c38e7ff3fdd8d0ad05ae8acc7607563d7a67a",
        "transactionHash": "0x92ddeb8df37c1b43debeecb07aa25a0ed96f13f8405aca9dba08ab204a8d2e29",
        "blockNumber": 3852626,
        "traceAddress": [
            2
        ],
        "subtraces": 0,
        "action": {
            "callType": "call",
            "from": "0xb57b9206d75c1bb02cb64f97fb5176eae731a62d",
            "gas": "0x8fc",
            "value": "0x18cbe7bb38000",
            "to": "0xfd97b6d9977cb009cdf95fac9dd4878b53233b94",
            "input": "0x"
        },
        "transactionPosition": 78,
        "type": "call",
        "result": {
            "output": "0x",
            "gasUsed": "0x0"
        }
    }
]
```
In the trace is only one ether transfer with 436224723746816 WEI (0x18cbe7bb38000).

Where is the other 21375011463593984 WEI from the address?

sorpaas:
Fixes #7765.

### DIFF
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -726,6 +726,12 @@ mod tests {
 		machine
 	}
 
+	fn make_byzantium_machine(max_depth: usize) -> EthereumMachine {
+		let mut machine = ::ethereum::new_byzantium_test_machine();
+		machine.set_schedule_creation_rules(Box::new(move |s, _| s.max_depth = max_depth));
+		machine
+	}
+
 	#[test]
 	fn test_contract_address() {
 		let address = Address::from_str("0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6").unwrap();
@@ -846,7 +852,7 @@ mod tests {
 		let mut state = get_temp_state();
 		state.add_balance(&sender, &U256::from(100), CleanupMode::NoEmpty).unwrap();
 		let info = EnvInfo::default();
-		let machine = make_frontier_machine(5);
+		let machine = make_byzantium_machine(5);
 		let mut substate = Substate::new();
 		let mut tracer = ExecutiveTracer::default();
 		let mut vm_tracer = ExecutiveVMTracer::toplevel();
@@ -865,7 +871,7 @@ mod tests {
 				call_type: CallType::Call
 			}),
 			result: trace::Res::Call(trace::CallResult {
-				gas_used: 32361.into(),
+				gas_used: 33021.into(),
 				output: vec![]
 			}),
 			subtraces: 1,
@@ -875,7 +881,7 @@ mod tests {
 				from: "5555555555555555555555555555555555555555".into(),
 				to: "0000000000000000000000000000000000000003".into(),
 				value: 1.into(),
-				gas: 67835.into(),
+				gas: 66560.into(),
 				input: vec![],
 				call_type: CallType::Call
 			}), result: trace::Res::Call(trace::CallResult {
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -428,8 +428,12 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 					self.state.discard_checkpoint();
 					output.write(0, &builtin_out_buffer);
 
-					// trace only top level calls to builtins to avoid DDoS attacks
-					if self.depth == 0 {
+					// trace only top level calls and calls with balance transfer to builtins to avoid DDoS attacks
+					let transferred = match params.value {
+						ActionValue::Transfer(value) => value,
+						ActionValue::Apparent(_) => U256::zero(),
+					};
+					if self.depth == 0 || transferred != U256::zero() {
 						let mut trace_output = tracer.prepare_trace_output();
 						if let Some(out) = trace_output.as_mut() {
 							*out = output.to_owned();
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -431,11 +431,11 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 					// Trace only top level calls and calls with balance transfer to builtins. The reason why we don't
 					// trace all internal calls to builtin contracts is that memcpy (IDENTITY) is a heavily used
 					// function.
-					let transferred = match params.value {
-						ActionValue::Transfer(value) => value,
-						ActionValue::Apparent(_) => U256::zero(),
+					let is_transferred = match params.value {
+						ActionValue::Transfer(value) => value != U256::zero(),
+						ActionValue::Apparent(_) => false,
 					};
-					if self.depth == 0 || transferred != U256::zero() {
+					if self.depth == 0 || is_transferred {
 						let mut trace_output = tracer.prepare_trace_output();
 						if let Some(out) = trace_output.as_mut() {
 							*out = output.to_owned();
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -818,6 +818,76 @@ mod tests {
 	}
 
 	#[test]
+	fn test_call_to_precompiled_tracing() {
+		// code:
+		//
+		// 60 00 - push 00 out size
+		// 60 00 - push 00 out offset
+		// 60 00 - push 00 in size
+		// 60 00 - push 00 in offset
+		// 60 01 - push 01 value
+		// 60 03 - push 03 to
+		// 61 ffff - push fff gas
+		// f1 - CALL
+
+		let code = "60006000600060006001600361fffff1".from_hex().unwrap();
+		let sender = Address::from_str("4444444444444444444444444444444444444444").unwrap();
+		let address = Address::from_str("5555555555555555555555555555555555555555").unwrap();
+
+		let mut params = ActionParams::default();
+		params.address = address.clone();
+		params.code_address = address.clone();
+		params.sender = sender.clone();
+		params.origin = sender.clone();
+		params.gas = U256::from(100_000);
+		params.code = Some(Arc::new(code));
+		params.value = ActionValue::Transfer(U256::from(100));
+		params.call_type = CallType::Call;
+		let mut state = get_temp_state();
+		state.add_balance(&sender, &U256::from(100), CleanupMode::NoEmpty).unwrap();
+		let info = EnvInfo::default();
+		let machine = make_frontier_machine(5);
+		let mut substate = Substate::new();
+		let mut tracer = ExecutiveTracer::default();
+		let mut vm_tracer = ExecutiveVMTracer::toplevel();
+
+		let mut ex = Executive::new(&mut state, &info, &machine);
+		let output = BytesRef::Fixed(&mut[0u8;0]);
+		ex.call(params, &mut substate, output, &mut tracer, &mut vm_tracer).unwrap();
+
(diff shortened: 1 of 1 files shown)


andresilva:
Can we add a simple regression test for this?

ethcore/src/executive.rs
```diff
@@ -428,8 +428,12 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
 					output.write(0, &builtin_out_buffer);
 
-					// trace only top level calls to builtins to avoid DDoS attacks
-					if self.depth == 0 {
+					// trace only top level calls and calls with balance transfer to builtins to avoid DDoS attacks
+					let transferred = match params.value {
+						ActionValue::Transfer(value) => value,
+						ActionValue::Apparent(_) => U256::zero(),
+					};
+					if self.depth == 0 || transferred != U256::zero() {
```
tomusdrw: Doesn't it open up an attack vector then? The comment clearly states that only top level calls should be traced, currently the workaround is to just send 1 wei with them, which seems pretty cheap.

Also I would simplify it to:
```
let is_value_transfer = match params.value {
  Transfer(value) => value.is_zero(),
  Apparent(_) => false,
};
```

sorpaas: @tomusdrw I'm a little bit confused here: isn't CALL/CALLCODE always pays the opcode fee regardless whether it's precompiled contract or not? If gases are properly paid then it's probably not an attack vector any more?

I noticed the original line was written by Gavin Wood before the EIP150 and Homestead hard fork. Given that we fixed most of the CALL gas attack vectors in those two hard forks, and the reasoning above, I actually think we may want to consider remove the `depth == 0` check altogether.

sorpaas:
@andresilva Someone might treat those precompiled contracts as normal accounts, and transfer value, for whatever reasons. So it might be `CALL` opcode. For example, in solidity I can write:

```
address target = 0x0000000000000000000000000000000000000003;
target.transfer(msg.value);
```

I'll try to add some tests for this!

ethcore/src/executive.rs
sorpaas: Yeah the math also checks out. Consider `0x00..03` and `0x00..ff` as two nonexisting accounts, using `byzantium_test` config, we have:

* `6000600060006000600160ff61fffff1`: CALL to `0x00..ff` consumes `32421` gases for toplevel (without finalization).
* `60006000600060006001600361fffff1`: CALL to `0x00..03` consumes `33021` gases for toplevel (without finalization). The additional 600 gases are the base gas for RIPEMD160.

So I think it's safe to remove `depth == 0` check altogether.

sorpaas:
Realized that changing input/output type will make it incompatible with existing tracing database. So I'm going to revert and simply record an empty Vector for precompiled contracts with depth greater than zero.

5chdn:
Would love to backport this to 1.11 - what do you think @andresilva @tomusdrw

ethcore/src/trace/types/trace.rs
```diff
@@ -18,19 +18,59 @@
...
+#[derive(Debug, Clone, PartialEq)]
 pub struct CallResult {
 	/// Gas used by call.
 	pub gas_used: U256,
 	/// Call Output.
-	pub output: Bytes,
+	pub output: Option<Bytes>,
+}
+
+impl rlp::Encodable for CallResult {
```
tomusdrw: I'm not 100% if backward compatibility is worth maintaining this custom RLP serialization, perhaps better to prepare a database migration?

ethcore/src/trace/types/trace.rs
```diff
@@ -18,19 +18,59 @@
...
 
 /// `Call` result.
-#[derive(Debug, Clone, PartialEq, Default, RlpEncodable, RlpDecodable)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct CallResult {
 	/// Gas used by call.
 	pub gas_used: U256,
 	/// Call Output.
-	pub output: Bytes,
+	pub output: Option<Bytes>,
```
tomusdrw: Is it worth changing the type here? It introduces additional complexity (custom RLP serialization or database migration) and will only be used when tracing calls to builtins - a feature that was not available before at all. @debris what's your thoughts on this?

ethcore/src/executive.rs
```diff
@@ -410,7 +410,7 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 			let default = [];
 			let data = if let Some(ref d) = params.data { d as &[u8] } else { &default as &[u8] };
 
-			let trace_info = tracer.prepare_trace_call(&params);
+			let trace_info = tracer.prepare_trace_call(&params, self.depth == 0);
```
tomusdrw: I'm not sure about the input. I think it's not cheaper than calling regular contract, is it? Why only trace input for regular contracts and not builtins then?

tomusdrw:
@5chdn makes sense, although it requires re-syncing the client to recreate the traces.

sorpaas:
@tomusdrw I chose to change the type to `Option<Bytes>` because otherwise we cannot distinguish a) traces with zero-length input/output, and b) traces with input/output stripped. This might be a confusion point on RPC (because stripped input/output will show as `0x`, which looks really as if it is just zero-length).

We can also just reset two commits and it will be the non-change-type behavior.

tomusdrw:
@sorpaas Currently we don't trace calls to builtins at all, so it's still an improvement if it displays `0x`. I think it's just a documentation matter and it's fine to keep the simpler format (at least it doesn't require migration and doesn't introduce complexity to the code). @debris - your call.

ethcore/src/executive.rs
tomusdrw: I would keep the input then, since if we don't consider that an issue on a contract level we probably should also keep the same behaviour for builtins as well - the cost is payed up front by allocating the memory, so I think the amount of data that can be generated in traces is acceptable.

sorpaas:
Based on Gav's comments, for precompiled contracts, this is changed back to trace a) direct calls, and b) internal calls where transferred value is not zero.

Issue #6010, opened 2017-07-06
Allow warp syncing for light clients


rphmeier:
1. Generalize snapshot restoration targets via a trait. Don't require a full `BlockChain`/`StateDB`, but maybe   associated types.
2. Don't require getting state chunks (all we care about are secondary/corroboration chunks)
3. Background header sync in order to get ancient CHT roots. This should be toggleable via RPC: mobile apps might set this to only run when connected to wifi. Only downside is headers in the gap will be unavailable until that sync completes

adria0:
Closing issue due to its stale state.

PR #8489, opened 2018-04-25
Update hardcodedSync for Ethereum, Kovan, and Ropsten

5chdn:


folsen:
Would really like a smaller format of this in the future :(

5chdn:
Yes, if anyone has any idea how to minify this, go ahead!

tomaka:
When #6010 is done, we will only need the latest header in the chain spec and not the CHTs.

PR #8490, opened 2018-04-25
Refactor block reward application and tracing

andresilva:
Refactors block reward application to use a common helper function. Add support for tracing different kinds of rewards.

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -1015,8 +1015,10 @@ impl Engine<EthereumMachine> for AuthorityRound {
 		let author = *block.header().author();
 		benefactors.push((author, RewardKind::Author));
 
-		let rewards = match self.block_reward_contract {
+		let rewards: Vec<_> = match self.block_reward_contract {
 			Some(ref c) if block.header().number() >= self.block_reward_contract_transition => {
+				// NOTE: this logic should be moved to a function when another
+				//       engine needs support for block reward contract.
 				let mut call = |to, data| {
 					let result = self.machine.execute_as_system(
 						block,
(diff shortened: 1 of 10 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1027,10 +1029,11 @@ impl Engine<EthereumMachine> for AuthorityRound {
 					result.map_err(|e| format!("{}", e))
 				};
 
-				c.reward(&benefactors, &mut call)?
+				let rewards = c.reward(&benefactors, &mut call)?;
+				rewards.into_iter().map(|(author, amount)| (author, RewardKind::External, amount)).collect()
```
andresilva: Unfortunately when using the block reward contract all rewards must be tagged as `External`, there's no way to appropriately map to benefactors list that was passed to the contract.

ethcore/src/trace/types/trace.rs
```diff
@@ -141,13 +141,19 @@ pub enum RewardType {
 	Block,
 	/// Uncle
 	Uncle,
+	/// Empty step (AuthorityRound)
```
andresilva: Should we remove this type from tracing and instead use `block_reward::RewardKind`?

ethcore/src/engines/block_reward.rs
```diff
@@ -14,13 +14,17 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
+//! A module with types for declaring block rewards and a client interface for interacting with a
```
andresilva: I'm not sure `engines` is the appropriate module for `block_reward`. Any thoughts?

ethcore/src/trace/types/trace.rs
tomusdrw: Sounds ok for me.

PR #8491, opened 2018-04-25
Pass on storage keys tracing to handle the case when it is not modified

### LINKED ISSUE #6520: Incorrect stateDiff in trace_replaytransaction
- **Parity version**: v1.8.0-unstable-d520aa2-20170829/x86_64-linux-gnu/rustc1.19.0
 - **Operating system**: Linux
 - **And installed**: from source

# Problem:
The problem relates to the [trace_replaytransaction](https://github.com/paritytech/parity/wiki/JSONRPC-trace-module#trace_replaytransaction) RPC call.

If A has balance 1000 and B has balance 1000, there are two storage keys with both a value of 1000 in the storage of the contract. In theory, if A sends 500 tokens to B, there should be a stateDiff like this:

```
          "storage": {
            "0x3122cf52ca415f61f9a275b9b68122794febbe33bf20650f5a17717ee0875719": {
              "*": {
                "from": "0x00000000000000000000000000000000000000000000000000000000000003e8",
                "to": "0x000000000000000000000000000000000000000000000000000000000000014f"
              }
            },
            "0xf54353a2b20ca39c6925309d40b328469cccb296e1e6cc2ee095dd38422a7526": {
              "*": {
                "from": "0x00000000000000000000000000000000000000000000000000000000000003e8",
                "to": "0x00000000000000000000000000000000000000000000000000000000000005dc"
              }
```
But what I get is this:

```
       "storage": {
          "0x3122cf52ca415f61f9a275b9b68122794febbe33bf20650f5a17717ee0875719": {
            "*": {
              "from": "0x0000000000000000000000000000000000000000000000000000000000000000",
              "to": "0x000000000000000000000000000000000000000000000000000000000000014f"
            }
          },
          "0xf54353a2b20ca39c6925309d40b328469cccb296e1e6cc2ee095dd38422a7526": {
            "*": {
              "from": "0x0000000000000000000000000000000000000000000000000000000000000000",
              "to": "0x00000000000000000000000000000000000000000000000000000000000005dc"
            }
          }
```
**The `from` fields are incorrect**

## Real on chain example (for [TX 0xbba6a...](https://etherscan.io/tx/0xbba6a4ca0ee856f6d0cab991e60fcade2db2b5745b6afee321b899dd37eedd80)):

The above Tx has the following [ethon:msgPayload](http://ethon.consensys.net/msgPayload):
```
Function: transfer(address _to, uint256 _value)

MethodID: 0xa9059cbb
[0]:0000000000000000000000002730826691bb0e970ef1608d5a3b944c29db7ed5
[1]:0000000000000000000000000000000000000000000002ad12e6a90c9c89f800
```
The token contract is Account `0x5e575279bf9f4acf0a130c186861454247394c06`
this means that Account `0x2730826691bb0e970ef1608d5a3b944c29db7ed5` sends `12637381652300000000000` Golem units to Account `0x2730826691bb0e970ef1608d5a3b944c29db7ed5`.

As can be seen [here](https://etherscan.io/address/0xa74476443119a942de498590fe1f2454d7d4ac0d#code) (line 32), the token contract that is the [ethon:to](http://ethon.consensys.net/to) of the [ethon:Tx](http://ethon.consensys.net/Tx) uses a `mapping (address => uint256) balances;) for storing the golem token 

If account `0x5e575279bf9f4acf0a130c186861454247394c06` successfully sends something, its balance cannot be 0 before the transfer. However, in the stateDiff it is:

```
       "storage": {
          "0x3122cf52ca415f61f9a275b9b68122794febbe33bf20650f5a17717ee0875719": {
            "*": {
              "from": "0x0000000000000000000000000000000000000000000000000000000000000000",
              "to": "0x00000000000000000000000000000000000000000008ed5239ab7f16f7e19df0"
            }
          },
          "0xf54353a2b20ca39c6925309d40b328469cccb296e1e6cc2ee095dd38422a7526": {
            "*": {
              "from": "0x0000000000000000000000000000000000000000000000000000000000000000",
              "to": "0x0000000000000000000000000000000000000000000005314f4a71a02b2bf400"
            }
          }
```
Here is the full result of `curl --data '{"method":"trace_replayTransaction","params":["0xbba6a4ca0ee856f6d0cab991e60fcade2db2b5745b6afee321b899dd37eedd80",["trace","stateDiff"]],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545`

https://gist.github.com/jo-tud/8efbe903e81f7271ffc79d8b133ddac2

sorpaas:
Fixes #6520 

The bug happens because in our current implementation, we only build PodAccount from modified storage keys. As a result, storage values before are not available in tracing, thus displayed `0x00..00`. To fix it, we need to merge all old state storage keys with new state storage keys.

`State::ensure_cached` is changed to take a `FnMut` closure. This should really be a `FnOnce` but the compiler thinks we're using it multiple times, so currently I use the supertrait `FnMut`.

### DIFF
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -863,12 +863,12 @@ impl<B: Backend> State<B> {
 	}
 
 	/// Populate a PodAccount map from this state, with another state as the account and storage query.
-	pub fn to_pod_diff(&mut self, query: &PodState) -> trie::Result<PodState> {
+	pub fn to_pod_diff<X: Backend>(&mut self, query: &State<X>) -> trie::Result<PodState> {
 		assert!(self.checkpoints.borrow().is_empty());
 
 		// Merge PodAccount::to_pod for cache of self and `query`.
 		let all_addresses = self.cache.borrow().keys().cloned()
-			.chain(query.get().keys().cloned())
+			.chain(query.cache.borrow().keys().cloned())
 			.collect::<BTreeSet<_>>();
 
 		Ok(PodState::from(all_addresses.into_iter().fold(Ok(BTreeMap::new()), |m: trie::Result<_>, address| {
@@ -882,11 +882,17 @@ impl<B: Backend> State<B> {
 								let self_keys = acc.storage_changes().keys().cloned()
 									.collect::<BTreeSet<_>>();
 
-								if let Some(ref query_acc) = query.get().get(&address) {
-									let query_keys = query_acc.storage.keys().cloned()
-										.collect::<BTreeSet<_>>();
-
-									self_keys.union(&query_keys).cloned().collect::<Vec<_>>()
+								if let Some(ref query_storage) = query.cache.borrow().get(&address)
+									.and_then(|opt| {
+										if let Some(ref query_acc) = opt.account {
+											Some(query_acc.storage_changes().keys().cloned()
+												 .collect::<BTreeSet<_>>())
+										} else {
+											None
+										}
+									})
+								{
+									self_keys.union(&query_storage).cloned().collect::<Vec<_>>()
 								} else {
 									self_keys.into_iter().collect::<Vec<_>>()
 								}
@@ -923,7 +929,7 @@ impl<B: Backend> State<B> {
 	/// Consumes self.
 	pub fn diff_from<X: Backend>(&self, mut orig: State<X>) -> trie::Result<StateDiff> {
 		let pod_state_post = self.to_pod();
-		let pod_state_pre = orig.to_pod_diff(&pod_state_post)?;
+		let pod_state_pre = orig.to_pod_diff(self)?;
 		Ok(pod_state::diff_pod(&pod_state_pre, &pod_state_post))
 	}
 
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -872,56 +872,47 @@ impl<B: Backend> State<B> {
 			.collect::<BTreeSet<_>>();
 
 		Ok(PodState::from(all_addresses.into_iter().fold(Ok(BTreeMap::new()), |m: trie::Result<_>, address| {
-			match m {
-				Err(e) => Err(e),
-				Ok(mut m) => {
-					let account = self.ensure_cached(&address, RequireCache::Code, true, |acc| {
-						acc.map(|acc| {
-							// Merge all modified storage keys.
-							let all_keys = {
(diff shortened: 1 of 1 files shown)


5chdn:
Would love to backport this to 1.11 - what do you think ... anyone? :)

ethcore/src/state/mod.rs
```diff
@@ -862,40 +862,75 @@ impl<B: Backend> State<B> {
...
 
-	fn query_pod(&mut self, query: &PodState, touched_addresses: &[Address]) -> trie::Result<()> {
-		let pod = query.get();
+		// Merge PodAccount::to_pod for cache of self and `query`.
+		let all_addresses = self.cache.borrow().keys().cloned()
+			.chain(query.cache.borrow().keys().cloned())
+			.collect::<BTreeSet<_>>();
+
+		Ok(PodState::from(all_addresses.into_iter().fold(Ok(BTreeMap::new()), |m: trie::Result<_>, address| {
+			match m {
```
tomusdrw: Could be just `let mut m = m?;`

ethcore/src/state/mod.rs
```diff
@@ -862,40 +862,75 @@ impl<B: Backend> State<B> {
...
+					let account = self.ensure_cached(&address, RequireCache::Code, true, |acc| {
+						acc.map(|acc| {
+							// Merge all modified storage keys.
+							let all_keys = {
+								let self_keys = acc.storage_changes().keys().cloned()
+									.collect::<BTreeSet<_>>();
+
+								if let Some(ref query_storage) = query.cache.borrow().get(&address)
+									.and_then(|opt| {
+										if let Some(ref query_acc) = opt.account {
```
tomusdrw: Could be just `?` since it now supports `Options`

ethcore/src/state/mod.rs
```diff
@@ -862,40 +862,75 @@ impl<B: Backend> State<B> {
...
+							};
+
+							// Storage must be fetched after ensure_cached to avoid borrow problem.
+							(*acc.balance(), *acc.nonce(), all_keys, acc.code().map(|x| x.to_vec()))
+						})
+					})?;
+
+					if let Some((balance, nonce, storage_keys, code)) = account {
+						let storage = storage_keys.into_iter().fold(Ok(BTreeMap::new()), |s: trie::Result<_>, key| {
+							match s {
```
tomusdrw: `let mut s = s?;` ?

Issue #6825, opened 2017-10-19
Node with no tracing DB returns empty result


keorn:
Node without tracing e.g. `parity --config dev`
```
curl --data '{"method":"trace_filter","params":[{"fromBlock":"0x2ed0c4","toBlock":"0x2ed128","toAddress":["0x8bbB73BCB5d553B5A556358d27625323Fd781D37"]}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
                       
{"jsonrpc":"2.0","result":[],"id":1}
```

Expected error.

folsen:
Closing after #8504

PR #8504, opened 2018-04-27
Directly return None if tracing is disabled

sorpaas:
(rel #6825, but this uses a different behavior to align with FatDB `parity_listAccounts`.)

This aligns the behavior of `trace_filter` with FatDB behavior -- when it is not enabled, return `null` in RPC instead of an empty array.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1873,9 +1873,7 @@ impl BlockChainClient for Client {
 	}
 
 	fn filter_traces(&self, filter: TraceFilter) -> Option<Vec<LocalizedTrace>> {
-		let tracedb = self.tracedb.read();
-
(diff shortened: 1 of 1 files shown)


ethcore/src/client/client.rs
```diff
@@ -1873,6 +1873,12 @@ impl BlockChainClient for Client {
 	}
 
 	fn filter_traces(&self, filter: TraceFilter) -> Option<Vec<LocalizedTrace>> {
+		let tracedb = self.tracedb.read();
```
tomusdrw: In all methods it changes lock order and may introduce deadlock with `restore_db` (where we lock `chain_db` and `trace_db` for writing)

Might be better to release the locks as fast as possible.

PR #8517, opened 2018-04-30
Show imported messages for light client

### LINKED ISSUE #8488: Light client does not report new blocks after being fully synchronized.
It does not have any "Imported ..." statements, it just looks like it stalls (but it isn't).

Can we have some logging that indicates it's not stalling?

sorpaas:
Fixes #8488.

Show "Imported" messages via `LightChainNotify`.


5chdn:
Cool thanks!

PR #8522, opened 2018-05-01
Don't panic in import_block if invalid rlp

ascjones:
Spotted by @tomusdrw: invalid block/header RLP passed to `Unverified::new` would cause a panic. Fortunately where we receive blocks from peers there is already a [guard](https://github.com/paritytech/parity/blob/7fdb87ad38f8ea969a4e733c7bbf621b1668a82d/ethcore/sync/src/block_sync.rs#L486).

Anyway I've changed it so that invalid RLP will result in a `DecoderError` instead of a panicking.

Related to #8033

### DIFF
--- a/ethcore/src/verification/queue/kind.rs
+++ b/ethcore/src/verification/queue/kind.rs
@@ -119,14 +119,13 @@ pub mod blocks {
 
 	impl Unverified {
 		/// Create an `Unverified` from raw bytes.
-		pub fn new(bytes: Bytes) -> Self {
-			use views::BlockView;
+		pub fn from_rlp(bytes: Bytes) -> Result<Self, ::rlp::DecoderError> {
 
-			let header = view!(BlockView, &bytes).header();
-			Unverified {
+			let header : Header = ::rlp::Rlp::new(&bytes).val_at(0)?; 
+			Ok(Unverified {
 				header: header,
 				bytes: bytes,
-			}
+			})
 		}
 	}
 
--- a/ethcore/src/verification/queue/kind.rs
+++ b/ethcore/src/verification/queue/kind.rs
@@ -121,7 +121,7 @@ pub mod blocks {
 		/// Create an `Unverified` from raw bytes.
 		pub fn from_rlp(bytes: Bytes) -> Result<Self, ::rlp::DecoderError> {
 
-			let header : Header = ::rlp::Rlp::new(&bytes).val_at(0)?; 
+			let header = ::rlp::Rlp::new(&bytes).val_at(0)?; 
 			Ok(Unverified {
 				header: header,
 				bytes: bytes,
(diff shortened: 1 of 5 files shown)


ethcore/src/verification/queue/kind.rs
```diff
@@ -119,14 +119,13 @@ pub mod blocks {
 
 	impl Unverified {
 		/// Create an `Unverified` from raw bytes.
-		pub fn new(bytes: Bytes) -> Self {
-			use views::BlockView;
+		pub fn from_rlp(bytes: Bytes) -> Result<Self, ::rlp::DecoderError> {
 
-			let header = view!(BlockView, &bytes).header();
-			Unverified {
+			let header : Header = ::rlp::Rlp::new(&bytes).val_at(0)?;
```
andresilva: Drop the space before the colon `let header: Header`.

ascjones: I've removed the type annotation since it was redundant anyway.

ethcore/src/verification/queue/kind.rs
```diff
@@ -119,14 +119,13 @@ pub mod blocks {
 
 	impl Unverified {
 		/// Create an `Unverified` from raw bytes.
-		pub fn new(bytes: Bytes) -> Self {
-			use views::BlockView;
+		pub fn from_rlp(bytes: Bytes) -> Result<Self, ::rlp::DecoderError> {
 
-			let header = view!(BlockView, &bytes).header();
-			Unverified {
+			let header = ::rlp::Rlp::new(&bytes).val_at(0)?;
```
tomusdrw: Do you mind fixing similar issue for `ancient block` (`block_with_receipt`) import?
AFAIR we also have a `BlockView` there.

andresilva: Also `collect_blocks` (`https://github.com/paritytech/parity/blob/master/ethcore/sync/src/block_sync.rs#L472`). It's the only place that calls `import_block_with_receipts` but it creates a header view before doing so.

andresilva: Ok I think what comes from `BlockCollection` is valid (invalid blocks/headers are silently ignored when inserting) so `collect_blocks` is safe, not sure if we should update it.

ascjones: In fact, invalid header/block body RLP is rejected further upstream in [`block_sync`](https://github.com/paritytech/parity/blob/master/ethcore/sync/src/block_sync.rs#L234), which results in the offending peer [being deactivated](https://github.com/paritytech/parity/blob/master/ethcore/sync/src/chain.rs#L779).

So blocks pulled from `BlockCollection` should all be valid RLP so safe to use the views inside `collect_blocks`. 

Seems to me that we could remove plenty of 'unsafe' view usages in this path since we could pass down the pre-decoded headers/block bodies, alongside/instead of the raw bytes. But that would be a separate PR.

ascjones: @tomusdrw see https://github.com/paritytech/parity/pull/8522/commits/22b2c3a7562ea1caa5379eb5d91b0b9d2957299a

PR #8524, opened 2018-05-01
Keep all enacted blocks notify in order

### LINKED ISSUE #8514: WS Subscriptions Missing Events and Block Headers
- When subscribed to new headers and events via ws api, parity will push incomplete event sequence and also headers and events are coming out of order.
- Chain reorg block headers are not double pushed (replacement header missing).
- `Imported X + another X blocks` will push in sequence (X, X-1, X-2...) instead of (...X-2, X-1, X)
- **Events from "synced" blocks between "imported" ones, in case parity missed some, are not pushed at all**

This is a completely unreliable behaviour for checking for events and removals.

> I'm running:
> - **Which Parity version?**: v1.9.7-stable

sorpaas:
Fixes #8514.

Collect ImportRoute paths to Vector instead of hash map, so that reorg hashes are correctly duplicated, and we don't get it out of order.


ethcore/src/client/client.rs
```diff
@@ -255,14 +255,9 @@ impl Importer {
...
 		// will be in the hashmap
-		let map = import_results.iter().fold(HashMap::new(), |mut map, route| {
-			for hash in &route.enacted {
-				map.insert(hash.clone(), true);
-			}
-			for hash in &route.retracted {
-				map.insert(hash.clone(), false);
-			}
-			map
+		let map = import_results.iter().flat_map(|route| {
```
tomusdrw: I think that won't work. The comment above explains what we tried to achieve with the previous code.
We need to remove duplicated blocks and make sure we only take last status of a block (i.e. it could go from `enacted` to `retracted` and then back to `enacted` in `ImportRoute`) - with current approach later on, after partition the block will end up in both `enacted` and `retracted`.

ethcore/src/client/client.rs
```diff
@@ -255,14 +255,9 @@ impl Importer {
...
+			route.enacted.iter().map(|h| (*h, true))
+				.chain(route.retracted.iter().map(|h| (*h, false))).collect::<Vec<_>>()
```
rphmeier: is collecting to a vector actually necessary here? and the purpose of using a `HashMap` was to make sure the most recent enacted/retracted value holds. The changes here seem as though you can get multiple entries for the same hash (= bad)

ethcore/src/client/client.rs
sorpaas: @tomusdrw I think it probably doesn't matter a lot. `calculate_enacted_retracted` is only used in two places in `client.rs` (L334 and L2080). The only usage for calculated `enacted` and `retracted` is to pass it to `notify.new_blocks`.

I can't think of a good way to remove duplicate while preserving the actual route history -- it looks like the simplest way is to call `tree_route` for another time with the first and last block, but that may be already too expensive?

So I'm thinking about, it maybe a better idea to change the signature of `ChainNotify`? Instead of keep `enacted: &[H256], retracted: &[H256]`, we change it to `route: &[(H256, Status::Enacted/Status::Retracted)]`. Many listeners would just need the later, like the RPC `pubsub` mod, where it's expected to re-push headers in every re-org. If listeners need the former, it's also trivial to construct from the later. What do yo think?

tomusdrw: ^^ Makes sense to me, that would be the best I suppose, but we need to review all cases where `enacted` and `retratected` is used to make sure we are not breaking anything by introducing that change.

sorpaas:
I added a new type `ChainRoute`, which wraps a `Vec<H256, ChainRouteType>`. The new type exposes methods to get the original `enacted` and `retracted` behavior. The old behavior is used in `miner` and `sync`, and the new behavior is used in `rpc`.

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -141,19 +141,20 @@ impl<C> ChainNotificationHandler<C> {
 		}
 	}
 
-	fn notify_logs<F, T>(&self, enacted: &[H256], logs: F) where
-		F: Fn(EthFilter) -> T,
+	fn notify_logs<F, T, Ex>(&self, enacted: &[(H256, Ex)], logs: F) where
```
sorpaas: It looks like we cannot change `enacted: &[(H256, Ex)]` to an generic `Iterator`. We need to iterate over `enacted` many times, but most `Iterator` are not clonable. This do mean that in `LightChainNotify`, there's one additional `collect::Vec<_>`, but I think this tradeoff is okay.

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -141,19 +141,20 @@ impl<C> ChainNotificationHandler<C> {
...
+		F: Fn(EthFilter, &Ex) -> T,
+	    Ex: Send,
```
tomusdrw: Funky indent.

ethcore/src/client/chain_notify.rs
```diff
@@ -29,15 +31,76 @@ pub enum ChainMessageType {
...
+		});
+
+		// Split to enacted retracted (using hashmap value)
+		let (enacted, retracted) = map.into_iter().partition(|&(_k, v)| v);
+		// And convert tuples to keys
+		(map_to_vec(enacted), map_to_vec(retracted))
+	}
+
+	/// Whether this particular route contains non-duplicate enacted blocks.
+	pub fn contains_enacted(&self) -> bool {
```
tomusdrw: Both functions don't seem to be used and are also quite inefficent. I'd remove them.

ethcore/src/client/chain_notify.rs
```diff
@@ -29,15 +31,76 @@ pub enum ChainMessageType {
...
+		ChainRoute(import_results.iter().flat_map(|route| {
+			route.retracted.iter().map(|h| (*h, ChainRouteType::Retracted))
+				.chain(route.enacted.iter().map(|h| (*h, ChainRouteType::Enacted)))
+		}).collect())
+	}
+}
+
+impl ChainRoute {
+	/// Gather all non-duplicate enacted and retracted blocks.
+	pub fn to_enacted_retracted(&self) -> (Vec<H256>, Vec<H256>) {
```
tomusdrw: It seems that we call it for every `ChainRoute` anyway and sometimes we call it multiple times. Maybe calculate that in the constructor as well and just return `(&[H256], &[H256])` here?

PR #8527, opened 2018-05-02
Rlp decode returns Result

dvdplm:
Part of the fix for issue #8033. The change of return types have a great many consequences on a lot of code and sometimes it is difficult to know what the right thing to do is when decoding fails, so careful review is important here. This is also my first PR on the code base, so pushing this early to collect feedback: is this roughly the right direction?


dvdplm:
[clabot:check]

dvdplm:
@ascjones do you mind taking a look at this and let me know if this is roughly the right direction here? I am currently going through crates and fixing things as I go, but it'd be great to have some feedback and comments on when it's ok to panic/log/transform `Err` to `None` etc.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -430,7 +430,10 @@ impl<'a> Iterator for EpochTransitionIter<'a> {
 				return None
 			}
 
-			let transitions: EpochTransitions = ::rlp::decode(&val[..]);
+			let transitions: EpochTransitions = match ::rlp::decode(&val[..]) {
+				Ok(decoded) => decoded,
+				Err(_) => return None
+			};
```
ascjones: Shouldn't break out of the iterator with invalid RLP. Since we wrote the `EpochTransitions` to the db in the first place then either the db is corrupted or the data structure has changed, so should panic.

ethcore/light/src/client/header_chain.rs
```diff
@@ -524,7 +524,16 @@ impl HeaderChain {
 											None
 										}
 										Ok(None) => panic!("stored candidates always have corresponding headers; qed"),
-										Ok(Some(header)) => Some((epoch_transition, ::rlp::decode(&header))),
+										Ok(Some(header)) => {
+											match ::rlp::decode(&header) {
```
debris: this value was fetched from the database. It should never be invalid, therefore we should use `expect` here

ethcore/light/src/client/header_chain.rs
```diff
@@ -742,7 +751,17 @@ impl HeaderChain {
 	/// so including it within a CHT would be redundant.
 	pub fn cht_root(&self, n: usize) -> Option<H256> {
 		match self.db.get(self.col, cht_key(n as u64).as_bytes()) {
-			Ok(val) => val.map(|x| ::rlp::decode(&x)),
+			Ok(db_fetch) => {
+				db_fetch.map_or(None, |bytes| {
+					match ::rlp::decode(&bytes) {
```
debris: same here

ethcore/light/src/client/header_chain.rs
```diff
@@ -793,7 +812,17 @@ impl HeaderChain {
 	pub fn pending_transition(&self, hash: H256) -> Option<PendingEpochTransition> {
 		let key = pending_transition_key(hash);
 		match self.db.get(self.col, &*key) {
-			Ok(val) => val.map(|x| ::rlp::decode(&x)),
+			Ok(db_fetch) => {
+				db_fetch.map_or(None, |bytes| {
+					match ::rlp::decode(&bytes) {
```
debris: and here

debris:
it's definitely the right direction. We want to use use `decode(..)?` when decoding data received from the network and want to use `decode(..).expect('')` for data fetched from disk :)

ethcore/src/db.rs
```diff
@@ -218,11 +218,18 @@ impl Writable for DBTransaction {
...
+		where T: rlp::Decodable, R: Deref<Target = [u8]> {
 		let result = self.get(col, &key.key());
 
 		match result {
-			Ok(option) => option.map(|v| rlp::decode(&v)),
+			Ok(option) => option.map(|v| match rlp::decode(&v){
+				Ok(decoded_rlp) => decoded_rlp,
+				//REVIEW: Should we panic here?
+				Err(decode_err) => {
+					panic!("decoding db value failed, key={:?}, err={}", &key.key() as &[u8], decode_err)
```
dvdplm: @debris the situation here is similar to the above, db value read from disc, so panicking is ok. Q: do you prefer me to shorten this up with a `rlp::decode(&v).expect("decoding db value failed")` or is it useful to have the additional info about which variant of the decode error and the key?

util/journaldb/src/overlaydb.rs
```diff
@@ -135,9 +135,13 @@ impl OverlayDB {
 
 	/// Get the refs and value of the given key.
 	fn payload(&self, key: &H256) -> Option<Payload> {
-		self.backing.get(self.column, key)
+		match self.backing.get(self.column, key)
```
dvdplm: @debris so here there's more of the same I take it? 

Better:

```rust
	fn payload(&self, key: &H256) -> Option<Payload> {
            self.backing.get(self.column, key)
	        .expect("Low-level database error. Some issue with your hard disk?")
	        .map(|d| decode(&d).expect("decoding db value failed"))
	}
```

ethcore/src/db.rs
debris: I believe it could be shortened with:

```rust
result.and_then(|option| match option {
  None => Ok(None),
  Some(v) => rlp::decode(&v).map(Some)
}).expect("decoding db value failed)
```

ethcore/src/encoded.rs
```diff
@@ -48,7 +48,7 @@ impl Header {
 	pub fn new(encoded: Vec<u8>) -> Self { Header(encoded) }
 
 	/// Upgrade this encoded view to a fully owned `Header` object.
-	pub fn decode(&self) -> FullHeader { ::rlp::decode(&self.0) }
+	pub fn decode(&self) -> FullHeader { ::rlp::decode(&self.0).unwrap_or(FullHeader::default()) } // REVIEW: is the default ok here? Useful? The real fix would be to make `decode` return a `Result` as well.
```
debris: It should definitely return `Result`. It it requires too many changes, let's replace it with `.expect` for now and let's create another issues to track this

ethcore/src/encoded.rs
```diff
@@ -205,7 +205,7 @@ impl Block {
 	pub fn header_view(&self) -> HeaderView { self.view().header_view() }
 
 	/// Decode to a full block.
-	pub fn decode(&self) -> FullBlock { ::rlp::decode(&self.0) }
+	pub fn decode(&self) -> FullBlock { ::rlp::decode(&self.0).unwrap_or(FullBlock::default()) }
```
debris: Same here. Returning `default` is not a good pattern and may lead too undefined behaviours

ethcore/src/engines/tendermint/mod.rs
```diff
@@ -142,8 +142,10 @@ impl <F> super::EpochVerifier<EthereumMachine> for EpochVerifier<F>
 	}
 
 	fn check_finality_proof(&self, proof: &[u8]) -> Option<Vec<H256>> {
-		let header: Header = ::rlp::decode(proof);
-		self.verify_light(&header).ok().map(|_| vec![header.hash()])
+		match ::rlp::decode(proof) {
+			Ok(header) => self.verify_light(&header).ok().map(|_| vec![header.hash()]),
+			Err(_) => None // REVIEW: log perhaps? Not sure what the policy is.
```
debris: This is a good question. According to interface documentation:

```rust
	/// Returns `Some(hashes)` if the proof proves finality of these hashes.
	/// Returns `None` if the proof doesn't prove anything.
```

it should be ok to return `None` in case of invalid proof. Can you confirm @rphmeier ?

ethcore/src/state/account.rs
```diff
@@ -145,8 +145,14 @@ impl Account {
 
 	/// Create a new account from RLP.
 	pub fn from_rlp(rlp: &[u8]) -> Account {
-		let basic: BasicAccount = ::rlp::decode(rlp);
-		basic.into()
+		match ::rlp::decode::<BasicAccount>(rlp) {
+			Ok(basic_account) => basic_account.into(),
+			Err(e) => {
+				// REVIEW: Is panicking ok here? I'd like to change the return value here to `Result<Account, SomeError>` but not sure if that's desirable or the right time.
```
debris: I believe it's not ok, cause we don't know where decoded slice of bytes comes from. It's better to change it to `Result<Account, SomeError>`

ethcore/src/state/account.rs
```diff
@@ -202,8 +208,8 @@ impl Account {
 			return Ok(value);
 		}
 		let db = SecTrieDB::new(db, &self.storage_root)?;
-
-		let item: U256 = db.get_with(key, ::rlp::decode)?.unwrap_or_else(U256::zero);
+		let unwrapping_decoder = |bytes:&[u8]| ::rlp::decode(&bytes).unwrap_or_else(|_| U256::zero());
```
debris: I believe we should have `.expect` instead of `unwrap_or_else` here. If the value exists in the database, we always expect it to be encoded properly.

ethcore/src/state/account.rs
```diff
@@ -478,7 +484,8 @@ impl Account {
 
 		let trie = TrieDB::new(db, &self.storage_root)?;
 		let item: U256 = {
-			let query = (&mut recorder, ::rlp::decode);
+			let unwrapping_decoder = |bytes:&[u8]| ::rlp::decode(bytes).unwrap_or_else(|_| U256::zero() );
```
debris: same here :)

util/journaldb/src/archivedb.rs
```diff
@@ -45,14 +45,14 @@ pub struct ArchiveDB {
 
 impl ArchiveDB {
 	/// Create a new instance from a key-value db.
-	pub fn new(backing: Arc<KeyValueDB>, col: Option<u32>) -> ArchiveDB {
-		let latest_era = backing.get(col, &LATEST_ERA_KEY).expect("Low-level database error.")
-			.map(|val| decode::<u64>(&val));
+	pub fn new(backing: Arc<KeyValueDB>, column: Option<u32>) -> ArchiveDB {
+		let latest_era = backing.get(column, &LATEST_ERA_KEY).expect("Low-level database error.")
+			.map(|val| decode::<u64>(&val).unwrap_or(0)); // REVIEW: is `0` an ok value for `latest_era` if the Rlp decoding fails? The alternative is to either panic or change the return value of the function to return `Result<ArchiveDB, SomeError>`.
```
debris: no, there should be another expect here

util/journaldb/src/earlymergedb.rs
```diff
@@ -263,7 +263,7 @@ impl EarlyMergeDB {
 		let mut refs = HashMap::new();
 		let mut latest_era = None;
 		if let Some(val) = db.get(col, &LATEST_ERA_KEY).expect("Low-level database error.") {
-			let mut era = decode::<u64>(&val);
+			let mut era = decode::<u64>(&val).unwrap_or(0); // REVIEW: The error is lost here. Log and return? Panic? Or is `0` an ok fallback value?
```
debris: there should be expect here

util/journaldb/src/overlayrecentdb.rs
```diff
@@ -186,7 +186,7 @@ impl OverlayRecentDB {
 		let mut earliest_era = None;
 		let mut cumulative_size = 0;
 		if let Some(val) = db.get(col, &LATEST_ERA_KEY).expect("Low-level database error.") {
-			let mut era = decode::<u64>(&val);
+			let mut era = decode::<u64>(&val).unwrap_or(0); // REVIEW: The error is lost here. Log and return? Panic? Or is `0` an ok fallback value?
```
debris: it should be expect

util/journaldb/src/refcounteddb.rs
```diff
@@ -62,17 +62,17 @@ pub struct RefCountedDB {
 
 impl RefCountedDB {
 	/// Create a new instance given a `backing` database.
-	pub fn new(backing: Arc<KeyValueDB>, col: Option<u32>) -> RefCountedDB {
-		let latest_era = backing.get(col, &LATEST_ERA_KEY).expect("Low-level database error.")
-			.map(|val| decode::<u64>(&val));
+	pub fn new(backing: Arc<KeyValueDB>, column: Option<u32>) -> RefCountedDB {
+		let latest_era = backing.get(column, &LATEST_ERA_KEY).expect("Low-level database error.")
+			.map(|val| decode::<u64>(&val).unwrap_or(0)); // REVIEW: is `0` an ok value for `latest_era` if the Rlp decoding fails? The alternative is to either panic or change the return value of the function to return `Result<ArchiveDB, SomeError>`.
```
debris: it should be expect

ethcore/light/src/client/header_chain.rs
```diff
@@ -228,15 +228,15 @@ impl HeaderChain {
 		let decoded_header = spec.genesis_header();
 
 		let chain = if let Some(current) = db.get(col, CURRENT_KEY)? {
-			let curr : BestAndLatest = ::rlp::decode(&current);
+			let curr : BestAndLatest = ::rlp::decode(&current)?;
```
debris: it should be expect, cause we're reading from the db :)

ethcore/light/src/client/header_chain.rs
```diff
@@ -228,15 +228,15 @@ impl HeaderChain {
...
 
 			let mut cur_number = curr.latest_num;
 			let mut candidates = BTreeMap::new();
 
 			// load all era entries, referenced headers within them,
 			// and live epoch proofs.
 			while let Some(entry) = db.get(col, era_key(cur_number).as_bytes())? {
-				let entry: Entry = ::rlp::decode(&entry);
+				let entry: Entry = ::rlp::decode(&entry)?;
```
debris: it should be expect, cause we're reading from the db :)

debris:
added several more grumbles. I believe that after they are resolved, it's ready to be merged :)

ethcore/src/encoded.rs
dvdplm: https://github.com/paritytech/parity/issues/8553

ethcore/src/encoded.rs
dvdplm: https://github.com/paritytech/parity/issues/8553

dvdplm:
@debris I need a second reviewer to approve this right? Should I ping someone myself or wait until someone gets around to it, what is the praxis? :)

(And thank you for all of your feedback on this, much appreciated and very useful.)

util/journaldb/src/overlayrecentdb.rs
```diff
@@ -195,7 +195,13 @@ impl OverlayRecentDB {
...
 					trace!("read_overlay: era={}, index={}", era, db_key.index);
-					let value = decode::<DatabaseValue>(&rlp_data);
+					let value = match decode::<DatabaseValue>(&rlp_data) {
+						Ok(v) => v,
+						Err(e) => {
+							trace!("read_overlay: Error decoding DatabaseValue err={}, era={}, index{}", e, era, db_key.index);
+							continue;
+						}
+					};
 					count += value.inserts.len();
```
ascjones: What's the reasoning for continuing the loop in this case, instead of the `expect` like in the other places? Couldn't figure it out from a quick read of the code.

dvdplm: @ascjones that's a good q but I don't have a great answer. I let it `continue` thinking that the previous code would have done the same but now that you point it out I think it would have panicked. Furthermore the code as I've written it is probably not going to yield a valid&usable `JournalOverlay` if some parts are missing. I'll change it to panic.

ascjones: Yeah at least that way it behaves exactly how it would have before (panic with invalid rlp)

PR #8528, opened 2018-05-02
Fix light sync with initial validator-set contract

### LINKED ISSUE #8468: Light clients broken on chains with genesis validator-set contract
> I'm running:
>
> - **Which Parity version?**: v1.10.0, v1.10.2, v1.11.0-unstable
> - **How installed?**: via binaries, from source
> - **Which network are you connected to?**: sokol

How to properly start `light` client on a custom sidechain, not using any hardcoded sync block? I tried using 1.10.0, 1.10.2, 1.11.0 (master)

I tried to set it in config toml file:
```
light = true
no_hardcoded_sync = true
```
also tried
```
light = true
hardcoded_sync = false
```
also with `--light --no-hardcoded-sync` cmd options,
but I get either
```
Loading config file from node.toml
You might have supplied invalid parameters in config file.
unknown field `no_hardcoded_sync`, ...
```
or
```
error: Found argument '--no-hardcoded-sync' which wasn't expected, or isn't valid in this context
	Did you mean --no-hardware-wallets?
```
Also compiled `master` and got the error message as if `--no-hardcoded-sync` is ignored:
```
2018-04-24 12:21:18 UTC Running in experimental Light Client mode.
Error starting light client: Core error: Cannot decode address[]
```

tomaka:
Fix #8468 

When initializing a light database from scratch, we create a fake transaction that performs the initial state transition. Right now this fake transition has a hardcoded gas of 50M. I don't really understand why this uses any gas at all or why this would need to be shared with the client (like the TODO mentions), so I used 0 instead.

cc @rphmeier who wrote this code I guess.

### DIFF
--- a/ethcore/src/spec/spec.rs
+++ b/ethcore/src/spec/spec.rs
@@ -776,7 +776,7 @@ impl Spec {
 				author: *genesis.author(),
 				timestamp: genesis.timestamp(),
 				difficulty: *genesis.difficulty(),
-				gas_limit: *genesis.gas_limit(),
+				gas_limit: U256::max_value(),
 				last_hashes: Arc::new(Vec::new()),
 				gas_used: 0.into(),
 			};
@@ -785,7 +785,7 @@ impl Spec {
 			let tx = Transaction {
 				nonce: self.engine.account_start_nonce(0),
 				action: Action::Call(a),
-				gas: U256::max_value(),
+				gas: U256::from(50_000_000), // TODO: share with client.
 				gas_price: U256::default(),
 				value: U256::default(),
 				data: d,
--- a/ethcore/src/spec/spec.rs
+++ b/ethcore/src/spec/spec.rs
@@ -785,7 +785,7 @@ impl Spec {
 			let tx = Transaction {
 				nonce: self.engine.account_start_nonce(0),
 				action: Action::Call(a),
-				gas: U256::from(50_000_000), // TODO: share with client.
+				gas: U256::max_value(),
 				gas_price: U256::default(),
 				value: U256::default(),
 				data: d,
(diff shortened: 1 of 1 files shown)


ethcore/src/spec/spec.rs
```diff
@@ -785,7 +785,7 @@ impl Spec {
 			let tx = Transaction {
 				nonce: self.engine.account_start_nonce(0),
 				action: Action::Call(a),
-				gas: U256::from(50_000_000), // TODO: share with client.
+				gas: U256::default(),
```
rphmeier: doesn't seem right to provide no gas. then no work will be done and we will almost certainly end up in the wrong state. 50M is an arbitrary amount but we could also set to `U256::max_value` -- the key thing is really that the `env_info`'s gas limit is at least the amount of gas provided to the transaction, which seems to be the failure here.

rphmeier: I think for `genesis_epoch_data`, the goal is to get an initial list of validators. when the `ValidatorSet` is a contract, we have to call a function on that contract to extract the validators. This function needs some gas to read out the validators from storage, allocated memory for them, and ABI-encode the return data.

The thought process I've been following so far with ValidatorSet contracts is that supplying infinite gas is fine because the contract is actually part of the consensus process and shouldn't be subject to the metering rules or have an escape-hatch from turing-completeness. The same should apply here.

tomaka:
I'm check whether this solves the issue.

tomaka:
This doesn't fix the issue, but we can merge anyway I guess?

ethcore/src/spec/spec.rs
```diff
@@ -785,7 +785,7 @@ impl Spec {
 			let tx = Transaction {
 				nonce: self.engine.account_start_nonce(0),
 				action: Action::Call(a),
-				gas: U256::max_value(),
+				gas: U256::from(50_000_000), // TODO: share with client.
```
rphmeier: this can also be `max_value` and actually IMO it makes more sense

tomaka:
Should be good now.
Tried again, issue gets fixed with the changes.

andresilva:
Sorry for some reason I thought the gas limit was already `U256::max_value()` (and I guess I didn't read your comment 🙈).

PR #8530, opened 2018-05-02
Don't block sync when importing old blocks

tomusdrw:
- Moves old blocks import outside of sync lock (other thread via IoService) to prevent sync stalls.
- Introduces a generic way to execute stuff on the IoService, so by accident partially addresses https://github.com/paritytech/parity/issues/8186
- Removes a requirement for `Message` of `IoService<Message>` to be `Clone` (introduces `Arc` for broadcasted `UserMessage`)
- Adds a bunch of perf timers.


ethcore/src/client/ancient_import.rs
```diff
@@ -53,17 +53,47 @@ impl AncientVerifier {
...
+
+		// when there is no verifier initialize it.
+		// We use a bool flag to avoid double locking in the happy case
+		if !verified {
+			let mut cur_verifier = self.cur_verifier.write();
+			if cur_verifier.is_none() {
+				*cur_verifier = Some(self.initial_verifier(header, chain)?);
+			}
+			// Call again to verify.
+			return self.verify(rng, header, chain);
```
ngotchac: Couldn't that go in the `else` case branch above directly?

tomusdrw: Unfortunatelly the lock for `cur_verifier` is still held in `else` branch above, so you cannot lock it again for writing.

But it seems that there is a deadlock here as well, since we call re-cursively `verify` when still holding `cur_verifier` lock, I've updated it.

Issue #8222, opened 2018-03-26
util/patricia_tree `eliminate expect()`


niklasad1:
https://github.com/paritytech/parity/blob/master/util/patricia_trie/src/lookup.rs#L58

ascjones:
I added this as part of #8033. Originally the call to Node::decoded was unsafe for invalid rlp, because of the unwraps in Rlp e.g. https://github.com/paritytech/parity/blob/master/util/rlp/src/rlpin.rs#L126

The DecoderErrors are now propogated and can be handled further up. In this case I originally added a DecoderError case to `TrieError` but ran into a couple of issues and decided to add the expect for expediency, reasoning that it behaves at worst as badly as before.

Now the wider changes are nearly done I can look at this again (or someone else). IIRC the slight issues I had were about the data to store in the DecoderError case since it is not Clonable, and the H256 key is not always available in the context. So should be simple once this is figured out.

PR #8536, opened 2018-05-03
Remove expect

dvdplm:
Fixes issue #8222

### DIFF
--- a/util/patricia_trie/src/lookup.rs
+++ b/util/patricia_trie/src/lookup.rs
@@ -55,29 +55,32 @@ impl<'a, Q: Query> Lookup<'a, Q> {
 			// without incrementing the depth.
 			let mut node_data = &node_data[..];
 			loop {
-				match Node::decoded(node_data).expect("rlp read from db; qed") {
-					Node::Leaf(slice, value) => {
-						return Ok(match slice == key {
-							true => Some(self.query.decode(value)),
-							false => None,
-						})
-					}
-					Node::Extension(slice, item) => {
-						if key.starts_with(&slice) {
-							node_data = item;
-							key = key.mid(slice.len());
-						} else {
-							return Ok(None)
+				match Node::decoded(node_data) {
+					Ok(decoded_node) => match decoded_node {
+						Node::Leaf(slice, value) => {
+							return Ok(match slice == key {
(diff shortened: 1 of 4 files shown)


ascjones:
Is returning `Ok(None)` the right thing in the case of a `DecoderError`? So the key exists but the data is invalid. I think we are losing some information there.

dvdplm:
> I think we are losing some information there.

Do you think I should dig deeper into this and see if I can make `DecoderErrors` bubble up as the boxed trie error?

ascjones:
That would be the ideal since this fn is already returning a Result, but if I recall it wasn't straightforward (can't remember why). 

Stepping back, we are decoding the Node RLP that was [self encoded](https://github.com/paritytech/parity/blob/master/util/patricia_trie/src/triedbmut.rs#L132). So again it's a case of either a corrupted DB or the node RLP encoding has changed and the DB wasn't migrated. In this exceptional scenario I think panicking would be better than returning `Ok(None)`. 

But if we can add the Result case and it is handled correctly further up then that would be best.

dvdplm:
> That would be the ideal since this fn is already returning a Result, but if I recall it wasn't straightforward (can't remember why).

Attempted this in https://github.com/paritytech/parity/pull/8536/commits/1d54fc973aa6db2a8586bf55f0536cc11696bb7a – do you mind taking a look? I am struggling to write a decent test for it though (not sure how to make the look up fail with a decode error!).

util/patricia_trie/src/lookup.rs
```diff
@@ -55,29 +55,32 @@ impl<'a, Q: Query> Lookup<'a, Q> {
...
-							false => None,
-						})
-					}
-					Node::Extension(slice, item) => {
-						if key.starts_with(&slice) {
-							node_data = item;
-							key = key.mid(slice.len());
-						} else {
-							return Ok(None)
+				match Node::decoded(node_data) {
```
ascjones: I think this nested match could be avoided with a `?` here and change the `From` impl to `Box<TrieError>`?

dvdplm: Good call, totally works! :)

ascjones:
I did the same when I first started but you need two reviewers before merging.

Issue #8185, opened 2018-03-22
Initial ethcore-io refactor


debris:
Currently ethcore-io event loop is based on deprecated part of `mio`. Let's replace it with `tokio`

tomaka:
Can you actually use tokio (actual question, I don't know ethcore-io enough to know)?
I think the `Core` and the `Handle` are not `Send` or `Sync`, so you cannot send sockets or (futures that contain a socket) amongst threads.

0x7CFE:
@tomaka, please correct me if I'm wrong, but I do not see a problem here.

Looks like Parity internals use `util::io` facade to get even loop functionality. And if we [look at the API](https://github.com/paritytech/parity/blob/master/util/io/src/service.rs#L44)  we may see, that most of the `Message` generics are bounded  by `Send + Sync`.

I even tried forcing `+ Sync` bound to `IoChannel`, `IoMessage`, `Handlers` and other stuff and it all compiled, which makes me think, that at the moment we do not have messages that are neither `Send` nor `Sync`.

Could you please give me some more details on a part of the system where you expect problems? What particular `Core` and `Handle` you've mentioned before?

tomaka:
Right now the sockets are identified by a `StreamToken`, which is just a `usize`, but it is essentially the mio events loop that *owns* the sockets.

If you switch to tokio, however, you would need to manipulate `TcpStream` objects instead of `usize`s, and these are not `Send` or `Sync` if I'm not mistaken (could be wrong here).
Since `TcpStream` implements traits from the `futures` crate, the code that uses `ethcore_io` is supposed to manipulate them directly. What I mean by that is that the role they play is more than simply identifiers.

It should be possible to wrap around tokio's API so that each `TcpStream` stays in a dedicated I/O processing threads, and use channels from `std::sync::mpsc` in order to communicate data, but that would add a large overhead I think.

> Could you please give me some more details on a part of the system where you expect problems? What particular Core and Handle you've mentioned before?

https://docs.rs/tokio-core/0.1.16/tokio_core/reactor/struct.Core.html
https://docs.rs/tokio-core/0.1.16/tokio_core/reactor/struct.Handle.html

tomaka:
Also, once you build a `future` you need to send it back to `ethcore_io` in the correct thread so that it is passed to the `Core::run` function. This further complicates the problem.

tomaka:
[...]

PR #8537, opened 2018-05-03
Make mio optional in ethcore-io

tomaka:
cc #8185 
cc #7915 

Makes the `mio` dependency optional in `ethcore-io`.
This should make it possible to compile `ethcore` without depending on `mio`, which is one step towards compiling Parity for WASM.

The `network-devp2p` crate is the only one that requires `mio`. The wasm/emscripten port of Parity will not use devp2p, therefore this is not a problem.


tomaka:
This PR essentially reimplements the existing timer and messages codes, except without using `mio`.

tomaka:
I didn't realize it, but `ethcore-sync` requires the `to_handler` function.
Since implementing it is a bit tricky, I decided not to reimplement it, and `ethcore-sync` requires `mio` as a dev-dependency.

util/io/src/lib.rs
```diff
@@ -106,8 +134,9 @@ impl From<::std::io::Error> for IoError {
 	}
 }
 
-impl<Message> From<NotifyError<service::IoMessage<Message>>> for IoError where Message: Send + Clone {
-	fn from(_err: NotifyError<service::IoMessage<Message>>) -> IoError {
+#[cfg(feature = "mio")]
+impl<Message> From<NotifyError<service_mio::IoMessage<Message>>> for IoError where Message: Send + Clone {
+	fn from(_err: NotifyError<service_mio::IoMessage<Message>>) -> IoError {
 		IoError::Mio(::std::io::Error::new(::std::io::ErrorKind::ConnectionAborted, "Network IO notification error"))
```
niklasad1: Move to `service_mio.rs`? Then you can get rid of the `annonation` too!

util/io/src/lib.rs
```diff
@@ -123,58 +152,120 @@ pub trait IoHandler<Message>: Send + Sync where Message: Send + Sync + Clone + '
...
 	/// Register a new stream with the event loop
+	#[cfg(feature = "mio")]
 	fn register_stream(&self, _stream: StreamToken, _reg: Token, _event_loop: &mut EventLoop<IoManager<Message>>) {}
 	/// Re-register a stream with the event loop
+	#[cfg(feature = "mio")]
 	fn update_stream(&self, _stream: StreamToken, _reg: Token, _event_loop: &mut EventLoop<IoManager<Message>>) {}
 	/// Deregister a stream. Called whenstream is removed from event loop
+	#[cfg(feature = "mio")]
 	fn deregister_stream(&self, _stream: StreamToken, _event_loop: &mut EventLoop<IoManager<Message>>) {}
 }
```
niklasad1: These `annotations` are not that nice, can we get rid off them by splitting the `IoHandler trait` into two separate traits in each module (service_mio and service_non_mio)?

tomusdrw: Sounds good to me, one trait should probably require the other and in `non-mio` implementation it can be just an empty marker trait.

util/io/src/service_mio.rs
```diff
@@ -267,7 +267,8 @@ impl<Message> Handler for IoManager<Message> where Message: Send + Clone + Sync
 				event_loop.shutdown();
 			},
 			IoMessage::AddHandler { handler } => {
-				let handler_id = self.handlers.write().insert(handler.clone()).unwrap_or_else(|_| panic!("Too many handlers registered"));
+				let handler_id = self.handlers.write().insert(handler.clone());
+				assert!(handler_id <= MAX_HANDLERS, "Too many handlers registered");
```
tomusdrw: Old handler ids are re-used?

tomaka: Yes, when we remove a handler its ID becomes available again.
In practice it doesn't really matter since we just add all the handlers at once at initialization, and remove them all at once at the end.

tomaka:
I don't see what is so bad about annotations. :confused: 
Splitting things in the two modules means code duplication, and since we don't test the non-mio path yet I think there are more chances of things breaking if we duplicate code. Someone might make a change in the mio code and forget to propagate it.

tomusdrw:
> Someone might make a change in the mio code and forget to propagate it.

Same is true about annotations. IMHO to make sure it's not broken we should add a `cargo check --features non-mio` or sth to a build step, otherwise both ways will cause issues.
I'm ok with any decision.

5chdn:
Can't merge :)

5chdn:
Test failed.

```
---- tests::send_message_to_handler stdout ----
	thread 'tests::send_message_to_handler' panicked at 'assertion failed: handler.0.load(atomic::Ordering::SeqCst)', util/io/src/lib.rs:213:3
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:390
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
   6: ethcore_io::tests::send_message_to_handler
   7: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1462
             at /checkout/src/libcore/ops/function.rs:223
             at /checkout/src/liballoc/boxed.rs:788
   8: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102


failures:
    tests::send_message_to_handler
```

PR #8538, opened 2018-05-03
Check that the Android build doesn't dep on c++_shared

tomaka:
The opinion of a shell expert would be appreciate to make sure that what I'm doing is correct.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -190,7 +190,7 @@ android-armv7:
     - triggers
   script:
     - cargo build --target=armv7-linux-androideabi
-    # TODO: check that `arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/release/parity | grep c++_shared` is empty
+    - if [ $(arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/debug/parity | grep -i 'c++_shared' | wc -l) -ne 0]; then echo "FAIL!!" fi
   tags:
     - rust-arm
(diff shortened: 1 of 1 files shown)


General-Beck:
CI LINT error 
remove [] pls
and build error https://gitlab.parity.io/parity/parity/-/jobs/86576

tomaka:
#8412 will fix the build error.

tomaka:
I'm fairly confident that my small script doesn't do what is intended.

.gitlab-ci.yml
```diff
@@ -190,7 +190,7 @@ android-armv7:
     - triggers
   script:
     - cargo build --target=armv7-linux-androideabi
-    # TODO: check that `arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/release/parity | grep c++_shared` is empty
+    - $(arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/debug/parity | grep c++_shared | wc -l) -eq 0
```
niklasad1: I think this should work:

```bash
if [ $(arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/debug/parity | grep c++_shared | wc -l) -ne 0]; then
   echo "FAIL!!"
fi
```
But I'm not really sure how this will be executed in docker?

General-Beck: this option is better

5chdn:
Please also _approve_ it :heavy_check_mark:

PR #8539, opened 2018-05-03
Remove inject.js server-side injection for dapps

axelchalon:
Closes https://github.com/paritytech/parity/issues/8529

The Electron shell (Parity UI) already injects `inject.js` into the dapp webview, so we don't need the server to inject `<script src="/parity-utils/inject.js"></script>` to the response when serving the dapp anymore.

cc @amaurymartiny @tomusdrw

### DIFF
--- a/dapps/src/page/handler.rs
+++ b/dapps/src/page/handler.rs
@@ -17,9 +17,8 @@
 use std::io;
 use std::time::{Duration, SystemTime};
 use hyper::{self, header, StatusCode};
-use hyper::mime::{self, Mime};
+use hyper::mime::{Mime};
 
-use apps;
 use handlers::{Reader, ContentHandler, add_security_headers};
 use {Embeddable};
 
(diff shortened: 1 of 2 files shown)


tomusdrw:
> The Electron shell (Parity UI) already injects inject.js into the dapp webview

Does it? If yes then PR looks good. I though we need it for backward compatibility though.

amaurymartiny:
@tomusdrw Yes, [here](https://github.com/parity-js/shell/blob/next/src/Dapp/dapp.js#L157). So until now inject.js was injected twice.

I'll double check that all dapps work on Parity UI.

amaurymartiny:
All dapps work correctly, dapp-develop renders correctly, dapp-wallet too. Looks good to me.

tomusdrw:
Test failure seems unrelated. Restarted the build, the code looks good.

amaurymartiny:
Once this gets merged, we should also remove the `parity ui` command, because dapps won't work anymore on :8180.

5chdn:
Test fails.

```

failures:

---- tests::home::should_inject_js stdout ----
	thread 'tests::home::should_inject_js' panicked at 'assertion failed: `(left == right)`
  left: `false`,
 right: `true`: Expected inject script tag in: 3BF
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Parity</title>
    <style>
      html {
        background: white;
        background-repeat: round;
      }

      html, body, #container {
        width: 100%;
        height: 100%;
        margin: 0;
        padding: 0;
        font-family: Sans-Serif;
      }

      .loading {
        text-align: center;
        padding-top: 5em;
        font-size: 2em;
        color: #999;
      }
    </style>
  <link rel="shortcut icon" href="parity-logo-black-no-text.png"><link href="commons.css" rel="stylesheet"><link href="bundle.css" rel="stylesheet"></head>
  <body>
    <div id="container">
      <div class="loading">Loading</div>
    </div>
  <script type="text/javascript" src="commons.js"></script><script type="text/javascript" src="bundle.js"></script></body>
</html>

0

', dapps/src/tests/home.rs:84:2
```

Issue #8428, opened 2018-04-18
Persistent time-based peer disabling


tomusdrw:
Currently we only "ban" peers in memory, when we detect they are on some different forks or different networks.

Every time the node is started though we again make a dozen of useless connections and requests, cause that info is not persisted.

I think it would make sense to have a file similar to `nodes.json` where we store peers that are temporarily banned (for some speciifc amount of time).

It would also help with dealing with warp-sync issues, currently you may end up with 25 peers that cannot provide you snapshot data that you are requesting (with `--warp-barrier`) - in such case we just idle and hope new peers will connect to use, or the old ones will get disconnected. The solution would be to temporarily ban current peers (say for 5 minutes), disconnect them and try to discover more.

CC @andresilva @ngotchac 

@rphmeier any thoughts?

adria0:
Closing issue due to its stale state.

PR #8541, opened 2018-05-03
Node table sorting according to last contact data

andresilva:
Keep track of the timestamp of the last contact we had with a node and record whether that was a failure or a success. A contact is considered a success after the session is initiated (successful exchange of hello packets). Contacts that are older than 1 week are ignored. When sorting the nodes we keep nodes with most recent successful contacts first, then we shuffle unknown nodes (either old contacts or new uncontacted nodes), and nodes with contact failures last. 

I've also fixed some cases where we weren't noting failures properly (there are still more places where we're not noting failure properly though but I wanted to keep this PR simple).

This is a minor attempt to improve #7531, I plan to extend this with #8428 in a future PR.

*I'm marking this as in progress because I'll be offline for a week and won't be able to address any comments. Feel free to commit to this branch.*

### DIFF
--- a/util/network-devp2p/src/host.rs
+++ b/util/network-devp2p/src/host.rs
@@ -601,6 +601,7 @@ impl Host {
 				},
 				Err(e) => {
 					debug!(target: "network", "{}: Can't connect to address {:?}: {:?}", id, address, e);
+					self.nodes.write().note_failure(&id);
 					return;
 				}
 			}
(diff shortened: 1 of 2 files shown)


andresilva:
I haven't tested this a lot, it should have no problem parsing `nodes.json` in the old format. A good test would be to start with a particularly problematic `nodes.json` file and see if this PR improves peer count compared to the current peer selection strategy (using the same initial `nodes.json`).

c0deright:
@andresilva If you would provide an ubuntu amd64 deb file for me with parity-1.9.7 with this patch applied I could test this over the weekend.

PR #8543, opened 2018-05-04
Refactoring `ethcore-sync` - Fixing warp-sync barrier

ngotchac:
This (pretty big) PR is mainly a refactoring of the `ethcore/sync/src/chain.rs` that divides the module into 4 components:
- The `handler`, which handles responses from other peers.
- The `propagator` which propagate messages to other peers (new blocks, ...)
- The `requester` which requests data to other peers
- The `supplier` which supplies data to other peers (responds to their requests)

It also fixes [here](https://github.com/paritytech/parity/commit/157cf3228d7be86b715edd1a1f8568b1e2d2c484) an issue with warp-sync that was wrongly resetting the ChainSync state to `Idle`.

This is (or will be) the base for other PRs that should improve warp-sync.


ethcore/sync/src/chain/requester.rs
```diff
@@ -0,0 +1,154 @@
...
+	GET_BLOCK_HEADERS_PACKET,
+	GET_RECEIPTS_PACKET,
+	GET_SNAPSHOT_DATA_PACKET,
+	GET_SNAPSHOT_MANIFEST_PACKET,
+};
+
+pub struct SyncRequester {}
+
+impl SyncRequester {
+	/// Perofrm block download request`
```
niklasad1: Perform block download `request`?

ethcore/sync/src/chain/requester.rs
```diff
@@ -0,0 +1,154 @@
...
+
+	ETH_PACKET_COUNT,
+	GET_BLOCK_BODIES_PACKET,
+	GET_BLOCK_HEADERS_PACKET,
+	GET_RECEIPTS_PACKET,
+	GET_SNAPSHOT_DATA_PACKET,
+	GET_SNAPSHOT_MANIFEST_PACKET,
+};
+
+pub struct SyncRequester {}
```
niklasad1: Docs

rphmeier: and just `pub struct SyncRequester;`, no braces

5chdn:
Adding the patch label, we might want to have this in beta if possible. Branches should not have diverged too much.

dvdplm:
I ran this branch against a clean DB and it seems to behave very well, respecting my barrier settings. 
Questions: 

a) What is the expected behavior when:

1. run with the barrier set at say 4 000 000
2. stop
3. run with the barrier set at 5 000 000 

Should I expect the sync to pick back up from 4M + whatever blocks were synched through the normal sync mode, or is that a separate issue?

b) how does the code figure out what the number of snapshot chunks to fetch is? I've seen numbers vary wildly (anything from 370 to 1338) so I'm guessing we're fetching the number of chunks that are available at the peer?

ngotchac:
@dvdplm 
So a Snapshot is divided into chunks (state chunks and block chunks), and is defined by a Manifest files that contains all the information needed. So it's from the Manifest that the number of files is determined : https://wiki.parity.io/Warp-Sync
So now, if you download a Snapshot for the state at block 4_000_000, stop, and try to get another one for the state at block 5_000_000, the chunks _will be_ totally different (because of how the snapshot format has been defined). It could be possible to find another format so that you have some redundant chunks, but that would require quite some work (cc @andresilva). You will thus need to redownload every chunks.
Now if you start downloading a snapshot, stop, and try to get the same one, it will be possible to retrieve the previously downloaded chunks with this PR : #8544

5chdn:
Test fails.

dvdplm:
@ngotchac thanks, that was useful. FYI I just managed to complete a warp sync from scratch using this branch. :)

rphmeier:
Regarding alternative snapshot formats, I've just filed https://github.com/paritytech/parity/issues/8565

ethcore/sync/src/chain/handler.rs
```diff
@@ -0,0 +1,827 @@
...
+	SNAPSHOT_MANIFEST_PACKET,
+	STATUS_PACKET,
+	TRANSACTIONS_PACKET,
+};
+
+/// The Chain Sync Handler: handles responses from peers
+pub struct SyncHandler;
+
+impl SyncHandler {
+	pub fn on_packet(sync: &mut ChainSync, io: &mut SyncIo, peer: PeerId, packet_id: u8, data: &[u8]) {
```
niklasad1: sorry for nitpicking, but this needs docs

ethcore/sync/src/chain/mod.rs
```diff
@@ -0,0 +1,1379 @@
...
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+/// `BlockChain` synchronization strategy.
```
niklasad1: Optional to fix: Module documentation should use `//!` instead of `///`

5chdn:
Can't merge.

Issue #7862, opened 2018-02-12
Share Snapshots between clients.


5chdn:
In addition to blockchain synchronization, the Parity client should also synchronize snapshots. Clients that are still able to generate snapshots should tell other nodes about their snapshot (height, size, etc) and encourage them to store a copy of the snapshot along with the blockchain. This will help with snapshot distribution in the network.

5chdn:
Partially addressed in #8544

Issue #6415, opened 2017-08-29
Snapshot sync is started from scratch after restart


kondratovich:
I'm starting new parity node with empty state. Snapshot sync in progress. When I restart node, all synced snapshots are ignored and counter starts from zero.

Does snapshot sync has some disk state?

5chdn:
Slightly related to #6350.

I'm not entirely sure how the snapshot sync works in this case. Did you restart the node immediately? Or was there some time passing between stopping and starting the node? It's quite possible that the client started fetching a _new_ snapshot and therefore started from scratch.

BreezeRo:
This just occurred to me running 1.7.0 on docker

5chdn:
@BreezeRo  Did you restart the node immediately? Or was there some time passing between stopping and starting the node?

kondratovich:
@5chdn In my case - few seconds, maybe minute

5chdn:
Fixed in #8544

PR #8544, opened 2018-05-04
Resumable warp-sync / Seed downloaded snapshots

ngotchac:
**This should be merge after #8543**

This PR makes warp-sync resumable, by keeping the downloaded chunks between restarts.
On restart, when a new snapshot restoration starts, it looks for the already downloaded chunks, and process them one to one. Then new chunks are asked to other peers.

It also enables nodes to seed the snapshot they just downloaded before having all the Old Blocks: before, it would first not send the SnapshotManifest if some old blocks would be downloading ; peers would also ask for old blocks to any other peers, which some peers couldn't respond to since they are still downloading them.

It also fixes a corner when your node would start downloading a Manifest, but the remote node would disconnect before the transfer is finished. Your node would be stuck on `SnapshotManifest` sync state and never recover from it.

Could @rphmeier and/or @tomusdrw confirm that theses ^^^ changes are OK?


ethcore/sync/src/chain/supplier.rs
```diff
@@ -237,20 +237,20 @@ impl SyncSupplier {
 	/// Respond to GetSnapshotManifest request
 	fn return_snapshot_manifest(io: &SyncIo, r: &Rlp, peer_id: PeerId) -> RlpResponseResult {
 		let count = r.item_count().unwrap_or(0);
-		trace!(target: "sync", "{} -> GetSnapshotManifest", peer_id);
+		trace!(target: "warp-sync", "{} -> GetSnapshotManifest", peer_id);
```
tomusdrw: Maybe `warp`? Would be shorter.

ethcore/sync/src/chain/mod.rs
```diff
@@ -648,8 +656,28 @@ impl ChainSync {
 		}
 	}
 
+	/// Check if the snapshot service is ready
+	fn is_snapshot_ready(&mut self, io: &SyncIo) -> bool {
```
tomusdrw: IMHO not a good name for anything else besides a getter for a boolean value

ethcore/sync/src/chain/mod.rs
```diff
@@ -648,8 +656,28 @@ impl ChainSync {
...
+		if io.snapshot_service().ready() {
+			trace!(target: "snapshot", "Snapshot Service is ready!");
+			// Sync the previously resumed chunks
+			self.snapshot.sync(io);
```
tomusdrw: Calling that method multiple times will overwrite `completed_chunks`, I don't think it's expected.

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
...
 				_ => return Err(e.into()),
 			}
 		}
 
-		fs::create_dir_all(&rest_dir)?;
+		// Move the previous recovery temp directory
+		// to `prev_chunks` to be able to restart restoring
+		// with previously downloaded blocks
+		// This step is optional, so don't fail on error
+		fs::rename(&recovery_temp, &prev_chunks).ok();
```
tomusdrw: IMHO this should be moved under `restoration.lock()`, previously it was ok to remove and create the same directory multiple times, now we are doing a rename which should probably be done only once.

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
 	/// Initialize the restoration synchronously.
 	/// The recover flag indicates whether to recover the restored snapshot.
 	pub fn init_restore(&self, manifest: ManifestData, recover: bool) -> Result<(), Error> {
-		let rest_dir = self.restoration_dir();
-
-		let mut res = self.restoration.lock();
+		self.ready.store(false, Ordering::SeqCst);
```
tomusdrw: Shouldn't that be moved under `restoration.lock()`?

ethcore/src/snapshot/service.rs
```diff
@@ -492,6 +580,7 @@ impl Service {
 
 		let _ = fs::remove_dir_all(self.restoration_dir());
 		*self.status.lock() = RestorationStatus::Inactive;
+		self.ready.store(false, Ordering::SeqCst);
```
tomusdrw: Is that fine? Why we are becoming not ready after the snapshot is restored? Maybe the name is not good enough and should encapsulate more regarding "readiness"

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
...
+		file.read_to_end(&mut buffer)?;
+
+		let hash = keccak(&buffer);
+
+		let (is_valid, is_state) = if manifest.block_hashes.contains(&hash) {
+			(true, false)
+		} else if manifest.state_hashes.contains(&hash) {
+			(true, true)
+		} else {
+			(false, false)
```
tomusdrw: Perhaps early exit here instead of having this `is_valid` flag?

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
...
+	fn import_prev_chunks(&self, manifest: ManifestData) -> Result<(), Error> {
+		let prev_chunks = self.prev_chunks_dir();
+
+		// Restore previous snapshot chunks
+		let files = fs::read_dir(prev_chunks.as_path())?;
+		let mut num_temp_chunks = 0;
+
+		for prev_chunk_file in files {
+			// Import the chunk, don't fail and continue if one fails
+			self.import_prev_chunk(&manifest, prev_chunk_file).ok();
```
tomusdrw: Maybe just log a warning here and not increment `num_temp_chunks`?

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
...
+
+		// Import previous chunks, continue if it fails
+		self.import_prev_chunks(manifest).ok();
+		self.ready.store(true, Ordering::SeqCst);
+
+		Ok(())
+	}
+
+	/// Import the previous chunks into the current restoration
+	fn import_prev_chunks(&self, manifest: ManifestData) -> Result<(), Error> {
```
tomusdrw: Isn't there a better way to figure out if the chunks in directory are really mean for this manifest?

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,133 @@ impl Service {
...
+				block_chunks: block_chunks as u32,
+				state_chunks_done: self.state_chunks.load(Ordering::SeqCst) as u32,
+				block_chunks_done: self.block_chunks.load(Ordering::SeqCst) as u32,
+			};
+		}
 
 		self.restoring_snapshot.store(true, Ordering::SeqCst);
+
+		// Import previous chunks, continue if it fails
+		self.import_prev_chunks(manifest).ok();
```
tomusdrw: What will happen if we start to sync a different manifest?

ethcore/src/snapshot/service.rs
ngotchac: We could check for the name first, but hashing the content of the chunks is really fast so I don't think it's a bottleneck here (and it shouldn't happen that much that you stop your warp sync to restart syncing another snapshot)

ethcore/src/snapshot/service.rs
ngotchac: It just won't import any of the available chunks and the directory will be deleted anyway.

ethcore/sync/src/chain/mod.rs
ngotchac: Yep but always with the right value. Is that an issue?

ethcore/src/snapshot/service.rs
ngotchac: Yeah so `ready` here kind of means `initialized`. The issue being that it should be set to `false` before `init_restore` is called, since it's called by message passing and otherwise there would be race condition with the Chain Sync reading the `ready` value before the message is passed and `init_restore` is called.

ethcore/src/snapshot/service.rs
tomusdrw: My question is more like: "Isn't that an issues?" How do we know if it's the right manifest to sync with?

ethcore/src/snapshot/service.rs
tomusdrw: I see. Is it possible to encapsulate that flow in the names somehow? Currently it's a little bit misleading, maybe call it `is_ready_for_restoration`?

ethcore/sync/src/chain/mod.rs
tomusdrw: Are we sure it's the right value? `completed_chunks` are modified in `snapshot`, no? Are we super sure this two are in sync?

If it's required only initially, maybe change the type of `completed_chunks` and only restore if it's `None`?

5chdn:
Does this address

- #6415 as in "resumable warp-sync"?
- #7862 as in "seed downloaded snapshots"?

ethcore/src/snapshot/service.rs
ngotchac: Mmmh, the `manifest` is the one that just got downloaded, so it must be the right one. I'm not sure I get your question.

ethcore/src/snapshot/service.rs
```diff
@@ -407,54 +417,129 @@ impl Service {
...
+		if let Err(e) = fs::remove_dir_all(&prev_chunks) {
 			match e.kind() {
 				ErrorKind::NotFound => {},
 				_ => return Err(e.into()),
 			}
 		}
 
-		fs::create_dir_all(&rest_dir)?;
+		{
+			self.ready.store(false, Ordering::SeqCst);
```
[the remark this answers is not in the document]
tomusdrw: Need to be careful with resetting the state to `false` in case of any error. IMHO better to make sure that it's only done after the lock is acquired and we know that restoration is still needed (i.e. `res != None`?)

ngotchac:
@5chdn it does address #6415 ; and partially #7862 : partially because it just seeds the snapshot you downloaded, but I think what we want as well is to download snapshots even if you are synced, so that you can propagate them, right?

5chdn:
> but I think what we want as well is to download snapshots even if you are synced, so that you can propagate them, right?

Yes, if this does not cause too much overhead (bandwidth, disk space), we should do this.

ngotchac:
It could be an opt-out option.

ethcore/src/snapshot/service.rs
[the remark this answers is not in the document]
ngotchac: I updated this bit so that the restoration stays locked until the initialization is done.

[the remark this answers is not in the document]
ngotchac: https://github.com/paritytech/parity/pull/8544/commits/0f1e5b414f74bad2b15d992c752d465b4832a21d

rpc/src/v1/tests/helpers/snapshot_service.rs
```diff
@@ -43,6 +43,8 @@ impl TestSnapshotService {
 impl SnapshotService for TestSnapshotService {
 	fn manifest(&self) -> Option<ManifestData> { None }
 	fn supported_versions(&self) -> Option<(u64, u64)> { None }
+	fn restoration_ready(&self) -> bool { true }
+	fn completed_chunks(&self) -> Option<Vec<H256>> { Some(vec![]) }
```
ordian: > IMO the Snapshot restoration state in the Chain Sync shouldn't exist anyway, and it should be read from the Snapshot Service when needed.

Maybe we could embed `Snapshot` in `Service` and add a bunch of methods, like `needed_chunk` to `SnapshotService` instead of `fn completed_chunks`? That way we wouldn't need to sync `Snapshot` manually.

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,136 @@ impl Service {
...
+					ErrorKind::NotFound => {},
+					_ => return Err(e.into()),
+				}
 			}
+
+			fs::create_dir_all(&rest_dir)?;
+
+			// make new restoration.
+			let writer = match recover {
+				true => Some(LooseWriter::new(self.temp_recovery_dir())?),
```
ordian: nit: `recovery_temp`?

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,136 @@ impl Service {
...
+				false => None
+			};
+
+			let params = RestorationParams {
+				manifest: manifest.clone(),
+				pruning: self.pruning,
+				db: self.restoration_db_handler.open(&self.restoration_db())?,
```
ordian: nit: `&rest_db`?

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,136 @@ impl Service {
...
+			// Import the chunk, don't fail and continue if one fails
+			match self.import_prev_chunk(restoration, &manifest, prev_chunk_file) {
+				Ok(_) => num_temp_chunks += 1,
+				Err(e) => trace!(target: "snapshot", "Error importing chunk: {:?}", e),
+			}
+		}
 
-		let state_chunks = params.manifest.state_hashes.len();
-		let block_chunks = params.manifest.block_hashes.len();
+		trace!(target:"snapshot", "Imported {} previous chunks", num_temp_chunks);
```
ordian: Judging by the way it is computed, I think, `num_temp_chunks` is a number of all chunks in temp dir. But since it's used only for debugging, it's not that important.

ethcore/src/snapshot/service.rs
```diff
@@ -406,55 +416,136 @@ impl Service {
 	/// Initialize the restoration synchronously.
 	/// The recover flag indicates whether to recover the restored snapshot.
 	pub fn init_restore(&self, manifest: ManifestData, recover: bool) -> Result<(), Error> {
-		let rest_dir = self.restoration_dir();
+		{
```
ordian: Why do we need this extra scope?

ethcore/sync/src/snapshot.rs
```diff
@@ -47,6 +51,19 @@ impl Snapshot {
 		}
 	}
 
+	/// Sync the Snapshot completed chunks with the Snapshot Service
+	pub fn sync (&mut self, io: &SyncIo) {
```
ordian: nit: space before `(`

ngotchac:
@ordian I had some points you raised fixed in another branch I just merged.
The main change is that now the Snapshot service has a new status `RestorationStatus::Initializing`. The Sync service checks for the status before asking peers for chunks, and the when it goes from `Initializing` to `Ongoing` it fetches the chunks done from the service once, and then go on as previously.

ethcore/src/snapshot/service.rs
```diff
@@ -424,37 +449,102 @@ impl Service {
...
+		file.read_to_end(&mut buffer)?;
+
+		let hash = keccak(&buffer);
+
+		let is_state = if manifest.block_hashes.contains(&hash) {
+			false
+		} else if manifest.state_hashes.contains(&hash) {
+			true
+		} else {
+			return Ok(());
```
ordian: nit: I believe something like `Result<bool,_>` here is needed to distinguish imported blocks from others.

ethcore/src/snapshot/service.rs
```diff
@@ -424,37 +449,102 @@ impl Service {
 			}
 		}
 
+		{
+			*self.status.lock() = RestorationStatus::Initializing {
```
ordian: nit: I don't think scope here is necessary since `lock()` is not left hand side of the assignment, so guard here is temporary ([play example](http://play.rust-lang.org/?gist=5885b84ebcd4740aaacc5616dcc19146&version=stable&mode=debug))

ethcore/sync/src/chain/mod.rs
```diff
@@ -755,26 +758,48 @@ impl ChainSync {
...
 				},
 				SyncState::SnapshotData => {
-					if let RestorationStatus::Ongoing { state_chunks_done, block_chunks_done, .. } = io.snapshot_service().status() {
-						if self.snapshot.done_chunks() - (state_chunks_done + block_chunks_done) as usize > MAX_SNAPSHOT_CHUNKS_DOWNLOAD_AHEAD {
-							trace!(target: "sync", "Snapshot queue full, pausing sync");
-							self.state = SyncState::SnapshotWaiting;
+					match io.snapshot_service().status() {
+						RestorationStatus::Ongoing { state_chunks_done, block_chunks_done, .. } => {
+							// Initialize the snapshot if not already done
+							if !self.snapshot.is_initialized() {
```
ordian: nit: the if is not necessary, since `initialize` does the checking

ethcore/sync/src/snapshot.rs
```diff
@@ -44,7 +49,27 @@ impl Snapshot {
 			completed_chunks: HashSet::new(),
 			snapshot_hash: None,
 			bad_hashes: HashSet::new(),
+			initialized: false,
+		}
+	}
+
+	/// Sync the Snapshot completed chunks with the Snapshot Service
+	pub fn initialize(&mut self, io: &SyncIo) {
```
ordian: nit: initialize only needs iterator over chunks, not the whole `SyncIo`

ethcore/sync/src/chain/handler.rs
```diff
@@ -101,14 +101,27 @@ impl SyncHandler {
...
+		trace!(target: "sync", "== Disconnecting {}: {}", peer_id, io.peer_info(peer_id));
+		sync.handshaking_peers.remove(&peer_id);
+		if sync.peers.contains_key(&peer_id) {
+			debug!(target: "sync", "Disconnected {}", peer_id);
+			sync.clear_peer_download(peer_id);
+			sync.peers.remove(&peer_id);
+			sync.active_peers.remove(&peer_id);
+
+			if sync.state == SyncState::SnapshotManifest {
+				// Check if other we are asking other peers for
```
ordian: typo: check if __ we are

ethcore/sync/src/chain/supplier.rs
```diff
@@ -135,7 +136,7 @@ impl SyncSupplier {
 		let inc = (skip + 1) as BlockNumber;
 		let overlay = io.chain_overlay().read();
 
-		while number <= last && count < max_count {
+		while (number <= last || overlay.contains_key(&number)) && count < max_count {
```
ordian: Is this change unrelated to the PR? Maybe a comment would help.

ngotchac: Yep kind of unrelated : it actually fixes a bug where you wouldn't send the _ForkBlock_ header before being synced even if we cached it for this exact reason : we want to share it with other node ASAP so that they can confirm we are on the right fork.

ethcore/sync/src/snapshot.rs
ordian: ok, a compromise variant would be to require on `&SnapshotService`, but I have to stop grumbling now :)

ethcore/src/snapshot/service.rs
rphmeier: (this is done now)

ethcore/sync/src/snapshot.rs
```diff
@@ -89,18 +115,22 @@ impl Snapshot {
...
+		let needed_chunks: Vec<H256> = self.pending_state_chunks.iter()
 			.chain(self.pending_block_chunks.iter())
-			.find(|&h| !self.downloading_chunks.contains(h) && !self.completed_chunks.contains(h))
-			.cloned();
+			.filter(|&h| !self.downloading_chunks.contains(h) && !self.completed_chunks.contains(h))
+			.map(|h| *h)
+			.collect();
+
+		// Get a random chunk
+		let chunk = thread_rng().choose(&needed_chunks);
```
rphmeier: why choose randomly?

ethcore/src/snapshot/service.rs
```diff
@@ -424,40 +449,104 @@ impl Service {
...
+	fn import_prev_chunks(&self, restoration: &mut Option<Restoration>, manifest: ManifestData) -> Result<(), Error> {
+		let prev_chunks = self.prev_chunks_dir();
+
+		// Restore previous snapshot chunks
+		let files = fs::read_dir(prev_chunks.as_path())?;
+		let mut num_temp_chunks = 0;
+
+		for prev_chunk_file in files {
+			// Don't go over all the files if the restoration has been aborted
+			if !self.restoring_snapshot.load(Ordering::SeqCst) {
```
rphmeier: I don't think restoration can be aborted since this function is called holding the restoration lock

ngotchac: Actually it first sets `restoring_snapshot` to `false` when aborting, then locks `restoration` (lines ~730)

ngotchac:
@rphmeier PR grumbles addressed :)

5chdn:
:tada:

PR #8545, opened 2018-05-04
Handle socket address parsing errors

### LINKED ISSUE #6972: `Node::from_str()` reports errors incorrectly
Impl `FromStr` for `Node` and `NodeEndpoint` return `NetworkError::AddressResolve` even on incorrectly formatted addresses, like `foo`, `enode://foo@bar` and even on an empty string.

Issue exists on 713bba00 and earlier.

dvdplm:
Unpack errors and check for io::ErrorKind::InvalidInput and return our own AddressParse error. Remove the foreign link to std::net::AddrParseError and add an `impl From` for that error. Test parsing properly.

Fix #6972


util/network-devp2p/src/node_table.rs
```diff
@@ -27,6 +27,7 @@ use network::{Error, ErrorKind, AllowIP, IpFilter};
 use discovery::{TableUpdates, NodeEntry};
 use ip_utils::*;
 use serde_json;
+use std::io;
```
debris: please place imports in order, ideally in line 23, together with other `std` imports

tip: if you are using vim, you can easily sort everything by typing `:sort` when in visual mode :)

util/network-devp2p/src/node_table.rs
```diff
@@ -442,11 +456,56 @@ mod tests {
...
+	#[test]
+	fn endpoint_parse_empty_ip_string_returns_error() {
+		let endpoint = NodeEndpoint::from_str("");
+		assert!(endpoint.is_err());
+		assert_matches!(
+			endpoint.unwrap_err().kind(),
+			// TODO: after 1.27 is stable, remove the `&`s and `ref`s, see https://github.com/rust-lang/rust/pull/49394
+			&ErrorKind::AddressParse(ref io_err) => {
+				if let &Some(ref e) = io_err {
+					assert_eq!(e.to_string(), "invalid socket address");
```
debris: what if `io_err` is `None`? This assertion will not catch that

util/network-devp2p/src/node_table.rs
```diff
@@ -122,18 +123,31 @@ impl FromStr for NodeEndpoint {
 				address: a,
 				udp_port: a.port()
 			}),
-			Ok(_) => Err(ErrorKind::AddressResolve(None).into()),
-			Err(e) => Err(ErrorKind::AddressResolve(Some(e)).into())
+			Ok(None) => {
+				// REVIEW: how is this case possible? I think we can remove this case.
+				Err(ErrorKind::AddressResolve(None).into())
```
debris: I think we shouldn't remote it, cause we cannot prove that std api, will not change underlaying implementation.

anyway, I think it's also more idiomatic to use `bail!` macro in cases like this one, e.g.

```rust
Ok(None) => bail!(ErrorKind::AddressResolve(None)),
```

util/network-devp2p/src/node_table.rs
```diff
@@ -122,18 +123,31 @@ impl FromStr for NodeEndpoint {
...
+			},
+			Err(e) => {
+				match e.kind() {
+					io::ErrorKind::InvalidInput => {
+						Err(ErrorKind::AddressParse(Some(e)).into())
+					},
+					_ => {
+						Err(e.into())
```
debris: this will be transformed into `Error::Io(...)`. Is it desired?

util/network/src/error.rs
```diff
@@ -84,11 +84,16 @@ error_chain! {
 	foreign_links {
 		SocketIo(IoError) #[doc = "Socket IO error."];
 		Io(io::Error) #[doc = "Error concerning the Rust standard library's IO subsystem."];
-		AddressParse(net::AddrParseError) #[doc = "Error concerning the network address parsing subsystem."];
 		Decompression(snappy::InvalidInput) #[doc = "Decompression error."];
 	}
 
 	errors {
+		#[doc = "Error concerning the network address parsing subsystem."]
+		AddressParse(err: Option<io::Error>) {
```
debris: I'm not sure if it's worth having `Option<io::Error>` here. Maybe it should be just `AddressParse()` ? `io:Error` is always the same https://doc.rust-lang.org/src/std/net/addr.rs.html#891

util/network-devp2p/src/node_table.rs
dvdplm: Can you elaborate on this a little? Not sure I understand what you mean by "shouldn't remote it". Thanks!

dvdplm: Oh, I think you meant to type "shouldn't REMOVE" it. Now I get it!

util/network-devp2p/src/node_table.rs
dvdplm: How about we return `Err(ErrorKind::AddressParse())` for all error cases here? In light of your other comment below and https://doc.rust-lang.org/src/std/net/addr.rs.html#891 it seems kind of silly to cover other cases (there aren't any, `to_socket_addrs` always only return `InvalidInput`). This match arm would become:

```rust
Err(e) => Err(ErrorKind::AddressParse())
```

Thoughts?

5chdn:
Please be more descriptive with PR titles :)

dvdplm:
> Please be more descriptive with PR titles :)

You are absolutely right, apologies.

util/network-devp2p/src/node_table.rs
```diff
@@ -463,6 +487,17 @@ mod tests {
...
 
+	#[test]
+	fn node_parse_fails_for_invalid_urls() {
+		let node = Node::from_str("foo");
+		assert!(node.is_err());
+		assert_matches!(node.unwrap_err().kind(), &ErrorKind::AddressParse);
+
+		let node = Node::from_str("enode://foo@bar");
+		assert!(node.is_err());
+		assert_matches!(node.unwrap_err().kind(), &ErrorKind::AddressParse);
```
debris: I believe `assert_matches` is no longer required :)

5chdn:
Can't merge :)

util/network-devp2p/src/node_table.rs
dvdplm: meh, sadly it seems like that's still up in the air: https://github.com/rust-lang-nursery/error-chain/issues/95

assert_matches is used elsewhere, so at least it's not a new-new dependency. :/

PR #8546, opened 2018-05-04
Add transaction hash to the RPC result of replay_block

allquantor:
Suggestion how the issue https://github.com/paritytech/parity/issues/8499 can be solved. 
The motivation is elaborated in the issue description. 

There are no tests, happy to add some if requiered but it looks like a small change to me that can go inside without testing.


allquantor:
[clabot:check]

rpc/src/v1/types/trace.rs
```diff
@@ -616,6 +617,7 @@ pub struct TraceResults {
 impl From<Executed> for TraceResults {
 	fn from(t: Executed) -> Self {
```
tomusdrw: IMHO we should avoid changing core types `Executed` and instead do it inside the RPC layer.
Change `From<Executed>` implementation into:
```
impl TraceResults {
 fn new(t: Executed, hash: H256) -> Self {
  ...
 }
}
```
And fix all occurences.

allquantor: Sure, it could be done this way. Do you have an idea how to get the transaction hash in following places:

https://github.com/paritytech/parity/blob/a04c5b180a821c709b0ea24c7d4f21c9c2511839/rpc/src/v1/impls/traces.rs#L176

https://github.com/paritytech/parity/blob/a04c5b180a821c709b0ea24c7d4f21c9c2511839/rpc/src/v1/impls/traces.rs#L134

It involves probably more complexity than the benefit to change this code?

allquantor: Avoiding DB migrations should be preferred imho. As well as I have a good feeling that txhash was missed in executed since any execution belongs to unique txhash.
I could add some tests?

5chdn:
Tests failing.

5chdn:
Hi @allquantor are you still working on this?

allquantor:
@5chdn 
Completely forgot. Will try to push a fix today.

5chdn:
Let me know when you are ready :)

foxfromabyss:
Extended the test support, see #8946

rpc/src/v1/types/trace.rs
sorpaas: Second @tomusdrw. Otherwise we have cases where fake-sign transactions (like if replay a fake call) can have invalid transaction hash values. We don't need any db migrations and the change should also be simple enough.

PR #8555, opened 2018-05-07
Fix packet count when talking with PAR2 peers

tomusdrw:
Fixes couple of issues:
- The issue is that light status packet is incorrectly mapped to private transaction packet if we speak with PAR2 peers. Due to this light protocol decides to disconnect peers because they never send status after handshake.
- Increases couple of timeouts for light protocol.
- Changes the eclipse protection to check if the peer is already known to us, so that if a remote peer is connecting and we know of it, we won't reject such connection (since we would attempt to connect to them anyway)


5chdn:
Could not compile `ethcore-network-devp2p`.

grbIzl:
Looks good. Only didn't understand the reasoning for timeouts. How do we choose values for them?

rphmeier:
I pulled the values from a few empirical observations about network latency and time to fetch from DB initially. In the future we can have them be adaptive based on observed response times.

5chdn:
Could not compile `whisper-cli`.

5chdn:
Could not compile `ethcore-sync`. :rofl:

util/network-devp2p/src/host.rs
```diff
@@ -722,11 +725,24 @@ impl Host {
...
-								(s.info.originated && egress_count > min_peers) ||
-								(!s.info.originated && ingress_count > max_ingress) {
-								// only proceed if the connecting peer is reserved.
-								if !self.reserved_nodes.read().contains(&id) {
+							if reserved_only
+								|| (s.info.originated && egress_count > min_peers)
+								|| (!s.info.originated && ingress_count > max_ingress)
+							{
+								// We didn't start the connection, but the node is known to us
+								// So eventually we will attempt to connect to it as well.
```
sorpaas: Have we checked that this is not subject to a variant on the table poisoning attack?

Let's say if there are N nodes controlled by the attacker in a client's node table. For outgoing connections, not all of those N nodes will be tried immediately. However, with this change, in incoming connections, the attacker can flood the client using all those N nodes, and controls a much larger portion of the ongoing connections.

util/network-devp2p/src/host.rs
```diff
@@ -722,11 +725,24 @@ impl Host {
...
+								let is_incoming_but_known = !s.info.originated && self.nodes.read().contains(&id);
+
+								if is_incoming_but_known {
+									warn!(target: "network", "Allowing incoming connection from a known node.");
```
dvdplm: s/warn!/info!/ perhaps?

ethcore/light/src/net/mod.rs
```diff
@@ -111,9 +114,9 @@ mod packet {
 mod timeout {
 	use std::time::Duration;
 
-	pub const HANDSHAKE: Duration =  Duration::from_millis(2500);
-	pub const ACKNOWLEDGE_UPDATE: Duration = Duration::from_millis(5000);
-	pub const BASE: u64 = 1500; // base timeout for packet.
+	pub const HANDSHAKE: Duration =  Duration::from_millis(4_000);
+	pub const ACKNOWLEDGE_UPDATE: Duration = Duration::from_millis(5_000);
+	pub const BASE: u64 = 2500; // base timeout for packet.
```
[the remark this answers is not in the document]
dvdplm: While we're in here, perhaps s/BASE/BASE_TIMEOUT/?

ethcore/sync/src/chain/requester.rs
```diff
@@ -140,7 +140,8 @@ impl SyncRequester {
 			}
 			peer.asking = asking;
 			peer.ask_time = Instant::now();
-			let result = if packet_id >= ETH_PACKET_COUNT {
+			// TODO [ToDr] This seems quite fragile. Be careful when protocol is updated.
+			let result = if packet_id >= ETH_PROTOCOL_VERSION_63.1 {
```
dvdplm: Perhaps a test here to help our future selves?

ethcore/light/src/net/mod.rs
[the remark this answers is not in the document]
tomusdrw: @dvdplm we use it always including module name, so it's `timeout::BASE` and I think it doesn't make sense to repeat it in the name.

ethcore/sync/src/chain/requester.rs
tomusdrw: Logged a separate issue.

tomusdrw:
I've reverted changes to eclipse protection since indeed the change needs more discussion and review and is not critical to this PR.

ethcore/light/src/net/mod.rs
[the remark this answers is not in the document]
dvdplm: fair enough

Issue #8190, opened 2018-03-22
Release next-beta 1.11


5chdn:
**Prepare Code:**
- [x] Decide on a rough timeline, feature freeze, release date
  - [x] Feature freeze: Apr 20, 2018
  - [x] Estimated release: **Apr 30, 2018**
- [x] Do a reality-check against the [roadmap-known-issues](https://paritytech.github.io/wiki/Known-Issues-Priorities)
- [x] **Make sure UI related changes are done**
  - [ ] Remove UI and Public Node completely
  - [x] Replace :8180 with static page (link to docs, simple stats)
  - [ ] Remove (rework?) installers (code, ci, ...) and promote separate UI
- [x] Notify devs about a new beta release, ask to submit PRs
- [x] Create backports for all PRs tagged with **B0-patch**
- [x] Merge all pending pull requests that should be included in the next beta

**Freeze Code:**
- [x] archive `stable` branch to something like `old-stable-1.9`
  - [x] change THIS_TRACK in code to `oldstable`
  - [x] protect that branch in Github settings
- [x] merge `beta` branch to `stable`
  - [x] and bump track (THIS_TRACK `stable`)
  - [x] check ci scripts for `stable` or `*1.9*` specifics -> `*1.10*`
  - [x] check the js/ui scripts from track specific stuff
- [x] merge `master` branch to `beta`
  - [x] and bump track (THIS_TRACK `beta`)
  - [x] check ci scripts for `beta` or `*1.10*` specifics -> `*1.11*`
  - [x] check the js/ui scripts from track specific stuff
- [x] On master: bump minor and patch version in code to 1.12.0
  - [x] Cargo.toml (`version` key)
  - [x] util/version/Cargo.toml (`version` key)
  - [x] nsis/installer.nsi (`VERSIONMINOR` and `VERSIONBUILD` keys)
  - [x] mac/Parity.pkgproj (`VERSION` key)
  - [x] mac/Parity/Info.plist (`CFBundleShortVersionString`)
  - [x] `cargo update -p parity-version`
- [x] Run `cargo update -p parity-version` on the new branches
- [x] On stable: bump and consider a stable release

**Conduct QA:**
- [x] Check if Gitlab pipeline on beta and latest HEAD is green
- [x] Create a test tag for testing the full release pipeline, e.g., `v1.11.0-rc0`
- [x] Fix all issues with the test tag until CI is all green
[...]

PR #8556, opened 2018-05-07
Update CHANGELOG for 1.9, 1.10, and 1.11

5chdn:
ref #8190


5chdn:
Waiting for docs for whisper-cli from @niklasad1 then this is good to go.

PR #8520, opened 2018-04-30
Enable WebAssembly and Byzantium for Ellaism

ellaismer:
Schedule a hard fork for Ellaism network at block 2,000,000, enabling:

* WebAssembly: pwasm as in Kovan. https://github.com/ellaism/specs/blob/master/specs/2018-0003-wasm-hardfork.md
* Byzantium: EIP100, EIP140, EIP196, EIP197, EIP198, EIP211, EIP214, EIP658. https://github.com/ellaism/specs/blob/master/specs/2018-0004-byzantium.md

PR #8557, opened 2018-05-07
[stable] Backports

tomusdrw:
- [ ] #8522 `Don't panic in import_block if invalid rlp`
  Requires quite big refactoring PRs to be backported. It's not critical since we checked that currently all paths are covered with additional checks. We can create a similar fix though for stable if required.
- [x] #8491 `Pass on storage keys tracing to handle the case when it is not modifie`
- [x] #8463 `Fetching logs by hash in blockchain database`
- [ ] #8486 `Remove depth == 0 check for tracing precompiled contracts`
  Not merged yet, and not critical
- [x] #8493 `Update wasmi and pwasm-utils`
- [x] #8520 `Enable WebAssembly and Byzantium for Ellaism`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1678,6 +1678,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "nan-preserving-float"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "net2"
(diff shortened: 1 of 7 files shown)


5chdn:
No need, we can release without them.

5chdn:
Forgot, and added #8520

Issue #8553, opened 2018-05-07
Decoding headers can fail


dvdplm:
In ethcore, encoder.rs does not handle decoding errors. Make `Header::decode`, `Block::decode` return `Result` and fix the resulting problems in consuming code.

- [x] `Header::decode` returns `Result` (PR #8527)
- [x] `Block::decode` returns `Result`

PR #8570, opened 2018-05-08
Decoding headers can fail

dvdplm:
Fix for issue #8553


ethcore/src/encoded.rs
```diff
@@ -47,7 +48,9 @@ impl Header {
 	pub fn new(encoded: Vec<u8>) -> Self { Header(encoded) }
 
 	/// Upgrade this encoded view to a fully owned `Header` object.
-	pub fn decode(&self) -> FullHeader { ::rlp::decode(&self.0).expect("decoding failure") }
+	pub fn decode(&self) -> Result<FullHeader, Error> {
+		rlp::decode(&self.0).map_err(|e|e.into() ) // REVIEW: I can't make the automatic conversion from ::rlp::DecoderError to error::Error work
```
debris: I believe it's better to return `rlp::Error`

ethcore/sync/src/light_sync/response.rs
```diff
@@ -45,6 +43,18 @@ impl From<DecoderError> for BasicError {
 	}
 }
 
+impl From<ethcore::error::Error> for BasicError {
+	fn from(err: ethcore::error::Error) -> BasicError {
+		match err {
+			ethcore::error::Error(ethcore::error::ErrorKind::Decoder(dec_err), _) => {
+				BasicError::Decoder(dec_err)
+			},
+			_ => panic!("unhandled ethcore error in light_sync")
```
debris: and this will be redundant

ethcore/src/encoded.rs
debris: using the same `Error` type for all ethcore crate is one of the biggest anti patterns that we have in our codebase. Unfortunately when this code was written, we didn't know that

rpc/src/v1/impls/light/parity.rs
```diff
@@ -395,7 +395,7 @@ impl Parity for ParityClient {
 
 		let engine = self.light_dispatch.client.engine().clone();
 		let from_encoded = move |encoded: encoded::Header| {
-			let header = encoded.decode();
+			let header = encoded.decode().expect("decoding error"); // REVIEW: not sure what to do here; what is a decent return value for the error case here?
```
debris: I believe it should be `BoxFuture<Result<RichHeader>>`. 

if `self.fetcher` is fetching headers from the network, this may be a critical vulnerability in light clients

ethcore/src/error.rs
```diff
@@ -310,11 +316,12 @@ impl From<AccountsError> for Error {
...
 		ErrorKind::AccountProvider(err).into()
 	} 
-} 
+}
 
+// REVIEW: I changed this from `UtilError::from(err).into()` to this because I could not get it to work and I couldn't find anything using it. Ok?
 impl From<::rlp::DecoderError> for Error {
 	fn from(err: ::rlp::DecoderError) -> Error {
-		UtilError::from(err).into()
+		ErrorKind::Decoder(err).into()
```
debris: 👍 imo `UtilError` as it is, is obsolete

ethcore/src/encoded.rs
dvdplm: Do you mean changing the signature from 
```rust
pub fn decode(&self) -> Result<FullHeader, Error>
```
…to:
```rust
pub fn decode(&self) -> Result<FullHeader, rlp::DecoderError>
```
?

Doing that works and surprisingly (to me) did not break a lot of code. There was one instance of having to `map_err` a `rlp::DecoderError` into an `ethcore::error::Error` to make `and_then` work, and that's a little awkward.

5chdn:
Oops :rofl: :fire:

PR #8580, opened 2018-05-09
Remove unnecessary cloning in overwrite_with

sorpaas:
Cloning can be fixed by using into_iter.

### DIFF
--- a/ethcore/src/state/account.rs
+++ b/ethcore/src/state/account.rs
@@ -459,7 +459,7 @@ impl Account {
 		self.code_size = other.code_size;
 		self.address_hash = other.address_hash;
 		let mut cache = self.storage_cache.borrow_mut();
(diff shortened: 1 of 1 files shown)


dvdplm:
lgtm

ethcore/src/state/account.rs
```diff
@@ -459,8 +459,8 @@ impl Account {
 		self.code_size = other.code_size;
 		self.address_hash = other.address_hash;
 		let mut cache = self.storage_cache.borrow_mut();
-		for (k, v) in other.storage_cache.into_inner() {
-			cache.insert(k.clone() , v.clone()); //TODO: cloning should not be required here
+		for (k, v) in other.storage_cache.into_inner().into_iter() {
```
rphmeier: isn't a call to `into_iter()` redundant then?

PR #8581, opened 2018-05-09
block_header can fail so return Result

dvdplm:
Part 2 of https://github.com/paritytech/parity/pull/8570 where we sort out the return type of the block_header method of the rpc Parity trait

### DIFF
--- a/rpc/src/v1/impls/light/parity.rs
+++ b/rpc/src/v1/impls/light/parity.rs
@@ -390,7 +390,7 @@ impl Parity for ParityClient {
 		})
 	}
 
-	fn block_header(&self, number: Trailing<BlockNumber>) -> BoxFuture<Result<RichHeader>> {
+	fn block_header(&self, number: Trailing<BlockNumber>) -> BoxFuture<RichHeader> {
 		use ethcore::encoded;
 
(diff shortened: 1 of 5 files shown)


dvdplm:
> where are those functions used?

Hmm, I looked for usages in my IDE but found none but I chalked it up to poor rust support (CLion). Hmm.

debris:
ah, sry. Haven't noticed that it is a part of the rpc crate. Yes, everything should stay ;)

rpc/src/v1/impls/light/parity.rs
```diff
@@ -390,14 +390,14 @@ impl Parity for ParityClient {
 		})
 	}
 
-	fn block_header(&self, number: Trailing<BlockNumber>) -> BoxFuture<RichHeader> {
+	fn block_header(&self, number: Trailing<BlockNumber>) -> BoxFuture<Result<RichHeader>> {
```
jimpo: This is redundant -- Future can already have an error value. I recommend keeping the return type as is and changing the `map` on line 432 to an `and_then`.

dvdplm: You are right, and your suggestion works great. @debris reverting the return type change and going with @jimpo 's suggestion! :)

PR #8588, opened 2018-05-09
Allow making direct RPC queries from the C API

### LINKED ISSUE #8577: Make it possible to perform RPC requests without a socket through parity-clib
Right now you have to connect through IPC or HTTP.
This doesn't really make sense if we're using Parity as a library.

tomaka:
Close #8577 

Allows performing RPC requests through the C API.
This makes the C API more usable in practice.

I initially wanted to propose a more complete interface where you'd create an RPC request object, then poll it from time to time (or some sort of similar mechanism), then retrieve the result, then destroy it.
However I realized that polling the `Future` of the RPC request would panic because we're not inside of a tokio task.

I then wanted to spawn a thread for each RPC request where I'd block waiting for the request to complete, then send the result through a `mpsc::channel()`. But then I realized that the user of the C API could do this themselves for the same overhead.

Therefore I decided to go with synchronous RPC queries only.

### DIFF
--- a/parity-clib-example/main.cpp
+++ b/parity-clib-example/main.cpp
@@ -1,4 +1,7 @@
 #include <cstddef>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 #include <unistd.h>
 #include <parity.h>
 
@@ -8,8 +11,8 @@ int main() {
     ParityParams cfg = { 0 };
     cfg.on_client_restart_cb = on_restart;
 
-    const char* args[] = {"--light"};
-    size_t str_lens[] = {7};
+    const char* args[] = {"--no-ipc"};
+    size_t str_lens[] = {8};
     if (parity_config_from_cli(args, str_lens, 1, &cfg.configuration) != 0) {
         return 1;
     }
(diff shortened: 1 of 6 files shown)


parity-clib-example/main.cpp
```diff
@@ -8,8 +11,8 @@ int main() {
     ParityParams cfg = { 0 };
     cfg.on_client_restart_cb = on_restart;
 
-    const char* args[] = {"--light"};
-    size_t str_lens[] = {7};
+    const char* args[] = {"--no-ipc"};
```
tomaka: I removed `--light` because `parity_versionInfo` isn't available for the light client (don't really know why).
The `--no-ipc` is here just to give an example usage for the arguments.

parity/run.rs
```diff
@@ -890,42 +895,64 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
...
 		keep_alive: Box<Any>,
 	},
 }
 
 impl RunningClient {
+	/// Performs a synchronous RPC query.
+	/// Blocks execution until the result is ready.
+	pub fn rpc_query_sync(&self, request: &str) -> Option<String> {
+		let metadata = Metadata {
+			origin: Origin::CApi,
```
debris: just a question

are these origins being used somewhere?

5chdn:
@debris, please maintain the A0 label until we have enough reviews so other devs see it still requires a sign-off.

debris:
I see no point in waiting for two reviews before changing the label. We've worked liked that for over 2 years.  'looksgood' always meant that at least one review was ok, but another one is still required.

also, there is a label -`review required`

screen shot 2018-05-22 at 19 30 50

5chdn:
Ok :disappointed:

PR #8589, opened 2018-05-09
Fix XOR distance calculation in discovery Kademlia impl

jimpo:
The distance function was incorrectly being computed as a popcount(XOR) instead of log2(XOR). Also the function to find nearest node entries was not correct as there was a loss of precision in the distance calculation before sorting. This fixes both bugs and reduces the time to find nearest node entries from O(n * log(BUCKET_SIZE)) in all bucket node entries to O(BUCKET_SIZE * log(BUCKET_SIZE)) + O(#buckets).


util/network-devp2p/src/math.rs
```diff
@@ -0,0 +1,97 @@
...
+		// See: http://graphics.stanford.edu/~seander/bithacks.html#IntegerLog
+		impl FastLog for $t {
+			#[inline]
+			fn log2(v: $t) -> usize {
+				if v == 0 {
+					panic!("FastLog::log2 cannot be called with 0");
+				}
+
+				let b: [u64; 6] = [0x2, 0xC, 0xF0, 0xFF00, 0xFFFF0000, 0xFFFFFFFF00000000];
+				let s: [usize; 6] = [1, 2, 4, 8, 16, 32];
```
sorpaas: Maybe change `b` and `s` to static so that we don't need to allocate stack each time `log2` is called?

util/network-devp2p/src/discovery.rs
```diff
@@ -233,17 +249,17 @@ impl Discovery {
...
-			while v != 0 {
-				v >>= 1;
-				ret += 1;
+	/// The base 2 log of the distance between a and b using the XOR metric.
+	fn distance(a: &H256, b: &H256) -> Option<usize> {
+		for i in (0..ADDRESS_BYTES_SIZE).rev() {
+			let byte_index = ADDRESS_BYTES_SIZE - i - 1;
+			let d: u8 = a[byte_index] ^ b[byte_index];
+			if d != 0 {
+				let high_bit_index = FastLog::log2(d);
```
sorpaas: Given that we're only using `FastLog::log2` for `u8`, and we've already had some fixed constant there, what about using something like go-ethereum's [lzcount](https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/node.go#L364)? Not sure about the performance impact though.

util/network-devp2p/src/discovery.rs
```diff
@@ -233,17 +249,17 @@ impl Discovery {
...
-			let mut v: u8 = d[i];
-			while v != 0 {
-				v >>= 1;
-				ret += 1;
+	/// The base 2 log of the distance between a and b using the XOR metric.
+	fn distance(a: &H256, b: &H256) -> Option<usize> {
+		for i in (0..ADDRESS_BYTES_SIZE).rev() {
+			let byte_index = ADDRESS_BYTES_SIZE - i - 1;
+			let d: u8 = a[byte_index] ^ b[byte_index];
+			if d != 0 {
```
sorpaas: Do we need to deal with timing attack here because I suppose this particular function is a crypto function? In the case where network connection is really stable, would it be possible for an attacker to conduct timing attack and leak some information through `PING` time? That function calls `update_node` and uses `distance` when inserting.

Don't know whether that's actually feasible, but if so, maybe we can also just deal with `d != 0` case and return it after 32 loops.

sorpaas: To elaborate more on how this may be an issue (while a really subtle one), an attacker can use timing attack through `PING` on a stable network to guess the bucket index in a range of 8 (because of early return). Then, even if we applied eclipse attack countermeasure 3, the bucket index is still possible to be guessed with an offset of 8.

Public bucket index is an attack factor in the eclipse attack, so it may be possible to then use this information to conduct some sort of similar attacks.

util/network-devp2p/src/math.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright 2018 Coinbase, Inc.
```
debris: I'm not sure if we can accept this license header, cc @5chdn

rphmeier: it's definitely not compatible with the CLA

jimpo: I'm willing to change it if necessary, but my reading of the [CLA](https://www.parity.io/cla.html) made me think it's OK. Section 2.1 a of the CLA says "You retain ownership of the Copyright and Patent Claims in your Contributions and have the same rights to use or license the Contributions which You would have had without entering into the Agreement."

util/network-devp2p/src/discovery.rs
jimpo: I could change it if you want, I think the performance difference is pretty negligible. Integer log is a generally useful operation, regardless, but on 1-byte values a lookup table works as well I suppose.

util/network-devp2p/src/discovery.rs
jimpo: Yes, as long as the distance metric is public, I don't see the risk. If the distance is salted as recommended by the paper, this could be easily defended by sending the pong response packet before calling `update_node` (which would probably be a good idea regardless). Also, I generally think salting the distance is a bad idea...

util/network-devp2p/src/math.rs
5chdn: If you don't mind, let's share the copyright header. This will remove the burdon of future contributors or core developers to keep it updated. What do you think?

    // Copyright 2018 Coinbase Inc., Parity Technologies (UK) Ltd.

jimpo: Updated

util/network-devp2p/src/discovery.rs
```diff
@@ -286,36 +302,53 @@ impl Discovery {
...
+				if !bucket.nodes.is_empty() && append_bucket(&mut ret, bucket) {
+					return ret;
 				}
 			}
 		}
-
-		let mut ret:Vec<NodeEntry> = Vec::new();
-		for nodes in found.values() {
-			ret.extend(nodes.iter().map(|&n| n.clone()));
+		for i in (0..ADDRESS_BITS).rev() {
```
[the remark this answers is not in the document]
jimpo: It partitions the buckets by whether the distance bit is set and searches one partition of buckets in the backwards pass and the other partition in the forward pass.

jimpo:
@tomaka I agree that the discovery code is far from robust. I wrote up some of the other issues I have noticed in https://github.com/paritytech/parity/issues/8757.

debris:
let's give it a try 😉

debris:
thanks @jimpo !

PR #8611, opened 2018-05-12
Fix BlockReward contract "arithmetic operation overflow"

sorpaas:
For calls to `Machine::execute_as_system`, if `gas` is `max_value`, then it is possible to cause this statement to overflow. This sometimes cause the new block reward contract to panic. Changing it to `saturating_add` would fix that.

### DIFF
--- a/ethcore/src/machine.rs
+++ b/ethcore/src/machine.rs
@@ -122,7 +122,13 @@ impl EthereumMachine {
 }
 
 impl EthereumMachine {
-	/// Execute a call as the system address.
+	/// Execute a call as the system address. Block environment information passed to the
+	/// VM is modified to have its gas limit bounded at the upper limit of possible used
+	/// gases including this system call, capped at the maximum value able to be
+	/// represented by U256. This system call modifies the block state, but discards other
+	/// informations. If suicides, logs or refunds happened within the system call, they
+	/// will not be executed or recorded. Gas used by this system call will not be counted
+	/// on the block.
 	pub fn execute_as_system(
 		&self,
 		block: &mut ExecutedBlock,
--- a/ethcore/src/machine.rs
+++ b/ethcore/src/machine.rs
@@ -132,7 +132,7 @@ impl EthereumMachine {
 	) -> Result<Vec<u8>, Error> {
 		let env_info = {
 			let mut env_info = block.env_info();
-			env_info.gas_limit = env_info.gas_used + gas;
+			env_info.gas_limit = env_info.gas_used.saturating_add(gas);
 			env_info
 		};
 
--- a/ethcore/src/machine.rs
+++ b/ethcore/src/machine.rs
@@ -126,7 +126,7 @@ impl EthereumMachine {
 	/// VM is modified to have its gas limit bounded at the upper limit of possible used
 	/// gases including this system call, capped at the maximum value able to be
 	/// represented by U256. This system call modifies the block state, but discards other
-	/// informations. If suicides, logs or refunds happened within the system call, they
+	/// information. If suicides, logs or refunds happen within the system call, they
 	/// will not be executed or recorded. Gas used by this system call will not be counted
 	/// on the block.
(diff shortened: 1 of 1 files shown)


ethcore/src/machine.rs
```diff
@@ -132,7 +132,7 @@ impl EthereumMachine {
 	) -> Result<Vec<u8>, Error> {
 		let env_info = {
 			let mut env_info = block.env_info();
-			env_info.gas_limit = env_info.gas_used + gas;
+			env_info.gas_limit = env_info.gas_used.saturating_add(gas);
```
dvdplm: It would be great to document (and/or add a test for it) what happens in case of an overflow with this change. It fixes the panic, but is the new non-panicky behaviour legal?

sorpaas: I added more docs on how execute_as_system function actually works. I think this is mostly a definition issue -- system transactions are not normal transactions after all.

ethcore/src/machine.rs
```diff
@@ -122,7 +122,13 @@ impl EthereumMachine {
 }
 
 impl EthereumMachine {
-	/// Execute a call as the system address.
+	/// Execute a call as the system address. Block environment information passed to the
+	/// VM is modified to have its gas limit bounded at the upper limit of possible used
+	/// gases including this system call, capped at the maximum value able to be
+	/// represented by U256. This system call modifies the block state, but discards other
+	/// informations. If suicides, logs or refunds happened within the system call, they
```
dvdplm: s/informations/information/
s/suicides, logs or refunds happened within the system call/suicides, logs or refunds happen within the system call/

ethcore/src/machine.rs
```diff
@@ -122,7 +122,13 @@ impl EthereumMachine {
 }
 
 impl EthereumMachine {
-	/// Execute a call as the system address.
+	/// Execute a call as the system address. Block environment information passed to the
+	/// VM is modified to have its gas limit bounded at the upper limit of possible used
+	/// gases including this system call, capped at the maximum value able to be
```
dvdplm: I am not sure if I understand this phrase fully. Can you rephrase the part around "…upper limit of possible used gases including this system call"? Thanks!

sorpaas: Can you explain more which parts looks confusing? Would help a lot for me to rephrase it.

dvdplm: Absolutely. And remember I'm new to much of this so it might just be me being ignorant.  To me it's "upper limit of possible used gases" that is hard to understand.
Maybe something like: "This call executes a call as the system address and modifies the VM's gas limits to be bound to the upper limit, given by the gas amounts provided in the call. Gas is capped to the maximum value represented by a U256" – is that close to what you meant to say?

dvdplm: Ok, I think I understand a little better now, thank you for taking the time. I like your suggested change!

ethcore/src/machine.rs
dvdplm: @sorpaas talking further about this with @ascjones we were wondering what the reason is this method accepts a `gas` param at all: as the system calls are "special" and can use however much gas they want, then perhaps this line could just become `env_info.gas_limit = U256::max_value()`?

cc @andresilva

sorpaas: It do make sense. Sometimes we want to limit the gas can be used by system transactions. Like BLOCKHASH contract (https://eips.ethereum.org/EIPS/eip-210).

andresilva: @dvdplm It's typically used for consensus-level transactions, e.g. finalizing changes to `ValidatorSet`, block reward contract transactions. In these cases we don't really need to cap the execution using gas since we assume the contracts are well behaved. Although we do provide `U256::max_value()` in most cases, for EIP210 we set the gas limit defined in the spec. So I think we should keep the `gas` parameter and don't assume it will always be `U256::max_value()`.

PR #8614, opened 2018-05-14
Refactor EIP150, EIP160 and EIP161 forks to be specified in CommonParams 

sorpaas:
Currently if a non-ethash engine is declared, it can only use post-eip150 schedule with EIP160 and EIP161 all set to true. Some blockchains might not want to do this. Because EIP150, EIP160 and EIP161 are hard fork transitions only related to state, they can be declared in CommonParams rather than specific to ethash engine.


folsen:
Are these changes backwards compatible for the old format of the spec file or would this require everyone to update their spec files?

sorpaas:
@folsen It's not backward compatible. This would require some users to update their spec files, just like what happened to Parity v1.9. So it's tagged `releasenotes`.

folsen:
> This would require some users to update their spec files, just like what happened to Parity v1.9

This was a pretty big deal, caused a fair amount of headache and support requests to come in. I'm afraid I don't have any concrete suggestions for how to easily make it backwards compatible though so, I'm not sure if there's anything we can do to alleviate this problem?

sorpaas:
Hmm I don't have any ideas for making it backward compatible, either. Need some help on this. Probably a versioning system for spec files? (But IMHO any backward compatible solution would instead add a good deal of complexity to the code instead.)

Tbaut:
Stupid question: Do we have concrete examples of chain/user that need this feature?

sorpaas:
@Tbaut One example would be those chains that do not want EIP161 but want to use PoA or PoS engines. Currently there's no way to define that in the current spec format.

5chdn:
We should not avoid changes because it causes an support overhead. We'll bold it in the changelog and fine.

folsen:
@5chdn that's not really a sustainable strategy :P This will break every single enterprise deployment, every private and consortium net everywhere, everyone who has a development net running somewhere. It's not just "support overhead", it's actually causing a massive amount of pain and confusion breaking existing projects. If there's a consortium chain running with 10 authorities, now someone needs to go out and explain to all those 10 partners why they have to go in and change some json file because "parity fucked it up".

If we have a good reason for doing it, like "without it we can't do X" or "party Y is asking for this because they can't launch their chain without EIP161", then I have absolutely no problem including this. If we're doing it because "it's cleaner" or there's some theoretical need, then I'm much more skeptical. 

We'll probably also need more changes in the future, so I don't think a spec migration system is something we should avoid putting in.

sorpaas:
@folsen Can you explain why this breaks "every single enterprise deployment, every private and consortium net everywhere, everyone who has a development net running somewhere"?

This only affects ethash engines chain specs. I don't see why PoA or PoS chain specs would need any migrations with this PR.

folsen:
@sorpaas ah then I have really misunderstood something, in the original post it says " if a non-ethash engine" so I thought it was the opposite, that it affected PoA and PoS.

sorpaas:
@folsen That's from functionality side. With this PR, you can declare EIP150, EIP160 and EIP161 in non-ethash engine specs. But from migration side, existing non-ethash specs are not affected. Only some ethash specs need upgrade.

folsen:
@sorpaas then I'm in favor of it :)

I still think we should have a spec migration "framework" though :P

ethcore/res/ethereum/social.json
```diff
@@ -26,6 +22,10 @@
     "minGasLimit": "0x1388",
     "networkID": "0x1C",
     "chainID": "0x1C",
+	"eip150Transition": "0x0",
```
debris: inconsistent indentation ;)

ethcore/res/ethereum/musicoin.json
```diff
@@ -33,6 +29,10 @@
     "networkID":"0x76740f",
     "forkBlock":"0x1d8015",
     "forkCanonHash":"0x380602acf82b629a0be6b5adb2b4a801e960a07dc8261bf196d21befdbb8f2f9",
+	"eip150Transition":"0x21e88e",
```
debris: inconsistent indentation ;)

ethcore/res/ethereum/mcip6_byz.json
```diff
@@ -33,6 +29,10 @@
     "networkID":"0x76740c",
     "forkBlock":"0x2b",
     "forkCanonHash":"0x23c3171e864a5d513a3ef85e4cf86dac4cc36b89e5b8e63bf0ebcca68b9e43c9",
+	"eip150Transition":"0x2a",
```
debris: inconsistent indentation ;)

ethcore/res/ethereum/mcip3_test.json
```diff
@@ -33,6 +29,10 @@
     "networkID":"0x76740b",
     "forkBlock":"0x5b6",
     "forkCanonHash":"0xa5e88ad9e34d113e264e307bc27e8471452c8fc13780324bb3abb96fd0558343",
+	"eip150Transition":"0x7fffffffffffff",
```
debris: inconsistent indentation ;)

ethcore/res/ethereum/easthub.json
```diff
@@ -26,7 +22,11 @@
     "minGasLimit": "0x1388",
     "networkID": "0x7",
     "chainID": "0x7",
+	"eip150Transition": "0x0",
```
debris: inconsistent indentation ;)

sorpaas: I'm going to fix all spec configs to use tabs instead of spaces given that's what's stated in `.editorconfig`, and the majority of our spec configs are tabbed.

folsen:
This PR did break one thing, http://hivetests.ethstats.net/

On future spec changes we should give a heads up there

holiman:
> On future spec changes we should give a heads up there 

Yeah, you can either ping me or just open a ticket on https://github.com/karalabe/hive

5chdn:
Are we still failing 875 tests?

folsen:
@5chdn I don't believe they've been re-run since this fix https://github.com/karalabe/hive/commit/12b61eb1827f8c87ad9163b780a65cffdcbe26e9

PR #8596, opened 2018-05-10
Fix account list double 0x display

sorpaas:
Fix #8547.

Use LowerHex for stable output.

PR #8623, opened 2018-05-14
[stable] Backports

andresilva:
* Fix account list double 0x display (#8596) 
* Trace precompiled contracts when the transfer value is not zero (#8486)
* Gitlab test script fixes (#8573)


5chdn:
Thanks, can you also apply this: #8573 (just CI)

PR #8625, opened 2018-05-14
Remove NetworkContext::io_channel()

tomaka:
Method is unused.


tomaka:
The tests fail at runtime on gitlab, which doesn't make sense considering the change.

niklasad1:
Agree, `No space left on device (os error 28)` /cc @General-Beck

5chdn:
Need a secondary reviewer.

5chdn:
Thanks

PR #8628, opened 2018-05-14
Use impl Future in the light client RPC helpers

tomaka:
cc @rphmeier who suggested the change

Modifies the light client RPC helpers to use `impl Future` instead of `BoxFuture`, which theoretically leads to slightly better performances.

More than the possible improvement, this change is nice in order to experiment with `impl Trait` and see its impact in general.

It turns out that the compilation is almost unchanged.
The binary size changed a bit (with `--release` then stripped)
Before: 54.73MB
After: 55.03MB

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -25,7 +25,7 @@ use ethcore::ids::BlockId;
 use ethcore::filter::Filter as EthcoreFilter;
 use ethcore::receipt::Receipt;
 
-use jsonrpc_core::{BoxFuture, Result};
+use jsonrpc_core::{Result, Error};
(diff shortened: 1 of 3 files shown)


tomaka:
This PR requires Rust 1.26.
It looks like the version used by gitlab is still 1.25.

rphmeier:
@kirushik do we still have issues with binary size and the auto-updater?

andresilva:
@rphmeier We increased the size limit to 128MB and that was released in 1.9.4 and 1.8.11. And 1.11.0 also includes some changes to the auto-updater. I think it's safe to push binaries larger than 50MB at this point (we probably already did).

5chdn:
Restarted CI with new rust version.

Issue #8620, opened 2018-05-14
Make patricia-trie usable from Polkadot (WAS: Extract patricia-trie to its own repo)


rphmeier:
Along with various bits of generalization in it and its dependencies to make generic over hash function, hash type, and node encoding and decoding.

CC https://github.com/paritytech/polkadot/pull/157

dvdplm:
Tasks:

- [ ] extract patricia_trie's in-tree dependencies into own repos
  - [ ] extract `ethcore-bytes`
  - [ ] extract `keccak-hash`
  - [ ] extract `hashdb`
  - [ ] extract `rlp` (maybe wait on this one, might be a big task)
  - [ ] extract `triehash`
  - [ ] extract `memorydb`
  - [ ] extract `ethcore-logger`
- [ ] extract patricia_trie into own repo
- [ ] make `trait Hash` generic over the hashing algorith (investigate doing this in the `hashdb` crate
- [ ] make `Node` generic over hash; add a `NodeCodec` trait
- [ ] implement `NodeCodec` for `rlp`
- [ ] make sure encoding does not suffer performance hit; investigate using `smallvec` here
- [ ] port structures like `TrieDB` to be generic `TrieDB<H: Hash, C: NodeCodec<H>>`

Open questions:
- Do we need to extract `rlp` as well?

rphmeier:
Looks about right. We might want to introduce a small crate just for the hash trait so other code (hashdb, triehash, etc.) can depend on it, and a separate crate for `rlp-node-codec`.

debris:
Sorry for not replying immediately to this thread. I don't agree with one of the points here, which is `extract patricia_trie's in-tree dependencies into own repos`. I believe it's redundant and will introduce **a lot** of maintenance burden.

> extract trie-standardmap – @debris this code does not have a license; you ok with GPL v3?

great catch, sure!

> extract plain_hasher – @debris this crate is owned by you so I can't push updates. Can you transfer it to parity?

historically, plain_hasher was a separate repo, but it's usacase is very specific for us, and after discussing it with Arkadiy it decided to move it to parity to avoid maintenance burden

5chdn:
I have no strong opinion here, but I want to highlight that we should rename `ethcore-*` crates to `parity-*` soon. If we touch this code now, let's keep it in mind. cc #8649

adria0:
Closing the issue due to its stale state

PR #8630, opened 2018-05-14
Use ethcore-bytes from crates.io

dvdplm:
Move ethcore-bytes out-of-tree and update uses.

Part of issue #8620.


Cargo.toml
```diff
@@ -32,7 +32,7 @@ fdlimit = "0.1"
 ctrlc = { git = "https://github.com/paritytech/rust-ctrlc.git" }
 jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
 ethcore = { path = "ethcore" }
-ethcore-bytes = { path = "util/bytes" }
+ethcore-bytes = "0.1.1"
```
debris: Imo, we should keep everything in one repo. This could be just

```toml
ethcore-bytes = { path = "util/bytes", version = "0.1.1" }
```

debris:
putting `onice` due to https://github.com/paritytech/parity/issues/8620#issuecomment-389741501

I believe it's not a good idea

PR #8633, opened 2018-05-15
Use rlp from crates

dvdplm:
Move the rlp out of tree and use the crate from crates.io. Part of issue #8620.


Cargo.toml
```diff
@@ -46,7 +46,7 @@ ethereum-types = "0.3"
 node-filter = { path = "ethcore/node_filter" }
 ethkey = { path = "ethkey" }
 node-health = { path = "dapps/node-health" }
-rlp = { path = "util/rlp" }
+rlp= "0.2.2"
```
rphmeier: prefer a space before `=` (in all other instances as well)

dvdplm: fixed

dvdplm:
closing

PR #8634, opened 2018-05-15
Move triehash and trie-standardmap out of tree

dvdplm:
Use triehash and trie-standardmap from crates.io

Part of #8620.


5chdn:
Please always apply at least one M-label to each PR

debris:
putting `onice` due to https://github.com/paritytech/parity/issues/8620#issuecomment-389741501

I believe it's not a good idea

PR #8641, opened 2018-05-16
Don't open Browser post-install on Mac, fixes #8194

### LINKED ISSUE #8194: 1.10-beta MacOS installer opens 127.0.0.1:8180
> I'm running:
>
> - **Which Parity version?**: 1.10.0 beta
> - **Which operating system?**: macos

Just download the 1.10.0-beta binary from parity.io. The installer on OSX opens 127.0.0.1:8180 in the browser (behavior in previous installers), but now ui is disabled by default, so it shows page not found.

gnunicorn:
Since we start parity with the UI disabled per default now, opening the browser post installation will show an annoying error message, confusing the user. This patch removes opening the browser to prevent that annoyance.

fixes #8194


gnunicorn:
[clabot:check]

5chdn:
We should remove all installers, too.

Issue #7008, opened 2017-11-09
Ancient block sync stalls


dip239:
> I'm running:
>
> - **Parity version**: Parity/v1.8.2-beta-1b6588c-20171025/x86_64-linux-gnu/rustc1.21.0
> - **Operating system**: Linux
> - **And installed**: sudo apt-get install ./parity_1.8.2_amd64.deb

configuration detail: chain is simlinked to another drive
```
human@EtherBox:~$ ls -lisa ./.local/share/io.parity.ethereum/chains
524382 0 lrwxrwxrwx 1 human human 28 Sep  3 21:25 ./.local/share/io.parity.ethereum/chains -> /media/human/CHAIN_DB/chains
```

After warp sync, I'm unable to get all missing blocks fetched. After new start Parity forgets the already fetched blocks and start downloading them again and again.

Here is a sample of 3 parity runs

![image](https://user-images.githubusercontent.com/4574735/32586099-1504cf62-c501-11e7-86cc-c44abe018ca3.png)

 
Here is the log file from above (as text file)
[parity-3x-sync.txt](https://github.com/paritytech/parity/files/1456141/parity-3x-sync.txt)

---

roninkaizen:
normal behaviour/ misunderstanding-
as you looked at the screen you posted, it became obvious-
confirmed = yes
unwanted, unintended=no
normal behavior, other counting on the transfers, 
also having that, being fully sync on that nodes, 
irritiating=yes
annoyance=no

Document it, then it is known to be normal. thx

5chdn:
That's the ancient block download.

Warp-sync downloads the latest snapshot and the last 30k blocks. After that, it starts downloading the full blockchain (yellow numbers).

dip239:
@5chdn @roninkaizen 
What is the reason to download ancient blocks again and again? This is the issue reported.

Please check the block numbers:
1st run: downloaded ancient blocks from #3574411 to #3614671
2st run: downloaded ancient blocks from #3573015 to #3576698    
3st run: downloaded ancient blocks from #3573269  to #3575935    

ancient blocks being downloaded in the same range again und again. 
What is the reason to do it if it is not a bug?

arkpar:
This is really weird indeed. Never seen this before. Could you restart with `-l sync=trace` and post logs?

dip239:
yes, please:
Here 4 subsequent runs, two last are with `-l sync=trace`.
[...]

PR #8642, opened 2018-05-16
Fix ancient blocks not downloading

ngotchac:
This fixes ancient blocks not being downloaded.

What happened is that the importing of ancient blocks was asynchronous through a `IoChannel`, and thus checking for the parent being in the chain would fail since the previous wouldn't be imported yet. This is fixed with the `pending_ancient_blocks` `HashSet`.
Then, the importing of blocks would be queued in the `IoChannel` but not unqueued in order, thus leading to the failure of importing blocks, since the parent was needed to compute the total difficulty.
This is fixed by using a proper `VecDequeue` queue in the `IoChannel` that acts as a FIFO.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -17,7 +17,7 @@
 use std::collections::{HashSet, BTreeMap, BTreeSet, VecDeque};
 use std::fmt;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


5chdn:
This? #7008

5chdn:
IMHO this is worth a patch release :)

PR #8643, opened 2018-05-16
Keep existing blocks when restoring a Snapshot

### LINKED ISSUE #6350: Keep existing blocks when restoring from a snapshot.
When restoring from a snapshot parity should try and re-import ancient blocks existing in the current database starting from genesis.

ngotchac:
Closes #6350

In order to restore existing blocks, it just iterates over the blocks of the current DB before swapping it with the snapshot one, from 1 (or first needed block) until all the blocks have been imported.

To test the feature, one can sync without warp-sync for a few thousand blocks, then restart with `--warp-barrier`

Blocks restoration seemed pretty fast (~15 seconds for 150_000 blocks), but if too slow this could be improved by skipping caches.


ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+						}
+					// Break if we already imported some blocks in the current batch and there
+					// are no more left
+					} else if parent_diff.is_some() {
+						break;
+					}
+
+					// Writting changes to DB and logging every now and then
+					if block_number % 1_000 == 0 {
+						next_db.write_buffered(batch);
```
rphmeier: `write_buffered` will keep them in memory. we need to flush periodically

ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+						let block_receipts = chain.block_receipts(&block_hash);
+
+						match (block, block_receipts) {
+							(Some(block), Some(block_receipts)) => {
+								let diff = block.difficulty();
+								let raw_block = block.into_inner();
+								let block_receipts = block_receipts.receipts;
+
+								next_chain.insert_unordered_block(&mut batch, &raw_block, block_receipts, parent_diff, false, true);
+								parent_diff = Some(diff);
```
rphmeier: `parent_diff` is meant to be a total difficulty

ngotchac: Oups, right...

ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+
+		let cur_chain = &self.client.chain();
+		let next_db = self.restoration_db_handler.open(&rest_db)?;
+		let next_chain = BlockChain::new(Default::default(), &[], next_db.clone());
+		let next_chain_info = next_chain.chain_info();
+
+		// Try to include every block that will need to be downloaded from the current chain
+		// Break when no more blocks are available from it.
+		match (next_chain_info.ancient_block_number, next_chain_info.first_block_number) {
+			(Some(next_ancient_block), Some(next_first_block)) if next_ancient_block + 1 < next_first_block => {
```
rphmeier: `ancient_block_number` is always going to be `1` (or `0`, can't remember) after a restoration, isn't it?

rphmeier: the blocks we want to import from the old chain: anything available from genesis to `first_block_number`, but maintaining the invariant that only a single gap exists.

ngotchac: Yep that's right, it's supposed to always be `1`. It does ensure that there is still only one gap of blocks.

ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+				trace!(target: "snapshot", "Trying to import ancient blocks from {} to {}",
+					next_ancient_block, next_first_block,
+				);
+
+				let mut block_number = next_ancient_block + 1;
```
rphmeier: really needs to check that the first ancient block we pull from the old client has the right `parent_hash`.

ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+								let block_receipts = block_receipts.receipts;
+
+								next_chain.insert_unordered_block(&mut batch, &raw_block, block_receipts, parent_diff, false, true);
+								parent_diff = Some(diff);
+							},
+							_ => (),
+						}
+					// Break if we already imported some blocks in the current batch and there
+					// are no more left
+					} else if parent_diff.is_some() {
```
rphmeier: why only break if `parent_diff` is `Some`? if there are no more blocks we can import it seems like it should _always_ break the loop.

ethcore/src/snapshot/service.rs
```diff
@@ -327,9 +328,74 @@ impl Service {
...
+				);
+
+				let mut block_number = next_ancient_block + 1;
+				let mut batch = DBTransaction::new();
+				let mut parent_diff = None;
+
+				while block_number < next_first_block {
+					let chain = cur_chain.read();
+
+					if let Some(block_hash) = chain.block_hash(block_number) {
```
rphmeier: we don't have an exclusive lock on the chain's DB. this is racing with client's normal block import. (in practice this isn't a problem right now but let's not lean on assumptions from outside the module)

ethcore/src/snapshot/service.rs
ngotchac: Yep that's right, didn't push this change yet.

ethcore/src/snapshot/service.rs
rphmeier: > The blocks won't change during it
why?
this is an assumption beyond the module.

rphmeier: This is easiest when we import blocks in reverse because the `parent_hash` is always available. But then we should always make sure that `next_first_block.parent_hash` corresponds to the hash of a block that we import as its parent.

ngotchac: So could it be just a check that for each block, the parent's hash is the one expected, ie. the one we got from the previous block ? If there is a mismatch, then just stop right there.

rphmeier:
Since GitHub marks that comment as hidden I'll continue here:

> So could it be just a check that for each block, the parent's hash is the one expected, ie. the one we got from the previous block ? If there is a mismatch, then just stop right there.

sure, but that's a little weird because we know that those blocks are still probably in the DB, just that there was a reorg in the meantime

ngotchac:
But do we keep blocks in DB that has been _reorged_?

rphmeier:
yes

ngotchac:
@rphmeier I update the PR so that it starts at the best available ancient block, and it iterates backwards from the parent's hash.
It actually also fixes an issue with resuming snapshot restoration with all the needed chunks (it would previously hang).

ethcore/src/client/client.rs
```diff
@@ -845,6 +845,11 @@ impl Client {
 		*self.exit_handler.lock() = Some(Box::new(f));
 	}
 
+	/// Returns the chain reference
+	pub fn chain(&self) -> &RwLock<Arc<BlockChain>> {
```
debris: It's a bad practice to expose `&RwLock`. There is no guarantee this will not deadlock

ethcore/src/snapshot/service.rs
```diff
@@ -16,6 +16,7 @@
 
 //! Snapshot network service implementation.
 
+// use std::cmp;
```
debris: should be removed

ethcore/src/snapshot/service.rs
```diff
@@ -220,7 +221,7 @@ pub struct ServiceParams {
 	/// Usually "<chain hash>/snapshot"
 	pub snapshot_root: PathBuf,
 	/// A handle for database restoration.
-	pub db_restore: Arc<DatabaseRestore>,
+	pub client: Arc<Client>,
```
debris: It's contradictory to @0x7CFE refactor of `ethcore`. We don't want any modules to require `Client`, but only the interface that is actually being used

ethcore/src/snapshot/tests/service.rs
```diff
@@ -103,6 +102,9 @@ fn restored_is_equivalent() {
 
 #[test]
 fn guards_delete_folders() {
+	let gas_prices = vec![1.into(), 2.into(), 3.into(), 999.into()];
+	let client = generate_dummy_client_with_spec_and_data(Spec::new_null, 400, 5, &gas_prices);
```
debris: iirc, this helper function generates client with 400 blocks. you should check in this test if they were actually migrated correctly

ethcore/src/blockchain/blockchain.rs
```diff
@@ -838,6 +827,80 @@ impl BlockChain {
 		}
 	}
 
+	/// Update the best ancient block to the given hash, after checking that
+	/// it's directly linked to the currently known best ancient block
+	pub fn update_best_ancient_block(&self, hash: &H256) {
```
debris: This function is called only from a single place. Is it possible for hash to not be linked to any known ancient block? If yes, what does it mean? Should it be handled somehow? Currently the result of this function execution is unknown which makes it very difficult to debug

ngotchac: So this function will go from the given hash, and will ensure there is a link between the block at the given hash and the last know best ancient block. Thus, it only update the best ancient block if there is a link.

ethcore/src/snapshot/tests/service.rs
ngotchac: I'm not sure I understand. This test only tests if the guarded folders are deleted.

debris: from #6350

> When restoring from a snapshot parity should try and re-import ancient blocks existing in the current database starting from genesis.

We need a test for that ;)

ngotchac: Ah yeah sure, I thought you were commenting on the `guards_delete_folders` test!

5chdn:
Marking as _stale_ :)

ngotchac:
Finally added the test case @debris !

5chdn:
@debris @tomusdrw could you have a look at this PR again?

5chdn:
> @debris @tomusdrw could you have a look at this PR again?

anyone?

ethcore/src/snapshot/service.rs
```diff
@@ -334,12 +338,144 @@ impl Service {
...
+			None => return Ok(count),
+		};
+
+		// Check that the intersection is non-null
+		if start_block_number >= end_block_number {
+			return Ok(count);
+		}
+
+		// Find the best block we can sync to
+		match (cur_chain_info.ancient_block_number, cur_chain_info.best_block_number) {
```
tomusdrw: Can you elaborate on the logic here a bit more? My understanding is that this is what the dbs look like:

```
current_chain:
[0, cur.ancient_block_number] ... [cur.first_block, cur.best_block]

new_chain:
[0, 0] ... [new.first_block, new.best_block]
```

So we need to import everything from the following ranges:
```
[0, min(cur.ancient_block_number, new.first_block)] if cur.ancient_block_number.is_some()
// or 
[0, min(cur.best_block, new.first_block)]
```

If that's the case we should try to simplify the logic here (and error handling especially - following this `match` statements makes it a bit hard to read).

I'd propose adding a closure that returns `-> Option<(from_hash, to_hash)>` and utilize `?` to early exit all `None`s, so the ccode would look like this:
```rust
/// Finds a range of blocks to import.
///
/// The old database looks like this:
/// [genesis, best_ancient_block] ... [first_block, best_block]
/// If we are fully synced neither `best_ancient_block` nor `first_block` is set, and we can assume that the whole range from [genesis, best_block] is imported.
/// The new database only contains the tip of the chain, so the useful set of blocks is defined as:
/// [0 ... min(new.first_block, best_ancient_block or best_block)]
fn find_range() -> Option<...> {
  let from = next_chain_info.ancient_block_number? + 1; // is this ever non 0?
  let to = next_chain_info.first_block_number? - 1;

  let available_to = cur_chain.ancient_block_number.unwrap_or(cur_chain_info.best_block_number);

  let to = cmp::min(to, available_to);

  if from >= to {
    return None;
  }

  let from_hash = self.client.block_hash(BlockId::Number(from - 1))?; //why -1?
  let to_hash = self.client.block_hash(BlockId::Number(to))?;

  Some((from_hash, to_hash))
}

```

ethcore/src/snapshot/service.rs
```diff
@@ -480,13 +616,22 @@ impl Service {
...
 		self.import_prev_chunks(&mut res, manifest).ok();
 
-		*self.status.lock() = RestorationStatus::Ongoing {
-			state_chunks: state_chunks as u32,
-			block_chunks: block_chunks as u32,
-			state_chunks_done: self.state_chunks.load(Ordering::SeqCst) as u32,
-			block_chunks_done: self.block_chunks.load(Ordering::SeqCst) as u32,
+		// It could be that the restoration failed or completed in the meanwhile
+		let mut restoration_status = self.status.lock();
+		let update_status = match *restoration_status {
```
tomusdrw: Would rather early exit here:
```rust
if let RestorationStatus::Initializing {..} = *restoration_status {
  return Ok(())
}

*restoration_status = RestorationStatus::Ongoing {
...
}

Ok(())
```

ethcore/src/snapshot/service.rs
```diff
@@ -756,26 +901,20 @@ impl Drop for Service {
...
 	use tempdir::TempDir;
-	use test_helpers::restoration_db_handler;
-
-	struct NoopDBRestore;
-	impl DatabaseRestore for NoopDBRestore {
-		fn restore_db(&self, _new_db: &str) -> Result<(), Error> {
-			Ok(())
-		}
-	}
+	// use test_helpers_internal::restoration_db_handler;
```
tomusdrw: Remove

ethcore/src/snapshot/tests/service.rs
```diff
@@ -146,3 +144,196 @@ fn guards_delete_folders() {
...
+		_ => panic!("Invalid Snapshot Service status."),
+	}
+
+	// Check that the latest block number is the right one
+	assert_eq!(client2.block(BlockId::Latest).unwrap().number(), NUM_BLOCKS as u64);
+
+	// Check that we have blocks in [NUM_BLOCKS - NUM_SNAPSHOT_BLOCKS + 1 ; NUM_BLOCKS]
+	// but none before
+	assert!(client2.block(BlockId::Number(NUM_BLOCKS - NUM_SNAPSHOT_BLOCKS + 1)).is_some());
+	assert!(client2.block(BlockId::Number(100)).is_none());
```
tomusdrw: Maybe do it in a loop?

ethcore/src/snapshot/tests/service.rs
```diff
@@ -146,3 +144,196 @@ fn guards_delete_folders() {
...
+
+	// Check that the first 50 blocks have been migrated
+	assert!(client2.block(BlockId::Number(49)).is_some());
```
tomusdrw: Same here.

ethcore/sync/src/chain/mod.rs
```diff
@@ -938,6 +938,16 @@ impl ChainSync {
 				self.state = SyncState::Blocks;
 				self.continue_sync(io);
 			},
+			SyncState::SnapshotData => {
+				match io.snapshot_service().status() {
+					RestorationStatus::Inactive |
```
tomusdrw: IMHO the line breaks are unnecessary, I'd also move the `match to line above so that it looks like this:
```
SyncState::SnapshotData => match io.snapshot_service().status() {
  RestorationStatus::Inactive | RestorationStatus::Failed => {
    ...
  },
  RestorationStatus::Initializing { .. } | RestorationStatus::Ongoing { .. } => (),
}

File: ethcore/Cargo.toml [hunk 18]
```diff
@@ -77,6 +77,9 @@ fake-hardware-wallet = { path = "../util/fake-hardware-wallet" }
 
 [dev-dependencies]
 trie-standardmap = { path = "../util/trie-standardmap" }
+kvdb-rocksdb = { path = "../util/kvdb-rocksdb" }
```
tomusdrw: Why is this needed now exactly?

ethcore/Cargo.toml
```diff
@@ -77,6 +77,9 @@ fake-hardware-wallet = { path = "../util/fake-hardware-wallet" }
...
+env_logger = "0.4"
+tempdir = "0.3"
```
tomusdrw: Same question as `kvdb-rocksdb`

ethcore/src/snapshot/service.rs
```diff
@@ -334,12 +338,144 @@ impl Service {
...
+		let last_block_hash = match self.client.block_hash(BlockId::Number(end_block_number)) {
+			Some(h) => h,
+			None => return Ok(count),
+		};
+
+		let mut success = false;
+		let mut batch = DBTransaction::new();
+		let mut block = self.client.block(BlockId::Hash(last_block_hash));
+
+		while let Some(block_view) = block {
```
tomusdrw: I'd simplify this loop a bit as well, maybe lower the number of variables and use stuff like `block_view.hash()` directly.
```rust
let mut parent_hash = last_block_hash;
while target_hash != parent_hash {
  let block = self.client.block(BlockId::Hash(parent_hash))?;
  // move to parent block
  parent_hash = block.parent_hash();

  let receipts = self.client.block_receipts(&block.hash())?;
  let parent_difficulty= self.client.block_total_difficulty(BlockId::Hash(block.parent_hash()))?;

   //insert block, etc

   if block.number() % 10_000 {
   }
}
```

^^ something like this

ngotchac:
Thanks @tomusdrw ; I'll address those in the coming day(s)!

ethcore/src/blockchain/blockchain.rs
```diff
@@ -843,6 +832,80 @@ impl BlockChain {
...
+
+			// Ensure that the new best ancient block is after the current one
+			if block_view.number() <= cur_ancient_block.number {
+				return;
+			}
+
+			let mut block_hash = *hash;
+			let mut is_linked = false;
+
+			loop {
```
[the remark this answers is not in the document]
ngotchac: Actually, it prevents the `update_best_ancient_block` function to be called twice in parallel, which could break the wanted behavior.

ethcore/src/client/traits.rs
```diff
@@ -287,6 +287,9 @@ pub trait BlockChainClient : Sync + Send + AccountData + BlockChain + CallContra
 	/// Get raw block receipts data by block header hash.
 	fn block_receipts(&self, hash: &H256) -> Option<Bytes>;
```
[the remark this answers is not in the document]
ngotchac: I'm not sure I get your point. `block_receipts` was already part of the trait. It was basically getting the decoded data and encoding it to RLP. I only added the `decoded` version which kind of splits the logic in two.

ethcore/src/snapshot/service.rs
```diff
@@ -203,6 +204,9 @@ impl Restoration {
 /// Type alias for client io channel.
 pub type Channel = IoChannel<ClientIoMessage>;
 
+/// Trait alias for the Client Service used
+pub trait CustomClient: BlockChainClient + BlockInfo + DatabaseRestore {}
```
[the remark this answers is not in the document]
ngotchac: +1 for SnapshotClient, but a type doesn't work, I get a `non-auto additional trait` error. I remember struggling with that but don't remember the details (I think I asked on Riot and that was the only solution that worked).

ethcore/src/snapshot/service.rs
```diff
@@ -334,12 +338,144 @@ impl Service {
...
+		let rest_db = self.restoration_db();
+
+		let cur_chain_info = self.client.chain_info();
+
+		let next_db = self.restoration_db_handler.open(&rest_db)?;
+		let next_chain = BlockChain::new(Default::default(), &[], next_db.clone());
+		let next_chain_info = next_chain.chain_info();
+
+		// Get the *inclusive* first and last blocks to iterate through
+		let start_block_number = match next_chain_info.ancient_block_number {
```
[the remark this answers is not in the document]
ngotchac: True, I actually wanted to write a generic `migrate_block` methods for any DB, not necessarily for Snapshots. However the new chain will have somewhere around 30_000 blocks from the snapshot.

ethcore/src/snapshot/tests/service.rs
ngotchac: I'm not sure it would add that much, considering we would have to loop over 5_000 blocks (which could maybe take quite some time?).

ethcore/src/snapshot/tests/service.rs
ngotchac: Same as above

ethcore/src/snapshot/service.rs
```diff
@@ -334,12 +338,144 @@ impl Service {
...
+			None => return Ok(count),
+		};
+
+		let mut success = false;
+		let mut batch = DBTransaction::new();
+		let mut block = self.client.block(BlockId::Hash(last_block_hash));
+
+		while let Some(block_view) = block {
+			// Early return if restoration is aborted
+			if !self.restoring_snapshot.load(Ordering::SeqCst) {
```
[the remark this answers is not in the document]
ngotchac: Do we care if some blocks are flushed to DB when the restoration is aborted? Anyway the DB should be deleted afterwards.

ethcore/src/blockchain/blockchain.rs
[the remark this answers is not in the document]
tomusdrw: It doesn't really. `RwLock::read` can be acquired from multiple threads, although it might deadlock if called from the same thread recursively.
Calling other functions dependent on `sefl` while holding a lock is imho asking for troubles, especially if the code is not thoroughly tested. Another person will try to access the same lock in some other function and it will break.

ethcore/src/client/traits.rs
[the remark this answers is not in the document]
tomusdrw: Yeah, so my point is that we don't really need both methods, especially that they are returning the same data, just in a different format.

I'd go for a small refactoring and:
1. Renaming `decoded_block_reeipts` to `block_receipts` (more in-line with other methods that return typed data from the client)
2. Changing existing occurences to encode the receipts on their own (I suspect there isn't many).

ethcore/src/blockchain/blockchain.rs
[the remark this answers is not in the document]
ngotchac: Hm, right. So should I use a `write` lock? The only function called here is `block_details` which we are sure doesn't use any lock.

ethcore/src/client/traits.rs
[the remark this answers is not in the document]
ngotchac: Got it.

ngotchac:
@tomusdrw @debris ?!

ethcore/src/blockchain/blockchain.rs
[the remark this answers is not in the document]
debris: I believe there is a logic error here (or in `insert_unordered_block`).

`insert_unordered_block` updates `best_ancient_block`, but does not update `block_details`

https://github.com/paritytech/parity-ethereum/blob/1b1941a896c8485a05fa3e9ffe8b251b498eb541/ethcore/src/blockchain/blockchain.rs#L782-L795

it updates only `pending_block_details`

https://github.com/paritytech/parity-ethereum/blob/1b1941a896c8485a05fa3e9ffe8b251b498eb541/ethcore/src/blockchain/blockchain.rs#L1092-L1098

`block_details` are updated only in `commit` function

https://github.com/paritytech/parity-ethereum/blob/1b1941a896c8485a05fa3e9ffe8b251b498eb541/ethcore/src/blockchain/blockchain.rs#L1127

So there will be a race condition if this function is called between `insert` and `commit`

ethcore/src/blockchain/blockchain.rs
```diff
@@ -843,6 +832,80 @@ impl BlockChain {
...
+		self.set_best_ancient_block(block_view.number(), hash, &mut batch);
+		self.db.key_value().write(batch).expect("Low level database error.");
+	}
+
+	/// Set the best ancient block with the given value: private method
+	/// `best_ancient_block` must not be locked, otherwise a DeadLock would occur
+	fn set_best_ancient_block(&self, block_number: BlockNumber, block_hash: &H256, batch: &mut DBTransaction) {
+		let mut best_ancient_block = self.best_ancient_block.write();
+		let ancient_number = best_ancient_block.as_ref().map_or(0, |b| b.number);
+		if self.block_hash(block_number + 1).is_some() {
```
debris: I see that this line hasn't been changed, but I believe there is a logic error here. Subsequent calls to `insert_unordered_block` store block hashes only in the pending collection. Therefore it may happen that `best_ancient_block` will be never set to `Some` and `self.block_hash()` will always return `None` leading the database into unknown state (maybe it gets stack exactly because of that?)

ngotchac: I think it's fine because the block hash is written to DB before calling `set_best_ancient_block` in `prepare_update`.

5chdn:
Let's get this reviewed soon @debris @tomusdrw :)

ethcore/src/snapshot/service.rs
```diff
@@ -333,12 +338,108 @@ impl Service {
...
+			let target_block_num = cmp::min(next_available_from - 1, cur_available_to);
+
+			if target_block_num == 0 {
+				return None;
+			}
+
+			trace!(target: "snapshot", "Trying to import ancient blocks until {}", target_block_num);
+
+			let target_hash = self.client.block_hash(BlockId::Number(0))?;
+			let start_hash = self.client.block_hash(BlockId::Number(target_block_num))?;
```
tomusdrw: Shouldn't that be the other way around, i.e.:
```rust
let start_hash = self.client.block_hash(BlockId::Number(0))?;
let target_hash = self.client.block_hash(BlockId::Number(target_block_num))?;
```
?

tomusdrw: Oh I see, we go the other way around, so `start_hash` is where we start the loop from. TBH It's quite confusing, I think it would actually be clearer to start from `target_hash` and check if we got to `start_hash`, cause somehow implicitly `num(start_hash) < num(target_hash)` and the fact that we are iterating backwards doesn't really change it. (maybe change `target_hash` to `end_hash`?)

ethcore/light/src/provider.rs
```diff
@@ -174,7 +174,7 @@ impl<T: ProvingBlockChainClient + ?Sized> Provider for T {
 	}
 
 	fn block_receipts(&self, req: request::CompleteReceiptsRequest) -> Option<request::ReceiptsResponse> {
-		BlockChainClient::block_receipts(self, &req.hash)
+		BlockChainClient::encoded_block_receipts(self, &req.hash)
```
tomusdrw: Doesn't make sense to use `encoded` method here, since we decode it right away, no?

ethcore/sync/src/chain/supplier.rs
```diff
@@ -226,7 +226,7 @@ impl SyncSupplier {
 		let mut added_receipts = 0usize;
 		let mut data = Bytes::new();
 		for i in 0..count {
-			if let Some(mut receipts_bytes) = io.chain().block_receipts(&rlp.val_at::<H256>(i)?) {
+			if let Some(mut receipts_bytes) = io.chain().encoded_block_receipts(&rlp.val_at::<H256>(i)?) {
```
tomusdrw: IF we remove the other occurence of `encoded_block_receipts`, it seems that's the only one. I'd remove that method from a trait completely, and just use `.map(::rlp::encode(&x)`` here instead.

ethcore/src/snapshot/service.rs
ngotchac: Well the `target_hash` is really the target here: we are checking that starting from `start_hash` we can reach the target.

5chdn:
Please rebase on master if possible

5chdn:
Please reopen when ready

ngotchac:
Rebased on master

5chdn:
@debris @tomusdrw could you give this a final review?

@ngotchac sorry, could you rebase this again?

ethcore/src/blockchain/blockchain.rs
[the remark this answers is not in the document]
tomusdrw: ^^ @ngotchac is this addressed?

ethcore/sync/src/chain/supplier.rs
```diff
@@ -226,7 +226,9 @@ impl SyncSupplier {
 		let mut added_receipts = 0usize;
 		let mut data = Bytes::new();
 		for i in 0..count {
-			if let Some(mut receipts_bytes) = io.chain().encoded_block_receipts(&rlp.val_at::<H256>(i)?) {
+			if let Some(receipts) = io.chain().block_receipts(&rlp.val_at::<H256>(i)?) {
+				let mut receipts_bytes = ::rlp::encode(&receipts).into_vec();
+			// if let Some(mut receipts_bytes) = io.chain().encoded_block_receipts(&rlp.val_at::<H256>(i)?) {
```
tomusdrw: remove please?

ethcore/src/snapshot/service.rs
```diff
@@ -334,12 +339,110 @@ impl Service {
...
+		let cur_chain_info = self.client.chain_info();
+
+		let next_db = self.restoration_db_handler.open(&rest_db)?;
+		let next_chain = BlockChain::new(Default::default(), &[], next_db.clone());
+		let next_chain_info = next_chain.chain_info();
+
+		// The old database looks like this:
+		// [genesis, best_ancient_block] ... [first_block, best_block]
+		// If we are fully synced neither `best_ancient_block` nor `first_block` is set, and we can assume that the whole range from [genesis, best_block] is imported.
+		// The new database only contains the tip of the chain ([forst_block, best_block]),
```
ascjones:
```suggestion
		// The new database only contains the tip of the chain ([first_block, best_block]),
```

ngotchac:
:tada: !!!

PR #8644, opened 2018-05-16
Remove HostInfo::next_nonce

tomaka:
Another unused method that can be moved out of its trait.


niklasad1:
@tomaka 
I don't understand this, it looks like you have a moved the method `next_nonce` from the trait `HostInfoTrait` into the struct `HostInfo`! 

Why, isn't `next_nonce` unused?

PR #8645, opened 2018-05-16
add crosscompile config

General-Beck:
target's for armv7, arm64 and android-armv7


General-Beck:
@tomaka we can take out as a patch in a separate script of CI `- before_script`

5chdn:
I don't have any strong opinion on this. @kirushik ?

kirushik:
@afri I'm with Pierre in believing this is not the proper place; also. l think that https://github.com/paritytech/scripts/blob/master/docker-files-for-Gitlab-CI-rust/cross/android/Dockerfile#L75-L80 is the way we go now? Then this PR is not needed anymore (given linkers for all the platforms are covered in our build servers/containers setup)

PR #8650, opened 2018-05-17
parity: disable ancient blocks download by default

5chdn:
putting this up for discussion as suggested by @folsen

- adds `--ancient-blocks` parameter (disabled by default)
- disables ancient blocks download by default
- move `--no-ancient-blocks` parameter to legacy (`Deprecated::DoesNothing`)
- cleans up some whitespaces (can't we automatically detect issues somehow?)

open questions raised

- shall we really default to a node without full block history
- are there any security implications

### DIFF
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
 // This file is part of Parity.
 
 // Parity is free software: you can redistribute it and/or modify
@@ -422,9 +422,9 @@ usage! {
 			"--reserved-only",
 			"Connect only to reserved nodes.",
 
-			FLAG flag_no_ancient_blocks: (bool) = false, or |_| None,
-			"--no-ancient-blocks",
-			"Disable downloading old blocks after snapshot restoration or warp sync.",
+			FLAG flag_ancient_blocks: (bool) = false, or |_| None,
+			"--ancient-blocks",
+			"Enable downloading old blocks after snapshot restoration or warp sync. Turns client into a full node.",
 
 			FLAG flag_no_serve_light: (bool) = false, or |c: &Config| c.network.as_ref()?.no_serve_light.clone(),
 			"--no-serve-light",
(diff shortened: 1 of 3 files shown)


folsen:
So, there's two aspects to this issue.

One is; what can we do right now. On this point I would actually say that we should add a `--semi-light` parameter that is a combination of `--no-ancient-blocks` and a computed `--warp-barrier`.

The second is; what can we do in the future. Here I would say we should use the hardcoded sync points used for the light client to verify the headers in the snapshot when it's downloaded, **and** add a new block-pruning algorithm that updated state but removes block bodies afterwards.

So you would download full blocks from your snapshot height to best block, but they'd be removed after being validated and the state is updated, and of course any new blocks created would be downloaded (as opposed to a light-client), validated, state updated and then body is pruned. 

This is what I'd call a "true" `--semi-light`, so the `--semi-light` option can be migrated from just a simple `--no-ancient-blocks` to a more real structure like this. I still can't think of any security flaws with this, but maybe someone else can. 

It hurts the data availability as much as light clients do, but I think that's just something we have to live with, and maybe it'll actually push the question of full-node incentivisation if too many people use this option.

folsen:
@tomusdrw So 1) and 2) are fixable. 

Point 3 is one I hadn't thought about. That is indeed a problem, and sort of negates the point of having this unless there can be a workaround.

> What are the use cases we are trying to address with such approach?

Again here I want to highlight that my point with this is not that light-client isn't "good enough" or that it doesn't work or anything, it does. It also isn't about addressing any problems of importing ancient blocks.

My point is basically that if I am a heavy dapp user, and want a good user experience with snappy interaction with a dapp, a light client will not be good enough because it requires network fetches.

My assumption here is that the light client would add a noticeable overhead to interaction with complex dapps, so if I want an optimal dapp experience I will need to run a full node. I haven't benchmarked this at all, I have no real evidence for this assumption, but my gut feel says that light client dapp experience will be sluggish compared to a full node experience.

Thus, it would be nice if we could offer a full-node experience that had the same security parameters as light client but didn't take the 100+GB (soon to be many 100's of GBs) of a regular full node.

-----

With this said, you are right that probably the slowest network request a light client has to make is fetching logs, and if we still have to fetch logs over the network then it's very questionable how much we win. If we can't offer a block-body-pruned mode that also has logs then yeah ¯\\_(ツ)_/¯

tomusdrw:
Perhaps we could store headers + some sort of aggregated receipts for every block? In case of a re-org we could just re-fetch the missing blocks from the network, storing aggregated receipts would allow us to query logs efficiently without storing much data.

folsen:
@tomusdrw I'm very interested in keeping this proposal alive and combining it with @rphmeier's ideas of storing some chunk of blocks based on your nodeid (to not hurt the network too much). That would also make it more useful to the network than a light client, and could potentially even serve light clients, while still making it "easy" to run.

I don't know what a good format for this discussion is though, clearly not a closed PR :P

5chdn:
I closed it because I don't want anyone to have the impression parity is removing the full node ability :)

Let's find a proper place to discuss this. Maybe an #issue?

PR #8653, opened 2018-05-17
Remove NetworkService::config()

tomaka:
More preparation for the libp2p backend.

Once we switch to devp2p and libp2p together, it won't really be possible to query the `NetworkConfiguration` anymore.
This PR therefore removes `config()` in favour of `num_peers_range()`, which is common to both devp2p and libp2p.

### DIFF
--- a/ethcore/sync/src/api.rs
+++ b/ethcore/sync/src/api.rs
@@ -17,6 +17,7 @@
 use std::sync::Arc;
 use std::collections::{HashMap, BTreeMap};
 use std::io;
+use std::ops::Range;
 use std::time::Duration;
 use bytes::Bytes;
 use devp2p::NetworkService;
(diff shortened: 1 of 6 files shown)


ethcore/sync/src/api.rs
```diff
@@ -849,12 +863,13 @@ impl ManageNetwork for LightSync {
...
 		let (connected, active) = self.proto.peer_count();
-		let config = self.network_config();
+		let peers_range = self.num_peers_range();
+		debug_assert!(peers_range.end > peers_range.start);
 		PeerNumbers {
 			connected: connected,
 			active: active,
-			max: config.max_peers as usize,
-			min: config.min_peers as usize,
+			max: peers_range.end as usize - 1,
```
debris: why is the range exclusive and not inclusive? This is a breaking change. If someone have `min: 10` and `max: 10` specified in the config, the new min and max will be 10 and 9.

ethcore/sync/src/api.rs
```diff
@@ -519,8 +527,10 @@ pub trait ManageNetwork : Send + Sync {
 	fn start_network(&self);
 	/// Stop network
 	fn stop_network(&self);
-	/// Query the current configuration of the network
-	fn network_config(&self) -> NetworkConfiguration;
```
debris: can we just remove `struct NetworkConfiguration` entirely?

ethcore/sync/src/api.rs
```diff
@@ -519,8 +527,10 @@ pub trait ManageNetwork : Send + Sync {
...
+	/// Returns the minimum and maximum peers.
+	/// Note that `range.end` is *exclusive*.
+	// TODO: Range should be changed to RangeInclusive once stable (https://github.com/rust-lang/rust/pull/50758)
```
debris: 👍

tomaka:
I went with `Range` because it's more "explicit", but it has the drawback that the `end` is exclusive according to the docs. The correct thing to do would be to use `RangeInclusive`, but it's not stable yet (it's being backported though => https://github.com/rust-lang/rust/pull/50758).

I see multiple possibilities:

- Use `Range` and make `end` exclusive (this PR currently).
- Use `Range` and make `end` inclusive in contradiction to the documentation.
- Use a tuple or our own struct.

dvdplm:
> Wait a month until June 21st for RangeInclusive to be stable in Rust 1.27.

Is the other related work likely to need a while to finish? If so this sounds like the best option to me. Otherwise I'd keep it as-is and loop back and fix it when 1.27 gets out. (I was surprised to see that it's ok to return a `RangeInclusive` in 1.26 but accessing `end` requires unstable. Weirdly half-baked!)

tomaka:
> Is the other related work likely to need a while to finish?

Probably not a lot of time to get something working.

PR #8654, opened 2018-05-18
[WIP] Hybrid Casper FFG Implementation

sorpaas:
cc #7162, https://github.com/djrtwo/EIPs/issues/5. This is a work-in-progress. Do not merge. Reviews/comments/grumbles are highly appreciated.

This implements EIP-1011. Currently we have initialization/deployment, vote transactions validation/sorting, fork choice and finalization implemented. Command line options are not added and there's no block reward issuance modifications for ease of testing. EIP-1011's current specification is not compatible with the old testnet, so only local testing is possible.

Current known issues:

- [ ] Snapshot, warp sync, light client won't work with Casper right now.
- [ ] Miner's vote transaction selection is really inefficient. It probably only works for local vote transactions now. We might need a new struct `TransactionPicker` which picks the next transaction to be included in a block.
- [x] Miner's block reopening won't work with Casper. Because of the introduction of vote transaction ordering, when a new transaction comes, we cannot just reopen the current sealing block, revert to the last transaction's executed state, execute the new transactions and then close the block. Whenever we need to do this, it seems that we need to re-execute at least all of the vote transactions. @djrtwo do you have any suggestions how this might be fixed?


5chdn:
> EIP-1011's current specification is not compatible with the old testnet, so only local testing is possible.

Well, let's start an EIP-1011 testnet then? Happy to set this up. Other clients working on that spec can join at any point.

sorpaas:
@5chdn Need some more polishing and testing, and in particular, we need to have a validator implementation (probably just an external one through JSONRPC endpoints). Otherwise we cannot interact with the Casper contract. Many of the current parameters used here will also likely need to be changed because EIP1011 has not been finalized. But anyway, no matter testing locally, privately or publicly, I think having a testnet is important, and I'll definitely try to get it done soon.

Besides, I think @djrtwo and in particular, the py-evm group is planning to launch another new testnet soon. So we may also consider to join that.

djrtwo:
@5chdn Yeah, the testnet will be pretty boring without validators 😝. There was one written in pyethapp that is now deprecated. Harmony was also implementing one, but it was for the pre-EIP spec. 

Vyper just released some updates that we needed on the python side of things to get the EIP implementation going in py-evm. Trying to get out any implementation next week and start coordinating a multi-client testnet from there.

@sorpaas You should theoretically be able to open up the block, add a non`vote` tx and not re-run the `vote`s. Casper `vote` txs _can_ affect block transaction execution, but normal block transactions _cannot_ affect casper `vote`s. Essentially the post block state is all the normal transactions + limited state modification from the votes. The limited state modification from the votes will be the same regardless of what normal transactions were placed in front of it. I'm not sure how the internals of parity work, but you should be able apply with `vote` state mod to any-prestate without re-running.

djrtwo:
That said, what I am describing is a very non-standard operation and might require some casper-specific dev/constructs. If it helps defining what that limited state is more explicitly, I can do so here and maybe add to EIP

djrtwo:
@sorpaas Also wanted to let you know that the Validator guide is getting a facelift early next week to make sure it is in line with the EIP.

sorpaas:
@djrtwo Unfortunately Parity cannot do that. Let me try to re-describe what you're saying to make sure we are on the same page.

So we have transactions A, B, C, V, where A, B, C are normal transactions, and V is a vote transaction. So now we form a block:

```
A -> B -> C -> V -> post_state
```

`post_state` are stuff like applying block rewards, etc.

Now the block is being mined, but another local normal transaction D pops up. In Parity we would then re-build the above block to make sure local transactions are mined relatively fast. Now consider without Casper, what we do is to revert `post_state`, and then apply D, then only re-apply post_state again, which is a relatively cheap operation. So we have:

```
A -> B -> C -> V -> D -> post_state
```

But with Casper we need to somehow get rid of V first. I think what you mentioned is that we store the executed state of C, then run transaction D on that state, collect all changes without applying it to merkle trie, but instead then apply it to the executed state of V.

This looks like a really complicated thing to do for any current client which apply transactions sequentially (including Parity). For them, it is always the case that changes are applied to merkle trie without storing intermediary changes. If we want to directly do things in parallel on merkle trie, I think it may require some really complicated algorithm beyond my knowledge.

So what I think is possible to do for Parity in this case is only to revert state back to `C`, apply changes of `D`, and then reapply all vote transactions. This, IMHO, would be a really unpleasant disadvantage for sequential transaction client.

I remember cpp-ethereum group mentioned that they're also having similar issues. I wonder given nearly all of our current Ethereum client implementations apply transactions sequentially, would there be any chance that you can lift the transaction ordering requirement in EIP1011?

djrtwo:
A key component of the transaction ordering requirement is so normal transactions and `vote` transactions can be run in parallel. If we do not force normal transactions before `vote` transactions, we lose this advantage.

Sorry, I realized that some pseudocode might have cleared up what I meant above. 
```python
state_N = pre_state_block_N
thread_1:
    tx_post_state = state_N.apply(normal_block_transactions)
thread_2:
    vote_post_state = state_N.apply(vote_transactions)

post_state_block_N = copy(tx_post_state)
post_state_block_N.casper.subsection_of_storage = vote_post_state.casper.subsection_of_storage
```

In this case, if we want to add a new block tx to our already processed block we can just do
```python
tx_post_state.apply(new_tx)
post_state_block_N = copy(tx_post_state)
post_state_block_N.casper.subsection_of_storage = vote_post_state.casper.subsection_of_storage
```

To make this clearer for all clients, I can include this pseudocode and include the exact portions of the casper contract state that `vote` can modify.

Does the above code clear up our intentions? and does it seem feasible to implement in parity?

sorpaas:
@djrtwo Can you explain what you mean by `subsection_of_storage`? Is it okay if we just set `casper.storage_root` to the new value? (Even in that case it would require some "hacks" in Parity TBH.)

I'm trying to understand why this "advantage" of "normal transactions and vote transactions can be run in parallel" is really important. IMHO, in addition to the fact that all clients run sequential transaction applying, I also think this particular optimization is too specific to Casper and is unlikely to benefit/advance other parallel transaction execution research. And actually, even with the ordering restriction lifted, transactions can still run in parallel -- you would just need to hold the really infrequent write lock on `casper.storage_root`, and the performance would just be the same for most cases.

djrtwo:
a portion of the casper storage, mainly related to the `checkpoints` struct can only be written to by `casper` votes. There are other portions of the casper storage that can be modified by normal block transactions (`deposit`, `logout`, `withdraw` tx). What I'm suggesting by `subsection_of_storage` is something like on some subset of variables.
```
post_state_block_N.casper.storage_trie.update("variable") = vote_post_state.casper.storage_trie.read("variable")
```

As I think about it though, it is complex. I will discuss with the research team.

As the EIP is currently written, we charge `vote` transactions zero gas in terms of validator payment and in terms of `gasUsed` in the block. If we sacrifice putting the transactions at the end of the block, we lose the paralellizabiliy and the reduced gas burden on the block gas_limit and validator. I'd be concerned with granting the gas exceptions to `vote`s just assuming that no one is reading from the casper state often. Opens us up for sorts of dos attacks.

sorpaas:
@djrtwo That would actually be more expensive compared with just updating the storage_root, though. If we don't have strong guarantee that casper.storage is not modified by normal transactions, I think we may have issues there.

You can actually keep all other guarantees (`vote_gas_used` reducing gas burden, no failure cases, etc) but only lifting the ordering requirement. IMHO this would be a much better option -- it suffers no more attacks compared with the other version, we save problems for sequential client implementations, and it doesn't sacrifice much paralellizabiliy. I don't think it has any additional DoS attack vectors. Using your version of the pseducode:

```
state_N = pre_state_block_N
normal_block_transactions = collect all normal transactions from block.transactions
vote_transactions = collect all vote transactions from block.transactions
thread_1:
    tx_post_state = state_N.apply(normal_block_transactions)
    trap on read casper.storage:
      if (transaction.index - 1) is vote transaction:
        wait for (transaction.index - 1) to finish on other thread.
thread_2:
    vote_post_state = state_N.apply(vote_transactions)

post_state_block_N = copy(tx_post_state)
post_state_block_N.casper.subsection_of_storage = vote_post_state.casper.subsection_of_storage
```

And my opinion is that this, in addition to save some issues, encourages clients to implement much more useful parallelization that can be potentially used in other cases.

ethcore/src/engines/mod.rs
```diff
@@ -451,6 +463,31 @@ pub trait EthEngine: Engine<::machine::EthereumMachine> {
 		self.machine().verify_transaction_basic(t, header)
 	}
 
+	/// Prepare the environment information passed for transaction execution.
+	fn prepare_env_info(&self, t: &SignedTransaction, block: &ExecutedBlock, env_info: &mut EnvInfo) {
```
rphmeier: seems like all these new functions could be replaced with a more generic one on the main engine trait `fn execution_rules(block: Block) -> Machine::ExecutionRules`. For EthereumMachine this could be an object that provides the ordering rules and transaction outcome.

ethcore/src/ethereum/ethash.rs
```diff
@@ -358,7 +365,32 @@ impl Engine<EthereumMachine> for Arc<Ethash> {
 	}
 
 	fn fork_choice(&self, new: &ExtendedHeader, current: &ExtendedHeader) -> engines::ForkChoice {
-		engines::total_difficulty_fork_choice(new, current)
+		let new_metadata: CasperMetadata = new.metadata().map(|d| rlp::decode(d).expect("Block metadata is valid; qed")).unwrap_or(Default::default());
+		let current_metadata: CasperMetadata = new.metadata().map(|d| rlp::decode(d).expect("Block metadata is valid; qed")).unwrap_or(Default::default());
```
rphmeier: _why_ is it valid? include that in the proof

ethcore/src/machine.rs
```diff
@@ -60,6 +69,39 @@ pub struct EthashExtensions {
 	pub dao_hardfork_beneficiary: Address,
 	/// DAO hard-fork DAO accounts list (L)
 	pub dao_hardfork_accounts: Vec<Address>,
+	/// Number of first block wehre Casper rules begin.
+	pub hybrid_casper_transition: u64,
```
rphmeier: Given that we want to use casper in other deployments, we should not implement it as an ethash extension.

vbuterin:
The simplest solution I can see to this is to disallow logout, withdraw and similar methods during the same blocks (13...49 mod 50 I believe) where voting is possible.

Will think more to see if this could be made more comprehensive and elegant.

vbuterin:
BTW regarding recalculating votes, it's worth noting that signature verifications (and calls to purity-verified addresses in general) can be cached.

vbuterin:
@sorpaas ok, yeah, that works too. And I do like how generic it is.

ethcore/src/machine.rs
niklasad1: s/wehre/where

sorpaas:
To give some updates and contexts for this, mostly from discussions from gitter `ethereum/casper`:

1. The transaction ordering requirement is being removed. So it solved one of the pain point. Thanks @djrtwo @vbuterin!
2. There's a potential point for an attacker to spam a node's tx pools which make new transactions not able to be included for a short period of time. It's still in discussion on gitter for how this might be solved.

ethcore/src/machine.rs
sorpaas: Will handle the refactoring soon, once all the protocol stuff is done. :)

ethcore/src/engines/mod.rs
sorpaas: @rphmeier Can you explain how you vision this to work? We don't need transaction ordering any more so there'll be two more functions -- `prepare_env_info` and `verify_transaction_outcome`. Both of them are actually really Casper-specific (it sets a block metadata value for `vote_gas_used`). Is the intention to create `ExecutionRules` struct to reduce size of `EthereumMachine`? Or would it be better if we just do https://github.com/paritytech/parity/pull/8654#discussion_r189244580 and delegate all Casper-related stuff there?

rphmeier: we may want to use the `Engine` trait in deployments where there are no transactions. Casper may even work in that case because it is sufficiently general. Leaking the `SignedTransaction` and `ExecutedBlock` type into the engine trait makes all prior generalization work irrelevant.

sorpaas:
Okay. I think I've got a simple validator implementation that allows a temporary testnet to work. I think the next step would be to polish the code and address some comments by @rphmeier.

If you want to try this branch:

1. Clone `casper` branch locally, and **merge #8656 to the branch you cloned**, as currently there's a small bug affecting local mining.
2. Get a miner, preferably an older version of ethminer which supports CPU mining.
3. Get [casper-validator](https://github.com/sorpaas/casper-validator), a dummy validator that works by manually submitting votes.
4. Run parity with chain spec at `ethcore/res/ethereum/casper_test.json`. Either create a new account or import the test private key `e75e747d3880d72ff459342a89b025ec59f763524049ba4666cc86f184f48216`.
5. Mine till block 5.
6. In `casper-validator`, run `node index.js`. This will create a validation contract and make a deposit to Casper contract.
7. Mine till block 30.
8. For every 5 blocks (Casper epoch length is set to 5 in the configs for testing), run `node vote.js`. This will submit a voting using the validator index 1 (which we just deployed).
9. Check highest justified epoch and highest finalized epoch should move upward as you submit more votes.
10. Profit!

gaia:
> Snapshot, wrap sync, light client won't work with Casper.

Are there plans to make _full fast_ or _full fatdb trace_ work with Casper for staking
or does the (current implementation of) Casper protocol requirements require an _full archive_ node?

sorpaas:
@gaia Yes, definitely. For staking validators, they would only need to be able to execute calls and submit transactions (signed and unsigned) against the current "best" block. So full node, wrap sync and light client will all work for Casper in the end.

gaia:
Thanks @sorpaas.

I'm planning hardware for the year - would you mind clarifying the following?

As of the current implementation, there are two proposed min deposit amounts:
1. Staking at 1500 ETH deposit
2. Shard Notary at 32 ETH deposit (https://tinyurl.com/y8sskfq6)

What is the least storage parity config for staking 1 (and if possible 2) that is most likely to be working on day 1 of mainnet Casper?

sorpaas:
@gaia For (2) you might need some special softwares to get on because it needs a beacon chain. But for normal Casper as specified in EIP1011, full fast node will surely work.

ethcore/src/engines/mod.rs
sorpaas: @rphmeier Currently the newly defined two functions `prepare_env_info` and `verify_transaction_outcome` is in the `EthEngine` trait. We didn't change `Engine` trait, nor leaked any additional types to `Engine`. So I think it would be okay from this point of view? I still don't have a good idea what the `Machine::ExecutionRules` type would look like.

And regarding Casper being general -- I think it may work in some future versions of Casper, but right now, the Hybrid Casper version is really dependent on executing transactions. The issue is that `simple_casper` being a contract, for many times it "bypass" engine -- functions like `deposit` and `vote` are just normal (signed or unsigned) transactions, so in some sense it's kind of "invisible" to engine.

sorpaas:
cc @5chdn

Just spinned up a testnet using this branch so that we can test out whether the networking and txpool propagation works alright, before we get the testnet from @djrtwo. :smiley: You can also use this to test any validator implementation against Casper revision 2.0. Please PM me if you need any testnet tokens. :)

* Network stats: http://139.162.71.88:3000/
* Validator log: https://seashells.io/v/Qbcqk3EV

Specifications:

* Use [this file](https://gist.github.com/sorpaas/dfa9c3e576ec9d5424a60b429af812c2#file-casper_test-json) and binary from this branch `casper` to connect to the network.
* The [SimpleCasper](https://github.com/paritytech/parity/blob/casper/ethcore/res/code/simple_casper.hex) contract is compiled using [master Vyper](https://github.com/ethereum/vyper/commit/a2ab03882b9e56a58057ed93643d086fc5e093b7) with default settings at tag v0.2.0, and is at address 0x40.
* The [PurityChecker](https://github.com/paritytech/parity/blob/casper/ethcore/res/code/purity_checker.hex) contract is at address
* The [MsgHasher](https://github.com/paritytech/parity/blob/casper/ethcore/res/code/msg_hasher.hex) contract is at address 0x42.
* The [RlpDecoder](https://github.com/paritytech/parity/blob/casper/ethcore/res/code/rlp_decoder.hex) contract is at address 0x43.
* Casper transition happened at block 50, and the warm up period is 50.
* Other parameters are the same as in EIP1011.
* EIP155 is `0x1ee7`.

akx20000a:
@sorpaas @5chdn Chainkorea will deploy the casper testnet based on the balance snpashot of ethereum mainnet. Timeline will be on June 10th 😄

ethcore/src/engines/hybrid_casper.rs
```diff
@@ -0,0 +1,371 @@
...
+		metadata.highest_justified_epoch = self.highest_justified_epoch(caller)?;
+		metadata.highest_finalized_epoch = self.highest_finalized_epoch(caller)?;
+		metadata.highest_finalized_hash = self.checkpoint_hashes(metadata.highest_finalized_epoch, caller)?;
+
+		Ok(())
+	}
+
+	/// Casper-specific fork choice.
+	pub fn fork_choice(&self, new: &ExtendedHeader, current: &ExtendedHeader) -> ForkChoice {
+		let new_metadata: HybridCasperMetadata = new.metadata().map(|d| rlp::decode(d).expect("Metadata is only set by serializing CasperMetadata struct; deserailzling CasperMetadata RLP always succeeds; qed")).unwrap_or(Default::default());
```
afck: `unwrap_or_else(Default::default)` would avoid unnecessarily instantiating `HybridCasperMetadata` (also a few times below).

5chdn:
Putting this on ice as per dev call last week. Are you continuing work on this branch or shall we close it?

sorpaas:
No. Let's close this for now.
I think the beacon chain design would require a completely different set of code.

5chdn:
:cry:

PR #8656, opened 2018-05-18
Fix PoW blockchains sealing notifications in chain_new_blocks

sorpaas:
On master, PoW mining is broken for me. Using any dummy ethash spec and `ethminer`, I would be stuck on mining the same block. `eth` APIs all returns the new best block number except `getWork`, which returns the block to be sealed at the time when I started Parity. This PR fixes this problem.

It looks to be that the issue is when we import a sealed block, for engines that require external sealing, we still need immediate resealing.


debris:
@tomusdrw please review and merge

5chdn:
@tomusdrw could you take a look at this one?

PR #8657, opened 2018-05-18
Remove the Keccak C library and use the pure Rust impl

dvdplm:
When the [recent changes to the Rust implementation of the keccak](https://github.com/debris/tiny-keccak/pull/29) hash lands we can get rid of the one-file C library currently used.

This PR adds a few methods to the `keccak-hash` that were previously calls into C. The current usage patterns in calling code is not very neat/consistent and this PR does nothing to improve on that, prefering to be a pure drop-in change with no concequences for the rest of Parity.

Pending merge of https://github.com/debris/tiny-keccak/pull/29


Vurich:
Looks good, but can you post a benchmark comparison? I think a small amount of slowdown is acceptable if it means getting rid of C but obviously we want to make sure we're not introducing a significant regression.

folsen:
@Vurich see https://github.com/paritytech/keccak-hash/pull/1

dvdplm:
> make sure we're not introducing a significant regression.

@Vurich I'd like to be sure of this too; as @folsen mentioned I ran what's in `ethash` and found no slowdown. If you know of other code that use this stuff a lot I'm happy to dig in and measure the performance impact more.

Vurich:
@dvdplm As far as I know that's the only performance-sensitive place that we use this

util/hash/src/lib.rs
```diff
@@ -30,28 +32,37 @@ pub const KECCAK_NULL_RLP: H256 = H256( [0x56, 0xe8, 0x1f, 0x17, 0x1b, 0xcc, 0x5
...
 
 pub fn keccak<T: AsRef<[u8]>>(s: T) -> H256 {
 	let mut result = [0u8; 32];
 	write_keccak(s, &mut result);
 	H256(result)
 }
 
-pub fn write_keccak<T: AsRef<[u8]>>(s: T, dest: &mut [u8]) {
-	let input = s.as_ref();
+pub fn keccak_256_unchecked(out: *mut u8, outlen: usize, input: *const u8, inputlen: usize) {
```
rphmeier: function should be marked unsafe -- it unconditionally takes the given pointers and assumes they are from valid data.

util/hash/src/lib.rs
```diff
@@ -30,28 +32,37 @@ pub const KECCAK_NULL_RLP: H256 = H256( [0x56, 0xe8, 0x1f, 0x17, 0x1b, 0xcc, 0x5
...
-		// and dest are properly allocated
-		keccak_256(dest.as_mut_ptr(), dest.len(), input.as_ptr(), input.len());
+		Keccak::keccak256(
+			slice::from_raw_parts(input, inputlen),
+			slice::from_raw_parts_mut(out, outlen)
+		);
 	}
 }
 
+pub fn keccak_512_unchecked(out: *mut u8, outlen: usize, input: *const u8, inputlen: usize) {
```
rphmeier: same here.

util/hash/src/lib.rs
```diff
@@ -39,22 +39,18 @@ pub fn keccak<T: AsRef<[u8]>>(s: T) -> H256 {
...
 }
 
-pub fn keccak_256_unchecked(out: *mut u8, outlen: usize, input: *const u8, inputlen: usize) {
-	unsafe {
-		Keccak::keccak256(
-			slice::from_raw_parts(input, inputlen),
-			slice::from_raw_parts_mut(out, outlen)
-		);
-	}
+pub unsafe fn keccak_256_unchecked(out: *mut u8, outlen: usize, input: *const u8, inputlen: usize) {
```
rphmeier: let's document these and their invariants as well.

ethash/src/keccak.rs
```diff
@@ -30,7 +30,9 @@ pub mod keccak_512 {
 	pub fn inplace(input: &mut [u8]) {
 		// This is safe since `sha3_*` uses an internal buffer and copies the result to the output. This
 		// means that we can reuse the input buffer for both input and output.
-		hash::keccak_512_unchecked(input.as_mut_ptr(), input.len(), input.as_ptr(), input.len());
+		unsafe {
+			hash::keccak_512_unchecked(input.as_mut_ptr(), input.len(), input.as_ptr(), input.len());
+		}
```
dvdplm: @rphmeier should these functions also be `pub unsafe …`?

rphmeier: nope. `&mut [T]` has type-system guarantees to point to valid data.

dvdplm:
> this pr updates tens of dependencies, many of which are now duplicated

Totally unintended, a git snafu I believe. Fixed [here](https://github.com/paritytech/parity/pull/8657/commits/421c922f03c33441505710b74ce80c0a22c59390).

PR #8658, opened 2018-05-18
Shutdown the Snapshot Service early

ngotchac:
This PR fixes an issue where the node would not quit if previously downloaded chunks would still be importing by introducing `shutdown` methods to the Client Service and the Snapshot Service.

### DIFF
--- a/parity/run.rs
+++ b/parity/run.rs
@@ -880,7 +880,8 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq: 
 		inner: RunningClientInner::Full {
 			informant,
 			client,
-			keep_alive: Box::new((watcher, service, updater, ws_server, http_server, ipc_server, ui_server, secretstore_key_server, ipfs_server, event_loop)),
+			service: Arc::new(service),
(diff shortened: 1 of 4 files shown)


parity/run.rs
```diff
@@ -902,6 +903,7 @@ enum RunningClientInner {
 	Full {
 		informant: Arc<Informant<FullNodeInformantData>>,
 		client: Arc<Client>,
+		service: Arc<ClientService>,
```
folsen: Can we rename this to something more descriptive, like `snapshot_service`, `service` in itself could be anything. Or is there any particular reason you picked `service`?

ngotchac: It was just because the variable was named `service`, but could change for sure to `client_service`

5chdn:
Does not compile

```
error[E0046]: not all trait items implemented, missing: `shutdown`
  --> ethcore/sync/src/tests/snapshot.rs:74:1
   |
74 | impl SnapshotService for TestSnapshotService {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `shutdown` in implementation
   |
   = note: `shutdown` from trait: `fn(&Self)`
```

PR #8659, opened 2018-05-18
Partially seed Snapshots

ngotchac:
This PR introduces the notion of `Bitfield` for snapshots: which is the currently available chunks of the snapshot that is being downloaded (cf. [here](https://wiki.theory.org/index.php/BitTorrentSpecification#bitfield:_.3Clen.3D0001.2BX.3E.3Cid.3D5.3E.3Cbitfield.3E) for a better explanation that I could provide).

It thus introduces two new packets to ask and respond to a `GetSnapshotBitfield` request.
It then asks peers for chunks that we know are available to him. If the peer doesn't have any more available chunks, his bitfield is updated at most every `BITFIELD_REFRESH_DURATION` second (here 10).


ngotchac:
[putting this one _on progress_ since I'm gone next week, but would be glad to collect comments]

ethcore/sync/src/chain/handler.rs
```diff
@@ -607,9 +645,10 @@ impl SyncHandler {
 		if sync.snapshot.is_complete() {
 			// wait for snapshot restoration process to complete
 			sync.state = SyncState::SnapshotWaiting;
+		} else {
```
folsen: Before it would give the task to the "same peer" _and_ to other peers, right. Now it's either or. Curious as to why this change, I don't have enough context to really understand most of this but it just struck me as something I wanted to know more about.

ethcore/sync/src/chain/mod.rs
```diff
@@ -302,7 +310,7 @@ pub enum ForkConfirmation {
 /// Syncing peer information
 pub struct PeerInfo {
 	/// eth protocol version
-	protocol_version: u8,
+	pub protocol_version: u8,
```
folsen: Why make this `pub`?

ethcore/sync/src/chain/mod.rs
```diff
@@ -649,24 +723,21 @@ impl ChainSync {
...
 				None
 			}
 		).collect();
-		let mut peers: Vec<(PeerId, u8)> = confirmed_peers.iter().filter(|&&(peer_id, _)|
-			self.active_peers.contains(&peer_id)
-		).map(|v| *v).collect();
 
 		random::new().shuffle(&mut peers); //TODO: sort by rating
 		// prefer peers with higher protocol version
 		peers.sort_by(|&(_, ref v1), &(_, ref v2)| v1.cmp(v2));
```
folsen: Don't we do some sort of peer sorting somewhere else as well? For connectivity issues, would it make sense to have a "global" peer sorting function that gets reused, that combines all the rules to find quality peers, like protocol version + connectivity + reputation kind of thing. Or do we literally only care about protocol version at this stage?

ethcore/sync/src/chain/mod.rs
```diff
@@ -698,104 +769,123 @@ impl ChainSync {
...
-							trace!(target: "warp", "Snapshot is stil initializing.");
-							return;
-						},
-						_ => {
-							return;
-						},
-					}
+					},
+					RestorationStatus::Initializing { .. } => {
+						trace!(target: "warp", "Snapshot is stil initializing.");
```
folsen: s/stil/still

ethcore/sync/src/chain/mod.rs
```diff
@@ -1232,25 +1235,11 @@ pub mod tests {
...
-				ask_time: Instant::now(),
-				last_sent_transactions: HashSet::new(),
-				expired: false,
-				confirmation: super::ForkConfirmation::Confirmed,
-				snapshot_number: None,
-				snapshot_hash: None,
-				asking_snapshot_data: None,
-				block_set: None,
-			});
+		let mut peer = PeerInfo::default();
```
folsen: ❤️

folsen:
Seems like a great improvement to me! 👍

ethcore/sync/src/chain/handler.rs
ngotchac: So it's only because if you look at the sync_peer method, it returns immediately if the state is Waiting. This is just to prevent this useless check I guess.

ethcore/sync/src/chain/mod.rs
ngotchac: I'm not quite sure about this. I think that the peers are only sorted here, so it would make sense to have some sort of sorting method. However I don't know if that's the right place, since we might have other useful data somewhere else, for example the first ping time or equivalent...

ethcore/sync/src/chain/handler.rs
```diff
@@ -625,10 +672,12 @@ impl SyncHandler {
 			asking_blocks: Vec::new(),
 			asking_hash: None,
 			ask_time: Instant::now(),
+			ask_bitfield_time: Instant::now() - BITFIELD_REFRESH_DURATION,
```
ordian: Is it safe to subtract `BITFIELD_REFRESH_DURATION` from `Instant::now()` and why is it needed?

ethcore/src/snapshot/service.rs
```diff
@@ -656,18 +689,65 @@ impl Service {
...
 		}
 	}
+
+	/// Get the completed Snapshot Manifest (if any), read from FS
+	fn completed_manifest(&self) -> Option<ManifestData> {
+		self.reader.read().as_ref().map(|r| r.manifest().clone())
+	}
+
+	/// Get the currently partially restored Snapshot Manifest (if any)
+	fn partial_manifest(&self) -> Option<ManifestData> {
```
ordian: `partial_manifest` name is slightly misleading, since it is a snapshot that is partial, not a manifest, if IIUC

rphmeier:
I'm not sure this is actually a good idea, because it will allow for bad snapshots to propagate through honest nodes.

ngotchac:
@rphmeier How about seeding them only after having downloaded the _block chunks_ and verified they are valid?

rphmeier:
I don't know if that's enough because even a single state entry being wrong compromises the whole snapshot. If we are going to take this route I would recommend that we write code for nodes to propagate snapshots even when not doing warp restore by:
  - downloading & verifying all block chunks
  - randomly sampling state data and refusing to propagate if bad

PR #8682, opened 2018-05-22
Fix cli signer

### LINKED ISSUE #8393: CLI signer does not work
The CLI signer does not work, sends invalid requests to RPC. IMHO it's important to get it working since we are separating the UI and there will be no other way to confrim the request.

ascjones:
Fixes #8393.

Problem was that the client wasn't prefixing uints with `0x`. It was using the alternate flag `{:#x}` which is supposed to add that prefix.

Version 0.3.2 of etheruem-types includes a fix to prefix `LowerHex` with `0x` if the alternate flag is used.


rpc/Cargo.toml
```diff
@@ -47,7 +47,7 @@ ethcore-miner = { path = "../miner" }
 ethcore-private-tx = { path = "../ethcore/private-tx" }
 ethcore-sync = { path = "../ethcore/sync" }
 ethcore-transaction = { path = "../ethcore/transaction" }
-ethereum-types = "0.3"
+ethereum-types = "0.3.2"
```
niklasad1: Grumble, if we should keep `ethereum-types = "0.3"` here!

ascjones: We need at least 0.3.2 to include the fix. How about `~0.3.2`?

tomaka: `0.3.2` is the same as `~0.3.2`

tomaka:
Are we sure that changing the serialization format doesn't break other things? :confused:

rpc/Cargo.toml
ascjones: You're right (`>=0.3.2 <0.4`). So it's just narrowing the range to include the version with the fix.

ascjones:
@tomaka the format remains the same. The change is that using the alternate flag with lowerhex `{:#x}` will prefix `uints` with `0x`, as it should for [LowerHex](https://doc.rust-lang.org/std/fmt/trait.LowerHex.html).

Most places are manually adding that prefix with `0x{:x}`. So the only danger would be where the alternate flag is already being used (now the behaviour is fixed), and a search of `parity` for `{:#x}` brings up only the cli signer (which was the cause of the bug) and the rpc uint serialisation wrapper (where I've just replaced `0x{:x}` with `{:#x}` which are equivalent).

PR #8683, opened 2018-05-22
Set the request index to that of the current request

### LINKED ISSUE #8678: Panic on invalid back-reference.
> I'm running:
>
> - **Which Parity version?**: v1.10.3-stable-b9ceda38a-20180507
> - **Which operating system?**: Linux (Arch Linux)
> - **How installed?**: via pacman (local package manager, main repository)

In the process of using Web3 to inspect the Ethereum main chain, I feed the Parity light client an RPC call that crashed it. Specifically, I asked for for the block `0x1784ac` (full hash given below), which is as an uncle of `0x248fe7`. Asking for `0x248fe7` successfully returns the block, including any requests for an arbitrary depth of ancestors.

My naive assessment of the situation is that because I am running a light client, the requested block was not recorded on my machine, leading to a chain of events which caused the client to panic. I propose this with a high degree of confidence, as this event only occurred after I first asked for an uncle block while asking for main-chain blocks continues to work.

I presume that the correct behavior would be to either retrieve the block from the Ethereum network _or_ answer with some standard "no such block" kind of RPC response. What follows is the complete transcript of the event for the Parity client:

```
$ parity --light
2018-05-21 16:10:43  Starting Parity/v1.10.3-stable-b9ceda38a-20180507/x86_64-linux-gnu/rustc1.25.0
2018-05-21 16:10:43  Keys path /home/levi/.local/share/io.parity.ethereum/keys/Foundation
2018-05-21 16:10:43  DB path /home/levi/.local/share/io.parity.ethereum/chains/ethereum/db/906a34e69aec8c0d
2018-05-21 16:10:43  Path to dapps /home/levi/.local/share/io.parity.ethereum/dapps
2018-05-21 16:10:43  Running in experimental Light Client mode.
2018-05-21 16:10:44  Removed existing file '/home/levi/.local/share/io.parity.ethereum/jsonrpc.ipc'.
2018-05-21 16:10:49  Public node URL: [REDACTED]

====================

stack backtrace:
   0:     0x56248f3ed93c - <unknown>
   1:     0x56248f3ec1f2 - <unknown>
   2:     0x56248ea357cb - <unknown>

Thread '<unnamed>' panicked at 'Fails only on invalid back-references; back-references here known to be valid; qed: NoSuchOutput', libcore/result.rs:945

Aborted (core dumped)
```

*Steps to reproduce*:

The following steps reliably allow me to reproduce the above error.

First, start Parity as such:
```
$ parity --light
```

Second, via NodeJS, ask for `0x1784ac`:
```
$ node
> const Web3 = require('web3');
> var web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));
> web3.eth.getBlock('0x1784ac7715ffe762a730f1dc07aca04138f0a028422b89d700e38717de5fa4bb')
```

The result is the following error on Node's end:
```
Error: Invalid JSON RPC response: undefined
    at Object.InvalidResponse (/home/levi/Desktop/test/node.js/node_modules/web3/lib/web3/errors.js:38:16)
    at HttpProvider.send (/home/levi/Desktop/test/node.js/node_modules/web3/lib/web3/httpprovider.js:101:18)
    at RequestManager.send (/home/levi/Desktop/test/node.js/node_modules/web3/lib/web3/requestmanager.js:58:32)
    at Eth.send [as getBlock] (/home/levi/Desktop/test/node.js/node_modules/web3/lib/web3/method.js:145:58)
```

At this time the instance of Parity has crashed with the previously given output.

dvdplm:
When setting up the chain of (two) requests to look up a block by hash, the second need to refer to the first. This fixes an issue where the back ref was set to the subsequent request, not the current one. When the [requests are executed](https://github.com/paritytech/parity/blob/master/ethcore/light/src/on_demand/mod.rs#L313) we loop through them in order and ensure the requests that should produce headers all match up. We do this by index so they better be right or we [panic](https://github.com/paritytech/parity/issues/8678).

In other words: off by one.

closes #8678

### DIFF
--- a/rpc/src/v1/tests/eth.rs
+++ b/rpc/src/v1/tests/eth.rs
@@ -204,6 +204,18 @@ fn eth_get_block() {
 	assert_eq!(tester.handler.handle_request_sync(req_block).unwrap(), res_block);
 }
 
+#[test]
+fn eth_get_block_by_hash() {
+	let chain = extract_chain!("BlockchainTests/bcGasPricerTest/RPC_API_Test");
+	let tester = EthTester::from_chain(&chain);
+
+	// We're looking for block number 4 from "RPC_API_Test_Frontier"
+	let req_block = r#"{"method":"eth_getBlockByHash","params":["0x9c9bdab4cb53fd834e790b13545597f026494d42112e84c0aca9dd6bcc545295",false],"id":1,"jsonrpc":"2.0"}"#;
+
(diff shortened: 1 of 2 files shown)


folsen:
Is it possible to get a test for this somehow? When a function depends on order of execution of the lines it always seems extremely brittle to me.

dvdplm:
> a test for this

Good pt. I'll see what I can do. :)

rpc/src/v1/tests/eth.rs
```diff
@@ -204,6 +204,18 @@ fn eth_get_block() {
 	assert_eq!(tester.handler.handle_request_sync(req_block).unwrap(), res_block);
 }
 
+#[test]
+fn eth_get_block_by_hash() {
```
rphmeier: I'm pretty sure this test doesn't actually invoke the light client code (there aren't RPC tests for the light client RPC)

dvdplm: Indeed it does not (see [this commit](https://github.com/paritytech/parity/pull/8683/commits/19017661ee8e70821abd20d7ffdc33ba6f19e3e6)) but when I discovered that I had already written the test so I pushed it here anyway. I opened [this](https://github.com/paritytech/parity/issues/8693) to address tests for the light client RPC.

PR #8692, opened 2018-05-23
parity: trim whitespace when parsing duration strings

andresilva:


parity/helpers.rs
```diff
@@ -47,11 +47,11 @@ fn to_seconds(s: &str) -> Result<u64, String> {
...
-		x if x.ends_with("seconds") => x[0..x.len() - 7].parse().map_err(bad),
-		x if x.ends_with("minutes") => x[0..x.len() - 7].parse::<u64>().map_err(bad).map(|x| x * 60),
-		x if x.ends_with("hours") => x[0..x.len() - 5].parse::<u64>().map_err(bad).map(|x| x * 60 * 60),
-		x if x.ends_with("days") => x[0..x.len() - 4].parse::<u64>().map_err(bad).map(|x| x * 24 * 60 * 60),
-		x => x.parse().map_err(bad),
+		x if x.ends_with("seconds") => x[0..x.len() - 7].trim().parse().map_err(bad),
+		x if x.ends_with("minutes") => x[0..x.len() - 7].trim().parse::<u64>().map_err(bad).map(|x| x * 60),
+		x if x.ends_with("hours") => x[0..x.len() - 5].trim().parse::<u64>().map_err(bad).map(|x| x * 60 * 60),
+		x if x.ends_with("days") => x[0..x.len() - 4].trim().parse::<u64>().map_err(bad).map(|x| x * 24 * 60 * 60),
+		x => x.trim().parse().map_err(bad),
```
niklasad1: Are we assuming that only a number `"123"` is in seconds?

For `explicitness` I would rather return an error in that case!

andresilva: Since this function converts to seconds, if no unit is provided I think it's ok to assume it is in seconds (this is currently used for parsing CLI parameters).

niklasad1: Ok, seems to only be used by `Tendermint` so let's merge this!

PR #8697, opened 2018-05-23
Implement recursive fmt::Debug for patrica_trie::TrieDB to print Node trees

### LINKED ISSUE #8184: use debug_struct to print trie
discussion: https://github.com/paritytech/parity/pull/8166#discussion_r175972677

gnunicorn:
fixes #8184

### DIFF
--- a/util/patricia_trie/src/triedb.rs
+++ b/util/patricia_trie/src/triedb.rs
@@ -78,64 +78,9 @@ impl<'db> TrieDB<'db> {
 
 	/// Get the data of the root node.
 	fn root_data(&self) -> super::Result<DBValue> {
-		self.db.get(self.root).ok_or_else(|| Box::new(TrieError::InvalidStateRoot(*self.root)))
-	}
-
-	/// Indentation helper for `format_all`.
-	fn fmt_indent(&self, f: &mut fmt::Formatter, size: usize) -> fmt::Result {
-		for _ in 0..size {
-			write!(f, "  ")?;
-		}
-		Ok(())
-	}
-
-	/// Recursion helper for implementation of formatting trait.
-	fn fmt_all(&self, node: Node, f: &mut fmt::Formatter, deepness: usize) -> fmt::Result {
-		match node {
-			Node::Leaf(slice, value) => writeln!(f, "'{:?}: {:?}.", slice, value.pretty())?,
-			Node::Extension(ref slice, ref item) => {
-				write!(f, "'{:?} ", slice)?;
-				if let Ok(node) = self.get_raw_or_lookup(&*item) {
-					match Node::decoded(&node) {
-						Ok(n) => self.fmt_all(n, f, deepness)?,
-						Err(err) => writeln!(f, "ERROR decoding node extension Rlp: {}", err)?,
-					}
-				}
-			},
-			Node::Branch(ref nodes, ref value) => {
-				writeln!(f, "")?;
-				if let Some(ref v) = *value {
-					self.fmt_indent(f, deepness + 1)?;
-					writeln!(f, "=: {:?}", v.pretty())?
-				}
-				for i in 0..16 {
-					let node = self.get_raw_or_lookup(&*nodes[i]);
-					match node.as_ref() { 
-						Ok(n) => {
-							match Node::decoded(&*n) {
-								Ok(Node::Empty) => {},
-								Ok(n) => {
-									self.fmt_indent(f, deepness + 1)?;
-									write!(f, "'{:x} ", i)?;
-									self.fmt_all(n, f, deepness + 1)?;
-								}
-								Err(e) => {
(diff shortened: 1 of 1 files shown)


gnunicorn:
ah. I didn't see this part. I had run `cargo fmt` in between and it had changed... _a lot_ . I can revert that part, too, if you want  @dvdplm . I don't really have a preference...

util/patricia_trie/src/node.rs
```diff
@@ -107,13 +108,43 @@ impl<'a> Node<'a> {
...
+				let item = item.fmt(f)?;
+				f.debug_struct("Node::Extension")
+					.field("slice", &slice)
+					.field("item", &item)
+					.finish()
+			}
+			Node::Branch(ref nodes, ref value) => {
+				let nodes_fmt = f.debug_list().entries(nodes).finish()?;
+				f.debug_struct("Node::Branch")
+					.field("nodes", &nodes_fmt)
```
debris: I think this is invalid, cause `nodes_fmt` is `()`.? [docs](https://doc.rust-lang.org/std/fmt/struct.DebugList.html#method.finish)

util/patricia_trie/src/node.rs
```diff
@@ -107,13 +108,43 @@ impl<'a> Node<'a> {
...
+		match &self {
+			Node::Leaf(slice, value) => f.debug_struct("Node::Leaf")
+				.field("slice", &slice)
+				.field("value", &value.pretty())
+				.finish(),
+			Node::Extension(ref slice, ref item) => {
+				let item = item.fmt(f)?;
+				f.debug_struct("Node::Extension")
+					.field("slice", &slice)
+					.field("item", &item)
```
debris: I think this is invalid, cause `item` is `()`.

util/patricia_trie/src/node.rs
```diff
@@ -107,13 +108,43 @@ impl<'a> Node<'a> {
...
 	}
 }
 
+impl<'a> fmt::Debug for Node<'a> {
+	/// Recursion helper for implementation of formatting trait.
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		match &self {
+			Node::Leaf(slice, value) => f.debug_struct("Node::Leaf")
+				.field("slice", &slice)
+				.field("value", &value.pretty())
```
debris: is it possible to avoid calling `.pretty()`? :)

gnunicorn: Yes, but that gives a different value: `00·01` (on pretty) vs `[0, 1]`. I am fine with the latter, but just to make sure that is okay.

gnunicorn:
@debris just come to realize that you are right, for once because of the reasons you describe (formatter isn't really meant to be used recursively and the way I thought it worked was ... unsequential), but more importantly: the errors that were previously possible in the formatter came from the lookup. My current approach isn't actually doing the look up but just prints the addresses. Thus we are not seeing any tree actually.

Let me get back to this.

util/patricia_trie/src/node.rs
```diff
@@ -177,3 +171,142 @@ impl<'a> From<Node<'a>> for OwnedNode {
...
+	use super::Node;
+	use nibbleslice::NibbleSlice;
+	static D: &'static [u8;3] = &[0x01u8, 0x23, 0x45];
+
+	#[test]
+	fn test_node_default_debug() {
+		let empty = Node::Empty;
+
+		assert_eq!(format!("{:?}", empty), "Node::Empty");
+        assert_eq!(format!("{:#?}", empty), "Node::Empty");
```
debris: bad indentation ;)

gnunicorn:
@debris yeah. I tried to push it into Node first (handing over a closure) and that might have worked but wouldn't actually take into account the right ident.

So, I moved the entire implementation into its own `TrieAwareDebugNode` that handles lookup and all nicely and added a test on the triedb-side of things.

But as I had to remove all I did before, I decided to rebase it entirely. I hope that is okay.

util/patricia_trie/src/triedb.rs
```diff
@@ -174,15 +119,55 @@ impl<'db> Trie for TrieDB<'db> {
...
+				Ok(Node::Leaf(slice, value)) => f.debug_struct("Node::Leaf")
+						.field("slice", &slice)
+						.field("value", &value)
+					.finish(),
+				Ok(Node::Extension(ref slice, ref item)) => f.debug_struct("Node::Extension")
+						.field("slice", &slice)
+						.field("item", &TrieAwareDebugNode{trie: self.trie, key: item})
+					.finish(),
+				Ok(Node::Branch(ref nodes, ref value)) => f.debug_struct("Node::Branch")
+						.field("nodes", &nodes)
```
gnunicorn: hold on a second, this still doesn't convert the entries.

PR #8710, opened 2018-05-24
ethcore-sync: fix connection to peers behind chain fork block

andresilva:
We're currently not connecting to any peers that are behind the chain's fork block (#1920000 for foundation). This is a regression introduced recently in #8543.

This PR reverts to the previous behavior, where we still accept connections from/to an unconfirmed peer. This behavior is guarded against by `PeerInfo#can_sync()` where we'll make sure that the given peer has confirmed the fork header before we try to sync from him.

### DIFF
--- a/ethcore/sync/src/chain/handler.rs
+++ b/ethcore/sync/src/chain/handler.rs
@@ -332,14 +332,6 @@ impl SyncHandler {
 		Ok(())
 	}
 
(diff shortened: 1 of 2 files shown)


folsen:
This is an important fix to get in asap, @ngotchac and @tomusdrw could you please prioritise reviewing?

Issue #8552, opened 2018-05-07
Blooms recomputation in case of a fork is very slow


usd-yamazaki:
> I'm running:
>
> - **Which Parity version?**: Parity/v1.9.5-stable-ff821da-20180321/x86_64-linux-gnu/rustc1.24.1
> - **Which operating system?**: Ubuntu 16.04.2 LTS
> - **How installed?**: via installer
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: private PoA (Aura)
> - **Did you try to restart the node?**: yes

### Node list
- AU1: Sealing node
- AU2: Sealing node
- OB1: eth_sendRawTransaction execution node (not seal)
- OB2: ordinary node (not seal)

I am sending externally created transactions to private PoA chain with eth_sendRawTransaction.
Send 6400 transactions from one account at a time and wait for synchronization.
Each transaction uses 1649 bytes and uses approximately 1.15 M gas.
If a temporary fork occurs during this sync, subsequent synchronization will be extremely slow.
In normal case, it takes about 7 minutes for all transactions to be processed.
If a temporary fork occurs, it takes about 30 minutes for all transactions to be processed.
During this time, the CPU usage rate will be 100%, but memory and storage are calm.

In this log, a temporary fork occurs between AU1 and AU2 at #1638747, and long time synchronization of OB1 starts at that timing.
At this time, there were about 3700 pending transactions for OB1.

*AU1*
```
2018-05-02 20:17:20  Imported #1638744 fa14…7c6e (2 txs, 2.37 Mgas, 100.78 ms, 3.79 KiB)
2018-05-02 20:17:26  Imported #1638745 9e7e…bdc9 (304 txs, 352.14 Mgas, 584.05 ms, 489.75 KiB)
2018-05-02 20:17:30  Imported #1638746 54d2…aa53 (0 txs, 0.00 Mgas, 4.88 ms, 0.57 KiB)
2018-05-02 20:17:42  Imported #1638747 6f11…18b0 (5 txs, 5.79 Mgas, 2354.12 ms, 8.63 KiB)
2018-05-02 20:17:45  Imported #1638748 aeaf…5edd (3 txs, 3.48 Mgas, 9.47 ms, 5.41 KiB) + another 1 block(s) containing 1170 tx(s)
2018-05-02 20:17:50  Imported #1638749 08b1…b7d0 (0 txs, 0.00 Mgas, 4.80 ms, 0.57 KiB)
[...]

PR #8712, opened 2018-05-25
new blooms database

debris:
This pr introduces new database for header blooms.

#### Why?
The only time when we access blooms is when we are filtering logs. Filtering logs requires iteration over all blooms in a database and this is very inefficient with RocksDB. As blockchain got bigger and bigger it took more and more time to effectively filter all blooms starting from the genesis block.

#### How?
I built increment only, sequential database for storing blooms. It consists of three database layers. `top`, `mid`, `bot`.

- `bot` layer contains all header blooms
- `mid` layer contains blooms created from 16 consecutive blooms of the `bot` layer. 
- `top` layer contains blooms from 16 consecutive blooms from the `mid` layer. 

Blooms at each layer are placed next to each other on a disk. So it's very efficient to find `O(1)~` and iterate `O(1)~` over them. It's a huge improvement compared to the previous database implementation where every lookup and step of iteration was probably `O(log n)`.

size of a database (for 1000000 blocks)
- `bot` layer - *256mb* (1000000 * size_of_bloom (256b))
- `mid` layer - *16mb* (`bot` / 16)
- `top` layer - *1mb* (`mid` /  16)

#### Other issues addressed
- #8552 - this pr also addresses #8552 by simply **not recomputing** blooms at higher levels. Why? Because in case of a fork, it's likely that the same transaction on a different chain will be included in block of the same number or in one of the following blocks. As long as the distance between block on chain a, and block on chain b, is on average significantly lower than 128, false-positive ration of bloom filter should not increase. And even if the `top` and `mid` bloom return the false-positive, the underlying bottom bloom is replaced with a new one, so this will not lead to incorrect results.

  https://github.com/paritytech/parity/blob/343b29866c4cc592afa9038995217b6ba99e70fa/util/blooms-db/src/db.rs#L96-L98

- this pr also addresses all issues with bloom recomputation on forks. Recomputation does not happen and there is no in-memory cache, so blooms are always valid

#### todo list (in this pr)
- [x] new blooms-db
- [x] new database interface for parity (aka `BlockChainDB`)
- [x] integrate new db interface with parity and tests (😓 )
- [x] `BlockChainDB` should not expose `&RwLock` in the interface
- [x] fix and tests for blooms-db flushing
- [x] migrate traces to the new blooms-db
- [x] remove old `bloomchain` crate
- [x] filter multiple bloom possibilities at the same time
- [x] restore blooms after snapshot
- [x] write database migration script

#### benchmarks

- So far I run only a single benchmark for blooms-db. The results are promising, but I believe that they do not fully show how much everything has improved.

  benchmark from old [stackoverflow thread](https://ethereum.stackexchange.com/questions/2190/is-searching-data-stored-in-event-logs-prohibitively-slow/2195#2195)

  ```
  time curl -X POST --data '{"id":8,"jsonrpc":"2.0","method":"eth_getLogs","params":[{"fromBlock":"0x0","toBlock":"0xf0be2", "address": "0x33990122638b9132ca29c723bdf037f1a891a70c"}]}' -H "Content-Type: application/json" http://127.0.0.1:8545 >> /dev/null
  ```

  old parity:

  ```
  0.77s
  ```

  this branch:

  ```
  0.01s
  ```

---

This pr, should also reduce disk writes by roughly **70gb** when doing full sync

```
3 * 5670000 * 16 * 256 bytes = 69.67296 gigabytes
```

- 3 - bloom levels
- 5670000 - number of blocks
- 16 - currently blooms are stored in groups containing 16 blooms
- 256 - size of a bloom


ethcore/src/blockchain/blockchain.rs
```diff
@@ -1459,8 +1459,7 @@ mod tests {
 	use std::sync::Arc;
 	use rustc_hex::FromHex;
 	use hash::keccak;
-	use kvdb::{KeyValueDB, DBTransaction};
-	use kvdb_memorydb;
+	use kvdb::{DBTransaction};
```
niklasad1: Remove blocks around a single `import`

ethcore/src/snapshot/tests/helpers.rs
```diff
@@ -33,7 +33,7 @@ use snapshot::io::{SnapshotReader, PackedWriter, PackedReader};
 use tempdir::TempDir;
 use rand::Rng;
 
-use kvdb::{KeyValueDB, DBValue};
+use kvdb::{DBValue};
```
niklasad1: Remove blocks around a single import

ethcore/src/snapshot/tests/proof_of_work.rs
```diff
@@ -28,8 +27,7 @@ use snapshot::io::{PackedReader, PackedWriter, SnapshotReader, SnapshotWriter};
 
 use parking_lot::Mutex;
 use snappy;
-use kvdb::{KeyValueDB, DBTransaction};
-use kvdb_memorydb;
+use kvdb::{DBTransaction};
```
niklasad1: Remove blocks around a single import

ethcore/src/snapshot/tests/service.rs
```diff
@@ -24,11 +24,11 @@ use ids::BlockId;
 use snapshot::service::{Service, ServiceParams};
 use snapshot::{self, ManifestData, SnapshotService};
 use spec::Spec;
-use test_helpers::{self, generate_dummy_client_with_spec_and_data};
+use test_helpers::{generate_dummy_client_with_spec_and_data};
```
niklasad1: Remove blocks around a single import

ethcore/src/snapshot/tests/service.rs
```diff
@@ -24,11 +24,11 @@ use ids::BlockId;
...
 use test_helpers_internal::restoration_db_handler;
 
 use io::IoChannel;
-use kvdb_rocksdb::{Database, DatabaseConfig};
+use kvdb_rocksdb::{DatabaseConfig};
```
niklasad1: Remove blocks around a single import

ethcore/src/trace/db.rs
```diff
@@ -108,18 +104,15 @@ impl<T> TraceDB<T> where T: DatabaseExtras {
 		let current_size = self.cache_size();
 
 		let mut traces = self.traces.write();
-		//let mut blooms = self.blooms.write();
 		let mut cache_manager = self.cache_manager.write();
 
 		cache_manager.collect_garbage(current_size, | ids | {
 			for id in &ids {
 				match *id {
```
niklasad1: Use `if let` instead of match of a single pattern?

debris:
@niklasad1 this pr is still in progress and quite far away from being ready for a review :)

@tjayrush 

> 1. Won't the higher level blooms, given that they are the same width and are getting filled with more and more data, get quickly saturated? 

They will, especially top level blooms. But the cost of checking them is so low, that it doesn't matter.

Currently we use exactly the same bloom filters with three layers for filtering events. The bottleneck is rocksdb io, and this pr is trying to address only that problem. :)

> Did you think about widening the mid-layer, top-layer blooms so as to keep a more consistent false-positive rate?

Yes, that's a possible improvement.

> How hard would it be to enhance the blooms to note every single address in a block (as opposed to the current list which, I think, doesn't include every address). The reason I'm asking is because I want to use the bloom to answer more questions than just "Is there an event on my address in this block."

You can use transaction traces exactly for that. They already utilize the same bloom filters.

tjayrush:
@debris 

> You can use transaction traces exactly for that. They already utilize the same bloom filters.

The reason I'm suggesting adding every address to the bloom is to avoid having to request traces. If one is trying to do a full accounting of a particular address (or group of addresses), one needs to visit every trace, but if the blooms included all addresses involved in a block, one could use them to avoid a huge number of trace requests. Requesting traces (especially across the 2016 dDos transactions) is brutal. Did you have a chance to glance at the paper I mentioned above? With that method, I can collect full lists of transactions per account about 100 times faster than reading every trace. The current blooms don't include every address, so they miss a lot of blocks if one wants a full list of transactions.

debris:
I finally came back to this pr and finished it. 

@tjayrush yes, I read the paper. Using adaptive bloom filters to note every single address is very interesting. We should consider adding it to parity

ethcore/src/test_helpers.rs
```diff
@@ -255,16 +262,97 @@ pub fn get_test_client_with_blocks(blocks: Vec<Bytes>) -> Arc<Client> {
...
+		fn trace_blooms(&self) -> &blooms_db::Database {
+			&self.trace_blooms
+		}
+	}
+
+	impl BlockChainDBHandler for RestorationDBHandler {
+		fn open(&self, db_path: &Path) -> Result<Arc<BlockChainDB>, Error> {
+			let key_value = Arc::new(kvdb_rocksdb::Database::open(&self.config, &db_path.to_string_lossy())?);
+			let blooms_path = db_path.join("blooms");
+			let trace_blooms_path = db_path.join("trace_blooms");
```
debris: I'm not sure if this is a good location for blooms db. I'm open for any other suggestions

ethcore/src/client/client.rs
```diff
@@ -1293,10 +1294,11 @@ impl snapshot::DatabaseRestore for Client {
 		let mut tracedb = self.tracedb.write();
 		self.importer.miner.clear();
 		let db = self.db.write();
-		db.restore(new_db)?;
+		db.key_value().restore(new_db)?;
```
debris: because blooms-db is placed in the same directory, it also moves it. Maybe we should move this function out of `KeyValueDB` interface to make it more descriptive

ethcore/src/client/client.rs
```diff
@@ -1293,10 +1294,11 @@ impl snapshot::DatabaseRestore for Client {
...
 
+		// TODO: restore blooms properly
```
debris: ah, I forgot to reopen the blooms db, will fix it

parity/db/rocksdb/blooms.rs
```diff
@@ -0,0 +1,84 @@
...
+	let db = open_database(&path.as_ref().to_string_lossy(), config)?;
+
+	// possible optimization:
+	// pre-allocate space on disk for faster migration
+
+	// iterate over header blooms and insert them in blooms-db
+	// Some(3) -> COL_EXTRA
+	// 3u8 -> ExtrasIndex::BlocksBlooms
+	// 0u8 -> level 0
+	let blooms_iterator = db.key_value()
```
debris: from `ethcore/src/blockchain/extras.rs`

parity/db/rocksdb/blooms.rs
```diff
@@ -0,0 +1,84 @@
...
+
+	for (number, blooms) in blooms_iterator {
+		db.blooms().insert_blooms(number, blooms.iter())?;
+	}
+
+	// iterate over trace blooms and insert them in blooms-db
+	// Some(4) -> COL_TRACE
+	// 1u8 -> TraceDBIndex::BloomGroups
+	// 0u8 -> level 0
+	let trace_blooms_iterator = db.key_value()
```
debris: from `ethcore/src/trace/db.rs`

parity/db/rocksdb/blooms.rs
```diff
@@ -0,0 +1,84 @@
...
+				(key[3] as u64) << 16 |
+				(key[4] as u64) << 8 |
+				(key[5] as u64);
+
+			let blooms = rlp::decode_list::<Bloom>(&group);
+			(number, blooms)
+		});
+
+	for (number, blooms) in blooms_iterator {
+		db.blooms().insert_blooms(number, blooms.iter())?;
```
sorpaas: `insert_blooms` is not atomic. If any of `top`, `mid`, or `bot` file write fails, the bloomdb basically enters a corrupt status -- we will have top-level blooms saying something exists but cannot be found in `mid` or `bot` (true negative). So I think it may be better to just panic `expect` here like we do in `blockchain.rs`?

util/blooms-db/src/db.rs
```diff
@@ -0,0 +1,277 @@
...
+	/// Insert consecutive blooms into database starting with positon from.
+	pub fn insert_blooms<'a, I, B>(&mut self, from: u64, blooms: I) -> io::Result<()>
+	where ethbloom::BloomRef<'a>: From<B>, I: Iterator<Item = B> {
+		for (index, bloom) in (from..).into_iter().zip(blooms.map(Into::into)) {
+			let pos = Positions::from_index(index);
+
+			// constant forks make lead to increased ration of false positives in bloom filters
+			// since we do not rebuild top or mid level, but we should not be worried about that
+			// most of the time events at block n(a) occur also on block n(b) or n+1(b)
+			self.top.accrue_bloom::<ethbloom::BloomRef>(pos.top, bloom)?;
```
sorpaas: Regarding possible bloomdb true negative corrupt situation, given that we only insert one bloom at a time, I think a simple solution maybe to create a metadata file, storing the current writing bloom position before attempting the insertion. Then, if we find any error case in below, next time when the program starts, we would only need to rewrite that particular bloom position to fix the corruption.

This doesn't solve the disk consistency issue. Otherwise we will need to `flush` within the for loop, which may be a bad performance penalty. So not sure whether we would want this.

util/blooms-db/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::path::Path;
+use parking_lot::Mutex;
+
+/// Threadsafe API for blooms database.
+///
+/// # Warning
+///
+/// This database does not guarantee atomic writes.
+pub struct Database {
+	database: Mutex<db::Database>,
```
sorpaas: Any reasons we use `Mutex` but not `RwLock`? I think it would be safe for multiple parties to read blooms at the same time?

parity/db/rocksdb/blooms.rs
debris: I don't think it matters. even if it ends up in the corrupt state. On next parity launch the migration will start from scratch. Corrupted data will be overwritten and database state will be consistent

util/blooms-db/src/db.rs
debris: > Then, if we find any error case in below, next time when the program starts, we would only need to rewrite that particular bloom position to fix the corruption.

That's what I did initially. Then I thought about the problem again and realised that all of this does not matter. Existence of blooms is not consensus critical. If one of the writes fail (or partially fail), this particular block will be reimported on next launch of parity and the database will overwrite corrupted data

util/blooms-db/src/lib.rs
debris: Great question! There is a reason, https://users.rust-lang.org/t/how-to-handle-match-with-irrelevant-ok--/6291/15

debris: tl;dr

> Reading a file is mutating the object that represent the file stream, namely altering the position. The very data being mutated is stored in the kernel space, and simultaneous mutation won’t cause the kernel any trouble because it is smart enough to synchronize accesses.
Still, two threads simultaneously reading the same file stream will break internal logic. (I guess even a single read() call on BufReader could yield a non-contiguous slice).

debris: That's why I made iterator function of `db::Database` `mut`.

sorpaas: Ah got it!

5chdn:
Please add a :heavy_check_mark: tick if you think this looks good.

ethcore/src/client/client.rs
niklasad1: @debris Fixed? If so remove the comment :wave:

ethcore/src/snapshot/service.rs
```diff
@@ -622,7 +621,8 @@ impl Service {
 
 							match is_done {
 								true => {
-									db.flush().map_err(UtilError::from)?;
+									// TODO: flush also blooms?
```
niklasad1: As I understand the `API` doesn't provide `flushing` but it does `flush` everytime new blooms are inserted?!

If so remove the comment alternatively change the API!

ethcore/src/trace/db.rs
```diff
@@ -52,144 +47,74 @@ impl Key<FlatBlockTraces> for H256 {
...
+	/// cache
 	traces: RwLock<HashMap<H256, FlatBlockTraces>>,
-	blooms: RwLock<HashMap<TraceGroupPosition, blooms::BloomGroup>>,
-	cache_manager: RwLock<CacheManager<CacheId>>,
-	// db
-	tracesdb: Arc<KeyValueDB>,
-	// config,
-	bloom_config: BloomConfig,
-	// tracing enabled
+	cache_manager: RwLock<CacheManager<H256>>,
```
niklasad1: Missing docs for `cache_manager`

ethcore/src/trace/db.rs
```diff
@@ -286,19 +209,8 @@ impl<T> TraceDatabase for TraceDB<T> where T: DatabaseExtras {
...
-				.collect::<HashMap<TraceGroupPosition, blooms::BloomGroup>>();
-
-			let blooms_keys: Vec<_> = blooms_to_insert.keys().cloned().collect();
-			let mut blooms = self.blooms.write();
-			batch.extend_with_cache(db::COL_TRACE, &mut *blooms, blooms_to_insert, CacheUpdatePolicy::Remove);
-			// note_used must be called after locking blooms to avoid cache/traces deadlock on garbage collection
-			for key in blooms_keys {
-				self.note_used(CacheId::Bloom(key));
-			}
+			// TODO: replace it with database for trace blooms
```
niklasad1: Is this supposed to be fixed in this `PR`?

ethcore/src/trace/db.rs
```diff
@@ -396,8 +308,10 @@ impl<T> TraceDatabase for TraceDB<T> where T: DatabaseExtras {
 	}
 
 	fn filter(&self, filter: &Filter) -> Vec<LocalizedTrace> {
-		let chain = BloomGroupChain::new(self.bloom_config, self);
-		let numbers = chain.filter(filter);
+		let possibilities = filter.bloom_possibilities();
+		let numbers = self.db.trace_blooms()
+			.filter(filter.range.start as u64, filter.range.end as u64, &possibilities).expect("TODO: blooms pr");
```
niklasad1: If this is fixed change the `expect` message!

parity/db/rocksdb/blooms.rs
```diff
@@ -0,0 +1,84 @@
...
+			let number =
+				(key[2] as u64) |
+				(key[3] as u64) << 8 |
+				(key[4] as u64) << 16 |
+				(key[5] as u64) << 24;
+
+			let blooms = rlp::decode_list::<Bloom>(&group);
+			(number, blooms)
+		});
+
```
niklasad1: This code is `identical` except the variables to #L38-#L52 consider moving it to a function instead?

util/blooms-db/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "blooms-db"
+version = "0.1.0"
+authors = ["debris <marek.kotewicz@gmail.com>"]
```
niklasad1: This should be `admin@parity.io`?

ethcore/src/snapshot/service.rs
debris: that's an obsolete comment ;)
Yes, this db doesn't provide flushing, cause writes are nonatomic and happen immediately when someone calls insert

ethcore/src/trace/db.rs
debris: ahh... it's not a public field and no change has been made here :p but I'll add a description ;p

ethcore/src/trace/db.rs
debris: oh, yes, obsolete comment 😅

parity/db/rocksdb/blooms.rs
debris: it's not exactly the same, notice that endianness is completely different. also the function would have 4 variables and imo would be less readable than what we have here. As this is a migration, this code will never be modified again

util/blooms-db/Cargo.toml
debris: 😅

parity/db/rocksdb/blooms.rs
niklasad1: Ok, makes sense then NVM!

util/blooms-db/Cargo.toml
niklasad1: Forgot one thing,  also add a license I guess it should be: `license = "GPL-3.0"`

debris: I believe it doesn't matter for subcrates :)

ethcore/src/client/client.rs
debris: not yet, that's the only remaining thing :)

util/blooms-db/Cargo.toml
debris: actually, you are right. We usually have a `license` field so I'll also add it :)

ethcore/src/client/client.rs
debris: done!

debris:
@niklasad1 @sorpaas I addressed grumbles. Can you review the pr again and mark as accepted (if it's good) :)

debris:
also, I merged `fs-swap` branch into this one. Thought that I'll need it, but didn't need it eventually. If it's a problem, I'll remove those prs

PR #8715, opened 2018-05-25
Private packets verification and queue refactoring

### LINKED ISSUE #8250: Add private transactions validation before propagating
In the first implementation all private transaction messages propagated blindly. Some basic validation needs to be added in order not to spam the network with not already valid transactions.

grbIzl:
This PR addresses two major TODOs from private transactions PR
- Verification of private transactions packets and prevention of packets resending (closes #8250 )
- Refactoring of queue's implementation in order to switch to direct usage of new tx queue


ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,128 +15,238 @@
...
+
+	fn sender(&self) -> &Address {
+		&self.transaction_sender
+	}
+}
+
+#[derive(Debug)]
+pub struct PrivateScorying;
+
+impl txpool::Scoring<VerifiedPrivateTransaction> for PrivateScorying {
```
grbIzl: @tomusdrw I reworked my queue's implementation in order to use txpool directly for storing transaction in it. The main problem, that I'm facing, is scoring implementation. Actually I'd like to fully re-use miner's scoring, but I don't know, how to do it, because it's bound with miner's VerifiedTransaction trait.

grbIzl: So as result I had to copy-past scoring code from miner.

Tbaut:
@grbIzl we should update the documentation for that one right?

grbIzl:
@Tbaut i don't think so. It's mostly the internal improvements and refactorings

ethcore/private-tx/src/lib.rs
```diff
@@ -256,48 +259,109 @@ impl Provider where {
...
-			verification_queue.remove_private_transaction(&transaction_hash);
 		}
 		Ok(())
 	}
 
+	/// Add signed private transaction into the store	/// Creates corresponding public transaction if last required signature collected and sends it to the chain
+	pub fn process_signatures(&self) -> Result<(), Error> {
+		trace!("Processing signed private transactions");
+		let mut signed_transactions = self.imported_signed_transactions.lock();
+		for tx in signed_transactions.drain(..) {
```
tomusdrw: Again error handling here:
We iterate over a list of transactions, but any error inside here (or `return` statement) will cause the draining iterator to be dropped and removing all elements anyway.

If the assumption is that there is always only one transaction then let's make it explicit and pass the transaction as function parameter as suggested in a comment above.

ethcore/private-tx/src/lib.rs
```diff
@@ -324,13 +388,13 @@ impl Provider where {
 	}
 
 	/// Broadcast the private transaction message to the chain
-	fn broadcast_private_transaction(&self, message: Bytes) {
-		self.notify(|notify| notify.broadcast(ChainMessageType::PrivateTransaction(message.clone())));
+	fn broadcast_private_transaction(&self, transaction_hash: H256, message: Bytes) {
+		self.notify(|notify| notify.broadcast(ChainMessageType::PrivateTransaction(transaction_hash.clone(), message.clone())));
```
tomusdrw: no need to `.clone()` the hash, it's `Copy` anyway

ethcore/private-tx/src/lib.rs
```diff
@@ -567,105 +631,44 @@ impl Importer for Arc<Provider> {
...
-				self.transactions_for_verification.lock().add_transaction(
-					original_tx,
-					contract,
-					validation_account,
-					hash,
-					self.pool_client(&nonce_cache),
-				)?;
-				let provider = Arc::downgrade(self);
-				self.channel.send(ClientIoMessage::execute(move |_| {
+		let original_tx = self.extract_original_transaction(private_tx.clone(), &contract)?;
```
tomusdrw: Why do we need to clone here and then in `add_transaction`? It seems pretty expensive.

ethcore/private-tx/src/lib.rs
```diff
@@ -567,105 +631,44 @@ impl Importer for Arc<Provider> {
...
+			original_tx,
+			validation_account.map(|&account| account),
+			private_tx.clone(),
+			self.pool_client(&nonce_cache),
+		)?;
+		let provider = Arc::downgrade(self);
+		if let Err(e) = self.channel.send(ClientIoMessage::execute(move |_| {
 					if let Some(provider) = provider.upgrade() {
-						if let Err(e) = provider.process_queue() {
+						if let Err(e) = provider.process_verification_queue() {
```
tomusdrw: Same as with `process_signatures` might be easier to just offload everything inside the `IoWorker` thread and just `move` the `private_tx` here and then import and process it. Currently we are almost always processing only one transaction, but the whole code assumes we are operating on lists.

ethcore/private-tx/src/lib.rs
```diff
@@ -683,7 +686,7 @@ impl ChainNotify for Provider {
 	fn new_blocks(&self, imported: Vec<H256>, _invalid: Vec<H256>, _route: ChainRoute, _sealed: Vec<H256>, _proposed: Vec<Bytes>, _duration: Duration) {
 		if !imported.is_empty() {
 			trace!("New blocks imported, try to prune the queue");
-			if let Err(err) = self.process_queue() {
+			if let Err(err) = self.process_verification_queue() {
```
tomusdrw: Ok, so it seems some transactions will become ready after a block, right? Makes sense to keep the `process_verification_queue` as a separate function then, but we need to fix the error handling there.

ethcore/private-tx/src/messages.rs
```diff
@@ -73,4 +73,9 @@ impl SignedPrivateTransaction {
 	pub fn private_transaction_hash(&self) -> H256 {
 		self.private_transaction_hash
 	}
+
+	/// Own hash
+	pub fn hash(&self) -> H256 {
+		keccak(&*self.rlp_bytes())
```
tomusdrw: It never changes, should be computed and memoized in `::new`

ethcore/private-tx/src/private_transactions.rs
tomusdrw: Let's make the two (`Scoring` and `Ready`) generic then.

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,128 +15,238 @@
...
+		self.transaction.heap_size_of_children()
+	}
+
+	fn sender(&self) -> &Address {
+		&self.transaction_sender
+	}
+}
+
+#[derive(Debug)]
+pub struct PrivateScorying;
```
tomusdrw: nit: typo, should be just `PrivateScoring`

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,128 +15,238 @@
...
-			.expect("One transaction inserted; one result returned; qed")?;
 
-		self.descriptors.insert(transaction_hash, PrivateTransactionDesc {
-			private_hash,
-			contract,
+		let options = self.verification_options.clone();
+		// Use pool's verifying pipeline for original transaction's verification
+		let verifier = pool::verifier::Verifier::new(client, options, Default::default());
+		let _verified_tx = verifier.verify_transaction(pool::verifier::Transaction::Unverified(transaction.clone()))?;
+		let signed_tx = SignedTransaction::new(transaction)?;
```
tomusdrw: Should take `SignedTransaction` from `_verified_tx`, otherwise you are recovering the signature twice

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,128 +15,238 @@
...
-	}
-
-	/// Remove transaction from the queue for verification
-	pub fn remove_private_transaction(&mut self, transaction_hash: &H256) {
-		self.descriptors.remove(transaction_hash);
-		self.transactions.remove(&[*transaction_hash], true);
+	pub fn drain<C: pool::client::NonceClient>(&self, client: C) -> Vec<Arc<VerifiedPrivateTransaction>> {
+		let ready = PrivateReadyState::new(client);
+		let mut hashes: Vec<H256> = Vec::new();
+		let res: Vec<Arc<VerifiedPrivateTransaction>> = self.verification_pool.read().pending(ready).collect();
```
tomusdrw:
```rust
let transactions: Vec<_> = self.verification_pool.read().pending(ready).collect()
let mut pool = self.verification_pool.write();
for tx in &transactions {
  pool.remove(tx.hash(), true);
}
transactions
```

doesn't make sense to allocate the vec twice.

ethcore/sync/src/chain/mod.rs
```diff
@@ -329,6 +329,10 @@ pub struct PeerInfo {
 	ask_time: Instant,
 	/// Holds a set of transactions recently sent to this peer to avoid spamming.
 	last_sent_transactions: HashSet<H256>,
+	/// Holds a set of private transactions recently sent to this peer to avoid spamming.
+	last_sent_private_transactions: HashSet<H256>,
```
tomusdrw: Why do we need to separate the two? The collision is not possibly anyway, for simplicity let's have a single set.

ethcore/sync/src/chain/mod.rs
```diff
@@ -1083,6 +1117,9 @@ impl ChainSync {
 				peer_info.last_sent_transactions.clear()
 			);
 		}
+
+		// reset stats for private transaction packets
+		self.clear_private_stats();
```
tomusdrw: So it seems that we always re-broadcast after the the block, right? Why can't we use the same logic as for regular transactions?

Also this statements should be guarded by similar ifs as the above code - we don't need to clear it when we import uncle block or we are syncing.

ethcore/sync/src/chain/mod.rs
```diff
@@ -1121,13 +1158,13 @@ impl ChainSync {
 	}
 
 	/// Broadcast private transaction message to peers.
-	pub fn propagate_private_transaction(&mut self, io: &mut SyncIo, packet: Bytes) {
-		SyncPropagator::propagate_private_transaction(self, io, packet);
+	pub fn propagate_private_transaction(&mut self, io: &mut SyncIo, transaction_hash: H256, packet: Bytes) {
```
tomusdrw: Might be worth merging the two methods (and the methods they are calling into one) and just discriminate using `enum` to avoid duplication.

5chdn:
Marking as _stale_ @grbIzl. :hatched_chick:

5chdn:
Will reopen once you continue working on it. :+1:

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,128 +15,188 @@
...
+	fn nonce(&self) -> U256 {
+		self.transaction.nonce
+	}
+}
+
+/// Checks readiness of transactions by comparing the nonce to state nonce.
+/// Guarantees only one transaction per sender
+#[derive(Debug)]
+pub struct PrivateReadyState<C> {
+	nonces: HashMap<Address, U256>,
```
tomusdrw: Doesn't really need to be a hashmap, we don't care about the values, cause every time it's occupied we return `Future` anyway. Would refactor to `HashSet` or sth (separate PR)

ethcore/sync/src/chain/mod.rs
```diff
@@ -1083,6 +1096,11 @@ impl ChainSync {
 			self.peers.values_mut().nth(peer).map(|peer_info|
 				peer_info.last_sent_transactions.clear()
 			);
+
+			//re-broadcast private packets as well
+			self.peers.values_mut().nth(peer).map(|peer_info|
+				peer_info.reset_private_stats()
```
tomusdrw: Why do we take `peers.values_mut().nth()` again, we have it just above. Please add a second line to `map`

miner/src/pool/scoring.rs
```diff
@@ -67,21 +67,21 @@ impl NonceAndGasPrice {
 	}
 }
 
-impl txpool::Scoring<VerifiedTransaction> for NonceAndGasPrice {
+impl<P: ?Sized> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txpool::VerifiedTransaction + Sized {
```
tomusdrw: So is it `?Sized` or `Sized`? I don't think we need either of the bounds (just rely on default `Sized`)

miner/src/pool/mod.rs
```diff
@@ -175,3 +182,19 @@ impl txpool::VerifiedTransaction for VerifiedTransaction {
...
 	}
 }
+
+impl ScoredTransaction for VerifiedTransaction {
+	fn priority(&self) -> Priority {
+		self.priority
+	}
+
+	/// Gets transaction gas price.
+	fn gas_price(&self) -> U256 {
```
tomusdrw: Might be better to return `&` here, to avoid copying.

miner/src/pool/mod.rs
```diff
@@ -108,6 +108,18 @@ impl Priority {
 	}
 }
 
+/// Scoring properties for verified transaction.
+pub trait ScoredTransaction {
```
tomusdrw: Maybe just extend `txpool::VerifiedTransaction` to simplify the bound in scoring implementation?

ethcore/private-tx/src/messages.rs
```diff
@@ -28,12 +28,28 @@ pub struct PrivateTransaction {
...
 	/// Address of the contract
 	pub contract: Address,
+	/// Hash
+	hash: H256,
 }
 
 impl PrivateTransaction {
-	/// Compute hash on private transaction
+	/// Constructor
+	pub fn new(encrypted: Bytes, contract: Address) -> Self {
```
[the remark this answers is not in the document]
debris: I actually like the idea that hash is cached in this way. We just need to make sure that this structure is never `mut`. Unfortunately rust does not allow us to enforce immutability of struct instances.

ethcore/private-tx/src/messages.rs
```diff
@@ -28,12 +28,28 @@ pub struct PrivateTransaction {
...
+
+	fn compute_hash(mut self) -> PrivateTransaction {
+		self.hash = keccak(&*self.rlp_bytes());
+		self
+	}
+
+	/// Hash of the private transaction
 	pub fn hash(&self) -> H256 {
-		keccak(&*self.rlp_bytes())
+		self.hash
```
debris: maybe just `pub hash` instead of the function?

ethcore/private-tx/src/messages.rs
```diff
@@ -28,12 +28,28 @@ pub struct PrivateTransaction {
...
 
 impl PrivateTransaction {
-	/// Compute hash on private transaction
+	/// Constructor
+	pub fn new(encrypted: Bytes, contract: Address) -> Self {
+		PrivateTransaction {
+			encrypted,
+			contract,
+			hash: 0.into(),
+		}.compute_hash()
```
debris: please get rid of the `compute_hash` function and just compute hash inline

ethcore/private-tx/src/messages.rs
```diff
@@ -59,7 +77,13 @@ impl SignedPrivateTransaction {
 			r: sig.r().into(),
 			s: sig.s().into(),
 			v: add_chain_replay_protection(sig.v() as u64, chain_id),
-		}
+			hash: 0.into(),
+		}.compute_hash()
```
debris: please get rid of the `compute_hash` function and just compute hash inline

ethcore/private-tx/src/lib.rs
```diff
@@ -261,48 +259,122 @@ impl Provider where {
...
+		let ready_transactions = self.transactions_for_verification.drain(self.pool_client(&nonce_cache));
 		for transaction in ready_transactions {
-			let transaction_hash = transaction.signed().hash();
-			match verification_queue.private_transaction_descriptor(&transaction_hash) {
-				Ok(desc) => {
-					if !self.validator_accounts.contains(&desc.validator_account) {
-						trace!("Cannot find validator account in config");
-						bail!(ErrorKind::ValidatorAccountNotSet);
+			let private_hash = transaction.private_transaction.hash();
+			match transaction.validator_account {
```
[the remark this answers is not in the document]
grbIzl: Reworked

ethcore/private-tx/src/lib.rs
```diff
@@ -562,115 +634,54 @@ pub trait Importer {
...
-			//Add signature to the store
-			match self.transactions_for_signing.lock().add_signature(&private_hash, tx.signature()) {
-				Ok(_) => trace!("Signature stored for private transaction"),
-				Err(err) => {
-					trace!("Failed to add signature to signing store, error: {:?}", err);
-					bail!(err);
-				}
-			}
+		let provider = Arc::downgrade(self);
+		if let Err(e) = self.channel.send(ClientIoMessage::execute(move |_| {
```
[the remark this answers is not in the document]
grbIzl: Reworked

ethcore/private-tx/src/lib.rs
```diff
@@ -688,7 +699,7 @@ impl ChainNotify for Provider {
 	fn new_blocks(&self, imported: Vec<H256>, _invalid: Vec<H256>, _route: ChainRoute, _sealed: Vec<H256>, _proposed: Vec<Bytes>, _duration: Duration) {
 		if !imported.is_empty() {
 			trace!("New blocks imported, try to prune the queue");
-			if let Err(err) = self.process_queue() {
+			if let Err(err) = self.process_verification_queue() {
 				trace!("Cannot prune private transactions queue. error: {:?}", err);
```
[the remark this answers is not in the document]
grbIzl: Thanks! I've looked through the whole file in order to fix such ambiguities and left only trace and warn

ethcore/private-tx/src/messages.rs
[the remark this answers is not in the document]
grbIzl: I agree with @tomusdrw Hashing was incorrectly added into API.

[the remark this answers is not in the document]
grbIzl: Reworked

ethcore/private-tx/src/messages.rs
```diff
@@ -73,4 +97,9 @@ impl SignedPrivateTransaction {
 	pub fn private_transaction_hash(&self) -> H256 {
 		self.private_transaction_hash
 	}
+
+	/// Own hash
+	pub fn hash(&self) -> H256 {
```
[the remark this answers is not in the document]
grbIzl: Don't understand this comment :-(

ethcore/private-tx/src/private_transactions.rs
grbIzl: Absolutely correct

ethcore/sync/src/chain/mod.rs
grbIzl: Moved to the proper place

miner/src/pool/scoring.rs
grbIzl: Removed

miner/src/pool/mod.rs
grbIzl: Reworked

miner/src/pool/mod.rs
grbIzl: I didn't want to add methods to the base interface because they are (IMO) related to the miner's functionality only and that's why you used them only in miner's scoring. So I didn't want to "garbage" API of base VerifiedTransaction

ethcore/private-tx/src/messages.rs
grbIzl: IMO such optimization in order to calculate hash only once is justified. I've reworked it in order to provide access to the fields via getters

ethcore/private-tx/src/lib.rs
```diff
@@ -261,48 +251,125 @@ impl Provider where {
...
+			if let Err(e) = process_transaction(&transaction) {
+				warn!("Error: {:?}", e);
+			}
 		}
 		Ok(())
 	}
 
+	/// Add signed private transaction into the store
+	/// Creates corresponding public transaction if last required signature collected and sends it to the chain
+	pub fn process_signature(&self, signed_tx: SignedPrivateTransaction) -> Result<(), Error> {
```
niklasad1: take ```signed_tx``` by reference instead and get rid of a ```clone()```: https://github.com/paritytech/parity-ethereum/blob/2778d4e99bb25c8e2df3f15617e57b16156fbc10/ethcore/private-tx/src/lib.rs#L678

ethcore/private-tx/src/lib.rs
```diff
@@ -261,48 +251,125 @@ impl Provider where {
...
-						warn!("Incorrect type of action for the transaction");
+						bail!("Incorrect type of action for the transaction");
 					}
-				},
-				Err(e) => {
-					warn!("Cannot retrieve descriptor for transaction with error {:?}", e);
 				}
 			}
-			verification_queue.remove_private_transaction(&transaction_hash);
+			return Ok(());
```
niklasad1: needless ```return``` i.e, enough with ```Ok(())```

ethcore/private-tx/src/lib.rs
```diff
@@ -261,48 +251,125 @@ impl Provider where {
...
+				}
+			}
+			//Remove from store for signing
+			match self.transactions_for_signing.write().remove(&private_hash) {
+				Ok(_) => {}
+				Err(err) => {
+					warn!("Failed to remove transaction from signing store, error: {:?}", err);
+					bail!(err);
+				}
+			}
```
niklasad1: Replace ```single-pattern match``` with:
 
```if let Err(err) = self.transactions_for_signing.write().remove(&private_hash) { ..... }```

File: ethcore/private-tx/src/lib.rs [hunk 30]
```diff
@@ -316,26 +383,26 @@ impl Provider where {
...
 				bail!(err);
 			}
 		}
 	}
 
 	/// Broadcast the private transaction message to the chain
-	fn broadcast_private_transaction(&self, message: Bytes) {
-		self.notify(|notify| notify.broadcast(ChainMessageType::PrivateTransaction(message.clone())));
+	fn broadcast_private_transaction(&self, transaction_hash: H256, message: Bytes) {
+		self.notify(|notify| notify.broadcast(ChainMessageType::PrivateTransaction(transaction_hash, message.clone())));
```
niklasad1: It looks suspect to `cloning` the vector here instead of moving into the closure! Any particular reason why?

ethcore/private-tx/src/lib.rs
```diff
@@ -690,8 +701,8 @@ impl ChainNotify for Provider {
 	fn new_blocks(&self, imported: Vec<H256>, _invalid: Vec<H256>, _route: ChainRoute, _sealed: Vec<H256>, _proposed: Vec<Bytes>, _duration: Duration) {
```
niklasad1: Out-of-scope for this PR:

Would be better to take the function parameters as slices to support more types than `Vec`, i.e, to use instead ```&[H256]```! 

But because this is part of a trait that would probably demand some work!

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
+use txpool;
+use txpool::{VerifiedTransaction, Verifier};
 use error::{Error, ErrorKind};
 
+type Pool = txpool::Pool<VerifiedPrivateTransaction, pool::scoring::NonceAndGasPrice>;
+
 /// Maximum length for private transactions queues.
 const MAX_QUEUE_LEN: usize = 8312;
 
 /// Desriptor for private transaction stored in queue for verification
```
niklasad1: Typo in ```Descriptor```

Rephrase this, something like `Private transaction type`!

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
-	/// Contract's address used in private transaction
-	pub contract: Address,
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VerifiedPrivateTransaction {
+	/// Original private transaction
+	pub private_transaction: PrivateTransaction,
 	/// Address that should be used for verification
-	pub validator_account: Address,
+	pub validator_account: Option<Address>,
+	/// Resulted verified
```
niklasad1: Verified result

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
+	pub transaction: SignedTransaction,
+	/// Original transaction's hash
```
niklasad1: Original transaction hash

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
+	pub transaction_hash: H256,
+	/// Original transaction's sender
```
niklasad1: Orginal transaction sender or Transaction origin address

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
-			.next()
-			.expect("One transaction inserted; one result returned; qed")?;
-
-		self.descriptors.insert(transaction_hash, PrivateTransactionDesc {
-			private_hash,
-			contract,
+		let options = self.verification_options.clone();
+		// Use pool's verifying pipeline for original transaction's verification
+		let verifier = pool::verifier::Verifier::new(client, options, Default::default(), None);
+		let verified_tx = verifier.verify_transaction(pool::verifier::Transaction::Unverified(transaction.clone()))?;
```
niklasad1: Needless `clone()` call 

You can either take the `transaction (UnverifiedTransaction)` by reference and clone it or take it by value and move it!

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -15,129 +15,182 @@
...
 		}
 	}
 }
 
 impl VerificationStore {
 	/// Adds private transaction for verification into the store
 	pub fn add_transaction<C: pool::client::Client>(
-		&mut self,
+		&self,
 		transaction: UnverifiedTransaction,
```
niklasad1: See comment below!

ethcore/sync/src/private_tx.rs
```diff
@@ -15,26 +15,29 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use parking_lot::Mutex;
+use ethereum_types::H256;
 
 /// Trait which should be implemented by a private transaction handler.
 pub trait PrivateTxHandler: Send + Sync + 'static {
 	/// Function called on new private transaction received.
-	fn import_private_transaction(&self, rlp: &[u8]) -> Result<(), String>;
+	/// Returns hash of the imported transaction
```
niklasad1: **the** hash

ethcore/sync/src/private_tx.rs
```diff
@@ -15,26 +15,29 @@
...
+	fn import_private_transaction(&self, rlp: &[u8]) -> Result<H256, String>;
 
 	/// Function called on new signed private transaction received.
-	fn import_signed_private_transaction(&self, rlp: &[u8]) -> Result<(), String>;
+	/// Returns hash of the imported transaction
```
niklasad1: **the** hash

miner/src/pool/scoring.rs
```diff
@@ -185,12 +185,8 @@ mod tests {
 		};
 
 		let keypair = Random.generate().unwrap();
-		let txs = vec![tx1, tx2, tx3, tx4].into_iter().enumerate().map(|(i, tx)| {
-			let verified = tx.unsigned().sign(keypair.secret(), None).verified();
-			txpool::Transaction {
-				insertion_id: i as u64,
-				transaction: Arc::new(verified),
-			}
+		let txs = vec![tx1, tx2, tx3, tx4].into_iter().enumerate().map(|(_, tx)| {
```
niklasad1: remove ```enumerate()``` because it is not used!

ethcore/private-tx/src/lib.rs
grbIzl: It's not possible to move captured outer variable (message)

sorpaas: Just a note: in the future we may consider to change type `ChainMessageType` to only take reference of `[u8]`. In that case this should solve this message clone issue.

PR #8717, opened 2018-05-26
Update dev chain

tomusdrw:
- Enable WASM
- Enable Metropolis features and in general make it more foundation-chain-like.


folsen:
Adding this to releasenotes since it will break all existing dev chains.

PR #8719, opened 2018-05-27
Remove -k/--insecure option from curl installer

vrde:
Piping `curl` to `bash` while **disabling** certificate verification can lead to security problems. This PR removes the `-k` option from the one-line installer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,13 +138,13 @@ first.
 ## Simple one-line installer for Mac and Ubuntu
 
 ```bash
-bash <(curl https://get.parity.io -Lk)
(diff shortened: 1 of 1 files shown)


vrde:
[clabot:check]

folsen:
Why was it added in the first place?

niklasad1:
@folsen I don't know but probably hosted on a website without https/tls/ssl?

folsen:
@niklasad1 ah, yeah probably, makes sense

5chdn:
We also need this updated on the wiki: https://github.com/paritytech/wiki

PR #8726, opened 2018-05-29
Ease tiny-keccak version requirements (1.4.1 -> 1.4)

svyatonik:
this should help with issues like [that one](https://travis-ci.org/paritytech/polkadot/builds/385095222?utm_source=github_status&utm_medium=notification)


niklasad1:
@svyatonik I think you have to change the version in `Cargo.lock` to fix that issue or have you tested this branch with Polkadot?

svyatonik:
Yeah - checked this locally. Guess `Cargo.lock` is not used here, when we're using subcrates.

PR #8727, opened 2018-05-29
Add a deadlock detection thread

ordian:
And expose it under a feature flag: `cargo build --features "deadlock_detection"`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1093,6 +1093,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "fixedbitset"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "flate2"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1940,6 +1945,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "ordermap"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "owning_ref"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


parity/lib.rs
```diff
@@ -141,6 +141,37 @@ fn print_hash_of(maybe_file: Option<String>) -> Result<String, String> {
 	}
 }
 
+fn run_deadlock_detection_thread() {
+	#[cfg(feature = "deadlock_detection")]
```
niklasad1: What is the reason of putting this `conditional compilation flag` inside the function rather than outside?

parity/lib.rs
```diff
@@ -161,6 +192,8 @@ fn execute<Cr, Rr>(command: Execute, on_client_rq: Cr, on_updater_rq: Rr) -> Res
 	// 		they want
 	let logger = setup_log(&command.logger).expect("Logger is initialized only once; qed");
 
+	run_deadlock_detection_thread();
```
niklasad1: I think it would be more verbose to have a compilation flag here 

```rust    
#[cfg(feature = "deadlock_detection")]
run_deadlock_detection_thread()
```

parity/lib.rs
```diff
@@ -141,6 +141,37 @@ fn print_hash_of(maybe_file: Option<String>) -> Result<String, String> {
...
+
+		info!("Starting deadlock detection thread.");
+		// Create a background thread which checks for deadlocks every 10s
+		thread::spawn(move || {
+			loop {
+				thread::sleep(Duration::from_secs(10));
+				let deadlocks = deadlock::check_deadlock();
+				if deadlocks.is_empty() {
+					continue;
+				}
```
niklasad1: Grumble on this `if statement` because `deadlock iterator` will handle this case but I guess it is ok if you want to print the number of iterators. 

No big deal for me I just want the code to be as simple as possible!

parity/lib.rs
```diff
@@ -141,6 +141,35 @@ fn print_hash_of(maybe_file: Option<String>) -> Result<String, String> {
...
+	info!("Starting deadlock detection thread.");
+	// Create a background thread which checks for deadlocks every 10s
+	thread::spawn(move || {
+		loop {
+			thread::sleep(Duration::from_secs(10));
+			let deadlocks = deadlock::check_deadlock();
+			if deadlocks.is_empty() {
+				continue;
+			}
+
```
niklasad1: Grumble on this if statement because deadlock iterator will handle this case but I guess it is ok if you want to print the number of deadlocks!

No big deal for me I just want the code to be as simple as possible!

ordian:
The code was copied from the docs: https://amanieu.github.io/parking_lot/parking_lot/deadlock/index.html#example. I guess it's simple enough.

Cargo.lock
```diff
@@ -3183,6 +3205,16 @@ dependencies = [
  "unicode-width 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "thread-id"
```
tomaka: This crate doesn't compile for wasm32-unknown-unknown, but since it's behind a feature flag it doesn't really matter.

ordian: I'm not whether the `Cargo.lock` should be updated, since this is under a feature flag. Should I revert the changes to `Cargo.lock`?

tomaka: No, the `Cargo.lock` always considers all options I think. So it's correct to update it.

PR #8731, opened 2018-05-29
Fix some nits using clippy

niklasad1:
- Got rid of some unsafe blocks
- Don't clone copy types
- Some other fixes

### DIFF
--- a/whisper/cli/src/main.rs
+++ b/whisper/cli/src/main.rs
@@ -19,6 +19,9 @@
 //! Spawns an Ethereum network instance and attaches the Whisper protocol RPCs to it.
 //!
 
+#![warn(missing_docs)]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy, clippy_pedantic))]
+
 extern crate docopt;
 extern crate ethcore_network_devp2p as devp2p;
 extern crate ethcore_network as net;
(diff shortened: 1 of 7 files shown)


ordian:
A general question: do we want to integrate clippy checks into CI? It can be pinned to a specific rust version like [this](https://github.com/killercup/cargo-edit/blob/88b3e49dfdac8640aac671c594362204c18c8802/.travis.yml#L15-L19).

niklasad1:
Yeah, I think we should enable clippy in the CI to enforce proper `coding guidelines` and so on but I guess it would require some work to fix all things in the code base and so on!

FYI, I heard some rumors that clippy will be shipped in `rustc nightly` in the future!

whisper/cli/src/main.rs
```diff
@@ -19,6 +19,9 @@
 //! Spawns an Ethereum network instance and attaches the Whisper protocol RPCs to it.
 //!
 
+#![warn(missing_docs)]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy, clippy_pedantic))]
```
dvdplm: Can you elaborate on what this does? Disables slippy for this file only?

ordian: I think `deny` turns warning into errors. The `!` applies this rule to the module (in this case only this file).

niklasad1: Yeah, @ordian I think you are right.

But, in addition, `cargo-clippy` is only available on the nightly toolchain so these will be ignored if you use the original `rust-toolchain` file in Parity (assuming no overrides)!

cargo-clippy is essentially a modified verision of `rustc` that you run instead of rustc with additional lints i.e, instead of running rustc you run clippy!

However, it might be better to use `warn` instead of `deny`!

PR #8737, opened 2018-05-30
A more helpful Error message when we can't handshake because of too many open files

### LINKED ISSUE #6791: Improve error message: "Too many open files in system"
> I'm running:
>
> - **Parity version**: 1.8.0
> - **And installed**: via homebrew

---

```
====================

stack backtrace:

Thread 'IO Worker #3' panicked at 'Can't create handshake: StdIo(Error { repr: Os { code: 23, message: "Too many open files in system" } })', src/libcore/result.rs:860

[1]    60411 abort      parity
```

gnunicorn:
Attemps to fix #6791 by providing a better error message when the handshake fails because
of limited resources given by the system around.

### DIFF
--- a/util/network-devp2p/src/session.rs
+++ b/util/network-devp2p/src/session.rs
@@ -104,29 +104,45 @@ impl Session {
 		nonce: &H256, host: &HostInfo) -> Result<Session, Error>
 		where Message: Send + Clone + Sync + 'static {
 		let originated = id.is_some();
-		let mut handshake = Handshake::new(token, id, socket, nonce).expect("Can't create handshake");
-		let local_addr = handshake.connection.local_addr_str();
-		handshake.start(io, host, originated)?;
-		Ok(Session {
-			state: State::Handshake(handshake),
-			had_hello: false,
-			info: SessionInfo {
-				id: id.cloned(),
-				client_version: String::new(),
-				protocol_version: 0,
-				capabilities: Vec::new(),
-				peer_capabilities: Vec::new(),
-				ping: None,
-				originated: originated,
-				remote_address: "Handshake".to_owned(),
-				local_address: local_addr,
+		match Handshake::new(token, id, socket, nonce) {
+			Ok(mut handshake) => {
+				let local_addr = handshake.connection.local_addr_str();
+				handshake.start(io, host, originated)?;
+				Ok(Session {
+					state: State::Handshake(handshake),
+					had_hello: false,
+					info: SessionInfo {
+						id: id.cloned(),
+						client_version: String::new(),
+						protocol_version: 0,
+						capabilities: Vec::new(),
+						peer_capabilities: Vec::new(),
+						ping: None,
+						originated: originated,
+						remote_address: "Handshake".to_owned(),
+						local_address: local_addr,
+					},
+					ping_time: Instant::now(),
+					pong_time: None,
+					expired: false,
+					protocol_states: HashMap::new(),
+					compression: false,
+				})
 			},
-			ping_time: Instant::now(),
-			pong_time: None,
-			expired: false,
-			protocol_states: HashMap::new(),
-			compression: false,
-		})
+			Err(err) => {
+				if let Error(ErrorKind::Io(inner), state) = err {
+					Err(Error(
+						if inner.raw_os_error() == Some(24) {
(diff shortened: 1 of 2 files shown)


util/network-devp2p/src/session.rs
```diff
@@ -104,29 +104,45 @@ impl Session {
...
-			ping_time: Instant::now(),
-			pong_time: None,
-			expired: false,
-			protocol_states: HashMap::new(),
-			compression: false,
-		})
+			Err(err) => {
+				if let Error(ErrorKind::Io(inner), state) = err {
+					Err(Error(
+						if inner.raw_os_error() == Some(24) {
```
gnunicorn: I am not super happy with this: upacking the error to call that function on it with a specific hard-coded CODE-Number is one thing, but then replacing that with a different error to give our message seems crude -- especially here.

However, I couldn't find any way in which `error_chains` supports linking different local errors upon the same type but discriminating by a function call or closure -- which I think, how it is should be done instead... Also because this error message is now only matched in this particular case. Which fixes the issue reported, but should we fail with a TOO_MANY_OPEN_FILES on any part of the code (which any call to `ethkey::generate` might does), we still see the rather cryptic previous error message :( .

Tips/ideas appreciated!

util/network-devp2p/src/session.rs
```diff
@@ -104,29 +104,45 @@ impl Session {
...
+							// handle sys error CODE 24; "Too many open files" more gracefully
+							ErrorKind::TooManyOpenFiles("Can't create handshake.".to_string())
```
dvdplm: Forgive me the silly question but what does this look like to the user in the console?

```
TooManyOpenFiles: Can't create handshake
```

Does the process exit at that point? Does it carry with it the error code when exiting? I'm thinking about people running parity on a server with monitoring attached looking for specific process exit codes to trigger alarms.

util/network-devp2p/src/session.rs
dvdplm: Not sure how to do this better (or if that is possible). A question though: is `24` platform agnostic? Looks like it's the Linux `EMFILE`.

util/network-devp2p/src/session.rs
gnunicorn: Honestly wasn't able to trigger this particular warning myself, but took the information provided in the corresponding ticket - https://github.com/paritytech/parity/issues/6791 .

From my understanding this happens in a Worker-Thread, which - as a result - then panics. The process itself keeps running and from the weirdness of the backtrace message, I doubt anyone tried to have any script do things when that error appears. So for the case you are thinking about this doesn't change anything.

util/network-devp2p/src/session.rs
dvdplm: Looking closer it seems like `EMFILE` is actually standard and defined in `errno.h` in the C stdlib. So that is probably fine. Reading the issue again I noticed they ran into error `23` and not `24` and then I found [this](https://stackoverflow.com/questions/24862733/difference-between-linux-errno-23-and-linux-errno-24).
I think we should try to respond sensibly to both cases (i.e. `ENFILE`/`23` is too many files in the *system* and `EMFILE`/`24` is too many files in the *process*).

dvdplm: We also should use an enum with the error numbers, I'd say the `libc` crate is appropriate here: [`ENFILE`](https://msdn.microsoft.com/en-us/library/t3ayayh1.aspx), [`EMFILE`](https://docs.rs/libc/0.2.36/libc/constant.EMFILE.html).

gnunicorn: @dvdplm oh, looks like they were running into both: 
 - [23 in the first comment](https://github.com/paritytech/parity/issues/6791#issue-265711449)
 - [24 in this one](https://github.com/paritytech/parity/issues/6791#issuecomment-378793502)

The Error Message is only appropriate for 24. In 23 I suppose you have to kill some processes or restart the system or something? I'll ask in the ticket, what message they'd like for that case.

dvdplm: > kill some processes or restart the system or something

well, you'll need to release some file handle descriptors! ;) Or increase the system-level ulimits. A system restart should not be necessary though.

gnunicorn:
Closing as I've found a better approach, submitted as #8744

PR #8738, opened 2018-05-30
Pause pruning during snapshotting

### LINKED ISSUE #7863: Pause pruning during snapshotting
Fast-pruned nodes should be able to generate snapshots before the history required for the snapshot is pruned away. One idea would be to disable the pruning during snapshotting and continue pruning once the snapshot is available.

ordian:
Fixes #7863.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -189,6 +189,9 @@ pub struct Client {
 	/// Database pruning strategy to use for StateDB
 	pruning: journaldb::Algorithm,
 
+	/// Whether `StateDB` pruning is paused.
+	pruning_paused: AtomicBool,
+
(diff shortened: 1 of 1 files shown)


ordian:
@rphmeier How about tracking the earliest state era needed for taking a snapshot?

rphmeier:
can you elaborate? I don't understand what you mean exactly

ordian:
> ordian> I was thinking we can track and constantly update the minimal block number needed to produce the current state chunk. Or we can't do that?
> rphmeier> in general it will always be the same block
> rphmeier> otherwise we would have to add a ton of metadata to each trie node saying the block number it was last modified
> rphmeier> that would be possible if we had an actual "trie" DB which is specialized for storing the state trie

PR #8739, opened 2018-05-30
Make HashDB generic

dvdplm:
The aim of this PR is to make `patricia-trie` usable from Polkadot by making it generic over both the hashing algorithm and the encoding scheme (currently Keccak-256 and RLP).

Part of https://github.com/paritytech/parity/issues/7019 and https://github.com/paritytech/parity/issues/8620

The [`HashDB` trait](https://github.com/paritytech/parity/pull/8739/files?utf8=%E2%9C%93&diff=split#diff-b7cc171f2158609f8381e6ce2db3e77eR37) – the basis of the various Trie DB types – is made generic over the hashing algorithm and output format by introducing a new [`Hasher` trait](https://github.com/paritytech/parity/pull/8739/files?utf8=%E2%9C%93&diff=split#diff-b7cc171f2158609f8381e6ce2db3e77eR26) and threading it throughout the crates using it. 
With the `NodeCodec` trait we encapsulate the information needed to encode/decode a trie node. The [`NodeCodec` trait](https://github.com/paritytech/parity/pull/8739/files?utf8=%E2%9C%93&diff=split#diff-f2ef68ff3f4dd78aa49292623abe3826R27) is generic over the `Hasher` as well.

The concrete implementations we need in Parity for Keccak-256 and RLP, are provided in two new mini-crates: [`keccak-hasher`](https://github.com/paritytech/parity/pull/8739/files?utf8=%E2%9C%93&diff=split#diff-48b380f843384eec8bea18feb56c2a31R27) and [`patricia-trie-ethereum`](https://github.com/paritytech/parity/pull/8739/files?utf8=%E2%9C%93&diff=split#diff-13db864be650a234bec8c7150a4230e7R1). Here we also provide type aliases for `Result` and `TrieError`.

Benchmark compared to master:

```
 name                        master ns/iter  PR ns/iter        diff ns/iter  diff %  speedup
 trie_insertions_32_mir_1k   2,230,153        2,217,779            -12,374  -0.55%   x 1.01
 trie_insertions_32_ran_1k   2,216,102        2,231,613             15,511   0.70%   x 0.99
 trie_insertions_random_mid  1,556,671        1,578,555             21,884   1.41%   x 0.99
 trie_insertions_six_high    1,604,928        1,643,805             38,877   2.42%   x 0.98
 trie_insertions_six_low     3,049,674        3,090,029             40,355   1.32%   x 0.99
 trie_insertions_six_mid     2,048,429        2,059,450             11,021   0.54%   x 0.99
 trie_iter                   1,581,328        1,549,457            -31,871  -2.02%   x 1.02
```


util/hashdb/src/lib.rs
```diff
@@ -17,62 +17,89 @@
...
+		out.into()
+	}
+}
 
 /// `HashDB` value type.
 pub type DBValue = ElasticArray128<u8>;
 
 /// Trait modelling datastore keyed by a 32-byte Keccak hash.
-pub trait HashDB: AsHashDB + Send + Sync {
+pub trait HashDB: Send + Sync {
```
dvdplm: How do I best fit in `AsHashDB` in this? It takes a `Hasher` now (see below), but not sure how to make pass the type argument.

util/plain_hasher/src/lib.rs
```diff
@@ -1,13 +1,15 @@
...
+extern crate hashdb;
 
 use std::{hash, mem};
 use std::collections::{HashMap, HashSet};
 use ethereum_types::H256;
+use hashdb::Hasher;
 
 /// Specialized version of `HashMap` with H256 keys and fast hashing function.
-pub type H256FastMap<T> = HashMap<H256, T, hash::BuildHasherDefault<PlainHasher>>;
+pub type H256FastMap<H, T> = HashMap<<H as Hasher>::Out, T, hash::BuildHasherDefault<PlainHasher>>;
```
dvdplm: Should we change the name of this? To just `FastMap`?

util/plain_hasher/src/lib.rs
```diff
@@ -1,13 +1,15 @@
...
 /// Specialized version of `HashSet` with H256 keys and fast hashing function.
 pub type H256FastSet = HashSet<H256, hash::BuildHasherDefault<PlainHasher>>;
```
dvdplm: Not sure what this is used for tbh. Anyone?

util/patricia_trie/src/lib.rs
```diff
@@ -49,155 +49,159 @@ mod lookup;
...
+pub use self::triedbmut::TrieDBMut;
 pub use self::sectriedbmut::SecTrieDBMut;
 pub use self::sectriedb::SecTrieDB;
 pub use self::fatdb::{FatDB, FatDBIterator};
 pub use self::fatdbmut::FatDBMut;
 pub use self::recorder::Recorder;
+use node_codec::NodeCodec;
+
+// REVIEW: is this overdoing it?
+pub type KeccakRlpNodeCodec = node_codec::RlpNodeCodec<hashdb::KeccakHasher>;
```
rphmeier: it might be overdoing it a bit. I think the most idiomatic thing is to have `RlpNodeCodec` stuff defined in its own crate, and `KeccakHasher` defined in its own crate as well, without `patricia-trie` depending on either. Then there might be some facade crate (`patricia-trie-ethereum`) which re-exports the trie types and traits from the `patricia-trie` crate and depends exports some typedefs using `KeccakHasher` and `RlpNodeCodec`. IMO those typedefs should have simple names, that don't mention `Keccak` or  `Rlp` since they are apparent from context.

But that's if you want to go all out. What's here is enough for generalization in Polkadot.

util/patricia_trie/src/lib.rs
```diff
@@ -49,155 +49,159 @@ mod lookup;
...
+pub enum TrieError<T> {
 	/// Attempted to create a trie with a state root not in the DB.
-	InvalidStateRoot(H256),
+	InvalidStateRoot(T),
 	/// Trie item not found in the database,
-	IncompleteDatabase(H256),
+	IncompleteDatabase(T),
 	/// Corrupt Trie item
-	DecoderError(rlp::DecoderError),
+	// REVIEW: what we'd really like to do here is include the `rlp::DecoderError` in the `TrieError`
```
rphmeier: what we'd specifically want to do is define `TrieError<T, E>` where `E=rlp::DecoderError` in Ethereum-like chains.

rphmeier: (and provide a typedef either here or in `patricia-trie-ethereum` if we go that route, `type TrieError = generic::TrieError<H256, rlp::DecoderError>`)

util/patricia_trie/src/lib.rs
```diff
@@ -49,155 +49,159 @@ mod lookup;
...
 	type Item;
 
 	/// Decode a byte-slice into the desired item.
-	fn decode(self, &[u8]) -> Self::Item;
+	fn decode(self, data: &[u8]) -> Self::Item;
 
 	/// Record that a node has been passed through.
-	fn record(&mut self, &H256, &[u8], u32) { }
+	#[allow(unused_variables)]
+	fn record(&mut self, hash: &H::Out, data: &[u8], depth: u32) {}
```
rphmeier: better just prefix the names with `_` (e.g. `_hash`) rather than silencing the lint

util/patricia_trie/src/lib.rs
```diff
@@ -49,155 +49,159 @@ mod lookup;
...
 /// A key-value datastore implemented as a database-backed modified Merkle tree.
 pub trait Trie {
+	type H: Hasher;
 	/// Return the root of the trie.
-	fn root(&self) -> &H256;
+	fn root(&self) -> &<Self::H as Hasher>::Out;
 
 	/// Is the trie empty?
-	fn is_empty(&self) -> bool { *self.root() == KECCAK_NULL_RLP }
+	fn is_empty(&self) -> bool { *self.root() == Self::H::HASHED_NULL_RLP }
```
rphmeier: specifically, this should draw from the node codec, because non-rlp-encoded tries will have non-rlp-encoded null nodes. so the hash will be different. this is an architectural issue:

`trait NodeCodec<H: Hasher>` should have the `const HASH_NULL_NODE: H::Out`. makes implementation of the trait in a generic context impossible w/o const functions but for now we can limit scope a bit.

util/patricia_trie/src/lib.rs
```diff
@@ -245,40 +251,41 @@ macro_rules! wrapper {
 	}
 }
 
-impl<'db> Trie for TrieKinds<'db> {
-	fn root(&self) -> &H256 {
+impl<'db, H: Hasher, C: NodeCodec<H>> Trie for TrieKinds<'db, H, C> where H::Out: rlp::Decodable + rlp::Encodable {
```
rphmeier: why are these `rlp` bounds necessary?

util/patricia_trie/src/lookup.rs
```diff
@@ -16,27 +16,35 @@
...
+pub struct Lookup<'a, H: Hasher + 'a, C: NodeCodec<H>, Q: Query<H>> {
 	/// database to query from.
-	pub db: &'a HashDB,
+	pub db: &'a HashDB<H>,
 	/// Query object to record nodes and transform data.
 	pub query: Q,
 	/// Hash to start at
-	pub hash: H256,
+	pub hash: H::Out,
+	pub marker: PhantomData<C>, // TODO: probably not needed when all is said and done? When Query is made generic?
```
rphmeier: `pub codec: &'a C` would probably be fine

util/patricia_trie/src/lookup.rs
```diff
@@ -55,7 +63,8 @@ impl<'a, Q: Query> Lookup<'a, Q> {
 			// without incrementing the depth.
 			let mut node_data = &node_data[..];
 			loop {
-				match Node::decoded(node_data)? {
+				// TODO: see comment on `TrieError::DecoderError` for what to do here
+				match C::decode(node_data).map_err(|e| TrieError::DecoderError(hash, format!("{:?}",e)))? {
```
rphmeier: if `TrieError::DecoderError` is generic over decoder error type as well, this makes things easy

util/patricia_trie/src/node_codec.rs
```diff
@@ -0,0 +1,107 @@
+use bytes::*;
+use nibbleslice::NibbleSlice;
+use rlp::{Prototype, Rlp, RlpStream, DecoderError, Decodable, Encodable};
+use hashdb::Hasher;
+use node::Node;
+use std::marker::PhantomData;
+
+pub trait NodeCodec<H: Hasher>: Sized {
```
rphmeier: docs!

util/patricia_trie/src/node_codec.rs
```diff
@@ -0,0 +1,107 @@
+use bytes::*;
```
rphmeier: 1. no glob imports
2. copyright header
3. module docs

util/patricia_trie/src/node_codec.rs
rphmeier: on all public items

util/patricia_trie/src/triedbmut.rs
```diff
@@ -129,29 +128,31 @@ impl Node {
 
 	// encode a node to RLP
 	// TODO: parallelize
-	fn into_rlp<F>(self, mut child_cb: F) -> ElasticArray1024<u8>
-		where F: FnMut(NodeHandle, &mut RlpStream)
+	fn into_rlp<F, C>(self, mut child_cb: F) -> ElasticArray1024<u8>
```
rphmeier: need to generalize so we don't necessarily refer to RLP

rphmeier:
...started to review. the `Hasher` stuff seems mostly fine the the changes that have been made so far seem fine. But the `NodeCodec` generalization seems halfway done since there are lots of references to RLP stuff all over the place.

My feeling is that for the initial generalization our need in Polkadot is more about hash function than encoding, so we could make the case for leaving the `NodeCodec` stuff out of this PR and doing additional generalization later. But I don't feel comfortable merging this PR unless:
  - no `NodeCodec` stuff is included and we just use RLP and generalize over hash function
  - `NodeCodec` stuff is included and the trie code makes no direct references to the RLP crate whatsoever. The crate should be able to compile without `rlp` as a dependency.

util/patricia_trie/src/lib.rs
dvdplm: Largely they are not needed (put there before `NodeCodec` was added) and I'll remove them.

However there's the problem of `into_rlp()` and `commit_node()` methods that still make use of `RlpStream` and they introduce the `Encodable` bound (which then bubble up).

Partly fixed in #5db5de901

ethcore/Cargo.toml
```diff
@@ -17,9 +17,9 @@ ethash = { path = "../ethash" }
 ethcore-bloom-journal = { path = "../util/bloom" }
 ethcore-bytes = { path = "../util/bytes" }
 fetch = { path = "../util/fetch" }
-hashdb = { path = "../util/hashdb" }
-memorydb = { path = "../util/memorydb" }
-patricia-trie = { path = "../util/patricia_trie" }
+hashdb = { version = "0.2", path = "../util/hashdb" }
+memorydb = { version = "0.2", path = "../util/memorydb" }
+patricia-trie = { version = "0.2", path = "../util/patricia_trie" }
```
debris: `version =` is not needed, unless you want to publish `ethcore` to crates.io. In all other cases, it's always ignored if it occurs together with `path =`

ethcore/light/src/provider.rs
```diff
@@ -386,7 +386,7 @@ impl<L: AsLightClient> AsLightClient for LightProvider<L> {
 
 #[cfg(test)]
 mod tests {
-	use ethcore::client::{EachBlockWith, TestBlockChainClient};
+	use ::ethcore::client::{EachBlockWith, TestBlockChainClient};
```
debris: any reason for that?

util/plain_hasher/src/lib.rs
```diff
@@ -17,13 +17,14 @@
...
 
 use ethereum_types::H256;
-use std::collections::{HashMap, HashSet};
+use hashdb::Hasher;
 use std::hash;
-
+use std::collections::{HashMap, HashSet};
 /// Specialized version of `HashMap` with H256 keys and fast hashing function.
-pub type H256FastMap<T> = HashMap<H256, T, hash::BuildHasherDefault<PlainHasher>>;
+pub type H256FastMap<H, T> = HashMap<<H as Hasher>::Out, T, hash::BuildHasherDefault<PlainHasher>>;
```
debris: This is invalid. `plain_hasher` has an assumption that key is 32 bytes long hash. It will not work for hashes of different length

util/hashdb/src/lib.rs
```diff
@@ -17,62 +17,77 @@
...
-use std::collections::HashMap;
 use elastic_array::ElasticArray128;
 use ethereum_types::H256;
+use heapsize::HeapSizeOf;
+use std::collections::HashMap;
+use std::{fmt::Debug, hash::Hash};
+use tiny_keccak::Keccak;
+
+pub trait Hasher: Sync + Send {
+	type Out: AsRef<[u8]> + Debug + PartialEq + Eq + Clone + Copy + Hash + Send + Sync /* REVIEW: how do I get around this? --> */ + HeapSizeOf;
```
debris: Most of these trait requirements are specific to particular functions using `Hasher` and imo only those functions should require a template type to implement them.

util/patricia_trie/benches/trie.rs
```diff
@@ -69,9 +74,9 @@ fn trie_insertions_32_mir_1k(b: &mut Bencher) {
 	};
 	let d = st.make();
 	b.iter(&mut ||{
-		let mut memdb = MemoryDB::new();
-		let mut root = H256::new();
-		let mut t = TrieDBMut::new(&mut memdb, &mut root);
+		let mut memdb = MemoryDB::<KeccakHasher>::new();
+		let mut root = <KeccakHasher as Hasher>::Out::new();
```
debris: `H256::new()` looks nicer and should still work ;)

util/patricia_trie/src/fatdbmut.rs
```diff
@@ -14,105 +14,119 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-use ethereum_types::H256;
-use keccak::keccak;
-use hashdb::{HashDB, DBValue};
+//use ethereum_types::H256;
+//use keccak::keccak;
```
debris: to remove

util/memorydb/src/lib.rs
```diff
@@ -320,4 +336,20 @@ mod tests {
...
+//			const HASHED_NULL_RLP: ethereum_types::H160= ethereum_types::H160([0; 20]);
+//			fn hash(_x: &[u8]) -> Self::Out { ethereum_types::H160(*b"01010202010102020101") }
+//		}
+//		impl HeapSizeOf for DummyHasher { fn heap_size_of_children(&self) -> usize { 0 } }
+//
+//		let mut db = MemoryDB::<DummyHasher>::new();
+//		// TODO: Fails. Trying to use a type that isn't H256 fails because of the tight coupling between memorydb and plain_hasher (specifically the assert on key length == 32)
+//		 let key = db.insert(b"32103210321032103210321032103210");
+//		 assert_eq!(key, ethereum_types::H264(*b"010102020101020201010202010102025"));
+//	}
```
debris: uncomment or remove

ethcore/light/src/on_demand/request.rs
```diff
@@ -650,9 +651,10 @@ impl From<::rlp::DecoderError> for Error {
 	}
 }
 
-impl From<Box<TrieError>> for Error {
-	fn from(err: Box<TrieError>) -> Self {
-		Error::Trie(*err)
+impl<T> From<Box<TrieError<T>>> for Error {
+	fn from(err: Box<TrieError<T>>) -> Self {
+		// Error::Trie(*err)
+		Error::Trie(TrieError::InvalidStateRoot(<KeccakHasher as Hasher>::Out::new())) // REVIEW: how do I fix this without making `Error` generic also?
```
debris: maybe `TrieError::InvalidStateRoot(String)` and implement conversion to it for every `Hasher::Out + Debug` ?

util/patricia_trie/src/triedb.rs
```diff
@@ -39,45 +42,51 @@ use bytes::Bytes;
...
 ///   assert!(t.contains(b"foo").unwrap());
 ///   assert_eq!(t.get(b"foo").unwrap().unwrap(), DBValue::from_slice(b"bar"));
 /// }
 /// ```
-pub struct TrieDB<'db> {
-	db: &'db HashDB,
-	root: &'db H256,
+pub struct TrieDB<'db, H, C>
+where 
+	H: Hasher + 'db,
```
debris: is this `+ 'db` really needed? imo, `H` requirement is a static lifetime and compiler should easily infer it

util/patricia_trie/src/triedb.rs
```diff
@@ -39,45 +42,51 @@ use bytes::Bytes;
...
+///   TrieDBMut::<_, RlpCodec>::new(&mut memdb, &mut root).insert(b"foo", b"bar").unwrap();
+///   let t = TrieDB::<_, RlpCodec>::new(&memdb, &root).unwrap();
 ///   assert!(t.contains(b"foo").unwrap());
 ///   assert_eq!(t.get(b"foo").unwrap().unwrap(), DBValue::from_slice(b"bar"));
 /// }
 /// ```
-pub struct TrieDB<'db> {
-	db: &'db HashDB,
-	root: &'db H256,
+pub struct TrieDB<'db, H, C>
```
debris: have you tried changing the order of generics?

```
struct TrieDB<'db, C, H> { }
```

I believe it will allow you to init `TrieDB` without boilerplate `_`.

`TrieDB::<_, RlpCodec>new` -> `TrieDB::<RlpCodec>::new`

util/patricia_trie/src/triedb.rs
```diff
@@ -39,45 +42,51 @@ use bytes::Bytes;
...
+{
+	db: &'db HashDB<H>,
+	root: &'db H::Out,
 	/// The number of hashes performed so far in operations on this trie.
 	hash_count: usize,
+	codec_marker: PhantomData<C>,
 }
 
-impl<'db> TrieDB<'db> {
+impl<'db, H, C> TrieDB<'db, H, C>
```
debris: you should add another `impl` to simplify init of this structure

```rust
impl<'db> TrieDB<'db, RlpCodec, KeccakHasher> {
  fn with_rlp_codec(db: &'db HashDB<KeccakHasher>, root: &db H256) -> super::Result<Self> {
    Self::new(db, root)
  }
}

```

util/patricia_trie/src/triedb.rs
```diff
@@ -39,45 +42,51 @@ use bytes::Bytes;
...
+where 
+	H: Hasher, 
+	C: NodeCodec<H>
+{
 	/// Create a new trie with the backing database `db` and `root`
 	/// Returns an error if `root` does not exist
-	pub fn new(db: &'db HashDB, root: &'db H256) -> super::Result<Self> {
+	pub fn new(db: &'db HashDB<H>, root: &'db H::Out) -> super::Result<Self, H::Out> {
```
debris: this `super::Result` is super weird. it should be either `Result<_, _>` or `x::Result<_>`

util/patricia_trie/src/triedb.rs
rphmeier: I disagree, that would happen naturally if we had a `type TrieDB<'db> = generic::TrieDB<'db, RlpCodec, KeccakHasher>` in the `patricia-trie-ethereum` crate I suggested

rphmeier: I favor a split because we want to depend on patricia trie logic but not keccak or rlp in Polkadot.

ethcore/Cargo.toml
dvdplm: Ok, that makes sense (not sure how it'd work otherwise). The idea here was mainly to signal that there are breaking changes.

Maybe this is a good time to pause and figure out what the versioning strategy should be here. When this work started we were planning to extract `patricia-trie` into its own crate outside the parity repo. This was abandoned (for good reason), but we never devised a precise plan for sharing the code with polkadot.  I think we have two options for this.

1. Leave everything in-tree and add a new `ethereum-patricia-trie` crate as @rphmeier suggests that acts as a facade to `patricia-trie` (reexports, convenience type aliases and whatnot). Add mini-crates for `KeccakHasher` and `RlpNodeCodec`.

2. Publish crates to crates.io from inside the parity repo and let parity use `v0.1` for as long as it needs to but ideally not indefinitely – avoiding a fork is the whole point of this work; polkadot can use `v0.2`. We'd still need a facade of some kind but it could potentially live inside the `patricia-trie` crate itself temporarily.

I think option 2 sounds like a maintenance nightmare and I'd like to go with option 1 (or hear about alternatives!).

util/plain_hasher/src/lib.rs
dvdplm: Indeed there's a REVIEW question for figuring out what to do about this [here](https://github.com/paritytech/parity/pull/8739/files#diff-2109d58f1e6b9a9260dc72aa1cd58979R184).
I know Polkadot wants to use `Blake2b` and will probably be mostly 32 bytes. Lifting the 32 byte assumption seems like it requires some care but perhaps we can tackle that in a separate PR where we figure out a way to make the length configurable (conditional compilation? macro? generics?)?

util/patricia_trie/benches/trie.rs
dvdplm: True. My plan was to make a final "esthetics" pass and fix up things like this once everything is nailed down and fully working.

ethcore/light/src/on_demand/request.rs
dvdplm: Not sure I follow. `TrieError::InvalidStateRoot` here is plain wrong and only temporary to make the code compile. I honestly don't know how to deal with the generic `TrieError<T>` in higher level crates that don't use generic errors (and error_chain – which cannot have generic errors as foreign link). @rphmeier suggested in PM that we make a type alias for `TrieError` but I am not entirely sure how that's going to work.

util/patricia_trie/src/triedb.rs
debris: you're right. I read your review after I did mine :)

ethcore/light/src/on_demand/request.rs
debris: > Not sure I follow. TrieError::InvalidStateRoot here is plain wrong and only temporary to make the code compile

I see

>  I honestly don't know how to deal with the generic TrieError<T> in higher level crates that don't use generic errors (and error_chain – which cannot have generic errors as foreign link). @rphmeier suggested in PM that we make a type alias for TrieError but I am not entirely sure how that's going to work.

What about not having TrieError generic at all. Just having one of the variants of it wrap `Box<io::Error>`?

util/plain_hasher/src/lib.rs
debris: can this type be defined as associated type of Hasher that you introduced? 

```rust
trait Hasher {
  type Out: AsRef<u8>;
  type StdHasher: std::hash::Hasher;
}
```

util/patricia_trie/src/lookup.rs
dvdplm: I can't get that to work. I need:

```rust
pub struct Lookup<'a, H: Hasher + 'a, C: NodeCodec<H> + 'a, Q: Query<H>> {
	pub db: &'a HashDB<H>,
	pub query: Q,
	pub hash: H::Out,
	pub codec: &'a C,
}
```

util/plain_hasher/src/lib.rs
dvdplm: Hmm. Not sure I understand. You mean something like this?

```rust
pub type HashedFastMap<H, T> = HashMap<<H as Hasher>::Out, T, hash::BuildHasherDefault<H::StdHasher>>;
```

It feels like which (hash map) hasher to use is orthogonal to the key construction algo.

Would it be hard/impossible to refactor `PlainHasher` to be generic over the length? e.g. `PlainHasher<L>` and then pass in `KeccakHasher::LENGTH`?

util/patricia_trie/src/triedb.rs
dvdplm: I think it is needed (or rather: the compiler says it is!).

ethcore/light/src/on_demand/request.rs
rphmeier: we tried something pretty close. the issue is we need `Clone` + `Debug` for `TrieError` further up in the hierarchy and `Box<error::Error>` isn't clone

rphmeier: > From<Box<TrieError<T>>> for Error

Using the generic vesion of `TrieError` is the issue here. The _actual_ conversion we want to support is `From<Box<TrieError<H256>>>`

util/patricia_trie/src/triedb.rs
dvdplm: Sadly it doesn't. It's kind of dumb actually: seems like there's enough info in `NodeCodec<H>` to infer that the `H` param must be the same type. :(

5chdn:
Please, indicate when this is ready to be reviewed.

util/rlp/src/stream.rs
```diff
@@ -52,17 +53,74 @@ impl RlpStream {
...
+
+		// try to finish and prepend the length
+		self.note_appended(1);
+
+		// return chainable self
+		self
+	}
+
+	/// Drain the object and return the underlying ElasticArray. Panics if it is not finished.
+	fn drain(self) -> ElasticArray1024<u8> {
```
ordian: What are the reasons for using `ElasticArray` and not `smallvec`? Are there any subtle differences?

rphmeier: I know @Vurich optimized smallvec somewhat recently, but there were noticable performance regressions when we tried to switch last year.

util/hashdb/src/lib.rs
```diff
@@ -14,65 +14,61 @@
...
 	/// Remove a datum previously inserted. Insertions can be "owed" such that the same number of `insert()`s may
 	/// happen without the data being eventually being inserted into the DB. It can be "owed" more than once.
-	fn remove(&mut self, key: &H256);
+	fn remove(&mut self, key: &H::Out);
 }
 
+//TODO: investigate if we can/should get rid of this trait altogether. There used to be a `impl<T> AsHashDB for T` but that does not work with generics so we currently have concrete impls in `ethcore` for this – perhaps we can just impl the needed methods where needed and get rid of this?
 /// Upcast trait.
-pub trait AsHashDB {
+pub trait AsHashDB<H: Hasher> {
```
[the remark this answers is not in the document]
dvdplm: I like the idea a lot, but there's a cycle in there so the compiler won't have it. :/

util/rlp/src/stream.rs
dvdplm: @ordian No reason other than sticking to what the previous code did trying to keep the PR on scope. :)
I am a little worried about the slight but consistent performance regression of this PR so I'm hesitant to toss `smallvec` into the mix at this point, especially after reading about the outcome of last years attempt ([here]https://github.com/paritytech/parity/pull/4874)).

ethcore/light/src/cht.rs
```diff
@@ -26,10 +26,13 @@
 use ethcore::ids::BlockId;
 use ethereum_types::{H256, U256};
 use hashdb::HashDB;
+use keccak_hasher::KeccakHasher;
 use memorydb::MemoryDB;
 use bytes::Bytes;
-use trie::{self, TrieMut, TrieDBMut, Trie, TrieDB, Recorder};
+use trie::{TrieMut, TrieDBMut, Trie, TrieDB, Recorder};
+use ethtrie::{self, RlpCodec};
```
rphmeier: (just commenting here because it's the first file I see this in)

It can be overly verbose to rely on 3 crates in order to use `patricia-trie-ethereum`. It would be easier to use if we re-exported the API of `keccak_hasher` and `patricia-trie` from `patricia-trie-ethereum`, maybe using `pub extern crate` declarations there.

ethcore/light/src/cht.rs
```diff
@@ -66,11 +69,11 @@ impl<DB: HashDB> CHT<DB> {
 	/// Generate an inclusion proof for the entry at a specific block.
 	/// Nodes before level `from_level` will be omitted.
 	/// Returns an error on an incomplete trie, and `Ok(None)` on an unprovable request.
-	pub fn prove(&self, num: u64, from_level: u32) -> trie::Result<Option<Vec<Bytes>>> {
+	pub fn prove(&self, num: u64, from_level: u32) -> ethtrie::Result<Option<Vec<Bytes>>> {
 		if block_to_cht_number(num) != Some(self.number) { return Ok(None) }
 
 		let mut recorder = Recorder::with_depth(from_level);
-		let t = TrieDB::new(&self.db, &self.root)?;
+		let t = TrieDB::<_, RlpCodec>::new(&self.db, &self.root)?;
```
rphmeier: there ought to be a type alias for `TrieDB` and `TrieDBMut` in `patricia-trie` so we don't need explicit type annotations

ethcore/src/executed.rs
```diff
@@ -117,11 +117,21 @@ pub enum ExecutionError {
...
+	fn from(err: Box<ethtrie::TrieError>) -> Self {
+		ExecutionError::Internal(format!("{:?}", err))
 	}
 }
+impl From<ethtrie::TrieError> for ExecutionError {
+	fn from(err: ethtrie::TrieError) -> Self {
+		ExecutionError::Internal(format!("{:?}", err))
+	}
+}
+// impl<T> From<Box<trie::TrieError<T>>> for ExecutionError where T: fmt::Debug {
```
rphmeier: some commented-out code here needs to be removed

util/hashdb/src/lib.rs
```diff
@@ -14,65 +14,61 @@
...
+extern crate heapsize;
 
-use std::collections::HashMap;
 use elastic_array::ElasticArray128;
-use ethereum_types::H256;
+use heapsize::HeapSizeOf;
+use std::collections::HashMap;
+use std::{fmt::Debug, hash::Hash};
+
+pub trait Hasher: Sync + Send {
```
rphmeier: docs! on all public items (associated types and constants included)

util/journaldb/src/lib.rs
```diff
@@ -53,6 +56,32 @@ pub mod overlaydb;
 /// Export the `JournalDB` trait.
 pub use self::traits::JournalDB;
 
+use archivedb::ArchiveDB;
+impl AsHashDB<KeccakHasher> for ArchiveDB {
+	fn as_hashdb(&self) -> &HashDB<KeccakHasher> { self }
+	fn as_hashdb_mut(&mut self) -> &mut HashDB<KeccakHasher> { self }
+}
+use earlymergedb::EarlyMergeDB;
```
rphmeier: style: i'm not a huge fan of these interleaved `use` and `impl` blocks. maybe we could do this all in an embedded submodule?

ethcore/light/src/cht.rs
dvdplm: > in patricia-trie

You mean in `patricia-trie-ethereum`?

util/keccak-hasher/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+extern crate hashdb;
+extern crate ethereum_types;
+extern crate tiny_keccak;
+
+use hashdb::Hasher;
+use ethereum_types::H256;
+use tiny_keccak::Keccak;
+
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct KeccakHasher;
```
rphmeier: docs!

util/memorydb/src/lib.rs
```diff
@@ -177,19 +186,12 @@ impl MemoryDB {
...
-			Some(&(ref d, rc)) if rc > 0 => Some(d.clone()),
-			_ => None
-		}
-	}
+impl<H: Hasher> HashDB<H> for MemoryDB<H> {
 
-	fn keys(&self) -> HashMap<H256, i32> {
+	// REVIEW: this method is what made it necessary to add a type param to H256FastMap, which I'd rather have avoided.
+	//         The problem is that the keys returned are `H256` and type inference fails on the `collect()` call.
+	//         I could not make it work without parameterizing H256FastMap too.
```
rphmeier: note that `H256FastMap` is just a parameterization of `HashMap` with a custom hash-map hasher. We could allow the `Hasher` trait to specify a custom hash map hasher type and not have `MemoryDB` use `H256FastMap` directly.

util/patricia_trie/src/lookup.rs
rphmeier: assuming the trie itself has a node-codec member then it can simply borrow it

ethcore/light/src/cht.rs
rphmeier: yeah

util/patricia_trie/src/nibbleslice.rs
```diff
@@ -154,6 +156,7 @@ impl<'a> NibbleSlice<'a> {
 	}
 
 	/// Encode while nibble slice in prefixed hex notation, noting whether it `is_leaf`.
+	#[inline(always)]
```
Vurich: This function is non-trivial and so I'd probably annotate it `#[inline]` so LLVM can make the choice to inline or not itself. `#[inline(always)]` is only for really trivial functions that are really hot but you have profiled to see are not being inlined automatically (like `len` below) and should be used with care. Although if you've benchmarked this and found that the `(always)` is necessary then ignore this comment, of course!

dvdplm: It did show up in the traces and looked suspicious, but now that I try again it doesn't seem to make any difference at all so dialing back to just `#[inline]` makes sense. :)

ethcore/src/state_db.rs
```diff
@@ -309,13 +311,15 @@ impl StateDB {
 		}
 	}
 
+	// TODO: needed?
```
ordian: Used in ethcore/src/client/client.rs:1688:52.

ethcore/src/state_db.rs
```diff
@@ -309,13 +311,15 @@ impl StateDB {
...
 	/// Conversion method to interpret self as `HashDB` reference
-	pub fn as_hashdb(&self) -> &HashDB {
+	pub fn as_hashdb(&self) -> &HashDB<KeccakHasher> {
 		self.db.as_hashdb()
 	}
 
+	// TODO: needed?
```
ordian: Used in ethcore/src/client/client.rs:613:62.

util/patricia_trie/src/triedbmut.rs
```diff
@@ -334,29 +331,35 @@ impl<'a> TrieDBMut<'a> {
...
 		self.db
 	}
 
 	// cache a node by hash
-	fn cache(&mut self, hash: H256) -> super::Result<StorageHandle> {
-		let node_rlp = self.db.get(&hash).ok_or_else(|| Box::new(TrieError::IncompleteDatabase(hash)))?;
-		let node = Node::from_rlp(&node_rlp, &*self.db, &mut self.storage);
+	fn cache(&mut self, hash: H::Out) -> Result<StorageHandle, H::Out, C::Error> {
+		let node_encoded = self.db.get(&hash).ok_or_else(|| TrieError::IncompleteDatabase(hash))?;
+		// let node_encoded = self.db.get(&hash).ok_or_else(|| Box::new(TrieError::IncompleteDatabase(hash)))?;
```
ordian: obsolete?

dvdplm: No this one's actually wrong, good catch.

ethcore/src/account_db.rs
```diff
@@ -77,27 +78,32 @@ impl Factory {
...
+	pub fn from_hash(db: &'db HashDB<KeccakHasher>, address_hash: H256) -> Self {
 		AccountDB {
 			db: db,
 			address_hash: address_hash,
 		}
 	}
 }
 
-impl<'db> HashDB for AccountDB<'db>{
+impl<'db> AsHashDB<KeccakHasher> for AccountDB<'db> {
```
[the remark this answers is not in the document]
rphmeier: We can file a refactor issue but I wouldn't call this a blocker for this PR

rphmeier:
LGTM after a merge and some extra whitespace cleanup.

ethcore/src/account_db.rs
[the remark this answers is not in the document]
dvdplm: Added here: https://github.com/paritytech/parity/issues/9022

5chdn:
@dvdplm please resolve conflicts :muscle:

PR #8744, opened 2018-05-31
Custom Error Messages on ENFILE and EMFILE IO Errors

### LINKED ISSUE #6791: Improve error message: "Too many open files in system"

gnunicorn:
Add custom mapping of ENFILE and EMFILE IO Errors (Failure because of missing system resource) right when chaining ioError into ::util::Network::Error to improve Error Messages given to user

Notes:
 - Adds libc as a dependency to util/network
 - Adds assert-matches as dev-dependency
 - fixes #6791
 - superseeds #8737

### DIFF
--- a/util/network/Cargo.toml
+++ b/util/network/Cargo.toml
@@ -16,3 +16,7 @@ ipnetwork = "0.12.6"
 rlp = { path = "../rlp" }
 libc = "0.2.7"
 snappy = { git = "https://github.com/paritytech/rust-snappy" }
+
+
+[dev-dependencies]
+assert_matches = "1.2"
--- a/util/network/Cargo.toml
+++ b/util/network/Cargo.toml
@@ -14,7 +14,7 @@ ethereum-types = "0.3"
 ethkey = { path = "../../ethkey" }
 ipnetwork = "0.12.6"
 rlp = { path = "../rlp" }
-libc = "0.2.7"
+libc = "0.2"
 snappy = { git = "https://github.com/paritytech/rust-snappy" }
 
 
--- a/util/network/Cargo.toml
+++ b/util/network/Cargo.toml
@@ -14,4 +14,5 @@ ethereum-types = "0.3"
 ethkey = { path = "../../ethkey" }
 ipnetwork = "0.12.6"
 rlp = { path = "../rlp" }
+libc = "0.2.7"
 snappy = { git = "https://github.com/paritytech/rust-snappy" }
--- a/util/network/src/error.rs
+++ b/util/network/src/error.rs
@@ -15,6 +15,7 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::{io, net, fmt};
+use libc::{ENFILE, EMFILE};
 use io::IoError;
 use {rlp, ethkey, crypto, snappy};
 
@@ -83,7 +84,6 @@ impl fmt::Display for DisconnectReason {
 error_chain! {
 	foreign_links {
 		SocketIo(IoError) #[doc = "Socket IO error."];
-		Io(io::Error) #[doc = "Error concerning the Rust standard library's IO subsystem."];
 		Decompression(snappy::InvalidInput) #[doc = "Decompression error."];
 	}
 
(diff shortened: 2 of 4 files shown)


gnunicorn:
Unlike the previous PR, which attempted to handle a specific case on the higher level, when it is noticed, this is a more general approach right in the networks library: when an IO error happens, this uses some custom mapping code to wrap it into a Network-Error-Kind, with specific support for ENFILE and EMFILE errors. 

Thus, rather than having the check on the outer layer whenever the error could potentially happen (which is _anywhere_ in the code base), we've moved the improved error message right onto one of the lowest levels within the stack, populating up everywhere it might reaches the user.

util/network/Cargo.toml
```diff
@@ -14,4 +14,5 @@ ethereum-types = "0.3"
 ethkey = { path = "../../ethkey" }
 ipnetwork = "0.12.6"
 rlp = { path = "../rlp" }
+libc = "0.2.7"
```
dvdplm: I think it's fine to use `0.2` here – this code have quite unspecific version requirements.

util/network/src/error.rs
```diff
@@ -141,6 +141,34 @@ error_chain! {
 			description("Packet is too large"),
 			display("Packet is too large"),
 		}
+
+		#[doc = "We've reached system resource limits for this process"]
```
dvdplm: I for one like it more impersonal: `Reached process resource limits`. :)

util/network/src/error.rs
```diff
@@ -141,6 +141,34 @@ error_chain! {
...
+		#[doc = "We've reached system wide resource limits"]
+		SystemTooManyFiles {
+			description("Too many open files on system."),
+			display("Too many open files on system. Consider closing some processes/release some file handlers or increas the system-wide resource limits and restart parity."),
+		}
+
+		#[doc = "An unknown IO error occurred."]
+		Io(err: io::Error) {
+			description("IO Error"),
+			display("We've encoutered an IO Error: {:}", err),
```
dvdplm: typo: `encountered`

And I'd prefer `Unexpected IO error: {}`.

util/network/src/error.rs
```diff
@@ -191,3 +219,32 @@ fn test_errors() {
...
 	}
 }
+
+#[test]
+fn test_io_errors() {
+	use libc::{EMFILE, ENFILE};
+
+	let en_file_error = io::Error::from_raw_os_error(ENFILE);
+
+	match *<Error as From<io::Error>>::from(en_file_error).kind() {
```
dvdplm: Use `assert_matches` here and below? This is a little hard to read imo.

util/network/src/error.rs
```diff
@@ -141,6 +141,34 @@ error_chain! {
...
+		#[doc = "An unknown IO error occurred."]
+		Io(err: io::Error) {
+			description("IO Error"),
+			display("We've encoutered an IO Error: {:}", err),
+		}
+	}
+}
+
+impl From<::std::io::Error> for Error {
+	fn from(err: ::std::io::Error) -> Self {
```
debris: `std::io` namespace is already imported. be consistent ;)

util/network/src/error.rs
```diff
@@ -224,27 +224,21 @@ fn test_errors() {
...
-
-	let usual_error = io::Error::from_raw_os_error(0);
-
-	match *<Error as From<io::Error>>::from(usual_error).kind() {
-		ErrorKind::Io(_) => {},
-		_ => panic!("Unexpected error"),
-	}
-
+	assert_matches!(
+		<Error as From<io::Error>>::from(
```
debris: also, is there any reason to do, `Error as From<...>>::from` instead of `Error::from(...)` over here? :)

gnunicorn: Not strictly, it does work without. I started the test by copy-pasting the tests from above and kept the same explicit statement. I suspect that it was being done there (and why I left it here) to ensure - and make explicit to other people reading the code later - which code part is expected to run in for the test case.

PR #8745, opened 2018-05-31
CI: Fixes for Android Pipeline

5chdn:


.gitlab-ci.yml
```diff
@@ -190,9 +190,9 @@ android-armv7:
     - triggers
   script:
     - cargo build --target=armv7-linux-androideabi
-    - if [ $(arm-linux-androideabi-objdump -x ./target/armv7-linux-androideabi/debug/parity | grep -i 'c++_shared' | wc -l) -ne 0]; then echo "FAIL!!" fi
   tags:
     - rust-arm
+  allow_failure: true
```
tomaka: It should no longer fail if you remove the `if`, but I guess it's safer to allow failure.

PR #8751, opened 2018-05-31
Fix ancient blocks queue deadlock

tomusdrw:
A more generic solution proposed by @ngotchac is replaced with a solution that handles only ancient blocks.

The problem with previous solution was that `queue.lock().pop_front()` was actually holding the lock for longer then needed and if `fun` was trying to use the queue recursively it would deadlock (since parking_lot mutexes are not re-entrant).

Avoiding holding the lock would cause race issues and we need to be super sure that old blocks are imported in order by the same thread.

Hence I reverted to the old `IoChannelQueue` implementation and introduced a separate `import_lock` for ancient blocks. We don't lock on the queue to prevent blocking the event loop when adding new items.


ethcore/src/client/client.rs
```diff
@@ -2039,39 +2047,59 @@ impl IoClient for Client {
...
+					let hash = header.hash();
+					client.importer.import_old_block(
+						&header,
+						&block_bytes,
+						&receipts_bytes,
+						&**client.db.read(),
+						&*client.chain.read()
+					).map(|_| ()).unwrap_or_else(|e| {
+						error!(target: "client", "Error importing ancient block: {}", e);
+					});
```
niklasad1: consider to use `map_or_else` here instead of `map().unwrap_or_else`

Issue #6948, opened 2017-10-31
Add `from` and `to` to `eth_getTransactionReceipt`


mmitech:
> I'm running:
>
> - **Parity version**: 1.8.2
> - **Operating system**: Linux: Ubuntu 16.04
> - **And installed**: via binaries 

I am switching from geth to parity, I wrote my APP following the docs, but for some reason [getTransactionreceipt](https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgettransactionreceipt) seems to be missing some fields, specifically from and to which is needed in my APP.

5chdn:
The example on that page does not contain `from` and `to`, was this added recently?

tomusdrw:
`from` and `to` (and all other fields) can be retrieved from the transaction itself (by hash) adding only those two fields seems a bit weird.

simon-temple:
There is a difference between the documented response to getTransactionReceipt and the actual response:

```
 --> POST http://localhost:8545/ http/1.1
Content-Type: application/json; charset=utf-8
Content-Length: 141 
{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0x60a37c38327d3f3b2cace6ad89546895bf49c1fb6c70a4b23c9027967232ecb8"],"id":9}
 --> END POST (141-byte body)

<-- 200 OK http://localhost:8545/ (3ms)
Content-Type: application/json
Transfer-Encoding: chunked
Date: Wed, 11 Apr 2018 08:28:11 GMT
[...]

PR #8756, opened 2018-05-31
added from and to to Receipt

ab320012:
#6948
Opening a PR to track the issue. 

I am having some issues testing this. After following initial set up for macos x I start parity and then call with the method eth_getTransactionReceipt; with any transaction it returns a null result 

curl -X POST localhost:8545 -H 'content-type: application/json'  \
-d '{
    "jsonrpc": "2.0",
    "method": "eth_getTransactionReceipt",
    "params": ["0x60a37c38327d3f3b2cace6ad89546895bf49c1fb6c70a4b23c9027967232ecb8"],
    "id": 9
}'
=>
{"jsonrpc":"2.0","result":null,"id":1}

I am thinking its a config issue not sure if anyone can assist.


Tbaut:
Your node is most probably not synced. You should start parity with the `--chain dev` flag, to launch the [development chain](https://wiki.parity.io/Private-development-chain), then make a transaction (you can use [Parity UI](https://github.com/Parity-JS/shell/releases to send eth to yourself with the pre-configured account) and finally query the transaction.

ab320012:
@Tbaut thanks for the explanation. 

Running into more issues: installed parity-ui. starting parity in the following way:
./target/release/parity --chain dev --ui-no-validation

then in parity-ui project: npm start

when i route to localhost:3000 i get a loading screen:
screen shot 2018-06-01 at 7 48 28 am

Tbaut:
Never saw that. Could you install it with the binaries?

ab320012:
@Tbaut , I am able to get a response, have been getting a little stuck on the implementation, hope you don't mind me querying! I am not sure where to get the 'to' property, It looks like calling action on the transaction returns an address but its not a H160. Also, having a hard time coercing the type Option<ethereum_types::H160> to Option<v1::types::hash::H160> by calling into on the ethcore object. I think I should create a coersion method in /rpc/src/v1/types/hash for the option type but want to bounce it off of someone. Again I appreciate your time, as you can see I am a bit new to rust :)

ab320012:
commenting for gitcoin tracking - the issue is almost done just working out the failing tests

ab320012:
@5chdn  not 100% sure what to do about this last test, it intermittently says failed. The to - from logic works as far as i can see. Is there something I am missing?

vs77bb:
Thanks for the update @ab320012! Snoozed Gitcoin Bot 👍

ab320012:
@5chdn all the tests pass i think this is done

folsen:
@tomusdrw  LGTM, this doesn't seem to introduce that second DB read you were talking about, could you please give this a look?

ethcore/types/src/receipt.rs
```diff
@@ -157,6 +157,10 @@ pub struct LocalizedReceipt {
 	pub log_bloom: Bloom,
 	/// Transaction outcome.
 	pub outcome: TransactionOutcome,
+	/// Receiver address
+	pub to: Option<H160>,
```
tomusdrw: Note we now have:
```
contract_address: Option<_>,
to: Option<_>,
```
and exactly one of the two will be `Some` for every transaction.

Just making sure that what's speced.

folsen:
@tomusdrw Actually, I see what the problem is not, the documentation itself is really inconsistent.

There's: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt
That says it shouldn't be included.
And there's: https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgettransactionreceipt

That includes it, but this could be interpreted as the javascript library should be returning the fields on its own (this is obviously not how it works though).

I'm actually not sure what the right thing to do here is.

ab320012:
Why not try to update the docs for eth get transaction receipt in ethereum

Tbaut:
Need a PR for the [wiki](https://github.com/paritytech/wiki) too

Issue #8649, opened 2018-05-17
Parity node clean-up tracker ticket.


5chdn:
The next release shall not happen if we did not address the following points.

---
- [x] Remove MacOS installers, trays, etc. from code (and CI) #8875
- [x] Remove Windows installers, trays, etc. from code (and CI) #8778

---
- [x] Remove all left-overs of User Interface from code, CLI, etc. #8783
- [x] Remove all left-overs of Public Node from code, CLI, etc. #8758 (#8763)

---
- [ ] ~~Enforce consistency of crate naming, rename `ethcore-*` crates to `parity-*`~~

folsen:
I would argue the last point is not a release blocker. That is strictly behind-the-scenes and could actually be a lot of work.

tomusdrw:
Yeah, I had also a discussion with @debris where we think that `parity-*` should be reserved for potential secondary product names (like `parity-evm`, etc) and not necessarily internal crates. Most of the crates are sort of "ethereum core" and that's where the name came from (suprisingly colliding later with the company name).

debris:
Crate renaming definitely shouldn't be a release blocker. I would also reorganize location of our executables. Now they are difficult to find for people not familiar with the project. It would be cool if we place all of them in a single directory

5chdn:
Makes sense, thanks for your input.

sorpaas:
I'm going to take a look at UI and Public Node removal if no one else is working on that.

PR #8758, opened 2018-06-01
Remove public node settings from cli

sorpaas:
rel #8649.

Remove public node related settings. Keep legacy `--public-node` args and `public_node` item in config.toml to keep it backward compatible.


parity/run.rs
```diff
@@ -721,11 +720,7 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 
 	// set up dependencies for rpc servers
 	let rpc_stats = Arc::new(informant::RpcStats::default());
-	let secret_store = match cmd.public_node {
-		true => None,
-		false => Some(account_provider.clone())
-	};
-
+	let secret_store = Some(account_provider.clone());
```
ascjones: Is it possible now to remove the `Option` from `secret_store` in the rpc dependencies? Then could remove these unwraps e.g. https://github.com/paritytech/parity/blob/master/rpc/src/v1/impls/signing_unsafe.rs#L111.

ascjones:
Just some compilation errors in some mocks

5chdn:
Question: is there anything left from public node after this is merged?

sorpaas:
@5chdn This remove all things enabled by the `--public-node` flag. So I think there shouldn't be anything else left for public node.

PR #8771, opened 2018-06-03
Multiple improvements to discovery ping handling

jimpo:
This addresses some of the problems listed in https://github.com/paritytech/parity/issues/8757. The PR is rather large, so it is best reviewed commit by commit and could be split into multiple PRs if reviewers would like.

There are a few behavior changes:
- Nodes are only added to the k-buckets *after* responding to a PING
- The echo hash on PONG messages is verified against the request packet
- FIND_NODE request timeouts are handled similarly to PING timeouts
- Selection of the last seen node in a k-bucket is improved
- A node is allowed 5 consecutive request timeouts before being evicted from a k-bucket

There are more improvements to come (the k-buckets are still not limited to k entries), but I think this is a step in the right direction.

### DIFF
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -65,6 +65,7 @@ pub struct NodeBucket {
 struct PendingRequest {
 	packet_id: u8,
 	sent_at: Instant,
+	packet_hash: H256,
 }
 
 impl Default for NodeBucket {
@@ -228,8 +229,10 @@ impl Discovery {
 				let mut rlp = RlpStream::new_list(2);
 				rlp.append(&self.discovery_id);
 				append_expiration(&mut rlp);
-				self.send_packet(PACKET_FIND_NODE, &r.endpoint.udp_address(), &rlp.drain())
-					.unwrap_or_else(|e| warn!("Error sending node discovery packet for {:?}: {:?}", &r.endpoint, e));
+				match self.send_packet(PACKET_FIND_NODE, &r.endpoint.udp_address(), &rlp.drain()) {
+					Ok(_) => (),
+					Err(e) => warn!("Error sending node discovery packet for {:?}: {:?}", &r.endpoint, e),
+				}
 				self.discovery_nodes.insert(r.id.clone());
 				tried_count += 1;
 				trace!(target: "discovery", "Sent FindNode to {:?}", &r.endpoint);
@@ -282,9 +285,13 @@ impl Discovery {
 		self.public_endpoint.to_rlp_list(&mut rlp);
 		node.endpoint.to_rlp_list(&mut rlp);
 		append_expiration(&mut rlp);
-		self.send_packet(PACKET_PING, &node.endpoint.udp_address(), &rlp.drain())?;
+		let hash = self.send_packet(PACKET_PING, &node.endpoint.udp_address(), &rlp.drain())?;
 
-		let request_info = PendingRequest { packet_id: PACKET_PING, sent_at: Instant::now() };
+		let request_info = PendingRequest {
+			packet_id: PACKET_PING,
+			sent_at: Instant::now(),
+			packet_hash: hash,
+		};
 		self.expiring_pings.push_back((node.id, request_info.sent_at));
 		self.in_flight_requests.insert(node.id, request_info);
 
@@ -292,10 +299,11 @@ impl Discovery {
 		Ok(())
 	}
 
-	fn send_packet(&mut self, packet_id: u8, address: &SocketAddr, payload: &[u8]) -> Result<(), Error> {
+	fn send_packet(&mut self, packet_id: u8, address: &SocketAddr, payload: &[u8]) -> Result<H256, Error> {
 		let packet = assemble_packet(packet_id, payload, &self.secret)?;
+		let hash = H256::from(&packet[0..32]);
 		self.send_to(packet, address.clone());
-		Ok(())
+		Ok(hash)
 	}
 
 	fn nearest_node_entries(&self, target: &NodeId) -> Vec<NodeEntry> {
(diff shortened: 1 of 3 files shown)


util/network-devp2p/src/discovery.rs
```diff
@@ -119,27 +149,35 @@ impl Discovery {
...
+			self.try_ping(e);
 		}
 	}
 
 	/// Add a list of nodes. Pings a few nodes each round
-	pub fn add_node_list(&mut self, nodes: Vec<NodeEntry>) {
-		self.adding_nodes = nodes;
-		self.update_new_nodes();
+	pub fn add_node_list(&mut self, mut nodes: Vec<NodeEntry>) {
+		for node in nodes.drain(..) {
```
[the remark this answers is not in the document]
jimpo: Will do. I assume the drain can be removed from `init_node_list` as well? (which is why I did it here).

util/network-devp2p/src/discovery.rs
```diff
@@ -95,10 +121,14 @@ pub struct Discovery {
...
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
+	in_flight_requests: HashMap<NodeId, PendingRequest>,
+	expiring_pings: VecDeque<(NodeId, Instant)>,
+	expiring_finds: VecDeque<(NodeId, Instant)>,
 	send_queue: VecDeque<Datagramm>,
 	check_timestamps: bool,
 	adding_nodes: Vec<NodeEntry>,
 	ip_filter: IpFilter,
+	request_backoff: Vec<Duration>,
```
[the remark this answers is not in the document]
twittner: > It's on Discovery, not NodeBucket.

Of course, sorry. Typo.

>  And the motivation is just injecting dependencies explicitly.

I was just thinking that since `Discovery::new` refers to `const REQUEST_BACKOFF` anyway only to map its integers to `Duration`s, why not declare `REQUEST_BACKOFF` itself as a `&'static [Duration]` and only assign the reference to `Duration::request_backoff`. After all, client code can not change the backoff setting and in test modules one can still define other backoff sequences and overwrite the default. But it was just a quick thought and certainly not important.

util/network-devp2p/src/discovery.rs
```diff
@@ -119,87 +149,81 @@ impl Discovery {
...
+	pub fn add_node_list(&mut self, mut nodes: Vec<NodeEntry>) {
+		for node in nodes {
+			self.add_node(node);
+		}
 	}
 
 	/// Add a list of known nodes to the table.
 	pub fn init_node_list(&mut self, mut nodes: Vec<NodeEntry>) {
-		for n in nodes.drain(..) {
+		for n in nodes {
```
twittner: The `nodes` parameter here and in `add_node_list` does not need to be declared as `mut`.

util/network-devp2p/src/discovery.rs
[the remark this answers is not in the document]
jimpo: Added 364c8e1001c367e45693459d0002c9eacd9f3621. Let me know if you prefer that or want me to revert. Shouldn't make any difference performance-wise because Discovery is only instantiated once.

5chdn:
@twittner @tomaka please use the files > review > approve functionality to review a PR :)

util/network-devp2p/src/discovery.rs
```diff
@@ -450,22 +528,49 @@ impl Discovery {
...
 		let packets = chunks.map(|c| {
-			let mut rlp = RlpStream::new_list(1);
+			let mut rlp = RlpStream::new_list(2);
 			rlp.begin_list(c.len());
 			for n in 0 .. c.len() {
 				rlp.begin_list(4);
 				c[n].endpoint.to_rlp(&mut rlp);
 				rlp.append(&c[n].id);
 			}
+			append_expiration(&mut rlp);
```
tomaka: This may be a stupid question but isn't this a modification of the protocol?
(because if so, what about compatibility?)

jimpo: No, this should affect behavior. The change is that the expiry timestamp used to be appended in `send_packet` and now it is appended to the RLP payload before calling that method.

util/network-devp2p/src/discovery.rs
```diff
@@ -450,22 +528,49 @@ impl Discovery {
...
+				}
+				result
+			}
+			Entry::Vacant(_) => false,
+		};
+
+		if !is_expected {
+			debug!(target: "discovery", "Got unexpected Neighbors from {:?}", &from);
+			return Ok(None);
+		}
```
twittner: I don't quite understand `response_count`. If we receive two responses with 10 results each, this would only consider the first one and ignore the second response? I am probably thick, but I would be grateful if you could explain the idea a bit more. Also, would datagram duplication not interfere with `response_count`s? (Not sure how much of a problem this would be in practice.)

jimpo: Yeah, this is not really ideal, but I don't have a better solution. Since the NEIGHBORS packet does not have any ID or hash that can be used to associate it with a particular request, the best we can do is ensure there is at most one FIND_NODE in flight to any node at all times and associate the response based on sender. The sender is not supposed to send more than k=16 results to any query, so if they do, this code will ignore the second 10 result packet in your example. As far as packet duplication, if it is a concern, we could dedup the results based on node ID and track unique results in the response count.

util/network-devp2p/src/discovery.rs
```diff
@@ -450,22 +528,49 @@ impl Discovery {
...
+						request.response_count + results_count <= BUCKET_SIZE
+					{
+						request.response_count += results_count;
+						true
+					} else {
+						false
+					}
+				};
+				if entry.get().response_count == BUCKET_SIZE {
+					entry.remove();
```
twittner: If a node reports less than `BUCKET_SIZE` responses it will not have it's entry removed here and instead timeout. This will eventually cause the node from being removed altogether, no?

jimpo: Yeah, ideally the packet would have an indication that it's the end of a series of NEIGHBORS packets, but since it doesn't I think this is the best strategy. From my reading, Geth has the same behavior (https://github.com/ethereum/go-ethereum/blob/master/p2p/discover/udp.go#L329) of triggering a timeout if it receives fewer than k results (and more than k would be incorrect).

I'm assuming that this won't be too much of a problem in practice because 1) you only return <16 if there's <16 nodes in the whole routing table (which I suppose is possible on a small network) 2) there are 4 retries and if any of them are a PING, it will clear the failure count. In a future PR, it would probably be good to put something in to ensure we PING nodes that have failed a few FIND_NODE requests.

Another alternative would be to keep the PendingRequest alive but not treat it as a timeout if it clears with <16 responses. On the other hand, I think it's more likely in that case that we stay connected to a high latency node that continues to give incomplete NEIGHBORS responses without ever booting them.

twittner: > Another alternative would be to keep the PendingRequest alive but not treat it as a timeout if it clears with <16 responses.

This is what I had in mind, but I agree that the approach chosen here is probably better in practice.

Issue #8696, opened 2018-05-23
Massive increase in CPU usage with 1.11.1


folsen:
- **Which Parity version?**: 1.11.1
- **Which operating system?**: Linux
- **How installed?**: via binaries
- **Are you fully synchronized?**: yes
- **Which network are you connected to?**: ethereum
- **Did you try to restart the node?**: yes

After upgrading from latest 1.10 to 1.11.1 I use way way more CPU, while network usage is down.

![Stats](http://s.3h.se/MkdvzqYtw3.png)

SwaroopH:
Same with us. Even did a full db reset couple of days later to ensure it wasn't some background db upgrade causing the load as you can see the graph.

<img width="892" alt="screenshot 2018-05-25 13 02 41" src="https://user-images.githubusercontent.com/62628/40564047-fff2d5a4-601b-11e8-808a-2670a9c87c53.png">

AyushyaChitransh:
> I'm running:
>
> - **Which Parity version?**: v1.10.0-unstable-66755be8f-20180206/x86_64-linux-gnu/rustc1.23.0 and v1.12.0-unstable-8057e8df4-20180604/x86_64-linux-gnu/rustc1.26.1
> - **Which operating system?**: Amazon Linux
> - **How installed?**: from source
> - **Are you fully synchronized?**: yes! Fat DB and trace enabled.
> - **Which network are you connected to?**: Private network, PoA chain.
> - **Did you try to restart the node?**: so many times

**More attempts to solve this issue**: Tried following parameters: 
 - increased the CPU from 2 core to 4 core
 -  `--no-hardware-wallets` did not help.
 - `--jsonrpc-threads 1 --jsonrpc-server-threads 1`
 - `--no-ancient-blocks` also did not resolve the issue.

We are using Amazon Linux, 2 core CPU. This has been a gradual issue in our v1.10.0 versions. We received warnings for CPU usage upto 60% and then those issues transformed gradually into critical issues with CPU usage upto 104%

I am not able to pin this issue to any of these specific things:
 - a problem with parity
 - Issue with increasing blockchain size
 - Issue with server hardware configuration.

Is there any other thing/configuration which may help to resolve these CPU usage problems?

AyushyaChitransh:
In my case, I was monitoring the node using https://github.com/cubedro/eth-net-intelligence-api and as soon as I stopped the monitoring process, CPU usage came to normal in about 10 seconds. This happened three to four times.

[...]

PR #8777, opened 2018-06-04
Limit the number of transactions in pending set

tomusdrw:
With large pools and in case the node is not `--force-sealing` the pending set was actually being computed in the sync thread (propagate transactions on timeout). That could cause a significant delay, since there was nothing in the cache and the pool was pretty huge, but for propagation we never needed more than 64 first transactions.

- This PR introduces `max_len` argument to avoid constructing too large pending sets.
- Additionally we introduce `UnorderedIterator` that can be computed faster than the priority one. This for now is only used for `parity_allTransactions` RPC, but in future could be used in places where we don't really care that much about strict ordering (RPC filters maybe)

CC @folsen 

Partially related: #8696


ethcore/sync/src/api.rs
```diff
@@ -373,7 +377,9 @@ struct SyncProtocolHandler {
 impl NetworkProtocolHandler for SyncProtocolHandler {
 	fn initialize(&self, io: &NetworkContext) {
 		if io.subprotocol_name() != WARP_SYNC_PROTOCOL_ID {
-			io.register_timer(0, Duration::from_secs(1)).expect("Error registering sync timer");
+			io.register_timer(PEERS_TIMER, Duration::from_millis(700)).expect("Error registering sync timer");
```
jimpo: s/sync/peers/ in the expect message

ethcore/sync/src/api.rs
```diff
@@ -373,7 +377,9 @@ struct SyncProtocolHandler {
...
+			io.register_timer(SYNC_TIMER, Duration::from_millis(1100)).expect("Error registering sync timer");
+			io.register_timer(TX_TIMER, Duration::from_millis(1300)).expect("Error registering sync timer");
```
jimpo: The expect message should be changed probably.

miner/src/pool/queue.rs
```diff
@@ -261,15 +280,27 @@ impl TransactionQueue {
 			scoring::NonceAndGasPrice,
 			Listener,
 		>) -> T,
+	{
+		debug!(target: "txqueue", "Re-computing pending set for block: {}", block_number);
+		let _timer = ::trace_time::PerfTimer::new("pool::collect_pending");
```
jimpo: Any reason not to use the `trace_time!` macro?

miner/src/pool/queue.rs
```diff
@@ -198,45 +209,53 @@ impl TransactionQueue {
...
 
 		// Double check after acquiring write lock
 		let mut cached_pending = self.cached_pending.write();
-		if let Some(pending) = cached_pending.pending(block_number, current_timestamp, nonce_cap.as_ref()) {
+		if let Some(pending) = cached_pending.pending(block_number, current_timestamp, nonce_cap.as_ref(), max_len) {
 			return pending;
 		}
 
-		let pending: Vec<_> = self.collect_pending(client, block_number, current_timestamp, nonce_cap, |i| i.collect());
+		// In case we don't have a cached set, but we don't care about order
```
jimpo: Might be nice to test this case, and assert unordered results and that the cache is left alone.

tomusdrw: Test added.

5chdn:
Does not compile.

```
   Compiling parity-rpc v1.12.0 (file:///builds/parity/parity/rpc)
error[E0050]: method `ready_transactions` has 2 parameters but the declaration in trait `ethcore::miner::MinerService::ready_transactions` has 4
   --> rpc/src/v1/tests/helpers/miner_service.rs:211:42
    |
211 |     fn ready_transactions<C>(&self, _chain: &C) -> Vec<Arc<VerifiedTransaction>> {
    |                                             ^^ expected 4 parameters, found 2
    |
    = note: `ready_transactions` from trait: `fn(&Self, &C, usize, ethcore::miner::PendingOrdering) -> std::vec::Vec<std::sync::Arc<miner::pool::VerifiedTransaction>>`

For more information about this error, try `rustc --explain E0050`.
error: Could not compile `parity-rpc`.
```

dvdplm:
> Does not compile.

Confirmed. And while the above is easy enough to fix (`fn ready_transactions<C>(&self, _chain: &C, max_len: usize, ordering: PendingOrdering) -> Vec<Arc<VerifiedTransaction>>`) there are other errors after it; they seem to stem from the `TestBlockChainClient` and `ChainNotificationHandler` not working well together anymore: the tests assume a `new_transcations()` method that is not available.

PR #8780, opened 2018-06-04
simplify kvdb-rocksdb restore function

debris:
`fs-swap` crate: https://github.com/debris/fs-swap


debris:
pr is ready for another review. `fs-swap` is now atomic, but it may fail if:

a) osx is not using APFS (which is the default)
b) windows directories or on different volumes

in these cases we're trying to do nonatomic swap

debris:
from http://man7.org/linux/man-pages/man2/rename.2.html

> If `oldpath` is relative and `olddirfd` is the special value `AT_FDCWD`,
       then `oldpath` is interpreted relative to the current working directory
       of the calling process (like `rename()`).

If I understand this correctly, by passing `AT_FDCWD`, the paths are being resolved as in `rename()`, so exactly as we want

andresilva:
Makes sense, we want to resolve relative paths using the cwd of the process.

util/kvdb-rocksdb/src/lib.rs
```diff
@@ -590,32 +594,24 @@ impl Database {
...
-				if existed {
-					fs::remove_dir_all(&backup_db)?;
-				}
-			}
-			Err(e) => {
-				// restore the backup.
-				if existed {
-					fs::rename(&backup_db, &self.path)?;
+				// ignore errors
+				let _ = fs::remove_dir_all(new_db);
```
ordian: The old code doesn't seem to remove `new_db`, is this the intended behaviour?

ordian: Oh, looks like I misunderstood it.

debris:
already merged as a part of #8712

PR #8783, opened 2018-06-04
Remove UI related settings from CLI

sorpaas:
rel #8649. This removes several stuff:

* Remove embeddable settings in dapps that allow it to be embedded in an iframe. This is not used by UI any more, and IMO it's unsafe.
* Remove all CLI UI settings (move then to legacy), except `ui-path` -- this is actually not really related to UI. It is used for signer authcode storing.
* Deprecation warnings are removed.
* UI endpoints are removed. Please use https://github.com/parity-js/shell/pull/126 branch to test integration.


dapps/src/handlers/fetch.rs
```diff
@@ -250,12 +242,11 @@ impl ContentFetcherHandler {
 		url: &str,
 		path: EndpointPath,
 		installer: H,
-		embeddable_on: Embeddable,
 		fetch: F,
 		pool: CpuPool,
 	) -> Self {
 		let fetch_control = FetchControl::default();
-		let errors = Errors { embeddable_on };
+		let errors = Errors { };
```
tomusdrw: Could be static methods instead now.

parity/signer.rs
```diff
@@ -49,45 +48,26 @@ pub fn codes_path(path: &Path) -> PathBuf {
...
-to authorize your browser.
-Or use the generated token:
-{}"#,
-			colored(auth_url),
-			code
-		)
+Generated token:
+{}
+"#,
+			colored(code)
```
debris: do we still need it?

sorpaas: Yes, I think we still need it. This token is needed to connect to Parity UI and for times users still need to run `parity signer new-token`,

PR #8796, opened 2018-06-05
Handle removed logs in filter changes and add geth compatibility field

### LINKED ISSUE #6880: Parity doesn't return `removed` flag on filter changes
> I'm running:
>
> - **Parity version**: 1.8.1
> - **And installed**: docker images

---
I've installed a log filter and try to get filter changes.
[Standard defines](https://github.com/ethereum/wiki/wiki/JSON-RPC#returns-42), that every emited log shouls have a `removed` field, but the logs I'm getting miss this field.
```
        {
            "address": "0x71d271f8b14adef568f8f28f1587ce7271ac4ca5",
            "blockHash": "0xe8ed671b455103a24c9c0efc1204416687a2ba2af00c397c350c011202f4d75b",
            "blockNumber": "0x43772e",
            "data": "0x000000000000000000000000db790e1d4a5d7f71db60e55aa7b3af7cf9bbabf60000000000000000000000002013201070451a038d99cbbecedb879b7bdb73e8",
            "logIndex": "0x20",
            "topics": [
                "0x09e48df7857bd0c1e0d31bb8a85d42cf1874817895f171c917f6ee2cea73ec20"
            ],
            "transactionHash": "0x3ab0317c18848368581da767ce0518cac675630f722d1ef8dc385762d5c07c67",
            "transactionIndex": "0x35",
            "transactionLogIndex": "0x0",
            "type": "mined"
        }
```

sorpaas:
Fixes #6880. This PR adds two things:

1. Add `removed` field, equivalent to `log.type == "removed"`, for geth compatibility.
2. Handle reorg removed logs in filter. We do this by keeping track of the current polling block hash. Whenever log changes are requested, we check whether this block hash is on the canon chain. If not, going through the parent hashes of the tracked block hash to find the canon chain, before getting a normal filter. Light client removed logs are disabled, as I think it's expensive to fetch this information.


sorpaas:
I updated this to address https://github.com/paritytech/parity/issues/6880#issuecomment-394867201, handling removed logs in filter changes.

rpc/src/v1/impls/eth_filter.rs
```diff
@@ -52,6 +52,9 @@ pub trait Filterable {
 
 	/// Get a reference to the poll manager.
 	fn polls(&self) -> &Mutex<PollManager<PollFilter>>;
+
+	/// Get removed logs within route from the given block to the nearest canon block, not including the canon block. Also returns how many logs have been traversed.
+	fn canon_logs(&self, block_hash: H256, filter: &EthcoreFilter) -> (Vec<Log>, u64);
```
andresilva: Can we rename this to something like `removed_logs` or `reorged_logs`? Something other than `canon_logs` since we're returning logs that are not canon.

rpc/src/v1/impls/eth_filter.rs
```diff
@@ -176,6 +213,10 @@ impl<T: Filterable + Send + Sync + 'static> EthFilter for T {
 					filter.from_block = BlockId::Number(*block_number);
 					filter.to_block = BlockId::Latest;
 
+					// retrieve reorg logs
+					let (mut reorg, reorg_len) = last_block_hash.map_or_else(|| (Vec::new(), 0), |h| self.canon_logs(h, &filter));
+					*block_number -= reorg_len as u64;
```
andresilva: This doesn't affect the filter that's used below when getting pending logs. And it's overridden afterwards with `current + 1`.

sorpaas: My mistake. Filter `from_block` should be set after we alter `block_number`.

PR #8803, opened 2018-06-05
Fixed AuthorityRound deadlock on shutdown

### LINKED ISSUE #8088: Panic (deadlock) on Parity shutdown when using AuthorityRound engine
Environment: linux + branch with up-to-date master merged-in (`origin/secretstore_doc_key_via_contract`) + PoA of three nodes (configs are generated using [this script](https://github.com/svyatonik/sstore_test/blob/master/poa_integration/config_generate_list.sh)).

When trying to stop parity instances, I often see:
```bash
thread '<unnamed>' panicked at 'failed to join thread: Resource deadlock avoided (os error 35)', /checkout/src/libstd/sys/unix/thread.rs:178:12
```

The reason is that we're trying to drop IO service from IO worker thread. Here are stacks of two threads that are trying to join each other: https://gist.github.com/svyatonik/5f1251da6990df46b7033ff843b7a9ee . In this case, IO worker thread was the last thread holding reference to the `Arc<AuthorityRound>` [in this block](https://github.com/paritytech/parity/blob/58a1671076246a8478581e0508ffc624217ea242/ethcore/src/engines/authority_round/mod.rs#L705). `Arc<AuthorityRound>,` in turn was the last holder of IO service reference => when `Arc<AuthorityRound>` goes out of scope, we're sending `IoMessage::Shutdown` message to [service thread](https://github.com/paritytech/parity/blob/58a1671076246a8478581e0508ffc624217ea242/util/io/src/service.rs#L446) and start joining this thread. The [reaction to this message]() is [clearing](https://github.com/paritytech/parity/blob/58a1671076246a8478581e0508ffc624217ea242/util/io/src/service.rs#L266) and [joining](https://github.com/paritytech/parity/blob/58a1671076246a8478581e0508ffc624217ea242/util/io/src/worker.rs#L136) all the IO workers (including the one who's currently dropping the service) => deadlock, which is detected by `pthread_join`.

debris:
closes #8088

follow up for @svyatonik explanation 

memory in our event loop looks as follows:

`IoService` -> `manager_thread` -> `IoManager` -> `Workers` -> `workers_threads`

and in case of `AuthorityRound` it was additionally:

`workers_threads` -> `TransitionHandler` -> `Weak<AuthorityRound>` -> `IoService` 

everything was fine, as long as we were not trying to shutdown one of the workers when he owned the strong reference of `IoService`

as a solution, I refactored the code so `TransitionHandler` thread never has ownership of `IoService`

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -382,12 +382,16 @@ impl Decodable for SealedEmptyStep {
 	}
 }
 
+struct PermissionedStep {
+	inner: Step,
+	can_propose: AtomicBool,
+}
+
 /// Engine using `AuthorityRound` proof-of-authority BFT consensus.
 pub struct AuthorityRound {
 	transition_service: IoService<()>,
-	step: Arc<Step>,
-	can_propose: AtomicBool,
-	client: RwLock<Option<Weak<EngineClient>>>,
+	step: Arc<PermissionedStep>,
+	client: Arc<RwLock<Option<Weak<EngineClient>>>>,
 	signer: RwLock<EngineSigner>,
 	validators: Box<ValidatorSet>,
 	validate_score_transition: u64,
@@ -407,7 +411,7 @@ pub struct AuthorityRound {
 
 // header-chain validator.
 struct EpochVerifier {
-	step: Arc<Step>,
+	step: Arc<PermissionedStep>,
 	subchain_validators: SimpleList,
 	empty_steps_transition: u64,
 }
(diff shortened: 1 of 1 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -698,34 +707,37 @@ fn unix_now() -> Duration {
...
-				// has not been called fast enough.
-				// Make sure to advance up to the actual step.
-				while engine.step.duration_remaining().as_millis() == 0 {
-					engine.step();
+			// NOTE we might be lagging by couple of steps in case the timeout
+			// has not been called fast enough.
+			// Make sure to advance up to the actual step.
+			while self.step.inner.duration_remaining().as_millis() == 0 {
+				self.step.inner.increment();
+				self.step.can_propose.store(true, AtomicOrdering::SeqCst);
```
ordian: Maybe it's worth extracting into a function:
```rust
fn step(step: &PermissionedStep, client: &RwLock<Option<Weak<EngineClient>>>) {
	step.inner.increment();
	step.can_propose.store(true, AtomicOrdering::SeqCst);
	if let Some(ref weak) = *client.read() {
		if let Some(c) = weak.upgrade() {
			c.update_sealing();
		}
	}
}
```
to avoid code duplication?

rphmeier:
`Engine`'s step function might be completely unnecessary now. it's currently used for tests but we can use `TransitionHandler` in the same way. Not a blocker for this PR but we could file to refactor.

ethcore/src/engines/authority_round/mod.rs
debris: this code would also leak, cause it upgrades the reference to `EngineClient` from worker thread

PR #8822, opened 2018-06-06
CI: Fix docker tags

5chdn:
- ~~latest should be beta (was master)~~
- beta should be beta (was disabled)
- stable should be stable (was disabled)


tomusdrw:
Just to be clear - this produces only two tags now: `latest` and `stable`, will never produce `beta`.

5chdn:
That's correct @tomusdrw - I restored latest to be master.

PR #8828, opened 2018-06-06
Clearing up the reason behind "Ethereum Signed Message"

elichai:
I think this makes more sense.
First because there aren't really chosen-plaintext attacks against asymmetric encryption keys out there.
Second because faking a transaction to be signed by the user is a viable attack vector.

### DIFF
--- a/rpc/src/v1/helpers/dispatch.rs
+++ b/rpc/src/v1/helpers/dispatch.rs
@@ -237,7 +237,7 @@ pub fn fetch_gas_price_corpus(
 
 /// Returns a eth_sign-compatible hash of data to sign.
 /// The data is prepended with special message to prevent
-/// chosen-plaintext attacks.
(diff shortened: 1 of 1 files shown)


elichai:
[clabot:check].

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -237,7 +237,7 @@ pub fn fetch_gas_price_corpus(
 
 /// Returns a eth_sign-compatible hash of data to sign.
 /// The data is prepended with special message to prevent
-/// chosen-plaintext attacks.
+/// malicious DApps from using the function to sign forge transactions.
```
5chdn: forge**d** transactions?

elichai: Fixed. Don't know how it happened :+1:

elichai:
Reverted. now it's good and clean.

Issue #6852, opened 2017-10-21
Identity Parameter no longer working


MysticRyuujin:
I've been using the `--identity=` for many releases now but it looks like somewhere around v1.7.6 or v1.7.7 (possibly earlier...) the parameter stopped being passed in the node information...

I was on v.1.7.7 earlier today when I noticed it, I updated to `Parity/v1.8.1-beta-0e4a06d-20171020/x86_64-linux-gnu/rustc1.21.0` and it still seems to be missing when I look at it in my peer list...

I can identify my own node by the use of the internal private IP but the name no longer includes my identity specified in the parameter

Office-Julia:
Hello. Could you please share the following information:
1. Which version of Parity are you running? ("master" or "git" is not a version)
2. Which operating system(s) are you running?
3. How did you install Parity? (binary, installer, from source, which branch?)
4. What flags are you running Parity with? (including passed --identity= value)
5. Do you use any custom config file?
6. Please, share the exact error message (if you get any), or provide logs

MysticRyuujin:
1. `Parity/v1.8.1-beta-0e4a06d-20171020/x86_64-linux-gnu/rustc1.21.0`

2. Ubuntu 16.04.3 x64

3. Official deb package

4. --identity MysticRyuujin-PARITY01 --ui-interface <redacted> --ui-hosts all --port 30302 --nat extip:$(curl -s ipinfo.io/ip) --reserved-peers /root/parity-reserved-peers --jsonrpc-interface <redacted> --jsonrpc-hosts all --ws-interface <redacted> --ws-origins all --ws-hosts all --no-secretstore

5. No

6. No errors

rphmeier:
Probably related to #2066. @MysticRyuujin can you try with the identity in quotes and can you clarify whether you run with `--identity=` or just `--identity`? the parser might not handle those cases exactly the same.

MysticRyuujin:
So my CLI options are exactly as above, just `--identity` with a space then the string. I modified the startup script to include quotes around the string but that did not change the behavior and in fact the running process shows no quotes (still) which makes sense as the bash interpreter is consuming the quotes anyway.

```
chase@PARITY01:~$ ps -ef | grep parity
root        993    963 59 08:18 ?        00:00:19 /usr/bin/parity --identity MysticRyuujin-PARITY01
```

Shortened for brevity

I then tested `--identity=MysticRyuujin-PARITY01` which results in:

```
[...]

PR #8830, opened 2018-06-06
Include node identity in the P2P advertised client version.

jimpo:
Fixes https://github.com/paritytech/parity/issues/6852.

Geth client versions are "Geth/NAME/VERSION/COMPILER", whereas this PR puts name at the front, like "NAME/Parity/VERSION/COMPILER". This is out of convenience and because I think it makes more sense that way, but I could change it to go after Parity if people prefer.

### DIFF
--- a/parity/configuration.rs
+++ b/parity/configuration.rs
@@ -754,7 +754,15 @@ impl Configuration {
 		ret.config_path = Some(net_path.to_str().unwrap().to_owned());
 		ret.reserved_nodes = self.init_reserved_nodes()?;
 		ret.allow_non_reserved = !self.args.flag_reserved_only;
-		ret.client_version = version();
+		ret.client_version = {
+			let mut client_version = version();
+			if !self.args.arg_identity.is_empty() {
+				// Insert name after the "Parity/" at the beginning of version string.
+				let idx = client_version.find('/').unwrap_or(client_version.len());
+				client_version.insert_str(idx, &format!("/{}", self.args.arg_identity));
+			}
+			client_version
+		};
 		Ok(ret)
 	}
 
@@ -1783,6 +1791,19 @@ mod tests {
 			Cmd::Run(c) => {
 				assert_eq!(c.net_conf.min_peers, 99);
 			},
+			_ => panic!("Should be Cmd::Run"),
+		}
+	}
+
+	#[test]
+	fn test_identity_arg() {
+		let args = vec!["parity", "--identity", "Somebody"];
+		let conf = Configuration::parse_cli(&args).unwrap();
+		match conf.into_command().unwrap().cmd {
+			Cmd::Run(c) => {
+				assert_eq!(c.name, "Somebody");
+				assert!(c.net_conf.client_version.starts_with("Parity/Somebody/"));
+			}
 			_ => panic!("Should be Cmd::Run"),
 		}
 	}


jimpo:
@5chdn By node key, do you mean the P2P network authentication key? There already is a separate `--node-key` flag for that. The help string for `--identity` is `Specify your node's name. (default: )`.

folsen:
We already read the `identity` flag into `NetworkSettings.name` but version is set in `NetworkConfiguration`. Do we have the `NetworkSettings` available where we actually report the version string, and if so should we use that rather than duplicating the arg to more places?

I'm slightly weary of doing this since I don't know what assumptions have been built around having a specific `client_version` on `NetworkConfiguration`. Maybe @arkpar can chime in?

jimpo:
@folsen `NetworkSettings` is for RPC settings, whereas `NetworkConfiguration` is for devp2p, which is why I duplicated the name into both.

Regarding assumptions about `client_version`, definitely would like some review from someone who can answer that. As far as I can tell, it is only exposed in the `Hello` packet to peers and RPC endpoints returning peer information.

andresilva:
In #6852 it is claimed that this used to work (apparently with "Parity/NAME/.." format). But I checked out some old v1.7 tags and the code is similar to what we have now, just using `version()`.

dvdplm:
It seems like @arkpar s [changes](https://github.com/paritytech/parity/commit/dde6baedeca4b2dccaf2eacfbd2453b3d191596e) are not directly related to this?

I'm having a hard time following along in the code as there are plenty of `version` variants floating around, but I agree with @jimpo and @andresilva that this will solve the immediate issue.

I'm a little wary of adding more logic (albeit simple) to handle the version string, e.g. it's not obvious that [`peer_client_version()`](https://github.com/paritytech/parity/blob/13bc922e546351e962c414b8adcf59eeb1041753/util/network-devp2p/src/host.rs#L179) uses the same info (or does it?). 

Ideally the fix should be in [`version()`](https://github.com/paritytech/parity/blob/master/util/version/src/lib.rs#L52) but that might have implications that I don't see.

5chdn:
Seems to work.

```
curl --data '{"method":"parity_nodeName","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545

{"jsonrpc":"2.0","result":"AfriTestsPr#8830-LoremIpsum","id":1}
```

> This is out of convenience and because I think it makes more sense that way, but I could change it to go after Parity if people prefer.

Could you put it after the `Parity/`? I think this would be more consistent. 

> had a look and it seems as @jimpo suggested is only used in the Hello packet (after this PR) and in the parity_nodeName RPC. 

Should this also be part of the official version string? I.e.,:

```
Parity/AfriTestsPr#8830-LoremIpsum/v1.12.0-nightly-9bd4bd6bb-20180613/x86_64-linux-gnu/rustc1.26.2
```

(Maybe limited to 8 chars)

andresilva:
@5chdn I don't think we should add it to the version string, only peer announces. It's *your* node but it's *our* version 😉.

@dvdplm I think we can add a `extra: Option<&str>` to the `version` function and then pass the `identity` when computing `client_version`. This is what @jimpo was trying to avoid by putting the name as a prefix.

dvdplm:
> add a extra: Option<&str> to the version function

I think this would be wise.

andresilva:
@folsen I meant more for things like `parity -v`. The geth nodes in ethstats.net don't seem to be showing any names (maybe could be that there are no nodes registered there with names). The parity nodes show the version string with a "//" e.g. "Parity//v1.8.12-unstable/x86_64-linux-gnu/rustc1.26.2", which is done [here](https://github.com/paritytech/parity/blob/master/rpc/src/v1/impls/web3.rs#L34) when implementing the `client_version` RPC. If we want to get the name in ethstats we also need to get the identity string into that method.

jimpo:
@dvdplm `peer_client_version` returns the version string of the remote peer in a session, not the one we advertise.

I don't think the `version()` method should return with the same -- it's the version. But I can definitely modify the client_version to put the name after Parity.

5chdn:
@dvdplm @andresilva @folsen needs a 2nd review

PR #8832, opened 2018-06-07
ethcore: fix ancient block import error handling

andresilva:
The existing code was discarding the error and printing the error message on every successful ancient block import.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -2073,15 +2073,16 @@ impl IoClient for Client {
 				let first = queued.write().1.pop_front();
 				if let Some((header, block_bytes, receipts_bytes)) = first {
 					let hash = header.hash();
-					client.importer.import_old_block(
+					let result = client.importer.import_old_block(
 						&header,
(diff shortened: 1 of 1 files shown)


ethcore/src/client/client.rs
```diff
@@ -2073,15 +2073,16 @@ impl IoClient for Client {
 				let first = queued.write().1.pop_front();
 				if let Some((header, block_bytes, receipts_bytes)) = first {
 					let hash = header.hash();
-					client.importer.import_old_block(
+					let result = client.importer.import_old_block(
```
niklasad1: Just to clarify that understand this correct, essentially `map_or` is used the inverse way here right?

Prints a error message when the Result is `Ok` and don't care when it is `Err`?

andresilva: Yes, exactly. We could change only `.ok()` to `.err()` and it would work, but I thought it was clearer using `if let Err`.

niklasad1: Yeah, that is fine!

PR #8834, opened 2018-06-07
Disable parallel verification and skip verifiying already imported txs.

tomusdrw:
Related: #8696


miner/src/pool/queue.rs
```diff
@@ -179,8 +178,14 @@ impl TransactionQueue {
 
 		let verifier = verifier::Verifier::new(client, options, self.insertion_id.clone());
 		let results = transactions
-			.into_par_iter()
-			.map(|transaction| verifier.verify_transaction(transaction))
+			.into_iter()
+			.map(|transaction| {
+				if self.pool.read().find(&transaction.hash()).is_some() {
+					bail!(transaction::Error::AlreadyImported)
+				}
```
ascjones: Was the primary cause of the high CPU usage the parallel verification or verifying already imported txs. Or a combination of both?

tomusdrw: So it seems like a combination of even more things, this brings the load down, but not 100% to the level of the one of 1.10. I'm still figuring out other culprits.

miner/src/pool/queue.rs
```diff
@@ -179,8 +178,14 @@ impl TransactionQueue {
 
 		let verifier = verifier::Verifier::new(client, options, self.insertion_id.clone());
 		let results = transactions
-			.into_par_iter()
-			.map(|transaction| verifier.verify_transaction(transaction))
+			.into_iter()
```
folsen: @tomusdrw Did you ever look at the size of the work for this and how `par_iter()` distributes it? Like if it's work-stealing it might actually cause more overhead to parallelise if the work-pieces are small enough (In some rayon examples for instance he shows a 0.95x speedup (so it's much worse) on a 4 core machine). It might be worth testing with rayons [`Chunks`](https://docs.rs/rayon/1.0.1/rayon/iter/struct.Chunks.html) iterator to chunk it up into larger work-pieces for each CPU to tackle.

PR #8844, opened 2018-06-08
Use sealing.enabled to emit eth_mining information

### LINKED ISSUE #7939: Mining false on authority nodes in POA chain
> I'm running:
>
> - **Which Parity version?**: v1.8.4-beta-c74c8c1-20171211/x86_64-linux-gnu/rustc1.22.1
> - **Are you fully synchronized?**: NA
> - **Which network are you connected to?**: Private
> - **Did you try to restart the node?**: NA

### Command:

    curl --data '{"method":"eth_mining","params":[],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545

### Expected Behavior
`eth_mining` returns true if validator node is actively issuing blocks.

Output:

> {"jsonrpc":"2.0","result":true,"id":1}

### Actual behavior

`eth_mining` returns false for authority nodes.

Output:

> {"jsonrpc":"2.0","result":false,"id":1}

sorpaas:
Fixes #7939. I think `sealing.enabled` is a more reliable way to get info on whether the node is mining. Tested against internal sealing mining/non-mining and PoW mining/non-mining clients. Looks like working.

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -1301,4 +1301,22 @@ mod tests {
 
 		assert!(!miner.is_currently_sealing());
 	}
+
+	#[test]
+	fn should_mine_if_fetch_work_request() {
+		struct DummyNotifyWork;
+
+		impl NotifyWork for DummyNotifyWork {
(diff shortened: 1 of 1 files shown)


folsen:
Is there any way to add tests for this?

PR #8853, opened 2018-06-11
Check whether we need resealing in miner and unwrap has_account in account_provider

sorpaas:
This PR does two things:

* Unwrap the result in `AccountProvider::has_account`. We never return Err for that.
* Check whether resealing is needed before update sealing for external transactions. We do this check for local transactions already.


sorpaas:
Ooops!

PR #8854, opened 2018-06-11
Fix concurrent access to signer queue

### LINKED ISSUE #8799: Fix concurrent access to signer queue
Currently it's possible that two parallel RPC requests to with password to sign a signer queue entry will cause two transactions to be sent to the network.

The issue exists since we are running a multi-threaded RPC, and the signer queue doesn't have explicit lock while accessing an item.

Code: https://github.com/paritytech/parity/blob/7d7d4822a5087b24b49f2a1ddb111f2daa51a017/rpc/src/v1/impls/signer.rs#L95

So we `peek` some ID, do signing and later removed it from the queue. If nonce is not provided explicitly both threads will get a unique one and cause two same transactions to be sent.

Related #8646

sorpaas:
Fixes #8799.

`SignerQueue::peek` is changed to `SignerQueue::take`, which takes the request out of the queue. Subsequent redundant confirmations will not find the sender. The caller then takes the request as the handle, and gives it back to `request_confirmed` or `request_rejected` to notify senders.

### DIFF
--- a/rpc/src/v1/helpers/signing_queue.rs
+++ b/rpc/src/v1/helpers/signing_queue.rs
@@ -77,14 +77,14 @@ pub trait SigningQueue: Send + Sync {
 
 	/// Removes a request from the queue.
 	/// Notifies possible token holders that request was rejected.
-	fn request_rejected(&self, id: U256) -> Option<ConfirmationRequest>;
+	fn request_rejected(&self, sender: ConfirmationSender) -> Option<ConfirmationRequest>;
 
 	/// Removes a request from the queue.
 	/// Notifies possible token holders that request was confirmed and given hash was assigned.
-	fn request_confirmed(&self, id: U256, result: ConfirmationResult) -> Option<ConfirmationRequest>;
+	fn request_confirmed(&self, sender: ConfirmationSender, result: ConfirmationResult) -> Option<ConfirmationRequest>;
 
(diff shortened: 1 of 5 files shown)


5chdn:
3 tests failed.

rpc/src/v1/helpers/signing_queue.rs
```diff
@@ -193,22 +190,26 @@ impl SigningQueue for ConfirmationsQueue {
...
+	fn request_rejected(&self, sender: ConfirmationSender) -> Option<ConfirmationRequest> {
+		debug!(target: "own_tx", "Signer: Request rejected ({:?}).", sender.request.id);
+		self.notify_result(sender, None)
 	}
 
-	fn request_rejected(&self, id: U256) -> Option<ConfirmationRequest> {
-		debug!(target: "own_tx", "Signer: Request rejected ({:?}).", id);
-		self.remove(id, None)
+	fn request_confirmed(&self, sender: ConfirmationSender, result: ConfirmationResult) -> Option<ConfirmationRequest> {
+		debug!(target: "own_tx", "Signer: Transaction confirmed ({:?}).", sender.request.id);
```
tomusdrw: It was `Request` since it's not necessarily a transaction (can be `eth_sign` request, or `parity_decrypt` request).

PR #8857, opened 2018-06-11
scripts: hardcode number of build jobs 😅🚀

5chdn:
- set default number of build jobs to 8
- update submodules before running tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -189,7 +189,7 @@ android-armv7:
     - stable
     - triggers
   script:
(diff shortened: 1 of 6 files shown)


sorpaas:
@5chdn According to [Cargo documents](https://doc.rust-lang.org/cargo/reference/config.html), I think the default number of jobs is the number of CPUs? So not sure whether adding `-j 8` is necessary.

5chdn:
Interesting. Why is `cargo build` so much slower than `cargo build -j $(nproc)` then?

PR #8864, opened 2018-06-11
README.md: update the list of dependencies

shdown:
`file` and `make` are also required.


shdown:
[clabot:check]

5chdn:
Thanks, while we are at it, can someone confirm we still need libssl?

debris:
> Thanks, while we are at it, can someone confirm we still need libssl?

Good questions. If we don't need `libssl`, we should also update our docker files in a separate pr

PR #8878, opened 2018-06-12
docker: Fix alpine build

andresilva:
- Switch to alpine edge for recent rustc version
- Remove openssl libs
- Add perl (needed for building ring)


andresilva:
@debris Logged in #8880.

PR #8879, opened 2018-06-12
Eliminate some more `transmute()`

niklasad1:


util/network-devp2p/src/ip_utils.rs
```diff
@@ -236,16 +236,15 @@ mod getinterfaces {
...
 				(IpAddr::V6(Ipv6Addr::new(
-					addr[0],
-					addr[1],
-					addr[2],
-					addr[3],
-					addr[4],
-					addr[5],
-					addr[6],
-					addr[7])),
+					((addr[1] as u16) << 8) | (addr[0] as u16),
```
niklasad1: `addr` is actually a [u8; 16] so no need to transmute it but requires some shifting to merge two bytes into a word (two bytes)

util/network-devp2p/src/ip_utils.rs
```diff
@@ -236,16 +236,15 @@ mod getinterfaces {
 				let sa: *const sockaddr_in6 = sa as *const sockaddr_in6;
 				let sa = & unsafe { *sa };
 				let (addr, port) = (sa.sin6_addr.s6_addr, sa.sin6_port);
```
debris: what is the type of `addr` here?

niklasad1: It is u32 but I saw that this was actually from the `std::net`  and it actually implements `from trait` with this so I changed it! Makes the code more readable IMO!

ethash/src/compute.rs
```diff
@@ -137,11 +137,11 @@ fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64)
 		($n:expr, $value:expr) => {{
 			// We use explicit lifetimes to ensure that val's borrow is invalidated until the
 			// transmuted val dies.
-			unsafe fn make_const_array<'a, T, U>(val: &'a mut [T]) -> &'a mut [U; $n] {
+			unsafe fn make_const_array<'a, T, U>(val: &mut [T]) -> &mut [U; $n] {
```
andresilva: Drop `unsafe`?

ethash/src/compute.rs
```diff
@@ -268,10 +268,10 @@ fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64)
 
 	let value: H256 = unsafe {
```
andresilva: Drop `unsafe`?

util/network-devp2p/src/ip_utils.rs
```diff
@@ -225,28 +225,13 @@ mod getinterfaces {
 				let sa: *const sockaddr_in = sa as *const sockaddr_in;
 				let sa = unsafe { &*sa };
 				let (addr, port) = (sa.sin_addr.s_addr, sa.sin_port);
-				(IpAddr::V4(Ipv4Addr::new(
-					(addr & 0x0000_00FF) as u8,
-					((addr & 0x0000_FF00) >>  8) as u8,
-					((addr & 0x00FF_0000) >> 16) as u8,
-					((addr & 0xFF00_0000) >> 24) as u8)),
-					port)
+				(IpAddr::V4(Ipv4Addr::from(addr)), port)
```
andresilva: Seems like we're encoding the octets in reverse order compared to `std`.

niklasad1: Ok, I will check but I assumed the tests actually tested this?!

ethash/src/compute.rs
niklasad1: `&mut *(val.as_mut_ptr() as *mut [U; $n])` is also unsafe because `deref` of raw pointer so I would like to keep the function as `unsafe`! Thoughts?

ethash/src/compute.rs
andresilva: Nevermind, `keccak_256::unchecked` is unsafe.

ethash/src/compute.rs
andresilva: Yeah won't compile without `unsafe` ignore my comment.

util/network-devp2p/src/ip_utils.rs
```diff
@@ -225,28 +225,13 @@ mod getinterfaces {
...
-					addr[0],
-					addr[1],
-					addr[2],
-					addr[3],
-					addr[4],
-					addr[5],
-					addr[6],
-					addr[7])),
-					port)
+				(IpAddr::V6(Ipv6Addr::from(addr)), port)
```
niklasad1: Yeah, this is wrong too!

util/network-devp2p/src/ip_utils.rs
niklasad1: Test: https://play.rust-lang.org/?gist=b26a380a7d08aa4b781508fc6d9f895e&version=stable&mode=debug

util/network-devp2p/src/ip_utils.rs
```diff
@@ -225,28 +225,17 @@ mod getinterfaces {
...
 				let (addr, port) = (sa.sin_addr.s_addr, sa.sin_port);
-				(IpAddr::V4(Ipv4Addr::new(
-					(addr & 0x0000_00FF) as u8,
-					((addr & 0x0000_FF00) >>  8) as u8,
-					((addr & 0x00FF_0000) >> 16) as u8,
-					((addr & 0xFF00_0000) >> 24) as u8)),
-					port)
+				// convert u32 to an `Ipv4 address`, but the u32 must be converted to `host-order`
+				// that's why `from_be` is used!
+				(IpAddr::V4(Ipv4Addr::from(<u32>::from_be(addr))), port)
```
niklasad1: This should now work both on `little endian` and `big-endian` machines!

ethash/src/compute.rs
rphmeier: looks like the explicit lifetimes have been dropped here. should be `&'a mut [T] -> &'a mut [U; $n]` right?

util/network-devp2p/src/ip_utils.rs
rphmeier: why wouldn't it before? The IPv6 code looks endian-reliant but simple integer operations in memory should be abstracted over endianness by LLVM

ethash/src/compute.rs
niklasad1: @rphmeier Explicit `lifetimes` here are not needed here, so I removed them but forgot `a` in the signature!

I can revert this if you want explicit lifetime annotations!

util/network-devp2p/src/ip_utils.rs
```diff
@@ -225,28 +225,17 @@ mod getinterfaces {
...
-					addr[2],
-					addr[3],
-					addr[4],
-					addr[5],
-					addr[6],
-					addr[7])),
-					port)
+				let ip_addr = Ipv6Addr::from(addr);
+				debug_assert!(addr == ip_addr.octets());
+				(IpAddr::V6(ip_addr), port)
```
niklasad1: I'm pretty convinced that this was a bug because 

`[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1] // ::1`

was parsed as `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0]`

Also with u8, endianness doesn´t matter

util/network-devp2p/src/ip_utils.rs
niklasad1: Oops, a nit from my side. You are right Rob!

PR #8882, opened 2018-06-13
Allow disabling local-by-default for transactions with new config entry

### LINKED ISSUE #8820: Option to not mark transactions as local when received over RPC
Currently the node will mark transactions submitted using `eth_sendRawTransaction` as local ones.

This means that such transactions will have a higher priority and also can go below configured minimal gas price.
Such situation can cause issues for public-facing nodes that expose `eth` namespace.

Solution:
1. Either allow disabling RPC methods name-by-name, so that you can choose to expose `eth` but not `eth_sendRawTransaction`
2. Or add a CLI flag to disable special treating of RAW transactions received over RPC

XertroV:
Fix #8820 

Transactions submitted through `import_own_transaction` are checked to see if we have the account (provided the `tx_queue_allow_unknown_local` flag is false).

The exact logic is:

``` rust
let treat_as_local = self.options.tx_queue_allow_unknown_local 
		|| self.accounts.and_then(|accts| accts.has_account(sender).ok()).unwrap_or(false);
```

BTW - I've never written Rust before, so please check this twice to make sure it is any good.

I'm also not sure if I should add any tests and/or where.


XertroV:
@5chdn - Do you guys want this to be squashed into one commit? I know some communities like that.

(Note: asking you because you just added the tag)

5chdn:
Yes, we squash-merge :)

XertroV:
image

XertroV:
Sweet - I think we're good to go. Thanks for being super on top of labels and such @5chdn. Parity really is one of the most functional and responsive github orgs I've ever come across. Massive props 👏.

Notes:

- At one stage in dev I explicitly returned from `import_own_transactions` but refactored to a bigger `if` statement so there are no explicit returns. Not sure what y'all preferences are in this regard, but this is cleaner to me (coming from purescript/haskell)
- This feels cleaner than the two options in #8820 and the behaviour is clearer / more predictable than either of those options.
- I chose to act exactly like `import_external_transactions` in the case we don't treat it as local to avoid any new or undefined behaviour.
- I think the issue arose because many moons ago the assumption was that transactions either came from the network or a trusted source (e.g. eth_sendRawTransaction).
- I'd like to see the default on `tx_queue_allow_unknown_local` be changed to `false` in an upcoming release. It's safer as the default and isn't likely to affect many people b/c it checks for local accounts. The edge case is people who submit low-fee transactions (which they want the node to keep) from an external source. I'm not sure if a watch-only wallet would solve that (though it looks like `AccountProvider` doesn't support that atm).
- It'd be nice to have some docs that could prevent new contributors going through the growing pains I did above.
- I was pleasantly surprised by Rust, having never used it before. Despite some unintuitive things it feels like a very productive and capable language (and ❤️ for anything monadic, like `Option` and `Result`). `do` syntax would be amazing if it doesn't exist yet.

*Bonus fun fact*: this is the first work I've done relating to anything core to Ethereum since *February 2014*.

5chdn:
> Parity really is one of the most functional and responsive github orgs I've ever come across. Massive props clap.

Haha, wait till you (eventually) get a _review_. :rofl:

ethcore/src/miner/miner.rs
```diff
@@ -799,21 +802,34 @@ impl miner::MinerService for Miner {
...
+            // | NOTE Code below requires sealing locks.                                |
+            // | Make sure to release the locks before calling that method.             |
+            // --------------------------------------------------------------------------
+            if imported.is_ok() && self.options.reseal_on_own_tx && self.sealing.lock().reseal_allowed() {
+                self.prepare_and_update_sealing(chain);
+            }
+
+
+			imported
+		} else {
```
sorpaas: I would suggest we:

* Leave `import_own_transaction` unchanged.
* Create another function, probably called `import_local_or_external_transaction` (or maybe a better name, can't think of right now), which dispatch to `import_own_transaction` or `import _external_transactions`.
* Make `eth_submitRawTransaction` call that function instead, through `Dispatcher`.

The issue is that `import_own_transaction` is used in many other places in `ethcore`, and this change may break some of the previous assumptions.

ethcore/src/miner/miner.rs
```diff
@@ -799,21 +802,34 @@ impl miner::MinerService for Miner {
...
+			// We want to replicate behaviour for external transactions if we're not going to treat
+			// this as local. This is important with regards to sealing blocks
+			self.import_external_transactions(chain,	vec![pending.transaction.into()])
```
sorpaas: Should be `<whitespace>` not `<tab>` right before `vec![`.

ethcore/src/miner/miner.rs
```diff
@@ -1146,37 +1162,44 @@ mod tests {
 		assert!(miner.submit_seal(hash, vec![]).is_ok());
 	}
 
+	fn  default_miner_opts() -> MinerOptions {
```
sorpaas: Extra whitespace after `fn`.

ethcore/src/miner/miner.rs
```diff
@@ -1146,37 +1162,44 @@ mod tests {
...
-				tx_queue_strategy: PrioritizationStrategy::GasPriceOnly,
-				refuse_service_transactions: false,
-				pool_limits: Default::default(),
-				pool_verification_options: pool::verifier::Options {
-					minimal_gas_price: 0.into(),
-					block_gas_limit: U256::max_value(),
-					tx_gas_limit: U256::max_value(),
-				},
-			},
+			default_miner_opts(),
```
sorpaas: Why having an extra `default_miner_opts`? I think this may be confusing as `MinerOptions` already implements `Default` trait.

parity/cli/mod.rs
```diff
@@ -712,6 +712,10 @@ usage! {
 			"--tx-queue-strategy=[S]",
 			"Prioritization strategy used to order transactions in the queue. S may be: gas_price - Prioritize txs with high gas price",
 
+			ARG arg_tx_queue_allow_unknown_local: (bool) = true, or |c: &Config| c.mining.as_ref()?.tx_queue_allow_unknown_local.clone(),
```
sorpaas: Maybe change this to a flag, as it is a boolean?

ethcore/src/miner/miner.rs
XertroV: Yup, it is confusing. The config returned is the default used in testing (which is not Default:: default()).

probs better to extract config from the miner produced by miner()

parity/cli/mod.rs
XertroV: I wanted the default (for the flag/art) to be true, which I can't do with a flag (seems like default must be false?) Also, unknown_local was the simplest name I could come up with that also explains the concept, and keeps all the flags/config items the same name (not negations). Length was also a cosideration. If there's a better name happy to change it.

(Like: allow_unknown_local is better than disallow_unknown_local (or deny_...), less thinking to do about what it means. and unknown_local is not great to start with, so already takes some thinking. Not super happy with the name, tbh)

ethcore/src/miner/miner.rs
XertroV: Yup, agreed. Much more elegant . Will get on that tomorrow.

ethcore/src/miner/miner.rs
XertroV: Alternative is to rename it to something like `std_test_miner_opts`

ethcore/src/miner/miner.rs
XertroV: How does `import_foreign_transaction` sound? Or maybe just `import_unknown_local_transaction`, which corresponds to the argument name very well. 

On the one hand introducing a new term is not good, but coming up with a good name is hard. Maybe `normalise_unknown_local`? like mixing them in with external instead of treating like known locals. I think the External/Local thing makes sense and should be left how it is (no redefining what a local tx is). (though maybe better names would be Indirect/Direct transactions (either that  you get them second-hand or directly.)

XertroV:
@sorpaas - The best name I think I've come up with is something like `demote_unfamiliar_locals` for the flag name, and `import_local_transaction` for the import function. By the definition of `local` the txs we get over `eth_sendRawTransaction` _are_ local, we just don't want to prioritise the ones that don't come from _us_. These names still feel clear, and `import_own_transaction` actually does mean our _own_ transaction now.

ethcore/src/miner/miner.rs
sorpaas: Both sound fine to me. :)

sorpaas:
In our current transaction pool implementation (in `miner/src/pool/verifier.rs`) we call transactions handled by `import_own_transaction` to be `Transaction::Local(...)`, and `import_external_transaction` to be `Transaction::Unverified(...)`. So I guess we may not want to overload the term "local" too much in our code.

But I think that's just a minor naming issue and both sound fine to me. :)

parity/cli/mod.rs
sorpaas: We have many "negative" flags already in cli, like `--no-warp`, `--no-discovery`, etc. And I searched the code, we don't have any boolean arg. So I still think that flag may be better for consistency, like `--disallow-unknown-local`/`--demote-unkown-local-transactions` or something like that.

Or maybe [like you said](https://github.com/paritytech/parity/pull/8882#issuecomment-396882476), we can set `--allow-unknown-local` default to false? I can't think of any case where this may break existing usages -- `eth_submitRawTransaction` specification doesn't require giving special promotion to the submitted transaction.

XertroV:
@sorpaas - Fixed up as per above. Note I went with `tx_queue_no_unfamiliar_locals` as the flag/config name, and `import_claimed_local_transaction` as the function name. These feel quite solid as names to me.

XertroV:
So I seem to be getting this error the last two tests that have run:

```
fatal: cannot create directory at 'BlockchainTests/GeneralStateTests/stQuadraticComplexityTest': No space left on device
Cloning into 'ethcore/res/wasm-tests'...
fatal: could not write to /builds/parity/parity/ethcore/res/wasm-tests/.git: No space left on device
fatal: clone of 'https://github.com/paritytech/wasm-tests' into submodule path 'ethcore/res/wasm-tests' failed
```

Ideas?

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -126,7 +126,10 @@ impl<C: miner::BlockChainClient, M: MinerService> FullDispatcher<C, M> {
 	pub fn dispatch_transaction(client: &C, miner: &M, signed_transaction: PendingTransaction) -> Result<H256> {
 		let hash = signed_transaction.transaction.hash();
 
-		miner.import_own_transaction(client, signed_transaction)
+		// use `import_claimed_local_transaction` so we can decide (based on config flags) if we want to treat
+		// it as local or not. Nodes with public RPC interfaces will want these transactions to be treated like
+		// external transactions.
+		miner.import_claimed_local_transaction(client, signed_transaction)
```
sorpaas: @XertroV Can we only use `import_claimed_local_transaction` for `eth_sendRawTransaction`? I'm worried about a race condition where:

* A new transaction, from a vault account, is sent through `eth_sendTransaction`.
* Before the dispatcher gets a chance, the vault is closed, and thus the account is removed from address list.
* The dispatcher gives the transaction to miner. Miner doesn't find it in the account list, so it's imported as external transaction -- clearly this is not what we want.

A simple way may be just to add a boolean param in `dispatch_transaction`, and only set that to true for `eth_sendRawTransaction`.

rpc/src/v1/tests/helpers/miner_service.rs
```diff
@@ -169,6 +169,21 @@ impl MinerService for TestMinerService {
 		Ok(())
 	}
 
+	/// Imports transactions to queue - treats as local based on config and tx source
+	fn import_claimed_local_transaction<C: Nonce + Sync>(&self, chain: &C, pending: PendingTransaction)
```
sorpaas: Maybe just call `import_own_transaction` below. The function body looks duplicate.

rpc/src/v1/helpers/dispatch.rs
XertroV: Yeah okay. I'm not too familiar with how all of this connects together, FYI.
What I'm doing atm is adding a `trusted: bool` param to `dispatch_transaction` that gets called with `false` from `eth_sendRawTx` and `true` from the other uses of `dispatch_transaction` (which seems to be called from `personal.rs` or `signer.rs`. That should mean all RPC methods that involve signing get a free pass regardless of config options.

Also going to add this flag to `import_claimed_local_transaction`.

rpc/src/v1/tests/helpers/miner_service.rs
XertroV: I tried that today and couldn't figure out how to do it (calling the instance of `import_own_transaction` above it).
The compiler told me to add `BlockChainClient` to `C` because `self.import_own_transaction` is called on `MinerService` not `TestMinerService`, but that didn't feel right.

Eventually I realised that `TestMinerService::import_own_transaction` isn't actually called anymore, so have left the body in my function and added `unimplemented!()` to `import_own_transaction`.

ethcore/src/miner/miner.rs
```diff
@@ -1191,6 +1220,34 @@ mod tests {
 		}.sign(keypair.secret(), Some(chain_id))
 	}
 
+	fn post_tx_assertions(
```
XertroV: This test refactor isn't correct.

5chdn:
Needs a 2nd review.

ethcore/src/miner/miner.rs
```diff
@@ -794,8 +797,9 @@ impl miner::MinerService for Miner {
 	fn import_own_transaction<C: miner::BlockChainClient>(
 		&self,
 		chain: &C,
-		pending: PendingTransaction,
+		pending: PendingTransaction
```
tomusdrw: We prefer to keep trailing commas in general.

PR #8883, opened 2018-06-13
Transactions hashes missing in trace_replayBlockTransactions method result #8725

### LINKED ISSUE #8725: Transactions hashes missing in trace_replayBlockTransactions method result

> I'm running:
>
> - **Which Parity version?**: 1.12.0
> - **Are you fully synchronized?**:  yes
> - **Which network are you connected to?**: ropsten 
> - **Did you try to restart the node?**:  yes

Hi, 

When running the method "trace_replayBlockTransactions" with "stateDiff" parameter I'm getting back a list of stateDiff of all transaction in a specific block.
The problem is that the result doesn't include the transactions hashes so we cant link the specific stateDiff to the related transaction.

For example, if running (on ropsten) the following:
`curl --data '{"method":"trace_replayBlockTransactions","params":["0x32d8e6",["trace"]],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545`

I'm getting:
```
{
    'jsonrpc':  '2.0',
     'result':  [
        {
            'output':  '0x000000000000000000000000202934ddeda0d38f30a70e9cf888c13afc2792bc',
             'stateDiff':  {
                '0x202934ddeda0d38f30a70e9cf888c13afc2792bc':  {
                    'balance':  { '+':  '0x0'},
                     'code':  { '+':  '0x6080..."},
                     'nonce':  { '+':  '0x1'},
                     'storage':  {
                        '0x0000000000000000000000000000000000000000000000000000000000000000':  {
                            '+':  '0x000000000000000000000000e7a043b5db22472df387d3cca173b6227b37f1f4' }
                    }
                },
                 '0xb2d9a807815d6e9afff805b835cd06cb990db423':  {
                    'balance':  '=',
                     'code':  '=',
                     'nonce':  { '*':  {'from':  '0x7', 'to':  '0x8' }},
                     'storage':  { }
                },
                 '0xe24246e6dcbb07bc15a1f9c3833fc1877df4c80e':  {
                    'balance':  { '*':  { 'from':  '0x7c5e833076332a74ad2','to':  '0x7c5e835123266a208d2' } },
                     'code':  '=',
                     'nonce':  '=',
                     'storage':  {}
                },
                 '0xe7a043b5db22472df387d3cca173b6227b37f1f4':  {
                    'balance':  { '*':  { 'from':  '0xd53e81b3de42cd4','to':  '0xd51dd4c09e96ed4' }},
                     'code':  '=',
                     'nonce':  { '*':  {'from':  '0x2a', 'to':  '0x2b' } },
                     'storage':  { }
                }
            },
             'trace':  [],
             'vmTrace':  None
        },
....],
     'id':  1
}
```
which you can see doesn't include the transaction hash.

Meanwhile (after checking with parity team on gitter) we can assume that the result of "trace_replayBlockTransactions" is in order with transactions index on the block.
Obvious, it isn't the straightforward way to handle this issue.

So adding the transaction hash to the result will be great.

shamardy:
closes #8725

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1565,12 +1565,12 @@ impl BlockChainClient for Client {
 
 		Ok(Box::new(txs.into_iter()
 			.map(move |t| {
-				let tx_hash = t.hash();
+				let transaction_hash = t.hash();
 				let t = SignedTransaction::new(t).expect(PROOF);
 				let machine = engine.machine();
 				let x = Self::do_virtual_call(machine, &env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
 				env_info.gas_used = env_info.gas_used + x.gas_used;
-				(tx_hash, x)
+				(transaction_hash, x)
 			})))
 	}
 
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1550,10 +1550,10 @@ impl BlockChainClient for Client {
 		let block = BlockId::Hash(address.block_hash);
 
 		const PROOF: &'static str = "The transaction address contains a valid index within block; qed";
-		Ok(self.replay_block_transactions(block, analytics)?.nth(address.index).expect(PROOF))
+		Ok(self.replay_block_transactions(block, analytics)?.nth(address.index).expect(PROOF).1)
 	}
 
-	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
+	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
 		let mut env_info = self.env_info(block).ok_or(CallError::StatePruned)?;
 		let body = self.block_body(block).ok_or(CallError::StatePruned)?;
 		let mut state = self.state_at_beginning(block).ok_or(CallError::StatePruned)?;
@@ -1565,11 +1565,12 @@ impl BlockChainClient for Client {
 
 		Ok(Box::new(txs.into_iter()
 			.map(move |t| {
+				let tx_hash = t.hash();
 				let t = SignedTransaction::new(t).expect(PROOF);
 				let machine = engine.machine();
 				let x = Self::do_virtual_call(machine, &env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
 				env_info.gas_used = env_info.gas_used + x.gas_used;
-				x
+				(tx_hash, x)
 			})))
 	}
 
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -612,7 +612,7 @@ impl BlockChainClient for TestBlockChainClient {
 	}
 
 	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
-		Ok(Box::new(vec!(H256::new()).into_iter().zip(self.execution_result.read().clone().unwrap().into_iter())))
+		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap_or(H256::new())).zip(self.execution_result.read().clone().unwrap().into_iter())))
 	}
 
 	fn block_total_difficulty(&self, _id: BlockId) -> Option<U256> {
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -611,8 +611,8 @@ impl BlockChainClient for TestBlockChainClient {
 		self.execution_result.read().clone().unwrap()
 	}
 
-	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		Ok(Box::new(self.execution_result.read().clone().unwrap().into_iter()))
+	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
+		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap_or(H256::new())).zip(self.execution_result.read().clone().unwrap().into_iter())))
 	}
 
 	fn block_total_difficulty(&self, _id: BlockId) -> Option<U256> {
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -612,7 +612,7 @@ impl BlockChainClient for TestBlockChainClient {
 	}
 
 	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
-		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap()).zip(self.execution_result.read().clone().unwrap().into_iter())))
+		Ok(Box::new(vec!(H256::new()).into_iter().zip(self.execution_result.read().clone().unwrap().into_iter())))
 	}
 
 	fn block_total_difficulty(&self, _id: BlockId) -> Option<U256> {
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -612,7 +612,7 @@ impl BlockChainClient for TestBlockChainClient {
 	}
 
 	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
-		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap_or(H256::new())).zip(self.execution_result.read().clone().unwrap().into_iter())))
+		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap()).zip(self.execution_result.read().clone().unwrap().into_iter())))
 	}
 
 	fn block_total_difficulty(&self, _id: BlockId) -> Option<U256> {
--- a/rpc/src/v1/types/trace.rs
+++ b/rpc/src/v1/types/trace.rs
@@ -22,6 +22,7 @@ use ethcore::trace as et;
 use ethcore::state_diff;
 use ethcore::account_diff;
 use ethcore::client::Executed;
+use ethereum_types::H256 as EthH256;
 use vm;
 use v1::types::{Bytes, H160, H256, U256};
 
@@ -627,6 +628,36 @@ impl From<Executed> for TraceResults {
 			trace: t.trace.into_iter().map(Into::into).collect(),
 			vm_trace: t.vm_trace.map(Into::into),
 			state_diff: t.state_diff.map(Into::into),
+		}
+	}
+}
+
+#[derive(Debug, Serialize)]
+/// A diff of some chunk of memory.
+pub struct TraceResultsWithTransactionHash {
+	/// The output of the call/create
+	pub output: Bytes,
+	/// The transaction trace.
+	pub trace: Vec<Trace>,
+	/// The transaction trace.
+	#[serde(rename="vmTrace")]
+	pub vm_trace: Option<VMTrace>,
+	/// The transaction trace.
+	#[serde(rename="stateDiff")]
+	pub state_diff: Option<StateDiff>,
+	/// The transaction Hash.
+	#[serde(rename="transactionHash")]
+	pub transaction_hash: H256,
+}
+
+impl From<(EthH256, Executed)> for TraceResultsWithTransactionHash {
+	fn from(t: (EthH256, Executed)) -> Self {
+		TraceResultsWithTransactionHash {
+			output: t.1.output.into(),
+			trace: t.1.trace.into_iter().map(Into::into).collect(),
+			vm_trace: t.1.vm_trace.map(Into::into),
+			state_diff: t.1.state_diff.map(Into::into),
+			transaction_hash: t.0.into(),
 		}
 	}
 }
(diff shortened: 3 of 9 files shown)


shamardy:
[clabot:check]

rpc/src/v1/types/trace.rs
```diff
@@ -195,20 +195,19 @@ impl From<account_diff::AccountDiff> for AccountDiff {
 	}
 }
 
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
```
debris: this is a breaking change

ethcore/types/src/state_diff.rs
```diff
@@ -19,28 +19,36 @@
...
 use account_diff::*;
 
 /// Expression for the delta between two system states. Encoded the
 /// delta of every altered account.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct StateDiff {
 	/// Raw diff key-value
-	pub raw: BTreeMap<Address, AccountDiff>
+	pub raw: BTreeMap<Address, AccountDiff>,
+	pub transaction_hash: H256
```
debris: `transaction_hash` is not a part of a `StateDiff` and should not be a property of a `struct StateDiff`

sorpaas:
@debris Yes. I think we should only change RPC types.

@shamardy Maybe we can consider to add an additional `Option` field in `TraceResults` representing the transaction hash (https://github.com/paritytech/parity/blob/master/rpc/src/v1/types/trace.rs#L610). When converting from `Executed` (https://github.com/paritytech/parity/blob/master/rpc/src/v1/types/trace.rs#L623), the transaction hash may not always be available (because we can replay a raw call) so it can be set to `None`. Then in replayTransaction (https://github.com/paritytech/parity/blob/master/rpc/src/v1/impls/traces.rs#L161) and replayBlockTransaction (https://github.com/paritytech/parity/blob/master/rpc/src/v1/impls/traces.rs#L167), manually set that field to the actual transaction hash. You can get the block transaction hashes via `client.block(id).transaction_hashes()`.

Or another idea might be to change replay_block_transactions in client (https://github.com/paritytech/parity/blob/master/ethcore/src/client/client.rs#L1552) to return an iterator of tuple of transaction hash and `Executed`. In that way there's no need to fetch transaction hashes again.

anyluck:
Is there a problem with this configuration
[parity]
db_path = "/ether/dat"
[rpc]
disable = false
port = 8545
interface = "local"
apis = ["web3", "eth", "net", "parity", "traces", "rpc", "secretstore", "personal"]

[ui]
interface = "all"
port = 8180

[websockets]
disable = false
port = 8546
interface = "all"
apis = ["web3", "eth", "pubsub", "net", "parity", "parity_pubsub", "traces", "rpc", "shh", "shh_pubsub"]

ethcore/src/client/client.rs
```diff
@@ -1240,6 +1241,8 @@ impl Client {
 
 			let mut ret = Executive::new(state, env_info, machine).transact_virtual(transaction, options)?;
 
+			ret.transaction_hash = if hash_trace { Some(transaction.hash()) } else { None };
```
debris: tbh, I don't know what to do with this. I generally don't like mutating structures like this. Making one field optional and then overwriting it in a different function is not idiomatic.

Unfortunately, it's not a new pattern and the same thing is done with `ret.state_diff` few lines below :/

shamardy: I don't know what can be done with this since it depends on if the user wants to trace transaction hashes or not.
Another way can be implementing it the same way but inside `replay_block_transactions` but I thought implementing this inside the same function that `state_diff` is implemented would look better.

shamardy: Maybe I can create a function that returns transaction hash inside the implementation of the `Executed` Struct like this:

`impl Executed {
	pub fn transaction_hash(transaction: &SignedTransaction) -> Option<H256> {
		...........................
	}
}`

then inside the do_virtual_call function I would set transaction_hash with this:
`if hash_trace { Some(ret.transaction_hash(transaction) };`

what do you think @debris

vs77bb:
Hi @debris I believe @shamardy might be waiting on your reply here. Hope you're doing well!

5chdn:
@sorpaas @debris could you respond to @shamardy here?

sorpaas:
My opinion on this is still that we should only change RPC types. The issue with adding `transactionHash` to `ethcore::CallAnalytics` type is that this "tracing" option doesn't have other uses (or allow users to set it) for callers other than `replayBlockTransactions`.

@shamardy Would you mind to take a look at https://github.com/paritytech/parity/pull/8883#issuecomment-396894373? We can directly fetch transaction hashes in the RPC call function, or if you want, return `Iterator<Item = (H256, Executed)>>` in `ethcore::Client::replay_block_transactions` would also be an idea. Or any other way you like to avoid changing non-RPC types. :)

shamardy:
@sorpaas please check updates. Thanks.

ethcore/src/client/client.rs
```diff
@@ -1541,15 +1541,15 @@ impl EngineInfo for Client {
 }
 
 impl BlockChainClient for Client {
-	fn replay(&self, id: TransactionId, analytics: CallAnalytics) -> Result<Executed, CallError> {
+	fn replay(&self, id: TransactionId, analytics: CallAnalytics) -> Result<(Option<H256>, Executed), CallError> {
```
sorpaas: Why is this Option? Isn't replay_block_transactions always return `Some(tx_hash)`?

ethcore/src/client/client.rs
```diff
@@ -1541,15 +1541,15 @@ impl EngineInfo for Client {
...
 		let address = self.transaction_address(id).ok_or(CallError::TransactionNotFound)?;
 		let block = BlockId::Hash(address.block_hash);
 
 		const PROOF: &'static str = "The transaction address contains a valid index within block; qed";
 		Ok(self.replay_block_transactions(block, analytics)?.nth(address.index).expect(PROOF))
 	}
 
-	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
+	fn replay_block_transactions(&self, block: BlockId, analytics: CallAnalytics) -> Result<Box<Iterator<Item = (Option<H256>,Executed)>>, CallError> {
```
sorpaas: Missing whitespace before `Executed`.

ethcore/src/client/client.rs
```diff
@@ -1561,11 +1561,12 @@ impl BlockChainClient for Client {
 
 		Ok(Box::new(txs.into_iter()
 			.map(move |t| {
+				let tx_hash = t.hash();
 				let t = SignedTransaction::new(t).expect(PROOF);
 				let machine = engine.machine();
 				let x = Self::do_virtual_call(machine, &env_info, &mut state, &t, analytics).expect(EXECUTE_PROOF);
 				env_info.gas_used = env_info.gas_used + x.gas_used;
-				x
+				(Some(tx_hash),x)
```
sorpaas: Missing whitespace before x.

ethcore/src/client/test_client.rs
```diff
@@ -607,12 +607,12 @@ impl EngineInfo for TestBlockChainClient {
 }
 
 impl BlockChainClient for TestBlockChainClient {
-	fn replay(&self, _id: TransactionId, _analytics: CallAnalytics) -> Result<Executed, CallError> {
-		self.execution_result.read().clone().unwrap()
+	fn replay(&self, _id: TransactionId, _analytics: CallAnalytics) -> Result<(Option<H256>,Executed), CallError> {
```
sorpaas: Missing whitespace for this and several cases below.

ethcore/src/client/test_client.rs
```diff
@@ -607,12 +607,12 @@ impl EngineInfo for TestBlockChainClient {
...
+		Ok((None,self.execution_result.read().clone().unwrap()?))
```
sorpaas: Can we fetch the transaction hash from `TransactionId`?

ethcore/src/client/test_client.rs
```diff
@@ -607,12 +607,12 @@ impl EngineInfo for TestBlockChainClient {
...
 	}
 
-	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		Ok(Box::new(self.execution_result.read().clone().unwrap().into_iter()))
+	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (Option<H256>,Executed)>>, CallError> {
+		Ok(Box::new(None.into_iter().zip(self.execution_result.read().clone().unwrap().into_iter())))
```
sorpaas: Same as above. Block info are available in TestBlockChainClient, so I think it may be better to fetch that available info.

rpc/src/v1/types/trace.rs
```diff
@@ -631,6 +632,36 @@ impl From<Executed> for TraceResults {
 	}
 }
 
+#[derive(Debug, Serialize)]
+/// A diff of some chunk of memory.
+pub struct TraceResultsWithTransHash {
```
sorpaas: Please use full name: `TraceResultsWithTransactionHash`.

rpc/src/v1/types/trace.rs
```diff
@@ -631,6 +632,36 @@ impl From<Executed> for TraceResults {
...
+	/// The transaction trace.
+	pub trace: Vec<Trace>,
+	/// The transaction trace.
+	#[serde(rename="vmTrace")]
+	pub vm_trace: Option<VMTrace>,
+	/// The transaction trace.
+	#[serde(rename="stateDiff")]
+	pub state_diff: Option<StateDiff>,
+	/// The transaction Hash.
+	#[serde(rename="TransactionHash")]
```
sorpaas: The usual json conversion is `transactionHash`.

rpc/src/v1/types/trace.rs
```diff
@@ -631,6 +632,36 @@ impl From<Executed> for TraceResults {
...
+	pub trans_hash: Option<H256>,
```
sorpaas: This should be changed to full name. And why is it Option? I think if we're using this type then the transaction hash should always be available, otherwise we'd rather use `TraceResults`?

rpc/src/v1/types/trace.rs
```diff
@@ -22,6 +22,7 @@ use ethcore::trace as et;
 use ethcore::state_diff;
 use ethcore::account_diff;
 use ethcore::client::Executed;
+use ethereum_types::H256 as ethH256;
```
sorpaas: We usually have Rust type with first letter to be capped.

5chdn:
Please rebase on latest master if possible, this will fix the CI issue.

ethcore/src/client/test_client.rs
```diff
@@ -611,8 +611,8 @@ impl BlockChainClient for TestBlockChainClient {
 		self.execution_result.read().clone().unwrap()
 	}
 
-	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		Ok(Box::new(self.execution_result.read().clone().unwrap().into_iter()))
+	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
+		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap_or(H256::new())).zip(self.execution_result.read().clone().unwrap().into_iter())))
```
sorpaas: The `self.execution_result` param is always set by `self.set_execution_result` function. We've never set the actual transaction hashes or modify `self.traces` when calling that function. So it always returns `H256::new` regardless. The `unwrap_or` can be removed.

ethcore/src/client/test_client.rs
```diff
@@ -611,8 +611,8 @@ impl BlockChainClient for TestBlockChainClient {
 		self.execution_result.read().clone().unwrap()
 	}
 
-	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = Executed>>, CallError> {
-		Ok(Box::new(self.execution_result.read().clone().unwrap().into_iter()))
+	fn replay_block_transactions(&self, _block: BlockId, _analytics: CallAnalytics) -> Result<Box<Iterator<Item = (H256, Executed)>>, CallError> {
+		Ok(Box::new(self.traces.read().clone().unwrap().into_iter().map(|t| t.transaction_hash.unwrap()).zip(self.execution_result.read().clone().unwrap().into_iter())))
```
sorpaas: @shamardy Sorry if I wasn't clear -- I think this should be changed **not** to use `self.traces` (but just return `H256::new`), because when we query `self.execution_result` in test, `self.traces` is never available -- `self.execution_result` is only set by `self.set_execution_result` and we only have a few appearance our code.

sorpaas: The above code will cause testing run of replayBlockTransactions to always return empty vector -- definitely not something we want.

shamardy:
@sorpaas in `rpc\src\v1\tests\mocked\traces.rs` the `self.traces` is set in line 38:
```
	*client.traces.write() = Some(vec![LocalizedTrace {
		action: Action::Call(Call {
			from: 0xf.into(),
			to: 0x10.into(),
			value: 0x1.into(),
			gas: 0x100.into(),
			input: vec![1, 2, 3],
			call_type: CallType::Call,
		}),
		result: Res::None,
		subtraces: 0,
		trace_address: vec![0],
		transaction_number: Some(0),
		transaction_hash: Some(5.into()),
		block_number: 10,
		block_hash: 10.into(),
	}]);
```
in line 236 there is a test for replay_block_transactions:
```
#[test]
fn rpc_trace_replay_block_transactions() {
	let tester = io();

	let request = r#"{"jsonrpc":"2.0","method":"trace_replayBlockTransactions","params":["0x10", ["trace", "stateDiff", "vmTrace"]],"id":1}"#;
	let response = r#"{"jsonrpc":"2.0","result":[{"output":"0x010203","stateDiff":null,"trace":[],"transactionHash":"0x0000000000000000000000000000000000000000000000000000000000000005","vmTrace":null}],"id":1}"#;

	assert_eq!(tester.io.handle_request_sync(request), Some(response.to_owned()));
}
```
in the response the `transactionHash `should be `5` like its set in the `*client.traces.write()` when I used `H256::new` in the last commit it returned a `transactionHash` of `0` maybe I should revert to the previous commit that uses `self.traces` where the output was right. what do you think?

shamardy:
also if thats the case should I revert to the one which uses `unwrap_or(H256::new())` or just `unwrap()`. is there a difference?

sorpaas:
@shamardy Ah you're right. Sorry it's my mistake.

Then I think your original one is correct -- we should revert to use `unwrap_or(H256::new())`.

shamardy:
@debris can you check the last update please. Thanks.

shamardy:
any update here?

5chdn:
> @debris can you check the last update, please. Thanks.

or anyone?

PR #8885, opened 2018-06-13
Allow Poll Lifetime to be configured via CLI

### LINKED ISSUE #5484: Make filter poll lifetime configurable
Poll lifetime is [currently hardcoded](https://github.com/paritytech/parity/blob/v1.6.6/rpc/src/v1/helpers/poll_manager.rs#L22) to 60 seconds. Scheduling software like cron runs at granularity of 60 seconds, and with high traffic, or a queue of filters to check, it's easy to exceed the 60 second lifetime.

Is it possible to make this configurable? Is there a downside?

gnunicorn:
... rather than it being a hard-coded constant. fixes #5484 .

However, in order to allow this to be configurable, we have to pass through the from the CLI through a range of structs, for which this is unrelated until it finally reaches the internally used PollManager. To achieve this I had to modify, among others, `EthClient` and `EthClientOptions` both of which are public structs exposed by `parity-rpc`. Thus means, these are technically backwards incompatible API-changes, which - under [SemVer](https://semver.org/) - would mean we have to bump the Major version of this crate...? Not sure if we want to do that.

Secondly, I am not super certain I traced the entire flow everywhere and there isn't still a `::Default()` somewhere that creates poll-lifetimes other than what the CLI specifies. Happy for any more experienced reviewer to point them out if I missed any!

### DIFF
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -732,6 +732,10 @@ usage! {
 			"--gas-price-percentile=[PCT]",
 			"Set PCT percentile gas price value from last 100 blocks as default gas price when sending transactions.",
 
+			ARG arg_poll_lifetime: (u32) = 60u32, or |c: &Config| c.mining.as_ref()?.poll_lifetime.clone(),
+			"--poll-lifetime=[S]",
+			"Set the lifetime of the internal index filter to S seconds.",
+
 			ARG arg_author: (Option<String>) = None, or |c: &Config| c.mining.as_ref()?.author.clone(),
 			"--author=[ADDRESS]",
 			"Specify the block author (aka \"coinbase\") address for sending block rewards from sealed blocks. NOTE: MINING WILL NOT WORK WITHOUT THIS OPTION.", // Sealing/Mining Option
@@ -1241,6 +1245,7 @@ struct Mining {
 	relay_set: Option<String>,
 	min_gas_price: Option<u64>,
 	gas_price_percentile: Option<usize>,
+	poll_lifetime: Option<u32>,
 	usd_per_tx: Option<String>,
 	usd_per_eth: Option<String>,
 	price_update_period: Option<String>,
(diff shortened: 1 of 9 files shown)


parity/cli/mod.rs
```diff
@@ -728,6 +728,10 @@ usage! {
 			"--gas-price-percentile=[PCT]",
 			"Set PCT percentile gas price value from last 100 blocks as default gas price when sending transactions.",
 
+			ARG arg_poll_lifetime: (u32) = 60u32, or |c: &Config| c.mining.as_ref()?.poll_lifetime.clone(),
+			"--poll-lifetime=[S]",
+			"Set the lifetime of the internal index filter to S seconds. Deafult: 60",
```
axelchalon: Didn't review; just a nitpick: you don't need to specify the default value in the help message; it's automatically appended.

debris:
> ... would mean we have to bump the Major version of this crate...? Not sure if we want to do that.

gnunicorn:
@debris I guess that is okay-ish... or do we know of any other people using this/our internal crates and thus, we'd break their build (as [cargo follow semver](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)?)... I mean, there is little downside for us to bump if, especially if it effects many people ...

also. the latest tests failures weren't mine but because of an old master. gitlab passed now.

debris:
> @debris I guess that is okay-ish... or do we know of any other people using this/our internal crates and thus

I don't think these crates are used directly by anyone. Usually people use parity over rpc

gnunicorn:
@5chdn it doesn't grumble anymore. might re-add the review-tag?

5chdn:
PR is already reviewed well. It just needs someone else to sign off.

Tbaut:
>LGTM, but "Set the lifetime of the internal index filter to S seconds." is going to trigger end-user questions imo so let's add a add-docs-issue when we merge this?

+1, this is not clear to me what this means: https://github.com/paritytech/parity/issues/8917

PR #8886, opened 2018-06-13
Add new ovh bootnodes and fix port for foundation bootnode 3.2

ddorgan:


5chdn:
Thanks. Do we have to remove old bootnodes btw?

ddorgan:
Better just remove them. Since some tweaking and updates 3 of the 4 are now serving snapshots, but best replace them with the new hosts.

5chdn:
But they are still working, I just tested that.

Could you remove the duplicate `1118980b...` entry?

5chdn:
Just tested: 30 out of 32 nodes (and 1 duplicate) are working :+1: 

I did not expect so much stability :)

folsen:
@5chdn how did you test them? (I'm looking at making this into a CI step)

5chdn:
I added them all to a bootnodes.txt and ran a parity client with --reserved-only and I logged all successful connections for some minutes (maybe 10?) and logged the enodes and figured a successful connection means the remote node must be valid and alive. No science involved :)

PR #8890, opened 2018-06-13
Hardware-wallets `Clean up things I missed in the latest PR`

niklasad1:


5chdn:
lol did I merge too fast?

niklasad1:
No, I didn’t want to introduce new changes after it got approved ;)

hw/src/lib.rs
```diff
@@ -181,16 +183,18 @@ impl From<trezor::Error> for Error {
...
 		Error::Usb(err)
 	}
 }
 
 /// Specifies the direction of the `HardwareWallet` i.e, whether it arrived or left
 #[derive(Debug, Copy, Clone)]
 pub enum DeviceDirection {
-    Arrived,
-    Left,
+	/// Device arrived
```
dvdplm: As in "user pulled out the wallet from the usb port"? Maybe be explicit here and say `Device became available`/`Device was removed from the system`?

hw/src/ledger.rs
```diff
@@ -363,7 +364,7 @@ impl Manager {
 }
 
 // Try to connect to the device using polling in at most the time specified by the `timeout`
-fn try_connect_polling(ledger: Arc<Manager>, timeout: &Duration, device_direction: DeviceDirection) -> bool {
+fn try_connect_polling(ledger: &Arc<Manager>, timeout: &Duration, device_direction: DeviceDirection) -> bool {
```
debris: maybe just `&Manager`? `&Arc<Manager>` will be automatically dereferenced to it

andresilva:
Do you mind merging master into this or rebasing? Makes it easier to review (also avoids having repeated commits in the changelog).

hw/src/lib.rs
niklasad1: @dvdplm do you want `s/Arrived/Available/ s/Left/Unavailable` renamed completely including the enum patterns? 

Happy to change, I have no strong opinions here!

hw/src/trezor.rs
```diff
@@ -439,7 +433,7 @@ impl <'a>Wallet<'a> for Manager {
 }
 
 // Try to connect to the device using polling in at most the time specified by the `timeout`
-fn try_connect_polling(trezor: Arc<Manager>, duration: &Duration, dir: DeviceDirection) -> bool {
+fn try_connect_polling(trezor: &Arc<Manager>, duration: &Duration, dir: DeviceDirection) -> bool {
```
andresilva: Use `&Manager` instead.

hw/src/ledger.rs
```diff
@@ -15,19 +15,20 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 //! Ledger hardware wallet module. Supports Ledger Blue and Nano S.
-/// See https://github.com/LedgerHQ/blue-app-eth/blob/master/doc/ethapp.asc for protocol details.
+/// See <https://github.com/LedgerHQ/blue-app-eth/blob/master/doc/ethapp.asc> for protocol details.
```
andresilva: Maybe replace "///" with "//!".

hw/src/ledger.rs
niklasad1: Goood catch thanks :1st_place_medal:

hw/src/lib.rs
dvdplm: I think that would make sense, yes, but my grumble was about the doc-comment. :)

niklasad1: Ok, I having second thoughts on this because `DeviceDirection` doesn't actually indicate whether a device became available or unavailable. Essentially, it only tells we got a `notification` that a `USB device` was connected or disconnected with manufacturer ID (Ledger or Trezor). To actually say that a device become available/unavailable we need to do additional checks, for instance, request an Ethereum address and so on!

So, I will keep this! Ok?!

Issue #8762, opened 2018-06-01
Bug_report: trace_replayBlockTransactions responds with error to set pruning to archive at block 0


ankitchiplunkar:
> I'm running:
>
> - **Which Parity version?**: Parity/v1.10.4-stable
> - **Which operating system?**: Ubuntu 18.04
> - **How installed?**: Simple one-line installer 
> - **Are you fully synchronized?**:  yes
> - **Which network are you connected to?**: ethereum 
> - **Did you try to restart the node?**: yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._
I am not able to use the JSON-RPC command  trace_replayBlockTransactions at block 0

```
(venv) blah@blah:~/ether_sql$ curl --data '{"method":"trace_replayBlockTransactions","params":["0x0",["trace"]],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":1}
```
The above command should return a null Array

```
{"jsonrpc":"2.0","result":[],"id":1}
```

This might be linked to a previous issue: https://github.com/paritytech/parity/issues/7547
The difference is that I can view traces using other functions:

```
(venv) blah@blah:~/ether_sql$ curl --data '{"method":"trace_block","params":["0x0"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","result":[],"id":1}

```

5chdn:
Fixed in #8891

PR #8891, opened 2018-06-13
Block 0 is valid in queries

### LINKED ISSUE #7547: rpc: trace_replayBlockTransactions respond with error to set pruning to archive even if it is set
parity version: Parity/v1.10.0-unstable-b6e61f403-20180112/x86_64-linux-gnu/rustc1.22.1
installed from source 
node is running with those flgs: `parity --config dev --tracing on --pruning=archive --rpcapi eth,personal,net,parity,traces`

there is issue with replayBlockTransactions:
curl --data '{"method":"trace_replayBlockTransactions","params":["0x0", ["trace", "stateDiff", "vmTrace"]],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
actual result:
{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":1}
expected result:
response with traces and not an error 

node startup stdout:
2018-01-12 18:25:28  Starting Parity/v1.10.0-unstable-b6e61f403-20180112/x86_64-linux-gnu/rustc1.22.1
2018-01-12 18:25:28  Keys path /home/*/.local/share/io.parity.ethereum/keys/DevelopmentChain
2018-01-12 18:25:28  DB path /home/*/.local/share/io.parity.ethereum/chains/DevelopmentChain/db/125ccdc70e48df86
2018-01-12 18:25:28  Path to dapps /home/*/.local/share/io.parity.ethereum/dapps
2018-01-12 18:25:28  State DB configuration: archive +Trace
2018-01-12 18:25:28  Operating mode: active
2018-01-12 18:25:28  Warning: Warp Sync is disabled because tracing is turned on.
2018-01-12 18:25:28  Configured for DevelopmentChain using InstantSeal engine

dvdplm:
Early exit for block nr 0 leads to spurious error about pruning: `…your node is running with state pruning…`.

Fixes #7547, #8762

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1047,7 +1047,8 @@ impl Client {
 	/// Otherwise, this can fail (but may not) if the DB prunes state.
 	pub fn state_at_beginning(&self, id: BlockId) -> Option<State<StateDB>> {
 		match self.block_number(id) {
-			None | Some(0) => None,
+			None => None,
+			Some(0) => self.state_at(id),
 			Some(n) => self.state_at(BlockId::Number(n - 1)),
 		}
 	}


ethcore/src/client/client.rs
```diff
@@ -1047,7 +1047,8 @@ impl Client {
 	/// Otherwise, this can fail (but may not) if the DB prunes state.
 	pub fn state_at_beginning(&self, id: BlockId) -> Option<State<StateDB>> {
 		match self.block_number(id) {
-			None | Some(0) => None,
+			None => None,
+			Some(0) => self.state_at(id),
 			Some(n) => self.state_at(BlockId::Number(n - 1)),
```
debris: any idea why is it `n - 1` ?

andresilva: I think it's the state at the beginning of the block, i.e. last block's end state.

dvdplm:
> I'm not sure whether this should be None or some empty state.

@andresilva I wasn't sure either tbh, but the RPC output is "some empty state": `{"jsonrpc":"2.0","result":[],"id":1}`, which seem to make more sense than the misleading `{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":1}` we return now.

Are you ok merging this?

ankitchiplunkar:
> I'm not sure whether this should be None or some empty state.

I think the result of using method `eth_traceReplayBlockTransactions` with mode `stateDiff` (for the mainnet) at block 0 should be the genesis rewards. More details here: https://ethereum.stackexchange.com/questions/44196/which-node-can-provide-transactions-in-block-0

andresilva:
@ankitchiplunkar That's what I was getting at when I said this:

"I think genesis state should only exist **after** genesis block (or is it beginning of genesis? 🤔)."

So if genesis state exists at the **beginning** of the genesis block then `stateDiff` should be empty, if it's applied at the **end** of the genesis block then as you said `stateDiff` should show those state changes. TBH I don't think there's a right answer here, it's just semantics for an edge case we can define it whatever way we want. The current implementation will not return the `stateDiff` as you suggested and will just return `[]` instead (I think).

dvdplm:
> just return [] instead (I think).

Correct.

Issue #6744, opened 2017-10-13
evm interface overhaul


debris:
Current evm interface has been great for us, but as the ethereum virtual machine gets more complicated, we need to revisit how it's done. We need to introduce changes which improve code readability and not decrease the performance.

assumptions:

- evm should never do recursive calls
- executive should not use `crossbeam` to reset the stack
- we should get rid of `Externalities` interface
- evm should be executed in a loop. the result of the execution should contain state changes, information if execution is done and what should be executed in the next iteration of the loop
- evm should not use callbacks
- executive public interface should not be changed
- builtins should be separated from executive

I think we could target `1.10` with those changes, so 3 months from now.

cc @paritytech/core-devs

tomusdrw:
Is it also possible to clearly separate builtins from executive?

debris:
@tomusdrw I've added it to the list

pepyakin:
I thought that these wasm vm use-cases might be useful to keep in mind if and when performing the overhaul:

For now, the wasm engine performs validation of the wasm code [each time before executing a contract](https://github.com/paritytech/parity/blob/fe5f5b28d9be6b894545cfbbb42580068e14524b/ethcore/wasm/src/lib.rs#L91). The validation run time depends on the code size. Since it depends only on a module itself and nothing else (i.e pure), it doesn't make sense to perform validation each times. It's sufficient to validate once and cache the result of the validation somewhere. Basically it's just a tri-state: not validated, invalid and valid.

Apart from that, there is an instrumentation (such as gas metering) of the wasm code that is also performed on each run. It's not so pure as validation, since it might depend on the exact algorithms used (i.e you can do metering by counting each instruction or by summing all instructions in a block. Same result but different algos), but IMO still something worthwhile caching.

And, of course, wasm→native compilation: recompiling each time before the execution might be a no-go and definitely needs some sort of caching.

So this boils down to the need of caching some data by the vm: ranging from a single flag to a big byte blob.

sorpaas:
I would like to take this refactoring, if that's okay.

[...]

PR #8914, opened 2018-06-18
Refactor evm Instruction to be a c-like enum

sorpaas:
rel #6744 

This allows some more type checking for the type, and also allows to put functions in a more Rust-y way.

* A small macro is used to implement `from_u8` using `match`. A small part of this marco is from `enum_derive` crate. But that crate defines other functions (like `from_u64`, etc) which we don't use.
* `INSTRUCTIONS` table is made private, and it's accessed from `Instruction::info(&self)`.
* Put `exec_stack_instruction` to `exec_instruction`. This allows Rust to type check match arms, and refuse to compile if new opcode is added, but its execution logic is not defined.


ethcore/evm/src/instructions.rs
```diff
@@ -165,436 +453,189 @@ impl InstructionInfo {
...
-		arr[CALLCODE as usize] =		InstructionInfo::new("CALLCODE",		7, 1, GasPriceTier::Special);
-		arr[RETURN as usize] =			InstructionInfo::new("RETURN",			2, 0, GasPriceTier::Zero);
-		arr[DELEGATECALL as usize] =	InstructionInfo::new("DELEGATECALL",	6, 1, GasPriceTier::Special);
-		arr[STATICCALL as usize] =		InstructionInfo::new("STATICCALL",		6, 1, GasPriceTier::Special);
-		arr[SUICIDE as usize] = 		InstructionInfo::new("SUICIDE",			1, 0, GasPriceTier::Special);
-		arr[CREATE2 as usize] = 		InstructionInfo::new("CREATE2",			3, 1, GasPriceTier::Special);
-		arr[REVERT as usize] =			InstructionInfo::new("REVERT",			2, 0, GasPriceTier::Zero);
+	static ref INSTRUCTIONS: [Option<InstructionInfo>; 0x100] = {
+		let mut arr = [None; 0x100];
+		arr[STOP as usize] =			Some(InstructionInfo::new("STOP",			0, 0, GasPriceTier::Zero));
```
sorpaas: This tab formatting displays correctly on my Emacs, but looks like it doesn't work on Github interface.
I'm thinking about maybe it's better to just use a single space as separator -- we don't gain much readability by using multiple tabs.

ethcore/evm/src/instructions.rs
```diff
@@ -16,83 +16,378 @@
...
-#[test]
-fn test_is_push() {
-	assert!(is_push(PUSH1));
-	assert!(is_push(PUSH32));
-	assert!(!is_push(DUP1));
+		impl $name {
+			#[doc = "Convert from u8 to the given enum"]
+			pub fn from_u8(value: u8) -> Option<Self> {
+				match value {
+					$( $discriminator => Some($variant) ),+,
```
tomusdrw: Are we sure it's equally performant to borrowing from the static array?

ethcore/evm/src/interpreter/shared_cache.rs
```diff
@@ -70,12 +70,14 @@ impl SharedCache {
...
-			if instruction == instructions::JUMPDEST {
-				jump_dests.insert(position);
-			} else if instructions::is_push(instruction) {
-				position += instructions::get_push_bytes(instruction);
+			let instruction = Instruction::from_u8(code[position]);
+
+			if let Some(instruction) = instruction {
+				if instruction == instructions::JUMPDEST {
+					jump_dests.insert(position);
+				} else if instruction.is_push() {
```
tomusdrw: Maybe change that to `if let Some(xx) = instructions.push_bytes() {` ?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -124,24 +123,31 @@ impl<Cost: CostType> vm::Vm for Interpreter<Cost> {
 		let mut gasometer = Gasometer::<Cost>::new(Cost::from_u256(params.gas)?);
 		let mut stack = VecStack::with_capacity(ext.schedule().stack_limit, U256::zero());
 		let mut reader = CodeReader::new(code);
-		let infos = &*instructions::INSTRUCTIONS;
 
 		while reader.position < code.len() {
-			let instruction = code[reader.position];
+			let opcode = code[reader.position];
+			let instruction = Instruction::from_u8(opcode);
```
tomusdrw: Might be worth mapping the `code` to `Instructions` during the first phase along with jumpdestination analysis.

sorpaas: The current issue is that we only calculate valid jump destinations upon first JUMP instruction, so on L129 that info may not be available.

I'll try and see (probably in a future PR) whether incrementally calculating jumpdests along with the exec loop would be possible, or another idea is to store jumpdest result to the db -- that info is static and never changes for a particular code array.

sorpaas:
Here's the result got from `evmbin/bench.sh`:

This PR:

```
Output: 0x606060405200
Gas used: 29fb170
Time: 0.511926861s
^^^^ usize
Output: 0x606060405200
Gas used: 29fb170
Time: 0.519975655s
^^^^ U256
Output: 0x606060405200
Gas used: 8865053
Time: 1.749029145s
^^^^ usize
Output: 0x606060405200
Gas used: 8865053
Time: 1.787731034s
^^^^ U256
```

Master Branch:

```
Output: 0x606060405200
Gas used: 29fb170
Time: 0.573969108s
^^^^ usize
Output: 0x606060405200
Gas used: 29fb170
Time: 0.562567334s
^^^^ U256
Output: 0x606060405200
Gas used: 8865053
Time: 2.61541081s
^^^^ usize
Output: 0x606060405200
Gas used: 8865053
Time: 2.23390354s
^^^^ U256
```

Looks like somehow the conversion to enum makes it run just slightly faster..

Don't know whether this is representative enough. We can also try some benchmarks using some of the jsontests.

sorpaas:
I got some benchmarks using #8944. Currently, running the comparison requires manually checking out two branches so it's not hassle-free. Another issue is that many tests only run for a really short time, and I cannot find a reliable way to make some of the results really comparable.

Below is a tuple of tests that run faster in this PR, 0-10% slower, 10-20% slower, 20-30% slower, 30-40% slower, 40-50% slower or 50%+ slower:

```
(256, 150, 66, 23, 43, 62, 73)
```

This tuple varies a lot each time I run the tests and gather the csv.

In particular, the performance tests in vm folder might be representive:

```
{'ackermann31-INT': 0.9897959183673469,
 'ackermann32-INT': 0.9167478091528725,
 'ackermann33-INT': 0.9596136962247586,
 'fibonacci10-INT': 1.0860832137733143,
 'fibonacci16-INT': 0.9300284017557449,
 'loop-add-10M-INT': 0.9226587206771758,
 'loop-divadd-10M-INT': 0.9229079794970425,
 'loop-divadd-unr100-10M-INT': 0.8998631955175739,
 'loop-exp-16b-100k-INT': 1.0188090390141178,
 'loop-exp-1b-1M-INT': 0.9923591588497622,
 'loop-exp-2b-100k-INT': 0.9466506030511994,
 'loop-exp-32b-100k-INT': 1.0991487959483188,
 'loop-exp-4b-100k-INT': 1.12556930650952,
 'loop-exp-8b-100k-INT': 1.1088218849334053,
 'loop-exp-nop-1M-INT': 0.8219694300785856,
 'loop-mul-INT': 0.8684106995830124,
 'loop-mulmod-2M-INT': 0.9754786210274101,
 'manyFunctions100-INT': 0.8409570724841661}
```

The value is `timing in this PR / timing on master`.

Those results are gathered by running `evm_jsontests_bench.sh` and then using a half-finished Jupyter notebook. I don't think that notebook is publishable, but I'll put it up as a gist once I cleared it up.

tomusdrw:
@sorpaas Great, that's good enough for me. I think the performance tests and perhaps `evmbin/benches.rs` should be a good indicator, so I'd focus on running those.
Maybe we add some more tests to `evmbin/benches` (separate PR) and use only this in future? Will simplify comparing the changes in the future.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -124,24 +123,31 @@ impl<Cost: CostType> vm::Vm for Interpreter<Cost> {
...
+			let instruction = instruction.expect("None case is checked above; qed");
+
+			let info = instruction.info();
 			self.verify_instruction(ext, instruction, info, &stack)?;
 
 			// Calculate gas cost
 			let requirements = gasometer.requirements(ext, instruction, info, &stack, self.mem.size())?;
 			if do_trace {
-				ext.trace_prepare_execute(reader.position - 1, instruction, requirements.gas_cost.as_u256());
+				ext.trace_prepare_execute(reader.position - 1, instruction as u8, requirements.gas_cost.as_u256());
```
jimpo: nit: Could use `opcode` instead of `instruction as u8`.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -227,13 +233,7 @@ impl<Cost: CostType> Interpreter<Cost> {
...
 
 			return Err(vm::Error::BadInstruction {
-				instruction: instruction
-			});
-		}
-
-		if info.tier == instructions::GasPriceTier::Invalid {
-			return Err(vm::Error::BadInstruction {
-				instruction: instruction
+				instruction: instruction as u8
```
jimpo: nit: Could use `opcode` instead of `instruction as u8`.

ethcore/evm/src/instructions.rs
```diff
@@ -564,24 +571,26 @@ pub enum GasPriceTier {
 	Invalid
 }
 
-impl Default for GasPriceTier {
-	fn default() -> Self {
-		GasPriceTier::Invalid
+impl GasPriceTier {
+	/// Returns the index in schedule for specific `GasPriceTier`
+	pub fn idx(&self) -> usize {
```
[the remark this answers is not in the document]
sorpaas: That "c-style enum" is actually just `GasPriceTier`. :)

We still need this `idx` function. It's used in `schedule`, where we dispatch based on the index to know about the default gas.

ethcore/evm/src/interpreter/mod.rs
sorpaas: We don't have opcode variable in `verify_instruction`. Because `instruction as u8` is basically a noop, I think this would be okay?

ethcore/evm/src/instructions.rs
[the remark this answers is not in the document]
jimpo: Ah, didn't realize the values were assigned implicitly. In that case, seems like this method could just be implemented as `*self as usize` instead of the match, or inlined.

5chdn:

ethcore/evm/src/instructions.rs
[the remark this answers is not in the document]
sorpaas: I'm not really sure about that -- the performance is mostly same, but I'm a little bit concerned about clarity.

Unlike `Instruction`, We don't have enum-to-number natural corresponding for `GasPriceTier`. I tried adding the index to enum and make `GasPriceTier` a c-like enum. The issue, however, is that there're other number corresponding for `GasPriceTier` (like default gas cost). And the result might be a little bit confusing.

PR #8920, opened 2018-06-19
Add type for passwords.

twittner:


ethcore/src/account_provider/mod.rs
```diff
@@ -560,29 +560,29 @@ impl AccountProvider {
...
 	/// Permanently removes an account.
-	pub fn kill_account(&self, address: &Address, password: &str) -> Result<(), Error> {
+	pub fn kill_account(&self, address: &Address, password: &Password) -> Result<(), Error> {
 		self.sstore.remove_account(&self.sstore.account_ref(&address)?, &password)?;
 		Ok(())
 	}
 
 	/// Changes the password of `account` from `password` to `new_password`. Fails if incorrect `password` given.
-	pub fn change_password(&self, address: &Address, password: String, new_password: String) -> Result<(), Error> {
+	pub fn change_password(&self, address: &Address, password: Password, new_password: Password) -> Result<(), Error> {
```
niklasad1: Out-of-scope for this PR:
Why not `pass-by_ref` here?

ethcore/src/account_provider/mod.rs
```diff
@@ -560,29 +560,29 @@ impl AccountProvider {
...
 		self.sstore.change_password(&self.sstore.account_ref(address)?, &password, &new_password)
 	}
 
 	/// Exports an account for given address.
-	pub fn export_account(&self, address: &Address, password: String) -> Result<KeyFile, Error> {
+	pub fn export_account(&self, address: &Address, password: Password) -> Result<KeyFile, Error> {
```
niklasad1: Out-of-scope for this PR:
Why not `pass-by_ref` here?

ethcore/src/account_provider/mod.rs
```diff
@@ -560,29 +560,29 @@ impl AccountProvider {
...
 		self.sstore.export_account(&self.sstore.account_ref(address)?, &password)
 	}
 
 	/// Helper method used for unlocking accounts.
-	fn unlock_account(&self, address: Address, password: String, unlock: Unlock) -> Result<(), Error> {
+	fn unlock_account(&self, address: Address, password: Password, unlock: Unlock) -> Result<(), Error> {
```
niklasad1: Out-of-scope for this PR:
Why not `pass-by_ref` here?

ethkey/src/password.rs
```diff
@@ -0,0 +1,58 @@
...
+impl Drop for Password {
+	fn drop(&mut self) {
+		let vec = unsafe {
+			self.0.as_mut_vec()
+		};
+		let n = vec.len();
+		let p = vec.as_mut_ptr();
+		for i in 0..n {
+			unsafe {
+				ptr::write_volatile(p.offset(i as isize), 0)
```
niklasad1: I think it could be worth to add a comment and explaining why `write_volatile` is used here. i.e., to actually make sure that `passwords` are erased and not LLVM performs some magic optimization!

Also consider to simply the code to:
```rust
    let ptr = unsafe { s.as_mut_vec() };
    for byte in ptr {
        unsafe { ptr::write_volatile(byte, 0) }
    }
```

5chdn:
Needs 2nd review.

ethkey/src/password.rs
```diff
@@ -0,0 +1,54 @@
...
+	pub fn as_str(&self) -> &str {
+		self.0.as_str()
+	}
+}
+
+// Custom drop impl to zero out memory.
+impl Drop for Password {
+	fn drop(&mut self) {
+		unsafe {
+            for byte_ref in self.0.as_mut_vec() {
```
andresilva: Replace indentation spaces with tabs.

util/mem/src/lib.rs
```diff
@@ -31,11 +31,9 @@ impl<T: AsMut<[u8]>> From<T> for Memzero<T> {
...
 impl<T: AsMut<[u8]>> Drop for Memzero<T> {
 	fn drop(&mut self) {
-		let n = self.mem.as_mut().len();
-		let p = self.mem.as_mut().as_mut_ptr();
-		for i in 0..n {
-			unsafe {
-				ptr::write_volatile(p.offset(i as isize), 0)
+		unsafe {
+			for byte_ref in self.mem.as_mut() {
+                ptr::write_volatile(byte_ref, 0)
```
andresilva: Tabs :)

niklasad1:
@dvdplm 

Explicit `mem::drop` is just to drop a value earlier than when it goes out-of-scope and in this case it doesn´t make sense to me! In this case, drop will called on the `String` after the drop on Password has been executed!

Also in context of this `issue` the essential part is the zero out the memory in RAM holding the password because after Rust is `dropping the value` (calling `free()` I guess), it's no guarantee that it actually gets destroyed it´s up the each Operating System's free() implementation to decide what to do with the memory. It might just be in memory until a `malloc()` call with the specific chunk size arrives for efficiency reasons!

dvdplm:
@niklasad1 makes sense, especially considering that `String` is not `Copy`. My (wrong) thinking was this: perhaps there are places in the code where the `Password` is copied as part of a method call and so if we are worried about the data lingering in memory we might want to zero out the mem of the copy asap instead of waiting for it to go out of scope. Brainfart. :)

andresilva:
@niklasad1 I think what @dvdplm meant was to use `drop` to make sure the value is dropped as soon as possible (before end of scope), the idea would be to minimize the amount of time sensitive information remains in memory. I didn't check the "owning" sites to see if there was any place where it made sense to drop eagerly.

PR #8924, opened 2018-06-19
simplify kvdb error types

debris:
We never recover from database errors, so there is no need to have a special type for them. Let's keep it simple and use `io::Error` instead


andresilva:
Please rebase or merge master.

debris:
rebased

PR #8926, opened 2018-06-19
Update Ropsten forkblock at 3383558

Tbaut:
Prevents clients from syncing the wrong side of the ropsten fork and get stuck.


andresilva:
Warp synced successfully with `--warp-barrier 3460000` and currently importing. 👍

PR #8932, opened 2018-06-20
Restrict vault.json permssion to owner and using random suffix for temp vault.json file

sorpaas:
In an unlucky situation, multiple process (which may be incorrectly started to run on the same chain) can decide to write to `vault_temp.json` file at the same time, resulting in a corrupted file. We don't protect this in key files right now, but given the original intention to write to `vault_temp.json` and then atomically rename to `vault.json` is to prevent similar type of thing, I think it would work better to use a random suffix if `vault_temp.json` file already exists.

This also restricts newly created `vault.json` file with permission owner only.


5chdn:
Needs a 2nd review.

PR #8934, opened 2018-06-20
Minimal effective gas price in the queue

tomusdrw:
For small (and saturated) queues this optimizes the verification pipeline to avoid recovering a sender in case we are quite sure that the transaction won't get into the pool anyway.


transaction-pool/src/pool.rs
```diff
@@ -393,6 +393,26 @@ impl<T, S, L> Pool<T, S, L> where
...
+		let is_full = {
+			self.by_hash.len() > threshold(self.options.max_count)
+			|| self.mem_usage > threshold(self.options.max_mem_usage)
+		};
+
+		if !is_full {
+			return None
+		}
+
+		self.worst_transactions.iter().next().map(|x| x.score.clone())
```
ascjones: Using `next()` here which looks correct (`worst_transactions` is ordered by ascending `score`). However I noticed: https://github.com/paritytech/parity/blob/0335e927e5d2133ceba4b54579094f1b4d810eac/transaction-pool/src/pool.rs#L277
Is the `next_back()` wrong then? Seems like it would check for the least worst transaction to replace.

tomusdrw: Great catch, I'll prepare a test and fix that issue with `remove_worst`

tomusdrw: Actually it's correct in `remove_worst`. `ScoreWithRef` has a custom `Ord` implementation that is ascending if the internal score is descending. So the error is actually here, will make sure to add a correct test to cover this.

ascjones: Looks good. Should this one also change to `next_back()` then?

https://github.com/paritytech/parity/blob/b56200916dfb90149ab22901dac5ef951688974b/transaction-pool/src/pool.rs#L393

5chdn:
Needs a 2nd review.

tomusdrw:
@ascjones thanks for the review, the issue with `worst_transaction` is addressed now.

jimpo:
Code looks fine, but I think this can be better encapsulated in scoring. Instead of passing the gas price into the verifier, could the queue just check `should_replace` against the pool's `worst_transaction` directly? This is more in line with the check that happens in `import` anyway, and this way the score <-> gas price equivalence doesn't leak as much.

tomusdrw:
@jimpo thanks for the review, I've updated the PR a bit:
- Removed the threshold (it didn't have much sense)
- Moved the decision making to the `Scoring` implementation.

Unfortunately we are not able to re-use `should_replace`, since the whole point is that it's just a heuristic to quickly reject some transactions, we want to do it without recovering the sender in verification, so there is no way to check if the nonce is lower/higher.

That said it's not really that much of an issue, the pool decisions are always local and are not part of the consensus itself. To have the most accurate image of the pending transactions one should run with big enough pool to store them all.

PR #8940, opened 2018-06-21
Allow transactions, when a configured transactionPermissionContract does not exist

wulfraem:
We wanted to use the `transactionPermissionContract` and deployed a matching contract. Then added the contract to the chain specification. Clients, that already had a synced chain were able to sync, but clients, that performed a full sync, stopped at the first transaction (in our case at block 5), because the configured `transactionPermissionContract` did not exist at this time. If the contract does not exist, the check defaults to "not permitted" due to an error, when calling the non-existing contract, this pull request checks if the contract exists and defaults to true in such a case.

This fixes our issue, but i guess a useful extension to the `transactionPermissionContract` logic would be to use a similar config format like the [Multi-Set Validator-Set](https://wiki.parity.io/Validator-Set.html#multi-set) and define starting blocks for valid permission contracts. This would allow to change the contracts and still be able to sync the chain.

### DIFF
--- a/ethcore/src/tx_filter.rs
+++ b/ethcore/src/tx_filter.rs
@@ -85,6 +85,11 @@ impl TransactionFilter {
 		}
 
 		let contract_address = self.contract_address;
+		if !client.code_hash(&contract_address, BlockId::Hash(*parent_hash)).map_or(false, |c| c != KECCAK_EMPTY) {
+			// transaction permission contract does not exist at current block, permit transaction
+			return true;
+		}
+
 		let contract_version = contract_version_cache.get_mut(parent_hash).and_then(|v| *v).or_else(||  {
 			self.contract.functions()
 				.contract_version()


ethcore/src/tx_filter.rs
```diff
@@ -85,6 +85,12 @@ impl TransactionFilter {
 		}
 
 		let contract_address = self.contract_address;
+		if !client.code_hash(&contract_address, BlockId::Hash(*parent_hash)).map_or(false, |c| c != KECCAK_EMPTY) {
+			trace!(target: "tx_filter", "transaction permission contract does not exist at current block, permitting transaction");
+			return true;
+		} else {
+			trace!(target: "tx_filter", "transaction permission contract exists at current block, checking permissions");
```
grbIzl: I suggest removing this trace. It will spam the log

wulfraem: Makes sense. I removed the logs.

wulfraem:
This pull request was intended to fix the syncing issue, that we had with the current implementation.

In the second part of my initial suggested a config similar to the config of the [Multi-Set Validator-Set](https://wiki.parity.io/Validator-Set.html#multi-set). This would allow to swap the permission contracts over time without creating multiple transitions for each new contract.

5chdn:
I'm not overly familiar with the transaction permissioning, but what @andresilva suggests is that you either have set contract enabled right from the start (would make this PR obsolete) or you want to make it highly configurable by defining some transaction permission contract "transition" logic allowing users to specify at which block which contract is being used.

5chdn:
Marking as _stale_, let me know if you have any further questions.

andresilva:
Replaced by #9275.

PR #8943, opened 2018-06-21
rpc: cap gas limit of local calls

### LINKED ISSUE #8311: eth_calls to some addresses seem to never return, they just hang indefinitely
> I'm running:
>
> - **Which Parity version?**: 1.8.7 , nightly, beta
> - **How installed?**: docker

**actual** 
making and `eth_call` for `name()` or `symbol()` on address `0x6a0a0fc761c612c340a0e98d33b37a75e5268472` just hangs forever. 
one test ran for 8 hours not returning anything. trying the same curl in geth seems instant.

**expected behavior**
to return something, be it the result or an error... anything

**steps to reproduce**
```
curl --data '{"method":"eth_call","params":[{"to":"0x6a0a0fc761c612c340a0e98d33b37a75e5268472","data":"0x06fdde03"}, "latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
```

thank you very much

/edit: tested various configurations: default, archive+tracing, no-ancient-blocks

### LINKED ISSUE #8904: eth_call some contract stuck and high CPU

> I'm running:
>
> - **Which Parity version?**: Parity/v1.10.6-stable-bc0d134-20180605/x86_64-linux-gnu/rustc1.26.1
> - **Which operating system?**:  Linux
> - **How installed?**: curl

curl --data '{"method":"eth_call","params":[{"data":"0x95d89b41","to":"0x532975d56cf18f597480e2521246b273ad9ae348"}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST http://localhost:8545

This will cause the parity CPU high and never respond. If I run this a few times, the parity won't accept any new RPC calls, I have to restart it.

andresilva:
Currently we have effectively unlimited gas for local calls, this leads to issues with unbounded loops in contracts that never end. I've decreased the cap to 500 million gas, which should still be more than enough for local calls.

Fixes #8904, fixes #8311.

### DIFF
--- a/rpc/src/v1/helpers/fake_sign.rs
+++ b/rpc/src/v1/helpers/fake_sign.rs
@@ -15,7 +15,6 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use transaction::{Transaction, SignedTransaction, Action};
-use ethereum_types::U256;
 
 use jsonrpc_core::Error;
 use v1::helpers::CallRequest;
@@ -29,7 +28,7 @@ pub fn sign_call(request: CallRequest, gas_cap: bool) -> Result<SignedTransactio
 		}
 		Some(gas) => gas,
 		None if gas_cap => max_gas,
-		None => U256::from(2) << 50,
+		None => max_gas * 10,
 	};
 	let from = request.from.unwrap_or(0.into());
 


debris:
@tomusdrw do you agree with this change?

PR #8944, opened 2018-06-21
EVM benchmark utilities

sorpaas:
rel #8914, #6744

Expose jsontests helpers from ethcore so that they can be used elsewhere. Allow hooking to start/stop point of each test so that time information can be recorded. Add a command in `evmbin`:

```
parity-evm stats-jsontests-vm <file> [--folder]
```

This runs executive standard jsontests against a specific folder/file, record the timing information, and then output them in tsv format. This would allow those benchmark info to be compared.

`evmbin/bench.sh` is moved to `scripts`, and a new script is added, called `evm_jsontests_bench.sh`, which runs the above command on all available vm jsontests.

I also changed all scripts to use `/usr/bin/env` instead of `/bin/bash`. The later would mean it breaks on all non-FHS Linux systems (like NixOS).


ethcore/src/json_tests/chain.rs
```diff
@@ -23,12 +24,26 @@ use miner::Miner;
 use io::IoChannel;
 use test_helpers;
 
-pub fn json_chain_test(json_data: &[u8]) -> Vec<String> {
+use super::HookType;
+
+/// Run chain jsontests on a given folder.
+pub fn run_test_path<H: FnMut(&str, HookType)>(p: &Path, skip: &[&'static str], h: &mut H) {
```
dvdplm: There are 7 identical `pub fn run_test_path` added in this PR – would it be possible to avoid some boilerplate and have a single copy?

ethcore/src/json_tests/test_common.rs
```diff
@@ -21,7 +21,20 @@ use std::path::Path;
 use std::ffi::OsString;
 pub use ethereum_types::{H256, U256, Address};
 
-pub fn run_test_path(p: &Path, skip: &[&'static str], runner: fn (json_data: &[u8]) -> Vec<String>) {
+/// Indicate the type of the hook passed to test functions.
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub enum HookType {
+	/// Hook passed on starting of a test.
```
dvdplm: Perhaps this is better? "Hook to code to run on test start"

ethcore/src/json_tests/test_common.rs
```diff
@@ -21,7 +21,20 @@ use std::path::Path;
...
+	OnStart,
+	/// Hook passed on a test is finished.
```
dvdplm: Perhaps this is better? "Hook to code to run on test end"

ethcore/src/json_tests/test_common.rs
```diff
@@ -21,7 +21,20 @@ use std::path::Path;
 use std::ffi::OsString;
 pub use ethereum_types::{H256, U256, Address};
 
-pub fn run_test_path(p: &Path, skip: &[&'static str], runner: fn (json_data: &[u8]) -> Vec<String>) {
+/// Indicate the type of the hook passed to test functions.
```
dvdplm: Perhaps "Indicate when to run the hook passed to test functions."?

evmbin/src/main.rs
```diff
@@ -76,6 +77,9 @@ State test options:
     --only NAME        Runs only a single test matching the name.
     --chain CHAIN      Run only tests from specific chain.
 
+Stats jsontests VM options:
+    --folder           Run jsontest for a folder instead of a single file.
```
dvdplm: Could this not be inferred from the input path? If it's a file => run the file; if it's a folder => run all tests within?

evmbin/src/main.rs
```diff
@@ -61,6 +61,7 @@ EVM implementation for Parity.
 Usage:
     parity-evm state-test <file> [--json --std-json --only NAME --chain CHAIN]
     parity-evm stats [options]
+    parity-evm stats-jsontests-vm <file> [--folder]
```
dvdplm: The name of the command in itself does not seem sufficient to understand what this does. Not sure where the best place to put a proper description of what it does and what the output format is would be, but I think it'd be very useful.

evmbin/src/main.rs
```diff
@@ -99,6 +105,40 @@ fn main() {
 	}
 }
 
+fn run_stats_jsontests_vm(args: Args) {
+	use json_tests::HookType;
+	use std::collections::HashMap;
+	use std::time::{Instant, Duration};
+
+	let file = args.arg_file.expect("FILE is required");
```
dvdplm: "FILE (or PATH) is required"?

evmbin/src/main.rs
```diff
@@ -99,6 +105,40 @@ fn main() {
...
+
+	let mut times: HashMap<String, (Instant, Option<Duration>)> = HashMap::new();
```
dvdplm: nitpick: I suggest calling this `timings`.

ethcore/src/json_tests/chain.rs
sorpaas: Those functions are wrappers for the parent `run_test_path` function. Before none of them are exposed, but I figured it would be helpful to add them there, as it simplify library user's work to figure out how those functions are meant to be used.

They're not entirely identical -- the function parameter passed is different, determining which type of tests it operates on. But I'm open to any idea if we can make this better and avoid duplications. :)

PR #8947, opened 2018-06-22
WIP Cython parity-clib example Ref Issue 8604

oogetyboogety:
Needs a makefile, but should be okay to test


oogetyboogety:
[clabot:check]

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        self.params = <ParityParams*> PyMem_Malloc(4*sizeof(double))
+
+    def __init__(self, args):
+        cdef size_t length 
+        cdef size_t* arg_lens  
+        cdef char** c_args 
+        cdef void* out
+        length = args.len()
+        for i in range(length):
+            py_byte_string = args[i].encode('UTF-8')
```
tomaka: I don't really know python/cython, but this looks like a dangling pointer.

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+#    cdef void* confguration
+#    def __cinit__(self, configuration, ):
+#        self.configuration = 
+from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
+cdef class ParityInterface:
+    cdef ParityParams* params
+    cdef void* parity
+
+    def __cinit__(self):
+        self.params = <ParityParams*> PyMem_Malloc(4*sizeof(double))
```
tomaka: That's not very future-proof! It should be based on the side of `ParityParams`.

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        cdef size_t* arg_lens  
+        cdef char** c_args 
+        cdef void* out
+        length = args.len()
+        for i in range(length):
+            py_byte_string = args[i].encode('UTF-8')
+            c_args[i] = py_byte_string
+            arg_lens[i] = py_byte_string.len()
+        params_success =  parity_config_from_cli(&(c_args[0]), &(arg_lens[0]), length, &out)
+        self.params = <ParityParams*>out
```
tomaka: That's not good! `out` should be written in `self.params.configuration`. There is no guarantee at all about the type of what is pointed by `out`.

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        #paritylib_config_from_cli(self, 
+        success = parity_start(self.params, &self.parity)
+        print(success)
+
+    def paritylib_destroy(self):
```
tomaka: This should be in a destructor (`__dealloc__()`), I think.

parity-clib-python-example/include/parity.h
```diff
@@ -0,0 +1,105 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
```
tomaka: Can't this `parity.h` file be included from the `parity-clib` directory?

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        return <object>parity_destroy(self.parity)
+
+    def paritylib_rpc(self, unicode rpc_call):
+        cdef char* c_rpc_call 
+        cdef size_t rpc_call_len = rpc_call.len()
+        cdef size_t* out_len
+        cdef char* out_str  
+        c_rpc_call[0] = rpc_call.encode('UTF-8')
```
tomaka: I guess this should be `c_rpc_call = ...` and not `c_rpc_call[0] = ...`.
Also, you probably have a dangling pointer here as well, although I don't know cython so I'm not sure if the value returned by `encode` is garbage collected or not.

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        #paritylib_config_from_cli(self, 
+        success = parity_start(self.params, &self.parity)
+        print(success)
+
+    def paritylib_destroy(self):
+        return <object>parity_destroy(self.parity)
+
+    def paritylib_rpc(self, unicode rpc_call):
+        cdef char* c_rpc_call 
+        cdef size_t rpc_call_len = rpc_call.len()
```
tomaka: I think this should be the length *after* the encoding as UTF-8, unless you are sure that `rpc_call` is already in UTF-8?

parity-clib-python-example/libparity.pyx
```diff
@@ -0,0 +1,58 @@
...
+        cdef size_t* out_len
+        cdef char* out_str  
+        c_rpc_call[0] = rpc_call.encode('UTF-8')
+        success = parity_rpc(&self.parity, c_rpc_call, rpc_call_len, out_str, out_len)
```
tomaka: You're passing uninitialized pointers :-/ This will likely crash.
You have to allocate a buffer for the output string, and pass its length and a pointer to it.

parity-clib-python-example/parity.h
```diff
@@ -0,0 +1,105 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
```
tomaka: Same remark as the other `parity.h`.
It would be better not to duplicate files.

oogetyboogety:
@tomaka So I thought I fixed most of the memory leaks, but now whenever Rust tries to access the pointer stored inside the Python class, like in parity_destroy or parity_rpc, it segfaults. I'll try stepping through with a debugger to see how to fix it.

parity-clib-python-example/include/parity.h
oogetyboogety: for some reason, this only worked when I patched this file to use a c typedef struct instead of the regular struct. Apparently, the type ParityParams requires it to be a typedef? I'm just a bit confused as well, can we use a struct as a type without marking it with typedef?

Thanks so much for all your help

5chdn:
@tomaka could you assist @oogetyboogety here?

PR #1765, opened 2016-07-29
Disable WAL

arkpar:
https://github.com/facebook/rocksdb/wiki/Known-Issues
"Atomicity is not guaranteed after DB recovery for more than one multiple column families and WAL is disabled."

This makes import 5% slower but guarantees DB recovery.
should also fix #1760

PR #8963, opened 2018-06-24
db: remove wal disabling / fast-and-loose option.

### LINKED ISSUE #8779: Change name/description of fast_and_loose.
The config option fast-and-loose has a name and documentation that currently implies that the node can be restarted, it just isn't resilient to failure.  This makes it sound like a great solution for running a dev chain or test chain, and lead me (and others #7910) to try to use it that way only to find out after much troubleshooting that that isn't the case.

My particular use case is to create a Parity docker image that has some base contracts pre-delpoyed to it that I can then test against.  For example, one might create a Dockerfile like:
```Dockerfile
FROM parity/parity:v1.11.1

WORKDIR /
COPY dev-key.json /parity/keys/DevChain/dev-key.json
COPY chain-spec.json /parity/chain-spec.json
COPY config.toml /parity/config.toml
RUN echo "" > /parity/password

COPY deploy-contracts.sh /app/deploy-contracts.sh
RUN chomd 555 /app/deploy-contracts.sh
RUN /parity/parity --config /parity/config.toml \
    && /app/deploy-contracts.sh \
    && kill -TERM %1

ENTRYPOINT [ "/parity/parity", "--config", "/parity/config.toml" ]
```

I would expect fast-and-lose to generally work here because I stop Parity once cleanly after deploying contracts, then when the container is run in the future it will start from that state (with contracts deployed).  However, with fast-and-loose in `config.toml`, this doesn't actually work.

The poor naming/documentation around this feature resulted in notable engineering time spent debugging.

Side note: It would be preferable if there were a mode that actually did what this implied, and supported clean exits but made no guarantees about unclean exits.  That would allow the above use-case to actually work.

5chdn:
fix #8779

ref #1765


andresilva:
I'd actually say we remove it since this is really unsafe and will just give us trouble (more issues with database corruption). It's better to just provide a bigger cache size for performance.

5chdn:
Hold on.

parity/cli/mod.rs
```diff
@@ -883,6 +879,10 @@ usage! {
 			"--warp",
 			"Does nothing; warp sync is enabled by default. Use --no-warp to disable.",
 
+			FLAG flag_fast_and_loose: (bool) = false, or |_| None,
+			"--fast-and-loose",
+			"Does nothing. DB WAL is always activated.",
```
andresilva: Maybe mention here that the option is deprecated?

5chdn: This is just a placeholder really since `--help` does not print these options anymore.

5chdn:
@andresilva done :blush:

5chdn:
- [ ] needs proper deprecated handling, ref #9036

5chdn:
ok, added to deprecated::does_nothing() :+1:

parity/deprecated.rs
```diff
@@ -203,6 +203,12 @@ pub fn find_deprecated(args: &Args) -> Vec<Deprecated> {
 		result.push(Deprecated::Removed("--tx-queue-ban-time"));
 	}
 
+	// Removed in 2.0.
+
+	if args.flag_fast_and_loose {
+		result.push(Deprecated::DoesNothing("--fast-and-loose"));
```
andresilva: Minor nit, but why `DoesNothing` instead of `Removed`?

5chdn: Correct. It should be removed.

5chdn:
FFFFFFFFFFFUUU... :rofl:

dvdplm:
**NOTE** We should merge https://github.com/paritytech/parity-common/pull/7 before merging this. If you came here to review, please consider reviewing that one instead.

Issue #2913, opened 2016-10-27
Config options without CLI behind them


tomusdrw:
It should be possible to create config entries without creating additional CLI options.
or
We should have `--help` and `--help --verbose` :)

guanqun:
This looks like small enough issue for me to dive in, but I don't quite get what you mean. Do you mean we should support `--help --verbose` instead of reporting `Unknown flag`?

rphmeier:
@guanqun Right now our help message is really big. The feature here is to reduce the CLI help to only essential options unless specifying `--verbose` to make the message more approachable.

guanqun:
Ah, I got it. It's like `git help` vs `git branch --help`. :)

rphmeier:
Yep, something like that. I think this would be an awesome opportunity to port our CLI to [clap](https://github.com/kbknapp/clap-rs) over docopt, but that would be a much larger change.

tomusdrw:
The issue here is also that currently any option that is available in config file needs to have a CLI argument, we can start with a possibility to have config options without CLI flag behind them and move some advanced CLI options to config files.

guanqun:
From the comments we want to do two things:

- convert docopt to clap
- remove some advanced CLI arguments. For this one, I don't think I'm in the right position to say which one should be removed. :) Hopefully someone can come up a list.

axelchalon:
I'm currently porting the CLI to clap (#2066) and it will then be possible to have options that can only be set from config files. It would also be possible to find a middle ground by having each subcommand have its own help message (i.e. ``parity account --help``) for example.

debris:
any update on this?

axelchalon:
Marek:

**Separating global help from help of subcommands**
`parity --help` only shows global options, and the options for a specific subcommand are shown by `parity account --help` for example, so this already reduces the size of the help outputs in general.

**Hiding some options from the CLI help output**
Currently, the `parity --help` output (for the global options only) is generated by us (this lets us divide the options into categories in the help output, etc.), so we could decide for example to not show the options if they, say, have an empty help message. This would be easy to implement. For this to also work for subcommands help outputs (`parity account --help`), we could generate those outputs ourselves rather than let Clap automatically do it, or more simply internally use `.hidden()` on clap args. Anyway, this would be doable.
I would imagine that it's best to show all the available options in `--help` but do you have some in mind that shouldn't appear?

**Short help vs long help**
[...]

PR #8967, opened 2018-06-24
parity: hide legacy options from cli --help

5chdn:
ref #2913


jimpo:
Looks fine. Maybe worth adding a `verbose: bool` param to `print_help` and displaying legacy options if it is true? (as per discussion in #2913)

5chdn:
I'd like to give the --verbose option more love. For now, I would like to just get rid of the legacy options.

Initially, the --verbose idea was to actually deliver some background information. This PR is rather unrelated to #2913 except for the point that I raised this in the discussion.

PR #8977, opened 2018-06-25
Fix deadlock in blockchain.

tomusdrw:
From https://docs.rs/parking_lot/0.6.2/parking_lot/type.RwLock.html

> This lock uses a task-fair locking policy which avoids both reader and writer starvation. This means that readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock. Because of this, **attempts to recursively acquire a read lock within a single thread may result in a deadlock.**

And it seems that `first_block_number()` actually goes through `block_number(_)` where we check the `best_block` again

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -1450,18 +1450,24 @@ impl BlockChain {
 
 	/// Returns general blockchain information
 	pub fn chain_info(&self) -> BlockChainInfo {
+		// Make sure to call internal methods first to avoid
+		// recursive locking of `best_block`.
+		let first_block_hash = self.first_block();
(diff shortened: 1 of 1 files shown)


5chdn:
Please indicate whether we need to backport this.

tomusdrw:
@5chdn we do, label added.

debris:
When was this bug introduced? It seems like parity should deadlock every time this function is called

tomusdrw:
@debris it deadlocks randomly and only when it happens that there is a `write()` lock request in between the two reads. I think the change was introduced in my header refactoring PR I remember touching some stuff with the `best_block` code.

5chdn:

PR #8980, opened 2018-06-25
Improve should_replace on NonceAndGasPrice

jimpo:
The current implementation of `should_replace` calls `choose` with transactions between potentially different senders, which violates the expectation. As a result, transactions may get replaced incorrectly with ones with a lower gas price if the nonces differ.

Noticed during review of #8934.


5chdn:

PR #8666, opened 2018-05-19
Update `license header` and `scripts`

niklasad1:
Sorry for a huge `PR` but most of the changes are caused by updated scripts.

The major changes in this PR are:

1. Updated `scripts/add_license.sh` it will now replace the existing header it starts with `// Copyright` and ends with `If not, see <http://www.gnu.org/licenses/>.`  or just add the header if no header is found
2. Added a new script that supresses duplicated empty lines 

So, the most of important parts of this PR to review are:

1. [scripts/add_license.sh](https://github.com/niklasad1/parity/blob/license-header/update/scripts/add_license.sh)
2. [scripts/remove_duplicate_empty_lines.sh](https://github.com/niklasad1/parity/blob/license-header/update/scripts/remove_duplicate_empty_lines.sh)

PR #3423, opened 2016-11-14
Enable Panic=abort

arkpar:

PR #499, opened 2016-02-22
Remove unneeded locking

arkpar:
Reduced lock contention a bit.
There is no global lock on the blockchain anymore, instead there is a lock on the `insert_block` function that guarantees that all updates are atomic.
Reads are are also atomic: database changes are committed with the transaction. Most of the caches that are addressed by block hash can be updated individually. The only exceptions are `best_block`, `block_hashes` and `transaction_addresses`. These are updated alltogether.
Block queue has been optimized to use more fine grained locking.
Global lock on the block queue has been removed. It was not needed in the first place since the module is fully reenterable.

PR #8983, opened 2018-06-26
Use local parity-dapps-glue instead of crate published at crates.io

sorpaas:


5chdn:
Does this only affect master?

sorpaas:
@5chdn We still need to wait for the CI to complete to know whether this is the cause of the problem. (And this change fixes a dependency issue regardless whether it's the cause of the CI failures.)

If so, then I think it's due to cargo/rust update, specifically, a problem for querying doctest dep versions. It probably affects every build using Rust 1.27. (But there's still a chance that this may not be what caused the CI failures and I might be wrong here.)

ordian:
I'm not sure whether this change can fix build issue, since all the [commits](https://github.com/paritytech/parity/commits/master/dapps/js-glue) to js-glue (#8666, #8791, #7355) after 1.9.1 was published, should not affect default build.

sorpaas:
Ah right. Then I guess this needs to wait for cache cleanup right now. I'm going to reset to HEAD^.

sorpaas:
Are there any more benefits to use panic=abort rather than binary size? I found out that it is introduced in #3423, and we had around 2mb binary gains, which, given our current binary size, doesn't look significant any more. It looks like we're one of the main project that suffers from the panic strategy error (searching google about this shows mostly parity repo on the front page). Would we consider to drop panic=abort?

I also noticed that lto optimization is disabled for `profile.release`, and it's introduced in #499. I didn't find explanations for this, however. Just curious, are there any particular reasons we are having this settings?

cc @arkpar

sorpaas:
Local testing worked for me. I'm going to push the commit that disables panic=abort and enable lto optimization, just to see how it works. Looks like this also has positive effect in incremental build.

ordian:
This is really strange, what command is failing? Is it `cargo test --features "json-tests" --all --release`? Because it works for me locally.

sorpaas:
@ordian I ran `./test.sh`. Without commit https://github.com/paritytech/parity/pull/8983/commits/8023e35a187ad195a9803431a887874ac2430590, it emits the same "... the panic strategy `abort` which is incompatible with this crate's strategy of `unwind` ..." error message. Removing panic=abort fixes the issue.

My opinion on this is that panic=abort is probably not mature enough on Rust yet. While aborting saves some binary size and compile time, unwinding allows possibly more control for how the program panics (ref https://users.rust-lang.org/t/panic-unwind-vs-abort/9928). There're also other projects, like ripgrep (https://github.com/BurntSushi/ripgrep/commit/d5c045469b8f59f007a1c5c265edc5598858d2e0) took the same reversion.

Cargo.toml
```diff
@@ -115,12 +115,9 @@ path = "parity/main.rs"
 name = "parity"
 
 [profile.dev]
-panic = "abort"
 
 [profile.release]
 debug = false
-lto = false
```
ordian: I'm afraid with lto enabled build machines can ran out of memory.

debris: according to [documentation](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) `lto = false` is the default

sorpaas:
@arkpar I think unwinding shouldn't impact runtime speed in any significant way, because it only affects the code generated when a panic actually happened (which won't do in normal execution). Can you be more specific on how you tested it's 5% faster?

arkpar:
Import 1m blocks from a file.
I'm not sure if rust stack unwinding is completely free for non-throwing code. Usually stack unwinding requires some additional steps in function prologue to setup the frame information. Reduced binary size is also a factor as it increases cache hit rate.

arkpar:
Database has to be resistant to power loss anyway.

sorpaas:
Reading this comment (https://github.com/rust-lang/rfcs/pull/1765#discussion_r82326901), for runtime performance, the difference would only be the landing pads (but it can still result in reasonable performance gain).

sorpaas:
Reading this story (https://www.reddit.com/r/rust/comments/55ns2m/safe_and_efficient_bidirectional_trees/) it looks like landing pads are inserted together with `unwrap` and `panic!`. So with less of them we get less landing pads.

No matter whether we choose to use `panic=abort` or `panic=unwind`, it may be a good optimization to try to reduce the amount of panics we have in code, and try to move them to be checked by type system instead.

PR #8995, opened 2018-06-27
Preserve the current abort behavior

sorpaas:
While removing panic=abort fixed the build, another issue is that our process is currently expected to be aborted if any of the threads fail. So this adds explicit process abort at the end of the panic hook.


sorpaas:
Realized this doesn't need to be patched to stable or beta. We used a different method to fix CI build there: https://github.com/paritytech/parity/pull/8986/commits/4c3410eb0c0098f4a7f379a6afcfe79e3d82a975 (but as it "hides" a crate in the workspace, we probably cannot take that approach on master.)

Issue #8994, opened 2018-06-27
Multi ValidatorSet not working as expected, doesn't validate new nodes but uses old contract


S3bb1:
> I'm running:
>
> - **Which Parity version?**: 1.11.4
> - **Which operating system?**: Windows / Linux
> - **How installed?**:  binaries / from source
> - **Are you fully synchronized?**:  yes
> - **Which network are you connected to?**: private PoA network
> - **Did you try to restart the node?**: yes

Hi there i'm currently testing the validator set multi option on AuRa ... my current workflow is

1. Deploy new OwnedSet Contract
2. Add to chain spec at a future block ( e.g.
```
     "multi": {
         "0": {
             "safeContract": "0x0000000000000000000000000000000000000005"
         },
         "37270": {
             "safeContract": "0x6a1c3cf62a5e5f36516f2687144445b64c9ecbdc"
         }
     }
```
3. restart mining nodes
4. wait until block is reached
see the log message that a new validator contract is active
5. add new validator to contract (with initiateChange event)
but now the new validator doesn't signs any new blocks ... instead the logs say that when emptysteps are done, the old validator set is active
```
signing nodes active in old contract: [0x5307…46e0, 0x4a67…a83d]
signing nodes active in new contract [0x4a67…a83d,0x00aC5...805B]
```
```
2018-06-27 10:49:27 Fetched proposer for step 510029789: 0x5307…46e0
2018-06-27 10:49:27 IO Worker TRACE engine generate_seal: 0x4a67…a83d not a proposer for step 510029789.
2018-06-27 10:49:30 IO Worker DEBUG engine Zooming to epoch for block 0x64d0…9ed4
2018-06-27 10:49:30 IO Worker TRACE engine Fetched proposer for step 510029790: 0x4a67…a83d
```
any ideas why for the empty steps the old validator contract is used?

andresilva:
I haven't tested using empty steps with a validator set yet. I think what you're seeing is a bug [here](https://github.com/paritytech/parity/blob/stable/ethcore/src/engines/authority_round/mod.rs#L1079). I'm working on this ATM so should have a fix soon.

S3bb1:
Cool, i will check if the problem still occurs if i setup a new chain without emtpy steps to proof that it comes from the empty step feature.

S3bb1:
i tried it again without empty steps and the issue is the same:

[...]

Issue #8065, opened 2018-03-06
Error when updating validators contract


afiorenza:
> I'm running:
>
> - **Which Parity version?**: 1.9.4-beta
> - **Which operating system?**: Linux (Ubuntu)
> - **How installed?**: via installer
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: Private
> - **Did you try to restart the node?**: no

I have a PoA chain running with 5 nodes. This is my current genesis:

```
{
  "name": "Bitsign",
  "engine": {
    "authorityRound": {
      "params": {
        "stepDuration": "5",
        "blockReward": "0xDE0B6B3A7640000",
        "validators" : {
          "safeContract": "0x0000000000000000000000000000000000000005"
        }
      }
    }
  },
  "params": {
    "gasLimitBoundDivisor": "0x400",
    "maximumExtraDataSize": "0x20",
    "minGasLimit": "0x1388",
    "networkID" : "0x2323"
  },
  "genesis": {
    "seal": {
      "authorityRound": {
        "step": "0x0",
        "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
      }
    },
    "difficulty": "0x20000",
    "gasLimit": "0x5B8D80",
    "timestamp": 1512771499
  },
  "accounts": {
    "0000000000000000000000000000000000000001": { "balance": "1000000000000000000", "builtin": { "name": "ecrecover", "pricing": { "linear": { "base": 3000, "word": 0 } } } },
    "0000000000000000000000000000000000000002": { "balance": "1000000000000000000", "builtin": { "name": "sha256", "pricing": { "linear": { "base": 60, "word": 12 } } } },
    "0000000000000000000000000000000000000003": { "balance": "1000000000000000000", "builtin": { "name": "ripemd160", "pricing": { "linear": { "base": 600, "word": 120 } } } },
    "0000000000000000000000000000000000000004": { "balance": "1000000000000000000", "builtin": { "name": "identity", "pricing": { "linear": { "base": 15, "word": 3 } } } },
[...]

PR #8998, opened 2018-06-27
Fixes for misbehavior reporting in AuthorityRound

andresilva:
- Use the epoch set when validating
- Always report on the contract (`self.validators`)
- Use correct epoch number when reporting
- Report skipped primaries when generating seal (otherwise the current sealer won't report misbehavior since it won't verify his own block)

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -1114,6 +1114,15 @@ impl Engine<EthereumMachine> for AuthorityRound {
 
 		match verify_timestamp(&self.step.inner, header_step(header, self.empty_steps_transition)?) {
 			Err(BlockError::InvalidSeal) => {
+				// This check runs in Phase 1 where there is no guarantee that the parent block is
+				// already imported, therefore the call to `epoch_set` may fail. In that case we
+				// won't report the misbehavior but this is not a concern because:
+				// - Only authorities can report and it's expected that they'll be up-to-date and
+				//   importing, therefore the parent header will most likely be available
+				// - Even if you are an authority that is syncing the chain, the contract will most
+				//   likely ignore old reports
+				// - This specific check is only relevant if you're importing (since it checks
+				//   against wall clock)
 				if let Ok((_, set_number)) = self.epoch_set(header) {
 					self.validators.report_benign(header.author(), set_number, header.number());
 				}
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -916,17 +916,17 @@ impl Engine<EthereumMachine> for AuthorityRound {
 		}
 
 		let active_set;
-		let validators = if self.immediate_transitions {
-			&*self.validators
+		let (validators, set_number) = if self.immediate_transitions {
+			(&*self.validators, header.number())
 		} else {
 			match self.epoch_set(header) {
 				Err(err) => {
 					warn!(target: "engine", "Unable to generate seal: {}", err);
 					return Seal::None;
 				},
-				Ok((set, _)) => {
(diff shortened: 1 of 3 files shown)


andresilva:
Maybe fixes #8994 and #8065. I haven't been able to replicate so far.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1060,12 +1098,14 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
-		// TODO [ToDr] Should this go from epoch manager?
-		// If yes then probably benign reporting needs to be moved further in the verification.
-		let set_number = header.number();
-
 		match verify_timestamp(&self.step.inner, header_step(header, self.empty_steps_transition)?) {
 			Err(BlockError::InvalidSeal) => {
+				let set_number = if self.immediate_transitions {
+					header.number()
+				} else {
+					self.epoch_set(header)?.1
```
tomusdrw: Is it guaranteed that parent block is already imported when we are doing `block_basic` verification? If not then zooming to the epoch set will most likely fail. Would be good to have a test case for this one.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1120,59 +1167,36 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
 	}
 
 	// Check the validators.
 	fn verify_block_external(&self, header: &Header) -> Result<(), Error> {
-		// fetch correct validator set for current epoch, taking into account
-		// finality of previous transitions.
 		let active_set;
-		let validators = if self.immediate_transitions {
-			&*self.validators
+		let (validators, set_number) = if self.immediate_transitions {
```
tomusdrw: This pattern seems to repeat couple of times, might be worth to extract it to a function?

andresilva: I've refactored part of it into the `epoch_set` function. Unfortunately it's not easy to get the whole pattern into a function since both branches are returning different types (`&ValidatorSet` vs `SimpleList`). It could work by returning `Box<ValidatorSet>` from the function but I can't clone `self.validators`.

ethcore/src/engines/authority_round/mod.rs
andresilva: Didn't think of this, so it's possible that `verify_block_basic` is done in parallel and therefore the parent might not be imported yet? This verification must be done in phase 1 since we want `TemporarilyInvalid` errors to be emitted at this phase (otherwise it won't recover the sync process).

5chdn:
Could you give me an ETA?

andresilva:
I think this is good to review:
- Created a `CowLike` type to return borrowed or owned data and refactor the `epoch_set` duplication
- Let reporting fail on `verify_block_basic`:
  - Only authorities can report and it's expected that they'll be up-to-date and importing
  - Even if you are an authority that is syncing the chain the contract will ignore old reports
  - This specific check is only relevant if you're importing (since it checks against wall clock)

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1114,8 +1114,10 @@ impl Engine<EthereumMachine> for AuthorityRound {
 
 		match verify_timestamp(&self.step.inner, header_step(header, self.empty_steps_transition)?) {
 			Err(BlockError::InvalidSeal) => {
-				let set_number = self.epoch_set(header)?.1;
-				self.validators.report_benign(header.author(), set_number, header.number());
+				if let Ok((_, set_number)) = self.epoch_set(header) {
```
tomusdrw: Maybe add a comment *why* it may fail and why it's ok?

andresilva: Done.

Issue #8981, opened 2018-06-26
Local transaction dropped


littleskunk:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Parity version?**: 1.11.4
> - **Which operating system?**: Windows
> - **How installed?**: via installer
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: no

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

**expected behavior**

I have a few pending transactions. They have a low fee and will need days to get confirmed.

A few days later I add a new transaction. The new transaction should be added to my local list of pending transactions. Drop any other transaction but not one of my pending transactions.

**actual**

The new transaction gets droped. Logfile `Transaction pushed out because of limit (hash x, replacement: y)`. UI message `Dropped because of queue limit`. Never ever drop any local transaction!

**steps to reproduce**

I did not try to reproduce it. I don't want to lose more pending transactions. I guess the following steps are needed:
1.) Add a transaction with the lowest possible fee.
2.) Restart node? The restart will load the transaction from disk. Maybe any important information gets lost?
3.) Wait until the mempool is full / configurate a tiny limit.
4.) Add a few more transaction and wait for the first transaction to get dropped.

**config**
```
[parity]
mode = "active"
auto_update = "none"
#no_persistent_txqueue = true
#public_node = true

[account]
disable_hardware = true

[rpc]
#interface = "local"
#cors = ["null"]
#server_threads = 4
#processing_threads = 16

[ipc]
path = "\\\\.\\pipe\\jsonrpc.ipc"

[ui]
force = true

[network]
no_serve_light = true
warp = true

[footprint]
pruning_history = 8
pruning_memory = 1
cache_size = 32768
db_compaction = "ssd"
scale_verifiers = true
num_verifiers = 16

[misc]
#logging = "rpc=trace"
log_file = "C:\\DEV\\Parity.log"
```

Tbaut:
Can you confirm that you are sending transactions with Parity UI?
Can you precise how you launch parity (which flags/toml file)?

littleskunk:
Config added above.

I am sending transactions with web3. `web3.eth.accounts.privateKeyToAccount(...).signTransaction(...)` and `web3.eth.sendSignedTransaction(...)`

On the logfile they are listed as local transactions. `Local tx ... below minimal gas price accepted`
[...]

PR #9002, opened 2018-06-28
Never drop local transactions from different senders.

tomusdrw:
#8981 

In case we submit multiple transactions from the same sender and `max_per_sender` is reached the transaction we attempt to insert should fail with an error, although for multiple senders, one local sender will never push out local transactions from another sender - those will be accepted over the limit.

## Warning!
Careful with #8820, exposing `sendRawTransaction` without running with the flag introduced in the other issue can lead to OOM.


sorpaas:
My main concern is just #8820. Should we enable flag `tx_queue_no_unfamiliar_locals` by default to protect against this?

tomusdrw:
@sorpaas we can, sure.
Although the JSON-RPC interface is not exposed by default anyway, so if someone does expose it they need to be aware that the node is vulnerable to all sorts of attacks, this one being one of them.\

That said I'm ok with enabling the flag by default, or maybe rather enabling the flag if the `interface` is not `localhost`.

PR #9005, opened 2018-06-28
Recently rejected cache for transaction queue

tomusdrw:
Tests shows that we receive the same transactions multiple times (especially when running with large number of peers).
This causes excessive verification overhead, when the same transaction is verified over and over again and then rejected for exactly the same reason.

This PR introduces a cache for rejections, which we clear after every block.

Also because of large amount of transactions the sync -> IoWorker queue was easily saturated and a lot of transactions were just dropped, this PR also addresses that by making the queue size depend on the transaction pool size.

In subsequent PR I plan to optimize cache eviction, cause initial tests indicate that the most common rejection reasons don't need to be cleared at every block:

```
$ cat recently.log | awk "{print \$12}" | sort | uniq -c | sort -n
      3 LimitReached
   6403 GasLimitExceeded
  39812 Old
 157679 InsufficientBalance
 291630 InsufficientGasPrice
 634291 TooCheapToReplace
 717894 AlreadyImported
```

Related #8696


miner/src/pool/queue.rs
```diff
@@ -138,6 +138,50 @@ impl CachedPending {
...
+			let to_remove: Vec<_> = inner.keys().take(self.limit / 2).cloned().collect();
+			for key in to_remove {
+				inner.remove(&key);
+			}
+		}
+	}
+}
+
+/// Minimal size of rejection cache, by default it's equal to queue size.
+const MIN_REJECTED_CACHE_SIZE: usize = 2048;
```
dvdplm: Maybe use this to instantiate the hashmap `with_capacity()`? Not sure how hot this cache is expected to be.

miner/src/pool/queue.rs
```diff
@@ -138,6 +138,50 @@ impl CachedPending {
 	}
 }
 
+#[derive(Debug)]
+struct RecentlyRejected {
+	inner: RwLock<HashMap<H256, transaction::Error>>,
```
dvdplm: Curious about why you're storing the error here as opposed to using a `HashSet`. Do you plan to use the error count stats to refine the logic further down?

miner/src/pool/queue.rs
```diff
@@ -138,6 +138,50 @@ impl CachedPending {
...
+	fn insert(&self, hash: H256, err: &transaction::Error) {
+		if self.inner.read().contains_key(&hash) {
+			return;
+		}
+
+		let mut inner = self.inner.write();
+		inner.insert(hash, err.clone());
+
+		// clean up
+		if inner.len() > self.limit {
```
dvdplm: If we never go above the limit the cache entries are never expired, is that ok?

5chdn:
@dvdplm or @andresilva for a final review?

miner/src/pool/queue.rs
dvdplm: Thank you for the explanation! :)

Issue #8961, opened 2018-06-23
5MB limit for RPC requests


littleskunk:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Parity version?**: 1.11.4 (issue first seen 1.9.5, working fine in 1.9.3)
> - **Which operating system?**: Windows
> - **How installed?**: via installer
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

**expected behavior**

I am running a script. I use web3 batch to send one single rpc request containing a few thousend balanceOf requests. In parity 1.9.3 it was working fine.

**actual**

In parity 1.9.5 a hard 5MB limit was implemented with this commit: https://github.com/paritytech/jsonrpc/commit/06bec6ab682ec4ce16d7e7daf2612870c4272028

1.) Please add an option. I would like to increase that limit.
2.) For some reasons I did not receive the error message `PayloadTooLarge`. The error message gets lost somewhere. Please return the error message.

**steps to reproduce**

parity config `logging = "rpc=trace"`

Run this script:
```
'use strict';

const Web3 = require('web3');
const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545', 0));
const batch = new web3.BatchRequest();

const erc20ABI = [
	{"constant":true,
		"inputs":[{"name":"_owner","type":"address"}],"name":"balanceOf",
		"outputs":[{"name":"balance","type":"uint256"}],"payable":false,"type":"function"},
	{"constant":false,
		"inputs":[{"name":"_to","type":"address"},{"name":"_value","type":"uint256"}],"name":"transfer",
		"outputs":[{"name":"success","type":"bool"}],"payable":false,"type":"function"}];

const Contract = '0xb64ef51c888972c908cfacf59b47c1afbc0ab8ac';
const ERC20 = new web3.eth.Contract(erc20ABI, Contract);

var count = 0;

function handleResult(error, ETHBalance) {
	if (error) {
		console.log('Error:', error.message);
		process.exit();
	}
	if (!ETHBalance) {
		console.log('Result:', ETHBalance);
		process.exit();
	}
	console.log(count++);
}

for(var i=0; i<25630; i++) {
	batch.add(ERC20.methods.balanceOf('0x0000000000000000000000000000000000000000').call.request(handleResult));
[...]

PR #9010, opened 2018-06-29
Add option for user to set max size limit for RPC requests

mttmartin:
Add option for user to set max size limit for RPC requests. This was requested in #8961.

### DIFF
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -482,6 +482,10 @@ usage! {
 			"--jsonrpc-server-threads=[NUM]",
 			"Enables multiple threads handling incoming connections for HTTP JSON-RPC server.",
 
+			ARG arg_jsonrpc_max_payload: (Option<usize>) = None, or |c: &Config| c.rpc.as_ref()?.max_payload,
+			"--jsonrpc-max-size=[NUM]",
+			"Specify maximum size for RPC requests in megabytes.",
+
 		["API and Console Options – WebSockets"]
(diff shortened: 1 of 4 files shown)


mttmartin:
[clabot:check]

parity/cli/mod.rs
```diff
@@ -482,6 +482,10 @@ usage! {
 			"--jsonrpc-server-threads=[NUM]",
 			"Enables multiple threads handling incoming connections for HTTP JSON-RPC server.",
 
+			ARG arg_jsonrpc_max_payload: (Option<usize>) = None, or |c: &Config| c.rpc.as_ref()?.max_payload,
+			"--jsonrpc-max-size=[NUM]",
```
tomusdrw: should be `MB` instead of `NUM`

5chdn:
Could not compile `parity-rpc`.

https://gitlab.parity.io/parity/parity/builds/90842

parity/cli/mod.rs
mttmartin: Thanks for looking over my PR. I've added both of your suggestions.

rpc/src/tests/rpc.rs
```diff
@@ -40,6 +40,7 @@ fn serve(handler: Option<MetaIoHandler<Metadata>>) -> Server<HttpServer> {
 			}
 		}),
 		1,
+        5,
```
tomusdrw: Incorrect indentation.

mttmartin: Thanks for the review. I've corrected the indentation issue.

PR #9017, opened 2018-06-29
Remove the dapps system

tomaka:
The UI has been removed, therefore we no longer need the dapps service.

Please let me know if I did something stupid by removing too much.


5chdn:
Test `v1::tests::mocked::parity_set::rpc_parity_set_dapps_list` failed.

axelchalon:
Nice! Let's wait for [Parity UI network dapps PR](https://github.com/parity-js/shell/pull/144) to be merged before we merge this

rpc/src/lib.rs
```diff
@@ -131,34 +131,55 @@ use http::tokio_core;
...
-		builder = builder.request_middleware(dapps)
-	}
+		.allowed_hosts(allowed_hosts.into())
+		.start_http(addr)?)
+}
 
-	Ok(builder.start_http(addr)?)
+/// Same as `start_http`, but takes an additional `middleware` parameter that is introduced as a
+/// hyper middleware.
+pub fn start_http_with_middleware<M, S, H, T, R>(
```
tomusdrw: Not sure about the duplication here, maybe pass a function that can add something to the builder: `F: FnOnce(&mut Builder)` function instead of `Option<R>`? (Cause I beleive the reason for duplication is that in case we are passing `None` the type cannot be inferred)

5chdn:
Ok this is ready

parity/configuration.rs
```diff
@@ -590,18 +586,6 @@ impl Configuration {
 		self.args.arg_ntp_servers.split(",").map(str::to_owned).collect()
 	}
 
-	fn dapps_config(&self) -> DappsConfiguration {
-		DappsConfiguration {
-			enabled: self.dapps_enabled(),
-			dapps_path: PathBuf::from(self.directories().dapps),
-			extra_dapps: if self.args.cmd_dapp {
```
sorpaas: Do we need to remove the corresponding `CMD` and `ARG` in `parity/cli/mod.rs`? I think it's basically those:

* https://github.com/tomaka/parity/blob/8a612b66b5918e704dbb0e5484d58638867f9ae4/parity/cli/mod.rs#L29
* https://github.com/tomaka/parity/blob/8a612b66b5918e704dbb0e5484d58638867f9ae4/parity/cli/mod.rs#L542

And we probably also need to add them to `parity/deprecated.rs`.

sorpaas: There's also the config serialization struct to be removed: https://github.com/tomaka/parity/blob/8a612b66b5918e704dbb0e5484d58638867f9ae4/parity/cli/mod.rs#L1223

5chdn:
I'm sorry for all the conflicts. This has top priority now and I will make sure nobody else merges anything else :D

tomaka:
> I'm sorry for all the conflicts. This has top priority now and I will make sure nobody else merges anything else :D

Thank you.

parity/deprecated.rs
```diff
@@ -209,6 +209,22 @@ pub fn find_deprecated(args: &Args) -> Vec<Deprecated> {
...
+
+	if args.arg_dapp_path.is_some() {
+		result.push(Deprecated::Removed("--dapp-path"));
+	}
+
+	if args.flag_no_dapps {
+		result.push(Deprecated::Removed("--no-dapps"));
+	}
+
+	if args.arg_dapps_path != "" {
```
sorpaas: It looks like the default is `"$BASE/dapps"` so this deprecation might be always printing?

My suggestion would be to change `arg_dapps_path` in `parity/cli/mod.rs` to be `Option<String>` with default to `None`. In that case, we can detect whether there's any attempt to set that arg and reliably print out deprecation warnings.

tomaka: Ah yeah, true. I put it as an empty string because the `deprecated` test builds the config with `Default::default()` (which sets the value to empty).
Fixed with an `Option`.

PR #9018, opened 2018-06-30
Make work-notify an optional feature

udoprog:
This is a step towards making depending on parity components a bit lighter.
I'm a bit unsure on if `Stratum` is still needed at all when the work-notify feature is disabled. I think in _principle_ yes, but `work-notify` seems to be all it is currently used for.

My ultimate goal is to make depending on hyper and tokio optional when we don't need to access external services, like when testing smart contracts. Removing those two dependencies transitively would remove a large chunks of crates being compiled when depending on ethcore.


udoprog:
[clabot:check]

udoprog:
Closing in favor of #9020 which contains all the features I want.

PR #9020, opened 2018-06-30
Introduce more optional features in ethcore

udoprog:
This introduces optional features to `ethcore`. If they are not enabled, they leave out the entire hyper/tokio ecosystem which is a lot of dependencies which are not needed when doing things like smart contract testing (e.g. [parables](https://github.com/primablock/parables)).

This is `cargo tree --verbose` in `./ethcore` before and after the change.

Before:
```
cargo tree --verbose | wc -l
691
```

After:
```
cargo tree --verbose | wc -l
493
```

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -213,8 +213,8 @@ pub struct Miner {
 }
 
 impl Miner {
-	#[cfg(feature = "work-notify")]
 	/// Push listener that will handle new jobs
+	#[cfg(feature = "work-notify")]
 	pub fn add_work_listener(&self, notifier: Box<NotifyWork>) {
 		self.listeners.write().push(notifier);
 		self.sealing.lock().enabled = true;
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -24,6 +24,7 @@ use engines::{EthEngine, Seal};
 use error::{Error, ErrorKind, ExecutionError};
 use ethcore_miner::gas_pricer::GasPricer;
 use ethcore_miner::pool::{self, TransactionQueue, VerifiedTransaction, QueueStatus, PrioritizationStrategy};
+#[cfg(feature = "work-notify")]
 use ethcore_miner::work_notify::NotifyWork;
 use ethereum_types::{H256, U256, Address};
 use parking_lot::{Mutex, RwLock};
@@ -200,6 +201,7 @@ pub struct Miner {
 	// NOTE [ToDr]  When locking always lock in this order!
 	sealing: Mutex<SealingWork>,
 	params: RwLock<AuthoringParams>,
+	#[cfg(feature = "work-notify")]
 	listeners: RwLock<Vec<Box<NotifyWork>>>,
 	nonce_cache: RwLock<HashMap<Address, U256>>,
 	gas_pricer: Mutex<GasPricer>,
@@ -211,6 +213,7 @@ pub struct Miner {
 }
 
 impl Miner {
+	#[cfg(feature = "work-notify")]
 	/// Push listener that will handle new jobs
 	pub fn add_work_listener(&self, notifier: Box<NotifyWork>) {
 		self.listeners.write().push(notifier);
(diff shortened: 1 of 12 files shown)


udoprog:
@5chdn Ah, tests are run with `--no-default-features`. So all the things provided by dependencies through features have to be conditionally supported as well?

udoprog:
@tomaka well... yeah :). That made it click. They should be part of dependency, not the local `[features]` for the parity app.

tomaka:
For what it's worth, this feature is desirable for #7915 because `hyper` and `tokio` don't work when inside a browser.

udoprog:
@dvdplm Hey!

> I am not sure I agree that this is the right approach

Increasing modularization seems to be the right long term approach. Features are more like a way to get it done right now, and they can be removed as things are broken up into more composable traits, types, and crates. But apart from not being as visible as a dependency, a missing feature causes a compile-time error similar to a missing dependency.

> "is this code compiled for the binary used?"`

Note that the features introduced here are not gated in the binary (thanks due to the comment by @tomaka). You would get a compile time error if you try to use a type or function that is feature-gated in `ethcore`. It's similar to if you were trying to use `serde` for a dependency without enabling the feature.

miner/src/gas_pricer.rs
```diff
@@ -16,70 +16,25 @@
...
-}
+#[cfg(feature = "price-info")]
+use gas_price_calibrator::GasPriceCalibrator;
 
 /// Struct to look after updating the acceptable gas price of a miner.
 #[derive(Debug, PartialEq)]
 pub enum GasPricer {
 	/// A fixed gas price in terms of Wei - always the argument given.
 	Fixed(U256),
+	#[cfg(feature = "price-info")]
```
tomaka: Nit: should be under the `///`.

miner/src/gas_pricer.rs
```diff
@@ -16,70 +16,25 @@
...
 	/// Gas price is calibrated according to a fixed amount of USD.
 	Calibrated(GasPriceCalibrator),
 }
 
 impl GasPricer {
+	#[cfg(feature = "price-info")]
```
tomaka: Nit: should be under the `///`.

ethcore/src/miner/stratum.rs
```diff
@@ -244,7 +247,19 @@ impl Stratum {
...
+		}
+
+		// NB: hack to avoid unused warning. `stratum` is kept for now since it's unclear if it has
+		// side-effects.
+		#[cfg(not(feature = "work-notify"))]
+		{
+			let _stratum = stratum;
+		}
+
 		Ok(())
```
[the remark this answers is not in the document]
udoprog: In that case, maybe we should just gate the `pub fn register` function for a compile-time error?

ethcore/src/miner/miner.rs
```diff
@@ -211,6 +213,7 @@ pub struct Miner {
 }
 
 impl Miner {
+	#[cfg(feature = "work-notify")]
```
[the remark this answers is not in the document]
udoprog: Fixed!

miner/src/gas_pricer.rs
udoprog: Fixed!

miner/src/gas_pricer.rs
udoprog: Fixed! ... Thanks!

dvdplm:
@udoprog Ok, I rest my case! :)

ethcore/src/miner/stratum.rs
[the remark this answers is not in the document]
udoprog: Done, thanks!

parity/run.rs
```diff
@@ -529,11 +530,16 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 	miner.set_author(cmd.miner_extras.author, None).expect("Fails only if password is Some; password is None; qed");
 	miner.set_gas_range_target(cmd.miner_extras.gas_range_target);
 	miner.set_extra_data(cmd.miner_extras.extra_data);
-	if !cmd.miner_extras.work_notify.is_empty() {
-		miner.add_work_listener(Box::new(
-			WorkPoster::new(&cmd.miner_extras.work_notify, fetch.clone(), event_loop.remote())
-		));
+
+	#[cfg(feature = "work-notify")]
```
tomusdrw: That's never enabled. main `Cargo.toml` doesn't have that feature.

PR #9024, opened 2018-07-02
A last bunch of transaction queue performance optimizations

tomusdrw:
- [x] Nonce cache size dependent on the pool size. Currently it was hardcoded to 4k entires, now it depends on the size of the pool (but never less than 4k). I've noticed that the cache was very often saturated when we run with large pool size.

- [x] Optimize order of `cull` and `update_sealing` in `miner::chain_new_blocks`. Since `cull` can take significant amount of time (seconds, for large pools) it's better to first attempt to construct pending block. We only need couple (hundreds) of best transactions and `Readiness` will make sure that they are not stale anyway. After pending block is created we attempt to clean up the pool. This should allow miners to start mining on the new block much earlier and reduce orphan rate because of that.

- [x] Split `cull` into chunks. To prevent acquiring `write()` lock of the queue for long time we are cleaning up the queue in chunks (1024 senders each chunk). This allows other threads to jump in between.

### DIFF
--- a/ethcore/src/miner/pool_client.rs
+++ b/ethcore/src/miner/pool_client.rs
@@ -36,10 +36,7 @@ use header::Header;
 use miner;
 use miner::service_transaction_checker::ServiceTransactionChecker;
 
-type NoncesCache = RwLock<HashMap<Address, U256>>;
-
-const MAX_NONCE_CACHE_SIZE: usize = 4096;
-const EXPECTED_NONCE_CACHE_SIZE: usize = 2048;
+type NoncesCache = (RwLock<HashMap<Address, U256>>, usize);
 
 /// Blockchain accesss for transaction pool.
 pub struct PoolClient<'a, C: 'a> {
@@ -176,7 +173,8 @@ impl<'a, C: 'a> Clone for CachedNonceClient<'a, C> {
 impl<'a, C: 'a> fmt::Debug for CachedNonceClient<'a, C> {
 	fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
 		fmt.debug_struct("CachedNonceClient")
-			.field("cache", &self.cache.read().len())
+			.field("cache", &self.cache.0.read().len())
+			.field("limit", &self.cache.1)
 			.finish()
 	}
 }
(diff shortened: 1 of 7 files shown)


miner/src/pool/queue.rs
```diff
@@ -340,9 +349,17 @@ impl TransactionQueue {
 			current_id.checked_sub(gap)
 		};
 
-		let state_readiness = ready::State::new(client, stale_id, nonce_cap);
-
-		let removed = self.pool.write().cull(None, state_readiness);
+		let mut removed = 0;
+		let senders: Vec<Address> = {
```
andresilva: Shouldn't this be a set? Otherwise we may cull the same sender more than once.

ethcore/src/miner/pool_client.rs
```diff
@@ -36,10 +36,7 @@ use header::Header;
 use miner;
 use miner::service_transaction_checker::ServiceTransactionChecker;
 
-type NoncesCache = RwLock<HashMap<Address, U256>>;
-
-const MAX_NONCE_CACHE_SIZE: usize = 4096;
-const EXPECTED_NONCE_CACHE_SIZE: usize = 2048;
+type NoncesCache = (RwLock<HashMap<Address, U256>>, usize);
```
andresilva: We should probably create a struct for this at some point.

ascjones:
Needs a merge/rebase from master

miner/src/pool/queue.rs
tomusdrw: We need the senders to be ordered by priority (so that we cull best senders first - more probability that it will have any effect on the queue), `best_transactions` in the pool (That we take the senders from) contains exactly one transaction per sender.

ethcore/src/miner/pool_client.rs
tomusdrw: Done

tomusdrw:
@ascjones Rebased.

miner/src/pool/queue.rs
ascjones: Is it worth adding a comment here to that effect?

andresilva:
Needs another rebase now that #9026 was merged.

miner/src/pool/queue.rs
andresilva: I didn't realize `best_transactions` only had the best tx from each sender 👍.

tomusdrw:
@ascjones yes, I was running the node and observing 3 things:
1. The average CPU load on the server
2. The peer count and sudden drops
3. The `perf=trace` logs and long pauses.

All three things are obviously related to each other, long pauses are usually caused by high CPU load in some place, which in turn leads to peer timeouts and drops. But overall chunkig the senders causes each chunk to be processed in roughly 200ms, and in between we can easily import incoming transactions or propagate the pending ones.

ascjones:
Thanks @tomusdrw. I see that node has now taken the top spot! 

Build needs fixing though.

niklasad1:
@tomusdrw  https://github.com/paritytech/parity/blob/832d171b2e8dc4257a9e5fe86f15c37e30f25734/ethcore/src/miner/pool_client.rs#L55-#L69

Duplicate method `get`

ascjones:
@niklasad1 caused by me hijacking @tomusdrw 's branch to fix the build (see above).

tomusdrw:
No worries, it's my fault. I've didn't try to compile after merge.

PR #9025, opened 2018-07-02
Only return error log for rustls

sorpaas:
`rustls` appears to be sending too many logs. Especially sometimes you will see:

```
Sending warning alert CloseNotify
```

However, according to RFC 5246 section 7.2 (https://tools.ietf.org/html/rfc5246#section-7.2), this close notify message is just indication that the connection is being closed. However, this sometimes frightens user.

This PR modifies `rustls` log level from warning to error.


5chdn:
But we should additionally consider fixing this upstream, it should not be a warning.

sorpaas:
@ordian I opened upstream PR: https://github.com/ctz/rustls/pull/182

@5chdn No we don't use it for anything critical -- it's used in fetcher, which handles price info fetch, and miner fixed endpoint notification (not the normal getwork and stratum protocol).

PR #9026, opened 2018-07-02
Optimize pending transactions filter

tomusdrw:
Re-opening #9014 

- [x] Also fix ethstats reporting of pending transactions (in case pending block is missing).

Context:
Currently pending transactions filter is using the very inefficient ready_transactions method from miner. For large pools it takes significant amount of time to construct the pending set, because:

We return transactions ordered by priority (gas price)
We check for readiness of transactions - i.e. we need to check the state nonce for every sender, and it's really expensive.
Because of that it can take seconds to construct an ordered pending set of 128k transactions (~110k senders).
The pending transactions filter doesn't require such pending set though, we are ok with unordered hashes and we don't need to be super sure about the nonce, so this PR:

Uses unordered transactions from the queue
Doesn't use the default Ready (which goes to the state), but rather assumes the transactions in the pool are ready - this may give us a couple of false-positives, but it's not critical for pending transactions hashes.
Thanks to this it works really fast (just constructs the BTreeSet and then does the difference).


ethcore/src/miner/miner.rs
```diff
@@ -1065,13 +1096,18 @@ impl miner::MinerService for Miner {
...
+		if has_new_best_block {
+			// Clear nonce cache
+			self.nonce_cache.write().clear();
+		}
 
 		// First update gas limit in transaction queue and minimal gas price.
 		let gas_limit = *chain.best_block_header().gas_limit();
 		self.update_transaction_queue_limits(gas_limit);
 
+
```
andresilva: Stray newline?

miner/src/pool/ready.rs
```diff
@@ -129,6 +129,38 @@ impl txpool::Ready<VerifiedTransaction> for Condition {
 	}
 }
 
+pub struct OptionalState<C> {
```
andresilva: Can we add some docs here?

"Checks readiness of transactions by comparing the nonce to state nonce. If nonce isn't found in provided state nonce store, defaults to the tx nonce and updates the nonce store. Useful for using with a state nonce cache when false positives are allowed." ?

dvdplm:
> this may give us a couple of false-positives, but it's not critical for pending transactions hashes.

Can you elaborate on this point: why is it not crucial?

rpc/src/v1/helpers/poll_filter.rs
```diff
@@ -16,19 +16,41 @@
...
+		f(&mut self.0.lock())
+	}
+}
+
 /// Filter state.
 #[derive(Clone)]
 pub enum PollFilter {
 	/// Number of last block which client was notified about.
 	Block(BlockNumber),
 	/// Hashes of all transactions which client was notified about.
```
dvdplm: Maybe `/// Hashes of all pending transaction the client knows about`?

ethcore/src/miner/miner.rs
```diff
@@ -851,6 +851,37 @@ impl miner::MinerService for Miner {
 		self.transaction_queue.all_transactions()
 	}
 
+	fn pending_transactions_hashes<C>(&self, chain: &C) -> BTreeSet<H256> where
```
dvdplm: I think it should be `pending_transaction_hashes` (no double plural `s`), see e.g. [this](https://english.stackexchange.com/questions/203985/hidden-layer-sizes-vs-hidden-layers-sizes).

(I am not a native speaker so take it with a grain of salt)

sorpaas: I checked and we're having both usages. This probably should be unified to a single one in the future:

* In `eth_filter` we have `pending_transactions_hashes`.
* In `ethcore`'s Block's views and `encoded`, we have `transaction_hashes`.

5chdn:
Please rebase.

tomusdrw:
> > this may give us a couple of false-positives, but it's not critical for pending transactions hashes.

> Can you elaborate on this point: why is it not crucial?

So the queue in general should contain only transactions that are `Ready`, however, if we miss a notification from `chain_new_blocks` (which may happen if the client goes from normal mode to sync mode (i.e. prints `Syncing..` instead of `Imported`) it's not necessarily the case, the queue will contain a bunch of transactions that are already included in the chain or old.
Another case is when transactions in the pool is `Future`, which means that there is a nonce gap between that particular transaction and the state nonce - that case is really rare though, and even if it happens usually it means that we will just receive the missing transaction soon(ish) and that transaction will become valid (Ready).

So because of the nature of the filters itself, when the caller is receiving the data it might be already outdated (since a new block could be imported, or the transaction could be replaced), that's why it doesn't make much sense to cover the case where the caller sometimes gets false-positives.


Regarding the set performance, TBH I wasn't benchmarking that I chose `BTreeSet` for consistency (we return `BTreeMap` in couple of other places) and testability (it has predictable order).
I'm not entirely sure how efficient `difference` is in both implementations, might be worth checking. I suspect that both are `O(n)`, but my gut feeling is that `BTreeSet` should have smaller constant in front.

5chdn:
Does not compile. 

Is this in progress or on ice? It already has 3 reviews.

tomusdrw:
@5chdn Fixed, sorry. It's neither in progress nor on ice, it's being reviewed. I thought that after issues are addressed we are going back to `pleasereview` label, so that people can review the changes I made again.

5chdn:
Ok, leaving this for @andresilva @sorpaas or @dvdplm to sign this off and merge.

dvdplm:
My review stays: LGTM.

@tomusdrw thank you for the extra explanations!

Issue #8604, opened 2018-05-11
Write examples for parity-clib in other languages


tomaka:
Right now there is a single example in C/C++, but we may want to show more example usages for `parity-clib`.

An open question is: do we want to show how to compile Parity in each example's package manager? For example, should `CMakeLists.txt` include the `cargo build` step? Technically it is the same, but when it comes to the educative value of the example both ways have advantages and disadvantages.

Individual languages issues:
- Go: #9338 
- Python: #9339 

If you want to claim a bounty, see the individual issues or ask here to open an issue for a specific language.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.25 ETH (141.28 USD @ $565.13/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/paritytech/parity/8604/520).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $8,949.35 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.

 1. [@oogetyboogety](https://gitcoin.co/profile/oogetyboogety) 

 has committed to working on this project to be completed 4 weeks, 1 day from now.


@vs77bb, __please see the below comments / questions regarding approach for this ticket from the bounty hunter(s):__ 
- [@oogetyboogety](https://gitcoin.co/profile/oogetyboogety): Starting example based on parity-clib-example. Calling methods in Cython (https://cython.readthedocs.io/en/latest/src/tutorial/external.html) for parity-clib/examples/python

gitcoinbot:
@oogetyboogety Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
* [ ] escalation to mods (6 days)

[...]

PR #9032, opened 2018-07-03
Move the C/C++ example to another directory

tomaka:
cc #8604 

Moves the C/C++ example to a subdirectory so that other examples can be added in the `parity-clib-examples` directory.


tomaka:
@niklasad1 I didn't do that because parity-clib is its own Rust crate with important code. I would have done it if it was just some sort of "SDK".

PR #9036, opened 2018-07-03
Clean up deprecated options and add CHECK macro

sorpaas:
We have added some deprecated options but forgot to include them in the `deprecated` mod. This makes users not be given any warnings if options are deprecated. This PR adds them back. In addition, options are organized by the version they were deprecated, this makes it possible to make a set of deprecation hard errors in the future.

A CHECK macro is also added in `usage!`. We had a hard coded check conditions in the macro before. Using the new `CHECK` syntax we can move it out.

Affects #9017. Deprecated `dapps` options will need to be added to the `deprecated` mod.


5chdn:
Thanks for that. I'll wait with #8963 for this.

parity/cli/mod.rs
```diff
@@ -879,57 +890,66 @@ usage! {
...
+			"Does nothing; HTTP JSON-RPC is on by default now.",
+
 			FLAG flag_jsonrpc_off: (bool) = false, or |_| None,
 			"--jsonrpc-off",
 			"Equivalent to --no-jsonrpc.",
 
 			FLAG flag_webapp: (bool) = false, or |_| None,
 			"-w, --webapp",
-			"Does nothing; dapps server is on by default now.",
+			"Does nothing; dapps server is removed.",
```
andresilva: `Does nothing; dapps server has been removed.`

parity/cli/mod.rs
```diff
@@ -1027,23 +1021,55 @@ usage! {
...
 			ARG arg_cache: (Option<u32>) = None, or |_| None,
 			"--cache=[MB]",
 			"Equivalent to --cache-size MB.",
 
-			ARG arg_tx_queue_ban_count: (u16) = 1u16, or |c: &Config| c.mining.as_ref()?.tx_queue_ban_count.clone(),
+			// ARG Removed in 1.7.
+
+			ARG arg_dapps_port: (Option<u16>) = None, or |c: &Config| c.dapps.as_ref()?.port.clone(),
+			"--dapps-port=[PORT]",
+			"Dapps server is merged with HTTP JSON-RPC server. Use --jsonrpc-port.",
```
andresilva: Should these arguments say instead that the dapps server has been removed?

sorpaas: Yeah right. I checked, those arguments indeed does nothing.

andresilva:
`deprecated::tests::test_find_deprecated` needs to be updated with dapps settings that were removed (instead of replaced).

PR #9046, opened 2018-07-04
Publish keccak-hasher to crates.io

dvdplm:
Related to https://github.com/paritytech/parity/issues/9011

The tests for `memorydb` has `keccak_hasher` as a dev-dependency. We don't want `keccak-hasher` in `parity-common`. To get around that I published `keccak-hasher` to crates.io and this PR simply tweaks the Cargo.toml and versions to make that possible.


dvdplm:
This is not needed anymore.

PR #9050, opened 2018-07-05
bump fs-swap to 0.2.3 so it is compatible with osx 10.11 again

debris:


Cargo.lock
```diff
@@ -558,7 +558,6 @@ dependencies = [
  "ethstore 0.2.0",
  "evm 0.1.0",
  "fake-hardware-wallet 0.0.1",
- "fetch 0.1.0",
```
dvdplm: I think this one pops up when `git submodule init`/`git submodule update` runs. Wasm-tests something something?

andresilva: It was removed in #9020.

PR #9051, opened 2018-07-05
Allow single opcode stepping for EVM

### LINKED ISSUE #9035: Add 'Step' to VM?
I am currently using the Parity EVM for a project of mine, and was wondering what the community thought of adding some extra functionality to the EVM.

Specifically, I think it would be useful to add a 'step' function to the vm::Vm trait (and consequently to the Interpreter struct that implements the trait) allowing a program to only execute one instruction at a time. Adding a public-facing function like this could be useful for a variety of reasons, but mostly for real-time debugging purposes and code-analysis. 

I have noticed that a few other VM's do have this 'hook' (SputnikVm, ethereumjs-vm)

(I could even implement this, it would be fairly simple)

I'm fairly new to this project, however, so I don't want to barge in and start suggesting things that would cause issues I am not aware of. I also don't want to add needless bloat to the EVM implementation, this is merely a suggestion.

sorpaas:
rel #6744, fixes #9035

This is a prerequisite for implementing EVM resume and callstack refactoring. In particular, a new function `Interpreter::step(&mut Ext)` is added. This function step one additional opcode, and can be used plainly. Other notable changes:

* ActionParams is taken when initializing a VM, but not when calling `exec`. This is because we need to allocate stack, gasometer, etc before `exec` is called.
* The ability to "reuse" a VM after an execution is removed. We never did that in our code. Allocation (either on stack or heap) for various VM structs need to happen anyway, so "reusing" doesn't provide performance benefits (except maybe reusing the previous VM's memory heap region).


sorpaas:
Here's the benchmark results:

jsontests benchmarks:

```
{'ackermann31-INT': 0.9897959183673469,
 'ackermann32-INT': 0.9167478091528725,
 'ackermann33-INT': 0.9596136962247586,
 'fibonacci10-INT': 1.0860832137733143,
 'fibonacci16-INT': 0.9300284017557449,
 'loop-add-10M-INT': 0.9226587206771758,
 'loop-divadd-10M-INT': 0.9229079794970425,
 'loop-divadd-unr100-10M-INT': 0.8998631955175739,
 'loop-exp-16b-100k-INT': 1.0188090390141178,
 'loop-exp-1b-1M-INT': 0.9923591588497622,
 'loop-exp-2b-100k-INT': 0.9466506030511994,
 'loop-exp-32b-100k-INT': 1.0991487959483188,
 'loop-exp-4b-100k-INT': 1.12556930650952,
 'loop-exp-8b-100k-INT': 1.1088218849334053,
 'loop-exp-nop-1M-INT': 0.8219694300785856,
 'loop-mul-INT': 0.8684106995830124,
 'loop-mulmod-2M-INT': 0.9754786210274101,
 'manyFunctions100-INT': 0.8409570724841661}
```

Uint benchmark (master):

```
Output: 0x606060405200
Gas used: 29fb170
Time: 0.407322134s
^^^^ usize
Output: 0x606060405200
Gas used: 29fb170
Time: 0.418064895s
^^^^ U256
Output: 0x606060405200
Gas used: 8865053
Time: 1.412694181s
^^^^ usize
Output: 0x606060405200
Gas used: 8865053
Time: 1.473797155s
^^^^ U256
```

Uint benchmark (PR):

```
Output: 0x606060405200
Gas used: 29fb170
Time: 0.472320852s
^^^^ usize
Output: 0x606060405200
Gas used: 29fb170
Time: 0.489357480s
^^^^ U256
Output: 0x606060405200
Gas used: 8865053
Time: 1.608301138s
^^^^ usize
Output: 0x606060405200
Gas used: 8865053
Time: 1.618444041s
^^^^ U256
```

[This gist](https://gist.github.com/sorpaas/2040b1fdc003a1a44ef6a1ec04ea6ed5) contains the python code used to format the result.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
...
+			_type: PhantomData,
+		})
+	}
 
-		let code = &params.code.as_ref().expect("exec always called with code; qed");
-		let mut valid_jump_destinations = None;
+	/// Execute a single step on the VM.
+	#[inline(always)]
+	pub fn step(&mut self, ext: &mut vm::Ext) -> InterpreterResult {
+		macro_rules! try_or_done {
```
tomusdrw: Is it really worth to have a macro for that?
Maybe sth like this:
```
#[inline(always)]
fn try_or_done<F>(x: Result<F, vm::Error>) -> Result<F, InterpreterResult> {
 x.map_err(|e| InterpreterResult::Done(Err(e))
}

// and used like this:
try_or_done(self.verify_instruction(ext, instruction, info))?;
```

Or maybe even add `From<vm::Error> for InterpreterResult` and convert errors automatically via `?`?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
...
-			}
-			let instruction = instruction.expect("None case is checked above; qed");
+				// TODO: make compile-time removable if too much of a performance hit.
+				self.do_trace = self.do_trace && ext.trace_next_instruction(
+					self.reader.position - 1, opcode, self.gasometer.current_gas.as_u256(),
+				);
 
-			let info = instruction.info();
-			self.verify_instruction(ext, instruction, info, &stack)?;
+				if instruction.is_none() {
```
tomusdrw:
```
let instruction = match instruction {
  Some(i) => i,
  None => return vm::Error::BadInstruction { .. }.into();
}
```

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
...
+		if self.done {
+			return InterpreterResult::Stopped;
+		}
 
-		while reader.position < code.len() {
-			let opcode = code[reader.position];
-			let instruction = Instruction::from_u8(opcode);
-			reader.position += 1;
+		let result = {
+			let mut inner = || {
```
tomusdrw: Why do we need the `inner` closure?

Oh, I see to be able to use `return` maybe it's worth to move it to a separate function?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
...
+				// This case is needed because code length can be zero.
+				if self.reader.position >= self.reader.len() {
```
tomusdrw: Seems that we are checking that twice now, can't this be moved as some initial step? The `reader.position` is changed only during regular execution and jumps, right?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
...
+		if let &InterpreterResult::Done(_) = &result {
+			self.done = true;
+			self.informant.done();
 		}
+		return result;
 	}
 
-	fn verify_instruction(&self, ext: &vm::Ext, instruction: Instruction, info: &InstructionInfo, stack: &Stack<U256>) -> vm::Result<()> {
+	fn verify_instruction(&self, ext: &vm::Ext, instruction: Instruction, info: &InstructionInfo) -> vm::Result<()> {
 		let schedule = ext.schedule();
```
tomusdrw: Schedule could probably be cached on `self` level, less virtual calls.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -102,127 +103,234 @@ enum InstructionResult<Gas> {
 	StopExecution,
 }
 
+/// ActionParams without code, so that it can be feed into CodeReader.
+#[derive(Debug)]
+struct InterpreterParams {
```
tomusdrw: Why we need this? Couldn't we just use `ActionParams` and add some info that the code is always `None`?

tomusdrw: Actually on a second though, it's better to have it like that - prevents future errors of accessing the code.

ethcore/evm/src/interpreter/mod.rs
sorpaas: The issue is that `?` only works for `Result` like things with `Ok` and `Err`. So we can't use any `?` syntax in `step` right now, unless we change `InterpreterResult` to be a `Result<A, B>`. This indeed can be done via:

* Return `Ok(result)`, which is the same case for `InterpreterResult::Done(_)`.
* Return error case for `InterpreterResult::Stopped` and `InterpreterResult::Continue`.

However, I think it may be a little bit confusing as `Ok`/`Err` do not really apply here.

ethcore/evm/src/interpreter/mod.rs
sorpaas: I moved it up to check just `reader.len() == 0`. That check still needs to happen every time `step` is called, otherwise we'll need an additional "interpreter state" (which might actually cost more memory).

ethcore/evm/src/interpreter/mod.rs
sorpaas: The issue is that for some use cases of `step`, like pausing, we would want to take back the `Ext` reference, possibly do some modifications to it, before passing it back to continue the execution.

1. If we take a reference of `Schedule` and put it on self level, it means that this single `Ext` is not usable again for the whole lifecycle of the Interpreter.
2. Or we need to clone the `Schedule`, which may also be expensive.

I think virtual calls might not be *that* expensive. And with (1), one optimization we can do is to split `Ext` up to a) interpreter-specific `Ext`, and b) transaction-level `Ext`. I think that may be a good trade-off compared with additional virtual calls during each `step`.

ethcore/evm/src/interpreter/mod.rs
tomusdrw: Oh, that's true. Forgot that it only works for `Result` on the return type as well, anyway might be worth extracting the stuff in `inner` to a separate function and maybe use a helper function instead of a macro. What do you think?

ethcore/evm/src/interpreter/mod.rs
tomusdrw: Makes sense. I thought that we already clone schedule on `ext.schedule()` but now saw it's only borrowed.

ethcore/evm/src/interpreter/mod.rs
sorpaas: I think I found a way: we can return `Result<(), InterpreterResult>` as the inner function and always return `Err`. The compiler would happily accept `?` syntax with that. And we would just need to make sure the inner function never returns `Ok(())`.

This detail is wrapped in `step_inner` so it doesn't affect public interface.

sorpaas:
A new change on this PR is that I made `Interpreter::new` to always succeeds. We only had one possible Err case -- `gasometer`, if `usize` cost type is used, can directly return out of gas error. This is changed to delay the error return to `step` -- by making `gasometer` an option, and `step`, if we found `gasometer` to be option, then directly return error.

The rationale for this is that I found out if `Interpreter::new` can return error, it would significantly complicate resumable `Executive` implementation.

PR #9052, opened 2018-07-05
Parity Ethereum 2.0.0

5chdn:
- rename `parity` ~~crate~~ to `parity-ethereum`
- bump version to 2.0.0


debris:
does`2.0.0` mean that we can break backwards compatibility? 🤔

5chdn:
> does `2.0.0` mean that we can break backwards compatibility?

hehe, go ahead.

ethcore/sync/src/chain/propagator.rs
```diff
@@ -48,7 +48,7 @@ fn accepts_service_transaction(client_id: &str) -> bool {
 	// Parity versions starting from this will accept service-transactions
 	const SERVICE_TRANSACTIONS_VERSION: (u32, u32) = (1u32, 6u32);
 	// Parity client string prefix
-	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity/v";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/v";
 
 	if !client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
```
5chdn: Are we seriously checking compatibility via version string? If yes, do you want me to check for both?

```
const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/v";
const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/v";
if !client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) && !client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
    // ...
}
```

ethcore/sync/src/chain/propagator.rs
```diff
@@ -48,16 +48,24 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
 	// Parity client string prefix
-	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity/v";
+	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/v";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/v";
 
-	if !client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		return false;
-	}
+	if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
+	let ver: Vec<u32> = client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
```
dvdplm: This line should be indented (and the 4 following lines as well).

ethcore/sync/src/chain/propagator.rs
```diff
@@ -48,16 +48,24 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
+		.take(2)
+		.filter_map(|s| s.parse().ok())
+		.collect();
+		ver.len() == 2 && (ver[0] > SERVICE_TRANSACTIONS_VERSION.0 || (ver[0] == SERVICE_TRANSACTIONS_VERSION.0 && ver[1] >= SERVICE_TRANSACTIONS_VERSION.1))
+	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
 	let ver: Vec<u32> = client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
```
dvdplm: indentation like above

ethcore/sync/src/chain/propagator.rs
```diff
@@ -48,16 +48,24 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
 
-	if !client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
+	if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
+		let ver: Vec<u32> = client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
+			.take(2)
+			.filter_map(|s| s.parse().ok())
+			.collect();
+			ver.len() == 2 && (ver[0] > SERVICE_TRANSACTIONS_VERSION.0 || (ver[0] == SERVICE_TRANSACTIONS_VERSION.0 && ver[1] >= SERVICE_TRANSACTIONS_VERSION.1))
+	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
+		let ver: Vec<u32> = client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
```
ngotchac: Would be cleaner to do the split in the ifs, and then the rest outside : 
```
let split = if ... { ...split() } else if { ...split() } else { return false };
let ver: Vec<u32> = ...
```

5chdn:
Ok, this finally works.

PR #9055, opened 2018-07-05
remove trait bounds from several structs

debris:
as described here: https://rust-lang-nursery.github.io/api-guidelines/future-proofing.html#c-struct-bounds

### DIFF
--- a/ethcore/types/src/account_diff.rs
+++ b/ethcore/types/src/account_diff.rs
@@ -24,7 +24,7 @@ use bytes::Bytes;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 /// Diff type for specifying a change (or not).
-pub enum Diff<T> where T: Eq {
+pub enum Diff<T> {
 	/// Both sides are the same.
 	Same,
 	/// Left (pre, source) side doesn't include value, right side (post, destination) does.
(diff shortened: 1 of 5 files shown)


ethcore/types/src/account_diff.rs
```diff
@@ -35,9 +35,15 @@ pub enum Diff<T> where T: Eq {
 	Died(T),
 }
 
-impl<T> Diff<T> where T: Eq {
+impl<T> Diff<T> {
 	/// Construct new object with given `pre` and `post`.
-	pub fn new(pre: T, post: T) -> Self { if pre == post { Diff::Same } else { Diff::Changed(pre, post) } }
+	pub fn new(pre: T, post: T) -> Self where T: Eq {
```
niklasad1: I'm not that familiar with `derive rules` for `Eq`, but looks a little strange that we have to specify in the constructor that Self (T: Eq) when we have specified it in the `derive` already!

Can somebody explain this? :)

debris: `derive(Debug)` means

```rust
impl<T> Debug for Diff<T> where T: Debug {
  // ...
}
```

so it is a separate `impl` and has nothing to do with this one

PR #9059, opened 2018-07-06
json-tests moved to tests/ directory

debris:
moved integration json-tests to `ethcore/tests` directory


5chdn:
evmbin does not compile.

debris:
I didn't know that `evmbin` depends on them. @tomusdrw what do you think about moving these tests into their own crate?

tomusdrw:
@debris Fine for me. There reason for dependency is that for cross-client testing there is an option to run json test from a command-line. We could split the part that runs the tests from the tests themselves.

sorpaas:
And we'll also need to update this file: `scripts/evm_jsontests_bench.sh`. Also contains reference to jsontests path.

ethcore/src/lib.rs
```diff
@@ -170,15 +171,12 @@ mod cache_manager;
...
 mod account_db;
 mod builtin;
-mod externalities;
 mod blockchain;
 mod factory;
 mod tx_filter;
 
 #[cfg(test)]
 mod tests;
-#[cfg(feature = "json-tests")]
```
sorpaas: This feature probably needs to be removed from `ethcore/Cargo.toml`.

5chdn:
@debris are you still working on this?

debris:
I'll try coming back to this pr later this week

debris:
If I do not, it can be closed

5chdn:
Ok

Issue #8927, opened 2018-06-19
Be able to run full and light client simultaneously (with portshift)


Tbaut:
As the title says, we should be able to run both clients on the same machine as long as there is a port-shift:

`$ parity --light # Run light client`
`$ parity  --ports-shift 100 # Run full client with ports-shift`

Light and full clients should use separated directories, the keys should be shared. 
A particular attention should be paid to the directory code as it varies between platforms. Proper tests should be performed.

Tbaut:
Not possible without specifying a different directory for each call.

Tbaut:
I guess we should make sure they run per default on different directories.
The keys should be shared.

5chdn:
I don't see why this is relevant. It has never been possible to run two instances with the same data directory.

roninkaizen:
works fine with 2 different config.toml files since version 1.9.x-
always tested with at least one node-
some examples within the wiki would be nice, to make the user aware of the 
exact handling while using parity db kill in such constellations

mttmartin:
@Tbaut, should all directories be separate(except keys) or just the chains directory?

Tbaut:
The keys and db are the 2 directories that matter. keys can prob. be shares (didn't try it) the db should not.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.7 ETH (320.84 USD @ $458.35/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/paritytech/parity/8927/742).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $16,282.08 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [mttmartin](https://gitcoin.co/profile/mttmartin) has started work. 	
 * Q: I'm working on this.

Tbaut:
@mttmartin wait, I think this is already addressed in another PR

Tbaut:
[...]

PR #9064, opened 2018-07-06
Add separate database directory for light client (#8927)

mttmartin:
This allows for the full and light client to run at the same time when using portshift. This adds a new database directory for the light client. I don't immediately see a reason to make unique directories for any others, but please let me know if I've missed something.  

I've tested this on both Linux and Windows.

### DIFF
--- a/parity/configuration.rs
+++ b/parity/configuration.rs
@@ -970,10 +970,12 @@ impl Configuration {
 		let data_path = replace_home("", &base_path);
 		let is_using_base_path = self.args.arg_base_path.is_some();
 		// If base_path is set and db_path is not we default to base path subdir instead of LOCAL.
-		let base_db_path = if is_using_base_path && self.args.arg_db_path.is_none() && self.args.flag_light {
-			"$BASE/chains_light"
-		} else if is_using_base_path && self.args.arg_db_path.is_none() {
-			"$BASE/chains"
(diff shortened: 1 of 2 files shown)


parity/configuration.rs
```diff
@@ -978,8 +978,12 @@ impl Configuration {
 		let data_path = replace_home("", &base_path);
 		let is_using_base_path = self.args.arg_base_path.is_some();
 		// If base_path is set and db_path is not we default to base path subdir instead of LOCAL.
-		let base_db_path = if is_using_base_path && self.args.arg_db_path.is_none() {
+		let base_db_path = if is_using_base_path && self.args.arg_db_path.is_none() && self.args.flag_light {
```
dvdplm: Maybe this is a little more readable?

```rust
let base_db_path = if is_using_base_path && self.args.arg_db_path.is_none() {
  if self.args.flag_light {
    …
  } else {
    …
  }
} else if self.args.flag_light {
  …
} else {
  …
}

```

mttmartin: Thanks for the reviews today. I updated this based on your suggestions.

tomaka:
This will make the light database vanish for anyone who upgrades their node to after this change.
But do we need a migration? The light client should sync really quickly, so I don't know if it's worth the effort.

Tbaut:
>I don't know if it's worth the effort

Definitely not needed, it's quick enough and not widely used yet.

PR #9077, opened 2018-07-09
Replace `std::env::home_dir` with `dirs::home_dir`

niklasad1:
`std::env::home_dir` is deprecated but will probably take a while until
it is deprecated on stable. For more info see https://github.com/rust-lang/rust/pull/51656


5chdn:
For the record, this broke android.

PR #9080, opened 2018-07-09
Fix nightly warnings

ordian:
The first commit switches to stable (since 1.28) API for global allocators when using `memory_profiling` feature.

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -20,7 +20,6 @@ use std::sync::Arc;
 
 use ethcore::basic_account::BasicAccount;
 use ethcore::encoded;
-use ethcore::executed::{Executed, ExecutionError};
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -87,6 +87,7 @@ pub fn extract_transaction_at_index(block: encoded::Block, index: usize, eip86_t
 }
 
 /// Type alias for convenience.
+#[allow(dead_code)]
```
dvdplm: Did you figure out why the compiler thinks `ExecutionResult` is not used? I had a stab a while back but didn't dig very deep.

ordian: @dvdplm looks like the problem was that the type was unnecessary public.
I used Marek's suggestion though.

5chdn:
Removing "on ice" :roll_eyes:

PR #9082, opened 2018-07-10
update snappy

debris:
contains: https://github.com/paritytech/rust-snappy/pull/4

unrelated changes for rand crate are caused by bug in cargo: https://github.com/rust-lang/cargo/issues/5702

cc @andresilva

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,6 +246,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "cmake"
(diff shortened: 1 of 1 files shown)


andresilva:
Yeah, was going to mention the `cmake` requirement. We should update the README to add it to the build dependencies and also other documentation on the wiki. We may also need to update some `Dockerfile`s. cc @5chdn @Tbaut.

PR #9085, opened 2018-07-10
Unify engine error to reject blocks

sorpaas:
Currently in our code, we have several places that deal with engine errors with different behavior. On engine error:

* If `on_new_block` fails, blocks are *rejected* if it's imported externally.
* If `on_new_block` fails, the client *panics* if it's created internally.
* If `on_close_block` fails, the error is *ignored* and we print a warning.

I think the *rejected* behavior is the most reasonable. So this PR changes all three cases to rejecting blocks.

* If we ignore the error and only print a warning, it might be ignored by the node operator as well. It may end up making the chain not having the same best block, which is bad.
* In some cases, a consensus failure might be recoverable. For example, we might have a certain combination of transactions that cause `on_close_block` to fail on a consensus smart contract. In that case, if *rejected* behavior is used, node operators may be able to push/replace transactions and cause the current node to create a new valid block, without bringing down the node.

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -489,8 +489,8 @@ impl Miner {
 
 		let block = match open_block.close() {
 			Ok(block) => block,
-			Err(_) => {
-				warn!(target: "miner", "Closing the block failed due to an engine error. Please check your chain specifications and consensus smart contracts.");
+			Err(err) => {
+				warn!(target: "miner", "Closing the block failed with error {:?}. This is likely an error in chain specificiations or on-chain consensus smart contracts.", err);
 				return None;
 			}
 		};
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -349,11 +349,15 @@ impl Miner {
 					// block not found - create it.
 					trace!(target: "miner", "prepare_block: No existing work - making new block");
 					let params = self.params.read().clone();
-					chain.prepare_open_block(
+
+					match chain.prepare_open_block(
 						params.author,
 						params.gas_range_target,
 						params.extra_data,
-					)
+					) {
+						Ok(block) => block,
+						Err(_) => return None,
+					}
 				}
 			};
 
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -705,7 +705,8 @@ impl Miner {
 		}
 	}
 
-	/// Returns true if we had to prepare new pending block.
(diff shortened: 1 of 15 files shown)


ethcore/src/miner/miner.rs
```diff
@@ -349,11 +349,15 @@ impl Miner {
...
-					chain.prepare_open_block(
+
+					match chain.prepare_open_block(
 						params.author,
 						params.gas_range_target,
 						params.extra_data,
-					)
+					) {
+						Ok(block) => block,
+						Err(_) => return None,
```
tomusdrw: Should we print a warning here?

ethcore/src/miner/miner.rs
```diff
@@ -480,15 +484,21 @@ impl Miner {
 		let elapsed = block_start.elapsed();
 		debug!(target: "miner", "Pushed {} transactions in {} ms", tx_count, took_ms(&elapsed));
 
-		let block = open_block.close();
+		let block = match open_block.close() {
+			Ok(block) => block,
+			Err(_) => {
+				warn!(target: "miner", "Closing the block failed due to an engine error. Please check your chain specifications and consensus smart contracts.");
```
tomusdrw: Maybe display the original error as well? Will give some additional input for debugging.

ethcore/src/miner/miner.rs
```diff
@@ -714,8 +724,14 @@ impl Miner {
...
 			// | Make sure to release the locks before calling that method.             |
 			// --------------------------------------------------------------------------
-			let (block, original_work_hash) = self.prepare_block(client);
-			self.prepare_work(block, original_work_hash);
+			match self.prepare_block(client) {
+				Some((block, original_work_hash)) => {
+					self.prepare_work(block, original_work_hash)
+				},
+				None => {
+					prepare_new = false;
```
tomusdrw: Isn't that a bit misleading? The function should return `true` if new pending block had to be prepared. And in that case it had to be prepared, but the preparation failed.
Just thinking if it won't break any other assumptions, maybe at least update the function comment.

sorpaas: I added `BlockPreparationStatus` which distinguishes three different status.

I think previously I set `prepare_new = false` here is indeed not quite right. We only use the result for `prepare_pending_block` in `prepare_and_update_sealing`. If we don't have to prepare a new block, the function then call `update_sealing`. So if the block had to be prepared but failed, then we shouldn't continue the if statement. (`update_sealing` will also just call `prepare_block` again and early return if it finds out that the conditions does not hold.)

5chdn:
Removing relnotes flag as this is not relevant for users. Or is it?

debris:
it is not

PR #9097, opened 2018-07-11
Revert "Replace `std::env::home_dir` with `dirs::home_dir` (#9077)"

tomaka:
This reverts commit 7e779327ebad5a60e068f39c60bcf944f3c99114.

Waiting for https://github.com/paritytech/parity-common/pull/11
After this PR is merged, will update the Cargo.lock


tomaka:
Note that this is not necessarily the *right* solution. Maybe `dirs` should be fixed, depending on @niklasad1's opinion.

tomaka:
Updated `parity-common`, should be ready for review.

PR #9104, opened 2018-07-12
Fix work-notify.

### LINKED ISSUE #9103: Optional ethcore features should be enabled by default
Recently some parts for ethcore and ethcore-miner were made optional: see #9020 

However, we import `ethcore` and `ethcore-miner` crate from multiple places, an all of those places should specify the required features. It is however easly forgotten.

We should make the features `default`, and other crates that don't require them should just include the crate with `no-default-features`.

Also using the features if they are disabled should trigger an error. Currently 2.0 and 2.1 just silently fail to work correctly. Seems that `work-notify`, `price-info` and  `stratum` are just broken currently.

tomusdrw:
Closes #9103 

It doesn't really enable the features by default, but makes a meta-feature to enable everything at once.

### DIFF
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -80,6 +80,13 @@ tempdir = "0.3"
 trie-standardmap = { git = "https://github.com/paritytech/parity-common" }
 
 [features]
+parity = ["work-notify", "price-info", "stratum"]
+# Large optional features that are enabled by default for Parity,
+# but might be omitted for other dependent crates.
+work-notify = ["ethcore-miner/work-notify"]
+price-info = ["ethcore-miner/price-info"]
+stratum = ["ethcore-stratum"]
+
 # Display EVM debug traces.
 evm-debug = ["evm/evm-debug"]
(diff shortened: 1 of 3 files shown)


ethcore/Cargo.toml
```diff
@@ -80,6 +80,13 @@ tempdir = "0.3"
 trie-standardmap = { git = "https://github.com/paritytech/parity-common" }
 
 [features]
+parity = ["work-notify", "price-info", "stratum"]
```
tomaka: Do we really need this?
I feel like we would get a compilation error in Parity anyway if we forget to enable a feature.

tomusdrw: I think this is just more future-proof in case we extract more optional features, because it's self-contained within `ethcore`. You don't have to go to all other crates and add your own feature, which is error-prone.
That would be especially important for things that might be disabled by default if the feature is not enabled.

Obviously can revert if you don't think it's necessary.

tomaka: It's not very proper, but it's fine I think.

PR #8887, opened 2018-06-13
Remove debian/.deb and centos/.rpm packaging scripts

5chdn:
After #8875 and finalizing #8649

- remove centos pipeline and remove rpm packaging
- remove debian pipeline and remove deb packaging
- rename ubuntu pipelines to generic linux and only build binaries, no packaging
- remove all unused libssl dependencies
- init git submodules in test suite
- do not publish every build from master, only from nightly tag

Test pipeline: https://gitlab.parity.io/parity/parity/pipelines/20791 :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: :rofl:

PR #8778, opened 2018-06-04
Remove windows tray and installer

ascjones:
Part of #8649. Maybe more to remove, creating this PR for now to test the gitlab CI build.

PR #9105, opened 2018-07-12
Changelogs for 1.11.7-stable and 2.0.0-beta

5chdn:
- [x] docs: mark 1.10 as end-of-life
- [x] docs: move changelog for 1.11
- [x] docs: Add changelog for 1.11.7-stable
- [x] docs: add changelog for 2.0.0-beta
- [ ] docs: add release notes for 2.0.0-beta

### DIFF
--- a/docs/CHANGELOG-1.11.md
+++ b/docs/CHANGELOG-1.11.md
@@ -1,3 +1,14 @@
+## Parity [v1.11.7](https://github.com/paritytech/parity/releases/tag/v1.11.7) (2018-07-16)
+
+Parity 1.11.7 is a bug-fix release to improve performance and stability and marks the 1.11 release track as `stable`.
+
+The full list of included changes:
+
+- Backports to 1.11.7-stable ([#9093](https://github.com/paritytech/parity/pull/9093))
+  - Parity-version: stabelize 1.11
+  - Parity-version: bump stable to 1.11.7
+  - Don't fetch snapshot chunks at random ([#9088](https://github.com/paritytech/parity/pull/9088))
+
 ## Parity [v1.11.6](https://github.com/paritytech/parity/releases/tag/v1.11.6) (2018-07-09)
 
 Parity 1.11.6 is a bug-fix release to improve performance and stability.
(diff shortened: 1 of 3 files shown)


Tbaut:
The first lines of the changelog are aggressive and sound like "we removed everything without alternatives". While the reality is more contrasted:
- Public node ability (`--public-node) has been removed
- Parity Wallet, a.k.a Browser UI has been removed, head to Parity UI if you want a UI
- Dapps have been removed, head to Parity UI if you want to see them
- Windows and MacOS versions are not available as installer anymore and the system trays were removed ([#8778](https://github.com/paritytech/parity/pull/8778)). There are still *.exe and *.dpkg binaries. 
- Linux versions are not available as deb-/rpm-packages anymore ([#8887](https://github.com/paritytech/parity/pull/8887)). You can use the binaries <-- we need a doc in the Readme for the binaries

5chdn:
> we need a doc in the Readme for the binaries

what do you mean?

> There are still *.exe and *.dpkg binaries.

dpkg?

Tbaut:
1. we need a linux-beginner-friendly explanation of how to DL, extract, make executable and launch parity in the Readme.
2. pkg maybe? I don't know the extension for mac :)

docs/CHANGELOG-1.11.md
```diff
@@ -0,0 +1,643 @@
+## Parity [v1.11.7](https://github.com/paritytech/parity/releases/tag/v1.11.7) (2018-07-16)
```
tomusdrw: Can we highlight couple of important changes to transaction queue that happened in 1.11?
Namely:
- The pool now limits transactions per-sender (see --tx-queue-per-sender), local transactions also have to obey that limit. Consider increasing the limit via CLI flag when running benchmarks or sending a lot of transactions at once.

- In case the pool is full transactions received over the network, but originating from accounts that we have private keys for might not get accepted to the pool any more with higher priority. Consider running with larger pool size or submitting the transactions directly on the node via `eth_sendRawTransaction`.

5chdn:
Thanks @Tbaut @tomusdrw I addressed your comments.

PR #9107, opened 2018-07-12
Completely remove all dapps struct from rpc

sorpaas:
Previously we return a dapps disabled error. Given that we don't plan to add dapps back, it makes sense to just completely remove those functions.


niklasad1:
I don't know this part of the codebase and don't feel comfortable reviewing this!

5chdn:
fair enough.

PR #9112, opened 2018-07-12
rpc: fix is_major_importing sync state condition

andresilva:
There's a similar definition [here](https://github.com/paritytech/parity/blob/master/ethcore/sync/src/chain/mod.rs#L271) and this makes it consistent with that.

When we're waiting for peers we're technically not syncing but we're waiting to do it.

### DIFF
--- a/parity/informant.rs
+++ b/parity/informant.rs
@@ -35,7 +35,7 @@ use io::{TimerToken, IoContext, IoHandler};
 use light::Cache as LightDataCache;
 use light::client::{LightChainClient, LightChainNotify};
 use number_prefix::{binary_prefix, Standalone, Prefixed};
-use parity_rpc::{is_major_importing};
+use parity_rpc::is_major_importing_or_waiting;
 use parity_rpc::informant::RpcStats;
 use ethereum_types::H256;
 use bytes::Bytes;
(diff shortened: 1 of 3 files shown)


5chdn:
Does "fix" indicate we should backport it?

andresilva:
Okay this has the unintended consequence of breaking the informant:

```
❯ ./target/release/parity --chain kovan --warp-barrier 7000000000
…
2018-07-13 11:45:39  Syncing       #0 0xa3c5…2cb9     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed        #0    5/25 peers   9 KiB chain 7 KiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,    0 µs
2018-07-13 11:45:44  Syncing       #0 0xa3c5…2cb9     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed        #0    5/25 peers   9 KiB chain 7 KiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,    0 µs
2018-07-13 11:45:49  Syncing       #0 0xa3c5…2cb9     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed        #0    5/25 peers   9 KiB chain 7 KiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,    0 µs
2018-07-13 11:45:54  Syncing       #0 0xa3c5…2cb9     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed        #0    5/25 peers   9 KiB chain 7 KiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,    0 µs
```

It just starts printing because it thinks it's syncing even though it's still waiting for warp sync. Will need to fix this.

andresilva:
Added a special case for the informant so that it doesn't print when waiting for peers.

PR #9113, opened 2018-07-12
AuthorityRound: finalize blocks

andresilva:
Implement `ancestry_actions` to mark finalized blocks using the existing rolling finality. I've also changed the `is_epoch_end` method so that it gets the recently finalized blocks as argument and can check the transition store using that.

### DIFF
--- a/ethcore/light/src/client/mod.rs
+++ b/ethcore/light/src/client/mod.rs
@@ -313,7 +313,7 @@ impl<T: ChainDataFetcher> Client<T> {
 					The node may not be able to synchronize further.", e);
 			}
 
-			let epoch_proof = self.engine.is_epoch_end(
+			let epoch_proof = self.engine.is_epoch_end_light(
 				&verified_header,
 				&|h| self.chain.block_header(BlockId::Hash(h)).and_then(|hdr| hdr.decode().ok()),
 				&|h| self.chain.pending_transition(h),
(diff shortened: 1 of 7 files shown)


andresilva:
The changes I did to `Engine::is_epoch_end` are not compatible with the way the light client works.

5chdn:
@andresilva are you still working on this?

rphmeier:
Seems like nobody reviewed it. @sorpaas can you take a look? I'll try to as well but I'm out-of-the-loop with parity-ethereum stuff.

sorpaas:
I remember the issue is with light client. Probably @andresilva can clarify more.

I didn't find anywhere that we write finalization info to light client currently (or do we?), so for the purpose of this refactoring PR, it may be okay to continue to ignore finality info for light client?

rphmeier:
The light client does track finality, but only for blocks which register validator set changes. Requires finality proofs though.

andresilva:
I haven't worked on this further after submitting the original PR. I think the changes for the full client make sense, but I changed signature of `check_epoch_end` to `check_epoch_end(Self, Header, [H256], BlockChain, Client)`, where the hashes are for recently finalized blocks since finality is tracked by the engine regardless of epoch change signals, but this is incompatible with the light client. The light client does track finality, but it tracks finality for all blocks in the current epoch and "starting from scratch" on each block import.

I think the easiest way to integrate this is to keep the existing finality tracking logic for the light client, this probably requires us to introduce a special light version of `check_epoch_end` that uses the same logic as previously. WDYT?

5chdn:
Adding patch label to make sure this ends up in 2.1-beta once branched out.

5chdn:
Please rebase on master

5chdn:
Please reopen when ready

PR #9115, opened 2018-07-13
Make sure to create full pending blocks

tomusdrw:
With recent optimizations we only query `block_gas_limit / min_tx_gas` transactions from the pending set.
This however might be a bit too little to fill up the entire block (since you can get couple of high-gas transactions that won't fit). This in turn produces blocks that have like 90% gas limit utilization.

To prevent that, the PR:
1. Increases the `MAX_SKIPPED_TRANSACTIONS` constant, so we will now try at most 128 transactions before quitting
2. Always gets `MAX_SKIPPED_TRANSACTIONS` more transactions from the pending set.

Also I've added a feature to the main `Cargo.toml` that allows you to test mined block import while syncing mainnet (so you get all the transactions and dynamics of mainnet), by sending `eth_submitWork` with any seal. The seal for local import is ignored.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ winapi = { version = "0.3.4", features = ["winsock2", "winuser", "shellapi"] }
 daemonize = { git = "https://github.com/paritytech/daemonize" }
 
 [features]
+miner-debug = ["ethcore/miner-debug"]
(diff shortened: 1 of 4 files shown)


Ri2k18:
0x00bbb76D5771060214CEBe5CA6152F7825ba7ADC

PR #9117, opened 2018-07-13
block cleanup

debris:
the goal of this pr is to reduce number of times block data is copied or hashed


ethcore/src/blockchain/blockchain.rs
```diff
@@ -1081,11 +1080,11 @@ impl BlockChain {
 			let mut best_block = self.pending_best_block.write();
 			if is_best && update.info.location != BlockLocation::Branch {
 				batch.put(db::COL_EXTRA, b"best", &update.info.hash);
-				let block = encoded::Block::new(update.block.to_vec());
```
debris: entire block was copied here

ethcore/src/block.rs
```diff
@@ -116,8 +116,6 @@ pub struct ExecutedBlock {
 	pub traces: Tracing,
 	/// Hashes of last 256 blocks.
 	pub last_hashes: Arc<LastHashes>,
-	/// Finalization flag.
-	pub is_finalized: bool,
```
debris: this flag is completely redundant and is never set to true

ethcore/src/block.rs
```diff
@@ -228,26 +225,6 @@ impl ::parity_machine::Transactions for ExecutedBlock {
...
-	fn is_finalized(&self) -> bool {
-		self.is_finalized
-	}
-
-	fn mark_finalized(&mut self) {
-		self.is_finalized = true;
-	}
-}
-
-impl ::parity_machine::WithMetadata for ExecutedBlock {
```
debris: this interface is never used

ethcore/src/block.rs
```diff
@@ -477,8 +447,8 @@ impl<'x> OpenBlock<'x> {
 		if s.block.header.transactions_root().is_zero() || s.block.header.transactions_root() == &KECCAK_NULL_RLP {
```
debris: I think we should also get rid of these checks, cause they may lead to invalid state. e.g

```rust
let closed_block = open_block.close();
let reopened_block = closed_block.reopen(engine);
reopened_block.push_transactions(transactions);
// close_and_lock is not recalculating `transactions_root`
let locked_block = reopened_block.close_and_lock();
```

ethcore/src/client/client.rs
```diff
@@ -298,7 +297,7 @@ impl Importer {
 
 						let transactions_len = closed_block.transactions().len();
 
-						let route = self.commit_block(closed_block, &header, &bytes, client);
+						let route = self.commit_block(closed_block, &header, encoded::Block::new(bytes), client);
```
debris: in future, I'd like this function (and blockchain.rs functions) to take as argument some `MetaBlock` that contains both plain block and it's encoded version. Currently we are encoding and then decoding stuff that we just encoded. That's not efficient

ethcore/src/client/client.rs
```diff
@@ -445,28 +443,29 @@ impl Importer {
...
 			let mut batch = DBTransaction::new();
-			chain.insert_unordered_block(&mut batch, block_bytes, receipts, None, false, true);
+			chain.insert_unordered_block(&mut batch, block, receipts, None, false, true);
 			// Final commit to the DB
 			db.write_buffered(batch);
 			chain.commit();
 		}
 		db.flush().expect("DB flush failed.");
-		Ok(hash)
+		Ok(())
```
debris: this value was never used, therefore not needed

ethcore/src/block.rs
sorpaas: I think we're going to use it #9113? cc @andresilva 

I think we won't implement Casper again in the short terms, so some of the #8401 refactorings might use a better interface. But I do think allowing finalized block is a feature that we may use in other engines.

sorpaas: Ah right. Aura only finalize ancestry blocks so it only needs `BlockChain::mark_finalized`. However, do we want to support instant finalization for some of our engines in the future? That's where we will need this flag.

I'm not sure whether POC-1 engine (which will replace the current Tendermint engine) will need this. [The spec](https://github.com/w3f/polkadot-spec/blob/master/spec.md#consensus) do mention instant finality. But I'm not really familiar with the algorithm.

debris: > However, do we want to support instant finalization for some of our engines in the future? That's where we will need this flag.

Than we should add this flag in the future. Imo, we shouldn't keep in our codebase anything that is not used right now

sorpaas: I can't say I fully agree with this, but it sounds okay to me. :+1: 

Based on the above rationale, however, another thing we need to remove is metadata (https://github.com/paritytech/parity/blob/21d3de236005732253dee442cfbcb52ebcbbc1e7/ethcore/src/block.rs#L119 and https://github.com/paritytech/parity/blob/21d3de236005732253dee442cfbcb52ebcbbc1e7/ethcore/src/blockchain/extras.rs#L155). Also introduced by #8401, but not used by current engines.

debris: metadata removed! :)

sorpaas: We can remove this field as well:

https://github.com/paritytech/parity-ethereum/blob/a598be7b6e3b0d35006d6c25d99e966b002749c7/ethcore/src/blockchain/extras.rs#L156

Metadata or finalization has not been set by anyone, so we can just change the `!is_short_version` to be 5 RLP fields. :)

debris: done ;)

ethcore/src/blockchain/extras.rs
```diff
@@ -180,7 +177,8 @@ impl rlp::Decodable for BlockDetails {
 	fn decode(rlp: &rlp::Rlp) -> Result<Self, rlp::DecoderError> {
 		let use_short_version = match rlp.item_count()? {
 			4 => true,
-			6 => false,
+			// 6, cause we used to store metadata
+			5 | 6 => false,
```
debris: to keep database backwards compatibility

sorpaas: Looks fine but I think we actually never had any `6` cases up to this point. `6` is possible after the Aura finality PR is merged. Right now we don't have any. It's also not possible to use any custom chain spec to reach the `6` case.

The only case where `6` was used is in https://github.com/paritytech/parity-ethereum/pull/8654, but we're not using that code in the short-term.

debris:
@andresilva please check the pr again and approve/reject changes

debris:
🎉

PR #9119, opened 2018-07-13
Remove node-health

amaurymartiny:
- Remove `node-health` and the `parity_nodeHealth` RPC
- Remove `ntp_servers`, as it was only used by `node-health` to check if local clock is sync.

---

Parity UI starting from v0.3.0 is free from `parity_nodeHealth` calls. The info returned by node-health has been replaced by:
- eth_syncing for sync state
- parity_netPeers for network state
- a [JS implementation](https://github.com/paritytech/fether/tree/master/packages/parity-electron#checkclocksync-promiseobjectisclocksync-boolean-timedrift-number) of whether the clock is sync or not.


parity/cli/mod.rs
```diff
@@ -794,10 +794,6 @@ usage! {
 			"--no-config",
 			"Don't load a configuration file.",
 
-			ARG arg_ntp_servers: (String) = "0.parity.pool.ntp.org:123,1.parity.pool.ntp.org:123,2.parity.pool.ntp.org:123,3.parity.pool.ntp.org:123", or |c: &Config| c.misc.as_ref()?.ntp_servers.clone().map(|vec| vec.join(",")),
-			"--ntp-servers=[HOSTS]",
-			"Comma separated list of NTP servers to provide current time (host:port). Used to verify node health. Parity uses pool.ntp.org NTP servers; consider joining the pool: http://www.pool.ntp.org/join.html",
-
```
ascjones: I believe this should be deprecated as per: https://github.com/paritytech/parity/pull/9036

amaurymartiny: Done.

amaurymartiny: Right, thanks!

parity/cli/mod.rs
```diff
@@ -1350,12 +1350,12 @@ struct Snapshots {
 #[derive(Default, Debug, PartialEq, Deserialize)]
 #[serde(deny_unknown_fields)]
 struct Misc {
-	ntp_servers: Option<Vec<String>>,
 	logging: Option<String>,
 	log_file: Option<String>,
 	color: Option<bool>,
 	ports_shift: Option<u16>,
 	unsafe_expose: Option<bool>,
+	_legacy_ntp_servers: Option<Vec<String>>,
```
andresilva: I know we did this for some other deprecated options, but do we need to keep this fields? Seems like it isn't used anywhere besides when parsing the argument. We can use `or |_| None` in that case.

rpc/src/v1/tests/mocked/parity.rs
```diff
@@ -21,7 +21,6 @@ use ethcore_logger::RotatingLogger;
 use ethereum_types::{Address, U256, H256};
 use ethstore::ethkey::{Generator, Random};
 use miner::pool::local_transactions::Status as LocalTransactionStatus;
-use node_health::{self, NodeHealth};
 use parity_reactor;
```
andresilva: Unused import.

andresilva:
The NTP code was my first contribution to parity 😛.

parity/cli/mod.rs
amaurymartiny: So I removed this field in my last commit. But the other ones are still there. Probably to be removed in the scope of another pr?

PR #9120, opened 2018-07-13
Remove all dapp permissions related settings

sorpaas:
Should only be merged after #9107. I submit this as a separate PR because this may need more polishing, or I may have removed too many stuff.

Completely remove dapp permission settings from AccountProvider and RPC:

* In RPC, all available accounts are returned, regardless of the origin. (Previously we return accounts based on dapps policy).
* It's not possible to set "default account" (as for dapps) any more. This is always the first account in the available account list.


5chdn:
there is conflict now in rpc.

rpc/src/v1/extractors.rs
```diff
@@ -36,13 +36,11 @@ pub struct RpcExtractor;
...
-	fn read_metadata(&self, origin: Option<String>, user_agent: Option<String>, dapps_origin: Option<String>) -> Metadata {
+	fn read_metadata(&self, _origin: Option<String>, user_agent: Option<String>, _dapps_origin: Option<String>) -> Metadata {
 		Metadata {
-			origin: match (origin.as_ref().map(|s| s.as_str()), user_agent, dapps_origin) {
-				(Some("null"), _, Some(dapp)) => Origin::Dapps(dapp.into()),
-				(Some(dapp), _, _) => Origin::Dapps(dapp.to_owned().into()),
-				(None, Some(service), _) => Origin::Rpc(service.into()),
-				(None, _, _) => Origin::Rpc("unknown".into()),
+			origin: match user_agent {
+				Some(service) => Origin::Rpc(service.into()),
```
tomusdrw: I think it might be good to actually use the `Origin` as well (if available)

rpc/src/v1/extractors.rs
```diff
@@ -36,13 +36,11 @@ pub struct RpcExtractor;
 impl HttpMetaExtractor for RpcExtractor {
 	type Metadata = Metadata;
 
-	fn read_metadata(&self, origin: Option<String>, user_agent: Option<String>, dapps_origin: Option<String>) -> Metadata {
+	fn read_metadata(&self, _origin: Option<String>, user_agent: Option<String>, _dapps_origin: Option<String>) -> Metadata {
```
tomusdrw: Should we get rid of `_dapps_origin` in the trait?

rpc/src/v1/impls/eth.rs
```diff
@@ -509,12 +502,10 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 		}
 	}
 
-	fn author(&self, meta: Metadata) -> Result<RpcH160> {
-		let dapp = meta.dapp_id();
-
+	fn author(&self, _meta: Metadata) -> Result<RpcH160> {
```
tomusdrw: Don't need the `Meta` on the trait level I suppose.

rpc/src/v1/impls/eth.rs
```diff
@@ -532,10 +523,9 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 		Ok(RpcU256::from(default_gas_price(&*self.client, &*self.miner, self.options.gas_price_percentile)))
 	}
 
-	fn accounts(&self, meta: Metadata) -> Result<Vec<RpcH160>> {
-		let dapp = meta.dapp_id();
-
-		let accounts = self.dapp_accounts(dapp.into())?;
+	fn accounts(&self, _meta: Metadata) -> Result<Vec<RpcH160>> {
```
tomusdrw: Same as above

rpc/src/v1/impls/light/parity.rs
```diff
@@ -105,13 +105,10 @@ impl ParityClient {
 impl Parity for ParityClient {
 	type Metadata = Metadata;
 
-	fn accounts_info(&self, dapp: Trailing<DappId>) -> Result<BTreeMap<H160, AccountInfo>> {
-		let dapp = dapp.unwrap_or_default();
-
+	fn accounts_info(&self) -> Result<BTreeMap<H160, AccountInfo>> {
```
tomusdrw: Should we accept (and ignore) the trailing param for backward compatibility?

rpc/src/v1/impls/parity.rs
```diff
@@ -110,12 +110,8 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 {
 	type Metadata = Metadata;
 
-	fn accounts_info(&self, dapp: Trailing<DappId>) -> Result<BTreeMap<H160, AccountInfo>> {
-		let dapp = dapp.unwrap_or_default();
-
-		let dapp_accounts = self.accounts
-			.note_dapp_used(dapp.clone().into())
-			.and_then(|_| self.accounts.dapp_addresses(dapp.into()))
+	fn accounts_info(&self) -> Result<BTreeMap<H160, AccountInfo>> {
```
tomusdrw: same as with `accounts()`

rpc/src/v1/impls/parity.rs
```diff
@@ -144,11 +140,8 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 		self.accounts.locked_hardware_accounts().map_err(|e| errors::account("Error communicating with hardware wallet.", e))
 	}
 
-	fn default_account(&self, meta: Self::Metadata) -> Result<H160> {
-		let dapp_id = meta.dapp_id();
-
-		Ok(self.accounts
-			.dapp_default_address(dapp_id.into())
+	fn default_account(&self, _meta: Self::Metadata) -> Result<H160> {
```
tomusdrw: Remove `meta` on the trait level

rpc/src/v1/metadata.rs
```diff
@@ -32,24 +32,9 @@ pub struct Metadata {
...
-		match self.origin {
-			Origin::Dapps(ref dapp) => dapp.clone(),
-			Origin::Ws { ref dapp, .. } => dapp.clone(),
-			Origin::Signer { ref dapp, .. } => dapp.clone(),
-			_ => DappId::default(),
-		}
-	}
-
 	/// Returns true if the request originates from a Dapp.
 	pub fn is_dapp(&self) -> bool {
```
tomusdrw: Maybe just remove completelty?

rpc/src/v1/traits/parity_accounts.rs
```diff
@@ -72,57 +72,6 @@ build_rpc_trait! {
 		#[rpc(name = "parity_setAccountMeta")]
 		fn set_account_meta(&self, H160, String) -> Result<bool>;
 
-		/// Sets addresses exposed for particular dapp.
-		/// Setting a non-empty list will also override default account.
-		/// Setting `None` will resets visible account to what's visible for new dapps
-		/// (does not affect default account though)
-		#[rpc(name = "parity_setDappAddresses")]
-		fn set_dapp_addresses(&self, DappId, Option<Vec<H160>>) -> Result<bool>;
```
tomusdrw: I wonder if we should mark the methods as deprecated first and only remove them in the next version?

rpc/src/v1/types/provenance.rs
```diff
@@ -67,115 +59,45 @@ impl fmt::Display for Origin {
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
 		match *self {
 			Origin::Rpc(ref origin) => write!(f, "{} via RPC", origin),
-			Origin::Dapps(ref origin) => write!(f, "Dapp {}", origin),
 			Origin::Ipc(ref session) => write!(f, "IPC (session: {})", session),
-			Origin::Ws { ref session, ref dapp } => write!(f, "{} via WebSocket (session: {})", dapp, session),
-			Origin::Signer { ref session, ref dapp } => write!(f, "{} via UI (session: {})", dapp, session),
+			Origin::Ws { ref session } => write!(f, "WebSocket (session: {})", session),
+			Origin::Signer { ref session } => write!(f, "Signer (session: {})", session),
```
tomusdrw: `Signer` might not be the most descriptive, maybe `Secure Session`

rpc/src/v1/impls/light/parity.rs
sorpaas: The issue is that we removed several methods, so it's already breaking backward compatibility no matter what. Would it still be useful to keep backward compatibility just for this function?

rpc/src/v1/traits/parity_accounts.rs
sorpaas: cc @5chdn. But I think we have 2.0 so it's okay to break backward compatibility from there? ;)

5chdn: Let's remove them for 2.0.1 and highlight that in release notes.

debris:
@tomusdrw @andresilva please review

rpc/src/v1/extractors.rs
```diff
@@ -36,9 +36,9 @@ pub struct RpcExtractor;
 impl HttpMetaExtractor for RpcExtractor {
 	type Metadata = Metadata;
 
-	fn read_metadata(&self, _origin: Option<String>, user_agent: Option<String>, _dapps_origin: Option<String>) -> Metadata {
+	fn read_metadata(&self, origin: Option<String>, user_agent: Option<String>) -> Metadata {
 		Metadata {
-			origin: match user_agent {
+			origin: match user_agent.or(origin) {
```
tomusdrw: IMHO should be other way around, `Origin` provides more info than `UserAgent`. Maybe best just do:
```
let service = format!("{} / {}", origin.unwrap_or("unknown origin"), user_agent.unwrap_or("unknown agent"));
```
?

amaurymartiny:
The wiki should also be updated: https://wiki.parity.io/JSONRPC-parity_accounts-module

Issue #9075, opened 2018-07-09
Add option to include transaction receipts in eth_getBlockByHash


medvedev1088:
In [some applications](https://github.com/medvedev1088/ethereum-etl) where performance is important it's convenient to query all receipts in a block in a single request. 

Since receipts are part of a block, as defined in the yellowpaper, it makes sense to allow including them in the block response for [eth_getBlockByHash](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbyhash) and [eth_getBlockByNumber](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbynumber) by adding the boolean parameter to the request:

> Parameters
> - QUANTITY|TAG - integer of a block number, or the string "earliest", "latest" or "pending", as in the default block parameter.
> - Boolean - If true it returns the full transaction objects, if false only the hashes of the transactions.
> - **Boolean - If true it returns the full receipts objects, if false the receipts objects are not returned.**

Here is an analogous issue for go-ethereum https://github.com/ethereum/go-ethereum/issues/17044

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 350.0 DAI (350.0 USD @ $1.0/DAI) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/paritytech/parity/9075/741).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $15,961.24 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [shamardy](https://gitcoin.co/profile/shamardy) has started work. 	
 * Q: Working on this

gitcoinbot:
@shamardy Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
[...]

PR #9126, opened 2018-07-15
Add option to include transaction receipts in eth_getBlockByHash #9075

shamardy:
closes https://github.com/paritytech/parity/issues/9075


sorpaas:
Do we need proper EIPs for this `eth_` namespace change?

shamardy:
@sorpaas @5chdn any update here?

sorpaas:
@shamardy Code looks good to me. The issue I see is whether we would need an EIP for this.

The current design can be confusing -- we have too booleans called in a line, and users can easily mix them up. Just an idea: maybe we can change this to accept either a boolean (which it is the old behavior), or a JSON dict with two optional fields `{ fullTransactions: bool, fullReceipts: bool }`.

tjayrush:
Personally, coming from the `user's` perspective, I think the original design works perfectly. Two bools is easy to understand, not confusing. 

If users mix the inputs up during the call, they will quickly discover the problem because the returned data won't parse. I'd rather see a simple, straightforward interface.

Also, I think this issue https://github.com/paritytech/parity-ethereum/issues/9279#issuecomment-410657893 is trying to accomplish almost exactly the same thing. They should be combined.

Finally, this should be an EIP not an issue against Parity. In that way, it would get added to other clients (particularly `geth`). We should be moving towards more consistent RPC interfaces across clients as opposed to less consistent.

debris:
> Do we need proper EIPs for this `eth_` namespace change?

This should definitely be an EIP fist. The issue was incorrectly labeled. Please open an EIP [here](https://github.com/ethereum/EIPs) and reopen the issue if accepted. I'm sorry @shamardy

PR #9130, opened 2018-07-16
parity: fix UserDefaults json parser

### LINKED ISSUE #8550: Delete `user_defaults` when doing `db kill`

> I'm running:
>
> - **Which Parity version?**:

```
Parity/v1.12.0-unstable-e30839e-20180505/x86_64-linux-gnu/rustc1.25.0-dev
```

> - **Which operating system?**: 

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.4 LTS
Release:        16.04
Codename:       xenial
```

> - **How installed?**: via installer / homebrew / binaries / from source

compiled rustc (1.25.dev) --> compiled cargo (0.26) --> compiled parity (git:e30839e85ff74cd5eb087ef19a48d70afd1fd285)

> - **Are you fully synchronized?**: 

Yes. Completely new setup using `--config dev`

> - **Which network are you connected to?**: ethereum / ropsten / kovan / ...

built-in dev config

> - **Did you try to restart the node?**: no / yes

n/a. The problem appears immediately after start

**expected behavior:** Export state to file with `parity export state foo.txt`. 

**actual behavior:** I re-started the node with `--fat-db` on and still received the error.

I run

```
parity --config dev --fat-db on
``` 
Starting with no db at all

```
~$ find .
.
./.bash_logout
./.profile
./.bashrc
```

Even with `--fat-db on` from the beginning, in another terminal I get:

```
~$ parity export state foo.txt
This command requires Parity to be synced with --fat-db on.
```

I notice the word "synced", is that something I need to do in addition to running with `--fat-db on` from an _empty_ database?

andresilva:
Fixes #8550.


ordian:
Dumb question: why can't we use serde derive here instead?

andresilva:
@ordian Seems like this code is a bit old so that's probably the reason. `Algorithm` and `Mode` currently do not implement `Serialize`/`Deserialize` and it doesn't feel right to add those implementations in `util/journaldb/lib.rs` or `ethcore/src/client/config.rs` since those crates don't really care about JSON. The alternative would be to introduce newtypes that wrap mode and algorithm in `UserDefaults`. Do you think we should do this?

ordian:
@andresilva I would prefer having derive, because it is more future-proof and less code to read - easier to reason about. As for `Algorithm` and `Mode`, we can e.g. use [`deserialize_with`](https://serde.rs/field-attrs.html#serdedeserializewith--path) or [`remote-derive`](https://serde.rs/remote-derive.html).

5chdn:
F-Label Society :rofl:

andresilva:
@5chdn Indeed 😆 I'm having a hard time learning not to use F-labels on PRs.

andresilva:
@ordian I've updated the PR to use `serde_derive`. The format of the `mode` field changed but only this value is lost when upgrading (defaults to active), so I don't think it's much of an issue.

parity/user_defaults.rs
```diff
@@ -14,110 +14,100 @@
...
 }
 
+pub fn mode_deserialize<'de, D>(deserializer: D) -> Result<Mode, D::Error>
+where D: Deserializer<'de> {
+	// this was added for migration purposes. since the format of the `mode`
+	// field in `UserDefaults` changed the deserialization of the whole struct
+	// would fail. this way we keep the existing values for the other fields and
+	// only provide a default for the `mode` field in case it fails to
+	// deserialize.
+	Ok(Mode::deserialize(deserializer).unwrap_or(Mode::Active))
```
ordian: Maybe we can preserve backwards compatibility like [this](http://play.rust-lang.org/?gist=2a6b710c8a82e83f2aba958bb0172e26&version=stable&mode=debug&edition=2015)?

andresilva: Nice! But apparently `flatten` can only be used with structs and maps:

`2018-07-17 13:11:47  Error loading user defaults file: ErrorImpl { code: Message("can only flatten structs and maps"), line: 1, column: 107 }`

Always prints that when starting although it seems to work. Also the serialization of `Duration` is different so also not backwards compatible.

(I didn't know you could use crates in rust playground that's super useful!)

ordian: Ah, looks like we're using `serde` `1.0.37` and support for flattening internally tagged enums was only implemented in [`1.0.46`](https://github.com/serde-rs/serde/issues/1189). The difference in `Duration` serialization is also probably due to different versions used in playground in parity codebase.

(rust playground was merged with `Integer32`'s playground implementation [one year ago](https://users.rust-lang.org/t/the-official-rust-playground-now-has-the-top-100-crates-available/11817))

andresilva:
Updated serde and added the `Seconds(Duration)` wrapper. It should be backwards compatible with the previous format now.

parity/user_defaults.rs
```diff
@@ -18,22 +18,65 @@ use std::fs::File;
...
 use journaldb::Algorithm;
 use ethcore::client::{Mode as ClientMode};
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "lowercase")]
+#[derive(Clone)]
+pub struct Seconds(Duration);
+
+impl Seconds {
+    pub fn value(&self) -> u64 {
```
ordian: There are some spaces left from copy-paste, please M-x tabify

PR #9138, opened 2018-07-17
Fix bugfix hard fork logic

sorpaas:
A bugfix hard fork is something that should always be on by default. When checking, it should use `||` instead of `&&`. In addition, EIP86, EIP98, etc do not belong to this category.

This PR also changes the function to return a string instead, so it can tell the operator what needs to be changed in the chain spec config.

### DIFF
--- a/ethcore/src/spec/spec.rs
+++ b/ethcore/src/spec/spec.rs
@@ -57,7 +57,7 @@ fn fmt_err<F: ::std::fmt::Display>(f: F) -> String {
 
 /// Parameters common to ethereum-like blockchains.
 /// NOTE: when adding bugfix hard-fork parameters,
-/// add to `contains_bugfix_hard_fork`
+/// add to `nonzero_bugfix_hard_fork`
 ///
 /// we define a "bugfix" hard fork as any hard fork which
 /// you would put on-by-default in a new chain.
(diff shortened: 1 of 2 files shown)


5chdn:
does this require a backport?

sorpaas:
Given this is a bugfix for bugfix hard fork and the change is simple, backporting it would be fine. :)

ethcore/src/spec/spec.rs
```diff
@@ -188,13 +188,21 @@ impl CommonParams {
...
 
-	/// Whether these params contain any bug-fix hard forks.
-	pub fn contains_bugfix_hard_fork(&self) -> bool {
-		self.eip98_transition != 0 && self.eip155_transition != 0 &&
-			self.validate_receipts_transition != 0 && self.eip86_transition != 0 &&
-			self.eip140_transition != 0 && self.eip210_transition != 0 &&
-			self.eip211_transition != 0 && self.eip214_transition != 0 &&
-			self.validate_chain_id_transition != 0 && self.dust_protection_transition != 0
+	/// Return Some if the current parameters contain a bugfix hard fork not on block 0.
+	pub fn nonzero_bugfix_hard_fork(&self) -> Option<&'static str> {
```
niklasad1: `static` lifetime annotation is not required anymore so ```Option<&str>``` should be enough!

PR #9140, opened 2018-07-17
Comply EIP-86 with the new definition

### LINKED ISSUE #9127: Incorrect implementation of EIP 86
The specification of [EIP 86](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-86.md) suggests that new contracts be created using the semantics of the `CREATE2(value, salt, mem_start, mem_size)` opcode. Primarily, this requires new contract addresses be created from `hash(sender + salt + code_hash)`.

Currently it seems that Parity does not match the spec since:
* [`FromSenderAndCodeHash`](https://github.com/paritytech/parity/blob/c7d21841a427b74c880d2d45010df51a764da8dd/ethcore/src/executive.rs#L70) does not include a salt, which [prevents a contract from calling `CREATE2` multiple times](https://github.com/ethereum/solidity/issues/2136)
* [`create_address_scheme`](https://github.com/paritytech/parity/blob/123b6ae62ef266131f24da9b184e68b16955fb06/ethcore/src/machine.rs#L312) uses `FromCodeHash` as its post-eip86 default
* the [implementation of CREATE2](https://github.com/paritytech/parity/blob/master/ethcore/evm/src/interpreter/mod.rs#L317) is not only unsalted but also fails to pop all four arguments off the stack 

Hopefully I'm not missing something very basic here.

sorpaas:
Fixes #9127

The current version of EIP-86 differs from the one when it was implemented in parity-ethereum in point 3 and point 4. For point 3, a new salt was added, allowing CREATE2 to create multiple contracts with the same init code. Point 4 is the same as EIP-684, which we have already implemented.

The default contract address scheme was also changed to be always the old one (sender and nonce), regardless of EIP-86.

Note that this is backward incompatible, with the assumption that none (or not many) working chains have attempted to set the `eip86Transition` flag, because it's not really useful by itself.

### DIFF
--- a/rpc/src/v1/types/transaction.rs
+++ b/rpc/src/v1/types/transaction.rs
@@ -161,8 +161,8 @@ pub struct RichRawTransaction {
 
 impl RichRawTransaction {
 	/// Creates new `RichRawTransaction` from `SignedTransaction`.
-	pub fn from_signed(tx: SignedTransaction, block_number: u64, eip86_transition: u64) -> Self {
-		let tx = Transaction::from_signed(tx, block_number, eip86_transition);
+	pub fn from_signed(tx: SignedTransaction) -> Self {
+		let tx = Transaction::from_signed(tx);
 		RichRawTransaction {
 			raw: tx.raw.clone(),
 			transaction: tx,
@@ -172,7 +172,7 @@ impl RichRawTransaction {
 
 impl Transaction {
 	/// Convert `LocalizedTransaction` into RPC Transaction.
-	pub fn from_localized(mut t: LocalizedTransaction, _eip86_transition: u64) -> Transaction {
+	pub fn from_localized(mut t: LocalizedTransaction) -> Transaction {
 		let signature = t.signature();
 		let scheme = CreateContractAddress::FromSenderAndNonce;
 		Transaction {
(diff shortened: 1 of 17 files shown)


5chdn:
Didn't we enable eip86 on kovan? Will this break existing networks that activated it?

sorpaas:
I'm really hoping we can get the concept of "nightly" as in Rust compiler. In that way, we can define features where you cannot activate on stable/beta parity-ethereum and be able to change their definitions.

5chdn:
Yes, that's certainly the idea of the nightly track. Just asking because I'll be the one writing the release notes at some point :-p

rpc/src/v1/types/transaction.rs
```diff
@@ -172,9 +172,9 @@ impl RichRawTransaction {
 
 impl Transaction {
 	/// Convert `LocalizedTransaction` into RPC Transaction.
-	pub fn from_localized(mut t: LocalizedTransaction, eip86_transition: u64) -> Transaction {
+	pub fn from_localized(mut t: LocalizedTransaction, _eip86_transition: u64) -> Transaction {
```
debris: if it's unused, can we just remove it?

rpc/src/v1/types/transaction.rs
```diff
@@ -206,9 +206,9 @@ impl Transaction {
 	}
 
 	/// Convert `SignedTransaction` into RPC Transaction.
-	pub fn from_signed(t: SignedTransaction, block_number: u64, eip86_transition: u64) -> Transaction {
+	pub fn from_signed(t: SignedTransaction, block_number: u64, _eip86_transition: u64) -> Transaction {
```
debris: same here

holiman:
How do I enalbe this in config? That is, what specific `eipxxTransition` name should I use in common/params ? Needed for hive.

debris:
@holiman it is `eip86Transition`

PR #9141, opened 2018-07-17
Insert PROOF messages for some cases in blockchain

sorpaas:
Change non-test code to use `expect` and avoid `unwrap`.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -536,7 +536,9 @@ impl BlockChain {
 		};
 
 		// load best block
-		let best_block_hash = match bc.db.key_value().get(db::COL_EXTRA, b"best").expect("Low level database error. Some issue with disk?") {
+		let best_block_hash = match bc.db.key_value().get(db::COL_EXTRA, b"best")
+			.expect("Low level database error. Some issue with disk?")
+		{
 			Some(best) => {
 				H256::from_slice(&best)
 			}
@@ -571,8 +573,11 @@ impl BlockChain {
 
 		{
 			// Fetch best block details
-			let best_block_total_difficulty = bc.block_details(&best_block_hash).expect("Best block is from a known block hash; a known block hash always comes with a known block detail; qed").total_difficulty;
-			let best_block_rlp = bc.block(&best_block_hash).expect("Best block is from a known block hash; qed");
+			let best_block_total_difficulty = bc.block_details(&best_block_hash)
+				.expect("Best block is from a known block hash; a known block hash always comes with a known block detail; qed")
+				.total_difficulty;
+			let best_block_rlp = bc.block(&best_block_hash)
+				.expect("Best block is from a known block hash; qed");
 
 			// and write them
 			let mut best_block = bc.best_block.write();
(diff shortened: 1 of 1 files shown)


ethcore/src/blockchain/blockchain.rs
```diff
@@ -571,8 +571,8 @@ impl BlockChain {
 
 		{
 			// Fetch best block details
-			let best_block_total_difficulty = bc.block_details(&best_block_hash).unwrap().total_difficulty;
-			let best_block_rlp = bc.block(&best_block_hash).unwrap();
+			let best_block_total_difficulty = bc.block_details(&best_block_hash).expect("Best block is from a known block hash; a known block hash always comes with a known block detail; qed").total_difficulty;
```
ascjones: This might benefit from breaking across lines, though not sure what our convention is here. There is some precedent of having `expect` on the following line in other parts of the codebase.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -536,7 +536,9 @@ impl BlockChain {
 		};
 
 		// load best block
-		let best_block_hash = match bc.db.key_value().get(db::COL_EXTRA, b"best").unwrap() {
+		let best_block_hash = match bc.db.key_value().get(db::COL_EXTRA, b"best")
+			.expect("Low level database error. Some issue with disk?")
```
andresilva: Maybe we could mention that we're trying to get the `best` block? 
"Low-level database error when fetching 'best' block." ?

ethcore/src/blockchain/blockchain.rs
```diff
@@ -586,8 +591,12 @@ impl BlockChain {
 		{
 			let best_block_number = bc.best_block.read().header.number();
 			// Fetch first and best ancient block details
-			let raw_first = bc.db.key_value().get(db::COL_EXTRA, b"first").unwrap().map(|v| v.into_vec());
-			let mut best_ancient = bc.db.key_value().get(db::COL_EXTRA, b"ancient").unwrap().map(|h| H256::from_slice(&h));
+			let raw_first = bc.db.key_value().get(db::COL_EXTRA, b"first")
+				.expect("Low level database error. Some issue with disk?")
```
andresilva: Same as above, we can mention that it's the "first" block and the "best ancient" block below.

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1350,11 +1359,13 @@ impl BlockChain {
 				}
 			},
 			BlockLocation::BranchBecomingCanonChain(ref data) => {
-				let ancestor_number = self.block_number(&data.ancestor).unwrap();
+				let ancestor_number = self.block_number(&data.ancestor)
+					.expect("Inserted block's ancestor number always exists; qed");
```
andresilva: "hash belongs to an ancestor of an inserted block; ancestors of an inserted block are always available; block number of an inserted block is always available; qed"

Actually while writing this I realised that this could fail because the second condition doesn't always hold true (because of warp sync) right?

ethcore/src/blockchain/blockchain.rs
```diff
@@ -1350,11 +1359,13 @@ impl BlockChain {
...
 				let start_number = ancestor_number + 1;
 
 				let mut blooms: Vec<Bloom> = data.enacted.iter()
-					.map(|hash| self.block_header_data(hash).unwrap())
+					.map(|hash| self.block_header_data(hash)
+						 .expect("Inserted block's header data always exists; qed"))
```
andresilva: "hash belongs to an inserted block; block header data of an inserted block is always available; qed"

ethcore/src/blockchain/blockchain.rs
sorpaas: Ah looks like that's indeed. This function is called by `insert_unordered_block` which can get unordered?

sorpaas: It seems that this is fine, although all the logic is not really apparent (so we may consider to do some refactorings for this): In wrap sync, the location is always set to `BlockLocation::CanonChain` so we will never reach this match statement.

ascjones:
Tests failing: `key_server_cluster::cluster::tests::ecdsa_signing_session_completes_if_node_does_not_have_a_share`

sorpaas:
Hmm this is weird. We shouldn't have changed anything in `key_server_cluster`. Even tried to restart the build once. :/

ethcore/src/blockchain/blockchain.rs
andresilva: Thanks for looking this up. Yes, it's not clear from just looking at this code that invariant holds. But it makes sense that we do consider the warped block to be canon (what else can we do if we already trusted the warp?).

PR #9143, opened 2018-07-17
Added --tx-queue-no-early-reject flag to disable early tx queue rejects

ppratscher:
This PR resolves https://github.com/paritytech/parity/issues/9087 by adding a CLI flag to selectively disable a transaction queue optimization that early rejects transactions below the minimal effective gas price, without comparing the sender to the list of local accounts of a node first.

This allows local transactions to always enter the pool, despite it being full, but requires an additional ecrecover operation on every transaction.

### DIFF
--- a/miner/src/pool/verifier.rs
+++ b/miner/src/pool/verifier.rs
@@ -43,8 +43,8 @@ pub struct Options {
 	pub block_gas_limit: U256,
 	/// Maximal gas limit for a single transaction.
 	pub tx_gas_limit: U256,
-	/// Skip check checks
-	pub tx_queue_no_early_reject: bool,
+	/// Skip checks for early rejection, to make sure that local transactions are always imported.
+	pub no_early_reject: bool,
 }
 
 #[cfg(test)]
@@ -54,7 +54,7 @@ impl Default for Options {
 			minimal_gas_price: 0.into(),
 			block_gas_limit: U256::max_value(),
 			tx_gas_limit: U256::max_value(),
-			tx_queue_no_early_reject: false,
+			no_early_reject: false,
 		}
 	}
 }
--- a/miner/src/pool/verifier.rs
+++ b/miner/src/pool/verifier.rs
@@ -43,6 +43,8 @@ pub struct Options {
 	pub block_gas_limit: U256,
 	/// Maximal gas limit for a single transaction.
 	pub tx_gas_limit: U256,
+	/// Skip check checks
+	pub tx_queue_no_early_reject: bool,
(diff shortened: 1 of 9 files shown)


ppratscher:
[clabot:check]

miner/src/pool/verifier.rs
```diff
@@ -43,6 +43,8 @@ pub struct Options {
 	pub block_gas_limit: U256,
 	/// Maximal gas limit for a single transaction.
 	pub tx_gas_limit: U256,
+	/// Skip check checks
+	pub tx_queue_no_early_reject: bool,
```
tomusdrw: Could be named just `no_early_reject`, we are in the context of `tx_queue` already.

Also the comment could be a bit more clear :) Maybe something like:
```
Skip checks for early rejection, to make sure that local transactions are always imported.
```

ppratscher: Sorry, that comment was a leftover type. Will fix it & also fix the failing tests.

ppratscher:
Sorry but I unable to find a test named `min_effective_gas_price`. Which test are you referring to?

tomusdrw:
@ppratscher was thinking about this test: https://github.com/paritytech/parity/blob/21e0cd7781f55f0007d5c7d87929f49513367e27/miner/src/pool/tests/mod.rs#L987-L1023

Should look somewhat like this:

```
#[test]
fn should_not_reject_early_in_case_gas_price_is_less_than_min_effective() {
	// given
	let txq = TransactionQueue::new(
		txpool::Options {
			max_count: 1,
			max_per_sender: 2,
			max_mem_usage: 50
		},
		verifier::Options {
			minimal_gas_price: 1.into(),
			block_gas_limit: 1_000_000.into(),
			tx_gas_limit: 1_000_000.into(),
		},
		PrioritizationStrategy::GasPriceOnly,
	);
	let client = TestClient::new().with_balance(1_000_000_000);
	let tx1 = Tx::gas_price(2).signed().local();

	let res = txq.import(client.clone(), vec![tx1]);
	assert_eq!(res, vec![Ok(())]);
	assert_eq!(txq.status().status.transaction_count, 1);
	assert!(client.was_verification_triggered());

	// when
	let client = TestClient::new();
	let tx1 = Tx::default().signed().local();
	let res = txq.import(client.clone(), vec![tx1]);

	// then
	assert_eq!(res, vec![Ok(())]);
	assert_eq!(txq.status().status.transaction_count, 2);
	assert!(client.was_verification_triggered());
}

PR #9148, opened 2018-07-17
Disable per-sender limit for local transactions.

tomusdrw:
Since it causes confusion for many people.


transaction-pool/src/tests/helpers.rs
```diff
@@ -77,6 +77,9 @@ impl Scoring<Transaction> for DummyScoring {
 			scoring::Choice::RejectNew
 		}
 	}
+	fn should_ignore_sender_limit(&self, _new: &Transaction) -> bool {
```
andresilva: Missing newline.

PR #9151, opened 2018-07-17
handle SyncHandler errors properly

debris:
This pr changes the way malformed rlp is handled by sync module. 

Error handling is now consistent and more strict. If foreign peer sends us malformed rlp, it is always disabled

### DIFF
--- a/ethcore/sync/src/chain/handler.rs
+++ b/ethcore/sync/src/chain/handler.rs
@@ -87,9 +87,26 @@ impl SyncHandler {
 				Ok(())
 			}
 		};
-		result.unwrap_or_else(|e| {
-			debug!(target:"sync", "{} -> Malformed packet {} : {}", peer, packet_id, e);
-		})
+
+		match result {
+			Err(DownloaderImportError::Invalid) => {
+				debug!(target:"sync", "{} -> Malformed packet {}", peer, packet_id);
+				io.disable_peer(peer);
+				sync.deactivate_peer(io, peer);
+				sync.continue_sync(io);
+			},
+			Err(DownloaderImportError::Useless) => {
+				sync.deactivate_peer(io, peer);
+				// give tasks to other peers
+				sync.continue_sync(io);
+			},
+			Ok(()) => {
+				// give a task to the same peer first
+				sync.sync_peer(io, peer, false);
+				// give tasks to other peers
+				sync.continue_sync(io);
+			},
+		}
 	}
 
 	/// Called when peer sends us new consensus packet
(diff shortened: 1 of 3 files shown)


ethcore/sync/src/chain/handler.rs
```diff
@@ -87,9 +87,26 @@ impl SyncHandler {
 				Ok(())
 			}
 		};
-		result.unwrap_or_else(|e| {
-			debug!(target:"sync", "{} -> Malformed packet {} : {}", peer, packet_id, e);
-		})
+
+		match result {
+			Err(DownloaderImportError::Invalid) => {
+				debug!(target:"sync", "{} -> Malformed packet {}", peer, packet_id);
```
svyatonik: Most of times `DownloaderImportError::Invalid` doesn't means malformed rlp, but some other issue which is separately traced (like `Ignored ancient new block hash`, `network id mismatch`, ...). Probably add separate variant for the decoder error (and why not to keep the decoder error like `DownloaderImportError::Decoder(DecoderError)` - isn't it helpful)?

ethcore/sync/src/chain/handler.rs
```diff
@@ -87,9 +87,26 @@ impl SyncHandler {
...
+				io.disable_peer(peer);
+				sync.deactivate_peer(io, peer);
+				sync.continue_sync(io);
```
ordian: nit: maybe move `continue_sync` out of the match, since it's repeated 3 times?

ethcore/sync/src/chain/handler.rs
```diff
@@ -87,9 +87,26 @@ impl SyncHandler {
...
+			},
+			Err(DownloaderImportError::Useless) => {
+				sync.deactivate_peer(io, peer);
+				// give tasks to other peers
+				sync.continue_sync(io);
+			},
+			Ok(()) => {
+				// give a task to the same peer first
+				sync.sync_peer(io, peer, false);
```
svyatonik: There are some methods where `sync_peer` is still called inside (both with force `true` and `false`). I guess this will be a reason for some extra `Skipping busy peer` messages in the log (since now it is called twice - inside methods and in this match). Not an issue, though :)

ethcore/sync/src/chain/handler.rs
debris: I don't think that `DecoderError` is useful here. There is nothing nothing useful we can do with this information.

And maybe, I'll just improve the error message, since the meaning of `Invalid` is broader than 'Malformed packet'. Would this be good enough?

ethcore/sync/src/chain/handler.rs
debris: Thank's for pointing out. I'll remove all remaining `sync.sync_peer(io, peer, false)` :)

ethcore/sync/src/chain/handler.rs
svyatonik: Sure :)

PR #9160, opened 2018-07-18
ci: update version strings for snaps

5chdn:


General-Beck:
everything is correct

PR #9161, opened 2018-07-18
deserialize block only once during verification

debris:
first pull request that tries to limit number of times the block is being deserialized. 

before:
- uncles were deserialized 3 times during verification

now:
- uncles are deserialized only once

### DIFF
--- a/ethcore/src/verification/verification.rs
+++ b/ethcore/src/verification/verification.rs
@@ -25,7 +25,6 @@ use std::collections::HashSet;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use bytes::Bytes;
-use ethereum_types::H256;
 use hash::keccak;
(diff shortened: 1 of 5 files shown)


ethcore/src/verification/verification.rs
```diff
@@ -59,63 +60,66 @@ impl HeapSizeOf for PreverifiedBlock {
...
+		None
+	};
+
+	let transactions = block.transactions
+		.into_iter()
+		.map(|t| {
 			let t = engine.verify_transaction_unordered(t, &header)?;
 			if let Some(max_nonce) = nonce_cap {
 				if t.nonce >= max_nonce {
 					return Err(BlockError::TooManyTransactions(t.sender()).into());
```
ordian: Any reasons why we have `return Err(....into())` in some places and `bail!(...)` in other ones apart from historical?

PR #9164, opened 2018-07-18
Be more graceful on Aura difficulty validation

sorpaas:


debris:
I'm not super familiar with an Aura Engine, but the code looks good

PR #9179, opened 2018-07-21
ethcore: add builtin benchmarks based on geth

andresilva:
This PR implements the builtins benchmarks using the same benchmarking strategy and test cases as in geth (https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts_test.go). This should help an on-going analysis on builtin pricing (https://gist.github.com/holiman/1b69d6361f490f71617c4e08dc125c07).

cc @holiman


holiman:
Works like a charm for me, thanks!

PR #9186, opened 2018-07-22
Better support for eth_getLogs in light mode

### LINKED ISSUE #9184: eth_getLogs RPC is unreliable in light mode
> I'm running:
>
> - **How installed?**:  from source

The eth_getLogs RPC only works for recent blocks in light mode. Specifically, it only works if logs are created on blocks where the headers are stored. If the headers are pruned, the client does not fetch them from the network on-demand to scan for logs.

I noticed this when using Augur and opened an [issue there](https://github.com/AugurProject/augur-app/issues/161). This [gist](https://gist.github.com/epheph/e9b0d5f033cfe9f71e5c111a3567a1ee) provides a test case.

While it could be too expensive to support `eth_getLogs` on a large range of headers, I think it makes sense for the light client to fetch small ranges of headers on demand from the network and scan them for matching logs.

Furthermore, the log entries responses are incomplete since `BlockReceipts` responds with a `Receipt` instead of a `RichReceipt`. `RichReceipt` includes the transaction hash, index, cumulative gas, etc., which are expected in the RPC response. Fixing this would require a modification to the PIP protocol, I assume.

jimpo:
Fixes #9184.

If the light client receives an `eth_getLogs` request for blocks that it does not have, it will fetch the headers from the network on demand rather than silently failing and returning 0 logs.

There is a limit of a 1,000 block range on requests to ensure they can be answered quickly.

Tested manually on mainnet because I don't see other automated tests for RPCs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,6 +2135,7 @@ dependencies = [
  "ethstore 0.2.0",
  "fake-fetch 0.0.1",
  "fake-hardware-wallet 0.0.1",
+ "fastmap 0.1.0",
  "fetch 0.1.0",
  "futures 0.1.21 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures-cpupool 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/ethcore/light/src/on_demand/request.rs
+++ b/ethcore/light/src/on_demand/request.rs
@@ -1191,46 +1191,53 @@ mod tests {
 
 		let cache = Mutex::new(make_cache());
 
+		let header_with_ancestors = |hash, count| {
+			HeaderWithAncestors {
+				block_hash: hash,
+				ancestor_count: count
+			}
+		};
+
 		// Correct responses
-		assert!(HeaderWithAncestors(headers[0].hash().into(), 0)
+		assert!(header_with_ancestors(headers[0].hash().into(), 0)
 				.check_response(&cache, &headers[0].hash().into(), &raw_headers[0..1]).is_ok());
-		assert!(HeaderWithAncestors(headers[0].hash().into(), 2)
+		assert!(header_with_ancestors(headers[0].hash().into(), 2)
 				.check_response(&cache, &headers[0].hash().into(), &raw_headers[0..3]).is_ok());
-		assert!(HeaderWithAncestors(headers[0].hash().into(), 10)
+		assert!(header_with_ancestors(headers[0].hash().into(), 10)
 				.check_response(&cache, &headers[0].hash().into(), &raw_headers[0..11]).is_ok());
-		assert!(HeaderWithAncestors(headers[2].hash().into(), 2)
+		assert!(header_with_ancestors(headers[2].hash().into(), 2)
 				.check_response(&cache, &headers[2].hash().into(), &raw_headers[2..5]).is_ok());
-		assert!(HeaderWithAncestors(headers[2].hash().into(), 10)
+		assert!(header_with_ancestors(headers[2].hash().into(), 10)
 				.check_response(&cache, &headers[2].hash().into(), &raw_headers[2..11]).is_ok());
-		assert!(HeaderWithAncestors(invalid_successor.hash().into(), 0)
+		assert!(header_with_ancestors(invalid_successor.hash().into(), 0)
 				.check_response(&cache, &invalid_successor.hash().into(), &[raw_invalid_successor.clone()]).is_ok());
 
 		// Incorrect responses
-		assert_eq!(HeaderWithAncestors(invalid_successor.hash().into(), 0)
+		assert_eq!(header_with_ancestors(invalid_successor.hash().into(), 0)
 				   .check_response(&cache, &headers[0].hash().into(), &raw_headers[0..1]),
 				   Err(Error::WrongHash(invalid_successor.hash(), headers[0].hash())));
-		assert_eq!(HeaderWithAncestors(headers[0].hash().into(), 0)
+		assert_eq!(header_with_ancestors(headers[0].hash().into(), 0)
 				   .check_response(&cache, &headers[0].hash().into(), &[]),
 				   Err(Error::Empty));
-		assert_eq!(HeaderWithAncestors(headers[0].hash().into(), 10)
+		assert_eq!(header_with_ancestors(headers[0].hash().into(), 10)
 				   .check_response(&cache, &headers[0].hash().into(), &raw_headers[0..10]),
 				   Err(Error::TooFewResults(11, 10)));
-		assert_eq!(HeaderWithAncestors(headers[0].hash().into(), 9)
+		assert_eq!(header_with_ancestors(headers[0].hash().into(), 9)
 				   .check_response(&cache, &headers[0].hash().into(), &raw_headers[0..11]),
 				   Err(Error::TooManyResults(10, 11)));
 
 		let response = &[raw_headers[0].clone(), raw_headers[2].clone()];
-		assert_eq!(HeaderWithAncestors(headers[0].hash().into(), 1)
+		assert_eq!(header_with_ancestors(headers[0].hash().into(), 1)
 				   .check_response(&cache, &headers[0].hash().into(), response),
 				   Err(Error::WrongHeaderSequence));
 
 		let response = &[raw_invalid_successor.clone(), raw_headers[0].clone()];
-		assert_eq!(HeaderWithAncestors(invalid_successor.hash().into(), 1)
+		assert_eq!(header_with_ancestors(invalid_successor.hash().into(), 1)
 				   .check_response(&cache, &invalid_successor.hash().into(), response),
 				   Err(Error::WrongHeaderSequence));
 
 		let response = &[raw_invalid_successor.clone(), raw_headers[1].clone()];
-		assert_eq!(HeaderWithAncestors(invalid_successor.hash().into(), 1)
+		assert_eq!(header_with_ancestors(invalid_successor.hash().into(), 1)
 				   .check_response(&cache, &invalid_successor.hash().into(), response),
 				   Err(Error::WrongHeaderSequence));
 	}
--- a/ethcore/light/src/on_demand/request.rs
+++ b/ethcore/light/src/on_demand/request.rs
@@ -423,8 +423,9 @@ impl CheckedRequest {
 				if let Some(&net_request::HashOrNumber::Hash(start)) = req.start.as_ref() {
 					let mut result = Vec::with_capacity(req.max as usize);
 					let mut hash = start;
+					let mut cache = cache.lock();
 					for _ in 0..req.max {
-						match cache.lock().block_header(&hash) {
+						match cache.block_header(&hash) {
 							Some(header) => {
 								hash = header.parent_hash();
 								result.push(header);
--- a/ethcore/light/src/on_demand/request.rs
+++ b/ethcore/light/src/on_demand/request.rs
@@ -16,6 +16,7 @@
 
 //! Request types, verification, and verification errors.
 
+use std::cmp;
 use std::sync::Arc;
 
 use bytes::Bytes;
@@ -47,6 +48,8 @@ pub enum Request {
 	HeaderProof(HeaderProof),
 	/// A request for a header by hash.
 	HeaderByHash(HeaderByHash),
+	/// A request for a header by hash with a range of its ancestors.
+	HeaderWithAncestors(HeaderWithAncestors),
 	/// A request for the index of a transaction.
 	TransactionIndex(TransactionIndex),
 	/// A request for block receipts.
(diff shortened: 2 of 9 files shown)


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -438,6 +435,154 @@ impl LightFetch {
...
+				let cht_root = match cht_root {
+					Some(cht_root) => cht_root,
+					None => return Either::A(future::err(errors::unknown_block())),
+				};
+
+				let header_proof = request::HeaderProof::new(to_number, cht_root)
+					.expect("HeaderProof::new is Some(_) if cht::block_to_cht_number() is Some(_); \
+							 this would return above if block_to_cht_number returned None; qed");
+				reqs.push(header_proof.into());
+				Field::back_ref(reqs.len() - 1, 0)
```
ordian: I would prefer
```rust
let idx = reqs.len();
let hash_ref = Field::back_ref(idx, 0);
reqs.push(header_proof.into());
```
Because it is easier to reason about overflows that way and it matches the style in other places.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -438,6 +435,154 @@ impl LightFetch {
...
+				match (last_hash, from_block) {
+					(Some(h1), BlockId::Hash(h2)) if h1 != h2 => Vec::new(),
+					_ => headers,
+				}
+			}))
+		})
+	}
+
+	fn headers_by_hash(&self, hashes: Vec<H256>) -> impl Future<Item = HashMap<H256, encoded::Header>, Error = Error> {
+		let mut refs = HashMap::with_capacity(hashes.len());
```
ordian: Maybe we could use [`H256FastMap`](https://github.com/paritytech/parity-common/blob/4481687a81f04993e5f54adef27852167d7114f7/plain_hasher/src/lib.rs#L27) instead?

ethcore/light/src/on_demand/request.rs
```diff
@@ -625,6 +657,10 @@ pub enum Error {
 	Decoder(::rlp::DecoderError),
 	/// Empty response.
 	Empty,
+	/// Response data length exceeds request max.
+	TooManyResults(u64, u64),
```
ordian: For future (doesn't need to be fixed in this PR): I would prefer having a struct-like enum variants, since it's not always clear w/o some context, what is the meaning and the order of the fields.

ethcore/light/src/provider.rs
```diff
@@ -33,6 +33,9 @@ use transaction_queue::TransactionQueue;
 
 use request;
 
+/// Maximum allowed size of a headers request.
+pub const MAX_HEADERS_LENGTH: u64 = 512;
```
ordian: I personally find the previous name (`MAX_HEADERS_TO_SEND`) more descriptive and less confusing than this one.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -438,6 +435,154 @@ impl LightFetch {
...
+			let to_header_hint = match to_block {
+				BlockId::Hash(ref h) => header_map.remove(h),
+				_ => None,
+			};
+			let headers_fut = fetcher.headers_range(from_block_num, to_block_num, to_header_hint);
+			Either::B(headers_fut.map(move |headers| {
+				// Validate from_block if it's a hash
+				let last_hash = headers.last().map(|hdr| hdr.hash());
+				match (last_hash, from_block) {
+					(Some(h1), BlockId::Hash(h2)) if h1 != h2 => Vec::new(),
```
ordian: We should return an error in this case (`Either::A`), right?

folsen:
> I'm not totally sure this is a good idea since it can only support small-ish ranges of blocks

This is something we've actually discussed internally, fetching logs for very large range of blocks is common behavior in a lot of DApps, we'd need to make sure this doesn't open up an "attack" vector by accidentally spamming those requests, or rather make sure that it fails in some suitably graceful way.

What would happen right now if a request was sent for a large range?

jimpo:
@folsen Right now it just takes a really long time then times out on the client side. This should not get merged without some sort of limit on the range length at least.

A more sophisticated approach could be to have two limits, one on the number of blocks within the past 2048 (CHT size) and one on the number of blocks before that. Or to weight them differently or something. But that's probably just an unnecessarily confusing UX.

ethcore/light/src/provider.rs
jimpo: The reason I changed it is that the variable was used in a context that was receiving headers, not sending them, so it seemed confusing. What about `MAX_HEADERS_PER_REQUEST`?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -438,6 +435,154 @@ impl LightFetch {
...
+		let mut reqs = Vec::with_capacity(hashes.len());
+
+		for hash in hashes {
+			refs.entry(hash).or_insert_with(|| {
+				if let Some(header) = self.client.block_header(BlockId::Hash(hash)) {
+					header.into()
+				} else {
+					reqs.push(request::HeaderByHash(hash.into()).into());
+					HeaderRef::Unresolved(reqs.len() - 1, hash.into())
+				}
```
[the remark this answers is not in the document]
jimpo: It seems kind of roundabout to wrap a value in an enum then unwrap the result because we know how that branch is handled in a helper function. I figured it was preferable here to avoid the `expect` and accept some code duplication here, but I'll change it if you really prefer.

As for the deduping, this method does not require the input hash vector to be sorted (or of length <= 2), so the `or_insert_with` is the easiest way to dedup IMO, rather than doing a sort first.

rpc/src/v1/helpers/light_fetch.rs
jimpo: The full node ethcore `Client` appears to just return an empty `Vec`, so I kept the same behavior here. https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/client/client.rs#L1874

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
ordian: Even though I agree that [sometimes](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) duplication is better, IMHO in this case using `make_header_requests` makes code more readable and DRY (e.g. we can change the logic in one place). As for possible performance impact, I think if `make_header_requests` is inlined, then the impact should be [negligible](https://godbolt.org/g/fLTaeS).

As for deduping, since this is a private method, we can control the requirements. But I'm fine with leaving it as is.

ethcore/light/src/provider.rs
ordian: I like `MAX_HEADERS_PER_REQUEST` much better than `MAX_HEADERS_LENGTH`.

rpc/src/v1/helpers/light_fetch.rs
ordian: Good point. There has been some discussion about changing the default behavior [here](https://github.com/paritytech/parity-ethereum/pull/9256#issuecomment-409221617):
>  user requests logs for some block and they get back an empty array they cannot differentiate between "the block you requested has no logs" and "we don't know about the block you requested"

 but I guess this particular case will remain the same.

rpc/src/v1/helpers/errors.rs
```diff
@@ -100,6 +100,14 @@ pub fn request_rejected_limit() -> Error {
 	}
 }
 
+pub fn request_rejected_param_limit() -> Error {
+	Error {
+		code: ErrorCode::ServerError(codes::REQUEST_REJECTED_LIMIT),
+		message: "Request has been rejected because requested data size exceeds limit.".into(),
```
ordian: How a user can find out the value of the limit?

jimpo: I'll put it in the error message.

Cargo.lock
```diff
@@ -2159,6 +2159,7 @@ dependencies = [
  "parity-version 2.1.0",
  "parking_lot 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "patricia-trie 0.2.1 (git+https://github.com/paritytech/parity-common)",
+ "plain_hasher 0.1.0 (git+https://github.com/paritytech/parity-common)",
```
ordian: `H256FastMap` has moved to `util/fastmap` (#9307).

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -478,6 +485,8 @@ impl LightFetch {
 			if to_block_num < from_block_num {
 				// early exit for "to" block before "from" block.
 				return Either::A(future::err(errors::filter_block_not_found(to_block)));
```
ordian: cc #9373

ethcore/light/src/on_demand/request.rs
```diff
@@ -701,6 +739,56 @@ impl HeaderProof {
 	}
 }
 
+/// Request for a header by hash with a range of ancestors.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HeaderWithAncestors(pub Field<H256>, pub u64);
```
niklasad1: I would prefer a `normal struct` instead of a `tuple struct` here because it will make easier to read/understand!

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -312,78 +314,67 @@ impl LightFetch {
...
-				.take_while(|ref hdr| hdr.number() != from_block_number)
-				.chain(Some(from_block_header))
-				.filter(|ref hdr| {
-					let hdr_bloom = hdr.log_bloom();
-					bit_combos.iter().any(|bloom| hdr_bloom.contains_bloom(bloom))
-				})
-				.map(|hdr| (hdr.number(), hdr.hash(), request::BlockReceipts(hdr.into())))
-				.map(|(num, hash, req)| self.on_demand.request(ctx, req).expect(NO_INVALID_BACK_REFS).map(move |x| (num, hash, x)))
-				.collect();
+		let fetcher: Self = self.clone();
```
niklasad1: let fetcher **~~: Self~~** = self.clone();

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -461,6 +452,147 @@ impl LightFetch {
...
+		let fetch_hashes = [from_block, to_block].iter()
+			.filter_map(|block_id| match block_id {
+				BlockId::Hash(hash) => Some(hash.clone()),
+				_ => None,
+			})
+			.collect::<Vec<_>>();
+
+		let best_number = self.client.chain_info().best_block_number;
+
+		let fetcher: Self = self.clone();
```
niklasad1: let fetcher ~~: Self~~ = self.clone();

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -461,6 +452,147 @@ impl LightFetch {
...
+				let last_hash = headers.last().map(|hdr| hdr.hash());
+				match (last_hash, from_block) {
+					(Some(h1), BlockId::Hash(h2)) if h1 != h2 => Vec::new(),
+					_ => headers,
+				}
+			}))
+		})
+	}
+
+	fn headers_by_hash(&self, hashes: Vec<H256>) -> impl Future<Item = H256FastMap<encoded::Header>, Error = Error> {
```
niklasad1: Take ```hashes``` by reference here instead because `H256` is ```Copy```?!

niklasad1:
Needs to be rebased/merged into master!

ethcore/light/src/on_demand/request.rs
```diff
@@ -390,6 +405,7 @@ impl CheckedRequest {
 
 				None
 			}
+			// TODO: CheckedRequest::HeaderWithAncestors arm
```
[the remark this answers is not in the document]
jimpo: Added caching in 586f46ccc04e108bf216c696ed995e3a59da7ea0.

ethcore/light/src/on_demand/request.rs
```diff
@@ -399,6 +415,26 @@ impl CheckedRequest {
...
+			CheckedRequest::HeaderWithAncestors(_, ref req) => {
+				if req.skip != 1 || !req.reverse {
+					return None;
+				}
+
+				if let Some(&net_request::HashOrNumber::Hash(start)) = req.start.as_ref() {
+					let mut result = Vec::with_capacity(req.max as usize);
+					let mut hash = start;
+					for _ in 0..req.max {
+						match cache.lock().block_header(&hash) {
```
ordian: I would rather move `cache.lock()` outside of the loop.

5chdn:
Cool. Needs a 2nd review.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -299,78 +301,67 @@ impl LightFetch {
...
-									log_type: "mined".into(),
-									removed: false,
-								});
+				let on_demand = &fetcher.on_demand;
+
+				let maybe_future = fetcher.sync.with_context(move |ctx| {
+					// find all headers which match the filter, and fetch the receipts for each one.
+					// match them with their numbers for easy sorting later.
+					let bit_combos = filter.bloom_possibilities();
+					let receipts_futures: Vec<_> = headers.drain(..)
```
niklasad1: A minor thing, isn't better to use `into_iter` instead of `drain` in this context because we are not reusing the `headers` after the operation?

(`drain` is borrowing the vector (ends up with an empty vector) vs `into_iter` is taking ownership)

PR #9187, opened 2018-07-22
ethash: implement EIP-1234

5chdn:
https://github.com/ethereum/EIPs/pull/1234


eosclassicteam:
@5chdn I think EIP-1234 is in draft state, shoudn't we implement this after the EIP is being active??

5chdn:
That's why there is an "on ice" label. :roll_eyes:

5chdn:
Reopening as per https://github.com/ethereum/pm/issues/55

PR #9194, opened 2018-07-23
calculate gas_used of the transaction trace correctly, 

### LINKED ISSUE #6964: Gas used got by trace_transaction
> I'm running:
>
> - **Parity version**: 1.8.2
> - **Operating system**: Windows
> - **And installed**: via installer

I try to get gas used of a transaction. The gas used got by eth_getTransactionReceipt is 63082 and got by trace_transaction is as below. The sum of gas used of traces is not equal to 63082. 
And all value transfer have gasUsed 0.
Is the gas used of traces just an estimate value?
Can anyone share knowledge of implementation details of this?

Gas used of traces:
```
call contract1                         41810
  ->contract1 call contract2           15149
    ->contract2 transfer 1             0
    ->contract2 transfer 2             0
  ->contract1 transfer                 0
  ->contract1 call contract3           15149
    ->contract3 transfer 1             0
    ->contract3 transfer 2             0
```

### LINKED ISSUE #9178: Gas used according to transaction trace doesn't match transaction receipt
> I'm running:
>
> - **Which Parity version?**: 1.10.4
> - **Which operating system?**: MacOS 

It appears that the amount of used gas according to the transaction traces (obtained by the RPC call `trace_block`) doesn't match the amount according to the transaction receipt (obtained by the RPC call `eth_getTransactionReceipt`).

Here is one concrete example, for the transaction hash `0xcbaa1b514a4565003fa0708a21cf0ddb115ab20d2c9373ffda0c129d1deaeebc`:

transaction trace:
```
{
  'action': {
   'callType': 'call',
   'from': '0xe09c68e61998d9c81b14e4ee802ba7adf6d74cdb',
   'gas': '0x6ee4',
   'input': '0x454e32413658483230',
   'to': '0x2910543af39aba0cd09dbb2d50200b3e800a63d2',
   'value': '0xde0b6b3a7640000'
  },
  'blockHash': '0xc77c370a5e4f12e8e0b20e8435132a04d474a02323c86077d678cbaba2af5905',
  'blockNumber': 47505,
  'result': {'gasUsed': '0x0', 'output': '0x'},
  'subtraces': 0,
  'traceAddress': [],
  'transactionHash': '0xcbaa1b514a4565003fa0708a21cf0ddb115ab20d2c9373ffda0c129d1deaeebc',
  'transactionPosition': 0,
  'type': 'call'
}
```

transaction receipt:
```
{
 'blockHash': '0xc77c370a5e4f12e8e0b20e8435132a04d474a02323c86077d678cbaba2af5905',
 'blockNumber': '0xb991',
 'contractAddress': None,
 'cumulativeGasUsed': '0x546c',
 'gasUsed': '0x546c',
 'logs': [],
 'logsBloom': '...',
 'root': '0x22a37344703dea06565d297e334469810a512ceded1617681df783ee1e1cb9f2',
 'status': None,
 'transactionHash': '0xcbaa1b514a4565003fa0708a21cf0ddb115ab20d2c9373ffda0c129d1deaeebc',
 'transactionIndex': '0x0'
}
```

According to the transaction receipt the amount of used gas is `0x546c` or `21612`. However the transaction trace contains the amount `gas = 0x6ee4` or `28388`, which is **larger** then the value in the transaction receipt. There is also the value `result.gasUsed = 0` which is, obviously, also not correct.

debris:
- fixes #9178 
- fixes #6964

description of the issue: https://github.com/paritytech/parity-ethereum/issues/9178#issuecomment-406897257


ethcore/vm/src/action_params.rs
```diff
@@ -76,7 +76,9 @@ pub struct ActionParams {
 	pub sender: Address,
 	/// Transaction initiator.
 	pub origin: Address,
-	/// Gas paid up front for transaction execution
+	/// Gas paid up front for entire transaction execution.
+	pub entire_gas: U256,
+	/// Gas paid up front for action execution.
```
andresilva: `Gas paid up front for action execution (entire gas minus action opcode and data cost).` ?

sorpaas:
I don't quite understand the meaning of `entire_gas` here:

* For top-level call/create, it makes sense -- those are invoked by a transaction, and it indeed has the concept of transaction "entire gas".
* However, for sub call/create, I think this concept might not have much meaning? It's currently always set to equal to just `gas`, and remain unused except in tracing.

Given that this `entire_gas` value is only for tracing, would it be possible to move it out of `ActionParams` struct?

sorpaas:
Another method to fix #9178 and #6964 is to directly insert the transaction gas to tracing mod. I think that may be the better way, because the current gas used is actually correct if we consider a transaction is a wrapper on a call/create executive.

Also a note, this PR is slightly conflicts with #9360.

5chdn:
What's the status of this?

5chdn:
Please reopen once you continue working on this.

PR #9203, opened 2018-07-23
Double up the number of possible sessions.

tomusdrw:
Allows to actually get more than ~900 peers.


tomusdrw:
Not much, I managed to get up to 1.1k for now (2k max). I guess that's already max for that particular node anyway.

PR #9211, opened 2018-07-24
 Remove pass-by-reference return data value from executive 

sorpaas:
rel #6744 

Return data is always available from `FinalizationResult`, so using pass-by-reference in `Executive` parameters doesn't give us any performance benefits. This reference also creates some trouble for resumable `Executive`.

This PR changes it so that output is read from `FinalizationResult::return_data`. An `unsafe` in `evm` crate was also able to be removed due to this change.


sorpaas:
Previously we trace call output via `cmp::min(parent_output.len(), actual_output.len())`, but because we don't know anything about `parent_output` in `executive` right now, so this needs to be changed to just `actual_output.len()`. One test is affected. Don't know whether there're better ways to handle this.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -409,20 +409,25 @@ impl<Cost: CostType> Interpreter<Cost> {
...
+					let input = self.mem.read_slice(in_off, in_size);
+					ext.call(&call_gas.as_u256(), sender_address, receive_address, value, input, &code_address, call_type)
 				};
 
+				let output = self.mem.writeable_slice(out_off, out_size);
+
 				return match call_result {
 					MessageCallResult::Success(gas_left, data) => {
+						let len = cmp::min(output.len(), data.len());
+						(&mut output[..len]).copy_from_slice(&data[..len]);
```
pepyakin: Wait, isn't this a copy that the former code tried to avoid?

ethcore/src/executive.rs
```diff
@@ -431,7 +438,6 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 					Err(evm_err)
 				} else {
 					self.state.discard_checkpoint();
-					output.write(0, &builtin_out_buffer);
```
debris: -1 copy 👍

PR #9220, opened 2018-07-26
Removes duplicate libudev-dev from Dockerfile

JohnnySheffield:
I've noticed that libudev-dev is installed twice in the hub/Dockerfile. This pull request just removes the duplicate.


JohnnySheffield:
[clabot:check].

PR #9232, opened 2018-07-27
fix typo

dongz9:
When a connection is closed, token should be `StreamToken` instead of `TimerToken`.


ascjones:
@dongz9 please sing the CLA (see above) and we can merge this.

dongz9:
[clabot:check]

PR #9234, opened 2018-07-27
Implement EIP-1052 (EXTCODEHASH) and fix several issues in state account cache

sorpaas:
* Previously, if a cache operation on an account failed, then we print an error message, but silently ignore the error, and act as if the code on that account is empty. This PR fixes that issue to properly return `TrieError`.
* In state interface, allow the query to code to return `Option`. This makes it possible to distinguish account that does not exist, and account that does not have code.
* Implement EIP-1052 EXTCODEHASH.

### DIFF
--- a/ethcore/evm/src/interpreter/mod.rs
+++ b/ethcore/evm/src/interpreter/mod.rs
@@ -574,7 +574,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 			},
 			instructions::EXTCODEHASH => {
 				let address = u256_to_address(&stack.pop_back());
-				let hash = ext.extcodehash(&address)?.unwrap_or_else(H256::default);
+				let hash = ext.extcodehash(&address)?.unwrap_or_else(H256::zero);
 				stack.push(U256::from(hash));
 			},
 			instructions::CALLDATACOPY => {
@@ -595,14 +595,12 @@ impl<Cost: CostType> Interpreter<Cost> {
 				Self::copy_data_to_memory(&mut self.mem, stack, params.code.as_ref().map_or_else(|| &[] as &[u8], |c| &**c as &[u8]));
 			},
 			instructions::EXTCODECOPY => {
-				use std::ops::Deref;
-
 				let address = u256_to_address(&stack.pop_back());
 				let code = ext.extcode(&address)?;
 				Self::copy_data_to_memory(
 					&mut self.mem,
 					stack,
-					code.as_ref().map(|c| &c.deref()[..]).unwrap_or(&[])
+					code.as_ref().map(|c| &(*c)[..]).unwrap_or(&[])
 				);
 			},
 			instructions::GASPRICE => {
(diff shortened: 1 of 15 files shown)


ethcore/evm/src/interpreter/mod.rs
```diff
@@ -590,7 +596,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 			},
 			instructions::EXTCODECOPY => {
 				let address = u256_to_address(&stack.pop_back());
-				let code = ext.extcode(&address)?;
+				let code = ext.extcode(&address)?.unwrap_or(Arc::new(Vec::new()));
```
debris: maybe `match` over here instead of `unwrap_or` ? I believe that we could avoid `Arc` and `Vec` allocations

ethcore/src/state/mod.rs
```diff
@@ -950,19 +957,27 @@ impl<B: Backend> State<B> {
...
+					return Err(Box::new(TrieError::IncompleteDatabase(H256::from(a))));
+				}
 			}
-			f(acc.map(|a| &*a))
-		});
+			Ok(f(acc.map(|a| &*a)))
+		}) {
+			Some(r) => Some(r?),
+			None => None,
+		};
```
debris: can you merge this match with the next match?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -589,9 +595,15 @@ impl<Cost: CostType> Interpreter<Cost> {
...
 			instructions::EXTCODECOPY => {
+				use std::ops::Deref;
+
 				let address = u256_to_address(&stack.pop_back());
 				let code = ext.extcode(&address)?;
-				Self::copy_data_to_memory(&mut self.mem, stack, &code);
+				Self::copy_data_to_memory(
+					&mut self.mem,
+					stack,
+					code.as_ref().map(|c| &c.deref()[..]).unwrap_or(&[])
```
andresilva: I think `code.as_ref().map(|c| &(*c)[..]).unwrap_or(&[])` should compile. Not sure if that's preferable though.

ethcore/src/state/mod.rs
```diff
@@ -911,31 +911,38 @@ impl<B: Backend> State<B> {
 		Ok(pod_state::diff_pod(&pod_state_pre, &pod_state_post))
 	}
 
-	// load required account data from the databases.
-	fn update_account_cache(require: RequireCache, account: &mut Account, state_db: &B, db: &HashDB<KeccakHasher>) {
+	/// Load required account data from the databases. Returns whether the cache succeeds.
+	#[must_use]
```
andresilva: :+1: We should probably use this more.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -568,9 +569,14 @@ impl<Cost: CostType> Interpreter<Cost> {
 			},
 			instructions::EXTCODESIZE => {
 				let address = u256_to_address(&stack.pop_back());
-				let len = ext.extcodesize(&address)?;
+				let len = ext.extcodesize(&address)?.unwrap_or(0);
 				stack.push(U256::from(len));
 			},
+			instructions::EXTCODEHASH => {
+				let address = u256_to_address(&stack.pop_back());
+				let hash = ext.extcodehash(&address)?.unwrap_or_else(H256::default);
```
andresilva: Prefer `H256::zero` to make it more explicit?

sorpaas:
@andresilva Yeah, I think testing this with jsontests would make most sense.

PR #9236, opened 2018-07-27
Ensure that we always call trace_call for builtins

udoprog:
I'm trying to implement a tracer for a simple debugger based on parity, but noticed that a contract where I'm calling `ecrecover` there is no `trace_call` to determine when a call is done which I need to maintain a call stack.

### DIFF
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -440,19 +440,23 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 						ActionValue::Transfer(value) => value != U256::zero(),
 						ActionValue::Apparent(_) => false,
 					};
-					if self.depth == 0 || is_transferred {
+
+					let trace_output = if self.depth == 0 || is_transferred {
 						let mut trace_output = tracer.prepare_trace_output();
 						if let Some(out) = trace_output.as_mut() {
 							*out = output.to_owned();
 						}
+						trace_output
+					} else {
+						None
+					};
 
-						tracer.trace_call(
-							trace_info,
-							cost,
-							trace_output,
-							vec![]
-						);
-					}
+					tracer.trace_call(
+						trace_info,
+						cost,
+						trace_output,
+						vec![]
+					);
 
 					let out_len = builtin_out_buffer.len();
 					Ok(FinalizationResult {


udoprog:
Issue with GitLab:
```
________Running Parity Full Test Suite________
 Downloading threadpool v1.7.1
error: unable to get packages from source

Caused by:
  No space left on device (os error 28)
```

sorpaas:
@udoprog I don't know whether we would want this. The issue is that some precompiled contracts are used quite often (like `IDENTITY`, which is used for `memcpy`). If we trace all of them, it creates really heavy workload for the client, with really small gas costs. So the previous solution was to only trace two kinds of precompiled calls:

* Trace top-level calls.
* Trace noticeable sub-level calls (i.e. any sub-level calls with non-zero value transfers).

You can find more on this at #8486.

sorpaas:
@udoprog Would there be any better ways to solve your use cases? You mentioned that you need to maintain a call stack. Can you explain more how/why you need that?

udoprog:
@sorpaas Sure. Below is a more detailed explanation of what I do.

As I enter a call, I lookup 1) the source map, and 2) the AST for the destination of that call and push it on a stack.
As instructions are processed I use the source map to locate AST constructs which matches in order to decode the value of expressions for debugging.

If the call finishes, I pop the stack so that the correct stack element is maintained on top.
If the call fails, I pop this stack to be able to pinpoint where the failure happened.

I rely on `prepare_trace_call` for pushing to the stack, since that is called before any instructions in the destination are executed, and `trace_call`/`trace_failed_call` for popping the stack. The imbalance caused by certain builtins throws this off.

In order to do the instruction processing, this call stack is shared (through a `Mutex`) with a VMTracer implementation. That is why it's important that it is maintained at exactly these points, since the top of the stack needs to be set correctly as instructions are processed.

I've tried using `subtracer` to push, and `drain` to pop. But [`prepare_trace_call` is called _before_ these](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/executive.rs#L473), so the interaction becomes messy. If `trace_call` was called on the subtracer instead it would work.

udoprog:
I pushed the implementation so you can take a look or make suggestions:
https://github.com/PrimaBlock/parables/blob/debugging/testing/trace.rs#L401

I realized I deal with this right now by checking `params.code_address == Address::from(0x1)`, if there was a more reliable way to test it, that would work as well:
https://github.com/PrimaBlock/parables/blob/debugging/testing/trace.rs#L407

Issue #8565, opened 2018-05-08
Warp Sync: Alternative snapshot formats


rphmeier:
We want to investigate new snapshot formats which are better for the following properties:
  - Generation of (partial) snapshots on nodes
  - Reusability of chunks between snapshots to make finding them easier
  - Trustworthiness of data

Snapshot chunks are currently divided into two categories:
  - state chunks encode the entire account state of the blockchain at some block
  - security chunks provide corroboration for whether that block is valid without syncing the whole chain

W.r.t. different consensus engines, the "security" chunks will look completely different but will usually contain reusable data. For example, validator-set based consensus systems prove security with a series of bootstrapped handoffs (as long as we relax the weak subjectivity model to assume that old validator sets remain uncorrupted). All finalized handoffs can be reused, although usually their proof is small enough that all the handoffs can fit in a single chunk. Depending on the state churn and snapshot distance we may also be able to reuse some state chunks. 

A keystone/delta model where we have intermittent "full" state snapshots every N*K blocks and the snapshots between them (every K blocks) only store deltas over that state is one possibility.

One major problem with the current snapshot system is that it is too heavy for most nodes to produce a snapshot before they prune the state that it encodes from their database. State chunks are currently very tightly packed using a method that makes it impossible to determine which account a chunk starts at or the exact data of the account entry without having produced all the chunks before. One possibility is to design predictable scheme for the boundaries of chunks will allow nodes to produce some of the state chunks but not all.

We can augment this scheme with random sampling: nodes which don't produce full snapshots will randomly sample some accounts and produce account entries for them, which they will keep on disk. They will refuse to propagate any snapshot where their random sample doesn't match the data in the snapshot. Assuming all repropagating nodes have their own random sample and a sufficiently large network, this makes it very unlikely for bad snapshot data to make its way through to unsynced nodes.

cc @ngotchac, @Vurich, @ordian

rphmeier:
@ngotchac and I had a conversation about more predictable state chunk starting points and came to this proposal:

We get rid of duplicate-code optimizations where contract code shared between accounts is only present once (this may not affect size significantly for a few reasons).

[...]

PR #9239, opened 2018-07-27
Multithreaded snapshot creation

ngotchac:
This PR introduces multithreaded snapshot creation. The idea is that a snapshot will be divided in N parts, and the user can choose to multithread the process into T threads, each thread processing `N / T` parts of the snapshot.

For now, it would be divided into 16 different parts. The more parts, the bigger the snapshot will be, since there are no deduplication between parts. On Kovan, the snapshot's size increased from 1GB to 1.1GB, so a 10% increase. I still need to run this on Foundation.

Regarding performance gains, here's a little graph:
snapshot cpu

The time spent creating the snapshot is not linear with the number of threads, but with `sqrt(num_threads)`.

This PR adds thus a CLI argument `--snapshot-threads` which specifies the number of threads.
If this is accepted, the snapshot manifest could be edited in a future PR to specify the differents sub-parts of the state snapshot, so that the snapshot recovery would also work in parrallel.

It should be discussed whether this performance increase (in creating, and in a near future in restoring) outweighs the size increase.

**Update** : on Mainnet, splitting the snapshot into 16 subparts increases the number of chunks by only 2%, from 1818 to 1859 chunks. I was able to produce a full snapshot on 8 threads in 1h10min instead of the previously 5h.


rphmeier:
@ngotchac related to #8565 ?

rphmeier:
Does it produce bit-for-bit the same snapshot regardless of the number of threads? That's a property of snapshots which we want to keep

ngotchac:
Yes it does ! It won't if you change the `SNAPSHOT_SUBPART` variable (which for now is 16), but if you run 1,4 or 16 threads it will produce the same chunks, so the same snapshot.

ordian:
@ngotchac I'm going to take look after https://github.com/paritytech/parity-common/pull/13 (looking at it now) is merged, since there are too many distracting changes associated with the `branch = "ng-fix-triedb-seek"` switch.

ethcore/src/snapshot/mod.rs
```diff
@@ -281,11 +338,33 @@ pub fn chunk_state<'a>(db: &HashDB<KeccakHasher>, root: &H256, writer: &Mutex<Sn
 	let mut used_code = HashSet::new();
 
 	// account_key here is the address' hash.
-	for item in account_trie.iter()? {
+	let mut account_iter = account_trie.iter()?;
+
+	let mut seek_to = None;
+
+	if let Some(part) = part {
+		let part_offset = 256 / SNAPSHOT_SUBPARTS;
```
ordian: Maybe name `256`? I don't like magic constants.

ethcore/src/snapshot/mod.rs
```diff
@@ -263,10 +318,12 @@ impl<'a> StateChunker<'a> {
 
 /// Walk the given state database starting from the given root,
 /// creating chunks and writing them out.
+/// `part` is a number between 0 and 15, which describe which part of
+/// the tree should be chunked.
 ///
 /// Returns a list of hashes of chunks created, or any error it may
 /// have encountered.
-pub fn chunk_state<'a>(db: &HashDB<KeccakHasher>, root: &H256, writer: &Mutex<SnapshotWriter + 'a>, progress: &'a Progress) -> Result<Vec<H256>, Error> {
+pub fn chunk_state<'a>(db: &HashDB<KeccakHasher>, root: &H256, writer: &Mutex<SnapshotWriter + 'a>, progress: &'a Progress, part: Option<usize>) -> Result<Vec<H256>, Error> {
```
ordian: It seems like `part` is always `Some`, maybe we don't need an `Option`, do we?

ethcore/src/snapshot/mod.rs
```diff
@@ -263,10 +318,12 @@ impl<'a> StateChunker<'a> {
 
 /// Walk the given state database starting from the given root,
 /// creating chunks and writing them out.
+/// `part` is a number between 0 and 15, which describe which part of
```
ordian: Maybe add an assert, that `part < 16`?

ethcore/src/snapshot/mod.rs
```diff
@@ -281,11 +338,33 @@ pub fn chunk_state<'a>(db: &HashDB<KeccakHasher>, root: &H256, writer: &Mutex<Sn
...
+
+	let mut seek_to = None;
+
+	if let Some(part) = part {
+		let part_offset = 256 / SNAPSHOT_SUBPARTS;
+		let mut seek_from = vec![0; 32];
+		seek_from[0] = (part * part_offset) as u8;
+		account_iter.seek(&seek_from)?;
+
+		// Set the upper-bond, except for the last part
```
ordian: typo: upper bound

ethcore/src/snapshot/mod.rs
```diff
@@ -281,11 +338,33 @@ pub fn chunk_state<'a>(db: &HashDB<KeccakHasher>, root: &H256, writer: &Mutex<Sn
...
+		if part < SNAPSHOT_SUBPARTS - 1 {
```
ordian: we could get rid of this `if`, if we make `seek_to` inclusive upper bound, e.g.  the check would be
```rust
			if account_key[0] > seek_to {
				break;
			}
```

ethcore/src/snapshot/mod.rs
```diff
@@ -142,14 +164,46 @@ pub fn take_snapshot<W: SnapshotWriter + Send>(
...
+			state_guards.push(state_guard);
+		}
+
+		let block_hashes = block_guard.join()?;
+		let mut state_hashes = Vec::new();
+
+		for guard in state_guards {
+			let mut part_state_hashes = guard.join()?.clone();
+			state_hashes.append(&mut part_state_hashes);
+		}
```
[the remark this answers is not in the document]
ngotchac: `extend` expects a mutable reference :/

ethcore/src/snapshot/mod.rs
ngotchac: It's actually useful for testing

ethcore/src/snapshot/mod.rs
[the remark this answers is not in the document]
ordian: Hm, not sure what you mean: http://play.rust-lang.org/?gist=855f74c9c518bb8777179e8e3daee376&version=stable&mode=debug&edition=2015

https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Extend%3CT%3E

ethcore/src/snapshot/mod.rs
ngotchac: But `seek_to` wouldn't make sense if `part == None`, so it might as well be an Option

ethcore/src/snapshot/mod.rs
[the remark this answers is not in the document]
ngotchac: Oups, you're right, sorry.

ngotchac:
@ordian I updated it so that the default number of snapshot threads is half the number of cpus

PR #9243, opened 2018-07-27
add snapcraft cmake build dependency

gabreal:
build failed for snaps because of missing cmake dependency.


gabreal:
[clabot:check]

PR #9244, opened 2018-07-28
Provide read-only access to storage during tracing

udoprog:
I'm putting together a debugger based on Parity and need access to storage during tracing to be able to decode solidity variables.

### DIFF
--- a/ethcore/src/externalities.rs
+++ b/ethcore/src/externalities.rs
@@ -29,6 +29,7 @@ use vm::{
 };
 use transaction::UNSIGNED_SENDER;
 use trace::{Tracer, VMTracer};
+use storage::{StorageAccess, StorageError};
 
 /// Policy for handling output data on `RETURN` opcode.
 pub enum OutputPolicy {
@@ -411,7 +412,25 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	}
 
 	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem: &[u8]) {
-		self.vm_tracer.trace_executed(gas_used, stack_push, mem)
+		let Externalities { ref state, ref mut vm_tracer, .. } = *self;
+
+		let access = ExtStorageAccess(*state, &self.origin_info.address);
+		vm_tracer.trace_executed(gas_used, stack_push, mem, &access)
+	}
+
+	fn trace_finalized(&mut self) {
+		let Externalities { ref state, ref mut vm_tracer, .. } = *self;
+
+		let access = ExtStorageAccess(*state, &self.origin_info.address);
+		vm_tracer.trace_finalized(&access);
+	}
+}
+
+struct ExtStorageAccess<'a, 'b, B: 'a>(&'a State<B>, &'b Address);
+
+impl<'a, 'b, B: 'a> StorageAccess for ExtStorageAccess<'a, 'b, B> where B: StateBackend {
+	fn storage_at(&self, key: &H256) -> Result<H256, StorageError> {
+		self.0.storage_at(self.1, key).map_err(Into::into)
 	}
 }
 
--- a/ethcore/src/trace/mod.rs
+++ b/ethcore/src/trace/mod.rs
@@ -40,6 +40,7 @@ use ethereum_types::{H256, U256, Address};
 use kvdb::DBTransaction;
 use vm::{Error as VmError, ActionParams};
 use header::BlockNumber;
+use storage::StorageAccess;
 
 /// This trait is used by executive to build traces.
 pub trait Tracer: Send {
@@ -86,7 +87,12 @@ pub trait VMTracer: Send {
 	fn trace_prepare_execute(&mut self, _pc: usize, _instruction: u8, _gas_cost: U256, _mem_written: Option<(usize, usize)>, _store_written: Option<(U256, U256)>) {}
 
 	/// Trace the finalised execution of a single valid instruction.
-	fn trace_executed(&mut self, _gas_used: U256, _stack_push: &[U256], _mem: &[u8]) {}
+	fn trace_executed(&mut self, _gas_used: U256, _stack_push: &[U256], _mem: &[u8], _storage: &StorageAccess) {}
+
+	/// Trace the end of a subtrace.
+	///
+	/// After this, no more instructions will be executed for this subtrace.
+	fn trace_finalized(&mut self, _storage: &StorageAccess) {}
 
 	/// Spawn subtracer which will be used to trace deeper levels of execution.
 	fn prepare_subtrace(&mut self, _code: &[u8]) {}
--- a/evmbin/src/display/json.rs
+++ b/evmbin/src/display/json.rs
@@ -21,7 +21,7 @@ use std::mem;
 
 use ethereum_types::{U256, H256};
 use bytes::ToPretty;
-use ethcore::trace;
+use ethcore::{trace, storage};
 
 use display;
 use info as vm;
@@ -122,7 +122,7 @@ impl trace::VMTracer for Informant {
 		});
 	}
 
-	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem: &[u8]) {
+	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem: &[u8], _storage: &storage::StorageAccess) {
 		let subdepth = self.subdepth;
 		Self::with_informant_in_depth(self, subdepth, |informant: &mut Informant| {
 			let mem_diff = informant.mem_written.clone().map(|(o, s)| (o, &(mem[o..o+s])));
@@ -168,6 +168,15 @@ impl trace::VMTracer for Informant {
 		});
 	}
 
+	fn trace_finalized(&mut self, storage: &storage::StorageAccess) {
+		if self.unmatched {
+			// print last line with final state:
+			self.gas_cost = 0.into();
+			let gas_used = self.gas_used;
+			self.trace_executed(gas_used, &[], &[], storage);
+		}
+	}
+
 	fn prepare_subtrace(&mut self, code: &[u8]) {
 		let subdepth = self.subdepth;
 		Self::with_informant_in_depth(self, subdepth, |informant: &mut Informant| {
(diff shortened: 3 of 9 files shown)


udoprog:
Gonna fix the error when I have a moment. I realized I also have to add another hook for when the execution context of a contract is done that is not drain.

sorpaas:
@udoprog Can you explain more on the use-case here? If you are doing a normal tracing via full `ethcore`, I think the storage should already be "accessible" through the `store_diff` parameter. You maintain a hashmap, and move values into the hashmap as `store_diff` comes in. If you want more fine grained control, you can directly use `evm::Interpreter::step` and inspect the full vm state at each opcode.

Would this cover your use-case?

udoprog:
@sorpaas IIUC, `store_diff` only includes storage that is _changed_. At least my tests seem to indicate this assuming I did it correctly. This does not include persistent storage across subsequent contract invocations.

Is this correct?

sorpaas:
Yes. `store_diff` only includes changed storage. If you want to access other storage slots, would this work for you:

* Before execution, call `Client::state_at` to grab another state handle.
* When you need the data, call `State::storage_at` to get the *original* storage value.

udoprog:
@sorpaas what do you mean with "before execution". Can you show what you mean in code?

Note that StorageAccess as introduced here provides uniform access to both committed and uncommitted storage as a contract is being executed.

I think what you propose would require me to keep a reference to state_db around for when other contracts are called, get a snapshot of storage as it was before the transaction, and overlay any modifications I observe through store_diff. Is that correct?

It might be difficult to share state_db _during_ a trace since I think it's already mutably borrowed to run the transaction. But I would have to check that.

Overall it seems a fair bit harder to get right.

5chdn:
please rebase on master

udoprog:
@5chdn done

5chdn:
Can you have a look at this @debris @ordian :)

evmbin/src/display/std_json.rs
```diff
@@ -148,7 +148,7 @@ impl<Trace: Writer, Out: Writer> trace::VMTracer for Informant<Trace, Out> {
 	fn trace_prepare_execute(&mut self, _pc: usize, _instruction: u8, _gas_cost: U256) {
 	}
 
-	fn trace_executed(&mut self, _gas_used: U256, stack_push: &[U256], _mem_diff: Option<(usize, &[u8])>, store_diff: Option<(U256, U256)>) {
+	fn trace_executed(&mut self, _gas_used: U256, stack_push: &[U256], _mem_diff: Option<(usize, &[u8])>, store_diff: Option<(U256, U256)>, storage: &storage::StorageAccess) {
```
ordian: `_storage`

ethcore/src/trace/mod.rs
```diff
@@ -118,7 +119,12 @@ pub trait VMTracer: Send {
 	fn trace_prepare_execute(&mut self, _pc: usize, _instruction: u8, _gas_cost: U256) {}
 
 	/// Trace the finalised execution of a single valid instruction.
-	fn trace_executed(&mut self, _gas_used: U256, _stack_push: &[U256], _mem_diff: Option<(usize, &[u8])>, _store_diff: Option<(U256, U256)>) {}
+	fn trace_executed(&mut self, _gas_used: U256, _stack_push: &[U256], _mem_diff: Option<(usize, &[u8])>, _store_diff: Option<(U256, U256)>, _storage_access: &StorageAccess) {}
+
+	/// Trace the end of a subtrace.
+	///
+	/// After this, no more instructions will be executed for this subtrace.
+	fn trace_done(&mut self, _storage_access: &StorageAccess) {}
```
ordian: I don't have a strong opinion on naming, but we already have `done_subtrace`, so maybe `trace_finalized`?

evmbin/src/display/json.rs
```diff
@@ -100,7 +100,7 @@ impl trace::VMTracer for Informant {
 		self.gas_cost = gas_cost;
 	}
 
-	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem_diff: Option<(usize, &[u8])>, store_diff: Option<(U256, U256)>) {
+	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem_diff: Option<(usize, &[u8])>, store_diff: Option<(U256, U256)>, storage: &storage::StorageAccess) {
```
ordian: `_storage`

sorpaas:
@ordian We don't need this tracer anywhere for ourselves, and it's not exposed in RPC. So this is solely about using `ethcore` as a library and specific to @udoprog's usecase. I still believe that this can be done by just directly getting a `State` reference, but probably indeed it adds more convenience for what @udoprog tries to do.

sorpaas:
@udoprog Sorry I realized that I missed your last message!

> I think what you propose would require me to keep a reference to state_db around for when other contracts are called, get a snapshot of storage as it was before the transaction, and overlay any modifications I observe through store_diff. Is that correct?
>
> It might be difficult to share state_db during a trace since I think it's already mutably borrowed to run the transaction. But I would have to check that.

Sharing the reference would indeed be hard. However, if your usecase is debugging and you can bare a little bit of performance, you can clone it! `State` struct implements `Clone`, and once cloned, it's completely detached from any state modifications.

udoprog:
@sorpaas OK! I tried cloning, and with the current amount of storage that I'm using in tests it seems to work fine. I haven't tried overlaying the changes, but all the APIs are there, so I don't think that should be a problem.

I'm a bit concerned that I have to do it for every call, and that might not scale long term if I ever build tests with a bit more state. So I'd still be interested to do it without cloning at some point.

sorpaas:
@udoprog It's actually not as expensive as you think. `State` initially are really small, containing only the storage root and other relevant info. Caches of account and storage items are only built up when accessing fields. What's more, when fetching info from the disk, `State` will check for a global cache, which is shared.

As a result, for your usecase, if you clone `State` before VM execution, it should scale well.

Using `Arc` to wrap checkpoints on `State` would permanently solve the reference/cloning issue you're facing (and it also fixes https://github.com/paritytech/parity-ethereum/pull/9319#discussion_r215208765), but we probably don't have a date for that right now.

sorpaas:
@udoprog I tried to look into the code but still don't yet understand the use case (probably I haven't looked this hard enough!). It would be great if you can further explain what you tries to do:

* Do you want to print all changed storage values after each opcode? You don't need `State` reference for that -- just accumulate all changed values from the trace is enough.
* If what you want to do is to print all storage values of the executing account -- the issue is that real-world storage can be really big, and for this, we probably need to explore more to support this use case.
* If, however, what you want to do is to provide a debug interface -- interrupt execution after each opcode, allow the inspector to freely look into any storage slot it wants, then I would recommend to use `evm::Interpreter::<U256>::step` (you'll still need to get an `Externalities` from `ethcore`). This allows the execution to pause at each step, and you can freely inspect memory, storage, stack, or anything you would like!

udoprog:
@sorpaas I trace the execution of a contract, record each time storage is accessed. If a revert occurs, I lookup as much context as possible at the point it was reverted. That includes the value of most expressions that have been evaluated. If any of these expressions are references into storage, I lookup and decode their value from storage.

Right now I'm trying to rebase on the latest version of ethereum, and a fair bit of the tracing stuff has changed (e.g. subtracers no longer seem to be a thing).

This is a gif showcasing the state dumping in general: https://github.com/PrimaBlock/parables/blob/master/images/expression-dump.gif

That works because the access to storage is an immediate expression that I can decode, but if I for example have a loop I'm recording all storage accesses in order to dump that as part of the context. These storage slots then need to be decoded at a later time, by accessing storage directly.

5chdn:
@udoprog sorry, could you rebase this again?

ethcore/src/externalities.rs
```diff
@@ -411,7 +412,25 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	}
 
 	fn trace_executed(&mut self, gas_used: U256, stack_push: &[U256], mem_diff: Option<(usize, &[u8])>, store_diff: Option<(U256, U256)>) {
-		self.vm_tracer.trace_executed(gas_used, stack_push, mem_diff, store_diff)
+		let Externalities { ref state, ref mut vm_tracer, .. } = *self;
+
+		let access = ExtStorageAccess(*state, &self.origin_info.address);
```
sorpaas: Would it be possible to change to a design where if `NoopTracer` is used, we don't create this struct? For example, this can be done by directly passing `State` reference into tracer.

udoprog: The benefit is pairing the storage access with the address of the contract so that we don't have to do this, making the storage access API more convenient.

What are the worries about from setting up this struct? The overhead should be negligible since we are mostly dealing with pointers.

udoprog:
@5chdn I'll rebase after it's been reviewed or not if it's rejected if that's fine by you?

I've already rebased a couple of times.

ethcore/src/externalities.rs
sorpaas: Yeah looks like the overhead is negligible. However, I wonder about other tracer usage for ethcore-as-a-library. People may want to build other tracer implementation that access to other state information (like other account's storage slot, account code, nonce, etc). So I'm just thinking maybe directly passing `State` would be more generalized and potentially fit more usage?

udoprog:
I don't think this applies cleanly since 1e9aebbc86

If I'm reading that correctly, calls can now be suspended and resumed (the details of which are not entirely clear to me yet). When this happens a new externality is created and finalized. This messes with the `trace_finalized` call introduced here since it expected the ext to live for the duration of a single call.

`done_subtrace` seems to be the correct hook to use, but that's not called through an ext any longer. But it does have access to the `state` so introducing storage here seems doable. But it's not something that can be suitable achieved through a rebase.

sorpaas:
@udoprog Yeah. It's possible to get the same interface as of now, but also not sure if it suits you need still -- there's no concept of subtracers any more.

sorpaas:
`done_subtrace` should be fine to introduce `State` here if you're sure that suits your use case!

5chdn:
Are you still working on this @udoprog ?

udoprog:
@5chdn No, unfortunately I currently don't have the time to familiarize myself with the changes done while this was pending. Either someone else can pick it up, or feel free to close it and I'll open a new one if I end up finding the time.

5chdn:
Ok, no worries.

PR #9254, opened 2018-07-31
Improve return data truncate logic

sorpaas:
Rust's `Vec` allocator will never change the init pointer, so `drain` will do copy anyway (with the unwanted tail data). This PR fixes it by only attempting to truncate if `offset` is `0`, and for other cases, just create a new `Vec`.


ethcore/evm/src/interpreter/memory.rs
```diff
@@ -119,14 +119,19 @@ impl Memory for Vec<u8> {
 	fn into_return_data(mut self, offset: U256, size: U256) -> ReturnData {
 		let mut offset = offset.low_u64() as usize;
 		let size = size.low_u64() as usize;
+
 		if !is_valid_range(offset, size) {
-			return ReturnData::empty()
+			return ReturnData::empty();
```
NikVolf: avoid redundant `;`

PR #9257, opened 2018-07-31
Update ref to `parity-common` and update `seek` behaviour

ngotchac:
This PR: https://github.com/paritytech/parity-common/pull/13 updates the behavior of `seek` in TrieDBIterators. 
This PR updates the reference to `parity-common` and updates the behavior of `seek` which places the iterator on the given element (if existing) or the closest next one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ fdlimit = "0.1"
 ctrlc = { git = "https://github.com/paritytech/rust-ctrlc.git" }
 jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
 ethcore = { path = "ethcore", features = ["parity"] }
-parity-bytes = { git = "https://github.com/paritytech/parity-common" }
+parity-bytes = { git = "https://github.com/paritytech/parity-common", branch = "ng-fix-triedb-seek" }
 ethcore-io = { path = "util/io" }
 ethcore-light = { path = "ethcore/light" }
 ethcore-logger = { path = "logger" }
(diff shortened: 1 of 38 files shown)


ngotchac:
@dvdplm That's right!

ethcore/src/client/client.rs
```diff
@@ -1692,6 +1692,9 @@ impl BlockChainClient for Client {
 		if let Some(after) = after {
 			if let Err(e) = iter.seek(after) {
 				trace!(target: "fatdb", "list_accounts: Couldn't seek the DB: {:?}", e);
+			} else {
+				// Position the iterator after the `after` element
+				iter.next();
```
ordian: Could you explain how this works? Why calling `next` after `seek` returned an `Err` would position the iterator after the `after`?

ngotchac: So the `else` branch is actually when there has been no errors

ordian: Ah, right, thanks for clarifying that. So we assume that `after` is present in the db, otherwise this is not correct?

ngotchac: So in this context, `last` is whether `None` or an address that is actually in DB.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ fdlimit = "0.1"
 ctrlc = { git = "https://github.com/paritytech/rust-ctrlc.git" }
 jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
 ethcore = { path = "ethcore", features = ["parity"] }
-parity-bytes = { git = "https://github.com/paritytech/parity-common" }
+parity-bytes = { git = "https://github.com/paritytech/parity-common", branch = "ng-fix-triedb-seek" }
```
sorpaas: Just an idea. To get "atomic" PRs, we can make `parity-common` repo a submodule of this repo. A submodule is pinned to a commit, so we can avoid always manually changing branches. (And it also makes browsing all ethereum-related code easier!)

ethcore/src/client/client.rs
debris: I think this could be simplified by writing:

```rust
if let Err(e) = iter.next(after) {
  trace!("...");
}
```

Cause `Err` is no used later on anyway

ngotchac:
Hm, the error is used in the trace log, and `next` doesn't give an error

debris:
ah, you are totally right! I confused seek with peek 🤦‍♂️

Cargo.toml
ngotchac: What if we just specify a commit hash for `parity-common`? Should be easier to implement, and should be safe.

PR #9252, opened 2018-07-30
decode block rlp less often

debris:
in total:
- removed 4 redundant rlp deserializations
- avoid 1 redundant block data copy

PR #9264, opened 2018-08-01
ethcore sync decodes rlp less often

debris:
should be merged after #9252


dvdplm:
@debris can you merge #9252 to make it easier to review the changes here?

andresilva:
Please rebase/merge master (which already includes #9252).

ethcore/sync/src/blocks.rs
```diff
@@ -278,30 +358,33 @@ impl BlockCollection {
...
-						}
-						_ => break,
+					match self.blocks.remove(&head) {
+						Some(block) => {
+							if block.body.is_some() && (!self.need_receipts || block.receipts.is_some()) {
+								blocks.push(block);
+								hashes.push(head);
+								self.head = Some(head);
+							} else {
+								self.blocks.insert(head, block);
```
ascjones: So previously this would have removed an already downloaded header? So adding it back is just preventing it from being downloaded again?

debris: so the logic has not changed here. previously we called `self.blocks.get(...)` (line 281) and then `self.blocks.remove(...)` line 303. Now we are instantly removing `self.blocks.remove(...)` (line 361), but we need to readd blocks that shouldn't be removed (this line)

debris: it lets us avoid unnecessary searching in the hashmap, and unnecessary copying of the blocks (few lines below)

ethcore/sync/src/block_sync.rs
```diff
@@ -236,45 +235,39 @@ impl BlockDownloader {
 		let mut valid_response = item_count == 0; //empty response is valid
 		let mut any_known = false;
 		for i in 0..item_count {
-			let info: BlockHeader = r.val_at(i).map_err(|e| {
-				trace!(target: "sync", "Error decoding block header RLP: {:?}", e);
-				BlockDownloaderImportError::Invalid
-			})?;
-			let number = BlockNumber::from(info.number());
+			let info = SyncHeader::from_rlp(r.at(i)?.as_raw().to_vec())?;
```
andresilva: Do you think it's important to keep the logging message? Or will this be logged elsewhere?

ethcore/sync/src/block_sync.rs
```diff
@@ -236,45 +235,39 @@ impl BlockDownloader {
...
+			if self.blocks.contains(&hash) {
+				trace!(target: "sync", "Skipping existing block header {} ({:?})", number, hash);
 				continue;
 			}
 
 			if self.highest_block.as_ref().map_or(true, |n| number > *n) {
 				self.highest_block = Some(number);
 			}
-			let hash = info.hash();
-			let hdr = r.at(i).map_err(|e| {
```
andresilva: We were decoding this twice in the same function just because... :scream:

ethcore/sync/src/block_sync.rs
debris: it's already logged upstream :)

Issue #9057, opened 2018-07-06
Add parity_getRawBlockByNumber and parity_submitRawBlock RPC API endpoints


ppratscher:
In order to allow for a little bit more flexibility when mining blocks it would be nice to have a more low level block API available. Currently the eth_getWork and eth_submitWork RPC calls do not allow miners to make any changes to the pending block which in turn make large scale mining setups quite complicated given the deficits of the Ethereum stratum protocol variants currently in use.

Thus I would like to propose to add two new RPC API endpoints:
parity_getRawBlockByNumber should return a RLP encoded block and should accept the same parameters as eth_getBlockByNumber (including returning the pending block)

parity_submitRawBlock should accept a RLP encoded block and publish the block to the network (similar to the eth_sendRawTransaction call)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.9 ETH (423.39 USD @ $470.43/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/paritytech/parity/9057/732).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $19,025.04 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 2 months, 4 weeks from now. Please review their questions below:


1. [kelvintyb](https://gitcoin.co/profile/kelvintyb) has started work. 	
 * Q: Hi, would like to try my hand at this while I'm picking up Rust. Noted also that this is planned for the 2.1 release, I will try to get it done by end of month so there's enough buffer time to work on feedback. 

Couple of questions to start:
1) Is there a particular dev I should be directing questions to in the chat for this issue?
2) Correct me if I'm wrong, but seems like the proposed API endpoints require RLP encoding/decoding, is there a preferred library for this that's already in use elsewhere in the codebase? Generally I try not to add on more dependencies than necessary.

gitcoinbot:
[...]

PR #9267, opened 2018-08-01
[WIP] Add parity_getRawBlockByNumber and parity_submitRawBlock RPC API endpoints

kelvintyb:
Currently WIP.

This fixes issue #9057.

### DIFF
--- a/rpc/src/v1/impls/parity.rs
+++ b/rpc/src/v1/impls/parity.rs
@@ -103,7 +103,7 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 {
 	type Metadata = Metadata;
 
-	fn rich_block(&self, id: BlockNumberOrId, include_txs: bool) -> Result<Option<RichBlock>> {
+	fn rich_block(&self, id: BlockNumberOrId, include_txs: bool) -> Result<Option<Bytes>> {
 		let client = &self.client;
 
 		let client_query = |id| (client.block(id), client.block_total_difficulty(id), client.block_extra_info(id), false);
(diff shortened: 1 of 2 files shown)


kelvintyb:
This is a rough implementation taking reference from the spec given in the linked issue, would like comments on whether this is correct directionally. 

I see that the Block object in ethcore has a method rlp_bytes() that seems to be aligned with the need for translating the block into bytes for submitting as a transaction -> is it accurate to say that this should be re-implemented for the Block type in rpc/src/v1/types/block.rs?

Hope to firm up the logic and start adding tests in my next commit to this.

rpc/src/v1/impls/parity.rs
```diff
@@ -103,6 +103,99 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 {
 	type Metadata = Metadata;
 
+	fn rich_block(&self, id: BlockNumberOrId, include_txs: bool) -> Result<Option<RichBlock>> {
```
andresilva: You should return the hex-encoded block RLP here instead of a `RichBlock`.

rpc/src/v1/impls/parity.rs
```diff
@@ -417,6 +510,29 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 		}))
 	}
 
+		fn raw_block_by_number(&self, num: BlockNumber, include_txs: bool) -> BoxFuture<Option<RichBlock>> {
```
andresilva: As commented above, the type of this function should be `fn raw_block_by_number(&self, num: BlockNumber, include_txs: bool) -> BoxFuture<Option<Bytes>>`.

andresilva: Indentation.

rpc/src/v1/impls/parity.rs
```diff
@@ -417,6 +510,29 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
...
+			Box::new(future::done(self.rich_block(num.into(), include_txs)))
+		}
+
+		fn submit_block(&self, block: RichBlock) -> Result<RpcH256> {
```
andresilva: Indentation.

rpc/src/v1/impls/parity.rs
```diff
@@ -417,6 +510,29 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
...
+			// TODO: Re-implement rlp_bytes for rpc/v1/src/types/block.rb 
+			//       or reuse the rlp_bytes from ethcore/src/block.rb?
+			rlp_encoded_block = block.rlp_bytes()
+			Rlp::new(&rlp_encoded_block.into_vec()).as_val()
+				.map_err(errors::rlp)
+				.and_then(|tx| SignedTransaction::new(tx).map_err(errors::transaction))
+				.and_then(|signed_transaction| {
+					FullDispatcher::dispatch_transaction(
```
andresilva: This wraps the block RLP and dispatches it as if it were a transaction, whereas in this case it is a full block. We should validate the block against our chain and then import it locally using the miner (which should trigger propagation afterwards).

I think we should validate the block and import it locally (this requires that the client is synced for this method to work) instead of blindingly broadcasting it, since otherwise this endpoint can be used to easily spam the network with invalid blocks.

rpc/src/v1/impls/parity.rs
andresilva: Should not have the `include_txs` parameter.

kelvintyb:
Thanks all for the comments, I will review and make the necessary changes tomorrow; may need a little more time for the last part to explore validation of the block, but the rest should be no issue.

kelvintyb:
clabot:check

rpc/src/v1/impls/parity.rs
kelvintyb: @andresilva did you mean that I should convert the result of this function into Bytes using .rlp_bytes here first, instead of in fn submit_block (in the same file)? 

Also, to clarify on the include_txs parameter, since this function is to be used to allow miners to make any changes to the pending block before submission, is it then the case that I should always assume we include localized transactions and set `transactions` of the block to 

BlockTransactions::Full(block.view().localized_transactions().into_iter().map(|t| Transaction::from_localized(t)).collect()) ?

rpc/src/v1/impls/parity.rs
kelvintyb: @ppratscher is there something in the current codebase that already does this or can be adapted to accept the block RLP? The code block within `impl Rebuilder for PowRebuilder` (starting at line 220 in ethcore/src/snapshot/consensus/work.rs) seems to have some relevance to this, but I'm unsure if I'm missing the mark on this one. Appreciate any further guidance you can give on this one!

kelvintyb:
@andresilva @ppratscher left some questions in the previous comment threads for the earlier commit, appreciate your input on those. Thanks!

rphmeier:
Submitting the block as a transaction doesn't make sense. The `submitRawBlock` should simply attempt to import the block the same way that the sync service does.

kelvintyb:
@rphmeier I think I took the spec laid out in the issue too literally for submitRawBlock; which sync service in the codebase are you referring to? (parity/stratum.rs, or someplace in /rpc?)

5chdn:

kelvintyb:
Apologies, I didn't get any follow ups on my previous comments and didn't get round to sending the notification to stop work. Please feel free to use my previous code if it's relevant for completing this task.

PR #9268, opened 2018-08-02
Add EIP-1014 transition config flag

sorpaas:
rel #8427


Tbaut:
Can we also update the doc please? https://github.com/paritytech/wiki/blob/master/Chain-specification.md

andresilva:
Maybe the EIPs should be updated so that both interfaces are compatible?

sorpaas:
Hmm indeed. I apparently missed this important difference. I added a comment on https://github.com/ethereum/EIPs/pull/1014. Let me put this on ice until the spec is cleared out.

sorpaas:
Per discussion during the last AllCoreDevs, the issue is that even if we keep EIP-86, it will need to be vastly changed. So it was decided to deprecate EIP-86:

* I removed all EIP-86 related flags.
* Null signed transaction functionalities are kept. We're still using null signed transactions some places in our code. In execution context, the old config `eip86` is only used to know whether we should keep null signed address nonce unchanged. So I renamed it to `keep_unsigned_nonce`.
* Comply the CREATE2 definition with the current EIP-1014 (Option 2). This also affects KIP-4, but I guess that's unreleased feature so it should be okay.

sorpaas:
> we can try to avoid the call to nonce for our new opcode by changing `contract_address` function prototype a bit.

Can't think of a good way to refactor that right now. I think it would need to give a state reference to `contract_address` fn but that'll certainly not worth it. I'll keep thinking though. :)

> if I understood correctly `keep_unsigned_nonce` is only here for deprecating a functionality in the future, but not related to this EIP

`keep_unsigned_nonce` is for null signed transaction, which was a part of EIP-86. I think that's a generally useful feature even though we don't want EIP-86 itself any more (for example, the old Casper spec needs null signed transaction). And yes, that change is not related to EIP-1014.

sorpaas:
Ooops! I apparently forgot to push the fix commit for @cheme's grumbles. Sorry!

cheme:
I also forgot the 'grumble' tag, I am glad I understand correctly the `keep_unsigned_once`, seems like I do not have other comments.

cheme:
Sorry for the misclick (I accidently close the pr).

sorpaas:
Yeah I think let's just wait till this Friday's meeting (https://github.com/ethereum/pm/issues/55), and merge this if there's no other issues.

With this PR, `create2` also uses the `sha3(0xff ++ msg.sender ++ salt ++ sha3(init_code)))`. I'll get the spec update done soon.

5chdn:
Putting this on ice till Friday.

PR #9231, opened 2018-07-26
nonroot CentOS Dockerfile

JohnnySheffield:
## Usage - CentOS

Builds a lightweight non-root Parity docker image:

```
git clone https://github.com/paritytech/parity-ethereum.git
cd parity-ethereum
./docker/centos/build.sh
```

Fully customised build:
```
PARITY_IMAGE_REPO=my-personal/parity \
PARITY_BUILDER_IMAGE_TAG=build-latest \
PARITY_RUNNER_IMAGE_TAG=centos-parity-experimental \
./docker/centos/build.sh
```

Default values:
```
# The image name
PARITY_IMAGE_REPO - parity/parity

# The tag to be used for builder image, git commit sha will be appended
PARITY_BUILDER_IMAGE_TAG - build

# The tag to be used for runner image
PARITY_RUNNER_IMAGE_TAG - latest
```

All default ports you might use will be exposed:
```
#           secret
#      ipfs store     ui   rpc  ws   listener  discovery
#      ↓    ↓         ↓    ↓    ↓    ↓         ↓
EXPOSE 5001 8082 8083 8180 8545 8546 30303/tcp 30303/udp
```
Related to #7374

PR #9269, opened 2018-08-02
Docker alpine: use multi-stage concept

fanatid:
This brings multi-stage concept and nonroot user.
Current image size is 1.42GB, with this Pull Request 58.2MB (20MB compressed on docker hub)

### DIFF
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -37,6 +37,7 @@ EXPOSE 8080 8545 8180
 
 WORKDIR /home/parity
 
+RUN mkdir -p /home/parity/.local/share/io.parity.ethereum/
 COPY --chown=parity:parity --from=builder /parity/target/x86_64-alpine-linux-musl/release/parity ./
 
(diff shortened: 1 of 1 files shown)


fanatid:
[clabot:check]

sorpaas:
@fanatid Merge #9270 would fix this CI build.

fanatid:
@sorpaas thanks, rebased

JohnnySheffield:
Do we care about supporting lower versions of Docker? For multistage builds one needs Docker 17.05.

fanatid:
Yes, multistage require Docker >= 17.05 :(
I can rewrite it without multistage without significant changes of image size, if this required.

JohnnySheffield:
I was kinda 50:50 with #9231 but then decided to go with supporting <17.05 I'm also not sure about different platform support. For instance, i think docker rpm package for RHEL is v1.13

I'm ready to rework  #9231 to go with multistage builds if needed, but i think this is more of a question for Parity team what their plans are

PR #9273, opened 2018-08-02
Avoid using $HOME if not necessary

tomaka:
When Parity is used within an Android application, the `$HOME` environment variable isn't set. This will cause Parity to panic.

This commit fixes it by avoiding to load the value of `$HOME` if it is not required.

### DIFF
--- a/parity/upgrade.rs
+++ b/parity/upgrade.rs
@@ -105,7 +105,9 @@ fn upgrade_from_version(previous_version: &Version) -> Result<usize, Error> {
 fn with_locked_version<F>(db_path: Option<&str>, script: F) -> Result<usize, Error>
 	where F: Fn(&Version) -> Result<usize, Error>
 {
-	let mut path = db_path.map_or({
+	let mut path = db_path.map_or_else(|| {
+		// Note: this can panic on device such as Android where $HOME isn't set
+		// That's why the `map_or_else` is important compared to `map_or`.
 		let mut path = env::home_dir().expect("Applications should have a home dir");
 		path.push(".parity");
 		path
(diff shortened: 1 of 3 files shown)


parity/upgrade.rs
```diff
@@ -105,7 +105,9 @@ fn upgrade_from_version(previous_version: &Version) -> Result<usize, Error> {
 fn with_locked_version<F>(db_path: Option<&str>, script: F) -> Result<usize, Error>
 	where F: Fn(&Version) -> Result<usize, Error>
 {
-	let mut path = db_path.map_or({
+	let mut path = db_path.map_or_else(|| {
```
ordian: It seems to me like the `db_path` is always `is_some()` (`with_locked_version` is only used in `upgrade`, and `db_path` in `upgrade` is always `Some`). Maybe we can remove `map_or_else` altogether, or am I wrong?

tomaka: I think you're right.

tomaka:
It turns out that this code doesn't work in practice (we get a panic on Android), so I'll work a bit more on it.

tomaka:
The code is now working. Parity runs even if you don't have `$HOME` set, provided you pass all the necessary `--no-*` options.
Also fixed the remark.

parity/upgrade.rs
```diff
@@ -205,6 +201,10 @@ fn upgrade_user_defaults(dirs: &DatabaseDirectories) {
 }
 
 pub fn upgrade_data_paths(base_path: &str, dirs: &DatabaseDirectories, pruning: Algorithm) {
+	if env::home_dir().is_none() {
+		return;
```
ordian: Does it make make sense to print a warning here?

tomaka: I don't think so. The function only does something if `$HOME/.parity` exists, and doesn't print any warning if it doesn't. If `$HOME` isn't defined we behave exactly as if `$HOME/.parity` doesn't exist.

PR #9275, opened 2018-08-02
ethcore: add transition flag for transaction permission contract

andresilva:
Replaces #8940.


andresilva:
cc @Tbaut update docs with this.

Tbaut:
@andresilva as a general reflex (and for any dev when updating such small things), if you can create a PR for the wiiki when introducing a new flag, that'd be great. I did it for this one though :)

PR #9277, opened 2018-08-03
Implement KIP4: create2 for wasm

### LINKED ISSUE #9139: pwasm with FromSenderSaltAndCodeHash scheme contract creation
rel #9140 

Currently, in parity-ethereum's wasm engine, we use `FromSenderAndCodeHash` scheme for `create` extern runtime. This seems to be from the old EIP-86 CREATE_P2SH opcode. The issue with this, as pointed out in EIP-86, is that one contract cannot create multiple "sub"-contracts with the same init code.

Would we consider to add a new `create2` extern runtime for pwasm to support `FromSenderSaltAndCodeHash` scheme (i.e. when generating contract address, keccak sender, a 32 byte "salt", and code hash, instead of only keccak sender and code hash)?

sorpaas:
Fixes #9139, implements https://github.com/kovan-testnet/KIPs/pull/4


ethcore/wasm/src/runtime.rs
```diff
@@ -571,6 +549,59 @@ impl<'a> Runtime<'a> {
...
+	/// * result_ptr - pointer to write an address of the newly created contract
+	pub fn create2(&mut self, args: RuntimeArgs) -> Result<RuntimeValue> {
+		//
+		// method signature:
+		//   fn create2(endowment: *const u8, salt: *const u8, code_ptr: *const u8, code_len: u32, result_ptr: *mut u8) -> i32;
+		//
+		trace!(target: "wasm", "runtime: CREATE2");
+		let endowment = self.u256_at(args.nth_checked(0)?)?;
+		trace!(target: "wasm", "       val: {:?}", endowment);
+		let salt: H256 = self.u256_at(args.nth_checked(0)?)?.into();
```
pepyakin: seems like param indexes should be offset by +1

NikVolf:
could be added to the existing creator.wasm

https://github.com/paritytech/wasm-tests/blob/474110de59a0f632b20615256c913b144c49354c/src/creator.rs

sorpaas:
Let me try to add the tests.

eosclassicteam:
@sorpaas This will make ellaism to fork again I guess 😄

sorpaas:
Added `create2` tests in `wasm-tests`.

sorpaas:
@pepyakin Hmm regarding gas costs, it looks like other tests are passing, so it simply means nothing has changed?

I used the same nightly version as specified as [rust-toolchain](https://github.com/paritytech/wasm-tests/blob/master/rust-toolchain), so it may indeed just be the case.

pepyakin:
> Hmm regarding gas costs, it looks like other tests are passing, so it simply means nothing has changed?

Yeah, I think even after rebuilding with the same nightly I sometimes experienced similar problems (maybe it was something non-deterministic like incremental compilation and/or codegen-units being more than 1, idk).

But I believe this might not longer be the case and green tests are a good evidence of that : )

PR #9280, opened 2018-08-03
nonroot CentOS Docker image

JohnnySheffield:
@niklasad1 @General-Beck Continuation of #9231, but with a much cleaner commit history.


docker/centos/build.Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM centos:latest
```
debris: why is this file called `build.Dockerfile`? Is it a common convention? (I'm not a docker expert)

docker/centos/build.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+WORKDIR /build
+
+ADD . /build/parity-ethereum
+
+RUN yum -y update && \
+    yum install -y systemd-devel git make gcc-c++ gcc file binutils && \
+    curl -L "https://cmake.org/files/v3.12/cmake-3.12.0-Linux-x86_64.tar.gz" -o cmake.tar.gz && \
```
debris: why not `yum install cmake` ?

JohnnySheffield: yum install cmake installs cmake 2.x, but then parity build fails as it needs 3.x, so i decided to install cmake directly. Ideally i would like to get it with yum, but had no luck

docker/centos/build.Dockerfile
JohnnySheffield: I'll rename it to Dockerfile.build, short google points that this is some kind of convention

PR #9281, opened 2018-08-05
Fix `env::home_dir()` deprecation warning.

c0gent:
* Import the `dirs` crate in `util/dir`.
* Replace uses of `env::home_dir()` with `dirs::home_dir()`.
* Reexport `dirs::home_dir`.
* Continue to use `std::env::home_dir` on Android.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,11 +357,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "dir"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "app_dirs 1.2.1 (git+https://github.com/paritytech/app-dirs-rs)",
+ "dirs 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "ethereum-types 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "journaldb 0.2.0",
+]
+
+[[package]]
+name = "dirs"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


c0gent:
[clabot:check]

tomaka:
Thanks for the PR. However [we reverted the exact same change](https://github.com/paritytech/parity-ethereum/pull/9097) some time ago because the `dirs` crate only supports Windows, Linux and Mac.

c0gent:
My mistake. Not sure how I missed that PR.

Should I add some cfg attributes to keep Android on `std::env` then?

c0gent:
Updated with exceptions for Android.

tomaka:
In my opinion the cost of making the code dirty is higher than seeing this small warning when you compile. I would personally prefer an `#[allow(deprecated)]` and an issue opened, compared to having platform-specific code.

For me the proper fix is to change the code so that the Parity binary introduces $HOME into the library part, but that's a much bigger change.

c0gent:
I see what you mean and I hear you 100% about the code dirtyness but because it will be such a large amount of work to make such a minor thing work spotlessly, and because the drawbacks of the quick fix are so few (other than it bothering us), I think it would be better to merge and leave an open issue until the time can be spent.

c0gent:
I can optionally just leave the deprecation warnings enabled on Android as reminders (they are suppressed currently).

c0gent:
I understand completely. I can add a `#![allow(deprecated)]` but I really don't think that a blanket allow is a good idea here and I don't think it solves much (considering we'd still need the `allow` when `dir::home_dir` is called -- which is only once in this case).  If they must be used, I prefer the 'warts' of piecemeal `allow`s in order to highlight precisely what's being suppressed and prevents anything else from accidentally going under the radar. (I know you know all of this -- by the way I'm a long time `glium`/`vulkano` user, I've long appreciated your work and contributions!)

I can close this PR for now. I won't have time to do a proper implementation right now.

PR #9285, opened 2018-08-06
Fix loop start value

Jongsic:
Fix loop start value of `to_hash` I think this is very little mistake.


Jongsic:
[clabot:check]

andresilva:
We should probably add some tests for this so we don't regress in the future. @Jongsic feel free to submit another PR if you want to add any tests for this. 🙂👍

PR #9292, opened 2018-08-06
Allow setting the panic hook with parity-clib

tomaka:
For example on Android, we want to present the panic message to the user with a dialog box, and put it in the logs (which are not the same as stderr).

This PR adds a `parity_set_panic_hook` function in the API of parity-as-a-library, so that we can choose what happens in case of a panic.

### DIFF
--- a/parity-clib/src/lib.rs
+++ b/parity-clib/src/lib.rs
@@ -18,6 +18,7 @@
 //! duplicating documentation.
 
 extern crate parity_ethereum;
+extern crate panic_hook;
 
 use std::os::raw::{c_char, c_void, c_int};
 use std::panic;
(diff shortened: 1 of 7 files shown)


parity-clib/src/lib.rs
```diff
@@ -162,3 +163,21 @@ pub extern fn parity_rpc(client: *mut c_void, query: *const char, len: usize, ou
 		}).unwrap_or(1)
 	}
 }
+
+#[no_mangle]
+pub extern fn parity_set_panic_hook(callback: extern "C" fn(*mut c_void, *const c_char, usize), param: *mut c_void) {
```
rphmeier: should this not be `unsafe` to invoke?

rphmeier: right now I can call this from Rust code with crap data and it will register it as a thread-safe panic hook

tomaka:
I made all the other FFI functions unsafe as well.

util/panic_hook/src/lib.rs
```diff
@@ -24,16 +24,26 @@ use std::thread;
...
+		process::abort()
+	});
 }
 
-/// Set the panic hook with a closure to be called afterwards.
-pub fn set_with<F: Fn() + Send + Sync + 'static>(f: F) {
+/// Set the panic hook with a closure to be called. The closure receives the panic message.
+///
+/// Depending on how Parity was compiled, after the closure either returns the process aborts or
+/// unwinding starts.
```
niklasad1: I don't understand this, can you rephrase?

I guess it should be something like `"after the closure has been executed it either returns the exit code when the process terminates or the unwinding starts"`

parity-clib/src/lib.rs
```diff
@@ -33,132 +34,140 @@ pub struct ParityParams {
...
+		let on_client_restart_cb = {
+			struct Cb(Option<extern "C" fn(*mut c_void, *const c_char, usize)>, *mut c_void);
+			unsafe impl Send for Cb {}
+			unsafe impl Sync for Cb {}
+			impl Cb {
+				fn call(&self, new_chain: String) {
+					if let Some(ref cb) = self.0 {
+						cb(self.1, new_chain.as_bytes().as_ptr() as *const _, new_chain.len())
 					}
 				}
```
niklasad1: "Duplicate struct definition", the only thing the differs is that this one is inside Option and https://github.com/paritytech/parity-ethereum/blob/4ce97f9554d2353f225d8f47f615e85a019e0316/parity-clib/src/lib.rs#L159-#L167 is not!

Can you move the `"local struct definitions"` to one `file scope definition` and rename it to `Callback`? I think the full name is worth the effort here!

PR #9293, opened 2018-08-06
Replace `std::env::home_dir()` with `home` crate impl.

c0gent:
* Import the `home` crate in `util/dir`.

* Replace uses of `env::home_dir()` with `home::home_dir()`.
  * `home` uses a 'correct' impl. on windows and the stdlib impl. of `::home_dir` otherwise.

* Reexport `home::home_dir` from `util/dir`.

* Bump `util/dir` to 0.1.2.

Addresses deficiencies of https://github.com/paritytech/parity-ethereum/pull/9077 and https://github.com/paritytech/parity-ethereum/pull/9281. Tested on Linux and Windows.


c0gent:
@tomaka Please let me know what you think about this.

I'm guessing `should_check_status_of_request_when_its_resolved` can spuriously fail?

dvdplm:
@niklasad1 this is very similar to your [PR from early July](https://github.com/paritytech/parity-ethereum/pull/9077) that was later reversed because of Android issues I think? I don't remember the conclusion of that though, to hold off?

tomaka:
Would merging this PR move the base directory of people using Windows?

c0gent:
What this does is make the home directory more consistent. It's possible that someone compiling and running Parity on msys/cygwin *could* now have a different home folder. I suppose I could add a check for existing base directories, in what would have been considered the old location, in the event that no base directory is found in the new. What do you think?

niklasad1:
@dvdplm This is slightly different instead of using `dirs crate` as rustc suggest. This `PR` is using the `home crate` which is using a different Windows implementation but the same implementation as the `stdlib` for UNIX. 

So, my educated guess is that this will work fine on Android (UNIX impl) because it doesn't have any conditional compilation guards to prevent platforms other than Linux, OSX and Windows. (I think the home_dir will be root folder on Android not sure though)

But this must absolutely be tested on ```Android``` before merging!

tomaka:
> But this must absolutely be tested on Android before merging!

Well, `home_dir()` will return `None` on Android because the HOME directory doesn't really make sense.

dvdplm:
@c0gent can you sort out the merge conflicts so we can merge this please? :)

c0gent:
Done.

niklasad1:
@c0gent 

```README.MD``` should not be deleted, can you revert that?

ordian:
Please revert `README.MD` deletion in the next PR.

PR #9294, opened 2018-08-06
Update `log` -> 0.4, `env_logger` -> 0.5.

c0gent:
* Rename a few types & methods.

* Change `(Log)Builder::format` (closure) arg.


niklasad1:
Hey @c0gent 

Thanks for your PR!

Do you know if `log v0.4` is compatible with `log v0.3`? Because a few crates still depend on `log v0.3`  and I want to make sure the logs still works for these!

For example `hyper-rustls` depends on `log 0.3` and last time I checked it was not trivial to bump that crate because we use our own fork of ring which ended up in conflicting versions!

c0gent:
According to the [`CHANGELOG`](https://github.com/rust-lang-nursery/log/blob/master/CHANGELOG.md), the only real incompatibility comes when consuming a 0.4-using crate from a 0.3-using one. In that case the module and file name information is lost. My understanding is (and in my limited testing) the reverse setup (as is the case here) works fine.

c0gent:
The latest version of `hyper-rustls` (0.13) is on `log` 0.4 (in its `rustls` dep.). The older version currently in use (0.11), currently uses `log` 3.9. Indeed, all `log` versions in use within Parity are at least 3.9 (according to `Cargo.lock`).

andresilva:
Thanks :+1:

PR #9303, opened 2018-08-07
Add HTTPS proxy env variable handling

kavu:
Hi! 

I want to purpose HTTPS proxy handling, which commonly passed with env variable `HTTPS_PROXY` (or lowercase `https_proxy`).

The reason I purpose it is that GasPriceCalibrator cannot fetch ETH price in my environment, which has quite strict proxy.

It's my first Pull Request into Parity and in Rust general, so I am looking forward on tips, advises and corrections here. 

By the way, I did not include `rustfmt` formatting of `util/fetch`, because there will be a lot of chunks to review.

Hope that helps!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,6 +273,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "core-foundation-sys 0.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "crossbeam"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1093,6 +1110,7 @@ dependencies = [
  "bytes 0.4.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.23 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper 0.11.27 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper-proxy 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper-rustls 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-core 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1119,6 +1137,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "fnv"
 version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "foreign-types"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "foreign-types-shared 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "foreign-types-shared"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -1293,6 +1324,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper-proxy"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bytes 0.4.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures 0.1.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper 0.11.27 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper-tls 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "native-tls 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-core 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-io 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-tls 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "hyper-rustls"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1307,6 +1353,20 @@ dependencies = [
  "tokio-rustls 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "tokio-service 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "webpki-roots 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "hyper-tls"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "futures 0.1.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "hyper 0.11.27 (registry+https://github.com/rust-lang/crates.io-index)",
+ "native-tls 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-core 0.1.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-io 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-service 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-tls 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1849,6 +1909,20 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "native-tls"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "lazy_static 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl 0.9.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "schannel 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "security-framework 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
+ "security-framework-sys 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tempdir 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "net2"
 version = "0.2.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


kavu:
[clabot:check]

dvdplm:
Welcome, and thank you for your contribution! :)

The Parity configuration options are passed as command line switches (e.g. `--warp-barrier=123`) or using the toml config file. We do not use `ENV` variables for config. I do see your point about `HTTPS_PROXY` being an informal standard of sorts so perhaps we can accept an exception, but the "normal" config paths must also be supported (and preferred): first for a `--https_proxy` command line switch, then check config file for `https_proxy`; lastly check if the env var is set.

As for the feature per sé, can you help me understand why you need `parity` to have the proxy functionality built-in, as opposed to using an external proxy (something like [Tinyproxy](https://tinyproxy.github.io/) perhaps?)?

dvdplm:
I understand your use case much better now, thank you for explaining. :)

This is something we want to support, and as mentioned above it'd be good to make the proxy option configurable in the same way other network options are configured (config file, overridden by cli options). Take a look at `run.rs` where the `fetch::Client` is instantiated and try to make the changes to use a `ProxyConnector`. The config options should probably go in the `NetworkConfiguration` struct.

kavu:
@dvdplm  Thanks, David, for pointing it out!

kavu:
@dvdplm Yeah, I am still on it! Thanks. I will inform you when it's be ready for review. BTW, I think I have no permissions to add or remove labels :)

dvdplm:
> BTW, I think I have no permissions to add or remove labels :)

Ah, ok sorry about that. Let us know in a comment then when you're ready for a review! :)

5chdn:

kavu:
@5chdn Yeah, sorry for my delay. I want to wrap it up this weekend.

Does it slow down some release pipeline? :(

5chdn:
No :)

ethcore/sync/src/api.rs
```diff
@@ -647,6 +647,8 @@ pub struct NetworkConfiguration {
 	pub ip_filter: IpFilter,
 	/// Client version string
 	pub client_version: String,
+	/// HTTPS Proxy
+	pub https_proxy: Option<String>,
```
kavu: I am not quite sure why there are two (here and https://github.com/paritytech/parity-ethereum/pull/9303/files#diff-034858a645d5e2357f9460db644109d3R207) NetworkConfiguration structs in the project, but I've added the fields to both.

Also, I am not sure if it really should be Option.

parity/helpers.rs
```diff
@@ -298,6 +300,12 @@ pub fn password_prompt() -> Result<Password, String> {
 	Ok(password)
 }
 
+pub fn new_fetch_client(https_proxy: Option<&String>) -> Result<fetch::Client, String> {
+	return https_proxy
+		.map_or_else(|| fetch::Client::new(), |s| fetch::Client::new_with_proxy(Some(s.to_string())))
```
kavu: Pretty sure that `Some(s.to_string())` is not good a practice, but could find a way to do this idiomatically.

parity/run.rs
```diff
@@ -241,6 +241,9 @@ fn execute_light_impl(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<Runnin
 		None
 	};
 
+	// fetch service
+	let fetch = new_fetch_client(net_conf.https_proxy.as_ref())?;
```
kavu: This and https://github.com/paritytech/parity-ethereum/pull/9303/files#diff-05d235380ad5b27b1e5ee6a84511a5e8R472 gave me a really hard time fighting borrow checkers. The only way I find to do this is to move `mut net_conf` and fetch creation based on `net_conf` a little bit earlier, before others moves occur.

util/fetch/src/client.rs
```diff
@@ -169,12 +170,42 @@ impl Drop for Client {
 }
 
 impl Client {
+	/// Create a new fetch client with a proxy.
+	pub fn new_with_proxy(https_proxy_url: Option<String>) -> Result<Self, Error> {
```
kavu: A bit of code duplication. Is there a better way?

The problem I see here is the background_thread that is started during Client creation, and we can't add `set_proxy()` here, because `https_connector` with a [possible] proxy should be already created, during the thread start.

That's why I've added a separate factory method, to build a Client, but passing a Proxy URL into the Thread creation function.

kavu:
@dvdplm @5chdn Phew! I think I've done it. I am still not quite sure if this patch can pass code quality check, but I am looking forward to tune it accordingly. Thanks!

Cargo.lock
```diff
@@ -1805,6 +1865,20 @@ name = "nan-preserving-float"
...
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[[package]]
+name = "native-tls"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "lazy_static 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl 0.9.24 (registry+https://github.com/rust-lang/crates.io-index)",
```
niklasad1: This dependency will require ```OpenSSL``` and is by default dynamically linked! For a while ago we got rid of ```OpenSSL``` thus, we need to update all docker images and similar!

/cc @kirushik 
Do you have any thoughts about this regarding the binary size etc?

kavu: Yeah, I agree it really can be a blocker here. I'll try to find out what can be done here. It's interesting that `hyper` on it own does not require OpenSSL to connect over HTTPS. But `hyper-proxy` seems to require it somehow. Strange.

ethcore/sync/src/api.rs
niklasad1: I think it makes sense because it is/should be optional to use a ```https proxy```

parity/helpers.rs
niklasad1: It is fine here because, in the end, you need to take ownership of the String in the ```background thread (closure)```.
However, you could pass to the `String` by reference instead and clone when you need it! A String in Rust is a ```Vec<u8>``` which takes 3 words on the stack and the rest on the heap!

However, remove needless `return` and the ending `;`

parity/run.rs
```diff
@@ -468,8 +469,8 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 	// spin up event loop
 	let event_loop = EventLoop::spawn();
 
-	// fetch service
-	let fetch = fetch::Client::new().map_err(|e| format!("Error starting fetch client: {:?}", e))?;
+	let mut net_conf = cmd.net_conf;
+	let fetch = new_fetch_client(net_conf.https_proxy.as_ref())?;
```
niklasad1: ```new_fetch_client(cmd.net_conf.https_proxy.as_ref())``` should be sufficient here because `new_fetch_client()` returns a value not a reference

Then we can have smaller scope for the ```let mut net_conf``` binding (i.e, as it was before)

util/fetch/src/client.rs
```diff
@@ -199,16 +230,34 @@ impl Client {
...
 		thread::Builder::new().name("fetch".into()).spawn(move || {
 			let mut core = match reactor::Core::new() {
 				Ok(c) => c,
 				Err(e) => return tx_start.send(Err(e)).unwrap_or(())
 			};
 
+			let connector = {
+				let https_connector = hyper_rustls::HttpsConnector::new(4, &core.handle());
+
+				match https_proxy_url.unwrap_or_default().parse() {
```
niklasad1: I don't like ```unwrap_or_default``` and rely on that an empty String will not be parsed as a SockAddr.

I would much prefer something like:
```rust

				match https_proxy_url.map(|urll| url.parse()) {
					Some(Ok(proxy_url)) => {
						debug!(target: "fetch", "https proxy is {:?}", proxy_url);
						ProxyConnector::from_proxy(
							https_connector,
							Proxy::new(Intercept::Https, proxy_url),
						).unwrap()
					}
					_ => {
						debug!(target: "fetch", "https proxy is not set");
						ProxyConnector::new(https_connector).unwrap()
					}
				}

```

util/fetch/src/client.rs
```diff
@@ -169,12 +170,42 @@ impl Drop for Client {
...
 
 impl Client {
+	/// Create a new fetch client with a proxy.
+	pub fn new_with_proxy(https_proxy_url: Option<String>) -> Result<Self, Error> {
+		let (tx_start, rx_start) = std::sync::mpsc::sync_channel(1);
+		let (tx_proto, rx_proto) = mpsc::channel(64);
+
+		Client::background_thread(tx_start, rx_proto, https_proxy_url)?;
+
+		match rx_start.recv_timeout(Duration::from_secs(10)) {
```
niklasad1: consider changing ```Duration::from_secs(10)``` to a constant instead with something like ```const MAX_FETCH_TIME: Duration = Duration::from_secs(10);```

util/fetch/src/client.rs
kavu: I decided to add new `new_with_proxy` because changing `new` definition will break current usages, e.g. in tests. But if you think it's ok, then I'll gladly change it.

util/fetch/src/client.rs
kavu: Got it! Thanks!

util/fetch/src/client.rs
kavu: Got it!

kavu:
@niklasad1 Added some fixes as you asked, hope it looks better now.

kavu:
@niklasad1 Speaking about `OpenSSL`. I've looked on `hyper-proxy` dependencies and found that it requires `tokio-tls` and `hyper-tls`, which both require `native-tls`. And the last one of course requires many other stuff, to deliver TLS no matter what system an application builds for. 

So, I guess this issue will block my PR, as I (or somebody else) need to implement `tokio-rustls`, `hyper-rustls` and after that `hyper-proxy-rustls` 😞 

Maybe, there is some other way, but I am not see it yet.

util/fetch/src/client.rs
```diff
@@ -169,14 +171,44 @@ impl Drop for Client {
...
+		})
+	}
+
 	/// Create a new fetch client.
 	pub fn new() -> Result<Self, Error> {
 		let (tx_start, rx_start) = std::sync::mpsc::sync_channel(1);
 		let (tx_proto, rx_proto) = mpsc::channel(64);
 
-		Client::background_thread(tx_start, rx_proto)?;
+		Client::background_thread(tx_start, rx_proto, String::default())?;
```
niklasad1: ```Client::background_thread(tx_start, rx_proto, &None)?;```

util/fetch/src/client.rs
```diff
@@ -199,16 +231,34 @@ impl Client {
 		})
 	}
 
-	fn background_thread(tx_start: TxStartup, rx_proto: mpsc::Receiver<ChanItem>) -> io::Result<thread::JoinHandle<()>> {
+	fn background_thread(tx_start: TxStartup, rx_proto: mpsc::Receiver<ChanItem>, https_proxy_url: String) -> io::Result<thread::JoinHandle<()>> {
```
niklasad1: ```https_proxy_url: &Option<String>```

Then you must clone ```https_proxy_url``` and move the `clone` into to the thread closure

util/fetch/src/client.rs
```diff
@@ -199,16 +231,34 @@ impl Client {
...
 		thread::Builder::new().name("fetch".into()).spawn(move || {
 			let mut core = match reactor::Core::new() {
 				Ok(c) => c,
 				Err(e) => return tx_start.send(Err(e)).unwrap_or(())
 			};
 
+			let connector = {
+				let https_connector = hyper_rustls::HttpsConnector::new(4, &core.handle());
+
+				match https_proxy_url.parse() {
```
niklasad1: use map().parse() as I suggested earlier i.e, if the String is `None` then directly indicate an error instead of creating an empty String (default)

parity/helpers.rs
```diff
@@ -298,6 +300,12 @@ pub fn password_prompt() -> Result<Password, String> {
 	Ok(password)
 }
 
+pub fn new_fetch_client(https_proxy: Option<&String>) -> Result<fetch::Client, String> {
+	https_proxy
```
niklasad1: if you change `fetch::Client::new(https_proxy: &Option<String>)` and remove `fetch::Client::new_with_proxy(https_proxy_url: &Option<String>)`

Then you can do:

```rust
pub fn new_fetch_client(https_proxy: &Option<String>) -> Result<fetch::Client, String> {
       fetch::Client::new(https_proxy)
           .map_err(|e| format!("Error starting fetch client: {:?}", e))
 }
```

However, it requires some changes by the caller of this method ofc!

niklasad1:
@kavu 
Thanks, the code looks good but now we **"only"** need to decide how the fix the ```OpenSSL``` dependency!

kirushik:
> we "only" need to decide how the fix the OpenSSL dependency!

Speaking of which, I would prefer not to reintroduce OpenSSL if that's avoidable.
It was a PITA to link with `libssl` for all platforms at once from our build pipeline in the past — I don't want that complexity to be reintroduced, if avoidable.

5chdn:
Please reopen once decided how to proceed.

Issue #9288, opened 2018-08-06
`pending` for defaultBlock causes issues during estimateGas


iFA88:
Greetings!

I have tried the following nodes:
`Parity-Ethereum/v2.0.0-beta-6eae372-20180717/x86_64-linux-gnu/rustc1.27.1`
`Parity-Ethereum/v2.0.1-beta-e7dc0be-20180726/x86_64-linux-gnu/rustc1.27.2`

where I have the same issue for `eth_estimateGas` when I use `pending` for defaultBlock parameter:
```
-> {"jsonrpc": "2.0", "params": [{"to": "0xa8711bf8f25e9d87498777e4b3b651f6e811cc7d", "from": "0xa8711bf8f25e9d87498777e4b3b651f6e811cc7d", "value": "0x2386f26fc10000", "data": "0x"}, "pending"], "method": "eth_estimateGas", "id": 1}
<- {"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":1}
```

This happens not every time, about 1/5 times. Maybe it happens during the block import process.

When it works then I receive:
```
-> [{"to": "0xa8711bf8f25e9d87498777e4b3b651f6e811cc7d", "from": "0xa8711bf8f25e9d87498777e4b3b651f6e811cc7d", "value": "0x2386f26fc10000", "data": "0x"}, "pending"], "method": "eth_estimateGas", "id": 1}
<- {"jsonrpc":"2.0","result":"0x5208","id":1}
```

Tbaut:
I could reproduce the problem.
>If I use latest they produces the same error.

I did **not** reproduce with "latest" however. In general @iFA88, for a non mining node, "pending" will be the same as "latest" since your node is not creating any block.

Tbaut:
Can you have a look @tomusdrw ?

iFA88:
> non mining node

iFA88:
I have forgot to say that I use the following parameters:
```
--no-persistent-txqueue --jsonrpc-server-threads 8 --no-ancient-blocks --min-gas-price=10000000 --tx-queue-mem-limit=4096 --tx-queue-size=256000 --reseal-on-txs=all --force-sealing --rpcport 8548 --port 30306 --no-ws --no-secretstore --cache-size 2048
```

Tbaut:
[...]

PR #9310, opened 2018-08-08
Prevent blockchain & miner racing when accessing pending block.

tomusdrw:
The issue is that `client` might have imported a new block already, but `on_chain_new_blocks` notification has not been run yet (so new pending block was not yet constructed). This is even more visible now, when pending block construction may take significant amount of time (hundreds millis).
So the miner has a pending block with the same number as the best_block returned from `client`.

We could either:
1. Return a bit stale data (implemented in this PR)
2. Force miner to create a new pending block (I would avoid that, we don't really want to create pending blocks from RPC threads, since it's quite heavy)
3. Wait for the new pending block (if we know we run with `--force-sealing` or sth). This would required to have some kind of lock on the pending data. (I would also avoid that to prevent locking down the RPC cpupool).

Since the inherent nature of `pending` is that it's ephemeral (and the data might actually be invalid by the time you receive a response) I believe that returning stale data is the least evil approach.

Should address #9288

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -319,14 +319,15 @@ impl Miner {
 	/// Retrieves an existing pending block iff it's not older than given block number.
 	///
 	/// NOTE: This will not prepare a new pending block if it's not existing.
-	/// See `map_pending_block` for alternative behaviour.
 	fn map_existing_pending_block<F, T>(&self, f: F, latest_block_number: BlockNumber) -> Option<T> where
 		F: FnOnce(&ClosedBlock) -> T,
 	{
 		self.sealing.lock().queue
 			.peek_last_ref()
 			.and_then(|b| {
-				if b.block().header().number() > latest_block_number {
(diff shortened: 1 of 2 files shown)


tomusdrw:
I've noticed another possible issue that causes incorrect behaviour and will attempt to fix it now.

tomusdrw:
Logs in #9288 indicated on another reason why it may be failing.
During reseal the pending block was actually `popped` of the queue, so it was no longer available. Added a fix for this, changing the semantics of `using_queue` a bit:
1. `pop_if` is replaced with `get_pending_if` (we always clone)
2. `push` was renamed to `set_pending`

So now we can push to an old block, but it's previous pending version is still available for the queries.

tomusdrw:
@dvdplm yup, I mean exactly this, where we already imported a block with number `N` to the blockchain, but we have a pending block `N` that might be a little bit different though (but it's the same block that has been returned for all previous queries).

debris:
does it need to backported?

PR #9315, opened 2018-08-08
Allow tx pool to be Send

arkpar:
This allows using transaction pool in `Arc` without specifying bounds all over the high level code.


dvdplm:
> We should probably move the pool to parity-common

Good idea: https://github.com/paritytech/parity-common/issues/19

PR #9316, opened 2018-08-08
ethcore/sync `Make view macro only visible to test`

niklasad1:
Fixes the following warning:

```bash
warning: unused `#[macro_use]` import                                                                                                                                                                                                          
  --> ethcore/sync/src/lib.rs:29:1                                                                                                                                                                                                             
   |                                                                                                                                                                                                                                           
29 | #[macro_use]                                                                                                                                                                                                                              
   | ^^^^^^^^^^^^                                                                                                                                                                                                                              
   |                                                                                                                                                                                                                                           
```

### DIFF
--- a/ethcore/sync/src/lib.rs
+++ b/ethcore/sync/src/lib.rs
@@ -26,6 +26,7 @@ extern crate ethcore_network_devp2p as devp2p;
 extern crate parity_bytes as bytes;
 extern crate ethcore_io as io;
 extern crate ethcore_transaction as transaction;
+#[cfg_attr(test, macro_use)]
 extern crate ethcore;
 extern crate ethereum_types;
 extern crate env_logger;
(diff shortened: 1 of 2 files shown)


ethcore/sync/src/lib.rs
```diff
@@ -26,7 +26,6 @@ extern crate ethcore_network_devp2p as devp2p;
 extern crate parity_bytes as bytes;
 extern crate ethcore_io as io;
 extern crate ethcore_transaction as transaction;
-#[macro_use]
```
ordian: Looks like this macro is used only in tests, so replacing this line with `#[cfg_attr(test, macro_use)] ` should do the trick.
see https://github.com/rust-lang/rust/issues/44342.

niklasad1: Thanks :+1:

PR #9318, opened 2018-08-08
Allow dropping light client RPC query with no results

cheme:
This PR helps with #9133 .

I did observe infinite requesting when sending rpc call like getTransactionByHash on the light client.

The initial problem is that the light client is going to query random peers indefinitely until it got a reply (the way the communication is done closing the query do not end this).
In the case of a query with incorrect hash it will just query forever random peers.

In this PR I change this by limiting the number of query to one for each peer, before failing.
It leaves a few open questions:

- Should we also use a timeout mechanism for the case where there are not many peers? This PR do break the `no_capability` test, with a timeout mechanism the test would make sense again.
- When failing I use the `Cancelled` error resulting in `{"jsonrpc":"2.0","error":{"code":-32603,"message":"Internal error occurred: on-demand sender cancelled","data":"\"\""},"id":1}`, is it acceptable?
- If there is new peers or removal of peers during the process some peers may not be queried or queried twice (the hashmap iterator keep its order so it is not that bad). Obviously, I can store the previous query peerids but I am not sure it is worth it.
- When testing on ropsten, there are definitely very few peers that can reply (one with full history). When we got a OnDemand query that passes for a peer and not for others, should we exclude those peers from the OnDemand list of peers?


ethcore/light/src/on_demand/mod.rs
```diff
@@ -81,6 +82,9 @@ struct Pending {
 	required_capabilities: Capabilities,
 	responses: Vec<Response>,
 	sender: oneshot::Sender<Vec<Response>>,
+	first_query: Option<usize>,
+	nb_query: usize,
```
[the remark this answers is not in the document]
cheme: Yes not explicit enough: I can switch to : 
- first_query : base_query_index
- nb_query : total_query_count
- rem_query : remaining_query_count

ethcore/light/src/on_demand/mod.rs
```diff
@@ -180,6 +185,14 @@ impl Pending {
 		self.net_requests = builder.build();
 		self.required_capabilities = capabilities;
 	}
+
+	// return no response, will result in an error
+	// consume object (similar to drop when no reply found)
```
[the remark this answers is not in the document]
cheme: Best comment (object is definitely a bad wording for rust) : 
// returning no reponse, it will result in an error.
// self is consumed on purpose.
The comment is just here to point out that we do not want self to be use back (used oneshot channel).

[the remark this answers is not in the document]
dvdplm: Yep, I think the comment is useful here. I like your suggestion for an update.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -145,7 +149,8 @@ impl Pending {
 	// if the requests are complete, send the result and consume self.
 	fn try_complete(self) -> Option<Self> {
 		if self.requests.is_complete() {
-			let _ = self.sender.send(self.responses);
+			self.sender.send(self.responses).map_err(|_|())
+				.expect("Non used one shot channel");
```
rphmeier: i don't think it should panic here

ethcore/light/src/on_demand/mod.rs
```diff
@@ -180,6 +185,14 @@ impl Pending {
 		self.net_requests = builder.build();
 		self.required_capabilities = capabilities;
 	}
+
+	// returning no reponse, it will result in an error.
+	// self is consumed on purpose.
+	fn no_response(self) {
+		trace!(target: "on_demand", "Dropping a pending query (no reply)");
+		self.sender.send(Vec::with_capacity(0)).map_err(|_|())
+			.expect("Non used one shot channel");
```
rphmeier: I think this will panic if the receiving end has hung up. better to ignore/handle the error than panic

ethcore/light/src/on_demand/mod.rs
```diff
@@ -240,7 +253,16 @@ impl<T: request::RequestAdapter> Future for OnResponses<T> {
 	type Error = Canceled;
 
 	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-		self.receiver.poll().map(|async| async.map(T::extract_from))
+		match self.receiver.poll() {
+			Ok(Async::Ready(v)) => {
+				if v.len() == 0 {
```
rphmeier: `is_empty`

rphmeier:
I'm not sure the logic in this PR is robust against the amount of peers changing over time, and it still seems that it may request from the same peer in that case. Why not just alter `Pending` to have a `attempts_remaining: usize` and a `attempted: HashSet<PeerId>`. When `attempts_remaining == 0` then return nothing. We can make the `attempts_remaining` initial value configurable by CLI or RPC. A reasonable default would be 10

cheme:
Hi Ralph, I am totally sure it is not robust against amount of peer changing (the third point in my initial PR).
I did not want to store queried peer, if you believe it is not an issue to store it for every OnDemand request it would certainly be more robust.

rphmeier:
Storing a `HashSet` of queried peers per request is probably fine, but if it turns out to be a performance issue in the future we can find some other heuristic.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -49,7 +50,14 @@ pub mod request;
 /// The result of execution
 pub type ExecutionResult = Result<Executed, ExecutionError>;
 
+/// The default number of retry for OnDemand queries send to other nodes
+pub const DEFAULT_NB_RETRY: usize = 10;
+
+/// The default time limit in milliseconds for inactive (no new peer to connect to) OnDemand queries (0 for unlimited)
+pub const DEFAULT_QUERY_TIME_LIMIT: u64 = 10000;
```
niklasad1: Use `const Duration` here instead!

ethcore/light/src/on_demand/mod.rs
```diff
@@ -49,7 +50,14 @@ pub mod request;
 /// The result of execution
 pub type ExecutionResult = Result<Executed, ExecutionError>;
 
+/// The default number of retry for OnDemand queries send to other nodes
```
niklasad1: The default number of retries for OnDemand queries to send to the other nodes

ethcore/light/src/on_demand/mod.rs
```diff
@@ -145,7 +157,9 @@ impl Pending {
 	// if the requests are complete, send the result and consume self.
 	fn try_complete(self) -> Option<Self> {
 		if self.requests.is_complete() {
-			let _ = self.sender.send(self.responses);
+			if self.sender.send(self.responses).is_err() {
+				debug!(target: "on_demand", "Dropped oneshot channel receiver on complet request");
```
niklasad1: complete?

parity/configuration.rs
```diff
@@ -205,7 +205,7 @@ impl Configuration {
 			};
 			Cmd::Account(account_cmd)
 		} else if self.args.flag_import_geth_keys {
-        	let account_cmd = AccountCmd::ImportFromGeth(
+					let account_cmd = AccountCmd::ImportFromGeth(
```
niklasad1: Fix indentation!

parity/configuration.rs
```diff
@@ -1331,10 +1333,10 @@ mod tests {
...
 			max_connections: 100,
 		}, LogConfig {
-            color: true,
-            mode: None,
-            file: None,
-        } ));
+						color: true,
+						mode: None,
+						file: None,
+				} ));
```
niklasad1: Fix indentation!

parity/cli/mod.rs
```diff
@@ -1364,12 +1374,21 @@ struct Whisper {
...
 
+#[derive(Default, Debug, PartialEq, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct Light {
+	on_demand_nb_retry: Option<usize>,
+	on_demand_inactive_time_limit: Option<u64>,
+}
+
+
+
```
niklasad1: Stray duplicate empty lines, one should be enough!

cheme:
The limit is here to avoid querying peer twice (in case of low number of peers connected). It is mainly related to the usecase where we query a transaction by hash with a hash that does not exists : with only a timeout we will query on many peers (or the same peers if not many connected peers) until timeout.

There is certainly something to do to remove peers (from OnDemand peer storage) that do not reply correctly (especially in the use case where we got a reply from a peer but other peers having capability for the blockhash do not reply), but it needs some investigation (my main testing case is getting a transaction by hash and some peers do not reply when the hash is in their capability range); it is probably not a binary decision.
I think a priority queue with a credit management and an historic of bad reply may be interesting.

5chdn:
What's the status of this?

parity/cli/mod.rs
```diff
@@ -1364,12 +1374,19 @@ struct Whisper {
 	pool_size: Option<usize>,
 }
 
+#[derive(Default, Debug, PartialEq, Deserialize)]
```
niklasad1: Make this `Copy` and `Clone`?

ethcore/light/src/on_demand/tests.rs
```diff
@@ -180,7 +180,7 @@ fn no_capabilities() {
 
 	harness.inject_peer(peer_id, Peer {
 		status: dummy_status(),
-		capabilities: capabilities,
+		capabilities: capabilities.clone(),
```
niklasad1: You can remove `clone` here `capabilities` is Copy!

rpc/src/v1/helpers/errors.rs
```diff
@@ -444,7 +445,41 @@ pub fn filter_block_not_found(id: BlockId) -> Error {
 	}
 }
 
+pub fn on_demand_error(err: OnDemandError) -> Error {
+	match err.kind() {
+		OnDemandErrorKind::ChannelCanceled(ref e) => return on_demand_cancel(e.clone()),
```
niklasad1: Needless `return`

Also, is it possible to take ownership of the `future::oneshot::Canceled´ instead and not clone it?

rphmeier: Should be possible with a `match err { Error(OnDemandErrorKind::ChannelCanceled(e), _) =>  e`

ethcore/light/src/on_demand/mod.rs
```diff
@@ -366,30 +447,68 @@ impl OnDemand {
...
-				let rng = rand::random::<usize>() % cmp::max(num_peers, 1);
-				for (peer_id, peer) in peers.iter().chain(peers.iter()).skip(rng).take(num_peers) {
+				let history_len = pending.query_id_history.len();
+				let start = if history_len == 0 {
+					pending.remaining_query_count = self.base_retry_number;
+					let rand = rand::random::<usize>();
+					pending.base_query_index = rand;
+					rand
+				} else {
+					pending.base_query_index + history_len
```
rphmeier: perhaps make this an explicit wrapping add?

ethcore/light/src/on_demand/mod.rs
```diff
@@ -366,30 +447,68 @@ impl OnDemand {
...
+				} % cmp::max(num_peers, 1);
```
rphmeier: if we lose peers along the way we might end up re-querying some of them. but generally the peer set will be decently stable so maybe not the worst thing

ethcore/light/src/on_demand/mod.rs
```diff
@@ -366,30 +447,68 @@ impl OnDemand {
...
+					pending.no_response();
+					None
+				} else if init_remaining_query_count == pending.remaining_query_count {
+					if let Some(query_inactive_time_limit) = self.query_inactive_time_limit {
+						let now = SystemTime::now();
+						if pending.inactive_time_limit.is_none() {
+							debug!(target: "on_demand", "No more peer to query, waiting for {} seconds until dropping query", query_inactive_time_limit.as_secs());
+							pending.inactive_time_limit = Some(now + query_inactive_time_limit);
+						} else {
+							if now > pending.inactive_time_limit.unwrap() {
```
rphmeier: we don't allow `unwrap` in the codebase. You should use `expect` with a proof that it is not going to panic. Maybe rewrite this block with an `if let Some(x) = pending.inactive_time_limit`?

rpc/src/v1/impls/light/eth.rs
```diff
@@ -216,7 +216,7 @@ impl<T: LightChainClient + 'static> EthClient<T> {
 								};
 
 								fill_rich(block, score)
-							}).map_err(errors::on_demand_cancel)),
+							}).map_err(errors::on_demand_error)),
```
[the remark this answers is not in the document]
cheme: I resolve merge conflict really badly previously, my fault.

ethcore/light/src/on_demand/mod.rs
cheme: I also did a test on hashset insertion a few line after, so it should be fine.

ethcore/light/src/on_demand/mod.rs
niklasad1: I think Rob is referring ```<integer>::wrapping_add()```, see https://doc.rust-lang.org/std/primitive.usize.html#method.wrapping_add

cheme: Oh, it's nice.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -74,13 +111,22 @@ impl Peer {
 	}
 }
 
+
+/// Either an array of response or a single error
+/// Currently fails globally.
```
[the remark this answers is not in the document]
cheme: The PR starts to be a bit old, I believe it meant that the query is dropped on error; I am removing this line.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -177,6 +225,25 @@ impl Pending {
...
+			debug!(target: "on_demand", "Dropped oneshot channel receiver on no response");
+		}
+	}
+
+	// returning a peer discovery timeout during queries attempts
+	fn time_out(self) {
+		trace!(target: "on_demand", "Dropping a pending query (no new peer time out)");
+		let err = self::error::ErrorKind::TimeoutOnNewPeers(self.requests.num_answered(), self.query_id_history.len());
+		if self.sender.send(Err(err.into())).is_err() {
+			debug!(target: "on_demand", "Dropped oneshot channel receiver on no response");
```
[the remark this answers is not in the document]
cheme: Oups, totally.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -363,30 +444,68 @@ impl OnDemand {
...
 					}
 
-					match ctx.request_from(*peer_id, pending.net_requests.clone()) {
-						Ok(req_id) => {
-							trace!(target: "on_demand", "Dispatched request {} to peer {}", req_id, peer_id);
-							self.in_transit.write().insert(req_id, pending);
-							return None
+					if pending.query_id_history.insert(peer_id.clone()) {
+
+						if !peer.can_fulfill(&pending.required_capabilities) {
```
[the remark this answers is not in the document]
cheme: We will keep the query in pending (returning Some(Pending)) at the end of the iteration.
Next call we will take the line 492 branch (no peer queried waiting delay for a new peer or drop if delay finished).
I realize that my count of query in trace can be a bit off (because peer with no capabilities are in the hash set), but fortunatly I do the number of query test on remaining_query_count.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -406,6 +525,21 @@ impl OnDemand {
...
 	}
+
+	/// Changes default number of retry for query.
+	pub fn default_retry_number(&mut self, nb_retry: usize) {
+		self.base_retry_number = nb_retry;
+	}
+
+	/// Changes default time limit for query.
+	pub fn query_inactive_time_limit(&mut self, inactive_time_limit: Duration) {
+		self.query_inactive_time_limit = if inactive_time_limit == Duration::new(0,0) {
```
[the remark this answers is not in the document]
cheme: There is Duration::Default() that could do it but I find it misleading, I will use a constant.

parity/cli/mod.rs
```diff
@@ -562,6 +562,15 @@ usage! {
...
 			"Specify CORS header for IPFS API responses. Special options: \"all\", \"none\".",
 
+		["Light Client Options"]
+			ARG arg_on_demand_nb_retry: (Option<usize>) = None, or |c: &Config| c.light.as_ref()?.on_demand_nb_retry,
+			"--on_demand_nb_retry=[RETRIES]",
+			"Specify maximum number of retry query to send to other node for a query.",
+
+			ARG arg_on_demand_inactive_time_limit: (Option<u64>) = None, or |c: &Config| c.light.as_ref()?.on_demand_inactive_time_limit,
+			"--on_demand_inactive_time_limit=[MS]",
+			"Specify maximum total time limit for a light client query to stay inactive. O for no limit.",
```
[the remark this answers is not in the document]
cheme: It is currently in MilliSeconds ([MS]), and yes 0 is no limit wait forever; it is not that unsafe because it is combined with the maximum number of retries (at first I implement it without any timeout).

parity/cli/mod.rs
```diff
@@ -562,6 +562,15 @@ usage! {
 			"--ipfs-api-cors=[URL]",
 			"Specify CORS header for IPFS API responses. Special options: \"all\", \"none\".",
 
+		["Light Client Options"]
+			ARG arg_on_demand_nb_retry: (Option<usize>) = None, or |c: &Config| c.light.as_ref()?.on_demand_nb_retry,
```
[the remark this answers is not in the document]
cheme: Yes I miss the - usage, it is major (displayed in parity --help).

parity/cli/mod.rs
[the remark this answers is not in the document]
cheme: Here it is not a hard timelimit, it is really an inactive time (if new peers are added regularly the request could run for way longer than this value). Inactive means that there was no new query for some time, if I got a new query in between the inactive time is reset to 0.

ethcore/light/src/on_demand/mod.rs
```diff
@@ -74,13 +113,21 @@ impl Peer {
...
 struct Pending {
 	requests: basic_request::Batch<CheckedRequest>,
 	net_requests: basic_request::Batch<NetworkRequest>,
 	required_capabilities: Capabilities,
 	responses: Vec<Response>,
-	sender: oneshot::Sender<Vec<Response>>,
+	sender: oneshot::Sender<PendingResponse>,
+	base_query_index: usize,
+	remaining_query_count: usize,
+	query_id_history: BTreeSet<PeerId>,
```
ordian: Any reason to use `BTreeSet` instead of `HashSet`? We don't rely on ids order as it seems, so `HashSet` should more appropriate maybe?

cheme: I expect it to be a small set, and I generally favor a btreeset in those case.

cheme:
Ordian, thanks for pointing out `failsafe-rs`, it seems like an interesting crate. At the time I was considering doing a very small change to fix it (not even a timeout), the other configurations were added afterward.

rphmeier:
Failsafe-rs seems like a fantastic way to proceed. Ideally we can make things like # of attempts or timeouts user-configurable. Although I would place that as a refactoring beyond the scope of this PR.

Issue #9040, opened 2018-07-04
Nonce increment when transaction failed


gituser:
> I'm running:
>
> - **Which Parity version?**: 1.10.8
> - **Which operating system?**: Linux
> - **How installed?**: from source
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum 
> - **Did you try to restart the node?**: yes

**Summary**
When you try to send a transaction with gas less than 21000 or (if you specify amount of the transaction bigger than your actual address balance) it fails, but nonce still increments, thus all future transactions will be forever in the queue, until someone fills the nonce gap between that last successful transaction and the one in the queue or restarts the parity.

**Here is how to reproduce:**
1) create some transaction without nonce (thus nonce will be picked up, let's say `nonce` is 10)
2) next, create a transaction with gas=20000, you'll get this error:
```
Error: Transaction gas is too low. There is not enough gas to cover minimal cost of the transaction (minimal: 21000, got: 20000). Try increasing supplied gas.
```
3) now try creating another transaction without specifying the nonce, `nonce` will be 12 and thus transaction will stay in the pool, until you send transaction with `nonce=11`

I think nonce shouldn't be incremented if transaction has failed due to low gas or any other error when there is no resulting hash.

gituser:
The issue here is not about gas amount, but rather about nonce increment. Why parity does increment the nonce when in fact there was no previous transaction?

gituser:
Also, `nonce` is being incremented if you try to send transaction which exceeds your current account balance. That means that after that all future transactions will be stuck as well.

seunlanlege:
can't replicate this on kovan.

Tbaut:
@gituser please provide a concrete example (with rpc or a web3 script) that could allow us to reproduce the behavior? Happy to reopen then.

gituser:
You want me to provide you with RPC call how to send transaction? It's all described in the issue description, re-read it.

Simply try to send transaction with gas=20000 or specify value bigger than your current balance and then try to create another transaction with good value, it will never be picked up, because there will be a gap in nonce.

Why close the issue without trying to reproduce it? I've written exact steps how to reproduce the issue.

gituser:
Also, I've just tested on newest parity release and the bug is still there on Ropsten, I'm pretty sure on Kovan it should be there too.

Tbaut:
[...]

PR #9319, opened 2018-08-08
EIP 1283: Net gas metering for SSTORE without dirty maps

sorpaas:
This implements [EIP-1283](https://eips.ethereum.org/EIPS/eip-1283).

This implementation only modifies EVM gas costs, wasm gas costs is unaffected by `eip1283Transition` flag.


ethcore/evm/src/tests.rs
```diff
@@ -716,7 +716,7 @@ fn test_jumps(factory: super::Factory) {
 		test_finalize(vm.exec(&mut ext)).unwrap()
 	};
 
-	assert_eq!(ext.sstore_clears, 1);
+	assert_eq!(ext.sstore_clears, U256::from(15_000));
```
cheme: ext.schedule().sstore_refund_gas instead of 15000

ethcore/src/executive.rs
```diff
@@ -184,18 +185,20 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 			schedule: schedule,
 			depth: 0,
 			static_flag: false,
+			transaction_checkpoint_index: None,
```
cheme: This needs a bit of comment I think. From what I understand it is to manage reentrant call by addressing the first checkpoint initial values; if it is the case renaming to `first_transaction_checkpoint_index` also makes sense.

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,49 @@ impl<B: Backend> State<B> {
 			|a| a.as_ref().and_then(|account| account.storage_root().cloned()))
 	}
 
-	/// Mutate storage of account `address` so that it is `value` for `key`.
-	pub fn storage_at(&self, address: &Address, key: &H256) -> TrieResult<H256> {
+	/// Get the value of storage at last checkpoint.
```
cheme: is 'at last' compatible with `checkpoint_index` parameter.

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,49 @@ impl<B: Backend> State<B> {
...
+	pub fn checkpoint_storage_at(&self, checkpoint_index: usize, address: &Address, key: &H256) -> TrieResult<Option<H256>> {
+		if let Some(ref checkpoint) = self.checkpoints.borrow().get(checkpoint_index) {
+			Ok(Some(if let Some(entry) = checkpoint.get(address) {
+				match entry {
+					Some(entry) => {
+						match entry.account {
+							Some(ref account) => {
+								if let Some(value) = account.cached_storage_at(key) {
```
cheme: I don't get this, the function is only used in `reverted_storage_at` to initiate `origin` value of the storage key, so why are we looking for the new value value of the key (`cached_storage_at` query first the storage changes). Maybe it is `cached_original_storage_at` that was meant to be called?

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,49 @@ impl<B: Backend> State<B> {
...
+							},
+						}
+					},
+					None => {
+						// The value was not cached at that checkpoint, meaning it was not modified.
+						self.original_storage_at(address, key)?
+					},
+				}
+			} else {
+				// This key does not have a checkpoint entry. We use the current value.
```
cheme: Is it possible here that the checkpoint do not have the value because of the indexed checkpoint mechanism (I am unsure about the indexed checkpoint mechanism seeing that it was added in a later commit, I am wondering), leading to issues?

cheme:
Hello sorpaas, I started to review your code but I'am a bit to unsure about the `current_checkpoint_index`. So I just left you a few comments.

ethcore/src/state/mod.rs
sorpaas: The checkpoint stores an account entry. When a reversion happens, it gets swapped with the current cache (or cause the current cache to get invalidated). So here, we first check whether we can get the storage key value in the checkpoint account entry, if not, then this value has not been changed, so we fetch the original value since last state commit.

ethcore/src/state/mod.rs
sorpaas: Yeah the logic seems to be flawed there. My current fix is recursive, and I can't think of a better way right now using the current checkpointing method.

I think we need to finish the state mod refactoring before this PR lands.

sorpaas: I realized that I [was](https://github.com/paritytech/parity-ethereum/pull/9319#discussion_r213246747) doing the index change in wrong direction. Just fixed this. Current solution is still recursive, but it has an upper bound. So I think this would work fine without #9427. However, #9427 would still be a really nice optimization.

sorpaas:
All TODO items for this PR are cleared.

ethcore/src/externalities.rs
```diff
@@ -106,13 +108,18 @@ impl<'a, T: 'a, V: 'a, B: 'a> Externalities<'a, T, V, B>
...
+			transaction_checkpoint_index: transaction_checkpoint_index,
 		}
 	}
 }
 
 impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	where T: Tracer, V: VMTracer, B: StateBackend
 {
+	fn reverted_storage_at(&self, key: &H256) -> vm::Result<H256> {
+		self.state.checkpoint_storage_at(self.transaction_checkpoint_index, &self.origin_info.address, key).map(|v| v.unwrap_or(H256::zero())).map_err(Into::into)
```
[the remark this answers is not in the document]
sorpaas: I think you're right! I don't think we had this assumption, but right now it's indeed the case -- checkpoints are only created via `Executive::call` and `Executive::create`, and we always made sure to remove all checkpoints before the top-level call/create is finished.

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,55 @@ impl<B: Backend> State<B> {
...
+				match entry {
+					Some(entry) => {
+						match entry.account {
+							Some(ref account) => {
+								if let Some(value) = account.cached_storage_at(key) {
+									Ok(Some(value))
+								} else {
+									// This key has checkpoint entry, but not in the entry's cache.
+									// So it's not modified at all.
+									Ok(Some(self.original_storage_at(address, key)?))
```
tomusdrw: Are you sure this is correct? We should probably check all the previous checkpoints, cause maybe the value is there?
I.e:
```
original: Some(Acc { Storage(x) = 0 })
checkpoint0: Some(Acc { Storage(x) = 1 })
checkpoint1: Some(Acc { Storage(x) = None, Storage(y) = 2 })
checkpoint2: Some(Acc { Storage(x) = 3 })
```
so now getting `checkpoint_storage_at(1, _, x)` should most likely return `1`, right? So the state of the account from the previous checkpoint, not the original storage?

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,55 @@ impl<B: Backend> State<B> {
...
+								}
+							},
+							None => {
+								// The account didn't exist at that point. Return empty value.
+								Ok(Some(H256::new()))
+							},
+						}
+					},
+					None => {
+						// The value was not cached at that checkpoint, meaning it was not modified.
```
tomusdrw: Perhaps it was modified in one of the previous checkpoints?

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,55 @@ impl<B: Backend> State<B> {
 			|a| a.as_ref().and_then(|account| account.storage_root().cloned()))
 	}
 
-	/// Mutate storage of account `address` so that it is `value` for `key`.
-	pub fn storage_at(&self, address: &Address, key: &H256) -> TrieResult<H256> {
+	/// Get the value of storage at a specific checkpoint.
+	pub fn checkpoint_storage_at(&self, checkpoint_index: usize, address: &Address, key: &H256) -> TrieResult<Option<H256>> {
```
tomusdrw: I'm unconvinced this code actually works correctly, do we have some higher-level tests to prove me wrong?

Best with scenarios, where we have deeply nested calls, the accounts are suicided in the meantime, the values are changed multiple times on different levels, etc.

ethcore/src/state/mod.rs
```diff
@@ -401,9 +401,12 @@ impl<B: Backend> State<B> {
 		self.factories.vm.clone()
 	}
 
-	/// Create a recoverable checkpoint of this state.
-	pub fn checkpoint(&mut self) {
-		self.checkpoints.get_mut().push(HashMap::new());
+	/// Create a recoverable checkpoint of this state. Return the checkpoint index.
+	pub fn checkpoint(&mut self) -> usize {
```
tomusdrw: I'm a bit hesitant on exposing the index here. If we merge / discard the checkpoint the index might point to a completely different one or be plain invalid.

ethcore/src/state/mod.rs
```diff
@@ -2196,6 +2266,27 @@ mod tests {
...
+		let a = Address::zero();
+		let k = H256::from(U256::from(0));
+		let c1 = state.checkpoint();
+		state.set_storage(&a, k, H256::from(U256::from(1))).unwrap();
+		let c2 = state.checkpoint();
+		let c3 = state.checkpoint();
+		state.set_storage(&a, k, H256::from(U256::from(3))).unwrap();
+		let c4 = state.checkpoint();
+		state.set_storage(&a, k, H256::from(U256::from(4))).unwrap();
+		let c5 = state.checkpoint();
```
tomusdrw: Can you check what happens when checkpoints are discarded / committed? I believe checkpoints represent:
"The changes that has to be made to the cache to return to a point in time." and because of that the logic at `checkpoint_storage_at` might be off.

ethcore/src/state/mod.rs
```diff
@@ -2196,6 +2266,27 @@ mod tests {
...
+
+		assert_eq!(state.checkpoint_storage_at(c1, &a, &k).unwrap(), Some(H256::from(U256::from(0))));
+		assert_eq!(state.checkpoint_storage_at(c2, &a, &k).unwrap(), Some(H256::from(U256::from(1))));
```
tomusdrw: Why is it `1`? If we discard all checkpoints `c3,c4,c5` I believe we should be back at `0`?
Also its a bit weird that checking `checkpoint_storage_at(c2)` before creating checkpoint `c3` would return different values, but perhaps i'm missing something.

ethcore/vm/src/ext.rs
```diff
@@ -63,6 +63,9 @@ pub enum CreateContractAddress {
 
 /// Externalities interface for EVMs
 pub trait Ext {
+	/// Returns the storage value for a given key if reversion happens on the current transaction.
+	fn reverted_storage_at(&self, key: &H256) -> Result<H256>;
```
tomusdrw: Maybe call it `previous_storage_at` or even `initial_storage_at`? With a comment explaining that it returns the storage value for given key before running current transaction?

ethcore/vm/src/ext.rs
```diff
@@ -136,8 +139,11 @@ pub trait Ext {
 	/// then A depth is 0, B is 1, C is 2 and so on.
 	fn depth(&self) -> usize;
 
-	/// Increments sstore refunds count by 1.
-	fn inc_sstore_clears(&mut self);
+	/// Increments sstore refunds counter.
+	fn inc_sstore_refund(&mut self, value: U256);
```
tomusdrw: I'd rather call it `add` and `sub` instead of `inc` and `dec`, since it takes a parameter now. (or change the comments to say "Increase/Decrease" instead of "Increment/Decrement")

ethcore/src/externalities.rs
```diff
@@ -390,8 +394,12 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 		self.depth
 	}
 
-	fn inc_sstore_clears(&mut self) {
-		self.substate.sstore_clears_count = self.substate.sstore_clears_count + U256::one();
+	fn inc_sstore_refund(&mut self, value: U256) {
+		self.substate.sstore_clears_refund = self.substate.sstore_clears_refund + value;
```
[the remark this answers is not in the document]
sorpaas: We're sure it won't overflow -- Refund will never go above gas provided, and gas provided has upper limit of `U256::max_value()`. I think the issue here is that if it actually overflows, that indicates a consensus bug, and I think it would be better to panic instead of silently saturating if that ever happens.

ethcore/src/externalities.rs
```diff
@@ -390,8 +394,12 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
...
+	}
+
+	fn dec_sstore_refund(&mut self, value: U256) {
+		self.substate.sstore_clears_refund = self.substate.sstore_clears_refund - value;
```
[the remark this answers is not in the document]
sorpaas: The same as above -- we're mostly sure it won't overflow, otherwise it's a consensus bug.

ethcore/src/state/mod.rs
sorpaas: Yes you're right. I'm planning to use #9427 (where we can get better references to checkpoints). However, right now I can't think of any better ways.

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,55 @@ impl<B: Backend> State<B> {
...
+						// The value was not cached at that checkpoint, meaning it was not modified.
+						Ok(Some(self.original_storage_at(address, key)?))
+					},
+				}
+			} else {
+				// This key does not have a checkpoint entry.
+				if checkpoint_index >= checkpoints_len - 1 {
+					Ok(Some(self.storage_at(address, key)?))
+				} else {
+					self.checkpoint_storage_at(checkpoint_index + 1, address, key)
```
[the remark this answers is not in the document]
sorpaas: @tomusdrw It will at most only be 1024 depth (where we only push several small stack items), but I think this is something we should fix for current state mod implementation.

ethcore/src/state/mod.rs
sorpaas: Can you explain more what you mean by "original" (what's the checkpoint index for "original", "checkpoint0")? If we didn't checkpoint before modifying the state, then the original state is forever lost?

ethcore/src/state/mod.rs
sorpaas: If we find `None` as existing for a checkpoint entry, then it means that previously when swapped from cache, the cache doesn't have that value. This means that it's not modified by this `State<B>` at all.

ethcore/src/state/mod.rs
sorpaas: The `c2` checkpoint is created after `state.set_storage(1)`. So in that case at `c2` the value should be `1`?

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
sorpaas: Consider the following checkpoint sequences:

```
orig -> c1 -> set 0 -> c2 -> c3 -> set 1
```

When calling `set 0`, `orig` would be swapped to `c1`'s entry. When calling `set 1`, `set 0` would be swapped to `c3`'s entry. So after this, we have:

```
c1: orig
c2: (empty)
c3: set 0
```

If we want to get `c2`'s value (which should be `set 0`), we go to `c2`'s entry, finding that it doesn't even contain the entry, so we need to go downward to `c3` and find `set 0`.

ethcore/src/state/mod.rs
sorpaas: Hmm that was L563. Let me check this.

ethcore/src/externalities.rs
```diff
@@ -390,8 +394,12 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 		self.depth
 	}
 
-	fn inc_sstore_clears(&mut self) {
-		self.substate.sstore_clears_count = self.substate.sstore_clears_count + U256::one();
+	fn inc_sstore_refund(&mut self, value: U256) {
```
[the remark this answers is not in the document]
tomusdrw: I didn't mean memory, but rather a performance improvement since we are not using our custom U256 arithmetic. But as mentioned, it's fine for me to see a separate PR with that if anyone is interested on working on this, it's probably not super hot code anyway, so it would be marginal.

ethcore/src/externalities.rs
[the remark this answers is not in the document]
tomusdrw: It would be a consensus bug only if it's implemented differently in other clients. Always panicking doesn't seem to me like a good idea (would actually prefer the network to split than to kill large part of it, or even the entire network if it's a private Parity-only chain).

Happy with whatever behavior that doesn't cause a panic (i.e. `overflowing_add` or `saturating_add`) - better be safe (and explicit) than sorry.

ethcore/src/state/mod.rs
tomusdrw: Makes sense!

ethcore/src/state/mod.rs
tomusdrw: You're right, my misunderstanding of how checkpoints work!

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
tomusdrw: You're right, I misunderstood the direction in which checkpoints work.

ethcore/src/state/mod.rs
sorpaas: Here's some added tests for discard/revert of checkpoints. https://github.com/paritytech/parity-ethereum/pull/9319/commits/c2b9e9741cb9d92c48eab2c200ba33f11de0290a
As @tomusdrw pointed out, we probably still need to check coverage to make sure all branches are tested.

ethcore/src/externalities.rs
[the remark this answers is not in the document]
sorpaas: I realized that I was wrong here. Theoretically it's indeed possible for this value to go over `U256::max_value()` (but practically that should never happen. So I'm going to change this to `saturating_add`.

sorpaas:
@5chdn Not yet. I'm still fixing @tomusdrw's grumbles!

ethcore/src/state/mod.rs
```diff
@@ -539,8 +542,55 @@ impl<B: Backend> State<B> {
...
+							},
+							None => {
+								// The account didn't exist at that point. Return empty value.
+								Ok(Some(H256::new()))
+							},
+						}
+					},
+					None => {
+						// The value was not cached at that checkpoint, meaning it was not modified.
+						Ok(Some(self.original_storage_at(address, key)?))
```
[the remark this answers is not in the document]
sorpaas: Fixed this by only calling `original_storage_at`/`storage_at` after releasing current `self.checkpoints` borrow.

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
sorpaas: Fixed!

ethcore/evm/src/interpreter/gasometer.rs
```diff
@@ -342,6 +347,64 @@ fn add_gas_usize<Gas: evm::CostType>(value: Gas, num: usize) -> (Gas, bool) {
...
+					schedule.sstore_reset_gas
+				}
+			} else {
+				schedule.sload_gas
+			}
+		}
+	)
+}
+
+pub fn handle_eip1283_sstore_clears_refund(ext: &mut vm::Ext, original: &U256, current: &U256, new: &U256) {
```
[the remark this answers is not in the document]
sorpaas: I think there may be issues doing this -- sometimes we need to issue two refunds at once, and the value itself can be either positive or negative. So returning a value instead of directly modifying ext would mean that we need extra codes to handle signed addition/subtraction.

ethcore/evm/src/interpreter/gasometer.rs
```diff
@@ -342,6 +347,64 @@ fn add_gas_usize<Gas: evm::CostType>(value: Gas, num: usize) -> (Gas, bool) {
 	value.overflow_add(Gas::from(num))
 }
 
+#[inline]
+fn calculate_eip1283_sstore_gas<Gas: evm::CostType>(schedule: &Schedule, original: &U256, current: &U256, new: &U256) -> Gas {
+	Gas::from(
+		if current == new {
```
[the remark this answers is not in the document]
sorpaas: TBH I think for this, using if branches may be better. In that way we can directly follow the spec of each cause. Using match is indeed shorter, but I think it may actually be more confusing.

ethcore/vm/src/ext.rs
sorpaas: Fixed!

sorpaas:
Addressed all current round of grumbles. This PR is again ready for review!

.gitlab-ci.yml
```diff
@@ -85,6 +85,7 @@ test-coverage-kcov:
   stage:                           test
   only:
     - master
+    - pr-9319
```
tomusdrw: To be removed before merge

ethcore/evm/src/interpreter/gasometer.rs
```diff
@@ -124,13 +124,18 @@ impl<Gas: evm::CostType> Gasometer<Gas> {
 				let address = H256::from(stack.peek(0));
 				let newval = stack.peek(1);
 				let val = U256::from(&*ext.storage_at(&address)?);
+				let orig = U256::from(&*ext.initial_storage_at(&address)?);
```
tomusdrw: I still believe it's better to move it inside the if banch

sorpaas: Sorry I missed this. Moved!

ethcore/src/executive.rs
```diff
@@ -1614,6 +1614,62 @@ mod tests {
...
 
+	evm_test!{test_eip1283: test_eip1283_int}
+	fn test_eip1283(factory: Factory) {
+		let x1 = Address::from(0x1000);
+		let x2 = Address::from(0x1001);
+		let y1 = Address::from(0x2001);
+		let y2 = Address::from(0x2002);
+
+		let mut state = get_temp_state_with_factory(factory.clone());
+		state.new_contract(&x1, U256::zero(), U256::from(1)).unwrap();
```
cheme: The `new_contract` does not seems to be mandatory, it contradicts the comment of `init_code`, but was use this way in other test. Let's keep it.

ethcore/src/state/account.rs
```diff
@@ -154,12 +162,17 @@ impl Account {
 
 	/// Create a new contract account.
 	/// NOTE: make sure you use `init_code` on this before `commit`ing.
-	pub fn new_contract(balance: U256, nonce: U256) -> Account {
+	pub fn new_contract(balance: U256, nonce: U256, original_storage_root: H256) -> Account {
```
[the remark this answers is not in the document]
cheme: I may have a wrong impression about the existing code (I noted that yesterday evening), still likes Option version better, but it is only a suggestion.

ethcore/src/state/account.rs
```diff
@@ -457,16 +553,21 @@ impl Account {
...
-		let mut cache = self.storage_cache.borrow_mut();
-		for (k, v) in other.storage_cache.into_inner() {
-			cache.insert(k, v);
+		if self.storage_root == other.storage_root {
+			let mut cache = self.storage_cache.borrow_mut();
+			for (k, v) in other.storage_cache.into_inner() {
+				cache.insert(k, v);
+			}
+		} else {
+			self.storage_cache = other.storage_cache;
```
cheme: This makes sense to me (the replacement logic for new contracts and cache), but I do not have the background to be confident with this new behavior in the context of the evm. (I am not requesting anything, jut pointing out that my review is a bit incomplete on this part of the code).

ethcore/evm/src/interpreter/gasometer.rs
```diff
@@ -124,13 +124,18 @@ impl<Gas: evm::CostType> Gasometer<Gas> {
...
 
-				let gas = if val.is_zero() && !newval.is_zero() {
-					schedule.sstore_set_gas
+				let gas = if schedule.eip1283 {
+					calculate_eip1283_sstore_gas(schedule, &orig, &val, &newval)
```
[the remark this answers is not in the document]
cheme: You address it in your last commit (my comment was unclear)

ethcore/src/executive.rs
```diff
@@ -1614,6 +1614,62 @@ mod tests {
...
+			params.code = Some(Arc::new("6001600055600060006000600061200263fffffffff4".from_hex().unwrap()));
+			params.gas = gas;
+			let mut substate = Substate::new();
+			let mut ex = Executive::new(&mut state, &info, &machine, &schedule);
+			let res = ex.call(params, &mut substate, &mut NoopTracer, &mut NoopVMTracer).unwrap();
+
+			(res, substate.sstore_clears_refund, gas)
+		};
+		let gas_used = gas - gas_left;
+		// sstore: 1 -> (1) -> () -> (0 -> 1 -> 0)
```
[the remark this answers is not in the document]
sorpaas: The first number is the original value. Added the tests!

ethcore/src/state/account.rs
```diff
@@ -59,6 +59,10 @@ pub struct Account {
 	// LRU Cache of the trie-backed storage.
 	// This is limited to `STORAGE_CACHE_ITEMS` recent queries
 	storage_cache: RefCell<LruCache<H256, H256>>,
+	// LRU Cache of the trie-backed storage for original value.
+	// This is only used when the initial storage root is different compared to
+	// what is in the database. That is, it is only used for new contracts.
+	original_storage_cache: Option<(H256, RefCell<LruCache<H256, H256>>)>,
```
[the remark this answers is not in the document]
sorpaas: Yeah, `original_storage_cache` is a lot like `storage_root` + `storage_cache`. However, the usage for them are really different so I'm not sure if refactoring would save code.

And TBH, after we reviewed and merged this PR, I think the next reasonable step would be to speed up finishing #9427 -- the current checkpoint logic is not easy to follow, so moving the code would only help a little. If we make it so that account/storage changes struct are cheap to clone (through PDS), then I think checkpoint logic can be simplified a lot.

ethcore/src/state/account.rs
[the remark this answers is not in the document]
sorpaas: I think it may actually be more confusing. In account model, we currently distinguish:

* Account that does not exist. (`None`)
* Account exists, and storage is empty. (`Some(KECCAK_NULL_RLP)`)
* Account exists, and storage is not empty. (`Some(x)` where `x != KECCAK_NULL_RLP`)

For the purpose of `original_storage_root`, we don't need to distinguish `None` and `Some(KECCAK_NULL_RLP)`. So without `Option` it may work better here.

ethcore/src/state/mod.rs
```diff
@@ -536,11 +552,86 @@ impl<B: Backend> State<B> {
...
+	/// Get the value of storage at a specific checkpoint.
+	pub fn checkpoint_storage_at(&self, checkpoint_index: usize, address: &Address, key: &H256) -> TrieResult<Option<H256>> {
+		enum ReturnKind {
+			/// Use original storage at value at this address.
+			OriginalAt,
+			/// Use the downward checkpoint value.
+			Downward,
+		}
+
+		let (checkpoints_len, kind) = {
```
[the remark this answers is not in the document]
sorpaas: The block is actually small and mostly consists of comments. :)

I think it would be better to keep `ReturnKind` to be inside `checkpoint_storage_at`, but putting another function would mean that's not possible.

ethcore/src/state/mod.rs
```diff
@@ -536,11 +552,86 @@ impl<B: Backend> State<B> {
...
+						// account's original storage root. Otherwise, we need to populate the checkpoint account. Note
+						// that the later case is not possible under current Ethereum consensus rules.
+						if account.base_storage_root() == self.original_storage_root(address)? {
+							ReturnKind::OriginalAt
+						} else {
+							// If account base storage root is different from the original storage root since last
+							// commit, then it can only be created from a new contract, where the base storage root
+							// would always be empty. Note that this branch is actually never called, because
+							// `cached_storage_at` handled this case.
+							return Ok(Some(H256::new()));
```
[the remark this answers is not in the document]
sorpaas: This should indeed be `unreachable!` if we're absolutely certain our code contains no bug. I think we try to make sure the logic is correct as best as we can, but if we made some mistakes elsewhere and it accidentally reached this cause, this reasoning logic provided in this cause is still correct, and we would avoid a client panic.

ethcore/src/executive.rs
sorpaas: Yeah, this is just to test whether checkpointing is working with `new_contract`. In reality we would not encounter `new_contract` called on an existing account.

ethcore/src/state/account.rs
sorpaas: In this case, `self.storage_cache` does not equal `other.storage_cache`. This means one of them is `KECCAK_NULL_RLP`. We cannot merge the storage cache here. Because this function is to "overwrite" `self` with `other`, we just move `other`'s storage cache to `self`.

ethcore/src/executive.rs
sorpaas: Ah sorry, I was confused on this and the other test: In this case, we want to create new contracts and init code for `x1`, `x2`, `y1`, `y2`. `new_contract` must be called before `init_code`.

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
cheme: That is what I thought, as I say I prefer to fail than having an unexpected behavior, but I understand that in a production context it is not really possible. Ideally we would have a test build with panics and a production build without. A log alert trace could be a good option for such case.

ethcore/src/state/account.rs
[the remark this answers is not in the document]
cheme: I guess it is good to keep it in mind, could be a point to consider in #9427 (Option<Option< is ugly but there is other ways).

ethcore/src/state/account.rs
cheme: Yes the logic is sound, thanks for confirming it, I was just worried about some 'assertion' that could have been made on existing code.

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
cheme: It's fine with me.

ethcore/src/state/mod.rs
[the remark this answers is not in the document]
sorpaas: Currently `codecov` shows that this cause is indeed not triggered by any tests. I would actually consider it more like a failsafe (it does not necessarily break consensus) -- if this cause is ever reached, returning `H256::new()` is the only possible correct value.

Added a `warn!` so that if it ever happens, we'll (hopefully) get reports!

Nathapon8899:
sp-eip1283

Nathapon8899:
#9040

PR #9321, opened 2018-08-08
Fix load share

ordian:
Share of the load given to a light peer didn't depend on the max number of connected peers.
This PR fixes it.

### DIFF
--- a/ethcore/sync/src/api.rs
+++ b/ethcore/sync/src/api.rs
@@ -265,7 +265,7 @@ fn light_params(
 	network_id: u64,
 	max_peers: u32,
 	pruning_info: PruningInfo,
-	sample_store: Option<Box<SampleStore>>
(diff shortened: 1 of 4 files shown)


ethcore/sync/src/api.rs
```diff
@@ -256,11 +261,35 @@ pub struct EthSync {
...
+		capabilities: Capabilities {
+			serve_headers: true,
+			serve_chain_since: Some(pruning_info.earliest_chain),
+			serve_state_since: Some(pruning_info.earliest_state),
+			tx_relay: true,
+		},
+		sample_store: sample_store,
+	};
+
+	let max_peers = ::std::cmp::max(max_peers, 1);
```
dvdplm: Bugfix is here in case anyone struggles to see it. 👍

PR #9324, opened 2018-08-09
More details in logs returned by light client

### LINKED ISSUE #9214: Light Node Sending Pending Events Only
> I'm running:
>
> - **How installed?**: installer

I am running a light node which is run through the following command "parity --no-ancient-blocks --jsonrpc-hosts "all" --unsafe-expose --tracing off --warp --light".

I listen to a contract's Transfer logs using the code below.

`var Transfer = contract.Transfer({}, {fromBlock: 'latest', toBlock: 'latest'});`

While I am using a full node, the event logs I receive are with the status/type "mined", but with the same code, when I use the light node, the logs are returned with a status/type "pending" and most of the information such as transaction hash is not available.

I don't plan to keep a full node running for Parity. So can anyone point us what is the issue? Is there any to get the light node to send me "only mined Transfer events"?

tomusdrw:
Previous code was just mapping `ethcore::LogEntry` into `rpc::Log`, which by default omits all localisation details of the log.
This PR changes the logic to create `LocalizedLogEntry` first and then convert to rpc type.

There is no easy way to fetch `transaction_hash` though afaict (we would have to query block body for that), so it's going to be `null` in the response.

Closes #9214

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -22,7 +22,6 @@ use ethcore::basic_account::BasicAccount;
 use ethcore::encoded;
 use ethcore::filter::Filter as EthcoreFilter;
 use ethcore::ids::BlockId;
-use ethcore::log_entry::LocalizedLogEntry;
 use ethcore::receipt::Receipt;
 
 use jsonrpc_core::{Result, Error};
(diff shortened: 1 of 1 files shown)


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -338,17 +339,34 @@ impl LightFetch {
...
+									entry: log,
+									block_hash: hash,
+									block_number: num,
+									// No way to easily retrieve transaction hash, so let's just skip it.
+									transaction_hash: Default::default(),
+									transaction_index,
+									log_index: block_index,
+									transaction_log_index,
+								};
+								let mut log = Log::from(log);
```
debris: Can we just create this `Log` instead and avoid conversion and mutable fields?

tomusdrw: Good idea. Done.

PR #9328, opened 2018-08-09
ethcore: fix pow difficulty validation

### LINKED ISSUE #8397: Stage 1 block verification failed for c913…53da: Block(InvalidProofOfWork(OutOfBounds { min: Some(307293), max: None, found: 307292 }))
Parity node stops syncing with block verification error:

```
2018-04-12 19:28:23 UTC Imported #307711 1f51…966a (0 txs, 0.00 Mgas, 0.97 ms, 0.52 KiB)
2018-04-12 19:28:30 UTC Stage 1 block verification failed for c913…53da: Block(InvalidProofOfWork(OutOfBounds { min: Some(307293), max: None, found: 307292 }))
2018-04-12 19:28:33 UTC    2/25 peers      5 MiB chain   21 KiB db  0 bytes queue    3 MiB sync  RPC:  0 conn,  0 req/s, 275 µs
```

> I'm running:
>
> - **Which Parity version?**: 1.9.5
> - **How installed?**: official docker images
> - **Which network are you connected to?**: local network
> - **Did you try to restart the node?**: yes, after restart I still get:

```
2018-04-13 10:22:12 UTC Stage 1 block verification failed for c913…53da: Block(InvalidProofOfWork(OutOfBounds { min: Some(307293), max: None, found: 307292 }))
```

I got this issue two times with brand new local chains. First time it happened around 100000 block. Then I simply started new chain and it happend again on block 307711. Network runs a few latest stable version Geth nodes and a single latest stable Parity node (various stable versions of both Geth and Parity nodes as chains runs for a few months and I upgrade once new is released).

I don't know any easy way to reproduce it.

andresilva:
Fixes #8397.


andresilva:
I moved these methods to the `ethash` crate, this way we can remove a duplicate definition from `miner` (since it already depends on `ethash` crate).

ethash/src/lib.rs
```diff
@@ -136,6 +138,32 @@ impl EthashManager {
 	}
 }
 
+/// Convert an Ethash boundary to its original difficulty. Basically just `f(x) = 2^256 / x`.
+pub fn boundary_to_difficulty(boundary: &ethereum_types::H256) -> U256 {
+	assert!(!boundary.is_zero());
+
+	let d = U512::from(&**boundary);
```
andresilva: This method could be rewritten to use `difficulty_to_boundary` but we would have to do an extra conversion. Still, both definitions are next to another and are tested in relation to one another, so I think it's fine to have this duplicated.

ordian: I think something like `fn(n: impl Into<U512>) -> U256` could work

miner/src/work_notify.rs
```diff
@@ -67,19 +67,10 @@ impl WorkPoster {
 	}
 }
 
-/// Convert an Ethash difficulty to the target boundary. Basically just `f(x) = 2^256 / x`.
-fn difficulty_to_boundary(difficulty: &U256) -> H256 {
```
debris: good catch! I didn't know that we have it in more than 1 place

debris:
please merge with master :)

ethash/src/lib.rs
```diff
@@ -136,6 +138,29 @@ impl EthashManager {
...
+	difficulty_to_boundary_aux(difficulty).into()
+}
+
+fn difficulty_to_boundary_aux<T: Into<U512>>(difficulty: T) -> ethereum_types::U256 {
+	let difficulty = difficulty.into();
+
+	assert!(!difficulty.is_zero());
+
+	if difficulty == U512::one() {
+		U256::max_value().into()
```
ordian: nit: we don't need that `.into()`

ethash/src/lib.rs
```diff
@@ -136,6 +138,29 @@ impl EthashManager {
...
+	} else {
+		// d > 1, so result should never overflow 256 bits
```
ordian: nit: s/d/difficulty

PR #9331, opened 2018-08-09
Light client logs should include 'from_block' when querying logs

### LINKED ISSUE #8560: light client: subscribe over websockets doesnt send new events
Not sure if its related to the light client or websocket. If I use a recent from block I get some events, if I use latest I get nothing, and either way I never get any new events (as I can open a second script attached to infura's websocket)

```
target/release/parity -d /Volumes/vms --light
2018-05-07 11:43:18  Starting Parity/v1.12.0-unstable-f0c6d17ad-20180504/x86_64-macos/rustc1.25.0
2018-05-07 11:43:18  Keys path /Volumes/vms/keys/Foundation
2018-05-07 11:43:18  DB path /Volumes/vms/chains/ethereum/db/906a34e69aec8c0d
2018-05-07 11:43:18  Path to dapps /Volumes/vms/dapps
2018-05-07 11:43:18  Running in experimental Light Client mode.
2018-05-07 11:43:20  Imported #5572784 0x5aa1…b1ba (7.98 Mgas) + another 1 header(s)
2018-05-07 11:43:23  Public node URL: enode://c6622236c33a3ae026da0622472c37d77ee6f8194908d73adb3d00505fb63d29521b49d5588d3f8e202a9acd5fbcc65646a1e1c30d9bdeee3af5740857065db4@192.168.0.35:30303
2018-05-07 11:43:31  Imported #5572785 0x2cb3…143c (0.40 Mgas)
2018-05-07 11:43:40  Imported #5572787 0xf283…03ec (7.99 Mgas)
2018-05-07 11:43:48    14/50 peers   4 KiB cache 0 bytes queue  RPC:  0 conn,  0 req/s,   0 µs
```

```
contract.events.allEvents(
{fromBlock: 5572785} //gets recent past events, but no new updates
//{} //nothing
//{fromBlock:'latest'} //nothing
, function(error, events){
console.log(error, events)
})
```

```
null { address: '0x06012c8cf97BEaD5deAe237070F9587f8E7A266d',
  blockHash: null,
  blockNumber: null,
  logIndex: null,
  transactionHash: null,
  transactionIndex: null,
  transactionLogIndex: null,
  type: 'pending',
  id: null,
  returnValues: 
   Result {
     '0': '0x5B49c5072750Ec45E63E1F300cc4A3a6C8CB7599',
     '1': '740439',
     '2': '740367',
     '3': '739104',
     '4': '511181029587041829902179471452593439812944709427753037147557425946068364',
     owner: '0x5B49c5072750Ec45E63E1F300cc4A3a6C8CB7599',
     kittyId: '740439',
     matronId: '740367',
     sireId: '739104',
     genes: '511181029587041829902179471452593439812944709427753037147557425946068364' },
  event: 'Birth',
  signature: '0x0a5311bd2a6608f08a180df2ee7c5946819a649b204b554bb8e39825b2c50ad5',
  raw: 
   { data: '0x0000000000000000000000005b49c5072750ec45e63e1f300cc4a3a6c8cb759900000000000000000000000000000000000000000000000000000000000b4c5700000000000000000000000000000000000000000000000000000000000b4c0f00000000000000000000000000000000000000000000000000000000000b472000004a10c218c74b8295c5ee180670c0c6210c40808734591394e52b8e90858c',
     topics: [ '0x0a5311bd2a6608f08a180df2ee7c5946819a649b204b554bb8e39825b2c50ad5' ] } }
```

> I'm running:
>
> - **Which Parity version?**: 0.0.0
a7a46f425369fd662042cb4a1c9ffc301b3b5b75
and "web3": "^1.0.0-beta.34"
> - **Which operating system?**: Windows / MacOS / Linux
macosx
> - **How installed?**: via installer / homebrew / binaries / from source
source
> - **Are you fully synchronized?**: no / yes
yes, light client
> - **Which network are you connected to?**: ethereum / ropsten / kovan / ...
mainnet
> - **Did you try to restart the node?**: no / yes
yes

cheme:
Fixes #8560 .

When using a websocket on logs, there was no entry displayed. The issue was that we called light client get logs method for each block (with a range of one block in the filter), but the method did not include the first block to query (with a range of one block it did include nothing).

This PR fixes this by making LightClient `log` works correctly (by including the start block of the filter).

For the websocket usecase it also increases OnDemand query a lot, so this other PR may be helpfull : #9318 (during my testing without #9318 the lightclient became a bit irresponsive after a while). There is also probably room for optimization in function 'notify_log', yet it should not directly impact network query (if OnDemand cache works as intended of course).

I did try to change as less as possible code, but from a personal point of view I would rather use : 

```rust
      let mut receipts_futures: Vec<_> = Vec::new();
      for hdr in self.client.ancestry_iter(filter.to_block) {
        let end_cond =  BlockId::Hash(hdr.hash()) == filter.from_block
          || BlockId::Number(hdr.number()) == filter.from_block;

        let hdr_bloom = hdr.log_bloom();
        if bit_combos.iter().find(|&bloom| hdr_bloom & *bloom == *bloom).is_some() {
          let hdr_number = hdr.number();
          receipts_futures.push(
            self.on_demand.request(ctx, request::BlockReceipts(hdr.into())).expect(NO_INVALID_BACK_REFS)
            .map(move |x|{ (hdr_number, x) })
            );
        }
        if end_cond { break; }
      }
      let receipts_futures = receipts_futures;
```
than

```rust
			let receipts_futures: Vec<_> = self.client.ancestry_iter(filter.to_block)
				.take_while(|ref hdr|{
					BlockId::Hash(hdr.hash()) != filter.from_block
					&& BlockId::Number(hdr.number()) != filter.from_block
				})
				.chain(::std::iter::once(self.client.block_header(filter.from_block).expect("checked before")))
				.filter(|ref hdr| {
					let hdr_bloom = hdr.log_bloom();
					bit_combos.iter().find(|&bloom| hdr_bloom & *bloom == *bloom).is_some()
				})
				.map(|hdr| (hdr.number(), request::BlockReceipts(hdr.into())))
				.map(|(num, req)| self.on_demand.request(ctx, req).expect(NO_INVALID_BACK_REFS).map(move |x| (num, x)))
				.collect();


```

So if the reviewers prefer the more imperative version please tell me.

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -337,10 +337,10 @@ impl LightFetch {
 					BlockId::Hash(hdr.hash()) != filter.from_block
 					&& BlockId::Number(hdr.number()) != filter.from_block
 				})
-				.chain(::std::iter::once(self.client.block_header(filter.from_block).expect("checked before")))
+				.chain(Some(self.client.block_header(filter.from_block).expect("checked before")))
 				.filter(|ref hdr| {
 					let hdr_bloom = hdr.log_bloom();
-					bit_combos.iter().find(|&bloom| hdr_bloom & *bloom == *bloom).is_some()
+					bit_combos.iter().any(|bloom| hdr_bloom.contains_bloom(bloom))
 				})
 				.map(|hdr| (hdr.number(), request::BlockReceipts(hdr.into())))
 				.map(|(num, req)| self.on_demand.request(ctx, req).expect(NO_INVALID_BACK_REFS).map(move |x| (num, x)))
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -321,10 +321,15 @@ impl LightFetch {
 			BlockId::Number(x) => Some(x),
 		};
 
-		match (block_number(filter.to_block), block_number(filter.from_block)) {
-			(Some(to), Some(from)) if to < from || from > best_number 
+		let (from_block_number, from_block_header) = match self.client.block_header(filter.from_block) {
+			Some(from) => (from.number(), from),
+			None => return Either::A(future::err(errors::unknown_block())),
+		};
+
+		match block_number(filter.to_block) {
+			Some(to) if to < from_block_number || from_block_number > best_number 
 				=> return Either::A(future::ok(Vec::new())),
-			(Some(_), Some(_)) => {},
+			Some(_) => (),
 			_ => return Either::A(future::err(errors::unknown_block())),
 		}
 
@@ -333,11 +338,8 @@ impl LightFetch {
 			// match them with their numbers for easy sorting later.
 			let bit_combos = filter.bloom_possibilities();
 			let receipts_futures: Vec<_> = self.client.ancestry_iter(filter.to_block)
-				.take_while(|ref hdr|{
-					BlockId::Hash(hdr.hash()) != filter.from_block
-					&& BlockId::Number(hdr.number()) != filter.from_block
-				})
-				.chain(Some(self.client.block_header(filter.from_block).expect("checked before")))
+				.take_while(|ref hdr| hdr.number() != from_block_number)
+				.chain(Some(from_block_header))
 				.filter(|ref hdr| {
 					let hdr_bloom = hdr.log_bloom();
 					bit_combos.iter().any(|bloom| hdr_bloom.contains_bloom(bloom))
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -322,7 +322,8 @@ impl LightFetch {
 		};
 
 		match (block_number(filter.to_block), block_number(filter.from_block)) {
-			(Some(to), Some(from)) if to < from => return Either::A(future::ok(Vec::new())),
+			(Some(to), Some(from)) if to < from || from > best_number 
+				=> return Either::A(future::ok(Vec::new())),
 			(Some(_), Some(_)) => {},
 			_ => return Either::A(future::err(errors::unknown_block())),
 		}
@@ -332,8 +333,11 @@ impl LightFetch {
 			// match them with their numbers for easy sorting later.
 			let bit_combos = filter.bloom_possibilities();
 			let receipts_futures: Vec<_> = self.client.ancestry_iter(filter.to_block)
-				.take_while(|ref hdr| BlockId::Number(hdr.number()) != filter.from_block)
-				.take_while(|ref hdr| BlockId::Hash(hdr.hash()) != filter.from_block)
+				.take_while(|ref hdr|{
+					BlockId::Hash(hdr.hash()) != filter.from_block
+					&& BlockId::Number(hdr.number()) != filter.from_block
+				})
+				.chain(::std::iter::once(self.client.block_header(filter.from_block).expect("checked before")))
 				.filter(|ref hdr| {
 					let hdr_bloom = hdr.log_bloom();
 					bit_combos.iter().find(|&bloom| hdr_bloom & *bloom == *bloom).is_some()


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -332,8 +333,11 @@ impl LightFetch {
 			// match them with their numbers for easy sorting later.
 			let bit_combos = filter.bloom_possibilities();
 			let receipts_futures: Vec<_> = self.client.ancestry_iter(filter.to_block)
-				.take_while(|ref hdr| BlockId::Number(hdr.number()) != filter.from_block)
-				.take_while(|ref hdr| BlockId::Hash(hdr.hash()) != filter.from_block)
+				.take_while(|ref hdr|{
+					BlockId::Hash(hdr.hash()) != filter.from_block
+					&& BlockId::Number(hdr.number()) != filter.from_block
```
debris: this is a bug. this take_while will never return `false` therefore this iterator will never end

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -332,8 +333,11 @@ impl LightFetch {
...
+				})
+				.chain(::std::iter::once(self.client.block_header(filter.from_block).expect("checked before")))
 				.filter(|ref hdr| {
 					let hdr_bloom = hdr.log_bloom();
 					bit_combos.iter().find(|&bloom| hdr_bloom & *bloom == *bloom).is_some()
```
debris: not related to this pr, but it's more optimal to use `contains_bloom`

rpc/src/v1/helpers/light_fetch.rs
ordian: `take_while(a).take_while(b)` is equivalent to `take_while(a && b)`, right? 
Should we check for the `BlockId::Latest` case as well?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -332,11 +333,14 @@ impl LightFetch {
 			// match them with their numbers for easy sorting later.
 			let bit_combos = filter.bloom_possibilities();
 			let receipts_futures: Vec<_> = self.client.ancestry_iter(filter.to_block)
-				.take_while(|ref hdr| BlockId::Number(hdr.number()) != filter.from_block)
-				.take_while(|ref hdr| BlockId::Hash(hdr.hash()) != filter.from_block)
+				.take_while(|ref hdr|{
+					BlockId::Hash(hdr.hash()) != filter.from_block
+					&& BlockId::Number(hdr.number()) != filter.from_block
+				})
+				.chain(Some(self.client.block_header(filter.from_block).expect("checked before")))
```
ordian: `block_header` can return `None` and I don't see where do we check it's not `None`, could you point me to that line?

rpc/src/v1/helpers/light_fetch.rs
cheme: Seems correct to me (also when testing). I can still switch back to using two `take_while` (does not make a lot of difference as it combines correctly).

rpc/src/v1/helpers/light_fetch.rs
cheme: Yes, this kind of logic should definitely use dedicated function.

rpc/src/v1/helpers/light_fetch.rs
cheme: It was tested in the case of a block hash (in `block_number` ) and I added a condition before to make sure the number case would be in range to, it is certainly not explicit enough.
I did a little change to test it explicitly in my last commit (plus remove the need to have two condition in takewhile).

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -321,9 +321,15 @@ impl LightFetch {
...
-			(Some(to), Some(from)) if to < from => return Either::A(future::ok(Vec::new())),
-			(Some(_), Some(_)) => {},
+		let (from_block_number, from_block_header) = match self.client.block_header(filter.from_block) {
+			Some(from) => (from.number(), from),
+			None => return Either::A(future::err(errors::unknown_block())),
+		};
+
+		match block_number(filter.to_block) {
+			Some(to) if to < from_block_number || from_block_number > best_number 
+				=> return Either::A(future::ok(Vec::new())),
```
andresilva: Minor nit but I would put the `=>` in the previous line, I don't know what's our style on this though.

rpc/src/v1/helpers/light_fetch.rs
debris: `take_while(a).take_while(b)` is equal to `take_while(a || b)` and this is a big difference

PR #9332, opened 2018-08-09
Cross-client Encrypt/Decrypt

topealabi:
To encrypt

- we obtain an encryption public key from the reciever
- we generate an ephemeral key pair for the sender
- we encrypt the message using the reciever public key and the ephemeral private key, and a nonce

- We serialize to json and send over a a payload of the form:

const encryptedData = { version: 'x25519-xsalsa20-poly1305',
nonce: '1dvWO7uOnBnO7iNDJ9kO9pTasLuKNlej',
ephemPublicKey: 'FBH1/pAEHOOW14Lu3FWkgV3qOEcuL78Zy+qW1RwzMXQ=',
ciphertext: 'f8kBcl/NCyf3sybfbwAKk/np2Bzt9lRVkZejr6uh5FgnNlH/ic62DZzy' };

To decrypt

- we use the receiver’s encryption private key and the ephemeral public key which is included in the payload as well as the nonce

Related to:
https://github.com/ethereum/EIPs/pull/1098
https://github.com/MetaMask/eth-sig-util/pull/18


debris:
@topealabi thanks for your contribution. These changes should be reviewed after #8715 is merged. Also, please reopen the PR on top of the master. We do not do accept PRs to the feature branches

PR #9333, opened 2018-08-09
Cross-client Encrypt/Decrypt

topealabi:
To encrypt

we obtain an encryption public key from the reciever

we generate an ephemeral key pair for the sender

we encrypt the message using the reciever public key and the ephemeral private key, and a nonce

We serialize to json and send over a a payload of the form:

const encryptedData = { version: 'x25519-xsalsa20-poly1305',
nonce: '1dvWO7uOnBnO7iNDJ9kO9pTasLuKNlej',
ephemPublicKey: 'FBH1/pAEHOOW14Lu3FWkgV3qOEcuL78Zy+qW1RwzMXQ=',
ciphertext: 'f8kBcl/NCyf3sybfbwAKk/np2Bzt9lRVkZejr6uh5FgnNlH/ic62DZzy' };

To decrypt

we use the receiver’s encryption private key and the ephemeral public key which is included in the payload as well as the nonce

Related to:
ethereum/EIPs#1098
MetaMask/eth-sig-util#18

With help from @javadevelopr


topealabi:
[clabot:check]

dvdplm:
@topealabi It looks like this is work-in-progress at this time so I added the "inprogress" label. When you're ready for a review, let us know by adding the "pleasereview" label. Thanks! :)

topealabi:
Yes, should have an update in the next few days

5chdn:
Ok, let's reopen this once you get back.

PR #9334, opened 2018-08-09
ethcore: remove unused macro_use

andresilva:


dvdplm:
See https://github.com/paritytech/parity-ethereum/pull/9316

PR #9347, opened 2018-08-13
TransactionPool: Expose UnorderedIterator.

tomusdrw:
Might be required to name the type returned by `fn unordered_pending`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -669,7 +669,7 @@ dependencies = [
  "rlp 0.2.1 (git+https://github.com/paritytech/parity-common)",
  "rustc-hex 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "trace-time 0.1.0",
- "transaction-pool 1.12.2",
+ "transaction-pool 1.12.3",
  "url 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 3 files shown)


tomusdrw:
@dvdplm We have two methods in the pool that return pending transactions:
1. `pending` https://docs.rs/transaction-pool/1.12.1/transaction_pool/struct.Pool.html#method.pending
2. `unordered_pending` https://docs.rs/transaction-pool/1.12.1/transaction_pool/struct.Pool.html#method.unordered_pending

The first one returns transactions ordered by priority (more costly) the second one just returns all transactions in semi-random order. Both functions return a type that `impl Iterator`. Unfortunately the type of the second one was not public (exported from the crate), so the caller could use and consume iterator, but couldn't name it.

I'm exposing this to be able to use it in Polkadot, and need to write something like this: `fn all<R, F: FnOnce(UnorderedIterator) -> R>(&self, f: F) -> R {}`. if it was just about consuming the iterator or passing it further I could use `impl Trait`, but unfortunatelly you can't use `impl Trait` in such definition, so I have to use the concrete type.

Hope it helps with the context :)

dvdplm:
That was really useful, thanks! :)

> I'm exposing this to be able to use it in Polkadot,

So this is a candidate for `parity-common` then?

tomusdrw:
@dvdplm yup, https://github.com/paritytech/parity-common/issues/19

PR #9353, opened 2018-08-14
Remove prepare_trace_output and make sure prepare_trace_call and trace*call are balanced

sorpaas:
This refactors `prepare_trace_output` to instead directly take the reference of return values, so that it's simpler and we save a stack item. This should also fixes [the issue](https://github.com/paritytech/parity-ethereum/pull/9236#issuecomment-408444995) @udoprog is facing. Replaces #9236


udoprog:
Nice, thanks!

PR #9359, opened 2018-08-15
Fix filter_block_not_found error message

sorpaas:
Address grumble https://github.com/paritytech/parity-ethereum/pull/9256#pullrequestreview-146520420
cc @jimpo

### DIFF
--- a/rpc/src/v1/helpers/errors.rs
+++ b/rpc/src/v1/helpers/errors.rs
@@ -426,7 +426,7 @@ pub fn filter_not_found() -> Error {
 pub fn filter_block_not_found(id: BlockId) -> Error {
 	Error {
 		code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST), // Specified in EIP-234.
-		message: "One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found".into(),
+		message: "One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found, or fromBlock and toBlock are out of order".into(),
 		data: Some(Value::String(match id {
 			BlockId::Hash(hash) => format!("0x{:x}", hash),
 			BlockId::Number(number) => format!("0x{:x}", number),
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/errors.rs
```diff
@@ -426,7 +426,7 @@ pub fn filter_not_found() -> Error {
 pub fn filter_block_not_found(id: BlockId) -> Error {
 	Error {
 		code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST), // Specified in EIP-234.
-		message: "One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found".into(),
```
ordian: I would prefer having a separate message for that error, something like that:
```rust
pub fn filter_block_error(id: BlockId, msg: &str) -> Error {
	Error {
		code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST), // Specified in EIP-234.
		message: msg.into(),
		data: Some(Value::String(match id {
			BlockId::Hash(hash) => format!("0x{:x}", hash),
			BlockId::Number(number) => format!("0x{:x}", number),
			BlockId::Earliest => "earliest".to_string(),
			BlockId::Latest => "latest".to_string(),
		})),
	}
}
pub fn filter_block_not_found(id: BlockId) -> Error {
	filter_block_error(id, "One of ...")
}
pub fn filter_block_out_of_order(id: BlockId) -> Error {
	filter_block_error(id, "fromBlock and toBlock are out of order")
}
```

sorpaas: I think this needs a custom error type in `ethcore`, and we cannot use `?` syntax any more for the `Client::logs` function. Not sure it would worth it or not, but let me give a try.

sorpaas: I think this would indeed need some amounts of change in `Client` error types, or we would need to duplicate the checks in `Client::logs` in RPC again (which I don't think is a good idea). For the former, I think it may be better to fix this together with a `Client` and `Blockchain` struct error type overhaul, so that we can preserve using `?` syntax there. So I'm going to temporarily close this.

I created an issue #9373.

PR #9360, opened 2018-08-16
Resumable EVM and heap-allocated callstack

### LINKED ISSUE #6744: evm interface overhaul

sorpaas:
Mostly fixes #6744. All basic things for this PR are fixed now.

* The VM interface is changed to three traits -- `Exec`, which starts a VM execution, and can return a resumable trap, `ResumeCall`/`ResumeCreate` to resume execution from call or create traps. Those are accessed using trait objects.
* `Externalities::call/create` now has an additional parameter `trap`. If `trap` is `true` (the case for EVM), then this function would return a trap. If `trap` is `false` (the case for Wasm), then it will directly execute the context.
* Split `call/create` calls in `Executive` to use a separate struct `CallCreateExecutive`. This is the main struct that handles all resumes. One can use `CallCreateExecutive::exec/resume_call/resume_create` to get the resume behavior, or use `CallCreateExecutive::consume` to get the non-resume behavior.
* A refactoring on tracing mod. This makes it to use a single struct to record subtracing infos so that executive does not need extra work to record them.

Things to be done:

- [x] Tracing mod is not working yet.
- [x] Fix all current tests and write tests for resumes.
- [x] Fix all `TODO` proof messages.
- [x] Find a way to insert `crossbeam` again. In rare cases we can still exceeds current call stack limit (if too many Wasm contexts are stacked).
- [x] ~Provide a compile feature flag to enable wasm trap. With pause PR merged to wasmi, that's doable.~ (This PR is already getting big, let's do this in another PR.)
- [x] Make sure we don't have extra clones.


sorpaas:
I think this PR is ready for review now.

debris:
@tomusdrw let's review this together by the end of this week :)

ethcore/src/executive.rs
```diff
@@ -165,6 +189,559 @@ impl TransactOptions<trace::NoopTracer, trace::NoopVMTracer> {
...
+						self.kind = CallCreateExecutiveKind::ResumeCreate(origin_info, resume, unconfirmed_substate);
+						return Err(TrapError::Create(subparams, address, self));
+					},
+				};
+
+				Self::enact_result(&res, state, substate, unconfirmed_substate);
+				Ok(res)
+			},
+			CallCreateExecutiveKind::ResumeCreate(..) =>
+				panic!("Resumable as create, but called resume_call"),
```
debris: is there any way we can avoid those panics?

ethcore/src/trace/executive_tracer.rs
```diff
@@ -17,159 +17,166 @@
...
+			*parentlen += 1;
+		}
+
 		let trace = FlatTrace {
-			subtraces: top_level_subtraces(&subs),
-			action: Action::Create(create.expect("self.prepare_trace_create().is_some(): so we must be tracing: qed")),
+			trace_address: self.index_stack.clone(),
+			subtraces: self.sublen_stack.last().cloned().unwrap_or(0),
+			action: Action::Create(Create::from(params.clone())),
 			result: Res::Create(CreateResult {
```
debris: maybe it's worth introducing a third variant? or redesigning the implementation to avoid dummy data creation?

ethcore/src/executive.rs
sorpaas: We can create traits similar to what we do for vm:

```
pub trait Exec: Send {
	fn exec(self: Box<Self>, state: &mut State<B>, ...) -> ExecutiveTrapResult<GasLeft>;
}

pub trait ResumeCall: Send {
	fn resume_call(self: Box<Self>, state: &mut State<B>, ...) -> Box<Exec>;
}

pub trait ResumeCreate: Send {
	fn resume_create(self: Box<Self>, state: &mut State<B>, ...) -> Box<Exec>;
}
```

However, they would need to use traits objects and dynamic dispatch, which is a small amount of runtime overhead. I think it may be a good idea to only use those when in the future we want to expose `Executive` to be some sort of public API.

ethcore/src/trace/executive_tracer.rs
sorpaas: I think this may be okay. `U256::zero()`, `Vec::new()`, `Address::default()` all would just zero out the fields, and wouldn't allocate anything on the heap.

* From performance's perspective, it wouldn't make any difference if we do dummy data creation or use `Option`, they would both just zeroing out the fields.
* From clarity's perspective, this dummy data creation is restraint in `ExecutiveTracer` and doesn't add any complexity outside of it.

5chdn:
Could I get a 2nd review here @fckt @andresilva :)

ethcore/src/trace/executive_tracer.rs
```diff
@@ -17,159 +17,166 @@
...
-		debug!(target: "trace", "Traced failed call {:?}", trace);
-		self.traces.push(trace);
-		self.traces.extend(prefix_subtrace_addresses(subs));
+	fn done_trace_call(&mut self, gas_used: U256, output: &[u8]) {
+		if self.skip_one {
+			self.skip_one = false;
+			return;
+		}
+
+		let vecindex = self.vecindex_stack.pop().expect("prepare/done_trace are not balanced");
```
NikVolf: would be nice to write why they are always balanced

debris:
🎉

Issue #8976, opened 2018-06-25
Light client gas estimation does not work


danuker:
> I'm running:
>
> - **Which Parity version?**: 1.11.4-beta-cc44ae9cb-20180619/x86_64-linux-gnu/rustc1.26.1
> - **Which operating system?**: Linux Debian
> - **How installed?**: via binaries (Deb packages)
> - **Are you fully synchronized?**: yes (in light mode)
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

I am running the light client (`--light` CLI option).

**Steps to reproduce**
When I try to get a gas estimation for a plain transaction, I get a weird "Unknown block number" error.
```
> curl --data '{"method":"eth_estimateGas","params":[{"from":"0x_my_address_here"}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545

{"jsonrpc":"2.0","error":{"code":-32602,"message":"Unknown block number"},"id":1}
```

**Actual behavior**
```{"jsonrpc":"2.0","error":{"code":-32602,"message":"Unknown block number"},"id":1}```

**Expected behavior**
I should either get a real gas estimate, or a useful error message.

**Comment by danuker** [comment]
Note: While waiting for a "real" sync, I can get a proper response to `eth_estimateGas`.

**Comment by Tbaut** [comment]
Thanks for reporting the issue.
Can you try with "latest":
```
curl --data '{"method":"eth_estimateGas","params":[{"from":"0x_my_address_here"}, "latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545

```

**Comment by danuker** [comment]
I tried it. Same response:
```
> curl --data '{"method":"eth_estimateGas","params":[{"from":"0x_my_address_here"}, "latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8544
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Unknown block number"},"id":1}
```

danuker:
Seems to work on Kovan for me as well.

niklasad1:
@danuker @Tbaut 

I can reproduce this on the  `main net` what I have discovered is that the merkle proof fails because your `account` is not in the `state db` and then it will just return an empty `account`!

[...]

PR #9370, opened 2018-08-17
Light client `Provide default nonce in transactions when it´s missing`

niklasad1:
Closes https://github.com/paritytech/parity-ethereum/issues/8976

When the `account nonce` is missing in a `EthTransaction` will cause it to fall in
these cases provide `default_nonce` value instead!

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -207,7 +207,7 @@ impl LightFetch {
 			}
 		};
 
-		let from = req.from.unwrap_or(Address::zero());
+		let from = req.from.unwrap_or_else(|| Address::zero());
 		let nonce_fut = match req.nonce {
 			Some(nonce) => Either::A(future::ok(Some(nonce))),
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -235,25 +235,25 @@ impl LightFetch {
 			let action = req.to.map_or(Action::Create, Action::Call);
 			let value = req.value.unwrap_or_else(U256::zero);
 			let data = req.data.unwrap_or_default();
+			let nonce = nonce.unwrap_or_default();
 
 			future::done(match (nonce, req.gas) {
-				(Some(n), Some(gas)) => Ok((true, EthTransaction {
+				(n, Some(gas)) => Ok((true, EthTransaction {
```
ordian: Can we avoid the match altogether? 
```rust
Ok((req.gas.is_some(), EthTransaction {
	gas: req.gas.unwrap_or_else(|| START_GAS.into()),
	...
}
```

Issue #9306, opened 2018-08-08
Invalid old blocks


ascjones:
> I'm running:
>
> - **Which Parity version?**: 2.0.1
> - **Which operating system?**: Linux
> - **How installed?**: binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

After warp sync completes, old/ancient blocks are downloaded. On my node, starting at `#1` some of those blocks are invalid, followed by valid blocks. Originally this was true only for `#1`, all other blocks that I queried were valid. After a fresh sync, the same occurred except the range `#1..#57` was invalid:

**Block 1** :x:
Expected [0x88e96d4537bea4d9c05d12549907b32561d3bf31f45aae734cdc119f13406cb6](https://etherscan.io/block/1)
```
curl -sX POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x1", false], "id": 75}' localhost:8545 | jq ".result.hash"
"0x49077beab2fa25f648dab021318e203f9f4dc6aed7e4b6b26526f63673b8ef41"
```

**Block 57** :x:
Expected [0xf394e8a094b2c701e66c8074ee7da247b676f71ba1633a68435f5e2aee975c5c](https://etherscan.io/block/57)
```
curl -sX POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x39", false], "id": 75}' localhost:8545 | jq ".result.hash"
"0xa638e3dd16207e5cebab1b057e0e9b97762321df22b7cd28ff0a6cdc7b78a7d1"
```

**Block 58** :heavy_check_mark: 
Expected [0xcd5b5c4cecd7f18a13fe974255badffd58e737dc67596d56bc01f063dd282e9e](https://etherscan.io/block/58)
```
curl -sX POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x3a", false], "id": 75}' localhost:8545 | jq ".result.hash"
"0xcd5b5c4cecd7f18a13fe974255badffd58e737dc67596d56bc01f063dd282e9e"
```
[...]

PR #9381, opened 2018-08-20
Prevent old blocks sync restart if import queue full

ascjones:
This PR prevents ancient blocks sync restarting when the `IoChannelQueue` becomes full. This was causing some invalid old blocks to be inserted (see #9306). This fix was suggested by @tomusdrw:

1. Handle the QueueFull error
2. Just ignore the blocks that couldn't be imported
3. Re-try importing from the same point a bit later (hoping that the blocks in the queue were imported in the meantime)

*edit* Only partially addresses #9306. It prevents the sync from restarting from the `queue full` error, but it will still eventually retract back to genesis (see #9306 for explanation). 

@andresilva has a fix for this and will submit a follow up PR.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -16,7 +16,6 @@
 
 use std::collections::{HashSet, BTreeMap, VecDeque};
 use std::cmp;
-use std::fmt;
 use std::str::FromStr;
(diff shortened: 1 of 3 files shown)


5chdn:
Is that an approval in context of #9407, @tomusdrw ?

andresilva:
@5chdn Yes, #9407 descends from this branch and is what I tested.

Issue #9379, opened 2018-08-20
docker build error Cargo.toml


sicute:
Try docker build with ubuntu .
`docker build -f Dockerfile --tag ethcore/parity:latest .`
```
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.10)
Removing intermediate container 0a533e4fd11c
 ---> 9f1643e1809b
Step 8/12 : ADD . /build/parity
 ---> df366e04bb98
Step 9/12 : RUN cd parity &&         cargo build --release --verbose &&         ls /build/parity/target/release/parity &&         strip /build/parity/target/release/parity
 ---> Running in a922b3dd1e0f
error: could not find `Cargo.toml` in `/build/parity` or any parent directory
The command '/bin/sh -c cd parity &&         cargo build --release --verbose &&         ls /build/parity/target/release/parity &&         strip /build/parity/target/release/parity' returned a non-zero code: 101
ubuntu@ip-172-31-93-216:~/docker/ethparity$
```

Tbaut:
Can you share the docker file you're using?

sicute:
i follow this Dockerfile 
https://github.com/paritytech/parity-ethereum/blob/master/docker/ubuntu/Dockerfile

Tbaut:
Thanks, this one is not maintained it seems, please use https://github.com/paritytech/parity-ethereum/blob/master/docker/hub/Dockerfile

PR #9386, opened 2018-08-21
Delete obsolete Dockerfile

Tbaut:
- This dockerfile was unmaintained and we have a similar (maintained) one in /hub
- Fix https://github.com/paritytech/parity-ethereum/issues/9379


Tbaut:
hub, alpine,android got created/updated in the past months. Regarding 14.04 that's your call..

PR #3903, opened 2016-12-19
fix deadlock in queue drop

rphmeier:
Closes #3885

PR #9392, opened 2018-08-21
Revert "Use std::sync::Condvar (#1732)"

ordian:
Reverts https://github.com/paritytech/parity-ethereum/pull/1732. This would hopefully allow us to detect deadlocks via `--feature "deadlock_detection"`.

`parking_lot_core`  now [uses](https://github.com/Amanieu/parking_lot/commit/c28f656d9efe3f233a078facdd9a19f4219ed9eb) WaitOnAddress/WakeByAddress on Windows, so I guess the reason not to use `parking_lot::Condvar` is no longer relevant.

### DIFF
--- a/ethcore/src/verification/queue/mod.rs
+++ b/ethcore/src/verification/queue/mod.rs
@@ -19,7 +19,7 @@
 
 use std::thread::{self, JoinHandle};
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering as AtomicOrdering};
-use std::sync::{Condvar as SCondvar, Mutex as SMutex, Arc};
+use std::sync::Arc;
 use std::cmp;
 use std::collections::{VecDeque, HashSet, HashMap};
 use heapsize::HeapSizeOf;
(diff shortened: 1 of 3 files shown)


ethcore/src/verification/queue/mod.rs
```diff
@@ -202,8 +202,8 @@ struct Verification<K: Kind> {
 	verifying: Mutex<VecDeque<Verifying<K>>>,
 	verified: Mutex<VecDeque<K::Verified>>,
 	bad: Mutex<HashSet<H256>>,
-	more_to_verify: SMutex<()>,
-	empty: SMutex<()>,
+	more_to_verify: Mutex<()>,
```
andresilva: It seems like these locks were previously added so that they could be used with `std::sync::Condvar`. Previously the `Condvar` waited on `unverified`. Do you think we could remove these?

ordian: Good question! Did that at first, but then I saw [another PR](https://github.com/paritytech/parity-ethereum/pull/3903/files?utf8=%E2%9C%93&diff=split&w=1#diff-afe0a092320d9405848cfdc98328afefR687) using `more_to_verify`, and it wasn't obvious if they can be safely removed. Anyway, it seems to work w/o them, so I'm going to push another commit with the removal.

5chdn:
wow, reverting a 26 months old PR :+1:

PR #9404, opened 2018-08-23
Add a new RPC `parity_submitWorkDetail` similar `eth_submitWork` but return block hash

YihaoPeng:
### Purpose

There is a problem with Ethereum's current RPC `eth_submitWork`: the miner cannot know the real block hash after a submission, so it is impossible to query the block from a blockchain browser or perform any subsequent automated processing.

So I added a new RPC `parity_submitWorkDetail` to Parity. It is similar to `eth_submitWork` but will return the block hash on success, and return an explicit error message on failure.

### Params and Response
```js
  // RPC parity_submitWorkDetail

  // Params (same as `eth_submitWork`):
       [
           "<nonce>",
           "<pow_hash>",
           "<mix_hash>"
       ]
   
  // Result on success:
       "block_hash"
  
  // Error on failure:
       {code: -32005, message: "Cannot submit work.", data: "<reason for submission failure>"]
```

### Example:
call RPC with curl:
```bash
curl --data-binary '{"jsonrpc": "2.0", "method": "parity_submitWorkDetail", "params": ["0xccfb958c9f08dc52","0x86f29a6fb5d3fa0b8e32e9c1c1d53f2496a91d414d5707f7c88da39c109ceb66","0x95fc749e53339cd9187a58d28d686ef7b1e9c3261c1c47046ea2873e8bf36684"], "id": 5}' -H 'content-type: application/json' http://localhost:8545/
```
response with error message:
```json
{"jsonrpc":"2.0","error":{"code":-32005,"message":"Cannot submit work.","data":"PoW hash is invalid or out of date."},"id":5}
```
response with block hash:
```json
{"jsonrpc":"2.0","result":"0x07a992176ab51ee50539c1ba287bef937fe49c9a96dafa03954fb6fefa594691","id":5}
```

### Personal Description
I am a complete Rust newcomer and this is the first time I read and change Rust's code. I have completed the function and tested it and it works fine. But I don't know if it is enough as a pull request.

If I need to do something else, please let me know.

### Testing and Usage
I tested it on the Ropsten chain and the submission worked fine. (This is a very simple patch and I think it may not require further testing.)

Open source POW pool BTCPool uses this patch to [record its submited block to database](https://github.com/btccom/btcpool/commit/02dcd607e73c39d6bd9a9fed4b9b52803714767f).

[BTC.COM pool](https://cn.pool.btc.com/pool-stats) has mined 17 blocks in Ethereum Mainnet with its Parity node with this patch.


sorpaas:
The issue is the same as #9126 -- it's recommended to have an EIP first before modifying the `eth_` jsonrpc namespace, in order to keep it tidy and to keep cross-client compatible.

However, if you want, you can change this PR to use `parity_` namespace and we may be able to accept it!

rpc/src/v1/impls/eth.rs
```diff
@@ -801,6 +801,36 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 		}
 	}
 
+	fn submit_work_detail(&self, nonce: RpcH64, pow_hash: RpcH256, mix_hash: RpcH256) -> Result<SubmitDetailResult> {
```
sorpaas: And just a small grumble -- we can refactor L806 to L813 to a helper function to avoid duplicate codes with [submit_work](https://github.com/paritytech/parity-ethereum/blob/34ea01b4804ab4b1020b097425d505e6af287fe9/rpc/src/v1/impls/eth.rs#L784).

YihaoPeng:
Does I need to move codes from `eth.rs` to `parity.rs`, or just rename RPC is ok?

rpc/src/v1/impls/eth.rs
YihaoPeng: thanks for your advice. I have modified it.

sorpaas:
@YihaoPeng For `parity_` namespace it would need to be moved to `parity.rs`.

YihaoPeng:
OK, I will try to move the code.

YihaoPeng:
Opps, a wrong click...

YihaoPeng:
@sorpaas The code is moved to `parity.rs` and the RPC is renamed to `parity_submitWorkDetail`.
However, I cannot avoid to duplicate codes because my low Rust skills. :sweat_smile:
I don't know how to call a function in another trait and whether this good.

rpc/src/v1/impls/parity.rs
```diff
@@ -449,4 +452,34 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
...
+		let nonce: EthH64 = nonce.into();
+		let pow_hash: EthH256 = pow_hash.into();
+		let mix_hash: EthH256 = mix_hash.into();
+		trace!(target: "miner", "submit_work_detail: Decoded: nonce={}, pow_hash={}, mix_hash={}", nonce, pow_hash, mix_hash);
+
+		let seal = vec![rlp::encode(&mix_hash).into_vec(), rlp::encode(&nonce).into_vec()];
+		let import = self.miner.submit_seal(pow_hash, seal)
+			.and_then(|block| self.client.import_sealed_block(block));
+
+		match import {
```
sorpaas: JSONRPC supports errors, so there's no need for `SubmitDetailResult` type. Add a new error helper in `rpc/src/v1/helpers/errors.rs`, and then change this function's return type to `Result<H256>`. You can then change the return value to something like `import.map_err(errors::submit_work)`.

YihaoPeng: I have completed the change.

rpc/src/v1/impls/parity.rs
```diff
@@ -46,7 +47,8 @@ use v1::types::{
 	BlockNumber, ConsensusCapability, VersionInfo,
 	OperationsInfo, ChainStatus,
 	AccountInfo, HwAccountInfo, RichHeader,
-	block_number_to_id
+	block_number_to_id,
+	H64 as RpcH64, H256 as RpcH256,
```
dvdplm: Why was this double-import of `v1::types::H256` necessary? The `v1::types::` primitives have conversion facilities in place to convert to and from `ethereum_types::H*` (unrelated to this PR: I suspect that is not optimal to have different hash types) so I think it might work converting between them directly.

SwimmingTiger: It's my wrong. I have removed the double imported `v1::types::H256`.

5chdn:
Can you have another look at this @sorpaas @dvdplm :)

rpc/src/v1/impls/parity.rs
```diff
@@ -449,4 +450,26 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 				.map(|res| res.into_iter().map(|res| res.output.into()).collect())
 				.map_err(errors::call)
 	}
+
+	fn submit_work_detail(&self, nonce: H64, pow_hash: H256, mix_hash: H256) -> Result<H256> {
```
sorpaas: There's currently some duplicate code for `parity_submitWorkDetail` and `eth_submitWork`. We can put this function into a helper and call it in both `parity_submitWorkDetail` and `eth_submitWork`. Something like this:

In `rpc/src/v1/helpers/...rs`:

```
pub fn submit_work_detail(miner: _, client: _, ...) -> Result<H256> {
  ...
}
```

In `parity.rs`:

```
fn submit_work_detail(&self, nonce: H64, pow_hash: H256, mix_hash: H256) -> Result<H256> {
  helpers::submit_work_detail(self.miner, self.client, nonce, pow_hash, mix_hash)
}
```

In `eth.rs`:

```
fn submit_work(&self, nonce: RpcH64, pow_hash: RpcH256, mix_hash: RpcH256) -> Result<bool> {
  match helpers::submit_work_detail(self.miner, self.client, nonce, pow_hash, mix_hash) {
    Ok(_) => Ok(true),
    Err(_) => Ok(false),
  }
}
```

YihaoPeng: I am trying to make such a change, but I am having trouble.
The question is what type of miner and client should be.
I noticed that in paraity.rs and eth.rs they are all defined as generics and two definitions are different.
I tried to use generics but I encountered an error like this:
```
pub fn submit_work_detail<C: BlockChainClient, M: MinerService>(client: C, miner: M, nonce: H64, pow_hash: H256, mix_hash: H256) -> Result<H256, Error> {...}
```

```
The trait bound `std::sync::Arc<C>: ethcore::miner::BlockChainClient` is not satisfied
rpc/src/v1/impls/eth.rs:785:9
the trait `ethcore::miner::BlockChainClient` is not implemented for `std::sync::Arc<C>`
```

YihaoPeng: OK, I have got it. This works:
```
pub fn submit_work_detail<C: BlockChainClient, M: MinerService>(client: &Arc<C>, miner: &Arc<M>, nonce: H64, pow_hash: H256, mix_hash: H256) -> Result<H256, Error> {...}
```

5chdn:
Please rebase on master

SwimmingTiger:
I tried rebase but every commit needs to resolve the conflict. A merge or sqush is obviously easier and can be done automatically.

sorpaas:
@SwimmingTiger I think you'll just need to do a merge (`git pull origin master`). We squash merge PRs anyway so merge commits won't matter. :)

SwimmingTiger:
OK, I have pushed the merging.

PR #9406, opened 2018-08-23
ethcore: kovan: delay activation of strict score validation

### LINKED ISSUE #9403: Stage 3 block verification failed for #1000002 on Kovan with tracing turned on
> I'm running:
>
> - **Which Parity version?**: Parity-Ethereum/v2.0.1-beta-e7dc0be-20180726/x86_64-macos/rustc1.27.2
> - **Which operating system?**: MacOS
> - **How installed?**: homebrew
> - **Which network are you connected to?**: kovan

State DB configuration: fast +Trace
Operating mode: dark
Warning: Warp Sync is disabled because tracing is turned on.
Configured for Kovan using AuthorityRound engine

Hitting this error:
```
Stage 3 block verification failed for #1000002 (0x2c27…b9a6)
Error: Error(Block(InvalidDifficulty(Mismatch { expected: 0xfffffffffffffffffffffffffffffffd, found: 0xfffffffffffffffffffffffffffffffe })), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })
```
This does not occur with tracing turned off. I've tried killing the database and restarting but it hits this error again.

andresilva:
Fix #9403.

Recently we added a stricter validation of block score for AuthorityRound chains (#9164). But in the past we had a race condition that lead to generating some blocks with invalid difficulty (#7198). Taking into account the date of that PR, it would be expected for the issue to not exist past around block #5000000 on Kovan. Still, I synced the chain, increasing the transition block every time I found an invalid block and the issue was rare enough that I think it explains why it seems to have stopped happening before the fix was released.

I'm currently syncing Kovan at around block #6000000 and haven't seen any block that fails validation past #4301764.

### DIFF
--- a/ethcore/res/ethereum/kovan.json
+++ b/ethcore/res/ethereum/kovan.json
@@ -19,7 +19,7 @@
 						"0x00a0a24b9f0e5ec7aa4c7389b8302fd0123194de"
 					]
 				},
-				"validateScoreTransition": 1000000,
+				"validateScoreTransition": 4301764,
 				"validateStepTransition": 1500000,
 				"maximumUncleCountTransition": 5067000,
 				"maximumUncleCount": 0


andresilva:
I would appreciate it if someone else could sync Kovan from scratch with this change.

5chdn:
with --no-warp and --tracing on?

andresilva:
Yes, with `--no-warp`. Tracing isn't relevant in this case, the original issue should happen with and without tracing.

andresilva:
Just finished a full sync with these changes.

PR #9410, opened 2018-08-24
Fetch `parity-common` crates from crates.io

dvdplm:
Also removes `transaction-pool` and `trace-time` which now are in `parity-common` (and published).

~~Blocked on crate ownership issues: `plain_hasher` (cc @debris)~~

~~Blocked on https://github.com/paritytech/ethabi/pull/117 (also see https://github.com/paritytech/ethabi/issues/118)~~

~~Blocked on https://github.com/paritytech/rust-rocksdb/issues/16~~


sorpaas:
I would worry about "atomic PR" issues, like the one we faced in #9257.

Have we considered the [git submodule approach](https://github.com/paritytech/parity-ethereum/pull/9257#discussion_r206544383)?

dvdplm:
> Have we considered the git submodule approach?

I for one really (really) dislike git submodules and I think once we're "over the hump" and all `parity-common` crates are published we'll be mostly fine. I think we overdid it a little on https://github.com/paritytech/parity-ethereum/pull/9257 tbh.

registrar/src/registrar.rs
```diff
@@ -52,7 +52,9 @@ impl Registrar {
 		};
 
 		let address_fetcher = self.registrar.functions().get_address();
-		let id = address_fetcher.input(keccak(key), DNS_A_RECORD);
+		// TODO: keccak(), from `keccak-hash`, uses `ethereum-types 0.4 which is different from what ethabi 5.x uses so keccak_hasher::H256 != ethereum_types::H256 – I think the only way out is to upgrade ethabi as well. :/
```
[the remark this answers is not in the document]
dvdplm: removed.

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -795,7 +795,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 						TWO_POW_96 => a >> 96,
 						TWO_POW_224 => a >> 224,
 						TWO_POW_248 => a >> 248,
-						_ => a.overflowing_div(b).0,
+						_ => a/b,
```
andresilva: `a / b`?

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -821,7 +821,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 				} else if a == min && b == !U256::zero() {
 					min
 				} else {
-					let c = a.overflowing_div(b).0;
+					let c = a/b;
```
andresilva: `a / b`

registrar/src/registrar.rs
```diff
@@ -52,7 +52,8 @@ impl Registrar {
 		};
 
 		let address_fetcher = self.registrar.functions().get_address();
-		let id = address_fetcher.input(keccak(key), DNS_A_RECORD);
+		let hashed_key : [u8; 32] = keccak(key).into();
```
andresilva: `let hashed_key: [u8; 32] = ...`

util/keccak-hasher/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 
 [dependencies]
-ethereum-types = "0.3"
+ethereum-types = "0.4"
 tiny-keccak = "1.4.2"
-hashdb = { git = "https://github.com/paritytech/parity-common" }
-plain_hasher = { git = "https://github.com/paritytech/parity-common" }
\ No newline at end of file
+hashdb = "0.2.1"
+plain_hasher = "0.2"
```
andresilva: Can we add a newline at the end of this file?

util/patricia-trie-ethereum/Cargo.toml
```diff
@@ -6,10 +6,14 @@ description = "Merkle-Patricia Trie (Ethereum Style)"
...
+keccak-hasher = { version = "0.1.1", path = "../keccak-hasher" }
+hashdb = "0.2"
+rlp = { version = "0.2.4", features = ["ethereum"] }
+parity-bytes = "0.1"
+ethereum-types = "0.4"
+elastic-array = "0.10"
+
+[dev-dependencies]
+memorydb = "0.2.1"
+keccak-hash = "0.1.2"
```
andresilva: Can we add a newline here as well?

util/triehash-ethereum/Cargo.toml
```diff
@@ -6,6 +6,6 @@ description = "Trie-root helpers, ethereum style"
 license = "GPL-3.0"
 
 [dependencies]
-triehash = { git = "https://github.com/paritytech/parity-common" }
-ethereum-types = "0.3"
+triehash = { version = "0.2.3", features = ["ethereum"] }
+ethereum-types = "0.4"
 keccak-hasher = { path = "../keccak-hasher" }
```
andresilva: Also newline at end of file?

ethcore/evm/src/interpreter/mod.rs
debris: why is it changed?

sorpaas: @dvdplm Please all those cases back. `Div` panics on default but we need overflowing behavior here, otherwise it breaks consensus.

andresilva: It seems like these methods were removed when going from `bigint` to `uint`, which makes sense since for unsigned integers div and rem should never overflow.

andresilva: I also had the same question though so it's better if @dvdplm confirms.

sorpaas: Yeah right. I think it's this: https://github.com/paritytech/parity-common/pull/32

dvdplm: Confirmed, that's the PR. I think the reason they were there to begin with was an attempt to fulfill a trait that is no longer in rust (removed ~1.13?).

dvdplm:
> all dependencies have proper version tags on parity-common repo or other places.

I agree with this strongly and have slapped myself several times already for not having done this already. I will go through `parity-common` and make sure it's done and make sure others do the same. In `parity-common` I have to make tags like `kvdb-v0.1` and it'll apply to the whole repo ofc but that's not a big problem I think.

andresilva:
I built a binary from this branch and synced mainnet (my node was about 4000 blocks away from the chain head) and I also warp-synced on kovan.

Issue #9389, opened 2018-08-21
Light client - no error when there is insufficient balance


aghamir:
> I'm running:
>
> - **Which Parity version?**: v2.0.1-beta-e7dc0be-20180726
> - **Which operating system?**: Linux
> - **How installed?**: via installer
> - **Are you fully synchronized?**: no
> - **Which network are you connected to?**: ropsten 
> - **Did you try to restart the node?**: no

When I use `personal_sendtransaction` via light client json rpc, and set value more than the balance of account, it creates transaction without any error. However, as expected, it won't go to blockchain. I expected that json rpc send an error due to lack of balance in account.

Tbaut:
Confirmed, on a full node I get :
```
{"jsonrpc":"2.0","error":{"code":-32010,"message":"Insufficient funds. The account you tried to send transaction from does not have enough funds. Required 4294967296009400000000000000 and got: 501690007696599545624."},"id":1}
```

niklasad1:
@aghamir 

Well, I think the `light-client` is using the ```full-client´s``` implementation unintentionally! It can be found here: https://github.com/paritytech/parity-ethereum/blob/master/parity/rpc_apis.rs#L444-#L528. 

Thus, this should not be supported yet if I have understood this correctly!

/cc @rphmeier @tomaka Is my understanding correct?

niklasad1:
@rphmeier 

Yes, that is true but it imports the `PersonalApi` from the `full node` implementation i.e, from ```rpc/src/v1/impls/personal.rs``` instead from `rpc/src/v1/impls/light/` (it not implemented here). 

Is it intended?

aghamir:
This issue is not fixed yet. It created transaction and wait to sufficient balance available.

niklasad1:
@aghamir Can you report which version of parity you are running and how you produced it please?

If I run a transaction on https://github.com/paritytech/parity-ethereum/commit/9f1af6b3e8c1689940dbbaab1ffacfee3ca16142 I get the following:

```bash
curl --data '{"method":"personal_sendTransaction","params":[{"from":"from", "to":"to", "data":"0x41cd5add4fd13aedd64521e363ea279923575ff39718065d38bd46f0e6632e8e","value":"0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"},""],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8555
[...]

PR #9417, opened 2018-08-27
[light] Validate `account balance` before importing transactions

niklasad1:
Attempt to fix https://github.com/paritytech/parity-ethereum/issues/9389

### Abstract
Basically, we want the same behavior in the `full client` and the `light client` when a transaction has insufficient funds we should deny the transaction early. However, the transaction queues differ because in `light-mode` we don't have access to the `state trie` and don´t cache it either.

Thus, we need to check the balance before importing transactions to the ```light transaction queue```!

### What does this PR
Before signing each transaction we need to fetch the `next_nonce` to fill in the transaction.
It can be performed in two ways:

1. The account is in the transaction queue or already provided, then just fill the nonce
2. Make a network request to get the account state

However, if we use option 1) we don't get the balance!

So instead, now we need always go out the network and can't use the cached things for signing new transactions inorder to validate the balance!

### Testing
```bash
curl --data '{"method":"personal_sendTransaction","params":[{"from":"from","to":"to","data":"0x41cd5add4fd13aedd64521e363ea279923575ff39718065d38bd46f0e6632e8e","value":"0x186a0"},"password"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","error":{"code":-32010,"message":"Insufficient funds. The account you tried to send transaction from does not have enough funds. Required 10645936000100000 and got: 0."},"id":1}
```


rphmeier:
> The account is in the transaction queue

The nonces in the transaction queue may be outdated, as they are only refreshed periodically. It might be possible that we always get an up-to-date nonce with this approach, but probably not if there are multiple computers that use the same account.

niklasad1:
@rphmeier 

Ok, I guess to only way to ensure that the nonce is ```fresh``` is to query to the network. 

Should we remove relying on `cached nonces` from the transaction queue and send a network request each time (i.e, remove ```cached_next_nonce()``` and only use ```next_nonce()```)?

rphmeier:
Fetching from the queue should be fine, but we should file an issue about migrating to the generic transaction-queue and querying the network as necessary.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -302,27 +302,37 @@ impl LightDispatcher {
...
+		let account_start_nonce = self.client.engine().account_start_nonce(self.client.best_block_header().number());
+		Box::new(self.account(addr.clone())
+			.and_then(move |maybe_account| {
+				future::ok(maybe_account.map_or(account_start_nonce, |account| account.nonce))
+			})
+		)
+	}
+
+	/// Try fetching the next nonce from the cache
+	pub fn cached_next_nonce(&self, addr: &Address) -> Result<U256> {
```
fckt: Is there any good reason why we use `Result` here instead of `Option`?

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -371,7 +381,13 @@ impl Dispatcher for LightDispatcher {
 		match (request_nonce, force_nonce) {
 			(_, false) | (Some(_), true) => Box::new(gas_price),
 			(None, true) => {
-				let next_nonce = self.next_nonce(from);
+				let next_nonce = {
+					if let Ok(nonce) = self.cached_next_nonce(&from) {
```
fckt: is there any chance to incapsulate this particular logic?

fckt:
> The nonces in the transaction queue may be outdated, as they are only refreshed periodically. It might be possible that we always get an up-to-date nonce with this approach, but probably not if there are multiple computers that use the same account.

So that means, in order to have a robust nonce sync, we should sync it between nodes..

rpc/src/v1/helpers/dispatch.rs
niklasad1: No, because the error is unused it makes no sense to use a Result but it was for convenience before refactoring!

But I will get rid of this method entirely instead to ensure to we have a fresh `nonce` instead!

rpc/src/v1/helpers/dispatch.rs
niklasad1: This will go away too, but I just realized that this might perform two network requests in case the nonce is not cached and I need to changer this, thanks

niklasad1:
Yepp, I think is sufficient to fetch the next nonce from the network because we still have to do it in order to check the account balance. However, I suppose it is possible that to have races before an account is actually present in the state trie or at least two concurrent queries/ transactions!

But, if we always query the network I guess the nonce cache becomes needless and the getNextNonce RPC might suffer from worse performace!

fckt:
> Syncing the nonce between nodes is probably tricky.

Just wonder, have we ever considered https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type to be applicable in such cases?

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -302,40 +302,42 @@ impl LightDispatcher {
...
-			),
-			None => Box::new(future::err(errors::network_disabled()))
+		match account_future {
+			Some(response) => Box::new(response.map_err(|_| errors::no_light_peers())),
+			None => Box::new(future::err(errors::network_disabled())),
 		}
 	}
+
+	/// Get an account's next nonce.
+	pub fn next_nonce(&self, addr: &Address) -> BoxFuture<U256> {
```
fckt: Minor: maybe it's better to move `addr` here ..

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -302,40 +302,42 @@ impl LightDispatcher {
...
+		let account_start_nonce = self.client.engine().account_start_nonce(self.client.best_block_header().number());
+		Box::new(self.account(addr.clone())
```
fckt: .. to avoid `.clone`?
User could clone it anyway if it'll be necessary.

rpc/src/v1/helpers/dispatch.rs
niklasad1: Yepp, that makes sense thanks!

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -368,39 +370,38 @@ impl Dispatcher for LightDispatcher {
...
-					})
-				))
-			},
-		}
+		let future_account = self.account(from);
+
+		Box::new(gas_price.and_then(move |mut filled| {
+			future_account
+				.and_then(move |maybe_account| {
+					let cost = filled.value + filled.gas_price * filled.gas;
```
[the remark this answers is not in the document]
niklasad1: Yes

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -368,39 +370,37 @@ impl Dispatcher for LightDispatcher {
...
-		let nonces = self.nonces.clone();
-		Box::new(self.next_nonce(filled.from)
-			.map_err(|_| errors::no_light_peers())
-			.and_then(move |nonce| {
-				let reserved = nonces.lock().reserve(filled.from, nonce);
-
-				ProspectiveSigner::new(accounts, filled, chain_id, reserved, password)
-			}))
+		let nonce = filled.nonce.expect("nonce is always provided; qed");
+		Box::new(future::done(sign_transaction(&*accounts, filled, chain_id, nonce, password)))
```
niklasad1: I think this is a minor performance improvement compared to creating a `ProspectiveSigner` also before in worst case we could actually perform two network request to fetch the ```next_nonce``` i.e,  in ```fill_optional_fields and sign``` now it's always one, the one in ```fill_optional_fields```!

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -302,40 +302,42 @@ impl LightDispatcher {
...
-			),
-			None => Box::new(future::err(errors::network_disabled()))
+		match account_future {
+			Some(response) => Box::new(response.map_err(|_| errors::no_light_peers())),
+			None => Box::new(future::err(errors::network_disabled())),
 		}
 	}
+
+	/// Get an account's next nonce.
+	pub fn next_nonce(&self, addr: Address) -> BoxFuture<U256> {
```
niklasad1: This might be slower now because of not relying on the cached nonces anymore!

PR #9433, opened 2018-08-29
Bad blocks RPC + reporting

### LINKED ISSUE #8761: Feature-request: debug_getBadBlocks
Geth has an endpoint called `debug_getBadBlocks`, which is very useful for two purposes.

1. It can be used to detect consensus flaws more quickly, by correlating what one client considers bad blocks and what the other client has in it's chain. 
2. It can be used to analyze consensus flaws more easily, since even if I run a client that does not import a rejected block, it can still provide me with the block for analysis on why it was considered 'bad'. 

Geth implementation is [here](https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L1397).  See UPPDATE below
Response definition [here](https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L1391) See UPPDATE below 

Geth uses an LRU-cache with up to `10` bad blocks, but that's an implementation detail.

tomusdrw:
Closes #8761 

- Prints a verbose error message in case bad block is detected.
- Stores bad blocks in a cache, the cache is limited to take at most 8MBs (should we rather have a count limit?)
- Exposes bad blocks over RPC attempting to maintain geth's compatibility, but also returns all the same fields we return for regular block RPCs.
- Adds `reason` to the RPC response to see internal error message that occurred during import.


ethcore/src/client/bad_blocks.rs
```diff
@@ -0,0 +1,81 @@
...
+	fn default() -> Self {
+		BadBlocks {
+			last_blocks: RwLock::new(MemoryLruCache::new(8 * 1024 * 1024)),
+		}
+	}
+}
+
+impl BadBlocks {
+	/// Reports given RLP as invalid block.
+	pub  fn report(&self, raw: Bytes, message: String) {
```
sorpaas: Extra whitespace.

ethcore/src/client/client.rs
```diff
@@ -291,22 +296,27 @@ impl Importer {
...
 
-						client.report.write().accrue_block(&header, transactions_len);
-					}
-				} else {
-					invalid_blocks.insert(header.hash());
+							client.report.write().accrue_block(&header, transactions_len);
+						}
+					},
+					Err(err) => {
+						self.bad_blocks.report(raw, format!("{:?}", err));
```
sorpaas: Can we make this configurable? Such as only enable bad blocks reporting when the debug namespace RPC is enabled, or just use another cli flag.

rpc/src/v1/impls/debug.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use ethcore::client::BlockChainClient;
+use transaction::LocalizedTransaction;
+
+use jsonrpc_core::Result;
+use v1::traits::Debug;
+use v1::types::{Block, Bytes, RichBlock, BlockTransactions, Transaction};
+
+/// Debug rpc implementation.
+pub struct DebugClient<C> {
```
sorpaas: I think this mod needs to be enabled in `parity/rpc_apis.rs`? Or did I missed that?

ethcore/src/client/client.rs
tomusdrw: I think the whole point is to have it running by default, so that anyone (that have seen such message) can be asked to get bad blocks from RPC and then report them for the analysis.

rpc/src/v1/impls/debug.rs
tomusdrw: Oh, that's 100% true. Haven't tested it properly, will add to CLI and actually check if it's queryable through RPC.

holiman:
The reason to have this functionality is so that we can run one geth node and one parity node in paralell, and detect _immediately_ if a consensus fork has occurred (one rejecting a block that the other has accepted). 

The later steps of analyzing the error is not very difficult, although it definitely helps if the client that rejecteded the block also prints out the rlp, for analysis. 
 
> We are not limited to 8mb bad blocks. For sophisticated consensus attackers, they may find vulnerabilities in networking/importing first, and flood the 8mb struct with non-consensus-critical bad blocks, forcing the consensus-critical bad blocks to be dropped out of the 8mb cache.

So, of the cases where real-world consensus issues have happened, none of them have been at that level of sophistication -- afiak all of them have been incidental. But even so -- how could they be flooded with non-consensus-critical bad blocks? Those would be rejected due to low difficulty even before going into actual block execution.

holiman:
> @holiman What's the definition of a "bad block" then? In this PR we are reporting any block that fails the validation (even before execution) - so even low difficulty, invalid seal, etc. blocks would be reported here.

Actually, I'm not quite certain -- it's possible we do too. However, as you've  pointed out -- an attacker needs to specifically target the node, in order to get his bad blocks propagated.

holiman:
We report bad blocks here:https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L1136 (pre-check) and here: https://github.com/ethereum/go-ethereum/blob/master/core/blockchain.go#L1154 (processing)

holiman:
Some more context: we've had 'reporting' of bad blocks now for quite some time. By 'reporting' I mean printing out a warning about a bad block being encountered, and some details, in the log stream. 

This has triggered only a couple of times over the last couple of years, and each time it has been very good info. IIRC, that's what made people notice when Parity mined bad blocks on Ropsten due to the consensus-bug with having EIP-86 enabled. 

The added ability to actually get the same info over RPC will make it possible to get instant alerts immediately if a fork occurs. If the fork-monitor needs to parse log output for the same info, it will be quite messy. 

If you see fit to limit the badblock-cache even more, then please do so, but I'd really love to have this feature in, even if it's somewhat limited. 

Right now, our fork monitor (http://mon04.ethdevops.io/forkmon) would immediately detect if parity head contains something that geth rejects, but the opposite scenario would not be detected until later.

5chdn:
How does this relate to #9207 ?

tomusdrw:
@5chdn #9207 is around for more than a month and it's nowhere close to be neither correct nor finished - it stores `bad_blocks` hashes in the `Blockchain`, and the RPC attempts to retrieve them from the database, but those blocks are bad so they never even get to the db. While I appreciate the contribution, reviewing the PR is pretty much implementing that stuff in comments.

parity/rpc_apis.rs
```diff
@@ -647,6 +658,7 @@ impl ApiSet {
 				public_list
 			},
 			ApiSet::SafeContext => {
+				public_list.insert(Api::Debug);
```
andresilva: Why enable this only for safe context (given that bad blocks cache is enabled by default)? I understand that the `debug` namespace might have other methods in the future which are unsafe to expose.

util/memory_cache/src/lib.rs
```diff
@@ -86,6 +86,12 @@ impl<K: Eq + Hash, V: HeapSizeOf> MemoryLruCache<K, V> {
 	pub fn current_size(&self) -> usize {
 		self.cur_size
 	}
+
+	/// Get backing LRU cache instance (read only)
+	pub fn backstore(&self) -> &LruCache<K, V> {
+		&self.inner
+	}
+
```
andresilva: Unnecessary newline?

ethcore/src/client/bad_blocks.rs
```diff
@@ -0,0 +1,81 @@
...
+use bytes::{Bytes, ToPretty};
+use ethereum_types::H256;
+use itertools::Itertools;
+use memory_cache::MemoryLruCache;
+use parking_lot::RwLock;
+use verification::queue::kind::blocks::Unverified;
+
+/// Recently seen bad blocks.
+pub struct BadBlocks {
+	last_blocks: RwLock<MemoryLruCache<H256, (Unverified, String)>>,
```
fckt: How about to add the raw_block data (Vec<u8>) into the tuple also, to be able to save "undecodable" BadBlocks? Is that makes any sense?

tomusdrw: Yeah, I though about it actually. It doesn't allow us to return detailed block in the RPC response though. I think currently it makes sense to have something with the same results as geth to start detecting issues asap.

That said, feel free to log it as a separate issue.

parity/rpc_apis.rs
tomusdrw: Didn't enable for two reasons:
1. `debug` is potentially unsafe
2. It should never be used by regular dapps
3. We have this implemented to support detecting forks, I don't want to make it a first-class namespace where we implement all the methods, cause that most likely won't happen - the methods are not going through EIPs, and they are pretty much geth-specific debug RPCs.

util/memory_cache/src/lib.rs
tomusdrw: Fixed

ethcore/src/client/bad_blocks.rs
andresilva: I also thought about this and checked the geth implementation and it seems they also don't store block rlp's that aren't decodable.

tomusdrw:
To follow up on @sorpaas comment. We spoke through a different channel about this and agreed that it doesn't necessarily suits the core client.

The idea that came up was to expose structured logs via RPC (we already to that for recent logs, we could add a subscription), where one would be able to subscribe to some particular events happening in the core client.
Using events data we could build a separate binary that would be able to expose this (and similar) RPCs. Event sourcing would also be useful for all other stats and attaching to events at the runtime would also be quite useful for debugging purposes - imagine attaching to a running node to inspect incoming transactions or networking communication (currently it requires restarting with different logging flag).

PR #9436, opened 2018-08-30
docker: install missing dependencies in arm target dockerfiles

andresilva:


andresilva:
On xenial the cmake package is already at version 3 (https://packages.ubuntu.com/xenial/cmake).

PR #9438, opened 2018-08-30
Test PR for CI tracing...

ddorgan:


ddorgan:
Ah: If you know the job doesn't alter the cached files, you can skip the upload step by setting policy: pull in the job specification. Typically, this would be twinned with an ordinary cache job at an earlier stage to ensure the cache is updated from time to time

PR #9440, opened 2018-08-30
Update state tests execution model

### LINKED ISSUE #9431: New statetest execution model
A couple of months ago, the `stateTest` execution model was slightly changed, which I noticed when updating the `tests` within the geth repo. 

Some gitter discussion on the topic: https://gitter.im/ethereum/tests?at=5b85004aff445156164b685f 
Changes is to add a `0`-value mining reward (touch). Normally, the coinbase gets txfees during a statetest -- making it exist already. So this change only makes a difference in the cases
where
- the coinbase suicided, or
- there are only 'bad' transactions, which aren't executed. In those cases, the coinbase gets no txfee, so isn't created, and thus needs to be touched

Here are the relevant changes to Geth that was needed in order to make it work: https://github.com/ethereum/go-ethereum/pull/17538/files/548127685648625d8acb3e6ea1b479a2f2a167a9#diff-f53696be8527ac422b8d4de7c8e945c1R149 . 

It would be good to know if parity will implement this, or not. 

- If parity does not implement this, then geth shouldn't either, since the execution model of statetests is at the heart of the evmlab-based fuzzing engine, and differences here will yield tons of false positives. 

CC @winsvega to answer any questions about the reason for change.

tomusdrw:
Closes #9431 

- Touches the coinbase at the end of the execution
- Updates the state tests to a version that supports that
- As a consequence updates the transactions tests to a new format


dvdplm:
For reviewers it's important to note that you need to check out rev `b8a21c193696976ca3b33b6d82107601063a5d26` in `ethcore/res/ethereum/tests` or bad things happen.

ethcore/res/ethereum/eip161_test.json
```diff
@@ -24,7 +24,7 @@
 		"eip161dTransition": "0x0",
 		"eip98Transition": "0x7fffffffffffffff",
 		"eip86Transition": "0x7fffffffffffffff",
-		"eip155Transition": "0x7fffffffffffffff",
+		"eip155Transition": "0x0",
```
dvdplm: Setting to `0` enables the feature right?

tomusdrw:
@dvdplm it should be fine to just do `git submodule update` after `git pull`, since the correct submodule revision should be fetched automagically.

ethcore/src/json_tests/test_common.rs
```diff
@@ -51,6 +52,8 @@ pub fn run_test_path<H: FnMut(&str, HookType)>(
 			}}) {
 			run_test_path(&p, skip, runner, start_stop_hook)
 		}
+	} else if extension == Some("swp") || extension == None {
+		// Ignore junk
```
dvdplm: :(

ethcore/src/json_tests/transaction.rs
```diff
@@ -34,52 +35,54 @@ pub fn run_test_file<H: FnMut(&str, HookType)>(p: &Path, h: &mut H) {
...
+			};
+
+			let rlp: Vec<u8> = test.rlp.clone().into();
+			let res = Rlp::new(&rlp)
+				.as_val()
+				.map_err(::error::Error::from)
+				.and_then(|t: UnverifiedTransaction| {
+					let mut header: Header = Default::default();
+					// Use high enough number to activate all required features.
+					header.set_number(0x6ffffffffffffe);
```
dvdplm: Perhaps make it this number a named `const` (for readability)? e.g. `FEATURE_ENABLING_BLOCKNR` or some such?

dvdplm:
> should be fine to just do git submodule update

Wouldn't that update `wasm-tests` as well? That might be undesireable?

ethcore/res/ethereum/eip161_test.json
tomusdrw: Yes, otherwise some transaction tests for `EIP158` spec (which for us is actaully eip161_test.json) fail, cause they assume that eip155 is enabled as well.

ethcore/src/json_tests/test_common.rs
tomusdrw: Yeah, there are some `swp` files or `.stub` files committed in the repo :(

ethcore/src/json_tests/transaction.rs
tomusdrw: Makes sense.

sorpaas:
Please merge master. I think we may face the same test fail as [py-evm](https://github.com/ethereum/py-evm/pull/1224#issuecomment-418775512) after #9319, where it fixed the bug in `discard_checkpoint` if a new contract is created but the account already exists.

PR #9444, opened 2018-08-31
Pyethereum keystore support

tworec:
This will enable to read keystores created via https://github.com/ethereum/pyethereum which we are using in https://github.com/golemfactory/golem.

This PR will allow to read only one such a keystore because there is hardcoded 0-address imputed, and address is a key in internal map of accounts. 

see https://github.com/ethereum/pyethereum/issues/739

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,7 +240,7 @@ dependencies = [
  "heapsize 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "keccak-hash 0.1.2 (git+https://github.com/paritytech/parity-common)",
  "parity-bytes 0.1.0 (git+https://github.com/paritytech/parity-common)",
- "rlp 0.2.1 (git+https://github.com/paritytech/parity-common)",
+ "rlp 0.2.2 (git+https://github.com/paritytech/parity-common)",
  "rlp_derive 0.1.0",
  "rustc-hex 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -532,7 +532,7 @@ dependencies = [
  "evm 0.1.0",
  "fake-hardware-wallet 0.0.1",
  "hardware-wallet 1.12.0",
- "hashdb 0.2.0 (git+https://github.com/paritytech/parity-common)",
+ "hashdb 0.2.1 (git+https://github.com/paritytech/parity-common)",
  "heapsize 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.5.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "journaldb 0.2.0",
@@ -557,7 +557,7 @@ dependencies = [
  "patricia-trie-ethereum 0.1.0",
  "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rayon 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "rlp 0.2.1 (git+https://github.com/paritytech/parity-common)",
+ "rlp 0.2.2 (git+https://github.com/paritytech/parity-common)",
  "rlp_compress 0.1.0",
  "rlp_derive 0.1.0",
  "rustc-hex 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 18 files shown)


ethstore/src/account/crypto.rs
```diff
@@ -79,12 +79,14 @@ impl Crypto {
 
 	/// Encrypt custom plain data
 	pub fn with_plain(plain: &[u8], password: &Password, iterations: u32) -> Result<Self, crypto::Error> {
-		let salt: [u8; 32] = Random::random();
+		let salt_arr: [u8; 32] = Random::random();
```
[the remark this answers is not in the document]
niklasad1: Yeah, I wonder the same since [crypto::derive_key_iterations](https://github.com/paritytech/parity-common/blob/master/parity-crypto/src/lib.rs#L58) requires a slice and it should be fine to allocate an array on the stack and pass a reference to it!

ethstore/cli/src/main.rs
```diff
@@ -145,7 +145,6 @@ impl fmt::Display for Error {
 }
 
 fn main() {
-	panic_hook::set_abort();
```
niklasad1: Why is this removed?

debris: same question

Cargo.lock
```diff
@@ -1965,11 +1964,11 @@ dependencies = [
 [[package]]
 name = "parity-crypto"
 version = "0.1.0"
-source = "git+https://github.com/paritytech/parity-common#0045887fecd2fec39e56c962a0b27e2caf3b9474"
+source = "git+https://github.com/paritytech/parity-common#51258799ee639783a7153d807ce9ddf4d49b29fa"
 dependencies = [
  "ethereum-types 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "quick-error 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "ring 0.12.1 (git+https://github.com/paritytech/ring)",
+ "ring 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
```
debris: what's the reason to change `ring` dependency? iirc the official repo does not contain several changes that we need

Cargo.lock
```diff
@@ -3684,6 +3679,11 @@ dependencies = [
  "xml-rs 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[patch.unused]]
+name = "ring"
+version = "0.12.1"
+source = "git+https://github.com/paritytech/ring#bae475e9f7ea7dd4ae671bef4b576089a9b06731"
```
debris: what is `patch.unused` property and how was it created? I've never seen this entry in `Cargo.lock` files

ethstore/src/json/key_file.rs
```diff
@@ -160,7 +160,7 @@ impl<'a> Visitor<'a> for KeyFileVisitor {
 
 		let address = match address {
 			Some(address) => address,
-			None => return Err(V::Error::missing_field("address")),
+			None => H160::zero(),
```
debris: this is unacceptable and may lead to errors

5chdn:
Please rebase on master.

ethstore/src/account/crypto.rs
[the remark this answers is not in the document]
tworec: bc `Pbkdf2` has `salt` as `Vec` now. But ok, I can pass `[u8]`  into [`crypto::derive_key_iterations`](https://github.com/paritytech/parity-common/blob/master/parity-crypto/src/lib.rs#L58)

ethstore/cli/src/main.rs
tworec: don't remember exactly, but i has some local issues with it. Restoring

Cargo.lock
tworec: `ring` is problematic since it does not allow to link multiple versions of itself into program, and at Golem we are using `ring 0.13`

 https://github.com/briansmith/ring/blob/7e3682c02170f5eafa423b63bacfa018de2878a9/Cargo.toml#L14-L15

tworec: my final workaround is to extremely strip parity-ethereum project under [my private branch](https://github.com/tworec/parity-ethereum/tree/strip-down-ethkeys) to just contain `ethkey` and `ethstore` which we are using as for now.

Is there any other solution to have linked ring 12 and 13 in single program?

tworec: So for this PR I'm reverting this change

ethstore/src/json/key_file.rs
tworec: ok, so I propose import with optional password to recover address when it is not in keystore

PR #9448, opened 2018-08-31
light: give free credits for reserved peers

sorpaas:
When connecting to a reserved peers in light client, assign free credits to them. This fixes the scenario where a home server is running a full node, and another light client connects to the full node as a reserved peer.


ethcore/light/src/net/mod.rs
```diff
@@ -706,8 +708,13 @@ impl LightProtocol {
 		};
 
 		let capabilities = self.capabilities.read().clone();
-		let local_flow = self.flow_params.read();
-		let status_packet = status::write_handshake(&status, &capabilities, Some(&**local_flow));
+		let cost_local_flow = self.flow_params.read();
```
dvdplm: A test for this would be nice.

ethcore/light/src/net/mod.rs
```diff
@@ -818,7 +825,11 @@ impl LightProtocol {
 		}
 
 		let remote_flow = flow_params.map(|params| (params.create_credits(), params));
-		let local_flow = self.flow_params.read().clone();
+		let local_flow = if io.is_reserved_peer(*peer) {
+			self.free_flow_params.clone()
```
andresilva: Why not always return `FlowParams::free()` instead of adding the `free_flow_params` field?

ordian:
Have you tested this? I was unable to connect light client on kovan to a reserved local full node using master, though it works on v2.0.1.

5chdn:
I assume you tested this, @sorpaas ?

Also, conflict in net.

sorpaas:
Not yet. Let me add some tests before merge.

ethcore/light/src/net/mod.rs
sorpaas: In https://github.com/paritytech/parity-ethereum/pull/9448/files#diff-114b98222286628ef4252b75052c348bR712 we only need a reference to `FlowParams::free()`, so while for L829 it's the same, on that one it saves one `FlowParams` stack allocation.

I don't think that's a big deal though so I'm happy to change it to always use `FlowParams::free()` directly.

PR #9457, opened 2018-09-03
Fix a bug in evmbin initial_gas display

sorpaas:
`initial_gas` should be transaction gas limit, not block gas limit.


sorpaas:
I think there's no issue with this change, so let me merge this. Please correct me if I'm wrong!

Issue #9427, opened 2018-08-28
StateDB overhaul


sorpaas:
Our current StateDB implementation is not fully optimized:

* For the first time writing a value not in a checkpoint, or when reverting a checkpoint, all values in the account entry needs to be swapped.
* Sometimes we call `clone_dirty` on an account to clone all dirty values, but we actually only need a few.
* When fetching a checkpoint storage value, in the case where the account entry is not in the checkpoint, we would need to cascade up towards the checkpoint chain. This is the reverse of the issue we face in https://github.com/paritytech/substrate/issues/480

To fix this, here're some ideas we can do:

* Remove indirection of storage values in account entry. Keep one change tree for storage entries, and keep another for other account values.
* Keep original account value caches (the value of the account since last state commit) as normal hash maps.
* Keep account value changes as complete `im::HashMap` (meaning, if anything is not in the hash trie, then we're sure the value has not changed since last commit). This makes creating/reverting checkpoints really cheap.

sorpaas:
I'm closing this. Check https://github.com/paritytech/parity-ethereum/pull/9465#issuecomment-419833578 for rationale.

PR #9465, opened 2018-09-04
 Use im::HashMap for storage_changes 

sorpaas:
rel #9427

This uses `im::HashMap` for `Account::storage_changes`. In the future we should refactor to use PDS for other account changes, but right now I'm still figuring out some entanglement related to `state` and `state_db` mod.

The previous bottleneck was in `Account::clone_dirty`, where we do a full clone of all changes at each checkpoint. This PR relieves that issue. The complexity comparison is as follows:

* Get: `O(1)` -> `O(logn)`
* Insert: `O(1)` -> `O(logn)`
* Clone: `O(n)` -> `O(1)`

Note that `O(logn)` is actually a small number -- if a storage uses up all 256-bit variations and we consider the complexity to be `O(log(2)n)`, then it will only take at most 256 hops.

I'm still thinking about how to do a benchmark on this change, but currently haven't thought of a good way.


ethcore/private-tx/src/lib.rs
```diff
@@ -68,10 +69,11 @@ pub use messages::{PrivateTransaction, SignedPrivateTransaction};
 pub use error::{Error, ErrorKind};
 
 use std::sync::{Arc, Weak};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::time::Duration;
 use ethereum_types::{H128, H256, U256, Address};
 use hash::keccak;
+use im::{HashMap as IMHashMap};
```
dvdplm: maybe just `use im::HashMap as IMHashMap;`

ethcore/src/state/account.rs
```diff
@@ -30,6 +30,7 @@ use trie::{Trie, Recorder};
 use ethtrie::{TrieFactory, TrieDB, SecTrieDB, Result as TrieResult};
 use pod_account::*;
 use rlp::{RlpStream, encode};
+use im::{HashMap as IMHashMap};
```
dvdplm: same here

dvdplm:
Do you plan to add benchmarks in this PR or in a separate one? The potential performance impact is my only concern at this point.

ethcore/src/state/account.rs
```diff
@@ -382,7 +383,12 @@ impl Account {
 	/// Commit the `storage_changes` to the backing DB and update `storage_root`.
 	pub fn commit_storage(&mut self, trie_factory: &TrieFactory, db: &mut HashDB<KeccakHasher>) -> TrieResult<()> {
 		let mut t = trie_factory.from_existing(db, &mut self.storage_root)?;
-		for (k, v) in self.storage_changes.drain() {
+		let old_storage_changes = {
+			let mut tmp = im::HashMap::new();
+			::std::mem::swap(&mut tmp, &mut self.storage_changes);
```
fckt: This is not quite related to this specific PR, but I just wonder, what will happen if it'll fail to commit? Should we concern about this?

sorpaas: Then that error will be propagate out and cause the block import to fail. I think we handled that correctly in `Client`.

sorpaas:
I did a benchmark on this, it's actually slower:

```
	#[test]
	fn storage_bench() {
		use std::time::{Duration, Instant};
		let now = Instant::now();

		let a = 10.into();

		let db = get_temp_state_db();
		let mut state = State::new(db, U256::from(0), Default::default());

		for i in 0..1000000 {
			state.set_storage(&a, H256::from(&U256::from(i)), H256::from(&U256::from(20u64))).unwrap();
		}

		state.checkpoint();
		state.set_storage(&a, H256::from(&U256::from(0)), H256::from(&U256::from(20u64))).unwrap();

		for i in 0..1000000 {
			state.storage_at(&a, &H256::from(&U256::from(i))).unwrap();
		}

		println!("Elapsed: {:?}", now.elapsed());
	}
```

I think the reason is due to `Arc`, whose locks give performance penalty. I'm re-thinking about the plan on #9427:

* With #9319, we have actually already get performance gain in some cases (in the above test, it's 1.5x speed up) because of this (https://github.com/paritytech/parity-ethereum/pull/9319/files#diff-f626e5fe7236214aee0072a7d8ee53d7R289).
* Most storage changes are actually really small, and our hardware is designed to make copying small chunks of memory really efficient.
* Checkpoint iteration (https://github.com/paritytech/parity-ethereum/pull/9319/files#diff-fff79319205229b71cd0b01cdcb4eedbR616) might not be an issue -- it has an upper bound and we're only checking for the first checkpoint where that entry exists.

So for the short-term, I think it would be okay to keep the current state checkpointing logic as is -- I don't think that's part of our current performance bottleneck right now.

dvdplm:
Thank you for trying this out, these things can be subtle! :)

> Most storage changes are actually really small,

Is it true that while most might be really small, there are some that are really big?

sorpaas:
> Is it true that while most might be really small, there are some that are really big?

Yes that's true. Some indeed can be big.

PR #9469, opened 2018-09-04
Support millisecond timestamp for instant seal engine

### LINKED ISSUE #9468: finer time resolution please
If two blocks are closer in time than (half) a second, parity returns identical `web3.eth.getBlock(blockNumber).timestamp` for *both* blocks ... 

.... which can then lead to [ZeroDivisionError like here](https://gitlab.com/electronDLT/chainhammer/blob/515033d42a0fab45cd53ae4e823501385344bf8f/tps.py#L99).

Of course, I can catch that, as you see. But ... but.

(For the context, please have a look at [these chainhammer benchmarks --> parity.md](https://gitlab.com/electronDLT/chainhammer/blob/5fa31f2999627f9a3858f443240fcaa1a2dfd23d/parity.md#run-8-instantseal))

**--> Please increase the precision of the blocktime, thanks.**

(`geth quorum raft consensus` is even [using nanoseconds](https://gitlab.com/electronDLT/chainhammer/blob/515033d42a0fab45cd53ae4e823501385344bf8f/tps.py#L92-93) which is IMHO too extreme.)

---

- **Parity Ethereum version**: v1.11.10
- **Operating system**: Linux
- **Installation**: docker
- **Fully synchronized**: not relevant, instantseal
- **Network**: instantseal dev chain
- **Restarted**: yes

sorpaas:
Fixes #9468 

In instant seal engine, one can set the `millisecondTimestamp` config to enable millisecond timestamp resolution. I didn't add nanosecond support because that overflows `u64`.

### DIFF
--- a/json/src/spec/instant_seal.rs
+++ b/json/src/spec/instant_seal.rs
@@ -21,7 +21,8 @@
 pub struct InstantSealParams {
 	/// Whether to enable millisecond timestamp.
 	#[serde(rename="millisecondTimestamp")]
-	pub millisecond_timestamp: Option<bool>,
+	#[serde(default)]
+	pub millisecond_timestamp: bool,
 }
 
 /// Instant seal engine descriptor.
--- a/json/src/spec/instant_seal.rs
+++ b/json/src/spec/instant_seal.rs
@@ -1,0 +1,32 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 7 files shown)


json/src/spec/instant_seal.rs
```diff
@@ -0,0 +1,32 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Instant seal engine params deserialization.
+
+/// Instant seal engine params deserialization.
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct InstantSealParams {
+	/// Whether to enable millisecond timestamp.
+	#[serde(rename="millisecondTimestamp")]
+	pub millisecond_timestamp: Option<bool>,
```
ordian: Maybe use `#[serde(default)]` and make it a `bool`? `Option<bool>` seems like an overkill.

Tbaut:
Can we add this to the wiki please @sorpaas ?

PR #9473, opened 2018-09-04
Light `clippy(fy)`

niklasad1:
Ran `clippy` and fixed some issues!


ethcore/light/src/cache.rs
```diff
@@ -31,7 +31,7 @@ use ethereum_types::{H256, U256};
 use memory_cache::MemoryLruCache;
 
 /// Configuration for how much data to cache.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
niklasad1: It is 5 usizes (max 40 bytes) and follows 
> Generally speaking, if your type can implement Copy, it should

niklasad1:
@ordian 

Yes, I had some `wasm-tests`  mismatches but I guess this is newer than master. Should I revert it to be on the safe side?

ethcore/light/src/cache.rs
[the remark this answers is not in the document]
niklasad1: I can change it ofc if you have other opinions!

ethcore/sync/src/light_sync/sync_round.rs
```diff
@@ -148,16 +148,11 @@ impl Fetcher {
...
-			let start_hash = match self.sparse.front() {
-				Some(first) => first.hash(),
-				None => break,
-			};
-
-			match self.complete_requests.remove(&start_hash) {
-				None => break,
+		while let Some(first) = self.sparse.pop_front() {
+			match self.complete_requests.remove(&first.hash()) {
+				None => { self.sparse.push_front(first); break; }
```
andresilva: Multi-line maybe?

ethcore/sync/src/light_sync/sync_round.rs
```diff
@@ -148,16 +148,11 @@ impl Fetcher {
...
 	fn collect_ready(&mut self) {
-		loop {
-			let start_hash = match self.sparse.front() {
-				Some(first) => first.hash(),
-				None => break,
-			};
-
-			match self.complete_requests.remove(&start_hash) {
-				None => break,
+		while let Some(first) = self.sparse.pop_front() {
```
andresilva: I don't know this code so I don't know if it makes sense to change this code to speculatively `pop` and then re-adding if `complete_requests.remove` fails. Is the success path the most common?

niklasad1: Good point, Andre!

I don't know either this was more a hack to get around having a reference and mutating at the same time!
I somehow convinced myself by 

> This queue has `O(1)` amortized inserts and removals from both ends of the container

But now I realize that the ```Header struct``` is big (~600 bytes) so I will revert this thanks :raised_hands:

Issue #7411, opened 2017-12-29
Warn when reading date from non-existing ancient blocks.


tomusdrw:
Currently after warp sync not all blocks might be present, although you can see the latest state in the UI.

This might be especiall frustrating for developers, since RPC is just simply not returning data for such blocks.

Example:
Fetching all the logs for particular address can render an empty list even though there are clearly blocks with logs out there.

Proposal:
1. Display a warning if such query is detected (blockNumber < latestBlock && block not found)
2. Return a "ancient block missing" RPC error.

epheph:
This issue is really critical for Augur, which sync's the blockchain via a series of getLogs, making the assumption that blocks prior to getBlock("latest").number are available. To illustrate the issue outlined above, here is an inspection of CryptoKitties Birth events from a recently warp-sync'd parity node (warp-barrier: `0x5f8521`):

```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[ {"fromBlock":"0x500000","toBlock":"0x5f853e","address":"0x06012c8cf97bead5deae237070f9587f8e7a266d","topics":["0x0a5311bd2a6608f08a180df2ee7c5946819a649b204b554bb8e39825b2c50ad5"]} ],"id": 74}' localhost:8545 | jq .result[].blockNumber | uniq -c
      1 "0x5f8523"
      2 "0x5f8527"
      2 "0x5f8528"
     13 "0x5f852b"
      2 "0x5f852d"
      3 "0x5f852e"
      7 "0x5f852f"
      2 "0x5f8530"
      3 "0x5f8532"
      2 "0x5f8533"
     18 "0x5f8535"
      1 "0x5f8536"
      2 "0x5f8537"
      6 "0x5f8539"
     14 "0x5f853a"
     16 "0x5f853c"
      1 "0x5f853d"
      1 "0x5f853e"
```

There were a great number of events between `0x500000` and `0x5f8523`, but they were simply omitted from this getLogs call, with no indication that anything went wrong inside Parity. Running this query again later, after ancient blocks downloaded, would return different results.

We will add a section to our README advising users not to use Parity with warp-sync (or to use warp-sync only with a specific `-warp-barrier` prior to any blocks that contain relevant logs), but changing this behavior in Parity seems like the right thing to do. Ideally, an error would be returned if Parity could not satisfy a request, instead of a silent or partial failure.

udoprog:
[...]

PR #9475, opened 2018-09-04
Add Error message when sync is still in progress.

### LINKED ISSUE #9188: Add error message for eth_getTransactionReceipt when ancient blocks haven't been downloaded yet (warp-sync)
> I'm running:
>

A warp-synced full node returns empty responses to `eth_getTransactionReceipt` RPC requests for transactions before the snapshot point.

```bash
# Request transaction from recent block
$ curl -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0xf8e026115bafda9efaee095d1f3342333826f53e332d4ffd7c9fb52bc499f9d5"],"id":0}' localhost:8545
{"jsonrpc":"2.0","result":{"blockHash":"0x7d24546c087fe34a595a250ba3aa3b4fa8c6cbfa20ec3963e122b1234170742b",...

# Request transaction from old block
$ curl -H 'Content-Type: application/json' --data '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0x511b14d534466fdc5163d2cd7e37910a570c8e012e9970c7206272b4f8b122c5"],"id":0}' localhost:8545
{"jsonrpc":"2.0","result":null,"id":0}
```

I assume Parity does not back fill transaction receipts from ancient blocks in warp mode, but the current behavior seems broken. The RPC should either return an error if the receipt is missing or fetch it on demand.

seunlanlege:
fix #9188

### DIFF
--- a/ethcore/src/client/config.rs
+++ b/ethcore/src/client/config.rs
@@ -125,9 +125,6 @@ pub struct ClientConfig {
 	pub max_round_blocks_to_import: usize,
 	/// Snapshot configuration
 	pub snapshot: SnapshotConfiguration,
-	/// used by the RPC, returns an error if block information was requested
-	/// and there are gaps in the db
-	pub allow_empty_block_result: bool
 }
 
 impl Default for ClientConfig {
@@ -154,7 +151,6 @@ impl Default for ClientConfig {
 			transaction_verification_queue_size: 8192,
 			max_round_blocks_to_import: 12,
 			snapshot: Default::default(),
-			allow_empty_block_result: false
 		}
 	}
 }
--- a/ethcore/src/client/config.rs
+++ b/ethcore/src/client/config.rs
@@ -125,6 +125,9 @@ pub struct ClientConfig {
 	pub max_round_blocks_to_import: usize,
 	/// Snapshot configuration
 	pub snapshot: SnapshotConfiguration,
+	/// used by the RPC, returns an error if block information was requested
+	/// and there are gaps in the db
+	pub allow_empty_block_result: bool
 }
 
 impl Default for ClientConfig {
@@ -151,6 +154,7 @@ impl Default for ClientConfig {
 			transaction_verification_queue_size: 8192,
 			max_round_blocks_to_import: 12,
 			snapshot: Default::default(),
+			allow_empty_block_result: false
 		}
 	}
 }
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -471,6 +471,10 @@ usage! {
 			"--jsonrpc-no-keep-alive",
 			"Disable HTTP/1.1 keep alive header. Disabling keep alive will prevent re-using the same TCP connection to fire multiple requests, recommended when using one request per connection.",
 
+			FLAG flag_jsonrpc_experimental: (bool) = false, or |c: &Config| c.rpc.as_ref()?.experimental_rpcs.clone(),
+			"--jsonrpc-experimental",
+			"Enable experimental RPCs. Enable to have access to methods from unfinalised EIPs in all namespaces",
+
 			ARG arg_jsonrpc_port: (u16) = 8545u16, or |c: &Config| c.rpc.as_ref()?.port.clone(),
 			"--jsonrpc-port=[PORT]",
 			"Specify the port portion of the HTTP JSON-RPC API server.",
@@ -1224,6 +1228,7 @@ struct Rpc {
 	processing_threads: Option<usize>,
 	max_payload: Option<usize>,
 	keep_alive: Option<bool>,
+	experimental_rpcs: Option<bool>,
 	poll_lifetime: Option<u32>,
 }
 
@@ -1682,6 +1687,7 @@ mod tests {
 			// RPC
 			flag_no_jsonrpc: false,
 			flag_jsonrpc_no_keep_alive: false,
+			flag_jsonrpc_experimental: false,
 			arg_jsonrpc_port: 8545u16,
 			arg_jsonrpc_interface: "local".into(),
 			arg_jsonrpc_cors: "null".into(),
@@ -1965,6 +1971,7 @@ mod tests {
 				processing_threads: None,
 				max_payload: None,
 				keep_alive: None,
+				experimental_rpcs: None,
 				poll_lifetime: None,
 			}),
 			ipc: Some(Ipc {
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -503,6 +503,10 @@ usage! {
 			"--jsonrpc-max-payload=[MB]",
 			"Specify maximum size for HTTP JSON-RPC requests in megabytes.",
 
+			ARG arg_poll_lifetime: (u32) = 60u32, or |c: &Config| c.rpc.as_ref()?.poll_lifetime.clone(),
+			"--poll-lifetime=[S]",
+			"Set the RPC filter lifetime to S seconds. The filter has to be polled at least every S seconds , otherwise it is removed.",
+
 		["API and Console Options – WebSockets"]
 			FLAG flag_no_ws: (bool) = false, or |c: &Config| c.websockets.as_ref()?.disable.clone(),
 			"--no-ws",
@@ -756,10 +760,6 @@ usage! {
 			ARG arg_gas_price_percentile: (usize) = 50usize, or |c: &Config| c.mining.as_ref()?.gas_price_percentile,
 			"--gas-price-percentile=[PCT]",
 			"Set PCT percentile gas price value from last 100 blocks as default gas price when sending transactions.",
-
-			ARG arg_poll_lifetime: (u32) = 60u32, or |c: &Config| c.mining.as_ref()?.poll_lifetime.clone(),
-			"--poll-lifetime=[S]",
-			"Set the lifetime of the internal index filter to S seconds.",
 
 			ARG arg_author: (Option<String>) = None, or |c: &Config| c.mining.as_ref()?.author.clone(),
 			"--author=[ADDRESS]",
@@ -1224,6 +1224,7 @@ struct Rpc {
 	processing_threads: Option<usize>,
 	max_payload: Option<usize>,
 	keep_alive: Option<bool>,
+	poll_lifetime: Option<u32>,
 }
 
 #[derive(Default, Debug, PartialEq, Deserialize)]
@@ -1316,7 +1317,6 @@ struct Mining {
 	relay_set: Option<String>,
 	min_gas_price: Option<u64>,
 	gas_price_percentile: Option<usize>,
-	poll_lifetime: Option<u32>,
 	usd_per_tx: Option<String>,
 	usd_per_eth: Option<String>,
 	price_update_period: Option<String>,
@@ -1690,6 +1690,7 @@ mod tests {
 			arg_jsonrpc_server_threads: None,
 			arg_jsonrpc_threads: 4,
 			arg_jsonrpc_max_payload: None,
+			arg_poll_lifetime: 60u32,
 
 			// WS
 			flag_no_ws: false,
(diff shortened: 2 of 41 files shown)


seunlanlege:
[clabot:check]

andresilva:
(reposting from private chat)

I guess for some RPCs we could implement a best-effort approach. For example to fetch transaction by hash:

- We try to fetch it from the current chain, if we find something we return it (same behavior as now)
- If we don't find it and we know that we're still syncing ancient blocks then we return with an error saying that the transaction wasn't found but we still don't have all the data yet, i.e. this could be a false negative (so the user can retry later when sync is finished)
- If we don't find the data but we are fully synced we just return `None` in this case

Can we use this strategy for all RPC methods? Do we want to implement something like this?

tomusdrw:
#7411 might be a bit more straightforward - we have the block number there so we can just return an error in case `block_number < latest_block_number` and the data is missing (would be worth to handle logs as well, as in the comment in that issue).

rpc/src/v1/impls/eth.rs
```diff
@@ -639,8 +639,25 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 		Box::new(future::done(self.rich_block(BlockId::Hash(hash.into()).into(), include_txs)))
 	}
 
+	fn check_block_existence(&self, num: BlockNumber) -> Result<()> {
+		if let BlockNumber::Num(number) = num {
+			match self.client.chain_info().best_block_number {
+				BlockNumber::Num(best_block_number) => {
+					if num < best_block_number {
```
tomusdrw: This is quite wrong, actually this are the only queries that make sense at all, since:
1. In database we store blocks from `[0...best_block_number]`
2. Querying anything from that range should succeed
3. Querying anything greater than `best_block_number` will always fail

With ancient blocks the case is that the database is not complete, so we have a situation where:
1. We have `[0..best_ancient_block] ... [first_block .. best_block]`
2. The latter range has been restored together with the snapshot
3. The former range has been downloaded during ancient block import
4. Any query that tries to get `num < best_block_number` and fails (i.e. returns `None`) most likely falls into `[best_ancient_block .. first_block]` range, so we should detect that and return the error only in such case.

rpc/src/v1/impls/eth.rs
```diff
@@ -639,8 +639,25 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
+				_ => {}
+			}
+		}
+		Ok(())
+	}
+
 	fn block_by_number(&self, num: BlockNumber, include_txs: bool) -> BoxFuture<Option<RichBlock>> {
-		Box::new(future::done(self.rich_block(num.into(), include_txs)))
+		let result = self.check_block_existence(num)
+			.and_then(|_| self.rich_block(num.into(), include_txs));
```
tomusdrw: To make it correct you should:
1. First query the block
2. If it's `None` and `num < best_block_number` return ancient block unavailable error.

rpc/src/v1/impls/eth.rs
```diff
@@ -667,7 +687,9 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 		};
 
 		let transaction_id = PendingTransactionId::Location(block_id, index.value());
-		Box::new(future::done(self.transaction(transaction_id)))
+		let result = self.transaction(transaction_id)
```
tomusdrw: Here we actually have `BlockNumber` so we don't have to fallback to this heuristic for hashes (just check if the block number is available and return error if not).

Tbaut:
Despite trying hard, I still can't manage to know all issues by number :) Having a human understandable description and title for a PR is therefore requested!

rpc/src/v1/helpers/errors.rs
```diff
@@ -196,6 +199,53 @@ pub fn no_work_required() -> Error {
...
+	move |response| {
+		if response.is_none() {
+			let BlockChainInfo { ancient_block_hash, first_block_hash, .. } = client.chain_info();
+			// block information was requested, but unfortunately we couldn't find it and there
+			// are gaps in the database ethcore/src/blockchain/blockchain.rs:202
+			if ancient_block_hash.is_some() && first_block_hash.is_some() {
+				return Err(Error {
+					code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST),
+					// this error message feels pretty straightforward to me. ¯\_(ツ)_/¯
+					message: "We cannot tell for sure if the thing you requested is not available or we just don't have it".into(),
```
andresilva: Probably better to be explicit about it? "Ancient block sync is still in progress."

rpc/src/v1/helpers/errors.rs
```diff
@@ -196,6 +199,53 @@ pub fn no_work_required() -> Error {
...
+
+pub fn check_block_gap<'a, T, C>(client: &'a C) -> impl Fn(Option<T>) -> RpcResult<Option<T>> + 'a
+	where C: BlockChainClient,
+{
+	move |response| {
+		if response.is_none() {
+			let BlockChainInfo { ancient_block_hash, first_block_hash, .. } = client.chain_info();
+			// block information was requested, but unfortunately we couldn't find it and there
+			// are gaps in the database ethcore/src/blockchain/blockchain.rs:202
+			if ancient_block_hash.is_some() && first_block_hash.is_some() {
```
andresilva: I think first_block_hash will always be some.

rpc/src/v1/helpers/errors.rs
```diff
@@ -196,6 +199,53 @@ pub fn no_work_required() -> Error {
 	}
 }
 
+pub fn unavailable_block() -> Error {
+	Error {
+		code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST),
+		message: "Block is unavailable".into(),
```
andresilva: Probably better to have a more explicit message about ancient block sync?

andresilva:
Please rebase/merge master to fix conflicts.

rpc/src/v1/helpers/errors.rs
```diff
@@ -197,6 +200,48 @@ pub fn no_work_required() -> Error {
...
+		Ok(response)
+	}
+}
+
+pub fn check_block_gap<'a, T, C>(client: &'a C) -> impl Fn(Option<T>) -> RpcResult<Option<T>> + 'a
+	where C: BlockChainClient,
+{
+	move |response| {
+		if response.is_none() {
+			let BlockChainInfo { ancient_block_hash, first_block_hash, .. } = client.chain_info();
```
tomusdrw: `first_block_hash` is unused

rpc/src/v1/helpers/errors.rs
```diff
@@ -197,6 +200,48 @@ pub fn no_work_required() -> Error {
...
+			// block information was requested, but unfortunately we couldn't find it and there
+			// are gaps in the database ethcore/src/blockchain/blockchain.rs:202
+			if ancient_block_hash.is_some()  {
+				return Err(unavailable_block())
```
tomusdrw: This is a bit different than `check_block_number_existence`. Here we are unsure if the block is not valid or was not found, in the former we actually have certainty that the block is missing because we are still syncing ancient blocks.

TBH I think we should express that in the error message as well.

seunlanlege:
@andresilva can you take a look at this again, already added the cli flag that disables this behaviour

ethcore/src/client/config.rs
```diff
@@ -125,6 +125,9 @@ pub struct ClientConfig {
 	pub max_round_blocks_to_import: usize,
 	/// Snapshot configuration
 	pub snapshot: SnapshotConfiguration,
+	/// used by the RPC, returns an error if block information was requested
+	/// and there are gaps in the db
+	pub allow_empty_block_result: bool
```
andresilva: I think this should be added to `EthClientOptions`, this way we don't need to expose `ClientConfig`.

parity/cli/mod.rs
```diff
@@ -238,6 +238,11 @@ usage! {
 	{
 		// Global flags and arguments
 		["Operating Options"]
+			FLAG flag_allow_empty_block_result: (bool) = false, or |c: &Config| c.parity.as_ref()?.allow_empty_block_result
+			.clone(),
+			"--no-empty-block-result",
```
andresilva: I think this flag should be renamed to `allow-empty-block-result` since enabling it allows empty block results (instead of disallowing).

seunlanlege:
Please do

andresilva:
This change should be documented in the release notes (`--allow-empty-block-result` restores the previous behavior).

parity/cli/mod.rs
```diff
@@ -238,6 +238,11 @@ usage! {
 	{
 		// Global flags and arguments
 		["Operating Options"]
+			FLAG flag_allow_empty_block_result: (bool) = false, or |c: &Config| c.parity.as_ref()?.allow_empty_block_result
```
tomusdrw: please move it to RPC section and call it `--jsonrpc-allow-missing-blocks`, maybe?

sorpaas:
Also there's a test failure:

```
error[E0063]: missing field `jsonrpc_allow_missing_blocks` in initializer of `v1::impls::eth::EthClientOptions`
   --> rpc/src/v1/tests/eth.rs:143:4
    |
143 |             EthClientOptions {
    |             ^^^^^^^^^^^^^^^^ missing `jsonrpc_allow_missing_blocks`
```

parity/cli/mod.rs
```diff
@@ -463,6 +463,10 @@ usage! {
 
 
 		["API and Console Options – HTTP JSON-RPC"]
+			FLAG flag_allow_missing_blocks: (bool) = false, or |c: &Config| c.rpc.as_ref()?.allow_missing_blocks.clone(),
```
sorpaas:
```suggestion
			FLAG flag_jsonrpc_allow_missing_blocks: (bool) = false, or |c: &Config| c.rpc.as_ref()?.allow_missing_blocks.clone(),
```

I think we should keep args to be `--jsonrpc-allow-missing-block`, but set configs to be only `allow_missing_blocks`, because the later can be identified inside the `rpc` category already.

parity/cli/mod.rs
```diff
@@ -463,6 +463,10 @@ usage! {
...
+			"--allow-missing-blocks",
```
sorpaas:
```suggestion
			"--jsonrpc-allow-missing-blocks",
```

rpc/src/v1/impls/eth.rs
```diff
@@ -718,28 +750,32 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
 
 	fn transaction_receipt(&self, hash: RpcH256) -> BoxFuture<Option<Receipt>> {
+		let best_block = self.client.chain_info().best_block_number;
 		let hash: H256 = hash.into();
 
-		if self.options.allow_pending_receipt_query {
-			let best_block = self.client.chain_info().best_block_number;
-			if let Some(receipt) = self.miner.pending_receipt(best_block, &hash) {
-				return Box::new(future::ok(Some(receipt.into())));
+		match (self.miner.pending_receipt(best_block, &hash), self.options.allow_pending_receipt_query) {
```
sorpaas: This will unnecessarily call `self.miner.pending_receipt` even when `self.options.allow_pending_receipt_query` is false. I would suggest to change it back to the old two `if` form.

rpc/src/v1/impls/eth.rs
```diff
@@ -718,28 +750,33 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
 
 		let transaction_id = PendingTransactionId::Location(block_id, index.value());
-		Box::new(future::done(self.transaction(transaction_id)))
+		let result = self.transaction(transaction_id).and_then(
+			errors::check_block_number_existence(&*self.client, num, self.options.allow_missing_blocks));
+		Box::new(future::done(result))
 	}
 
 	fn transaction_receipt(&self, hash: RpcH256) -> BoxFuture<Option<Receipt>> {
+		let best_block = self.client.chain_info().best_block_number;
```
sorpaas: This is only used inside `if self.options.allow_pending_receipt_query {`, so maybe it's better to move it back.

PR #9479, opened 2018-09-05
upload will fail if a file with the same hash is already up

gabreal:
still about this issue: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/96821

which happens if a snap with the same checksum is already uploaded. allow_failure here will at least the pipeline proceed to the next stage.

(the before_script statements with *determine_version are duplicate)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,9 +41,17 @@ cache:
     paths:
     - artifacts/
 
-.determine_version: &determine_version |
-  export VERSION=$(grep -m 1 "version =" Cargo.toml | awk '{print $3}' | tr -d '"' | tr -d "\n")
-  echo "Version" $VERSION
+.determine_version:
+  before_script:                   &determine_version
+    - >
+      VERSION="$(sed -r -n '1,/^version/s/^version = "([^"]+)".*$/\1/p' < Cargo.toml)";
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -222,33 +222,28 @@ publish-linux-snap-amd64:          &publish_snap
     BUILD_ARCH:                    amd64
   dependencies:
     - package-linux-snap-amd64
+  allow_failure:                   true
```
5chdn: How is this possible, we are building nightly only once per day. This should ideally not fail. Can't we fix this on snap side?

gabreal: The upload only fails for uploads of exactly the same snap for the same version/architecture. So if there are no changes in the code nightly builds will still produce a snap which will be upload and hash check on snapcraft refuses it then. I found no way of changing this behaviour.

IMHO the easiest way to fix this would be to have a version like the current timestamp (e.g. 20180905) or git commit like before: https://imgur.com/a/bgpoG7T

5chdn: The version is wrong anyways, this is not 2.1.0. This need to be fixed in the script that figures out the version. Adding a commit hash and date would help here if `nightly`.

PR #9210, opened 2018-07-24
Check if synced when using eth_getWork (#9193)

mttmartin:
This should fix  #9193.

PR #9484, opened 2018-09-06
Only check warp syncing for eth_getWorks

sorpaas:


5chdn:
That only affects master 2.1.0+, does it?

sorpaas:
It affects anything after #9210. So I think that means beta and stable.

fckt:
I'd like to add my review, but I need to understand some context first.. Why we don't need to check for `is_major_importing` but check only for `is_warp_syncing` here?

sorpaas:
@fckt the issue is that using `is_major_importing` may break this RPC call for some miners. A peer may cause the client to occasionally switch to `SyncState::Blocks` (considered syncing by `is_major_importing`), while the client is really close to chain head.

sorpaas:
I'm going to add the blocker tag for this PR, because if this is indeed an issue, then pow for some chains may have been broken.

Issue #9459, opened 2018-09-03
evmbin: replace hardcoded json output logic with serde serialization


sorpaas:
rel https://github.com/paritytech/parity-ethereum/pull/9458#pullrequestreview-151808745

ordian:
closed by #9489

PR #9489, opened 2018-09-07
Replace hardcoded JSON with serde json! macro

HCastano:
Looks to address #9459.

I'm having a problem getting the tests to pass though, and could use some suggestions. The json! macro creates JSON with all the fields in alphabetical order. This is causing three of the evmbin tests to fail since the expected tests strings are not in the same order. 

This is from one of the failed tests (`should_trace_failure`). You can see all the fields are correctly serialized, just in the "wrong" order.

```
"{\"pc\":0,\"op\":96,\"opName\":\"PUSH1\",\"gas\":\"0xffff\",\"stack\":[],\"storage\":{},\"depth\":1}\n
{\"pc\":2,\"op\":214,\"opName\":\"\",\"gas\":\"0xfffc\",\"stack\":[\"0xf8\"],\"storage\":{},\"depth\":1}\n"

"{\"depth\":1,\"gas\":\"0xffff\",\"op\":96,\"opName\":\"PUSH1\",\"pc\":0,\"stack\":[],\"storage\":{}\n
{\"depth\":1,\"gas\":\"0xfffc\",\"op\":214,\"opName\":\"\",\"pc\":2,\"stack\":[\"0xf8\"],\"storage\":{}}\n"
```

I can change the fields of the raw test strings to be in alphabetical order and get the tests to pass, but that seems like a hack since the order of JSON fields shouldn't matter.

### DIFF
--- a/evmbin/src/display/json.rs
+++ b/evmbin/src/display/json.rs
@@ -45,7 +45,7 @@ pub struct Informant {
 
 impl vm::Informant for Informant {
 	fn before_test(&mut self, name: &str, action: &str) {
-		println!("{}", json!({"action": action, "test": name}).to_string());
+		println!("{}", json!({"action": action, "test": name}));
 	}
 
 	fn set_gas(&mut self, gas: U256) {
@@ -65,7 +65,7 @@ impl vm::Informant for Informant {
 					"time": display::as_micros(&success.time),
 				});
 
-				println!("{}", success_msg.to_string())
+				println!("{}", success_msg)
 			},
 			Err(failure) => {
 				for trace in failure.traces.unwrap_or_else(Vec::new) {
@@ -78,7 +78,7 @@ impl vm::Informant for Informant {
 					"time": display::as_micros(&failure.time),
 				});
 
-				println!("{}", failure_msg.to_string())
+				println!("{}", failure_msg)
 			},
 		}
 	}
--- a/evmbin/src/display/json.rs
+++ b/evmbin/src/display/json.rs
@@ -61,7 +61,7 @@ impl vm::Informant for Informant {
 
 				let success_msg = json!({
 					"output": format!("0x{}", success.output.to_hex()),
-					"gasUsed": format!("0x{:x}", success.gas_used),
+					"gasUsed": format!("{:#x}", success.gas_used),
 					"time": display::as_micros(&success.time),
 				});
 
@@ -74,7 +74,7 @@ impl vm::Informant for Informant {
 
 				let failure_msg = json!({
 					"error": display::escape_newlines(&failure.error),
-					"gasUsed": format!("0x{:x}", failure.gas_used),
+					"gasUsed": format!("{:#x}", failure.gas_used),
 					"time": display::as_micros(&failure.time),
 				});
 
@@ -107,8 +107,8 @@ impl trace::VMTracer for Informant {
 			"pc": self.pc,
 			"op": self.instruction,
 			"opName": info.map(|i| i.name).unwrap_or(""),
-			"gas": format!("0x{:x}", gas_used.saturating_add(self.gas_cost)),
-			"gasCost": format!("0x{:x}", self.gas_cost),
+			"gas": format!("{:#x}", gas_used.saturating_add(self.gas_cost)),
+			"gasCost": format!("{:#x}", self.gas_cost),
 			"memory": format!("0x{}", self.memory.to_hex()),
 			"stack": self.stack,
 			"storage": self.storage,
--- a/evmbin/src/display/json.rs
+++ b/evmbin/src/display/json.rs
@@ -280,7 +280,20 @@ mod tests {
 {"pc":5,"op":88,"opName":"PC","gas":"0x2102","gasCost":"0x2","memory":"0x","stack":["0x0","0x0","0x0","0x0","0x0"],"storage":{},"depth":2}
 {"pc":6,"op":48,"opName":"ADDRESS","gas":"0x2100","gasCost":"0x2","memory":"0x","stack":["0x0","0x0","0x0","0x0","0x0","0x5"],"storage":{},"depth":2}
 {"pc":7,"op":241,"opName":"CALL","gas":"0x20fe","gasCost":"0x0","memory":"0x","stack":["0x0","0x0","0x0","0x0","0x0","0x5","0xbd770416a3345f91e4b34576cb804a576fa48eb1"],"storage":{},"depth":2}
-			"#,
+"#,
+		);
+
+		run_test(
+			Informant::default(),
+			&compare_json,
+			"3260D85554",
+			0xffff,
+			r#"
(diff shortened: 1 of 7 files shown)


evmbin/src/display/json.rs
```diff
@@ -81,24 +59,26 @@ impl vm::Informant for Informant {
...
 
-				println!(
-					"{{\"error\":\"{error}\",\"gasUsed\":\"{gas:x}\",\"time\":{time}}}",
-					error = display::escape_newlines(&failure.error),
-					gas = failure.gas_used,
-					time = display::as_micros(&failure.time),
-				)
+				let failure_msg = json!({
+					"error": display::escape_newlines(&failure.error),
+					"gasUsed": format!("0x{:x}", failure.gas_used),
```
niklasad1: Use ```format!("{:#x}", x)``` instead ```format!("{0x{:x}", x)```

evmbin/src/display/json.rs
```diff
@@ -81,24 +59,26 @@ impl vm::Informant for Informant {
...
 
-				println!(
-					"{{\"output\":\"0x{output}\",\"gasUsed\":\"{gas:x}\",\"time\":{time}}}",
-					output = success.output.to_hex(),
-					gas = success.gas_used,
-					time = display::as_micros(&success.time),
-				)
+				let success_msg = json!({
+					"output": format!("0x{}", success.output.to_hex()),
+					"gasUsed": format!("0x{:x}", success.gas_used),
```
niklasad1: Use format!("{:#x}", x) instead format!("{0x{:x}", x)

evmbin/src/display/json.rs
```diff
@@ -123,19 +103,19 @@ impl trace::VMTracer for Informant {
...
-			stack = self.stack(),
-			storage = self.storage(),
-			depth = self.depth,
-		);
-		self.traces.push(trace);
+		let trace = json!({
+			"pc": self.pc,
+			"op": self.instruction,
+			"opName": info.map(|i| i.name).unwrap_or(""),
+			"gas": format!("0x{:x}", gas_used.saturating_add(self.gas_cost)),
```
niklasad1: Use ```format!("{:#x}", x)``` instead ```format!("{0x{:x}", x)```

evmbin/src/display/std_json.rs
```diff
@@ -113,26 +100,28 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
-					gas = success.gas_used,
-					time = display::as_micros(&success.time),
-				).expect("The sink must be writeable.");
+				let trace_data = json!({"stateRoot": success.state_root});
+				writeln!(&mut trace_sink, "{}", trace_data.to_string())
+					.expect("The sink must be writeable.");
+
+				let out_data = json!({
+					"output": format!("0x{}", success.output.to_hex()),
+					"gasUsed": format!("0x{:x}", success.gas_used),
```
niklasad1: Use ```format!("{:#x}", x)``` instead ```format!("{0x{:x}", x)```

evmbin/src/display/std_json.rs
```diff
@@ -113,26 +100,28 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
-				writeln!(
-					&mut out_sink,
-					"{{\"error\":\"{error}\",\"gasUsed\":\"0x{gas:x}\",\"time\":{time}}}",
-					error = display::escape_newlines(&failure.error),
-					gas = failure.gas_used,
-					time = display::as_micros(&failure.time),
-				).expect("The sink must be writeable.");
+				let out_data = json!({
+					"error": display::escape_newlines(&failure.error),
+					"gasUsed": format!("0x{:x}", failure.gas_used),
```
niklasad1: Use ```format!("{:#x}", x)``` instead ```format!("{0x{:x}", x)```

sorpaas:
@HCastano 

> I can change the fields of the raw test strings to be in alphabetical order and get the tests to pass, but that seems like a hack since the order of JSON fields shouldn't matter.

I think that's okay, as long as you think the output is correct. We have many mocked RPC tests and they're all changed this way.

evmbin/src/display/std_json.rs
```diff
@@ -113,26 +100,28 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
 			Err(failure) => {
-				writeln!(
-					&mut out_sink,
-					"{{\"error\":\"{error}\",\"gasUsed\":\"0x{gas:x}\",\"time\":{time}}}",
-					error = display::escape_newlines(&failure.error),
-					gas = failure.gas_used,
-					time = display::as_micros(&failure.time),
-				).expect("The sink must be writeable.");
+				let out_data = json!({
+					"error": display::escape_newlines(&failure.error),
```
sorpaas: `escape_newlines` is not necessary any more with `serde_json`, please remove that function.

evmbin/src/display/json.rs
```diff
@@ -123,19 +103,19 @@ impl trace::VMTracer for Informant {
...
-		self.traces.push(trace);
+		let trace = json!({
+			"pc": self.pc,
+			"op": self.instruction,
+			"opName": info.map(|i| i.name).unwrap_or(""),
+			"gas": format!("{:#x}", gas_used.saturating_add(self.gas_cost)),
+			"gasCost": format!("{:#x}", self.gas_cost),
+			"memory": format!("0x{}", self.memory.to_hex()),
+			"stack": self.stack,
+			"storage": self.storage,
```
ordian: I think this implementation is different than it used to be (`Display` vs `Debug`). It's a shame we don't have tests with non-empty `storage`.

HCastano: Why don't any of the tests have a non-empty `storage`?

HCastano: @ordian So both `stack` and `storage` don't implement `Display` (since I removed the `stack()` and `storage()` functions). Why is having `Display` important for these variables?

ordian: Sorry for not being clear. What I mean was, that the `fn storage()` was using `Debug` (`0x{:?}`) key/value representation, and I thought that serde impl was using `Display` one, but I was wrong.
Thanks for adding an additional test!

HCastano: Ah, I see. No problem!

evmbin/src/display/json.rs
```diff
@@ -123,19 +103,19 @@ impl trace::VMTracer for Informant {
...
-			memory = self.memory(),
-			stack = self.stack(),
-			storage = self.storage(),
-			depth = self.depth,
-		);
-		self.traces.push(trace);
+		let trace = json!({
+			"pc": self.pc,
+			"op": self.instruction,
+			"opName": info.map(|i| i.name).unwrap_or(""),
```
niklasad1: minor thing this could simplified to:

```info.map_or("", |i| i.name)```

PR #9491, opened 2018-09-07
Fix Snapshot restoration failure on Windows

### LINKED ISSUE #9106: DB atomic swap failed: renameat2 failed with code: 22 at the end of a warp sync

> I'm running:
>
> - **Which Parity version?**: 2.0

At the end of a warp-sync and before having the normal sync, I get the error: `DB atomic swap failed: renameat2 failed with code: 22`

This does not prevent the sync to continue.

```bash
2018-07-12 12:51:39  Syncing snapshot 1723/1726     #6901   25/25 peers   5 MiB chain 31 MiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,  102 µs
2018-07-12 12:51:46  Syncing snapshot 1724/1726     #6901   25/25 peers   5 MiB chain 31 MiB db 0 bytes queue 10 KiB sync  RPC:  0 conn,    0 req/s,  102 µs
2018-07-12 12:51:50  DB atomic swap failed: renameat2 failed with code: 22
2018-07-12 12:51:56  Syncing #5945000 0xbaf7…07d7     0.00 blk/s    0.0 tx/s    0 Mgas/s     28+   62 Qed  #5945095   25/25 peers   103 KiB chain 34 KiB db 6 MiB queue 264 KiB sync  RPC:  0 conn,    0 req/s,  102 µs
2018-07-12 12:52:01  Syncing #5945000 0xbaf7…07d7     0.00 blk/s    0.0 tx/s    0 Mgas/s     44+  301 Qed  #5945350   25/25 peers   520 KiB chain 240 KiB db 32 MiB queue 1 MiB sync  RPC:  0 conn,    0 req/s,  102 µs
2018-07-12 12:52:06  Syncing #5945001 0xbdc6…bd91     0.20 blk/s   31.2 tx/s    1 Mgas/s      0+  505 Qed  #5945508   25/25 peers   836 KiB chain 2 MiB db 51 MiB queue 297 KiB sync  RPC:  0 conn,    0 req/s,  102 µs
```

ngotchac:
Fixes #9106 

The issue would only arise on Windows: the files used for the Blooms database were not properly closed before restoring the snapshot DB. Thus windows would fail to move the directories around.

In addition to this PR, having https://github.com/paritytech/parity-common/pull/53 and https://github.com/debris/fs-swap/pull/8 would be great!

### DIFF
--- a/util/blooms-db/src/db.rs
+++ b/util/blooms-db/src/db.rs
@@ -14,12 +14,16 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::{io, fmt};
+use std::{error, io, fmt};
 use std::path::{Path, PathBuf};
 
 use ethbloom;
 
 use file::{File, FileIterator};
+
+fn other_io_err<E>(e: E) -> io::Error where E: Into<Box<error::Error + Send + Sync>> {
+	io::Error::new(io::ErrorKind::Other, e)
+}
 
 /// Bloom positions in database files.
 #[derive(Debug)]
@@ -39,8 +43,14 @@ impl Positions {
 	}
 }
 
-/// Blooms database.
-pub struct Database {
+struct DatabaseFilesIterator<'a> {
+	pub top: FileIterator<'a>,
+	pub mid: FileIterator<'a>,
+	pub bot: FileIterator<'a>,
+}
+
+/// Blooms database files.
+struct DatabaseFiles {
 	/// Top level bloom file
 	///
 	/// Every bloom represents 16 blooms on mid level
@@ -53,6 +63,46 @@ pub struct Database {
 	///
 	/// Every bloom is an ethereum header bloom
 	bot: File,
+}
+
+impl DatabaseFiles {
+	/// Open the blooms db files
+	pub fn open(path: &PathBuf) -> io::Result<DatabaseFiles> {
+		// let path = path.as_ref();
+		Ok(DatabaseFiles {
+			top: File::open(path.join("top.bdb"))?,
+			mid: File::open(path.join("mid.bdb"))?,
+			bot: File::open(path.join("bot.bdb"))?,
+		})
+	}
+
+	pub fn accrue_bloom(&mut self, pos: Positions, bloom: ethbloom::BloomRef) -> io::Result<()> {
+		self.top.accrue_bloom::<ethbloom::BloomRef>(pos.top, bloom)?;
+		self.mid.accrue_bloom::<ethbloom::BloomRef>(pos.mid, bloom)?;
+		self.bot.replace_bloom::<ethbloom::BloomRef>(pos.bot, bloom)?;
+		Ok(())
+	}
+
+	pub fn iterator_from(&mut self, pos: Positions) -> io::Result<DatabaseFilesIterator> {
+		Ok(DatabaseFilesIterator{
+			top: self.top.iterator_from(pos.top)?,
+			mid: self.mid.iterator_from(pos.mid)?,
+			bot: self.bot.iterator_from(pos.bot)?,
+		})
+	}
+
+	pub fn flush(&mut self) -> io::Result<()> {
+		self.top.flush()?;
+		self.mid.flush()?;
+		self.bot.flush()?;
+		Ok(())
+	}
+}
+
+/// Blooms database.
+pub struct Database {
+	/// Database files
+	db_files: Option<DatabaseFiles>,
 	/// Database path
 	path: PathBuf,
 }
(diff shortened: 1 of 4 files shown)


ForauDNB:
Thankyou for bringing up this issue for investigation. Is anybody aware of any manual workarounds in the mean time?

ngotchac:
Sadly there are no manual fixes other than not using Windows or using an older Parity version :/

ForauDNB:
Ok thanks. I'm not too worried as for me this only affects mainnet, ropsten is syncing fine and that's all I need at the moment for solidity development purposes :). Although I may just end up dual booting my new machine. Thanks for your help, keep up the great work 👍

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
...
+	}
+
+	pub fn accrue_bloom(&mut self, pos: Positions, bloom: ethbloom::BloomRef) -> io::Result<()> {
+		self.top.accrue_bloom::<ethbloom::BloomRef>(pos.top, bloom)?;
+		self.mid.accrue_bloom::<ethbloom::BloomRef>(pos.mid, bloom)?;
+		self.bot.replace_bloom::<ethbloom::BloomRef>(pos.bot, bloom)?;
+		Ok(())
+	}
+
+	pub fn iterator_from(&mut self, pos: Positions) -> io::Result<DatabaseFilesIterator> {
```
dvdplm: Just a curiosity: This method looks a little like the `std::io::Seek` trait. Would it be possible to refactor things so that `Positions` become `Position` and consist of single `u64` (which is then shifted properly to form the three variants needed for top, middle and bottom, kind of what happens in the `insert_blooms` method) and then impl `seek` for `DatabaseFilesIterator`? Not sure what we'd gain, and not in scope for this bug-fix PR but interested in your thoughts! :)

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
...
-			// most of the time events at block n(a) occur also on block n(b) or n+1(b)
-			self.top.accrue_bloom::<ethbloom::BloomRef>(pos.top, bloom)?;
-			self.mid.accrue_bloom::<ethbloom::BloomRef>(pos.mid, bloom)?;
-			self.bot.replace_bloom::<ethbloom::BloomRef>(pos.bot, bloom)?;
+		match self.db_files {
+			Some(ref mut db_files) => {
+				for (index, bloom) in (from..).into_iter().zip(blooms.map(Into::into)) {
+					let pos = Positions::from_index(index);
+
+					// constant forks make lead to increased ration of false positives in bloom filters
```
dvdplm: s/constant forks make lead to increased ration of /Constant forks may lead to increased ratio of/

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
...
+					// since we do not rebuild top or mid level, but we should not be worried about that
+					// most of the time events at block n(a) occur also on block n(b) or n+1(b)
```
dvdplm: s/most of the times/because most of the time/

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
...
+
+	pub fn iterator_from(&mut self, pos: Positions) -> io::Result<DatabaseFilesIterator> {
+		Ok(DatabaseFilesIterator{
+			top: self.top.iterator_from(pos.top)?,
+			mid: self.mid.iterator_from(pos.mid)?,
+			bot: self.bot.iterator_from(pos.bot)?,
+		})
+	}
+
+	pub fn flush(&mut self) -> io::Result<()> {
```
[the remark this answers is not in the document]
andresilva: We should probably flush on drop, but also keep the explicit flushing.

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
...
 
 		Ok(database)
 	}
 
+	/// Close the inner-files
+	pub fn close(&mut self) -> io::Result<()> {
+		if let Some(ref mut db_files) = self.db_files {
+			db_files.flush()?;
+		}
+		self.db_files = None;
```
andresilva: If we impl `Drop` for `DatabaseFiles` we can avoid the explicit flush above right?

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,121 @@ pub struct Database {
 	///
 	/// Every bloom is an ethereum header bloom
 	bot: File,
+}
+
+impl DatabaseFiles {
+	/// Open the blooms db files
```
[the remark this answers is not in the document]
ngotchac: Well it's called `blooms-db` :p

util/blooms-db/src/db.rs
ngotchac: Yep !

util/blooms-db/src/db.rs
ngotchac: I'm not actually quite sure how this `blooms-db` work so I'd say it's (as you said) out-of-scop for this PR, but @debris might have an opinion on this

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,124 @@ pub struct Database {
...
+	db_files: Option<DatabaseFiles>,
 	/// Database path
 	path: PathBuf,
 }
 
 impl Database {
 	/// Opens blooms database.
 	pub fn open<P>(path: P) -> io::Result<Database> where P: AsRef<Path> {
-		let path = path.as_ref();
+		let path: PathBuf = path.as_ref().to_owned();
```
sorpaas: Why is this changed to take owned value? `Path` also has `join` which may just be enough.

ngotchac:
@sorpaas So the issue is that (on Windows) we need to first close the handles to the `blooms-db` files before being able to move the `key_value` DB (by calling the `restore` method): so first close the `blooms` handles, call `restore` on the `kvdb` and then re-open the handles. So you can't just drop the values and re-open them, since we need to do something between the two.
I found that the minimal changes would be to add this `close` method to the `blooms-db`, which already has the `path` value stored in `struct Database`.
Does this answer your comment?

util/blooms-db/src/db.rs
```diff
@@ -53,68 +63,124 @@ pub struct Database {
...
+
+	pub fn accrue_bloom(&mut self, pos: Positions, bloom: ethbloom::BloomRef) -> io::Result<()> {
+		self.top.accrue_bloom::<ethbloom::BloomRef>(pos.top, bloom)?;
+		self.mid.accrue_bloom::<ethbloom::BloomRef>(pos.mid, bloom)?;
+		self.bot.replace_bloom::<ethbloom::BloomRef>(pos.bot, bloom)?;
+		Ok(())
+	}
+
+	pub fn iterator_from(&mut self, pos: Positions) -> io::Result<DatabaseFilesIterator> {
+		Ok(DatabaseFilesIterator{
```
sorpaas: Missing a whitespace between `DatabaseFilesIterator` and `{`.

sorpaas:
Understood. Indeed we need `close`. Have two final grumbles!

PR #9492, opened 2018-09-07
Don't error when Snapshot is aborted

ngotchac:
This would previously cause to delete all downloaded chunks if the snapshot was aborted while restoring a chunk.

### DIFF
--- a/ethcore/src/snapshot/service.rs
+++ b/ethcore/src/snapshot/service.rs
@@ -589,7 +589,7 @@ impl Service {
 		// TODO: be able to process block chunks and state chunks at same time?
 		let mut restoration = self.restoration.lock();
 		match self.feed_chunk_with_restoration(&mut restoration, hash, chunk, is_state) {
-			Ok(()) => (),
+			Ok(()) |
 			Err(Error(SnapshotErrorKind::Snapshot(SnapshotError::RestorationAborted), _)) => (),
 			Err(e) => {
 				warn!("Encountered error during snapshot restoration: {}", e);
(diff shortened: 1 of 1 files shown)


ethcore/src/snapshot/service.rs
```diff
@@ -584,10 +585,19 @@ impl Service {
...
 
 	/// Feed a chunk of either kind. no-op if no restoration or status is wrong.
-	fn feed_chunk(&self, hash: H256, chunk: &[u8], is_state: bool) -> Result<(), Error> {
+	fn feed_chunk(&self, hash: H256, chunk: &[u8], is_state: bool) {
 		// TODO: be able to process block chunks and state chunks at same time?
 		let mut restoration = self.restoration.lock();
-		self.feed_chunk_with_restoration(&mut restoration, hash, chunk, is_state)
+		match self.feed_chunk_with_restoration(&mut restoration, hash, chunk, is_state) {
+			Ok(()) => (),
+			Err(Error(SnapshotErrorKind::Snapshot(SnapshotError::RestorationAborted), _)) => (),
```
ordian: We could unify this cases with 
```rust
Ok(()) | 
Err(Error(<snip>)) => (),
```

ethcore/src/snapshot/service.rs
```diff
@@ -584,10 +585,19 @@ impl Service {
 	}
 
 	/// Feed a chunk of either kind. no-op if no restoration or status is wrong.
```
dvdplm: Maybe update the doc-string to "Feed a chunk of either Block or Status kind. No-op if … …"?

PR #9480, opened 2018-09-05
Generalized blockReward and difficultyBombDelays config

sorpaas:
Replaces #9187. Rel https://github.com/paritytech/parity-ethereum/issues/8427

Make `blockReward` config in ethash to accept `multi`, and add `difficultyBombDelays` config. Removes EIP649 configurations.

PR #9505, opened 2018-09-10
Enable all Constantinople hard fork changes in constantinople_test.json

sorpaas:
Address grumbles on https://github.com/paritytech/parity-ethereum/pull/9480#pullrequestreview-153553566


sorpaas:
Yeah right. Sorry!

ethcore/res/ethereum/constantinople_test.json
```diff
@@ -1,16 +1,16 @@
...
 				"minimumDifficulty": "0x020000",
 				"difficultyBoundDivisor": "0x0800",
 				"durationLimit": "0x0d",
-				"blockReward": "0x29A2241AF62C0000",
+				"blockReward": "0x1BC16D674EC80000",
 				"homesteadTransition": "0x0",
 				"eip100bTransition": "0x0",
 				"difficultyBombDelays": {
-					"0": 3000000
+					"0": 8000000
```
5chdn: Why 8_000_000? What about

```
				"difficultyBombDelays": {
					"0": 3000000
					"5": 2000000
				}
```

to match eip 649, 1234?

cheme: I think those json test are used in a all or nothing basis (eg from ethereum test), so putting the sum of eip 649 and 1234 together at 0 makes sense (on two line it would be better but I think json dont allow to identical keys). I got a doubt 649 was a 3 million delay and 1234 another additional 5 million ?

5chdn: fine by me. 649 is 3 million, 1234 is 2 million. sum is 5 million.

5chdn:
tests::evm::test_blockhash_eip210_int


fails

ethcore/res/ethereum/constantinople_test.json
cheme: My mistake, when reading 1234, " 5 million blocks later than previously specified with the Homestead fork" gave me the wrong impression of additional 5M but it would have been only true if refering to byzantium fork. 
So yes we need to use 5M instead of 8M for this line.

sorpaas:
@cheme We still have eip210 config flag though. They're not yet deprecated. Before we remove it, how about let's just continue to use this json file to test out `test_blockhash_eip210_int`? I think creating a new test config just for eip210 might be a little bit overkill.

cheme:
I got the same feeling (a new json is a bit overkill), but the `constantinople_test.json` is going to be used by 'ethereum/tests' json rpc tests. So we may need a eip210_test.json file, or simply disable the test.

sorpaas:
Added `eip210_test.json`!

tolgatrz:
what date will it take place?

PR #9507, opened 2018-09-10
adds parity_verifySignature RPC method

### LINKED ISSUE #7009: RPC method to verify the signature of a message
```
parity_verifySignature(is_prefixed: bool, message: Bytes, signature: H512) -> Result<(Address, PublicKey), Error>
```

ec recovers the public key and returns it along with ethereum address (might be better to use some rich struct instead of a tuple.

seunlanlege:
Closes #7009


rpc/src/v1/helpers/mod.rs
```diff
@@ -34,8 +34,10 @@ mod signer;
 mod signing_queue;
 mod subscribers;
 mod subscription_manager;
+mod verify_signature;
```
niklasad1: this file doesn't exist change it to `mod signature` or change the file name itself?

rpc/src/v1/helpers/errors.rs
```diff
@@ -276,6 +276,14 @@ pub fn signing(error: AccountError) -> Error {
 	}
 }
 
+pub fn verification_error<T: fmt::Display>(data: T) -> Error {
+	Error {
+		code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST),
+		message: format!("{}", data).into(),
```
niklasad1: Needless `into()`

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,27 @@
...
+use jsonrpc_core::Result;
+use v1::types::{Bytes, H520, RichBasicAccount, BasicAccount};
+use v1::helpers::errors;
+use ethereum_types::H256;
+use tiny_keccak::Keccak;
+
+pub fn verify_signature(is_prefixed: bool, message: Bytes, signature: H520) -> Result<RichBasicAccount> {
+	let mut buf = [0; 32];
+	buf.copy_from_slice(&message.0[..]);
+	let mut message = H256(buf);
```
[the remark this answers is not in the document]
seunlanlege: did not think of this :sweat_smile:

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,27 @@
...
+
+	if is_prefixed {
+		let mut buf = [0; 32];
+		let mut keccak = Keccak::new_keccak256();
```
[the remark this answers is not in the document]
seunlanlege: thanks!

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,27 @@
...
+pub fn verify_signature(is_prefixed: bool, message: Bytes, signature: H520, chain_id: Option<u64>) -> Result<RichBasicAccount> {
+	let mut hash = keccak(message.0.clone());
+	if is_prefixed {
+		hash = eth_data_hash(message.0);
+	}
+
+	let signature = Signature::from(signature.0);
+	let is_valid_for_current_chain = chain_id.map(|chain| {
+		let v = signature.v();
+		if v > 1 && (v as u64) == chain {
```
seunlanlege: I'm pretty sure this is wrong

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,27 @@
...
+use v1::types::{Bytes, H520, RichBasicAccount, BasicAccount};
+use v1::helpers::errors;
+use v1::helpers::dispatch::eth_data_hash;
+use hash::keccak;
+
+pub fn verify_signature(is_prefixed: bool, message: Bytes, signature: H520, chain_id: Option<u64>) -> Result<RichBasicAccount> {
+	let mut hash = keccak(message.0.clone());
+	if is_prefixed {
+		hash = eth_data_hash(message.0);
+	}
```
niklasad1: Since you are not using the message after you have computed the `hash` the ```clone``` seems needless!

Consider: 
```let hash = if is_prefixed { eth_data_hash(message.0) } else { keccak(message.0) }; ```

Also I think keccak supports slices so should be possible to pass-by-reference too!

File: rpc/src/v1/impls/light/parity.rs [hunk 7]
```diff
@@ -33,20 +33,21 @@ use light::client::LightChainClient;
...
 use jsonrpc_macros::Trailing;
-use v1::helpers::{self, errors, ipfs, SigningQueue, SignerService, NetworkSettings};
+use v1::helpers::{self, errors, ipfs, SigningQueue, SignerService, NetworkSettings, verify_signature};
 use v1::helpers::dispatch::LightDispatcher;
 use v1::helpers::light_fetch::LightFetch;
 use v1::metadata::Metadata;
 use v1::traits::Parity;
 use v1::types::{
-	Bytes, U256, U64, H160, H256, H512, CallRequest,
+	Bytes, U256, U64, H160,H520, H256, H512, CallRequest,
```
niklasad1: whitespace after comma :smile_cat:

seunlanlege:
@tomusdrw @niklasad1 i think this is ready to be merged

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,31 @@
...
+		eth_data_hash(message.0)
+	} else {
+		keccak(message.0.clone())
+	};
+
+	let signature = Signature::from(signature.0);
+	let is_valid_for_current_chain = chain_id.map(|chain_id| {
+		let v = signature.v() as u64;
+		if v > 36  {
+			let decoded_v = (v - 35) - (chain_id * 2);
```
niklasad1: this can still overflow ```if chain_id*2 is bigger than v-35``` (maybe only theoretical but still ...)

Consider using checked_sub and saturated_mul for this use-case instead!

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,31 @@
...
+use v1::types::{Bytes, H520, RichBasicAccount, BasicAccount};
+use v1::helpers::errors;
+use v1::helpers::dispatch::eth_data_hash;
+use hash::keccak;
+
+pub fn verify_signature(is_prefixed: bool, message: Bytes, signature: H520, chain_id: Option<u64>) -> Result<RichBasicAccount> {
+	let hash = if is_prefixed {
+		eth_data_hash(message.0)
+	} else {
+		keccak(message.0.clone())
```
niklasad1: sorry to nag but please remove ```clone()``` here :1st_place_medal:

rpc/src/v1/helpers/signature.rs
seunlanlege: not sure why we need saturating_mul with chain_id, isn't it supposed to be a compile-time constant?

rpc/src/v1/helpers/signature.rs
```diff
@@ -9,16 +9,16 @@ pub fn verify_signature(is_prefixed: bool, message: Bytes, signature: H520, chai
...
 		eth_data_hash(message.0)
 	} else {
-		keccak(message.0.clone())
+		keccak(message.0)
 	};
 
 	let signature = Signature::from(signature.0);
 	let is_valid_for_current_chain = chain_id.map(|chain_id| {
 		let v = signature.v() as u64;
 		if v > 36  {
```
niklasad1: @seunlanlege 

this if expression is unnecessary now so remove it and return false from the match instead :+1:

Also I would prefer if you can perform ```saturated_mul``` on chain_id * 2 because it can also overflow sorry if I was unclear

rpc/src/v1/tests/helpers/signature.rs
```diff
@@ -0,0 +1,57 @@
+use std::sync::Arc;
```
tomusdrw: New files need a preamble, but I think it would be much better to add the tests in the same file as the function-under-test. Just add `#[cfg(test)] mod tests {` and put the content there.

rpc/src/v1/tests/helpers/signature.rs
```diff
@@ -0,0 +1,57 @@
...
+	let (v, r, s) = (sig.v(), sig.r(), sig.s());
+	let v = add_chain_replay_protection(v as u64, signining_chain_id);
+	let account = verify_signature(is_prefixed, data.into(), v, r.into(), s.into(), rpc_chain_id).unwrap();
+	assert_eq!(account.inner.address, address);
+	assert_eq!(account.inner.is_valid_for_current_chain, is_valid_for_current_chain)
+}
+
+#[test]
+fn test_verify_signature_prefixed_mainnet() {
+	setup(true, Some(1), Some(1), true)
```
tomusdrw: Test in a current form is pretty unreadable - I think it would be way better to avoid passing so many bools and numbers without some additional description.
I think the `setup` should only be producing the signature, the assertion and call to `verify_signature` should be inlined and duplicated in every test, like this:

```rust
#[test]
fn test_verify_signature_not_prefixed_mainnet() {
  let should_prefix = true;
  let data = vec![5u8];
  let chain_id = Some(1);
  let (address, v, r, s) = sign(should_prefix, &data, chain_id);

  let account = verify_signature(should_prefix, &data, v, r, s, chain_id).unwrap();

  assert_eq!(account.inner.address, address);
  assert_eq!(account.inner.is_valid_for_current_chain, true);
}
```

IMHO that would really help readability/understandability of the test. If you want to keep the de-duplication, then consider using this:

```rust
#[test]
fn test_verify_signature_not_prefixed_mainnet() {
  run_test(TestCase {
    should_prefix: true,
    signing_chain_id: Some(1),
    rpc_chain_id: Some(1),
    is_valid_for_current_chain: true,
  });
}
```

rpc/src/v1/tests/mocked/parity.rs
```diff
@@ -531,3 +531,26 @@ fn rpc_parity_call() {
...
+
+	let request = r#"{
+		"jsonrpc": "2.0",
+		"method": "parity_verifySignature",
+		"params": [
+			false,
+			"0xe552acf4caabe9661893fd48c7b5e68af20bf007193442f8ca05ce836699d75e",
+			"0x2089e84151c3cdc45255c07557b349f5bf2ed3e68f6098723eaa90a0f8b2b3e5",
+			"0x5f70e8df7bd0c4417afb5f5a39d82e15d03adeff8796725d8b14889ed1d1aa8a",
+			1
```
tomusdrw: In RPC we only use non hex-encoded numbers for indices/limits. For constistency this should be `U64` type (hex-encoded ser/de)

rpc/src/v1/traits/parity.rs
```diff
@@ -218,5 +218,9 @@ build_rpc_trait! {
 		/// Call contract, returning the output data.
 		#[rpc(name = "parity_call")]
 		fn call(&self, Vec<CallRequest>, Trailing<BlockNumber>) -> Result<Vec<Bytes>>;
+
+		/// ecrecover signature
+		#[rpc(name = "parity_verifySignature")]
+		fn verify_signature(&self, bool, Bytes, H256, H256, u64) -> Result<RichBasicAccount>;
```
tomusdrw: I still don't understand why we need `RichBasicAccount` here, if we don't pass any "rich" data. Just return the `BasicAccount` and get rid of the `to_rich_struct` helper func.

5chdn:
Please, rebase on master

rpc/src/v1/traits/parity.rs
```diff
@@ -218,5 +217,9 @@ build_rpc_trait! {
 		/// Call contract, returning the output data.
 		#[rpc(name = "parity_call")]
 		fn call(&self, Vec<CallRequest>, Trailing<BlockNumber>) -> Result<Vec<Bytes>>;
+
+		/// ecrecover signature
```
dvdplm: The docs here are a bit…short? I don't see the params being documented, but perhaps you'll address that in a separate PR? Maybe something like: "/// Extracts Address and public key from signature using the r, s and v params. Equivalent to Solidity erecover"?

rpc/src/v1/types/account_info.rs
```diff
@@ -41,3 +43,19 @@ pub struct HwAccountInfo {
 	/// Device manufacturer.
 	pub manufacturer: String,
 }
+
+/// account derived from a signature
+/// as well as information that tells if it is valid for
+/// the current chain
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
```
dvdplm: Do we really need all these derives? It's hard to tell from the code in this PR if that is the case. Thanks!

rpc/src/v1/types/account_info.rs
```diff
@@ -41,3 +43,19 @@ pub struct HwAccountInfo {
...
+#[serde(rename_all="camelCase")]
+pub struct BasicAccount {
+	/// address of the recovered account
+	pub address: Address,
+	/// public key of the recovered account
+	pub public_key: Public,
+	/// if the signature contains chain replay protection
+	/// and the chain_id encoded within the signature matches the current chain
+	/// this would be true, otherwise false.
```
dvdplm: Good docs on this type; would be great if you could fix the line lengths and start all sentences with a capital letter.

rpc/src/v1/types/account_info.rs
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
```
dvdplm: Would be great with some module docs here; I think these types are all return types for RPC calls? If so, perhaps "Return types for RPC calls" is enough.

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,176 @@
...
+pub fn verify_signature(is_prefixed: bool, message: Bytes, v: U64, r: H256, s: H256, chain_id: Option<u64>) -> Result<BasicAccount> {
+	let hash = if is_prefixed {
+		eth_data_hash(message.0)
+	} else {
+		keccak(message.0)
+	};
+	let v: u64 = v.into();
+	let is_valid_for_current_chain = match (chain_id, v) {
+		(None, v) if v == 0 || v == 1 => true,
+		(Some(chain_id), v) if v > 36 => (v - 35) / 2 == chain_id,
```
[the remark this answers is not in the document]
seunlanlege: if there's no `chain_id`, then `v` must either 1/0 to be a valid for the current chain.
otherwise, if there's a `chain_id` and `v` > 36; (36 is the lowest value, where `v` = 0, `chain_id` = 1)
we check if the encoded `chain_id` is valid for the current chain.

(`v` - 35) / 2 == `chain_id`

rpc/src/v1/helpers/errors.rs
```diff
@@ -309,22 +309,22 @@ pub fn transaction_message(error: &TransactionError) -> String {
...
 			format!("Transaction gas price is too low. It does not satisfy your node's minimal gas price (minimal: {}, got: {}). Try increasing the gas price.", minimal, got)
-		},
+		}
 		InsufficientBalance { balance, cost } => {
 			format!("Insufficient funds. The account you tried to send transaction from does not have enough funds. Required {} and got: {}.", cost, balance)
-		},
+		}
 		GasLimitExceeded { limit, got } => {
 			format!("Transaction cost exceeds current gas limit. Limit: {}, got: {}. Try decreasing supplied gas.", limit, got)
-		},
```
NikVolf: all those above comma removal are unrelated to the PR topic

seunlanlege: My editor removes trailing commas, do want me to revert?

rpc/src/v1/helpers/signature.rs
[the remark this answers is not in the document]
niklasad1: This is related to extract to chain_id and remove the replay protection right?

This code is hard to read, can you add an inline comment and refer `EIP 155` and perhaps add a one-liner function named `remove_replay_protection` or something similar?

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,182 @@
...
+		_ => false,
+	};
+
+	let v = if v > 36 { (v - 1) % 2 } else { v };
```
niklasad1: Can you explain why you are not using https://github.com/paritytech/parity-ethereum/blob/master/ethcore/transaction/src/transaction.rs#L89-#L97 here?

Also, I think it should be `v >= 35` if we want to allow chain_id 0 here

rpc/src/v1/helpers/signature.rs
```diff
@@ -0,0 +1,182 @@
...
+
+	let signature = Signature::from_rsv(&r.into(), &s.into(), v as u8);
+	let public = recover(&signature, &hash).map_err(errors::encryption)?;
+	let address = public_to_address(&public);
+	Ok(BasicAccount { address, public_key: public, is_valid_for_current_chain })
```
niklasad1: hrm, if `is_valid_for_current_chain == false` isn't better to return Error here? Or please explain :1st_place_medal:

rpc/src/v1/types/account_info.rs
```diff
@@ -41,3 +44,19 @@ pub struct HwAccountInfo {
 	/// Device manufacturer.
 	pub manufacturer: String,
 }
+
+/// account derived from a signature
+/// as well as information that tells if it is valid for
+/// the current chain
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all="camelCase")]
+pub struct BasicAccount {
```
niklasad1: Why not rename this to `RecoveredAccount` instead of `BasicAccount` which can be confused with `ethcore::BasicAccount` (state of an account) ?

rpc/src/v1/helpers/signature.rs
seunlanlege: no it's included in the return payload.

rpc/src/v1/helpers/signature.rs
seunlanlege: v isn't going to be in electrum notation.

niklasad1:
@dvdplm Can you have look at his one again? I think we should merge it

@seunlanlege  
Rebase please :P

5chdn:
@dvdplm Can you have a look at this one again?

5chdn:
Happy to merge this if you could resolve the conflicts again

PR #9510, opened 2018-09-10
Update tokio-uds

tomaka:
tokio-uds 0.1 doesn't compile on Android anymore.
Updates to tokio-uds 0.2.

I've said it before, but it would be really nice to detect the moment the build breaks, and not later.

cc @General-Beck

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1788,7 +1788,7 @@ dependencies = [
 [[package]]
 name = "mio-named-pipes"
 version = "0.1.6"
-source = "git+https://github.com/alexcrichton/mio-named-pipes#57a6298769fce2c66531d4085bdaf8c48ce64bce"
+source = "git+https://github.com/alexcrichton/mio-named-pipes#2072ae0de5b3632dbb065fcd9c8be6c6a2fc39ae"
 dependencies = [
  "log 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


tomaka:
All I did was `cargo update -p parity-tokio-ipc` :confused:

tomaka:
This is in fact the JSON-RPC crate not compiling anymore.

tomusdrw:
You need to use a different branch of `parity-tokio-ipc` and only bump the uds stuff. `parity-tokio-ipc` git master branch have migrated to use latest `tokio` while rest of parity stack doesn't so we shouldn't really update yet. Plus it has a different API that is not supported in `jsonrpc`.

ordian:
Note, that `parity-tokio-ipc` [pinned revision](https://github.com/paritytech/jsonrpc/pull/311) still depends on `tokio-uds ` version `0.1`, and its master has quite a few breaking changes, so the migration might be quite involved.

5chdn:
can you rebase on master to see if this fixes android?

5chdn:
Can we put some priority on this?

tomaka:
I don't think this PR doesn't make sense anymore, and I'm not even sure what's broken.
What is sure is that `tokio-uds 0.1` doesn't compile on Android.
I don't know anything about the jsonrpc crate, so someone better informed should handle that.

PR #9511, opened 2018-09-10
new ethabi

debris:
this pull request updates the repo to use the latest version of ethabi.

@svyatonik please carefully review changes in secret store as they were the most complicated :)


secret_store/src/listener/service_contract.rs
```diff
@@ -449,318 +446,280 @@ pub fn mask_topics(mask: &ApiMask) -> Vec<H256> {
...
 	/// Check if response from key server is required.
-	pub fn is_response_required(client: &Client, contract_address: &Address, contract: &service::Service, server_key_id: &ServerKeyId, key_server: &Address) -> bool {
+	pub fn is_response_required(client: &Client, contract_address: &Address, server_key_id: &ServerKeyId, key_server: &Address) -> bool {
 		// we're checking confirmation in Latest block, because we're interested in latest contract state here
-		let do_call = |data| client.call_contract(BlockId::Latest, *contract_address, data);
-		contract.functions()
-			.is_document_key_store_response_required()
-			.call(*server_key_id, key_server.clone(), &do_call)
-			.unwrap_or(true)
+		let (encoded, decoder) = service::functions::is_server_key_generation_response_required::call(*server_key_id, *key_server);
```
svyatonik: `is_server_key_generation_response_required` -> `is_document_key_store_response_required`

svyatonik:
`ServiceTransactionsChecker` + SS-related changes are looking good (except for 1 typo above). I'll take a look again tomorrow with a clear head (probably missed something today - not feeling good).

hash-fetch/src/urlhint.rs
```diff
@@ -109,7 +108,6 @@ impl URLHintContract {
 	/// Creates new `URLHintContract`
 		pub fn new(client: Arc<RegistrarClient<Call=Asynchronous>>) -> Self {
```
dvdplm: indentation is off here

dvdplm:
I am way too unfamiliar with the this code to be able to say something intelligent about it. As a general comment it seems like the new interface is slightly less ergonomic than the old (I believe the change was dictated by performance reasons though, so that's fine).
While reading I wondered if we're opening up for nasty bugs now that the contract functions are free standing functions and not tied to the type as struct members; imagine we have two contracts with identical call signatures and we happen to call the wrong one. But as said, I'm way too unfamiliar with this stuff to tell if it's a real concern or not.

ethcore/src/engines/block_reward.rs
```diff
@@ -81,15 +81,13 @@ impl Into<trace::RewardType> for RewardKind {
 #[derive(PartialEq, Debug)]
 pub struct BlockRewardContract {
 	kind: SystemOrCodeCallKind,
-	block_reward_contract: block_reward_contract::BlockReward,
 }
```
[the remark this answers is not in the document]
debris: I'd prefer to do it in a separate pr. It seems to be a trivial, but a relatively big change

debris:
> While reading I wondered if we're opening up for nasty bugs now that the contract functions are free standing functions and not tied to the type as struct members;

tbh, I don't see how tying it to a struct could is any different than calling a freestanding method

> imagine we have two contracts with identical call signatures and we happen to call the wrong one. 

if the signatures of *contract functions* are the same, nothing bad would happen cause the result of such encoding would be the same.

but I guess you wanted to ask about calling two functions that have the accept params of the same type. well... there is not much we can do to prevent it... we just need to write appropriate tests for it

PR #9514, opened 2018-09-10
Fix checkpointing when creating contract failed

sorpaas:
In `revert_to_checkpoint` case, when an entry exists but `None`, it should invalidate current cache. So in `discard_checkpoint` we shouldn't merge it with downward checkpoint entries.


5chdn:
Adding patch label to make sure this ends up in 2.1-beta once branched out.

PR #9522, opened 2018-09-11
Fix a panic on create2 new opcode

cheme:
Memory calculation for create2 is the same as for create because the additional parameter was popped before.

Alternate fix would be to stop poping the salt in interpreter `exec_instruction` function. This is a bit more robust (this fix only is fine if `exec_instruction` is called before the `requirement` call), but require more change to fix.

I did not add a test case because it is covered by src/GeneralStateTestsFiller/stCreate2/create2InitCodesFiller.json of ethereum/tests. 

Please note that this json test case does not currently run in our CI (constantinople test case are not yet all running).


debris:
> Please note that this json test case does not currently run in our CI (constantinople test case are not yet all running).

Can we enable it?

PR #9526, opened 2018-09-11
Improve P2P discovery

### LINKED ISSUE #9549: Network fragmentation on Ropsten?
- **Parity Ethereum version**: 2.0.5, 2.1.0
- **Operating system**: Linux, Ubuntu
- **Fully synchronized**: no
- **Network**: ropsten
- **Restarted**: yes

I can't sync ropsten with either `beta` or `stable`, looks like a peering issue, I can only find 3-4 Parity peers and sync is stuck at block `599_264`, list of unique peers after two days running from two different clients on Ropsten:

- `Parity-Ethereum/v2.0.5-stable-f7c9c27-20180911/x86_64-linux-gnu/rustc1.28.0`
- `Parity-Ethereum/v2.1.0-beta-f59ba3d-20180911/x86_64-linux-gnu/rustc1.28.0`
- `Parity/v1.10.6-stable-bc0d134-20180605/x86_64-linux-gnu/rustc1.26.2`
- `Parity/v1.11.5-beta-0487c5b-20180629/x86_64-linux-gnu/rustc1.27.0`
- `Parity/v1.11.7-stable-085035f-20180717/x86_64-linux-gnu/rustc1.27.1`
- `Parity/v1.9.2-beta-0feb0bb-20180201/x86_64-linux-gnu/rustc1.23.0`

See also https://ropsten-stats.parity.io/

ngotchac:
This PR fixes a few things in the P2P discovery protocol.

- It fixes the wrongfuly read `To` field in the `Pong` packet that is actually useless
- It starts the first discovery round as soon as the first peers (bootnodes) have been pinged
- It adds a new _fast_ discovery timer that runs when the number of peers < minimum peers
- It doesn't times out `FindNode` requests that have been answered but with fewer nodes than expected (basically it just checks that some packets have been received)
- It takes into account the **wrong** hash that older `parity-ethereum` versions used to send back in `Pong` packets
- It doesn't remove timed out discovery nodes from the main sync, as it could be that UDP is just disabled and that would lead to removing all the nodes from the main sync

Here are some comparaisons between current master and this branch. Look at the number of peers growing.

Running node for 1m30 on `master`:
peers - master

Running node for 1m30 on `ng-p2p-discovery`
peers - ng-p2p-discovery

This PR enlights an issue that I still haven't been able to resolve: when there are many peers connected, there are from time to time some peers dropping all at the same time because of timeouts. It seems that something is holding the timer's thread for heavy processing, and when the thread is free again, many requests timed out and thus those peers are removed. If anyone is willing to look into it, that would be great.

fixes #9549


5chdn:
Wow :+1:

util/network-devp2p/src/discovery.rs
```diff
@@ -1096,14 +1202,6 @@ mod tests {
 			panic!("Expected discovery1 to be added to discovery1's table");
 		}
 
-		let ping_back = discovery2.dequeue_send().unwrap();
-		assert!(!discovery2.any_sends_queued());
-		let data = &ping_back.payload[(32 + 65)..];
-		assert_eq!(data[0], PACKET_PING);
-		let rlp = Rlp::new(&data[1..]);
-		assert_eq!(ep2, NodeEndpoint::from_rlp(&rlp.at(1).unwrap()).unwrap());
-		assert_eq!(ep1, NodeEndpoint::from_rlp(&rlp.at(2).unwrap()).unwrap());
```
ngotchac: On `Ping` packet received, the node that sent the request is now added to the NodeTable without going through another `Ping/Pong` round.

eosclassicteam:
Amazing feature, thanks 👍

util/network-devp2p/src/discovery.rs
```diff
@@ -601,51 +664,56 @@ impl<'a> Discovery<'a> {
...
-			self.expire_in_flight_request(node_id, sent_at);
-		}
-		while let Some((node_id, sent_at)) = self.expiring_finds.pop_front() {
-			if time.duration_since(sent_at) <= FIND_NODE_TIMEOUT {
-				self.expiring_finds.push_front((node_id, sent_at));
-				break;
+		});
+		self.in_flight_find_nodes.retain(|node_id, find_node_request| {
+			if time.duration_since(find_node_request.sent_at) > FIND_NODE_TIMEOUT {
+				if !find_node_request.answered {
```
jimpo: The rationale for requiring exactly BUCKET_SIZE results is discussed [here](https://github.com/paritytech/parity-ethereum/pull/8771#discussion_r201172193) and matches behavior in Geth. Can you explain why you prefer not to expire if there is an insufficient number of responses?

util/network-devp2p/src/discovery.rs
```diff
@@ -457,61 +471,105 @@ impl<'a> Discovery<'a> {
...
+						None
+					} else {
+						if request.deprecated_echo_hash == echo_hash {
+							info!(target: "discovery", "Got a Pong from old Parity version.");
+						}
+						// Get the UDP port from the saved request, and
+						// the address from the incoming connection
+						Some(NodeEntry {
+							id: *node_id,
+							endpoint: NodeEndpoint {
```
jimpo: Hmm, why use `from` at all? If the ping was sent to `request.node` and we got a response, seems like we should just keep using that.

util/network-devp2p/src/discovery.rs
```diff
@@ -80,15 +80,28 @@ impl BucketEntry {
...
+struct PingRequest {
+	// Time when the request was sent
 	sent_at: Instant,
-	packet_hash: H256,
-	response_count: usize, // Some requests (eg. FIND_NODE) have multi-packet responses
+	// The node to which the request was sent
+	node: NodeEntry,
+	// The hash sent in the Ping request
+	echo_hash: H256,
+	// The hash Parity used to respond with
```
jimpo: Can you include in the comment the Parity version that this was fixed in so this can be removed eventually?

util/network-devp2p/src/discovery.rs
```diff
@@ -317,35 +334,32 @@ impl<'a> Discovery<'a> {
 		self.public_endpoint.to_rlp_list(&mut rlp);
 		node.endpoint.to_rlp_list(&mut rlp);
 		append_expiration(&mut rlp);
+		let old_parity_hash = keccak(rlp.as_raw());
 		let hash = self.send_packet(PACKET_PING, &node.endpoint.udp_address(), &rlp.drain())?;
 
-		let request_info = PendingRequest {
-			packet_id: PACKET_PING,
+		// Add the request to inflight pings
```
jimpo: nit: This comment seems obvious to me from the code below.

util/network-devp2p/src/discovery.rs
```diff
@@ -457,61 +471,105 @@ impl<'a> Discovery<'a> {
...
 
 		let mut response = RlpStream::new_list(3);
-		dest.to_rlp_list(&mut response);
+		let pong_to = NodeEndpoint {
+			address: from.clone(),
+			udp_port: ping_from.udp_port
+		};
+		// Here the PONG's `To` field should be the node we are
+		// sending the request to
+		ping_to.to_rlp_list(&mut response);
```
jimpo: So this line should be deleted and replaced with the one below, right?

util/network-devp2p/src/discovery.rs
```diff
@@ -457,61 +471,105 @@ impl<'a> Discovery<'a> {
...
+			return Ok(None);
 		} else if !self.is_allowed(&entry) {
 			debug!(target: "discovery", "Address not allowed: {:?}", entry);
-		} else {
-			self.add_node(entry.clone());
+			return Ok(None);
 		}
-
-		Ok(None)
+		Ok(self.update_node(entry.clone()))
```
jimpo: Why the change from `add_node` to `update_node`? According to [this](https://github.com/ethereum/devp2p/blob/master/discv4.md#endpoint-proof), the node should only be added to the routing table after we receive a PONG (not just a PING).

util/network-devp2p/src/discovery.rs
```diff
@@ -457,61 +471,105 @@ impl<'a> Discovery<'a> {
...
-		let source = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
-		let dest = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
+		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_to = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
 		let timestamp: u64 = rlp.val_at(3)?;
 		self.check_timestamp(timestamp)?;
 
 		let mut response = RlpStream::new_list(3);
-		dest.to_rlp_list(&mut response);
+		let pong_to = NodeEndpoint {
```
jimpo: Why is this different from `ping_from`? It makes more sense to me to respond to the endpoint that the sender claims is their address than mixing and matching it with the `from` SocketAddr.

util/network-devp2p/src/discovery.rs
```diff
@@ -529,38 +531,40 @@ impl<'a> Discovery<'a> {
...
-		// 			});
-		// 			false
-		// 		} else {
-		// 			true
-		// 		}
-		// 	});
-		// }
+		if expected_node.is_none() {
+			let mut found = false;
+			self.in_flight_pings.retain(|_, ping_request| {
```
jimpo: When would this be the case? Iterating through the whole `in_flight_pings` map on every unexpected PONG doesn't seem like a good idea to me.

If there's a good reason for this, the map might as well be keyed on the ping hash (except that it would be harder to maintain compatibility with the `deprecated_echo_hash` thing).

util/network-devp2p/src/discovery.rs
```diff
@@ -591,65 +658,74 @@ impl<'a> Discovery<'a> {
...
+				false
+			} else {
+				true
 			}
-			self.expire_in_flight_request(node_id, sent_at, &mut removed);
+		});
+		for node_id in nodes_to_expire {
+			self.expire_node_request(node_id);
 		}
-		removed
```
jimpo: So when are unreliable nodes removed from the stored table now?

util/network-devp2p/src/discovery.rs
```diff
@@ -544,29 +606,34 @@ impl<'a> Discovery<'a> {
...
-					{
+					// Mark the request as answered
+					if !request.answered {
+						request.answered = true;
+					}
+					if request.response_count + results_count <= BUCKET_SIZE {
 						request.response_count += results_count;
 						true
 					} else {
+						debug!(target: "discovery", "Got unexpected Neighbors from {:?} ; oversized packet ({} + {}) node_id=0x{:x}", &from, request.response_count, results_count, node_id);
```
dvdplm: nit: you can use `{#:x}` instead of `0x{:x}`

util/network-devp2p/src/discovery.rs
```diff
@@ -115,13 +128,14 @@ pub struct Discovery<'a> {
 	id_hash: H256,
 	secret: Secret,
 	public_endpoint: NodeEndpoint,
+	started_discovering: bool,
+	discovering: bool,
```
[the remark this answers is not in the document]
ngotchac: So `started_discovery` is used in order to start the first round of discovery _ever_. The idea is that it is set to `false` at the beginning, then we add some peers (bootnodes), ping them, and as soon as the pings are answered or timedout, we start the first discovery round. Then we start a new round every 10s or 30s depending on the number of peers. Without this, we would have to wait for an extra 10s for it to run for the first time.
The `discovering` bool is used to prevent starting a new discovery round while there is already one going.

util/network-devp2p/src/discovery.rs
ngotchac: The idea is that they are not removed via the discovery protocol, but the actual TCP protocol has some mechanisms to remove unresponsive nodes (or ones on another chain, etc.)
The issue was for nodes that have UDP blocked and don't run with `--no-discovery`. The number of peers would slowly drop to 0 and we won't be able to recover from that.
I can run some more tests to see how it affects the sync though.

util/network-devp2p/src/discovery.rs
ngotchac: So this is to cater for the fact that the node answering your Ping might not have the NodeID you've been told. Another mechanism would be to Ping back any unknown received Pong ; I can try that if it sounds better.

util/network-devp2p/src/discovery.rs
ngotchac: To quote your link: _When a ping packet is received, the recipient should reply with a pong packet. It may also consider the sender for addition into the node table._

ngotchac: IMO it's just another unnecessary round of Ping-Pong.

util/network-devp2p/src/discovery.rs
```diff
@@ -115,6 +115,7 @@ pub struct Discovery<'a> {
 	id_hash: H256,
 	secret: Secret,
 	public_endpoint: NodeEndpoint,
+	discovering: bool,
```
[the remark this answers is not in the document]
ngotchac: Definitly, thanks!

util/network-devp2p/src/discovery.rs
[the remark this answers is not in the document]
dvdplm: I see, much clearer now. Can't really think of a better name, maybe s/`started_discovering`/`discovery_initiated`/. Or just a comment here?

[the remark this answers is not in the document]
ngotchac: Sounds better, thanks!

util/network-devp2p/src/discovery.rs
jimpo: Yes, `add_node` considers the sender for inclusion, but only after obtaining Endpoint Proof, as it is referred to in the doc.

I disagree that it is unnecessary. The purpose of PINGing the node is to ensure that it is listening at the network address it claims to be and participating in the protocol. When receiving a single UDP packet, the origin address can be spoofed, or the packet could be sent without the sender actually listening on the network. The linked document details the difficulty of verifying that a peer is honestly running a discovery server, but PINGing is a good heuristic.

util/network-devp2p/src/discovery.rs
jimpo: If the NodeID is wrong, then the address (NodeID + Network Endpoint) is wrong, so I think it's fine to just drop the PONG.

Consider also that the sender of the PONG will likely PING back to us after receiving the PING, from which we can get the correct NodeID, so this unlikely case should resolve itself anyway.

util/network-devp2p/src/discovery.rs
jimpo: Cool, seems reasonable.

util/network-devp2p/src/discovery.rs
ngotchac: Okay I'll remove it from now. Note that Geth doesn't Ping back an received Ping, it just adds it to their NodeTable .

util/network-devp2p/src/discovery.rs
ngotchac: Note that Geth doesn't Ping back an received Ping, it just adds it to their NodeTable .

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +464,107 @@ impl<'a> Discovery<'a> {
...
-
-		Ok(None)
+		Ok(self.update_node(entry.clone()))
 	}
 
 	fn on_pong(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr) -> Result<Option<TableUpdates>, Error> {
-		trace!(target: "discovery", "Got Pong from {:?}", &from);
-		let dest = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
+		trace!(target: "discovery", "Got Pong from {:?} ; node_id=0x{:x}", &from, node_id);
+		let _pong_to = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
```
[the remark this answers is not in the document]
ngotchac: It was actually just in case we figure out why this value is useful, as I cannot.

util/network-devp2p/src/discovery.rs
tomaka: I don't really know how devp2p works, but I think that it is useful to wait for the PONG, as nodes behind a NAT (which is probably not uncommon) may report an unreachable address.

util/network-devp2p/src/discovery.rs
ngotchac: Actually, this isn't really used from the implementations that I've read, but I think the idea is that the recipient of the Pong knows what we think his address is. There is no point in copying a field of the packet he just sent us, he probably knows it already.

util/network-devp2p/src/discovery.rs
andresilva: I think it was added in https://github.com/paritytech/parity-ethereum/pull/8042 and shipped in 1.11.0. AFAIK geth just ignores this value.

util/network-devp2p/src/discovery.rs
jimpo: OK, that makes sense.

[This](https://github.com/ethereum/go-ethereum/blob/c73b654fd172464f598c239b4e7c4bc07a641e3b/p2p/discover/udp.go#L329) [code](https://github.com/ethereum/go-ethereum/blob/c73b654fd172464f598c239b4e7c4bc07a641e3b/p2p/discover/udp.go#L438) in Geth is where it seems to me it would timeout FIND_NODE requests with insufficient results.

util/network-devp2p/src/discovery.rs
jimpo: I'm pretty see Geth checks the hash. [See here](https://github.com/ethereum/go-ethereum/blob/c73b654fd172464f598c239b4e7c4bc07a641e3b/p2p/discover/udp.go#L292).

util/network-devp2p/src/discovery.rs
ngotchac: So I think that the first bit is checking that the result is actually < 16 (it errors if `nreceived >= bucketSize`) ; and the second I'm not 100% sure but I think it's a general timeout machanism, not related to `FindNode`

util/network-devp2p/src/discovery.rs
jimpo: Well, it's used below as the entry that `add_node` is called on. Whereas I think `ping_from` is a better choice.

util/network-devp2p/src/discovery.rs
```diff
@@ -544,29 +575,34 @@ impl<'a> Discovery<'a> {
...
+		let is_expected = match self.in_flight_find_nodes.entry(*node_id) {
 			Entry::Occupied(mut entry) => {
-				let result = {
+				let expected = {
 					let request = entry.get_mut();
-					if request.packet_id == PACKET_FIND_NODE &&
-						request.response_count + results_count <= BUCKET_SIZE
-					{
+					// Mark the request as answered
+					if !request.answered {
```
jimpo: mega-nit: The `if !request.answered` is not necessary -- setting to true is idempotent.

util/network-devp2p/src/discovery.rs
jimpo: Where do you see that it errors? In my reading of it, [this code](https://github.com/ethereum/go-ethereum/blob/c73b654fd172464f598c239b4e7c4bc07a641e3b/p2p/discover/udp.go#L422), marks the request as complete once `nreceived >= bucketSize` and does nothing (waiting for more responses) if it's not.

ngotchac: Hm I think you are right, sorry I misread the code... Even so, I think the mechanism in this PR is better.

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +463,77 @@ impl<'a> Discovery<'a> {
...
 		}
-
 		Ok(None)
 	}
 
 	fn on_pong(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr) -> Result<Option<TableUpdates>, Error> {
-		trace!(target: "discovery", "Got Pong from {:?}", &from);
-		let dest = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
+		trace!(target: "discovery", "Got Pong from {:?} ; node_id={:#x}", &from, node_id);
+		let _pong_to = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
```
cheme: Not checking `_pong_to` seems possible (if ping_hash is checked), personally I would find it nice to have a short comment explaining why.

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +463,77 @@ impl<'a> Discovery<'a> {
...
 			Entry::Occupied(entry) => {
-				let is_expected = {
+				let expected_node = {
 					let request = entry.get();
-					request.packet_id == PACKET_PING && request.packet_hash == echo_hash
+					if request.echo_hash != echo_hash && request.deprecated_echo_hash != echo_hash {
+						debug!(target: "discovery", "Got unexpected Pong from {:?} ; packet_hash={:#x} ; expected_hash={:#x}", &from, request.echo_hash, echo_hash);
+						None
+					} else {
+						if request.deprecated_echo_hash == echo_hash {
```
cheme: I do not know how such deprecated functionalities are managed, ideally this should disappear in a few parity version I think. It is an open question, maybe we can use things like rust deprecated tag or other to keep trace of it (and not forget it in the base code ad vita eternam)

cheme: But maybe we want to keep it for better compatibility (not really my point of view)

util/network-devp2p/src/discovery.rs
```diff
@@ -591,65 +627,74 @@ impl<'a> Discovery<'a> {
...
+		// Start discovering if the first pings have been sent (or timed out)
+		} else if self.in_flight_pings.len() == 0 && !self.discovery_initiated {
+			self.discovery_initiated = true;
+			self.refresh();
+		}
 	}
 
 	pub fn refresh(&mut self) {
-		self.start();
+		if self.discovery_round.is_none() {
```
cheme: I don't really get it, whether it is fast or slow discovery timeout we end up on this same function : wouldn't this function be for fast only and slow discovery use the original one (with state clear and all).

util/network-devp2p/src/host.rs
```diff
@@ -70,7 +71,9 @@ const SYS_TIMER: TimerToken = LAST_SESSION + 1;
 // for IDLE TimerToken
 const MAINTENANCE_TIMEOUT: Duration = Duration::from_secs(1);
 // for DISCOVERY_REFRESH TimerToken
-const DISCOVERY_REFRESH_TIMEOUT: Duration = Duration::from_secs(60);
+const DISCOVERY_REFRESH_TIMEOUT: Duration = Duration::from_secs(30);
```
cheme: Not sure about the rationale for lowering this value.

util/network-devp2p/src/discovery.rs
```diff
@@ -42,9 +42,9 @@ const PACKET_PONG: u8 = 2;
 const PACKET_FIND_NODE: u8 = 3;
 const PACKET_NEIGHBOURS: u8 = 4;
 
-const PING_TIMEOUT: Duration = Duration::from_millis(300);
+const PING_TIMEOUT: Duration = Duration::from_millis(500);
```
cheme: From my point of view half a second does not seems a lot, and it is the same as geth, probably fine.

util/network-devp2p/src/discovery.rs
```diff
@@ -42,9 +42,9 @@ const PACKET_PONG: u8 = 2;
...
 const FIND_NODE_TIMEOUT: Duration = Duration::from_secs(2);
-const EXPIRY_TIME: Duration = Duration::from_secs(60);
+const EXPIRY_TIME: Duration = Duration::from_secs(20);
```
cheme: I am totally in favor of this change of value :+1:

util/network-devp2p/src/discovery.rs
```diff
@@ -239,8 +250,12 @@ impl<'a> Discovery<'a> {
 	}
 
 	fn discover(&mut self) {
-		self.update_new_nodes();
-		if self.discovery_round == DISCOVERY_MAX_STEPS {
+		if self.discovery_round.is_none() {
+			return;
+		}
+		let discovery_round = self.discovery_round.expect("discovery_round is not None ; qed");
```
cheme: Could use an `if let` pattern, but at the same time it indents +1 , keeping it as it is seems fine.

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +463,77 @@ impl<'a> Discovery<'a> {
...
-		let dest = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
+		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_to = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
 		let timestamp: u64 = rlp.val_at(3)?;
 		self.check_timestamp(timestamp)?;
 
 		let mut response = RlpStream::new_list(3);
-		dest.to_rlp_list(&mut response);
+		let pong_to = NodeEndpoint {
+			address: from.clone(),
```
cheme: Here we change a bit our code. Previously (if I am not mistaken), it was done in `on_pong` only if `!node.endpoint.is_valid()`. I think it is better to use socket address directly like done here (getting info from the transport layer), yet let's keep this change in mind.

util/network-devp2p/src/discovery.rs
ngotchac: Sorry I thought it was to `on_pong` method.

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +463,77 @@ impl<'a> Discovery<'a> {
...
 	}
 
 	fn on_pong(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr) -> Result<Option<TableUpdates>, Error> {
-		trace!(target: "discovery", "Got Pong from {:?}", &from);
-		let dest = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
+		trace!(target: "discovery", "Got Pong from {:?} ; node_id={:#x}", &from, node_id);
+		let _pong_to = NodeEndpoint::from_rlp(&rlp.at(0)?)?;
 		let echo_hash: H256 = rlp.val_at(1)?;
 		let timestamp: u64 = rlp.val_at(2)?;
 		self.check_timestamp(timestamp)?;
```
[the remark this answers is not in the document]
ngotchac: It will just go expired after a few 100ms, so I think it's fine as-is.

util/network-devp2p/src/discovery.rs
ngotchac: The real issue is that I'm not sure why is this field in the specs ; so I'm not sure I could write why we omit it :/

util/network-devp2p/src/discovery.rs
ngotchac: Not sure either... @5chdn maybe?

util/network-devp2p/src/discovery.rs
ngotchac: So _fast_ discovery is actually just calling this method more often, that's all.

util/network-devp2p/src/host.rs
ngotchac: Now that we have the _fast_ one, I guess I can increase it.

util/network-devp2p/src/discovery.rs
ngotchac: From the specs (https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery): _request timeouts are 300ms_ ; but I agree it' pretty low, that's why the 500ms

util/network-devp2p/src/discovery.rs
ngotchac: It would be actually quite a bad idea to use the `ping_from` that we have been sent, since it's mostly the peer local address, such as `V4(0.0.0.0:42786)` or `V6([::]:30303)`

util/network-devp2p/src/discovery.rs
5chdn: Please, add a `#[deprecated()]` to `deprecated_echo_hash` and move on :)

I really need this for the release:)

util/network-devp2p/src/discovery.rs
cheme: Yes I was thinking to avoid some replay, but it seems already cover by ping_hash. Don't know for sure either.

5chdn:
failing tests are unrelated. anyone else want to give this a final review?

util/network-devp2p/src/discovery.rs
ngotchac: So both _fast_ and _normal_ call the same thing:
```rust
self.discovery.lock().as_mut().map(|d| d.refresh());
```

The `refresh` method won't actually do anything if there is already a discovery going on:
```rust
pub fn refresh(&mut self) {
	if self.discovery_round.is_none() {
		self.start();
	}
}
```

Does that answers your concerns?

cheme: Not totally, but being rational and not looking at previous behaviour it seems better.

ngotchac: Ah sorry, yeah the previous implementation was a bit borked at this level, but since it was called every minute, it was very unlikely that the previous discovery was still going on when called.

util/network-devp2p/src/discovery.rs
ordian: Can be this though: 
```rust
let discovery_round = match self.discovery_round {
	Some(r) => r,
	None => return,
};
```

util/network-devp2p/src/discovery.rs
```diff
@@ -448,61 +464,77 @@ impl<'a> Discovery<'a> {
...
-		dest.to_rlp_list(&mut response);
+		let pong_to = NodeEndpoint {
+			address: from.clone(),
+			udp_port: ping_from.udp_port
+		};
+		// Here the PONG's `To` field should be the node we are
+		// sending the request to
+		// WARNING: this field _should not be used_, but old Parity versions
+		// use it in order to get the node's address.
+		// So this is a temporary fix so that older Parity versions don't brake completely.
```
jimpo: s/brake/break/

util/network-devp2p/src/discovery.rs
```diff
@@ -544,29 +576,32 @@ impl<'a> Discovery<'a> {
...
-		let is_expected = match self.in_flight_requests.entry(*node_id) {
+		let is_expected = match self.in_flight_find_nodes.entry(*node_id) {
 			Entry::Occupied(mut entry) => {
-				let result = {
+				let expected = {
 					let request = entry.get_mut();
-					if request.packet_id == PACKET_FIND_NODE &&
-						request.response_count + results_count <= BUCKET_SIZE
-					{
+					// Mark the request as answered
```
jimpo: nit: This comment feels unnecessary to me given the line below.

alexdupre:
It's unbelievable, finally someone was able to find the P2P discovery issues and fix them. Great job!

5chdn:
Hell, yes!

Issue #9430, opened 2018-08-28
Fetching transaction receipt performance


adyshimony:
Hi,

**100 blocks txs receipts - total of 15324 - takes 102 seconds.**

I am with latest Parity, on ubuntu, this is a constant issue.

I am indexing all transaction, and getting very slow performance when fetching transaction receipt.

For example: 

blocks 4832000 - 4832099
To get the 100 blocks, the duration in sec:, 8.921 (block include transactions body)

// now get the receipts for each transaction per block (I am waiting for all txs per block. waiting for all txs  for all blocks didn't improve performance)

getTransactionsFromBlockAsync for block: 4832000, txs: 252, duration in sec:, 1.929
getTransactionsFromBlockAsync for block: 4832001, txs: 103, duration in sec:, 0.190
getTransactionsFromBlockAsync for block: 4832002, txs: 79, duration in sec:, 0.134
getTransactionsFromBlockAsync for block: 4832003, txs: 64, duration in sec:, 0.064
getTransactionsFromBlockAsync for block: 4832004, txs: 251, duration in sec:, 2.090
getTransactionsFromBlockAsync for block: 4832005, txs: 258, duration in sec:, 2.317
getTransactionsFromBlockAsync for block: 4832006, txs: 302, duration in sec:, 3.510
getTransactionsFromBlockAsync for block: 4832007, txs: 87, duration in sec:, 0.162
getTransactionsFromBlockAsync for block: 4832008, txs: 192, duration in sec:, 1.355
getTransactionsFromBlockAsync for block: 4832009, txs: 323, duration in sec:, 4.140
getTransactionsFromBlockAsync for block: 4832010, txs: 162, duration in sec:, 0.754
getTransactionsFromBlockAsync for block: 4832011, txs: 54, duration in sec:, 0.047
getTransactionsFromBlockAsync for block: 4832012, txs: 43, duration in sec:, 0.032
getTransactionsFromBlockAsync for block: 4832013, txs: 24, duration in sec:, 0.014
getTransactionsFromBlockAsync for block: 4832014, txs: 68, duration in sec:, 0.084
getTransactionsFromBlockAsync for block: 4832015, txs: 121, duration in sec:, 0.376
getTransactionsFromBlockAsync for block: 4832016, txs: 15, duration in sec:, 0.011
getTransactionsFromBlockAsync for block: 4832017, txs: 247, duration in sec:, 2.373
[...]

PR #9527, opened 2018-09-11
RPC: parity_getBlockReceipts

tomusdrw:
Closes: #9430 

Some performance benchmarks:
After: http://gist.github.com/tomusdrw/05242fb0a7d42ed13474218e6a83f517
Before: https://gist.github.com/tomusdrw/f664ff78ce2714fadc2128fa5539cd86

(Not realy meant to compare `eth_getTransactionReceipt` performance before and after, cause obviously here is obviously a huge variance here, but I think it's clearly visible that getBlockReceipts is way faster and single receipt performance has not degraded)

Script used to check the performance:
https://gist.github.com/tomusdrw/ffd102326522d415ddea1b2aa1938cc9

EDIT: realised it's a bit unfair if we test `parity_getBlockReceipts` as the second one, cause the first call fills up the cache. Updated the results, but even if `getBlockReceipts` fetches data from disk it's faster, so it's not really a contributing factor here.


rpc/src/v1/traits/parity.rs
```diff
@@ -211,6 +211,11 @@ build_rpc_trait! {
 		#[rpc(name = "parity_getBlockHeaderByNumber")]
 		fn block_header(&self, Trailing<BlockNumber>) -> BoxFuture<RichHeader>;
 
+		/// Get block receipts.
+		/// Allows you to fetch receipts from the entire block at once.
+		#[rpc(name = "parity_getBlockReceipts")]
+		fn block_receipts(&self, Trailing<BlockNumber>) -> BoxFuture<Vec<Receipt>>;
```
debris: `Trailing` == latest?

rpc/src/v1/impls/light/parity.rs
```diff
@@ -415,6 +415,19 @@ impl Parity for ParityClient {
 		Box::new(self.fetcher().header(id).and_then(from_encoded))
 	}
 
+	fn block_receipts(&self, number: Trailing<BlockNumber>) -> BoxFuture<Vec<Receipt>> {
+		// Note: Here we treat `Pending` as `Latest`.
+		//       Since light clients don't produce pending blocks
+		//       (they don't have state) we can safely fallback to `Latest`.
```
debris: redundant extra spaces

rpc/src/v1/impls/eth.rs
```diff
@@ -419,11 +418,11 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
...
 
-	let pending_logs = receipts.into_iter()
-		.flat_map(|(hash, r)| r.logs.into_iter().map(|l| (hash.clone(), l)).collect::<Vec<(H256, LogEntry)>>())
-		.collect::<Vec<(H256, LogEntry)>>();
-
-	pending_logs.into_iter()
+	receipts.into_iter()
+		.flat_map(|r| {
+			let hash = r.transaction_hash;
+			r.logs.into_iter().map(move |l| (hash.clone(), l))
```
debris: `move` and `clone()` are probably redundant, cause `hash` implements `Copy`

ethcore/src/client/client.rs
```diff
@@ -1785,26 +1785,49 @@ impl BlockChainClient for Client {
...
-			}))
+		let number = chain.block_number(&hash)?;
+		let body = chain.block_body(&hash)?;
+		let mut receipts = chain.block_receipts(&hash)?.receipts;
+		receipts.truncate(address.index + 1);
+
+		let transaction = body.view().localized_transaction_at(&hash, number, address.index)?;
+		let receipt = receipts.pop()?;
+		let gas_used = receipts.last().map_or_else(|| 0.into(), |r| r.gas_used);
+		let no_of_logs = receipts.into_iter().map(|receipt| receipt.logs.len()).sum::<usize>();
```
debris: I don't understand these 3 lines. why do we pop a receipt and then take a `gas_used` from the one before it?

ethcore/src/client/client.rs
```diff
@@ -1785,26 +1785,49 @@ impl BlockChainClient for Client {
...
+
+		let receipt = transaction_receipt(self.engine().machine(), transaction, receipt, gas_used, no_of_logs);
+		Some(receipt)
+	}
+
+	fn block_receipts(&self, id: BlockId) -> Option<Vec<LocalizedReceipt>> {
```
jimpo: This method could probably use a test.

rpc/src/v1/impls/light/parity.rs
```diff
@@ -415,6 +415,19 @@ impl Parity for ParityClient {
...
+		let id = match number.unwrap_or_default() {
+			BlockNumber::Num(n) => BlockId::Number(n),
+			BlockNumber::Earliest => BlockId::Earliest,
+			BlockNumber::Latest | BlockNumber::Pending => BlockId::Latest,
```
jimpo: Since there is no Pending state, I'd say either return an empty vector or keep the [deprecation log](https://github.com/paritytech/parity-ethereum/blob/6e62d77e4d6211fca1b66a34f3b5047f3a406e77/rpc/src/v1/impls/light/eth.rs#L80) from `num_to_id`.

ethcore/src/client/client.rs
tomusdrw: So to actually generate a `LocalizedReceipt` we need 3 things:
1. Actual receipt
2. Gas used by that transaction
3. Number of logs generate prior to this transaction (to calculate log_index_in_block).

So the logic is as follows:
1. Fetch all receipts from 0 to `index` (inclusive)
2. Take the last one (`receipts.pop()`; point 1.)
3. Take the previous to last one to calculate (point 2.) (receipts only store cumulative gas, so we calculate `current.gas_used - previous.gas_used` to figure out what `current` gas usage of transaction was)
4. Use all previous receipts to calculate total number of logs. (point 3.)

ethcore/src/client/client.rs
tomusdrw: Added for both `block_receipt` and `transaction_receipt`

rpc/src/v1/types/block_number.rs
```diff
@@ -52,6 +52,22 @@ impl BlockNumber {
...
+	pub fn to_block_id(self) -> BlockId {
+		// NOTE Here we treat `Pending` as `Latest`.
+		// Since light clients don't produce pending blocks
+		// (they don't have state) we can safely fallback to `Latest`.
+		match self {
+			BlockNumber::Num(n) => BlockId::Number(n),
+			BlockNumber::Earliest => BlockId::Earliest,
+			BlockNumber::Latest => BlockId::Latest,
+			BlockNumber::Pending => {
+				warn!("`Pending` is deprecated and may be removed in future versions. Falling back to `Latest`");
```
jimpo: Hmm, this should only get logged when accessed from the light client impl. So maybe move back to `num_to_id` on the light client or add a `light_client: bool` method param?

tomusdrw: Moved it to a trait, so it explicitly says it's for light client only.

5chdn:
please rebase on master

PR #9407, opened 2018-08-23
ethcore: fix ancient block sync

andresilva:
Fixes #9300 and fixes #7008. Merge after #9381.

Issue #9225, opened 2018-07-26
Fails to sync ancient blocks from peer with the same difficulty


ascjones:
> I'm running:
>
> - **Which Parity version?**: 2.0.0
> - **Which operating system?**: Linux
> - **How installed?**: binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes

**Reproduce:**

Create a `reserved-peers` file with a node you know is fast and up to date with the best block.

`parity -l sync=trace --log-file=log-file --reserved-peers=reserved-peers --reserved-only`

Logs like the following should be produced:

```
1751-2018-07-26 15:51:39  IO Worker #3 TRACE sync  Considering peer 0, force=false, td=Some(0x12f966affeebce9f577), our td=5600202211188051408247, latest=0x9d9f…34a5, have_latest=false, state=Idle
1752:2018-07-26 15:51:39  IO Worker #3 TRACE sync  peer 0 is not suitable for asking old blocks
1753-2018-07-26 15:51:39  IO Worker #3 TRACE sync  Deactivating peer 0
```

Note that `0x12f966affeebce9f577 == 5600202211188051408247`

We would expect to be able to download old blocks from this peer. This also has the effect of not downloading old blocks from peers even if they are only a few behind.

PR #9531, opened 2018-09-11
Fix ancient blocks sync

### LINKED ISSUE #7008: Ancient block sync stalls

### LINKED ISSUE #9300: Background verifier sometimes goes back
> I'm running:
>
> - **Which Parity version?**: 1.11.8
> - **Which network are you connected to?**: core, sokol

We run a POA-based networks (sokol and core). After upgrade to 1.11.8 I'm getting hard time syncing with the network using warp-sync.
After all snapshots were downloaded, a verifier in the background starts processing blocks (if I understand correctly), here's a piece of log: 
```
2018-08-07 20:21:42    #1320360   15/25 peers   7 MiB chain 87 MiB db 0 bytes queue 5 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
2018-08-07 20:21:45  Imported #4050210 0x6620…f959 (2 txs, 0.25 Mgas, 16 ms, 0.85 KiB)
2018-08-07 20:21:51  Imported #4050211 0x7253…604b (1 txs, 0.02 Mgas, 0 ms, 0.67 KiB)
2018-08-07 20:21:55  Imported #4050212 0x3af3…1b8a (0 txs, 0.00 Mgas, 1 ms, 0.57 KiB)
2018-08-07 20:22:01  Imported #4050213 0xc563…69f7 (1 txs, 0.23 Mgas, 3 ms, 0.75 KiB)
2018-08-07 20:22:05  Imported #4050214 0x3187…d824 (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2018-08-07 20:22:10  Imported #4050215 0x688d…38f4 (1 txs, 0.23 Mgas, 4 ms, 0.75 KiB)
2018-08-07 20:22:12    #1345587   15/25 peers   6 MiB chain 87 MiB db 0 bytes queue 5 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
2018-08-07 20:22:15  Imported #4050216 0xef42…9d84 (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2018-08-07 20:22:20  Imported #4050217 0x9754…49d1 (2 txs, 0.40 Mgas, 4 ms, 0.99 KiB)
2018-08-07 20:22:25  Imported #4050218 0x5528…1323 (0 txs, 0.00 Mgas, 1 ms, 0.57 KiB)
2018-08-07 20:22:30  Imported #4050219 0x30e4…589e (1 txs, 0.23 Mgas, 2 ms, 0.75 KiB)
2018-08-07 20:22:41  Imported #4050221 0x3dbc…4ce5 (1 txs, 0.08 Mgas, 1 ms, 0.80 KiB) + another 1 block(s) containing 1 tx(s)
2018-08-07 20:22:46  Imported #4050222 0x28b2…f2fa (1 txs, 0.07 Mgas, 5 ms, 0.80 KiB)
2018-08-07 20:22:50    #1375534   15/25 peers   7 MiB chain 87 MiB db 0 bytes queue 85 KiB sync  RPC:  0 conn,  0 req/s,   0 µs
2018-08-07 20:22:51  Imported #4050223 0x57a7…9be5 (0 txs, 0.00 Mgas, 4 ms, 0.57 KiB)
2018-08-07 20:22:55  Imported #4050224 0x6b50…3828 (0 txs, 0.00 Mgas, 8 ms, 0.57 KiB)
2018-08-07 20:23:00  Imported #4050225 0x8898…a33b (1 txs, 0.18 Mgas, 8 ms, 0.71 KiB)
2018-08-07 20:23:05  Imported #4050226 0xca15…f9eb (0 txs, 0.00 Mgas, 10 ms, 0.57 KiB)
2018-08-07 20:23:10  Imported #4050227 0x1b17…91f5 (2 txs, 0.40 Mgas, 9 ms, 0.99 KiB)
2018-08-07 20:23:15  Imported #4050228 0x6726…0861 (2 txs, 0.24 Mgas, 4 ms, 1.04 KiB)
2018-08-07 20:23:20  Imported #4050229 0x466d…f469 (3 txs, 0.69 Mgas, 14 ms, 1.10 KiB)
2018-08-07 20:23:22       #8636   15/25 peers   6 MiB chain 87 MiB db 0 bytes queue 1 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
2018-08-07 20:23:26  Imported #4050230 0x4092…dace (4 txs, 0.79 Mgas, 10 ms, 1.41 KiB)
2018-08-07 20:23:32  Imported #4050231 0xd356…9b8f (1 txs, 0.18 Mgas, 6 ms, 0.71 KiB)
2018-08-07 20:23:35  Imported #4050232 0xbd55…a91e (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2018-08-07 20:23:41  Imported #4050233 0xd0de…d6d8 (6 txs, 0.84 Mgas, 4 ms, 1.62 KiB)
2018-08-07 20:23:46  Imported #4050234 0x6b69…b6c5 (1 txs, 0.07 Mgas, 3 ms, 0.80 KiB)
2018-08-07 20:23:51  Imported #4050235 0x51e7…665f (0 txs, 0.00 Mgas, 7 ms, 0.57 KiB)
2018-08-07 20:23:56  Imported #4050236 0x20c4…db14 (2 txs, 0.46 Mgas, 22 ms, 0.93 KiB)
2018-08-07 20:23:57      #13716   15/25 peers   6 MiB chain 87 MiB db 0 bytes queue 1 MiB sync  RPC:  0 conn,  0 req/s,   0 µs
2018-08-07 20:24:00  Imported #4050237 0x4672…a188 (0 txs, 0.00 Mgas, 8 ms, 0.57 KiB)
2018-08-07 20:24:06  Imported #4050238 0x9e69…9455 (3 txs, 0.22 Mgas, 12 ms, 0.93 KiB)
2018-08-07 20:24:10  Imported #4050239 0x884a…2481 (2 txs, 0.04 Mgas, 2 ms, 0.78 KiB)
2018-08-07 20:24:15  Imported #4050240 0x63cc…ab77 (5 txs, 0.51 Mgas, 24 ms, 1.43 KiB)
2018-08-07 20:24:20  Imported #4050241 0x86a3…d6ba (3 txs, 0.21 Mgas, 4 ms, 1.02 KiB)
2018-08-07 20:24:25  Imported #4050242 0xa90c…aa55 (3 txs, 0.27 Mgas, 3 ms, 0.96 KiB)
2018-08-07 20:24:31  Imported #4050243 0x51ee…3def (2 txs, 0.04 Mgas, 3 ms, 0.78 KiB)
```
As you can see it goes `1345587 -> 1375534` and then drops to `8636`. This can happen multiple times, so it's hard to get the verification process to complete. It happens on different block numbers

To test it yourself, please use our [spec.json](https://raw.githubusercontent.com/poanetwork/poa-chain-spec/core/spec.json) and [list of reserved peers](https://raw.githubusercontent.com/poanetwork/poa-chain-spec/core/bootnodes.txt)
```
$ parity --chain spec.json --reserved-peers bootnodes.txt
```

### LINKED ISSUE #9306: Invalid old blocks

ascjones:
Fixes #9300, fixes #7008 and fixes #9306. Rel #9407.

*Edit: there were several issues here*

## TLDR;
- Ancient blocks queue full, download resets, multiple stale responses cause download rounds with 0 blocks imported, retracting to block `#0`
- After retracting to `#0`, can download a non canonical set of blocks 
- It can get stuck on the head of a non canonical chain, when it should retract to find a common block and resume the sync.

### Retraction after queue full

1. `BlockDownloader` in `State::Blocks` [requesting headers](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L452) for subchains: 
2. Ancient blocks [queue gets full](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L516) and the [old blocks download is restarted](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/mod.rs#L641)
3. `BlockDownloader` in `State::ChainHead`. In flight header response from abandoned round arrives, and [resets subchain headers](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L287) to consecutive headers (see the todo comment for a clue!). These headers may be orphaned from their parents which were discarded after the queue became full above.
4. Then we get multiple sync rounds in quick succession which import 0 blocks [because of `UnknownParent`](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L509) after, causing the sync to quickly [retract to 0](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L387). This was also being caused by stale body responses.

The fix is to expire all pending requests when resetting the sync.

### Getting stuck on a non canonical block

1. Some peers have bad blocks in the range `#1..#57`, probably propogated by the above bug. Also around the fork block `#1920000` it was getting stuck on some non canon blocks.
2. The hash gets [stuck](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/blocks.rs#L566) in the list of heads so it never gets [empty](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/blocks.rs#L396) and the round never [completes](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs).

The fix is to reset the sync if we receive consecutive header responses for a requesetd hash with no useful headers that advance the subchain.

### Further improvements

The ancient blocks queue still fills up fairly frequently, resulting in many downloaded blocks being discarded. I will address this in a future PR, by pausing the download of old blocks similar to the normal sync. For now though this no longer triggers the retraction, and the ancient blocks sync now completes.

I also have some further improvements/refactorings for this code stashed away for future PRs. But trying to keep this PR to the minimum that will actually get the ancient block sync working again.


ethcore/sync/src/block_sync.rs
```diff
@@ -515,6 +528,7 @@ impl BlockDownloader {
 				},
 				Err(BlockImportError(BlockImportErrorKind::Queue(QueueErrorKind::Full(limit)), _)) => {
 					debug!(target: "sync", "Block import queue full ({}), restarting sync", limit);
+					bad = true;
```
andresilva: This is done to trigger resetting the sync right? Do you think we should return a different error type other than `BlockDownloaderImportError::Invalid`? (And also handle it in `ChainSync::collect_blocks`).

ascjones: Yes I agree and I had previously done that. However my idea with this initial PR was to do the minimum amount to fix this bug.

However looks like this fix doesn't work entirely so I will add that back while waiting for another mainnet sync...

ascjones:
Unfortunately this does not appear to fix the issue when syncing mainnet. I'm currently investigating.

ethcore/sync/src/block_sync.rs
```diff
@@ -490,50 +519,51 @@ impl BlockDownloader {
...
 				},
 				Err(BlockImportError(BlockImportErrorKind::Block(BlockError::TemporarilyInvalid(_)), _)) => {
-					debug!(target: "sync", "Block temporarily invalid, restarting sync");
+					debug_sync!(self, "Block temporarily invalid: {:?}, restarting sync", h);
 					break;
 				},
 				Err(BlockImportError(BlockImportErrorKind::Queue(QueueErrorKind::Full(limit)), _)) => {
-					debug!(target: "sync", "Block import queue full ({}), restarting sync", limit);
+					debug_sync!(self, "Block import queue full ({}), restarting sync", limit);
+					err = Some(CollectBlocksError::QueueFull);
```
ascjones: Returning an error here in case of the queue being full will trigger the sync to restart

ethcore/sync/src/chain/handler.rs
```diff
@@ -292,7 +292,9 @@ impl SyncHandler {
 		let block_set = sync.peers.get(&peer_id)
 			.and_then(|p| p.block_set)
 			.unwrap_or(BlockSet::NewBlocks);
-		if !sync.reset_peer_asking(peer_id, PeerAsking::BlockBodies) {
+		let allowed = sync.peers.get(&peer_id).map(|p| p.is_allowed()).unwrap_or(false);
+
+		if !sync.reset_peer_asking(peer_id, PeerAsking::BlockBodies) || !allowed {
```
ascjones: Prevents expired body requests being handled, was causing retraction.

ethcore/sync/src/chain/mod.rs
```diff
@@ -762,22 +762,20 @@ impl ChainSync {
 						}
 					}
 
-					// Only ask for old blocks if the peer has a higher difficulty than the last imported old block
-					let last_imported_old_block_difficulty = self.old_blocks.as_mut().and_then(|d| {
-						io.chain().block_total_difficulty(BlockId::Number(d.last_imported_block_number()))
-					});
+					// Only ask for old blocks if the peer has an equal or higher difficulty
+					let equal_or_higher_difficulty = peer_difficulty.map_or(false, |pd| pd >= syncing_difficulty);
```
ascjones: Post fork block we were getting some dodgy blocks from some nodes, so tightening this back up. Still addresses https://github.com/paritytech/parity-ethereum/issues/9225

ethcore/sync/src/chain/mod.rs
```diff
@@ -855,14 +853,29 @@ impl ChainSync {
...
+					if let Err(err) = downloader.collect_blocks(io, false) {
+						trace!(target: "sync", "Error collecting blocks {:?}. Restarting OldBlocks download", err);
+						// reset in flight requests in order to prevent them being handled in the next round
+						for (pid, ref mut p) in &mut self.peers {
+							if p.block_set == Some(BlockSet::OldBlocks) {
+								p.reset_asking();
+								debug!(target: "sync", "Reset peer asking OldBlocks {:?}", pid);
+							}
+						}
+						downloader.reset();
```
ascjones: This fixes the issue of stale responses causing the retraction after reset. Peer reset must happen before download reset.

ethcore/sync/src/block_sync.rs
```diff
@@ -292,22 +312,31 @@ impl BlockDownloader {
...
 			},
 			State::Blocks => {
 				let count = headers.len();
-				// At least one of the heades must advance the subchain. Otherwise they are all useless.
+				// At least one of the headers must advance the subchain. Otherwise they are all useless.
 				if count == 0 || !any_known {
-					trace!(target: "sync", "No useful headers");
+					trace_sync!(self, "No useful headers, expected hash {:?}", expected_hash);
+					if let Some(eh) = expected_hash {
+						if self.blocks.heads_len() > 1 && !self.useless_expected_hashes.insert(eh) {
```
andresilva: I'm not entirely sure I understand this. This is to avoid importing the bad #57 block from which we won't be able to progress. But when we reset the sync it means we'll clear the `useless_expected_hashes`, what's preventing us from getting into the same bad fork again?

ethcore/sync/src/block_sync.rs
```diff
@@ -549,4 +579,127 @@ impl BlockDownloader {
...
+	use ethcore::header::*;
+	use ethcore::client::{TestBlockChainClient};
+	use parking_lot::RwLock;
+	use rlp::*;
+	use std::collections::{VecDeque};
+	use tests::helpers::{TestIo};
+	use tests::snapshot::TestSnapshotService;
+
+	use super::*;
+	// use chain::tests::{
```
andresilva: Remove this?

ethcore/sync/src/chain/handler.rs
```diff
@@ -292,7 +292,9 @@ impl SyncHandler {
 		let block_set = sync.peers.get(&peer_id)
 			.and_then(|p| p.block_set)
 			.unwrap_or(BlockSet::NewBlocks);
-		if !sync.reset_peer_asking(peer_id, PeerAsking::BlockBodies) {
+		let allowed = sync.peers.get(&peer_id).map(|p| p.is_allowed()).unwrap_or(false);
```
andresilva: Note to self: `is_allowed()` will check for require expiration.

ethcore/sync/src/chain/mod.rs
```diff
@@ -855,14 +853,29 @@ impl ChainSync {
...
 					self.restart(io);
 				}
 			},
 			BlockSet::OldBlocks => {
-				if self.old_blocks.as_mut().map_or(false, |downloader| { downloader.collect_blocks(io, false) == Err(DownloaderImportError::Invalid) }) {
-					self.restart(io);
-				} else if self.old_blocks.as_ref().map_or(false, |downloader| { downloader.is_complete() }) {
+				let mut is_complete = false;
+				if let Some(downloader) = self.old_blocks.as_mut() {
+					if let Err(err) = downloader.collect_blocks(io, false) {
```
andresilva: We used to restart on `DownloaderImportError::Invalid`. Do we still want to do this?

ethcore/sync/src/block_sync.rs
ascjones: The idea is that it should eventually find the correct blocks - it might try and reset a couple of times but once it gets a good peer it will move past the bad fork.

I clear `useless_expected_hashes` because I don't want to block further progress on a valid chain. It's possible to receive consecutive useless headers for some other reason other than being on the bad fork. The thinking is to be slightly overzealous on detecting being stuck (i.e. only 2 consecutive useless responses) but allowing it to keep trying again. Again on the assumption that it will retract and find the canon chain eventually. Happy to hear any better ideas though!

ascjones: The check is to prevent unnecessary resets when we're at the head of the chain (i.e. with just one 'subchain' head) with `NewBlocks` sync, which was happening often. I believe this is harmless anyway, but my approach here was to try and reset only when necessary.

The counter would probably work and is slightly simpler, but again might force some unnecessary resets.

ethcore/sync/src/chain/mod.rs
```diff
@@ -855,14 +853,29 @@ impl ChainSync {
...
+						trace!(target: "sync", "Error collecting blocks {:?}. Restarting OldBlocks download", err);
+						// reset in flight requests in order to prevent them being handled in the next round
+						for (pid, ref mut p) in &mut self.peers {
+							if p.block_set == Some(BlockSet::OldBlocks) {
+								p.reset_asking();
```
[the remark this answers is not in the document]
ascjones: Addressed in https://github.com/paritytech/parity-ethereum/pull/9531/commits/48e548359504bed3c8d863263e8b608e5a190826. But have used the `DownloadAction` enum we already have instead of bool.

ethcore/sync/src/chain/mod.rs
ascjones: `DownloaderImportError::Invalid` was the only error instance returned, so it was effectively a bool return. Related: https://github.com/paritytech/parity-ethereum/pull/9531/commits/48e548359504bed3c8d863263e8b608e5a190826

ascjones:
I've tested 4-5 fresh syncs successfully on mainnet, and also 4-5 times on POA network so this looks good.

ddorgan:
@ascjones looks like these test jobs aren't finishing .... does it work for you locally?

ascjones:
@ddorgan looks like my last commit has caused the tests to hang. Looking.

ddorgan:
@ascjones was wrong about that .... looks like the tests are running ok it's just taking a long time.... currently the four jobs at nearly three hours running, the normal test time (from say another build this morning on the same system) is 20 minutes.

ethcore/sync/src/block_sync.rs
jimpo: That makes sense. Can you leave a comment in the code here explaining that (the purpose of the `heads_len() > 1` check?). And perhaps it should check if the expected_hash is the last entry in `heads`, even if `heads_len()` is greater than 0.

If you think tracking useless responses by subchain head is preferable to a single counter, might I suggest changing `useless_expected_hashes` to a `H256FastMap<usize>` which tracks the number of useless responses per hash. That way, the reset threshold can be raised from 2 if necessary.

Regarding the counter,

ethcore/sync/src/block_sync.rs
```diff
@@ -490,53 +513,50 @@ impl BlockDownloader {
...
 
-		if bad {
-			return Err(BlockDownloaderImportError::Invalid);
-		}
-
 		if self.blocks.is_empty() {
 			// complete sync round
-			trace!(target: "sync", "Sync round complete");
+			trace_sync!(self, "Sync round complete");
 			self.reset();
```
jimpo: `reset()` doesn't need to be called here now since it's called in `ChainSync::collect_blocks`.

ethcore/sync/src/chain/mod.rs
```diff
@@ -855,14 +853,29 @@ impl ChainSync {
 	fn collect_blocks(&mut self, io: &mut SyncIo, block_set: BlockSet) {
 		match block_set {
 			BlockSet::NewBlocks => {
-				if self.new_blocks.collect_blocks(io, self.state == SyncState::NewBlocks) == Err(DownloaderImportError::Invalid) {
+				if self.new_blocks.collect_blocks(io, self.state == SyncState::NewBlocks) == DownloadAction::Reset {
+					trace!(target: "sync", "Restarting NewBlocks download");
 					self.restart(io);
```
jimpo: In the case of a successfully completed sync round, this probably shouldn't `restart()`? Why not handle it similar to OldBlocks by expiring peer requests and calling `reset()`?

ethcore/sync/src/chain/mod.rs
```diff
@@ -855,14 +853,29 @@ impl ChainSync {
...
 				}
 			},
 			BlockSet::OldBlocks => {
-				if self.old_blocks.as_mut().map_or(false, |downloader| { downloader.collect_blocks(io, false) == Err(DownloaderImportError::Invalid) }) {
-					self.restart(io);
```
jimpo: Should it restart in the case of a bad old block?

ethcore/sync/src/block_sync.rs
ascjones: I did this but it was causing tests to hang and the sync to break. I need to do some debugging as to why that is, but at the moment it seems that you can't reset the peers after the sync round is complete.

ethcore/sync/src/chain/mod.rs
ascjones: Indeed we shouldn't. I've reverted the change which returns `DownloadAction::Reset` in case of successful round completion. 

I also agree this could be unified. My thinking for this PR was to minimize changes that might break `NewBlocks` sync, and just get something to fix ancient blocks.

ethcore/sync/src/chain/mod.rs
ascjones: Yes and that behaviour is retained.

ascjones:
@jimpo #9654 partially addresses some of the issues here, so we need to coordinate how we merge these. 

But as it is this set of changes does succeed in doing a full mainnet sync, run 7 times now. So I believe this is ready to be merged.

If the other one gets in first I will just rebase off that and remove the 'consecutive useless headers' stuff.

Either way I think we should try and get it in soon since it is currently broken.

ethcore/sync/src/chain/mod.rs
jimpo: Now it resets peer requests and the downloader state instead of calling `restart()`, so it looks like the behavior after an invalid ancient block has changed. But I think the new logic is better (it doesn't mess with the new blocks downloader state).

ascjones: You're right, I misunderstood that comment last night. Yes that was intentional to stop the NewBlocks downloader being reset. I think that similarly the OldBlocks download perhaps shouldn't be reset (via restart) when we get a bad block for NewBlocks. Plenty of improvements still to make here!

ascjones:
@jimpo have implemented your suggestions with the useless counter and the reset logic. Have run it 3 times against mainnet and it manages to get past the corrupt blocks. Also ran a couple of full syncs against POA network and it works okay.

andresilva:
@ascjones Are the changes to the tests submodule expected? (maybe you updated the submodule unknowingly?)

ascjones:
Yes just noticed that too, I must've done that by accident. Fixing.

PR #9537, opened 2018-09-12
ci: always run build pipelines for win, mac, linux, and android

5chdn:
- Runs build pipelines for windows, macos, and android on every PR in addition to the test suite on linux.
- ~~Runs test pipelines in debug mode to speed up things.~~
- Enable tests on master :roll_eyes:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -141,19 +141,23 @@ build-linux-ubuntu-armhf:
     - rust-arm
 
 build-linux-android-armhf:
-  <<:                              *build
+  stage:                           build
   image:                           parity/rust-android:gitlab-ci
   variables:
     CARGO_TARGET:                  armv7-linux-androideabi
+  script:
+    - scripts/gitlab/build-unix.sh
   tags:
     - rust-arm
 
 build-darwin-macos-x86_64:
-  <<:                              *build
+  stage:                           build
   variables:
     CARGO_TARGET:                  x86_64-apple-darwin
     CC:                            gcc
     CXX:                           g++
+  script:
+    - scripts/gitlab/build-unix.sh
   tags:
     - osx
   <<:                              *collect_artifacts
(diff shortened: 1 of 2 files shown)


niklasad1:
This PR should be blocked this until #9510 fixed?

niklasad1:
@5chdn that's cool :+1:

.gitlab-ci.yml
```diff
@@ -104,18 +104,17 @@ test-coverage-kcov:
 
 build-linux-ubuntu-amd64:          &build
   stage:                           build
-  only:                            *releaseable_branches
```
gabreal: In my understanding this means the linux binary is in fact build twice for every branch on github. One for testing and one regular version.

General-Beck: only:                            &releaseable_branches
    - master
    - stable
    - beta
    - tags

General-Beck: This line should be left

5chdn: I'm a bit pedantic here, I want to see build results for all platforms.

Are you saying, the test stage and the build stage are both running linux and therefore duplicate pipelines?

scripts/gitlab/test.sh
```diff
@@ -19,7 +19,7 @@ git submodule update --init --recursive
 rustup show
 if [[ "${RUST_FILES_MODIFIED}" == "0" ]];
 then echo "__________Skipping Rust tests since no Rust files modified__________";
-else ./test.sh || exit $?;
+else ./test.sh --no-release || exit $?;
```
ordian: I thought we were running test with `--release` to speed things up, also some tests may time out like this `has been running for over 60 seconds`?

ordian: It would be nice to see some (compile + run time) numbers and to check if
```rust
[profile.dev]
opt-level = 1
```
is faster for tests.

5chdn: > It would be nice to see some (compile + run time) numbers

Yes that what I was doing last night but fell asleep eventually :laughing: 

Stay tuned for numbers. You are probably right with the timeouts.

5chdn:
@ordian ok, here's the thing: debug mode speeds up compile times but increases test time, release mode increases compile time and speeds up test time. this is a zero-sum game, so I reverted it.

Issue #8746, opened 2018-05-31
up to 30 seconds delay to rpc response 


mikhail-manuilov:
I'm running:

- **Which Parity version?**: 1.10.4
- **Which operating system?**: Linux
- **How installed?**: docker
- **Are you fully synchronized?**: yes
- **Which network are you connected to?**: ethereum
- **Did you try to restart the node?**: yes

I'm running parity with nginx as a reverse tls-proxy on Digital Ocean instance with ssd disks 8GB ram, 4 vcpu.

It's fully synced, but I ran rsync to copy db from local parity.

Running 1800 requests using Chrome. For some reason I got up to 30 seconds delays to some "easy" requests with no shortage on io/ram/cpu as docker stats/ iostat / top tells me.

parity is using ~550 MB running for several days with no limit set in docker.

This request always tops responce time {jsonrpc: "2.0", id: 249, method: "net_version", params: []}.

But production parity v1.9.7 running in docker on VERY slow disks in azure and getting desynced sometimes have much better average and worst responce times with iowait for 20-30%, BUT using 10GB+ memory.

I can test 1.9.7 on DO, but I suppose 10GB is some db cache that makes parity reply much faster, what I am I doing wrong here?

rudolfix:
we have exactly same issue, 4 nodes behind load balancer on Digital Ocean with significant traffic being handled nicely EXCEPT `net_version` method call which takes 10-30 seconds.
Is there a plan to fix it?

ankitarora121:
Facing the same issue, `net_version` takes ~30 seconds. Strange!

debris:
`net_version` was waiting to acquire read lock from networking thread. it does not wait for it any more

PR #9544, opened 2018-09-12
net_version caches network_id to avoid redundant aquire of sync read lock

debris:
issue #8746

This pull request does not resolve the underlying issue with `net_version` rpc method, but it is simply trying to workaround the problem. We should backport it and see if the issue with `net_version` still exists. 

If it does, it is caused by rpc library and we should fix the problem over there.

If it does not, we should fix `sync.status()` locking and/or all other rpc methods that do not respond immediately

### DIFF
--- a/rpc/src/v1/impls/net.rs
+++ b/rpc/src/v1/impls/net.rs
@@ -22,7 +22,12 @@ use v1::traits::Net;
 
 /// Net rpc implementation.
 pub struct NetClient<S: ?Sized> {
-	sync: Arc<S>
+	sync: Arc<S>,
+	/// Cached `network_id`.
+	///
+	/// We cache it to avoid redundant aquire of sync read lock.
+	/// https://github.com/paritytech/parity-ethereum/issues/8746
+	network_id: u64,
 }
 
 impl<S: ?Sized> NetClient<S> where S: SyncProvider {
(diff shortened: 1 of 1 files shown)


rpc/src/v1/impls/net.rs
```diff
@@ -22,25 +22,31 @@ use v1::traits::Net;
...
 
 impl<S: ?Sized> Net for NetClient<S> where S: SyncProvider + 'static {
 	fn version(&self) -> Result<String> {
-		Ok(format!("{}", self.sync.status().network_id).to_owned())
+		Ok(format!("{}", self.network_id))
 	}
 
 	fn peer_count(&self) -> Result<String> {
-		Ok(format!("0x{:x}", self.sync.status().num_peers as u64).to_owned())
+		Ok(format!("0x{:x}", self.sync.status().num_peers as u64))
```
niklasad1: use {:#x} instead of 0x{:x}

tomusdrw: This is probably also pretty slow, we could maybe exposes a specific method to return only number of peers (best would be if it didnt require locking at all) - but it's a separate issue.

debris: @tomusdrw definitely, but let's fix it in a separate pr :)

niklasad1: @tomusdrw 
Just so I understand this, do you mean to return like ```u64``` instead of a String here to get rid of allocation including probably something like ```itoa``` under the hood to convert the integer to ascii/unicode?

debris: I'm pretty sure @tomusdrw was talking about accessing `sync.status()` object which aquires lock :)

tomusdrw: ^^ that @niklasad1 :)

Issue #9490, opened 2018-09-07
Security: Issue: "[websockets]apis" settings are ignored i.e. "parity_accounts" is not getting disabled i.e. it is possible to invoke "parity_newAccountFromPhrase"!


becke-ch-s0-v1:
_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: 1.11.8
- **Operating system**: Linux - Ubuntu 18.04.1 LTS
- **Installation**: Docker - parity/parity:v1.11.8
- **Fully synchronized**: yes
- **Network**: private PoA network - "networkID" : "0x777"
- **Restarted**: yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._
My node configuration:
`/data/becke-ch--parity-node--s0-v1/configuration/node-config.toml`
Has the following value:
```
[network]
bootnodes = ["enode://8aa082af49915716b275cc9ab16e1b2d57a67ab8182490fafb9b0b4b4f9f87eee0731e70f5c42f1f1da3e07893d7ed1f43f5f26d9cdec9d280e49111071195cd@172.17.0.2:30303"]
[rpc]
apis = ["web3", "eth", "pubsub", "net", "shh", "shh_pubsub"]
[websockets]
apis = ["web3", "eth", "pubsub", "net", "shh", "shh_pubsub"]
```
And when I start parity as follows:
`docker run -ti -v /data/becke-ch--parity-node--s0-v1/:/data/becke-ch--parity-node--s0-v1/ parity/parity:v1.11.8 --chain /data/becke-ch--parity-node--s0-v1/configuration/chain-config.json --config /data/becke-ch--parity-node--s0-v1/configuration/node-config.toml --base-path /data/becke-ch--parity-node--s0-v1/data/ --jsonrpc-interface 172.17.0.2 --ws-interface 172.17.0.2`
**I expect that the API `parity_accounts` is disabled i.e. it should not be possible to invoke `parity_newAccountFromPhrase`!**
This is actually true for "rpc" i.e. when I call the following:
`curl --data '{"method":"parity_newAccountFromPhrase","params":["stylus outing overhand dime radial seducing harmless uselessly evasive tastiness eradicate imperfect","hunter2"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST 172.17.0.2:8545`
I'm getting as expected the error:
[...]

PR #9545, opened 2018-09-13
Remove initial token for WS

tomusdrw:
Since the UI has been phased out we can remove the initial token that allows one to connect to "secure" version of WebSockets RPC for the first time without actually generating any tokens.

Related #9490

### DIFF
--- a/rpc/src/tests/ws.rs
+++ b/rpc/src/tests/ws.rs
@@ -136,7 +136,7 @@ mod testing {
 	}
 
 	#[test]
-	fn should_allow_initial_connection_but_only_once() {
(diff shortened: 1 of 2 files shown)


rpc/src/tests/ws.rs
```diff
@@ -136,7 +136,7 @@ mod testing {
 	}
 
 	#[test]
-	fn should_allow_initial_connection_but_only_once() {
+	fn should_not_allow_initial_connection_even_once() {
```
cheme: I like the reference to the old test, but 'should_not_allow_initial_connection' would be simplier. You can ignore this grumble.

PR #9547, opened 2018-09-13
ethereum libfuzzer integration small change

cheme:
This PR, contains very few change to avoid maintaining a parity branch for ethereum libfuzzer project (https://github.com/ethereum/ethereum-vm-fuzzer).

Currently parity fuzzing with libfuzzer requires manual changes (or maintaining their own branch which is also difficult and I see it as kind of a blocker for integrating the libfuzzer operation to others ethereum testing projects).

In this pr we expose the `pod_account` module as  public. I do not believe that it is a big issue (`pod_state` module for instance is already public), I can also reexport 'PodAccount' from pod_state or look for other ways of doing it.

We also add a new test function for `EvmTestClient` : call_envinfo which allows the fuzzer to fuzz on envinfo. Since it only impact `EvmTestClient` (test), I think it is a minor change.


dvdplm:
> Currently parity fuzzing with libfuzzer requires manual changes

Can you elaborate on this? Not sure I understand what this entails.

Not sure if it makes sense, but it'd be easier (for me) to review this if I knew what the testing situation (adding new tests? running the current ones?) looks like now and how it'd improve with this change. Thanks!

cheme:
It's an ethereum project (I realize the linked repo is private) that uses https://llvm.org/docs/LibFuzzer.html on both geth and parity (I think aleth also plugged on it at some point but I am not sure). The fuzzer is mainly targeting the evm interpreter. 

The manual changes are what is in this PR (not much but enough to be a bad constraint). There is also some rust code in the fuzzer project (which need those small changes and expose the C fuzz target), but as a first step removing this constraint will be good.

ethcore/src/client/evm_test_client.rs
```diff
@@ -194,6 +207,7 @@ impl<'a> EvmTestClient<'a> {
 		).map_err(EvmTestError::Evm)
 	}
 
+
```
sorpaas: Extra blank line. Please remove!

5chdn:
please merge master to update pipelines

PR #9550, opened 2018-09-13
ci: fix rpc docs generation, take 2

ordian:
Fixes paritytech/devops/issues/204. Looks like in #9515 I didn't push the latest commit, sorry about that.


5chdn:
Thanks

PR #9552, opened 2018-09-13
ethcore: fix detection of major import

### LINKED ISSUE #9428: When importing new blocks Parity goes into sync mode very briefly, giving false ethstats info
> I'm running:
>
> - **Which Parity version?**: 1.11.8
> - **How installed?**: Docker
> - **Which network are you connected to?**: Ethereum & Ropsten

This issue is only seen in 1.11.8, and not in any of the previous versions.

Upon upgrading to 1.11.8, I noticed that our Parity nodes are reporting large block propagation times occasionally. These block propagation times would always equal the time it taken to get the next block, which raised my suspicion there was an issue.

This can be seen on https://ethstats.net/. If you look at all the Parity 1.11.8 nodes, they mostly all have large spikes in propagation times.

----

Upon investigating, it seemed that all the nodes that were reporting to not have the latest block did actually have the latest block and were reporting the correct block number when ethstats said they weren't.

This caused me to investigate into the [eth-net-intelligence-api](https://github.com/cubedro/eth-net-intelligence-api), and see what could be happening to cause this. Line 720 in the node.js file:
https://github.com/cubedro/eth-net-intelligence-api/blob/master/lib/node.js#L720

There's only one way the service can miss a new block, and that's if a Parity node goes into sync mode and then comes out of it straight away. This is due to the blocknumber only being updated when syncing is actively in progress, if a node comes straight back out of sync then it only then listens for further new blocks.

This aligns with what is being seen in ethstats, with the wrong blocknumber being reported until a further new block comes in.

---

To test this theory, I sat spamming `up + enter` with `eth.syncing` in the Geth JS console:
```
> eth.syncing
false
> eth.syncing
false
> eth.syncing
false
> eth.syncing
{
  currentBlock: 6228721,
  highestBlock: 6228721,
  startingBlock: 6206171,
  warpChunksAmount: null,
  warpChunksProcessed: null
}
> eth.syncing
false
> eth.syncing
false
```

The above output is from when the Parity instance just starts reporting a wrong blocknumber on ethstats. The duration between each of the `eth.syncing` commands is milliseconds, as fast as I can hit `up + enter`.

andresilva:
After a sync round completes successfully we call `Chainsync::reset`, which in turn will set the sync state to the initial state by calling `ChainSync::get_init_state`. If warp sync is enabled the initial state will be set to `WaitingPeers`, which would be incorrect since it would make the client try to start a warp restore and also leads to incorrect detection of major import state. The fix is to force the state to `Idle` after the sync is complete. Also reverted changes introduced in #9112. I tested this by hammering the `eth_syncing` RPC and comparing the results to current master. On current master whenever a new block is imported `eth_syncing` would temporarily return that the node is syncing.

Fixes #9428.

### DIFF
--- a/ethcore/sync/src/chain/mod.rs
+++ b/ethcore/sync/src/chain/mod.rs
@@ -700,6 +700,7 @@ impl ChainSync {
 	fn complete_sync(&mut self, io: &mut SyncIo) {
 		trace!(target: "sync", "Sync complete");
 		self.reset(io);
+		self.state = SyncState::Idle;
 	}
 
 	/// Enter waiting state
--- a/ethcore/sync/src/chain/mod.rs
+++ b/ethcore/sync/src/chain/mod.rs
@@ -507,8 +507,9 @@ impl ChainSync {
 		self.peers.clear();
 	}
 
-	/// Reset sync. Clear all downloaded data but keep the queue
-	fn reset(&mut self, io: &mut SyncIo) {
+	/// Reset sync. Clear all downloaded data but keep the queue.
+	/// Set sync state to the given state or to the initial state if `None` is provided.
+	fn reset(&mut self, io: &mut SyncIo, state: Option<SyncState>) {
 		self.new_blocks.reset();
 		let chain_info = io.chain().chain_info();
 		for (_, ref mut p) in &mut self.peers {
@@ -520,7 +521,7 @@ impl ChainSync {
 				}
 			}
 		}
-		self.state = ChainSync::get_init_state(self.warp_sync, io.chain());
+		self.state = state.unwrap_or_else(|| ChainSync::get_init_state(self.warp_sync, io.chain()));
 		// Reactivate peers only if some progress has been made
 		// since the last sync round of if starting fresh.
 		self.active_peers = self.peers.keys().cloned().collect();
@@ -534,7 +535,7 @@ impl ChainSync {
 			io.snapshot_service().abort_restore();
 		}
 		self.snapshot.clear();
-		self.reset(io);
+		self.reset(io, None);
 		self.continue_sync(io);
 	}
 
@@ -699,8 +700,7 @@ impl ChainSync {
 	/// Called after all blocks have been downloaded
 	fn complete_sync(&mut self, io: &mut SyncIo) {
 		trace!(target: "sync", "Sync complete");
-		self.reset(io);
-		self.state = SyncState::Idle;
+		self.reset(io, Some(SyncState::Idle));
 	}
 
 	/// Enter waiting state


ngotchac:
LGTM, but I wonder if it wouldn't make more sense to pass a `state` to `reset` then, instead of overwriting it instantly @andresilva ?

ethcore/sync/src/chain/mod.rs
```diff
@@ -700,6 +700,7 @@ impl ChainSync {
 	fn complete_sync(&mut self, io: &mut SyncIo) {
 		trace!(target: "sync", "Sync complete");
 		self.reset(io);
+		self.state = SyncState::Idle;
```
tomusdrw: Reset is setting the state to `ChainSync::get_init_state(self.warp_sync, io.chain());` already. Setting it to `Idle` here will break `warp-barrier`.

rpc/src/v1/helpers/block_import.rs
```diff
@@ -19,23 +19,16 @@
 use ethcore::client::BlockQueueInfo;
 use sync::SyncState;
 
-/// Check if client is during major sync or during block import and allows defining whether 'waiting for peers' should
-/// be considered a syncing state.
-pub fn is_major_importing_or_waiting(sync_state: Option<SyncState>, queue_info: BlockQueueInfo, waiting_is_syncing_state: bool) -> bool {
+/// Check if client is during major sync or during block import.
+pub fn is_major_importing(sync_state: Option<SyncState>, queue_info: BlockQueueInfo) -> bool {
```
tomusdrw: I think the flag was introduced quite recently, no? `WaitingForPeers` is the initial state in case we are attempting warp sync, and we don't really want to start regular sync just yet (we first wait for at least couple of warp-peers before falling back to regular sync).

ethcore/sync/src/chain/mod.rs
ngotchac: Oups, you're right @tomusdrw !

andresilva: You're right it does break the warp barrier, but I also don't think the current behavior of setting to `get_init_state` is correct, that's why I forced it to `Idle`. Every time we're synced with the tip of the chain we go into `WaitingPeers` state which tries to trigger a snapshot sync. If you monitor `eth_syncing` it goes into syncing state briefly every time a new block is imported.

andresilva: Does it make sense to set it to `get_init_state` until `WAIT_PEERS_TIMEOUT` is elapsed, and force it to idle afterwards?

ngotchac: I think this line: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/mod.rs#L451 should be modified as you are saying, set to `Idle` when `WAIT_PEERS_TIMEOUT` is elapsed

rpc/src/v1/helpers/block_import.rs
andresilva: Yes it was added recently (#9112), but it's not necessary if we don't switch into `WaitingPeers` state on the beginning of each sync round, and in that case we should always consider `WaitingPeers` to be a syncing state.

5chdn:
Please rebase on master, sorry for the inconvenience.

ethcore/sync/src/chain/mod.rs
andresilva: Okay I was wrong. This doesn't break the warp-barrier, it keeps on searching for the given snapshot until it finds it. It doesn't break the warp barrier because it will be stuck in `WaitingPeers` state and therefore will never complete any sync round (i.e. `complete_sync` won't be called which would force the state to `Idle`). It does break the informant because of the changes I made to `is_major_importing`, it will print as if the node was syncing while it is actually searching for the snapshot. Though I'm not sure what is the correct semantics here? When `eth_syncing` returns `false` should it mean that the node just isn't syncing, or that it's not syncing because it doesn't need to? I.e. it is already synced to the tip of the chain.

ordian: Yeah, I guess if we're in `complete_sync`, then we've already reached warp barrier. We can make `reset` take an additional state param and pass either `ChainSync::get_init_state(self.warp_sync, io.chain())` or `SyncState::Idle`, but I don't see it as blocking this PR.

dvdplm: I think the fix should be moved from `complete_sync()` to `reset()` – it's brittle to have the state first set in `reset()` and then clobbered right afterwards in `complete_sync()`. 
I am new to this code and might be completely wrong here, but it seems like we're conflating state with configuration (`WarpSync::Enabled`). Maybe I'm making the wrong assumption here, but isn't it true that once warp syncing is done we want it turned off until the node is restarted? Or are there cases when a node reaches the tip of the chain at one point t and starts syncing normally and then at t + n need to go back to warp syncing? If not, then `Idle` would be the correct state as soon as `complete_sync` is called?
In other words: if we're beyond the warp barrier or at the tip, calling `complete_sync()` should put us in `Idle` (but I'd move that logic to `reset()` if possible).

ordian: > Or are there cases when a node reaches the tip of the chain at one point t and starts syncing normally and then at t + n need to go back to warp syncing?

Yes, we can switch back to warp syncing e.g. if our node went offline for a while (or just slow) and is too far away (`SNAPSHOT_RESTORE_THRESHOLD`) from the current tip (see [`maybe_start_snapshot_sync`](https://github.com/paritytech/parity-ethereum/blob/6d06c7ad183a363c8c470278fc5f37704af5f789/ethcore/sync/src/chain/mod.rs#L548-L613), which is called regularly on timer).

PR #9557, opened 2018-09-13
Simultaneous platform tests WIP

gabreal:
please see how builds could work for testing in parallel.
macos looks pretty slow.


scripts/gitlab/check-changes.sh
```diff
@@ -0,0 +1,34 @@
...
+
+
+export RUST_FILES_MODIFIED="$(git --no-pager diff --name-only $GIT_COMPARE...$CI_COMMIT_SHA | grep -v -e ^\\. -e ^LICENSE -e ^README.md -e ^test.sh -e ^windows/ -e ^scripts/ -e ^mac/ -e ^nsis/ | wc -l)"
+echo "RUST_FILES_MODIFIED: $RUST_FILES_MODIFIED"
+
+
+if [ "${RUST_FILES_MODIFIED}" = "0" ]
+then
+  echo "__________Skipping Rust tests since no Rust files modified__________";
+  exit 1
```
5chdn: Is there a nicer way to do this? "no files modified" is a valid result and should not result in a failing job

gabreal:
Stages still need to be debugged, prob one could add allow_failure directives for some platforms.

PR #9563, opened 2018-09-14
while working on the platform tests make them non-breaking

gabreal:
This is just a proposal.

During debugging and potentially upgrading macos hardware pr checks might not be verified at all. So if you get bored by that simply merge it and you can have the previous behaviour (except for the long build times).


5chdn:
I unified the test job names and enabled them all but marked them as fallible. What do you think @gabreal @General-Beck

PR #9564, opened 2018-09-15
Increase Gas-floor-target and Gas Cap

eosclassicteam:
go-ethereum already increased their default gastarget and gaslimit to 8M see https://github.com/ethereum/go-ethereum/pull/17546/files#diff-21609b3ea13b7a6695113bee0fd055acR51 and https://github.com/ethereum/go-ethereum/releases/tag/v1.8.15


5chdn:
Test fails

```
    configuration::tests::test_run_cmd
```

PR #9570, opened 2018-09-17
Add constantinople conf to EvmTestClient.

cheme:
Fixes 9568.
Constantinople spec was not resolve when using EvmTestClient. This adds it.

### DIFF
--- a/ethcore/src/client/evm_test_client.rs
+++ b/ethcore/src/client/evm_test_client.rs
@@ -86,9 +86,9 @@ impl<'a> EvmTestClient<'a> {
 			ForkSpec::EIP150 => Some(ethereum::new_eip150_test()),
 			ForkSpec::EIP158 => Some(ethereum::new_eip161_test()),
 			ForkSpec::Byzantium => Some(ethereum::new_byzantium_test()),
+			ForkSpec::Constantinople => Some(ethereum::new_constantinople_test()),
 			ForkSpec::EIP158ToByzantiumAt5 => Some(ethereum::new_transition_test()),
 			ForkSpec::FrontierToHomesteadAt5 | ForkSpec::HomesteadToDaoAt5 | ForkSpec::HomesteadToEIP150At5 => None,
-			_ => None,
 		}
 	}
 


debris:
lgtm, but some of the constantinople json tests are failing @sorpaas do you know what's missing / should be fixed?

cheme:
I did not see the failing test, those are partly related to outdated ethtest submodule reference (last time I checked the test did not include some eip and I had to go back to some old block reward value and others changes to make them pass). I think I will try to update the test or at least makes change required to avoid pass CI.

sorpaas:
Tried to look into this via `CreateOOGafterInitCodeReturndata2_d0g0v0`, and it turned out to be a `blockReward` error. It can be fixed by changing the Constantinople block reward transition to a block other than zero. I suspect many of the other test failures are also due to this.

cheme:
From my tests many error pass with old byzantium conf instead. I also got too many error to consider fixing them under this PR.
I started to put a filter to skip tests that are considered 'under issue' : this is to help keeping our state tests up to date. I believe it is better to have the state test close to latest commit from 'ethereum/tests' and having identified failing one than keeping an old branch for the submodule.
Skipped test got a 'reference' field (in the json conf) that must refer to a pending issue (in our case a hub issue needs to be created to update all those field).
I see two categories of 'under issue' test : 
- those that requires a fix from parity 
- those that requires stabilization from the submodule (currently there is other issue with geth or aleth).

sorpaas:
Indeed, right now it looks like we don't have any `skip` method for `jsontests`. It seems that we just passed all tests for `ethereum/tests` in the past?

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                        "reference": 9568,
+                        "failing": "stCreateTest",
+                        "subtests": [
+                                "CreateOOGafterInitCodeRevert_d0g0v0_Constantinople",
+                                "CreateOOGafterInitCodeReturndata3_d0g0v0_Constantinople",
+                                "CreateOOGafterInitCodeRevert2_d0g0v0_Constantinople",
+                                "CreateOOGafterInitCode_d0g1v0_Constantinople",
+                                "CreateOOGafterInitCodeReturndataSize_d0g0v0_Constantinople",
+                                "CreateOOGafterInitCode_d0g0v0_Constantinople",
+                                "CreateOOGafterInitCodeReturndata2_d0g0v0_Constantinople",
```
sorpaas: Please cherry-pick https://github.com/paritytech/parity-ethereum/pull/9577/commits/544f6a3d0c9fa61f626b8bb3d137f7c5039d7af7
I think the only one left here right now should be `CreateOOGafterInitCodeReturndata2`.

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                                "modexp_modsize0_returndatasize_d0g1v0_Constantinople",
+                                "modexp_modsize0_returndatasize_d0g2v0_Constantinople",
+                                "modexp_modsize0_returndatasize_d0g3v0_Constantinople"
+                        ]
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stRevertTest",
+                        "subtests": [
+                                "RevertOnEmptyStack_d0g0v0_Constantinople"
```
sorpaas: Looks like this is fixed? Right now what I get from `GeneralStateTest_stRevertTest` is only `["LoopCallsDepthThenRevert2"]`.

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                        ]
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stSpecialTest",
+                        "subtests": [
+                                "FailedCreateRevertsDeletion_d0g0v0_Constantinople",
```
sorpaas: This one doesn't appear any more with https://github.com/paritytech/parity-ethereum/pull/9577/commits/544f6a3d0c9fa61f626b8bb3d137f7c5039d7af7

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                                "push32withoutByte_d0g0v0_Constantinople"
+                        ]
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stSystemOperationsTest",
```
sorpaas: This can be removed as well with https://github.com/paritytech/parity-ethereum/pull/9577/commits/544f6a3d0c9fa61f626b8bb3d137f7c5039d7af7

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                                },
+                                "RevertDepthCreateAddressCollision": {
+                                        "subnumbers": ["3","4"],
+                                        "chain": "Constantinople (test)"
+                                }
+                        }
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stStaticCall",
```
sorpaas: Those are passing except `static_checkOpcodes2`.

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                                        "subnumbers": ["5","6","7","8"],
+                                        "chain": "Constantinople (test)"
+                                }
+                        }
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stZeroKnowledge",
+                        "subtests": {
+                                "pointAddTrunc": {
```
sorpaas: Looks like those are passing except `pointAdd`!

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
+{ "block":
+        [
+
+                {
+                        "reference": 9568,
```
sorpaas: Can you change this type to a string? Sometimes it may be easier to reference it via `ethereum/tests`'s issue number.

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,215 @@
...
+                },
+                {
+                        "reference": 9568,
+                        "failing": "stZeroKnowledge",
+                        "subtests": {
+                                "pointAddTrunc": {
+                                        "subnumbers": ["*"],
+                                        "chain": "Constantinople (test)"
+                                },
+                                "pointAdd": {
```
sorpaas: This one is indeed `ethereum/tests`'s issue: https://github.com/ethereum/tests/issues/512

cheme:
With #9440 we got test passing up to the date of august, but obviously without constantinople test.
Clearly the synch with ethereum/tests is not really smooth (I need to check but there might still be some tests that are not included), and having a bit of flexibility by skipping some test that are in an ongoing issue could be good.

There is currently another way to skip tests : using 'skip => ' as first parameter of 'declare_test' macro, but it is not currently use, and it is truly a skip feature.

It is important that this 'skip' is limitted to our CI and that the feature is not include in tools such as 'hive'.

ethcore/res/ethereum/tests-issues/currents.json
cheme: There is something not so good with the way the test are currently running : having a first failure is hiding other tests failure (it reports only one failure but if I skip it, it shows the other ones). So the other test in this branch are still failing for me.

holiman:
> and having a bit of flexibility by skipping some test that are in an ongoing issue could be good.

We have that in geth too. 

> It is important that this 'skip' is limitted to our CI and that the feature is not include in tools such as 'hive'. 

No worries, that's impossible. Hive runs it black-box, so the node doesn't even know it's being tested :)

sorpaas:
@cheme Do you think it would be possible to extend the `skip =>` macro for all test skipping functionalities? I think that may be a good thing to do for a future PR.

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,190 @@
...
+                                        "chain": "Constantinople (test)"
+                                }
+ 
+                        }
+                },
+                {
+                        "reference": "https://github.com/ethereum/tests/issues/512",
+                        "failing": "stZeroKnowledge",
+                        "subtests": {
+                                "pointAddTrunc": {
```
sorpaas: I think those one are all actually passing except `pointAdd`? @cheme Do you mind to check this?

cheme: Not passing :( , there is something bad with our current code : if you run the test without skipping anything it fails on only one test, but as soon as you skip it, it reveals others failure.

cheme:
@sorpaas , extending the test macro to skip the tests seems possible, but I fear that it will somehow make those skips less visible. I like having a json (I do not say that its current schema is right) to keep trace of those skips because it is accessible quickly to anyone (non rust developpers included).

ethcore/res/ethereum/tests-issues/currents.json
sorpaas: Got it! Can you also cherry-pick https://github.com/paritytech/parity-ethereum/pull/9577/commits/16f661a59518bf0d949b0b15ea0191fa4df5ba95? Our previous definition of `modexp` and `bn128_mul` has wrong gas price. I think it may fix some other tests.

sorpaas:
@cheme Yeah that's certainly a trade-off! I just think having a global variable doesn't really look nice for the long-term (we also need to look into the `declare_test!` macros to know which folders are being ran!) -- and IMHO we really should aim to only use skip tests for the short-term. :)

Anyway, regarding current skip tests, if we add something like below somewhere in the code, it should avoid the situation where someone accidentally enabled `ci-skip-tests` for non-testing build. This may be important because those jsontests functions are also used in `evmbin`:

```
#[cfg(all(not(test), feature = "ci-skip-tests"))]
compile_error!("ci-skip-tests can only be enabled for testing builds.")
```

ethcore/res/ethereum/tests-issues/currents.json
cheme: Indeed, it does reduce the number of failing zk tests. Many still don't pass, but it is due to https://github.com/ethereum/tests/issues/512 (with 1283 disabled everything is fine).

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -0,0 +1,374 @@
+{ "block":
```
sorpaas: Can you change this file to use tabs for indentation.

sorpaas:
https://github.com/ethereum/tests/pull/511 should be the one that fixes most of those test failures.

PR #9578, opened 2018-09-17
docs(rpc): push the branch along with tags

ordian:


ordian:
I believe `scripts/gitlab-rpc-docs.sh` is not used anymore, so it was removed.

PR #9589, opened 2018-09-19
Remove unused dependencies in `ethcore-sync`

niklasad1:
Remove unused dependencies and move `rustc-hex` as `test-dependency` because it is
only used for tests


dvdplm:
Late to the party here, but consider switching to `rustc-hex` 2.0 or, when possible, use `hex-literal` (e.g. in tests).

Issue #6155, opened 2017-07-26
Differentiate between manual throw and natural gas exhaustion


rphmeier:
In the light client, when executing `eth_call` or `eth_estimateGas`, we start with a small amount of gas and double it each time the result is exceptional, until the block gas limit is reached. Being able to differentiate between the request actually running out of gas and a deliberate throw would let us avoid making many redundant requests.

rphmeier:
I think this is already handled but we don't have a test for it. We should just try the `rpc::light_helpers::execute_tx` with a proof of some call that throws. Probably easier said than done -- setting up the contract, state, and state proof necessary for that will probably be a nuisance.

folsen:
@tomaka Anything you'd be interested in picking up?

tomaka:
I can pick this up in the next few weeks.

PR #9591, opened 2018-09-19
fix (light/provider) : Make `read_only executions` read-only

### LINKED ISSUE #9535: [light client] RPC `eth_estimateGas` don't behave as in `full client`
### What is the problem

In the context of a light-client, for example, we want to use the `eth_estimateGas RPC` for an `unknown account` and this doesn't work currently because the `ExecutionProof` always falls and this don't happen in `full client` implementation. 

```bash
2018-09-12 10:54:02   TRACE rpc  Request: {"method":"eth_estimateGas","params":[{"from":"acc", "gas":"0xffff", "gasPrice":"0xffff", "nonce":"0x0"}, "latest"],"id":1,"jsonrpc":"2.0"}.           
2018-09-12 10:54:02   DEBUG on_demand  Execution request, Incomplete { block_hash: Scalar(0x97db89353cc3abb9d51a6690338cb66a2f10dd58f350394e9f1f980201320b2c), from: acc, action: Create, gas: 65535, gas_price: 65535, value: 0, data: [] }
execute_tx finished wait on the future
2018-09-12 10:54:04  IO Worker #1 DEBUG on_demand  #1 requests have not been dispatched
2018-09-12 10:54:04  IO Worker #1 DEBUG on_demand  ReqID: Request #11
2018-09-12 10:54:04  IO Worker #1 DEBUG on_demand  Request Batch { outputs: {}, requests: [Execution(Incomplete { block_hash: Scalar(0x97db89353cc3abb9d51a6690338cb66a2f10dd58f350394e9f1f980201320b2c), from: x, action: Create, gas: 65535, gas_price: 65535, value: 0, data: [] })], answered: 0 }
2018-09-12 10:54:07  IO Worker #2 DEBUG on_demand  Execution request completed Incomplete { block_hash: Scalar(0x97db89353cc3abb9d51a6690338cb66a2f10dd58f350394e9f1f980201320b2c), from: account, action: Create, gas: 65535, gas_price: 65535, value: 0, data: [] }
2018-09-12 10:54:07  IO Worker #2 DEBUG on_demand  BadExecution Proof
2018-09-12 10:54:07  IO Worker #2 DEBUG on_demand  Peer 43 gave bad response: Validity(BadProof)
2018-09-12 10:54:07  IO Worker #2 DEBUG on_demand  BadExecution Proof
2018-09-12 10:54:07  IO Worker #2 DEBUG on_demand  Peer 36 gave bad response: Validity(BadProof)
...
and so on!
```
<sup>1</sup> The request batch log is not part of parity, I manually added it which prints out the number of requests and indiviudal request which have not been anwered yet!

### Educated Guess

I'm not exactly sure the reason why but I think it is because the actual execution in `full client` seems to boost the actual account balance to execute the transaction successfully (transact_virtual will provide enough funds), https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/client/client.rs#L1473-#L1545. 

In the light client `validation` (executed by one or more full nodes) we use are using the actual account balance for validation and should be changed to do same as the `full client` (boost the account balance)!

However, in context as a validator, I'm not sure how to distinguish between different executions and it might be better to leave it as it is!

Related files:

- https://github.com/paritytech/parity-ethereum/blob/master/ethcore/light/src/provider.rs#L267-#L286 (call to provider `full node`)
- https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/client/client.rs#L2346-#L2364
- https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/state/mod.rs#L230-#L263

### LINKED ISSUE #9551: [Light client] Disconnect all its peers subsequently to an eth_call
- **Parity Ethereum version**:v2.2.0-unstable-6b39131-20180913
- **Operating system**: Linux
- **Installation**:  built from source
- **Fully synchronized**: yes
- **Network**:  kovan
- **Restarted**: yes

How to reproduce :
Launching parity with:
`parity --light --chain kovan -l on_demand,rpc`
and calling the bellow RPC a coupe times (it hapened after 4-5 times), it ends up in a state where all peers have been disconnected.
edit: I sometimes can't reproduce it even after ~20 calls

```
$ curl --data '{"method":"eth_call","params":[{"to":"0x4a6e6c3868a279e1d9047b42c3fb356ff4680003","data":"0x70a08231000000000000000000000000004702bdcc3c7dbffd943136107e70b827028600"},"latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","result":"0x0000000000000000000000000000000000000000019d944653184202a34c0000","id":1}
```
After getting the above result (all good), I get:
```
$ curl --data '{"method":"eth_call","params":[{"to":"0x4a6e6c3868a279e1d9047b42c3fb356ff4680003","data":"0x70a08231000000000000000000000000004702bdcc3c7dbffd943136107e70b827028600"},"latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","error":{"code":-32065,"message":"Timeout for On-demand query; 1 query attempts remain for query #0"},"id":1}
```

This makes the light client disconnect to all its peers (from originally 5 to 0), [here are the full logs with `rpc` and `on_demand` traces](https://gist.github.com/Tbaut/8b8ea8e3c7ae733aab6f499836eb6849)

niklasad1:
Attempt to close #9551 and fix #9535 

/cc @Tbaut can you try this with `fether`? 

This changes so all `ExecutionRequests` from light-clients are executed
as read-only which the `virtual flag` == true ensures. 

This boost up the current transaction to always succeed such as account balance and similar!

Note, this only affects `eth_estimateGas` and `eth_call` AFAIK.

## Manual tests on (kovan)

1. Started a `provider` (full node)  
2. Request from light client with unknown account ```0x0066Dc48bb833d2B59f730F33952B3c29fE926ff```

```bash
curl --data '{"method":"eth_estimateGas","params":[{"from":"0x0066Dc48bb833d2B59f730F33952B3c29fE926ff", "gas":"0xffff" }, "latest"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8555
{"jsonrpc":"2.0","result":"0xcf08","id":1}
```

### DIFF
--- a/ethcore/light/src/on_demand/request.rs
+++ b/ethcore/light/src/on_demand/request.rs
@@ -61,7 +61,7 @@ pub enum Request {
 	/// A request for a contract's code.
 	Code(Code),
 	/// A request for proof of execution.
-	Execution(TransactionProof),
+	Execution(ReadOnlyTransactionProof),
 	/// A request for epoch change signal.
 	Signal(Signal),
 }
@@ -145,7 +145,7 @@ impl_single!(Receipts, BlockReceipts, Vec<Receipt>);
 impl_single!(Body, Body, encoded::Block);
 impl_single!(Account, Account, Option<BasicAccount>);
 impl_single!(Code, Code, Bytes);
-impl_single!(Execution, TransactionProof, super::ExecutionResult);
+impl_single!(Execution, ReadOnlyTransactionProof, super::ExecutionResult);
 impl_single!(Signal, Signal, Vec<u8>);
 
 macro_rules! impl_args {
@@ -256,7 +256,7 @@ pub enum CheckedRequest {
 	Body(Body, net_request::IncompleteBodyRequest),
 	Account(Account, net_request::IncompleteAccountRequest),
 	Code(Code, net_request::IncompleteCodeRequest),
-	Execution(TransactionProof, net_request::IncompleteExecutionRequest),
+	Execution(ReadOnlyTransactionProof, net_request::IncompleteExecutionRequest),
 	Signal(Signal, net_request::IncompleteSignalRequest)
 }
 
@@ -1030,7 +1030,7 @@ impl Code {
 
 /// Request for transaction execution, along with the parts necessary to verify the proof.
 #[derive(Clone)]
-pub struct TransactionProof {
+pub struct ReadOnlyTransactionProof {
 	/// The transaction to request proof of.
 	pub tx: SignedTransaction,
 	/// Block header.
@@ -1042,7 +1042,7 @@ pub struct TransactionProof {
 	pub engine: Arc<EthEngine>,
 }
 
-impl TransactionProof {
+impl ReadOnlyTransactionProof {
 	/// Check the proof, returning the proved execution or indicate that the proof was bad.
 	pub fn check_response(&self, _: &Mutex<::cache::Cache>, state_items: &[DBValue]) -> Result<super::ExecutionResult, Error> {
 		let root = self.header.as_ref()?.state_root();
--- a/ethcore/light/src/on_demand/request.rs
+++ b/ethcore/light/src/on_demand/request.rs
@@ -61,7 +61,7 @@ pub enum Request {
 	/// A request for a contract's code.
 	Code(Code),
 	/// A request for proof of execution.
-	Execution(ReadOnlyTransactionProof),
+	Execution(TransactionProof),
 	/// A request for epoch change signal.
 	Signal(Signal),
 }
@@ -145,7 +145,7 @@ impl_single!(Receipts, BlockReceipts, Vec<Receipt>);
 impl_single!(Body, Body, encoded::Block);
 impl_single!(Account, Account, Option<BasicAccount>);
 impl_single!(Code, Code, Bytes);
-impl_single!(Execution, ReadOnlyTransactionProof, super::ExecutionResult);
+impl_single!(Execution, TransactionProof, super::ExecutionResult);
 impl_single!(Signal, Signal, Vec<u8>);
 
 macro_rules! impl_args {
@@ -256,7 +256,7 @@ pub enum CheckedRequest {
 	Body(Body, net_request::IncompleteBodyRequest),
 	Account(Account, net_request::IncompleteAccountRequest),
 	Code(Code, net_request::IncompleteCodeRequest),
-	Execution(ReadOnlyTransactionProof, net_request::IncompleteExecutionRequest),
+	Execution(TransactionProof, net_request::IncompleteExecutionRequest),
 	Signal(Signal, net_request::IncompleteSignalRequest)
 }
 
@@ -1030,7 +1030,7 @@ impl Code {
 
 /// Request for transaction execution, along with the parts necessary to verify the proof.
 #[derive(Clone)]
-pub struct ReadOnlyTransactionProof {
+pub struct TransactionProof {
 	/// The transaction to request proof of.
 	pub tx: SignedTransaction,
 	/// Block header.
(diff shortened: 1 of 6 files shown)


ethcore/src/client/client.rs
```diff
@@ -2357,7 +2358,7 @@ impl ProvingBlockChainClient for Client {
 			self.engine.machine(),
 			&env_info,
 			self.factories.clone(),
-			false,
+			true,
```
niklasad1: This will now use [transact_virtual](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/executive.rs#L227l) to ensure that the account has enough balance for the execution

ethcore/light/src/on_demand/request.rs
```diff
@@ -61,7 +61,7 @@ pub enum Request {
 	/// A request for a contract's code.
 	Code(Code),
 	/// A request for proof of execution.
-	Execution(TransactionProof),
+	Execution(ReadOnlyTransactionProof),
```
niklasad1: More explicit name

Tbaut:
Thanks for tackling that @niklasad1 
I gave it a shot and unfortunately this did not succeed. Here is a trace with the RPCs: https://gist.github.com/Tbaut/f13f6c26e80d20227cb46d5c3765daab

ordian:
@Tbaut note, that it fixes full node's side of light client, so to properly test this, you need to connect light client to a patched full node (reserved and reserved-only connection is broken on master)

Tbaut:
Thanks @ordian, I definitely did not test it (and I couldn't have had indeed :( ).

niklasad1:
@Tbaut @ordian 

I brute-forced this and spinned up a regular full-node (not reserved) along with a light client and it works (eventually found) my full node!

From your logs it looks like ```eth_call``` is the problem which this PR fixes.
Also, ```eth_getBalance``` will return as long as it gets at least one reply from its peers regardless whether the account exists or not!

Tbaut:
FWIW, I tested Fether about 1-2 months ago by connecting it directly to a light client that was connected to a full node using reserved peers, and I had fewer problems than when the light client was connecting to nodes in the wild.

ethcore/src/client/client.rs
ordian: What do you think about removing `virt` parameter from `state::prove_transaction` (always set to true) and add docs to `prove_transaction` trait method?

ethcore/light/src/on_demand/request.rs
niklasad1: Ok, I will revert this!

cheme:
I am starting to wonder if it is really fine to run transaction as virtual in this case. For instance if I want to send a transaction and do not put enough gas maybe I would rather have a failure than a success (success leading me to send my transaction and lose my gas).

What I am saying, is that we probably want light client 'eth_estimateGas' call to run as virtual, but 'eth_call' to run as normal. With this PR it always run as virtual.
I am also not sure with the renaming (for the same reason as stated before).

I need confirmation that I am incorrect before approving, maybe @rphmeier .

Even if I am correct, the problem remains: in 'rpc/src/v1/helpers/light_fetch.rs' both `call` and `estimategas` run on the same rpc call (`proved_execution`) and we probably want to run virtual only for  `estimateGas` which does not seems possible with current LES definition. So the change from this PR will make `estimateGas` works for light client but might break `call` for light client. I do not really see a good solution that does not required changing LES a bit (if changing LES there is also certainly something to do about #6155 ).

cheme:
Indeed, so it is the same behavior, must be correct.

PR #9602, opened 2018-09-20
Update ropsten bootnodes

eosclassicteam:
AWS instance moved to m5.large for better capicity 🍾 

If anyone is planning to host ethereum node on AWS please use m5 instead of t2 or m4


5chdn:
Couldn't you also migrate the IP address?

eosclassicteam:
@5chdn Yes it was impossible to maintain old ip address since AWS doesn't support it ( Using Elastic ip would help but I am not considering to use it since there are limitations for amount per account ) 

Update complete for the following PR

https://github.com/ethereum/go-ethereum/pull/17682
https://github.com/ethereum/ropsten/pull/32

PR #9603, opened 2018-09-20
Use static call and apparent value transfer for block reward contract code

sorpaas:
Make block reward contract code to use static call and apparent value transfer, so that system address will never be touched by it.


andresilva:
We need to document that block reward contract deployed with `blockRewardContractCode` must be pure.

5chdn:
Do we want to backport this?

sorpaas:
This is a potentially breaking change (but also a bugfix that makes block reward contract code work!). So I think we may need to backport this, but it's not urgent. Sorry I missed the backport tag!

PR #9572, opened 2018-09-17
[beta] Revert to old parity-tokio-ipc.

tomusdrw:

PR #9561, opened 2018-09-14
deps: cargo update-p jsonrpc-core

ordian:
Thanks to paritytech/jsonrpc/pull/311 it should fix compilation on windows.

PR #9604, opened 2018-09-20
Revert to old parity-tokio-ipc.

cheme:
checking if changes from #9572 can be applied to master (currently windows non passing CI block all PR)

### DIFF
--- a/ethcore/light/src/net/request_credits.rs
+++ b/ethcore/light/src/net/request_credits.rs
@@ -411,6 +411,13 @@ mod tests {
 		assert_eq!(costs, new_costs);
 	}
 
+	/// additional time delay to avoid non deterministic CI breakage on some systems
+	#[cfg(target_os = "windows")]
+	const ACCURATE_SYSTEM_TIME : bool = false;
+	#[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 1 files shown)


cheme:
yes and as in #9561 it is totally unclear which test did fail.

cheme:
Under this state (a few test disable), test.sh run correctly in my windows 10 VM. Something else to notice about windows CI 
```
/c/r/builds/effe9fc5/0/parity/parity-ethereum/test.sh: line 50: make: command not found
```
That ends the `Running the C++ example`, but it does not seem to be the root cause of the CI failure.

cheme:
This does compile, but the main issue after that is that some test do not pass. I spotted a first category of failing test (typically related to wrong path translation on windows), and I had to deactivate them for the test case to run on my windows 10 VM. Yet on the windows 7 vm of the CI there seems to be other errors, but the main problem is that they are not accessible (not in the log : maybe some log flushing does not happen).

ordian:
@cheme I think @debris is referring to unrelated changes to `Cargo.lock`

cheme:
Oh, of course, sorry. (using windows makes me grumpy)

5chdn:
Tests fail

PR #9605, opened 2018-09-20
ethcore: mcip3 block reward contract

andresilva:
Merge only after #9603.

Implements the Musicoin block reward logic in a contract using `blockRewardContractCode`. The contract and instructions on how to build can be found at https://gist.github.com/andresilva/6f2afaf9486732a0797f4bdeae018ee9. I have successfully synced the musicoin chain using this contract but would appreciate it if someone else could try as well.


eosclassicteam:
@sorpaas @andresilva Hello may I ask it is possible to implement this contract with block height counts? We have two hardfork exists and we need to change the block reward at the specific height

ethcore/res/ethereum/musicoin.json
```diff
@@ -10,12 +10,8 @@
 				"homesteadTransition":"0x118c30",
 				"eip100bTransition":"0x21e88e",
 				"blockReward":"0x1105a0185b50a80000",
-				"mcip3Transition":"0x124f81",
-				"mcip3MinerReward":"0xd8d726b7177a80000",
-				"mcip3UbiReward":"0x2b5e3af16b1880000",
-				"mcip3UbiContract":"0x00efdd5883ec628983e9063c7d969fe268bbf310",
-				"mcip3DevReward":"0xc249fdd327780000",
-				"mcip3DevContract":"0x00756cf8159095948496617f5fb17ed95059f536",
+				"blockRewardContractTransition": "0x124f81",
```
eosclassicteam: It will be great when we have transition height inside contract

andresilva:
@eosclassicteam The contract has access to the block number so the transitions should be implemented inside the contract.

```
if (block.number > 1000) {
  // new reward logic
} else {
  ...
}
```

eosclassicteam:
@andresilva Thanks, also I would like to suggest changing ECIP-1017 to contract also 😄

ascjones:
@andresilva I think you need to merge latest CI changes

andresilva:
I added a comment pointing to the source of the block reward contract. I also removed the `mcip6_byz` chain spec which seemed to be unused.

PR #9611, opened 2018-09-20
rpc-docs should push the branch

gabreal:


ordian:
@debris 
Currently we push only tags to wiki, see https://github.com/paritytech/wiki/commits/v2.0.5.
Pushing branches along with tags is no worse in terms of conflicts.

5chdn:
CI_COMMIT_REF_NAME is always the tag and tags are unique, so in the rare case someone force-tags this, it's ok if this pipeline fails. @debris

PR #9613, opened 2018-09-20
stable(rpc-docs): avoid recursive copy

ordian:
`cp -r $PROJECT_DIR jsonrpc/.parity` tries to copy `.` into subdir, which is not nice.

### DIFF
--- a/scripts/gitlab-rpc-docs.sh
+++ b/scripts/gitlab-rpc-docs.sh
@@ -41,12 +41,13 @@ upload_files() {
     git push --tags
 }
 
-PROJECT_DIR=$(pwd)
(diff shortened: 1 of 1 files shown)


scripts/gitlab-rpc-docs.sh
```diff
@@ -41,12 +41,13 @@ upload_files() {
     git push --tags
 }
 
-PROJECT_DIR=$(pwd)
+RPC_TRAITS_DIR="rpc/src/v1/traits"
 
 setup_git
 cd ..
```
ordian: This `cd ..` is not present in master, not sure what it's doing here.

5chdn: the cd does not make any sense

Issue #8675, opened 2018-05-21
eth_estimateGas - Requires higher than upper limit of X


edevil:
_Before filing a new issue, please **provide the following information**._

> I'm running:
>
> - **Which Parity version?**: 1.9.7 AND 1.10.4
> - **Which operating system?**: Linux
> - **How installed?**: Docker
> - **Are you fully synchronized?**: yes (private chain from scratch)
> - **Which network are you connected to?**: private PoA chain
> - **Did you try to restart the node?**: yes

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

I have deployed a contract based on OpenZeppelin's MintableToken and HasNoEther, and I can mint tokens with no problems using a Ganache test node. However, when I do the same against a parity node I get an error when calling eth_estimateGas for a simple mint operation:

**actual**

```
POST / HTTP/1.1
Host: 127.0.0.1:8545
User-Agent: Web3.py/4.2.1/<class 'web3.providers.rpc.HTTPProvider'>
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Type: application/json
Content-Length: 273

{"jsonrpc": "2.0", "method": "eth_estimateGas", "params": [{"to": "0xafF2e125054693758CB14fD7Ee03c4a8800aaAAA", "data": "0x40c10f19000000000000000000000000c33ced9f1d89e51abee70840303c403e5950925b000000000000000000000000000000000000000000000000000000000000000a"}], "id": 19}

HTTP/1.1 200 OK
Content-Type: application/json
Transfer-Encoding: chunked
Date: Mon, 21 May 2018 15:37:34 GMT

A2
{"jsonrpc":"2.0","error":{"code":-32015,"message":"Transaction execution error.","data":"Internal(\"Requires higher than upper limit of 59591043530\")"},"id":19}

0
```

**expected (Ganache)**
```
POST / HTTP/1.1
Host: 127.0.0.1:7545
User-Agent: Web3.py/4.2.1/<class 'web3.providers.rpc.HTTPProvider'>
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Type: application/json
Content-Length: 272

[...]

PR #9615, opened 2018-09-20
ethcore: handle vm exception when estimating gas

andresilva:
Could help with #8675.


debris:
I scheduled `retry` on darwin machine, but the job is still pending

ordian:
Shall we backport this to stable and beta?

PR #9620, opened 2018-09-21
ethcore/VerificationQueue don't spawn up extra `worker-threads` when explictly specified not to

niklasad1:
In the verification queue we spawn up maximum worker threads to do the work and don't honor if `num-verifiers` i.e, we still spawn the maximum number of threads which consume extra memory. Thus, this PR changes that!

However, there is one catch though when `--scale-verifiers` is specified then
we must spawn all the threads because all threads are created upon initialization AFAIK and eventually might be scaled (this could probably be fixed but out-of-scope for this PR)

OT:
In my opinion, it doesn't make sense to use both `num-verifiers` and
`scale-verifiers` they are kind of contradictory!

## Examples how this works
See added test cases

### DIFF
--- a/ethcore/src/verification/queue/mod.rs
+++ b/ethcore/src/verification/queue/mod.rs
@@ -39,9 +39,6 @@ pub mod kind;
 const MIN_MEM_LIMIT: usize = 16384;
 const MIN_QUEUE_LIMIT: usize = 512;
 
-// maximum possible number of verification threads.
-const MAX_VERIFIERS: usize = 8;
-
 /// Type alias for block queue convenience.
 pub type BlockQueue = VerificationQueue<self::kind::Blocks>;
 
@@ -85,7 +82,7 @@ impl Default for VerifierSettings {
 	fn default() -> Self {
 		VerifierSettings {
 			scale_verifiers: false,
-			num_verifiers: MAX_VERIFIERS,
+			num_verifiers: ::num_cpus::get(),
 		}
 	}
 }
@@ -231,9 +228,7 @@ impl<K: Kind> VerificationQueue<K> {
 		let empty = Arc::new(Condvar::new());
 		let scale_verifiers = config.verifier_settings.scale_verifiers;
 
-		let num_cpus = ::num_cpus::get();
-
-		let max_verifiers = cmp::max(num_cpus, MAX_VERIFIERS);
+		let max_verifiers = ::num_cpus::get();
 		let default_amount = cmp::max(1, cmp::min(max_verifiers, config.verifier_settings.num_verifiers));
 
 		// if `auto-scaling` is enabled spawn up extra threads as they might be needed
@@ -753,6 +748,13 @@ mod tests {
 		BlockQueue::new(config, engine, IoChannel::disconnected(), true)
 	}
 
+	fn get_test_config(num_verifiers: usize, is_auto_scale: bool) -> Config {
+		let mut config = Config::default();
+		config.verifier_settings.num_verifiers = num_verifiers;
+		config.verifier_settings.scale_verifiers = is_auto_scale;
+		config
+	}
+
 	fn new_unverified(bytes: Bytes) -> Unverified {
 		Unverified::from_rlp(bytes).expect("Should be valid rlp")
 	}
@@ -853,12 +855,11 @@ mod tests {
 
 	#[test]
 	fn scaling_limits() {
-		use super::MAX_VERIFIERS;
-
+		let max_verifiers = ::num_cpus::get();
 		let queue = get_test_queue(true);
-		queue.scale_verifiers(MAX_VERIFIERS + 1);
+		queue.scale_verifiers(max_verifiers + 1);
 
-		assert!(queue.num_verifiers() < MAX_VERIFIERS + 1);
+		assert!(queue.num_verifiers() < max_verifiers + 1);
 
 		queue.scale_verifiers(0);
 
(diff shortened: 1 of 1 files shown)


ethcore/src/verification/queue/mod.rs
```diff
@@ -232,12 +232,20 @@ impl<K: Kind> VerificationQueue<K> {
...
 		let num_cpus = ::num_cpus::get();
-		let max_verifiers = cmp::min(num_cpus, MAX_VERIFIERS);
-		let default_amount = cmp::max(1, cmp::min(max_verifiers, config.verifier_settings.num_verifiers));
-		let state = Arc::new((Mutex::new(State::Work(default_amount)), Condvar::new()));
+		let cli_max_verifiers = cmp::max(1, config.verifier_settings.num_verifiers);
+
+		// if `auto-scaling` is enabled spawn up extra threads as they might be needed
+		// otherwise just spawn the number of threads specified by the config
+		let max_verifiers = if scale_verifiers {
+			cmp::min(num_cpus, MAX_VERIFIERS)
```
ordian: Do you think it makes sense to print a WARN or INFO, that `num_verifiers` is not used in this case?

ethcore/src/verification/queue/mod.rs
```diff
@@ -232,12 +232,20 @@ impl<K: Kind> VerificationQueue<K> {
...
+		} else {
+			cmp::min(cli_max_verifiers, cmp::min(num_cpus, MAX_VERIFIERS))
+		};
+
+		let state = Arc::new((Mutex::new(State::Work(cli_max_verifiers)), Condvar::new()));
```
ordian: Should this be `max_verifiers`?

ordian: What I actually meant is `cli_max_verifiers` should be <= `max_verifiers`

niklasad1: I need to rework this logic cli_max_verifiers can be bigger than max_verifiers. I will add tests 🦀🦀

ethcore/src/verification/queue/mod.rs
```diff
@@ -877,4 +887,57 @@ mod tests {
...
+			config.verifier_settings.num_verifiers = 3;
+			config.verifier_settings.scale_verifiers = false;
+
+			let queue = BlockQueue::new(config, engine, IoChannel::disconnected(), true);
+
+			assert_eq!(queue.num_verifiers(), 3);
+		}
+
+		#[test]
+		fn worker_threads_specifyed_to_zero_should_set_to_one() {
```
dvdplm: typo: s/specifyed/specified/

ethcore/src/verification/queue/mod.rs
```diff
@@ -877,4 +887,57 @@ mod tests {
...
+			let spec = Spec::new_test();
+			let engine = spec.engine;
+			let mut config = Config::default();
+			config.verifier_settings.num_verifiers = 1;
```
dvdplm: Shouldn't this be `0`?

ethcore/src/verification/queue/mod.rs
```diff
@@ -877,4 +887,57 @@ mod tests {
...
+			assert_eq!(queue.num_verifiers(), 8);
+		}
+
+		#[test]
+		fn worker_threads_scaling_with_specifed_num_of_workers() {
+			let spec = Spec::new_test();
+			let engine = spec.engine;
+			let mut config = Config::default();
+			config.verifier_settings.num_verifiers = 5;
+			config.verifier_settings.scale_verifiers = true;
```
dvdplm: Maybe add a helper to reduce the boiler plate code a bit, e.g.:

```rust
fn test_config(verifiers: usize, scale: bool) -> Config {

}
```
```

File: ethcore/src/verification/queue/mod.rs [hunk 6]
```diff
@@ -877,4 +887,57 @@ mod tests {
...
+
+			let queue = BlockQueue::new(config, engine, IoChannel::disconnected(), true);
+			queue.scale_verifiers(8);
+
+			assert_eq!(queue.num_verifiers(), 8);
```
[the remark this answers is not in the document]
niklasad1: That's a good question and the answer is yes. I coded this because we assume that the maximum number of CPUs is 8 by the ```constant MAX_VERIFIERS``` but I think it doesn't make sense when thinking about it now! 

I will change this test and remove the constant!

ethcore/src/verification/queue/mod.rs
```diff
@@ -39,9 +39,6 @@ pub mod kind;
 const MIN_MEM_LIMIT: usize = 16384;
 const MIN_QUEUE_LIMIT: usize = 512;
 
-// maximum possible number of verification threads.
-const MAX_VERIFIERS: usize = 8;
```
niklasad1: This is removed because it assumes max 8 CPUs now instead use on ```num_cpus::get()``` instead

ethcore/src/verification/queue/mod.rs
```diff
@@ -877,4 +888,50 @@ mod tests {
...
+		}
+
+		#[test]
+		fn worker_threads_scaling_with_specifed_num_of_workers() {
+			let num_cpus = ::num_cpus::get();
+			// only run the test with at least 2 CPUs
+			if num_cpus > 1 {
+				let spec = Spec::new_test();
+				let engine = spec.engine;
+				let config = get_test_config(num_cpus - 1, true);
```
niklasad1: This can't overflow because ```num_cpus``` is at least 2

Issue #9558, opened 2018-09-14
Deduplicate smallvec dependency


ordian:
Currently we have several versions of the same (transitive) dependency, e.g.:
```text
smallvec v0.2.1
smallvec v0.4.5
smallvec v0.6.5
```
It would be nice to remove duplicates, which in turn will reduce compile time and maybe binary size.
To see the list of dependencies, run: `cargo tree --no-indent | sort | rg -v "\(\*\)$" | uniq`

debris:
I don't think we should have a generic issue for **all** duplicate dependencies in parity. It should be a policy to avoid them, but with so many of them, the duplication is sometimes inevitable

ordian:
closed by #9657

PR #9631, opened 2018-09-24
EIP-712 implementation

### LINKED ISSUE #9424: Implement EIP-712 Signing API
Parity lacks an API to perform an EIP-712 conforming signature for a given account. An API should be implemented that allows for arbitrary data to be signed by a specified account. This API will encompass signing as well as verifying functionalities.

Current API: https://wiki.parity.io/JSONRPC-personal-module#personal_sign

seunlanlege:
closes #9424

**Missing**
- [x] encode nested array types


util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+			t if message_types.get(t).is_some() => {
+				let encoded = encode_data(&field.type_, &message_types, &value)?;
+				let hash = (&keccak(encoded)).to_vec();
+				tokens.push(Token::FixedBytes(hash));
+			}
+			// (un)signed integers
+			ty if ty.starts_with("uint") || ty.starts_with("int") => {
+				print!("{}", value);
+				// try to deserialize as a number first, then a string
+				let uint = match (value.as_u64(), value.as_str()){
```
seunlanlege: i guess i should make this `u128`

util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+	let mut tokens = vec![Token::FixedBytes(type_hash)];
+	for field in message_types.get(message_type).ok_or_else(|| ErrorKind::NonExistentType)? {
+		let value = &message[&field.name];
+		match &*field.type_ {
+			// Array type e.g uint256[], string[]
+			ty if ty.rfind(']') == Some(ty.len() - 1) => {
+				let array_type = ty.split('[').collect::<Vec<_>>()[0];
+				let mut items = vec![];
+				for item in value.as_array().ok_or_else(|| serde_error("array", &field.name))? {
+					let encoded = encode_data(array_type.into(), &message_types, item)?;
```
tomusdrw: This is also recursive, might be good to actually limit the depth to prevent stack overflows.

util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+				tokens.push(Token::Address(address));
+			}
+			// custom type defined in message types
+			t if message_types.get(t).is_some() => {
+				let encoded = encode_data(&field.type_, &message_types, &value)?;
+				let hash = (&keccak(encoded)).to_vec();
+				tokens.push(Token::FixedBytes(hash));
+			}
+			// (un)signed integers
+			ty if ty.starts_with("uint") || ty.starts_with("int") => {
```
tomusdrw: Is `uinthello` a valid type? We should reject such stuff.

util/EIP-712/src/eip712.rs
```diff
@@ -0,0 +1,135 @@
...
+pub struct EIP712 {
+	pub types: MessageTypes,
+	pub primary_type: String,
+	pub message: Value,
+	pub domain: Value,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct FieldType {
+	pub name: String,
```
tomusdrw: We should make sure that:
1. Every `name` does not have any spaces or commas (otherwise you could mess up the encoding and do: `Address(string na, uint256 me)` and instead of encoding one field you encode two (i.e. the encoding becomes ambigous)
2. The type is one from the list of known/supported types - might be good to actually have an enum for that if possible.

util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+				print!("{}", value);
```
tomusdrw: to remove

util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+use ethabi::{encode, Token};
+use ethereum_types::{Address, U256};
+use keccak_hash::{keccak, H256};
+use serde_json::Value;
+use std::collections::HashSet;
+use std::str::FromStr;
+
+/// given a type and HashMap<String, Vec<FieldType>>
+/// returns a HashSet of dependent types of the given type
+fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes) -> Option<(HashSet<&'a str>)>
```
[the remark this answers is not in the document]
seunlanlege: the compiler can't choose the lifetime of the return type

5chdn:
Can you set git user name and mail?

https://help.github.com/articles/setting-your-commit-email-address-in-git/

5chdn:
Can you associate the email with your Github account? It does not show your identity for commits

Cargo.lock
```diff
@@ -641,9 +661,9 @@ dependencies = [
  "rand 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "rlp 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "rlp_derive 0.1.0",
- "serde 1.0.78 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.78 (registry+https://github.com/rust-lang/crates.io-index)",
- "smallvec 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.79 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.79 (registry+https://github.com/rust-lang/crates.io-index)",
+ "smallvec 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
```
ordian: Could you revert unrelated changes like this (we try to avoid duplicate deps, see #9558)?

Also please revert submodules updates.

util/EIP-712/src/parser.rs
```diff
@@ -0,0 +1,133 @@
...
+						return Err(ErrorKind::UnexpectedToken(lexer.token_as_str().to_owned(), field_type.to_owned()))?
+					}
+				}
+				Token::BracketClose if array_depth == 10 => {
+					return Err(ErrorKind::UnsupportedArrayDepth)?
+				}
+				_  => return Err(ErrorKind::UnexpectedToken(lexer.token_as_str().to_owned(), field_type.to_owned()))?
+			};
+
+			token = Some(type_);
```
ordian: So it parses the last token type (e.g. `"byte uint byte"` is parsed as `Bytes(1)`). Is this the expected behavior?

seunlanlege: actually no, the type string would be validated before it's parsed.

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,356 @@
...
+		}
+
+		Type::Bytes(size) if *size < 32 => {
+			let string = value.as_str().ok_or_else(|| serde_error("string", field_name))?;
+			if string.len() < 2 {
+				return Err(ErrorKind::HexParseError(
+					format!("Expected a 0x-prefixed string of even length, found {} length string", string.len()))
+				)?
+			}
+			let string = string.get(2..).expect("line 200 checks for length; qed");
```
ordian: * Why not just `[2..]` (and in other places)?
* Can we reuse code with case `if *size == 32`?

util/EIP-712/src/eip712.rs
seunlanlege: @ordian thanks a lot for that crate :handshake:

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,356 @@
...
+			// try to deserialize as a number first, then a string
+			let uint = match (value.as_u64(), value.as_str()) {
+				(Some(number), _) => U256::from(number),
+				(_, Some(string)) => {
+					if string.len() <= 2 {
+						return Err(ErrorKind::HexParseError(
+							format!("Expected a 0x-prefixed string of even length, found {} length string", string.len()))
+						)?
+					}
+					let string = string.get(2..).expect("line 200 checks for length");
```
pepyakin: I wouldn't recommend refering checks by line, because of maintenance burden imposed on the future editors. Chances are that nobody actually will notice breaking this invariant by adding some lines before it.

In fact, this already is outdated since line 200 is occupied by this `/// encodes and hashes the given EIP712 struct`.

rpc/src/v1/traits/personal.rs
```diff
@@ -42,6 +42,11 @@ build_rpc_trait! {
 		#[rpc(name = "personal_sign")]
 		fn sign(&self, Bytes, H160, String) -> BoxFuture<H520>;
 
+		/// Produces an EIP-720 compliant signature with given account using the given password to unlock the
```
tomusdrw: eip-712?

util/EIP-712/src/eip712.rs
tomusdrw: @seunlanlege I think you still need to call `.validate()` to perform validation - I haven't seen it anywhere in the code. If I'm mistaken please add a test case that fails to deserialize invalid values.

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,339 @@
...
+		}
+
+		Type::Bytes(size) => {
+			let string = value.as_str().ok_or_else(|| serde_error("string", field_name))?;
+			if string.len() < 2 {
+				return Err(ErrorKind::HexParseError(
+					format!("Expected a 0x-prefixed string of even length, found {} length string", string.len()))
+				)?
+			}
+			let mut bytes = H256::from_str(&string[2..]).map_err(|err| ErrorKind::HexParseError(format!("{}", err)))?;
```
tomusdrw: This seems incorrect. Anything greater than 32-bytes will fail to be decoded, why do you even use `H256` here, it should be arbitrary bytes, no?

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,339 @@
...
+			if *size == 32 {
```
tomusdrw: I don't understand what's going on here, we hash only in case the size is 32? Can you point to a part of the spec defining that?

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,339 @@
...
+use itertools::Itertools;
+use linked_hash_set::LinkedHashSet;
+use serde_json::to_value;
+use parser::{Parser, Type};
+use error::{Result, ErrorKind, serde_error};
+use eip712::{EIP712, MessageTypes};
+
+/// given a type and HashMap<String, Vec<FieldType>>
+/// returns a HashSet of dependent types of the given type
+fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes) -> Option<(LinkedHashSet<&'a str>)>
```
tomusdrw: My previous question is still valid: Can you argument why we need `LinkedHashSet` here?

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,339 @@
...
+				return Err(ErrorKind::InvalidAddressLength(addr.len()))?;
+			}
+			let address = EthAddress::from_str(&addr[2..]).map_err(|err| ErrorKind::HexParseError(format!("{}", err)))?;
+			encode(&[EthAbiToken::Address(address)])
+		}
+
+		Type::Uint | Type::Int => {
+			// try to deserialize as a number first, then a string
+			let uint = match (value.as_u64(), value.as_str()) {
+				(Some(number), _) => U256::from(number),
```
tomusdrw: Rised that earlier already - are we sure we want to support amibgious formats for numbers (i.e. regular numbers and hex-encoded strings?) The reason we don't do that in RPC is because JavaScript doesn't support large numbers natively, and accepting numbers fools the developers that they can just pass arbitrary number (even U256) to the API and it will work, while JS might decide differently, so it's very error-prone.
I'm strongly in favor of supporting only hex-encoded (`QUANTITY`) numbers

util/EIP-712/src/lib.rs
```diff
@@ -0,0 +1,303 @@
+extern crate serde;
```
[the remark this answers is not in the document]
tomusdrw: Could you add `unused_extern_crates` as well?

util/EIP-712/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+keccak-hash = "0.1"
+ethereum-types = "0.4"
+hex = "0.3"
+failure = "0.1"
+itertools="0.7"
+failure_derive = "0.1"
+lazy_static = "1.1"
+linked_hash_set = "0.1.3"
+toolshed = "0.4"
+lunarity = { git = "https://github.com/paritytech/lunarity" }
```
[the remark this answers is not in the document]
tomusdrw: @maciejhirsz any progress on this, I think it would be better to avoid git-deps.

util/EIP-712/src/encode.rs
seunlanlege: Yeah, you're right, will remove it.

util/EIP-712/src/encode.rs
seunlanlege: `std::collections::HashSet` doesn't have a `pop` method, ``LinkedHashSet` has a `pop_back` method,

util/EIP-712/src/encode.rs
seunlanlege: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-typed-structured-data-%F0%9D%95%8A

seunlanlege: @maciejhirsz looks like lunarity parses `bytes` as `Bytes32`

util/EIP-712/src/encode.rs
tomusdrw: Still not convinced:
1. we use `pop_back` for `types` not `deps` (the return type)
2. Why do we need to process `types` in order?

util/EIP-712/src/encode.rs
tomusdrw: Don't see anyhing in the link that would suggest that we hash in case the size is 32 bytes. It actually says somehing completely different:

> bytes1 to bytes31 are arrays with a beginning (index 0) and an end (index length - 1), they are zero-padded at the end to bytes32 and encoded in beginning to end order. This corresponds to their encoding in ABI v1 and v2.
> The dynamic values bytes and string are encoded as a keccak256 hash of their contents.

So we should only hash if bytes vector is (strictly) greater than 32 bytes and just pad the results in case it's smaller.

seunlanlege: > looks like lunarity parses bytes as Bytes32

@tomusdrw yeah i made the incorrect assumption that `bytes32` == `bytes`

util/EIP-712/src/encode.rs
seunlanlege: To answer "why does `types` have to be  `HashSet`"?, It's an optimization.

I could either 
1. iterate through the `types` to assert that the `type` i'm adding to the array doesn't already exist.
2. I could just use a hashset.

util/EIP-712/src/parser.rs
```diff
@@ -0,0 +1,138 @@
...
+	fn from(field_type: Type) -> String {
+		match field_type {
+			Type::Address => "address".into(),
+			Type::Uint => "uint".into(),
+			Type::Int => "int".into(),
+			Type::String => "string".into(),
+			Type::Bool => "bool".into(),
+			Type::Bytes(len) => format!("bytes{}", len),
+			Type::Custom(custom) => custom,
+			Type::Array(type_) => (*type_).into()
```
seunlanlege: should be `format!("{}[]", (*type_).into())`

util/EIP-712/src/encode.rs
tomusdrw: @seunlanlege I was asking more about `Linked` variant of `HashSet` (which requires an external crate). I'm completely ok with how it is implemented right now, just trying to understand the design decision behind `LinkedHashSet`, like:
1. Is it required by the spec (the order of items matter)
2. Is it required for tests (makes asserting that something is on particular position in the hashset)

The minimalist in me is just reluctant to include an external crate if it's not strictly required :)

seunlanlege: Actually no the spec doesn't in any way require a `LinkedHashSet`, I require a `HashSet` that i can `pop` from; which the one in the stdlib lacks.

seunlanlege: I can make `deps` `std::collection::HashSet`; it doesn't actually need to be `LinkedHashSet`

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,414 @@
...
+				.from_hex::<Vec<u8>>()
+				.map_err(|err| ErrorKind::HexParseError(format!("{}", err)))?;
+			let bytes = keccak(&bytes).to_vec();
+
+			encode(&[EthAbiToken::FixedBytes(bytes)])
+		}
+
+		Type::Byte(_) => {
+			let string = value.as_str().ok_or_else(|| serde_error("string", field_name))?;
+			if string.len() < 2 {
```
ordian: do we want to check if the prefix is actually `"0x"`? and maybe extract this check into a function

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,414 @@
...
+				return Err(ErrorKind::HexParseError(
+					format!("Expected a 0x-prefixed string of even length, found {} length string", string.len()))
+				)?
+			}
+			let mut bytes = (&string[2..])
+				.from_hex::<Vec<u8>>()
+				.map_err(|err| ErrorKind::HexParseError(format!("{}", err)))?;
+
+			if *message_type == Type::Bytes {
```
ordian: was this case supposed to be merged with `Type::Bytes`?

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,414 @@
...
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! EIP712 Encoder
```
ordian: It would be nice to add api layout to the docs, something like this:
https://github.com/PaulRBerg/go-ethereum/blob/0b952c17c47d62381bdbb74ec1e1434a1906c7bf/signer/core/api_layout.md, although it's documented well in the EIP.

util/EIP-712/src/encode.rs
```diff
@@ -0,0 +1,414 @@
...
+
+/// given a type and HashMap<String, Vec<FieldType>>
+/// returns a HashSet of dependent types of the given type
+fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes) -> Option<(HashSet<&'a str>)>
+{
+	if message_types.get(message_type).is_none() {
+		return None;
+	}
+
+	let mut types = LinkedHashSet::new();
```
ordian: [indexmap](https://github.com/bluss/indexmap) should perform better here (and it does have `pop`), although it's definitely not a bottleneck.

util/EIP-712/src/parser.rs
```diff
@@ -0,0 +1,176 @@
...
+				Token::BracketOpen if token.is_some() && state == State::Close => {
+					state = State::Open;
+					lexer.consume();
+					continue
+				}
+				Token::BracketClose if array_depth < 10 => {
+					if state == State::Open && token.is_some() {
+						let length = match current_array_length.take() {
+							None => 0,
+							Some(len) => len
```
tomusdrw: Seems that both `Type[0]` and `Type[]` would be parsed to the same value, is that correct?

seunlanlege: but i think the code could be more declarative.

util/EIP-712/src/eip712.rs
```diff
@@ -0,0 +1,177 @@
...
+pub struct EIP712 {
+	pub(crate) types: MessageTypes,
+	pub(crate) primary_type: String,
+	pub(crate) message: Value,
+	pub(crate) domain: EIP712Domain,
+}
+
+impl Validate for EIP712 {
+	fn validate(&self) -> Result<(), ValidationErrors> {
+		for (_, field_types) in self.types.iter() {
```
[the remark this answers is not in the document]
seunlanlege: not entirely sure what you mean.

util/EIP-712/src/eip712.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use serde_json::from_str;
+	#[test]
+	fn test_deserialization() {
+		let string = r#"{
+            "primaryType": "Mail",
```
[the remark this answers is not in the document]
seunlanlege: it is

util/EIP-712/src/eip712.rs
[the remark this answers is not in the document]
tomusdrw: `self.types` is a `HashMap`, you are using only values from `Iterator`, so you can just use `.values()` to get iterator for that:

```suggestion
		for field_types in self.types.values() {
```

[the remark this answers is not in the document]
seunlanlege: ah ok, didn't realize i could do that, thanks!

PR #9638, opened 2018-09-24
fix bad-block reporting no reason

debris:
related to #9433


debris:
#9433 also introduced another one redundant clone of `block.bytes`

> https://github.com/paritytech/parity-ethereum/blob/adcbfcf8d652ecb99c139dc3c478a2b9830697d1/ethcore/src/client/client.rs#L1405-L1410

I'll fix it in a separate pull request

Issue #9635, opened 2018-09-24
key_server_cluster randomly failing


debris:
job: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/100432

pr: #9633 

```
---- key_server_cluster::cluster::tests::ecdsa_signing_session_completes_if_node_does_not_have_a_share stdout ----
thread 'key_server_cluster::cluster::tests::ecdsa_signing_session_completes_if_node_does_not_have_a_share' panicked at 'no result in 300ms', secret_store/src/key_server_cluster/cluster.rs:1253:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:221
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:475
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:345
   7: ethcore_secretstore::key_server_cluster::cluster::tests::ecdsa_signing_session_completes_if_node_does_not_have_a_share
   8: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1451
             at /checkout/src/libcore/ops/function.rs:223
             at /checkout/src/liballoc/boxed.rs:642
   9: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:105
```

job: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/100420

pr: #9360

```
---- key_server_cluster::cluster::tests::schnorr_signing_session_completes_if_node_does_not_have_a_share stdout ----
thread 'key_server_cluster::cluster::tests::schnorr_signing_session_completes_if_node_does_not_have_a_share' panicked at 'no result in 300ms', secret_store/src/key_server_cluster/cluster.rs:1253:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
[...]

PR #9639, opened 2018-09-24
ignore key_server_cluster randomly failing tests

debris:
related to #9635


secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1432,6 +1432,7 @@ pub mod tests {
 	}
 
 	#[test]
+	#[ignore]
```
tomusdrw: I like to see a comment explaining why a test ignored, this helps later on checking on this case and also understanding the reasoning behind this.
I think the link to #9635 would be good here.

PR #9644, opened 2018-09-25
refactor (hardware wallet) : reduce the number of threads

### LINKED ISSUE #9622: Refactor hardware-wallet EventHandler to only use 1 thread
The [hardware wallet manager](https://github.com/paritytech/parity-ethereum/tree/master/hw) is handling usb events in two different threads for Trezor and Ledger.
This should be replaced by using only one thread for less memory usage!

niklasad1:
Attempt to close #9622 along with fixing some minor formatting things!

It introduces some boiler-plate code though such as:
* Some errors are duplicated in trezor and ledger but I think these makes sense because it indicate the errors more clearly
* Device arrived and left methods
* Checking the device is valid
* Fixed the Ledger logs which was faulty (added inline comments for more info)

Tested successfully both on `Trezor` and `Ledger` on Linux

Proof that number of threads have been reduced:

```bash
# This branch
$  ps hH p 19919 | wc -l
57
# master
$ ps hH p 23329 | wc -l 
58
```


hw/src/ledger.rs
```diff
@@ -110,65 +110,36 @@ impl From<libusb::Error> for Error {
...
 	devices: RwLock<Vec<Device>>,
 	key_path: RwLock<KeyPath>,
 }
 
 impl Manager {
 	/// Create a new instance.
-	pub fn new(hidapi: Arc<Mutex<hidapi::HidApi>>, exiting: Arc<AtomicBool>) -> Result<Arc<Self>, libusb::Error> {
-		let manager = Arc::new(Self {
-			usb: hidapi,
+	pub fn new(usb: Arc<Mutex<hidapi::HidApi>>) -> Arc<Self> {
```
niklasad1: doesn't make sense that this function returns Result any because it ```can't fail``` anymore

hw/src/ledger.rs
```diff
@@ -110,65 +110,36 @@ impl From<libusb::Error> for Error {
...
-					}
-				}
-			})
-			.ok();
-
-		Ok(manager)
+		})
 	}
 
 	// Transport Protocol:
```
niklasad1: fix formatting!!!

niklasad1:
@5chdn Need grab to hardware wallets and test this first. I will change the label when it is good for reviewing hopefully later today

hw/src/ledger.rs
```diff
@@ -363,17 +323,6 @@ impl Manager {
...
-	while start_time.elapsed() <= *timeout {
-		if let Ok(num_devices) = ledger.update_devices(device_direction) {
-			trace!(target: "hw", "{} number of Ledger(s) {}", num_devices, device_direction);
-			return true;
-		}
-	}
-	false
-}
 
 impl <'a>Wallet<'a> for Manager {
```
sorpaas: nit: not introduced in this PR, but space should be after `<'a>`.

hw/src/lib.rs
```diff
@@ -292,10 +328,75 @@ impl HardwareWalletManager {
 
 impl Drop for HardwareWalletManager {
 	fn drop(&mut self) {
-		// Indicate to the USB Hotplug handlers that they
-		// shall terminate but don't wait for them to terminate.
-		// If they don't terminate for some reason USB Hotplug events will be handled
+		// Indicate to the USB Hotplug handlers that it
+		// shall terminate but don't wait for it to terminate.
+		// If it don't terminate for some reason USB Hotplug events will be handled
```
sorpaas: nit: `s/don't/doesn't`

5chdn:
:+1: needs a 2nd review

hw/src/lib.rs
```diff
@@ -292,10 +328,75 @@ impl HardwareWalletManager {
...
+	fn extract_device_info(device: &libusb::Device) -> Result<(u16, u16), Error> {
+		let desc = device.device_descriptor()?;
+		Ok((desc.vendor_id(), desc.product_id()))
+	}
+}
+
+impl libusb::Hotplug for EventHandler {
+	fn device_arrived(&mut self, device: libusb::Device) {
+		// Upgrade reference to an Arc
+		if let (Some(ledger), Some(trezor)) = (self.ledger.upgrade(), self.trezor.upgrade()) {
```
dvdplm: q: is it impossible to have one upgrade call return `Some` and the other `None`? Wouldn't that cause the whole block to be skipped? In other words, are both `upgrade()` calls going to return `Some` if one of them does?

hw/src/lib.rs
```diff
@@ -292,10 +328,75 @@ impl HardwareWalletManager {
...
+						trace!(target: "hw", "Ledger device was detected but connection failed");
+					}
+				}
+			}
+		}
+	}
+
+	fn device_left(&mut self, device: libusb::Device) {
+		// Upgrade reference to an Arc
+		if let (Some(ledger), Some(trezor)) = (self.ledger.upgrade(), self.trezor.upgrade()) {
```
dvdplm: Same question here.

hw/src/lib.rs
```diff
@@ -292,10 +328,75 @@ impl HardwareWalletManager {
...
+				}
+			}
+		}
+	}
+}
+
+
+/// Helper to determine if a device is a valid HID
+pub fn is_valid_hid_device(usage_page: u16, interface_number: i32) -> bool {
+	usage_page == 0xFF00 || interface_number == 0
```
dvdplm: Can you use `const`s here? Would make it easier to read and document.

hw/src/ledger.rs
```diff
@@ -393,8 +342,17 @@ impl <'a>Wallet<'a> for Manager {
 		let devices = usb.devices();
 		let num_prev_devices = self.devices.read().len();
 
+		// Wrong state, sometimes when a ledger is connected at run-time (with no other devices connected)
```
dvdplm:
```suggestion
		// Sometimes when a ledger is connected at run-time with no other devices connected it will case a `disconnected` event. To work around this, ignore such spurious events and poll a couple of extra times in order to get the correct state.
```

hw/src/lib.rs
niklasad1: Good question, because the HardwareWalletManager keeps an `Arc` for both `Ledger` and `Trezor` for the entire lifetime of Parity unless it is disabled by the cli arg `--no-hardware-wallet`! upgrade()` will always succeed as long the `HardwareWalletManager` is enabled (the strong count will be at least 1). So, either both `Trezor` and `Ledger` is enabled or none is enabled.

However, I admit it is a bit sloppy. Theoretically, interleaving "could" happen if one of the `hw-wallets` has been dropped and the other has not been dropped but this is matter in granularity in terms of microseconds and I don't worry too much about this.

I can change it if you like? :P

5chdn:
Please resolve conflicts :)

Issue #9623, opened 2018-09-21
CLI argument to specify the number of `dns worker threads` in `fetch`


niklasad1:
Add functionality to control the number of `DNS worker threads` in [fetch](https://github.com/paritytech/parity-ethereum/tree/master/util/fetch), it is currently hardcoded to 4

This is the line that specifies the number of `DNS worker threads`:
https://github.com/paritytech/parity-ethereum/blob/master/util/fetch/src/client.rs#L211

debris:
is it really needed? a difference between footprint of a single thread and a few threads in this case is probably so small that no user will bother to change this CLI param

debris:
@tomusdrw what do you think?

niklasad1:
@debris this is part of an attempt to make the light client `lighter` and AFAIK the default-stack size for pthread is somewhere between 2MB - 32 MB (on x86) not sure on other platforms and Windows though.

In the light-client mode I would say that this is worth the effort!

debris:
so maybe let's just set it by default to 1 thread for `light` client and keep 4 for normal run :)

tomusdrw:
I second @debris, 1 thread for light client and 4 for full node seems like a good idea. Possilby we could even have always 1 thread, since we don't really talk to random addreses at all - anything that get's resolved on a first request is probably cached forever.

niklasad1:
Sure, it sounds good to me!

However, because we compile light and full the same I think a simple flag to ```fetch``` is the way to go here for now and I'm not a big fan of feature flags!

tomusdrw:
@niklasad1 we can check the `--light` flag, no? And run with 1 thread if that's enabled.

niklasad1:
@tomusdrw 
Yeah, I meant in `parity::run::excute` we check the light flag either to run as full or light. In those different methods, we need to pass this flag into the ```fetch::Client``` constructor whether if it `should` be light or not! So, yes we could use the light flag or simple bool!

I submit a PR to this probably easier to understand then!

PR #9647, opened 2018-09-25
refactor(fetch) : light use only one `DNS` thread

niklasad1:
Attempt to close https://github.com/paritytech/parity-ethereum/issues/9623

### DIFF
--- a/parity/run.rs
+++ b/parity/run.rs
@@ -283,7 +283,7 @@ fn execute_light_impl(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<Runnin
 	let cpu_pool = CpuPool::new(4);
 
 	// fetch service
-	let fetch = fetch::Client::new().map_err(|e| format!("Error starting fetch client: {:?}", e))?;
+	let fetch = fetch::Client::new(cmd.light).map_err(|e| format!("Error starting fetch client: {:?}", e))?;
 	let passwords = passwords_from_files(&cmd.acc_conf.password_files)?;
 
 	// prepare account provider
@@ -477,7 +477,7 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq: 
 	let event_loop = EventLoop::spawn();
 
 	// fetch service
-	let fetch = fetch::Client::new().map_err(|e| format!("Error starting fetch client: {:?}", e))?;
+	let fetch = fetch::Client::new(cmd.light).map_err(|e| format!("Error starting fetch client: {:?}", e))?;
 
 	let txpool_size = cmd.miner_options.pool_limits.max_count;
 	// create miner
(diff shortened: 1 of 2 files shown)


util/fetch/src/client.rs
```diff
@@ -199,16 +199,17 @@ impl Client {
...
-	fn background_thread(tx_start: TxStartup, rx_proto: mpsc::Receiver<ChanItem>) -> io::Result<thread::JoinHandle<()>> {
+	fn background_thread(tx_start: TxStartup, rx_proto: mpsc::Receiver<ChanItem>, is_light: bool) -> io::Result<thread::JoinHandle<()>> {
 		thread::Builder::new().name("fetch".into()).spawn(move || {
 			let mut core = match reactor::Core::new() {
 				Ok(c) => c,
 				Err(e) => return tx_start.send(Err(e)).unwrap_or(())
 			};
 
 			let handle = core.handle();
+			let number_dns_threads = if is_light { 1 } else { 4 };
```
ascjones: Instead of passing the flag down you could pass the number of threads as a parameter.

parity/run.rs
```diff
@@ -283,7 +283,7 @@ fn execute_light_impl(cmd: RunCmd, logger: Arc<RotatingLogger>) -> Result<Runnin
 	let cpu_pool = CpuPool::new(4);
 
 	// fetch service
-	let fetch = fetch::Client::new().map_err(|e| format!("Error starting fetch client: {:?}", e))?;
+	let fetch = fetch::Client::new(cmd.light).map_err(|e| format!("Error starting fetch client: {:?}", e))?;
```
ascjones: This will always be true so could just pass in `1` here

parity/run.rs
```diff
@@ -477,7 +477,7 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 	let event_loop = EventLoop::spawn();
 
 	// fetch service
-	let fetch = fetch::Client::new().map_err(|e| format!("Error starting fetch client: {:?}", e))?;
+	let fetch = fetch::Client::new(cmd.light).map_err(|e| format!("Error starting fetch client: {:?}", e))?;
```
ascjones: This will always be false so could just pass in `4` here

util/fetch/src/client.rs
niklasad1: That's better, will change it!

parity/run.rs
```diff
@@ -77,7 +77,13 @@ const SNAPSHOT_HISTORY: u64 = 100;
 const GAS_CORPUS_EXPIRATION_MINUTES: u64 = 60 * 6;
 
 // Pops along with error messages when a password is missing or invalid.
-const VERIFY_PASSWORD_HINT: &'static str = "Make sure valid password is present in files passed using `--password` or in the configuration file.";
+const VERIFY_PASSWORD_HINT: &str = "Make sure valid password is present in files passed using `--password` or in the configuration file.";
+
+// Full client number of DNS threads
+const FETCH_FULL_NUM_DNS_THREADS: usize = 4;
```
tomusdrw: I doubt we actually need all 4 for the full node either. But fine with keeping it the same for backward compatibility.

PR #9607, opened 2018-09-20
HF in POA Sokol (2018-09-19)

varasev:
Hard fork in POA Sokol was done on `September 19th, 2018`. The proposed changes correspond to the changes in our `spec.json` for `Sokol`: https://github.com/poanetwork/poa-chain-spec/pull/86/files#diff-42eb5109ad96d4ac46cdcbf18f2938de

PR #9648, opened 2018-09-25
Backports for stable 2.0.7

5chdn:
- [x] parity-version: bump stable to 2.0.7
- [x] Fix path to parity.h (#9274)
- [ ] ~Fix ancient blocks sync (#9531)~ - too hairy to backport
- [x] ethcore: fix detection of major import (#9552)
- [x] fix (light/provider) : Make `read_only executions` only read-only (#9591)
- [x] HF in POA Sokol (2018-09-19) (#9607)
- [x] fix failing node-table tests on mac os (#9633)
- [x] fix(light_fetch): avoid race with BlockNumber::Latest (#9665)
- [x] CI: Remove unnecessary pipes (#9681)
- [x] Docker: run parity as normal user (#9689)
- [x] CI: Skip docs job for master and nightly (#9693)
- [ ] ~shutdown devp2p connections (#9711)~ - too late
- [x] ethcore-io retries failed work steal (#9651)


ordian:
I believe backporting #9638 (fixes problems of #9433) here doesn't make sense, since #9433 is not in stable yet. Same applies to #9608.

scripts/gitlab/build-unix.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
niklasad1: I think we should have ```#!/usr/env bash``` here instead

5chdn: Agreed. Need to be fixed in master though

PR #9650, opened 2018-09-25
test.sh: use cargo --target for platforms other than linux, win or mac

gabreal:
test runs need cargo --target option for cross-compilation and should then build only but not run the tests.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -4,6 +4,7 @@
 FEATURES="json-tests,ci-skip-issue"
 OPTIONS="--release"
 VALIDATE=1
+THREADS=8
 
 case $1 in
   --no-json)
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -4,6 +4,7 @@
 FEATURES="json-tests,ci-skip-issue"
 OPTIONS="--release"
 VALIDATE=1
+THREADS=8
```
5chdn: is there a reason to hardcode the number of threads?

test.sh
```diff
@@ -29,31 +30,58 @@ esac
...
+
+case $CARGO_TARGET in
+  (x86_64-unknown-linux-gnu|x86_64-apple-darwin|x86_64-pc-windows-msvc|'')
+    # native builds
+    if [ "$VALIDATE" -eq "1" ]
+    then
+      echo "________Validate build________"
+      time cargo check --no-default-features
+      time cargo check --manifest-path util/io/Cargo.toml --no-default-features
+      time cargo check --manifest-path util/io/Cargo.toml --features "mio"
```
5chdn: why do they not require a $TARGET?

test.sh
```diff
@@ -29,31 +30,58 @@ esac
...
+    
+      # Validate chainspecs
+      echo "________Validate chainspecs________"
+      time ./scripts/validate_chainspecs.sh
+    fi
+
+
+    # Running the C++ example
+    echo "________Running the C++ example________"
```
5chdn: this is only run in win,mac, and linux?

test.sh
```diff
@@ -29,31 +30,58 @@ esac
...
+  (*)
+    if [ "$VALIDATE" -eq "1" ]
+    then
+      echo "________Validate build________"
+      time cargo check --target $CARGO_TARGET --no-default-features
+      time cargo check --target $CARGO_TARGET --manifest-path util/io/Cargo.toml --no-default-features
+      time cargo check --target $CARGO_TARGET --manifest-path util/io/Cargo.toml --features "mio"
+    
+      # Validate chainspecs
+      echo "________Validate chainspecs________"
```
5chdn: script duplicates a bit here, can we simplify it?

test.sh
gabreal: I just set it as a variable here and did not want to change the behaviour in this pr. But generally I think it's okay to have a fixed number - unless we decide not to run jobs concurrently and then dynamically set it to the number of cores available on a runner node.

test.sh
gabreal: They use the default target. I don't want to break the script for those who run it locally and probably don't have the CARGO_TARGET variable set because it's not required and specific to the CI.

test.sh
gabreal: Yes, @tomaka told me that that only makes sense.

test.sh
gabreal: Okay, I can do that.

gabreal:
>  does this mean we don't run tests for anything not linux, windows, macos?
exactly, cross-compiled binaries won't be run but only compiled.

test.sh
General-Beck: cargo help test
If you want to control the
number of simultaneous running test cases, pass the `--test-threads` option
to the test binaries:

    cargo test -- --test-threads=1
MB 
if darwin -> sysctl -n hw.ncpu
if windows -> echo %NUMBER_OF_PROCESSORS%
if linux -> $(nproc)

PR #9651, opened 2018-09-25
ethcore-io retries failed work steal

debris:


ethcore/src/engines/tendermint/mod.rs
```diff
@@ -693,7 +693,6 @@ impl Engine<EthereumMachine> for Tendermint {
 	}
 
 	fn stop(&self) {
-		self.step_service.stop()
```
debris: `stop` doesn't nee to be called as it is already called on drop

util/io/src/service_mio.rs
```diff
@@ -430,7 +430,7 @@ impl<Message> IoChannel<Message> where Message: Send + Sync + 'static {
 /// General IO Service. Starts an event loop and dispatches IO requests.
 /// 'Message' is a notification message type
 pub struct IoService<Message> where Message: Send + Sync + 'static {
-	thread: Mutex<Option<JoinHandle<()>>>,
+	thread: Option<JoinHandle<()>>,
```
debris: since `rust 1.29` `JoinHandle` is `Send + Sync`

util/io/src/worker.rs
```diff
@@ -91,8 +92,9 @@ impl Worker {
 
 			while !deleting.load(AtomicOrdering::Acquire) {
 				match stealer.steal() {
-					chase_lev::Steal::Data(work) => Worker::do_work(work, channel.clone()),
-					_ => break,
+					deque::Steal::Data(work) => Worker::do_work(work, channel.clone()),
+					deque::Steal::Retry => {},
```
debris: this was a bug, if worker lost the race for stealing data, worker was incorrectly killed

ordian: the break is from the inner `while` loop

tomusdrw: I'm pretty sure it's going to increase CPU usage, but should also improve performance, cause previously if a thread lost the race it would go to sleep vs re-trying and get the work done. If it went to sleep other threads had to finish whatever they were doing and then pick up the new task.

debris: there is a case when `Worker` is being dropped. If `steal` returns `Retry`, work will never be resumed

niklasad1:
Let's merge it?

PR #9655, opened 2018-09-26
remove master from releasable branches

General-Beck:
need backporting in beta 
fix https://gitlab.parity.io/parity/parity-ethereum/-/jobs/101065 etc


5chdn:
This probably stops releasing docker `latest`

gabreal:
Ah I see the update with the except clause.
I wonder do we really want to run the full pipeline on every merge to the master branch i.e. https://gitlab.parity.io/parity/parity-ethereum/pipelines/24033

We have nightly builds anyway.

5chdn:
I think master should only run tests for the platforms

PR #9657, opened 2018-09-26
Replace `tokio_core` with `tokio` (`ring` -> 0.13)

### LINKED ISSUE #9533: Update ring to 0.13 or submit upstream changes
[The version of ring currently in use within parity](https://github.com/paritytech/ring) is a fork.

@tomaka, Is there any reason these changes have not been submitted upstream? If not, I, or someone on my team can take care of it. Please advise.

### LINKED ISSUE #9722: error: Vulnerable crates found!
- **Parity Ethereum version**: 2.1.2
- **Operating system**: Linux
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: yes

```
    Scanning Cargo.lock for vulnerabilities (385 crate dependencies)
error: Vulnerable crates found!

ID:	 RUSTSEC-2018-0003
Crate:	 smallvec
Version: 0.2.1
Date:	 2018-07-19
URL:	 https://github.com/servo/rust-smallvec/issues/96
Title:	 Possible double free during unwinding in SmallVec::insert_many
Solution: upgrade to: >= 0.6.3 OR ^0.3.4 OR ^0.4.5 OR ^0.5.1

ID:	 RUSTSEC-2018-0001
Crate:	 untrusted
Version: 0.5.1
Date:	 2018-06-21
URL:	 https://github.com/briansmith/untrusted/pull/20
Title:	 An integer underflow could lead to panic
Solution: upgrade to: >= 0.6.2

error: 2 vulnerabilities found!
```

c0gent:
# Changes

### Replace `tokio_core` with `tokio`.

* Remove `tokio-core` and replace with `tokio` in

    - `ethcore/stratum`

    - `secret_store`

    - `util/fetch`

    - `util/reactor`

* Bump hyper to 0.12 in

    - `miner`

    - `util/fake-fetch`

    - `util/fetch`

    - `secret_store`

* Bump `jsonrpc-***` to 0.9 in

    - `parity`

    - `ethcore/stratum`

    - `ipfs`

    - `rpc`

    - `rpc_client`

    - `whisper`

* Bump `ring` to 0.13

* Use a more graceful shutdown process in `secret_store` tests.

* Convert some mutexes to rwlocks in `secret_store`.

### Consolidate Tokio Runtime use, remove `CpuPool`.

* Rename and move the `tokio_reactor` crate (`util/reactor`) to
  `tokio_runtime` (`util/runtime`).

* Rename `EventLoop` to `Runtime`.

    - Rename `EventLoop::spawn` to `Runtime::with_default_thread_count`.

    - Add the `Runtime::with_thread_count` method.

    - Rename `Remote` to `Executor`.

* Remove uses of `CpuPool` and spawn all tasks via the `Runtime` executor
  instead.

* Other changes related to `CpuPool` removal:

    - Remove `Reservations::with_pool`. `::new` now takes an `Executor` as an argument.

    - Remove `SenderReservations::with_pool`. `::new` now takes an `Executor` as an argument.

# Still Incomplete

* [X] Troubleshoot the following failing tests:
    - json_tests::state::state_tests::GeneralStateTest_stCallCreateCallCodeTest
    - json_tests::state::state_tests::GeneralStateTest_stDelegatecallTestHomestead
    - json_tests::state::state_tests::GeneralStateTest_stInitCodeTest
    - json_tests::state::state_tests::GeneralStateTest_stRandom
    - json_tests::state::state_tests::GeneralStateTest_stRevertTest
    - json_tests::state::state_tests::GeneralStateTest_stSpecialTest
    - json_tests::state::state_tests::GeneralStateTest_stSystemOperationsTest
    - json_tests::state::state_tests::GeneralStateTest_stTransactionTest
    - json_tests::state::state_tests::GeneralStateTest_stTransitionTest
    - json_tests::state::state_tests::GeneralStateTest_stZeroCallsRevert
* [X] Replace/Update dependency sources:
    - [X] `parity-common` (Needs a release, tag, or branch other than master)
    - [X] `ring` (unclear if upstream works properly on Android)
    - [X] `jsonrpc-***` 
      - [X] awaiting merge on: https://github.com/paritytech/jsonrpc/pull/303
      - [X] awaiting merge on: https://github.com/paritytech/jsonrpc/pull/327
      - [X] need a tag, branch, or release to depend on (pinned to: https://github.com/paritytech/jsonrpc/commit/7a5977610fea105331ad578a5e2689776a0818a1)
    - [X] `cid`
* [X] Determine what to do with the `it_should_not_read_too_much_data_sync`
  test (`util/fetch/src/client.rs:807`). This can be left alone and deferred to a future PR.
* [X] Address intermittent stalling/timeouts in `secret_store` tests.
* [X] Address issues with fetch client.

# Related Upcoming TODOs (separate PR)

* Convert many boxed future return types to `impl Future` and clean up extra types.
* Ensure no errors are discarded in `secret_store/src/key_server_cluster/cluster.rs` (and elsewhere).
* Remove or modify the `it_should_not_read_too_much_data_sync` test (`util/fetch/src/client.rs`).

Closes #9533
Closes #9722


Cargo.toml
```diff
@@ -140,4 +140,7 @@ members = [
 ]
 
 [patch.crates-io]
-ring = { git = "https://github.com/paritytech/ring" }
+# ring = { git = "https://github.com/paritytech/ring" }
+ring = { git = "https://github.com/briansmith/ring" }
+parity-crypto = { git = "https://github.com/c0gent/parity-common", branch = "ring-0.13" }
```
ordian: parity-crypto = "0.2"

Cargo.toml
```diff
@@ -140,4 +140,7 @@ members = [
 ]
 
 [patch.crates-io]
-ring = { git = "https://github.com/paritytech/ring" }
+# ring = { git = "https://github.com/paritytech/ring" }
+ring = { git = "https://github.com/briansmith/ring" }
```
ordian: we don't need this

ipfs/src/error.rs
```diff
@@ -95,3 +95,15 @@ impl From<ServerError> for String {
...
 }
+
+impl ::std::fmt::Display for ServerError {
+	fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl ::std::error::Error for ServerError {
+	fn description(&self) -> &str { "ServerError" }
```
tomaka: The methods of the `Error` trait are deprecated. Instead, it's the implementation of `Display` that should write `ServerError`.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -59,7 +59,7 @@ const KEEP_ALIVE_SEND_INTERVAL: Duration = Duration::from_secs(30);
 const KEEP_ALIVE_DISCONNECT_INTERVAL: Duration = Duration::from_secs(60);
 
 /// Empty future.
-pub type BoxedEmptyFuture = Box<Future<Item = (), Error = ()> + Send>;
+pub type BoxedEmptyFuture = Box<Future<Item = (), Error = ()> + Send + 'static>;
```
tomaka: Nit: change shouldn't be needed.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -235,15 +233,15 @@ pub struct Connection {
...
 		let listen_address = make_socket_address(&config.listen_address.0, config.listen_address.1)?;
 		let connections = ClusterConnections::new(&config)?;
 		let servers_set_change_creator_connector = connections.connector.clone();
 		let sessions = ClusterSessions::new(&config, servers_set_change_creator_connector);
-		let data = ClusterData::new(&handle, config, connections, sessions);
+		let data = ClusterData::new(&executor, config, connections, sessions);
 
 		Ok(Arc::new(ClusterCore {
-			handle: handle,
+			// executor: executor,
```
tomaka: Commented out code.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1241,18 +1247,25 @@ pub mod tests {
...
-			if predicate() {
-				break;
-			}
 
-			if Instant::now() - start > timeout {
-				panic!("no result in {:?}", timeout);
-			}
-		}
+		let check_predicate = future::loop_fn((), move|_| {
+				if predicate() {
```
tomaka: Wrong indentation.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1119,7 +1117,15 @@ pub mod tests {
 	use std::time::{Duration, Instant};
 	use std::collections::{BTreeSet, VecDeque};
 	use parking_lot::Mutex;
-	use tokio_core::reactor::Core;
+	use tokio::{
+		// reactor::Reactor,
```
tomaka: Commented out code.

secret_store/src/key_server_cluster/io/deadline.rs
```diff
@@ -16,18 +16,28 @@
...
 
 /// Complete a passed future or fail if it is not completed within timeout.
-pub fn deadline<F, T>(duration: Duration, handle: &Handle, future: F) -> Result<Deadline<F>, io::Error>
-	where F: Future<Item = T, Error = io::Error> + Send + 'static, T: 'static {
-	let timeout: DeadlineBox<F> = Box::new(Timeout::new(duration, handle)?.map(|_| DeadlineStatus::Timeout));
-	let future: DeadlineBox<F> = Box::new(future.map(DeadlineStatus::Meet));
+pub fn deadline<F, T>(duration: Duration, future: F) -> Result<Deadline<F>, io::Error>
+	where F: Future<Item = T, Error = io::Error> + Send + 'static, T: Send + 'static
+{
+	// let timeout: DeadlineBox<F> = Box::new(Timeout::new((), duration)?.map(|_| DeadlineStatus::Timeout));
```
tomaka: Commented out code.

ipfs/src/error.rs
c0gent: Indeed, (and I'm inclined to move errors towards `failure` depending on what happens with the new std error trait) but `std::error::Error` is required when [building a hyper server](https://docs.rs/hyper/0.12.10/hyper/server/struct.Builder.html#method.serve).

tomaka: `Error` itself is not deprecated, just implementing the methods on it.
You still should write `impl Error for ServerError {}`.

c0gent: Ah, my mistake. It's been a while since I've manually implemented it at all. :)

I thought maybe you were referring to potential future deprecation: https://github.com/rust-lang/rfcs/pull/2504.

Cargo.toml
```diff
@@ -31,7 +31,7 @@ futures = "0.1"
 futures-cpupool = "0.1"
 fdlimit = "0.1"
 ctrlc = { git = "https://github.com/paritytech/rust-ctrlc.git" }
-jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
+jsonrpc-core = { git = "https://github.com/c0gent/jsonrpc.git", branch = "c0gent-hyper" }
```
dvdplm: This will change back to `"https://github.com/paritytech/jsonrpc.git"` at some point before the PR lands right?

ethcore/stratum/Cargo.toml
```diff
@@ -8,14 +8,14 @@ authors = ["Parity Technologies <admin@parity.io>"]
 [dependencies]
 ethereum-types = "0.4"
 keccak-hash = "0.1"
-jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
-jsonrpc-macros = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
-jsonrpc-tcp-server = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
+jsonrpc-core = { git = "https://github.com/c0gent/jsonrpc.git", branch = "c0gent-hyper" }
+jsonrpc-macros = { git = "https://github.com/c0gent/jsonrpc.git", branch = "c0gent-hyper" }
+jsonrpc-tcp-server = { git = "https://github.com/c0gent/jsonrpc.git", branch = "c0gent-hyper" }
```
dvdplm: These will change back to "https://github.com/paritytech/jsonrpc.git" at some point before the PR lands right?

rpc/src/tests/ws.rs
```diff
@@ -34,10 +34,9 @@ pub fn serve() -> (Server<ws::Server>, usize, GuardedAuthCodes) {
 	let authcodes = GuardedAuthCodes::new();
 	let stats = Arc::new(informant::RpcStats::default());
 
-	let res = Server::new(|remote| ::start_ws(
+	let res = Server::new(|_remote| ::start_ws(
```
dvdplm: Why not just `Server::new(|_| …`?

rpc/src/v1/extractors.rs
```diff
@@ -216,26 +217,25 @@ impl<M: core::Middleware<Metadata>> WsDispatcher<M> {
...
+		core::FutureRpcResult<M::Future>,
 		core::FutureResponse,
 	>;
 
-	fn on_request<F, X>(&self, request: core::Request, meta: Metadata, process: F) -> Self::Future where
-		F: FnOnce(core::Request, Metadata) -> X,
-		X: core::futures::Future<Item=Option<core::Response>, Error=()> + Send + 'static,
+	fn on_request<F, X>(&self, request: core::Request, meta: Metadata, process: F)
+		-> Either<Self::Future, X>
+		where F: FnOnce(core::Request, Metadata) -> X,
```
dvdplm: Funny indentation here.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -137,7 +137,7 @@ pub struct ClusterConfiguration {
 	/// Administrator public key.
 	pub admin_public: Option<Public>,
 	/// Should key servers set change session should be started when servers set changes.
-	/// This will only work when servers set is configured using KeyServerSet contract.
+	/// This will only work when servers set is configured  using KeyServerSet contract.
```
dvdplm: Double spaces here. Also: s/when servers set is/when the server set is/

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -149,8 +149,8 @@ pub struct ClusterState {
 
 /// Network cluster implementation.
 pub struct ClusterCore {
-	/// Handle to the event loop.
-	handle: Handle,
+	// /// Handle to the event loop.
+	// executor: TaskExecutor,
```
dvdplm: commented out lines

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -293,53 +290,49 @@ impl ClusterCore {
...
+		Box::new(net_connect(&node_address, data.self_key_pair.clone(), disconnected_nodes)
 			.then(move |result| ClusterCore::process_connection_result(data, Some(node_address), result))
-			.then(|_| finished(())))
+			.then(|_| future::ok(())))
 	}
 
 	/// Start listening for incoming connections.
-	fn listen(handle: &Handle, data: Arc<ClusterData>, listen_address: SocketAddr) -> Result<BoxedEmptyFuture, Error> {
-		Ok(Box::new(TcpListener::bind(&listen_address, &handle)?
+	fn listen(data: Arc<ClusterData>, listen_address: SocketAddr) -> Result<BoxedEmptyFuture, Error> {
```
dvdplm: I haven't traced through all the calls but aren't we dropping a few possible errors in this method? Perhaps out of scope for this PR though...

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -803,12 +795,17 @@ impl ClusterData {
...
 
-	/// Spawns a future using thread pool and schedules execution of it with event loop handle.
-	pub fn spawn<F>(&self, f: F) where F: Future + Send + 'static, F::Item: Send + 'static, F::Error: Send + 'static {
-		let pool_work = self.pool.spawn(f);
-		self.handle.spawn(move |_handle| {
-			pool_work.then(|_| finished(()))
-		})
+	/// Spawns a future on the runtime.
+	//
+	// TODO: Consider implementing a more graceful shutdown process using an
```
dvdplm: +1

util/fetch/Cargo.toml
```diff
@@ -8,11 +8,11 @@ authors = ["Parity Technologies <admin@parity.io>"]
 
 [dependencies]
 futures = "0.1"
-hyper = "0.11"
-hyper-rustls = "0.11"
+hyper = "~0.12.9"
```
dvdplm: Do we really need to be this specific?

c0gent: I can't remember exactly what but I'm depending on something that was added in `0.12.9`.

Cargo.toml
c0gent: Absolutely. Currently just waiting on a specific release/branch/tag etc. to depend on (It's probably wise to hold off on that until this PR is confirmed to work properly).

rpc/src/tests/ws.rs
c0gent: No reason really. I like to label those unused variables just to make it easy to know what's there but I'm happy to change it.

rpc/src/v1/extractors.rs
c0gent: I'm a funny guy!

secret_store/src/key_server_cluster/cluster.rs
c0gent: Yes, the errors are not handled the way I would do it. I've tried not to change too much functionality from the original in this PR in order to keep things as straightforward as possible. I'll make a note to have a look at that in an upcoming PR :)

c0gent:
I haven't been able to spend time on it but I could still use help from someone more familiar with json_tests to give me some idea of where I might start looking for the reason why those are failing.

dvdplm:
re-json_tests: @sorpaas are you able to provide a few pointers here? :)

sorpaas:
@c0gent The issue looks like that you updated `ethereum/tests` submodule in this PR. Revert back that part should fix jsontests.

ethcore/Cargo.toml
```diff
@@ -20,7 +20,7 @@ hashdb = "0.2.1"
 memorydb = "0.2.1"
 patricia-trie = "0.2"
 patricia-trie-ethereum = { path = "../util/patricia-trie-ethereum" }
-parity-crypto = "0.1"
+parity-crypto = "0.2"
```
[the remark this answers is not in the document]
ordian: in `parity-crypto` 0.2 we only bumped ring to `0.13`, the problem is we can't compile both ring 0.12 and 0.13 due to https://github.com/briansmith/ring/blob/7e3682c02170f5eafa423b63bacfa018de2878a9/Cargo.toml#L14-L15

c0gent:
I'm re-running the tests and I have to leave for an appointment. I'll push the changes this afternoon (PST).

It will leave two items remaining (see OP), 1.) A jsonrpc tag/branch/release to depend on, 2.) Decide what to do about that test -- this can be deferred to a future PR.

cheme:
About the json_tests, as @sorpaas  point out, it is a submodule issue (I just run them successfully with your PR).
About jsonrpc, I ran the test with latest master dependencies, and got some failures, mainly related to secret_store keyserver:

```
failures:
    key_server_cluster::client_sessions::generation_session::tests::encryption_session_works_over_network
    key_server_cluster::cluster::tests::error_in_generation_session_broadcasted_to_all_other_nodes
    key_server_cluster::cluster::tests::generation_session_completion_signalled_if_failed_on_master
    key_server_cluster::cluster::tests::generation_session_is_removed_when_succeeded
    key_server_cluster::cluster::tests::sessions_are_removed_when_initialization_fails

```
Plus stratum 'can_push_work' also seem to fail, but mainly when running with other tests. 
But others than those 6 tests all seems to pass :)

c0gent:
Yup thanks. I'll fix all of that when I get back in a few hrs. It's just little stuff.

c0gent:
This PR is ready to go, modulo the two points noted above (see OP).

c0gent:
I'm still working on this. My theory is that the remaining test failures are due to bad assumptions made about the ordering of spawned futures during the original design of `secret_store`. I'm going to reset all of the changes I've made to that crate and reimplement the updates, better preserving the behavior of the original Tokio reactor (as much as is possible).

I realize that the recent discovery of a security vulnerability has put a rush on this PR but this may take me some time. Any other ideas on how to tackle this more quickly are very welcome.

c0gent:
Ok well I've found a solution for the `key_server_cluster` issues that doesn't require me to create a simulated `tokio_core` reactor.

I now realize that a few other components are still using hyper 0.11 which I'll now begin updating.

There's also apparently a problem compiling on Android which I'm not sure what to do about.

c0gent:
Looks like I got all of the hyper 0.11 deps out. (`rust-idg` depends on `tokio_core` but not `hyper`).

ordian:
> Looks like I got all of the hyper 0.11 deps out. (`rust-idg` depends on `tokio_core` but not `hyper`).

I don't think so: https://github.com/sbstp/rust-igd/blob/master/Cargo.toml#L24
also
```
❯ cargo tree -i -p hyper:0.11.27
hyper v0.11.27
└── igd v0.7.0
```

c0gent:
Whoops, my bad. For whatever reason though it's not listing hyper 0.11 in `Cargo.lock`.

c0gent:
And... it looks like my fix to `secret_store` still intermittently fails. 

Back to the drawing board.

c0gent:
I'm thinking the `secret_store` issue is due ultimately to deadlock issues with one/some of the mutexes or rwlocks. If so, the ordering upheld (though not guaranteed) by the tokio-core reactor might not have ever exposed them...

ordian:
@c0gent how elaborate is updating secret_store to hyper 0.12, maybe it'll fix the issue?

c0gent:
It's not too tough. I was planning to do it at a later time since it's only used for testing.

I'll go ahead and update it once I'm through trying a few other things out.

c0gent:
There's still some weird problem with `fetch` and the timer that I haven't completely sorted out...

secret_store/src/key_server_cluster/io/deadline.rs
```diff
@@ -16,18 +16,24 @@
...
+use tokio::timer::Timeout;
 
 type DeadlineBox<F> = Box<Future<Item = DeadlineStatus<<F as Future>::Item>, Error = <F as Future>::Error> + Send>;
 
 /// Complete a passed future or fail if it is not completed within timeout.
-pub fn deadline<F, T>(duration: Duration, handle: &Handle, future: F) -> Result<Deadline<F>, io::Error>
-	where F: Future<Item = T, Error = io::Error> + Send + 'static, T: 'static {
-	let timeout: DeadlineBox<F> = Box::new(Timeout::new(duration, handle)?.map(|_| DeadlineStatus::Timeout));
-	let future: DeadlineBox<F> = Box::new(future.map(DeadlineStatus::Meet));
+pub fn deadline<F, T>(duration: Duration, future: F) -> Result<Deadline<F>, io::Error>
```
svyatonik: Seems like logic has changed here: previously it was:
future completes with Ok before timeout => Ok(DeadlineStatus::Meet)
future completes with  Err before timeout => Err
future doesn't complete before timeout => Ok(DeadlineStatus::Timeout)

Now it always returns Ok

secret_store/src/key_server_cluster/io/deadline.rs
```diff
@@ -52,9 +58,9 @@ impl<F, T> Future for Deadline<F> where F: Future<Item = T, Error = io::Error> {
 
 	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
 		match self.future.poll() {
-			Ok(Async::Ready((result, _other))) => Ok(Async::Ready(result)),
+			Ok(Async::Ready(result)) => Ok(Async::Ready(result)),
```
svyatonik: `self.future.poll()` will be enough here

svyatonik: Probably worth removing `Deadline` at all and replace it with `Timeout` (not a part of this PR, though)

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1219,40 +1242,46 @@ pub mod tests {
...
+					Ok(Loop::Continue(()))
+				}
 			}
+		}).map_err(|_: ()| ());
 
-			if Instant::now() - start > timeout {
-				panic!("no result in {:?}", timeout);
-			}
-		}
+		runtime.block_on(check_predicate).unwrap();
```
svyatonik: afaiu, this will block all other futures execution if there's only one thread in `Runtime`' `ThreadPool` - `block_on` would just call `predicate()` until it timeouts. Probably worth specifying number of threads using `Builder` explicitly? Or just choose some other check mechanism.

secret_store/src/key_server.rs
```diff
@@ -191,18 +191,22 @@ impl KeyServerCore {
 		let (stop, stopped) = futures::oneshot();
 		let (tx, rx) = mpsc::channel();
 		let handle = thread::Builder::new().name("KeyServerLoop".into()).spawn(move || {
-			let mut el = match Core::new() {
+			let runtime_res = runtime::Builder::new()
+				.core_threads(4)
```
svyatonik: should be config.threads

secret_store/src/key_server_cluster/cluster.rs
svyatonik: @dvdplm Not sure I got it - all connection-establish (both for incoming && outcoming) related errors are processed in `process_connection_result` by tracing? Probably some other errors are dropped? Or do you have idea how to handle these errors in other way? Could you please, elaborate?

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -803,12 +804,28 @@ impl ClusterData {
...
+		} else {
+			error!("Secret store runtime unable to spawn task. Shutdown has been started.");
+		}
+	}
+
+	/// Sets the `is_shutdown` flag which prevents future tasks from being
+	/// spawned via `::spawn`.
+	#[cfg(test)]
+	pub fn shutdown(&self) {
+		self.is_shutdown.store(true, Ordering::Release);
```
svyatonik: What's the reason for this method? Something bad happened in tests?

secret_store/src/key_server_cluster/net/accept_connection.rs
```diff
@@ -19,20 +19,21 @@ use std::sync::Arc;
...
-use tokio_core::net::TcpStream;
+use tokio::net::TcpStream;
 use key_server_cluster::{Error, NodeKeyPair};
 use key_server_cluster::io::{accept_handshake, Handshake, Deadline, deadline};
 use key_server_cluster::net::Connection;
 
 /// Create future for accepting incoming connection.
-pub fn accept_connection(address: SocketAddr, stream: TcpStream, handle: &Handle, self_key_pair: Arc<NodeKeyPair>) -> Deadline<AcceptConnection> {
+pub fn accept_connection(stream: TcpStream, self_key_pair: Arc<NodeKeyPair>) -> Deadline<AcceptConnection> {
+	let address = stream.peer_addr().expect("Unable to determine tcp peer address");
```
svyatonik: Could you please add TODO to remove this expect in the future? Not sure that it can't fail here

secret_store/src/key_server_cluster/cluster.rs
svyatonik: I've been able to fix test-hangs-issue, but still not sure about the reason :/ Probably polling never-ending LoopFn causes some changes in other futures poll schedule - could be investigated separately.
So the fix is:
1. replace your version of `loop_until` with:
```rust
	pub fn loop_until<F>(runtime: &mut Runtime, timeout: Duration, predicate: F)
		where F: Send + 'static + Fn() -> bool
	{
		use futures::Stream;
		use tokio::timer::Interval;

		let start = Instant::now();

		runtime.block_on(Interval::new_interval(Duration::new(0, 1_000_000))
			.and_then(move |_| {
				if Instant::now() - start > timeout {
					panic!("no result in {:?}", timeout);
				}

				Ok(())
			})
			.take_while(move |_| future::ok(!predicate()))
			.for_each(|_| Ok(()))
			.then(|_| future::ok::<(), ()>(()))
		).unwrap();
	}
```
2. increase TIMEOUT constant to 1s (i.e. `const TIMEOUT: Duration = Duration::from_millis(1000);`) - probably new implementation of `loop_until` causes multiple thread-context switches because I got timeout several times in simplest `cluster_connects_to_other_nodes` test, even though it has been near-to-completion.

Ran `while reset && cargo test -p ethcore-secretstore --features secretstore; do :; done` for last hour without any issues (except single one - unrelated, will open a PR later today).

c0gent: Ok looks great. I'll make those adjustments.

Just to clarify, `Runtime::block_on` is basically equivalent to `Core::run`, `tokio::run` etc. It basically starts the runtime up. It should never be called from within a future, etc. so it wouldn't actually block anything else from happening, it just blocks the containing thread (which, again, should never be a future/task).

I'll explicitly set the thread pool count to 4 (which is a good round number to use in general) but if there are problems created by using other counts (like 1) it usually indicates some deeper problem (one of the futures is blocking the entire thread waiting on a mutex, etc.). As an aside, I have a crate, [Qutex](https://github.com/cogciprocate/qutex), which provides non-thread-blocking futures-aware mutexes which can be useful when dealing with IO and shared data.

secret_store/src/key_server_cluster/cluster.rs
c0gent: Basically, what was happening without that is that due to the recursive nature of task spawning, one task could be spawning new tasks as the runtime is shutting down, causing errors. Those errors sometimes caused tests to fail but usually didn't.

This probably shouldn't be just a test only thing but let's leave that for later.

secret_store/src/key_server_cluster/net/accept_connection.rs
c0gent: It could fail which probably means it would be better either to accept the address as a separate argument or return a result. I'll add a note to that effect.

secret_store/src/key_server_cluster/io/deadline.rs
c0gent: Yeah I probably did that as a temporary measure and forgot to go back and correct it. Double-check my corrections to make sure it looks ok.

c0gent:
The problem with the fetch client is due to the use of a `futures_cpupool` to spawn the fetch request. I was hoping to put this off until a future PR but I'll now need to go through and replace all of the `CpuPools` (or at least the ones which spawn anything with a timer) with a handle to the runtime or a `tokio_threadpool`.

ordian:
@tomusdrw is the plan to spin off a branch of jsonrpc with recent changes (perhaps `parity-2.1`) or to publish it to crates.io?

c0gent:
Updated to point to the latest commit on jsonrpc master.

util/runtime/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Tokio Core Reactor wrapper.
```
ordian: Should this doc be updated?
```suggestion
//! Tokio Runtime wrapper.
```

util/runtime/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Executor {
+	inner: Mode,
+}
+
+impl Executor {
+	/// Executor for existing runtime.
+	///
+	/// Deprecated: Exists only to connect with current JSONRPC implementation.
```
ordian: Maybe use deprecation attribute?
```suggestion
	#[deprecated(since = "0.1.0", note = "Exists only to connect with current JSONRPC implementation.")]

```

util/runtime/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+				});
+			},
+		}
+	}
+
+	/// Spawn a new future returned by given closure.
+	pub fn spawn_fn<F, R>(&self, f: F) where
+		F: FnOnce() -> R + Send + 'static,
+        R: IntoFuture<Item=(), Error=()> + Send + 'static,
+        R::Future: Send + 'static,
```
ordian: spaces

util/runtime/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+	pub fn with_thread_count(thread_count: usize) -> Self {
+		let mut runtime_bldr = TokioRuntimeBuilder::new();
+		runtime_bldr.core_threads(thread_count);
+
+		Self::new(&mut runtime_bldr)
+	}
+
+	/// Returns this runtime raw executor.
+	///
+	/// Deprecated: Exists only to connect with current JSONRPC implementation.
```
ordian: Maybe use deprecation attribute?
```suggestion
	#[deprecated(since = "0.1.0", note = "Exists only to connect with current JSONRPC implementation.")]

```

price-info/src/lib.rs
```diff
@@ -108,50 +108,56 @@ impl<F: Fetch> Client<F> {
...
-					.and_then(|value| value.pointer("/result/ethusd"))
-					.and_then(|obj| obj.as_str())
-					.and_then(|s| s.parse().ok());
-
-				match ethusd {
-					Some(ethusd) => {
-						set_price(PriceInfo { ethusd });
-						Ok(())
+			.then(move |body_res| {
+				match body_res {
```
tomaka: This could be replaced with `let body_res = body_res?;` to remove two levels of indentation.

tomaka:
Nothing bad to me, however I'm not sure so familiar with the threading model of Parity and JSONRPC in general.

c0gent:
Updated.

util/fetch/Cargo.toml
dvdplm: Seems to work with `hyper v0.12.11` (which is what I got after changing to `"0.12"` in Cargo.toml)

ipfs/src/error.rs
```diff
@@ -95,3 +95,11 @@ impl From<ServerError> for String {
 		}
 	}
 }
+
+impl ::std::fmt::Display for ServerError {
+	fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
```
niklasad1: Nitpick, Display and Debug provide the same imp.

More intuitive to have different impls IMO.

util/fetch/src/client.rs
```diff
@@ -291,7 +289,15 @@ impl Fetch for Client {
 			.and_then(|_| rx_res.map_err(|oneshot::Canceled| Error::BackgroundThreadDead))
 			.and_then(future::result);
 
-		Box::new(self.timer.timeout(future, maxdur))
+		Box::new(future.timeout(maxdur)
+			.map_err(|err| {
+				if err.is_inner() {
+					Error::from(err.into_inner().unwrap())
```
niklasad1: Handle the error instead of `unwrapping here`

Something like should work:

```rustc
if let Some(inner) = err.into_inner() {
   Error::from(inner)
} else {
   Error::TokioTimeoutInnerVal("".into())
}

```

Cargo.toml
```diff
@@ -28,10 +28,9 @@ serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
 futures = "0.1"
-futures-cpupool = "0.1"
 fdlimit = "0.1"
 ctrlc = { git = "https://github.com/paritytech/rust-ctrlc.git" }
-jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-1.11" }
+jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", rev = "7a59776" }
```
ordian: please update to `branch = "parity-2.2"`

5chdn: just testing this amazing new feature

```suggestion
jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc.git", branch = "parity-2.2" }
```

util/fetch/src/client.rs
```diff
@@ -585,23 +592,43 @@ impl fmt::Display for Error {
...
+			Error::TokioTimer(ref e) => write!(fmt, "tokio timer error: {:?}", e),
 			Error::Timeout => write!(fmt, "request timed out"),
 			Error::SizeLimit => write!(fmt, "size limit reached"),
 		}
 	}
 }
 
+impl ::std::error::Error for Error {
+	fn description(&self) -> &str { "Fetch client error" }
+    fn cause(&self) -> Option<&::std::error::Error> { None }
```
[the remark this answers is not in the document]
5chdn:
```suggestion
	fn cause(&self) -> Option<&::std::error::Error> { None }
```

5chdn:
Will merge once last grumbles are addressed

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -1219,40 +1241,49 @@ pub mod tests {
...
-			}
 
-			if Instant::now() - start > timeout {
-				panic!("no result in {:?}", timeout);
-			}
-		}
+		runtime.block_on(Interval::new_interval(Duration::new(0, 1_000_000))
+			.and_then(move |_| {
+				if Instant::now() - start > timeout {
+					panic!("no result in {:?}", timeout);
```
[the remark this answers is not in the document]
svyatonik: Could confirm that it looks OK to me (I've suggested this change, actually :) ). Tested both positive && negative cases - works OK.

secret_store/src/key_server_cluster/cluster.rs
```diff
@@ -165,7 +165,7 @@ pub struct ClusterClientImpl {
 
 /// Network cluster view. It is a communication channel, required in single session.
 pub struct ClusterView {
-	core: Arc<Mutex<ClusterViewCore>>,
+	core: Arc<RwLock<ClusterViewCore>>,
```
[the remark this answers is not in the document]
svyatonik: It doesn't break anything. I would also prefer not to mix different changes in single PR, but I guess these are from times when @c0gent tried to fix SS tests. Since this changes doesn't break anything + this PR is too old && big already, I'm OK with merging as-is.

util/fetch/src/client.rs
c0gent: The problem with doing it like that is that [this particular error](https://docs.rs/tokio/0.1.11/tokio/timer/timeout/struct.Error.html) basically has three variants, inner, timer, and elapsed. By calling `into_inner` first, you can't rely on `Error`'s `From` impl to determine which of the other two variants it is.

Basically this one's a bit tricky because in this case I want to use the `From` impl for the inner error type (to get the right variant), not the `timeout::Error` itself.

c0gent:
Should be all addressed.

secret_store/src/key_server_cluster/cluster.rs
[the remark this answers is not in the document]
svyatonik: Could you provide some deadlocked stacks? I haven't seen any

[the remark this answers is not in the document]
c0gent: I was going to add that the change you suggested may have redundantly fixed the problem. Should I switch back then?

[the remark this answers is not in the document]
c0gent: Remember the deadlocking was intermittent and it only seemed to happen on CI, which is why I left both changes just to be sure.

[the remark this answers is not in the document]
svyatonik: I haven't seen any deadlocked threads while working on this change, but I have seen that tokio has stopped waking up tasks. Have you seen some? If so - please share. I'm not insisting to revert this, as I said, but am warned about deadlocking.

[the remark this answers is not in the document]
svyatonik: Hanging tests were easy reproducible locally on my laptop, but again - I haven't seen any deadlocks in SS code (related to `Mutex` -> `RwLock` changes). So I'm not sure this change is about deadlock-fix - just possible change (we could even call it optimization :) ).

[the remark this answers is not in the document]
c0gent: Well should we just call it an optimization and leave it be?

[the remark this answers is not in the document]
svyatonik: Yes - let's leave as is

niklasad1:
@tomusdrw are you ok with this now? if that's the case let's merge it

tomusdrw:
Looks good but the tests are failing and some of the failures seems legit. @5chdn is it a known issue or something specific to this PR?

5chdn:
Failures are unrelated. Thanks, everyone.

5chdn:
Failures are unrelated. Thanks, everyone.

5chdn:
Merged 4x :man_facepalming:

PR #9658, opened 2018-09-27
Test fix for windows cache name...

ddorgan:


ddorgan:
See job: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/101262 

You now see: Checking cache for test-windows-rust-stable-17...

Instead of: Checking cache for %CI_JOB_NAME%-17

.gitlab-ci.yml
```diff
@@ -100,7 +100,7 @@ test-android-rust-stable:
 test-windows-rust-stable:
   stage:                           test
   cache:
-    key:                           "%CI_JOB_NAME%"
+    key:                           "${CI_JOB_NAME}"
     paths:
       - "%CI_PROJECT_DIR%/target/"
       - "%CI_PROJECT_DIR%/cargo/"
```
5chdn: Do we need to change the notation everywhere from `%VAR%` to `${VAR}` ?

ddorgan:
The other locations where the %VAR% notation is used is in the path locations for the caches which seem to work fine (but with the wrong cache name).

```
_____ Calculating checksums _____
Creating cache %CI_JOB_NAME%-17...
c:/r/builds/f8bd3cf5/0/parity/parity-ethereum/target/: found 4273 matching files 
c:/r/builds/f8bd3cf5/0/parity/parity-ethereum/cargo/: found 10759 matching files 
```

I'll check if we can move the windows variables to linux like ones to keep things consistent.

.gitlab-ci.yml
kirushik: As far as I understand — no, since scripts for caching are evaluated in linux environment (on a Gitlab server side) when we convert yaml into an actual executable; but _paths_ are expanded on the target machine when it's actually executed.

But I might as well be mistaken

PR #9663, opened 2018-09-27
Update a few parity-common dependencies

dvdplm:
Update `parity-ethereum` to use the latest hashdb, rlp, memorydb and patricia_trie crates. While the PR touch many files, the relevant changes are:

1. Change the RLP return type from `ElasticArrayXXX` to `Vec<u8>`
1. As hashdb is now generic over the DB data type, `hashdb::DBValue` is gone and replaced with `kvdb::DBValue` (wrapper for `ElasticArray128<u8>` for the time being)
1. The hashdb refactorings bubble up in memorydb as well

Related to https://github.com/paritytech/parity-common/pull/67

A (rather big) caveat to this PR is that recent correctness fixes in `rlp` broke tests in `util/network-devp2p`. The commits are:

1. https://github.com/paritytech/parity-common/pull/60 which breaks [`util/network-devp2p/src/discovery.rs`](https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/discovery.rs#L1068)
1. https://github.com/paritytech/parity-common/pull/59 which breaks 4 handshake tests: [1](https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/discovery.rs#L1068), [2](https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/handshake.rs#L411), [3](https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/handshake.rs#L461), [4](https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/handshake.rs#L490)

Work is under way to fix the errors and should not stop reviewers.


ethcore/light/Cargo.toml
```diff
@@ -12,16 +12,16 @@ ethcore = { path = ".."}
 parity-bytes = "0.1"
 ethcore-transaction = { path = "../transaction" }
 ethereum-types = "0.4"
-memorydb = "0.2.1"
-patricia-trie = "0.2"
+memorydb  = "0.3.0-beta"
```
NikVolf: why such strange versioning?

versions below 1.0 better to name just numerically, they are all betas

util/network/src/error.rs
```diff
@@ -210,7 +205,7 @@ fn test_errors() {
 	assert_eq!(DisconnectReason::Unknown, r);
 
 	match *<Error as From<rlp::DecoderError>>::from(rlp::DecoderError::RlpIsTooBig).kind() {
-		ErrorKind::Auth => {},
```
dvdplm: Oh nice; I was pretty baffled by the `Auth` error, didn't really make sense.

ethcore/light/Cargo.toml
dvdplm: To be clear: once https://github.com/paritytech/parity-common/pull/67 is reviewed but before it lands, I'll fix the versions and publish the proper `0.3.0` versions.

dvdplm:
@debris unfortunately it looks like we now have a new error in `ethcore`. :/

debris:
I'm working on it

ethcore/sync/src/chain/handler.rs
```diff
@@ -745,9 +745,7 @@ mod tests {
 
 		let block = Rlp::new(&block_data);
 
-		let result = SyncHandler::on_peer_new_block(&mut sync, &mut io, 0, &block);
-
-		assert!(result.is_ok());
+		SyncHandler::on_peer_new_block(&mut sync, &mut io, 0, &block).expect("result to be ok");
```
dvdplm: I think I'd prefer using an assert here like before: it's easier to spot what the test is doing if I can scan the code and look for the `assert`. Just a personal preference, but yeah, please consider it.

ethcore/sync/src/chain/mod.rs
```diff
@@ -1183,7 +1183,7 @@ pub mod tests {
 	}
 
 	pub fn get_dummy_blocks(order: u32, parent_hash: H256) -> Bytes {
-		let mut rlp = RlpStream::new_list(1);
+		let mut rlp = RlpStream::new_list(2);
```
dvdplm: So this was it? Our test was wrong the whole time? :)

ethcore/sync/src/chain/handler.rs
debris: the problem with `assert!(result.is_ok()` is that it gives no information what went wrong. It will report only `expected true to be equal false`, whereas `expect` will use `fmt::Debug` to print unwrapped error

ethcore/sync/src/chain/mod.rs
debris: yes :(

ethcore/sync/src/chain/handler.rs
dvdplm: Gotcha, didn't think about that. Fine by me then.

util/journaldb/src/overlaydb.rs
```diff
@@ -112,6 +112,7 @@ impl OverlayDB {
 							return Err(error_negatively_reference_hash(&key));
 						}
 						let payload = Payload::new(total_rc as u32, x.value);
+						// let payload = Payload::new(total_rc as u32, x.value.clone());
```
ordian: commented out code

dvdplm:
@ordian sure, good idea. Feel free to review anyway ofc.

ethcore/Cargo.toml
```diff
@@ -16,9 +16,9 @@ crossbeam = "0.3"
 ethash = { path = "../ethash" }
 ethcore-bloom-journal = { path = "../util/bloom" }
 parity-bytes = "0.1"
-hashdb = "0.2.1"
-memorydb = "0.2.1"
-patricia-trie = "0.2"
+hashdb  = "0.3.0"
```
ordian: why double space before `=` here and below?

dvdplm: Because I can't grep so well. Sigh. Fixing.

PR #9664, opened 2018-09-27
Remove caching for osx tests.

ddorgan:
Some jobs are spending longer on caching than the actual running of tests, e.g. https://gitlab.parity.io/parity/parity-ethereum/-/jobs/101206

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -56,4 +56,4 @@ cd parity-clib-examples/cpp && \
 # Running tests
 echo "________Running Parity Full Test Suite________"
 git submodule update --init --recursive
-time cargo test  $OPTIONS --features "$FEATURES" --all $1 -- --test-threads 8
+time cargo test  $OPTIONS --features "$FEATURES" --all $1 -- --test-threads 4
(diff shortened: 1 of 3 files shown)


ddorgan:
Not looking great so far. Will need to re-test this with a new system to rule out this being an issue with the disks on our existing runners.

test.sh
```diff
@@ -56,4 +56,4 @@ cd parity-clib-examples/cpp && \
 # Running tests
 echo "________Running Parity Full Test Suite________"
 git submodule update --init --recursive
-time cargo test  $OPTIONS --features "$FEATURES" --all $1 -- --test-threads 8
+time cargo test  $OPTIONS --features "$FEATURES" --all $1 -- --test-threads 4
```
5chdn: why's that?

ddorgan: I want to time this to check if the builds went faster. Right now the load average is about 16, which is double the amount of CPU's available. Was hoping reducing the amount of threads would cause less resource contention. But the times look about the same :/

5chdn:
Unfortunately, I think this does not solve any of our issues, also conflicts with #9681

PR #9670, opened 2018-09-28
Verify block syncing responses against requests

jimpo:
This uses the `asking_hash` and `asking_blocks` fields set on the peer before making a request to ensure their responses are expected and valid. One major issue I noticed is clients with corrupted chains returning a sequence of block headers that do not connect with one another. The additional validation on block headers especially has improved the reliability of ancient block syncing in my testing.

### DIFF
--- a/ethcore/src/client/test_client.rs
+++ b/ethcore/src/client/test_client.rs
@@ -242,6 +242,7 @@ impl TestBlockChainClient {
 		*self.error_on_logs.write() = val;
 	}
 
+	/// Add a block to test client.
 	pub fn add_block<F>(&self, with: EachBlockWith, hook: F)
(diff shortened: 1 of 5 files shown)


jimpo:
@ascjones Thanks for the review. This PR is complimentary to #9531 and definitely does not fix all of the issues that that one addresses. I believe it mostly does not conflict in a way that will complicate merging. There are some other changes in [this branch](https://github.com/jimpo/parity/tree/block-ancient-sync) that conflict more directly that we can look into after your PR is merged.

ordian:
@jimpo Thanks for your awesome contributions! Could you rebase on master to include recent CI changes?

5chdn:
(sorry for the CI mess, once rebased we can merge this)

PR #9671, opened 2018-09-28
Tests modification for windows CI

cheme:
Running tests on windows, fixing some path to string conversion and others windows specifics :
- change of duration value in `corpus_inaccessible` because of substract overflow happening on windows
- right color configuration for `test_command_signer_new_token`
- extend delay for keyserver test.
- deactivate `guards_delete_folder` due to error on removal when guards is dropped
- deactivate `should_allow_if_authorization_is_correct` (timeout error is thrown)

At the time those changes allows the test to run on windows 10 vm, but on the windows CI there is still a segfault on specific tests (probably rocksdb related or related to temp path issue) :
```
test snapshot::service::tests::cannot_finish_with_invalid_chunks ... error: test failed, to rerun pass '--lib'
/c/r/builds/f8bd3cf5/0/parity/parity-ethereum/test.sh: line 59:  2224 Segmentation fault      cargo test $OPTIONS --features "$FEATURES" --all $1 -- --test-threads 1

```


ethstore/src/accounts_dir/vault.rs
```diff
@@ -310,8 +310,10 @@ mod test {
 
 	#[test]
 	fn make_vault_dir_path_succeeds() {
-		assert_eq!(make_vault_dir_path("/home/user/parity", "vault", true).unwrap().to_str().unwrap(), "/home/user/parity/vault");
-		assert_eq!(make_vault_dir_path("/home/user/parity", "*bad-name*", false).unwrap().to_str().unwrap(), "/home/user/parity/*bad-name*");
+    use std::path::Path;
```
NikVolf: spaces here

ethcore/light/src/cache.rs
```diff
@@ -186,7 +186,7 @@ mod tests {
 
 		{
 			let corpus_time = &mut cache.corpus.as_mut().unwrap().1;
-			*corpus_time = *corpus_time - Duration::from_secs(6 * 3600);
+			*corpus_time = *corpus_time - Duration::from_secs(5 * 3600);
```
debris: why is this change needed?

ethcore/src/snapshot/tests/service.rs
```diff
@@ -104,6 +104,11 @@ fn restored_is_equivalent() {
 	}
 }
 
+
+// on windows the guards deletion (remove_dir_all)
+// is not happening (error directory is not empty).
+// So the test is disabled until windows api behave.
+#[cfg(not(target_os = "windows"))]
```
debris: Can you tell something more about the error returned? Why is it not happening? According to the [documentation](https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html)

> Removes a directory at this path, after removing all its contents. Use carefully!

It should happen

ethcore/src/views/mod.rs
```diff
@@ -31,11 +31,11 @@ pub use self::transaction::TransactionView;
...
-    use super::HeaderView;
+	use super::HeaderView;
 
-    #[test]
-    #[should_panic(expected="View rlp is trusted and should be valid. Constructed in ethcore/src/views/mod.rs on line 39: RlpExpectedToBeList")]
-    fn should_include_file_line_number_in_panic_for_invalid_rlp() {
-        let _ = view!(HeaderView, &[]).parent_hash();
-    }
+	#[test]
+	#[should_panic]
```
debris: why did you remove the panic message?

rpc/src/tests/ws.rs
```diff
@@ -105,6 +105,7 @@ mod testing {
 		http_client::assert_security_headers_present(&response.headers, None);
 	}
 
+	#[cfg(not(target_os = "windows"))]
```
debris: why is this test ignored? :)

ethstore/src/accounts_dir/vault.rs
```diff
@@ -310,8 +310,10 @@ mod test {
 
 	#[test]
 	fn make_vault_dir_path_succeeds() {
-		assert_eq!(make_vault_dir_path("/home/user/parity", "vault", true).unwrap().to_str().unwrap(), "/home/user/parity/vault");
-		assert_eq!(make_vault_dir_path("/home/user/parity", "*bad-name*", false).unwrap().to_str().unwrap(), "/home/user/parity/*bad-name*");
+		use std::path::Path;
+
+		assert_eq!(&make_vault_dir_path("/home/user/parity", "vault", true).unwrap(), &Path::new("/home/user/parity/vault"));
+		assert_eq!(&make_vault_dir_path("/home/user/parity", "*bad-name*", false).unwrap(), &Path::new("/home/user/parity/*bad-name*"));
```
debris: 👍

parity/configuration.rs
```diff
@@ -1860,13 +1860,15 @@ mod tests {
...
+		use std::path;
+
 		let std = parse(&["parity"]);
 		let base = parse(&["parity", "--base-path", "/test"]);
 
 		let base_path = ::dir::default_data_path();
 		let local_path = ::dir::default_local_path();
 		assert_eq!(std.directories().cache, dir::helpers::replace_home_and_local(&base_path, &local_path, ::dir::CACHE_PATH));
-		assert_eq!(base.directories().cache, "/test/cache");
+		assert_eq!(path::Path::new(&base.directories().cache), path::Path::new("/test/cache"));
```
debris: 👍

ethcore/light/src/cache.rs
cheme: It substracts overflow otherwhise (of course that is only happening on windows). That looks like a windows bug/inconsistency with duration.

debris: 🤦‍♂️ please describe it in the comment

ethcore/src/views/mod.rs
cheme: The windows panic message contains '\' instead of '/'  in the file path of the error message. Another option could be to have a copy of this tests for windows.

rpc/src/tests/ws.rs
cheme: I did not identify the root of the issue, the test is failing on a connection timeout.

ethcore/src/snapshot/tests/service.rs
cheme: The error is a simple 'directory is not empty', looking at https://github.com/rust-lang/rust/issues/29497 which exists since 2015, I may have been a bit optimistic in saying that there will be a fix someday, we may probably want to use  https://crates.io/crates/remove_dir_all .

cheme: Ok, I ran a test with the 'remove_dir_all' crate, and it is not better : we obtain an "os 32" error "the process cannot access this file because it is used by another process". Solving it will be probably a bit harder than expected, it should probably have its own issue and PR.

5chdn:
what's the status of this?

cheme:
I think there is valid fixes for windows test that should be merged, there is also two skip test, so before merging a issue should be created (one is about directory deletion failing on windows and is not that trivial, the other one is not identified).
With this PR I could run test on a windows 10 VM, but the integration seems still broken : to see the issue in integration you need to change test.sh to run on one thread and you see a 'segfault' error. This segfault may be related to ci envt or related to #9702 . 
For this PR I am in favor of merging it and creating another issue for the directory deletion test and maybe for the ci segfault in devops.

5chdn:
Could you review this @ordian @ascjones ?

5chdn:
Merging, it's Windows ;)

PR #9673, opened 2018-09-28
Calculate sha3 instead of sha256 for push-release.

### LINKED ISSUE #9668: Different hashing algorithms in updater and building script
- **Parity Ethereum version**: 2.2.0-nightly
- **Fully synchronized**: no
- **Network**: POA Sokol
- **Restarted**: yes

We have contracts and auto updater server configured for the POA Sokol Network, it worked with the old version of CI. Now after building it pushes a [sha256 hash](https://github.com/paritytech/parity-ethereum/blob/93e1040d07e385d1219d00af71c46c720b0a1acf/scripts/gitlab/build-unix.sh#L34) of binary to the Operations contract.
But updater calculates a `sha3` hash of binary and it doesn't match. Here is log from our test build:
```
parity_updater::updater  Unable to fetch update (2.2.0-nightly-0x44c4…21c4): HashMismatch { expected: 0x38db89616c5776c033465d6b7d6401bfab7412e47c1c8db4a2a95e8a8704987b, got: 0x06c38cfb06e9d6d33fba852167c7ed51e45fc4597bd66cc5517e3daeae76a566 }, retrying in 8 seconds.
```
Tried to use then `parity tools hash` in the build script , this way it works.

For reproducing you will probably need to add a push-release request to the auto updater server.

tomusdrw:
Closes #9668

Don't really know how to test this, so will be good if all our bash magicians could have a look.

### DIFF
--- a/scripts/gitlab/push.sh
+++ b/scripts/gitlab/push.sh
@@ -35,9 +35,10 @@ do
   for binary in $(ls parity.sha256)
   do
     sha256=$(cat $binary | awk '{ print $1}' )
+    sha3=$(cat ${binary/sha256/sha3} | awk '{ print $1}' )
     case $DIR in
       x86_64* )
-        DATA="commit=$CI_BUILD_REF&sha3=$sha256&filename=parity$WIN&secret=$RELEASES_SECRET"
+        DATA="commit=$CI_BUILD_REF&sha3=$sha3&filename=parity$WIN&secret=$RELEASES_SECRET"
         ../../scripts/gitlab/safe_curl.sh $DATA "http://update.parity.io:1337/push-build/$CI_BUILD_REF_NAME/$DIR"
         # Kovan
         ../../scripts/gitlab/safe_curl.sh $DATA "http://update.parity.io:1338/push-build/$CI_BUILD_REF_NAME/$DIR"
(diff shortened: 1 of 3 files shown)


natlg:
You will need probably to add push release step, [this part](https://github.com/paritytech/parity-ethereum/blob/7abe9ec4ccb3136052cea56f36022c3f8e9c0fd6/scripts/gitlab-push-release.sh#L6-L12) seems missed.
Not sure where it's supposed to be now, we added [here](https://github.com/paritytech/parity-ethereum/blob/984493db30f4ba9a078861d979a5f8d3cc26402a/scripts/gitlab/push.sh#L8) for our network and it worked.

5chdn:
@natlg is correct from what I can tell

scripts/gitlab/push.sh
```diff
@@ -2,11 +2,17 @@
...
-  echo "push release"
-  # Mainnet
 
-}
+
+echo "__________Register Release__________"
+DATA="secret=$RELEASES_SECRET"
+
+echo "Pushing release to Mainnet"
+../../scripts/gitlab/safe_curl.sh $DATA "http://update.parity.io:1337/push-release/$CI_BUILD_REF_NAME/$CI_BUILD_REF"
```
natlg: Thank you for the fix! Just the path should be `./scripts/gitlab/safe_curl.sh` here (same for the line 14)

5chdn: wish we had someone at parity who knows bash :man_facepalming: k

PR #9679, opened 2018-10-01
ci: fix push script

ordian:
Fixes https://github.com/paritytech/parity-ethereum/pull/9673#discussion_r221478221, thanks @natlg!

I guess `../..` was copied from lines below, but it was used in the context of `cd artifacts`, `cd $DIR`.

### DIFF
--- a/scripts/gitlab/build-unix.sh
+++ b/scripts/gitlab/build-unix.sh
@@ -26,7 +26,11 @@ mkdir -p artifacts
 cd artifacts
 mkdir -p $CARGO_TARGET
 cd $CARGO_TARGET
-cp ../../target/$CARGO_TARGET/release/{parity,parity-evm,ethstore,ethkey,whisper} .
+cp ../../target/$CARGO_TARGET/release/parity ./parity
+cp ../../target/$CARGO_TARGET/release/parity-evm ./parity-evm
+cp ../../target/$CARGO_TARGET/release/ethstore ./ethstore
+cp ../../target/$CARGO_TARGET/release/ethkey ./ethkey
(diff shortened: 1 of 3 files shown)


scripts/gitlab/build-unix.sh
```diff
@@ -26,7 +26,11 @@ mkdir -p artifacts
 cd artifacts
 mkdir -p $CARGO_TARGET
 cd $CARGO_TARGET
-cp ../../target/$CARGO_TARGET/release/{parity,parity-evm,ethstore,ethkey,whisper} .
+cp ../../target/$CARGO_TARGET/release/parity ./parity
+cp ../../target/$CARGO_TARGET/release/parity-evm ./parity-evm
+cp ../../target/$CARGO_TARGET/release/ethstore ./ethstore
+cp ../../target/$CARGO_TARGET/release/ethkey ./ethkey
+cp ../../target/$CARGO_TARGET/release/whisper ./whisper
```
tomusdrw: This should fix random paths on ARM build (I guess there is some crazy version of `cp` out there)

scripts/gitlab/build-windows.sh
```diff
@@ -31,7 +31,7 @@ echo "_____ Calculating checksums _____"
 for binary in $(ls)
 do
   rhash --sha256 $binary -o $binary.sha256
-  parity.exe tools hash $binary > $binary.sha3
+  ./parity.exe tools hash $binary > $binary.sha3
```
tomusdrw: `parity.exe` is for sure in the correct directory, but I guess it has to be invoked in a unix-like fashion

PR #9683, opened 2018-10-01
removed redundant clone before each block import

debris:


ethcore/src/client/client.rs
```diff
@@ -1405,15 +1405,14 @@ impl ImportBlock for Client {
 			bail!(EthcoreErrorKind::Block(BlockError::UnknownParent(unverified.parent_hash())));
 		}
 
-		let raw = unverified.bytes.clone();
```
debris: redundant block clone

ethcore/src/client/client.rs
```diff
@@ -1405,15 +1405,14 @@ impl ImportBlock for Client {
...
 		match self.importer.block_queue.import(unverified) {
 			Ok(res) => Ok(res),
 			// we only care about block errors (not import errors)
-			Err(EthcoreError(EthcoreErrorKind::Block(err), _))=> {
-				self.importer.bad_blocks.report(raw, format!("{:?}", err));
+			Err((block, EthcoreError(EthcoreErrorKind::Block(err), _)))=> {
```
ordian: nit: space before `=>`

debris:
> Another approach would be to move let raw = unverified.bytes.clone(); under bad_block reporting match, it would be ok, since it's not a common path.

@ordian you cannot do that cause `unverified` is already moved to a different context

PR #9686, opened 2018-10-01
disable cpp lib test for mac, win and beta

gabreal:


5chdn:
does this make sense with 0ababae360033a6c782cb97e48a1ddb790507113 ?

PR #9687, opened 2018-10-02
log(light-provider) - more verbose logs

niklasad1:
Help for debugging the light-provider interface


ethcore/light/src/provider.rs
```diff
@@ -143,67 +143,87 @@ pub trait Provider: Send + Sync {
...
 			::request::AccountResponse {
 				proof,
 				nonce: acc.nonce,
 				balance: acc.balance,
 				code_hash: acc.code_hash,
 				storage_root: acc.storage_root,
 			}
-		})
+		});
+		trace!(target: "pip_provider", "account_proof: {:?}", account_proof);
```
niklasad1: Well, I don't think is necessary to print out all the raw bytes here so I will change these to be more "human friendly"

5chdn:
Please reopen when you are ready.

PR #9688, opened 2018-10-02
Don't hash the init_code of CREATE.

tomusdrw:
It's usually wasteful to cache it anyway, since it's rarely re-used. Also for large init codes it hashing might be slower then jump destination computation anyway.

Similar PR for geth: https://github.com/ethereum/go-ethereum/pull/17806

### DIFF
--- a/ethcore/evm/src/interpreter/shared_cache.rs
+++ b/ethcore/evm/src/interpreter/shared_cache.rs
@@ -50,17 +50,22 @@ impl SharedCache {
 	}
 
 	/// Get jump destinations bitmap for a contract.
-	pub fn jump_destinations(&self, code_hash: &H256, code: &[u8]) -> Arc<BitSet> {
-		if code_hash == &KECCAK_EMPTY {
-			return Self::find_jump_destinations(code);
-		}
+	pub fn jump_destinations(&self, code_hash: &Option<H256>, code: &[u8]) -> Arc<BitSet> {
+		if let Some(ref code_hash) = code_hash {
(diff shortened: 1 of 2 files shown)


ethcore/evm/src/interpreter/shared_cache.rs
```diff
@@ -50,17 +50,22 @@ impl SharedCache {
...
-		}
+	pub fn jump_destinations(&self, code_hash: &Option<H256>, code: &[u8]) -> Arc<BitSet> {
+		if let Some(ref code_hash) = code_hash {
+			if code_hash == &KECCAK_EMPTY {
+				return Self::find_jump_destinations(code);
+			}
 
-		if let Some(d) = self.jump_destinations.lock().get_mut(code_hash) {
-			return d.0.clone();
+			if let Some(d) = self.jump_destinations.lock().get_mut(code_hash) {
```
sorpaas: Not caused by this PR, but I think here we can use `get` instead of `get_mut`?

holiman:
>  but per-instance local cache

That's a pretty good idea, actually. Probably a must-have, even

holiman:
I added that to the geth PR. I originally intended to have it, but forgot it :)

ethcore/evm/src/interpreter/shared_cache.rs
tomusdrw: It's a `MemoryLruCache`, `get_mut` is used to update the LRU list.

PR #9694, opened 2018-10-03
Implement CREATE2 gas changes and fix some potential overflowing

sorpaas:
rel https://github.com/ethereum/go-ethereum/pull/17812

This also changes SHA3 and memcpy costs to check overflowing. Currently it doesn't matter only because the unit gas cost is less than 32, but this restriction can be easily forgotten.

### DIFF
--- a/ethcore/evm/src/interpreter/gasometer.rs
+++ b/ethcore/evm/src/interpreter/gasometer.rs
@@ -176,8 +176,7 @@ impl<Gas: evm::CostType> Gasometer<Gas> {
 				Request::GasMem(default_gas, mem_needed(stack.peek(0), stack.peek(1))?)
 			},
 			instructions::SHA3 => {
-				let w = overflowing!(add_gas_usize(Gas::from_u256(*stack.peek(1))?, 31));
-				let words = w >> 5;
+				let words = overflowing!(to_word_size(Gas::from_u256(*stack.peek(1))?));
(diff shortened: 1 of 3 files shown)


sorpaas:
Hmm apparently gas cost changes break CREATE2 jsontests... We'll either need to disable them or wait for upstream fixes.

ethcore/evm/src/interpreter/gasometer.rs
```diff
@@ -232,8 +232,20 @@ impl<Gas: evm::CostType> Gasometer<Gas> {
 				Request::GasMemProvide(gas, mem, Some(requested))
 			},
 			instructions::CREATE | instructions::CREATE2 => {
-				let gas = Gas::from(schedule.create_gas);
-				let mem = mem_needed(stack.peek(1), stack.peek(2))?;
+				let start = stack.peek(1);
+				let len = stack.peek(2);
+
+				let gas = match instruction {
```
tomusdrw: might be better to just split `instructions::CREATE` and `instructions::CREATE2` in the upper match instead of matching twice.

tomusdrw:
IMHO Let's ignore `CREATE2` testcases for now and get that merged.

cheme:
For the test to skip, https://github.com/cheme/parity-ethereum/blob/e573a5d30c97dfcd8acee39889cf99af24d13cfb/ethcore/res/ethereum/tests-issues/currents.json#L369 might do it.

PR #9705, opened 2018-10-05
[fix] Incompatible match arms in parity-clib

dvdplm:
This error only shows up when building with `--all-features` so the `jni` code is built.


dvdplm:
> Basically, this changes the behavior if we get a "malformed" CLI argument we return an error (returns early) instead of calling parity?

Well, hm, without this the code doesn't compile (match arms differ in return type: the first is `()` and the second was `jlong`.
I suspect this was a just a typo inspired by the `match` just below it that also raises a java exception and returns `0`. TBH I don't know what `0` means in this context, it can't really be the shell "all good" retval right?

tomaka:
Throwing an exception short-circuits everything, so the actual return value doesn't matter.

Issue #9408, opened 2018-08-24
Parity-evm: handle failing statetests


holiman:
I'm doing some fuzzing using statetests that are not properly 'filled', meaning that they do not have a proper post-state. This makes Parity (`parity-evm`) behave a bit oddly. 

## How parity-evm does
When `parity-evm` is executed against a bad test, it behaves oddly.

```
# /parity-evm state-test --std-json /testfiles/martin-Thu_22_42_11-15624-7-test.json

/ $  /parity-evm state-test --std-json /testfiles/martin-Fri_08_26_26-10380-8-test.json
{"test":"randomStatetestmartin-Fri_08_26_26-10380-8:byzantium:0","action":"starting"}
{"pc":0,"op":0,"opName":"STOP","gas":"0x28b38d","stack":[],"storage":{},"depth":1}
{"error":"State root mismatch (got: 0x6bba…662f, expected: 0x0000…c0de)","gasUsed":"1c5c4b36b03a23","time":114}
```

1. It has an erroneous `gasUsed`-value: `1c5c4b36b03a23` 
2. The calculated `stateRoot` cannot be read from the output, since it's obfuscated by ellipsis (`0x6bba…662f`). 

For reference, here's how it looks if the test is fixed with a correct `postState`:

```
/ $  /parity-evm state-test --std-json /testfiles/fixed.json 
{"test":"randomStatetestmartin-Fri_08_26_26-10380-8:byzantium:0","action":"starting"}
{"pc":0,"op":0,"opName":"STOP","gas":"0x28b38d","stack":[],"storage":{},"depth":1}
{"stateRoot":"0x6bba9f22296aa9817a905f737f7edc8d6ab00c337d333a90e972778746a9662f"}
{"output":"0x","gasUsed":"1c5c4b36878696","time":108}
```
## How geth does

Geth separates the `jsonl` output and the actual test results, so that if we want to compare traces, we only need to look at the `stderr`. 

Geth (`stdout`):
```
/ # evm --json --nomemory statetest /testfiles/fixed.json 2>/dev/null
[
  {
    "name": "randomStatetestmartin-Fri_08_26_26-10380-8",
    "pass": true,
    "fork": "Byzantium"
  }
]
```
Geth (`stderr`)
```
/ # evm --json --nomemory statetest /testfiles/fixed.json 1>/dev/null
[...]

PR #9706, opened 2018-10-05
Add a optional json dump state to evm-bin

cheme:
This PR allows evm-bin to produce a json dump of the end state.
This is activated by using '--std-dump-json' which is a variant of '--std-json'.

This feature was asked for on ethereum/fuzztests gitter by @cdetrio and @holiman and geth related pr is : https://github.com/ethereum/go-ethereum/pull/17832 .

Two other variants are added for '--std-json' to allow some flexibility on the output streams : '--std-err-only' and '--std-out-only' that either puts everything on err or on out, I think it is one of the point of https://github.com/paritytech/parity-ethereum/issues/9408 .

The produce json for : 
```parity-evm state-test --std-json --std-dump-json --std-out-only ../../randomtests/Constantinople.minified.DEFAULT-Wed_13_52_03-13586-34149-test.json ``` is :

```
{"action":"starting","test":"randomStatetest:constantinople:0"}
{"depth":1,"gas":"0x91ba51","op":97,"opName":"PUSH2","pc":0,"stack":[],"storage":{}}
{"depth":1,"gas":"0x91ba4e","op":97,"opName":"PUSH2","pc":3,"stack":["0x376"],"storage":{}}
{"depth":1,"gas":"0x91ba4b","op":97,"opName":"PUSH2","pc":6,"stack":["0x376","0x2c0"],"storage":{}}
{"depth":1,"gas":"0x91ba48","op":97,"opName":"PUSH2","pc":9,"stack":["0x376","0x2c0","0x142"],"storage":{}}
...
{"depth":1,"gas":"0x8ace36","op":0,"opName":"STOP","pc":346,"stack":["0x1","0x0","0xdc04ee1a6673a05e7db196f61feb73d9ddad8ee","0x79de509cbe4d","0xac3eaebadf3075","0x0","0x0","0x640","0x1","0x20000","0x0","0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b","0x25f9d382ec5be36f","0x822fe20dcc76697c77d01a6c8039fd913b781ca4"],"storage":{}}
{"accounts":{"0x0dc04ee1a6673a05e7db196f61feb73d9ddad8ee":{"balance":"0xf8884ac77f5a629","code":"0x","nonce":"0x1","storage":{}},"0x822fe20dcc76697c77d01a6c8039fd913b781ca4":{"balance":"0x33077a8da750d683","code":"0x","nonce":"0x1","storage":{}},"0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b":{"balance":"0xf21861f7e5420393","code":"0x","nonce":"0x1e","storage":{}},"0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b":{"balance":"0x42a7299fac2152ad","code":"0x6103766102c06101426102eb733390556786a07591437e1e48bcb35bc54ce98dc3674cd2cc7fe2310224fa67d646e22e9f0a851367933b8f2d33b171551c677363517ed8d1b03661025d61033c670f8884ac77f5a629f565d6555072b2997a3bc3bcdcd779041c35bd45bd6c031588e970f5e5c3ef70346978ad701d9afd2a65f0389b7860acc6701acc23ef09714a2ca8db7226b50bf4d777cb6ea3df0072c863174d4c9066ac3eaea391e3e5086bbe676b3a6c1aaaa5adb967016102ee52676cdcd38160c8266a6796a9634d013d1e521d67da27088773e36138671c763d7f6601dd2a1d59434473072911627a9f357ebe3d2963031daba1d94ee7183f3067337bbba2c059192167168268202c02fa4e186710b5f63510867cab60cc6101316733077a8da750d683f579ab22d6bf2767f81735fb46671b5c9259cce7693b5a79207294ca6cd3c0c9f4b02fb34fb6fd062d6861b1b961a72ea200","nonce":"0x1f","storage":{}}},"root":"0xe568e9556f8729ce1aa89ef50538e17adce0c37ab8d0ba2b2f4c9eca85f4a9d5"}
{"error":"State root mismatch (got: 0xe568e9556f8729ce1aa89ef50538e17adce0c37ab8d0ba2b2f4c9eca85f4a9d5, expected: 0x00000000000000000000000000000000000000000000000000000000deadc0de)","gasUsed":"0x73e23","time":10979}
```

Note for reviewer:
- this pr adds serde and serde-derive as dependency of ethcore, if needed I can move the serde serializing code to a newtype implementation in another crate. At the time I am not sure it is worth it (there is already some serde in some subcrates of ethcore).
- the new associated type 'Sink' might look awkward but required to allow 'finish' method to stay static (other changes are much more impacting) and use the right streams.
- reading from triedb means switching to fatdb, it should only be active with "--std-dump-state"

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -108,6 +108,8 @@ memory_profiling = []
 # in order to manually test that parity fall-over to the local version
 # in case of invalid or deprecated command line arguments are entered
 test-updater = ["parity-updater/test-updater"]
+# Allow functionalities for evmbin or test only.
+evmbin-only = ["ethcore/evmbin-only"]
```
sorpaas: Is this feature flag really needed?

ethcore/Cargo.toml
```diff
@@ -110,3 +112,5 @@ test-heavy = []
 benches = []
 # Compile test helpers
 test-helpers = ["tempdir"]
+# Slow functionalities for evmbin or test only.
+evmbin-only = []
```
sorpaas: Just a nitpick -- suggest to call it something more direct, or maybe just `to-pod-slow`, or `to-pod-full`.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +951,58 @@ impl<B: Backend> State<B> {
...
+			}
+		}
+
+		PodState::from(result)
+	}
+
+
+	/// Only for use with tests (to slow for real use),
+	/// panic if feature 'evmbin-only' is not activated.
+	#[cfg(not(feature="evmbin-only"))]
```
sorpaas: Why do we need an explicit panic here? We should be able to just remove this section all together if `to_pod_slow` is indeed not used elsewhere.

ethcore/src/state/mod.rs
```diff
@@ -939,8 +939,7 @@ impl<B: Backend> State<B> {
 		}
 	}
 
-	/// Populate a PodAccount map from this state.
-	pub fn to_pod(&self) -> PodState {
+	fn cache_to_pod(&self) -> PodState {
```
sorpaas: We can keep the docs `/// Populate a PodAccount map from this state.`.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +951,58 @@ impl<B: Backend> State<B> {
 		}))
 	}
 
+	#[cfg(feature="evmbin-only")]
+	/// Populate a PodAccount map from this state.
+	/// Warning this is not for real time use
+	pub fn to_pod_slow(&self) -> PodState {
```
sorpaas: nitpick -- maybe name this something more about its functionality `to_pod_full` or others.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +951,58 @@ impl<B: Backend> State<B> {
 		}))
 	}
 
+	#[cfg(feature="evmbin-only")]
+	/// Populate a PodAccount map from this state.
```
sorpaas: Add docs here that this requires FatDB.

ethcore/wasm/run/src/runner.rs
```diff
@@ -66,7 +66,7 @@ impl Fail {
 }
 
 impl fmt::Display for Fail {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
sorpaas: typo -- looks like incorrect indentation.

evmbin/src/info.rs
```diff
@@ -19,19 +19,31 @@
...
 	/// Display a single run init message
 	fn before_test(&mut self, test: &str, action: &str);
 	/// Set initial gas.
 	fn set_gas(&mut self, _gas: U256) {}
+	/// Clone sink.
+	fn clone_sink(&self) -> Self::Sink;
 	/// Display final result.
-	fn finish(result: RunResult<Self::Output>);
+	fn finish(result: RunResult<Self::Output>, &mut Self::Sink);
+  /// Get trie spec to use.
```
sorpaas: Indentation is wrong here.

Cargo.toml
cheme: No just a reexport, can remove.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +951,58 @@ impl<B: Backend> State<B> {
...
+		let iter = match trie.iter() {
+			Ok(iter) => iter,
+			_ => return PodState::from(result),
+		};
+
+		let from_rlp = |b: &[u8]| Account::from_rlp(b).expect("decoding db value failed");
+
+		for item in iter {
+			if let Ok((addr, dbval)) = item {
+				let mut account = from_rlp(&dbval[..]);
```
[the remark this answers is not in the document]
cheme: Indeed it needs a previous 'commit' call to be correct, let's make cache then trie version.

ethcore/src/state/mod.rs
cheme: Only way, is to move call in evmbin, makes thing a bit more complicated. In fact it removes the need for a 'do_dump' fn in informant trait, it is better.

sorpaas: Didn't get you on this part. Isn't evmbin always built with this feature on? And if this function is not used anywhere else, I think we can remove it without getting any compiling errors?

cheme: The function was call from EvmTestClient (in ethcore module). EvmTestClient is used in other context than evmbin only (json test only at the time, plus libfuzzer, and anything because public). And the function call depend on a trait function (could not be in the evmbin informant trait cause this one is dropped at the time of the function call). So I needed an alternate implementation to compile. Anyway, with last commit it is not needed anymore.

ethcore/src/state/mod.rs
```diff
@@ -940,20 +940,71 @@ impl<B: Backend> State<B> {
...
+		};
+
+		let from_rlp = |b: &[u8]| Account::from_rlp(b).expect("decoding db value failed");
+
+		for item in iter {
+			if let Ok((addr, dbval)) = item {
+				result.entry(Address::from_slice(&addr)).or_insert_with(|| {
+					let mut account = from_rlp(&dbval[..]);
+					if account.code_size().is_none() {
+						let addr_hash = account.address_hash(&H160::from(&addr[..]));
```
sorpaas: Sorry I missed this -- this will also not completely work -- for a fresh account object, we would miss all storage items! And I do believe we have plenty of tests that deal with storage items. So I think we would also need to fix this.

We do already have logics in `cli::blockchain::execute_export_state` that do this. You can refer to its logic, and I would recommend one of the two ways to deal with this.

#### Refactor `Client::list_accounts` and `Client::list_storage`

Currently they only work for real client, but not EVM test client, but I believe it would be possible to extend this interface, and change the `count` argument to accept `Option`. With this, we change the signature to `to_pod_full(&self, accounts: HashMap<Address, Vec<H256>>)`, where for the hashmap, keys are account addresses, and values are existing storage items.

Within this function, call `State::require` for all addresses, and call `State::storage_at` for all storage items. After that, we can use `State::to_pod_cache` safely to get all addresses.

#### Simpler without Refactoring

Simply use the same logic as `Client::list_accounts` and `Client::list_storage`, but copy them to `State`, gather `HashMap<Address, Vec<H256>>` there. And use `State::require`, `State::storage_at`, and `State::to_pod_cache`.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,78 @@ impl<B: Backend> State<B> {
...
+					.map(|iter| iter.for_each(|o_kv|{
+						if let Ok((key, val)) = o_kv {
+							pod_storage.insert(H256::from(&key[..]), H256::from(&val[..]));
+						}
+					}
+					))));
+
+		let mut pod_account = PodAccount::from_account(&account);
+		// cached one first
+		pod_storage.append(&mut pod_account.storage);
```
sorpaas: Please use `self.storage_at` instead. The issue is that what we have in trie and what we have in cache might differ, and we need to use the value in cache.

sorpaas: > Reuse of `storage_at` was not easy in this case (could not iterate and
use the method at the same time (refcell mutable borrow panics) so
keeping code as is.

Would you mind to try this:

* Collect all storage items from trie first in a `Vec<H256>`.
* Call `storage_at` at the end of the round.

Not sure whether that would solve the lifetime issue.

cheme: Yes this will works (but I kinda wanted to avoid that :) ), I will switch to this version for the sake of using similar code.

sorpaas:
@cheme Just in case. Here's one way that I think this may work for us (untested code!):

```
fn fetch_trie_accounts(&self) -> Result<HashMap<Address, Vec<H256>>> {
  let mut result = HashMap::new();

  let trie = self.factories.trie.readonly(self.db.as_hashdb(), &self.root)?;
  for item in trie.iter() {
    if let Ok((addr, dbval)) = item { // Should we just return with Err case?
      let address = Address::from_slice(&addr);
      let account = Account::from_rlp(&dbval);

      let mut storage = Vec::new();
      let storage_trie = self.factories.trie.readonly(accountdb.as_hashdb(), &account.storage_root);
      for item in storage_trie.iter() {
        if let Ok((key, _)) = item {
          storage.insert(H256::from(key));
        }
      }
      result.insert(address, storage);
  }
  result
}

pub fn to_pod_full(&self) -> Result<PodState> {
  assert!(self.checkpoints.borrow().is_empty());
  assert!(self.factories.trie.is_fat());

  let trie_accounts = self.fetch_trie_accounts()?;
  for (address, storage) in trie_accounts {
    self.require(address, true)?;
    for key in storage {
      self.storage_at(address, key)?;
    }
  }
  Ok(self.to_pod_cache())
}
```

ethcore/src/state/mod.rs
cheme: > Please use self.storage_at instead. The issue is that what we have in trie and what we have in cache might differ, and we need to use the value in cache.

Doesn't the merge ('append') of both storage do that? (pod_account got values from storage_change replacing value read from trie). Can value from 'original_storage_cache' of 'Account' differs from trie value (this field is lost in the process) ? 

Switching to 'storage_at' anyway.

sorpaas: @cheme Ah sorry. Yeah you're right. `PodAccount::from` would only print things on `storage_changes`, so using `State::storage_at` won't be sufficient for our case. I think the current method works.

cheme:
@sorpaas , I did test your code, the problem is that 'storage_at' just update the lru cache, and when using 'to_pod_cache' -> 'PodAccount::from_account' will only display 'storage_changes', so we do not have any storage displayed.
For instance when using '/ethereum/tests/GeneralStateTests/stCallCodes/callcall_00.json' : 
I previously had:
```
{"accounts":{"0x1000000000000000000000000000000000000000":{"balance":"0xde0b6b3a763ffff","code":"0x6040600060406000600173100000000000000000000000000000000000000162055730f1600055","nonce":"0x0","storage":{"0x0000000000000000000000000000000000000000000000000000000000000000":"0x0100000000000000000000000000000000000000000000000000000000000000"}},"0x1000000000000000000000000000000000000001":{"balance":"0xde0b6b3a763ffff","code":"0x604060006040600060027310000000000000000000000000000000000000026203d090f1600155","nonce":"0x0","storage":{"0x0000000000000000000000000000000000000000000000000000000000000001":"0x0100000000000000000000000000000000000000000000000000000000000000"}},"0x1000000000000000000000000000000000000002":{"balance":"0x2","code":"0x600160025533600455346007553060e6553260e8553660ec553860ee553a60f055","nonce":"0x0","storage":{"0x0000000000000000000000000000000000000000000000000000000000000002":"0x0100000000000000000000000000000000000000000000000000000000000000","0x0000000000000000000000000000000000000000000000000000000000000004":"0x9410000000000000000000000000000000000000010000000000000000000000","0x0000000000000000000000000000000000000000000000000000000000000007":"0x0200000000000000000000000000000000000000000000000000000000000000","0x00000000000000000000000000000000000000000000000000000000000000e6":"0x9410000000000000000000000000000000000000020000000000000000000000","0x00000000000000000000000000000000000000000000000000000000000000e8":"0x94a94f5374fce5edbc8e2a8697c15331677e6ebf0b0000000000000000000000","0x00000000000000000000000000000000000000000000000000000000000000ec":"0x4000000000000000000000000000000000000000000000000000000000000000","0x00000000000000000000000000000000000000000000000000000000000000ee":"0x2100000000000000000000000000000000000000000000000000000000000000","0x00000000000000000000000000000000000000000000000000000000000000f0":"0x0100000000000000000000000000000000000000000000000000000000000000"}},"0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba":{"balance":"0x3997d","code":"0x","nonce":"0x0","storage":{}},"0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b":{"balance":"0xde0b6b3a7606683","code":"0x","nonce":"0x1","storage":{}}},"root":"0x7ebd8fdaaeb5e96fd2ad3b11c675e3b16e5902ae4fe100710f39925e59ccd914"}
```
and with 'storage_at', I got:
```
{"accounts":{"0x1000000000000000000000000000000000000000":{"balance":"0xde0b6b3a763ffff","code":"0x6040600060406000600173100000000000000000000000000000000000000162055730f1600055","nonce":"0x0","storage":{}},"0x1000000000000000000000000000000000000001":{"balance":"0xde0b6b3a763ffff","code":"0x604060006040600060027310000000000000000000000000000000000000026203d090f1600155","nonce":"0x0","storage":{}},"0x1000000000000000000000000000000000000002":{"balance":"0x2","code":"0x600160025533600455346007553060e6553260e8553660ec553860ee553a60f055","nonce":"0x0","storage":{}},"0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba":{"balance":"0x3997d","code":"0x","nonce":"0x0","storage":{}},"0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b":{"balance":"0xde0b6b3a7606683","code":"0x","nonce":"0x1","storage":{}}},"root":"0x7ebd8fdaaeb5e96fd2ad3b11c675e3b16e5902ae4fe100710f39925e59ccd914"}
{"gasUsed":"0x3997d","output":"0x","time":7236}
```
So I am wondering, do we want to display all storage value or only the changed one?
To use 'storage_at' I can still make an alternate version of 'from_account' that also use the lru cache values.

sorpaas:
@cheme Yeah you're right. I was wrong previously -- we cannot reuse `State::storage_at` and then `State::to_pod_cache` because it only prints changed storage items. So your current solution looks like working!

sorpaas:
And indeed we want to display all storage values just to avoid any confusion.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,78 @@ impl<B: Backend> State<B> {
...
+
+		PodState::from(result)
+	}
+
+	fn account_to_pod_account(&self, account: &mut Account, address: &Address) -> PodAccount {
+		let mut pod_storage = BTreeMap::new();
+		let addr_hash = account.address_hash(address);
+		let accountdb = self.factories.accountdb.readonly(self.db.as_hashdb(), addr_hash);
+
+		if account.code_size().is_none() {
```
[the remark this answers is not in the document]
sorpaas: Looks like if this ever happens, then it means previously we had a DB failure caching the code. So below we shouldn't call `cache_code` again. Would suggest we either change this to just skip code and print an warning or return an error.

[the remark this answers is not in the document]
cheme: You are right, seems impossible, it is a remnant from my previous code where I did not use 'require'.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,78 @@ impl<B: Backend> State<B> {
...
+		// put trie in cache 
+		for item in iter {
+			if let Ok((addr, _dbval)) = item {
+				let address = Address::from_slice(&addr);
+				let _ = self.require(&address, true);
+			}
+		}
+
+		// Resolve missing part
+		for (add, opt) in self.cache.borrow_mut().iter_mut() {
```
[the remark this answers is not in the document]
cheme: Same thing, it is from previous code that did not use qrequire', will be removed.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,78 @@ impl<B: Backend> State<B> {
...
+		let mut pod_storage = BTreeMap::new();
+		let addr_hash = account.address_hash(address);
+		let accountdb = self.factories.accountdb.readonly(self.db.as_hashdb(), addr_hash);
+
+		if account.code_size().is_none() {
+			let _ = account.cache_code(accountdb.as_hashdb());
+		}
+
+		account.storage_root()
+			.map(|root| self.factories.trie.readonly(accountdb.as_hashdb(), &root)
```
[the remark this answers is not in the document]
cheme: I will change to returning Result, it makes things looks a bit nicer.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,78 @@ impl<B: Backend> State<B> {
...
+	fn account_to_pod_account(&self, account: &mut Account, address: &Address) -> PodAccount {
+		let mut pod_storage = BTreeMap::new();
+		let addr_hash = account.address_hash(address);
+		let accountdb = self.factories.accountdb.readonly(self.db.as_hashdb(), addr_hash);
+
+		if account.code_size().is_none() {
+			let _ = account.cache_code(accountdb.as_hashdb());
+		}
+
+		account.storage_root()
```
sorpaas: Should use `base_storage_root` here.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,69 @@ impl<B: Backend> State<B> {
...
+			}
+		}
+
+		Ok(PodState::from(result))
+	}
+
+	/// Create a PodAccount from an account.
+	/// Differs from existing method by adding all storage
+	/// values of the account to the PodAccount.
+	/// This function shall therefore only be used for small
```
dvdplm: Maybe "This function is only intended for use in small tests or with fresh accounts"?

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,69 @@ impl<B: Backend> State<B> {
...
+	fn account_to_pod_account(&self, account: &Account, address: &Address) -> Result<PodAccount, Error> {
+		let mut pod_storage = BTreeMap::new();
+		let addr_hash = account.address_hash(address);
+		let accountdb = self.factories.accountdb.readonly(self.db.as_hashdb(), addr_hash);
+		let root = account.base_storage_root();
+
+		let trie = self.factories.trie.readonly(accountdb.as_hashdb(), &root)?;
+		for o_kv in trie.iter()? {
+			if let Ok((key, val)) = o_kv {
+				pod_storage.insert(H256::from(&key[..]), H256::from(U256::from(&val[..])));
```
dvdplm: Maybe you can do `pod_storage.insert(&key[..].into(), U256::from(&val[..]).into());` if you declare the `pod_storage` with `BTreeMap::new::<H256, H256>::new()`?

ethcore/src/state/mod.rs
```diff
@@ -2615,18 +2676,64 @@ mod tests {
...
-					   nonce: U256::zero(),
-					   code: Some(Default::default()),
-					   storage: vec![(H256::from(&U256::from(1u64)), H256::from(&U256::from(100u64)))]
-						   .into_iter().collect(),
-				   })).as_ref());
+				pod_account::diff_pod(Some(&PodAccount {
+					balance: U256::zero(),
+					nonce: U256::zero(),
+					code: Some(Default::default()),
+					storage: vec![(H256::from(&U256::from(1u64)), H256::from(&U256::from(20u64)))]
```
dvdplm: Are you sure you have to pass over `U256` here?

evmbin/src/info.rs
```diff
@@ -114,64 +126,82 @@ pub fn run_transaction<T: Informant>(
...
-			TransactResult::Ok { state_root, gas_left, .. } if state_root != post_root => {
-				(Err(EvmTestError::PostCondition(format!(
-					"State root mismatch (got: 0x{:x}, expected: 0x{:x})",
-					state_root,
-					post_root,
-				))), Some(gas_left), None)
+			TransactResult::Ok { state_root, gas_left, output, vm_trace, end_state, .. } => {
+				if state_root != post_root {
+					(Err(EvmTestError::PostCondition(format!(
+						"State root mismatch (got: 0x{:x}, expected: 0x{:x})",
```
dvdplm: Maybe use `{:#x}` here?

evmbin/src/main.rs
```diff
@@ -89,6 +89,9 @@ State test options:
 General options:
     --json             Display verbose results in JSON.
     --std-json         Display results in standardized JSON format.
+    --dump-json        Display result state dump in standardized JSON format.
+    --err-only         With --std-json redirect to err output only.
+    --out-only         With --std-json redirect to out output only.
```
dvdplm: Maybe this is a dumb idea, but there's a risk users get it wrong and use `--err-only` without `--dump-json` so we could do

```
--dump-json … …
--dump-json-err-only … …
--dump-json-out-only … …
```
i.e. have the three options mutually exclusive instead of combining.

ethcore/src/state/mod.rs
cheme: Not sure at all, copy paste from other tests :)  But it makes me understand a previous conversation on riot about casting usize to H256 : it appends 0 at the end of the array, so we need the intermediate U256 casting to get the expected value (I got a bug from this at first).
Personally I would make the cast from usize to H256 preppends 0, it is a really big source of error.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +952,69 @@ impl<B: Backend> State<B> {
...
 
+	#[cfg(feature="to-pod-full")]
+	/// Populate a PodAccount map from this state.
+	/// Warning this is not for real time use.
+	/// Use of this method requires FatDB mode to be able 
+	/// to iterate on accounts.
+	pub fn to_pod_full(&self) -> Result<PodState, Error> {
+
+		assert!(self.checkpoints.borrow().is_empty());
+		assert!(self.factories.trie.is_fat());
```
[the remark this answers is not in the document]
cheme: I'd rather panic here, unless/untill we got coding rules that force error checking (I did not check it currently), forcing some 'clippy' rules in ci could really be an idea.
In the context of this function (very specific use, not in critical context), I think the assert is ok (the function should never be called if not 'fat' and if it is called it is a bug to fix).

ethcore/src/state/mod.rs
```diff
@@ -940,7 +940,7 @@ impl<B: Backend> State<B> {
 	}
 
 	/// Populate a PodAccount map from this state.
-	pub fn to_pod(&self) -> PodState {
+	fn to_pod_cache(&self) -> PodState {
 		assert!(self.checkpoints.borrow().is_empty());
 		// TODO: handle database rather than just the cache.
```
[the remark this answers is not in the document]
cheme: You're right, the new method and renaming makes it somehow explicit. (in fact I should have read it, it was a good comment because it spotted the fact that fatdb would be need).

ethcore/src/pod_account.rs
```diff
@@ -31,21 +31,29 @@ use state::Account;
...
+	#[serde(serialize_with="opt_bytes_to_hex")]
 	/// The code of the account or `None` in the special case that it is unknown.
 	pub code: Option<Bytes>,
 	/// The storage of the account.
 	pub storage: BTreeMap<H256, H256>,
 }
 
+fn opt_bytes_to_hex<S>(opt_bytes: &Option<Bytes>, serializer: S) -> Result<S::Ok, S::Error>
+	where S: Serializer {
+	serializer.collect_str(&format_args!("0x{}",opt_bytes.as_ref().map(|b|b.to_hex()).unwrap_or("".to_string())))
```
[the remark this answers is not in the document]
cheme: It does not implement it, alas `Bytes` is only a type alias for `Vec<u8>`, so it should be quite a refact to fix it.

ethcore/src/client/evm_test_client.rs
```diff
@@ -92,32 +97,51 @@ impl<'a> EvmTestClient<'a> {
...
 
+	/// Change default function for dump state (default does not dump)
+	pub fn set_dump_state_fn(&mut self, dump_state: fn(&state::State<state_db::StateDB>) -> Option<pod_state::PodState>) {
+		self.dump_state = dump_state;
+	}
+
 	/// Creates new EVM test client with in-memory DB initialized with genesis of given Spec.
-	pub fn new(spec: &'a spec::Spec) -> Result<Self, EvmTestError> {
-		let factories = Self::factories();
+	/// You can specify a specific trie kind.
```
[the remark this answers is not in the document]
cheme: I am skipping the different type of trie (chance of leading to outdated comment in case of change on the variants).

ethcore/src/pod_state.rs
```diff
@@ -26,7 +26,7 @@ use types::state_diff::StateDiff;
 use ethjson;
 
 /// State of all accounts in the system expressed in Plain Old Data.
-#[derive(Debug, Clone, PartialEq, Eq, Default)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize)]
 pub struct PodState (BTreeMap<Address, PodAccount>);
```
sorpaas: Not related to this PR, but there's an extra space between `PodState` and `(BTreeMap`.

ethcore/src/state/mod.rs
```diff
@@ -952,8 +950,68 @@ impl<B: Backend> State<B> {
...
+				let address = Address::from_slice(&addr);
+				let _ = self.require(&address, true);
+			}
+		}
+
+		// Resolve missing part
+		for (add, opt) in self.cache.borrow().iter() {
+			if let Some(ref acc) = opt.account {
+				let pod_account = self.account_to_pod_account(acc, add)?;
+				result.insert(add.clone(), pod_account);
```
sorpaas: nit: `Address` is copy, so we can just do `*add` here.

ethcore/src/pod_account.rs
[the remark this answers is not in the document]
niklasad1: Consider to replace ```map().unwrap_or()``` with ```map_or_else```!

5chdn:
Could you give this a final review @dvdplm @niklasad1

5chdn:
Could you give this a final review @dvdplm @niklasad1 @tomusdrw

PR #9710, opened 2018-10-05
Pyethereum keystore support

tworec:
Continuation of #9444
(sorry for duplicate, but I can not reopen #9444)


ethstore/src/account/safe_account.rs
```diff
@@ -76,17 +76,28 @@ impl SafeAccount {
...
-	/// can be left `None`.
-	pub fn from_file(json: json::KeyFile, filename: Option<String>) -> Self {
-		SafeAccount {
+	/// can be left `None`. `password` is used for reading keystore files w/o address.
+	pub fn from_file(json: json::KeyFile, filename: Option<String>, password: &Option<Password>) -> Result<Self, Error> {
+		let crypto = Crypto::from(json.crypto);
+		let address = match json.address {
+			Some(address) => address.into(),
+			None => match password {
+                Some(password) =>
```
debris: invalid indentation

ethstore/src/accounts_dir/disk.rs
```diff
@@ -104,8 +105,16 @@ pub type RootDiskDirectory = DiskDirectory<DiskKeyFileManager>;
 
 /// Disk directory key file manager
 pub trait KeyFileManager: Send + Sync {
+	/// Read `SafeAccount` from given key file stream with password
+	#[allow(unused_variables)]
+	fn read_with_password<T>(&self, filename: Option<String>, reader: T, password: &Option<Password>)
+		-> Result<SafeAccount, Error> where T: io::Read {
+		unimplemented!()
```
debris: this is a badly designed interface. what you should do is:

```rust
pub struct KeyFileManagerWithPassword {
  password: Password,
}

impl KeyFileManager for KeyFileManagerWithPassword {
  fn read<T>(&self, filename: Option<String>, reader: T) -> Result<SafeAccount, Error> where T: io::Read;
}
```

ethstore/src/accounts_dir/mod.rs
```diff
@@ -46,8 +46,15 @@ pub struct VaultKey {
 
 /// Keys directory
 pub trait KeyDirectory: Send + Sync {
+	/// Read keys from directory with password
+	#[allow(unused_variables)]
+	fn load_with_password(&self, password: &Option<Password>) -> Result<Vec<SafeAccount>, Error> {
+		unimplemented!()
```
debris: same issue. badly designed interface

ethstore/src/json/crypto.rs
```diff
@@ -81,7 +82,7 @@ impl<'a> Visitor<'a> for CryptoFieldVisitor {
 			"kdf" => Ok(CryptoField::Kdf),
 			"kdfparams" => Ok(CryptoField::KdfParams),
 			"mac" => Ok(CryptoField::Mac),
-			_ => Err(Error::custom(format!("Unknown field: '{}'", value))),
+			_ => Ok(CryptoField::Ignored),
```
debris: I don't agree with this change. There is definitely a finite number of special cases. Please handle all of them and describe why they exist

ethstore/src/json/key_file.rs
```diff
@@ -46,7 +46,7 @@ pub struct KeyFile {
 	pub id: Uuid,
 	pub version: Version,
 	pub crypto: Crypto,
-	pub address: H160,
+	pub address: Option<H160>,
```
debris: ~I'm not sure about this change either. I don't know how this affects runtime. Without `address` field we don't know that what account we just loaded and I am afraid that it may be simply unusable.~

nvm, I see that you load the password in `from_file`. Please describe in the code why this field is optional.

ethstore/src/account/safe_account.rs
```diff
@@ -76,17 +76,28 @@ impl SafeAccount {
...
 	/// file, the `filename` should be `Some` name. If it is as yet anonymous, then it
-	/// can be left `None`.
-	pub fn from_file(json: json::KeyFile, filename: Option<String>) -> Self {
-		SafeAccount {
+	/// can be left `None`. `password` is used for reading keystore files w/o address.
+	pub fn from_file(json: json::KeyFile, filename: Option<String>, password: &Option<Password>) -> Result<Self, Error> {
+		let crypto = Crypto::from(json.crypto);
+		let address = match json.address {
+			Some(address) => address.into(),
+			None => match password {
```
debris: what if the json file contains `address` and we provided a password? imo, this should result in an `Error`

ethstore/src/account/kdf.rs
```diff
@@ -26,7 +26,7 @@ pub struct Pbkdf2 {
 	pub c: u32,
 	pub dklen: u32,
 	pub prf: Prf,
-	pub salt: [u8; 32],
+	pub salt: Vec<u8>,
```
debris: why did you change it to a `Vec<u8>`? I see no reason to do that

ethstore/src/json/bytes.rs
```diff
@@ -19,7 +19,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde::de::Error;
 use rustc_hex::{ToHex, FromHex, FromHexError};
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
debris: why did you add `derive(Clone)`?

ethstore/src/json/kdf.rs
```diff
@@ -111,7 +111,7 @@ pub struct Pbkdf2 {
 	pub c: u32,
 	pub dklen: u32,
 	pub prf: Prf,
-	pub salt: H256,
+	pub salt: Bytes,
```
debris: why is it changed to `Bytes`?

ethkey/cli/src/main.rs
```diff
@@ -168,7 +168,7 @@ fn main() {
 		Ok(ok) => println!("{}", ok),
 		Err(Error::Docopt(ref e)) => e.exit(),
 		Err(err) => {
-			println!("{}", err);
+			error!("{}", err);
```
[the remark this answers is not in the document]
tworec: :+1:

ethstore/cli/src/main.rs
```diff
@@ -146,12 +149,13 @@ impl fmt::Display for Error {
...
 fn main() {
 	panic_hook::set_abort();
+	env_logger::try_init().expect("Logger initialized only once.");
 
 	match execute(env::args()) {
 		Ok(result) => println!("{}", result),
 		Err(Error::Docopt(ref e)) => e.exit(),
 		Err(err) => {
-			println!("{}", err);
+			error!("{}", err);
```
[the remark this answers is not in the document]
tworec: :+1:

ethstore/cli/src/main.rs
```diff
@@ -241,7 +245,8 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
 	} else if args.cmd_import {
 		let src = key_dir(&args.flag_src)?;
 		let dst = key_dir(&args.flag_dir)?;
-		let accounts = import_accounts(&*src, &*dst)?;
+		let password = load_password(&args.arg_password).ok();
```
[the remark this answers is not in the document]
tworec: yup, empty string.

error will be printed when there will be a keystore to import w/o address. Otherwise I would silently ignore invalid pwd path.

ethstore/src/account/safe_account.rs
tworec: Again, I would silently ignore this. As for now all errors reading kestores are silently ignored, which is confusing me too, but to be consistent I propose to ignore it.

ethstore/src/json/bytes.rs
tworec: its probably aftermath my debugging... sorry

ethstore/src/json/kdf.rs
tworec: again, it's bc we do not know length of salt

ethstore/src/json/crypto.rs
tworec: :+1:

ethstore/src/accounts_dir/disk.rs
tworec: thx, will rethink it

5chdn:
Are you still working on this @tworec

tworec:
I want to finish this. I was looking to see @debris answers...

ethstore/src/account/kdf.rs
niklasad1: @tworec just revert the changes

tworec: but we do not the salt length, it can exceed 32 bytes. What then?

niklasad1: All right, sorry did not see your comment there then it makes sense to use a vector instead of an array.

tworec: 👍

ethstore/src/account/safe_account.rs
tworec: fixed

ethstore/src/json/crypto.rs
tworec: done

ethstore/src/json/crypto.rs
```diff
@@ -122,6 +124,8 @@ impl<'a> Visitor<'a> for CryptoVisitor {
 				Some(CryptoField::Kdf) => { kdf = Some(visitor.next_value()?); }
 				Some(CryptoField::KdfParams) => { kdfparams = Some(visitor.next_value()?); }
 				Some(CryptoField::Mac) => { mac = Some(visitor.next_value()?); }
+				// skip not required version field (it appears in pyethereum generated keystores)
```
sorpaas: :(

ethstore/src/lib.rs
```diff
@@ -36,7 +36,6 @@ extern crate ethereum_types;
 extern crate ethkey as _ethkey;
 extern crate parity_wordlist;
 
-#[macro_use]
 extern crate log;
```
sorpaas: Are we using `log` crate in `ethstore` at all? If not, maybe remove this altogether?

ethstore/cli/src/main.rs
```diff
@@ -239,8 +239,13 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
 			.collect();
 		Ok(format_accounts(&accounts))
 	} else if args.cmd_import {
-		let src = key_dir(&args.flag_src)?;
-		let dst = key_dir(&args.flag_dir)?;
+		let password = match &args.arg_password.as_ref() {
+			&"" => None,
```
niklasad1: You can remove `&` here 
```suggestion
			"" => None,
```

ethstore/cli/src/main.rs
```diff
@@ -239,8 +239,13 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
 			.collect();
 		Ok(format_accounts(&accounts))
 	} else if args.cmd_import {
-		let src = key_dir(&args.flag_src)?;
-		let dst = key_dir(&args.flag_dir)?;
+		let password = match &args.arg_password.as_ref() {
```
niklasad1:
```suggestion
		let password = match args.arg_password.as_ref() {
```

ethstore/cli/src/main.rs
```diff
@@ -239,8 +239,13 @@ fn execute<S, I>(command: I) -> Result<String, Error> where I: IntoIterator<Item
...
+			&"" => None,
+			_ => Some(load_password(&args.arg_password)?)
```
niklasad1: I don't like that you wrap potential errors in Option, use `ok()` instead to convert the result to an `Option`

niklasad1:
```suggestion
			_ => load_password(&args.arg_password).ok()
```

ethstore/src/accounts_dir/disk.rs
```diff
@@ -117,16 +119,30 @@ pub struct DiskDirectory<T> where T: KeyFileManager {
...
+	}
+}
 
 impl RootDiskDirectory {
 	pub fn create<P>(path: P) -> Result<Self, Error> where P: AsRef<Path> {
 		fs::create_dir_all(&path)?;
 		Ok(Self::at(path))
 	}
 
+	pub fn with_password(&self, password: Option<Password>) -> Self {
```
niklasad1: Add documentation here, please!

ethstore/src/accounts_dir/disk.rs
```diff
@@ -191,7 +207,7 @@ impl<T> DiskDirectory<T> where T: KeyFileManager {
 					.map_err(Into::into)
 					.and_then(|file| self.key_manager.read(filename, file))
 					.map_err(|err| {
-						warn!("Invalid key file: {:?} ({})", path, err);
+						eprintln!("Invalid key file: {:?} ({})", path, err);
```
niklasad1: hrm, why `eprintln` here instead of rustlog? I believe we want it to be rustlog

ethstore/src/accounts_dir/disk.rs
```diff
@@ -117,16 +119,30 @@ pub struct DiskDirectory<T> where T: KeyFileManager {
...
 
 /// Keys file manager for root keys directory
-pub struct DiskKeyFileManager;
+pub struct DiskKeyFileManager {
+	password: Option<Password>,
+}
+
+impl DiskKeyFileManager {
+	/// Create new disk key manager w/o password
+	pub fn new() -> Self {
```
niklasad1: You could derive this impl instead with a default implementation which I prefer in this context!

ethstore/src/lib.rs
niklasad1: No, not after the change in this PR:  https://github.com/paritytech/parity-ethereum/blob/43c12d676d4b5fad7b4964d5d55b477087f33c86/ethstore/src/accounts_dir/disk.rs#L210

sorpaas: @niklasad1 `eprintln!` is from `std`. I see that `macro_use` is completely removed so we aren't using any logging macros from `log` crate?

sorpaas: Ah I see, you meant that we should change back to use `log` macros instead for those `eprintln`.

ethstore/src/json/crypto.rs
tworec: any constructive criticism @sorpaas? :)

ethstore/src/accounts_dir/disk.rs
tworec: because otherwise, the error message is swallowed by `.ok()` 4 lines below, and if you are running `ethstore/cli` user has no idea what went wrong since there is no logging enabled...

ethstore/src/lib.rs
tworec: I also opt to remove `log` crate since it is not used any more. Unless you want me to restore `warn!` where I've added `eprintln`

5chdn:
:+1: needs a 2nd review

ethstore/src/accounts_dir/disk.rs
niklasad1: Ok, I see but not the `ok()'s` fault IMO. The method is badly designed for your purpose because it only propagates the error if the  `Vec<PathBuf>` fails.

/cc  @sorpaas @debris are we concerned by the performance penalty that `eprintln!` comes with in this context? I'm not exactly sure when this is code is used in the `core client` but I guess it is when we sign transactions and similar?

tworec: I haven't expected that keyfiles may be read upon every sign (ECC) ops... some dive into parity code has not confirmed this, but if it is the case then I should restore `warn!`.

My first version of this enables logging in `ethstore/cli`. This is also acceptable to me.

tworec:
@niklasad1 done, but I've initialized log level to `warn` also. It will enable logs for invalid password.

This is sample CLI output with invalid password
```
parity-ethereum/ethstore/cli pyethereum-keystore-support (0282536a4)$ cargo run import pwd.txx --src ~/ethstore/
   Compiling ethstore-cli v0.1.1 (/Users/tworec/git/parity-ethereum/ethstore/cli)                                                                                                    
    Finished dev [unoptimized + debuginfo] target(s) in 3.30s                                                                                                                        
     Running `/Users/tworec/git/parity-ethereum/target/debug/ethstore import pwd.txx --src /Users/tworec/ethstore/`
 WARN 2018-12-19T11:05:15Z: ethstore::accounts_dir::disk: Invalid key file: "/Users/tworec/ethstore/keystore-mainnet-backup.json" (Invalid password)
 WARN 2018-12-19T11:05:15Z: ethstore::accounts_dir::disk: Invalid key file: "/Users/tworec/ethstore/mainnet_keystore.json" (Invalid password)
```

5chdn:
@tworec could you rebase this?

@niklasad1 could you  review this again?

tworec:
@5chdn I think rebase is overkill here. I've merged instead.

5chdn:
yes, whatever you prefer :)

tworec:
So why do we need a review? Merge introduced no changes in this PR

5chdn:
yes. one more review. please wait till this holiday craze is over.

tworec:
ok

accounts/ethstore/src/accounts_dir/disk.rs
```diff
@@ -117,16 +119,25 @@ pub struct DiskDirectory<T> where T: KeyFileManager {
...
+}
 
 impl RootDiskDirectory {
 	pub fn create<P>(path: P) -> Result<Self, Error> where P: AsRef<Path> {
 		fs::create_dir_all(&path)?;
 		Ok(Self::at(path))
 	}
 
+	/// allows to read keyfiles with given password (needed for keyfiles w/o address)
+	pub fn with_password(&self, password: Option<Password>) -> Self {
```
tomusdrw: I don't think it's a good idea to implement that here. Reasons:
1. The assumption is that all accounts inside the directory share the same password which is most likely incorrect.
2. It's used only for python keyfiles import, but will polute (already convoluted) codebase

Can we not use `DiskDirectory` for python keyfiles import and just allow importing particular account files?
Alternatively can't we treat python keyfiles as a vault (they share the same password actually) and use structs from this code part?

accounts/ethstore/src/account/safe_account.rs
```diff
@@ -77,16 +77,32 @@ impl SafeAccount {
...
-		SafeAccount {
+	pub fn from_file(json: json::KeyFile, filename: Option<String>) -> Result<Self, Error> {
+		SafeAccount::from_file_with_password(json, filename, &None)
+	}
+
+	/// Same as [`from_file`](#method.from_file), but with `password` which enables reading keystore files w/o address.
+	pub fn from_file_with_password(json: json::KeyFile, filename: Option<String>, password: &Option<Password>) -> Result<Self, Error> {
+		let crypto = Crypto::from(json.crypto);
+		let address = match json.address {
+			Some(address) => address.into(),
```
tomusdrw: imho if the `password` is provided we should validate that the `adress` is correct. Maybe switch the order here: first check if we have password and after deriving the address optionally compare it with the one in `json`.

accounts/ethstore/src/account/safe_account.rs
```diff
@@ -77,16 +77,32 @@ impl SafeAccount {
 	/// Create a new `SafeAccount` from the given `json`; if it was read from a
 	/// file, the `filename` should be `Some` name. If it is as yet anonymous, then it
 	/// can be left `None`.
-	pub fn from_file(json: json::KeyFile, filename: Option<String>) -> Self {
-		SafeAccount {
+	pub fn from_file(json: json::KeyFile, filename: Option<String>) -> Result<Self, Error> {
```
tomusdrw: This now errors for no good reason, afaict it will never fail. Can we please get the proper return type back?

accounts/ethstore/src/account/safe_account.rs
```diff
@@ -77,16 +77,32 @@ impl SafeAccount {
...
+		SafeAccount::from_file_with_password(json, filename, &None)
+	}
+
+	/// Same as [`from_file`](#method.from_file), but with `password` which enables reading keystore files w/o address.
+	pub fn from_file_with_password(json: json::KeyFile, filename: Option<String>, password: &Option<Password>) -> Result<Self, Error> {
```
tomusdrw: I would make `password` mandatory here (see above comment). The decision which function to use and what errors to handle can be left to the caller.

accounts/ethstore/cli/src/main.rs
```diff
@@ -146,30 +148,34 @@ impl fmt::Display for Error {
 
 fn main() {
 	panic_hook::set_abort();
+	if env::var("RUST_LOG").is_err() {
+		env::set_var("RUST_LOG", "warn")
```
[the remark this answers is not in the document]
tworec: nope, see https://docs.rs/env_logger/0.5.0/env_logger/#enabling-logging
> by default all logging is disabled except for `error!`

And I want to see [this warning](https://github.com/paritytech/parity-ethereum/pull/9710/files#diff-4b50b928e10747ccbb5b89b25be9e318R205):
`warn!("Invalid key file: {:?} ({})", path, err);`

see discussion above

accounts/ethstore/cli/src/main.rs
```diff
@@ -146,30 +148,34 @@ impl fmt::Display for Error {
...
+	}
+	env_logger::try_init().expect("Logger initialized only once.");
 
 	match execute(env::args()) {
 		Ok(result) => println!("{}", result),
 		Err(Error::Docopt(ref e)) => e.exit(),
 		Err(err) => {
-			println!("{}", err);
+			eprintln!("{}", err);
```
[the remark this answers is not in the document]
tworec: maybe, but see this https://github.com/paritytech/parity-ethereum/pull/9710#discussion_r224011177 by @debris 

guy's pls, can we just proceed and not nitpicking about error messages?

accounts/ethstore/src/accounts_dir/disk.rs
tworec: ad1. I agree, it is not very generic, but as a user I just create special dir for my keyfiles (even if its single one) before import.

ad 2. it is used for all keyfiles w/o address. Not only pyethereum ones. As any new feature it introduces some noice. 

> Can we not use DiskDirectory for python keyfiles import and just allow importing particular account files?

I've tried to be consistent with current approach which is importing whole directories. Introduction of new `import-single` in CLI seems overkill to me.

>Alternatively can't we treat python keyfiles as a vault (they share the same password actually) and use structs from this code part?

I've analyzed vault's, but it does not fit here. You need to repack your keyfiles into valuts, and in orted to repack you need to import it first, right?

accounts/ethstore/src/account/safe_account.rs
tworec: nice feature. It will need to throw an error when addresses do not match, but in next comment you're proposing to revert `Result`.

accounts/ethstore/src/account/safe_account.rs
tworec: it will fail for example when the password is incorrect.

accounts/ethstore/src/account/safe_account.rs
tworec: sure, it can be done, but for price of code repetition, because I use it to implement `from_file`

accounts/ethstore/src/accounts_dir/disk.rs
tworec: ad1. if your keyfiles have different pwds you will see warnings, but import will succeed with at least this file with `pwd` you've entered. You can reexecute `import` with another `pwd` to import others.

tomusdrw: We have `import-wallet`, but it seems it only works for presale wallets currently. I'd be more inclined to handle it there than to hack around `DiskKeyFileManager` (since the former command allows importing directly to a vault, etc).

That said I'm fine with this if you address the `SafeAccount` semantics in a way I proposed in the other comment.

tworec:
@tomusdrw you're suggestion is incorporated.
@niklasad1 can you pls approve?

PR #9717, opened 2018-10-08
Schedule nightly builds

gabreal:
with this pr one can add a scheduled pipeline with gitlab for building nightly.

the scheduled pipeline has to have the $SCHEDULE_TAG variable set to nightly in order to make this work. after that no nightly tags on the repositories itself are needed anymore.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,10 +24,13 @@ cache:
     - stable
     - beta
     - tags
+    variables:
+      - $SCHEDULE_TAG == "nightly"
+
 
 .collect_artifacts:                &collect_artifacts
   artifacts:
-    name:                          "${CI_JOB_NAME}_${CI_COMMIT_REF_NAME}"
+    name:                          "${CI_JOB_NAME}_${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}"
     when:                          on_success
     expire_in:                     1 mos
     paths:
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -21,13 +21,17 @@ cache:
...
   only:                            &releaseable_branches
-    - stable
-    - beta
-    - tags
+    refs:
+      - stable
+      - beta
+      - tags
+    variables:
+      - $SCHEDULE_TAG == "nightly"
```
5chdn: does this just return true if ref is nightly?

.gitlab-ci.yml
```diff
@@ -21,13 +21,17 @@ cache:
...
+
 
 .collect_artifacts:                &collect_artifacts
   artifacts:
-    name:                          "${CI_JOB_NAME}_${CI_COMMIT_REF_NAME}"
+    name:                          "${CI_JOB_NAME}_${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}"
```
5chdn: what happens if schedule_tag is empty?

scripts/gitlab/publish-awss3.sh
```diff
@@ -40,10 +40,15 @@ done
 echo "__________Push binaries to AWS S3____________"
 aws configure set aws_access_key_id $s3_key
 aws configure set aws_secret_access_key $s3_secret
-if [[ "$CI_COMMIT_REF_NAME" = "beta" || "$CI_COMMIT_REF_NAME" = "stable" || "$CI_COMMIT_REF_NAME" = "nightly" ]];
-  then
+
+case "${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}" in
+  (beta|stable|nightly)
```
5chdn: is that even possible?

.gitlab-ci.yml
gabreal: It returns true if the variable $SCHEDULE_TAG is set to nightly, which will only be the case if this variable is explicitly set either in scheduleded builds or in manually triggered bulids.
For $CI_COMMIT_REF_NAME the behaviour will remain the same as before when pushing a nightly tag (but not needed anymore) - otherwise this variable will be the name of the branch which for scheduled nightly builds will be master.

.gitlab-ci.yml
gabreal: this is posix parameter expansion ':-' does return the value of the variable $SCHEDULE_TAG if set, otherwise the value of $CI_COMMIT_REF_NAME} will be given.

scripts/gitlab/publish-awss3.sh
gabreal: for sure, tested :)
the value of the case statement is evaluated before matching the cases. but we can also keep the 3-fold if conditional if you like.

Issue #9684, opened 2018-10-01
2.1.1-beta crashes with "illegal instruction"


iSOcH:
- **Parity Ethereum version**: 2.1.1-beta / 2.0.6
- **Operating system**: Linux 4.10.0-42-generic #46~16.04.1-Ubuntu x86_64
- **Installation**: binary from release
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: yes

Wanted to launch my first ethereum full node for quite some time, but I cannot start the latest version of parity.

Call without arguments (but also with `--help`) crashes with error:
`Illegal instruction (core dumped)`

The failing binary has sha256sum `17ec8d1ae4ed4eeccf1ea142c396a9410dc18575ddeee82cfec1b5347f7b1668`. On another machine with a more recent Ubuntu installation, 2.1.1-beta seems to work so far.

dmesg on the affected system contains the following error when the binary is started:
```
[   81.316333] traps: parity[1216] trap invalid opcode ip:5629f0f6e933 sp:7fffd847d130 error:0
[   81.316340]  in parity[5629efb91000+26f8000]
```

The affected system seems to be able to run 2.0.6 without this problem.

I'm sure you need more information, please tell me how to gather that.

joshua-mir:
Running parity with the `--log-file` flag will let you specify where parity will save it's logs, and you can specify *what* to log with `--logging` or `-l` and specifying what to log. https://wiki.parity.io/FAQ.html#troubleshooting-yelp11 has a list of things that you can log, but I'd suggest running `-l debug` for now to get as much information as possible. 

Other projects around the web seem to have similar issues with that version of Ubuntu, you might want to try building from source and seeing if that works.

ordian:
Looks like you tried to run the binary on a machine with old x86_64 CPU, could you share some details of your CPU, e.g. does it support SSE instructions (`cat /proc/cpuinfo`)?

iSOcH:
@joshua-mir thx, i'll try those flags later on.

yes, @ordian, the cpu is not the newest (but it supportes multiple versions of SSE)

```processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 23
model name	: Intel(R) Core(TM)2 Quad CPU    Q8400  @ 2.66GHz
stepping	: 10
microcode	: 0xa0b
cpu MHz		: 2003.000
cache size	: 2048 KB
physical id	: 0
siblings	: 4
core id		: 0
cpu cores	: 4
[...]

PR #9725, opened 2018-10-09
produce portable binaries

ordian:
Fixes (when we'll create binaries) #9684.

With this change, we now produce portable binaries, but it may incur some perf degradation (I haven't measured it).
Maybe it's worth adding relnotes, that for ultimate performance it's now better to compile `parity-ethereum` from source with `PORTABLE=OFF` env var.

`cargo update -p parity-rocksdb-sys` downgraded `rand` due to rust-lang/cargo/issues/5702, I manually reverted that change.


ordian:
@5chdn is it too late to include in 2.1.2?

5chdn:
Not at all.

PR #9731, opened 2018-10-10
removed "rustup" & added new runner tag

fevo1971:
* added new tag **windows** for the new windows-runner(s)
* removed the *rustup* comand from build-script. rustversion & target installed via ansible
  (https://gitlab.parity.io/parity/devops/blob/master/ansible/roles/win-gitlab-runner/tasks/main.yml)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -85,6 +85,7 @@ build-windows:
     - sh scripts/gitlab/build-windows.sh
   tags:
    - rust-windows
+   - windows
   <<:                              *collect_artifacts
 
 publish-docker:
@@ -162,6 +163,7 @@ test-windows:
     - sh scripts/gitlab/test-all.sh stable
   tags:
    - rust-windows
+   - windows
 
 test-beta:
   stage:                           optional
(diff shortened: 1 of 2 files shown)


gabreal:
this will switch the windows build to the new azure windows 10 runners - in case something goes wrong one can simply apply the windows tag to the old runners :+1:

.gitlab-ci.yml
```diff
@@ -162,6 +163,7 @@ test-windows:
     - sh scripts/gitlab/test-all.sh stable
   tags:
    - rust-windows
+   - windows
```
5chdn: what's the difference?

5chdn: I mean, why two tags, just tag the new runners with rust-windows and fine?

General-Beck:
new tags for runners are planned, more standard and understandable. like `docker`, `darwin`, `windows` & `shell`

fevo1971:
ok, i changed the tag to (just) **windows** and added the new tag to the runner settings of *betelgeuse* and *bellatrix* so we can managed the migration from the old to the new runners nodes in gitlab instead of changing the *gitlab-ci.yml* in the future.

5chdn:
Hehe, but why aren't we just applying the rust-windows tag to the new runners? This way we won't need any changes to the current gitlab-ci YML? :)

ddorgan:
@5chdn good to go now? The tag name was reverted. This is required to test the new windows CI hosts..

PR #9732, opened 2018-10-10
 remove unused expired value from Handshake

debris:
no logic has been changed in this pull request


niklasad1:
@debris 

The ```expired``` flag is removed because it was actually never enabled?

debris:
> The expired flag is removed because it was actually never enabled?

@niklasad1 yes

PR #9734, opened 2018-10-11
Make checkpoint_storage_at use plain loop instead of recursion

### LINKED ISSUE #9506: Change checkpoint_storage_at downward recursion to a plain loop
It has an upper bound (1024) so it shouldn't be an issue, but we can change it to a plain loop just to make sure.

https://github.com/paritytech/parity-ethereum/pull/9319/files#diff-fff79319205229b71cd0b01cdcb4eedbR616

sorpaas:
Fixes #9506

Unwrap the recursion at checkpoint_storage_at to be a plain loop.

### DIFF
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -572,9 +572,7 @@ impl<B: Backend> State<B> {
 
 			let mut kind = None;
 
-			for checkpoint_index in start_checkpoint_index..checkpoints.len() {
-				let checkpoint = &checkpoints[checkpoint_index];
-
+			for checkpoint in checkpoints.iter().skip(start_checkpoint_index) {
 				match checkpoint.get(address) {
 					// The account exists at this checkpoint.
 					Some(Some(AccountEntry { account: Some(ref account), .. })) => {
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -553,62 +553,72 @@ impl<B: Backend> State<B> {
 	}
 
 	/// Get the value of storage at a specific checkpoint.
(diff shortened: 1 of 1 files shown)


jimpo:
LGTM

ethcore/src/state/mod.rs
```diff
@@ -553,62 +553,72 @@ impl<B: Backend> State<B> {
...
+							// account's original storage root. Otherwise, the account must be a newly created contract.
+							if account.base_storage_root() == self.original_storage_root(address)? {
+								kind = Some(ReturnKind::OriginalAt);
+								break
+							} else {
+								// If account base storage root is different from the original storage root since last
+								// commit, then it can only be created from a new contract, where the base storage root
+								// would always be empty. Note that this branch is actually never called, because
+								// `cached_storage_at` handled this case.
+								warn!("Trying to get an account's cached storage value, but base storage root does not equal to original storage root! Assuming the value is empty.");
```
ordian: Should we use target: "state" here?

ethcore/src/state/mod.rs
```diff
@@ -553,62 +553,72 @@ impl<B: Backend> State<B> {
...
-						// This account has checkpoint entry, but the key is not in the entry's cache. We can use
-						// original_storage_at if current account's original storage root is the same as checkpoint
-						// account's original storage root. Otherwise, the account must be a newly created contract.
-						if account.base_storage_root() == self.original_storage_root(address)? {
-							ReturnKind::OriginalAt
+			let mut kind = None;
+
+			for checkpoint_index in start_checkpoint_index..checkpoints.len() {
+				let checkpoint = &checkpoints[checkpoint_index];
+
```
niklasad1: Simplify this by: ```for checkpoint in checkpoints.iter().skip(start_checkpoint_index) { }```

PR #9744, opened 2018-10-12
Implement NoProof for json tests and update tests reference

### LINKED ISSUE #9590: Currently failing JSON tests
** Note **

Currently this issue refers to an unmerged PR #9570 (early creation to allow reference from the exception json). It should only be considered active after #9570.

** end Note **

Umbrella issue to follow current json-test issues.

All those failing test are currently **skipped** by the CI.

When one test is fixed (either by an update of ethereum/test submodule or an actual parity fix), the corresponding skipped exception should be removed from the [file](https://github.com/paritytech/parity-ethereum/ethcore/res/ethereum/test-issues/currents.json) configuring those exceptions.

Current tags to qualify those issues are:

- `ToInvestigate` : no investigation done yet.
- `IssueWaiting` : the issue is likely to be solve in future ethereum/test generations. 

An indicative number of failure is also added. 

**Last synch**

https://github.com/ethereum/tests/commit/3f5febc901913ef698f1b09dda8705babd729e4a

**Block issues**

- [ ] stCreateTest : `ToInvestigate` : 1
- [ ] stCreate2 : `IssueWaiting` : 151 (all), the issue seems to be present for other client to (see ethereum/tests gitter).
- [ ] bcStateTest: `ToInvestigate` : 10  Note that those are not only for constantinople and are related to contract sucide #9440 could be related or a bugfix from #9319
- [ ] stEIP158Specific: `ToInvestigate` : 3
- [ ] stPreCompiledContracts: `ToInvestigate` : 2
- [ ] stReturnDataTest: `ToInvestigate` : 3
- [ ] stReturnDataTest: `ToInvestigate` : 3
- [ ] stSpecialTest: `ToInvestigate` : 1 ('push32withoutByte_d0g0v0_Constantinople')

**State issues**

- [ ] `stCreateTest`: `IssueWaiting` : 1 (passing with https://github.com/ethereum/tests/pull/511 at baf0eb45c1c02fb56612028bef3ce50932779c05)
- [ ] `stCreateTest2`: `ToInvestigate` : 1 (RevertInCreateInInitCreate2 - constantiople 1/1)
- [ ] `stEIP150Specific`: `IssueWaiting` : 1 (passing with https://github.com/ethereum/tests/pull/511 at baf0eb45c1c02fb56612028bef3ce50932779c05)
- [ ] `stInitCodeTest`: `IssueWaiting` : 1 Running test with https://github.com/ethereum/tests/pull/511 at 67f90e893326f9bde359e74492313269708d5e9a seems to fix it
- [ ] `stPreCompiledContracts` : `IssueWaiting` : 80 all modexp test failling. Running test with https://github.com/ethereum/tests/pull/511 at 67f90e893326f9bde359e74492313269708d5e9a seems to fix it
- [ ] `stRevertTest` : `ToInvestigate` : 14 (all constantinople related). (only 1 error ('RevertInCreateInInit') with https://github.com/ethereum/tests/pull/511 at baf0eb45c1c02fb56612028bef3ce50932779c05)
- [ ] `stRevertTest - RevertOpcodeMultipleSubCalls` : `IssueWaiting` >20 (passing with https://github.com/ethereum/tests/pull/511 at baf0eb45c1c02fb56612028bef3ce50932779c05)
- [ ] `stStaticCall` : `IssueWaiting` : 14 (all constantinople related) . Running test with https://github.com/ethereum/tests/pull/511 at 67f90e893326f9bde359e74492313269708d5e9a seems to fix it, but at next commit baf0eb45c1c02fb56612028bef3ce50932779c05 back to 13. Ok at  8a1e249a739aa0f2f3532a679993067b3bbb8f3c.
- [ ] `stZeroKnowledge` : `IssueWaiting` : all https://github.com/ethereum/tests/issues/512

cheme:
Closes #9590 .

Update to latest ethereum/tests .
Fix cost for constantinople test json.
Implements 'NoProof' engine from ethereum/tests#464 .

At this point with this PR everything pass except the case spotted by /cc @holiman this week in https://gist.github.com/holiman/0154f00d5fcec5f89e85894cbb46fcb2 . 

I skip those tests. For completeness, @sorpaas indicated that it is not an easy fix : so since those case cannot happen on a mainchain I did not open and reference an issue for them (in the json defining tests to skip I only added a comment but not issue reference).

Note for reviewer:
- when running NoProof with CanonNoSeal, the upper gas limit verification is required : I switch it from verify_block_basic to the header check. I got  to put another trait method for it, a lighter change would be to make this check for every engine but it would change a bit the existing behavior (might not be an issue).


ethcore/src/json_tests/chain.rs
```diff
@@ -114,6 +117,72 @@ pub fn json_chain_test<H: FnMut(&str, HookType)>(json_data: &[u8], start_stop_ho
...
+
+			if header.gas_limit() > &0x7fffffffffffffffu64.into() {
+				return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: None, max: Some(0x7fffffffffffffffu64.into()), found: header.gas_limit().clone() })));
+			}
+
+			Ok(())
+		}
+
+		/// No nonce handle, skip test.
+		fn verify_block_unordered(&self, _header: &Header) -> Result<(), Error> {
```
sorpaas: What's the difference of this and `VerifierType::CanonNoSeal`? It looks like here in addition we skipped `verify_block_unordered` transaction basics verification, which I worry would give us some true negatives of failing tests.

cheme: Yes 'check_seal' false will probably do it, I'll check that on monday, on ice till then.

5chdn:
we need to backport this, don't we?

cheme:
@sorpaas I just fix it properly, basically no handling of NoProof for one and change of querried blocks for the other.

@5chdn it could be good to backport, just to be able to run latest json-tests suite.

ethcore/src/ethereum/ethash.rs
```diff
@@ -322,10 +324,6 @@ impl Engine<EthereumMachine> for Arc<Ethash> {
 			return Err(From::from(BlockError::InvalidProofOfWork(OutOfBounds { min: Some(header.difficulty().clone()), max: None, found: difficulty })));
 		}
 
-		if header.gas_limit() > &0x7fffffffffffffffu64.into() {
```
sorpaas: Just a note on this: this would allow miner to craft blocks that have difficulty higher than this limit, but later when this block is importing, it will be rejected. It means that this will be one more place where if a miner configured its client incorrectly, mining will be wasted.

cheme: I did not think about that, I am restoring this test, it will be redundant with header check, but it is not a costy test.

sorpaas: I think this should be fine right now, and I don't think this would be an issue for any live networks. But just wanted to write a note here in case anything wrong happens. :)

cheme:
I close this request (I really messed up my rebase), #9814 replaces this.

Issue #9713, opened 2018-10-07
Feature request: parity_allTransactionHashes / parity_watchPendingTransactions


fanatid:
- **Parity Ethereum version**: 2.0.5
- **Operating system**: Linux
- **Installation**: binaries from vanity-service.parity.io
- **Fully synchronized**: yes
- **Network**: ethereum

My Parity instance run with `--tracing on --tx-queue-size=20000`

I'd like watch on transactions in mempool, I can do this with [eth_newPendingTransactionFilter](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newpendingtransactionfilter), but there is no filter for removed transactions, so I decide call [parity_allTransactions](https://wiki.parity.io/JSONRPC-parity-module#parity_alltransactions) in loop ([parity_pendingTransactions](https://wiki.parity.io/JSONRPC-parity-module#parity_pendingtransactions) is very slow).

My problem is, that `parity_allTransactions` is slow too, if I run Parity with default tx queue size (8192) -- call usually takes ~0.2-0.3s, but with mempool 20k call takes up to 1.5s. Is there any ways how improve performance? Or maybe right way for observing on transactions in mempool? :)

Command for testing:
```bash
$ time curl -s -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"parity_allTransactions","params":[],"id":0}' http://localhost:8545 | jq .result | jq length
18860

real	0m1.456s
user	0m1.269s
sys	0m0.137s
```

Thank you!

Tbaut:
Can you tell precisely what you wish to achieve? 
>I'd like watch on transactions in mempool

What do you mean exactly?

fanatid:
I'm working on service which provide API to ethereum wallets and for notifying users about new transactions (or removed transactions) I need watch on mempool. I can proceed with default mempool size (8192), but from other side I'd like have bigger mempool in Parity, because it's big mess when our Parity node kick tx from mempool and etherscan still show that this tx exists.
Initially I used `eth_newPendingTransactionFilter`, but switched to `parity_allTransactions` because want hold service mempool in actual state.

Tbaut:
What you wish to do sounds like an edge case.
`parity_allTransactions` taking time with a 20K pool is probably expected. 
>and etherscan still show that this tx exists

This can and will happen even if the time to query the mempool is reduced.

@tomusdrw can you confirm that this order of performance is expected?

tomusdrw:
[...]

PR #9745, opened 2018-10-13
RPC: parity_allTransactionHashes

fanatid:
Issue: https://github.com/paritytech/parity-ethereum/issues/9713
Sorry, probably you will say that code is shitty and I'll agree :) this is my first attempt write something on Rust and I made quick patch.
I tested with pool size 20k, `parity_allTransactionHashes` need ~100-200ms.
I'll try make measurements how `eth_getTransactionByHash` for new transactions will slow process when will have time.

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -50,13 +50,11 @@ use transaction::{Action, Transaction as EthTransaction, PendingTransaction, Sig
 use v1::helpers::{CallRequest as CallRequestHelper, errors, dispatch};
 use v1::types::{BlockNumber, CallRequest, Log, Transaction};
 
-pub use self::dispatch::{LightDispatcher};
-
 const NO_INVALID_BACK_REFS_PROOF: &str = "Fails only on invalid back-references; back-references here known to be valid; qed";
 
 const WRONG_RESPONSE_AMOUNT_TYPE_PROOF: &str = "responses correspond directly with requests in amount and type; qed";
 
-pub fn light_all_transactions(dispatch: &Arc<LightDispatcher>) -> impl Iterator<Item=PendingTransaction> {
+pub fn light_all_transactions(dispatch: &Arc<dispatch::LightDispatcher>) -> impl Iterator<Item=PendingTransaction> {
 	let txq = dispatch.transaction_queue.read();
 	let chain_info = dispatch.client.chain_info();
 
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -45,14 +45,25 @@ use ethereum_types::{U256, Address};
 use hash::H256;
 use parking_lot::Mutex;
 use fastmap::H256FastMap;
-use transaction::{Action, Transaction as EthTransaction, SignedTransaction, LocalizedTransaction};
+use transaction::{Action, Transaction as EthTransaction, PendingTransaction, SignedTransaction, LocalizedTransaction};
 
 use v1::helpers::{CallRequest as CallRequestHelper, errors, dispatch};
 use v1::types::{BlockNumber, CallRequest, Log, Transaction};
 
+pub use self::dispatch::{LightDispatcher};
+
 const NO_INVALID_BACK_REFS_PROOF: &str = "Fails only on invalid back-references; back-references here known to be valid; qed";
 
 const WRONG_RESPONSE_AMOUNT_TYPE_PROOF: &str = "responses correspond directly with requests in amount and type; qed";
+
+pub fn light_all_transactions(dispatch: &Arc<LightDispatcher>) -> impl Iterator<Item=PendingTransaction> {
+	let txq = dispatch.transaction_queue.read();
+	let chain_info = dispatch.client.chain_info();
+
+	let current = txq.ready_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+	let future = txq.future_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+	current.into_iter().chain(future.into_iter())
+}
 
 /// Helper for fetching blockchain data either from the light client or the network
 /// as necessary.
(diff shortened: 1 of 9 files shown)


rpc/src/v1/impls/light/parity.rs
```diff
@@ -277,6 +277,21 @@ impl Parity for ParityClient {
...
+		let current = txq.ready_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+		let future = txq.future_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+		Ok(
+			current
+				.into_iter()
+				.chain(future.into_iter())
+				.map(|tx| Transaction::from_pending(tx).hash.into())
+				.collect::<Vec<_>>()
+		)
+	}
```
fanatid: @seunlanlege also here is copy-paste from `all_transactions`, can this be solved through `all_transactions` or some shared function? or current is ok?

seunlanlege: you can move the shared logic into a function in `helpers/mod.rs` that returns the chained iterator of `ready_transactions` and `future_transactions`

```rust
/// returns an Iterator of pending transactions, intended for use by light clients
pub fn all_transactions(ligh_dispatch: &Arc<LightDispatch>) -> impl Iterator<Item = PendingTransaction> {
    /// shared logic here, return the chained iterator
}
```

then for `all_transaction_hashes` you can `map` out the hashes, and for `all_transactions` you can `map` into `Transaction::from_pending`

seunlanlege:
> @seunlanlege not sure that I fixed it, how I should run build locally for receiving same error?

rebuild the whole code base

5chdn:
Does not compile.

fanatid:
Should be fixed now.

rpc/src/v1/impls/light/parity.rs
seunlanlege: what you can do is move the shared logic into `helpers/all_transactions.rs`

basically have a function that returns an iterator of all transactions. 

```rust
pub fn all_transactions(dispatch: &Arc<LightDispatch>) -> impl Iterator<Item=PendingTransaction> {
    // fetch future_transactions and ready_transactions here
    // return the chained iterator of both
}
```

Then you can use `helpers::all_transactions()` depending on the rpc call

for `all_transaction_hashes`

```rust
fn all_transaction_hashes(&self) -> Result<Vec<H256>> {
   // personally i would prefer you access the hash directly on PendingTransaction instead of converting to Transaction first
    helpers::all_transactions(&self.light_dispatch)
        .map(|tx| Transaction::from_pending(tx).hash.into())
        .collect()
}
```

`all_transactions`

```rust
fn all_transactions(&self) -> Result<Vec<Transaction>> {
    helpers::all_transactions(&self.light_dispatch)
        .map(|tx| Transaction::from_pending(tx))
        .collect()
}
```

fanatid: Thanks for examples, this was very helpful. Not sure about helper location, but should be done now.

5chdn:
builds fail

fanatid:
=( I'd like fix it, but can you give a tip how I can debug build fail locally?

5chdn:
Sorry, the issue was on our side, restarted the jobs. Ignore me

rpc/src/v1/helpers/mod.rs
```diff
@@ -36,7 +36,11 @@ mod subscribers;
 mod subscription_manager;
 mod work;
 
-pub use self::dispatch::{Dispatcher, FullDispatcher};
+use std::sync::Arc;
+
+use transaction::{PendingTransaction};
```
niklasad1: Single `use` doesn't need parenthesis

fanatid: Fixed

rpc/src/v1/helpers/mod.rs
```diff
@@ -56,3 +60,12 @@ pub use self::work::submit_work_detail;
...
 }
+
+pub fn light_all_transactions(dispatch: &Arc<LightDispatcher>) -> impl Iterator<Item=PendingTransaction> {
+	let txq = dispatch.transaction_queue.read();
+	let chain_info = dispatch.client.chain_info();
+
+	let current = txq.ready_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+	let future = txq.future_transactions(chain_info.best_block_number, chain_info.best_block_timestamp);
+	current.into_iter().chain(future.into_iter())
+}
```
niklasad1: hrm, I would prefer to have this in a separate module such as `light_fetch` not intuitive to me to have this in the `root` to have a clear distinction between `light` and `full`

5chdn:
@niklasad1 like this? https://github.com/paritytech/parity-ethereum/pull/9745/commits/6ab1af00ecb6aea66a330ae966581b636e4f0c7f

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -45,15 +45,26 @@ use ethereum_types::{U256, Address};
 use hash::H256;
 use parking_lot::Mutex;
 use fastmap::H256FastMap;
-use transaction::{Action, Transaction as EthTransaction, SignedTransaction, LocalizedTransaction};
+use transaction::{Action, Transaction as EthTransaction, PendingTransaction, SignedTransaction, LocalizedTransaction};
 
 use v1::helpers::{CallRequest as CallRequestHelper, errors, dispatch};
 use v1::types::{BlockNumber, CallRequest, Log, Transaction};
 
+pub use self::dispatch::{LightDispatcher};
```
niklasad1: One last thing, this doesn't have to be public so I suggest 

```suggestion
use dispatch::LightDispatcher;
```

fanatid: Thanks, I noticed that `dispatch` already defined, so removed this line at all.

fanatid:
do you need rebase/squash?

niklasad1:
Since we are squash-merging and we don't have any conflicts it should be fine

PR #9746, opened 2018-10-14
Use signed 256-bit integer for sstore gas refund substate 

sorpaas:
While the overall transaction sstore gas refund cannot go below zero, on an individual execution frame it can. This replaces `saturating_sub` by a signed integer version to fix the issue.


ethcore/src/signed.rs
```diff
@@ -0,0 +1,158 @@
...
+		self.add_assign(other);
+		self
+	}
+}
+
+impl AddAssign<U256> for I256 {
+	fn add_assign(&mut self, other: U256) {
+		match self.0 {
+			Sign::Positive => {
+				self.1 += other;
```
tomusdrw: should we care about overflows/panics here (and in other cases/SubAssign)?

ethcore/src/executive.rs
```diff
@@ -1091,7 +1091,8 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 		let schedule = self.schedule;
 
 		// refunds from SSTORE nonzero -> zero
-		let sstore_refunds = substate.sstore_clears_refund;
+		assert!(substate.sstore_clears_refund.is_nonnegative(), "On transaction level, sstore clears refund cannot go below zero.");
```
cheme: Setting to 0 in case of negative plus a trace could be another way to do it.

ethcore/src/signed.rs
```diff
@@ -0,0 +1,158 @@
...
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum Sign {
+	Positive,
+	Zero,
+	Negative
+}
+
+/// Representation of a signed 256-bit integer.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub struct I256(Sign, U256);
```
cheme: I would not call it I256 but U256_And_Signed or anything that shows that we are not on 256bit only.

ethcore/src/externalities.rs
```diff
@@ -395,11 +395,11 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	}
 
 	fn add_sstore_refund(&mut self, value: U256) {
-		self.substate.sstore_clears_refund = self.substate.sstore_clears_refund.saturating_add(value);
```
cheme: We are loosing the saturating property here, maybe it could be saturating operation in I256 implementation by default (we need to renaim I256 to make it clear that it does not behave as U256)

cheme: I few nit, maybe also some super basic arithmetic tests case on I256 would be welcome.

ethcore/src/executive.rs
sorpaas: I would rather panic here -- our recent experience shows that it's easier to find consensus bug this way, and if negative ever happens (which we do have some informal proof that it shouldn't), then the network is broken anyway!

ethcore/src/executive.rs
```diff
@@ -1091,7 +1091,8 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 		let schedule = self.schedule;
 
 		// refunds from SSTORE nonzero -> zero
-		let sstore_refunds = substate.sstore_clears_refund;
+		assert!(substate.sstore_clears_refund >= 0, "On transaction level, sstore clears refund cannot go below zero.");
+		let sstore_refunds = U256::from(substate.sstore_clears_refund as u64);
```
sorpaas: `i128::max_value() as u64 == u64::max_value()`, and we checked above that it's always non-negative.

cheme:
Seems good using i128, but I would also switch sstore_refund and others usize parameters to explicit u64.

sorpaas:
@cheme The issue is that all our gas definitions use `usize`. So if we do that it means we do another casting from `usize` to `u64`.

PR #9752, opened 2018-10-15
Ignore global cache for patched accounts

### LINKED ISSUE #9494: [PrivateTransactions] wrong return value when calling not yet initialized variables of a private smart contract
- **Parity Ethereum version**: Parity-Ethereum/v2.0.3-unstable-a0a2beddf-20180831/x86_64-macos/rustc1.27.0
- **Operating system**:   MacOS 
- **Network**: dev
- **Restarted**: yes

Currently, when trying to call a not yet initialized variable of a private Contract it will return some value (in the tutorial it returned the number of validators). However, a not yet initialized variable should always return 0x0 or false. 

see documentation of secret transactions: 
https://github.com/paritytech/wiki/blob/master/Private-Transactions-Tutorial-4.md

> The result is the current value of x. As it wasn't initialized, it defaults to 1.
> {"jsonrpc":"2.0","result":"0x0000000000000000000000000000000000000000000000000000000000000001","id":1}

grbIzl:
For private transactions we patch the state in order to apply encrypted data upon it. Such state is not commited and used only in order to calculate the results of the private transaction. In terms of this PR patched state was introduced explicitly for the state and the account in order to work with cache properly. For such patched accounts only local cache should be used (as it was changed during "patching").

Closes #9494

### DIFF
--- a/ethcore/src/state/account.rs
+++ b/ethcore/src/state/account.rs
@@ -76,9 +76,6 @@ pub struct Account {
 	code_filth: Filth,
 	// Cached address hash.
 	address_hash: Cell<Option<H256>>,
-	// Patched account used to operate on the patched state with all changes in memory only
-	// Such accounts should not be committed
-	patched_account: bool,
 }
 
 impl From<BasicAccount> for Account {
@@ -95,7 +92,6 @@ impl From<BasicAccount> for Account {
 			code_cache: Arc::new(vec![]),
 			code_filth: Filth::Clean,
 			address_hash: Cell::new(None),
-			patched_account: false,
 		}
 	}
 }
@@ -116,7 +112,6 @@ impl Account {
 			code_cache: Arc::new(code),
 			code_filth: Filth::Dirty,
 			address_hash: Cell::new(None),
-			patched_account: false,
 		}
 	}
 
@@ -138,7 +133,6 @@ impl Account {
 			code_size: Some(pod.code.as_ref().map_or(0, |c| c.len())),
 			code_cache: Arc::new(pod.code.map_or_else(|| { warn!("POD account with unknown code is being created! Assuming no code."); vec![] }, |c| c)),
 			address_hash: Cell::new(None),
-			patched_account: false,
 		}
 	}
 
@@ -156,7 +150,6 @@ impl Account {
 			code_size: Some(0),
 			code_filth: Filth::Clean,
 			address_hash: Cell::new(None),
-			patched_account: false,
 		}
 	}
 
@@ -186,7 +179,6 @@ impl Account {
 			code_size: None,
 			code_filth: Filth::Clean,
 			address_hash: Cell::new(None),
-			patched_account: false,
 		}
 	}
 
@@ -212,16 +204,7 @@ impl Account {
 		self.code_filth = Filth::Dirty;
 		self.storage_cache = Self::empty_storage_cache();
 		self.storage_changes = storage;
-	}
-
-	/// Swtich account to the patched mode
-	pub fn switch_to_patched(&mut self) {
-		self.patched_account = true;
-	}
-
-	/// Returns patched mode flag
-	pub fn patched(&self) -> bool {
-		self.patched_account
+		self.storage_root = KECCAK_NULL_RLP;
 	}
 
 	/// Set (and cache) the contents of the trie's storage at `key` to `value`.
@@ -280,10 +263,6 @@ impl Account {
 	pub fn cached_storage_at(&self, key: &H256) -> Option<H256> {
 		if let Some(value) = self.storage_changes.get(key) {
 			return Some(value.clone())
-		}
-		//Use only in memory state for the patched account
-		if self.patched_account {
-			return None;
 		}
 		self.cached_moved_original_storage_at(key)
 	}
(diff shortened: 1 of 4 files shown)


5chdn:
Please review @andresilva @cheme :)

ethcore/src/state/mod.rs
```diff
@@ -1154,7 +1158,9 @@ impl<B: Backend> State<B> {
 
 	/// Replace account code and storage. Creates account if it does not exist.
 	pub fn patch_account(&self, a: &Address, code: Arc<Bytes>, storage: HashMap<H256, H256>) -> TrieResult<()> {
-		Ok(self.require(a, false)?.reset_code_and_storage(code, storage))
+		let mut account = self.require(a, false)?;
+		account.switch_to_patched();
+		Ok(account.reset_code_and_storage(code, storage))
```
cheme: I got a question regarding 'reset_code_and_storage' : currently it does not touch account fields 'original_storage_cache', shouldn't we (I am not 100% sure but it seems safer) ?

In fact the PR seems to try to ignore lru storage cache since storage is ignored in private_tx, but to get eip-1283 running we probably want to keep trace of original_storage_value, I am starting to wonder how it will behave.

ethcore/src/state/account.rs
```diff
@@ -263,6 +281,10 @@ impl Account {
 		if let Some(value) = self.storage_changes.get(key) {
 			return Some(value.clone())
 		}
+		//Use only in memory state for the patched account
+		if self.patched_account {
```
cheme: That is probably true for others function ('storage_at' will query the triedb for instance), do we need this kind of guard for them too?

I wonder if it wouldn't be ok to use the storage cache and original cache value (init them in 'reset_code_and_storage') and just prevent querying the triedb at 'get_and_cache_storage' ?

ethcore/src/state/account.rs
```diff
@@ -206,6 +214,16 @@ impl Account {
 		self.storage_changes = storage;
 	}
 
+	/// Swtich account to the patched mode
+	pub fn switch_to_patched(&mut self) {
+		self.patched_account = true;
```
cheme: Reseting some account values here could make sense. Maybe this function could be combine with 'patch_account' somehow to avoid strange account state if future code forgot to call 'patch_account'.

andresilva:
Asked review from @sorpaas since I don't know the implications this has related to EIP-1283 (as @cheme mentioned).

ethcore/src/state/account.rs
grbIzl: I was thinking heavily on how to achieve this, because I also don't like the this kind of dangling state for account. Didn't find a good solution. This information (if account is patched) is required both in state and account, but the external code works with the state only. So in such constraints this was the only suitable solution (IMO). 
Meanwhile 'forget' to patch account is pretty difficult, the more dangerous situation may appear, if someone starts to use patched accounts as regular ones. Here I tried to prevent this with asserts in commit methods.

ethcore/src/state/account.rs
grbIzl: Actually it is the guard for storage_at, because it directly calls cached_storage_at:

pub fn storage_at(&self, db: &HashDB<KeccakHasher, DBValue>, key: &H256) -> TrieResult<H256> {
		if let Some(value) = self.cached_storage_at(key) {
			return Ok(value);
		}
		Self::get_and_cache_storage(
			&self.storage_root,
			&mut self.storage_cache.borrow_mut(),
			db,
			key)
	}

ethcore/src/state/mod.rs
grbIzl: My understanding of eip-1283 is gas optimization of SSTORE calls. For this case original cache may be required. In terms of this PR I fixed caching logic for SLOAD, which doesn't use original cache. 
But again it's only my understanding. I don't know, how original cache can be further evolved. So it would be great, if @sorpaas can comment on this.

ethcore/src/state/account.rs
cheme: I think I am missing some code/context, I only find 'switch_to_patched' function call in 'patch_account' account function, so I was simply thinking of removing this fn and make the bool switch in account fn 'reset_code_and_storage' (which seems to be also called at a single place just after switch_to_patched).

ethcore/src/state/account.rs
cheme: I think 'Self::get_and_cache_storage' will be call due to this function returning 'None' value on missing key, then it will query the triedb (from the 'storage_at' code).

ethcore/src/state/mod.rs
cheme: I imagine plugging the encrypted state as a replacement for the trieDB storage could make it run ok, but there may be simpler way to do it.
Unrelated question, about those private contract, what are the limitation, it seems logic that they cannot change the state of other contracts (kinda break the confidentiality), but are they allowed to call external contract?

ethcore/src/state/account.rs
```diff
@@ -474,6 +496,7 @@ impl Account {
 
 	/// Commit the `storage_changes` to the backing DB and update `storage_root`.
 	pub fn commit_storage(&mut self, trie_factory: &TrieFactory, db: &mut HashDB<KeccakHasher, DBValue>) -> TrieResult<()> {
+		assert!(!self.patched_account, "Patched account should not be committed");
```
sorpaas: Why are we sure here that patched account won't be committed?

grbIzl:
@sorpaas Thanks for help! Yes, resetting storage root does the trick. I reverted my changes in account logic.

As for your question: we have a regular smart contract - "wrapper" for the private contract, that stores private contract's state and code (encrypted). We call this wrapper "public contract". For the execution, we temporarily patch public contract's state and code with decrypted state and code from private contract and run transaction on such patched state. Nonce and balance during execution taken from the public contract, they are in a way "shared" between private and public contracts. So we don't need to care about them for this code. Do you see any issues with such approach?

sorpaas:
@grbIzl Thanks for the explanation. Sounds good! I don't see any issues at least for now. :)

PR #9756, opened 2018-10-15
fix(light/response) : handle bad responses

niklasad1:
#### This PR adds the following:

* Closes https://github.com/paritytech/parity-ethereum/issues/9133
* Registers empty responses as a `bad`
* Register bad responses or at least reponses that are validated as bad and only takes a decision when the  majority of the peers have come to the same conclusion.

This change the assumptions slightly i.e, before we assumed that all requests are valid and only dropped on a timeout or too many request attempts to each peer.

Might be overlapping abit with the timeouts and I plan the clean it up that code abit if/after this gets merged! Should be sufficient with this and timeouts only

### Unresolved questions:
How this will work `in a real``` network only verified by unit tests and running a light and full node locally

### DIFF
--- a/ethcore/light/src/on_demand/mod.rs
+++ b/ethcore/light/src/on_demand/mod.rs
@@ -19,7 +19,7 @@
 //! will take the raw data received here and extract meaningful results from it.
 
 use std::cmp;
-use std::collections::{HashMap, BTreeSet};
+use std::collections::{HashMap, HashSet, BTreeSet};
 use std::marker::PhantomData;
 use std::sync::Arc;
 
@@ -41,6 +41,9 @@ use request::{self as basic_request, Request as NetworkRequest};
 use self::request::CheckedRequest;
 
 pub use self::request::{Request, Response, HeaderRef};
+
+use types::request::ResponseError;
+use self::request::Error as ValidityError;
 
 #[cfg(test)]
 mod tests;
@@ -69,6 +72,12 @@ pub mod error {
 		}
 
 		errors {
+			#[doc = "Request was faulty"]
+			FaultyRequest(req_id: super::ReqId, bad_responses: usize, num_providers: usize) {
+				description("Faulty request found")
+				display("The request: {} was determined as faulty, {}/{} peer(s) gave bad response", req_id, bad_responses, num_providers)
+			}
+
 			#[doc = "Max number of on-demand query attempts reached without result."]
 			MaxAttemptReach(query_index: usize) {
 				description("On-demand query limit reached")
@@ -124,6 +133,10 @@ struct Pending {
 	required_capabilities: Capabilities,
 	responses: Vec<Response>,
 	sender: oneshot::Sender<PendingResponse>,
+	/// This will collect how many bad responses we get from each peer per request
+	// When we get `|bad_responses| > peers / 2` then regard the request as `faulty`
+	// This, can happen for several reasons such as a request for a hash that doesn't exist
+	bad_responses: HashSet<PeerId>,
 	base_query_index: usize,
 	remaining_query_count: usize,
 	query_id_history: BTreeSet<PeerId>,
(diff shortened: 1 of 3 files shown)


ethcore/light/src/on_demand/mod.rs
```diff
@@ -608,12 +664,53 @@ impl Handler for OnDemand {
...
+						| ValidityError::WrongHash(_, _)
+						| ValidityError::WrongHeaderSequence => {
+							let peer = ctx.peer();
+							pending.add_bad_response(peer);
+							debug!(target: "on_demand", "Peer {} gave bad response", peer);
+							ctx.disable_peer(peer);
+						}
+					}
+				}
+
```
niklasad1: Verbosity, I think it helps to know which errors can occur

ethcore/light/src/on_demand/mod.rs
```diff
@@ -608,12 +664,53 @@ impl Handler for OnDemand {
...
-			if let Err(e) = pending.supply_response(&*self.cache, response) {
-				let peer = ctx.peer();
-				debug!(target: "on_demand", "Peer {} gave bad response: {:?}", peer, e);
-				ctx.disable_peer(peer);
+			match pending.supply_response(&*self.cache, response) {
+				Err(ResponseError::Validity(err)) => {
+					match err {
+						// This can be a malformed request or bad response but we can't determine which at the moment!
+						// Thus, register the response as bad and wait for more responses in order to take a decision
+						ValidityError::BadProof | ValidityError::Empty => {
```
niklasad1: We can get a `BadProof` for an execution on an account with an insufficient balance so better to be sure when we punish peers!

ethcore/light/src/on_demand/tests.rs
```diff
@@ -18,6 +18,7 @@
 
 use cache::Cache;
 use ethcore::header::Header;
+use ethcore::encoded::Header as EncodedHeader;
```
niklasad1: not used remove!!

ethcore/light/src/on_demand/tests.rs
```diff
@@ -28,7 +29,7 @@ use ::request::{self as basic_request, Response};
 
 use std::sync::Arc;
 
-use super::{request, OnDemand, Peer, HeaderRef};
+use super::{request, OnDemand, Peer, HeaderRef, ResponseError, ValidityError};
```
niklasad1: not used remove!

5chdn:
Please label your PR :)

ethcore/light/src/on_demand/mod.rs
```diff
@@ -237,14 +259,24 @@ impl Pending {
...
 		let err = self::error::ErrorKind::TimeoutOnNewPeers(self.requests.num_answered(), self.query_id_history.len());
 		if self.sender.send(Err(err.into())).is_err() {
 			debug!(target: "on_demand", "Dropped oneshot channel receiver on time out");
 		}
 	}
+
+	// The given request is determined as faulty and drop it accordingly
+	fn set_as_faulty_request(self, total_peers: usize, req_id: ReqId) {
+		let bad_peers = self.bad_responses.len();
+		warn!(target: "on_demand", "The request: {} was determined as faulty, {}/{} peer(s) gave bad response", req_id, bad_peers, total_peers);
```
ordian: Do we need a `warn!` here? A user can see a similar message in the response.

niklasad1:
@ordian 

Great, comments

1. Basically, timeout/retries are based on user configuration assumptions. This should improve throughput as long the peers and don't spam the network. We can get more accurate error message also but those are unforgently not propagated currently I will change that.
2. ...
3. Yes, that is awesome but I think it's even more beneficial for timeouts because of the ```exponential backoff``` i.e, to limit the number of attempts on a request by exponential decaying. Currently, we just wake up every 5 seconds and retry the request (I haven't double checked the actual interval but something like that)

ethcore/light/src/on_demand/mod.rs
ordian: Not sure I understand this. Why would it happen now that we have #9591 (unless someone is running node, older than 2.0.7-stable/2.1.2-beta)? And how can we distinguish this from invalid proof?

PR #9761, opened 2018-10-16
Version and notification for private contract wrapper added

### LINKED ISSUE #9719: [Private transactions] Public wrapper contract: emit an event on successful private state change with the initiator's address?
Hi all,

Is it possible to add an Event that gets emitted on a state change of the private contract?
Is it possible to know the address who initiated the state change (the caller of private_sendTransaction). What about including the initiator address in such an event?

Might be related to this [other issue](https://github.com/paritytech/parity-ethereum/issues/9641) I already opened.

grbIzl:
Special (Solidity) event is emitted, when changes in private contract state are approved and corresponding transaction sent to local miner.
The code of the contract wrapper is also updated, review can be found https://github.com/parity-contracts/private-tx/pull/7

Closes #9719


5chdn:
please review @dvdplm @andresilva

ethcore/private-tx/src/lib.rs
```diff
@@ -599,6 +625,21 @@ impl Provider where {
 		let value = self.client.call_contract(block, *address, data)?;
 		decoder.decode(&value).map_err(|e| ErrorKind::Call(format!("Contract call failed {:?}", e)).into())
 	}
+
+	fn get_contract_version(&self, block: BlockId, address: &Address) -> Result<usize, Error> {
+		let (data, decoder) = private_contract::functions::get_version::call();
+		let value = self.client.call_contract(block, *address, data)?;
```
andresilva: Should we also fallback to `INITIAL_PRIVATE_CONTRACT_VER` if this call fails? (this way this function would return `usize` instead of `Result`).

ethcore/private-tx/src/lib.rs
```diff
@@ -268,33 +272,33 @@ impl Provider where {
...
-						let signed_state = self.accounts.sign(validator_account, password, private_state_hash);
-						if let Err(e) = signed_state {
-							bail!("Cannot sign the state: {:?}", e);
-						}
-						let signed_state = signed_state.expect("Error was checked before");
-						let signed_private_transaction = SignedPrivateTransaction::new(private_hash, signed_state, None);
-						trace!(target: "privatetx", "Sending signature for private transaction: {:?}", signed_private_transaction);
-						self.broadcast_signed_private_transaction(signed_private_transaction.hash(), signed_private_transaction.rlp_bytes());
-					} else {
+					let contract = Self::contract_address_from_transaction(&transaction.transaction);
```
andresilva: Why not use `?` here?

ethcore/private-tx/src/lib.rs
```diff
@@ -198,36 +204,34 @@ impl Provider where {
...
-				trace!(target: "privatetx", "Hashed effective private state for sender: {:?}", private_state_hash);
-				self.transactions_for_signing.write().add_transaction(private.hash(), signed_transaction, contract_validators, private_state, contract_nonce)?;
-				self.broadcast_private_transaction(private.hash(), private.rlp_bytes());
-				Ok(Receipt {
-					hash: tx_hash,
-					contract_address: Some(contract),
-					status_code: 0,
-				})
-			}
+		let contract = Self::contract_address_from_transaction(&signed_transaction);
```
andresilva: Prefer `.map_err` and `?` here, instead of using `expect` below.

ethcore/private-tx/src/lib.rs
```diff
@@ -350,6 +354,21 @@ impl Provider where {
 					bail!(err);
 				}
 			}
+			//Notify about state changes
+			let contract = Self::contract_address_from_transaction(&desc.original_transaction);
+			if let Err(_) = contract {
+				bail!("Incorrect type of action for the transaction");
+			}
+			let contract = contract.expect("Error was checked before");
+			// TODO Usage of BlockId::Latest
```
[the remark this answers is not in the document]
grbIzl: See for example @tomusdrw comment here: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/private-tx/src/lib.rs#L209 We still use Latest in several modules, but I agree with him, that it's error-prone.
BTW I created a separate issue for addressing BlockId::Latest todo in private tx code (https://github.com/paritytech/parity-ethereum/issues/9825)

ethcore/private-tx/src/lib.rs
```diff
@@ -350,6 +347,15 @@ impl Provider where {
 					bail!(err);
 				}
 			}
+			//Notify about state changes
```
sorpaas: nit: missing space after `//`

ethcore/private-tx/src/lib.rs
sorpaas: My suggestion is that we may want to fail early when possible. If someone deploys a contract that has illegal interface, then stop early. It wouldn't crash the client anyway!

ethcore/private-tx/src/lib.rs
```diff
@@ -198,36 +204,30 @@ impl Provider where {
...
-					contract_address: Some(contract),
-					status_code: 0,
-				})
-			}
-		}
+		let contract = Self::contract_address_from_transaction(&signed_transaction).map_err(|_| ErrorKind::BadTransactonType)?;
+		let data = signed_transaction.rlp_bytes();
+		let encrypted_transaction = self.encrypt(&contract, &Self::iv_from_transaction(&signed_transaction), &data)?;
+		let private = PrivateTransaction::new(encrypted_transaction, contract);
+		// TODO [ToDr] Using BlockId::Latest is bad here,
```
sorpaas:
```suggestion
		// TODO #9825 [ToDr] Using BlockId::Latest is bad here,
```

Please add issue number to help navigate!

ethcore/private-tx/src/lib.rs
```diff
@@ -268,33 +268,30 @@ impl Provider where {
...
-						}
-						let signed_state = signed_state.expect("Error was checked before");
-						let signed_private_transaction = SignedPrivateTransaction::new(private_hash, signed_state, None);
-						trace!(target: "privatetx", "Sending signature for private transaction: {:?}", signed_private_transaction);
-						self.broadcast_signed_private_transaction(signed_private_transaction.hash(), signed_private_transaction.rlp_bytes());
-					} else {
-						bail!("Incorrect type of action for the transaction");
+					let contract = Self::contract_address_from_transaction(&transaction.transaction)
+						.map_err(|_| "Incorrect type of action for the transaction")?;
+					// TODO [ToDr] Usage of BlockId::Latest
```
sorpaas:
```suggestion
					// TODO #9825 [ToDr] Usage of BlockId::Latest
```

PR #9762, opened 2018-10-16
ethash: implement Progpow

andresilva:
This is a proof-of-concept implementation of progpow, this is still work in progress and I don't intend this to be merged yet.
- [x] Progpow light verification
- [x] Pass existing test vectors (https://github.com/chfast/ethash/blob/master/test/unittests/progpow_test_vectors.hpp)
- [x] Benchmark and optimize
- [x] Integrate with ethash engine
- [x] Update to latest spec

References:
- EIP 1057 (https://eips.ethereum.org/EIPS/eip-1057)
- reference implementation (https://github.com/ifdefelse/ProgPOW)
- go-ethereum implementation (https://github.com/ethereum/go-ethereum/pull/17731)
- aleth's ethash implementation (https://github.com/chfast/ethash)


andresilva:
Benchmarks of baseline performance:

```
test benchmarks::bench_hashimoto_light       ... bench:     981,214 ns/iter (+/- 32,031)
test benchmarks::bench_progpow_light         ... bench:  16,766,912 ns/iter (+/- 1,429,102)
test benchmarks::bench_progpow_optimal_light ... bench:   5,055,690 ns/iter (+/- 2,379,231)
```
`bench_progpow_optimal_light` - doesn't take into account cdag creation, this is realistic since the cdag can be cached during it's epoch (50 blocks).

Will have a look at what can be optimized now that it's passing all test vectors.

holiman:
> doesn't take into account cdag creation, this is realistic since the cdag can be cached during it's epoch (50 blocks). 

Isn't the `cDag` epoch the same as dag epoch, `30k` blocks?

andresilva:
@holiman Yes, you're right, the only thing that changes every 50 blocks is the program generated from the `cDag`, but the `cDag` itself has the same epoch as the ethash dag.

andresilva:
I updated the benchmarks to use criterion and made some optimizations. Current results:

```
hashimoto_light         time:   [900.11 us 906.55 us 914.25 us]
progpow_light           time:   [9.2520 ms 9.2781 ms 9.3053 ms]
progpow_optimal_light   time:   [2.4438 ms 2.4742 ms 2.5080 ms]
```

If someone wants to have a look at the generated assembly install `cargo-asm` and run `cargo asm ethash::progpow::progpow --rust --build-type release`.

ethash/benches/ethash.rs
```diff
@@ -0,0 +1,87 @@
...
+use tempdir::TempDir;
+use rustc_hex::FromHex;
+use ethash::{NodeCacheBuilder, OptimizeFor};
+use ethash::compute::light_compute;
+
+fn bench_hashimoto_light(c: &mut Criterion) {
+	let builder = NodeCacheBuilder::new(OptimizeFor::Memory);
+	let tempdir = TempDir::new("").unwrap();
+	let light = builder.light(&tempdir.path(), 1);
+	let h = FromHex::from_hex("c9149cc0386e689d789a1c2f3d5d169a61a6218ed30e74414dc736e442ef3d1f").unwrap();
```
dvdplm: Consider using `hex!` from `hex_literal`.

ethash/src/compute.rs
```diff
@@ -51,32 +53,55 @@ impl Light {
 		builder: &NodeCacheBuilder,
 		cache_dir: &Path,
 		block_number: u64,
+		progpow_transtion: u64,
```
ordian: typo here and below `s/transtion/transition/g`

5chdn:
Does not compile :)

andresilva:
I just launched a local testnet with parity and geth, the geth node is doing CPU mining and parity is happily importing the progpow blocks :relaxed:. There have been some changes to the progpow algorithm (https://github.com/ifdefelse/ProgPOW/commit/a3f62349a1513f0393524683f9671cfe17cca895), I will update my implementation once there are new test vectors available.

hackmod:
I guess you already know,
go-ethereum ProgPoW  PR updated! https://github.com/ethereum/go-ethereum/pull/17731
ProgPoW PR for ethash by chfast updated https://github.com/chfast/ethash/pull/108

naikmyeong:
@andresilva May I ask is it possible to update this PR for new spec of ProgPOW?

Here is the updated spec from other implementations

+ Reference implementation update: https://github.com/ifdefelse/ProgPOW/compare/620b4c7aafe60391f863372814d7517e94386379...master
+ go-ethereum ProgPoW PR updated! ethereum/go-ethereum#17731
+ ProgPoW PR for ethash by chfast updated chfast/ethash#108

naikmyeong:
https://github.com/ethereum/EIPs/pull/1588

naikmyeong:
Spec v0.9.2 released, may I expect this branch to be updated by this week? @andresilva 😄

naikmyeong:
@andresilva Whole rebase with master branch would be appreciated 😄

5chdn:
What's the plan here, waiting for the spec to finalize? Or shall we merge early and iterate from there? @andresilva

ifdefelse:
We consider the v0.9.2 to be final.  It's been running on the test network for a while now and no issues have been raised.

naikmyeong:
@5chdn @ifdefelse @andresilva I believe this pr can be merged now, other coins like ellaism or classic is waiting for this PR to be merged

ethash/src/lib.rs
```diff
@@ -25,14 +25,22 @@ extern crate crunchy;
...
+extern crate rustc_hex;
+
+#[cfg(test)]
+extern crate serde_json;
+
 #[cfg(test)]
 extern crate tempdir;
 
-mod compute;
+// FIXME [andre]: should all be private
```
veox: FIXME still present.

naikmyeong:
Rebased version if anyone needed https://github.com/gangnamtestnet/parity-ethereum/tree/new3

ethash/src/compute.rs
```diff
@@ -43,6 +44,7 @@ pub struct ProofOfWork {
 pub struct Light {
 	block_number: u64,
 	cache: NodeCache,
+	c_dag: Option<CDag>,
```
cheme: This could be an enum like 'EthashVariant', if there is some progpow2 or other. Yet it means that it needs to be build initiated from a fn parameter in new_with_builder.
I don't know if it is worth the added complexity (enum makes it explicit that c_dag condition usage of progpow but it may be better to keep it simple).

ethash/src/progpow.rs
```diff
@@ -0,0 +1,636 @@
...
+	}
+
+	for i in (1..mix_seq_dst.len()).rev() {
+		unsafe {
+			// NOTE: `i` takes values from the range [1..PROGPOW_REGS] and `j`
+			// takes values from the the range [0..i]. This way it is guaranteed
+			// that the indices are always within the range of `mix_seq_dst` and
+			// `mix_seq_cache` and we can skip the bounds checking.
+			let j = rnd.next_u32() as usize % (i + 1);
+			::std::ptr::swap(&mut mix_seq_dst[i], mix_seq_dst.get_unchecked_mut(j));
```
cheme: Does swap improve performance over just u32 assigns (I hoped the compiler would optimize that)?

ethash/src/progpow.rs
```diff
@@ -0,0 +1,636 @@
...
+	for l in 0..PROGPOW_DAG_LOADS {
+		let index = g_offset * PROGPOW_LANES * PROGPOW_DAG_LOADS + l * 16;
+		let node = calculate_dag_item(index as u32 / 16, cache);
+		dag_item[l * 16..(l + 1) * 16].clone_from_slice(node.as_words());
+	}
+
+	// Lanes can execute in parallel and will be convergent
+	for l in 0..mix.len() {
+		// Initialize the seed and mix destination sequence
+		let (mut rnd, mix_seq_dst, mix_seq_cache) = progpow_init(seed);
```
cheme: Here I think it would be better if this line was out of the loop and we just clone `rnd` on every iteration (kiss need derive clone).

ethash/src/progpow.rs
```diff
@@ -0,0 +1,636 @@
...
+	}
+
+	for r in 0..22 {
+		keccak_f800_round(&mut st, r);
+	}
+
+	(st[0].swap_bytes() as u64) << 32 | st[1].swap_bytes() as u64
+}
+
+pub fn keccak_f800_long(header_hash: H256, nonce: u64, result: [u32; 8]) -> H256 {
```
niklasad1: @andresilva AFAIK, `H256` implements from/into `[u8; 32]` why not use in instead of `[u32; 8]`?

Then you can get rid of the `transmute` I suppose :P

5chdn:
@andresilva do you have time to finalize this?

andresilva:
Yes, I'll try to finish this up in the weekend. :+1:

ethash/src/progpow.rs
andresilva: In this code `H256` is a type alias to `[u8; 32]` (defined in `ethash/keccak.rs`). The reason I'm doing the transmute is to avoid all the bit shuffling.

ethash/src/progpow.rs
andresilva: `unchecked_mut` was only used on `j`, since the compiler can't prove that it will always be in-bounds, so it must add a bounds check. You can check this by looking at the generated assembly, at least it generated different code when I wrote this. Anyway, even with the bounds check it probably doesn't make a difference performance-wise because of branch prediction, I couldn't reliably microbenchmark this on my laptop so I'm removing all unsafe code.

ethash/src/progpow.rs
andresilva: Good catch!

ethash/src/progpow.rs
niklasad1: Right, I verified it with this [godbolt example](https://rust.godbolt.org/z/JE5GZU) so you are right cool!

andresilva:
I think I've addressed all grumbles. The only unsafe code that is kept is for transmuting from `[u32; 8]` to `[u8; 32]`, I think doing the byte shuffling is unnecessary but can remove it if you think otherwise. There is probably a lot of low hanging fruit for optimization here. Completely unrolling the hash function `kecak_f800` like we do for `keccak_f1600` would probably speed it up a bit.

andresilva:
There's a test with a bunch of test vectors (shared among implementations) that might take a little while to run. We might want to disable this test on the CI.

ethash/src/compute.rs
```diff
@@ -40,9 +41,15 @@ pub struct ProofOfWork {
 	pub mix_hash: H256,
 }
 
+enum Algorithm {
+	Hashimoto,
+	Progpow(CDag),
```
niklasad1: Does it make sense to `Box CDag` in the Algorithm enum? 
It is quite big (16384 bytes), thus it will make the enum `Algorithm` as big.

andresilva: Good idea. No I didn't see any regression in the benchmarks with all the changes I've done so far.

andresilva:
Boxed the cdag inside algorithm and adapted an existing cargo feature to skip the slow progpow test vectors in CI.

andresilva:
Actually since CI is running the tests with `--release` the test vectors don't take that long to run (20s on my laptop vs like 4 minutes for a debug build), so I'm going to keep them.

ethash/src/progpow.rs
```diff
@@ -0,0 +1,595 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
niklasad1:
```suggestion
// Copyright 2015-2019 Parity Technologies (UK) Ltd.
```

ethash/src/progpow.rs
niklasad1: ok, since we don't support `big endian` it should be fine

andresilva:
I removed the verification that triggered the compile error, it wasn't working in the first place the reason it wasn't being triggered is because the feature flags were using different names. As far as I understand since `json_tests` is a part of `ethcore` it will be compiled in non-test builds (e.g. `evmbin` directly uses `json_tests` functionality).

5chdn:
:tada:

PR #9763, opened 2018-10-16
prevent zero networkID 

seunlanlege:


ordian:
looks like some tests need to be fixed

5chdn:
Hold on? This pull request should not have been merged.

1. It does not explain **why** network ID can't be zero
2. The network ID of Olympic actually _is_ zero, changing it basically creates a new network.

ethcore/res/ethereum/olympic.json
```diff
@@ -17,7 +17,7 @@
 		"accountStartNonce": "0x00",
 		"maximumExtraDataSize": "0x0400",
 		"minGasLimit": "125000",
-		"networkID" : "0x0",
+		"networkID" : "0xf0",
```
5chdn: Should be reverted

sorpaas: Yeah sorry I missed this. This line should indeed be reverted and fixed (or we may just consider removing olympics support).

5chdn:
Ok, so we might want to remove Olympic completely

seunlanlege:
alright will create an issue

Issue #9102, opened 2018-07-12
RPC threads dying w/ "Too many open files"


XertroV:
> I'm running:
>
> - **Which Parity version?**: 1.10.9 (observed on 1.10.8 too, though haven't dug through logs)
> - **Which operating system?**: Linux
> - **How installed?**: binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: yes (workaround)

pretty self explanatory - all RCP threads die simultaneously, but everything else keeps working (including websockets according to netstat; RPC ports disappear but websocket ports don't)

(Most of those requests are 2x eth-stats running via pm2)

Logs: 

```
Jul 11 12:16:35 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:16:35 UTC   39/50 peers    978 MiB chain   92 MiB db  0 bytes queue  270 KiB sync  RPC:  0 conn, 15 req/s,  42 µs
Jul 11 12:17:14 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:17:14 UTC Imported #5944667 2191…9c17 (201 txs, 7.99 Mgas, 577.74 ms, 28.41 KiB)
Jul 11 12:17:23 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:17:23 UTC Imported #5944668 8712…a45d (225 txs, 8.00 Mgas, 401.54 ms, 32.90 KiB)
Jul 11 12:17:23 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:17:23 UTC   40/50 peers    979 MiB chain   93 MiB db  0 bytes queue  270 KiB sync  RPC:  0 conn, 15 req/s, 200849 µs
Jul 11 12:17:34 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:17:34 UTC Imported #5944652 e9ce…5142 (158 txs, 7.98 Mgas, 530.73 ms, 27.99 KiB)
Jul 11 12:17:43 eth-aws-nv-node-03 parity[11186]: 2018-07-11 12:17:43 UTC Imported #5944669 3f6e…aa3e (123 txs, 7.98 Mgas, 319.24 ms, 23.49 KiB)
[...]

PR #9780, opened 2018-10-19
deps: update jsonrpc-core

ordian:
includes recent changes by @seunlanlege, cc #9102

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1412,7 +1412,7 @@ dependencies = [
 [[package]]
 name = "jsonrpc-core"
 version = "8.0.1"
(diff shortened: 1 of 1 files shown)


ordian:
I don't think we should merge it into master, since it conflicts with #9657. Just wanted to see if it works on Windows before we backport this.

ordian:
@seunlanlege yes, I know that. In order to incorporate recent changes in jsonrpc-core, we can either  update jsonrpc-core#parity-0.11 in stable and beta, or backport #9657. The latter is huge, so I'm in favor of the first approach.

PR #9788, opened 2018-10-20
ci: remove failing tests for android, windows, and macos

5chdn:


niklasad1:
FYI, it looks like the `android` pipeline is fixed now i.e, it compiles. However, the issue is now running the tests but should be enough to compile-test it IMO and seems to be some mismatch between which linker to use since we are cross-compiling!

5chdn:
Where is Android fixed?

5chdn:
Ok let me rebase this.

niklasad1:
that's weird must be something with configuration because locally on my machine:

```bash
$ cargo build --target armv7-linux-androideabi
$ file target/armv7-linux-androideabi/debug/parity
target/armv7-linux-androideabi/debug/parity: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /system/bin/linker, not stripped
```

5chdn:
It compiles fine, you are right, it's a CI issue, probably not correctly cleaning up the builds between different pipes. I'll escalate to devops.

5chdn:
@andresilva we will get a new pipeline soon anyways. I just want to get rid of the tests on these machines as this clogs the pipelines too much.

PR #9790, opened 2018-10-21
light client : failsafe crate (circuit breaker)

### LINKED ISSUE #9536: Refactor on-demand queries to use failsafe-rs
See https://github.com/paritytech/parity-ethereum/pull/9318#issuecomment-419519770. 
Other things worth doing:
- [ ] Use wrapping adds where possible
- [ ] The logic in [`dispatch_pending`](https://github.com/paritytech/parity-ethereum/blob/383107e6b62e12c9088d303ae85927568bbea653/ethcore/light/src/on_demand/mod.rs#L437) is quite complex. It would be nice to extract some functions to reduce cyclomatic complexity.

niklasad1:
Attempt to close #9536

### Major changes in this PR:
1.  Instead of limit the request by the number query retries and empty responses this introduces a ```exponentially moving average (provided by failsafe-rs)``` during a fixed time period when a request evaluated as faulty or not (but only failures are recorded) and a simple max_timeout_time for the maximum time that responses are evaluated
2. Adds functionality to determine the failure reason for bad responses which are propagated via the RPC interface
3. New CLI args:
   - response_time_window
   - request_time_window
   - request_backoff_start
   - request_backoff_max
   - request_backoff_rounds_max

### Benefits
1. A user can determine for over how long time period a request should be evaluated as a failure without specifying the number of responses or queries
2. The actual failure response is propagated to the user via the RPC interface
3. The request backs off `exponentially` upon failures and retries less frequently to avoid spamming the network

### Circuit breaker

image

- closed - calls are accepted
- open - failed over some threshold and calls are rejected
- half-open - calls are accepted after backoff from failure 

### Example
```bash
curl --data '{"method":"eth_getTransactionReceipt","params":["0x444172bef57ad978655171a8af2cfd89baa02a97fcb773067aef7794d6913fff"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8555

{"jsonrpc":"2.0","error":{"code":-32042,"message":"Bad response on request: [ TransactionIndex ]. Error cause was EmptyResponse, (majority count: 91 / total: 91)"},"id":1}
```


ethcore/light/src/on_demand/request_guard.rs
```diff
@@ -0,0 +1,89 @@
...
+	Rejected,
+	/// The request reached the maximum of backoff iterations
+	ReachedLimit,
+}
+
+/// Handle and register requests that can fail
+#[derive(Debug)]
+pub struct RequestGuard {
+	num_failures: usize,
+	max_failures: usize,
```
ordian: Can it be implemented as `success_rate_over_time_window(...).or_else(consecutive_failures(...))`?

ethcore/light/src/on_demand/tests.rs
```diff
@@ -495,3 +506,98 @@ fn fill_from_cache() {
...
+
+	thread::sleep(Duration::from_secs(1));
+
+	for (i, &backoff) in binary_exp_backoff.iter().enumerate() {
+		harness.service.dispatch_pending(&Context::FaultyRequest);
+		let now = SystemTime::now();
+		while now.elapsed().unwrap() <= harness.service.time_window_dur {}
+		let now = SystemTime::now();
+		while now.elapsed().unwrap() < Duration::from_secs(backoff + 2) {}
+		if i < 9 {
```
ordian: `binary_exp_backoff.len() - 1`?

ethcore/light/src/types/request/mod.rs
```diff
@@ -90,6 +90,8 @@ pub enum ResponseError<T> {
 	Validity(T),
 	/// No responses expected.
 	Unexpected,
+	/// Empty response
+	EmptyResponse,
```
ordian: How is it different from `ResponseError::Validity(ValidityError::Empty)`?

parity/cli/mod.rs
```diff
@@ -563,14 +563,26 @@ usage! {
...
+			"Specify light client start backoff time",
+			
+			ARG arg_on_demand_end_backoff: (Option<u64>) = None, or |c: &Config| c.light.as_ref()?.on_demand_end_backoff,
+			"--on-demand-end-backoff=[MS]",
+			"Specify light client end backoff time",
+			
+			ARG arg_on_demand_max_backoff_rounds: (Option<usize>) = None, or |c: &Config| c.light.as_ref()?.on_demand_max_backoff_rounds,
+			"--on-demand-max-backoff-rounds=[TIMES]",
+			"Specify light client maximum number of backoff iterations",
+
```
ordian: Not sure a user wants to have control over all those knobs, maybe it's ok to just have good (empirical research?) defaults, but I don't have a strong opinion on that.

ethcore/light/src/types/request/mod.rs
niklasad1: Excellent question, this is a bit blurry

- ```ResponseError:EmptyResponse``` - A response that is found `empty` before ```on_supply_responses``` https://github.com/paritytech/parity-ethereum/blob/master/ethcore/light/src/on_demand/mod.rs#L610-#L618
- ```ResponseError::Validity(ValidityError::Empty)``` - A response that validated by ```check_response``` https://github.com/paritytech/parity-ethereum/blob/master/ethcore/light/src/on_demand/request.rs#L620

Shall I change only to use ```ResponseError::Validity(ValidityError::Empty)```?

ethcore/light/src/on_demand/request_guard.rs
niklasad1: AFAIK, it can not because `failsafe-rs` will never drop a request it will just `backoff` and continue forever and we can't access the internal state. Thus, we can't distinguish between the `two failures`, i.e, if the request is in `Rejected state` just backed off or if it should be dropped. But, I will double check this!

However, this could also be implemented by a specific failure policy but this was pretty easy to implement so I didn't bother

ordian: ok, makes sense

ethcore/light/src/on_demand/mod.rs
```diff
@@ -403,10 +435,14 @@ impl OnDemand {
 			required_capabilities: capabilities,
 			responses,
 			sender,
-			base_query_index: 0,
-			remaining_query_count: 0,
-			query_id_history: BTreeSet::new(),
-			inactive_time_limit: None,
+			request_guard: RequestGuard::new(
```
cheme: I really have question about circuit breakers, for instance it seems to require managing a kind of global state (otherwise why would there be a need for mutex), but here aren't we spawning a new circuit breaker state for each request?

I was expecting that the request guard would be a clone of an existing state (like a state stored for on_demand per user or something more global).
I do not really see how we would get success rate aggregate this way. From what I remember, I did store error query history and remaining query count per request but because I was considering errors (there can be multiple error for a request due to empty reply (maybe one of the latest pr does fix that by slashing those kind of peers) but I think the request lifetime ends at the first request success.

I am not sure at which level the circuit breaker is meant to be applied : 
- for a peer and any of its request?
- for on_demand_service globally (probably not)?
- for a peer and a particular kind of query? (the fact that we can have multiple query per request does not make it easy)
- for a kind of request globaly?
- for every single request and its subsequent replies (that is what I feel is currently done, but I may be wrong)

We should probably document that.

It could make sense to use the other 'failure_policy' with backoff (counting error) at a per request level to basically do the same as before (but if we do it we would need to be able to get rid of the circuit safe mutex that becomes useless).

ethcore/light/src/on_demand/mod.rs
```diff
@@ -435,82 +471,65 @@ impl OnDemand {
...
-					.skip(offset).take(num_peers) {
-					// TODO: see which requests can be answered by the cache?
-					if pending.remaining_query_count == 0 {
-						break
+				// The first peer to dispatch the request is chosen at random
+				let rand = rand::random::<usize>() % cmp::max(num_peers, 1);
+
+				for (peer_id, peer) in peers
+					.iter()
+					.cycle()
```
cheme: Is the switch to 'cycle' needed here, I mean it feels to me like previous 'chain' call does ensure we did not infiniloop (I am not saying we can (I did not check) but just that it makes it easier to be sure of the fact). I do not think there is a need to loop more than once other all peers to find next dispatch.

ethcore/light/src/on_demand/response_guard.rs
```diff
@@ -0,0 +1,167 @@
...
+	WrongHeaderSequence,
+	/// Wrong response kind
+	WrongKind,
+	/// Wrong number
+	WrongNumber,
+	/// Wrong Trie Root
+	WrongTrieRoot,
+}
+
+/// Handle and register responses that can fail
```
cheme: Maybe comments on the why we need a response handler would be interesting.

Ok I start to get an intuition of how it works. Basically we add this as a way to aggregate the values (response) for the circuit breaker.
Which would mean we want a global circuit breaker (which make sense considering the success strategy instead of error), and this is to handle results (so no backoff strategy needed).
My concern here is that the state machine use is not related with any other state machine (I may have not read the circuit breaker crate correctly), shouldn't the constructor should use a clone of an existing circuit breaker state?

parity/cli/mod.rs
cheme: I think it is good to expose the primitives to give full control, but some alternate preset (basics configuration) could also be a nice thing.

 (at this point it probably lacks documentation of the impact of every option, but maybe the wiki or another source is better for that : it seems difficult to explain quickly).

Cargo.lock
```diff
@@ -3135,6 +3146,11 @@ dependencies = [
  "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
```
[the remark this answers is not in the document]
niklasad1: Submitted PR to fix [this](https://github.com/dmexe/failsafe-rs/pull/8)

I would not prefer not to fork that repo just for the dependency....

ethcore/light/src/on_demand/mod.rs
ordian: `.take(num_peers)` ensures we don't have infiniloop
`.iter().cycle().skip(rand).take(num_peers)` essentially rotates peers by `rand`

ethcore/light/src/on_demand/mod.rs
```diff
@@ -227,20 +234,17 @@ impl Pending {
...
-	// returning no reponse, it will result in an error.
-	// self is consumed on purpose.
-	fn no_response(self) {
-		trace!(target: "on_demand", "Dropping a pending query (no reply) at query #{}", self.query_id_history.len());
-		let err = self::error::ErrorKind::MaxAttemptReach(self.requests.num_answered());
+	// received too many empty responses, may be away to indicate a faulty request
+	fn bad_response(self, err: ResponseGuardError) {
+		let err = self::error::ErrorKind::BadResponse(err);
 		if self.sender.send(Err(err.into())).is_err() {
 			debug!(target: "on_demand", "Dropped oneshot channel receiver on no response");
```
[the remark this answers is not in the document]
niklasad1: I could implement a logger for the circuit break that register every event:

```rust
pub trait Instrument {
    /// Calls when state machine reject a call.
    fn on_call_rejected(&self);

    /// Calls when the circuit breaker become to open state.
    fn on_open(&self);

    /// Calls when the circuit breaker become to half open state.
    fn on_half_open(&self);

    /// Calls when the circuit breaker become to closed state.
    fn on_closed(&self);
}
```

Sufficient?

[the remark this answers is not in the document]
niklasad1: Otherwise, the error is actually propgated via RPC response but exactly how many bad responses and the particular reason is not propogated

ethcore/light/src/on_demand/mod.rs
niklasad1: The circuit breakers are on per request basis, and yes it will introduce two mutexes per request and I wanted to separate responses and attempts per request to have a simple logic.

I think `failsafe crate` want to be thread-safe and ensure that only one event can modify the state at the time.

ethcore/light/src/on_demand/response_guard.rs
niklasad1: I have partly already answered this concern in another similar question.

This creates a separate `clean` state machine and should not be inferred by anything?
The entire object is moved once either `RequestGuard` `ResponseGuard` have reached `our failure state` or upon success!

Thoughts?

ethcore/light/src/on_demand/mod.rs
cheme: The thing I do not understand with a per request basis is that the only relevant event that we will be registering into the circuit breaker will be request failure, so won't the failure policy "SuccessRateOverTimeWindows" act like a simple timeout (average success being always 0)? That is why the use of the other policy "ConsecutiveFailure" seems more intuitive to me.

On another hand, if we choose to use back our circuit state like single circuit state for each peers, we would register some successes and I would understand (but I do not expect it to be easy, especially in this future based code). Plus we need to be careful to not mess to much with the random selection of peers to avoid overloading some nodes, but that is another story.

For instance adding this test in response_guard illustrate that (the test is passing) : 

```rust
	#[test]
	fn act_like_timeout() {
		let mut guard = ResponseGuard::new(0.8, Duration::from_secs(5));
		thread::sleep(Duration::from_secs(5));
		assert!(guard.register_error(&ResponseError::Validity(ValidityError::WrongKind)).is_err());

    let mut guard = ResponseGuard::new(0.2, Duration::from_secs(5));
		assert!(guard.register_error(&ResponseError::Unexpected).is_ok());
		thread::sleep(Duration::from_secs(5));
		assert!(guard.register_error(&ResponseError::Validity(ValidityError::WrongKind)).is_err());

    let mut guard = ResponseGuard::new(0.8, Duration::from_secs(5));
    
    for i in 0..1000 {
		  assert!(guard.register_error(&ResponseError::Unexpected).is_ok());
    }
		thread::sleep(Duration::from_secs(5));
		assert!(guard.register_error(&ResponseError::Validity(ValidityError::WrongKind)).is_err());
	}

```

ethcore/light/src/on_demand/response_guard.rs
cheme: You're right, both comments are really similar, so I reply in previous comment.

ethcore/light/src/on_demand/mod.rs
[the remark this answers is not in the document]
cheme: Having an 'Instrument' implementation writing in our logger looks like a good idea.

I also believe that being able to display the inner state of the circuit breaker (especially FailurePolicy debug info), is something that would greatly benefit the crate.

5chdn:
Please rebase

ethcore/light/src/on_demand/mod.rs
niklasad1: > The thing I do not understand with a per request basis is that the only relevant event that we will be registering into the circuit breaker will be request failure, so won't the failure policy "SuccessRateOverTimeWindows" act like a simple timeout (average success being always 0)? That is why the use of the other policy "ConsecutiveFailure" seems more intuitive to me.

Yes, that is correct but I argue that is more intuitive for a user to specify a certain time window then the number of `bad-responses` because the number of peers is dynamic. However, that the first assert is passing was not expected from my side. I thought that the time window would be `reset` here and only be registered after the time window had passed. I need to re-think this maybe total number of failures or consecutive failures are better anyhow.

> single circuit state for each peer

No, that is too fine-grained IMHO

ethcore/light/src/on_demand/mod.rs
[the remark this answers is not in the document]
niklasad1: Well, I found it useless because it doesn't get the internal state but I added better logs at least. 
Should be fine now, can you confirm? 

Example Request:
```bash
2018-10-31 12:12:38  jsonrpc-eventloop-0 TRACE on_demand  TransactionIndex Request, Incomplete { hash: Scalar(0xa17b45b011dad358740ccb65383e0900b086643d15192060f3983160492ccaff) }                                                           
2018-10-31 12:12:38  jsonrpc-eventloop-0 TRACE circuit_breaker  RequestGuard; failures: 1/10, state StateMachine { state: "closed" }                                                                                                          
2018-10-31 12:12:38  jsonrpc-eventloop-0 TRACE on_demand  Was unable to dispatch 1 requests.
2018-10-31 12:12:39  IO Worker #3 TRACE circuit_breaker  RequestGuard; failures: 2/10, state StateMachine { state: "closed" }                                                                                                                 
2018-10-31 12:12:39  IO Worker #3 TRACE on_demand  Was unable to dispatch 1 requests.
2018-10-31 12:12:44  IO Worker #0 TRACE circuit_breaker  RequestGuard; failures: 3/10, state StateMachine { state: "closed" }
2018-10-31 12:12:44  IO Worker #0 TRACE on_demand  Was unable to dispatch 1 requests.
2018-10-31 12:12:49  IO Worker #0 TRACE circuit_breaker  RequestGuard; failures: 4/10, state StateMachine { state: "open" }
2018-10-31 12:12:49  IO Worker #0 TRACE on_demand  Was unable to dispatch 1 requests.
2018-10-31 12:12:54  IO Worker #3 TRACE circuit_breaker  RequestGuard; failures: 4/10, state StateMachine { state: "open" }
2018-10-31 12:12:54  IO Worker #3 TRACE on_demand  Was unable to dispatch 1 requests.
2018-10-31 12:12:59  IO Worker #0 TRACE circuit_breaker  RequestGuard; failures: 5/10, state StateMachine { state: "open" }
....
2018-10-31 12:13:49  IO Worker #1 TRACE circuit_breaker  RequestGuard: reached_limit, failures: 10/10, state StateMachine { state: "open" }                                                                                                   
2018-10-31 12:13:49  IO Worker #1 TRACE on_demand  Was unable to dispatch 0 requests.
2018-10-31 12:13:49  jsonrpc-eventloop-0 DEBUG rpc  [Some(Num(1))] Took 70439ms
2018-10-31 12:13:49  jsonrpc-eventloop-0 DEBUG rpc  Response: {"jsonrpc":"2.0","error":{"code":-32065,"message":"Request time out"},"id":1}.   
```

Example Response:
```bash
2018-10-31 12:15:28  jsonrpc-eventloop-0 TRACE rpc  Request: {"method":"eth_getTransactionByHash","params":["0xa17b45b011dad358740ccb65383e0900b086643d15192060f3983160492ccaff"],"id":1,"jsonrpc":"2.0"}.
2018-10-31 12:15:28  jsonrpc-eventloop-0 TRACE on_demand  TransactionIndex Request, Incomplete { hash: Scalar(0xa17b45b011dad358740ccb65383e0900b086643d15192060f3983160492ccaff) }
2018-10-31 12:15:28  jsonrpc-eventloop-0 TRACE on_demand  Was unable to dispatch 0 requests.
2018-10-31 12:15:28  IO Worker #3 TRACE circuit_breaker  ResponseGuard: StateMachine { state: "closed" }, responses {EmptyResponse: 1}
2018-10-31 12:15:28  IO Worker #3 TRACE on_demand  Was unable to dispatch 0 requests.
2018-10-31 12:15:29  IO Worker #0 TRACE circuit_breaker  ResponseGuard: StateMachine { state: "closed" }, responses {EmptyResponse: 2}
2018-10-31 12:15:29  IO Worker #0 TRACE on_demand  Was unable to dispatch 0 requests.
2018-10-31 12:15:29  IO Worker #0 TRACE circuit_breaker  ResponseGuard: StateMachine { state: "closed" }, responses {EmptyResponse: 3}
...
2018-10-31 12:15:38  IO Worker #0 TRACE circuit_breaker  ResponseGuard: StateMachine { state: "open" }, responses: {EmptyResponse: 71}
2018-10-31 12:15:38  jsonrpc-eventloop-1 DEBUG rpc  [Some(Num(1))] Took 10310ms
2018-10-31 12:15:38  jsonrpc-eventloop-1 DEBUG rpc  Response: {"jsonrpc":"2.0","error":{"code":-32042,"message":"Bad response timeout, error cause was EmptyResponse, (majority count: 71 / total: 71)"},"id":1}.
```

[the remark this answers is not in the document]
cheme: Nice to have some actual information about what happen :+1:

niklasad1:
@ordian @cheme 

I refactored this and the major thing to sort out is whether `the timeout feature` on the responses is the way to go? IMHO, I think this is the more user-friendly than limit the number of `bad_responses`, basically, a user just configures "I want to wait at most x seconds for this request"

However, the response guard doesn't need the use failsafe mechanism that you pointed our earlier which I can remove if I should continue with this.

Dump of the `new RPC error response`:
```bash
curl --data '{"method":"eth_getTransactionReceipt","params":["0x444172bef57ad978655171a8af2cfd89baa02a97fcb773067aef7794d6913fff"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8555

{"jsonrpc":"2.0","error":{"code":-32042,"message":"Bad response on request: [ TransactionIndex ]. Error cause was EmptyResponse, (majority count: 91 / total: 91)"},"id":1}
```

5chdn:
Could you give this a final review @ordian @cheme?

ethcore/light/src/on_demand/request_guard.rs
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	/// Update the state after a `faulty` call
+	pub fn register_error(&mut self) -> Error {
+		// Circuit breaker is `closed`, count as a failure
+		if self.num_failures >= self.max_failures {
+				trace!(target: "circuit_breaker", "RequestGuard: reached_limit, failures: {}/{}, state {:?}",
+					self.num_failures, self.max_failures, self.state);
+				Error::ReachedLimit
+		}
```
sorpaas: nit: don't know whether we have style guides for this, but usually we put `}` and `else {` in the same line?

```
if {
  do(something);
} else if {
  do(something_else);
} else {
  do(other_things);
}
```

ethcore/light/src/on_demand/response_guard.rs
```diff
@@ -0,0 +1,184 @@
...
+		let err = self.into_reason(err);
+		*self.responses.entry(err).or_insert(0) += 1;
+		if self.state.is_call_permitted() {
+			trace!(target: "circuit_breaker", "ResponseGuard: {:?}, responses {:?}", self.state, self.responses);
+			Ok(())
+		} else {
+			trace!(target: "circuit_breaker", "ResponseGuard: {:?}, responses: {:?}", self.state, self.responses);
+			let (&err, &max_count) = self.responses.iter().max_by_key(|(_k, v)| *v).expect("got at least one element; qed");
+			let majority = self.responses.values().filter(|v| **v == max_count).count() == 1;
+			// FIXME: more efficient with a separate counter
```
sorpaas: Would be nice if you can create an new issue and include an issue number there!

ethcore/light/src/on_demand/request_guard.rs
niklasad1: Yeah, I agree. I think I had a comment there before that's why the formatting was off!

niklasad1: @sorpass Yeah, that is true but how do you suggest to deal with the comment which is the reason I've put the else on a new line?

ethcore/light/src/on_demand/response_guard.rs
niklasad1: I will fix this if/when @cheme signs off of this :P

ethcore/light/src/on_demand/mod.rs
```diff
@@ -435,82 +501,56 @@ impl OnDemand {
...
-					pending.base_query_index + history_len
-				} % cmp::max(num_peers, 1);
-				let init_remaining_query_count = pending.remaining_query_count; // to fail in case of big reduction of nb of peers
-				for (peer_id, peer) in peers.iter().chain(peers.iter())
-					.skip(offset).take(num_peers) {
-					// TODO: see which requests can be answered by the cache?
-					if pending.remaining_query_count == 0 {
-						break
+				// The first peer to dispatch the request is chosen at random
+				let rand = rand::random::<usize>() % cmp::max(num_peers, 1);
```
ordian: Can we use here `gen_range` instead?
```suggestion
				let rand = rand::thread_rng().gen_range(0, cmp::max(num_peers, 1));
```

ethcore/light/src/on_demand/mod.rs
```diff
@@ -326,30 +351,68 @@ pub struct OnDemand {
...
+			request_backoff_start: Self::santize_circuit_breaker_input(request_backoff_start, "Request initial backoff time window"),
+			request_backoff_max: Self::santize_circuit_breaker_input(request_backoff_max, "Request maximum backoff time window"),
+			request_backoff_rounds_max,
+		}
+	}
+
+	fn santize_circuit_breaker_input(dur: Duration, name: &'static str) -> Duration {
+		if dur.as_secs() < 1 {
+			warn!(target: "on_demand",
+				"{} is too short must be at least 1 second, configuring it to 1 second", name);
```
ordian: If it's not allowed to be less than 1 second, maybe we should use seconds instead of ms for duration params?

ethcore/light/src/on_demand/request_guard.rs
cheme: Putting the comment as the first line of the else clause wouldn't really shock me.

ethcore/light/src/on_demand/response_guard.rs
cheme: Sorry to ask, did I lock something?

ethcore/light/src/on_demand/request_guard.rs
sorpaas: Yeah found an previous example: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/evm/src/interpreter/gasometer.rs#L411
It does look alright.

ethcore/light/src/on_demand/response_guard.rs
niklasad1: @cheme 

haha, I was referring to the functionality that this PR breaks i.e, `replacing the response limit per peer` and to use `a simple timeout` instead? If, that is not sorted there is no point and implementing other things IMHO!

ethcore/light/src/on_demand/mod.rs
niklasad1: Thanks, I rushed that part but it is fixed now at least!

ethcore/light/src/on_demand/response_guard.rs
niklasad1: Ok, thanks for taking the time but to verify that I understood your concerns:

1. This PR still might be useful (I want to add that it gives some information to the user why the request to the full-node failed or least what the majority response was)
2. Circuit breaker for `responses` are useless and we should replace it with a simple timeout as stated you pointed out earlier thanks for that!
3. Exponentially backoff for requests is useful but because we only register `bad responses` here to it is better to use `consecutive failures` (also easiser to understand)

>  I just observe that there is a hard limit at 30 attempts for exponential backoff), here I think that using consecutive failure strategy could make sense.

I think you are wrong there as you can see, https://github.com/dmexe/failsafe-rs/blob/master/src/backoff.rs#L136 the backoff is always regenerated. I think the `MAX_RETRIES` is used to calcuate "binary exponential backoff" by shifting the value `1 <<  MAX_RETRIES` and bigger value than 63 will overflow. That's my best guess 

> Still, I do not really get why we use success rate over consecutive failure (success rate can only be 0 or 100 (request end on first success and we use our constructions on a per request basis) whereas consecutive failure takes a lot of possible states and is somehow managed as the error counter in request_guards (I may be wrong I did not read the crate implementation)).

Basically, the request will run `forever` until it gets a successful response in the `circuit breaker`. That is why we use an upperbound  `max_failures` in the RequestGuard when to drop the request

It is okay for me to try with `consecutive failures instead` but then I suggest setting to value to 1 and then backoff and repeat until we either got a successful response or backoff #max_number in order to avoid doing too many N/W requests.

cheme: > I think you are wrong there as you can see, https://github.com/dmexe/failsafe-rs/blob/master/src/backoff.rs#L136 the backoff is always regenerated. I think the MAX_RETRIES is used to calcuate "binary exponential backoff" by shifting the value 1 << MAX_RETRIES and bigger value than 63 will overflow. 

Oups, sorry, I should have check more cautiously, thanks for looking at it.

> Basically, the request will run forever until it gets a successful response in the circuit breaker. That is why we use an upperbound max_failures in the RequestGuard when to drop the request

That is what I understood :-)

I feel like consecutive failure before backoff could make sense as a percentage of the current number of peers, but I do not really have things to back that (except that when testing a few month back I usually got a lot of peers returning invalid value but when we do not get lot of peers we want to wait). Isn't it a kind of error rates strategy. I start to wonder if it may make sense to write our own tailored 'failure policy' like 'ErrorRateOverTimeWindow' (just thinking loud).

I do not know about dropping response_guard, am I correct if I say that 'request_guard' and 'response_guard' are both in 'Pending' struct ? (just wondering if they can be merged, it probably only make sense with custom 'failure policy')

niklasad1: > I do not know about dropping response_guard, am I correct if I say that 'request_guard' and 'response_guard' are both in 'Pending' struct ? (just wondering if they can be merged, it probably only make sense with custom 'failure policy')

Yes, they are. Thus the first of `response_guard` and `request_guard` to detect a failure will drop the `Pending`. The idea from the beginning was to have a two failure policies of the `circuit-breaker` and combine them but because the state can't be accessed and it will never terminate I created two wrappers instead. The problem is that we can just query if the request is permitted or not thus can't tell if it will backoff will be closed/half-open or stay open forever. So, we can either an additional timeout or keep some state by our selves. 

I don't really understand what you are suggesting and the relationship between `requests` and `responses` when you talk about `our own faliure policy`.

What I can pick off is this:
- We should start with consecutive failures based on the number of peers
- When the peers are under some threshold we should backoff?
 - ?

Well, I think the circuit breaker API doesn't provide the functinality to change the peers for example after the circuit breaker has been created which makes that solution `hard` to get right (unless we want to write an circuit breaker ourselves)

Personally, I don't have any inutution on what this values should be and suggest just to keep it simple (e.g, consecutive failures then backoff and continue backoff and so on). The consecutive failures shall be configured by the user

niklasad1:
Thanks, Afri, it's far from being merged :P

ethcore/light/src/on_demand/response_guard.rs
cheme: :+1: for keeping it simple, I don't either have strong intuition on value but I identify two case where my previous configuration where not good : 
- start of a light client : there is to few nodes connected and we probably want to wait
- well established connection in light client : there will be no new node so waiting does not make a lot of sense.

About  'our own failure policy', I was mainly interpolating that our need may be  some 'RateOverTimeWindow', but as we count error and not success our own implementation of failsafe trait `FailurePolicy` could be an idea. Thinking twice, going for the crate success rate counting could be another target, but it would involve managing circuit breaker at a per peer level instead of a per request level. So I see two scenario (I do not say that they should be include in this pr):
- we will keep per request management, so we count errors only -> go with error count policy or a custom error rate one.
- we plan to switch to per peerid circuit so we can count multiple success (having mutex in the circuit does make sense with this scenario) and keeping a rate success is explain by this future move. The issue here is that hour request handler are not threadsafe (not sure they need to be, maybe they can stay per request specific but using per user id circuit breakers).

niklasad1: All right, yeah I think it is out-of-scope for this PR but we can create an issue and investigate it further!

amaurymartiny:
The error format is good for me

ethcore/light/src/on_demand/mod.rs
```diff
@@ -19,44 +19,54 @@
...
-/// The default time limit in milliseconds for inactive (no new peer to connect to) OnDemand queries (0 for unlimited)
-pub const DEFAULT_QUERY_TIME_LIMIT: Duration = Duration::from_millis(10000);
-
-const NULL_DURATION: Duration = Duration::from_secs(0);
+/// The initial backoff interval for OnDemand queries
+pub const DEFAULT_REQUEST_MIN_BACKOFF_DURATION: Duration = Duration::from_secs(10);
+/// The maximum request interval for OnDemand queries
+pub const DEFAULT_REQUEST_MAX_BACKOFF_DURATION: Duration = Duration::from_secs(100);
+/// The default window length a response is evaluated
+pub const DEFAULT_RESPONSE_TIME_TO_LIVE: Duration = Duration::from_secs(60);
```
ordian: Why was it changed from 10 secs to 1min, isn't 1min too long for a user to wait?

parity/cli/mod.rs
```diff
@@ -1820,8 +1835,11 @@ mod tests {
 			arg_snapshot_threads: None,
 
 			// -- Light options.
-			arg_on_demand_retry_count: Some(15),
-			arg_on_demand_inactive_time_limit: Some(15000),
+			arg_on_demand_response_time_window: Some(2000),
```
ordian: Could you update these numbers to seconds (`/ 1000`)?

ethcore/light/src/on_demand/mod.rs
sorpaas: I think we still need to (but haven't) revert this line at this moment, right?

ordian: Yeah, I should have commented instead of approving the PR to avoid the merge before fixing the comment, sorry.

PR #9792, opened 2018-10-22
Allow zero chain id in EIP155 signing process

sorpaas:
EIP155 actually didn't specify chainid 0 is invalid, and it can be valid -- `35` means refined v to be `0`, and `36` means refined v to be `1`. This PR fixes that.


ascjones:
Tests are failing

sorpaas:
We need to either:

* Disable those tests.
* Or force EIP155 being disabled when network id is zero.

sorpaas:
Reopening this for a while as I'm really confused why those tests fail.

PR #9804, opened 2018-10-23
Add readiness check for docker container

mazamats:
Since parity is built for "mission critical use", I thought other operators may see the need for this.

Adding the `curl` and `jq` commands allows for an extremely simple health check to be usable in container orchestrators.

For ease of use, I've added a simple health check script that is ready to go.

Here is an example of it being used in Kubernetes

```yaml
    spec:
      containers:
        image: parity/parity:stable
        imagePullPolicy: Always
        name: parity
        readinessProbe:
          exec:
            command:
            - /check_sync.sh
          failureThreshold: 3
          initialDelaySeconds: 5
          periodSeconds: 5
          successThreshold: 1
          timeoutSeconds: 1
        resources:
          limits:
            cpu: "1"
            memory: 4Gi
          requests:
            cpu: 500m
            memory: 2Gi
```

### DIFF
--- a/scripts/docker/hub/Dockerfile
+++ b/scripts/docker/hub/Dockerfile
@@ -31,6 +31,8 @@ COPY artifacts/x86_64-unknown-linux-gnu/$TARGET ./bin/$TARGET
 RUN echo "#!/bin/bash \n ${TARGET} \$@" > ./entrypoint.sh
 RUN chmod +x ./entrypoint.sh
 
+ADD check_sync.sh /check_sync.sh
+
 # setup ENTRYPOINT
 EXPOSE 5001 8080 8082 8083 8545 8546 8180 30303/tcp 30303/udp
 ENTRYPOINT ["./entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


mazamats:
[clabot:check]

scripts/docker/hub/Dockerfile
```diff
@@ -31,6 +31,8 @@ COPY artifacts/x86_64-unknown-linux-gnu/$TARGET ./bin/$TARGET
 RUN echo "#!/bin/bash \n ${TARGET} \$@" > ./entrypoint.sh
 RUN chmod +x ./entrypoint.sh
 
+ADD check_sync.sh /check_sync.sh
```
5chdn: missing a leading dot here ./check_sync.sh

mazamats: The script should still be executable and I would prefer to invoke it from `/check_sync.sh` rather than `/home/parity/check_sync.sh`

Either works though, the latter is just more cumbersome.

Issue #9737, opened 2018-10-11
Dev account can not be unlocked on first launch of parity


berndbohmeier:
_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: v2.0.7-stable
- **Operating system**: Linux
- **Installation**: docker hub
- **Fully synchronized**: yes
- **Network**: dev chain
- **Restarted**: yes

I use the the option `--config dev` to run a chain with dev configuration. This creates the default account `0x00a329c0648769a73afac7f9381e08fb43dbea72`. I try to unlock the account with `--unlock=0x00a329c0648769a73afac7f9381e08fb43dbea72`. Sadly this does not work at the first run and gives the error message: `Account 0x00a3…ea72 not found for the current chain. You can create an account via RPC, UI or parity account new --chain dev --keys-path /home/parity/chaindata/keys.`
If I run it without the unlock option first, and after with the unlock option, it works, so it seems it needs to write the account first to disk. This is annoying as it does not work with a fresh docker image. 

Proposed solution: Make it possible to already unlock the account at the first run.

debris:
> This creates the default account `0x00a329c0648769a73afac7f9381e08fb43dbea72`

Is there any documentation that describes this?

> If I run it without the unlock option first, and after with the unlock option, it works, so it seems it needs to write the account first to disk. This is annoying as it does not work with a fresh docker image.

To be honest, I'm not aware of this feature. It seems more like a bug as `0x00a329c0648769a73afac7f9381e08fb43dbea72` is an account created for an empty seedphrase. @tomusdrw do you know anything about this?

berndbohmeier:
Yes, it is documented here https://wiki.parity.io/Private-development-chain

debris:
Thanks! I believe the problems lies in fact that even though the password is an empty string, we still expect the password to be provided.

> An address containing a lot of Ether (`0x00a329c0648769a73afac7f9381e08fb43dbea72`) has automatically added with a password being an empty string.

I believe this should be clarified in the documentation. And parity should also fix wrong behaviour on the first run.

joshua-mir:
@berndbohmeier for a workaround, you can use a simple config file like

```
# This config should be placed in following path:
#   ~/.local/share/io.parity.ethereum/config.toml

[parity]
# Local Development Chain
chain = "dev"

[account]
# From: [0x00a329c0648769a73afac7f9381e08fb43dbea72] you'll be able to send transactions without password.
[...]

PR #9813, opened 2018-10-26
Insert dev account before unlocking

sorpaas:
So that dev account can be unlocked on the first run.

Fixes https://github.com/paritytech/parity-ethereum/issues/9737


ordian:
@sorpaas shall we backport this?

sorpaas:
@ordian Yes. Sorry I missed the tag!

Issue #9031, opened 2018-07-03
[Light client] NotEnoughBaseGas for RPC without specifying any Gas.


Tbaut:
> I'm running:
>
> - **Which Parity version?**: ~full node 1.11.5~, now light node v2.1.3
> - **Which operating system?**: Linux
> - **How installed?**: binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: kovan and mainnet
> - **Did you try to restart the node?**: yes

Edit:
I could reproduce with Parity Ethereum `--light` v2.1.3 this call on Kovan and mainner:
```bash
curl --data '{"method":"eth_estimateGas","params":[{"value":"0x0","data":"0x","from":"0x00dfc93112abd2578503b667b95491b101281f2b"}],"id":"c31e5d5f401bea24cf8b40340f532cee","jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
{"jsonrpc":"2.0","error":{"code":-32015,"message":"Transaction execution error.","data":"NotEnoughBaseGas { required: 53000, got: 50000 }"},"id":"c31e5d5f401bea24cf8b40340f532cee"}
```

The node with `rpc` logging:
```bash
2018-10-22 20:27:28   TRACE rpc  Request: {"method":"eth_estimateGas","params":[{"value":"0x0","data":"0x","from":"0x00dfc93112abd2578503b667b95491b101281f2b"}],"id":"c31e5d5f401bea24cf8b40340f532cee","jsonrpc":"2.0"}.
2018-10-22 20:27:28   DEBUG rpc  [Some(Str("c31e5d5f401bea24cf8b40340f532cee"))] Took 231ms
2018-10-22 20:27:28   DEBUG rpc  Response: {"jsonrpc":"2.0","error":{"code":-32015,"message":"Transaction execution error.","data":"NotEnoughBaseGas { required: 53000, got: 50000 }"},"id":"c31e5d5f401bea24cf8b40340f532cee"}.
```

----------------------
I get the error `NotEnoughBaseGas` when calling:
```bash
$ curl --data '{"method":"eth_estimateGas","params":[{"from":"0x004702bdc..028600"}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
[...]

PR #9824, opened 2018-10-29
light-fetch: Differentiate between out-of-gas/manual throw and use required gas from response on failure

niklasad1:
Attempt to close https://github.com/paritytech/parity-ethereum/issues/6155 & close https://github.com/paritytech/parity-ethereum/issues/9031

What does this PR:

1. Differentiate between out-of-gas/manual throw
2. Changes the start_gas from 50_000 to 60_000
3. If too little gas is submitted in a transaction then we use to the required `start_gas` replied to us to set the next `start_gas` if it doesn't exceed the block's gas limit (however this could be changed to set the `start_gas` regardless of block_limit and continue until we get `BlockGasLimitReached` error instead, will probably require at least one extra network request)

Example execution with 50_000 as start_gas:
```bash
2018-10-29 15:45:05  jsonrpc-eventloop-1 TRACE light_fetch  Placing execution request for 50000 gas in on_demand
2018-10-29 15:45:05  jsonrpc-eventloop-1 TRACE light_fetch  Not enough start gas provided required: 53000, got: 50000
2018-10-29 15:45:05  jsonrpc-eventloop-1 TRACE light_fetch  Placing execution request for 53000 gas in on_demand
```

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -627,7 +627,7 @@ fn execute_read_only_tx(gas_known: bool, params: ExecuteParams) -> impl Future<I
 						// `OutOfGas` exception, try double the gas
 						if let Some(vm::Error::OutOfGas) = executed.exception {
 							// block gas limit already tried, regard as an error and don't retry
-							if params.tx.gas == params.hdr.gas_limit() {
+							if params.tx.gas >= params.hdr.gas_limit() {
 								trace!(target: "light_fetch", "OutOutGas exception received, gas increase: failed");
 							} else {
 								params.tx.gas = cmp::min(params.tx.gas * 2_u32, params.hdr.gas_limit());
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -19,12 +19,12 @@
 use std::cmp;
 use std::sync::Arc;
 
-use light::on_demand::error::Error as OnDemandError;
 use ethcore::basic_account::BasicAccount;
 use ethcore::encoded;
 use ethcore::filter::Filter as EthcoreFilter;
 use ethcore::ids::BlockId;
 use ethcore::receipt::Receipt;
+use ethcore::executed::ExecutionError;
 
 use jsonrpc_core::{Result, Error};
 use jsonrpc_core::futures::{future, Future};
@@ -38,6 +38,7 @@ use light::on_demand::{
 	request, OnDemand, HeaderRef, Request as OnDemandRequest,
 	Response as OnDemandResponse, ExecutionResult,
 };
+use light::on_demand::error::Error as OnDemandError;
 use light::request::Field;
 
 use sync::LightSync;
@@ -202,8 +203,8 @@ impl LightFetch {
 	/// Helper for getting proved execution.
 	pub fn proved_read_only_execution(&self, req: CallRequest, num: Trailing<BlockNumber>) -> impl Future<Item = ExecutionResult, Error = Error> + Send {
 		const DEFAULT_GAS_PRICE: u64 = 21_000;
-		// starting gas when gas not provided.
-		const START_GAS: u64 = 50_000;
+		// (21000 G_transaction + 32000 G_create + some marginal to allow a few operations)
+		const START_GAS: u64 = 60_000;
 
 		let (sync, on_demand, client) = (self.sync.clone(), self.on_demand.clone(), self.client.clone());
 		let req: CallRequestHelper = req.into();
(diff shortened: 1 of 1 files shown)


ordian:
@niklasad1 looks nice, but I'm yet to test it. Could you add a unit test? :)

niklasad1:
@ordian 
Yes, ideally I should have but it would require to mock a `Full Node` which is a much bigger task. Is manual testing ok for this one?

Might be a little annoying but change the `start_gas` in `proved_read_only_execution` to some value and create `eth_call` request without gas from an account with zero balance!

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,31 @@ struct ExecuteParams {
...
+						trace!(target: "light_fetch", "Not enough start gas provided required: {}, got: {}", required, got);
+						if required <= params.hdr.gas_limit() {
+							params.tx.gas = required;
+							return Ok(future::Loop::Continue(params))
+						} else {
+							warn!(target: "light_fetch", "Required gas is bigger than block header's gas dropping the request");
+							Ok(future::Loop::Break(Err(ExecutionError::NotEnoughBaseGas { required, got })))
+						}
+					}
+					// Non-recoverable execution error
```
cheme: Here (we need to test to assert the error actually occurs), there is possibly 'ExecutionError::BlockGasLimitReached', maybe we do not need the previous test.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,31 @@ struct ExecuteParams {
...
+							trace!(target: "light_fetch", "Execution exception: {}", exception);
 						}
-
 						Ok(future::Loop::Break(Ok(executed)))
 					}
+					Err(ExecutionError::NotEnoughBaseGas { required, got }) => {
+						trace!(target: "light_fetch", "Not enough start gas provided required: {}, got: {}", required, got);
+						if required <= params.hdr.gas_limit() {
+							params.tx.gas = required;
+							return Ok(future::Loop::Continue(params))
```
cheme: I am just thinking about it, but we probably want to apply a margin (a n% fix additional cost) : since the transaction may not run in the same block (and even in the same block it can vary) the cost may change when running the actual transaction.
It is not really related to this PR, but likely to happen with some contracts, this PR may makes it more obvious (previously by doubling gas cost there was lesser chance for it to happen).

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,31 @@ struct ExecuteParams {
...
-						// exception?
-						if executed.exception.is_some() {
-							let old_gas = params.tx.gas;
-							params.tx.gas = params.tx.gas * 2u32;
-							if params.tx.gas > params.hdr.gas_limit() {
-								params.tx.gas = old_gas;
-							} else {
-								return Ok(future::Loop::Continue(params))
-							}
+						// `OutOfGas` exception can't occur here because of virtual execution context
```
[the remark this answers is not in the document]
niklasad1: I don't understand this. `NotEnoughBasGas` is not an exception, see https://github.com/paritytech/parity-ethereum/blob/master/ethcore/vm/src/error.rs#L38-#L86

Clarify please :P

rpc/src/v1/helpers/light_fetch.rs
niklasad1: Well, If we get ```BlockGasLimitReached``` then it is defintetely a failuty request and AFAIK any of the others errors can't be recovered!

The possible errors can be found:
https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/executed.rs#L74-#L118

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -202,8 +203,8 @@ impl LightFetch {
 	/// Helper for getting proved execution.
 	pub fn proved_read_only_execution(&self, req: CallRequest, num: Trailing<BlockNumber>) -> impl Future<Item = ExecutionResult, Error = Error> + Send {
 		const DEFAULT_GAS_PRICE: u64 = 21_000;
-		// starting gas when gas not provided.
-		const START_GAS: u64 = 50_000;
+		// (21000 G_transaction + 32000 G_create)
```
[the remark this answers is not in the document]
niklasad1: Sure, let's change it 60 then!

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
cheme: :+1:  indeed, we could use vm::Error::OutOfGas instead.

rpc/src/v1/helpers/light_fetch.rs
cheme: Yes I was just wondering if the test above could be remove, but it seems safe to keep it.

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
niklasad1: Right, got it very good catch :+1:

rpc/src/v1/helpers/light_fetch.rs
rphmeier: this is something that we have historically decided not to handle on the node side in order to give the most faithful estimate we can. Wrappers around the RPCs typically add a few percent to the result of `estimateGas`.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,40 @@ struct ExecuteParams {
...
 fn execute_read_only_tx(gas_known: bool, params: ExecuteParams) -> impl Future<Item = ExecutionResult, Error = Error> + Send {
 	if !gas_known {
 		Box::new(future::loop_fn(params, |mut params| {
 			execute_read_only_tx(true, params.clone()).and_then(move |res| {
 				match res {
 					Ok(executed) => {
-						// TODO: how to distinguish between actual OOG and
-						// exception?
-						if executed.exception.is_some() {
+						// `OutOfGas` exception can only occur on full-nodes that run `Parity-Ethereum v2.1.1` or older
```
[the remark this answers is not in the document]
niklasad1: Thanks, but should be sufficient just to change the comment right?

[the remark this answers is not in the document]
cheme: I think so, (we thankfully kept code for iterative querying).

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,39 @@ struct ExecuteParams {
...
-						// TODO: how to distinguish between actual OOG and
-						// exception?
-						if executed.exception.is_some() {
+						// `OutOfGas` exception, try double the gas
+						if let Some(vm::Error::OutOfGas) = executed.exception {
 							let old_gas = params.tx.gas;
-							params.tx.gas = params.tx.gas * 2u32;
+							params.tx.gas = params.tx.gas * 2_u32;
 							if params.tx.gas > params.hdr.gas_limit() {
+								trace!(target: "light_fetch", "OutOutGas exception received, gas increase: failed");
```
tomusdrw: Why dont we try with the `params.hdr.gas_limit()` at the very end? I can easily imagine a situation, where `block_gas_limit / 2 + 1` is not sufficient, but `block_gas_limit / 2 + 2` is enough.

ordian: [there is todo](https://github.com/paritytech/parity-ethereum/blob/5f3ae4dee34ede2738473b35db22f8d65351b69b/rpc/src/v1/impls/light/eth.rs#L398) to use binary search, maybe we should implement that (copy https://github.com/paritytech/parity-ethereum/blob/5f3ae4dee34ede2738473b35db22f8d65351b69b/ethcore/src/client/client.rs#L1539 ?).

cheme:
Same thing here, I just feel that @tomusdrw comment should be addressed (when we multiply gas by 2 we should finish with max block size (if not already eq to max block size)).

Something along the line of : 
```rust
if params.tx.gas == params.hdr.gas_limit() {
  trace!(target: "light_fetch", "OutOutGas exception received, gas increase: failed");
} else {
  params.tx.gas = params.tx.gas * 2_u32;

  if params.tx.gas > params.hdr.gas_limit() {
    params.tx.gas = params.hdr.gas_limit();
  }
  trace!(target: "light_fetch", "OutOutGas exception received, gas increase: succeed");
  return Ok(future::Loop::Continue(params))
}
  
```

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -615,28 +616,41 @@ struct ExecuteParams {
...
-						// exception?
-						if executed.exception.is_some() {
-							let old_gas = params.tx.gas;
-							params.tx.gas = params.tx.gas * 2u32;
-							if params.tx.gas > params.hdr.gas_limit() {
-								params.tx.gas = old_gas;
+						// `OutOfGas` exception, try double the gas
+						if let Some(vm::Error::OutOfGas) = executed.exception {
+							// block gas limit already tried, regard as an error and don't retry
+							if params.tx.gas == params.hdr.gas_limit() {
```
cheme:
```suggestion
							if params.tx.gas >= params.hdr.gas_limit() {
```

niklasad1: Yeah, `tx.gas` should never be bigger than gas_limit() if the logic is correct

PR #9834, opened 2018-10-30
Do not use the home directory as the working dir in docker

fubhy:
Moving the binary and the entry point to the home directory makes using the docker image in Kubernetes really hard. At the same time, there is no good reason to do so as far as I can tell (all that we want is to use a non-root user which we can absolutely do even without the home directory).

### DIFF
--- a/scripts/docker/hub/Dockerfile
+++ b/scripts/docker/hub/Dockerfile
@@ -16,7 +16,7 @@ RUN apt clean -y
 RUN rm -rf /tmp/* /var/tmp/* /var/lib/apt/lists/*
 
 RUN groupadd -g 1000 parity \
-  && useradd -u 1000 -g parity -s /bin/sh parity
+  && useradd -m -u 1000 -g parity -s /bin/sh parity
 
 USER parity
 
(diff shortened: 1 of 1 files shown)


scripts/docker/hub/Dockerfile
```diff
@@ -16,16 +16,12 @@ RUN apt clean -y
 RUN rm -rf /tmp/* /var/tmp/* /var/lib/apt/lists/*
 
 RUN groupadd -g 1000 parity \
-  && useradd -m -u 1000 -g parity -s /bin/sh parity
+  && useradd -u 1000 -g parity -s /bin/sh parity
```
asymmetric: I think it would still make sense to create the home directory, so that we can then create `/home/parity/.local/share/io.parity.ethereum/` in the Dockerfile. This would allow people to then mount a chain on the host into the container  without [permission issues](https://github.com/paritytech/parity-ethereum/pull/9689#issuecomment-429565718) and without having to pass `--base-path`.

TL;DR: I thnk we should leave the `-m` flag.

fubhy:
To be honest, I can't think of a scenario where one would want to run the docker image without mounting the base path and passing a custom path to --base-path. 

On the other hand, it does not hurt to create the home directory as long as we don't move any binaries or such there.

asymmetric:
For my curiosity, what was the problem in k8s with having the binary under `~/.bin`?

fubhy:
Because by default, kubernetes runs under a random UID and GID as an additional security precaution. 

K8s suggests to not rely.on the USER directive at all.

It's possible to configure your way around it of course but that's a bit of a uneccesary hassle

5chdn:
This broke our docker builds

PR #9835, opened 2018-10-30
avoid bloom_hash(x, 0)

bennetyee:
bloom_hash(x, 0) == bloom_hash(x, 1) for all x, so we shift the range
of k_i values up to avoid the k_i == 0 case.

The current buggy code tries to have k independent hashes, but in actuality is using k-1 different hashes.


bennetyee:
[clabot:check]

bennetyee:
my apologies.  apparently backwards compatibility is very important, so i'll close this (and the other alternative fix).  it is still bad that the statistical estimates based on `k` is going to be wrong.

PR #9836, opened 2018-10-30
Alternative fix for hash function independence.

bennetyee:
This simplifies the bloom_hash function to not special case k_i = 0 or
1.  Because we are using wrapping_add and wrapping_mul, the 0 case
just returns the base_hash and the 1 case returns base_hash +
(base_hash % 0xff...fc5), so these two cases should differ.

The current buggy code tries to have k independent hashes, but in actuality is using k-1 different hashes.

### DIFF
--- a/util/bloom/src/lib.rs
+++ b/util/bloom/src/lib.rs
@@ -178,11 +178,7 @@ impl Bloom {
 	}
 
 	fn bloom_hash(base_hash: u64, k_i: u32) -> u64 {
-		if k_i < 2 {
-			base_hash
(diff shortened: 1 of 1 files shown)


bennetyee:
https://github.com/paritytech/parity-ethereum/pull/9835 is the alternative fix.

bennetyee:
my apologies.  apparently backwards compatibility is very important, so i'll close this (and the other alternative fix).  it is still bad that the statistical estimates based on `k` is going to be wrong.

PR #9839, opened 2018-10-31
Clean up existing benchmarks

sorpaas:
* Move `benches` feature to `benches/` subfolder. `benches` feature is not considered special for `cargo`, so plain `cargo +nightly bench` won't compile with it. Moving benchmarks to subfolder fixes this issue.
* Fix `evm` benches compile.


ordian:
Consider using [criterion](https://github.com/japaric/criterion.rs) as it provides the following benefits:
* statistical analysis, detects regressions
* gnuplot charts for free
* stable-compatible

Also #9762 started to use it.

ethcore/src/state/mod.rs
```diff
@@ -1151,6 +1151,8 @@ impl<B: Backend> State<B> {
 					if require_code {
 						let addr_hash = account.address_hash(a);
 						let accountdb = self.factories.accountdb.readonly(self.db.as_hashdb(), addr_hash);
+
+						// FIXME (Issue #9838): update_account_cache can fail in rare cases, but we cannot return error in RefMut wrapper.
```
ordian: this is out of scope of this PR, but nice to have :+1:

PR #9840, opened 2018-10-31
Calculate gas for deployment transaction

### LINKED ISSUE #9708: [Private transactions] Cannot deploy private smart contract
Parity Ethereum version: Parity-Ethereum
Operating system: Linux
Installation: built from source with secretstore feature
Fully synchronized: yes
Network: Tobalaba
Restarted: yes

**actual** 
I'm tested the private transactions tutorial https://github.com/ngyam/tutorial-secretstore-privatetx
on tobalaba_local 
All steps on private transactions works well, but when i change the smart contract sample (Test.sol) by a new one, i cannot deploy the contracts with the error:

```
UnhandledPromiseRejectionWarning: Error: Returned error: Transaction cost exceeds current gas limit. Limit: 8012172, got: 10200000. Try decreasing supplied gas.
```

But the deployment of this contract works well in public mode (and cost half time gas)
Imported #133 0xab28…9de3 (1 txs, 0.44 Mgas, 1 ms, 2.12 KiB)

Code of the smart contract that cause the issue.

```solidity
pragma solidity ^0.4.24;

contract Consumption {

   mapping (address => mapping (bytes32 => Data))  consumption;
   mapping (address => mapping (address => bool))  authorization;

   struct Data {
       uint[] timestamp;
       uint256[] value;
   }

   function consume(uint256 _consumption, uint _timestamp,  string _date) public {
       bytes32 hash = keccak256(_date);
       consumption[msg.sender][hash].timestamp.push(_timestamp);
       consumption[msg.sender][hash].value.push(_consumption);
   }

   function getConsumption(address _consumptionAdress, string _date) public returns (uint[],uint256[]) {
       require(authorization[_consumptionAdress][msg.sender]);
       bytes32 hash = keccak256(_date);
       return (consumption[_consumptionAdress][hash].timestamp,consumption[_consumptionAdress][hash].value);
   }

   function authorize(address _address, bool _authorize) public {
       authorization[msg.sender][_address] = _authorize;
   }
```

 **expected behavior** 
Smart contracts can be deployed 

**steps to reproduce** 
Follow the https://github.com/ngyam/tutorial-secretstore-privatetx, by replacing Test.Sol by the code above

grbIzl:
Gas for the deployment transaction for the private contract was overestimated. More accurate method with client.estimate_gas implemented.
Closes #9708

### DIFF
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -556,28 +556,20 @@ impl Provider where {
 			.and_then(|h| h.decode().map_err(|_| ErrorKind::StateIncorrect).into())?;
 		let (executed_code, executed_state) = (executed.code.unwrap_or_default(), executed.state);
 		let tx_data = Self::generate_constructor(validators, executed_code.clone(), executed_state.clone());
-		let gas = match self.client.estimate_gas(&Transaction {
+		let mut tx = Transaction {
 			nonce: nonce,
 			action: Action::Create,
 			gas: u64::max_value().into(),
 			gas_price: gas_price,
 			value: source.value,
-			data: tx_data.clone(),
-			}.fake_sign(sender),
-		&state,
-		&header) {
+			data: tx_data,
+		};
+		tx.gas = match self.client.estimate_gas(&tx.clone().fake_sign(sender), &state, &header) {
 			Ok(estimated_gas) => estimated_gas,
 			Err(_) => self.estimate_tx_gas(validators, &executed_code, &executed_state, &[]),
 		};
-		Ok((Transaction {
-			nonce: nonce,
-			action: Action::Create,
-			gas: gas.into(),
-			gas_price: gas_price,
-			value: source.value,
-			data: tx_data,
-		},
-		executed.contract_address))
+
+		Ok((tx, executed.contract_address))
 	}
 
 	fn estimate_tx_gas(&self, validators: &[Address], code: &Bytes, state: &Bytes, signatures: &[Signature]) -> U256 {
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -556,7 +556,7 @@ impl Provider where {
 			.and_then(|h| h.decode().map_err(|_| ErrorKind::StateIncorrect).into())?;
 		let (executed_code, executed_state) = (executed.code.unwrap_or_default(), executed.state);
 		let tx_data = Self::generate_constructor(validators, executed_code.clone(), executed_state.clone());
-		let gas = match self.client.estimate_gas(&Transaction{
+		let gas = match self.client.estimate_gas(&Transaction {
 			nonce: nonce,
 			action: Action::Create,
 			gas: u64::max_value().into(),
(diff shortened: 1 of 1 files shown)


ethcore/private-tx/src/lib.rs
```diff
@@ -550,21 +551,44 @@ impl Provider where {
...
-		let gas: u64 = 650000 +
-			validators.len() as u64 * 30000 +
-			executed.code.as_ref().map_or(0, |c| c.len() as u64) * 8000 +
-			executed.state.len() as u64 * 8000;
+		let header = self.client.block_header(block)
+			.ok_or(ErrorKind::StatePruned)
+			.and_then(|h| h.decode().map_err(|_| ErrorKind::StateIncorrect).into())?;
+		let (executed_code, executed_state) = (executed.code.unwrap_or_default(), executed.state);
+		let tx_data = Self::generate_constructor(validators, executed_code.clone(), executed_state.clone());
+		let gas = match self.client.estimate_gas(&Transaction{
```
sorpaas: nit: missing space

grbIzl: For the deployment transaction the speed is not so important as it's one-time action. At the same time overestimation for the gas can (and actually is, see the corresponding bug) cause problems as wrapper contract is pretty complex and stores private contracts inside. In case, when the private contract is big enough, this overestimation exceeds gas_limit.

ethcore/private-tx/src/lib.rs
```diff
@@ -550,21 +551,44 @@ impl Provider where {
...
+		let (executed_code, executed_state) = (executed.code.unwrap_or_default(), executed.state);
+		let tx_data = Self::generate_constructor(validators, executed_code.clone(), executed_state.clone());
+		let gas = match self.client.estimate_gas(&Transaction {
+			nonce: nonce,
+			action: Action::Create,
+			gas: u64::max_value().into(),
+			gas_price: gas_price,
+			value: source.value,
+			data: tx_data.clone(),
+			}.fake_sign(sender),
```
andresilva: Nit: I think indentation is a bit weird like this

grbIzl:
@andresilva yes, usage of max_gas will address the issue, but IMHO such solution will damage "informativeness" of this API call (I mean, create deployment transaction method). Right now this call returns estimation for the gas and it makes a lot of sense since it may be totally unclear for the external user, how much gas it can take (because a lot of things hidden under the hood). At the same time, as I've already explained, this call is one-time action and will be done only once during system setup. So IMHO the performance is less important here, than the mentioned "informativeness"

andresilva:
@grbIzl Sorry for not moving this PR forward, I pushed a commit to fix the style nit, approve and merge if you agree :+1:.

sorpaas:
@grbIzl I still worry about use cases such as chains that may want to deploy new private transaction account for every user. In those cases we potentially can have a lot of deployments! I understand that probably no one uses it like that currently, but it indeed feels feasible to have this use cases in certain situations, and just want to make sure we don't shot ourselves in the foot!

* If we use `Client::estimate_gas`, it may be slow for them. And for those users there's probably no way to opt out if this is a bottleneck.
* If we just increase the current gas cap estimate. For users that do need the informativeness of estimated gas, they can take the transaction payload, feed it to `eth_estimateGas` and get what they want. For users that just want to get the payload and deploy things fast, it won't affect them.

PR #9814, opened 2018-10-26
Implement NoProof for json tests and update tests reference (replaces #9744)

cheme:
Closes #9590 .

Update to latest ethereum/tests .
Fix cost for constantinople test json.
Implements 'NoProof' engine from ethereum/tests#464 .

At this point with this PR everything pass except the case spotted by @holiman this week in https://gist.github.com/holiman/0154f00d5fcec5f89e85894cbb46fcb2 . 

I skip those tests. For completeness, @sorpaas indicated that it is not an easy fix : so since those case cannot happen on a mainchain I did not open and reference an issue for them (in the json defining tests to skip I only added a comment but not issue reference).

Note for reviewer:
- an other version of this PR exists with previous review comments is #9744 , this PR replaces it . 
- when running NoProof with CanonNoSeal, the upper gas limit verification is required : I switch it from verify_block_basic to the header check. I got  to put another trait method for it, a lighter change would be to make this check for every engine but it would change a bit the existing behavior (might not be an issue).

PR #9841, opened 2018-10-31
Move state root verification before gas used

sorpaas:
State root is usually the first indication of consensus errors. By this change, it hopefully makes consensus debugging from bug reports a little bit easier.

### DIFF
--- a/ethcore/src/verification/verification.rs
+++ b/ethcore/src/verification/verification.rs
@@ -245,14 +245,14 @@ fn verify_uncles(block: &PreverifiedBlock, bc: &BlockProvider, engine: &EthEngin
 
 /// Phase 4 verification. Check block information against transaction enactment results,
 pub fn verify_block_final(expected: &Header, got: &Header) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


sorpaas:
@ordian Yeah backporting this would indeed help!

ethcore/src/verification/verification.rs
```diff
@@ -245,15 +245,15 @@ fn verify_uncles(block: &PreverifiedBlock, bc: &BlockProvider, engine: &EthEngin
 
 /// Phase 4 verification. Check block information against transaction enactment results,
 pub fn verify_block_final(expected: &Header, got: &Header) -> Result<(), Error> {
+	if expected.state_root() != got.state_root() {
+		return Err(From::from(BlockError::InvalidStateRoot(Mismatch { expected: expected.state_root().clone(), found: got.state_root().clone() })))
```
niklasad1: Not introduced by this PR: clone() on `Copy-type` (H256 is copy)

sorpaas: Already fixed in #9814!

PR #9847, opened 2018-11-01
Health endpoint 

### LINKED ISSUE #7560: Feature Request: Return 200 OK for HTTP GET to root RPC endpoint
At the moment when I `curl http://localhost:8545` it 302's to the wallet UI at `http://localhost:8180`.

I'm trying to expose parity nodes running in Google Container Engine (GKE) that are sitting behind a Google Load Balancer.  Google's load balancer will not forward traffic unless the underlying service is healthy and the only option for a Google LB health check is a HTTP GET.  It would be super useful if parity could just respond 200 OK instead of the 302.

### LINKED ISSUE #9401: Need GET health check endpoint for deployment status
Running Parity 1.11.8

There is no documented health check which we can use for checking for service presence. Most services have a static GET /health, /status or /ping which returns either the current time as JSON or a static up:true or something that load balancers and monitoring systems can use to determine the service is running. 

We have been trying to use the peer count endpoint to determine if it's up but that endpoint is unreliable, often timing out after several seconds and leading to false service outage errors for us.

tomusdrw:
Closes #7560 
Closes #9401

Introduces new RPC method `parity_nodeStatus` that return OK in case the node is not syncing and you have at least one peer (exception: development chain).

This method is then used as a health API of HTTP server.


niklasad1:
@tomusdrw shall this be backported?

tomusdrw:
@niklasad1 no, I don't think backporting is required - it's a new feature not really a bugfix.

rpc/src/v1/impls/parity.rs
```diff
@@ -481,4 +487,21 @@ impl<C, M, U, S> Parity for ParityClient<C, M, U> where
 	fn submit_work_detail(&self, nonce: H64, pow_hash: H256, mix_hash: H256) -> Result<H256> {
 		helpers::submit_work_detail(&self.client, &self.miner, nonce, pow_hash, mix_hash)
 	}
+
+	fn status(&self) -> Result<()> {
+		let has_peers = self.client.spec_name() == "developmentchain" || self.sync.status().num_peers > 0;
```
ngotchac: Can't we have a more generic method to check if it's the `dev` chain? Or couldn't it be the same check as for the light implementation?

tomusdrw: Good point, was trying to look for any detection we do already, but haven't found anything.
In light we don't really have `spec_name` exposed, but I agree that current way is suboptimal.
I'll try to figure something out.

tomusdrw: @ngotchac addressed, could you have a look?

Tbaut:
Also please don't forget the doc :angel:

PR #9854, opened 2018-11-02
Fix docker script

5chdn:


gabreal:
https://gitlab.parity.io/parity/parity-ethereum/-/jobs/106905
and 
https://github.com/paritytech/parity-ethereum/blob/master/scripts/gitlab/publish-docker.sh#L20

5chdn:
Errhmmm, that's confusing, what would be the correct path then?

gabreal:
the line that I posted earlier shoult work.

PR #9855, opened 2018-11-02
ci: nuke the gitlab caches

5chdn:
For comparison: 

- **70 minutes** with cache: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/107291
- **33 seconds** without cache: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/107307

Both don't build anything. Just plain pipeline without and with cache.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,12 +13,6 @@ variables:
   BUILD_ARCH:                      amd64
   CARGO_TARGET:                    x86_64-unknown-linux-gnu
 
-cache:
-  key:                             "${CI_JOB_NAME}"
-  paths:
-    - ./target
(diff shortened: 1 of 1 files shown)


5chdn:
LOL this reduced the job time to 33 seconds.

gabreal:
the 33 seconds job didn't build anything. doesn't it slow down actual builds?

5chdn:
Honestly, I don't care. If caching takes more than 20 minutes, then we don't need caching. A fresh build on Linux takes around 20 minutes.

gabreal:
IMO it would be interesting to check how much speed-up the cache would give (easily comparable locally) and then think whether it's worth to use a different caching approach then.

PR #9845, opened 2018-10-31
[Almost Done] Clique POA/Rinkeby Read Support

thefallentree:
#5687 
do not merge yet, this is just for testing.

PR #9862, opened 2018-11-04
[WIP] Clique POA Implementation

jwasinger:
Turns out @thefallentree and myself were working on this simultaneously XD .  Anyways, here is what I have so far.  It's mainly just boilerplate code necessary to spin up a node using the *Clique* configuration which tries to continuously seal blocks.

Still needs a lot of work but I'll use this as a start.  I've also modified https://github.com/paritytech/parity-deploy to accomodate the changes here.  Will PR these soon.

Let me know what you think @thefallentree .  We can try to split the work of implementing the consensus logic between us.


jwasinger:
[clabot:check]

jwasinger:
Okay.  I had forgotten to push several files before.  It should build now.  I am able to run it but crashes after sealing several blocks XD .. Btw, I'll probably smash all these commits into one when everything is ready.

jwasinger:
> I am able to run it but crashes after sealing several blocks
Ah... This is some kind of race condition in my code where the engine is trying to seal several blocks with the same number simultaneously...

Unfortunately, it's going to be several weeks until I can continue on this.

jwasinger:
Ah...  I think this could be the issue: https://github.com/paritytech/parity-ethereum/pull/9862/files#diff-597dcb240a550e97d352501043e8f657R10 .

5chdn:
> Btw, I'll probably smash all these commits into one when everything is ready.

We will squash-merge this anyways, so no worries.

Apart from sealing, what's working? I tried to connect this branch to a Geth and Pantheon client running Clique, can you test my config? See https://github.com/goerli/testnet

thefallentree:
I have some questions about the code here: 

1. what is the purpose with StepEngine?  As my understanding clique engine either sign or wait for another singer forever. not sure if this is needed here.
2. Clique singer list is  stored in "extra_data" in the header,  there should be no validator list in the params here. and read from last checkpoint data.
3.  in Clique,  block author is always set to "0x0"  and we need to extract singer data from seal and change block author to singer , then process txns with gas fees. I think this maybe should be done in on_close_block (), can anyone with more understanding of parity codebase say if this is right path?

In either case, thx for the code, and I plan to continue from here. 

Cheers.

jwasinger:
> what is the purpose with StepEngine? As my understanding clique engine either sign or wait for another singer forever. not sure if this is needed here.

Maybe I don't fully understand the consensus engine.  My understanding is that the step engine is needed to trigger sealing.  Otherwise, it won't occur unless there are transactions.

> Clique singer list is stored in "extra_data" in the header, there should be no validator list in the params here. and read from last checkpoint data.

Agreed.

jwasinger:
> Apart from sealing, what's working?

Nothing at the moment.  it just compiles and tries to seal blocks.  I am dealing with an issue where my step engine (which is supposed to cause the client to seal blocks periodically, is encountering a race condition.  Unfortunately, I'm about to fly out on a vacation and probably don't have much time to look at this for the next two weeks.

jwasinger:
I think https://github.com/paritytech/parity-ethereum/pull/9862/files#diff-597dcb240a550e97d352501043e8f657R38 is a race condition because the client may not have been set using `register_client` when the timeout occurs.  I'm still trying to wrap my head around the exact purpose of these different components.

This code would also explain why the engine seems to not seal about 1/3 of the time that I start it up.

jwasinger:
> what is the purpose with StepEngine? As my understanding clique engine either sign or wait for another singer forever. not sure if this is needed here.

@thefallentree I have to study clique a bit more but this seems like it would be a DOS vulnerability.  Also, we need the step service to trigger sealing even when transactions aren't present because the signer may want to cast a vote.

Also, I looked at the step service code a bit more.  I think that we actually only want to trigger `client.update_sealing` once per prospective block.  The function is not reentrant as I thought.

5chdn:
Can you both submit your PR against https://github.com/goerli/parity-goerli ?

jwasinger:
Yup!  I want to test out a fix (for the race condition) first.  but then I will PR it in an hour or so.

thefallentree:
hey, I've updated my PR #9845 with correct rinkeby genesis files.  I will try to fixup the validation/sync part first, without worrying about sealing.

5chdn:
Merged into https://github.com/goerli/parity-goerli for easier collaboration. Creating a new PR once closer to be reviewable.

PR #9881, opened 2018-11-07
Make CALLCODE to trace value to be the code address

### LINKED ISSUE #9878: Trace returns inconsistent values for CALLCODE
- **Parity Ethereum version**: 2.1.4
- **Operating system**: Windows
- **Installation**: executable
- **Fully synchronized**: no
- **Network**: ethereum and dev test net

In issue #7166 it was reported that the trace API was reporting incorrect to/from address for delegatecall. In this issue it seemed that the to/from addresses were set to the address of the context (storage etc.) and not the address of the code. For a callType of "call" this made sense as those two are always the same, but in "delegatecall" it meant that address of the calling/called code were not reported but the contexts of execution, and in fact the trace did not report which code was being delegatecalled to at all.

It was commented in that issue that this was not the expected behaviour and it should report the parameters of the call. A patch was applied in f1eb26dad4900ac136c3c2dd25ab57eb52eb2681 which changed this behaviour (matching that of go-ethereum).

This patch only applied to delegatecall (it added added `match` that only changed behaviour on delegatecall). callcode still reports the contexts rather than the address of the code we are calling. In the case of callcode this means that the to and from addresses are always the same.

https://etherscan.io/vmtrace?txhash=0xb87b345fa76a1bcf44769115058dcc07f3acf96e2f47bd4bea960ad94cf5b433&type=parity#decoded is an example of callcode in a trace (example provided in the #7166 thread).

sorpaas:
Fixes #9878

### DIFF
--- a/ethcore/src/state/mod.rs
+++ b/ethcore/src/state/mod.rs
@@ -1533,7 +1533,7 @@ mod tests {
 	}
 
 	#[test]
-	fn should_not_trace_callcode() {
+	fn should_trace_callcode_properly() {
 		init_log();
 
 		let mut state = get_temp_state();
@@ -1575,7 +1575,7 @@ mod tests {
 			subtraces: 0,
 			action: trace::Action::Call(trace::Call {
 				from: 0xa.into(),
-				to: 0xa.into(),
+				to: 0xb.into(),
 				value: 0.into(),
 				gas: 4096.into(),
 				input: vec![],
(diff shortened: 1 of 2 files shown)


sorpaas:
Just a note on re-syncing: nothing would break if a node does not re-sync, it would just have old tracing data using the old scheme, and new tracing data using the new scheme. So I'd say re-syncing is only required if a node really wants to fix its older data.

PR #9886, opened 2018-11-09
Fix Parity not closing on Ctrl-C

### LINKED ISSUE #9807: Client doesn't shutdown gracefully on Ctrl-C
- **Parity Ethereum version**: master fe84718b5
- **Operating system**: Linux
- **Fully synchronized**: yes, but shouldn't be relevant
- **Network**: kovan, but shouldn't be relevant
- **Restarted**: yes

Works fine with b8da38f4e, so it was introduced with #9657:
```
^C2018-10-24 15:45:38  main WARN parity_ethereum::run  Shutdown is taking longer than expected.
2018-10-24 15:49:38  main WARN parity_ethereum::run  Shutdown timeout reached, exiting uncleanly.
```

ngotchac:
Fixes #9807 

The issue was the references of `Client` in `pm2` (in `PubSubClient::new`) were not being dropped. Using a Weak reference fixes the issue.

### DIFF
--- a/rpc/src/v1/impls/pubsub.rs
+++ b/rpc/src/v1/impls/pubsub.rs
@@ -21,7 +21,7 @@ use std::time::Duration;
 use parking_lot::RwLock;
 
 use jsonrpc_core::{self as core, Result, MetaIoHandler};
-use jsonrpc_core::futures::{Future, Stream, Sink};
+use jsonrpc_core::futures::{future, Future, Stream, Sink};
 use jsonrpc_macros::Trailing;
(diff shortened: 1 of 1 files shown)


tomusdrw:
I think it requires a patch, at least to beta.

Tbaut:
:tada: 
yes only beta needs it.

c0gent:
Nice work. Our high tech solution (replacing each `Arc` with a wrapper that tracks the origin of each reference) was just about to be put to use but you beat us to it.

At least if we ever run into this problem again I've got a solution ready.

PR #9900, opened 2018-11-13
ethcore: use Machine::verify_transaction on parent block

andresilva:
`Machine::verify_transaction` is meant to be used against the parent header. When verifying transactions in a block we would pass the block header instead of the parent, it would be correct for this case since we'd use `header.parent_hash()`, but when the [miner called it with `best_block`](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/miner/pool_client.rs#L122) it wouldn't work properly, e.g. transaction permission contract activation would be off-by-one and the `parent_hash` passed to `TransactionFilter` would be incorrect.

### DIFF
--- a/ethcore/src/verification/verification.rs
+++ b/ethcore/src/verification/verification.rs
@@ -147,7 +147,7 @@ pub fn verify_block_family<C: BlockInfo + CallContract>(header: &Header, parent:
 	verify_uncles(params.block, params.block_provider, engine)?;
 
 	for tx in &params.block.transactions {
-		engine.machine().verify_transaction(tx, header, params.client)?;
+		engine.machine().verify_transaction(tx, parent, params.client)?;
 	}
 
 	Ok(())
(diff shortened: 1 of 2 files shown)


ethcore/src/verification/verification.rs
```diff
@@ -147,7 +147,7 @@ pub fn verify_block_family<C: BlockInfo + CallContract>(header: &Header, parent:
 	verify_uncles(params.block, params.block_provider, engine)?;
 
 	for tx in &params.block.transactions {
-		engine.machine().verify_transaction(tx, header, params.client)?;
+		engine.machine().verify_transaction(tx, parent, params.client)?;
```
tomusdrw: Could you add a comment here why we validate against `parent`? I guess it might be obvious that current state is not yet available, but I think it's best to mention that explicitly.

andresilva:
Done. I also changed the name of the parameter in `verify_transaction` to make it more clear that it should be the parent.

ordian:
Why only backport to beta?

andresilva:
@ordian I think it can be backported to stable as well, but the change to `transaction_allowed` won't apply cleanly I think (should be easy to fix just remove the `header.number() + 1` parameter).

PR #9906, opened 2018-11-13
Remove rust-toolchain file

tomaka:
A reliquary from past times where managing Rust versions was a mess.

This file is a surprise, because even if your defaulting to the nightly version, you will still use stable to compile Parity instead.

CI will make sure that the code keeps compiling with the stable Rust.

The only drawback of removing this file is that people may use old versions of Rust to compile, and open issues.


niklasad1:
```cargo +nightly build``` works fine

IMHO, I prefer having the rust-toolchain file

ordian:
Alternatively, we could set `RUSTUP_TOOLCHAIN` variable (https://github.com/rust-lang-nursery/rustup.rs#override-precedence):
> * An explicit toolchain, e.g. cargo +beta,
> *    The RUSTUP_TOOLCHAIN environment variable,
> *    A directory override, ala rustup override set beta,
> *    The rust-toolchain file,
> *    The default toolchain,
> and they are prefered by rustup in that order

PR #9692, opened 2018-10-03
 aura: finalize blocks

andresilva:
Reopen stale #9113. I couldn't reopen #9113 since I rebased the branch and force-pushed.

- The Aura engine was updated to emit ancestry actions to finalize blocks
- The full client stores block finality in the database, the engine builds finality from an ancestry of `ExtendedHeader`
- `is_epoch_end` was updated to take a vec of recently finalized headers
- `is_epoch_end_light` was added which maintains the previous interface and is used by the light client since the client itself doesn't track finality

PR #9914, opened 2018-11-13
Fix performance issue importing Kovan blocks

ngotchac:
The `block` passed to `self.engine.ancestry_actions` didn't have all the Header's data, thus the `block.header.hash()` was wrong, and the condition `finality_checker.subchain_head() != Some(*chain_head.parent_hash())` in `AuthorityRound::build_finality` would never be `false`, so we would check all ancestry for every imported blocks, until a finalised one was to be found. This was very costly in same cases, eg. when first syncing the chain: on current `master` I would sync at 50 blocks/s, with this PR (as on stable) it goes at about 1_000 blocks/s.

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -798,6 +798,7 @@ impl AuthorityRound {
 				}
 			};
 
+			let epoch_transition_hash = epoch_manager.epoch_transition_hash;
 			let ancestry_iter = ancestry.map(|header| {
 				let mut signers = vec![*header.author()];
 				signers.extend(parent_empty_steps_signers.drain(..));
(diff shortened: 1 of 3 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -815,10 +816,11 @@ impl AuthorityRound {
 					None
 				}
 			})
-				.while_some();
+				.while_some()
+				.take_while(|&(h, _)| h != epoch_transition_hash);
```
tomusdrw: So that limits the iterator, can you elaborate a little bit why it's valid to do so?

andresilva: I discussed this with @ngotchac. Aura only used to track finality for safely triggering validator set changes, therefore we were only interested in finalizing blocks within the current epoch, so this `take_while` was the original behavior. Recently we added finality tracking to the client so we updated Aura to track finality until the last finalized block. On chains where there isn't a majority of validators online the unfinalized subchain just keeps growing and this can take some time to build (although if the subchain is properly cached this cost is avoided).

ordian:
#9692 is in beta, so should we backport this PR to beta as well?

andresilva:
@ordian yes we should 👍

PR #9918, opened 2018-11-14
simplify cargo audit

TriplEight:
`cargo audit` won't be using cache and will be installed by any available toolchain.


TriplEight:
[clabot:check]

PR #9919, opened 2018-11-14
add --force, it's installed somewhere

TriplEight:
it appeared that `cargo audit` was installed previously, that failed the beta and stable builds:
```
cargo-audit v0.5.2
error: binary `cargo-audit` already exists in destination as part of `cargo-audit v0.5.2`
Add --force to overwrite
```


ordian:
I think we don't need it on master, since we've removed caches in #9855, so we either backport this or that PR.

bkchr:
I support @ordian opinion on this topic.

PR #9920, opened 2018-11-14
parity-clib: `async C bindings to RPC requests` + `subscribe/unsubscribe to websocket events`

### LINKED ISSUE #9271: Add a method in the C bindings to replace the websocket API
If you want to embed Parity in a binary, there should be some interface in the bindings in order to use the websocket capabilities of Parity without having to open a dummy TCP connection.

niklasad1:
Attempt to close #9271 

This PR changes to following in parity-clib (C and Java bindings):
- asynchronous RPC queries (allocates a background thread that runs for at most X seconds)
- subscribe to WebSocket events (runs in a background thread until it is cancled)
- unsubscribe from WebSocket events

Additional changes:
- modify `party-cpp-example`
- add Java example application

TODOs:
- [x] Update Java bindings (Parity.java) and test it with a Java application
- [x] Change callback structure

Pitfalls/suggested improvements, should be fixed in this PR?!
- [ ] Dropping a GlobalRef in a detached thread (JNI related https://docs.rs/jni/0.10.2/jni/objects/struct.GlobalRef.html)
- [ ] Use a shared thread for WebSocket subscriptions (really hard without major refactoring because the parity-clib doesn't keep state)
- [ ] Use of deprecated `finalize()` in Parity.Java (I have looked into it very briefly and the way to go is to use PhantomReference instead but requires more boilerplate code and I like currently how simple the Parity class is)


logger/src/lib.rs
```diff
@@ -71,6 +71,7 @@ pub fn setup_log(config: &Config) -> Result<Arc<RotatingLogger>, String> {
 	builder.filter(Some("ws"), LevelFilter::Warn);
 	builder.filter(Some("hyper"), LevelFilter::Warn);
 	builder.filter(Some("rustls"), LevelFilter::Error);
+	builder.filter(Some("rpc"), LevelFilter::Trace);
```
niklasad1: Used for testing because the log can't be configured from the `C lib`. Should be removed before merging

parity-clib/src/lib.rs
```diff
@@ -136,31 +159,48 @@ pub unsafe extern fn parity_rpc(client: *mut c_void, query: *const c_char, len:
...
 
-			ptr::copy_nonoverlapping(output.as_bytes().as_ptr(), out_str as *mut u8, q_out_len);
-			*out_len = q_out_len;
-			0
-		} else {
-			1
-		}
+		let cb = callback.clone();
+
+		// FIXME: provide session object here, if we want to support the `PubSub`
```
tomaka: Well, the whole point of #9271 is to implement this. It's also the most difficult part, and why I gave up doing this myself.

tomaka: > everytime we get a notification we should send it to the callback?

Yes!

logger/src/lib.rs
niklasad1: Will keep it, if people want to test it helps to see what happens :P

niklasad1:
The PR should be in `reviewable` stable state now seem to work both for the light and full client.

The easiest way if you want to test it, grab the [update cpp example](https://github.com/paritytech/parity-ethereum/tree/parity-clib/jsonrpc-callback/parity-clib-examples/cpp) and then make sure that you are fully-synced otherwise `ChainNotify` doesn't notify the `PubSub Client`

parity-clib/parity.h
```diff
@@ -86,21 +88,52 @@ int parity_start(const ParityParams* params, void** out);
...
-/// a new thread for each RPC request that requires accessing the blockchain.
+/// - On success	: The parity client reference and the query string were valid
+/// - On error		: The parity client reference and the query string were not valid
 ///
-/// - `rpc` and `len` must contain the JSON string representing the RPC request.
-/// - `out_str` and `out_len` point to a buffer where the output JSON result will be stored. If the
-///	  buffer is not large enough, the function fails.
-/// - `out_len` will receive the final length of the string.
-/// - On success, the function returns 0. On failure, it returns 1.
+int parity_rpc(const void *const parity, const char* rpc_query, size_t rpc_len, size_t timeout_ms, callback response);
```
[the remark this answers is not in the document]
niklasad1: Are we talking about distinguishing between `ordinary RPC queries` and `websocket subscriptions in this context? If so, then I would expect people to use to different callbacks but sure I can add it anyway probably

When it comes to actual WebSocket responses I just use `a dummy regex` to get the subscription id which must be used to disable the subscription! I expect people to use some JSON parser in that case

parity-clib/parity.h
```diff
@@ -86,21 +88,52 @@ int parity_start(const ParityParams* params, void** out);
...
+
+
+/// Subscribes to the specified websocket event
+///
+///	 - parity		: Reference to the running parity client
+///	 - ws_query		: JSON encoded string representing the websocket and which event to subscribe to
```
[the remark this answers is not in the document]
niklasad1: Will rephrase

parity-clib-examples/cpp/main.cpp
```diff
@@ -20,38 +20,179 @@
...
+
+// callback that gets invoked when the client restarts
 void on_restart(void*, const char*, size_t) {}
 
+// callback that is invoked on ws responses
+void ws_response(void* _unused, const char* response, size_t len) {
+	printf("ws_response: %s\r\n", response);
+	std::regex is_subscription ("\\{\"jsonrpc\":\"2.0\",\"result\":\"0[xX][a-fA-F0-9]{16}\",\"id\":1\\}");
+	if (std::regex_match(response, is_subscription) == true) {
+		strncpy(g_str, response, 55);
```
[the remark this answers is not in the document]
niklasad1: I think it follows `our editor config` which is tab indention of width 4. Looks "ok" to me in my editor.

parity-clib/parity.h
```diff
@@ -86,21 +88,52 @@ int parity_start(const ParityParams* params, void** out);
...
+///	 - ws_query		: JSON encoded string representing the websocket event to unsubscribe from
+///	 - len			: Length of the query
+///	 - timeout		: Maximum time in milliseconds to wait for a response
+///	 - response		: Callback to invoke when a websocket event is received
 ///
-int parity_rpc(void* parity, const char* rpc, size_t len, char* out_str, size_t* out_len);
+///	 - On success	: The function return 0
+///	 - On error		: The function returns non-zero
+//
+int parity_unsubscribe_ws(const void *const parity, const void *const session, const char* ws_query,
```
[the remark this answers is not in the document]
niklasad1: hrm, the `PubSubSession API` doesn't allows us to cancel subscription because these are private, https://github.com/paritytech/jsonrpc/blob/master/pubsub/src/subscription.rs#L32-#L73.

Thus, we need to use parity client to so. @tomusdrw can you confirm or suggest how do it?

parity-clib/parity.h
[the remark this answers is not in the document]
tomaka: > to use to different callbacks but sure I can add it anyway probably

Well, you can't really do that correctly in C. The "callback-with-a-void*-parameter" pattern is very common in C.

parity-clib/parity.h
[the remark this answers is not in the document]
tomusdrw: @niklasad1 It's enough to just drop the `Session` and it will unsubscribe automagically (of course if you want to unsubscribe from all subscriptions and you don't need to wait for that to finish).

parity-clib/parity.h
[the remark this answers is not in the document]
niklasad1: hmm, I don't understand your comment, you should be able this this in C right?

```c
static void f1(void) {
    printf("f1\r\n");
}

static void f2(void) {
    printf("f2\r\n");
}

void register_callback(void (*callback)(void))
{
    printf("inside register_callback");
    (callback)();
}

int main(void) {
    register_callback(f1);
    register_callback(f2);
}
```

Or what do you mean? :P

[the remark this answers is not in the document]
tomaka: Imagine you want to query the block header of a certain block number.
In the callback, you parse the JSON response and you obtain the header and number. But then you probably want to do with that information, but since you don't have any context you have no other way than to use global variables. This is not great.

Instead, if you have this `void*` parameter you can pass any information required for the callback to do what it needs to do.

niklasad1:
@tomaka 

I have started to look at the Java bindings and the synchronous part works fine. However, the asynchronous part is a different story in order to invoke a java method I have to do by using the JNI/JVM environment AFAIU

Something like 
```rust
    env.call_method(
        callback, 
        "callback", 
        "(JLjava/lang/Object;Ljava/lang/Object;J)V", 
        &[JValue::Long(0), JValue::Object(null), JValue::Object(null), JValue::Long(0)]
  ).expect("Java method call failed");
```
Thus, using the `C API` "out-of-the-box" won't work. So as I see it I can either:

1. Introduce `CCallback` and `JCallback` in the the C-API
2. Re-implement the functionality in the `java module`

Thoughts?

parity-clib-examples/java/Main.java
```diff
@@ -0,0 +1,91 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import java.util.Vector;
+import java.util.concurrent.atomic.AtomicInteger;
+import io.parity.ethereum.Parity;
+import com.sun.jna.Pointer;
```
niklasad1: You need `jna-5.1.0.jar` in your CLASSPATH in order to compile this!

parity-clib/Parity.java
```diff
@@ -28,36 +28,59 @@
...
+    /** Unsubscribes to a specific WebSocket event
      *
-     * @param query The JSON-encoded RPC query to perform
-     * @return A JSON-encoded result
+     * @param session	Pointer the the session to terminate
      */
-    public String rpcQuery(String query) {
-        return rpcQueryNative(inner, query);
-    }
+	public void unsubscribeWebSocket(long session) {
```
niklasad1: This could be a static method

logger/src/lib.rs
sorpaas: @niklasad1 But this will propagate to release build, wouldn't it? I think we still need to remove this before merging, right?

niklasad1: Yeah, it will be removed :smile:

parity-clib/Parity.java
```diff
@@ -28,36 +28,59 @@
...
+
+    /** Subscribes to a specific WebSocket event which will run in a background thread until it is canceled.
+     *
+     * @param query     The JSON-encoded RPC query to perform
+     * @param callback  An instance of class which must have a instance method named `callback` that will be invoked
+     *                  when the result is ready
+     *
+    * @return A pointer to the current sessions which can be used to terminate the session later
+    */
+	public long subscribeWebSocket(String query, Object callback) {
```
tomaka: Indentation mismatch in this file.

parity-clib/parity.h
```diff
@@ -17,6 +17,11 @@
 #ifndef _PARITY_H_INCLUDED_
 #define _PARITY_H_INCLUDED_
 
+#define		PARITY_CALLBACK_RESTART			0
```
tomaka: I think you misunderstood my concern. We don't need this. All we need is an extra parameter that the user can pass to `parity_rpc` and `parity_subscribe_ws` and that is transparently passed to the callback.

niklasad1: Ok, one more dumb question from my side then!

How do you propose to safely cast the ```void *``` to some object then? 
Or are we assuming that the user will always pass the same type as `user_data` such as an user-defined struct?

tomaka: It is the user's job to interpret this `void*` the way they want, we only guarantee to pass back the same value. But yeah normally it is pointing to a user-defined structure.

DemiMarie:
Can we use asynchronous I/O in Java and C?  Ideally, it should be agnostic regarding the actual event loop used.

niklasad1:
@DemiMarie Yes, we can. It is agnostic for the actual event loop it just that invoking the actual _callback_ that is different. Because, we need some special `types` and `methods` from the jni crate I decided to duplicate that code in order to avoid having `jni` as mandatory dependency for `parity-clib`.

niklasad1:
@tomaka do you have time review this again?

parity/run.rs
```diff
@@ -851,20 +852,21 @@ enum RunningClientInner {
 }
 
 impl RunningClient {
-	/// Performs a synchronous RPC query.
-	/// Blocks execution until the result is ready.
-	pub fn rpc_query_sync(&self, request: &str) -> Option<String> {
+	/// Performs an asynchronous RPC query.
+	pub fn rpc_query(&self, request: &str, session: Option<Arc<PubSubSession>>)
```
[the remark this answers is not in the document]
niklasad1: Yeah, let us create an issue after this is merged then?

parity-clib/Parity.java
```diff
@@ -20,44 +20,67 @@
...
+
+	/** Unsubscribes to a specific WebSocket event
+	 *
+	 * @param session	Pointer the the session to terminate
+	 */
+	public void unsubscribeWebSocket(long session) {
+		unsubscribeWebSocketNative(session);
+	}
+
+	// FIXME: `finalize` is deprecated
```
niklasad1: Create a separate easy issue for this!

parity-clib/src/java.rs
```diff
@@ -0,0 +1,211 @@
...
+				.name("rpc_query".to_string())
+				.spawn(move || {
+					let mut current_thread = CurrentThread::new();
+					current_thread.spawn(future);
+					let _ = current_thread.run_timeout(Duration::from_millis(timeout_ms as u64))
+						.map_err(|_e| {
+							cb.call(error::TIMEOUT);
+						});
+				})
+			.expect("rpc-query thread shouldn't fail; qed");
```
tomaka: Nit: indentation
```suggestion
				.expect("rpc-query thread shouldn't fail; qed");
```

5chdn:
@niklasad1 could you rebase it?

@tomusdrw @sorpaas could you give a 2nd review here?

parity-clib/src/java.rs
```diff
@@ -0,0 +1,211 @@
...
+		let java_str = env.new_string(msg.to_string()).expect("Rust String is valid JString; qed");
+		let val = &[JValue::Object(JObject::from(java_str))];
+		env.call_method(self.callback.as_obj(), self.method_name, self.method_descriptor, val).expect(
+			"The callback must be an instance method and be named \"void callback(Object)\"; qed)");
+	}
+}
+
+#[no_mangle]
+pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_configFromCli(env: JNIEnv, _: JClass, cli: jobjectArray) -> jlong {
+	let cli_len = env.get_array_length(cli).expect("invalid Java bindings");
```
tomusdrw: maybe cast to `usize` her instead of repeating it multiple times?

parity-clib/src/java.rs
```diff
@@ -0,0 +1,211 @@
...
+		let env = self.jvm.attach_current_thread().expect("JavaVM should have an environment; qed");
+		let java_str = env.new_string(msg.to_string()).expect("Rust String is valid JString; qed");
+		let val = &[JValue::Object(JObject::from(java_str))];
+		env.call_method(self.callback.as_obj(), self.method_name, self.method_descriptor, val).expect(
+			"The callback must be an instance method and be named \"void callback(Object)\"; qed)");
+	}
+}
+
+#[no_mangle]
+pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_configFromCli(env: JNIEnv, _: JClass, cli: jobjectArray) -> jlong {
```
tomusdrw: Can we extract safe parts to some other method? Having such a big function with `unsafe` seems worrying to me.

parity-clib/src/java.rs
```diff
@@ -0,0 +1,211 @@
...
+	let _ = async_checker(parity, rpc, callback, &env)
+		.map(|(client, query, jvm, global_ref)| {
+			let callback = Arc::new(Callback::new(jvm, global_ref));
+			let cb = callback.clone();
+			let future = client.rpc_query(&query, None).map(move |response| {
+				let response = response.unwrap_or_else(|| error::EMPTY.to_string());
+				callback.call(&response);
+			});
+
+			let _handle = thread::Builder::new()
```
tomusdrw: We spawn a thread per call? Can't we have a single thread for all timeouts?

parity-clib/src/java.rs
```diff
@@ -0,0 +1,211 @@
...
+			let (tx, mut rx) = mpsc::channel(1);
+			let session = Arc::new(PubSubSession::new(tx));
+			let weak_session = Arc::downgrade(&session);
+			let query_future = client.rpc_query(&query, Some(session.clone()));;
+
+			let _handle = thread::Builder::new()
+				.name("ws-subscriber".into())
+				.spawn(move || {
+					// Wait for subscription ID
+					// Note this may block forever and be can't destroyed using the session object
```
tomusdrw: `s/be can't/can't be`

parity-clib/src/lib.rs
```diff
@@ -152,116 +196,86 @@ pub unsafe extern fn parity_rpc(client: *mut c_void, query: *const c_char, len:
...
-// Internal structure for handling callbacks that get passed a string.
-struct CallbackStr(Option<extern "C" fn(*mut c_void, *const c_char, usize)>, *mut c_void);
-unsafe impl Send for CallbackStr {}
-unsafe impl Sync for CallbackStr {}
-impl CallbackStr {
-	fn call(&self, new_chain: &str) {
-		if let Some(ref cb) = self.0 {
-			cb(self.1, new_chain.as_bytes().as_ptr() as *const _, new_chain.len())
+	panic::catch_unwind(|| {
+		if let Some((client, query)) = parity_rpc_query_checker(client, query, len) {
```
tomusdrw: Seems pretty similar to the previous file, can we share some logic (hopefuly doesn't need to be `unsafe`)

niklasad1: Yeah, that is true. The only thing that stopped me from sharing this is how the callbacks are invoked and I didn't want to have a dependency to `jni` in the parity-clib some special types are required. However, I will double check if I can come up with a better solution.

5chdn:
Can't merge, CI issue is unrelated though

niklasad1:
@5chdn that's was good in the case because I want to address @tomusdrw grumbles :P

tomusdrw:
@niklasad1 I'm totally fine with a separate PR to improve the code. It probably works now, we can make it more performant as the next step. Please create an issue if you'd like to split the work this way.

niklasad1:
Ok, let's merge this and then I clean it up in another PR

niklasad1:
@5chdn The parity-clib/example was moved so the source path didn't work. I will fix this soon

PR #9701, opened 2018-10-04
EIP 191

seunlanlege:
implemented EIP-191 signing (`personal_sign191`)

PR #9001, opened 2018-06-28
EIP-1186: add `eth_getProof` RPC-Method

simon-jentzsch:
Implementation for the [EIP-1186](https://github.com/ethereum/EIPs/issues/1186)

Ethereum uses MerkleTrees to store the state of accounts and their storage. This allows verification of each value by simply creating a MerkleProof. But currently the eth-Module does not give you access to these proofs. This EIP suggests a additional RPC-Method, which creates MerkleProofs for Accounts and Storage-Values. 

Combined with a stateRoot (from the blockheader) it enables offline verification of any account or storage-value. This allows especially IOT-Devices or even mobile apps which are not able to run a light client to verify responses from a untrusted source.

PR #9928, opened 2018-11-15
Add experimental RPCs flag

tomusdrw:
Recently we have introduced two RPCs in `personal_` namespace (#9631, #9701), the EIPs for those are still in draft, and the final RPC name is not decided yet. Another one is coming soon: #9001 and this one is in a "more standardized" namespace.

For that reason I'm in favour of explicitly marking such RPCs as "experimental" and also require opt-in to enable them.


5chdn:
Adding to release notes

PR #9932, opened 2018-11-16
Fix light client informant while syncing

ngotchac:
From #9385 , the informant wasn't displaying correctly the syncing state. This is because the `state` of `LightSync` had deadlock issues, and the default to `is_major_importing` was set to false if the lock couldn't be acquired. This lead to false logs in the informant.

In this PR, `is_idle` is added, and gets updated everytime the `state` is updated, such as `is_major_importing` can check this boolean instead of waiting for `state` to be unlocked.

### DIFF
--- a/ethcore/sync/src/light_sync/mod.rs
+++ b/ethcore/sync/src/light_sync/mod.rs
@@ -236,6 +236,7 @@ pub struct LightSync<L: AsLightClient> {
 	client: Arc<L>,
 	rng: Mutex<OsRng>,
 	state: Mutex<SyncState>,
+	is_idle: Mutex<bool>,
 }
 
 #[derive(Debug, Clone)]
@@ -309,16 +310,17 @@ impl<L: AsLightClient + Send + Sync> Handler for LightSync<L> {
 
 		if new_best.is_none() {
 			debug!(target: "sync", "No peers remain. Reverting to idle");
-			*self.state.lock() = SyncState::Idle;
+			self.set_state(&mut self.state.lock(), SyncState::Idle);
 		} else {
 			let mut state = self.state.lock();
 
-			*state = match mem::replace(&mut *state, SyncState::Idle) {
+			let next_state = match mem::replace(&mut *state, SyncState::Idle) {
 				SyncState::Idle => SyncState::Idle,
 				SyncState::AncestorSearch(search) =>
 					SyncState::AncestorSearch(search.requests_abandoned(unfulfilled)),
 				SyncState::Rounds(round) => SyncState::Rounds(round.requests_abandoned(unfulfilled)),
 			};
+			self.set_state(&mut state, next_state);
 		}
 
 		self.maintain_sync(ctx.as_basic());
@@ -390,12 +392,13 @@ impl<L: AsLightClient + Send + Sync> Handler for LightSync<L> {
 				data: headers,
 			};
 
-			*state = match mem::replace(&mut *state, SyncState::Idle) {
+			let next_state = match mem::replace(&mut *state, SyncState::Idle) {
 				SyncState::Idle => SyncState::Idle,
 				SyncState::AncestorSearch(search) =>
 					SyncState::AncestorSearch(search.process_response(&ctx, &*self.client)),
 				SyncState::Rounds(round) => SyncState::Rounds(round.process_response(&ctx)),
 			};
+			self.set_state(&mut state, next_state);
 		}
 
 		self.maintain_sync(ctx.as_basic());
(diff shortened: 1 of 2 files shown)


ethcore/sync/src/light_sync/mod.rs
```diff
@@ -235,7 +273,8 @@ pub struct LightSync<L: AsLightClient> {
 	pending_reqs: Mutex<HashMap<ReqId, PendingReq>>, // requests from this handler
 	client: Arc<L>,
 	rng: Mutex<OsRng>,
-	state: Mutex<SyncState>,
+	state: Mutex<SyncStateWrapper>,
+	is_idle: Mutex<bool>,
```
ordian: Maybe it's worth adding a comment, like that:
```suggestion
	// We duplicate this state tracking to avoid deadlocks in `is_major_importing`.
	is_idle: Mutex<bool>,
```
Also I was thinking of using an `std::sync::atomic::AtomicBool` to make sure the deadlock doesn't happen, but maybe it's just me being paranoid.

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -213,6 +214,43 @@ enum SyncState {
 	Rounds(SyncRound),
 }
 
+/// A wrapper around the SyncState that makes sure to
+/// update the giving reference to `is_idle`
+struct SyncStateWrapper {
```
ordian:
```suggestion
#[derive(Debug)]
struct SyncStateWrapper {
```

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -213,6 +214,43 @@ enum SyncState {
...
+	pub fn set(&mut self, state: SyncState, is_idle_handle: &mut bool) {
+		*is_idle_handle = match state {
+			SyncState::Idle => true,
+			_ => false,
+		};
+		self.state = state;
+	}
+
+	/// Returns the internal state's value
+	pub fn get(self) -> SyncState {
```
ordian: `get` usually takes a reference, for [Cell](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.into_inner) this is called `into_inner`
```suggestion
	pub fn into_inner(self) -> SyncState {
```

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -681,11 +709,13 @@ impl<L: AsLightClient> SyncInfo for LightSync<L> {
 	}
 
 	fn is_major_importing(&self) -> bool {
-		self.is_major_importing_do_wait(true)
-	}
+		const EMPTY_QUEUE: usize = 3;
+
+		let queue_info = self.client.as_light_client().queue_info();
+		let is_verifying = queue_info.unverified_queue_size + queue_info.verified_queue_size > EMPTY_QUEUE;
```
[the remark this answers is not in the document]
ngotchac: Not sure about that. But from my local testing, it lead to wrong information being displayed (`Import` logs too soon)

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -213,6 +214,44 @@ enum SyncState {
 	Rounds(SyncRound),
 }
 
+/// A wrapper around the SyncState that makes sure to
+/// update the giving reference to `is_idle`
+#[derive(Debug)]
+struct SyncStateWrapper {
```
cheme: I am not really sure about the usefulness of this struct, it seems that `set` function is only called once in `set_state` of LightClient, since fields are private. 
Ok, I got it, it avoid bad use in future client inner implementation, maybe then we could move `is_idle` to it:
```rust
struct SyncStateWrapper {
  state: Mutex::new(SyncState::Idle),
  is_idle: Mutex::new(true),
}
```
and keep deref to `state`. Then we would have an easier `set` function (without handle ref).
This could be a bad idea if we need to keep is_idle mutex open (does not seems so).

ngotchac: The thing is that you want to keep an open `SyncState` `Mutex` while `is_idle` still being accessible. With the suggested changes, you cannot have a lock on `SyncStateWrapper` (otherwise `is_idle` wouldn't be accessible), so you would have to expose the inner `SyncState` lock, which could be used to directly change the state, without updating `is_idle`. Does that make sense?

cheme: Make sense, my suggestion was your second hypothesis (no mutex around SyncStateWrapper) , it does need to exposes the inner lock, on the other hand it keeps both locks together.
There is maybe less chance to mistake with current version, a mistake would be to pass any `&mut bool` to `set_state` which would be clearly intentional.

PR #9934, opened 2018-11-17
Unbreak parity-rpc build on rust -stable

reyk:
```
   Compiling parity-rpc v1.12.0 (file:///home/reyk/devel/parity-ethereum/rpc)
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
   --> rpc/src/v1/helpers/light_fetch.rs:628:19
    |
628 |                         if let Some(vm::Error::OutOfGas) = executed.exception {
    |                                     ^^

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `parity-rpc`.

```

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -625,7 +625,7 @@ fn execute_read_only_tx(gas_known: bool, params: ExecuteParams) -> impl Future<I
 				match res {
 					Ok(executed) => {
 						// `OutOfGas` exception, try double the gas
-						if let Some(vm::Error::OutOfGas) = executed.exception {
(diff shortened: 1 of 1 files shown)


reyk:
[clabot:check]

niklasad1:
Hey, thanks but it works on Rust stable 1.30. Please update your toolchain instead

reyk:
It's a simple fix that works on both versions, why speaks against merging it?

I'm using OpenBSD 6.4 that ships rust 1.29.2 - OpenBSD -current includes 1.30.x or later but the next OpenBSD -stable release (6.5) will only be released in about 6 months.

reyk:
Thanks!

PR #9938, opened 2018-11-19
Fill transaction hash on ethGetLog of light client.

### LINKED ISSUE #9929: Light node response to eth_getLogs missing transactionHash

- **Parity Ethereum version**: v2.2.1
- **Operating system**: Linux
- **Installation**: Official docker image
- **Fully synchronized**: What's that even mean for a warping light node? 😜
- **Network**: ethereum
- **Restarted**: yes

Issuing a getLogs request to a parity node that is in --light:
```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[ {"fromBlock":"0x5a6da7","toBlock":"0x5a6da7","address":"0x75228dce4d82566d93068a8d5d49435216551599"} ],"id":74}' localhost:8545 | jq .result[0]

{
  "address": "0x75228dce4d82566d93068a8d5d49435216551599",
  "blockHash": "0x4ff6030cce082be8f5a56d7f7bbe5e32e8d98295f31c8e070285fbdd022a06a7",
  "blockNumber": "0x5a6da7",
  "data": "0x000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "logIndex": "0x9",
  "removed": false,
  "topics": [
    "0x299eaafd0d27519eda3fe7195b73e5269e442b3d80928f19afa32b6db2f352b6",
    "0x0000000000000000000000000000000000000000000000000000000000000000",
    "0x000000000000000000000000e991247b78f937d7b69cfc00f1a487a293557677"
  ],
  "transactionHash": null,
  "transactionIndex": "0x2a",
  "transactionLogIndex": "0x0",
  "type": "mined"
}
```

The data is correct and complete, but 
```
  "transactionHash": null,
```

That response, from a parity full node or geth node (light or full) is:
```
$  curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[ {"fromBlock":"0x5a6da7","toBlock":"0x5a6da7","address":"0x75228dce4d82566d93068a8d5d49435216551599"} ],"id":74}' my-geth-server | jq .result[0]
{
  "address": "0x75228dce4d82566d93068a8d5d49435216551599",
  "topics": [
    "0x299eaafd0d27519eda3fe7195b73e5269e442b3d80928f19afa32b6db2f352b6",
    "0x0000000000000000000000000000000000000000000000000000000000000000",
    "0x000000000000000000000000e991247b78f937d7b69cfc00f1a487a293557677"
  ],
  "data": "0x000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
  "blockNumber": "0x5a6da7",
  "transactionHash": "0x44c09f8eeff886723b79890e14743192a8c8d8a8eac158ed17600c94e502cce8",
  "transactionIndex": "0x2a",
  "blockHash": "0x4ff6030cce082be8f5a56d7f7bbe5e32e8d98295f31c8e070285fbdd022a06a7",
  "logIndex": "0x9",
  "removed": false
}
```

cheme:
This is highly inefficient, yet required to respect standard eth rpc.
Closes #9929 .

It exposes a new rpc method to keep old behavior  : `parity_getLogsNoTransactionHash`


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
 							}
 							future::ok::<_,OnDemandError>(matches)
 						}) // and then collect them into a vector.
-						.map(|matches| matches.into_iter().map(|(_, v)| v).collect())
 						.map_err(errors::on_demand_error)
+						// retrieve transaction hash.
+						.and_then(move |matches| {
+							let mut blocks = BTreeMap::new();
+							let mut result: Vec<Log> = matches.into_iter().map(|(_, v)| {
+								{
```
niklasad1: indentation is off here

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
...
+								blocks.entry(block_hash.clone()).or_insert_with(|| {
+									fetcher_block.block(BlockId::Hash(block_hash.clone().into()))
+								});
+								}
+								v
+							}).collect();
+							// future get blocks (unordered it)
+							stream::futures_unordered(blocks.into_iter().map(|(_,v)|v)).collect().map(move |blocks| {
+								let mut tr_per_bl = BTreeMap::new();
+								for enc_bl in blocks.iter() {
```
niklasad1: I don't understand the name `enc_block` can you rephrase?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
...
+								let block_hash = v.block_hash.as_ref().expect("Previously initialized with value; qed");
+								blocks.entry(block_hash.clone()).or_insert_with(|| {
+									fetcher_block.block(BlockId::Hash(block_hash.clone().into()))
+								});
+								}
+								v
+							}).collect();
+							// future get blocks (unordered it)
+							stream::futures_unordered(blocks.into_iter().map(|(_,v)|v)).collect().map(move |blocks| {
+								let mut tr_per_bl = BTreeMap::new();
```
niklasad1: `tree_per_block`? Seems to store all blocks as <Hash, Transactions>

And why is this not a HashMap? BTreeMap performs linear search O(n)?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
...
+									let tr = enc_bl.transactions();
+									tr_per_bl.insert(bl_hash, tr);
+								}
+								for log in result.iter_mut() {
+									let log_index: U256 = log.transaction_index.expect("Previously initialized with value; qed").into();
+									if log_index < usize::max_value().into() {
+										let block_hash = log.block_hash.clone().expect("Previously initialized with value; qed").into();
+										let btr = tr_per_bl.get(&block_hash)
+											.and_then(|arr_tr| arr_tr.get(log_index.as_usize()))
+											.map(|tr| tr.hash().into());
```
niklasad1: I think the acronym tx is better than tr!

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
...
+									let bl_hash = enc_bl.hash();
+									let tr = enc_bl.transactions();
+									tr_per_bl.insert(bl_hash, tr);
+								}
+								for log in result.iter_mut() {
+									let log_index: U256 = log.transaction_index.expect("Previously initialized with value; qed").into();
+									if log_index < usize::max_value().into() {
+										let block_hash = log.block_hash.clone().expect("Previously initialized with value; qed").into();
+										let btr = tr_per_bl.get(&block_hash)
+											.and_then(|arr_tr| arr_tr.get(log_index.as_usize()))
```
niklasad1: txs or transactions IMHO

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,42 @@ impl LightFetch {
...
+								for enc_bl in blocks.iter() {
+									let bl_hash = enc_bl.hash();
+									let tr = enc_bl.transactions();
+									tr_per_bl.insert(bl_hash, tr);
+								}
+								for log in result.iter_mut() {
+									let log_index: U256 = log.transaction_index.expect("Previously initialized with value; qed").into();
+									if log_index < usize::max_value().into() {
+										let block_hash = log.block_hash.clone().expect("Previously initialized with value; qed").into();
+										let btr = tr_per_bl.get(&block_hash)
```
niklasad1: tx_hash

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -367,8 +368,38 @@ impl LightFetch {
...
 						}) // and then collect them into a vector.
-						.map(|matches| matches.into_iter().map(|(_, v)| v).collect())
 						.map_err(errors::on_demand_error)
+						// retrieve transaction hash.
+						.and_then(move |matches| {
+							let mut blocks = BTreeMap::new();
+							let mut result: Vec<Log> = matches.into_iter().map(|(_, v)| {
+								{
+									let block_hash = v.block_hash.as_ref().expect("Previously initialized with value; qed");
+									blocks.entry(block_hash.clone()).or_insert_with(|| {
```
sorpaas: `block_hash` is copy, so we don't need `clone`.

niklasad1: Actually, I think this is `rpc::v1::types::hash` which is not `Copy` but we should refactor that to be consistent in the codebase!

niklasad1:
Yeah, could be good to provide the old behavior too but I don't think a flag for `eth namespace` is a good idea instead I would prefer to provide the old behavior in the `parity namespace` or somewhere else instead! Thoughts?

cheme:
At first I wanted to do it, but then I got lazy :)  The truth is that I started wondering if it would not be an additional unused rpc or flag.
You are right a flag is not a good idea.
I am more in favor of a undocumented parity ns rpc. Undocumented means not in wiki because it is very specific use.
So before putting it in place I jut want to check again that it will be useful (depending on use case blocks may be download and put in cache anyway due to other calls). /cc @Tbaut @amaurymartiny

Tbaut:
My 2cts 
>I am more in favor of a undocumented parity ns rpc. Undocumented means not in wiki because it is very specific use.

If it's available, it should be documented, no matter if it's used or not by the outside world, if we developed it (for Fether) it means it can be valuable for others, be it transparently in `light.js`. I think it's nice to keep the original one in a `parity_` rpc, but as any other RPC, to document it (it's part of the CI now anyways).

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -310,8 +310,8 @@ impl LightFetch {
 		}))
 	}
 
-	/// Get transaction logs
-	pub fn logs(&self, filter: EthcoreFilter) -> impl Future<Item = Vec<Log>, Error = Error> + Send {
+  /// Variant of get transaction logs that does not fetch log transactions hash
```
ordian: spaces? :)
```suggestion
	/// Variant of get transaction logs that does not fetch log transactions hash
```

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -378,6 +378,105 @@ impl LightFetch {
...
+										});
+									}
+									block_index += 1;
+								}
+							}
+							future::ok::<_,OnDemandError>(matches)
+						}) // and then collect them into a vector.
+						.map_err(errors::on_demand_error)
+						// retrieve transaction hash.
+						.and_then(move |matches| {
```
ordian: Is it possible to extract the code above into a function and reuse it in `logs_light`, so that if we would want to change it, we could do that in one place?

cheme:
So the method with old behavior can be call by switching : 
```
curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[ {"fromBlock":"0x5a6da7","toBlock":"0x5a6da7"} ],"id":74}' localhost:8545
```
to 
```
curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"parity_getLogsLight","params":[ {"fromBlock":"0x5a6da7","toBlock":"0x5a6da7"} ],"id":74}' localhost:8545
```
(new name is 'parity_getLogsLight')

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -376,6 +376,52 @@ impl LightFetch {
...
+				for log in result.iter_mut() {
+					let log_index: U256 = log.transaction_index.expect("Previously initialized with value; qed").into();
+					if log_index < usize::max_value().into() {
+						let block_hash = log.block_hash.clone().expect("Previously initialized with value; qed").into();
+						let tx_hash = transactions_per_block.get(&block_hash)
+							.and_then(|txs| txs.get(log_index.as_usize()))
+							.map(|tr| tr.hash().into());
+						log.transaction_hash = tx_hash;
+					} else {
+						trace!(target: "light_fetch", "A received Receipts indexed other usize length ignored");
```
ordian: I think that's unreachable, since `iter.enumerate()` always returns `(_: usize, _)`.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -308,11 +309,11 @@ impl LightFetch {
...
 			_ => panic!(WRONG_RESPONSE_AMOUNT_TYPE_PROOF),
 		}))
+
+	
 	}
 
-	/// Get transaction logs
-	pub fn logs(&self, filter: EthcoreFilter) -> impl Future<Item = Vec<Log>, Error = Error> + Send {
-		use std::collections::BTreeMap;
+	fn logs_common(&self, filter: EthcoreFilter) -> impl Future<Item = BTreeMap<(u64,usize),Log>, Error = Error> + Send {
```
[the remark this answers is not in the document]
cheme: There is a slight difference between both logs method, the collect from the second one is different, so I probably did it to avoid a vec iteration (which may be optimize anyway, I do not know).

rpc/src/v1/helpers/light_fetch.rs
cheme: :+1: I did not realize (I did see U256 so I did test)

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
niklasad1: Hrm, do we use a `BTreeMap` because we want the logs sorted by (block_num, block_index)? Then it makes sense to me at least!

Still both `fn logs` and `fn logs_light` collect the values of BTreeMap into a Vec so I would guess it would end up with the same number allocations as before (don't tx hashes retrieval)?

So why not return a Vec<Log> instead or am I missing something?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -376,6 +376,49 @@ impl LightFetch {
 					None => Either::B(Either::B(future::err(errors::network_disabled()))),
 				}
 			})
+
+	}
+
```
niklasad1: extra whitespace

rpc/src/v1/helpers/light_fetch.rs
niklasad1: missing spaces

```suggestion
	fn logs_common(&self, filter: EthcoreFilter) -> impl Future<Item = BTreeMap<(u64, usize), Log>, Error = Error> + Send {
```

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -308,11 +309,11 @@ impl LightFetch {
 			Some(OnDemandResponse::Receipts(b)) => b,
 			_ => panic!(WRONG_RESPONSE_AMOUNT_TYPE_PROOF),
 		}))
+
+
```
niklasad1: extra whitespace

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -376,6 +376,49 @@ impl LightFetch {
...
+	/// Get transaction logs
+	pub fn logs(&self, filter: EthcoreFilter) -> impl Future<Item = Vec<Log>, Error = Error> + Send {
+		use jsonrpc_core::futures::stream::{self, Stream};
+		let fetcher_block = self.clone();
+		self.logs_common(filter)
+		// retrieve transaction hash.
+			.and_then(move |matches| {
+			let mut blocks = BTreeMap::new();
+			let mut result: Vec<Log> = matches.into_iter().map(|(_, v)| {
```
niklasad1: indentation off

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -376,6 +376,49 @@ impl LightFetch {
...
+	}
+
+	/// Get transaction logs
+	pub fn logs(&self, filter: EthcoreFilter) -> impl Future<Item = Vec<Log>, Error = Error> + Send {
+		use jsonrpc_core::futures::stream::{self, Stream};
+		let fetcher_block = self.clone();
+		self.logs_common(filter)
+		// retrieve transaction hash.
+			.and_then(move |matches| {
+			let mut blocks = BTreeMap::new();
```
niklasad1: indentation off

rpc/src/v1/impls/eth.rs
```diff
@@ -146,6 +146,33 @@ enum PendingTransactionId {
 	Location(PendingOrBlock, usize)
 }
 
+pub fn base_logs<C, M, T: StateInfo + 'static> (client: &C, miner: &M, filter: Filter) -> BoxFuture<Vec<Log>> where
+	C: miner::BlockChainClient + BlockChainClient + StateClient<State=T> + Call<State=T>,
+	M: MinerService<State=T> {
```
niklasad1: Sorry for nitpicking but would prefer formatting the where thingy little differently.

```rust
pub fn base_logs<C, M, T: StateInfo + 'static>(client: &C, miner: &M, filter: Filter)
where
    C: miner::BlockChainClient + BlockChainClient + StateClient<State = T> + Call<State = T>,
    M: MinerService<State = T>,
{}
```

rpc/src/v1/traits/parity.rs
```diff
@@ -241,5 +241,11 @@ build_rpc_trait! {
 		/// as well as checks the signature for chain replay protection
 		#[rpc(name = "parity_verifySignature")]
 		fn verify_signature(&self, bool, Bytes, H256, H256, U64) -> Result<RecoveredAccount>;
+
+		/// Returns logs matching given filter object.
+		/// Skip filling transaction hash for faster query.
+		#[rpc(name = "parity_getLogsLight")]
```
niklasad1: I have no strong opinions on the naming but `getLogsLight` is not self-explanatory as long the documentation explains this I'm satisfied.

Otherwise, I propose `parity_getLogsNoTransactionHash`

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
cheme: Yes I think that is the purpose.
It is the same allocation in both case, just that with the btree I avoid running an iteration on the vec to create the block query futures.

cheme:
@niklasad1 I did switch rpc naming as suggested.

rpc/src/v1/helpers/light_fetch.rs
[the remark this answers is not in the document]
ordian: > I avoid running an iteration on the vec to create the block query futures

Not sure what you mean by that, we have created these futures in `BTreeMap` already.
I meant why don't we call `logs_no_tx_hash` from `logs` (no need for `logs_common`)? I see now we're allocating the same `Vec` into `result` in `logs`. And iterating over a `Vec` should be faster, since it's a sequential memory read.

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -378,6 +377,42 @@ impl LightFetch {
...
+
+	/// Get transaction logs
+	pub fn logs(&self, filter: EthcoreFilter) -> impl Future<Item = Vec<Log>, Error = Error> + Send {
+		use jsonrpc_core::futures::stream::{self, Stream};
+		let fetcher_block = self.clone();
+		self.logs_no_tx_hash(filter)
+			// retrieve transaction hash.
+			.and_then(move |matches| {
+				let mut blocks = BTreeMap::new();
+				let mut result: Vec<Log> = matches.into_iter().map(|v| {
```
ordian: this code can now be simplified a bit:
```rust
			.and_then(move |mut result| {
				let mut blocks = BTreeMap::new();
				result.iter().for_each(|v| {
					let block_hash = v.block_hash.as_ref().expect("Previously initialized with value; qed");
					blocks.entry(block_hash.clone()).or_insert_with(|| {
						fetcher_block.block(BlockId::Hash(block_hash.clone().into()))
					});
				});
```

Issue #9926, opened 2018-11-15
Stage 1 block verification failed - DifficultyOutOfBounds


mirek:
- **Parity Ethereum version**: Parity/v1.11.11-stable-cb03f38-20180910/x86_64-linux-gnu/rustc1.28.0
- **Operating system**: Red Hat Enterprise Linux Server release 7.5 (Maipo)
- **Installation**: ?
- **Fully synchronized**: N/A
- **Network**: internal
- **Restarted**: no

We've got authority round with 4 nodes via safe contract, everything was working fine until recently we've got 3 different histories - only 2 nodes are in sync, other two diverged into their own histories.

Related log entry at the moment they diverged is:

```
2018-11-13 10:05:33 UTC IO Worker #1 INFO import  Imported #13724 0x7133…9559 (1 txs, 0.05 Mgas, 0 ms, 0.97 KiB) + another 1 block(s) containing 1 tx(s)
2018-11-13 10:05:34 UTC IO Worker #3 WARN client  Stage 1 block verification failed for 0xe9e0…4ff9: Error(Block(DifficultyOutOfBounds(OutOfBounds { min: None, max: Some(0xffffffffffffffffffffffffffffffff), found: 0xffffffffffffffffffffffffffffffff })), State { next_error: None, backtrace: None })
2018-11-13 10:05:35 UTC Verifier #0 INFO import  Imported #13724 0xa9ff…e302 (2 txs, 0.52 Mgas, 3 ms, 1.80 KiB)
```

We have quite verbose/trace logs available on all 4 nodes if needed to investigate it further.

joshua-mir:
It would be helpful to have more logs from around the time this happened, especially things like system time on your nodes, and any sealing logs. However, I'd like to note that there aren't any more backports to 1.11.x; you'll mostly be on your own in fixing this issue.

tomusdrw:
@mirek are all 4 nodes running the exact same version and spec file?

mirek:
Yes except signers, which are different.

tomusdrw:
@mirek do you have empty steps enabled? The authorities should re-organize to longer chain anyway, but seems you have a 2/2 split. Did you try stopping the "faulty" nodes that produced the incorrect block(s) and restart them after a while to check if they re-organize?

tomusdrw:
Can you also make sure that time on all authorities is perfectly in sync?

mirek:
We have:
```
        "emptyStepsTransition": "0",
[...]

PR #9939, opened 2018-11-19
Fix empty steps

tomusdrw:
- prevent sending empty step message twice
- prevent sending empty step and then block in the same step
- don't accept double empty steps
- do basic validation of self-sealed blocks

Closes: #9926


andresilva:
We discussed some possible improvements to this. Logged in #9950.

PR #9943, opened 2018-11-20
fix: Missing blocks in filter_changes RPC

mattrutherford:
Fixes: [#9858](https://github.com/paritytech/parity-ethereum/issues/9858)

Implement validity check of reported hashes (in case of re-org)
to determine the last block reported that is still valid. This will return hashes for 
block numbers that have previously been reported if the hash is different.

Add const to set history length for validity check. Currently set to 32 blocks.

### DIFF
--- a/rpc/src/v1/helpers/poll_filter.rs
+++ b/rpc/src/v1/helpers/poll_filter.rs
@@ -17,7 +17,7 @@
 //! Helper type with all filter state data.
 
 use std::{
-	collections::{BTreeSet, HashSet},
+	collections::{BTreeSet, HashSet, VecDeque},
 	sync::Arc,
 };
 use ethereum_types::H256;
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/poll_filter.rs
```diff
@@ -49,7 +49,11 @@ impl SyncPollFilter {
 #[derive(Clone)]
 pub enum PollFilter {
 	/// Number of last block which client was notified about.
-	Block(BlockNumber),
+	Block {
+		last_block_number: BlockNumber,
+		#[doc(hidden)]
```
tomusdrw: why hidden in docs?

mattrutherford: Thanks - I was deliberating whether to hide it or not; I decided it was an implementation detail, but open to removing that.

PR #9944, opened 2018-11-20
allow rust-nightly builds fail in nightly builds

gabreal:
https://doc.rust-lang.org/cargo/guide/continuous-integration.html#gitlab-ci

allow rust-nightly builds to fail

https://gitlab.parity.io/parity/parity-ethereum/pipelines/26136

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -150,3 +150,4 @@ test-nightly:
     - scripts/gitlab/test-all.sh nightly
   tags:
     - rust-nightly
+  allow_failure:                   true


ordian:
I filed an issue for rust https://github.com/rust-lang/rust/issues/56069.

ordian:
@gabreal Since test-nightly is in optional stage, is it technically allowed to fail already?

gabreal:
The whole pipeline is marked as failed regardless of a stage being called optional - that's just a name.

gabreal:
but even with allow_failure the job will still be run, its result just won't be counted to the final result. I think it's more about notifications here.

PR #9946, opened 2018-11-20
prevent silent errors in daemon mode

seunlanlege:


c0deright:
If this is in 2.1.7-stable and the cause for https://github.com/paritytech/parity-ethereum/issues/9991 then I'm really glad I skipped 2.1.7-stable so far... *peh*

5chdn:
Yes it is.

PR #9947, opened 2018-11-20
Missing blocks in filter_changes RPC

mattrutherford:
Fixes: [#9858](https://github.com/paritytech/parity-ethereum/issues/9858)

Implement validity check of reported hashes (in case of re-org)
to determine the last block reported that is still valid. This will return hashes for
block numbers that have previously been reported if the hash is different.

Add const to set history length for validity check. Currently set to 32 blocks.

### DIFF
--- a/rpc/src/v1/tests/mocked/eth.rs
+++ b/rpc/src/v1/tests/mocked/eth.rs
@@ -317,10 +317,27 @@ fn rpc_blocks_filter() {
 
 	tester.client.add_blocks(2, EachBlockWith::Nothing);
 
+	let hash1 = tester.client.block_hash(BlockId::Number(1)).unwrap();
+	let hash2 = tester.client.block_hash(BlockId::Number(2)).unwrap();
 	let response = format!(
 		r#"{{"jsonrpc":"2.0","result":["0x{:x}","0x{:x}"],"id":1}}"#,
-		tester.client.block_hash(BlockId::Number(1)).unwrap(),
-		tester.client.block_hash(BlockId::Number(2)).unwrap());
+		hash1,
+		hash2);
+
+	assert_eq!(tester.io.handle_request_sync(request_changes), Some(response.to_owned()));
+
+	// in the case of a re-org we get same block number if hash is different - BlockId::Number(2)
+	tester.client.blocks.write().remove(&hash2).unwrap();
+	tester.client.numbers.write().remove(&2).unwrap();
+	tester.client.numbers.write().insert(1, hash2);
+	*tester.client.last_hash.write() = hash1;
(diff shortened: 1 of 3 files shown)


rpc/src/v1/tests/mocked/eth.rs
```diff
@@ -317,10 +317,27 @@ fn rpc_blocks_filter() {
...
-		tester.client.block_hash(BlockId::Number(2)).unwrap());
+		hash1,
+		hash2);
+
+	assert_eq!(tester.io.handle_request_sync(request_changes), Some(response.to_owned()));
+
+	// in the case of a re-org we get same block number if hash is different - BlockId::Number(2)
+	tester.client.blocks.write().remove(&hash2).unwrap();
+	tester.client.numbers.write().remove(&2).unwrap();
+	tester.client.numbers.write().insert(1, hash2);
```
tomusdrw: Seems that this line is actually changing hash of block1, why isn't that block part of the response?

mattrutherford: Good catch, that would be the case if the hash was different, but actually in ```import_block()``` in ```test_client.rs```,  that value gets overwritten again by the old hash! So I just need to take that line out because it effectively does nothing.

mattrutherford:
OK, thanks for the comments @andresilva . I like the elegant solution you mentioned, but depends in practice on how often a re-org would exceed 1 block ago; and if so whether duplicating returned hashes would be less less nice from an end-user perspective? Personally I think on balance I would prefer to keep as it is, but I am open to re-working it, if it's deemed the additional overhead is undesirable?

mattrutherford:
OK -  noted, thank you. I see tree_root will however include (previously reported) forked blocks, so depends on whether we care about duplication or not, because we would be able to check for it.
```
A1 -> A2 -> A3 -> A4
                       -> B3 -> B4
 { blocks: [B4, B3, A3, A4], ancestor: A2, index: 2 }
```

mattrutherford:
OK, noted, thanks. I notice that tree_root also includes forked blocks so we'd need to remove those if we care about duplication of the returned hashes.

andresilva:
Ah, I see what you mean, indeed if there was a re-org `tree_route` would also report the retracted route. In this case we would just ignore the retracted part of the route, this method would be more useful if we could notify about previously reported blocks that were retracted (we do this for logs) but the API doesn't allow this right now and we can't break compatibility.

mattrutherford:
I guess we could also rename ```tree_root```, add the extra param and then make a new ```fn tree_root(&self, from: H256, to: H256)```  that basically just delegates to the modified ```fn```.

PR #9952, opened 2018-11-22
Fix a deadlock

ngotchac:
A deadlock has been introduced in #8643:
In `fn commit(&self)` there were:
https://github.com/paritytech/parity-ethereum/blob/9475a2e474d9175abf05b30167bc543bd25a4e3c/ethcore/src/blockchain/blockchain.rs#L1190-L1191

And in `fn chain_info(&self) ` there were:
https://github.com/paritytech/parity-ethereum/blob/9475a2e474d9175abf05b30167bc543bd25a4e3c/ethcore/src/blockchain/blockchain.rs#L1533-L1534

So basically, if `chain_info` was called while `commit` was undergoing, it could happen that `commit` would W-lock `self.best_ancient_block`, `chain_info` would R-lock `self.best_block`, and then `commit` would wait to lock `self.best_block` while `chain_info` would wait to lock `self.best_ancient_block`.

I'm not sure if there is a better way to fix this, since it's not the first time we had an issue with this (#9385 fixed a deadlock, but broke the informant, then fixed in #9932).

I try to update `chain_info` so that it could return an `Option` with `try_read`, and the informant would just keep trying to get a `Report` until one is available, but that breaks a dozen of other files. If anyone has a better idea...

This PR also updates the informant, so that it prints info every 5 seconds, instead of previously varying randomly between 5s and 10s.

### DIFF
--- a/ethcore/src/blockchain/blockchain.rs
+++ b/ethcore/src/blockchain/blockchain.rs
@@ -1187,8 +1187,8 @@ impl BlockChain {
 		let mut pending_block_details = self.pending_block_details.write();
 		let mut pending_write_txs = self.pending_transaction_addresses.write();
 
-		let mut best_ancient_block = self.best_ancient_block.write();
 		let mut best_block = self.best_block.write();
+		let mut best_ancient_block = self.best_ancient_block.write();
 		let mut write_block_details = self.block_details.write();
 		let mut write_hashes = self.block_hashes.write();
 		let mut write_txs = self.transaction_addresses.write();
@@ -1530,8 +1530,8 @@ impl BlockChain {
 		let genesis_hash = self.genesis_hash();
 
 		// ensure data consistencly by locking everything first
-		let best_ancient_block = self.best_ancient_block.read();
 		let best_block = self.best_block.read();
+		let best_ancient_block = self.best_ancient_block.read();
 		BlockChainInfo {
 			total_difficulty: best_block.total_difficulty,
 			pending_total_difficulty: best_block.total_difficulty,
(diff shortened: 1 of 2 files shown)


ethcore/src/blockchain/blockchain.rs
```diff
@@ -1530,8 +1530,8 @@ impl BlockChain {
 		let genesis_hash = self.genesis_hash();
 
 		// ensure data consistencly by locking everything first
-		let best_block = self.best_block.read();
 		let best_ancient_block = self.best_ancient_block.read();
+		let best_block = self.best_block.read();
```
tomusdrw: I don't think it's a good solution. The general rule is described here:
https://github.com/paritytech/parity-ethereum/blob/9475a2e474d9175abf05b30167bc543bd25a4e3c/ethcore/src/blockchain/blockchain.rs#L209

So simply always capture locks in the order of declaration within a struct.

parity/informant.rs
```diff
@@ -289,7 +286,8 @@ impl<T: InformantData> Informant<T> {
 			return;
 		}
 
-		*self.last_tick.write() = Instant::now();
+		*self.last_tick.write() = now;
+		*self.last_report.lock() = full_report.client_report.clone();
```
tomusdrw: What's the gain of moving this here?

ethcore/src/blockchain/blockchain.rs
ngotchac: Hm, OK... But for example I wasn't aware of this comment, nor did the ones who review the PR, so I was wondering if we could find a better solution.

PR #9955, opened 2018-11-22
Disable EIP-98 transition by default

sorpaas:
This PR replaces the default value of `eip98Transition` to `BlockNumber::max_value`, to reduce the chance of chain mis-configs. Note that:

* If a chain has EIP-658 enabled at any point, then it "disables" EIP-98. So we don't need to specify them in `poacore` and `poasokol`.
* This leaves only Kovan configs need to be changed.
* Removing existing `"eip98Transition": "0xffffffffffffffff"` is entirely optional.

For release note, we need to specify that:

> For custom chain config, if you don't have `eip98Transition` specifically set, and you don't have `eip658Transition` enabled at block 0, then you need to add `"eip98Transition": 0`.

### DIFF
--- a/ethcore/res/ethereum/byzantium_test.json
+++ b/ethcore/res/ethereum/byzantium_test.json
@@ -28,7 +28,6 @@
 		"eip160Transition": "0x0",
 		"eip161abcTransition": "0x0",
 		"eip161dTransition": "0x0",
-		"eip98Transition": "0xffffffffffffffff",
 		"eip140Transition": "0x0",
 		"eip211Transition": "0x0",
 		"eip214Transition": "0x0",
--- a/ethcore/res/ethereum/callisto.json
+++ b/ethcore/res/ethereum/callisto.json
@@ -32,7 +32,6 @@
 		"eip161abcTransition": 10,
 		"eip161dTransition": 10,
 		"eip155Transition": 10,
-		"eip98Transition": "0xffffffffffffffff",
 		"eip140Transition": 20,
 		"eip211Transition": 20,
 		"eip214Transition": 20,
--- a/ethcore/res/ethereum/classic.json
+++ b/ethcore/res/ethereum/classic.json
@@ -30,8 +30,7 @@
 		"eip160Transition": 3000000,
 		"eip161abcTransition": "0x7fffffffffffffff",
 		"eip161dTransition": "0x7fffffffffffffff",
-		"eip155Transition": 3000000,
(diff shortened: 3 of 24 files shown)


stone212:
@5chdn Can you please help explain the language used here:

```
If a chain has EIP-658 enabled at any point, then it "disables" EIP-98. So we don't need to specify them in poacore and poasokol.
```

The phrase "we don't need to specify them in poacore and poasokol" is confusing for a few reasons.

1. "don't need to" or must not?

2. is this only for PoA chains?  If I have a private blockchain with PoW and EIP-658, is it necessary to add EIP-98?  I just do not understand the language of this release note.

joshua-mir:
@stone212 the two eips are mutually exclusive - they should never both be enabled on the same chain.

stone212:
@joshua-mir Okay so if I have `eip658Transition` (and I do) in my chain spec, I should *not* have `eip98transition` (and I do not) correct?

Because eip98transition is dis-abled by default after 2.3.0 so `eip658Transition` should be happy on a chain running Parity 2.3.x.  Do I understand this?

stone212:
@joshua-mir 

>  there is no special case for Aura.

What is this "Aura" you keep talking about?  I do not know what this is.

joshua-mir:
Aura is our PoA consensus engine.

> Because eip98transition is dis-abled by default after 2.3.0 so eip658Transition should be happy on a chain running Parity 2.3.x. Do I understand this?

Correct.

stone212:
@joshua-mir 

> Aura is our PoA consensus engine.

Okay but now this goes back in a circle to my earlier question, the vague language.  The release notes above make it seem like this issue only applies to PoA.  So let me ask again and mention PoW again:

Because eip98transition is dis-abled by default after 2.3.0 so eip658Transition should be happy on a chain running Parity 2.3.x *and using PoW / ethash*. Do I continue to understand this?

joshua-mir:
Yes, you're not misunderstanding anything. 

The reason why there are no PoW networks mentioned in the comments above is because very few people run private PoW networks. Unless a param is in the params section within the engine section of your config, it is unrelated to the consensus engine of your chain.

stone212:
@joshua-mir 

> Yes, you're not misunderstanding anything.

I am getting confused by the language.  Please, a yes or no:

If I have a private blockchain with PoW and EIP-658 (at a block that is past but is not block 0), is it necessary to add eip98transition at 0?

Issue #9865, opened 2018-11-05
Websocket: newBlockHeaders misses several blocks


ankitchiplunkar:
_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: v2.0.8-stable
- **Operating system**: Ubuntu 18.04
- **Installation**:  one-line installer 
- **Fully synchronized**: yes
- **Network**: ethereum 
- **Restarted**: no

#### Expected behaviour
The websocket when subscribed using `newBlockHeaders` should return the forked blocks and also  the canonical blocks. 

#### Actual behaviour
The websocket when subscribed using `newBlockHeaders`  does not return the canonical blocks (for almost every ~50 blocks)

#### Steps to reproduce the behaviour
I have a local parity node, am using web3.js for subscribing to `newBlockHeaders`, below is the code to get new blocks.

```
var Web3 = require('web3')
const fs = require('fs');
var web3 = new Web3();

web3.setProvider('ws://localhost:8546');

var subscription = web3.eth.subscribe('newBlockHeaders', function(error, result){
    if (!error) {
        console.log(result['hash']);

        return;
    }

    console.error(error);
}).on("data", function(blockHeader){
    console.log(blockHeader['number']);
    strData = "data, " + blockHeader['hash'] +', ' + blockHeader['number'] + ', '+ blockHeader['parentHash'] + "\n"
    fs.appendFile('test_pubsub.txt', strData , function (err) {
        if (err) throw err;
      });
}).on("changed", function(blockHeader){
    console.log(blockHeader['number']);
    strData = "change, " + blockHeader['hash'] +', ' + blockHeader['number'] + ', '+ blockHeader['parentHash'] + "\n"
    fs.appendFile('test_pubsub.txt', strData , function (err) {
        if (err) throw err;
      });
}).on("error", console.error);
```

The subscription service regularly misses blocks, in the example below we get the blocks `6630351` and `6630353` but not the block in between `6630352`.

```
0x9f3204da53f748aa52db99811bef51378ca303629211a5420f9b40aa1782cf2a, 6630351, 0x95e806a60dab8d9611668a485d1d98fc4e66525657a97ff75788e4252ddf6c71
0x532180a19001eff004853b625fa9e537db7a1174291a0229332dd5d94f6cbd97, 6630353, 0x784fb94564c6803334fa51f22fdc426c2186b73fe11a1dd14a50ca2e159b4c95
```

This has been documented before: https://github.com/ethereum/web3.js/issues/1375

[...]

PR #9964, opened 2018-11-26
[WIP] - Reliable sync check for new_blocks notifications

mattrutherford:
New block notifications are sometimes dropped during normal operation due to a false-positive check for whether node is in state of (major) syncing. 

- [x] Create new type to cache the size of collection in an AtomicUsize when releasing Mutex.
- [ ] Implement for `Verification`.
- [ ] Use this to determine sync state in `ethcore` `Client`

Will fix issue [9865](https://github.com/paritytech/parity-ethereum/issues/9865)


mattrutherford:
Closing to split into two separate PRs

Issue #646, opened 2016-03-09
Peer rating and reputation system


arkpar:

rphmeier:
Would be useful for light client -- peers can be ranked by how useful they were at serving data and multiplexed across a limited number of connection slots.

5chdn:
Is this implemented for the light client?

rphmeier:
No, this would most likely be done at the devp2p level.

PR #9967, opened 2018-11-26
Fix unstable peers and slowness in sync

ngotchac:
This PR fixes some issues with the stability of peer counts, and overall slowness in block sync.

Here are graphs that compares block height against time for the current `stable` and this branch:
block_heights
block_heights
Here for the peer counts:
peer_counts
peer_counts

The issue was that `sync.continue_sync` was called every time a packet was received, so it had to go through the whole list of peers, randomize it, etc.. Which could be quite expensive when connected to a lot of peers.

Now it is called on a timer, every 2.5 seconds.


ethcore/sync/src/api.rs
```diff
@@ -397,7 +398,8 @@ impl NetworkProtocolHandler for SyncProtocolHandler {
 	fn initialize(&self, io: &NetworkContext) {
 		if io.subprotocol_name() != WARP_SYNC_PROTOCOL_ID {
 			io.register_timer(PEERS_TIMER, Duration::from_millis(700)).expect("Error registering peers timer");
-			io.register_timer(SYNC_TIMER, Duration::from_millis(1100)).expect("Error registering sync timer");
+			io.register_timer(MAINTAIN_SYNC_TIMER, Duration::from_millis(1100)).expect("Error registering sync timer");
+			io.register_timer(CONTINUE_SYNC_TIMER, Duration::from_millis(2_500)).expect("Error registering sync timer");
```
sorpaas: Either `2500` or change the above and below line to be with `_`.

ethcore/sync/src/chain/mod.rs
```diff
@@ -655,37 +655,35 @@ impl ChainSync {
...
+			if peers.len() > 0 {
+				trace!(
+					target: "sync",
+					"Syncing with peers: {} active, {} available, {} total",
+					self.active_peers.len(), peers.len(), self.peers.len()
+				);
 
-			for (peer_id, _) in peers {
-				self.sync_peer(io, peer_id, false);
+				random::new().shuffle(&mut peers); //TODO: sort by rating
```
sorpaas: Not caused by this PR, but please create an issue and include an issue number for this TODO!

sorpaas: nit: missing space `// TODO`.

ngotchac: I think it's this one: https://github.com/paritytech/parity-ethereum/issues/646

ethcore/sync/src/chain/mod.rs
```diff
@@ -655,37 +655,35 @@ impl ChainSync {
...
+			if peers.len() > 0 {
+				trace!(
+					target: "sync",
+					"Syncing with peers: {} active, {} available, {} total",
+					self.active_peers.len(), peers.len(), self.peers.len()
+				);
 
-			for (peer_id, _) in peers {
-				self.sync_peer(io, peer_id, false);
+				random::new().shuffle(&mut peers); // TODO: sort by rating
```
sorpaas: @ngotchac If it's #646, would you mind to add the number here? Like this:

```suggestion
				random::new().shuffle(&mut peers); // TODO (#646): sort by rating
```

This was originally proposed for Substrate development (which now being strictly followed). Having `TODO` with only short description for many times makes it no easy to guess the original intention of the author, and thus make it difficulty for others to pick up the `TODO`. So it was decided that for PRs, all `TODO` and `FIXME` must either be removed, or include an issue number.

andresilva:
Should we backport this as well?

folsen:
@ngotchac just out of curiosity, how did you generate the graphs?

ngotchac:
@folsen I wrote a small Rust script that runs the given binary with a given config basically, then prints the graphs using `gnuplot`. Are you interested? I can publish it on GH

folsen:
@ngotchac definitely interested, seems to cross over with what some of the QA guys are working on

ngotchac:
The script used to get the graphs : https://github.com/ngotchac/eth-metrics
It's very PoC for now.

PR #9970, opened 2018-11-27
 ci: rearrange pipeline by logic

5chdn:
- audit job is renamed and moved to test stage (is cheap to run and should never fail)
- docs job is renamed and moved to the publish stage (where it belongs)
- all jobs in the optional stage are allowed to fail


ordian:
Note, that `cargo audit` fails when a vulnerability was discovered in a crate we're using. It may happen at any point in time. It shouldn't block merging a PR, that has changes unrelated to this failure (even if CI is failing), unless it introduces a dependency with a vulnerability, that is not yet fixed.

5chdn:
@ordian I see, but it's not required - so it's not blocking the PR to be merge. It just shows a big red X once something is discovered, wherever it may come from.

PR #9972, opened 2018-11-27
Deny unknown fields for chainspec

sorpaas:
This adds `deny_unknown_fields` to all spec deserialiser. In this way, a newer chainspec cannot be accidentially used in an older parity-ethereum version. If any transition is deprecated/removed, we also ensure that an older chainspec that contains those won't accidentially be loaded on newer parity-ethereum version.

After this, the chainspec tool doesn't need to manually check ignored fields any more. So we can remove `serde_ignored` dep.


stone212:
@sorpaas 

> If any transition is deprecated/removed, we also ensure that an older chainspec that contains those won't accidentially be loaded on newer parity-ethereum version.

In what way will this "ensure" that the older chain spec can not be loaded on newer parity-ethereum?  I have a problem with nodes running 2.4.x+ syncing with a healthy blockchain running 2.3.9.  Could this pull request be preventing sync?  My issue: https://github.com/paritytech/parity-ethereum/issues/11397

Issue #9953, opened 2018-11-22
Make tests pass on nightly again


ordian:
I guess the problem is that rust nightly no longer uses jemalloc and we're using `heapsize` crate, which assumes it does (cc rust-lang/rust/issues/56069). A temporary solution would be to use jemalloc as a global allocator (https://github.com/alexcrichton/jemallocator ?).

ordian:
This is worth doing soon, since this change will make its way into the rust stable eventually. 
I remember @tomaka and @cheme had some thoughts on this.

cheme:
For short term I can put jemalloc as default allocator with crate jemallocator, I currently need to fork heapsize with very few changes:
```
diff --git a/Cargo.toml b/Cargo.toml
index ec2747a..b3c2c0e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,10 @@ license = "MIT/Apache-2.0"
 repository = "https://github.com/servo/heapsize"
 build = "build.rs"
 
+
+[dependencies]
+jemallocator = "0.1.9"
+
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.4", features = ["std", "heapapi"] }
 
diff --git a/src/lib.rs b/src/lib.rs
index 7eecbce..85e85e0 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,13 @@
 #[cfg(target_os = "windows")]
 extern crate winapi;
 
+#[cfg(not(target_os = "windows"))]
+extern crate jemallocator;
+
+#[cfg(not(target_os = "windows"))]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
+
 #[cfg(target_os = "windows")]
 use winapi::um::heapapi::{GetProcessHeap, HeapSize, HeapValidate};
 use std::borrow::Cow;
@@ -36,15 +43,7 @@ pub unsafe fn heap_size_of<T>(ptr: *const T) -> usize {
 
 #[cfg(not(target_os = "windows"))]
 unsafe fn heap_size_of_impl(ptr: *const c_void) -> usize {
-    // The C prototype is `je_malloc_usable_size(JEMALLOC_USABLE_SIZE_CONST void *ptr)`. On some
-    // platforms `JEMALLOC_USABLE_SIZE_CONST` is `const` and on some it is empty. But in practice
[...]

PR #9977, opened 2018-11-27
Getting rid of heapsize step 1: miner

mocsy:
size_of_val in std seems to work fine.
This would allow this project to get rid of heapsize crate dep.

May or may not help out with https://github.com/paritytech/parity-ethereum/issues/9953

### DIFF
--- a/ethcore/transaction/tests/mod.rs
+++ b/ethcore/transaction/tests/mod.rs
@@ -1,0 +1,77 @@
+extern crate ethcore_transaction;
+extern crate ethereum_types;
+extern crate ethkey;
+#[cfg(test)]
+extern crate heapsize;
+extern crate rustc_hex;
+
+mod tests {
+	use super::*;
+	use ethcore_transaction::*;
+	use ethereum_types::U256;
+
+	#[test]
+	fn heapsize_should_match_std() {
+		assert_eq!(8, std::mem::size_of_val(&5u64));
+		assert_eq!(4, std::mem::size_of_val(&5u32));
+		assert_eq!(1, std::mem::size_of_val(&5u8));
+
+		use heapsize::HeapSizeOf;
+		let bytes: Vec<u8> = vec![1, 15, 31, 63, 127, 255];
+		let u32s: Vec<u32> = vec![1, 15, 31, 63, 127, 255];
+		let u64s: Vec<u64> = vec![1, 15, 31, 63, 127, 255];
+
+		let z: Vec<u64> = Vec::new();
+		assert_eq!(0, z.heap_size_of_children());
+		assert_eq!(0, std::mem::size_of_val(&*z));
+
+		let exp8 = 6;
+		let exp32 = 24;
+		let exp64 = 48;
+
+		assert_eq!(exp8, std::mem::size_of_val(&*bytes));
+		assert_eq!(exp32, std::mem::size_of_val(&*u32s));
+		assert_eq!(exp64, std::mem::size_of_val(&*u64s));
+
+		let byte_slice: &[u8] = &[1, 15, 31, 63, 127, 255];
+		let u32_slice: &[u32] = &[1, 15, 31, 63, 127, 255];
+		let u64_slice: &[u64] = &[1, 15, 31, 63, 127, 255];
+		assert_eq!(exp8, std::mem::size_of_val(&*byte_slice));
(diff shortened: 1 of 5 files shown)


mocsy:
[clabot:check]

ethcore/transaction/tests/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    use ethcore_transaction::*;
+    use ethereum_types::U256;
+
+    #[test]
+    fn heapsize_should_match_std() {
+		assert_eq!(8, std::mem::size_of_val(&5u64));
+		assert_eq!(4, std::mem::size_of_val(&5u32));
+		assert_eq!(1, std::mem::size_of_val(&5u8));
+
+        use heapsize::HeapSizeOf;
```
sorpaas: Please clean this to use tab for indentation!

mocsy:
6 is the size of the data of:
`let bytes: Vec<u8> = vec![1, 15, 31, 63, 127, 255];`
On windows both heap_size and std::mem give 6, that's fine.

Using heapsize the linux story is different:
On stable it claims data is padded to 2^n, which is 8 in this case.
On nightly reports 32bit which is wrong, since nightly no longe uses jemalloc I could actually measure using heaptrack and std::mem is spot on.

I propose to measure the data size using std::mem and remove heapsize.
I updated the test I provided accordingly.

mem_size

Note: heap_size_of_children needs to be removed later when all packages use heap_size_of_data instead.
In this PR I updated miner only.

ethcore/transaction/tests/mod.rs
mocsy: Done

lazaridiscom:
In my attempt to understand all this, I hit on a comment which is possibly relevant:

servo/heapsize#80 (comment)

mocsy:
> In my attempt to understand all this, I hit on a comment which is possibly relevant:
> 
> [servo/heapsize#80](https://github.com/servo/heapsize/issues/80) (comment)

My understanding is the goal here is to generate statistics and for that size_of_val should work fine.
size_of_val tells us the size of the data, and not how many bytes the allocator would allocate.
A custom allocator could reserve more bytes than the data needs. This might not tell if the allocator uses extra spacce for something or not. The std feature is still useful to recognize if data has been freed or not, and the number would be more consistent between platforms.
The other option would be to use allocator specific functions.
Heapsize remains jemalloc specific, and malloc_size_of looks specific to me as well.

lazaridiscom:
I read somewhere that heapsize is more precise and faster (due to jemalloc beeing faster).

Shouldn't the goal be to remove any barriers (fix issues) that block the usage of heapsize?

Or is it save to say that "heapsize has nothing positive to offer, and should be removed" ?

(I ask very general and more out of curiosity! Please do not feel annoyed, its not about your work/PR)

mocsy:
What block the usage of heapsize is rustc switching migrating off of jemalloc.
heapsize remains jeamalloc only as you found https://github.com/servo/heapsize/issues/80

On https://github.com/paritytech/parity-ethereum/issues/9953 two approaches were discussed, switching back to jemalloc and keep using heapsize, or embracing the new default and stop depending on heapsize.

This PR proposes a way to do the second.

About heapsize being precise: it assumes jemalloc and it's precise as long as jemalloc is used, but if you look at the screenshot and data I posted above, it shows it's not precise with the new version of rust.

lazaridiscom:
Thank you for taking the time to elaborate.

I don't know the Rust Domain enough in order to know good reference projects. The standard-procedure is too look at the reference-projects, in order to see how they deal with the issue.

As an example, servo removes heapsize, and one could simply lookup the way it is replaced and adopt it:

* https://github.com/servo/webrender/pull/2192
* https://github.com/servo/servo/pull/19510
* servo/euclid@fb143ba5

So, I assume that if a heapsize-related change is similar to what the servo project (or other reference-projects) do, then it should be fine.

(clarifying again: me commenting abstract/general)

5chdn:
needs a 2nd reviewer, @ordian @seunlanlege anyone? :)

ordian:
@mocsy thanks for your contribution. The fix for #9953 is being worked on in paritytech/parity-common/pull/89.

I believe `mem::size_of_val` doesn't fully fit our purposes, e.g. it shows its size and not capacity (the actual amount of allocated memory). So I'm closing this PR.

PR #9987, opened 2018-11-29
fix: new_blocks notifications sometimes not triggered during normal conditions

mattrutherford:
Remove duplicate check for ```is_empty```. Check total queue size is < 10 to send new blocks notifications (to indicate not doing major sync). Tried initially with queue size of 3 (as used in ```pub fn is_major_importing_or_waiting``` in ```rpc/src/v1/helpers/block_import.rs```) but during overnight testing it missed a block; increasing to 10 returned all canonical blocks in the same test setup.

Closes [9865](https://github.com/paritytech/parity-ethereum/issues/9865)

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -327,23 +327,25 @@ impl Importer {
 		};
 
 		{
-			if !imported_blocks.is_empty() && is_empty {
+			if !imported_blocks.is_empty() {
 				let route = ChainRoute::from(import_results.as_ref());
 
 				if is_empty {
 					self.miner.chain_new_blocks(client, &imported_blocks, &invalid_blocks, route.enacted(), route.retracted(), false);
 				}
 
-				client.notify(|notify| {
-					notify.new_blocks(
-						imported_blocks.clone(),
-						invalid_blocks.clone(),
-						route.clone(),
-						Vec::new(),
-						proposed_blocks.clone(),
-						duration,
-					);
-				});
+				if client.queue_info().total_queue_size() < 10 {
+					client.notify(|notify| {
+						notify.new_blocks(
+							imported_blocks.clone(),
+							invalid_blocks.clone(),
+							route.clone(),
+							Vec::new(),
+							proposed_blocks.clone(),
+							duration,
+						);
+					});
+				}
 			}
 		}
 


ethcore/src/client/client.rs
```diff
@@ -327,23 +327,25 @@ impl Importer {
...
-					notify.new_blocks(
-						imported_blocks.clone(),
-						invalid_blocks.clone(),
-						route.clone(),
-						Vec::new(),
-						proposed_blocks.clone(),
-						duration,
-					);
-				});
+				if client.queue_info().total_queue_size() < 10 {
```
sorpaas: Not sure whether calling `queue_info` here is expensive. We fetch many other not-used info by this and it acquires three atomics. And I suppose this function is something that needs to be really fast, so it may matter.

sorpaas: This hard-coding of `0` won't fit slower network conditions. Can we parameterize this?

mattrutherford: You're right that it includes a bunch of stuff that we don't need. It currently requires 3 Mutex locks  - I have a separate suggestion for a way to negate the need to lock [WIP](https://github.com/paritytech/parity-ethereum/pull/9988), which could mitigate risk here.

mattrutherford: You mean 10 might be too high? Maybe checking for distance from most recent network block might be a better way to do this? Do we have a reference for that?

mattrutherford:
I don't think increasing ```max_round_blocks_to_import``` will fix this because we miss around 1 in 50 block notifications in normal operation.

sorpaas:
@mattrutherford Do you mind to explain? I still don't see why we would get different results compared with increasing `max_round_blocks_to_import`.

sorpaas:
@mattrutherford From the log you provided, isn't it the case that when `queue_size` is empty, `is_empty` is always `true`? Are we really sure the bug is caused by `is_empty` check?

From the code, it looks like we add items to `VerificationQueue::processing` when adding blocks to the queue, and when we drained blocks from the queue via `VerificationQueue::drain`, we always call `mark_as_good` or `mark_as_bad` which clears items from `VerificationQueue::processing`. So it looks like when the queue is empty, we should always have `processing` to be empty. Otherwise we probably have a bug there.

mattrutherford:
@sorpaas Thanks - what you say makes sense. @tomusdrw  Also I think this is good to allow more flexibility.

mattrutherford:
I decided to take the suggestion by @tomusdrw because I think it's pragmatic to let each implementation have the information and decide what appropriate action to take. All existing functionality (except for the updated pubsub impl) remains as before.

The current solution has been tested over a 48hr period and appears stable with no missed notifications, previously missing a block notification every ~10 min or so.

I think we should open a new issue to examine the way the queue works and decide if it's working as expected. 

Another note - it may be possible to optimise ```NewBlocks```and have the members be references rather than owned values, allowing each impl to clone when necessary.

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -220,18 +218,9 @@ impl<C: LightClient> LightChainNotify for ChainNotificationHandler<C> {
...
-		&self,
-		_imported: Vec<H256>,
-		_invalid: Vec<H256>,
-		route: ChainRoute,
-		_sealed: Vec<H256>,
-		// Block bytes.
-		_proposed: Vec<Bytes>,
-		_duration: Duration,
-	) {
+	fn new_blocks(&self, new_blocks: NewBlocks) {
```
sorpaas: Can you add a check below this line that if we have no log_subscribers and no head_subscribers then early return? We do init some structs below regardless of whether there're subscribes. They may make normal syncing slow.

sorpaas:
@mattrutherford Also please update PR description to make @5chdn's job easier!

ethcore/src/client/client.rs
```diff
@@ -322,26 +322,29 @@ impl Importer {
...
 				client.notify(|notify| {
 					notify.new_blocks(
-						imported_blocks.clone(),
-						invalid_blocks.clone(),
-						route.clone(),
-						Vec::new(),
-						proposed_blocks.clone(),
-						duration,
+						NewBlocks::new(
+							imported_blocks.clone(),
```
tomusdrw: Could we pass references here instead of cloning the `Vec` every time?
Or maybe better pass `Arc<NewBlocks>` to avoid cloning the content if reference is not possible?

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -220,18 +218,10 @@ impl<C: LightClient> LightChainNotify for ChainNotificationHandler<C> {
...
+		if self.heads_subscribers.read().len() == 0 && self.logs_subscribers.read().len() == 0 { return }
```
tomusdrw: `is_empty()` instead of `len() == 0`? Or maybe if it's a hot code consider storing `AtomicBool` intead of acquring those two locks

ethcore/src/client/client.rs
mattrutherford: That's what I was thinking about too (comment above). There might be some more improvements to make, but I was wondering where to draw the line with this PR. I think this is worth doing this however.

tomusdrw: I'm fine with a separate PR improving this, just make sure to create an issue for this and add `patch` labels.

rpc/src/v1/impls/eth_pubsub.rs
mattrutherford: I think this is OK to leave as a `read()`, the only time we need a write lock is when adding or removing subscribers. Updated to use `is_empty()`

ethcore/src/client/chain_notify.rs
```diff
@@ -114,19 +114,51 @@ impl ChainRoute {
...
+	/// Route
+	pub route: ChainRoute,
+	/// Sealed
+	pub sealed: Vec<H256>,
+	/// Block bytes.
+	pub proposed: Vec<Bytes>,
+	/// Duration
+	pub duration: Duration,
+	/// `Validation.processing` is empty?
+	pub processing_is_empty: bool,
```
tomusdrw: I'm not fully sold on the name, I'd prefer something less technical and more descriptive, like maybe: `has_more_blocks_to_import`?

mattrutherford: actually that's a good call, because it also reverses the polarity of the bool which makes it more intuitive to use

ethcore/src/client/chain_notify.rs
```diff
@@ -114,19 +114,51 @@ impl ChainRoute {
...
-		&self,
-		_imported: Vec<H256>,
-		_invalid: Vec<H256>,
-		_route: ChainRoute,
-		_sealed: Vec<H256>,
-		// Block bytes.
-		_proposed: Vec<Bytes>,
-		_duration: Duration,
-	) {
+	fn new_blocks( &self, _new_blocks: NewBlocks) {
```
sorpaas:
```suggestion
 	fn new_blocks(&self, _new_blocks: NewBlocks) {
```

PR #9988, opened 2018-11-29
LenCachingMutex

mattrutherford:
While working on another issue I developed a way to query the size of a collection held in a ```Mutex``` without requiring a lock. Ultimately I found a a different way to solve the issue without needing this, but it may be useful anyway. I've included a commit to show it used in ```Verification```. If another thread has the lock while ```load_len()``` is called, the cached value may not be accurate. However this is often not too different to normal operation - stored len() may be stale at any point after the Guard is dropped.

### DIFF
--- a/util/len-caching-mutex/src/lib.rs
+++ b/util/len-caching-mutex/src/lib.rs
@@ -123,6 +123,13 @@ mod tests {
 
 	#[test]
 	fn works_with_vec() {
+		let v: Vec<i32> = Vec::new();
+		let lcm = LenCachingMutex::new(v);
+		assert_eq!(lcm.lock().shrink_to_fit(), ());
+	}
+
+	#[test]
+	fn works_with_vecdeque() {
 		let v: VecDeque<i32> = VecDeque::new();
 		let lcm = LenCachingMutex::new(v);
 		assert_eq!(lcm.lock().shrink_to_fit(), ());
--- a/util/len-caching-mutex/src/lib.rs
+++ b/util/len-caching-mutex/src/lib.rs
@@ -13,14 +13,17 @@
 
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+extern crate parking_lot;
 
 use std::collections::VecDeque;
 use std::ops::{Deref, DerefMut};
-use std::sync::Mutex;
-use std::sync::MutexGuard;
-use std::sync::PoisonError;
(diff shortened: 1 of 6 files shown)


util/len-caching-mutex/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+	pub fn new(data: T) -> LenCachingMutex<T> {
+		LenCachingMutex {
+			len: AtomicUsize::new(data.len()),
+			data: Mutex::new(data),
+		}
+	}
+
+	/// Load the most recent value returned from your `T`'s `len()`
+	pub fn load_len(&self) -> usize {
+		self.len.load(Ordering::Relaxed)
```
tomusdrw: I'd prefer using `SeqCst` ordering here (everywhere). I remember we had some discussions early in Parity Ethereum and afair `Relaxed` is `SeqCst` on x86 anyway, and on ARM we might get some heissenbugs because we use `Relaxed`. The performance gain is probably neglegible, so I'd prefer consistency and bug-less code.

ordian: We don't actually need an `AtomicUsize` here, since we're protecting both the data and the `len` with the given mutex. So technically we could write something like this
```rust
pub struct LenCachingMutex<T> {
  lock: Mutex<()>,
  data: T,
  len: usize,
}
```
But that would involve some amount of unsafe code. 
So IMHO using `Relaxed` here is fine.

tomusdrw: @ordian the whole point is to have a simple rule of thumb: never use `Relaxed`, cause it has neglegible performance gains and may cause heisenbugs. I'm fine with `AcqRel` ordering in both places.

mattrutherford: @tomusdrw @ordian  Thanks for the feedback.  I'm definitely happy to use ```SeqCst```, particularly if it helps people reason about the code.

util/len-caching-mutex/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+			data: Mutex::new(data),
+		}
+	}
+
+	/// Load the most recent value returned from your `T`'s `len()`
+	pub fn load_len(&self) -> usize {
+		self.len.load(Ordering::SeqCst)
+	}
+
+	pub fn lock(&self) -> Guard<T> {
```
sorpaas: Please remember to add docs for public functions in the future!

PR #9992, opened 2018-11-29
Fix Bloom migration

### LINKED ISSUE #9629: DB migration from 1.11.1 to 2.x causes event logs to disappear
- **Parity Ethereum version**: 2.1.1
- **Operating system**: Linux
- **Installation**: homebrew
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

It worked fine in Parity 1.11.x branch... Now it returns zero results.  The request is asking for all transfer events over a large period of time, so yes it should return a large amount of data and it usually takes a good 30-40 seconds to run.  But its not working at all now, and returns instantly with zero events. No errors.

Using web3@0.20.7

ngotchac:
Fixes #9629

The bloom's block number where wrongly converted from the previous format (the index was in fact `block_number / 16`, because there were `16` blooms per entry).

### DIFF
--- a/parity/db/rocksdb/blooms.rs
+++ b/parity/db/rocksdb/blooms.rs
@@ -23,7 +23,7 @@ use rlp;
 use super::kvdb_rocksdb::DatabaseConfig;
 use super::open_database;
 
-const LOG_BLOOMS_ELEMENTS_PER_INDEX: usize = 16;
+const LOG_BLOOMS_ELEMENTS_PER_INDEX: u64 = 16;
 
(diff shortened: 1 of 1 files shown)


andresilva:
Sorry misclick. In the future if we see issues with missing logs we should ask users whether their db migrated from 1.x to 2.x (since blooms were borked in the process).

naikmyeong:
@andresilva May I ask is there any ETA to update https://github.com/paritytech/parity-ethereum/pull/9762 PR? Thanks.

sorpaas:
@naikmyeong Let's move discussions about progpow to #9762.

PR #9996, opened 2018-11-30
Limit sync reorg to 20 blocks.

ethernomad:
Parity briefly had this feature 2 years ago: https://github.com/paritytech/parity-ethereum/pull/3519/commits/f396caa16336ec3a3c2d317ff6e20fcd954de29d

It is a crude but effective means of preventing a 51% attack. Similar to what has been recently implemented on BCH.

If an attacker does manage to fork the chain by sending different blocks to different nodes after 19 blocks, it could be detected and exchanges could halt deposits until the situation is rectified.

This is very useful for chains that do not have strong hash rate.

Should this be a chainspec parameter?

An even better technique might be something like this: https://www.reddit.com/r/btc/comments/9zrmtx/a_better_antireorg_algorithm_using_firstseen/

### DIFF
--- a/ethcore/sync/src/block_sync.rs
+++ b/ethcore/sync/src/block_sync.rs
@@ -37,6 +37,7 @@ const SUBCHAIN_SIZE: u64 = 256;
 const MAX_ROUND_PARENTS: usize = 16;
 const MAX_PARALLEL_SUBCHAIN_DOWNLOAD: usize = 5;
 const MAX_USELESS_HEADERS_PER_ROUND: usize = 3;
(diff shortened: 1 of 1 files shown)


5chdn:
Honestly, a client should be able to reorg to any length of chain with higher total difficulty.

And, if 51% of the network agree to change the rules, I don't see why a client should prevent that? Could you clarify what exact scenario you have in mind?

ethernomad:
Actually Wei Tang pointed out that --pruning-history can be used to accomplish the same thing.

PR #9999, opened 2018-11-30
Private tx enabled flag added into STATUS packet

grbIzl:
This PR is aimed to fix the following situation:
- Private tx functionality is enabled via launch flag.
- It's disabled by default
- In order to process private tx packets node has to be launched with this flag enabled 
As a result, on big (public) networks node might not have peers with private tx enabled among connected peers and private tx functionality doesn't work (because packets are not delivered among nodes).
The requirement to have nodes in reserved peers doesn't solve the issue, because we randomly select subset of connected peers for packets propagation and packet can be still lost.

In terms of this PR private tx enabled flag added into STATUS packet and node can explicitly select for propagation only nodes with private tx enabled.


rphmeier:
is it compatible with geth nodes who aren't aware of this packet?

arkpar:
All custom fields should only be added in the parity protocol.
The protocol version should be bumped if it wasn't bumped already since previous release.
Code should handle backwards compatibility with the previous version of the protocol. I.e. it should only add that field if both nodes use the latest protocol version and expect to read it only from nodes that are supporting the latest version.

ethcore/sync/src/api.rs
```diff
@@ -349,7 +349,7 @@ impl EthSync {
 		let sync = ChainSyncApi::new(
 			params.config,
 			&*params.chain,
-			params.private_tx_handler.clone(),
+			params.private_tx_handler.map(|handler| handler.clone()),
```
niklasad1:
```suggestion
			params.private_tx_handler.cloned(),
```

ethcore/sync/src/chain/handler.rs
```diff
@@ -654,9 +672,13 @@ impl SyncHandler {
 			trace!(target: "sync", "{} Ignoring packet from unconfirmed/unknown peer", peer_id);
 			return Ok(());
 		}
-
+		if sync.private_tx_handler.is_none() {
```
niklasad1: why not use pattern matching/if let to eliminate those `expect()` i.e, something like

```rust
let private_handler = match sync.private_tx_handler {
  Some(p) => p,
  None => return Ok(()), 
}
```

ethcore/sync/src/chain/mod.rs
```diff
@@ -1120,9 +1122,19 @@ impl ChainSync {
 	fn send_status(&mut self, io: &mut SyncIo, peer: PeerId) -> Result<(), network::Error> {
 		let warp_protocol_version = io.protocol_version(&WARP_SYNC_PROTOCOL_ID, peer);
 		let warp_protocol = warp_protocol_version != 0;
+		let private_tx_protocol = warp_protocol_version >= PAR_PROTOCOL_VERSION_3.0;
 		let protocol = if warp_protocol { warp_protocol_version } else { ETH_PROTOCOL_VERSION_63.0 };
 		trace!(target: "sync", "Sending status to {}, protocol version {}", peer, protocol);
-		let mut packet = RlpStream::new_list(if warp_protocol { 7 } else { 5 });
+		let rlp_len = if warp_protocol {
```
arkpar: Would be best to just use `begin_unbounded_list` and get rid of `rlp_len`

5chdn:
could you look at this again @niklasad1 ?

ethcore/sync/src/chain/propagator.rs
```diff
@@ -312,12 +312,16 @@ impl SyncPropagator {
...
 	pub fn propagate_private_transaction(sync: &mut ChainSync, io: &mut SyncIo, transaction_hash: H256, packet_id: PacketId, packet: Bytes) {
 		let lucky_peers = ChainSync::select_random_peers(&sync.get_private_transaction_peers(&transaction_hash));
-		trace!(target: "sync", "Sending private transaction packet to {:?}", lucky_peers);
-		for peer_id in lucky_peers {
-			if let Some(ref mut peer) = sync.peers.get_mut(&peer_id) {
-				peer.last_sent_private_transactions.insert(transaction_hash);
+		if lucky_peers.is_empty() {
+			error!(target: "privatetx", "Cannot propagate the packet, no peers with private tx enabled connected");
+		} else {
+			trace!(target: "sync", "Sending private transaction packet to {:?}", lucky_peers);
```
niklasad1: I'm a little confused on the `log target` would be nice to be consistent and use `privatetx` here? Or am I missing something?

grbIzl: Yeah, it looks weird indeed, I agree. I'll change it to privatetx for both cases

5chdn:
Thanks. Could you resolve the conflicts?

PR #10000, opened 2018-11-30
Fix daemonize

ordian:
Closes #9991. (Sorry, I screwed up in https://github.com/paritytech/parity-ethereum/pull/9946#pullrequestreview-176800230, but it didn't segfault for me)

Nice find by @andresilva:
> I don't know how rust handles file descriptors and forks
they're probably closed before forking?
https://github.com/rust-lang/rust/pull/24034
so this means any file descriptor is closed before forking/exec I think

The first commit moves fork instruction just before we open the db (so #9367 remains fixed).
The second commit switches `daemonize` to crates.io version (@tomaka is https://github.com/paritytech/daemonize/pull/1 no longer relevant (cf. https://github.com/knsd/daemonize/commit/4c5f9d0dccc17715ff847257c3b732406ef911c7#diff-49a4b4d69f811d5619c481802227b86e)?), it's somewhat unrelated, happy to revert if needed.

Issue #9367, opened 2018-08-17
When an account password file is not found and parity is run as a daemon it silently fails


XertroV:
> I'm running:
>
> - **Which Parity version?**: 2.0.1-beta
> - **Which operating system?**: Linux
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: no
> - **Which network are you connected to?**: custom poa
> - **Did you try to restart the node?**: yes

To test set 

```
[account]
password = ["/does-not-exist.txt"]
```

and run `parity --config config.toml daemon pidfile`, then run `parity --config config.toml` and you'll see the error.

PR #10003, opened 2018-11-30
move daemonize before creating account provider

ordian:
Follow-up for #10000, thanks @andresilva for spotting the issue.

The fetch service didn't work in #10000, since daemonize closed its open fds. We could've moved `daemonize` at the beginning, but we're trying to print as much errors (cf. #9367) as possible before starting a daemon.

### DIFF
--- a/parity/run.rs
+++ b/parity/run.rs
@@ -464,6 +464,12 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq: 
 
 	let passwords = passwords_from_files(&cmd.acc_conf.password_files)?;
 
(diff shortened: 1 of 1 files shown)


parity/run.rs
```diff
@@ -464,6 +464,12 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 
 	let passwords = passwords_from_files(&cmd.acc_conf.password_files)?;
 
+	// run in daemon mode
```
andresilva: Maybe let's add a small comment here regarding moving the `daemonize` call, just so we don't make the same mistake in the future.

Issue #8574, opened 2018-05-09
Make it possible to control logging in parity-clib


tomaka:
Right now the logs are always printed on stdout with no way of retrieving them or filtering them.

tomaka:
This is complicated because of the `parity_devLogs` RPC query which requires returning the logs.

tomaka:
Also there's a "conflict" with the CLI options that allow controlling the mode, color and file of the logging.

seunlanlege:
this should be closed with #10007

adria0:
Closing issue due to its stale state.

PR #10007, opened 2018-12-03
prevent silent errors in daemon mode

seunlanlege:
This PR replaces daemonize from crates.io, with a custom daemonize

The custom `daemonize` pipes the `daemon`'s STDOUT/STDERR to the parent process, after `parity-ethereum` successfully starts the daemon process detaches from the parent and the parent exists.

TODO:

- [x] Update Java bindings


niklasad1:
If we want to maintain this it should go in a separate repo preferable forked under the `parity org` as we have done with other forks such as `ring`, `rocksdb` etc

seunlanlege:
not sure why the c++ example isn't compiling @5chdn

parity/lib.rs
```diff
@@ -219,14 +220,18 @@ fn execute<Cr, Rr>(command: Execute, on_client_rq: Cr, on_updater_rq: Rr) -> Res
 /// binary.
 ///
 /// On error, returns what to print on stderr.
-pub fn start<Cr, Rr>(conf: Configuration, on_client_rq: Cr, on_updater_rq: Rr) -> Result<ExecutionAction, String>
+pub fn start<Cr, Rr>(
+	conf: Configuration,
+	logger: Arc<RotatingLogger>,
```
niklasad1: This breaks the C-bindings in `parity-clib` you need to update them accordingly

niklasad1:
@seunlanlege The changes you did `parity::start` breaks the C bindings and you need to add a logger object there both in `parity.h` and in `main.cpp`. I can help you if you are not familiar with C/C++

seunlanlege:
@niklasad1 please do, thanks.

seunlanlege:
moved the implementation to [here](https://github.com/paritytech/daemonize-rs)

ordian:
FYI: The naming convention for rust crates from the [api-guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html):
> Crate names should not use -rs or -rust as a suffix or prefix. Every crate is Rust! It serves no purpose to remind users of this constantly.

Could you remove `util/daemonize`, also clib logger API needs to be sorted out.

parity-clib/src/lib.rs
```diff
@@ -87,19 +92,42 @@ pub unsafe extern fn parity_config_destroy(cfg: *mut c_void) {
...
 
 #[no_mangle]
-pub unsafe extern fn parity_start(cfg: *const ParityParams, output: *mut *mut c_void) -> c_int {
+pub unsafe extern fn parity_start(cfg: *const ParityParams, logger: Logger, output: *mut *mut c_void) -> c_int {
 	panic::catch_unwind(|| {
 		*output = ptr::null_mut();
 		let cfg: &ParityParams = &*cfg;
 
+		let mode = {
+			if logger.mode_len == 0 {
```
niklasad1: Make sure no needless allocation is performed but after looking this up neither `slice::from_raw_parts` or `String::from_utf8` does allocate remove this!

parity-clib/src/lib.rs
```diff
@@ -87,19 +92,42 @@ pub unsafe extern fn parity_config_destroy(cfg: *mut c_void) {
...
+				None
+			} else {
+				let mode = slice::from_raw_parts(logger.mode as *const u8, logger.mode_len);
+				String::from_utf8(mode.to_owned()).ok()
+			}
+		};
+
+		let file = {
+			if logger.file_len == 0 {
```
niklasad1: remove this if too

seunlanlege:
@5chdn I can't sign in to gitlab, could you please restart the android build?

niklasad1:
@seunlanlege I restarted it but it looks like you need to bump `parity-daemonize` to get the latest changes

parity-clib/src/lib.rs
ordian: We're not using this check in `parity_config_from_cli`, I think it's safe since `mem::align_of::<u8> == 1` and also `String::from_utf8(slice.to_owned())` does not allocate if the `slice.is_empty()`.

parity/main.rs
```diff
@@ -184,11 +188,34 @@ fn main_direct(force_can_restart: bool) -> i32 {
 		parity_ethereum::Configuration::parse_cli(&args).unwrap_or_else(|e| e.exit())
 	};
 
+	// TODO: expose in the C API so that users can setup logging the way they want
```
ordian: This comment is now obsolete?
```suggestion
```

parity-clib/parity.h
```diff
@@ -37,6 +37,18 @@ struct ParityParams {
 	void *on_client_restart_cb_custom;
 };
 
+/// Logger options for `parity_start`
+struct Logger {
```
tomaka: Why not put the fields inside `ParityParams`?

parity-clib/parity.h
```diff
@@ -37,6 +37,18 @@ struct ParityParams {
...
+	/// Logger mode (Rust log format, see module docs for `ethcore-logger`
+	const char* mode;
```
tomaka: It should be documented that this can be `NULL` if `mode_len` is 0.

parity-clib/parity.h
```diff
@@ -37,6 +37,18 @@ struct ParityParams {
...
+	/// Length of the `Logger mode`
+	size_t mode_len;
+	/// File name to write the log to
+	const char* file;
```
tomaka: Same remark.

parity/lib.rs
tomaka: The difficult part of #8574 (and the reason why it wasn't done immediately when `parity-clib` was created) is that ideally the `RotatingLogger` wouldn't be exposed in the API.

However I guess we can ignore the problem in the name of "meh, who cares".

ordian: We're not exposing `RotatingLogger` in `C API`, we expose only `Logger` params struct.

tomaka: I mentioned "when parity-clib was created", because the Rust lib and the C library were created at the same time.

The logging system is unfortunately a global variable, and thus it shouldn't be the job of a library to set it.
The consequence is that if a program wants to use parity-as-a-library, they cannot setup their own logging compatible with the `log` crate.
An ideal world, it is `main.rs` that would setup the logging and that's it. The C library would have an independent function (not tied to parity, just like `parity_set_panic_hook` is totally independent as well) that sets up the Rust logging and explicitly mentions in its documentation that this applies to the `log` Rust crate as a whole.

parity-clib/src/lib.rs
niklasad1: Yeah, it is redundant thus it should be removed

However, when it comes to log-file we need the check because an empty string as `file name` will be a run-time error/panic

niklasad1:
@seunlanlege I can take care of the grumbles introduced by me.

niklasad1:
@tomaka @ordian I tried to unify the bindings a bit by adding the logger in the `ParityParams struct` and the build the logger in a separate function but it required an extra `from::raw` conversion.

Also it might be annoying to explicitly to call parity_setup_log when no logging is used otherwise the logger will be `NULL` and will end up in a `deref NULL`!

Thoughts?

To conclude this the `jni` bindings needs to updated too!

parity-clib/examples/cpp/main.cpp
```diff
@@ -53,10 +53,8 @@ const std::vector<std::string> ws_subscriptions {
 void callback(void* user_data, const char* response, size_t _len) {
 	Callback* cb = static_cast<Callback*>(user_data);
 	if (cb->type == CALLBACK_RPC) {
-		printf("rpc response: %s\r\n", response);
 		cb->counter -= 1;
 	} else if (cb->type == CALLBACK_WS) {
-		printf("websocket response: %s\r\n", response);
 		std::regex is_subscription ("\\{\"jsonrpc\":\"2.0\",\"result\":\"0[xX][a-fA-F0-9]{16}\",\"id\":1\\}");
```
niklasad1: Removed these printouts doesn't make sense anymore when these can be controlled via the `logger`

5chdn:
are you still working on this @seunlanlege ?

seunlanlege:
It's just the logger interface that's left on this, @tomaka @ordian could you look at this one more time?

parity-clib/parity.h
```diff
@@ -63,6 +66,33 @@ extern "C" {
...
+/// # Example
+///
+/// ```no_run
+/// void* cfg;
+/// const char *args[] = {"--light", "--can-restart"};
+/// size_t str_lens[] = {7, 13};
+/// if (parity_config_from_cli(args, str_lens, 2, &cfg) != 0) {
+///		return 1;
+/// }
+///	char[] logger_mode = "rpc=trace";
```
ordian:
```suggestion
/// char[] logger_mode = "rpc=trace";
```

parity-clib/parity.h
```diff
@@ -63,6 +66,33 @@ extern "C" {
...
+/// parity_set_logger(logger_mode, strlen(logger_mode), nullptr, 0, &cfg.logger)
```
ordian:
```suggestion
/// parity_set_logger(logger_mode, strlen(logger_mode), nullptr, 0, &cfg.logger);
```

parity/lib.rs
```diff
@@ -221,14 +223,21 @@ fn execute<Cr, Rr>(command: Execute, on_client_rq: Cr, on_updater_rq: Rr) -> Res
 /// binary.
 ///
 /// On error, returns what to print on stderr.
-pub fn start<Cr, Rr>(conf: Configuration, on_client_rq: Cr, on_updater_rq: Rr) -> Result<ExecutionAction, String>
-	where Cr: Fn(String) + 'static + Send,
-			Rr: Fn() + 'static + Send
+// FIXME: @tomaka - totally independent logging capability
```
ordian: Could you elaborate on what's missing?

parity-clib/parity.h
```diff
@@ -63,6 +66,33 @@ extern "C" {
...
+///
+///	- log_mode		: String representing the log mode according to `Rust LOG` or nullptr to disable logging.
+///					  See module documentation for `ethcore-logger` for more info.
+///	- log_mode_len	: Length of the log_mode or zero to disable logging
+///	- log_file		: String respresenting the file name to write to log to or nullptr to disable logging to a file
+///	- log_mode_len	: Length of the log_file or zero to disable logging to a file
+///	- logger		: Pointer to point to the created `Logger` object
+
+/// **Important**: This function must only be called exactly once otherwise it will panic. If you want to disable a
+///					logging mode or logging to a file make sure that you pass the `length` as zero
```
ordian:
```suggestion
/// logging mode or logging to a file make sure that you pass the `length` as zero
```

tomaka:
I'm fine with leaving it as it is, because doing it properly requires a lot of work.
I opened #10252.

parity/lib.rs
niklasad1: https://github.com/paritytech/parity-ethereum/issues/10252

Issue #9118, opened 2018-07-13
Alternative scoring implementation


tomusdrw:
## Why
Currently our scoring implementation is using a simple mapping of:
```
score = gas price
```
and every transaction is considered independently.

However a chain of transactions from a single sender:
```
Tx1(sender=X, nonce=N, gas_price=10)
Tx2(sender=X, nonce=N+1, gas_price=1_000)
Tx3(sender=X, nonce=N+2, gas_price=1_000_000)
```
is blocked by the score of the first transaction (we will never include `Tx1` if there are other transactions with `gas_price > 10` that can be included).
Including though `Tx1` would (most likely) allow us to include `Tx2` and `Tx3` and get the high fee that is payed there.

## How

The idea is to implement an alternative (experimental) scoring (should be controlled via CLI flag) that will bump a score of `Tx1` to reflect that there are more transactions in the chain.

Possible algorithm:
```rust
let last_index = txs.len() - 1;
score[last_index] = gas_price[last_index];
for i in last_index..1 {
   score[i - 1] = gas_price[i - 1]   // base score is always equal to gas price
         + is_consecutive(i - 1)     // only apply to consecutive nonces
         * 21_000 / gas[i - 1]       // "heavy" transactions are harder to bump
         * score[i]                  // Use the score of next transaction
         / 1_000                     // But require it to be order of magnitude greater
}

fn is_consecutive(i) {
 if nonce[i] + 1 == nonce[i+1] { 1 } else { 0 }
}
```

## Examples
Couple of examples of how it behaves:
```
# Simple transaction, increasing gas price

{"gp":1,"gas":21000,"score":3}
{"gp":1000,"gas":21000,"score":2000}
{"gp":1000000,"gas":21000,"score":1000000}
-----------------------
{"gp":1,"gas":21000,"score":2}
{"gp":1000,"gas":21000,"score":1000}
-----------------------
{"gp":1,"gas":21000,"score":2.001}
{"gp":1000,"gas":21000,"score":1001}
{"gp":1000,"gas":21000,"score":1000}

# Large transaction is harder to bump

{"gp":1,"gas":1000000,"score":1.021}
{"gp":1000,"gas":21000,"score":1000}
-----------------------
{"gp":1,"gas":1000000,"score":1.042}
{"gp":1000,"gas":21000,"score":2000}
{"gp":1000000,"gas":1000000,"score":1000000}
-----------------------
{"gp":1,"gas":1000000,"score":22}
[...]

PR #10008, opened 2018-12-04
Alternate Scoring

insipx:
WIP for #9118


miner/src/pool/scoring.rs
```diff
@@ -108,6 +204,16 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
 					super::Priority::Regular => 0,
 				};
 				scores[i] = scores[i] << boost;
+				let last_index = txs.len() - 1;
+				for idx in 1..last_index {
```
tomusdrw: Note the score depends on the next score (`score[idx]`) so we should compute it starting from the end

miner/src/pool/scoring.rs
```diff
@@ -108,6 +204,16 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+					scores[idx - 1] = *txs[idx-1].transaction.gas_price()
+						+ is_consecutive(&txs[idx-1..=idx])
+						+ (U256::from(21_000) / scores[idx - 1])
```
tomusdrw: I suppose this should be `gas` not `score`?

parity/cli/mod.rs
```diff
@@ -1346,6 +1347,7 @@ struct Mining {
 	refuse_service_transactions: Option<bool>,
 	infinite_pending_block: Option<bool>,
 	max_round_blocks_to_import: Option<usize>,
+	experimental_scoring: Option<bool>,
```
tomusdrw: not needed?

parity/helpers.rs
```diff
@@ -100,6 +100,7 @@ pub fn to_pending_set(s: &str) -> Result<PendingSet, String> {
 pub fn to_queue_strategy(s: &str) -> Result<PrioritizationStrategy, String> {
 	match s {
 		"gas_price" => Ok(PrioritizationStrategy::GasPriceOnly),
+                "consecutive" => Ok(PrioritizationStrategy::Consecutive),
```
tomusdrw: Indentation

miner/src/pool/scoring.rs
insipx: Yeahp, just didn't add gas() to `ScoredTransaction` yet, so used score as placeholder

miner/src/pool/scoring.rs
```diff
@@ -67,6 +72,7 @@ impl NonceAndGasPrice {
 	}
 }
 
+// differs in the `update_scores` function
```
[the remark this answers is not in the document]
insipx: Good point

I've combined them via the PrioritizationStrategy enum for now. However, I think it makes the update_scores function a bit too large imo, so will probably separate the two strategies out once i'm done testing the algorithm for consecutive transactions

5chdn:
Restarted CI

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,69 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+
+		let boost = |i: usize| {
+			match txs[i].priority() {
+				super::Priority::Local => 15,
+				super::Priority::Retracted => 10,
+				super::Priority::Regular => 0,
+			}
+		};
+
+		let event = |scores: &mut [U256]| {
```
tomusdrw: maybe call this `penalize`, it took me few seconds to figure out what this function is about (explanation comment on what `event` means is below)

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,69 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+				let is_consecutive = |i: usize| {
+					if (txs[i].nonce() + U256::one()) == txs[i+1].nonce() {
+						U256::one()
+					} else {
+						U256::zero()
 					}
+				};
+				match change {
+					Change::Culled(_) => {},
+					Change::RemovedAt(_) => {}
```
tomusdrw: So actually it's fine to ignore those events in `GasPriceOnly`, but with `Consecutive` strategy we should recalculate as well.

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,69 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+					Change::InsertedAt(i) | Change::ReplacedAt(i) => {
+						assert!(i < txs.len());
+						assert!(i < scores.len());
+						let last_index = txs.len() - 1;
+						scores[last_index] = *txs[last_index].transaction.gas_price();
+						for idx in (1..txs.len()).rev() {
+							let consecutive_bump = is_consecutive(idx - 1) *
```
tomusdrw: `idx - 1` is quite common, maybe do `let prev_idx = idx - 1`?

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,69 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+								((U256::from(21_000) * scores[idx]) /
+								txs[idx - 1].transaction.gas());
+							scores[idx - 1] = txs[idx - 1].transaction.gas_price() + (consecutive_bump / 1000);
+							scores[idx] <<= boost(idx - 1);
+						}
+						scores[0] <<= boost(last_index);
```
tomusdrw: I don't think this is necessary, we already cover that in case `idx == 1`.

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,69 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
 					}
+				};
+				match change {
+					Change::Culled(_) => {},
+					Change::RemovedAt(_) => {}
+					Change::InsertedAt(i) | Change::ReplacedAt(i) => {
+						assert!(i < txs.len());
+						assert!(i < scores.len());
+						let last_index = txs.len() - 1;
+						scores[last_index] = *txs[last_index].transaction.gas_price();
```
tomusdrw: I think you should include `boost` here

tomusdrw: Actually, currently `boost` seems to have exponential effect, can you instead apply it either:
1. Doing an additional loop after you calculate base scores (that will boost the scores, but boosting `idx+1` won't affect `idx`)
2. Instead of using `gas_price` use `gas_price` with boost applied.

I'm more inclined to option (2). Would be good to plot some graphs how they differ.

parity/cli/mod.rs
tomusdrw: I think it's still not needed:)

parity/cli/mod.rs
```diff
@@ -761,9 +761,10 @@ usage! {
 			"--tx-queue-per-sender=[LIMIT]",
 			"Maximum number of transactions per sender in the queue. By default it's 1% of the entire queue, but not less than 16.",
 
+			// INSI: Not sure about the name for this strategy -- 'dynamic'?
 			ARG arg_tx_queue_strategy: (String) = "gas_price", or |c: &Config| c.mining.as_ref()?.tx_queue_strategy.clone(),
 			"--tx-queue-strategy=[S]",
-			"Prioritization strategy used to order transactions in the queue. S may be: gas_price - Prioritize txs with high gas price",
+			"Prioritization strategy used to order transactions in the queue. S may be: gas_price - Prioritize txs with high gas price, consecutive - Prioritize consecutive transactions. Score such that the first transaction has a higher priority to reflect more transactions after it",
```
tomusdrw: I'd rephrase it a bit:
Include the entire chain of transactions when calculating their score. It means that one can increase priority of a transaction in the pool by sending next transaction after it.

miner/src/pool/scoring.rs
insipx: This is something I considered doing before but wasn't sure if it was necessary, plotting all the different algorithms now! :+1:

parity/cli/mod.rs
insipx: changed it to `Include the entire chain of transactions when calculating their score. This means one may increase the priority of a transaction in the pool by sending another transaction from the same account`

miner/src/pool/scoring.rs
tomusdrw: Seems that the 3rd graph has a different scale.

Note that we onyl care about transactions above 21k (base gas), would be good to plot starting from there.
Do you mind plotting all options on the same chart and add a line for regular scoring as well (i.e. the one with gas price only?)

miner/src/pool/scoring.rs
tomusdrw: Still valid?

5chdn:
does not compile @insipx

miner/src/pool/scoring.rs
insipx: Merged all the graphs into two, one where Gas price was kept constant and another where Gas was kept at a constant. 


Legend for all the graphs:
Algorithm from PR:  Black,
Option 1: Blue
Option 2: Red
GasPriceOnly Scoring:  Green

## Constant Gas Price. Gas price kept at 10,000

Constant Gas Price

#### Same graph, but with an extended X-Axis
Constant Gas Price

## Constant Gas. Gas kept at 21,000
Constant Gas

5chdn:
Still seems to fail

grbIzl:
@insipx as far as I've understood, the changes do not cover the attack with clearing all the balance (described in the original proposal). Am I correct? Do we have plans to address it later?

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,67 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
-				for (score, tx) in scores.iter_mut().zip(txs) {
-					// Never penalize local transactions.
-					if !tx.priority().is_local() {
-						*score = *score >> 3;
+			PrioritizationStrategy::Consecutive => {
+				let is_consecutive = |i: usize| {
+					if (txs[i].nonce() + U256::one()) == txs[i+1].nonce() {
+						U256::one()
+					} else {
+						U256::zero()
```
ngotchac: Are the transactions sorted here? What happens if I send consecutive transactions in the wrong order?

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,67 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
-			// So just lower the priority of all non-local transactions.
-			Change::Event(_) => {
-				for (score, tx) in scores.iter_mut().zip(txs) {
-					// Never penalize local transactions.
-					if !tx.priority().is_local() {
-						*score = *score >> 3;
+			PrioritizationStrategy::Consecutive => {
+				let is_consecutive = |i: usize| {
+					if (txs[i].nonce() + U256::one()) == txs[i+1].nonce() {
+						U256::one()
```
ngotchac: Why not directly return the `consecutive_bump` here?

miner/src/pool/scoring.rs
insipx: this part of the code should be dealing with sorted transactions. As I understand it, the sorting/ordering of transactions occurs in the `compare`, `choose` and `should_replace` methods as defined by the `Scoring` trait.

miner/src/pool/scoring.rs
```diff
@@ -92,33 +92,67 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
...
+					} else {
+						U256::zero()
 					}
+				};
+				match change {
+					Change::Culled(_) => {},
+					Change::RemovedAt(i) | Change::InsertedAt(i) | Change::ReplacedAt(i) => {
+						assert!(i < txs.len());
+						assert!(i < scores.len());
```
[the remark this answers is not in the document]
insipx: Nope, it's not. Fixed that in my latest commit

dvdplm:
@insipx I'm going to put this on ice for the time being. When you're ready for review again, just holler ok? :)

debris:
closing as stale

Issue #9978, opened 2018-11-28
panic on proof validation 


ngotiendatdinosys:
_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: Parity-Ethereum/v2.2.1-beta-5c56fc5-20181114/x86_64-macos/rustc1.30.1
- **Operating system**: MacOS
- **Installation**: one-line installer
- **Fully synchronized**: no
- **Network**: private
- **Restarted**: no

<img width="639" alt="screen shot 2018-11-28 at 10 10 51 am" src="https://user-images.githubusercontent.com/26533319/49126438-261b4c80-f2f6-11e8-9108-f53fb9736df9.png">

5chdn:
Can you share your chain specification?

adria0:
Closing the issue due to its stale state

PR #10015, opened 2018-12-05
Properly handle check_epoch_end_signal errors

sorpaas:
The function `check_epoch_end_signal` can fail, but currently its errors are silently ignored. This can cause other parts of engine to panic, and gets us issues like #9978. This PR moves `check_epoch_end_signal` out of `commit_block` so that its errors are properly handled.

When on error, the block is rejected. Nearly all times this is users' specification issue. With rejection, no further blocks will proceed thus allowing the user to rescue the chain without the need to kill the db.


5chdn:
Conflict in client, and needs a 2nd review :+1:

ordian:
@sorpaas please resolve conflicts

5chdn:
Does not compile

sorpaas:
Sorry I missed this!

PR #10017, opened 2018-12-05
ci: move future releases to ethereum subdir on s3

5chdn:
- move binaries from releases.parity.io to releases.parity.io/ethereum
- allow for future products to be released on same domain, e.g. releases.parity.io/substrate
- ~~only affects future binaries~~ will sync over old releases manually
- on ice: vanity service needs a patch before we release this


5chdn:
Yes, this needs more thoughts, give me a second.

PR #10031, opened 2018-12-07
ethcore: enable constantinople on ethereum

5chdn:
block 7080000, ref https://github.com/ethereum/pm/issues/64


5chdn:
please release this week as miners need to upgrade in time, ideally before xmas

PR #10033, opened 2018-12-07
First attempt at threshold signatures

DemiMarie:
This doesn’t even compile at the moment, but I would like to know if it is on the right track.


5chdn:
It seems you want feedback on 1e3736d3a96b6dde9ed9ad97cd9e0bd7f785c42e but apparently you pulled in the entire changes from the HBBFT integration in POA, do you mind rebasing your PR on master?

Issue #9770, opened 2018-10-17
Ropsten stuck at #4242985 block


alexdupre:
- **Parity Ethereum version**: 2.0.8
- **Operating system**: FreeBSD
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: ropsten
- **Restarted**: yes

Even after the upgrade from 2.0.7 to 2.0.8 release the block syncing is stuck at block #4242985.
What am I expected to do to recover from the Constantinople fork?

5chdn:
`parity db kill` and resync with `parity --chain ropsten --no-warp`

Issue #9726, opened 2018-10-10
Parity stuck at Block 310191


donamk:
_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: Parity-Ethereum/v2.1.1-beta-cb09330-20180919/x86_64-linux-gnu/rustc1.29.0
- **Operating system**: Linux
- **Installation**: one-line installer
- **Fully synchronized**:  yes
- **Network**: ethereum
- **Restarted**: yes

just download parity using one-line installer and set up like as following
```
[parity]
mode = "active"

[footprint]
tracing = "on"
db_compaction = "ssd"
pruning = "archive"
fat_db = "on"
```
However parity stuck at Block 310191 for 12 hours.

Here is my parity log tail -n 1000
```
2018-10-10 06:20:22  Syncing  #310191 0x8ab5…9e45     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed    #14732   36/50 peers      2 MiB chain    6 MiB db  0 bytes queue   54 MiB sync  RPC:  0 conn,    5 req/s,   52 µs
2018-10-10 06:20:32  Syncing  #310191 0x8ab5…9e45     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed    #28831   36/50 peers      2 MiB chain    6 MiB db  0 bytes queue   54 MiB sync  RPC:  0 conn,    6 req/s,   58 µs
2018-10-10 06:20:37  Syncing  #310191 0x8ab5…9e45     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed    #37210   36/50 peers      2 MiB chain    6 MiB db  0 bytes queue   54 MiB sync  RPC:  0 conn,    5 req/s,   58 µs
2018-10-10 06:20:42  Syncing  #310191 0x8ab5…9e45     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed    #44449   36/50 peers      2 MiB chain    6 MiB db  0 bytes queue   54 MiB sync  RPC:  0 conn,    5 req/s,   55 µs
2018-10-10 06:20:47  Syncing  #310191 0x8ab5…9e45     0.00 blk/s    0.0 tx/s    0 Mgas/s      0+    0 Qed    #25393   36/50 peers      2 MiB chain    6 MiB db  0 bytes queue   54 MiB sync  RPC:  0 conn,    3 req/s,   55 µs


...


[...]

PR #10042, opened 2018-12-10
Don't try to send oversized packets

ngotchac:
It seems that when answering a `get_block_bodies`, `get_block_headers`, etc. request, we only check the size of the response payload when sending it, instead of when building it. This means that if a peer asks for 100 big (meaning w/ a lot of transactions) blocks, we would construct the payload of the 100 block bodies even though it will never be sent because it would exceed the packet size.

With this PR, we stop adding blocks to the response when we know the size will exceed a certain limit.

For now, this limit is lower than the official protocol packet limit (set to 2MB), but if needed this can be changed.

I think it would solve these issues: #9770 and #9726 which would occur when a node is connected to only Parity nodes and is asking for big blocks (and thus never receive any answer).

When this PR is merged, I think that the next step would be to increase the number of blocks we ask our peers, since they shouldn't stop when the answer gets too big (Geth already does this).


ethcore/sync/src/chain/supplier.rs
```diff
@@ -238,6 +249,10 @@ impl SyncSupplier {
 			if let Some(node) = io.chain().state_data(&r.val_at::<H256>(i)?) {
 				data.push(node);
 				added += 1;
+				// Check that the packet won't be oversized
+				if data.len() > payload_soft_limit {
```
tomusdrw: This seems incorrect - it's just a vector of state data objects, not bytes.

Maybe worth computing RLP inline and checking the actual size (maybe every `N` items to avoid unecessary work)? It seems that `begin_list(len)` and `begin_unbounded_list()` is exactly the same - the only difference is that you need to `complete_unbounded_list` and it will panic if you exceed declared size (TBH the bounded list API seems completely useless, CC @debris).

ethcore/sync/src/chain/supplier.rs
```diff
@@ -266,6 +282,10 @@ impl SyncSupplier {
 				added_receipts += receipts_bytes.len();
 				added_headers += 1;
 				if added_receipts > MAX_RECEIPTS_TO_SEND { break; }
+				// Check that the packet won't be oversized
+				if data.len() > payload_soft_limit {
```
tomusdrw: We already check `MAX_RECEIPTS_TO_SEND` which actually seems to operate on bytes not the receitps (?)

ethcore/sync/src/chain/supplier.rs
```diff
@@ -215,6 +221,10 @@ impl SyncSupplier {
 			if let Some(body) = io.chain().block_body(BlockId::Hash(r.val_at::<H256>(i)?)) {
 				data.append(&mut body.into_inner());
 				added += 1;
+				// Check that the packet won't be oversized
+				if data.len() > payload_soft_limit {
```
tomusdrw: Same as with headers, not sure how reliable is `RlpStream::len`

ethcore/sync/src/chain/supplier.rs
```diff
@@ -182,6 +183,10 @@ impl SyncSupplier {
 			} else if let Some(hdr) = io.chain().block_header(BlockId::Number(number)) {
 				data.append(&mut hdr.into_inner());
 				count += 1;
+				// Check that the packet won't be oversized
+				if data.len() > payload_soft_limit {
```
tomusdrw: Why not sum up the raw header bytes that we append to RLP a line earlier?

ethcore/sync/src/chain/supplier.rs
ngotchac: Yes you're right, sorry.

ethcore/sync/src/chain/supplier.rs
ngotchac: Here `data` is of type `Bytes`

ethcore/sync/src/chain/supplier.rs
ngotchac: Do you think it would be more efficient?

ethcore/sync/src/chain/supplier.rs
```diff
@@ -266,6 +282,10 @@ impl SyncSupplier {
 				added_receipts += receipts_bytes.len();
 				added_headers += 1;
 				if added_receipts > MAX_RECEIPTS_TO_SEND { break; }
```
tomusdrw: Can we get rid of this? It seems redudant now.

ngotchac: Acutally, I kept `MAX_RECEIPTS_TO_SEND` which as you pointed is already in bytes (cf. https://github.com/paritytech/parity-ethereum/pull/10042#discussion_r240349764).

tomusdrw: Right, by do you know why do we have this limit in the first place? Maybe let's use soft limit instead?

ngotchac: Yeah I don't know either, let me do that.

ethcore/sync/src/chain/supplier.rs
andresilva: Nitpick: this won't include the size of the fork header. we can move this to just before the `if reverse`.

PR #10047, opened 2018-12-11
Backports for beta 2.2.5

Tbaut:
- [x] Bump beta to 2.2.5
- [x] Fix empty steps #9939
- [x] Strict empty steps validation #10041
- [x] enable constantinople on ethereum #10031
- [x] Change test miner max memory to support malloc allocator #10024
- [x] Fix test: corpus_inaccessible (panic) #10019

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2162,7 +2162,7 @@ version = "1.12.0"
 dependencies = [
  "jni 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "panic_hook 0.1.0",
- "parity-ethereum 2.2.4",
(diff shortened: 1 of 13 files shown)


niklasad1:
@tomusdrw shall we backport the `cross-beam` fix?

tomusdrw:
@niklasad1 I believe so, it will probably fix the audit build.

5chdn:
Yes, let's backport that to beta

andresilva:
#9939 and #10041 should be mentioned in the release notes (for stable as well) since they introduce new behavior and users must update their chain specs if they want to keep the existing behavior.

PR #10050, opened 2018-12-11
Update parking_lot to 0.7

tomaka:


niklasad1:
We should also bump `parking_lot` in:
* jsonrpc
* kvdb-memorydb
* kvdb-rocksdb

PR #10051, opened 2018-12-11
import rpc transactions sequentially

seunlanlege:
Previously the `ProspectiveSigner` marks nonces as used even though they weren't successfully imported into the transaction queue
which caused issues like: #9040

This PR auguments the nonce reservation system such that a reserved nonce is only marked as used if
the `PostSign` action executes successfully eg transaction is successfully imported into the tx queue

### DIFF
--- a/rpc/src/v1/helpers/dispatch.rs
+++ b/rpc/src/v1/helpers/dispatch.rs
@@ -45,7 +45,7 @@ use jsonrpc_core::futures::{future, Future, Poll, Async};
 use jsonrpc_core::futures::future::Either;
 use v1::helpers::{errors, nonce, TransactionRequest, FilledTransactionRequest, ConfirmationPayload};
 use v1::types::{
-	H256 as RpcH256, H520 as RpcH520, Bytes as RpcBytes,
+	H520 as RpcH520, Bytes as RpcBytes,
 	RichRawTransaction as RpcRichRawTransaction,
 	ConfirmationPayload as RpcConfirmationPayload,
 	ConfirmationResponse,
@@ -56,6 +56,7 @@ use v1::types::{
 use rlp;
 
 pub use self::nonce::Reservations;
+use v1::types::TransactionCondition;
 
 /// Has the capability to dispatch, sign, and decrypt.
 ///
@@ -70,8 +71,9 @@ pub trait Dispatcher: Send + Sync + Clone {
 		-> BoxFuture<FilledTransactionRequest>;
 
 	/// Sign the given transaction request without dispatching, fetching appropriate nonce.
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
-		-> BoxFuture<WithToken<SignedTransaction>>;
+	fn sign<T>(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith, post_sign: T)
+		-> BoxFuture<WithToken<SignedTransaction>>
+		where T: 'static + PostSign;
 
 	/// Converts a `SignedTransaction` into `RichRawTransaction`
 	fn enrich(&self, SignedTransaction) -> RpcRichRawTransaction;
@@ -164,8 +166,9 @@ impl<C: miner::BlockChainClient + BlockChainClient, M: MinerService> Dispatcher 
 		}))
 	}
 
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
+	fn sign<T>(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith, post_sign: T)
 		-> BoxFuture<WithToken<SignedTransaction>>
+		where T: 'static + PostSign
 	{
 		let chain_id = self.client.signing_chain_id();
 
@@ -176,7 +179,10 @@ impl<C: miner::BlockChainClient + BlockChainClient, M: MinerService> Dispatcher 
 		let state = self.state_nonce(&filled.from);
 		let reserved = self.nonces.lock().reserve(filled.from, state);
 
-		Box::new(ProspectiveSigner::new(accounts, filled, chain_id, reserved, password))
+		Box::new(ProspectiveDispatcher {
+			signer: ProspectiveSigner::new(accounts, filled, chain_id, reserved, password),
+			post_sign: Box::new(post_sign)
+		})
 	}
 
 	fn enrich(&self, signed_transaction: SignedTransaction) -> RpcRichRawTransaction {
@@ -396,8 +402,9 @@ impl Dispatcher for LightDispatcher {
 		}))
 	}
 
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
+	fn sign<T>(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith, _: T)
 		-> BoxFuture<WithToken<SignedTransaction>>
+		where T: 'static + PostSign
 	{
 		let chain_id = self.client.signing_chain_id();
 		let nonce = filled.nonce.expect("nonce is always provided; qed");
@@ -464,6 +471,66 @@ struct ProspectiveSigner {
 	ready: Option<nonce::Ready>,
 }
 
+/// action to execute after signing
+/// e.g importing a transaction into the chain
+pub trait PostSign: Send {
+	/// perform an action with the signed transaction
+	fn execute(&mut self, signer: SignedTransaction) -> Result<()>;
+}
+
+/// performs no action in PostSign
+pub struct NoopPostSign;
+
+impl PostSign for NoopPostSign {
+	fn execute(&mut self, _: SignedTransaction) -> Result<()> {
+		Ok(())
+	}
+}
+
+/// thsis imports the signed transaction in PostSign
+pub struct ImportTransactionPostSign<D: Dispatcher> {
+	dispatcher: D,
+	condition: Option<TransactionCondition>
+}
+
+impl<D: Dispatcher> ImportTransactionPostSign<D> {
+	/// given a dispatcher and a transaction condition
+	/// contructs a PostSign trait object that will import the signed transaction
+	pub fn new(dispatcher: D, condition: Option<TransactionCondition>) -> Self {
+		ImportTransactionPostSign {
+			dispatcher,
+			condition
+		}
+	}
+}
+
+impl<D: Dispatcher> PostSign for ImportTransactionPostSign<D> {
+	fn execute(&mut self, signed: SignedTransaction) -> Result<()> {
+		let condition = self.condition.take().map(Into::into);
+		self.dispatcher.dispatch_transaction(PendingTransaction::new(signed, condition))?;
+		Ok(())
+	}
+}
+
+struct ProspectiveDispatcher {
+	signer: ProspectiveSigner,
+	post_sign: Box<PostSign + Send>
+}
+
+impl Future for ProspectiveDispatcher {
+	type Item = WithToken<SignedTransaction>;
+	type Error = Error;
+
+	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		let (signed, ready) = try_ready!(self.signer.poll());
+		if let Err(e) = self.post_sign.execute((*signed).clone()) {
+			return Err(e)
+		}
+		ready.mark_used();
+		return Ok(Async::Ready(signed))
+	}
+}
+
 impl ProspectiveSigner {
 	pub fn new(
 		accounts: Arc<AccountProvider>,
@@ -510,7 +577,7 @@ impl ProspectiveSigner {
 }
 
 impl Future for ProspectiveSigner {
-	type Item = WithToken<SignedTransaction>;
+	type Item = (WithToken<SignedTransaction>, nonce::Ready);
 	type Error = Error;
 
 	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
@@ -546,8 +613,7 @@ impl Future for ProspectiveSigner {
 					if let (Some(result), Some(nonce)) = (self.prospective.take(), self.ready.take()) {
 						// Mark nonce as used on successful signing
 						return result.map(move |tx| {
-							nonce.mark_used();
-							Async::Ready(tx)
+							Async::Ready((tx, nonce))
 						})
 					} else {
 						panic!("Poll after ready.");
@@ -654,20 +720,15 @@ pub fn execute<D: Dispatcher + 'static>(
 ) -> BoxFuture<WithToken<ConfirmationResponse>> {
 	match payload {
 		ConfirmationPayload::SendTransaction(request) => {
-			let condition = request.condition.clone().map(Into::into);
-			Box::new(dispatcher.sign(accounts, request, pass)
-				.map(move |v| v.map(move |tx| PendingTransaction::new(tx, condition)))
-				.map(WithToken::into_tuple)
-				.map(|(tx, token)| (tx, token, dispatcher))
-				.and_then(|(tx, tok, dispatcher)| {
-					dispatcher.dispatch_transaction(tx)
-						.map(RpcH256::from)
-						.map(ConfirmationResponse::SendTransaction)
-						.map(move |h| WithToken::from((h, tok)))
-				}))
+			let post_sign = ImportTransactionPostSign::new(dispatcher.clone(), request.condition.clone());
+			Box::new(dispatcher.sign(accounts, request, pass, post_sign)
+				.map(|with_token| {
+					with_token.map(|tx| ConfirmationResponse::SendTransaction(tx.hash().into()))
+				})
+			)
 		},
 		ConfirmationPayload::SignTransaction(request) => {
-			Box::new(dispatcher.sign(accounts, request, pass)
+			Box::new(dispatcher.sign(accounts, request, pass, NoopPostSign)
 				.map(move |result| result
 					.map(move |tx| dispatcher.enrich(tx))
 					.map(ConfirmationResponse::SignTransaction)
--- a/rpc/src/v1/helpers/dispatch.rs
+++ b/rpc/src/v1/helpers/dispatch.rs
@@ -52,11 +52,11 @@ use v1::types::{
 	EthSignRequest as RpcEthSignRequest,
 	EIP191SignRequest as RpcSignRequest,
 	DecryptRequest as RpcDecryptRequest,
+	TransactionCondition
 };
 use rlp;
 
 pub use self::nonce::Reservations;
-use v1::types::TransactionCondition;
 
 /// Has the capability to dispatch, sign, and decrypt.
 ///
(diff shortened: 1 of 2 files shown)


rpc/src/v1/helpers/dispatch.rs
```diff
@@ -56,6 +56,7 @@ use v1::types::{
 use rlp;
 
 pub use self::nonce::Reservations;
+use v1::types::TransactionCondition;
```
seunlanlege: should be grouped with the others

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -464,6 +471,66 @@ struct ProspectiveSigner {
 	ready: Option<nonce::Ready>,
 }
 
+/// action to execute after signing
+/// e.g importing a transaction into the chain
+pub trait PostSign: Send {
```
seunlanlege: shouldn't this be `dyn Trait`?

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -464,6 +471,66 @@ struct ProspectiveSigner {
...
+
+impl<D: Dispatcher> PostSign for ImportTransactionPostSign<D> {
+	fn execute(&mut self, signed: SignedTransaction) -> Result<()> {
+		let condition = self.condition.take().map(Into::into);
+		self.dispatcher.dispatch_transaction(PendingTransaction::new(signed, condition))?;
+		Ok(())
+	}
+}
+
+struct ProspectiveDispatcher {
```
seunlanlege: feel like this should be `WithPostSign`

seunlanlege:
There should be some tests that affirm this new behaviour, just how to design them properly :thinking:

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -464,6 +483,66 @@ struct ProspectiveSigner {
...
+			dispatcher,
+			condition
+		}
+	}
+}
+
+impl<D: Dispatcher> PostSign for ImportTransactionPostSign<D> {
+	fn execute(&mut self, signed: SignedTransaction) -> Result<()> {
+		let condition = self.condition.take().map(Into::into);
+		self.dispatcher.dispatch_transaction(PendingTransaction::new(signed, condition))?;
```
tomusdrw: I don't like the logic is now splitted here from the main file.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -396,8 +410,13 @@ impl Dispatcher for LightDispatcher {
...
 	}
 
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
-		-> BoxFuture<WithToken<SignedTransaction>>
+	fn sign(
+		&self,
+		accounts: Arc<AccountProvider>,
+		filled: FilledTransactionRequest,
+		password: SignWith,
+		_: impl PostSign + 'static
```
tomusdrw: Why is post sign ignored for light clients? That's completely breaking sending transactions.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -464,6 +483,66 @@ struct ProspectiveSigner {
...
 
+/// action to execute after signing
+/// e.g importing a transaction into the chain
+pub trait PostSign: Send + Sync {
+	/// perform an action with the signed transaction
+	fn execute(&mut self, signer: SignedTransaction) -> Result<()>;
+}
+
+/// performs no action in PostSign
+pub struct NoopPostSign;
```
tomusdrw: Maybe just implement for `()`

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -523,32 +586,62 @@ impl Future for ProspectiveSigner {
...
-						return result.map(move |tx| {
-							nonce.mark_used();
-							Async::Ready(tx)
-						})
+						let mut post_sign_future = self.post_sign.take()
+							.expect("post_sign is set on creation; qed")
+							.execute(result)
+							.into_future();
+
+						match post_sign_future.poll()? {
```
tomusdrw: Can we just populate `self.post_sign_future` and then `loop` again so that the logic doesn't need to be duplicated?

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -523,32 +586,62 @@ impl Future for ProspectiveSigner {
...
+								nonce.mark_used();
+								return Ok(Async::Ready(item))
+							},
+						}
+					} else {
+						panic!("Poll after ready.");
+					}
+				},
+				WaitForPostSign => {
+					if let Some(mut fut) = self.post_sign_future.take() {
```
tomusdrw: I think you don't have to `.take()` it and then put back, maybe just `as_mut()` ? That would allow you to do `try_ready!(fut.poll())` I suppose.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -449,28 +477,61 @@ fn sign_transaction(
...
 	accounts: Arc<AccountProvider>,
 	filled: FilledTransactionRequest,
 	chain_id: Option<u64>,
 	reserved: nonce::Reserved,
 	password: SignWith,
 	state: ProspectiveSignerState,
-	prospective: Option<Result<WithToken<SignedTransaction>>>,
+	prospective: Option<WithToken<SignedTransaction>>,
 	ready: Option<nonce::Ready>,
+	post_sign: Option<P>,
```
tomusdrw: I know that it's aligned with how currently the state is passed in the `ProspectiveSigner`, but we might consider refactoring this in a way that the required items are passed in the `ProspectiveSignerState` enum instead to avoid so many `Option<` and `.expects()`.
Not required for this PR (or not required at all), unless someone wants to pick it up :)

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -449,28 +477,61 @@ fn sign_transaction(
...
+impl PostSign for () {
+	type Item = WithToken<SignedTransaction>;
+	type Out = Result<Self::Item>;
+	fn execute(self, signed: WithToken<SignedTransaction>) -> Self::Out {
+		Ok(signed)
+	}
 }
 
-impl ProspectiveSigner {
+impl<F: Send + Sync, T: Send + Sync> PostSign for F
```
tomusdrw: I think `Sync` bound are not really required.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -449,28 +477,61 @@ fn sign_transaction(
...
-	prospective: Option<Result<WithToken<SignedTransaction>>>,
+	prospective: Option<WithToken<SignedTransaction>>,
 	ready: Option<nonce::Ready>,
+	post_sign: Option<P>,
+	post_sign_future: Option<<<P as PostSign>::Out as IntoFuture>::Future>
+}
+
+/// action to execute after signing
+/// e.g importing a transaction into the chain
+pub trait PostSign: Send + Sync {
```
tomusdrw: Is `Sync` required here? It can be called only once anyway.

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -655,19 +748,21 @@ pub fn execute<D: Dispatcher + 'static>(
...
-					dispatcher.dispatch_transaction(tx)
-						.map(RpcH256::from)
-						.map(ConfirmationResponse::SendTransaction)
-						.map(move |h| WithToken::from((h, tok)))
-				}))
+			let cloned_dispatcher = dispatcher.clone();
+			let post_sign = move |with_token_signed: WithToken<SignedTransaction>| {
+				let (signed, token) = with_token_signed.into_tuple();
+				let signed_transaction = PendingTransaction::new(signed, condition);
+				cloned_dispatcher.dispatch_transaction(signed_transaction)
```
tomusdrw: Do we really have to clone the dispatcher? Wouldn't it work (without `move`) if you just pass it as is?

rpc/src/v1/impls/personal.rs
```diff
@@ -223,18 +229,26 @@ impl<D: Dispatcher + 'static> Personal for PersonalClient<D> {
...
 	fn send_transaction(&self, meta: Metadata, request: TransactionRequest, password: String) -> BoxFuture<RpcH256> {
-		Box::new(self.do_sign_transaction(meta, request, password)
-			.and_then(|(pending_tx, dispatcher)| {
-				let chain_id = pending_tx.chain_id();
-				trace!(target: "miner", "send_transaction: dispatching tx: {} for chain ID {:?}",
-					::rlp::encode(&*pending_tx).pretty(), chain_id);
-
-				dispatcher.dispatch_transaction(pending_tx).map(Into::into)
+		let condition = request.condition.clone().map(Into::into);
+		let dispatcher = self.dispatcher.clone();
```
tomusdrw: Possibly same as with the other `PostSign`

rpc/src/v1/impls/personal.rs
```diff
@@ -223,18 +229,26 @@ impl<D: Dispatcher + 'static> Personal for PersonalClient<D> {
...
+		Box::new(self.do_sign_transaction(meta, request, password,  move |signed: WithToken<SignedTransaction>| {
+			dispatcher.dispatch_transaction(
```
tomusdrw: I think we lost `trace` during refactoring, but I think it's fine :)

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -69,12 +69,19 @@ pub trait Dispatcher: Send + Sync + Clone {
...
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
-		-> BoxFuture<WithToken<SignedTransaction>>;
+	/// Sign the given transaction request, fetching appropriate nonce and executing the PostSign action
+	fn sign<P>(
+		&self,
+		accounts: Arc<AccountProvider>,
+		filled: FilledTransactionRequest,
+		password: SignWith,
+		post_sign: P
+	) -> BoxFuture<<P as PostSign>::Item>
```
niklasad1: should not need typecast here
 
```suggestion
	) -> BoxFuture<P::Item>
```

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -69,12 +69,19 @@ pub trait Dispatcher: Send + Sync + Clone {
...
+		where P: PostSign + 'static,
+		      <<P as PostSign>::Out as futures::future::IntoFuture>::Future: Send;
```
niklasad1:
```suggestion
		      <P::Out as futures::future::IntoFuture>::Future: Send;
```

niklasad1: nit spaces

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -164,19 +171,29 @@ impl<C: miner::BlockChainClient + BlockChainClient, M: MinerService> Dispatcher
...
 	}
 
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
-		-> BoxFuture<WithToken<SignedTransaction>>
+	fn sign<P>(
+		&self,
+		accounts: Arc<AccountProvider>,
+		filled: FilledTransactionRequest,
+		password: SignWith,
+		post_sign:  P
```
niklasad1: nit extra space

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -164,19 +171,29 @@ impl<C: miner::BlockChainClient + BlockChainClient, M: MinerService> Dispatcher
...
+				.and_then(move |signed| post_sign.execute(signed));
+			return Box::new(future);
 		}
 
 		let state = self.state_nonce(&filled.from);
 		let reserved = self.nonces.lock().reserve(filled.from, state);
 
-		Box::new(ProspectiveSigner::new(accounts, filled, chain_id, reserved, password))
+		Box::new(ProspectiveSigner::new(accounts, filled, chain_id, reserved, password, post_sign))
 	}
```
niklasad1: Not really introduced by this PR but why not if else and remove the explicit `return`?

I argue that it is easier to read but in this case optional because this method is simple!

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -396,12 +413,23 @@ impl Dispatcher for LightDispatcher {
 		}))
 	}
 
-	fn sign(&self, accounts: Arc<AccountProvider>, filled: FilledTransactionRequest, password: SignWith)
-		-> BoxFuture<WithToken<SignedTransaction>>
+	fn sign<P>(
```
niklasad1: missing space

```suggestion
	fn sign<P> (
```

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -396,12 +413,23 @@ impl Dispatcher for LightDispatcher {
...
+		&self,
+		accounts: Arc<AccountProvider>,
+		filled: FilledTransactionRequest,
+		password: SignWith,
+		post_sign: P
+	) -> BoxFuture<<P as PostSign>::Item>
```
niklasad1:
```suggestion
	) -> BoxFuture<P::Item>
```

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -396,12 +413,23 @@ impl Dispatcher for LightDispatcher {
...
+		where P: PostSign + 'static,
+		      <<P as PostSign>::Out as futures::future::IntoFuture>::Future: Send
```
niklasad1:
```suggestion
		      <P::Out as futures::future::IntoFuture>::Future: Send
```

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -449,28 +477,61 @@ fn sign_transaction(
...
 	filled: FilledTransactionRequest,
 	chain_id: Option<u64>,
 	reserved: nonce::Reserved,
 	password: SignWith,
 	state: ProspectiveSignerState,
-	prospective: Option<Result<WithToken<SignedTransaction>>>,
+	prospective: Option<WithToken<SignedTransaction>>,
 	ready: Option<nonce::Ready>,
+	post_sign: Option<P>,
+	post_sign_future: Option<<<P as PostSign>::Out as IntoFuture>::Future>
```
niklasad1:
```suggestion
	post_sign_future: Option<<P::Out as IntoFuture>::Future>
```

rpc/src/v1/helpers/dispatch.rs
niklasad1: Yeah, I think you are right

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -509,8 +572,8 @@ impl ProspectiveSigner {
 	}
 }
 
-impl Future for ProspectiveSigner {
-	type Item = WithToken<SignedTransaction>;
+impl<P: PostSign> Future for ProspectiveSigner<P> {
+	type Item = <P as PostSign>::Item;
```
niklasad1:
```suggestion
	type Item = P::Item;
```

rpc/src/v1/impls/personal.rs
```diff
@@ -18,7 +18,7 @@
 use std::sync::Arc;
 use std::time::Duration;
 
-use bytes::{Bytes, ToPretty};
+use bytes::{Bytes};
```
niklasad1:
```suggestion
use bytes::Bytes;
```

rpc/src/v1/helpers/dispatch.rs
seunlanlege: dispatcher.sign borrows dispatcher, so i can't move it

rpc/src/v1/helpers/dispatch.rs
seunlanlege: this isn't used consistently across the codebase

niklasad1: nvm, I was wrong I think your style is correct sorry!

seunlanlege:
looks like this looks good. needs one more approving review

rpc/src/v1/impls/personal.rs
```diff
@@ -68,7 +69,16 @@ impl<D: Dispatcher> PersonalClient<D> {
...
-	fn do_sign_transaction(&self, _meta: Metadata, request: TransactionRequest, password: String) -> BoxFuture<(PendingTransaction, D)> {
+	fn do_sign_transaction<P>(
+		&self,
+		_meta: Metadata,
+		request: TransactionRequest,
+		password: String,
+		post_sign: P
+	) -> BoxFuture<<P as PostSign>::Item>
+		where P: PostSign + 'static,
+		      <<P as PostSign>::Out as futures::future::IntoFuture>::Future: Send
```
seunlanlege:
```suggestion
 		      <P::Out as futures::future::IntoFuture>::Future: Send
```

rpc/src/v1/impls/personal.rs
```diff
@@ -68,7 +69,16 @@ impl<D: Dispatcher> PersonalClient<D> {
...
 
 impl<D: Dispatcher + 'static> PersonalClient<D> {
-	fn do_sign_transaction(&self, _meta: Metadata, request: TransactionRequest, password: String) -> BoxFuture<(PendingTransaction, D)> {
+	fn do_sign_transaction<P>(
+		&self,
+		_meta: Metadata,
+		request: TransactionRequest,
+		password: String,
+		post_sign: P
+	) -> BoxFuture<<P as PostSign>::Item>
```
seunlanlege:
```suggestion
 	) -> BoxFuture<P::Item>
```

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -449,28 +480,60 @@ fn sign_transaction(
...
+	prospective: Option<WithToken<SignedTransaction>>,
 	ready: Option<nonce::Ready>,
+	post_sign: Option<P>,
+	post_sign_future: Option<<P::Out as IntoFuture>::Future>
 }
 
-impl ProspectiveSigner {
+/// action to execute after signing
+/// e.g importing a transaction into the chain
+pub trait PostSign: Send {
```
cheme: Just a quick idea (maybe dumb idea :) , maybe this trait could be more generic. PostSign is specific name, but as I understand it is a definition for a future post action on a signed transaction, if we make the signed transaction (`WithToken<SignedTransaction>`) an inner type (input) we got  something that is totally generic (I wonder if it maybe already exists in the future api something like a trait for implementing a combination of `future::ok` and `and_then`).

seunlanlege: the futures crate has no such api

5chdn:
Needs a 2md review! :+1:

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -491,6 +554,8 @@ impl ProspectiveSigner {
 			},
 			prospective: None,
 			ready: None,
+			post_sign: Some(post_sign),
```
niklasad1: It doesn't make sense that `post_sign` is an Option anymore because `poll` assumes that it is always is `Some` otherwise panic and `new` requires it too!

seunlanlege: the `execute` method on `PostSign` takes `self`, so inorder to take ownership of `post_sign` from `ProspectiveSigner` it needs to be an option. refer to this line

https://github.com/paritytech/parity-ethereum/blob/001a42a98a557edc7c602265a580dd552c871a1d/rpc/src/v1/helpers/dispatch.rs#L592-L599

joshua-mir:
@5chdn we have had many users running into related issues, do you think we could backport this for this or next release?

5chdn:
yes, label at will

PR #10055, opened 2018-12-12
Update some dependencies for compilation with pc-windows-gnu

tomaka:
Integrates https://github.com/paritytech/rust-rocksdb/pull/25, https://github.com/paritytech/rust-snappy/pull/9, https://github.com/paritytech/libusb-sys/pull/2 and https://github.com/paritytech/libusb-sys/pull/4

cc @TriplEight

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1930,7 +1930,7 @@ dependencies = [
 [[package]]
 name = "libusb-sys"
 version = "0.2.4"
-source = "git+https://github.com/paritytech/libusb-sys#14bdb698003731b6344a79e1d814704e44363e7c"
+source = "git+https://github.com/paritytech/libusb-sys#4418e92c82509d1f049f7942dffe62a0bf5dd871"
 dependencies = [
  "cc 1.0.25 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -2222,7 +2222,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "num-integer 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.22 (registry+https://github.com/rust-lang/crates.io-index)",
```
ordian: This is probably downgrade from https://github.com/rust-lang/cargo/issues/5702, do we want to revert it?

tomaka: I'm surprised we're not using `cargo build --frozen` in our CI to detect that kind of problems.

ordian:
Also android build is failing.

tomaka:
> Also android build is failing.

That's quite funny. We used to erroneously build libusb-sys for x86-64-linux on the Android pipeline, but I assume that it worked because we don't actually use libusb-sys from within Parity when on Android.

Now that this is fixed, we reach a `#error "Unsupported OS"`.

niklasad1:
Ok, shall we just not use the `hardware-wallet stuff` for android temporary?

ordian:
I suggest to revert rand downgrade and deal with android in a separate PR. Any opinions?

Cargo.lock
ordian: I guess `cargo update -p rand:0.3.22` should fix that. `cargo check --frozen` won't fail, the downgrade only happens if you do `cargo update -p <some dep>`.

tomaka:
I don't think we should make `pc-windows-gnu` work by breaking Android.
I'll submit another PR.

PR #10056, opened 2018-12-12
Trivial journal for private transactions

### LINKED ISSUE #9641: [Private transactions] Returning informative private transaction status info/error instead of silent failiure
- **Parity Ethereum version**: Parity-Ethereum/v2.2.0-unstable-93e1040-20180921/x86_64-linux-gnu/rustc1.29.0
- **Operating system**: Linux
- **Installation**: built from source with secretstore feature
- **Network**: Tobalaba/dev/local PoA
- **Restarted**: yes

**actual**

When sending a private transaction with `private_sendTransaction`, e.g. modifying a value as in the [Parity private transactions tutorial](https://wiki.parity.io/Private-Transactions-Tutorial-4.html), a similar response is returned:

```
{ contractAddress: '0x5f72cc768e8de3d78e071c05a9a41a44e88a779a',
  status: 0,
  transactionHash: '0x1c3ae1a8897fc3c1da6e442006c8d35a77e8518803f4604c21c6653a8ba631a9' }
```
This return message was not found to be helpful, because:
 1. In case some error happens, the transaction fails silently (e.g. some error with secret store). There is no way to figure out what happened without digging the logs.
 2. There is no way to figure out if your transaction is ever gonna happen and when is it gonna happen. Should I wait some more? Retry? How can I track the private transaction status?

This is a limitation for a production grade application.

**expected behavior**

Is it possible to improve user experience with:
 - returning an error message in case of a failiure
 - provide a means of tracking the status of the private transaction. Is it still being verified by the private validators? Did it get rejected? Reached the transaction pool already? etc. 

**steps to reproduce**

I made a repo to easily play around with secretstore/private transactions: https://github.com/ngyam/tutorial-secretstore-privatetx 

I couldn't make private transactions work on live Tobalaba (start nodes in `nodes_ss_tobalaba/` folder), so that one is worth to check out. Modifying the value of x as in [tutorial part 4](https://wiki.parity.io/Private-Transactions-Tutorial-4.html) doesn't work.

grbIzl:
For private tx diagnostic purposes trivial journal implemented.
Details of the solution:
- All logs are put into the local json file, created in client's root directory (with name private_tx.log)
- So only the node created the private transaction can provide information about its status
- The following information is stored for every private transaction created on the node:
	/// Original signed transaction hash (used as a source for private tx)
	pub tx_hash: H256,
	/// Current status of the private transaction
	pub status: PrivateTxStatus,
	/// Creation timestamp
	pub creation_timestamp: u64,
	/// List of validations
	pub validators: Vec<ValidatorLog>,
	/// Timestamp of the resulting public tx deployment
	pub deployment_timestamp: Option<u64>,
	/// Hash of the resulting public tx
	pub public_tx_hash: Option<H256>,
- Status has one of the following values:
        /// Private tx was created but no validation received yet
	Created,
	/// Several validators (but not all) validated the transaction
	Validating,
	/// All validators validated the private tx
	/// Corresponding public tx was created and added into the pool
	Deployed, 
- Example of the log:
[{"tx_hash":"0x957251b3acfa1cabaed21234afa40cdb0c51952c6ac426d538f8b559bc50c271","status":"Created","creation_timestamp":1544627108,"validators":[{"account":"0x7ffbe3512782069be388f41be4d8eb350672d3a5","validated":false,"validation_timestamp":null}],"deployment_timestamp":null,"public_tx_hash":null},{"tx_hash":"0xc868b873b74259935f74b8b858bf44f692ad0ead3339844aaad1e6df6a29c982","status":"Deployed","creation_timestamp":1544626975,"validators":[{"account":"0x7ffbe3512782069be388f41be4d8eb350672d3a5","validated":true,"validation_timestamp":1544626975}],"deployment_timestamp":1544626975,"public_tx_hash":"0xe56b48fd040e14e38d6f6dae7846fec10ec20a7edc903b31ba87d3b1bd9413f0"}]
- Corresponding RPC method private_log added in order to access the information
- In order to preserve sensible size of the logs the following limitations implemented:
-- Amount of logs available (currently hardcoded 1000)
-- Lifetime of the logs (currently 20 days hardcoded), the older logs will be dropped with the next processing

Closes #9641

### DIFF
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -25,6 +25,7 @@ mod key_server_keys;
 mod private_transactions;
 mod messages;
 mod error;
+mod log;
 
 extern crate common_types as types;
 extern crate ethabi;
@@ -45,11 +46,15 @@ extern crate parking_lot;
 extern crate trie_db as trie;
 extern crate patricia_trie_ethereum as ethtrie;
 extern crate rlp;
+#[macro_use]
+extern crate serde_derive;
+extern crate serde;
+extern crate serde_json;
 extern crate rustc_hex;
 extern crate transaction_pool as txpool;
 extern crate url;
 #[macro_use]
-extern crate log;
+extern crate log as ethlog;
 #[macro_use]
 extern crate ethabi_derive;
 #[macro_use]
@@ -67,7 +72,8 @@ pub use encryptor::{Encryptor, SecretStoreEncryptor, EncryptorConfig, NoopEncryp
 pub use key_server_keys::{KeyProvider, SecretStoreKeys, StoringKeyProvider};
 pub use private_transactions::{VerifiedPrivateTransaction, VerificationStore, PrivateTransactionSigningDesc, SigningStore};
 pub use messages::{PrivateTransaction, SignedPrivateTransaction};
-pub use error::Error;
+pub use error::{Error, ErrorKind};
+use log::{Logging};
 
 use std::sync::{Arc, Weak};
 use std::collections::{HashMap, HashSet, BTreeMap};
@@ -117,6 +123,8 @@ pub struct ProviderConfig {
 	pub validator_accounts: Vec<Address>,
 	/// Account used for signing public transactions created from private transactions
 	pub signer_account: Option<Address>,
+	/// Path to private tx logs
+	pub logs_path: Option<String>,
 }
 
 #[derive(Debug)]
@@ -177,6 +185,7 @@ pub struct Provider {
 	accounts: Arc<Signer>,
 	channel: IoChannel<ClientIoMessage>,
 	keys_provider: Arc<KeyProvider>,
+	logging: Logging,
 }
 
 #[derive(Debug)]
@@ -211,6 +220,7 @@ impl Provider {
 			accounts,
 			channel,
 			keys_provider,
+			logging: Logging::new(config.logs_path),
 		}
 	}
 
@@ -257,8 +267,9 @@ impl Provider {
 		trace!(target: "privatetx", "Required validators: {:?}", contract_validators);
 		let private_state_hash = self.calculate_state_hash(&private_state, contract_nonce);
 		trace!(target: "privatetx", "Hashed effective private state for sender: {:?}", private_state_hash);
-		self.transactions_for_signing.write().add_transaction(private.hash(), signed_transaction, contract_validators, private_state, contract_nonce)?;
+		self.transactions_for_signing.write().add_transaction(private.hash(), signed_transaction, &contract_validators, private_state, contract_nonce)?;
 		self.broadcast_private_transaction(private.hash(), private.rlp_bytes());
+		self.logging.private_tx_created(tx_hash, &contract_validators);
 		Ok(Receipt {
 			hash: tx_hash,
 			contract_address: contract,
@@ -354,8 +365,9 @@ impl Provider {
 			Some(desc) => desc,
 		};
 		let last = self.last_required_signature(&desc, signed_tx.signature())?;
+		let original_tx_hash = desc.original_transaction.hash();
 
-		if last {
+		if last.0 {
 			let mut signatures = desc.received_signatures.clone();
 			signatures.push(signed_tx.signature());
 			let rsv: Vec<Signature> = signatures.into_iter().map(|sign| sign.into_electrum().into()).collect();
@@ -373,7 +385,7 @@ impl Provider {
 			trace!(target: "privatetx", "Last required signature received, public transaction created: {:?}", public_tx);
 			// Sign and add it to the queue
 			let chain_id = desc.original_transaction.chain_id();
-			let hash = public_tx.hash(chain_id);
+			let public_tx_hash = public_tx.hash(chain_id);
 			let signature = self.accounts.sign(signer_account, hash)?;
 			let signed = SignedTransaction::new(public_tx.with_signature(signature, chain_id))?;
 			match self.miner.import_own_transaction(&*self.client, signed.into()) {
@@ -392,6 +404,9 @@ impl Provider {
 					Err(err) => warn!(target: "privatetx", "Failed to send private state changed notification, error: {:?}", err),
 				}
 			}
+			// Store logs
+			self.logging.signature_added(original_tx_hash, last.1);
+			self.logging.tx_deployed(original_tx_hash, public_tx_hash);
 			// Remove from store for signing
 			if let Err(err) = self.transactions_for_signing.write().remove(&private_hash) {
 				warn!(target: "privatetx", "Failed to remove transaction from signing store, error: {:?}", err);
@@ -400,7 +415,10 @@ impl Provider {
 		} else {
 			// Add signature to the store
 			match self.transactions_for_signing.write().add_signature(&private_hash, signed_tx.signature()) {
-				Ok(_) => trace!(target: "privatetx", "Signature stored for private transaction"),
+				Ok(_) => {
+					trace!(target: "privatetx", "Signature stored for private transaction");
+					self.logging.signature_added(original_tx_hash, last.1);
+				}
 				Err(err) => {
 					warn!(target: "privatetx", "Failed to add signature to signing store, error: {:?}", err);
 					return Err(err);
@@ -420,17 +438,14 @@ impl Provider {
 		}
 	}
 
-	fn last_required_signature(&self, desc: &PrivateTransactionSigningDesc, sign: Signature) -> Result<bool, Error>  {
-		if desc.received_signatures.contains(&sign) {
-			return Ok(false);
-		}
+	fn last_required_signature(&self, desc: &PrivateTransactionSigningDesc, sign: Signature) -> Result<(bool, Address), Error>  {
 		let state_hash = self.calculate_state_hash(&desc.state, desc.contract_nonce);
 		match recover(&sign, &state_hash) {
 			Ok(public) => {
 				let sender = public_to_address(&public);
 				match desc.validators.contains(&sender) {
 					true => {
-						Ok(desc.received_signatures.len() + 1 == desc.validators.len())
+						Ok((desc.received_signatures.len() + 1 == desc.validators.len(), sender))
 					}
 					false => {
 						warn!(target: "privatetx", "Sender's state doesn't correspond to validator's");
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -73,9 +73,10 @@ pub use key_server_keys::{KeyProvider, SecretStoreKeys, StoringKeyProvider};
 pub use private_transactions::{VerifiedPrivateTransaction, VerificationStore, PrivateTransactionSigningDesc, SigningStore};
 pub use messages::{PrivateTransaction, SignedPrivateTransaction};
 pub use error::Error;
-pub use log::{Logging, TransactionLog, ValidatorLog, PrivateTxStatus, FileLogsSerializer, SystemTimestamp};
+pub use log::{Logging, TransactionLog, ValidatorLog, PrivateTxStatus, FileLogsSerializer};
 
 use std::sync::{Arc, Weak};
+use std::time::SystemTime;
 use std::collections::{HashMap, HashSet, BTreeMap};
 use ethereum_types::{H128, H256, U256, Address};
 use hash::keccak;
@@ -220,7 +221,7 @@ impl Provider {
 			accounts,
 			channel,
 			keys_provider,
-			logging: Logging::new(Arc::new(FileLogsSerializer::with_path(config.logs_path)), Box::new(SystemTimestamp {})),
+			logging: Logging::new(Arc::new(FileLogsSerializer::with_path(config.logs_path)), Box::new(SystemTime::now())),
 		}
 	}
 
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -63,6 +63,9 @@ extern crate derive_more;
 #[macro_use]
 extern crate rlp_derive;
 
+#[cfg(not(time_checked_add))]
+extern crate time_utils;
+
 #[cfg(test)]
 extern crate rand;
 #[cfg(test)]
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -701,7 +701,7 @@ impl Provider {
 	/// Retrieves log information about private transaction
 	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
 		match self.logging {
-			Some(ref logging) => logging.tx_log(&tx_hash).ok_or_else(|| Error::TxNotFoundInLog.into()),
+			Some(ref logging) => logging.tx_log(&tx_hash).ok_or(Error::TxNotFoundInLog.into()),
 			None => Err(Error::LoggingPathNotSet),
 		}
 	}
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -73,7 +73,7 @@ pub use key_server_keys::{KeyProvider, SecretStoreKeys, StoringKeyProvider};
 pub use private_transactions::{VerifiedPrivateTransaction, VerificationStore, PrivateTransactionSigningDesc, SigningStore};
 pub use messages::{PrivateTransaction, SignedPrivateTransaction};
 pub use error::{Error, ErrorKind};
-use log::{Logging};
+pub use log::{Logging, TransactionLog, ValidatorLog, PrivateTxStatus};
 
 use std::sync::{Arc, Weak};
 use std::collections::{HashMap, HashSet, BTreeMap};
@@ -687,6 +687,11 @@ impl Provider {
 	pub fn private_call(&self, block: BlockId, transaction: &SignedTransaction) -> Result<Executed, Error> {
 		let result = self.execute_private(transaction, TransactOptions::with_no_tracing(), block)?;
 		Ok(result.result)
+	}
+
+	/// Retrieves log information about private transaction
+	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
+		self.logging.tx_log(tx_hash).ok_or(ErrorKind::TxNotFoundInLog.into())
 	}
 
 	/// Returns private validators for a contract.
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -701,7 +701,7 @@ impl Provider {
 	/// Retrieves log information about private transaction
 	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
 		match self.logging {
-			Some(ref logging) => logging.tx_log(&tx_hash).ok_or(Error::TxNotFoundInLog.into()),
+			Some(ref logging) => logging.tx_log(&tx_hash).ok_or(Error::TxNotFoundInLog),
 			None => Err(Error::LoggingPathNotSet),
 		}
 	}
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -269,7 +269,7 @@ impl Provider {
 		trace!(target: "privatetx", "Hashed effective private state for sender: {:?}", private_state_hash);
 		self.transactions_for_signing.write().add_transaction(private.hash(), signed_transaction, &contract_validators, private_state, contract_nonce)?;
 		self.broadcast_private_transaction(private.hash(), private.rlp_bytes());
-		self.logging.private_tx_created(tx_hash, &contract_validators);
+		self.logging.private_tx_created(&tx_hash, &contract_validators);
 		Ok(Receipt {
 			hash: tx_hash,
 			contract_address: contract,
@@ -405,8 +405,8 @@ impl Provider {
 				}
 			}
 			// Store logs
-			self.logging.signature_added(original_tx_hash, last.1);
-			self.logging.tx_deployed(original_tx_hash, public_tx_hash);
+			self.logging.signature_added(&original_tx_hash, &last.1);
+			self.logging.tx_deployed(&original_tx_hash, &public_tx_hash);
 			// Remove from store for signing
 			if let Err(err) = self.transactions_for_signing.write().remove(&private_hash) {
 				warn!(target: "privatetx", "Failed to remove transaction from signing store, error: {:?}", err);
@@ -417,7 +417,7 @@ impl Provider {
 			match self.transactions_for_signing.write().add_signature(&private_hash, signed_tx.signature()) {
 				Ok(_) => {
 					trace!(target: "privatetx", "Signature stored for private transaction");
-					self.logging.signature_added(original_tx_hash, last.1);
+					self.logging.signature_added(&original_tx_hash, &last.1);
 				}
 				Err(err) => {
 					warn!(target: "privatetx", "Failed to add signature to signing store, error: {:?}", err);
@@ -691,7 +691,7 @@ impl Provider {
 
 	/// Retrieves log information about private transaction
 	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
-		self.logging.tx_log(tx_hash).ok_or(ErrorKind::TxNotFoundInLog.into())
+		self.logging.tx_log(&tx_hash).ok_or(ErrorKind::TxNotFoundInLog.into())
 	}
 
 	/// Returns private validators for a contract.
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -73,7 +73,7 @@ pub use key_server_keys::{KeyProvider, SecretStoreKeys, StoringKeyProvider};
 pub use private_transactions::{VerifiedPrivateTransaction, VerificationStore, PrivateTransactionSigningDesc, SigningStore};
 pub use messages::{PrivateTransaction, SignedPrivateTransaction};
 pub use error::{Error, ErrorKind};
-pub use log::{Logging, TransactionLog, ValidatorLog, PrivateTxStatus, FileLogsSerializer};
+pub use log::{Logging, TransactionLog, ValidatorLog, PrivateTxStatus, FileLogsSerializer, SystemTimestamp};
 
 use std::sync::{Arc, Weak};
 use std::collections::{HashMap, HashSet, BTreeMap};
@@ -220,7 +220,7 @@ impl Provider {
 			accounts,
 			channel,
 			keys_provider,
-			logging: Logging::new(Box::new(FileLogsSerializer::new(config.logs_path))),
+			logging: Logging::new(Arc::new(FileLogsSerializer::new(config.logs_path)), Box::new(SystemTimestamp {})),
 		}
 	}
 
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -188,7 +188,7 @@ pub struct Provider {
 	accounts: Arc<Signer>,
 	channel: IoChannel<ClientIoMessage>,
 	keys_provider: Arc<KeyProvider>,
-	logging: Logging,
+	logging: Option<Logging>,
 }
 
 #[derive(Debug)]
@@ -223,7 +223,7 @@ impl Provider {
 			accounts,
 			channel,
 			keys_provider,
-			logging: Logging::new(Arc::new(FileLogsSerializer::with_path(config.logs_path))),
+			logging: config.logs_path.map(|path| Logging::new(Arc::new(FileLogsSerializer::with_path(path)))),
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


5chdn:
needs reviews, could you have a look at this @twittner @sorpaas ?

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+	pub fn tx_deployed(&self, tx_hash: H256, public_tx_hash: H256) {
+		let mut logs = self.logs.write();
+		if let Some(log) = logs.get_mut(&tx_hash) {
+			log.status = PrivateTxStatus::Deployed;
+			log.deployment_timestamp = Some(SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs());
+			log.public_tx_hash = Some(public_tx_hash);
+		}
+	}
+
+	fn read_logs(&self) {
```
twittner: Should this not return a `Result`?

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+		// Drop old logs
+		let current_timestamp = SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs();
+		transaction_logs.retain(|tx_log| current_timestamp - tx_log.creation_timestamp < MAX_STORING_TIME);
+		let mut logs = self.logs.write();
+		for log in transaction_logs {
+			logs.insert(log.tx_hash, log);
+		}
+	}
+
+	fn flush_logs(&self) {
```
twittner: Should this not return a `Result`?

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+						trace!(target: "privatetx", "Cannot open logs file for writing: {}", err);
+						return;
+					}
+				}
+			}
+			None => {
+				return;
+			}
+		};
+		if let Err(err) = serde_json::to_writer(log_file, &self.logs.read().values().cloned().collect::<Vec<TransactionLog>>()) {
```
twittner: Instead of cloning the values, how about:

```rust
let logs = self.logs.read();
let mut json = serde_json::Serializer::new(log_file);
let mut json_array = json.serialize_seq(Some(logs.len()))?;
for v in logs.values() {
	json_array.serialize_element(v)?;
}
json_array.end()?;
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+			transaction_log.status = PrivateTxStatus::Validating;
+			if let Some(ref mut validator_log) = transaction_log.validators.iter_mut().find(|log| log.account == validator) {
+				validator_log.validated = true;
+				validator_log.validation_timestamp = Some(SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs());
+			}
+		}
+	}
+
+	/// Logs the final deployment of the resulting public transaction
+	pub fn tx_deployed(&self, tx_hash: H256, public_tx_hash: H256) {
```
twittner:
```suggestion
	pub fn tx_deployed(&self, tx_hash: &H256, public_tx_hash: H256) {
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+		if let Err(err) = serde_json::to_writer(log_file, &self.logs.read().values().cloned().collect::<Vec<TransactionLog>>()) {
+			error!(target: "privatetx", "Error during logs serialisation: {}", err);
+		}
+	}
+}
+
+// Flush all logs on drop
+impl Drop for Logging {
+	fn drop(&mut self) {
+		self.flush_logs();
```
twittner: I am not sure if writing the JSON file in a destructor is the best approach. One can not return errors back and as a destructor it executes quite slowly. Would it not be possible to offer explicit read/write functionality in the logging API?

5chdn:
@grbIzl

grbIzl:
Refactoring this code in order to address the comments from @twittner

ethcore/private-tx/src/log.rs
grbIzl: I was heavily thinking between these two approaches (log every event into the file vs flush all logs in the destructor). The problem with logging every event is, that these events are not sequential, as a result I would need to rewrite the whole log file in order to update one log record. Also not very efficient. So I've decided to stop on the flush in the dtor. But I'm open to the discussion ;-) and I've also refactored this part and added a test for serialization in order to proceed with any changes easily.

ethcore/private-tx/src/private_transactions.rs
```diff
@@ -223,7 +223,7 @@ impl SigningStore {
 		&mut self,
 		private_hash: H256,
 		transaction: SignedTransaction,
-		validators: Vec<Address>,
+		validators: &Vec<Address>,
```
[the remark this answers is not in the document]
grbIzl: Unfortunately these validators array is required in two different places, so I cannot just move it to the one of them.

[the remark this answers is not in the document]
twittner: If necessary the `Vec` can be cloned on the call site, no? My point was that `add_transaction` needs `validators` by value.

ethcore/private-tx/src/log.rs
twittner: It was only concerned with offering explicit read/write functions which return `Result` instead of reading and writing in constructor and destructor with no way to report errors. The granularity of writes should be unaffected.

grbIzl: Discussed this moment offline. As a result, I would like to re-phrase @twittner 's concern (as I understand its application): should we return the error during RPC call of private transaction methods in the case, when flushing\reading logs during this call fails. My opinion is that we should not: 1) logging is just a secondary functionality and the primary functionality must not suffer from its flaws. The informing of the client via error warnings is enough IMO for such cases, 2) half of the cases, when private transaction methods are called, is the automatic processing of incoming packets. In this case returning the error doesn't make any sense at all, because it cannot be handled by anybody.

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+		let logging = Logging {
+			logs: RwLock::new(HashMap::new()),
+			logs_dir: logs_dir.map(|dir| PathBuf::from(dir)),
+		};
+		logging.read_logs();
+		logging
+	}
+
+	/// Retrieves log for the corresponding tx hash
+	pub fn tx_log(&self, tx_hash: H256) -> Option<TransactionLog> {
```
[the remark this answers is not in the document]
grbIzl: Reworked

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,247 @@
...
+			status: PrivateTxStatus::Created,
+			creation_timestamp: SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs(),
+			validators: validator_logs,
+			deployment_timestamp: None,
+			public_tx_hash: None,
+		});
+	}
+
+	/// Logs the obtaining of the signature for the private transaction
+	pub fn signature_added(&self, tx_hash: H256, validator: Address) {
```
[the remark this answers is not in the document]
grbIzl: Reworked

ethcore/private-tx/src/log.rs
grbIzl: Yes, result is added. As for its usage, see the discussion below

ethcore/private-tx/src/log.rs
grbIzl: Reworked

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
+use parking_lot::{RwLock};
+use serde::ser::{Serializer, SerializeSeq};
+use error::{Error};
+
+/// Maximum amount of stored private transaction logs.
+const MAX_JOURNAL_LEN: usize = 1000;
+
+/// Maximum period for storing private transaction logs.
+/// Older logs will not be processed, 20 days
```
niklasad1: Please rephrase
```suggestion
/// Logs older than 20 days will not be processed
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Private transactions logs.
+
+use ethereum_types::{H256, Address};
+use std::collections::{HashMap};
```
niklasad1: Remove needless blocks for single imports

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+const MAX_STORING_TIME: u64 = 60 * 60 * 24 * 20;
+
+/// Current status of the private transaction
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub enum PrivateTxStatus {
+	/// Private tx was created but no validation received yet
+	Created,
+	/// Several validators (but not all) validated the transaction
+	Validating,
+	/// All validators validated the private tx
```
niklasad1:
```suggestion
	/// All validators has validated the private tx
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+	Deployed,
+}
+
+/// Information about private tx validation
+#[derive(Clone, Serialize, Deserialize)]
+pub struct ValidatorLog {
+	/// Account of the validator
+	pub account: Address,
+	/// Validation flag
+	pub validated: bool,
```
niklasad1: Why are both `validated` and `validation_timestamp` needed? 

AFAIU, you should not have a `validation_timestamp` unless validated it true?

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+pub trait LogsSerializer: Send + Sync + 'static {
+	/// Read logs from the source
+	fn read_logs(&self) -> Result<Vec<TransactionLog>, Error>;
+
+	/// Write all logs to the source
+	fn flush_logs(&self, logs: &HashMap<H256, TransactionLog>) -> Result<(), Error>;
+}
+
+pub struct FileLogsSerializer {
+	logs_dir: Option<PathBuf>,
```
niklasad1: Personally I would prefer a little different API:

```rust
#[derive(Default)]
pub struct FileLogsSerializer { .... }

impl FileLogsSerializer {
  pub fn with_path(path: PathBuf) -> Self { ... }
}

```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+/// Timestamp source for logs
+pub trait TimestampSource: Send + Sync + 'static {
+	/// Returns current timestamp in seconds
+	fn current_timestamp(&self) -> u64;
+}
+
+/// Timesource on the base of system time
+pub struct SystemTimestamp {}
+
+impl TimestampSource for SystemTimestamp {
```
niklasad1: Why not implement this directly on `SystemTime`?

I also think the trait name is not ideal something like `CurrentTime` is better because the actual implementer provides the `source`

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// Timestamp source for logs
+pub trait TimestampSource: Send + Sync + 'static {
+	/// Returns current timestamp in seconds
+	fn current_timestamp(&self) -> u64;
+}
+
+/// Timesource on the base of system time
+pub struct SystemTimestamp {}
```
niklasad1: maybe remove see comment below

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+		logging
+	}
+
+	/// Retrieves log for the corresponding tx hash
+	pub fn tx_log(&self, tx_hash: &H256) -> Option<TransactionLog> {
+		self.logs.read().get(&tx_hash).cloned()
+	}
+
+	/// Logs the creation of private transaction
+	pub fn private_tx_created<'a>(&self, tx_hash: &H256, validators: &Vec<Address>) {
```
niklasad1:
```suggestion
	pub fn private_tx_created(&self, tx_hash: &H256, validators: &[Address]) {
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+			deployment_timestamp: None,
+			public_tx_hash: None,
+		});
+	}
+
+	/// Logs the obtaining of the signature for the private transaction
+	pub fn signature_added(&self, tx_hash: &H256, validator: &Address) {
+		let mut logs = self.logs.write();
+		if let Some(transaction_log) = logs.get_mut(&tx_hash) {
+			transaction_log.status = PrivateTxStatus::Validating;
```
niklasad1: shall this state really be changed before the validator account has been found?

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+			status: PrivateTxStatus::Created,
+			creation_timestamp: self.timestamp_source.current_timestamp(),
+			validators: validator_logs,
+			deployment_timestamp: None,
+			public_tx_hash: None,
+		});
+	}
+
+	/// Logs the obtaining of the signature for the private transaction
+	pub fn signature_added(&self, tx_hash: &H256, validator: &Address) {
```
niklasad1: I don't understand the documentation and naming of this method.

To me it seems like it tries to validate a transaction and append the timestamp to it

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,412 @@
...
+		fn flush_logs(&self, logs: &HashMap<H256, TransactionLog>) -> Result<(), Error> {
+			// Sort logs in order to have the same order
+			let sorted_logs: BTreeMap<&H256, &TransactionLog> = logs.iter().collect();
+			*self.string_log.write() = serde_json::to_string(&sorted_logs.values().collect::<Vec<&&TransactionLog>>())?;
+			Ok(())
+		}
+	}
+
+	struct CounterTimestamp {
+		counter: RwLock<u64>,
```
niklasad1: This could be replaced `AtomicU64` when rust 1.34 is a thing.

ethcore/private-tx/src/private_transactions.rs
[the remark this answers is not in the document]
niklasad1: Yes, it could be cloned on the `call-site` but maybe @grbIzl is concerned by the massive call overhead of passing 3 words instead 1 :smile: 

Also, I noticed that transaction is cloned which is not needed because it moved into function but it is probably out-of-scope for this PR

ethcore/private-tx/src/lib.rs
```diff
@@ -674,6 +689,11 @@ impl Provider {
 		Ok(result.result)
 	}
 
+	/// Retrieves log information about private transaction
+	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
+		self.logging.tx_log(&tx_hash).ok_or(ErrorKind::TxNotFoundInLog.into())
```
niklasad1: `ok_or` is eager (always evaluated) could be replaced by `ok_or_else` because it is lazy

ethcore/private-tx/src/log.rs
grbIzl: Timestamp removed

ethcore/private-tx/src/log.rs
grbIzl: This wrapper was done in order to use in test. I have a test for serialization, where I need to have an exact timestamp value, that cannot be guaranteed with system timestamp

ethcore/private-tx/src/log.rs
grbIzl: Documentation re-phrased

ethcore/private-tx/src/log.rs
niklasad1: hmm, I don't understand your answer, basically I think the trait is fine which is used to create a test helper for to serialization AFAIU.

What I meant was that you can remove the `struct SystemTimeStamp` and do this instead:

```rust
impl TimestampSource for SystemTime {
    fn timestamp(&self) -> u64 {
        self.duration_since(UNIX_EPOCH).unwrap_or_default().as_secs()
    }
}
```

But it is fine either way!

ethcore/private-tx/src/log.rs
niklasad1: any reason why it isn't `#[derive(Default)]`?

soc1c:
Please rebase

ethcore/private-tx/src/log.rs
grbIzl: Got it! Reworked.

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,411 @@
...
+}
+
+/// Timesource on the base of system time
+impl CurrentTime for SystemTime {
+	fn timestamp(&self) -> u64 {
+		self.duration_since(UNIX_EPOCH).unwrap_or_default().as_secs()
+	}
+}
+
+/*/// Timesource on the base of system time
```
niklasad1: Remove?

grbIzl: Sorry, my bad

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,402 @@
...
+	Deployed,
+}
+
+/// Information about private tx validation
+#[derive(Clone, Serialize, Deserialize)]
+pub struct ValidatorLog {
+	/// Account of the validator
+	pub account: Address,
+	/// Validation timestamp, None if the transaction is not validated
+	pub validation_timestamp: Option<u64>,
```
twittner:
```suggestion
	pub validation_timestamp: Option<SystemTime>,
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,402 @@
...
+
+/// Information about the private transaction
+#[derive(Clone, Serialize, Deserialize)]
+pub struct TransactionLog {
+	/// Original signed transaction hash (used as a source for private tx)
+	pub tx_hash: H256,
+	/// Current status of the private transaction
+	pub status: PrivateTxStatus,
+	/// Creation timestamp
+	pub creation_timestamp: u64,
```
twittner:
```suggestion
	pub creation_timestamp: SystemTime,
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,402 @@
...
+	/// List of validations
+	pub validators: Vec<ValidatorLog>,
+	/// Timestamp of the resulting public tx deployment
+	pub deployment_timestamp: Option<u64>,
```
twittner:
```suggestion
	pub deployment_timestamp: Option<SystemTime>,
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,402 @@
...
+			log.deployment_timestamp = Some(self.timestamp_source.timestamp());
+			log.public_tx_hash = Some(*public_tx_hash);
+		}
+	}
+
+	fn read_logs(&self) -> Result<(), Error> {
+		let mut transaction_logs = self.logs_serializer.read_logs()?;
+		// Drop old logs
+		let current_timestamp = self.timestamp_source.timestamp();
+		transaction_logs.retain(|tx_log| current_timestamp - tx_log.creation_timestamp < MAX_STORING_TIME);
```
twittner: Looking at the `CurrentTime` implementation for `SystemTime` [above](https://github.com/paritytech/parity-ethereum/pull/10056/files#diff-bc698921b3e50488ef5d7b2875e5436bR163) it seems that errors are mapped to 0. Assuming `tx_log.creation_timestamp > 0` this could lead to an underflow here (or a panic in `debug` mode).

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,402 @@
...
+		let mut logs = self.logs.write();
+		if let Some(log) = logs.get_mut(&tx_hash) {
+			log.status = PrivateTxStatus::Deployed;
+			log.deployment_timestamp = Some(self.timestamp_source.timestamp());
+			log.public_tx_hash = Some(*public_tx_hash);
+		}
+	}
+
+	fn read_logs(&self) -> Result<(), Error> {
+		let mut transaction_logs = self.logs_serializer.read_logs()?;
```
twittner: `Logging` is sensitive to system clock changes which happen all too often unfortunately. Maybe upon instantiation of `Logging` a timestamp should be kept around and all subsequent timestamps are generated by `self.timestamp + Instant::now().elapsed()`. `Instant` is guaranteed to be monotonic. The initial timestamp could be set to `max(SystemTime::now(), latest of the log entries read)`.

ethcore/private-tx/src/lib.rs
```diff
@@ -674,6 +698,14 @@ impl Provider {
 		Ok(result.result)
 	}
 
+	/// Retrieves log information about private transaction
+	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
+		match self.logging {
+			Some(ref logging) => logging.tx_log(&tx_hash).ok_or_else(|| Error::TxNotFoundInLog.into()),
```
niklasad1:
```suggestion
			Some(ref logging) => logging.tx_log(&tx_hash).ok_or(Error::TxNotFoundInLog),
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+		} else {
+			File::open(&file_path).map_err(From::from)
+		}
+	}
+}
+
+impl LogsSerializer for FileLogsSerializer {
+	fn read_logs(&self) -> Result<Vec<TransactionLog>, Error> {
+		let log_file = self.open_file(false)?;
+		let transaction_logs: Vec<TransactionLog> = match serde_json::from_reader(log_file) {
```
niklasad1: The match could be simplified and `return directly` such as:

```rust
		match serde_json::from_reader(log_file) {
			Ok(logs) => Ok(logs),
			Err(err) => {
				error!(target: "privatetx", "Cannot deserialize logs from file: {}", err);
				return Err(format!("Cannot deserialize logs from file: {:?}", err).into());
			}
		}
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+		json_array.end()?;
+		Ok(())
+	}
+}
+
+/// Private transactions logging
+pub struct Logging {
+	logs: RwLock<HashMap<H256, TransactionLog>>,
+	logs_serializer: Arc<LogsSerializer>,
+	mono_time: Mutex<MonoTime>,
```
niklasad1: Why is this wrapped in a mutex? 

I mean it is never mutated AFAIU and just reads `Instant::elapsed()` or `Instand::elapsed() + start_time`?

twittner: I think the Mutex is not needed. The existing`RwLock` implies sufficient synchronisation, i.e. all calls to `MonoTime` are guarded by a `logs.write()`.

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	fn read_logs(&self) -> Result<SystemTime, Error> {
+		let mut transaction_logs = self.logs_serializer.read_logs()?;
+		// Drop old logs
+		let earliest_possible = SystemTime::now().checked_sub(MAX_STORING_TIME).ok_or(Error::TimestampOverflow)?;
+		transaction_logs.retain(|tx_log| tx_log.creation_timestamp > earliest_possible);
+		// Sort logs by their creation time in order to find the most recent
+		transaction_logs.sort_by(|a, b| b.creation_timestamp.cmp(&a.creation_timestamp));
+		let initial_timestamp = transaction_logs.first()
```
niklasad1: hrm, I don't understand the reasoning about returning max(l.creation_timestamp, SystemTime::now())

will not `SystemTIme::now() > l.creation_timestamp` always be true?

Why are we return `SystemTime::now` instead `most_recent` BTW?

Also, map and map_or could be merged to:

```.map_or(SystemTime::now(), |l| std::cmp::max(SystemTime::now(), l.creation_timestamp));```

Comment (twittner) [reply ethcore/private-tx/src/log.rs]: > will not SystemTIme::now() > l.creation_timestamp always be true?

No, `SystemTime` is not monotonic, see: https://doc.rust-lang.org/std/time/struct.SystemTime.html

File: rpc/src/v1/types/private_log.rs [hunk 31]
```diff
@@ -0,0 +1,93 @@
...
+	/// Timestamp of the resulting public tx deployment
+	pub deployment_timestamp: Option<u64>,
+	/// Hash of the resulting public tx
+	pub public_tx_hash: Option<H256>,
+}
+
+impl From<EthTransactionLog> for PrivateTransactionLog {
+	fn from(r: EthTransactionLog) -> Self {
+		PrivateTransactionLog {
+			tx_hash: r.tx_hash.into(),
```
niklasad1:
```suggestion
			tx_hash: r.tx_hash,
```

rpc/src/v1/types/private_log.rs
```diff
@@ -0,0 +1,93 @@
...
+			status: r.status.into(),
+			creation_timestamp: r.creation_timestamp.duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs(),
+			validators: r.validators.into_iter().map(Into::into).collect(),
+			deployment_timestamp: r.deployment_timestamp.map(|t| t.duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs()),
+			public_tx_hash: r.public_tx_hash.map(Into::into),
```
niklasad1:
```suggestion
			public_tx_hash: r.public_tx_hash,
```

rpc/src/v1/types/private_log.rs
```diff
@@ -0,0 +1,93 @@
...
+	/// Account of the validator
+	pub account: H160,
+	/// Validation timestamp, None, if the transaction is not validated yet
+	pub validation_timestamp: Option<u64>,
+}
+
+impl From<EthValidatorLog> for ValidatorLog {
+	fn from(r: EthValidatorLog) -> Self {
+		ValidatorLog {
+			account: r.account.into(),
```
niklasad1:
```suggestion
			account: r.account,
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Wrapper other JSON serializer
+pub trait LogsSerializer: Send + Sync + 'static {
+	/// Read logs from the source
+	fn read_logs(&self) -> Result<Vec<TransactionLog>, Error>;
+
+	/// Write all logs to the source
+	fn flush_logs(&self, logs: &HashMap<H256, TransactionLog>) -> Result<(), Error>;
+}
+
+pub struct FileLogsSerializer {
```
niklasad1: Missing docs, perhaps add `#![deny(missing_docs)]` in `src/lib.rs` to enforce it

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+		if let Err(err) = self.flush_logs() {
+			warn!(target: "privatetx", "Cannot write logs: {:?}", err);
+		}
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use serde_json;
+	use error::{Error};
```
niklasad1:
```suggestion
	use error::Error;
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+	use ethereum_types::{H256};
```
niklasad1:
```suggestion
	use ethereum_types::H256;
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+	use std::collections::{HashMap, BTreeMap};
+	use std::sync::{Arc};
```
niklasad1:
```suggestion
	use std::sync::Arc;
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+	use std::time::{SystemTime, Duration};
+	use types::transaction::{Transaction};
```
niklasad1:
```suggestion
	use types::transaction::Transaction;
```

ethcore/private-tx/src/log.rs
```diff
@@ -0,0 +1,409 @@
...
+	use parking_lot::{RwLock};
```
niklasad1:
```suggestion
	use parking_lot::RwLock;
```

ethcore/private-tx/src/lib.rs
```diff
@@ -674,6 +698,14 @@ impl Provider {
 		Ok(result.result)
 	}
 
+	/// Retrieves log information about private transaction
+	pub fn private_log(&self, tx_hash: H256) -> Result<TransactionLog, Error> {
+		match self.logging {
+			Some(ref logging) => logging.tx_log(&tx_hash).ok_or(Error::TxNotFoundInLog.into()),
```
niklasad1: This `into()`  is not needed

rpc/src/v1/impls/private.rs
```diff
@@ -119,4 +120,10 @@ impl Private for PrivateClient {
 		let key = client.contract_key_id(&contract_address).map_err(errors::private_message)?;
 		Ok(key)
 	}
+
+	fn private_log(&self, tx_hash: H256) -> Result<PrivateTransactionLog, Error> {
+		let client = self.unwrap_manager()?;
```
niklasad1: sorry for late comment Anton but this piece could be simplified to:

```rust
		self.unwrap_manager()?
			.private_log(tx_hash)
			.map_err(errors::private_message)
			.map(Into::into)

``

niklasad1:
@grbIzl shall this be mentioned in the release notes? If so, please add the label :P

PR #10060, opened 2018-12-13
Align personal_unlockAccount behaviour when permanent unlock is disabled

joshua-mir:
> If permanent unlocking is disabled (the default) then the duration argument will be **ignored**, and the account will be unlocked for a single signing.

Current behaviour throws an error.

### DIFF
--- a/rpc/src/v1/impls/personal.rs
+++ b/rpc/src/v1/impls/personal.rs
@@ -128,10 +128,7 @@ impl<D: Dispatcher + 'static> Personal for PersonalClient<D> {
 
 		let r = match (self.allow_perm_unlock, duration) {
 			(false, None) => store.unlock_account_temporarily(account, account_pass.into()),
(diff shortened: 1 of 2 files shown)


tomusdrw:
Flashbacks: https://github.com/paritytech/parity-ethereum/pull/6777
IMHO it's a bit misleading to ignore the parameter and it's better to explicitly error, but I'm open for arguments why not.

joshua-mir:
I can change this so it throws a *different* error, we no longer have geth compatibility mode so the error wasn't helpful. In addition I feel that the documentation is clear enough that we are ignoring duration.

tomusdrw:
I'd prefer an error, cause in the past we had issues with some implicit, but unexpected behaviour (even if documented).

joshua-mir:
(I was working with the documentation script, which is why github claims that the previous commit was made by devops, oops, please ignore.)

PR #10065, opened 2018-12-14
Handle the case for contract creation on an empty but exist account with storage items

sorpaas:
In spec it's technically possible to create empty account with existing storage items, thus break the net metering model. This PR creates `is_base_storage_root_unchanged` function in state. However, any **sane person** wouldn't do that so practically `is_base_storage_root_unchanged` should always be `true` with the only exception of tests.

For any items in the checkpoint, once we set `original_storage_cache`, we carry it on for any further items in the checkpoint. So we can check the newest items in the checkpoint list to determine whether there's any changes of the `base_storage_root` value. It works for current model because the value of `base_storage_root` can only ever be changed once.

### DIFF
--- a/ethcore/src/externalities.rs
+++ b/ethcore/src/externalities.rs
@@ -119,7 +119,7 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 		if self.state.is_base_storage_root_unchanged(&self.origin_info.address)? {
 			self.state.checkpoint_storage_at(0, &self.origin_info.address, key).map(|v| v.unwrap_or(H256::zero())).map_err(Into::into)
 		} else {
-			warn!(target: "externalities", "Detected existing account 0x{:x} where a forced contract creation happened.", self.origin_info.address);
+			warn!(target: "externalities", "Detected existing account {:#x} where a forced contract creation happened.", self.origin_info.address);
 			Ok(H256::zero())
 		}
 	}
--- a/ethcore/src/externalities.rs
+++ b/ethcore/src/externalities.rs
@@ -116,7 +116,11 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	where T: Tracer, V: VMTracer, B: StateBackend
 {
 	fn initial_storage_at(&self, key: &H256) -> vm::Result<H256> {
-		self.state.checkpoint_storage_at(0, &self.origin_info.address, key).map(|v| v.unwrap_or(H256::zero())).map_err(Into::into)
+		if self.state.is_base_storage_root_unchanged(&self.origin_info.address)? {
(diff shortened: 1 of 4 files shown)


ethcore/src/externalities.rs
```diff
@@ -116,6 +116,10 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	where T: Tracer, V: VMTracer, B: StateBackend
 {
 	fn initial_storage_at(&self, key: &H256) -> vm::Result<H256> {
+		if self.substate.new_contract_with_previously_non_empty_storage.contains(&self.origin_info.address) {
```
tomusdrw: Is it possible that the contract was empty initially, but then it is somehow initialized via `CREATE` within the same scope?
I.e. asking if returning `zero()` here wouldn't break something else

sorpaas: I changed this to use `is_base_storage_root_unchanged`. If base storage root is changed from an original storage root value, then the changed base storage root can only be `KECCAK_NULL_RLP` (because we only change it in contract creation). From any point after that, we discard the actual original storage value, but use `H256::zero()`.

I think it's safe and can't think of any way this will break things, but would love to have more eyes for review.

ethcore/src/state/mod.rs
```diff
@@ -539,6 +539,13 @@ impl<B: Backend> State<B> {
 			|a| a.as_ref().map_or(self.account_start_nonce, |account| *account.nonce()))
 	}
 
+	/// Whether the base storage root of an account remains unchanged.
+	pub fn is_base_storage_root_unchanged(&self, a: &Address) -> TrieResult<bool> {
```
niklasad1: why not just return `bool` here?

ethcore/src/externalities.rs
```diff
@@ -116,7 +116,12 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 	where T: Tracer, V: VMTracer, B: StateBackend
 {
 	fn initial_storage_at(&self, key: &H256) -> vm::Result<H256> {
-		self.state.checkpoint_storage_at(0, &self.origin_info.address, key).map(|v| v.unwrap_or(H256::zero())).map_err(Into::into)
+		if self.state.is_base_storage_root_unchanged(&self.origin_info.address)? {
+			self.state.checkpoint_storage_at(0, &self.origin_info.address, key).map(|v| v.unwrap_or(H256::zero())).map_err(Into::into)
+		} else {
+			warn!(target: "externalities", "Detected existing account 0x{:x} where a forced contract creation happened.", self.origin_info.address);
```
niklasad1:
```suggestion
			warn!(target: "externalities", "Detected existing account {:#x} where a forced contract creation happened.", self.origin_info.address);
```

ethcore/src/state/mod.rs
sorpaas: Below `State::ensure_cached` can fail, and in that case I think it's better to bail because it's a DB failure.

PR #10068, opened 2018-12-14
Merge Machine and WithRewards and make Machine an associated type of Engine 

tomaka:
Cleaning up some code.
Doesn't change any logic.


tomaka:
Oh, I didn't realize that my change requires a newer version of Rust.

tomaka:
Replace with #10071

PR #10071, opened 2018-12-15
Merge Machine and WithBalances

tomaka:
Cleaning up some code.
Doesn't change any logic.


tomaka:
Not everything requires the `WithRewards` bound.

PR #10081, opened 2018-12-18
Add --frozen when running cargo

tomaka:
This detects when a PR modifies a dependency in a `Cargo.toml` but doesn't update the `Cargo.lock`.


ordian:
Looks like this PR broke our CI:
> ________Validate build________
error: attempting to make an HTTP request, but --frozen was specified

I'm not sure why though, `test.sh` works locally for me.

PR #10086, opened 2018-12-19
Call private contract methods from another private contract (read-only) 

### LINKED ISSUE #9199: [PrivateTransactions] Call of private contract's methods inside the other private contract
The author of the private contract should be able to call methods of other private contracts in the methods of original private contract

grbIzl:
This PR adds possibility to make calls of private contract methods from another private contract (with condition, that caller has access to the keys for both private contracts).
In order to achieve this, ABI of permissioning contract of Secret Store was extended with availableKeys method, that returns all keys, available for the account.
The further solution is rather straightforward: after patching the state for the current private contract, we also patch states for all private contracts, available for the node, and execute the private transaction on this state.
Some details of the solution:
- Only read-only calls are supported. Client can modify the state of other private contracts during execution of the private transaction, but these changes won't be saved
- All code, responsible for work with Secret Store keys, moved to the separate module

Closes #9199

### DIFF
--- a/ethcore/private-tx/src/key_server_keys.rs
+++ b/ethcore/private-tx/src/key_server_keys.rs
@@ -1,0 +1,135 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Wrapper around key server responsible for access keys processing.
+
+use std::sync::Arc;
+use parking_lot::RwLock;
+use ethereum_types::{H256, Address};
+use ethcore::client::{BlockId, CallContract, Client, RegistryInfo};
+use ethabi::FunctionOutputDecoder;
+
+const ACL_CHECKER_CONTRACT_REGISTRY_NAME: &'static str = "secretstore_acl_checker";
+
+use_contract!(keys_acl_contract, "res/keys_acl.json");
+
+/// Returns the address (of the contract), that corresponds to the key
+pub fn key_to_address(key: &H256) -> Address {
+	Address::from_slice(&key.to_vec()[..10])
+}
+
+/// Returns the key from the key server associated with the contract
+pub fn address_to_key(contract_address: &Address) -> H256 {
+	// Current solution uses contract address extended with 0 as id
+	let contract_address_extended: H256 = contract_address.into();
+
+	H256::from_slice(&contract_address_extended)
+}
+
+/// Trait for keys server keys provider.
+pub trait KeyProvider: Send + Sync + 'static {
+	/// Account, that is used for communication with key server
+	fn key_server_account(&self) -> Option<Address>;
+
+	/// List of keys available for the account
+	fn available_keys(&self, block: BlockId, account: &Address) -> Option<Vec<Address>>;
+
+	/// Update permissioning contract
+	fn update_acl_contract(&self);
+}
+
+/// Secret Store keys provider
+pub struct SecretStoreKeys {
+	client: Arc<Client>,
+	key_server_account: Option<Address>,
+	keys_acl_contract: RwLock<Option<Address>>,
+}
+
+impl SecretStoreKeys {
+	/// Create provider
+	pub fn new(client: Arc<Client>, key_server_account: Option<Address>) -> Self {
+		SecretStoreKeys {
+			client,
+			key_server_account,
+			keys_acl_contract: RwLock::new(None),
+		}
+	}
+
+	fn keys_to_addresses(&self, keys: Option<Vec<H256>>) -> Option<Vec<Address>> {
+		keys.map(|key_values| {
+			let mut addresses: Vec<Address> = Vec::new();
+			for key in key_values {
+				addresses.push(key_to_address(&key));
+			}
+			addresses
+		})
+	}
+}
+
+impl KeyProvider for SecretStoreKeys {
+	fn key_server_account(&self) -> Option<Address> {
+		self.key_server_account
+	}
+
+	fn available_keys(&self, block: BlockId, account: &Address) -> Option<Vec<Address>> {
+		match *self.keys_acl_contract.read() {
+			Some(acl_contract_address) => {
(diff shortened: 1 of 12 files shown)


ethcore/sync/src/tests/private.rs
```diff
@@ -24,7 +24,7 @@ use ethcore::CreateContractAddress;
 use transaction::{Transaction, Action};
 use ethcore::executive::{contract_address};
 use ethcore::test_helpers::{push_block_with_transactions};
-use ethcore_private_tx::{Provider, ProviderConfig, NoopEncryptor, Importer, SignedPrivateTransaction};
+use ethcore_private_tx::{Provider, ProviderConfig, NoopEncryptor, Importer, SignedPrivateTransaction, StoringKeyProvider	};
```
sorpaas: nit: extra tab

ethcore/private-tx/src/lib.rs
```diff
@@ -511,25 +524,34 @@ impl Provider where {
 			let (new_address, _) = ethcore_contract_address(engine.create_address_scheme(env_info.number), &sender, &nonce, &transaction.data);
 			Some(new_address)
 		});
+		let contract_address = contract_address.expect("Private contract address should be non zero by this moment; qed");
+		// Patch other available private contracts' states as well
+		if let Some(key_server_account) = self.keys_provider.key_server_account() {
+			if let Some(available_contracts) = self.keys_provider.available_keys(block, &key_server_account) {
+				for private_contract in available_contracts {
```
sorpaas: This may be a performance bottleneck. I don't think we need to fix it immediately, but what we can do is:

* Add a hook in `Externalities::call` that checks the contract address.
* From there, only patch an account in the state if it's actually called.

grbIzl: Yes, you're right! TODO added into the code and the corresponding task #10133 created

5chdn:
Needs a 2nd review, could you have a look at this @dvdplm @tomusdrw ?

ethcore/private-tx/src/key_server_keys.rs
```diff
@@ -0,0 +1,148 @@
...
+	fn update_acl_contract(&self) {
+		let contract_address = self.client.registry_address(ACL_CHECKER_CONTRACT_REGISTRY_NAME.into(), BlockId::Latest);
+		let current_address = self.keys_acl_contract.read();
+
+		if *current_address != contract_address {
+			trace!(target: "privatetx", "Configuring for ACL checker contract from address {:?}",
+				contract_address);
+
+			let keys_acl_contract = self.keys_acl_contract.write();
+			keys_acl_contract.and(contract_address);
```
tomusdrw: This seems like a no-op, do we actually want to update the `keys_acl_contract`? Can we test this function?

ethcore/private-tx/src/key_server_keys.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Dummy keys provider.
+pub struct StoringKeyProvider {
+	available_keys: RwLock<Option<Vec<Address>>>,
+	key_server_account: Option<Address>,
+}
+
+impl StoringKeyProvider {
+	/// Store available keys
+	pub fn set_available_keys(&self, keys: &Vec<Address>) {
+		let mut current = self.available_keys.write();
```
tomusdrw: why not just `*self.available_keys.write() = Some(keys.clone())`?

ethcore/private-tx/src/key_server_keys.rs
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	fn available_keys(&self, block: BlockId, account: &Address) -> Option<Vec<Address>> {
+		match *self.keys_acl_contract.read() {
+			Some(acl_contract_address) => {
+				let (data, decoder) = keys_acl_contract::functions::available_keys::call(*account);
+				if let Ok(value) = self.client.call_contract(block, acl_contract_address, data) {
+					self.keys_to_addresses(decoder.decode(&value).ok())
+				} else {
+					None
```
tomusdrw: Should we log something here?

ethcore/private-tx/src/key_server_keys.rs
```diff
@@ -0,0 +1,148 @@
...
+		SecretStoreKeys {
+			client,
+			key_server_account,
+			keys_acl_contract: RwLock::new(None),
+		}
+	}
+
+	fn keys_to_addresses(&self, keys: Option<Vec<H256>>) -> Option<Vec<Address>> {
+		keys.map(|key_values| {
+			let mut addresses: Vec<Address> = Vec::new();
```
tomusdrw: `key_values.iter().map(key_to_address).collect()`?

ethcore/private-tx/src/lib.rs
```diff
@@ -511,25 +524,35 @@ impl Provider where {
 			let (new_address, _) = ethcore_contract_address(engine.create_address_scheme(env_info.number), &sender, &nonce, &transaction.data);
 			Some(new_address)
 		});
+		let contract_address = contract_address.expect("Private contract address should be non zero by this moment; qed");
```
tomusdrw: `should be` doesn't sound like a proof, where exactly is it supposed to be initialised? Can we reach this line if private transactions are not part of the registry?

ethcore/private-tx/src/lib.rs
```diff
@@ -511,25 +524,35 @@ impl Provider where {
...
+				Some(c) => Some(self.encrypt(&contract_address, &Self::iv_from_address(&contract_address), &c)?),
+				None => None,
+			};
+			(enc_code, self.encrypt(&contract_address, &Self::iv_from_transaction(transaction), &Self::snapshot_from_storage(&storage))?)
 		};
 		Ok(PrivateExecutionResult {
 			code: encrypted_code,
 			state: encrypted_storage,
-			contract_address,
+			contract_address: Some(contract_address),
```
tomusdrw: Maybe it's worth to change the type if it's always `Some`?

ethcore/private-tx/src/key_server_keys.rs
grbIzl: You're right. Sorry about that. Fixed, test added

ethcore/private-tx/src/key_server_keys.rs
grbIzl: No, it's not an exceptional situation

ethcore/private-tx/src/lib.rs
grbIzl: Re-phrased with more explicit proof.

ethcore/private-tx/src/lib.rs
grbIzl: Yep

ethcore/private-tx/src/lib.rs
```diff
@@ -512,25 +525,36 @@ impl Provider where {
 			let (new_address, _) = ethcore_contract_address(engine.create_address_scheme(env_info.number), &sender, &nonce, &transaction.data);
 			Some(new_address)
 		});
+		let contract_address = contract_address.expect("Private contract address is non zero by this point, \
```
tomusdrw: Maybe change the above line to `unwrap_or_else(||` ? 
The proof won't be necessary and also the contract address will be evaluated lazily.

PR #10088, opened 2018-12-19
Implement caching for service transactions checker

VladLupashevskyi:
In cases when certain address is allowed for executing free transactions it makes sense in my opinion to submit transactions which are automatically submitted by the client (such as validators reporting etc) with the gas price set to zero.

Here I propose my code improvement to achieve this by adding check for free tx certification in transact_contract fn which is called while doing validators reporting and other actions.

Looking forward for the feedback :)


ethcore/src/client/client.rs
```diff
@@ -2112,10 +2113,15 @@ impl BlockChainClient for Client {
...
+		let service_transaction_checker = ServiceTransactionChecker::default();
+		let gas_price = match service_transaction_checker.check_address(self, address) {
+			Ok(true) => U256::zero(),
+			_ => self.importer.miner.sensible_gas_price(),
+		};
 		let transaction = Transaction {
 			nonce: self.latest_nonce(&authoring_params.author),
 			action: Action::Call(address),
-			gas: self.importer.miner.sensible_gas_limit(),
+			gas: gas_price,
```
tomusdrw: I think you meant to change `gas_price` not `gas`.

ethcore/src/miner/service_transaction_checker.rs
```diff
@@ -48,4 +49,13 @@ impl ServiceTransactionChecker {
 		let value = client.call_contract(BlockId::Latest, address, data)?;
 		decoder.decode(&value).map_err(|e| e.to_string())
 	}
+
+	/// Checks if given address is whitelisted to send service transactions.
+	pub fn check_address<C: CallContract + RegistryInfo>(&self, client: &C, sender: Address) -> Result<bool, String> {
```
tomusdrw: This should be used internally by `check` to avoid code duplication

ethcore/src/client/client.rs
```diff
@@ -2112,10 +2113,15 @@ impl BlockChainClient for Client {
 
 	fn transact_contract(&self, address: Address, data: Bytes) -> Result<(), transaction::Error> {
 		let authoring_params = self.importer.miner.authoring_params();
+		let service_transaction_checker = ServiceTransactionChecker::default();
+		let gas_price = match service_transaction_checker.check_address(self, address) {
```
tomusdrw: `address` is contract address you try to call, not the sender (which should be checked here).

ethcore/src/miner/mod.rs
```diff
@@ -20,7 +20,7 @@
 //! Keeps track of transactions and currently sealed pending block.
 
 mod miner;
-mod service_transaction_checker;
+pub mod service_transaction_checker;
```
tomusdrw: please group with other `pub mod`

VladLupashevskyi:
@tomusdrw I pushed the fixes and sorry for these typos.

I am not sure what kind of deadlocks or data races could occur because free tx permission is checked for the latest mined block and not for the pending one. What I can think of is that issues could occur if we would call `transact_contract` inside of `check_address` function, but calling state modifying function inside state checking one does not make sense for me.

Maybe I am missing something, so I would be happy to get more feedback and dig into solving it :)

ethcore/src/miner/service_transaction_checker.rs
```diff
@@ -29,23 +30,24 @@ const SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME: &'static str = "service_transa
...
-		let address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest)
-			.ok_or_else(|| "contract is not configured")?;
-
-		trace!(target: "txqueue", "[{:?}] Checking service transaction checker contract from {}", hash, sender);
+		self.check_address(client, sender)
+	}
 
+	/// Checks if given address is whitelisted to send service transactions.
+	pub fn check_address<C: CallContract + RegistryInfo>(&self, client: &C, sender: Address) -> Result<bool, String> {
+		let contract_address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest)
```
sorpaas: Would the performance be a concern here? `registry_address` would need to query the state if `registrar` is set.

tomusdrw: I don't think that an issue, some cache would be nice (best would be if previously queried registry entries would be refreshed after a block is imported), but the code is not super hot (only in `\.transact` which is called by the engine, or by the miner, called only for zero-gas-price transactions attempting to enter the pool).

ethcore/src/client/client.rs
```diff
@@ -353,6 +354,14 @@ impl Importer {
 
 		let db = client.db.read();
 		db.key_value().flush().expect("DB flush failed.");
+		match ServiceTransactionChecker::default().refresh_cache(client) {
```
tomusdrw: I think this should rather be handled as `ChainNotify`

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+struct CertifiedAddressesCache {
+	cache: RwLock<BTreeMap<Address, bool>>,
+	last_block: u64
+}
+
 /// Service transactions checker.
 #[derive(Default, Clone)]
-pub struct ServiceTransactionChecker;
+pub struct ServiceTransactionChecker {
+	certified_addresses_cache: CertifiedAddressesCache
```
tomusdrw: This means that cache is present only for a single instance. We create a new instance every time using `ServiceTransactionChecker::default()` both in `transact_contract` and `import_block` so in the first case the cache is always dropped after the check and in the second case we `refresh_cache` that is empty.

To make it work we either have to make the cache static (not really recommended) or hold a single instance of cache for the whole time. (either by sharing one instance of `ServiceTransactionChecker` or by passing the cache as `Arc<RwLock<..>>` in the `ServiceTransactionChecker::new()` function (so we get rid of `Default`).

To be honest I think I'd prefer to have a single instance (or rather `Arc<ServiceTxChecker>`; ideally shared between miner and client).

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
 		}
 
-		let address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest)
+		self.check_address(client, sender)
+	}
+
+	/// Checks if given address is whitelisted to send service transactions.
+	pub fn check_address<C: CallContract + RegistryInfo + ChainInfo>(&self, client: &C, sender: Address) -> Result<bool, String> {
+		let cache = self.certified_addresses_cache.cache.try_read();
+		if (client.chain_info().best_block_number as u64 == self.certified_addresses_cache.last_block) && cache.is_some() {
```
tomusdrw: to avoid `unwrap` in the second line you can try this:
```
if client.chain_info().best_block_number == self.cache_last_block {
   let cache = self.cache.cache.try_read();
   if let Some(allowed) = cache.and_then(|x| x.get(&sender)) {
        return Ok(*allowed);
   }
}
```

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+			}
+		}
+		let contract_address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest)
 			.ok_or_else(|| "contract is not configured")?;
+		trace!(target: "txqueue", "Checking service transaction checker contract from {}", sender);
+		let (data, decoder) = service_transaction::functions::certified::call(sender);
+		let value = client.call_contract(BlockId::Latest, contract_address, data)?;
+		decoder.decode(&value).and_then(|allowed| {
+			let cache = self.certified_addresses_cache.cache.try_write();
+			if cache.is_some() {
```
tomusdrw:
```
if let Some(cache) = cache {
  cache.insert(sender, allowed);
}
```

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+	pub fn check_address<C: CallContract + RegistryInfo + ChainInfo>(&self, client: &C, sender: Address) -> Result<bool, String> {
+		let cache = self.certified_addresses_cache.cache.try_read();
+		if (client.chain_info().best_block_number as u64 == self.certified_addresses_cache.last_block) && cache.is_some() {
+			if let Some(allowed) = cache.unwrap().get(&sender) {
+				return Ok(*allowed);
+			}
+		}
+		let contract_address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest)
 			.ok_or_else(|| "contract is not configured")?;
+		trace!(target: "txqueue", "Checking service transaction checker contract from {}", sender);
```
tomusdrw: I think the log should be displayed even if we hit the cache

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+			Ok(allowed)
+		}).map_err(|e| e.to_string())
+	}
 
-		trace!(target: "txqueue", "[{:?}] Checking service transaction checker contract from {}", hash, sender);
+	/// Refresh certified addresses cache
+	pub fn refresh_cache<C: CallContract + RegistryInfo + ChainInfo>(&mut self, client: &C) -> Result<bool, String> {
+		trace!(target: "txqueue", "Refreshing certified addresses cache");
+		let contract_address = client.registry_address(SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME.to_owned(), BlockId::Latest);
+		if contract_address.is_none() {
```
tomusdrw: We should clear the cache in that case.

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+			return Ok(false)
+		}
+		let mut updated_addresses: BTreeMap<Address, bool> = BTreeMap::default();
+		let cache = self.certified_addresses_cache.cache.try_read();
+		if cache.is_some() {
```
tomusdrw: Using `try_read()` here means that we won't update the cache if we are unlucky. I'd rather go for more predictable behavior.

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+			for (address, allowed_before) in cache.unwrap().iter() {
+				let (data, decoder) = service_transaction::functions::certified::call(*address);
+				let value = client.call_contract(BlockId::Latest, contract_address.unwrap(), data)?;
```
tomusdrw: This logic should be reused

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+					updated_addresses.insert(*address, allowed);
+				}
+			};
+			let cache = self.certified_addresses_cache.cache.try_write();
+			if cache.is_some() {
+				let mut unwrapped_cache = cache.unwrap();
+				for (address, allowed) in updated_addresses.iter() {
+					unwrapped_cache.insert(*address, *allowed);
+				}
+				self.certified_addresses_cache.last_block = client.chain_info().best_block_number as u64;
```
tomusdrw: So this normally would need to be under a lock as well, but since your function is `&mut self` (you are modifying only one instance) you don't need it.

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+		} else {
+			Ok(false)
+		}
+	}
+}
 
-		let (data, decoder) = service_transaction::functions::certified::call(sender);
-		let value = client.call_contract(BlockId::Latest, address, data)?;
-		decoder.decode(&value).map_err(|e| e.to_string())
+impl Clone for CertifiedAddressesCache {
```
tomusdrw: We probably don't need `Clone`, remove the `derive(Clone)` from the entire checker.

miner/src/service_transaction_checker.rs
tomusdrw: and btw `HashMap` should be fine instead of `BTreeMap`.

miner/src/service_transaction_checker.rs
tomusdrw: Actually calling `chain_info()` might be quite heavy, let's just ditch that?

miner/src/service_transaction_checker.rs
```diff
@@ -16,37 +16,111 @@
...
+		if cache.is_some() {
+			for (address, allowed_before) in cache.unwrap().iter() {
+				let (data, decoder) = service_transaction::functions::certified::call(*address);
+				let value = client.call_contract(BlockId::Latest, contract_address.unwrap(), data)?;
+				let allowed = decoder.decode(&value).map_err(|e| e.to_string())?;
+				if *allowed_before != allowed {
+					updated_addresses.insert(*address, allowed);
+				}
+			};
+			let cache = self.certified_addresses_cache.cache.try_write();
```
tomusdrw: I would suggest a little bit different logic to refresh the cache:
```rust
// replace the cache with an empty list,
// since it's not recent it won't be used anyway.
let mut cache = mem::replace(&mut self.cache.write(), HashMap::default());
let addresses: Vec<_> = cache.keys().collect();
for address in addresses {
   // make a call
   let is_allowed = ...;
   // update the entry
   cache.insert(address, is_allowed);
}
mem::replace(&mut self.cache.write(),  cache);
```
This means that during the update cache is not used at all.

tomusdrw: On a second though if the cache is reused between client and miner we would get into a potential attack vector if the cache is never cleared and not limited.

VladLupashevskyi:
The PR #10232 without cache has been created.

I will continue with cache implementation in this PR.

miner/src/service_transaction_checker.rs
VladLupashevskyi: The `Clone` is used here: 

https://github.com/paritytech/parity-ethereum/blob/4b11d79829d028e091b19af077767051574cd837/ethcore/src/miner/pool_client.rs#L89

Should we use something like `ServiceTransactionChecker::new()` instead to get rid of `.clone()` there?

miner/src/service_transaction_checker.rs
VladLupashevskyi: I used `try_read()` and `try_write()` because it was not clear to me how many threads will be using the cache, as far as I can see now - there will be two of them: one for client and one for miner, so I'm still not sure how to guarantee that cache won't be accessed while another thread will be writing into it.

I can think about implementing some waiting when the cache is busy, but still I doubt whether it's a proper solution.

miner/src/service_transaction_checker.rs
VladLupashevskyi: I agree, once it's possible to guarantee that the cache is 100% updated on the new block we can get rid of `cache_last_block `.

miner/src/service_transaction_checker.rs
VladLupashevskyi: Sorry, I think I misunderstood a bit how the RwLock works. I have done some tests now and I saw that even if the cache is used by another thread writing into it, the `read` or `write` won't fail and they will be just waiting for the lock to release, so I believe it's safe to use `read` and `write` instead of `try_read` and `try_write`. Correct me please if I'm wrong.

miner/src/service_transaction_checker.rs
tomusdrw: So if we `Clone` the checker we should actually share the cache between instances - wrap it in `Arc` please.

miner/src/service_transaction_checker.rs
tomusdrw: @VladLupashevskyi that's correct `read()` will block until it can acquire the lock, `try_read()` is a non-blocking version of it, but may fail in case cache is being currently updated (i.e. someone holds `.write()` lock).
While I think `try_read` is fine when we are checking the cache (if we are unlucky it means that the cache is being updated anyway - so probably it contains obsolete data), updating should block.

Regarding threads - we actually run inside a threadpool, so even though we have two structs using this checker (`Miner`/`Client`) they might be invoked on multiple threads in the pool in the same time.

miner/src/service_transaction_checker.rs
tomusdrw: I'd go for `try_read()` here and `.write()` when cache is updated - that way instead of blocking for the cache to be ready we will just do additional state call, which shouldn't really be that heavy.

VladLupashevskyi:
@tomusdrw I think idea with moving ServiceTxChecker to one place sounds reasonable! I would create a public method to get the checker instance from miner, so it's done more generic than just exposing a `check` method.
So I'm gonna dig into this and will notify about the results!

VladLupashevskyi:
@tomusdrw  I updated the code according to the comments, now it looks much cleaner! I'm thinking it would be nice to introduce some cache limiting, I would propose to use the solution I introduced in another PR for NodeFilter: 

https://github.com/paritytech/parity-ethereum/blob/2202646acefc04e4a26c6fed3952728f0c355762/ethcore/node-filter/src/lib.rs#L107

where basically I used `Vec` in order to store sequence of recent calls to the contract and remove the oldest ones when cache hits limit.

What do you think about it?

5chdn:
Could you revisit this,  @tomusdrw

ethcore/src/miner/miner.rs
```diff
@@ -258,6 +260,7 @@ impl Miner {
 		let verifier_options = options.pool_verification_options.clone();
 		let tx_queue_strategy = options.tx_queue_strategy;
 		let nonce_cache_size = cmp::max(4096, limits.max_count / 4);
+		let refuse_service_transactions = options.refuse_service_transactions.clone();
```
tomusdrw: nit: `.clone()` not required, `bool` is `Copy`.

ethcore/src/miner/miner.rs
```diff
@@ -1268,6 +1281,17 @@ impl miner::MinerService for Miner {
 				self.transaction_queue.cull(client);
 			}
 		}
+		if self.service_transaction_checker.is_some() {
```
tomusdrw: please use `if let Some(checker) = self.service_transaction_checker`

ethcore/src/miner/miner.rs
```diff
@@ -1268,6 +1281,17 @@ impl miner::MinerService for Miner {
...
+			match self.service_transaction_checker.clone().unwrap().refresh_cache(chain) {
```
tomusdrw: If you need to `clone()` there is most likely something bad happening here.

ethcore/src/miner/pool_client.rs
```diff
@@ -109,11 +109,7 @@ C: BlockInfo + CallContract,
 			engine,
 			accounts,
 			best_block_header,
-			service_transaction_checker: if refuse_service_transactions {
-				None
-			} else {
-				Some(Default::default())
-			},
+			service_transaction_checker
```
tomusdrw: nit: missing trailing comma

miner/src/service_transaction_checker.rs
```diff
@@ -16,21 +16,32 @@
...
+use parking_lot::RwLock;
 
 use_contract!(service_transaction, "res/contracts/service_transaction.json");
 
 const SERVICE_TRANSACTION_CONTRACT_REGISTRY_NAME: &'static str = "service_transaction_checker";
 
 /// Service transactions checker.
-#[derive(Default, Clone)]
-pub struct ServiceTransactionChecker;
+#[derive(Clone)]
```
tomusdrw: Do we need `Clone`?

miner/src/service_transaction_checker.rs
```diff
@@ -16,21 +16,32 @@
...
+pub struct ServiceTransactionChecker {
+	certified_addresses_cache: Arc<RwLock<HashMap<Address, bool>>>
+}
 
 impl ServiceTransactionChecker {
+	/// Create new ServiceTransactionChecker instance
+	pub fn new() -> ServiceTransactionChecker {
```
tomusdrw: Why not `Default` as it was previously?

miner/src/service_transaction_checker.rs
```diff
@@ -44,9 +55,42 @@ impl ServiceTransactionChecker {
 
 	/// Checks if given address is whitelisted to send service transactions.
 	pub fn check_address<C: CallContract + RegistryInfo>(&self, client: &C, sender: Address) -> Result<bool, String> {
+		trace!(target: "txqueue", "Checking service transaction checker contract from {}", sender);
+		if let Some(allowed) = self.certified_addresses_cache.try_read().as_ref().and_then(|c| c.get(&sender)){
```
tomusdrw: `.as_ref()` seems redundant, also missing space before `{`

miner/src/service_transaction_checker.rs
VladLupashevskyi: We need `Clone` in order to share `ServiceTransactionChecker` with `Client` and `PoolClient`.

miner/src/service_transaction_checker.rs
VladLupashevskyi: I cannot do this without `.as_ref()` since  the `c` doesn't live long enough in this case. I tried to do it with `.map`, but in this case `Option<Option<&bool>>` is returned, so one more check is needed. Not sure what would be better solution here.

soc1c:
Needs a 2nd review :+1:

PR #10094, opened 2018-12-21
[WIP] Add EIP1024 methods

### LINKED ISSUE #9893: Implement EIP 1024 encryption methods
- **Parity Ethereum version**: 2.1.4
- **Operating system**: all
- **Installation**: all
- **Fully synchronized**: yes
- **Network**: all
- **Restarted**: yes

## Current Behavior

Geth does not currently provide any encryption methods

Parity currently provides a data encryption method using elliptic curve secp256k1. 

## Wanted Behavior

A new encryption method that works across different key management softwares. For example a sender could encrypt data with their MetaMask wallet and the receiver could decrypt it with Parity. This new encryption method will use curve25519 as specified in [EIP 1024](https://github.com/ethereum/EIPs/pull/1098). 

 [EIP 1024](https://github.com/ethereum/EIPs/pull/1098) uses a well audited library called [nacl](https://nacl.cr.yp.to/). The rust implementation of the necessary nacl functions can be found [here](https://github.com/javadevelopr/nacl-mini) 

##Definition of Done

You have successfully completed this bounty, when:

1. A developer can request an accounts encryption public key
2. A developer can encrypt any arbitrary data with using elliptical curve25519
3. A developer can decrypt any arbitrary data encrypted with their accounts encryption public key
4. These methods are compatible with MetaMasks encryption methods listed (here)[https://github.com/MetaMask/eth-sig-util]

zachzundel:
Still in progress, but would like some review on current approach

Fixes #9893


zachzundel:
[clabot:check]

rpc/src/v1/impls/eth.rs
```diff
@@ -995,4 +996,27 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
+		Err(errors::light_unimplemented(None))
+    }
+
+    fn encrypt(&self, encryption_public_key: RpcH256, version: String, data: Bytes, padding: Trailing<RpcU64>) -> Result<Bytes> {
+        let mut data_vec: Vec<u8> = data.into();
+        let mut message = data_vec.as_mut_slice();
+        let public_key_bytes: [u8; 32] = encryption_public_key.0;
+        let ad: [u8; 0] = [0; 0];
+        let mut nonce = vec![0; 12];
+        let rand = SystemRandom::new();
```
tomusdrw: Is there a way to use a deterministic source for tests or supply `nonce` externally for the function so that we can write some tests?

rpc/src/v1/impls/eth.rs
```diff
@@ -995,4 +996,27 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
+
+        let sealing_key = aead::SealingKey::new(&aead::AES_256_GCM, &public_key_bytes).unwrap();
```
tomusdrw: All `unwraps` should be turned into RPC errors (most likely add requried `From<_>/Into<_>` implementation

rpc/src/v1/impls/eth.rs
```diff
@@ -995,4 +996,27 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
+        rand.fill(&mut nonce).unwrap();
+
+        aead::seal_in_place(&sealing_key, &nonce, &ad, &mut message, 0);
```
tomusdrw: why do we use `aead`? Afair the issue is about a different scheme?

tomusdrw: @topealabi just joined, same handle as github. Could you invite me to appropriate room?

5chdn:

zachzundel:
Yes, sorry, I will work for a bit longer and if I can't get it figured out
I will release it to the main pool.

soc1c:
Hey Zach, please reopen once you continue working on this :+1:

Issue #9997, opened 2018-11-30
Remove/split out account management


5chdn:
Ideally, an Ethereum client should not contain account management and keystore.

Keys and transactions should be managed by external applications. Now that we removed the wallet from the client, this is a breaking, but easy to implement change. Maybe we double down our efforts on `ethstore` and `ethkey`.

5chdn:
also: hardware-wallet support is unused: #9622

lazaridiscom:
related: #10027 - (Missing Subsystems-Overview)

lazaridiscom:
> Do you want to remove whole account (e.g. personal RPC api endpoint) management from parity?
It seems to me you love to break things for users.

No worries, even larger breaking changes can be introduced in increments and with a deprecation-phase.

Although seeing the lack of structure withing this project, I doubt that this will be done this way.

lazaridiscom:
> please follow the general discussion that is being had around this topic within the community for more information.

My comments affect the processing (how to apply the change, providing an architectural overview to simplify explaining things/changes). As for "separating verifier from account-manager", it seems like an obvious "must do".

gituser:
@folsen, regarding RPC API endpoints you can always disable those before starting parity, I don't see who will benefit if you remove/split those from the client.

For example: the UI split resulted in negative impact for parity many users moved out of it because of the complications in the installation and maintenance.

Regarding, security - you can always enforce best security default settings for node operators and let them decide what they want to open via settings.

I'm not saying I'm against changes - but they should be rolled out slowly and imo there are many other more important bugs out there which needs fixing. 

Geth still have signing and account management inside the client I believe, at least `v1.8.2` does:

```
VERSION:
   1.8.2-stable-b8b9f7f4
   
COMMANDS:
   account     Manage accounts
```

cemkod:
What is the alternative to personal RPC api endpoint? What should we use? The depreciation message says to look at this issue for alternatives, yet I couldn't see any alternatives here.

tomusdrw:
@cemkod The recommended way is to manage the accounts on your own (preferably via secure hardware wallet) and submit signed transactions via `eth_sendRawTransaction`. 

That said before removing the methods completely we will most likely provide an alternative via [JSON-RPC](https://github.com/tomusdrw/jsonrpc-proxy) proxy that would handle signing (for instance using https://github.com/ethereum/go-ethereum/tree/master/cmd/clef)

cemkod:
[...]

PR #10101, opened 2018-12-24
Move a bunch of stuff around

tomusdrw:
An attempt to clear the directory structure a bit:
- if crate is specific to one of the client "components" move it inside (e.g move `price-info` to `miner`)
- dissolve tiny utility crates (`stop_guard` & `devtools`)
- create a subdir for account management (#9997)
- unify directory names for crates (use hyphen instead of underscore)
- rename some crates (e.g. `mem -> memzero`, additionaly this one could potentially use `memsec` crate)


ordian:
Please update test.sh (should be `cd parity-clib/examples/cpp`): https://github.com/paritytech/parity-ethereum/blob/bf9fedc4ee2eaa4dcbc6fcb9ef73bdf6967ee071/test.sh#L62

PR #10102, opened 2018-12-24
Replace ethcore-logger with env-logger.

tomusdrw:
- `env_logger` is sufficient and we can avoid bringing additional path dependency.


tomaka:
Shouldn't we remove the `logger` crate entirely as well?

tomusdrw:
@tomaka It's still useful - allows to share the log format between main parity binary and whisper CLI + exposes latest logs via RPC method. None of this is super critical, so I guess we could remove it entirely (and integrate the setup with the main crate) if you believe it's the way to go.

tomaka:
> if you believe it's the way to go.

I think this is the way to go, but we can do it later.

sorpaas:
Note that if we're really in the path to remove `logger` completely, we need to at least preserve those lines (https://github.com/paritytech/parity-ethereum/blob/b5f510ead742b1ac4c8e398339091e31926b07aa/logger/src/lib.rs#L70).

secret-store/src/key_server.rs
```diff
@@ -325,7 +325,7 @@ pub mod tests {
 
 	#[test]
 	fn document_key_generation_and_retrievement_works_over_network_with_single_node() {
-		//::logger::init_log();
+		//::env_logger::try_init();
```
sorpaas: Not caused by this PR, but can we remove those lines as they are commented out?

tomusdrw:
I've removed `ethcore-logger` from `whisper/cli` as well and moved it to `parity` directory.

So now it's only required by `rpc` as well (to get the `RotatingLogger`), but I've added a warning to that RPC method that it's going to be removed in the future, so we remove it gracefuly.

secret-store/src/key_server.rs
```diff
@@ -347,7 +347,7 @@ pub mod tests {
 
 	#[test]
 	fn document_key_generation_and_retrievement_works_over_network_with_3_nodes() {
-		//::logger::init_log();
+		::env_logger::try_init().ok();
```
sorpaas: Maybe `let _ = ::env_logger::try_init();`?

secret-store/src/key_server_cluster/cluster.rs
```diff
@@ -1360,7 +1360,7 @@ pub mod tests {
 
 	#[test]
 	fn error_in_generation_session_broadcasted_to_all_other_nodes() {
-		//::logger::init_log();
+		//::env_logger::try_init();
```
sorpaas: And uncomment those as well!

5chdn:
Thanks. Could you resolve the conflicts?

abunsen:
I was brought here by the RPC method `parity_devLogs` - will logging still be in the RPC?

PR #10109, opened 2018-12-28
Allow sealing when bugfix hardforks activate later than block 0

jwasinger:
Some networks (like Rinkeby) rely on this.

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -1095,13 +1095,12 @@ impl miner::MinerService for Miner {
 			None => return,
 		};
 
-		// refuse to seal the first block of the chain if it contains hard forks
+		// warn if the first block of the chain contains hard forks
 		// which should be on by default.
 		if block.block().header().number() == 1 {
 			if let Some(name) = self.engine.params().nonzero_bugfix_hard_fork() {
 				warn!("Your chain specification contains one or more hard forks which are required to be \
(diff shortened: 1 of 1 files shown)


jwasinger:
I'd also suggest making the message a trace instead of warning.  That way, it won't clog the output unless tracing for the miner is turned on.

sorpaas:
Or at least, @jwasinger, if you really need this, I would advise to do it like this:

* Add a new command line flag, like `--allow-insecure-first-block-sealing`.
* Keep current behavior, and only skip the check if the flag is on.

In this way, we make sure the configurators know what they're doing before proceeding. There're only three bugfix hard forks EIP155, `validate_receipts`, `validate_chain_id`. All of which if a network disables it, it can be really insecure.

sorpaas:
This if branch can only be triggered on Rinkeby if it fails to connect to any other peers (and tries to seal again another block 1). In all other cases, it's either a bug in the client or the config file.

The check won't affect any established networks (Foundation, Rinkeby, etc), but will only warn and stop sealing block 1 for a new network.

5chdn:
@jwasinger does this do anything beyond printing the warning?

jwasinger:
@sorpaas @5chdn I think I was mistaken and that my client was actually just taking an unusually long amount of time to find an connect to peers...

One side effect of not including this PR is that users will be able to sync to Rinkeby, but if they try and spin up a private network with the same configuration, it will fail.

PR #10114, opened 2018-12-28
Extract blockchain from ethcore

tomusdrw:
- Move out `blockchain` database from `ethcore`
- Separate` ethcore-db`
- Move a bunch of types to `common-types`
- Integrate `ethcore-transaction` with `common-types`
- Remove a bunch of re-exports


ethcore/blockchain/src/lib.rs
```diff
@@ -16,22 +16,22 @@
 
 //! Blockchain database.
 
+#[warn(missing_docs)]
```
niklasad1:
```suggestion
#![warn(missing_docs)]
```

ethcore/src/lib.rs
```diff
@@ -57,61 +57,64 @@
...
+extern crate crossbeam;
+extern crate ethabi;
 extern crate ethash;
+extern crate ethcore_blockchain as blockchain;
 extern crate ethcore_bloom_journal as bloom_journal;
-extern crate parity_crypto;
+extern crate ethcore_db as db;
 extern crate ethcore_io as io;
-extern crate parity_bytes as bytes;
-extern crate ethcore_logger;
```
cheme: Running on nightly 1.33 I had to restore this line (extern ethcore_logger) to run test from ethcore.

cheme: In fact removing #[cfg(feature = 'json-tests'] from the new import works better.

ethcore/src/transaction_ext.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
cheme: If I understand correctly, those method were not moved to common-types because of evm::Schedule dependency. Is moving evm::Schedule to common-type a possiblity (deps seems ok for that)?

cheme: I made a mistake when checking deps, indeed it would require evm to depend on common-type.

ethcore/vm/src/env_info.rs
```diff
@@ -20,9 +20,10 @@ use std::cmp;
 use std::sync::Arc;
 use hash::keccak;
 use ethereum_types::{U256, H256, Address};
-use types::BlockNumber;
 use ethjson;
 
+type BlockNumber = u64;
```
cheme: Is including deps on common-type for vm an issue ?

ethcore/blockchain/src/lib.rs
```diff
@@ -16,22 +16,22 @@
...
 mod best_block;
 mod block_info;
 mod blockchain;
 mod cache;
 mod config;
-mod extras;
 mod import_route;
 mod update;
 
-#[cfg(test)]
```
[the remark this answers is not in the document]
tomusdrw: I think it's not really worth it - we use it only in a single place outside of `ethcore-blockchain` (snapshot tests).
Feature flags are difficult to maintain and the crate has to be compiled twice during the test run.
Unused code will probably be stripped away by the compiler anyway.

ethcore/src/lib.rs
tomusdrw: 1. It's better to run `cargo test -p ethcore` from the maindir (you get the same `Cargo.lock`).
2. It actually requires `#[cfg(any(test, feature = "json-tests"))]`

ethcore/src/transaction_ext.rs
tomusdrw: The thing is that I don't think it fits the transaction interface at all. IMHO it's better to have "dumb" types and implement domain-specific logic in a separate structs.
Transaction gas calculation fits more into `machine` than to transaction itself, hence the `Schedule` dependency. I didn't want to refactor the code (the PR tries to just move things around), but I think it's something worth doing in the (probably distant) future.

ethcore/vm/src/env_info.rs
tomusdrw: I think it's pretty redundant since we only need to import a typedef for `u64`. It doesn't really bring any value imho.
Maybe we can put `BlockNumber` into `ethereum-types`?

ethcore/blockchain/src/generator.rs
```diff
@@ -19,51 +19,65 @@
...
 	}
 
+	/// Get block difficulty
 	#[inline]
 	pub fn difficulty(&self) -> U256 {
 		*self.header.difficulty()
 	}
 }
 
+/// Specify block options for generatro
```
sorpaas:
```suggestion
/// Specify block options for generator
```

ethcore/src/engines/mod.rs
```diff
@@ -24,28 +24,30 @@ mod signer;
...
 pub mod block_reward;
-pub mod epoch;
 
 pub use self::authority_round::AuthorityRound;
 pub use self::basic_authority::BasicAuthority;
 pub use self::epoch::{EpochVerifier, Transition as EpochTransition};
 pub use self::instant_seal::{InstantSeal, InstantSealParams};
 pub use self::null_engine::NullEngine;
 
+// TODO [ToDr] Remove re-export
```
sorpaas: It would be really helpful for others to pick up this TODO if you can create a new issue for this and refer the issue number here. And I think we should do this for any newly added TODOs!

ethcore/private-tx/src/lib.rs
```diff
@@ -25,27 +25,27 @@ mod private_transactions;
 mod messages;
 mod error;
 
+extern crate common_types as types;
```
ngotchac: Would it make sense to always use `common_types` (or always `types`) for consistency?

ngotchac: Sounds good to me! Maybe then `s/extern crate common_types;/extern crate common_types as types;/` ?

ethcore/blockchain/src/generator.rs
```diff
@@ -19,51 +19,65 @@
...
+use common_types::encoded;
+use common_types::header::Header;
+use common_types::transaction::SignedTransaction;
+use common_types::view;
+use common_types::views::BlockView;
 use rlp::encode;
-use transaction::SignedTransaction;
-use views::BlockView;
-use encoded;
+use rlp_derive::{RlpEncodable};
```
niklasad1:
```suggestion
use rlp_derive::RlpEncodable;
```

ethcore/db/src/keys.rs
```diff
@@ -19,14 +19,16 @@
 use std::io::Write;
 use std::ops;
 
-use db::Key;
-use engines::epoch::{Transition as EpochTransition};
+use common_types::BlockNumber;
+use common_types::engines::epoch::{Transition as EpochTransition};
```
niklasad1:
```suggestion
use common_types::engines::epoch::Transition as EpochTransition;
```

PR #10128, opened 2019-01-03
Identity fix

### LINKED ISSUE #10125: Service transactions not propagated when remote peers use `identity` config parameter
- **Parity Ethereum version**: master, also occurs with `v2.1.7-stable`.

- **Operating system**:  MacOS
- **Fully synchronized**: not applicable
- **Network**: private network
- **Restarted**: yes

Service transactions (with zero gas price) are not propagated to other nodes when using `identity` configuration flag.

This is due to the `accepts_service_transaction`function, which determine if remote peer can accept service transaction based on the client_id : 

https://github.com/paritytech/parity-ethereum/blob/b4f8bba8430b0f9e050cdac425b632b30f0f331c/ethcore/sync/src/chain/propagator.rs#L44-L57

With `identity` parameter, client_id does not start with `Parity-Ethereum/v` and peer is considered to not supporting service transactions.

Steps to reproduce: 
- Start a private chain
- Use `identity` parameter for nodes
- Use `eth_sendRawTransaction` to send a transaction to a non-mining node
- Transaction is never propagated nor mined.

joshua-mir:
fixes #10125 by looking at the right part of the client id.

### DIFF
--- a/ethcore/sync/src/chain/propagator.rs
+++ b/ethcore/sync/src/chain/propagator.rs
@@ -49,7 +49,7 @@ fn accepts_service_transaction(client_id: &str) -> bool {
 	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
 	const VERSION_PREFIX: &'static str = "/v";
 	
- 	let idx = client_id.rfind(VERSION_PREFIX).map(|idx| idx + VERSION_PREFIX.len()).unwrap_or(client_id.len());
+	let idx = client_id.rfind(VERSION_PREFIX).map(|idx| idx + VERSION_PREFIX.len()).unwrap_or(client_id.len());
 	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) || client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
 		client_id[idx..].split('.')
 	} else {
--- a/ethcore/sync/src/chain/propagator.rs
+++ b/ethcore/sync/src/chain/propagator.rs
@@ -50,7 +50,7 @@ fn accepts_service_transaction(client_id: &str) -> bool {
 	
 	let idx = client_id.find("/v").unwrap_or(client_id.len()) + 1;
 	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) || client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		client_id[idx.len()..].split('.')
+		client_id[idx..].split('.')
 	} else {
 		return false;
 	};
--- a/ethcore/sync/src/chain/propagator.rs
+++ b/ethcore/sync/src/chain/propagator.rs
@@ -48,7 +48,7 @@ fn accepts_service_transaction(client_id: &str) -> bool {
 	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
 	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
 	
-	let idx = client_id.rfind("/v").map(|idx| idx + 1).unwrap_or(client_id.len());
+ 	let idx = client_id.rfind("/v").map(|idx| idx + 2).unwrap_or(client_id.len());
 	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) || client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
 		client_id[idx..].split('.')
 	} else {
--- a/ethcore/sync/src/chain/propagator.rs
+++ b/ethcore/sync/src/chain/propagator.rs
@@ -48,7 +48,7 @@ fn accepts_service_transaction(client_id: &str) -> bool {
 	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
 	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
 	
-	let idx = client_id.find("/v").unwrap_or(client_id.len()) + 1;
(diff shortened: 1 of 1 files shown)


ethcore/sync/src/chain/propagator.rs
```diff
@@ -45,13 +45,12 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
-	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/v";
-
-	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
-		client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
-	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
+	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
+	
+	let idx = client_id.find("/v").unwrap_or(client_id.len()) + 1;
```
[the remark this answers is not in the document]
ordian: Also can identity arg start with `v` or even `v2.10`? 
I suggest
```suggestion
	let idx = client_id.rfind("/v").map(|idx| idx + 1).unwrap_or(client_id.len());
```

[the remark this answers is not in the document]
5chdn: This keeps confusing me for more than 3 years now. the `/v` is part of the version string directly following the client name. When I hacked this, I was assured the identity string will never be in that location.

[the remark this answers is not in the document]
ordian: @5chdn the code suggests identity is inserted before `/v`: https://github.com/paritytech/parity-ethereum/blob/801b8191ef14d6f7f9c63f9c41ac78ed360583c0/parity/configuration.rs#L743-L746

ethcore/sync/src/chain/propagator.rs
```diff
@@ -45,13 +45,12 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
-	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/v";
-
-	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
-		client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
-	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
+	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
+	
+	let idx = client_id.rfind("/v").map(|idx| idx + 1).unwrap_or(client_id.len());
```
ordian: Actually it should be `idx + 2`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8a040ab9427a3c201706194fcca13d7c
```suggestion
 	let idx = client_id.rfind("/v").map(|idx| idx + 2).unwrap_or(client_id.len());
```

joshua-mir: Should we change it to idx+"/v" .len() in this case? And change "/v" to a constant?

ordian: Yes, makes sense.

ethcore/sync/src/chain/propagator.rs
```diff
@@ -45,13 +45,13 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
-
-	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
-		client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
-	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
+	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
+	const VERSION_PREFIX: &'static str = "/v";
+	
+ 	let idx = client_id.rfind(VERSION_PREFIX).map(|idx| idx + VERSION_PREFIX.len()).unwrap_or(client_id.len());
```
ordian: space
```suggestion
	let idx = client_id.rfind(VERSION_PREFIX).map(|idx| idx + VERSION_PREFIX.len()).unwrap_or(client_id.len());
```

ethcore/sync/src/chain/propagator.rs
```diff
@@ -45,13 +45,13 @@ fn accepts_service_transaction(client_id: &str) -> bool {
...
-
-	let splitted = if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) {
-		client_id[LEGACY_CLIENT_ID_PREFIX.len()..].split('.')
-	} else if client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
-		client_id[PARITY_CLIENT_ID_PREFIX.len()..].split('.')
+	const LEGACY_CLIENT_ID_PREFIX: &'static str = "Parity/";
+	const PARITY_CLIENT_ID_PREFIX: &'static str = "Parity-Ethereum/";
+	const VERSION_PREFIX: &'static str = "/v";
+	
+	let idx = client_id.rfind(VERSION_PREFIX).map(|idx| idx + VERSION_PREFIX.len()).unwrap_or(client_id.len());
```
sorpaas: Would this break if a client's identity starts with `v`? How about split by `/` first and take the last part as version string?

PR #10135, opened 2019-01-04
 misc: bump license header to 2019

5chdn:
I'm sorry :laughing:


sorpaas:
I think the Cargo.lock issue is not caused by this PR. It should be able to be fixed by running `cargo` locally once.

5chdn:
We are doing this every year so developers can focus on code. #8666

And technically, these files just changed ;-)

PR #10137, opened 2019-01-06
Remove double fetch of block status

sorpaas:
`block_status` calls into `is_known` already, but we check (and lock) it twice. This PR fixes that.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -2160,18 +2160,16 @@ impl IoClient for Client {
 		let hash = unverified.hash();
 		{
 			// check block order
-			if self.chain.read().is_known(&hash) {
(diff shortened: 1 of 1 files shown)


ethcore/src/client/client.rs
```diff
@@ -2160,18 +2160,16 @@ impl IoClient for Client {
 		let hash = unverified.hash();
 		{
 			// check block order
-			if self.chain.read().is_known(&hash) {
+			let parent_hash = unverified.parent_hash();
+			let status = self.block_status(BlockId::Hash(parent_hash));
```
ascjones: Fetching the status of `parent_hash` now instead of `hash`

sorpaas: Yeah you're right. I think I got this PR wrong.

Issue #10127, opened 2019-01-03
ancient block sync panic: parent total difficulty always supplied for first block in chunk. only first block can have missing parent.


5chdn:
- **Parity Ethereum version**: 2.3.0 nightly @ 515108714
- **Operating system**: Arch Linux
- **Installation**: built from source (PR #10123 - but should be unrelated)
- **Fully synchronized**: no
- **Network**: classic
- **Restarted**: yes

```
❯ ./target/release/parity --chain classic --ports-shift 200 --reserved-peers ~/Desktop/classic
2019-01-03 12:15:14  Starting Parity-Ethereum/v2.3.0-nightly-515108714-20190103/x86_64-linux-gnu/rustc1.31.1
2019-01-03 12:15:14  Keys path /home/user/.local/share/io.parity.ethereum/keys/classic
2019-01-03 12:15:14  DB path /home/user/.local/share/io.parity.ethereum/chains/classic/db/906a34e69aec8c0d
2019-01-03 12:15:14  State DB configuration: fast
2019-01-03 12:15:14  Operating mode: active
2019-01-03 12:15:14  Configured for Ethereum Classic using Ethash engine
2019-01-03 12:15:21  Public node URL: enode://8dc97200ad854694b5c3ddfc9759907bcc1a7a817f750c0e80785dffab23d330428d04818e9540d2bbd9bc8294b0d2ff08b5975308418ea81d8c49c6faf091a3@192.168.43.131:30503
2019-01-03 12:15:24  Imported #7232868 0x5ab7…ef45 (5 txs, 0.26 Mgas, 38 ms, 1.16 KiB)
2019-01-03 12:15:29  Error importing ancient block: Block error (Seal element out of bounds: Expected 0xe0d0…bc7a, found 0x900b…b42c)


====================

stack backtrace:
   0:     0x5642c484c7cd - backtrace::backtrace::trace::h6a895d5d279a8d2f
   1:     0x5642c484b9e2 - <backtrace::capture::Backtrace as core::default::Default>::default::h03a8d770395d927e
   2:     0x5642c484ba59 - backtrace::capture::Backtrace::new::h2eb1c11c92f206be
   3:     0x5642c484ab96 - panic_hook::gen_panic_msg::hf38c962aa96f7e35
   4:     0x5642c3013e88 - panic_hook::set_with::{{closure}}::h2adb1ed3fd612fa1
[...]

PR #10138, opened 2019-01-06
Make sure parent block is not in importing queue when importing ancient blocks

sorpaas:
rel #10127. Partially fixes it.

* When queuing ancient blocks, we first check to make sure that the parent block is not unknown. However, this can also mean that the block is in the import queue. If this happens, ancient block importing would fail because it cannot find the parent block's total difficulty. This PR fixes the issue by only checking whether the parent block has been completely imported.
* in case an ancient import fails, all further imports wouldn't have parents, thus fetching parent total difficulty would panic. This PR fixes the issue by clearing the ancient queue if error happens.


ethcore/src/client/client.rs
```diff
@@ -2191,6 +2192,13 @@ impl IoClient for Client {
 				let first = queued.write().1.pop_front();
 				if let Some((unverified, receipts_bytes)) = first {
 					let hash = unverified.hash();
+					let parent_hash = unverified.parent_hash();
+					let status = client.importer.block_queue.status(&parent_hash);
```
ascjones: This queue is for new blocks. Ancient blocks are queued in `self.queued_ancient_blocks`. Could check `queued.read().0.contains(&parent_hash);` instead.

sorpaas: This checks whether the parent hash is in `importer`. The issue is that above when we report parent status, the block can be in the `importer`, thus break the assumption we have for ancient queues. I don't think it has anything to do with ancient block queues.

sorpaas:
I updated the PR with another case that would cause panic: in case an ancient import fails, all further imports wouldn't have parents, thus fetching parent total difficulty would panic.

ascjones:
Yes, and indeed this looks more likely to be the cause in #10127, just before the panic we get: `2019-01-03 12:15:29  Error importing ancient block: Block error (Seal element out of bounds: Expected 0xe0d0…bc7a, found 0x900b…b42c)`

ethcore/src/client/client.rs
```diff
@@ -2199,6 +2207,10 @@ impl IoClient for Client {
 					);
 					if let Err(e) = result {
 						error!(target: "client", "Error importing ancient block: {}", e);
+
+						let mut queued = queued.write();
+						queued.0.clear();
+						queued.1.clear();
```
ascjones: This should work - I believe in the next round it will attempt to download the next batch, which will fail with `UnknownParent`, and then start again from the last successfully imported ancient block. See https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L522

ethcore/src/client/client.rs
ascjones: Can you add a comment for why we are checking this? I understood after you explained the scenario, but it was not obvious to me reading the code initially.

ascjones: On second thoughts would it not be better to replace 

https://github.com/paritytech/parity-ethereum/blob/18fdcb429e514dc64f1ce0b16399768432a8b235/ethcore/src/client/client.rs#L2170-L2171

with `!self.chain.read().is_known(&parent_hash)`?  Currently `self.block_status` would indeed return `Known` if the block is in the normal queue, and then it would be rejected later by your check.

sorpaas: Yeah sounds like a better idea. In this scenario it would probably better to fail and let syncing to redo ancient block importing from a known good point completely.

PR #10154, opened 2019-01-08
fix(parity-clib): grumbles that were not addressed in #9920

niklasad1:
This PR introduces the following:

- remove a couple of needless unsafe functions
- some style nits
- eliminate repetitive event loops for `lib` and `java` (introduces a trait Callback to pass instead of the concrete type)


parity-clib/src/java.rs
```diff
@@ -62,22 +61,22 @@ impl<'a> Callback<'a> {
 }
 
 #[no_mangle]
-pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_configFromCli(env: JNIEnv, _: JClass, cli: jobjectArray) -> jlong {
-	let cli_len = env.get_array_length(cli).expect("invalid Java bindings");
+pub extern "system" fn Java_io_parity_ethereum_Parity_configFromCli(env: JNIEnv, _: JClass, cli: jobjectArray) -> jlong {
```
niklasad1: nit: this can end-up in UB if cli is malformed should be marked as unsafe

parity-clib/src/java.rs
```diff
@@ -96,19 +96,20 @@ pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_configFromCli(env:
 }
 
 #[no_mangle]
-pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_build(env: JNIEnv, _: JClass, config: va_list) -> jlong {
+pub extern "system" fn Java_io_parity_ethereum_Parity_build(env: JNIEnv, _: JClass, config: va_list) -> jlong {
```
niklasad1: nit: this can end-up in UB if `config` is malformed should be marked as unsafe

5chdn:
Needs a 2nd review

parity-clib/src/lib.rs
```diff
@@ -52,10 +52,32 @@ pub mod error {
 #[repr(C)]
 pub struct ParityParams {
 	pub configuration: *mut c_void,
-	pub on_client_restart_cb: Callback,
+	pub on_client_restart_cb: CCallback,
 	pub on_client_restart_cb_custom: *mut c_void,
 }
 
+/// Trait representing a callback that passes a string
+pub trait Callback: Send + Sync {
```
ordian: Does it need to be public?

parity-clib/src/java.rs
```diff
@@ -15,34 +15,31 @@
...
 	callback: GlobalRef,
 	method_name: &'a str,
 	method_descriptor: &'a str,
 }
 
-unsafe impl<'a> Send for Callback<'a> {}
-unsafe impl<'a> Sync for Callback<'a> {}
-impl<'a> Callback<'a> {
+unsafe impl<'a> Send for JavaCallback<'a> {}
+unsafe impl<'a> Sync for JavaCallback<'a> {}
```
ordian: It wasn't introduced by this PR, but could you explain why it's safe for [`GlobalRef`](https://docs.rs/jni/0.10.2/jni/objects/struct.GlobalRef.html)?

parity-clib/src/lib.rs
```diff
@@ -255,26 +214,70 @@ pub unsafe extern fn parity_unsubscribe_ws(session: *const c_void) {
...
+			// Note this may block forever and be can't destroyed using the session object
+			// However, this will likely timeout or be catched the RPC layer
+			if let Ok(Some(response)) = query_future.wait() {
+				callback.call(&response);
+			} else {
+				callback.call(error::SUBSCRIBE);
+				return;
+			}
+
+			while weak_session.upgrade().map_or(0, |session| Arc::strong_count(&session)) > 1 {
```
ordian: This is not equivalent to the loop used before if the condition returns false at the first iteration (is that ok?).

parity-clib/src/java.rs
```diff
@@ -139,26 +138,10 @@ pub unsafe extern "system" fn Java_io_parity_ethereum_Parity_rpcQueryNative(
...
-					let mut current_thread = CurrentThread::new();
-					current_thread.spawn(future);
-					let _ = current_thread.run_timeout(Duration::from_millis(timeout_ms as u64))
-						.map_err(|_e| {
-							cb.call(error::TIMEOUT);
-						});
-				})
-				.expect("rpc-query thread shouldn't fail; qed");
+			let callback = Arc::new(JavaCallback::new(jvm, global_ref));
+			parity_rpc_worker(client, &query, callback, timeout_ms as u64);
```
ordian: We still spawn a thread per call, correct?

niklasad1: Correct, I opened a ticket for this so we don't forget: https://github.com/paritytech/parity-ethereum/issues/10310

parity-clib/src/java.rs
niklasad1: I don't understand what you mean. Can you elaborate a bit?

ordian: Sorry, I meant why is it safe to impl Sync and Send for JavaCallback?

parity-clib/src/lib.rs
niklasad1: Yes, the `java mod` is using it too.

ordian: isn't `pub(crate)` enough?

parity-clib/src/lib.rs
niklasad1: Nice catch

Yeah, that is ok that means that the `sink` has been dropped and we should terminate. Before we entered loop but `rx.wait()` should unblock when sender is dropped and return `Ok(Ready(None))`

parity-clib/src/java.rs
niklasad1: That is a good question I have to admit I'm not completely sure (I more or less copy-pasted it from the `lib.rs`) but I argue that `JavaVM` and `GlobalRef` is never mutated and the `Java code` is thread-safe!

Also both [JavaVM](https://github.com/jni-rs/jni-rs/blob/master/src/wrapper/java_vm/vm.rs#L13-#L18) [GlobalRef](https://github.com/jni-rs/jni-rs/blob/master/src/wrapper/objects/global_ref.rs#L28-#L66) have unsafe `Send + Sync` impls but an inner-type `GlobalRefGuard` doesn't with the unsafe impls I get the following error:

```bash
53 | impl<'a> Callback for JavaCallback<'a> {
   |          ^^^^^^^^ `*mut jni::sys::_jobject` cannot be shared between threads safely
   |
   = help: within `jni::objects::global_ref::GlobalRefGuard`, the trait `std::marker::Sync` is not implemented for `*mut jni::sys::_jobject`
   = note: required because it appears within the type `jni::objects::JObject<'static>`
   = note: required because it appears within the type `jni::objects::global_ref::GlobalRefGuard`
   = note: required because of the requirements on the impl of `std::marker::Sync` for `std::sync::Arc<jni::objects::global_ref::GlobalRefGuard>`
   = note: required because it appears within the type `jni::objects::GlobalRef`
   = note: required because it appears within the type `java::JavaCallback<'a>`
```
Thoughts?

ordian: The reason I asked, is because the [docs for GlobalRef 0.10.2](https://docs.rs/jni/0.10.2/jni/objects/struct.GlobalRef.html) showed 
``` rust
impl !Sync for GlobalRef 
```
It seems like `Sync` was added later (https://github.com/jni-rs/jni-rs/pull/145), and the implementation hasn't changed for `GlobalRef`.

niklasad1: Thanks good point let’s ask them to bump the version so we can get rid of the unsafe part then :)

ordian: cc @DarkEld3r

PR #10157, opened 2019-01-09
snap: fix path in script

5chdn:
assumed the wrong path

https://gitlab.parity.io/parity/parity-ethereum/-/jobs/115402

### DIFF
--- a/scripts/gitlab/publish-snap.sh
+++ b/scripts/gitlab/publish-snap.sh
@@ -3,7 +3,10 @@
 set -e # fail on any error
 set -u # treat unset variables as error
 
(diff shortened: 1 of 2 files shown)


scripts/gitlab/publish-snap.sh
```diff
@@ -3,13 +3,13 @@
 set -e # fail on any error
 set -u # treat unset variables as error
 
-TRACK=`awk -F '=' '/^track/ {print $2}' ./util/version/Cargo.toml`
+TRACK=`awk -F '=' '/^track/ {print $2}' ../../util/version/Cargo.toml`
```
TriplEight: - fixed the choice of the track
- the path should start from the pkg source since `.gitlab-ci.yml` launches this file there.

Issue #10058, opened 2018-12-12
Remove dependency on libusb on Android


tomaka:
Cross-compiling to Android is quite tricky because of libusb-sys.
https://github.com/paritytech/libusb-sys/pull/4 fixed the fact that we were in reality compiling for Linux when trying to compile libusb for Android.

However this unveiled the problem that libusb actually doesn't support Android that well: https://github.com/paritytech/libusb-sys/blob/4418e92c82509d1f049f7942dffe62a0bf5dd871/libusb/libusb/core.c#L60

There are instructions to build libusb for Android here: https://github.com/paritytech/libusb-sys/blob/4418e92c82509d1f049f7942dffe62a0bf5dd871/libusb/android/README
However you are supposed to use `ndk-build` to do that, and we don't necessarily want to rewrite this tool in Rust.

Right now on Android we are [building libusb manually](https://github.com/paritytech/scripts/blob/a241620cb3410eea2903425111bb67e85463c61b/docker-files-for-Gitlab-CI-rust/cross/android/Dockerfile#L55-L70), but in order to do so we apply [a patch](https://github.com/paritytech/scripts/blob/a241620cb3410eea2903425111bb67e85463c61b/docker-files-for-Gitlab-CI-rust/cross/android/libudev.patch) on the source code, which is extremely hacky.
Libusb on Android seems to have quite [a few runtime issues](https://github.com/paritytech/libusb-sys/blob/4418e92c82509d1f049f7942dffe62a0bf5dd871/libusb/android/README#L92) as well.

A final solution is to remove the dependency on libusb on Android. This is probably the most simple solution, and the one I would suggest. Nobody uses our Android build yet, so we won't break anything.

I'm opening this issue to gather opinion. Any comment on this?

cc #9893 @TriplEight

ordian:
> However this unveiled the problem that libusb actually doesn't support Android that well: https://github.com/paritytech/libusb-sys/blob/4418e92c82509d1f049f7942dffe62a0bf5dd871/libusb/libusb/core.c#L60

I think for Android it defined OS_LINUX:
https://github.com/paritytech/libusb-sys/blob/4418e92c82509d1f049f7942dffe62a0bf5dd871/libusb/android/config.h#L36

PR #10161, opened 2019-01-09
fix(android): remove dependency to libusb

niklasad1:
Closes https://github.com/paritytech/parity-ethereum/issues/10058

```bash
# android
$  nm target/armv7-linux-androideabi/debug/parity | c++filt | grep libusb | wc -l
0
# native linux
$ nm target/debug/parity | c++filt | grep libusb | wc -l
45
```

### DIFF
--- a/ethcore/src/lib.rs
+++ b/ethcore/src/lib.rs
@@ -115,10 +115,10 @@ extern crate kvdb_rocksdb;
 #[cfg(any(test, feature = "blooms-db"))]
 extern crate blooms_db;
 
-#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "android"))]
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
 extern crate hardware_wallet;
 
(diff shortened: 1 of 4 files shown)


ethcore/src/lib.rs
```diff
@@ -115,10 +115,10 @@ extern crate kvdb_rocksdb;
 #[cfg(any(test, feature = "blooms-db"))]
 extern crate blooms_db;
 
-#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "android"))]
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
 extern crate hardware_wallet;
```
niklasad1: we could eliminate those with `rust 2018` instead of both have those feature flags in both `Cargo.toml` and `lib.rs`, see https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#cargo-features

TriplEight:
Now nothing (except of Android studio patching) stops a normal Android cross-compilation.
Soon I'll do a PR with CI changes.

PR #10164, opened 2019-01-10
backports for 2.3.0 beta

5chdn:
- [x] version: mark 2.3 track beta
- [x] version: mark update critical on all networks 
- [x] Handle the case for contract creation on an empty but exist account with storage items (#10065)
- [x] Fix _cannot recursively call into `Core`_ issue (#10144)
- [x] Ping nodes from discovery (#10167)
- [ ] snap: official image / test (#10168)
- [x] version: bump fork blocks for kovan and foundation, mark releases non critical (#10186)
- [x] pull constantinople on ethereum network (#10189)

### DIFF
--- a/util/version/Cargo.toml
+++ b/util/version/Cargo.toml
@@ -10,7 +10,7 @@ build = "build.rs"
 [package.metadata]
 # This versions track. Should be changed to `stable` or `beta` when on respective branches.
 # Used by auto-updater and for Parity version string.
-track = "nightly"
+track = "beta"
 
 # Network specific settings, used ONLY by auto-updater.
 # Latest supported fork blocks.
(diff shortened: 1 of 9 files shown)


util/version/Cargo.toml
```diff
@@ -10,15 +10,15 @@ build = "build.rs"
...
+track = "beta"
 
 # Network specific settings, used ONLY by auto-updater.
 # Latest supported fork blocks.
 # Indicates a critical release in this track (i.e. consensus issue).
 [package.metadata.networks]
-foundation = { forkBlock = 4370000, critical = false }
-ropsten = { forkBlock = 4230000, critical = false }
-kovan = { forkBlock = 6600000, critical = false }
+foundation = { forkBlock = 4370000, critical = true }
```
tomusdrw: `forkBlock` most likely needs an update

util/version/Cargo.toml
```diff
@@ -10,15 +10,15 @@ build = "build.rs"
...
+ropsten = { forkBlock = 4230000, critical = true }
```
tomusdrw: Why is it critical on testnets? Because of #10167 ?

5chdn: Yes, we might need a better definition of critical here

PR #10170, opened 2019-01-11
Remove reference to ui-interface command-line option

shoffmeister:
The ui-interface command-line option was apparently removed at some stage.

Remove mentions of it from command-line help output.


shoffmeister:
[clabot:check]

PR #10172, opened 2019-01-12
fix(parity-clib-example): remove websocket

niklasad1:
I suspect there is something wrong with the `WebSocket subscription` in the light client so let's remove it for now in order not to deny well-formed PRs!

/cc @tomusdrw @ngotchac

### DIFF
--- a/parity-clib/examples/cpp/main.cpp
+++ b/parity-clib/examples/cpp/main.cpp
@@ -14,20 +14,16 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
-#include <chrono>
+#include <cstring>
 #include <parity.h>
-#include <regex>
 #include <string>
-#include <thread>
+#include <vector>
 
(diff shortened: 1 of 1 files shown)


tomusdrw:
@niklasad1 you can subscribe to `eth_subscribe(newHeads)` @mattrutherford recently changed this so it's emitted even when syncing (see #9987)

niklasad1:
Thanks @tomusdrw 

I found out this yesterday too when I tested it locally too.

> I was wrong the websocket works without being fully synced tested locally now but it seems to some problem with ChainNotify in the light-client sometimes it works and sometimes not. I think I had some problem with it before but has worked locally for me for a while now

Thus, I need to debug if there is something weird with the `ChainNotify` for the light client (I have not confirmed it yet) and should probably be disabled for the CI to prevent other PRs to fail randomly.

tomaka:
I can also be the C++ example or C bindings that have an undefined behaviour, as opposed to the Rust code having a bug.

niklasad1:
Closing until the error-cause is confirmed which could be UB as @tomaka stated!

Issue #10130, opened 2019-01-03
Remove re-exports from ethcore


tomusdrw:
Currently we have a bunch of re-exports in `ethcore` crate. Some of them make sense (if we tie together the types together with a module), but are currently there only because the types were moved to `common-types` to avoid importing the entire crate.

Extract traits+types to separate crates or remove re-exports (whatever makes more sense in a particular case).

PR #10178, opened 2019-01-14
Extract CallContract and RegistryInfo traits into their own crate

HCastano:
This PR takes a stab at part of #10130.

As part of this PR `service-transaction-checker` (which depended on CallContract and RegistryInfo) got moved from `ethcore` to `ethcore-miner`.


ethcore/call-contract/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+#![warn(missing_docs)]
+
+//! Call Contract module
+//! Provides CallContract and RegistryInfo traits
```
tomusdrw: I would add a bit more explanation, maybe something like:

```suggestion
//! 
//! This crate exposes traits required to call contracts at particular block.
//! All utilities that depend on on-chain data should use those traits to access it.
```

ethcore/call-contract/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "call-contract"
```
tomusdrw: I wonder if it should be called `ethcore-call-contract` for consistency. While I don't like `ethcore-` prefix, I think it's good to prepend our internal crates with something. Any second opinion on this?

sorpaas: Second this!

5chdn:
```suggestion
name = "ethcore-call-contract"
```

PR #10180, opened 2019-01-14
Update for Android cross-compilation.

TriplEight:
Hey @tomaka , could you take a look at the changes in the script.

Unfortunately, there's nothing can be done with this monstrous Dockerfile: https://github.com/paritytech/scripts/blob/master/docker-files-for-Gitlab-CI-rust/cross/android/Dockerfile (size: 4111). I've also updated and pushed it to the docker hub.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -96,9 +96,9 @@ publish-docker:
     - scripts/gitlab/publish-docker.sh parity
 
 publish-snap:
-  stage:                           publish
-  only:                            *releaseable_branches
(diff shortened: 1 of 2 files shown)


niklasad1:
@TriplEight shouldn't we link with `arm-linux-androideabi-clang` according to the log `/usr/bin/ld` is used a linker in the for android cross-compilation?!

TriplEight:
@niklasad1 thanks for the tip, the linker you've mentioned was used, but incorrectly added in the Dockerfile.

PR #10182, opened 2019-01-14
chore(snapshot): improve message when taking snapshots & variable names

yxliang01:
When we are taking snapshots, we are taking the snapshot of the state at a particular block. The old message says that it "takes snapshot **starting** at block X" but it is actually a snapshot of state **at** block X.

Referring to https://wiki.parity.io/CLI-Sub-commands.html#snapshot

Please correct me if I understand the snapshot functionality wrongly.

Thanks


5chdn:
Please reopen once you signed the CLA

PR #10189, opened 2019-01-15
pull constantinople on ethereum network

5chdn:
- ethcore: pull constantinople on ethereum network
- version: revert fork block for ethereum
- version: mark update as critical on kovan (#10134) and ethereum networks (#10189)

### DIFF
--- a/util/version/Cargo.toml
+++ b/util/version/Cargo.toml
@@ -16,7 +16,7 @@ track = "nightly"
 # Latest supported fork blocks.
 # Indicates a critical release in this track (i.e. consensus issue).
 [package.metadata.networks]
-foundation = { forkBlock = 7080000, critical = false }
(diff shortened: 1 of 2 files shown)


util/version/Cargo.toml
```diff
@@ -16,9 +16,9 @@ track = "nightly"
 # Latest supported fork blocks.
 # Indicates a critical release in this track (i.e. consensus issue).
 [package.metadata.networks]
-foundation = { forkBlock = 7080000, critical = false }
+foundation = { forkBlock = 7080000, critical = true }
```
andresilva: Shouldn't we revert the `forkBlock` as well to `4370000`?

5chdn: Good catch!

PR #10196, opened 2019-01-16
fix(rpc-types/hash): make `Copy` to unify with ethereum-types

niklasad1:
The crates `ethereum-types` and `ethbloom` their types are `Copy` and `Rpc Hash/Uint` and based on those types so it is more convenient to have the same properties as those.


tomusdrw:
@niklasad1 aren't the types in `ethereum-types` serializable already? Maybe we can get rid of RPC wrappers completely?

niklasad1:
@tomusdrw Yeah, they do. I can check how painful it is to remove those RPC wrappers.

niklasad1:
Closed in favor of https://github.com/paritytech/parity-ethereum/pull/10217

PR #10198, opened 2019-01-16
Cancel Constantinople HF on POA Core

vbaranov:
Originally Constantinople HF was scheduled on **2019-01-18** for POA Core network

This PR changes spec for POA Core:
 - Removing of Constantinople EIPs transitions

[Related revert in POA Core spec.json](https://github.com/poanetwork/poa-chain-spec/pull/104)

PR #10199, opened 2019-01-16
Revert Constantinople activation in POA Core

phahulin:
Due to Constantinople HF being postponed in Eth mainnet,
POA Network also postpones HF in POA Core network that was planned on 18 Jan.

Reference PR: https://github.com/poanetwork/poa-chain-spec/pull/104/files


phahulin:
Same as #10198 sorry :)

PR #10209, opened 2019-01-17
fix(ManageNetwork): replace Range with RangeInclusive

niklasad1:
Fixes `TODO: Range should be changed to RangeInclusive once stable (https://github.com/rust-lang/rust/pull/50758)`

### DIFF
--- a/ethcore/sync/src/api.rs
+++ b/ethcore/sync/src/api.rs
@@ -17,7 +17,7 @@
 use std::sync::{Arc, mpsc, atomic};
 use std::collections::{HashMap, BTreeMap};
 use std::io;
-use std::ops::Range;
+use std::ops::RangeInclusive;
 use std::time::Duration;
 use bytes::Bytes;
 use devp2p::NetworkService;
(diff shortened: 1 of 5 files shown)


ethcore/sync/src/api.rs
```diff
@@ -615,9 +615,7 @@ pub trait ManageNetwork : Send + Sync {
 	/// Stop network
 	fn stop_network(&self);
 	/// Returns the minimum and maximum peers.
-	/// Note that `range.end` is *exclusive*.
-	// TODO: Range should be changed to RangeInclusive once stable (https://github.com/rust-lang/rust/pull/50758)
-	fn num_peers_range(&self) -> Range<u32>;
+	fn num_peers_range(&self) -> RangeInclusive<u32>;
 	/// Get network context for protocol.
```
niklasad1: IMHO, return type `RangeInclusive` makes it clear that the range is inclusive thus I removed the comment

ethcore/sync/src/api.rs
```diff
@@ -948,12 +946,12 @@ impl LightSyncProvider for LightSync {
...
 		let (connected, active) = self.proto.peer_count();
 		let peers_range = self.num_peers_range();
-		debug_assert!(peers_range.end > peers_range.start);
+		debug_assert!(peers_range.end() > peers_range.start());
 		PeerNumbers {
 			connected: connected,
 			active: active,
-			max: peers_range.end as usize - 1,
-			min: peers_range.start as usize,
+			max: *peers_range.end() as usize - 1,
```
niklasad1: Not addressed by this PR: why do we subtract `end` by 1?

ethcore/sync/src/api.rs
```diff
@@ -948,12 +946,12 @@ impl LightSyncProvider for LightSync {
 	fn peer_numbers(&self) -> PeerNumbers {
 		let (connected, active) = self.proto.peer_count();
 		let peers_range = self.num_peers_range();
-		debug_assert!(peers_range.end > peers_range.start);
+		debug_assert!(peers_range.end() > peers_range.start());
```
ordian: Should it be `>=` here and below?

niklasad1: Yeah, I think so otherwise this fail if `start == end` because end was exclusive before

PR #10213, opened 2019-01-18
Deprecate account management

tomusdrw:
- Move account management to a single crate (`ethcore-accounts`)
- Avoid depenending on accounts directly (only main binary, `parity-rpc` and `secret-store` (one struct) depends on it currently)
- Rely on raw secrets wherever possible (`engine-signer` and `secret-store`)
- Add compilation feature to allow easier removal in the future (diff in binary size is 5MB (61 vs 56 on my machine))
- Feature-gate RPCs that deal with accounts, add deprecation notice to others.

Related: #9997 

Future work:
1. Consider extracting accounts to a jsonrpc-proxy
2. Integrate with CLEF (https://github.com/ethereum/go-ethereum/tree/master/cmd/clef)


accounts/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+common-types = { path = "../ethcore/types" }
+ethkey = { path = "ethkey" }
+ethstore = { path = "ethstore" }
+log = "0.4"
+parking_lot = "0.7"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+
+[target.'cfg(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "android"))'.dependencies]
```
ordian: we removed hw support for android in #10161

niklasad1:
hrm, I don't understand why the android build fails

accounts/src/lib.rs
```diff
@@ -14,94 +14,62 @@
...
-			SignError::Hardware(ref e) => write!(f, "{}", e),
-			SignError::SStore(ref e) => write!(f, "{}", e),
+impl Default for AccountProviderSettings {
+	fn default() -> Self {
+		AccountProviderSettings {
+			enable_hardware_wallets: false,
+			hardware_wallet_classic_key: false,
+			unlock_keep_secret: false,
+			blacklisted_accounts: vec![],
 		}
```
niklasad1: `derive Default` instead for `AccountProviderSettings`

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1492,12 +1491,16 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
-		self.signer.write().set(ap, address, password);
+	fn set_signer(&self, signer: Box<EngineSigner>) {
+		*self.signer.write() = Some(signer);
 	}
 
 	fn sign(&self, hash: H256) -> Result<Signature, Error> {
-		Ok(self.signer.read().sign(hash)?)
+		Ok(self.signer.read()
+			.as_ref()
+			.ok_or_else(|| ethkey::Error::InvalidAddress)?
```
niklasad1: This could be `ok_or` because InvalidAddress is just a value (enum pattern)

parity/account.rs
```diff
@@ -61,83 +55,102 @@ pub struct ImportFromGethAccounts {
...
+
+#[cfg(not(feature = "accounts"))]
 pub fn execute(cmd: AccountCmd) -> Result<String, String> {
-	match cmd {
-		AccountCmd::New(new_cmd) => new(new_cmd),
-		AccountCmd::List(list_cmd) => list(list_cmd),
-		AccountCmd::Import(import_cmd) => import(import_cmd),
-		AccountCmd::ImportFromGeth(import_geth_cmd) => import_geth(import_geth_cmd)
-	}
+	Err("Account management is deprecated. Please see #9997 for alternatives.".into())
```
niklasad1: Consider to add link to the issue

Cargo.toml
```diff
@@ -87,6 +88,8 @@ winapi = { version = "0.3.4", features = ["winsock2", "winuser", "shellapi"] }
 daemonize = "0.3"
 
 [features]
+default = ["accounts"]
+accounts = ["ethcore-accounts", "ethcore-secretstore/accounts", "parity-rpc/accounts"]
```
svyatonik: This causes Parity to be built with "secretstore" feature enabled by default (since default features are "accounts" and it implies "ethcore-secretstore/accounts"). I guess we need something like `accounts AND secretstore = ["ethcore-secretstore/accounts"]`, but I'm not sure it could be defined within Cargo.toml. Probably replace `accounts => [..., "ethcore-secretstore/accounts", ...]` with `secretstore => [""ethcore-secretstore/accounts", "ethcore-accounts"]`?

rpc/src/v1/impls/secretstore.rs
```diff
@@ -21,7 +21,7 @@ use std::sync::Arc;
 
 use crypto::DEFAULT_MAC;
 use ethkey::Secret;
-use ethcore::account_provider::AccountProvider;
+use accounts::AccountProvider;
```
svyatonik: Just curios - are we going to leave account management under `accounts` feature forever (i.e. for some long period), or it'll be totally purged from the codebase in the nearest future? In the latter case, we'll need to replace all `address: H160, password: Password` in SS RPCs with something like `secret: H256` or do something else.

Cargo.toml
svyatonik: or even better - just remove "ethcore-secretstore/accounts" from "accounts" - so that there could be `--features ""`, `--features "accounts"`, `--features "secretstore"` and `--features "secretstore accounts"`.

tomusdrw: I don't think that would work, we need to enable `ethcore-secretstore/accounts` if both `secretstore` and `accounts` features are enabled. I changed it so that we require `accounts` in case we compile with `secretstore`

rpc/src/v1/impls/secretstore.rs
tomusdrw: It's not defined yet, perhaps we are going to move some functionality to the proxy, but if we decide to remove accounts completely we need to provide some non-account-based alternatives like using secret directly. Don't know the definitive answer yet though.

parity/account.rs
tomusdrw: Added

Cargo.lock
```diff
@@ -1970,7 +1987,7 @@ dependencies = [
 [[package]]
 name = "libusb-sys"
 version = "0.2.4"
-source = "git+https://github.com/paritytech/libusb-sys#1d33d9840a82adaf4d6a1a0f5141f022e5676802"
+source = "git+https://github.com/paritytech/libusb-sys#4418e92c82509d1f049f7942dffe62a0bf5dd871"
 dependencies = [
```
niklasad1: hrm, is this change intentional?

tomusdrw: Nope, let me restore `Cargo.lock` from master.

rpc/src/v1/helpers/dispatch/mod.rs
```diff
@@ -0,0 +1,381 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Utilities and helpers for transaction dispatch.
+
+pub(crate) mod light;
+mod full;
+mod prospective_signer;
+
```
niklasad1: This is a good change i.e., to break-up `dispatcher` into different mods however I haven't reviewed that code. I assume it is just moved?

tomusdrw: Yeah, mostly moved, but I had to do some small changes to support the new `Accounts` trait instead of using `AccountProvider` directly.

niklasad1: all right, we should probably run some manual tests on the `light-client` before merging because rpc integration tests are still missing

5chdn:
could you sign this off @ordian @svyatonik ?

5chdn:
Does not compile @tomusdrw

PR #10217, opened 2019-01-18
fix(rpc-types): replace uint and hash with `ethereum_types v0.4`

niklasad1:
This PR removes the RPC wrapper types `uint` and `hash` because `ethereum-types v0.4` implements serde. Also nice to remove all this conversions between `ethereum-types` to `RpcTypes` (I have not removed all those in this PR I will clean it up with another PR)

Most changes are imports with a few exceptions also I decided not to `re-export` `ethereum-types` instead each crate that is using them has to use `ethereum-types` instead!


rpc/src/v1/helpers/subscribers.rs
```diff
@@ -37,7 +39,7 @@ impl str::FromStr for Id {
 }
 impl Id {
 	pub fn as_string(&self) -> String {
-		format!("0x{:?}", self.0)
+		format!("{:?}", self.0)
```
niklasad1: This should be changed to `serde_json::to_string_pretty(&self.0)` for performance reasons but I think it is out-of-scope of this PR and should be fixed in another one

rpc/src/v1/traits/eth.rs
```diff
@@ -31,7 +31,7 @@ build_rpc_trait! {
 		#[rpc(name = "eth_protocolVersion")]
 		fn protocol_version(&self) -> Result<String>;
 
-		/// Returns an object with data about the sync status or false. (wtf?)
+		/// Returns an object with data about the sync status or false.
```
niklasad1: out-of-scope IMO but I couldn't resist. 

we should not such comments in the code base :small_airplane:

rpc/src/v1/helpers/dispatch.rs
```diff
@@ -677,9 +677,7 @@ pub fn execute<D: Dispatcher + 'static>(
 			if accounts.is_hardware_address(&address) {
 				let signature = accounts.sign_message_with_hardware(&address, &data)
 					.map(|s| H520(s.into_electrum()))
-					.map(RpcH520::from)
 					.map(ConfirmationResponse::Signature)
-					// TODO: is this correct? I guess the `token` is the wallet in this context
```
niklasad1: Out-of-scope from this PR but this comment was introduced by me for a long time ago and should be removed!

rpc/src/v1/traits/parity.rs
```diff
@@ -240,7 +241,7 @@ build_rpc_trait! {
 		/// Extracts Address and public key from signature using the r, s and v params. Equivalent to Solidity erecover
 		/// as well as checks the signature for chain replay protection
 		#[rpc(name = "parity_verifySignature")]
-		fn verify_signature(&self, bool, Bytes, H256, H256, U64) -> Result<RecoveredAccount>;
+		fn verify_signature(&self, bool, Bytes, H256, H256, u64) -> Result<RecoveredAccount>;
```
niklasad1: I changed the type to `u64` because `ethereum_types::U64` can't be converted into a `u64` or some other type to perform arithmetics on

tomusdrw: This is going to change the signature of the method - we should be careful here. I would prefer to have all numbers (or at least the ones that can't be express in JS with enough precision) encoded as hex

tomusdrw: Wouldn't `low_u64()` or `as_u64()` work?

niklasad1: Cool, both work just looked at the `From/Into` impls will change it back

seunlanlege:
tons on conflicts. please rebase, would like to get this merged as soon as possible.

niklasad1:
> Most changes are imports with a few exceptions also I decided not to `re-export` `ethereum-types` instead each crate that is using them has to use `ethereum-types` instead!

> not sure what this means

I meant before we used the wrapper types and exported them from the `rpc crate` as  for example `use rpc::signer::U256`. If you look at `cli-signer crate` it must explicitly import  `ethereum-types` instead. Clear enough?

niklasad1:
This is PR is blocked because `ethereum-types v0.4` contains the `de-serialize bug` that we recently fixed, see https://github.com/paritytech/primitives/blob/v0.4.0/serialize/src/lib.rs#L89

Cargo.lock
```diff
@@ -4517,7 +4520,7 @@ dependencies = [
 "checksum ethabi-contract 6.0.0 (registry+https://github.com/rust-lang/crates.io-index)" = "795e25fd868e12a59ca235dbe1f6cc8f1eba8f67d6a39438b29535e0126e0c27"
 "checksum ethabi-derive 6.0.2 (registry+https://github.com/rust-lang/crates.io-index)" = "66a587250c8190be9d6ae28d67b8957ed97cb9eee2e272173a20593ab054a075"
 "checksum ethbloom 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)" = "1a93a43ce2e9f09071449da36bfa7a1b20b950ee344b6904ff23de493b03b386"
-"checksum ethereum-types 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)" = "35b3c5a18bc5e73a32a110ac743ec04b02bbbcd3b71d3118d40a6113d509378a"
+"checksum ethereum-types 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)" = "5ef1b8b23cb62440bad25bfb5d0f5ba97f18ebabd9b777ef0ef5d37c72cca3c0"
 "checksum ethereum-types-serialize 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)" = "4ac59a21a9ce98e188f3dace9eb67a6c4a3c67ec7fbc7218cb827852679dc002"
```
niklasad1: This is a bummer need to update `ethereum-types-serialize` to 0.2.2 it is under progress.

niklasad1: Bad `cargo publish`

niklasad1:
@tomusdrw can you review this again? I think it is ready now,  I bumped `ethereum-types` and added some tests in this codebase even if `ethereum-types` have those more or less !

soc1c:
Needs a 2nd review :+1: 

cc @tomusdrw

rpc/src/v1/helpers/subscribers.rs
```diff
@@ -36,8 +36,9 @@ impl str::FromStr for Id {
 	}
 }
 impl Id {
+	// FIXME: [niklasad1] replace `format` with `serde_json::to_string_pretty`
```
tomusdrw: Should make this an issue?

rpc/src/v1/types/uint.rs
```diff
@@ -1,172 +0,0 @@
...
-}
-
-impl serde::Serialize for U64 {
-	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: serde::Serializer {
-		serializer.serialize_str(&format!("{:#x}", self))
-	}
-}
-
-#[cfg(test)]
-mod tests {
```
tomusdrw: Can we leave the tests to make sure that our assumptions are not broken when the library is updated? Also does 0.4 contain the fix for non-ascii characters?

niklasad1: Moved them to a [new_file](https://github.com/paritytech/parity-ethereum/blob/f319a687aa81d8a7c5f7224726d70778756a5cf0/rpc/src/v1/types/eth_types.rs) ok?

rpc/src/v1/helpers/subscribers.rs
niklasad1: Yes, when I benched it `serde::deser` was 10 times faster than `format!`

rpc/src/v1/types/uint.rs
tomusdrw: yup :ok_hand:

Issue #6616, opened 2017-10-02
Parity rejects MetaMask's `x-metamask-origin` header, and consequently all MetaMask/Firefox requests, during CORS preflight


interfect:
I'm running:

- **Parity version**: 1.6.10
- **Operating system**: Linux
- **And installed**: via `parity/parity:stable` Docker image

When using MetaMask 3.10.8 on Firefox 55 on Windows 7, pointed at my Parity node running on another Ubuntu machine, I observe an HTTP exchange that goeth thusly:

```
OPTIONS / HTTP/1.1
Host: 10.1.0.8:8545
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:55.0) Gecko/20100101 Firefox/55.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Access-Control-Request-Method: POST
Access-Control-Request-Headers: content-type,x-metamask-origin
Origin: moz-extension://c2480724-8a32-4afc-a7aa-c1d18a8bf229
Connection: keep-alive

HTTP/1.1 200 OK
Content-Type: application/json
Allow: OPTIONS, POST
Access-Control-Allow-Headers: origin, content-type, accept
Access-Control-Allow-Origin: moz-extension://c2480724-8a32-4afc-a7aa-c1d18a8bf229
Date: Mon, 02 Oct 2017 00:58:04 GMT
Transfer-Encoding: chunked
```

This is a CORS preflight for an actual request, but the actual request is never sent.

MetaMask wants to send an `x-metamask-origin` header with its request, giving the actual page (as opposed to `moz-extension://c2480724-8a32-4afc-a7aa-c1d18a8bf229`) that it is proxying requests for. Presumably this is so that you could potentially configure your node to handle requests only for certain duly authorized actual page origins, even when using MetaMask.

Parity doesn't seem to know about this header, and in particular is not returning it in the `Access-Control-Allow-Headers`. According to the CORS spec, that means that the browser must not permit a request to be made with that header on it. MetaMask (on Firefox, at least) insists on trying to send the header, so Firefox blocks its request altogether. The net result of all this is that MetaMask on Firefox does not work with a Parity node as the backend.

[...]

PR #10221, opened 2019-01-21
Echo CORS request headers by default

cmichi:
This PR addresses #6616:

> For maximum compatibility, Parity should echo back all the headers from Access-Control-Request-Headers in Access-Control-Allow-Headers. 

I just noticed that this is already the default behavior of the jsonrpc `with_meta_extractor(…)`: https://github.com/paritytech/jsonrpc/blob/parity-2.2/http/src/lib.rs#L240. I still think it's a good idea to merge this, since it makes the behavior more explicit and adds a test which would fail if the default behavior would ever change.

### DIFF
--- a/rpc/src/lib.rs
+++ b/rpc/src/lib.rs
@@ -151,9 +151,7 @@ pub fn start_http<M, S, H, T>(
 		.cors(cors_domains.into())
 		.allowed_hosts(allowed_hosts.into())
 		.health_api(("/api/health", "parity_nodeStatus"))
-		.cors_allow_headers(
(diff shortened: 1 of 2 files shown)


rpc/src/lib.rs
```diff
@@ -151,6 +151,9 @@ pub fn start_http<M, S, H, T>(
 		.cors(cors_domains.into())
 		.allowed_hosts(allowed_hosts.into())
 		.health_api(("/api/health", "parity_nodeStatus"))
+		.cors_allow_headers(
```
tomusdrw: Would keep both single line

PR #10222, opened 2019-01-21
Extract BlockInfo, ChainInfo, and TransactionInfo traits

HCastano:
This PR moves the `BlockInfo`, `ChainInfo`, and `TransactionInfo` traits into their own crate called `ethcore-blockchain-info`.

Related to: #10130


HCastano:
I just took a look at `BlockProvider`, and it does look like it has a lot of the same functionality as `BlockInfo`. It doesn't look like `BlockProvider` has a method like `TransactionInfo::transaction_block`, so should that just become part of the `BlockProvider` trait?

Are you suggesting something like this:

```rs
// In ethcore/src/client/traits.rs
pub trait BlockChainClient: ... {
    ...
    fn blockchain() -> &BlockChain { ... }
    ...
}
```
And then removing all the methods we can access through the `BlockChain` struct from `BlockChainClient`?

HCastano:
I'm going to close this in favour of doing something like in #10251 - in which the methods of this new `ethcore-blockchain-info` crate would go unused (intentionally).

If it that alternative approach doesn't work, this can be re-opened.

PR #10223, opened 2019-01-21
Enable St-Peters-Fork ("Constantinople Fix")

5chdn:
- disable eip-1283 on kovan block 10255201
- disable eip-1283 on ropsten block 4939394
- enable st-peters-fork on mainnet block 7280000
- disable eip-1283 on sokol block 7026400

---

- ref #10214 
- ref https://github.com/ethereum/go-ethereum/pull/18486

### DIFF
--- a/ethcore/res/ethereum/kovan.json
+++ b/ethcore/res/ethereum/kovan.json
@@ -49,8 +49,8 @@
 		"eip1014Transition": "0x8c6180",
 		"eip1052Transition": "0x8c6180",
 		"eip1283Transition": "0x8c6180",
-		"eip1283DisableTransition": "0x8c6180",
-		"kip4Transition": "0x9c7b61",
+		"eip1283DisableTransition": "0x9c7b61",
+		"kip4Transition": "0x8c6180",
 		"kip6Transition": "0x8c6180"
 	},
 	"genesis": {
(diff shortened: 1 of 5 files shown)


5chdn:
paging @phahulin @varasev @vbaranov - which block number for Sokol and xDai?

ethcore/res/ethereum/kovan.json
```diff
@@ -49,7 +49,8 @@
 		"eip1014Transition": "0x8c6180",
 		"eip1052Transition": "0x8c6180",
 		"eip1283Transition": "0x8c6180",
-		"kip4Transition": "0x8c6180",
+		"eip1283DisableTransition": "0x8c6180",
```
tomusdrw: I suppsoe we want to keep `kip4` at the same block and just schedule disabling of `eip1283`? The numbers should be swapped.

5chdn: Good someone is actually looking at the specs :+1:

varasev:
> paging @phahulin @varasev @vbaranov - which block number for Sokol and xDai?

For Sokol, the block number is `7026400` (6 February).

For xDai we don't have spec json in `ethcore/res/ethereum`, so we'll define the block number for xDai and update our [spec.json](https://github.com/poanetwork/poa-chain-spec/blob/dai/spec.json) a bit later.

5chdn:
ref https://github.com/ethereum/EIPs/pull/1716

PR #10234, opened 2019-01-22
Macos heapsize force jemalloc

cheme:
This PR is a temporary fix for heapsize issue on mac.

It uses a fork of heapsize that uses jemallocator crate. The fork is currently on my repo, maybe creating paritytech/heapsize could be an idea, but it conflicts with target that is having 
https://github.com/paritytech/parity-common/pull/93 merged and remove heapsize from parity-ethereum (needs an other PR merged and crates published so this commit is probably safer regarding current deadlines).

The build pass https://gitlab.parity.io/parity/parity-ethereum/-/jobs/118084 on a modified ci config (I restore the original config in latest commit).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,18 +35,6 @@ variables:
   - export VERSION
   - echo "Version = ${VERSION}"
 
-test-macos:
-  stage:                           test
-  variables:
-    RUN_TESTS:                     all
-    CARGO_TARGET:                  x86_64-apple-darwin
-    CC:                            gcc
-    CXX:                           g++
-  script:
-    - scripts/gitlab/test-all.sh stable
-  tags:
-    - rust-osx
-
 test-linux:
   stage:                           test
   variables:


ordian:
run `cargo update -p heapsize` to update `Cargo.lock` (it uses old heapsize commit 4dcb303d)

cheme:
@ordian thanks I just realize it, I was running test on a deprecated hash.

5chdn:
do you need a macos pipeline here?

cheme:
yes, I temporarilly put a 'test-build-darwin' as single test target, strange it doesn't launch on the last commit.

cheme:
@ordian it ends up that the problem was indeed with 
```
[target.'cfg(macos)'.dependencies]
```
it needed
```
[target.'cfg(target_os = "macos")'.dependencies]
```
and in code it also needed the target_os check.

rpc/src/v1/impls/personal.rs
```diff
@@ -21,6 +21,7 @@ use std::time::Duration;
 use bytes::Bytes;
 use ethcore::account_provider::AccountProvider;
 use types::transaction::PendingTransaction;
+use types::transaction::SignedTransaction;
```
ordian: Could you merge with master to eliminate this unrelated change?

cheme:
Yes it is temporary, the ongoing issue in 'parity-common' is https://github.com/paritytech/parity-common/pull/93 it is mainly waiting for review, the actual fix will be https://github.com/paritytech/parity-ethereum/compare/master...cheme:rem_heapsize but it requires the crates update from parity-common pr (it has been a long time since I did not update the branch so it is a bit out of sync).

ordian:
If we're going to release binaries with this fix, I think we should force jemalloc on every supported platform ([not only on macOS](https://github.com/servo/heapsize/compare/master...cheme:ec-macfix#diff-b4aea3e418ccdb71239b96952d9cddb6R25)). Alternatively, we could use rust 1.31, as suggested by @HCastano.

cheme:
The code will indeed work also for linux. I activated it only for macos to be cautious (and tbh also to make sure we do not forget it is temporary).

cheme:
One other thing to consider with also including linux in the fix is that it will avoid the performance drop related to the allocator switch.
If there is no voice against it, I will add it this afternoon.

cheme:
I am facing an issue (probably some cornercase between proc_macro, patching a crate and changing global alloc) when using the same method on linux, so I won't apply to linux too (using malloc_size_of it still behave properly).

.gitlab-ci.yml
```diff
@@ -35,6 +35,18 @@ variables:
   - export VERSION
   - echo "Version = ${VERSION}"
 
+test-macos:
```
5chdn: do we want to keep this?

cheme: Probably not if we merge the PR (will really slowdown CI of every PR), but I activated it last week because in some case with heapsize failures occurs after build and I wanted to assert it was not the case.
If we only use the PR to patch for the build (not sure if it is possible or even a good idea) we can keep it.

cheme: I switch it off, for reference https://gitlab.parity.io/parity/parity-ethereum/-/jobs/118672

cheme:
@tomaka it actually run the same way as the parity-util-mem does with 'jemalloc-global' feature so any issue with this will be an issue in parity-common/parity-util-mem (except the linux issue from previous comments that I think is related to the use of cargo patch somehow).

ordian:
Note, that it only enables jemalloc on macOS and not on linux (which may incur some perf regression). That was my only concern for approving this PR. But since this is a temporary fix, it's fine I guess?

PR #10238, opened 2019-01-23
fix(light-rpc): Make `light_sync` generic

niklasad1:
The motivation behind this change is to easily mock `light-sync` to make it possible to enable `rpc-integration` tests for the light-client.

Currently the `rpc's` requires the concrete type `sync::LightSync` which makes it very hard to do so


ascjones:
^^^ Conflicts

PR #10241, opened 2019-01-24
SecretStore: expose restore_key_public in HTTP API

### LINKED ISSUE #10240: [Secret Store] support for retrieving public portion of already generated server key
**Feature request**: it would be useful to be able to retrieve the public portion of an already generated server key through the HTTP API.

**Problem that would be solved**
After running a successful `server key generation session`, this public portion is returned. If the key already exists with that ID though, the following error is thrown (which is fine) but there is no way to query for this piece of information again:
```
Request failed
StatusCode: 400
StatusMessage: Bad Request
Body: "\"Server key with this ID is already generated\""
Request options: {"url":"http://52.74.41.192:8082/shadow/45ce99addb0f8385bd24f30da619ddcc0cadadab73e2a4ffb7801083086b3fc2/61dab3b08229e6aaaefb16eefc746217bf80f8df38d88e12e4e9f3b5158f0aaf671ab618614e01ee429784533b0e012b6c8c4abeee90db7b58c24e67508cbd2400/1","method":"POST"}
```

svyatonik:
closes #10240 

This is already exposed in on-chain API. Server key' public is returned unencrypted (as in `GenerateServerKey` call).

Upon receiving request, key version negotiation session starts (this looks like an redundant overhead, when key is available on the node, but it solves some problems internally). The session ends up in version + common part of key data agreement.

The key is returned only if requested by its author. This is controversial, because it differs from the on-chain API (anyone can request public portion of server key). But in on-chain API server key' public is already available to anyone who's watching contract. In HTTP API it is only available to the author.

cc @ngyam


dvdplm:
@svyatonik need rebase

vieiralc:
How can I retrieve the public portion of the server key?

svyatonik:
@vieiralc Something like that:
```bash
curl -X GET http://localhost:8082/server/0000000000000000000000000000000000000000000000000000000000005002/24fb44509d9b7770915568621fe64bd6dd7035c6e15878a795e324489ba79daa43fb2c49da5ed46eeca2357d3acb5ddf61d6afc1c90ca3f2a5791f72ab093cda01
```
where `0000000000000000000000000000000000000000000000000000000000005002` is the server key id and `24fb44509d9b7770915568621fe64bd6dd7035c6e15878a795e324489ba79daa43fb2c49da5ed46eeca2357d3acb5ddf61d6afc1c90ca3f2a5791f72ab093cda01` is the key author' signature server key id

PR #10243, opened 2019-01-24
add cli option to write contract storage diffs

rmulhol:
This PR creates a CLI option to enable writing contract storage diffs to a secondary datastore. This initial implementation enables writing to a CSV file, but is structured such that another filetype/datastore could be used by creating another implementation of the `StorageWriter` trait.

Thinking this could help folks who want to audit contract storage while not running an archive node. You can optionally configure which contracts to watch, such that you could maintain a full history of storage changes for contracts that interest you while running only a full node.

Path to CSV is also configurable.


rmulhol:
Please let me know if there are any blockers to having this reviewed. I noticed a CI failure on `gitlab-test-beta` but I think that's just because the `can push work` test is a bit flaky - would expect it to pass after a retry or two (but I don't have permissions to kick that off).

ethcore/src/block.rs
```diff
@@ -107,13 +108,17 @@ pub struct ExecutedBlock {
 	pub state: State<StateDB>,
 	/// Transaction traces.
 	pub traces: Tracing,
+	/// Writer for persisting storage changes.
+	pub storage_writer: Box<storage_writer::StorageWriter>,
```
tomusdrw: We should reuse `Tracing` for this instead of introducing additional entries.

ethcore/src/block.rs
```diff
@@ -107,13 +108,17 @@ pub struct ExecutedBlock {
...
+	/// Block hash.
+	pub header_hash: H256,
```
tomusdrw: It's already part of `self.header.hash`

ethcore/src/block.rs
```diff
@@ -371,6 +380,9 @@ impl<'x> OpenBlock<'x> {
 		let mut s = self;
 
 		s.engine.on_close_block(&mut s.block)?;
+		if s.block.storage_writer.enabled() {
+			s.block.state.write_watched_state(s.block.header_hash, s.block.header.number(), s.block.storage_writer.clone())?;
```
tomusdrw: That's a performance killer.

rmulhol:
@tomusdrw thanks for taking a look! The comments about improving performance and reducing complexity are very helpful.

It sounds like this sort of feature - writing raw data to a secondary datastore during the sync - may not be something you'd like to include in this codebase, in which case I may just try and incorporate your feedback into a separate fork and close this PR. Please let me know if I'm misunderstanding you there.

I like your suggested approach. But, as you noted, consuming over RPC is pretty cumbersome - it's very nice to have the node pushing diffs during the sync, even if I need to separately handle reorgs. If I went ahead with developing a separate process for exporting parts of the Parity database to CSV or another format alongside a sync, is that something you could envision being merged into this project?

ethcore/src/block.rs
rmulhol: The hash in `self.header.hash` does not match the actual header's hash, since the header being constructed here is [populated from](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/block.rs#L343) a subset of fields on the header. That works for verification since `self.header` only needs to match a [subset of fields](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/verification/verification.rs#L250) on the imported header. But it does not work for the CSV if we want to have the actual header hash.

ethcore/src/block.rs
rmulhol: Totally hear you there, but you only take the hit if the writer is enabled, right? Thinking that it might make sense to have an opt in approach where you take hit on performance in your sync to get significantly improved performance on accessing the data.

ethcore/src/block.rs
rmulhol: Cool, point taken. I had originally put this adjacent to `Tracing` because it felt similar but different - i.e. if a block can include knowledge of tracing, maybe it can also include knowledge of storage writing, but they shouldn't be combined if they're conceptually distinct. Happy to take a look at combining them if that seems like a better way to avoid introducing complexity.

rmulhol:
Just to be clear - I'm not at all wedded to CSV as the persistence layer. I could see a case for streaming the data via websockets or writing to any number of other storage options. The core feature I'm hoping to get a feel for is basically just publishing diffs during a sync so that they can be consumed quickly and in real time.

tomusdrw:
Hi @rmulhol, as I said earlier I'm not against that feature at all and like it.

I have issues with the storage format though - would prefer to have it inserted into our existing RocksDB instances or some other low-level, binary, key-value db. Unfortunatelly rocksdb can't be accessed by two processes at the same time so it's not possible to read from it while Parity is running, which probably doesn't work for your use case.

If you need this data pushed to you, Pub-Sub RPC seems like a perfect fit here - all diff changes can be pushed to subscribed clients, which is something I'd be willing to include as well (if implemented correctly, i.e. pushed asynchronously so that block import doesn't have to wait for IO). Pub-Sub could be implemented either with or without additional DB storage and RPC method to access past data if db persistence is enabled.

rmulhol:
Thanks for the feedback! Closing this PR since the approach will need to be significantly reworked before it can be merged.

PR #10246, opened 2019-01-25
Secure WS-RPC: grant access to all apis

axelchalon:
We're updating Fether/light.js to use personal_signTransaction rather than signer_*. However, currently on Parity Ethereum, ws connections with a secure token have the ApiSet::SafeContext, which is every api except personal. This PR makes ws connections with a secure token have access to all apis. This way we can launch Parity Ethereum with the default apis enabled and have personal_ only be exposed in the ws connection with secure token.


grbIzl:
Looks good

PR #10247, opened 2019-01-25
Increase number of requested block bodies in chain sync

elferdo:
In order to speed up syncing and to be on par with geth, we want to increase the number of block bodies requested from a peer. It happens, though, that older parity software would not send any reply at all when a packet went over a given size. To play it safe, if syncing with such a peer we want to keep the lower numbers we have today.

In addition to increasing the number of block bodies per request, I have imlemented a struct, ClientVersion, and associated functionality to help parsing peer Id strings and querying for specific capabilities of the peer software.

Preliminary tests syncing with only geth nodes show an increase in synced blocks for a given fixed length of time of around 30% (with variations between runs).

Thanks to @ngotchac for support and reviewing.


util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+			ClientVersion::ParityUnknownFormat(_) => false,
+			ClientVersion::Other(_) => false
+		}
+	}
+
+}
+
+
+fn is_parity(client_id: &str) -> bool {
+	if client_id.starts_with(LEGACY_CLIENT_ID_PREFIX) || client_id.starts_with(PARITY_CLIENT_ID_PREFIX) {
```
tomusdrw: do we really need `if`? and `returns`?

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+// This is really not robust: parse four arguments and
+// allow for an extra argument between identifier and
+// version
+// TODO implement a better logic
+fn parse_parity_format(client_version: &str) -> Result<ParityClientData, ()> {
+	let tokens: Vec<&str> = client_version.split("/").collect();
+
+
+	// Basically strip leading 'v'
+	if let Some(version_number) = &get_number_from_version(tokens[1]) {
```
tomusdrw: `tokens[1]` may panic in case there is not enough tokens, let's move the len check earlier.

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+		if tokens.len() < 4 {
+			return Err(());
+		}
+
+		return Ok(
+			ParityClientData {
+				name: tokens[0].to_string(),
+				variant: None,
+				semver: Version::parse(version_number).unwrap(),
```
tomusdrw: Please either use `expect()` or turn into error (I think the latter should be done).

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+
+		if let Ok(data) = parse_parity_format(client_version) {
+			ClientVersion::ParityClient(data)
+		} else {
+			ClientVersion::ParityUnknownFormat(client_version.to_string())
+		}
+	}
+}
+
+impl From<String> for ClientVersion {
```
tomusdrw: Might be unnecessary if we implement for `T: AsRef<str>` I think

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+	fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
+		match self {
+			ClientVersion::ParityClient(data) => format_parity_version_string(data, f),
+			ClientVersion::ParityUnknownFormat(id) => write!(f, "{}", id),
+			ClientVersion::Other(id) => write!(f, "{}", id)
+		}
+	}
+}
+
+fn get_number_from_version(version: &str) -> Option<String> {
```
tomusdrw: Do we really need to allocate new `String` here? Maybe `&str` would do?

ethcore/sync/src/lib.rs
```diff
@@ -34,6 +34,7 @@ extern crate parity_bytes as bytes;
 extern crate parking_lot;
 extern crate rand;
 extern crate rlp;
+extern crate semver;
```
tomusdrw: Seems unused in `sync`

ethcore/sync/src/sync_io.rs
```diff
@@ -40,9 +41,9 @@ pub trait SyncIo {
 	fn chain(&self) -> &BlockChainClient;
 	/// Get the snapshot service.
 	fn snapshot_service(&self) -> &SnapshotService;
-	/// Returns peer identifier string
-	fn peer_info(&self, peer_id: PeerId) -> String {
-		peer_id.to_string()
+	/// Returns peer version identifier
+	fn peer_info(&self, peer_id: PeerId) -> ClientVersion {
```
tomusdrw: let's rename it to `peer_version` since it doesn't really return the Id any more

rpc/src/v1/types/sync.rs
```diff
@@ -157,7 +157,7 @@ impl From<SyncPeerInfo> for PeerInfo {
 	fn from(p: SyncPeerInfo) -> Self {
 		PeerInfo {
 			id: p.id,
-			name: p.client_version,
+			name: p.client_version.to_string(), // TODO reimplement with ClientVersion directly
```
tomusdrw: Either address before merge or create an issue and add a number here

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
```
tomusdrw: Please add `#[!warn(missing_docs)]` and add docs for public stuff

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+	}
+}
+
+// Parse known parity formats.
+//
+// This is really not robust: parse four arguments and
+// allow for an extra argument between identifier and
+// version
+// TODO implement a better logic
+fn parse_parity_format(client_version: &str) -> Result<ParityClientData, ()> {
```
tomusdrw: The logic seems a bit tricky. if `variant` starts with `v` we are going to treat it as version and panic.

Why not change it so that we check the lengths first and then based on the fact that it has either 5 or 4 elements do different parsing of components?

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+// Parity client string prefix
+const LEGACY_CLIENT_ID_PREFIX: &str = "Parity";
+const PARITY_CLIENT_ID_PREFIX: &str = "Parity-Ethereum";
+
+// Parity versions starting from this will accept block bodies requests
+// of 256 bodies
+const PARITY_CLIENT_LARGE_REQUESTS_VERSION: &str = "2.3.0";
+
+// Parity versions starting from this will accept service-transactions
+const SERVICE_TRANSACTIONS_VERSION: &str = "1.6.0";
```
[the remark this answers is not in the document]
elferdo: That means getting rid of the `accepts_service_transaction` check as well, and in `SyncPropagator::propagate_new_transactions` simply pass a tautological filter to `select_peers_for_transactions`. That would leave no use for the filter parameter and it could simply be removed, right?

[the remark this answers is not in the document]
tomusdrw: > That means getting rid of the accepts_service_transaction check as well,

Yes (although see below)

> in SyncPropagator::propagate_new_transactions simply pass a tautological filter to select_peers_for_transactions

We should still send only to Parity Ethereum peers, so might be easier to have `accepts_service_transaction` return `true` only for Parity Ethereum (no matter what version), but not Geth/other impls

util/network/src/client_version.rs
elferdo: Right, no excuses here

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+/// This structure as it is represents the format used by Parity clients. Other
+/// vendors may provide additional fields.
+///
+/// TODO support formats with extra fields, e.g.:
+/// "Geth/main.jnode.network/v1.8.21-stable-9dc5d1a9/linux-amd64/go1.11.4"
+
+#[derive(Clone,Debug,PartialEq,Eq)]
+pub struct ParityClientData {
+	name: String,
+	variant: Option<String>,
```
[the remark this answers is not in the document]
tomusdrw: I think `--identity` is used to generate private key to encrypt communication and is not really used for version identitfier. This is reserved for parity-ethereum derivatives (like ewf client, etc) afair

util/network/src/client_version.rs
elferdo: Good point.

util/network/src/client_version.rs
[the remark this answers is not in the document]
tomusdrw: You're right. For some reason I though that `--identity` is `--node-key`.

util/network/src/client_version.rs
elferdo: Sure, good point.

util/network/src/client_version.rs
elferdo: I'm not happy with it either. The specification of the format is very loose and every client is more or less free to write their ID as they want (https://github.com/ethereum/wiki/wiki/Client-Version-Strings). We could use a regex to match the version part instead of just checking for a prefixing 'v'.

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+	pub fn client_capabilities_when_parity_old_version_then_handles_large_requests_false() {
+		let client_version_string: String = make_old_semver_version_string();
+
+		let client_version = ClientVersion::from(client_version_string.as_str());
+
+		assert!(!client_version.can_handle_large_requests());
+	}
+
+	// FIXME For some reason the version in this test is considered older than 2.3.0.
+	// A client with this ID _should_ actually be able to handle large requests
```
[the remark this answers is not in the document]
elferdo: Hmmm, that makes sense, but on some tests I did 2.3.0-beta was reported as actually older... I may be recalling wrong, and indeed this seems to be the case (https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=228bb4524e998e451a6ba9ab1c3fe1d8). I'll change the comment.

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+	let service_transactions_version = Version::parse(SERVICE_TRANSACTIONS_VERSION).unwrap();
+
+	*parity_client_data.semver() >= service_transactions_version
+}
+
+impl ClientCapabilities for ClientVersion {
+	fn can_handle_large_requests(&self) -> bool {
+		match self {
+			ClientVersion::ParityClient(data) => {
+				if *data.semver() < Version::parse(PARITY_CLIENT_LARGE_REQUESTS_VERSION).unwrap() {
```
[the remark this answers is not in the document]
elferdo: I considered the option of making ClientVersion a struct and have a field for the provider instead of variants Parity/Geth/Other. When making choices about "what kind of ..." I prefer pattern matching instead of ifs.

Sure, in this case I would try to avoid parsing and comparing the whole time too. Since the comparison is only made in case it's parity, I would add a constructor and a field to ParityClientData for now, and leave ClientVersion as an enum. What do you think?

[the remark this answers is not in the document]
ordian: To be clear, I was actually thinking about preserving the enum and creating a wrapping struct, that contains an enum and a field `can_handle_large_requests`, which is initialized in the constructor.

util/network/src/client_version.rs
tomusdrw: @elferdo I'd say to focus on the variants we know exist and we want to support. So afaict we have either:
- 4-parts client version id (version on the second position)
- 5-parts client version id (variant on the second position and version on the third position)

If we can't parse the string as one of the two variants then we can just assume "unknown". Note that the stuff that changes depending on the version is just some optimizations, i.e.:
1. Don't send service transactions to peers that you know will reject them
2. Don't request more headers if you know that the peer will reject them (we can always fine a different peer to request from anyway).

So we don't have to be super strict on parsing all possible cases here.

util/network/src/client_version.rs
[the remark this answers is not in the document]
tomusdrw: sounds good to me

util/network/src/client_version.rs
elferdo: Ok, I'll look into that.

elferdo: What do you mean? `impl From<T: AsRef<str>> for CV`? Or you mean implementing AsRef<str> for CV? I can see the first alternative being more general than this implementation, and the second one would be an alternative for `fmt::Display` I guess?

util/network/src/client_version.rs
elferdo: Right, agreed. Let's focus on specific, known instances and only try to generalize when the need comes.

util/network/src/client_version.rs
tomusdrw: I meant the former, to just have one `From` implementation (instead of `From<String>` and `From<&str>` as we have now), but it's nitpicking :)

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+use std::fmt;
+
+
+/// Description of the software version running in a peer
+/// according to https://github.com/ethereum/wiki/wiki/Client-Version-Strings
+/// This structure as it is represents the format used by Parity clients. Other
+/// vendors may provide additional fields.
+///
+/// TODO support formats with extra fields, e.g.:
+/// "Geth/main.jnode.network/v1.8.21-stable-9dc5d1a9/linux-amd64/go1.11.4"
```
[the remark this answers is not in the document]
elferdo: It is not implemented and right now we only actually parse "parity" strings. As discussed with @tomusdrw elsewhere, we will focus on parsing the specific cases we know exist (parity with 4 or 5 fields) and simply tag Geth as "Other". I will remove this TODO note if you're ok with that.

util/network/src/client_version.rs
elferdo: Oh, I see. I'm relatively new to rust, so I'm glad to learn the tricks ;-)

elferdo: Yep, it did the trick. Nice :-)

util/network/src/client_version.rs
```diff
@@ -0,0 +1,473 @@
...
+		);
+	} else if let Some(version_number) = &get_number_from_version(tokens[2]) {
+		if tokens.len() < 5 {
+			return Err(());
+		}
+
+		return Ok(
+			ParityClientData {
+				name: tokens[0].to_string(),
+				variant: Some(tokens[1].to_string()),
```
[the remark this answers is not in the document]
elferdo: You are right here. We have two options: we stick to the two basic alternatives we have right now, namely a four and a five argument ID strings with versions in second and third position respectively, or we go to a more generic solution that tries to locate version-like tokens wherever they are. I would go with the simplest one right now.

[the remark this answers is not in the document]
ordian: I think [finding the rightmost `/v`](https://github.com/paritytech/parity-ethereum/pull/10247/files#diff-68aa829147cff9109153923db5442c70L53) was good enough generic solution, alternatively, we can disallow having `/` in `--identity` and go for the simplest one.

util/network/src/client_version.rs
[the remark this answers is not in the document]
elferdo: Now I understand! It took me some time to understand this comment. You mean what I call "variant", ok, I'll change it. Thanks for the pointer.

util/network/src/client_version.rs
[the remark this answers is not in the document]
elferdo: How safe is it to assume version is the third token from the right? Os/target/platform and compiler are always there in parity IDs, right? Because if not, say the string is "Parity/linux/rustc/v2.3.0". Is "linux/rustc" platform and compiler or identity?

For parity, I would expect, from right to left: compiler, os, version, then the string should begin with "Parity/" or "Parity-Ethereum/" and everything in between is identity.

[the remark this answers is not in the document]
ordian: > Os/target/platform and compiler are always there in parity IDs, right?

According to git history of [util/version](https://github.com/paritytech/parity-ethereum/blob/eb3d33ed6ff2a697e9c48eb025c044e6b872f468/util/version/src/lib.rs#L52-L57), yes, at least for the versions we care about.

> Is "linux/rustc" platform and compiler or identity?

Identity (if present) is always inserted before version, so it would be identity (though the string is an invalid version).

> For parity, I would expect, from right to left: compiler, os, version, then the string should begin with "Parity/" or "Parity-Ethereum/" and everything in between is identity.

sounds right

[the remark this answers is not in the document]
elferdo: Then we have an agreement :-)

util/network/src/client_version.rs
```diff
@@ -0,0 +1,516 @@
...
+fn parse_parity_format(client_version: &str) -> Result<ParityClientData, ()> {
+	let tokens: Vec<&str> = client_version.split("/").collect();
+
+	if tokens.len() < PARITY_ID_STRING_MINIMUM_LENGTH {
+		return Err(())
+	}
+
+	let name = tokens[0];
+
+	let identity = if tokens.len() - 3 > 1 {
```
ordian: This feels simpler:
```suggestion
	let identity = if tokens.len() > PARITY_ID_STRING_MINIMUM_LENGTH {
```

util/network/src/client_version.rs
```diff
@@ -0,0 +1,516 @@
...
+//! Parse ethereum client ID strings and provide querying functionality
+
+use semver::Version;
+use std::fmt;
+
+/// Parity client string prefix
+const LEGACY_CLIENT_ID_PREFIX: &str = "Parity";
+const PARITY_CLIENT_ID_PREFIX: &str = "Parity-Ethereum";
+
+const PARITY_ID_STRING_MINIMUM_LENGTH: usize = 4;
```
ordian: I would suggest renaming this constant (`s/LENGTH/TOKENS` ?), since it's not clear w/o a context, that the length is in tokens, separated by `/`, and not bytes.
Also consider making it a local const in `fn parse_parity_format`, it's the only place where this constant is used.

rpc/src/v1/tests/helpers/sync_provider.rs
```diff
@@ -20,6 +20,9 @@ use std::collections::BTreeMap;
 use ethereum_types::H256;
 use parking_lot::RwLock;
 use sync::{SyncProvider, EthProtocolInfo, SyncStatus, SyncState, PeerInfo, TransactionStats};
+use self::ethcore_network::client_version::ClientVersion;
+
+extern crate ethcore_network;
```
ordian: Is this line actually needed?

rpc/src/v1/tests/helpers/sync_provider.rs
```diff
@@ -20,6 +20,9 @@ use std::collections::BTreeMap;
 use ethereum_types::H256;
 use parking_lot::RwLock;
 use sync::{SyncProvider, EthProtocolInfo, SyncStatus, SyncState, PeerInfo, TransactionStats};
+use self::ethcore_network::client_version::ClientVersion;
```
ordian: Doesn't this work?
```suggestion
use network::client_version::ClientVersion;
```

util/network/src/client_version.rs
elferdo: I actually prefer the first alternative for consistency, because in other places we can find `tokens.len() - n`, so it fits better with the rest of the code. The semantics is more clear to me: if the position where identity should be is greater than one, then actually find identity. What I feel really unhappy about is using hardwired numbers. I considered using constants or macros but could not come up with anything beautiful.

util/network/src/client_version.rs
elferdo: Ok, good points both.

rpc/src/v1/tests/helpers/sync_provider.rs
elferdo: Indeed not.

rpc/src/v1/tests/helpers/sync_provider.rs
elferdo: Indeed it does.

util/network/src/client_version.rs
```diff
@@ -0,0 +1,516 @@
...
+use std::fmt;
+
+/// Parity client string prefix
+const LEGACY_CLIENT_ID_PREFIX: &str = "Parity";
+const PARITY_CLIENT_ID_PREFIX: &str = "Parity-Ethereum";
+
+lazy_static! {
+/// Parity versions starting from this will accept block bodies requests
+/// of 256 bodies
+	static ref PARITY_CLIENT_LARGE_REQUESTS_VERSION: Version = Version::parse("2.3.0").unwrap();
```
ngotchac: I think this should be `2.4.0`

elferdo: Done.

ethcore/sync/src/block_sync.rs
```diff
@@ -29,10 +29,13 @@ use ethcore::error::{ImportErrorKind, QueueErrorKind, BlockError, Error as Ethco
...
 use blocks::{BlockCollection, SyncBody, SyncHeader};
 use chain::BlockSet;
+use network::PeerId;
+use network::client_version::ClientCapabilities;
 
 const MAX_HEADERS_TO_REQUEST: usize = 128;
-const MAX_BODIES_TO_REQUEST: usize = 32;
-const MAX_RECEPITS_TO_REQUEST: usize = 128;
+const MAX_BODIES_TO_REQUEST_LARGE: usize = 128;
+const MAX_BODIES_TO_REQUEST_SMALL: usize = 32; // Size request for parity clients prior to 2.3.0
```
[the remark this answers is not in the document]
elferdo: Oh, sorry, my bad. Should've checked before pushing.

5chdn:
@elferdo could you resolve the conflicts?

@tomusdrw are your comments addressed?

ethcore/sync/src/chain/mod.rs
```diff
@@ -342,6 +343,8 @@ pub struct PeerInfo {
 	snapshot_number: Option<BlockNumber>,
 	/// Block set requested
 	block_set: Option<BlockSet>,
+	/// Version of the software the peer is running
+	client_version: ClientVersion
```
tomusdrw: nit: missing comma

rpc/src/v1/tests/mod.rs
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
+
+
```
tomusdrw: nit: not really necessary

util/network/src/client_version.rs
```diff
@@ -0,0 +1,516 @@
...
+	// result. Otherwise return an error.
+	get_number_from_version(tokens[tokens.len() - 3])
+		.and_then(|v| Version::parse(v).ok())
+		.map(|semver| ParityClientData::new(
+			name.to_owned(),
+			identity,
+			semver,
+			os.to_owned(),
+			compiler.to_owned(),
+		)
```
tomusdrw: join the two lines maybe?

5chdn:
It's squash-and-merge day! Sorry for the conflicts again.

ethcore/sync/src/chain/mod.rs
elferdo: Good eye ;-)

util/network/src/client_version.rs
elferdo: I guess you mean the two ')' ? Done.

PR #10254, opened 2019-01-28
Fix v2.2.7 release date in the changelog.

gmajoulet:
Fix v2.2.7 release date in the changelog.

I hate being that guy who sends one line PRs to update the doc, but I installed the beta 2.3 version instead of the stable 2.2.7 because I thought it was one year old... before checking again.


gmajoulet:
[clabot:check]

PR #10258, opened 2019-01-29
Stable: Macos heapsize force jemalloc (#10234)

5chdn:
* Switch to non prefixed malloc_size_of on macos

* Fix

* Testing darwin build

* Fix import

* conflict

* switch heapsize deps commit

* switch heapsize commit

* Rename branch

* Restore gitlab ci to origin

* test for mac

* mac tests?

* Switch of macos CI tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -276,7 +276,7 @@ version = "0.1.0"
 dependencies = [
  "ethereum-types 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ethjson 0.1.0",
- "heapsize 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "heapsize 0.4.2 (git+https://github.com/cheme/heapsize.git?branch=ec-macfix)",
  "keccak-hash 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "parity-bytes 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1361,8 +1366,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "heapsize"
 version = "0.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/cheme/heapsize.git?branch=ec-macfix#aa0fa65a207ad8d66f94c7aec510b2f9dc984dbe"
```
cheme: Here it is the wrong commit (in merged  it is 421df390a930cb523a09e5528e6fe57b534b3b26 ), we need to force it. 
I am not sure how it update, but I should not have let the latest commit of my branch on a bad state (it uses jemalloc linux and was failling on building some proc_macro (when building registrar) in my tests).

cheme: Seems like it is passing on the Ci, but it is because the ci is using rust 1.31 (thanks @ordian for noticing).

5chdn: Uh, ok? Well spotted, how to force a certain commit here?

cheme:
```suggestion
source = "git+https://github.com/cheme/heapsize.git?branch=ec-macfix#421df390a930cb523a09e5528e6fe57b534b3b26"
```

PR #10262, opened 2019-01-30
update ring to 0.14

ordian:
This update is necessary, since ring 0.13 is going to be [yanked soon](https://github.com/briansmith/ring/issues/774#issuecomment-457862149).

Change from `u32` to `NonZeroU32` in parity-crypto was pretty invasive (b347599)


accounts/ethstore/src/ethstore.rs
```diff
@@ -37,11 +38,11 @@ pub struct EthStore {
 impl EthStore {
 	/// Open a new accounts store with given key directory backend.
 	pub fn open(directory: Box<KeyDirectory>) -> Result<Self, Error> {
-		Self::open_with_iterations(directory, KEY_ITERATIONS as u32)
+		Self::open_with_iterations(directory, NonZeroU32::new(KEY_ITERATIONS as u32).expect("KEY_ITERATIONS > 0; qed"))
```
niklasad1: This is just a `constant` and could be `unchecked` too? 
What's the reasoning on determining whether `unchecked` or not?

I guess it makes sense in this case because `KEY_ITERATIONS` is imported from `parity-crypto` right?

ordian: Yes, you're right. I initially used `new_unchecked` only for literals, where is was clear it's safe. But since `crypto::KEY_ITERATIONS` is a constant, it's safe too.

accounts/ethstore/src/account/crypto.rs
```diff
@@ -159,29 +160,31 @@ impl Crypto {
 #[cfg(test)]
 mod tests {
 	use ethkey::{Generator, Random};
-	use super::{Crypto, Error};
+	use super::{Crypto, Error, NonZeroU32};
+
+	const ITERATIONS: NonZeroU32 = unsafe { NonZeroU32::new_unchecked(10240) };
```
tomaka: Although this is technically safe, by ideology it's probably better to not use any unsafe anywhere.

ordian: > by ideology it's probably better to not use any unsafe anywhere

Well, I don't think we should treat "don't use unsafe" as a dogma. The are different kinds of unsafe code. Here the unsafety of `new_unchecked` comes from `const_fn` (and probably `const_panic`) being unstable. 
And, in this case, for literals and constants it's easy to reason about the safety.

Nevertheless, I've eliminated all the calls to `new_unchecked`.

PR #10279, opened 2019-02-03
Additional tests for uint/hash/bytes deserialization.

tomusdrw:


wkarshat:
@5chdn, can you write a single function to test for valid 'hex' format and use it throughout?
Along the lines of:
value.len() >= 4 && value.len() & 1 == 0 && value.starts_with("0x")

niklasad1:
@wkarshat 
> value.len() >= 4 && value.len() & 1 == 0 && value.starts_with("0x")

No, it won't cover `0x`. See https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding

PR #10286, opened 2019-02-03
Backport #10285 to beta

tomaka:


5chdn:
#10285 ?

5chdn:
Do we also need #10284 in beta?

Issue #10121, opened 2019-01-02
Node permission contract enabling leads to CPU overload and peers loosing


VladLupashevskyi:
- **Parity Ethereum version**: 1.11.11 and 2.1.10
- **Operating system**: MacOS / Linux
- **Installation**: homebrew and one-line installer
- **Fully synchronized**: yes
- **Restarted**: yes

After setting up the chain with node permission contract all parity nodes are loosing all peers within a couple of hours. The contract allows access only to few enode addresses and has the following logic:

```solidity
function connectionAllowed(bytes32 /*sl*/, bytes32 /*sh*/, bytes32 pl, bytes32 ph) public view returns (bool) {
        bool plAllowed = peerAllowedL[pl];
        bool phAllowed = peerAllowedH[ph];
        return plAllowed && phAllowed;
}
```

While debugging this issue I found out that when parity has zero peers, one of threads eats 100% CPU, which means that parity does not have enough time to react on trying to connect nodes (which was learned here from similar issues about 0 peers).

After removing `nodePermissionContract` from the json spec - the connections get stabilized as well as CPU usage.

I went further to the source code and tried to figure out what's going wrong.

I have removed the caching part from the code (LruCache) and the node started to work stable even with node permission contract enabled:

https://github.com/paritytech/parity-ethereum/blob/c077dc652d92535cff0afc9ae096c795c4cf39e3/ethcore/node-filter/src/lib.rs#L87-L90

I do not have too much experience with caching, but it seems to me that either the problem starts to appear when cache gets full or some kind of deadlock happens.

I would add a boot flag like `--disable-node-permission-caching` in order to provide a quick fix for blockchains that currently rely on this contract.

What do you think about that?

P.S.: During debugging I found out that the connection allowed check happens BEFORE network ID checking, which leads to huge amount of unnecessary calls to the contract. Was this done intentionally? I would consider move the check to the place when network ID is already checked, if that is possible of course!

joshua-mir:
cc @grbIzl?

tomusdrw:
@VladLupashevskyi 
1. Moving the check after `Status` message seems reasonable to me, I think we don't exchange any info with uncofirmed peers anyway.
2. Regarding the cache - I think the issue is caused by the `Mutex` - it requires exclusive access to check the cache, while the rest of the code could easily run in parallel. Instead of disalbing the cache via CLI flag I would either:
2.1 Disable it completely
2.2. Introduce a different cache behind `RwLock`:
```
[...]

PR #10288, opened 2019-02-03
Node table limiting and cache for node filter

VladLupashevskyi:
After removing node filter cache (#10121), the nodes were still loosing the the peers, however in this case the time when it starts to loose them was very dependent on the CPU power, so when I was testing the fix it was working well for couple of days on my i7 4th gen, but when I left for longer time peers loosing started to occur again.

I have done some research and found out that the problem was in this line:
https://github.com/paritytech/parity-ethereum/blob/06cae8a53556dede67e163325657bee3fba8feda/util/network-devp2p/src/host.rs#L598

where it was querying the contract on every state. But the interesting thing I found that the node table was not limited in size and was always growing. So after 4 hrs of running there were already 10k entries, so the state was queried 10k times each second. In this case introducing the cache is not really a solution, because once we hit the limit of it the querying starts again.

First of all I thought I found a typo in `update_table` fn where `;` was missing after for loop:
https://github.com/paritytech/parity-ethereum/blob/06cae8a53556dede67e163325657bee3fba8feda/util/network-devp2p/src/node_table.rs#L349
But because of lack of experience in rust adding `;` was not helpful.

So I decided to introduce the node table limiting which would allow to use caching for Node Filter, but also guarantee more predictable behaviour for nodes that don't use it, so the large size won't be causing any performance issues when node runs for very long time.

The limiting is done in the way that in case of adding a new node and hitting the limit, the least important node gets removed (either failure one with the most recent last_connected state, or node with unknown last_connected state, or the node with the oldest successful last_connected state).

I introduced simple cache for Node Filter based on `RwLock` which is reset on the new block. It's not a perfect solution, but I would say more generic, and especially useful for the chains where the filter gets dynamically updated. As further improvement I would see an opportunity to introduce a new version of node permission contract where the event `NodeTableUpdated` would be emitted when the list of allowed nodes is updated and clear the cache in this case, or for current contract version - checking the txes to the node permission contract on each block and clear the cache if they are present there (not sure how it would affect the performance though, but I believe it's better than just clearing the cache on each block).

I have done some checks and on current 2.3.0-beta when node filter contract is enabled, the CPU has 30% of load in the beginning and keeps growing, with this update the CPU load is constantly about 15-30% on AuRa chain with block time 4 seconds. When the contract is disabled the load is 5-10%.

Looking forward to get any comments or suggestions to these changes :)


util/network-devp2p/src/node_table.rs
```diff
@@ -235,47 +236,100 @@ impl Hash for Node {
...
+		if self.nodes.insert(node.id, node).is_none() {
+			self.ordered_ids.insert(index, id);
+		};
 	}
 
-	/// Returns a list of ordered nodes according to their most recent contact
-	/// and filtering useless nodes. The algorithm for creating the sorted nodes
-	/// is:
+	/// Get index in the ordered entries vector to insert node based on its last contact value
+	fn get_index_to_insert(&self, last_contact: Option<NodeContact>) -> usize {
```
grbIzl: Sorry, but I don't understand. Why are you introducing so complicated and fragile logic with keeping the order of nodes instead of just to limit the output number of ordered_entries?

ethcore/node-filter/src/lib.rs
```diff
@@ -37,28 +37,44 @@ extern crate tempdir;
...
 	contract_address: Address,
+	cache: RwLock<Cache>
+}
+
+struct Cache {
+	cache: HashMap<NodeId, bool>,
+	order: VecDeque<NodeId>
 }
 
+pub const CACHE_SIZE: usize = MAX_NODES_IN_TABLE + 1024;
```
grbIzl: Just curious, why +1024? :-)

ethcore/node-filter/src/lib.rs
```diff
@@ -37,28 +37,44 @@ extern crate tempdir;
...
+use devp2p::MAX_NODES_IN_TABLE;
+use parking_lot::RwLock;
 
 use_contract!(peer_set, "res/peer_set.json");
 
 /// Connection filter that uses a contract to manage permissions.
 pub struct NodeFilter {
 	client: Weak<BlockChainClient>,
 	contract_address: Address,
+	cache: RwLock<Cache>
```
grbIzl: If we understand the real reason for CPU overload (not limited Node table), can we just rollback #10143 and return previous LRUCache, that was used here? I frankly don't see the advantages of proposed solution in comparison with the previous one.

grbIzl:
@VladLupashevskyi In this part of code: https://github.com/paritytech/parity-ethereum/blob/d89b8d904ff28382d3692290a936706304627255/util/network-devp2p/src/host.rs#L595 we create an iterator over the node table and take some amount of nodes from it (~30 nodes). So the only scenario, when such performance degradation is possible, is when you have a lot of nodes forbidden in your permission contract and iterator goes over them, looking for allowed. I think, that it will be true for whitelisting permission contracts. For blacklisting contracts this code should return the required amount of nodes quicker.

util/network-devp2p/src/node_table.rs
VladLupashevskyi: That's what I was thinking about in the beginning, and of course that would be the easiest solution. However I assume that having limited table, gives us more predictable behaviour from performance point of view. As I said after running node for 4 hrs the table had 10k records and was growing.

ethcore/node-filter/src/lib.rs
VladLupashevskyi: We are chaining reserved nodes with nodes from table here:

https://github.com/paritytech/parity-ethereum/blob/d89b8d904ff28382d3692290a936706304627255/util/network-devp2p/src/host.rs#L587-L591

So if the cache size would be the same as node table size, in case when node table is full, the `reserved_peers.len` amount of unnecessary calls will be done to the contract, since we hit limits of cache.

ethcore/node-filter/src/lib.rs
VladLupashevskyi: The idea of implementing it via `RwLock` was to make it possible to read the cache from different threads without blocking each other.

VladLupashevskyi:
> @VladLupashevskyi In this part of code:
> 
> [parity-ethereum/util/network-devp2p/src/host.rs](https://github.com/paritytech/parity-ethereum/blob/d89b8d904ff28382d3692290a936706304627255/util/network-devp2p/src/host.rs#L595)
> 
> Line 595 in [d89b8d9](/paritytech/parity-ethereum/commit/d89b8d904ff28382d3692290a936706304627255)
> 
>  for id in nodes.filter(|id| 
> we create an iterator over the node table and take some amount of nodes from it (~30 nodes). So the only scenario, when such performance degradation is possible, is when you have a lot of nodes forbidden in your permission contract and iterator goes over them, looking for allowed. I think, that it will be true for whitelisting permission contracts. For blacklisting contracts this code should return the required amount of nodes quicker.

Yes, you are absolutely right - I've experienced this behaviour when was using whitelisting permissioning in the contract. But shouldn't this contract have a generic design, allowing to do whitelisting and blacklisting depending on the design of the blockchain? I believe that first is very crucial in permissioned consortium chains.

grbIzl:
> But shouldn't this contract have a generic design, allowing to do whitelisting and blacklisting depending on the design of the blockchain? I believe that first is very crucial in permissioned consortium chains.
I'm not arguing, it was just an analysis to help better understand the issue. We indeed must support any type of permissioning contract.

ethcore/node-filter/src/lib.rs
grbIzl: Got it. But IMO it's also worth some explanation

VladLupashevskyi: I agree! I would just add a comment there.

ethcore/node-filter/src/lib.rs
grbIzl: Ok. Than may be just change previous Mutex<LruCache> to RwLock<LruCache>? Because order field in Cache is just the implementation of Lru

VladLupashevskyi: Got it! I think that makes sense. Sorry I didn't notice and I was thinking that `LruCache` comes already with with built-in `Mutex` 😅

VladLupashevskyi: Actually I cannot really use `LruCache` since it does not provide non-mutable `get` method, so it's not possible to use `RwLock` with it

util/network-devp2p/src/node_table.rs
grbIzl: Ok. Can you at least create a couple tests for this method? (I mean get_index_to_insert)

util/network-devp2p/src/node_table.rs
```diff
@@ -344,27 +398,52 @@ impl NodeTable {
 	/// Apply table changes coming from discovery
 	pub fn update(&mut self, mut update: TableUpdates, reserved: &HashSet<NodeId>) {
 		for (_, node) in update.added.drain() {
-			let entry = self.nodes.entry(node.id).or_insert_with(|| Node::new(node.id, node.endpoint.clone()));
-			entry.endpoint = node.endpoint;
-		}
+			let mut add = false;
+			{
+				let entry = self.nodes.entry(node.id).or_insert_with(|| {add = true; Node::new(node.id, node.endpoint.clone())});
```
grbIzl: Can you please write it as multiline?

util/network-devp2p/src/node_table.rs
```diff
@@ -344,27 +398,52 @@ impl NodeTable {
...
 	}
 
 	/// Set last contact as success for a node
 	pub fn note_success(&mut self, id: &NodeId) {
+		let mut last_contact = None;
 		if let Some(node) = self.nodes.get_mut(id) {
 			node.last_contact = Some(NodeContact::success());
+			last_contact = node.last_contact
+		}
+		if let Some(pos) = self.ordered_ids.iter().position(|i| id == i) {
```
grbIzl: Code duplication with previous method. Please move it to the separate method (like update_ordered_ids(last_contact))

VladLupashevskyi:
Here I fixed some bugs with get_index_to_insert and added tests for it. I think the most appropriate place to test this function is in `table_last_contact_order` test, so in order to prevent its failing on macos due to lost nanoseconds precision I had to revert the fix 5a1dc3eb8a638d556248799d6a1f60e528331f40 was done by @debris, because the test become more complicated there and use `sleep(Duration::from_micros(1))` instead as simple fix. I think it shouldn't affect the test execution time, however the code become slightly bulky.

VladLupashevskyi:
I also removed `get_mut` method for NodeTable and introduced `get` instead to prevent occasional modification of the nodes HashMap.

VladLupashevskyi:
Could somebody review this? We are looking forward to that PR being reviewed and merged if possible :)

ethcore/node-filter/src/lib.rs
ngotchac: What about the `peek` method: https://docs.rs/lru/0.1.13/lru/struct.LruCache.html#method.peek ?

PR #10291, opened 2019-02-04
fix(osx and windows builds): bump parity-daemonize

niklasad1:
Attempt to fix `Windows` and `OSX` builds

/cc @5chdn @ddorgan can you help me run this on Windows and OSX? Thanks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2423,7 +2423,7 @@ dependencies = [
 
 [[package]]
 name = "parity-daemonize"
-version = "0.1.1"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


HCastano:
Looks like it's building fine on my Mac now.

niklasad1:
It fails because of the CI is running `rust 1.31.1`, see https://github.com/rust-lang/rust/issues/53130. Thus, it is fixed in `rust 1.32`

If we can bump to `rust 1.32` it would be find or any objections that we should support earlier rust versions?

5chdn:
why is CI using outdated rust versions @fevo1971 ?

fevo1971:
> why is CI using outdated rust versions @fevo1971 ?

on mac & windows rust is updated daylay (at ~23:00h), for the docker image parity/rust:gitlab-ci used for all other builds this is not the case (embarrassingly for some quite some time). i coordinated with @TriplEight to finish his work on the new docker image  (based on the *rust:stable* image).

niklasad1:
FYI, I've built `android` successfully locally, [log here](https://gist.githubusercontent.com/niklasad1/6e008b919ccb7864a107b49bfbc7381d/raw/cdc51ebe290bc3dadc130e5f15bfc3334525e428/build.log)

Thus, the `android pipeline` runs also an old version of Rust

Issue #9125, opened 2018-07-15
[Feature request] parity_setMinGasPrice


phahulin:
> I'm running:
>
> - **Which Parity version?**: 1.11.6
> - **Which operating system?**: Linux
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: -
> - **Did you try to restart the node?**: yes

It looks like `parity_setMinGasPrice` and some other apis are deprecated now

https://github.com/paritytech/parity/blob/494eb4ab6b518278fe7bf029de35f4e6de58b464/rpc/src/v1/impls/parity_set.rs#L78-L91

Could someone clarify why they are deprecated and how can those values (gas price, transactions limit, gas limit) be dynamically updated now?

phahulin:
I'm mostly interested in using this for custom chain (not Foundation), to set gasprice depending on the current exchange price, so this is related to https://github.com/paritytech/parity/issues/9089

Tbaut:
They are deprecated and not used anymore in v1.11 in favor of the following config:
```
    --tx-queue-mem-limit=[MB]
        Maximum amount of memory that can be used by the
        transaction queue. Setting this parameter to 0 disables
        limiting. (default: 4)

    --tx-queue-size=[LIMIT]
        Maximum amount of transactions in the queue (waiting to be
        included in next block). (default: 8192)

    --tx-queue-per-sender=[LIMIT]
        Maximum number of transactions per sender in the queue. By
        default it's 1% of the entire queue, but not less than 16.

    --tx-queue-gas=[LIMIT]
        Maximum amount of total gas for external transactions in
        the queue. LIMIT can be either an amount of gas or 'auto'
        or 'off'. 'auto' sets the limit to be 20x the current block
        gas limit. (default: off)
```

More info on the new Tx queue here https://wiki.parity.io/Transactions-Queue

phahulin:
@Tbaut these are configuration options. My question is about updating those values via api (especially gas price), without client restarts.

Tbaut:
>Could someone clarify why they are deprecated 

Answered above

>how can those values (gas price, transactions limit, gas limit) be dynamically updated now?

You can't fo it dynamically, it's a feature request :)

Tbaut:
Please modify the title accordingly.

phahulin:
[...]

PR #10294, opened 2019-02-04
Revive parity_setMinGasPrice RPC call

HCastano:
This RPC call was deprecated a while back during a refactoring, but there's been a feature request (#9125) wanting it to be restored.

This PR gives the people what they want.


ethcore/src/miner/miner.rs
```diff
@@ -865,6 +865,26 @@ impl miner::MinerService for Miner {
 		self.params.read().gas_range_target.0 / 5
 	}
 
+	fn set_minimal_gas_price(&self, new_price: U256) -> Result<bool, ()> {
+		if cfg!(feature = "price-info") {
+			warn!("Can't update fixed gas price while automatic gas calibration is enabled.");
```
tomusdrw: That's not 100% true, the fact that there is an option to set non-fixed `GasPricer` doesn't imply that it's actually set.
I think this `if` should rather be replaced with something like:
```
if let GasPricer::Fixed(mut val) = *self.gas_pricer.lock() {
    // update and recalibrate
   *val = new_price;
   // ...
} else {
  // Stringified error would be returned to RPC caller which imho would be nice.
  return Err("...");
}
```

ethcore/src/miner/miner.rs
```diff
@@ -865,6 +865,26 @@ impl miner::MinerService for Miner {
...
+			return Err(());
+		}
+
+		trace!(target: "miner", "minimal_gas_price: recalibrating fixed...");
+		*self.gas_pricer.lock() = GasPricer::new_fixed(new_price);
+
+		let txq = self.transaction_queue.clone();
+		let mut options = self.options.pool_verification_options.clone();
+		self.gas_pricer.lock().recalibrate(move |gas_price| {
```
tomusdrw: Is it worth keeping it DRY? and re-using the same method as for per-block update?

ethcore/src/miner/miner.rs
```diff
@@ -1646,4 +1666,19 @@ mod tests {
 
 		assert!(miner.is_currently_sealing());
 	}
+
+	#[test]
+	#[cfg(not(feature = "price-info"))]
```
tomusdrw: Would be good to have test for both cases.

rpc/src/v1/impls/parity_set.rs
```diff
@@ -69,9 +69,11 @@ impl<C, M, U, F> ParitySet for ParitySetClient<C, M, U, F> where
 	F: Fetch + 'static,
 {
 
-	fn set_min_gas_price(&self, _gas_price: U256) -> Result<bool> {
-		warn!("setMinGasPrice is deprecated. Ignoring request.");
-		Ok(false)
+	fn set_min_gas_price(&self, gas_price: U256) -> Result<bool> {
+		match self.miner.set_minimal_gas_price(gas_price.into()) {
+			Ok(success) => Ok(success),
+			Err(_) => Err(errors::request_rejected()),
```
tomusdrw: I think this error is reserved for signer, maybe let's return `unsupported` with explanation here?

rpc/src/v1/tests/mocked/parity_set.rs
```diff
@@ -106,7 +106,24 @@ fn rpc_parity_set_min_gas_price() {
...
 
 	let request = r#"{"jsonrpc": "2.0", "method": "parity_setMinGasPrice", "params":["0xcd1722f3947def4cf144679da39c4c32bdc35681"], "id": 1}"#;
-	let response = r#"{"jsonrpc":"2.0","result":false,"id":1}"#;
+	let response = r#"{"jsonrpc":"2.0","result":true,"id":1}"#;
+
+	assert_eq!(io.handle_request_sync(request), Some(response.to_owned()));
+}
+
+#[test]
+#[cfg(feature = "price-info")]
```
tomusdrw: I think `rpc` doesn't' have this feature so the test should be enable by default, but `TestMiner` might need a way to mock the error response.

ethcore/src/miner/miner.rs
tomusdrw: I think the features setting is irrelevant for this test, let's just remove this line.

rpc/src/v1/tests/mocked/parity_set.rs
tomusdrw: Still valid?

HCastano: I've removed the need for the `cfg` flag in [8f04c73](/paritytech/parity-ethereum/pull/10294/commits/8f04c73199f1314c952e6ef24c5af986973f14c2).

ethcore/src/miner/miner.rs
HCastano: I think it was left over from earlier, but let me double check.

ethcore/src/miner/miner.rs
HCastano: We only call `recalibrate()` twice, so for now I'm going to say no unless somebody else has a strong opinion about it.

tomusdrw:
@HCastano seems like the merge went wrong, can you check?

PR #10295, opened 2019-02-04
Make specification of protocol in SyncRequester::send_request explicit

### LINKED ISSUE #8605: Generic request sender in sync is not generic enough
It seems that `Requester::send_request` only supports two protocols (ETH, PAR), while claiming it's generic:
https://github.com/paritytech/parity/blob/master/ethcore/sync/src/chain/requester.rs#L136

Actually it should take under consideration the protocols supported by remote peer and also other protocols than ETH and PAR.

elferdo:
The function `SyncRequester::send_request` has a simple logic to match a given `packet_id` to its corresponding protocol. The logic does not seem to be robust, though, and expecting clients to provide the protocol themselves is not too much burden. If needed, we can still reimplement the matching logic separately.

As a consequence, `SyncIO` needs no longer provide `send` and `send_protocol`.

Fixes #8605


elferdo:
[clabot:check]

ethcore/sync/src/chain/propagator.rs
```diff
@@ -342,8 +343,8 @@ impl SyncPropagator {
 	}
 
 	/// Generic packet sender
-	pub fn send_packet(sync: &mut SyncIo, peer_id: PeerId, packet_id: PacketId, packet: Bytes) {
-		if let Err(e) = sync.send(peer_id, packet_id, packet) {
+	pub fn send_packet(sync: &mut SyncIo, protocol: ProtocolId, peer_id: PeerId, packet_id: PacketId, packet: Bytes) {
+		if let Err(e) = sync.send_protocol(protocol, peer_id, packet_id, packet) {
```
tomusdrw: Interesting find, afaict this change is actually breaking (i.e. `propagate_private_transaction` and `propagate_consensus_packet`) will now always use `WARP_PROTOCOL_ID` while before it was up to the network to decide on the protocol.
Anyone knows how it worked before and if we are not breaking anything here? CC @grbIzl

ethcore/sync/src/chain/mod.rs
```diff
@@ -103,15 +103,15 @@ use fastmap::{H256FastMap, H256FastSet};
...
+use network::{self, PeerId, PacketId, ProtocolId};
 use ethcore::client::{BlockChainClient, BlockStatus, BlockId, BlockChainInfo, BlockQueueInfo};
 use ethcore::snapshot::{RestorationStatus};
 use sync_io::SyncIo;
 use super::{WarpSync, SyncConfig};
 use block_sync::{BlockDownloader, DownloadAction};
 use rand::Rng;
 use snapshot::{Snapshot};
-use api::{EthProtocolInfo as PeerInfoDigest, WARP_SYNC_PROTOCOL_ID, PriorityTask};
+use api::{EthProtocolInfo as PeerInfoDigest, ETH_PROTOCOL, WARP_SYNC_PROTOCOL_ID, PriorityTask};
```
ngotchac: I think we should rename `WARP_SYNC_PROTOCOL_ID` to `WARP_SYNC_PROTOCOL` (as there are already `ETH_PROTOCOL` and `LIGHT_PROTOCOL`)

ethcore/sync/src/chain/mod.rs
```diff
@@ -1328,8 +1328,8 @@ impl ChainSync {
 	}
 
 	/// Broadcast private transaction message to peers.
-	pub fn propagate_private_transaction(&mut self, io: &mut SyncIo, transaction_hash: H256, packet_id: PacketId, packet: Bytes) {
-		SyncPropagator::propagate_private_transaction(self, io, transaction_hash, packet_id, packet);
+	pub fn propagate_private_transaction(&mut self, io: &mut SyncIo, transaction_hash: H256, protocol: ProtocolId, packet_id: PacketId, packet: Bytes) {
```
ngotchac: I think they only use `WARP_SYNC_PROTOCOL`, right?

ethcore/sync/src/chain/propagator.rs
elferdo: Hmmm, I guess I made the assumption that:
```
pub const CONSENSUS_DATA_PACKET: u8 = 0x15;
pub const PRIVATE_TRANSACTION_PACKET: u8 = 0x16;`
```
somehow implied `WARP`, but sure, it's just an assumption which may not be true and break.

grbIzl: @tomusdrw if i understand correctly, this logic relied upon the deleted if case in requester module. So with this change the sender specifies protocol explicitly but in principal nothing has changed.

ethcore/sync/src/chain/requester.rs
```diff
@@ -129,23 +128,20 @@ impl SyncRequester {
...
-	fn send_request(sync: &mut ChainSync, io: &mut SyncIo, peer_id: PeerId, asking: PeerAsking,  packet_id: PacketId, packet: Bytes) {
+	fn send_request(sync: &mut ChainSync, io: &mut SyncIo, peer_id: PeerId, asking: PeerAsking,  protocol: ProtocolId, packet_id: PacketId, packet: Bytes) {
 		if let Some(ref mut peer) = sync.peers.get_mut(&peer_id) {
 			if peer.asking != PeerAsking::Nothing {
 				warn!(target:"sync", "Asking {:?} while requesting {:?}", peer.asking, asking);
 			}
 			peer.asking = asking;
 			peer.ask_time = Instant::now();
-			// TODO [ToDr] This seems quite fragile. Be careful when protocol is updated.
-			let result = if packet_id >= ETH_PROTOCOL_VERSION_63.1 {
```
grbIzl: I meant this if case

grbIzl: @tomusdrw ^^

ethcore/sync/src/chain/mod.rs
elferdo: In fact, the calling functions specify `WARP_...`, yes, and it looks like they're only called in one place. Maybe we can move the constant here.

ethcore/sync/src/chain/requester.rs
tomusdrw: That's a different module (and different case), that's why I asked about the original one.
The latter is using `sync.send(` directly which defaults to `network.send(` which in turn defaults to `self.send_protocl(self.protocol)` afair.
So unless I'm missing something both of the methods should get sent using `self.protocol` (which is rpobably ETH_63, not WARP) but somehow get decoded correctly (cause I don't assume they are broken, just some quirk or my misunderstanding causes them to work correctly.)

grbIzl: Ok. Made more investigation. As far as I've understood, protocol id for sending packets is set here https://github.com/paritytech/parity-ethereum/blob/06cae8a53556dede67e163325657bee3fba8feda/util/network-devp2p/src/host.rs#L839 in terms of connected callback. I'm not familiar with this code, but it seems, that this PR may break this logic (when protocol id is defined during the session by libp2p).

grbIzl: Or may be I'm wrong again :-( For private and consensus packets it's specified explicitly here: https://github.com/paritytech/parity-ethereum/blob/06cae8a53556dede67e163325657bee3fba8feda/ethcore/sync/src/api.rs#L574

elferdo: To be honest, I feel uncomfortable refactoring without enough tests, and from the code alone I think it is not easy to reason about what is happening.

How about I close this PR and begin writing tests for what's already in place? That should give us some peace of mind.

elferdo:
Any thoughts how to proceed on this? How can I test that the changes work? I have tried running a node and connecting to just one reserved peer, a parity node, and synchronization seems to proceed fine both with and without warp.

ethcore/sync/src/tests/helpers.rs
```diff
@@ -80,6 +80,16 @@ impl<'p, C> Drop for TestIo<'p, C> where C: FlushingBlockChainClient, C: 'p {
 	}
 }
 
+fn assert_packet_id_matches_protocol(protocol: &ProtocolId, packet_id: &PacketId) {
+	match packet_id {
```
tomusdrw: I don't think we should crash in such case, can we just print warnings/errors if that happens? In our particular case crashing in most cases is the worst thing we can do. If the node can partially (without some features) operate it should, otherwise one can easily kill the entire network by exploiting a secondary feature.

elferdo: You mean the `assert`s? Actually this code is only used inside tests, as far as I can tell. It is called from `TestIo` which is a double for an actual `SyncIo`. I would rather have this expectations stated explicitly in the tests, but some refactoring would be needed for that.

tomusdrw: Oh, sorry didn't notice it's a test code, in such case it's all good!

grbIzl: We have kind of integration test for private transactions in the sync module. Corresponding test implementations for underlying message processing was changed in this PR as well (in EthPeer/process_io_message). So it's the best, that we can do without real e2e.

elferdo: Yeah, though there are some problems with those tests. For instance https://github.com/paritytech/parity-ethereum/blob/8ab6d8981027426dd5a6d4de6c2ed301d0a724ae/ethcore/sync/src/api.rs#L573-L584 does not really get tested. Instead, the tests call https://github.com/paritytech/parity-ethereum/blob/8ab6d8981027426dd5a6d4de6c2ed301d0a724ae/ethcore/sync/src/tests/helpers.rs#L232-L241

I have been looking into testing `EthSync` more thoroughly, but it's not so easy right now. One of the reasons, for instance, is that right now we cannot easily inject a `TestIo` into it: https://github.com/paritytech/parity-ethereum/blob/8ab6d8981027426dd5a6d4de6c2ed301d0a724ae/ethcore/sync/src/api.rs#L575

tomusdrw:
@elferdo I just noticed those new commits here, I think it would be better to merge this PR asap and then do any follow up work in susbsequent prs. This one is already reviewed and marked as good.

elferdo:
> @elferdo I'm pretty sure we won't break anything in this PR. As @grbIzl noticed, the `protocol` of the network is often overwritten using `with_context` function - IMHO this seems super confusing, I'd prefer to be explicit and use `send_protocol` always, alternatively as we spoke in DM we could tangle packet and protocol together (using an enum and `trait Packet{ fn id() -> u8, fn protocol() -> T }` for instance) to prevent any amibguity.

I implemented something like what you propose with `trait Packet[...]`, just in the context of syncing, that means `ethcore-devp2p` still uses both `protocol` and `packet_id` as arguments.

elferdo:
> @elferdo I just noticed those new commits here, I think it would be better to merge this PR asap and then do any follow up work in susbsequent prs. This one is already reviewed and marked as good.

Oh, ok, sorry. Should I revert these commits and open a new PR then?

tomusdrw:
> Should I revert these commits and open a new PR then?

Yes, if you don't mind. I think that will be better, cause otherwise we may end up with never ending story of commit-review cycle and never get anything merged. And it's going to be harder and harder to keep that PR mergeable to master.

elferdo:
Sure, that makes sense. I'll revert the last changes.

ethcore/sync/src/chain/propagator.rs
```diff
@@ -42,6 +43,28 @@ use super::{
 	TRANSACTIONS_PACKET,
 };
 
+/// Checks if peer is able to process service transactions
+fn accepts_service_transaction(client_id: &str) -> bool {
```
grbIzl: Didn't find the usage for this. What was the goal for this method?

elferdo: My bad, merged the wrong version resolving the conflict.

grbIzl:
Ok. Let's try to close this one and go further

PR #10168, opened 2019-01-10
snap: official image / test

TriplEight:
snapcraft has the official image `snapcore/snapcraft`, makes no sense to maintain our version.

PR #10297, opened 2019-02-04
CI optimizations

TriplEight:
DONE:
new cache handler: `sccache`
`npm(nodejs v10)` and `yarn` for autodocs
`test-all.sh` - del lines checking changes in rust code
`gitlab-ci` - `test-all.sh + nightly/beta` removed, we won't need them
`GIT_STRATEGY` and `GIT_SUBMODULE_STRATEGY` - changed to fetch
this `Dockerfile` for testing only, will change the usual one as soon as it works

TODO:
update the initial container in CI
offer a piplene with cargo check for every commit, feedback into personal chat
sccache to work in Gcloud
https://vfoley.xyz/rust-compile-speed-tips/ LTO - can speed up

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ stages:
   - publish
   - optional
 
-image:                             parity/rust-linux:stretch
+image:                             parity/rust-parity-ethereum-build:stretch
 
 variables:
   GIT_STRATEGY:                    fetch
@@ -44,7 +44,7 @@ test-linux:
   script:
     - scripts/gitlab/test-all.sh
   tags:
-    - rust-stable
+    - linux-docker
 
 test-audit:
   stage:                           test
@@ -53,7 +53,7 @@ test-audit:
     - set -u
     - cargo audit
   tags:
-    - rust-stable
+    - linux-docker
 
 build-linux:
   stage:                           build
@@ -62,7 +62,7 @@ build-linux:
     - scripts/gitlab/build-unix.sh
   <<:                              *collect_artifacts
   tags:
-    - rust-stable
+    - linux-docker
 
 build-darwin:
   stage:                           build
@@ -116,19 +116,65 @@ publish-snap:
   allow_failure:                   true
   <<:                              *collect_artifacts
 
-publish-awss3:
+publish:onnet:update:
   stage:                           publish
   only:                            *releaseable_branches
-  cache: {}
+  cache:                           {}
   dependencies:
     - build-linux
     - build-darwin
     - build-windows
   before_script:                   *determine_version
   script:
-    - scripts/gitlab/publish-awss3.sh
+    - scripts/gitlab/publish-onnet-update.sh
   tags:
-    - shell
+    - linux-docker
+
+# configures aws for fast uploads/syncs
+.s3_before_script:                &s3_before_script
+  before_script:
+    - mkdir -p ${HOME}/.aws
+    - |
+      cat > ${HOME}/.aws/config <<EOC
+      [default]
+      s3 =
+        max_concurrent_requests = 20
+        max_queue_size = 10000
+        multipart_threshold = 64MB
+        multipart_chunksize = 16MB
+        max_bandwidth = 50MB/s
+        use_accelerate_endpoint = false
+        addressing_style = path
+      EOC
+
+publish:awss3:release:
+  image:                           parity/awscli:latest
+  stage:                           publish
+  only:                            *releaseable_branches
+  cache:                           {}
+  dependencies:
+    - build-linux
+    - build-darwin
+    - build-windows
+  variables:
+    GIT_STRATEGY:                  none
+  <<:                              *s3_before_script
+  script:
+    - echo "__________Push binaries to AWS S3____________"
+    - case "${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}" in
+        (beta|stable|nightly)
+          export BUCKET=releases.parity.io/ethereum;
+          ;;
+        (*)
+          export BUCKET=builds-parity;
+          ;;
+      esac
+    - aws s3 sync ./artifacts s3://${BUCKET}/${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}/
+  after_script:
+    - aws s3 ls s3://${BUCKET}/latest/
+        --recursive --human-readable --summarize
+  tags:
+  - linux-docker
 
 publish-docs:
   stage:                            publish
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ stages:
   - publish
   - optional
 
-image:                             parity/rust-ci:latest
+image:                             parity/rust:gitlab-ci
 
 variables:
   GIT_STRATEGY:                    fetch
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,9 +4,11 @@ stages:
   - publish
   - optional
 
-image:                             parity/rust:gitlab-ci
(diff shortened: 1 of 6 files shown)


scripts/gitlab/test-all.sh
```diff
@@ -1,34 +1,11 @@
...
-
-if [ "${RUST_FILES_MODIFIED}" = "0" ]
-then
-  echo "__________Skipping Rust tests since no Rust files modified__________";
-  exit 0
-fi
-
-rustup default $1
-
+# FIXME:
```
kirushik: Is this intended?
Seems like you have something to finalize here (or it's just a matter of a lost commit).

scripts/gitlab/build-unix.sh
```diff
@@ -27,20 +27,18 @@ fi
...
-cd artifacts
-mkdir -p $CARGO_TARGET
-cd $CARGO_TARGET
+mkdir -p artifacts/$CARGO_TARGET
+
 if [ "${CARGO_TARGET}" = "armv7-linux-androideabi" ]
 then
 # only thing we need for android
- cp -v ../../target/$CARGO_TARGET/release/libparity.so ./libparity.so
+  cp -v target/$CARGO_TARGET/release/libparity.so ./libparity.so
```
gabreal: binaries have to go into the artifacts directory.
initially the script cd'ed into it now maybe
```
cp -v target/$CARGO_TARGET/release/libparity.so ./artifacts/libparity.so
```

gabreal: same below for `for binary in $(ls)`
should work with
```
for binary in $(ls artifacts)
```
but better test.

test.sh
```diff
@@ -1,33 +1,12 @@
 #!/bin/sh
 # Running Parity Full Test Suite
+echo "________Running test.sh________"
 
 FEATURES="json-tests,ci-skip-issue"
 OPTIONS="--release"
 VALIDATE=1
 THREADS=8
 
-case $1 in
```
gabreal: this looks like no problem for the ci, but may break test on developers machines. maybe worth an announcement.

scripts/gitlab/test-all.sh
```diff
@@ -1,34 +1,11 @@
...
-    ;;
-  (*)
-    export GIT_COMPARE=master
-    ;;
-esac
-
-export RUST_FILES_MODIFIED="$(git --no-pager diff --name-only $GIT_COMPARE...$CI_COMMIT_SHA | grep -v -e ^\\. -e ^LICENSE -e ^README.md -e ^CHANGELOG.md -e ^test.sh -e ^scripts/ -e ^docs/ -e ^docker/ -e ^snap/ | wc -l | tr -d ' ')"
-echo "RUST_FILES_MODIFIED: $RUST_FILES_MODIFIED"
-
-if [ "${RUST_FILES_MODIFIED}" = "0" ]
```
gabreal: why remove this? looks quite useful to only run the test when the rust source code has changed.

.gitlab-ci.yml
```diff
@@ -7,6 +7,8 @@ stages:
 image:                             parity/rust:gitlab-ci
```
gabreal: didn't you want to switch the build container, too?

gabreal: I see, just read your comment afterwards.

gabreal:
where is the configuration of `sccache`?

test.sh
TriplEight: removed it as never saw any parameters being passed to this script, but you're right, devs could make use of it.

TriplEight: I remember, Pierre told it's not needed a while ago.

scripts/gitlab/test-all.sh
TriplEight: With cache, it won't be needed as it won't compile while cargo knows it already was.
But let's return it for now, makes sense.

scripts/gitlab/build-unix.sh
TriplEight: thanks for pointing that out, fixed in next commit

5chdn:
snap misses libdb5.3, didn't we fix that already?

```
Failed to fetch stage packages: Error downloading packages for part 'parity': The package 'libdb' was not found..
```

https://gitlab.parity.io/parity/parity-ethereum/-/jobs/121253

---

AWS login failed 

```
Unable to locate credentials. You can configure credentials by running "aws configure".
```

https://gitlab.parity.io/parity/parity-ethereum/-/jobs/121254

.gitlab-ci.yml
```diff
@@ -2,11 +2,14 @@ stages:
   - test
   - build
   - publish
+  - publish-onchain
```
5chdn: why is this an additional stage?

.gitlab-ci.yml
```diff
@@ -112,19 +118,66 @@ publish-snap:
...
+publish:onnet:update:
+  stage:                           publish-onchain
   only:                            *releaseable_branches
-  cache: {}
+  cache:                           {}
   dependencies:
     - build-linux
     - build-darwin
     - build-windows
+    - publish:awss3:release
```
5chdn: can we use dashs rather than colons?

.gitlab-ci.yml
```diff
@@ -150,22 +203,3 @@ build-android:
   allow_failure:                   true
   <<:                              *collect_artifacts
 
-test-beta:
-  stage:                           optional
```
5chdn: why did we remove this?

.gitlab-ci.yml
```diff
@@ -150,22 +203,3 @@ build-android:
...
-  variables:
-    RUN_TESTS:                     cargo
-  script:
-    - scripts/gitlab/test-all.sh beta
-  tags:
-    - rust-beta
-  allow_failure:                   true
-
-test-nightly:
-  stage:                           optional
```
5chdn: why did we remove this?

.gitlab-ci.yml
TriplEight: should depend on publishing to AWS.
This is not the final version of CI though.

.gitlab-ci.yml
TriplEight: Wanted to move to colons in the final CI version, what's wrong with them? I really do not care, just thought it's nicer.

kirushik: @TriplEight colons don't work for Windows jobs. So either we have some mix, or it's dashes-only.

kirushik: AFAIR it's related to Windows FS limitations: gitlab-runner wants to create a file with the job name somewhere, and if it fails — the whole job fails.

.gitlab-ci.yml
TriplEight: These two jobs were to test the code on `stable` Rust against `beta` and `nightly` Rust.
Developers answered they code only on `stable` Rust so it's not needed. What could be a bit more helpful is nightly build.

.gitlab-ci.yml
TriplEight: ^

TriplEight:
> snap misses libdb5.3, didn't we fix that already?
> 
> ```
> Failed to fetch stage packages: Error downloading packages for part 'parity': The package 'libdb' was not found..
> ```
snap fix is not merged yet: https://github.com/paritytech/parity-ethereum/pull/10168/files 
> https://gitlab.parity.io/parity/parity-ethereum/-/jobs/121253
> 
> AWS login failed
fixed, in GitLab's env variables.

.gitlab-ci.yml
TriplEight: oh, didn't know, will definitely fix later.

5chdn:
This should not have been merged. I'm still on the fence with changes introduced here.

PR #10298, opened 2019-02-05
Convert to jsonrpc-derive, use jsonrpc-* from crates.io

ascjones:
Replaces `jsonrpc-macros` with `jsonrpc-derive` custom trait attribute. With the added bonus of now using published crates.

### DIFF
--- a/rpc/src/v1/traits/debug.rs
+++ b/rpc/src/v1/traits/debug.rs
@@ -21,9 +21,8 @@ use jsonrpc_derive::rpc;
 
 use v1::types::RichBlock;
 
-#[rpc]
 /// Debug RPC interface.
-
+#[rpc]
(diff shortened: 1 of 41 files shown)


rpc/src/v1/traits/debug.rs
```diff
@@ -17,14 +17,15 @@
...
 use v1::types::RichBlock;
 
-build_rpc_trait! {
-	/// Debug RPC interface.
-	pub trait Debug {
-		/// Returns recently seen bad blocks.
-		#[rpc(name = "debug_getBadBlocks")]
-		fn bad_blocks(&self) -> Result<Vec<RichBlock>>;
-	}
+#[rpc]
```
tomusdrw: attributes should go after docs

PR #10305, opened 2019-02-06
Don't add discovery initiators to the node table

### LINKED ISSUE #10230: Performance degression on Kovan
- **Parity Ethereum version**: 2.3.0
- **Operating system**: Linux
- **Installation**: auto-updater
- **Fully synchronized**: yes
- **Network**: kovan
- **Restarted**: yes

```
 > tac parity-error-1.log | grep Starting
2019-01-16 01:08 +00:00: 2019-01-16 01:08:51 UTC Starting Parity-Ethereum/v2.3.0-beta-10657d9-20190115/x86_64-linux-gnu/rustc1.31.1
2018-12-25 19:11 +00:00: 2018-12-25 19:11:07 UTC Starting Parity-Ethereum/v2.2.5-beta-7fbcdfe-20181213/x86_64-linux-gnu/rustc1.31.0
```

After upgrading from 2.2.5 to 2.3.0 on Kovan, the following happened:

screenshot from 2019-01-22 13-55-51

kirushik:
@TriplEight please verify that this closes #10230

### DIFF
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -56,6 +56,8 @@ const REQUEST_BACKOFF: [Duration; 4] = [
 
 const NODE_LAST_SEEN_TIMEOUT: Duration = Duration::from_secs(24*60*60);
 
+const OBSERVED_NODES_MAX_SIZE: usize = 10000;
+
 #[derive(Clone, Debug)]
 pub struct NodeEntry {
 	pub id: NodeId,
@@ -96,7 +98,21 @@ struct FindNodeRequest {
 #[derive(Clone, Copy)]
 enum PingReason {
 	Default,
-	FromDiscoveryRequest(NodeId)
+	FromDiscoveryRequest(NodeId, NodeValidity)
+}
+
+#[derive(Clone, Copy, PartialEq)]
+enum NodeCategory {
+	Bucket,
+	Observed
+}
+
+#[derive(Clone, Copy, PartialEq)]
+enum NodeValidity {
+	Ourselves,
+	ValidNode(NodeCategory),
+	ExpiredNode(NodeCategory),
+	UnknownNode
 }
 
 struct PingRequest {
@@ -146,7 +162,8 @@ pub struct Discovery<'a> {
 	discovery_id: NodeId,
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
-	other_observerd_nodes: LruCache<NodeId, NodeEndpoint>, // Sometimes we don't want to add nodes to the NodeTable, but still want to
+	other_observed_nodes: LruCache<NodeId, (NodeEndpoint, Instant)>, // Sometimes we don't want to add nodes to the NodeTable, but still want to
+	// keep track of them to avoid excessive pinging (happens when an unknown node sends a discovery request to us -- the node might be on a different net).
 	in_flight_pings: HashMap<NodeId, PingRequest>,
 	in_flight_find_nodes: HashMap<NodeId, FindNodeRequest>,
 	send_queue: VecDeque<Datagram>,
@@ -173,7 +190,7 @@ impl<'a> Discovery<'a> {
 			discovery_id: NodeId::new(),
 			discovery_nodes: HashSet::new(),
 			node_buckets: (0..ADDRESS_BITS).map(|_| NodeBucket::new()).collect(),
-			other_observerd_nodes: LruCache::new(256),
+			other_observed_nodes: LruCache::new(OBSERVED_NODES_MAX_SIZE),
 			in_flight_pings: HashMap::new(),
 			in_flight_find_nodes: HashMap::new(),
 			send_queue: VecDeque::new(),
@@ -544,13 +561,19 @@ impl<'a> Discovery<'a> {
 		};
 
 		if let Some((node, ping_reason)) = expected_node {
-			if let PingReason::FromDiscoveryRequest(target) = ping_reason {
+			if let PingReason::FromDiscoveryRequest(target, validity) = ping_reason {
 				self.respond_with_discovery(target, &node)?;
 				// kirushik: I would prefer to probe the network id of the remote node here, and add it to the nodes list if it's on "our" net --
 				// but `on_packet` happens synchronously, so doing the full TCP handshake ceremony here is a bad idea.
-				// So instead we just LRU-caching 256 most recently seen nodes to avoid unnecessary pinging
-				self.other_observerd_nodes.insert(node.id, node.endpoint);
-				Ok(None)
+				// So instead we just LRU-caching most recently seen nodes to avoid unnecessary pinging
+				match validity {
+					NodeValidity::UnknownNode | NodeValidity::ExpiredNode(NodeCategory::Observed) => {
+						self.other_observed_nodes.insert(node.id, (node.endpoint, Instant::now()));
+						Ok(None)
+					},
+					NodeValidity::ExpiredNode(NodeCategory::Bucket) => Ok(self.update_node(node)),
+					_ => Ok(None)
+				}
 			} else {
 				Ok(self.update_node(node))
 			}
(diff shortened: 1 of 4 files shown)


util/network-devp2p/src/discovery.rs
```diff
@@ -145,6 +146,7 @@ pub struct Discovery<'a> {
 	discovery_id: NodeId,
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
+	other_observerd_nodes: LruCache<NodeId, NodeEndpoint>, // Sometimes we don't want to add nodes to the NodeTable, but still want to
```
tomusdrw: typo: `observed` 
`[..] want to` what?

util/network-devp2p/src/discovery.rs
```diff
@@ -171,6 +173,7 @@ impl<'a> Discovery<'a> {
 			discovery_id: NodeId::new(),
 			discovery_nodes: HashSet::new(),
 			node_buckets: (0..ADDRESS_BITS).map(|_| NodeBucket::new()).collect(),
+			other_observerd_nodes: LruCache::new(256),
```
tomusdrw: Do you think it's going to be enough? Maybe let's keep this number higher, I don't think it requires a lot of memory.

util/network-devp2p/src/discovery.rs
```diff
@@ -589,7 +598,7 @@ impl<'a> Discovery<'a> {
 			debug!(target: "discovery", "Found a known node in a bucket when processing discovery: {:?}/{:?}", known_node, node);
 			(known_node.address.endpoint == node.endpoint) && (known_node.last_seen.elapsed() < NODE_LAST_SEEN_TIMEOUT)
```
tomusdrw: I wonder if when we are in this case (and get `false`) we should still update the `node_buckets` when ping succeeds. Currently it's going to be cached in `other_observed_nodes`, but because it's also on the bucket list we will never get `true` for that node again.
Alternatively maybe change the order of checking the cache first and then node_buckets?

util/network-devp2p/src/discovery.rs
```diff
@@ -574,7 +583,7 @@ impl<'a> Discovery<'a> {
 		Ok(None)
 	}
 
-	fn is_a_valid_known_node(&self, node: &NodeEntry) -> bool {
+	fn is_a_valid_known_node(&mut self, node: &NodeEntry) -> bool {
 		let id_hash = keccak(node.id);
 		let dist = match Discovery::distance(&self.id_hash, &id_hash) {
 			Some(dist) => dist,
```
tomusdrw: Doesn't returning `false` in case of ourselves cause a ping to be sent anyway?

kirushik: Good catch! Will make it an enum

util/network-devp2p/src/discovery.rs
kirushik: Shit, second line of a multiline comment is lost; will fix, thanks for spotting.

util/network-devp2p/src/discovery.rs
kirushik: It's just an _uneducated_ guess.
How many unrecognized nodes can ask us for peers in a real-life scenario?

kirushik: Hmmm, I think for our bootnodes it might be an order magnitude higher; I think something like 10000 should be enough (given in account a total number of live nodes in Ethereum mainnet =) And our records are about 80-90 bytes (64 bytes for NodeId, plus ipv6 address (16 bytes) and port (two bytes), plus a couple of bytes for the enum tracking and internal state keeping for the cache structure itself); so 10K records will end up in 1 megabyte of Ram in the worst case, which I'd say should be OK.

Thanks for bringing this up.

ngotchac:
Wouldn't it be easier to read and _maybe_ safer to just ignore `FindNode` requests from nodes that haven't sent a Ping yet (instead of adding them to a list of nodes we should Ping and so on)?

util/network-devp2p/src/discovery.rs
```diff
@@ -55,6 +56,8 @@ const REQUEST_BACKOFF: [Duration; 4] = [
 
 const NODE_LAST_SEEN_TIMEOUT: Duration = Duration::from_secs(24*60*60);
 
+const OBSERVED_NODES_MAX_SIZE: usize = 10000;
```
tomusdrw: `10_000` would be more readable

util/network-devp2p/src/discovery.rs
```diff
@@ -95,7 +98,21 @@ struct FindNodeRequest {
 #[derive(Clone, Copy)]
 enum PingReason {
 	Default,
-	FromDiscoveryRequest(NodeId)
+	FromDiscoveryRequest(NodeId, NodeValidity)
```
tomusdrw: nit: comma (also below)

util/network-devp2p/src/discovery.rs
```diff
@@ -145,6 +162,8 @@ pub struct Discovery<'a> {
 	discovery_id: NodeId,
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
+	other_observed_nodes: LruCache<NodeId, (NodeEndpoint, Instant)>, // Sometimes we don't want to add nodes to the NodeTable, but still want to
```
tomusdrw: let's make this comment regular rust comment?, i.e.
```
/// Sometimes we dont'...
/// sdf
other_observed_nodes: _,
```

util/network-devp2p/src/discovery.rs
```diff
@@ -541,10 +561,22 @@ impl<'a> Discovery<'a> {
...
 
 		if let Some((node, ping_reason)) = expected_node {
-			if let PingReason::FromDiscoveryRequest(target) = ping_reason {
+			if let PingReason::FromDiscoveryRequest(target, validity) = ping_reason {
 				self.respond_with_discovery(target, &node)?;
+				// kirushik: I would prefer to probe the network id of the remote node here, and add it to the nodes list if it's on "our" net --
+				// but `on_packet` happens synchronously, so doing the full TCP handshake ceremony here is a bad idea.
+				// So instead we just LRU-caching most recently seen nodes to avoid unnecessary pinging
+				match validity {
+					NodeValidity::UnknownNode | NodeValidity::ExpiredNode(NodeCategory::Observed) => {
```
tomusdrw: I guess we should update `observed_at` and `endpoint` in case of `ValidNode`s for completeness as well, no?

util/network-devp2p/src/discovery.rs
```diff
@@ -541,10 +561,22 @@ impl<'a> Discovery<'a> {
...
+						self.other_observed_nodes.insert(node.id, (node.endpoint, Instant::now()));
+						Ok(None)
+					},
+					NodeValidity::ExpiredNode(NodeCategory::Bucket) => Ok(self.update_node(node)),
```
tomusdrw: I think we should call `update_node` but still return `Ok(None)` here.
We only want to update `node_buckets`, but not the peer table (i.e. we don't want `TableUpdates`).

tomusdrw: Note that this call may ping new nodes (it could cause amplification as well).
Should we pass the reason inside to avoid that?

kirushik: `update_node` doesn't ping the passed node directly — instead, it finds a least-recently-pinged with `select_bucket_ping` and pings that one. I don't think that this opens any space for targeted amplification.

The whole `update_node` logic can be refactored, though.

util/network-devp2p/src/discovery.rs
```diff
@@ -200,41 +230,51 @@ impl<'a> Discovery<'a> {
...
-				added_map.insert(e.id, e.clone());
-				bucket.nodes.push_front(BucketEntry::new(e));
+        match self.update_bucket_record(e) {
+            Ok(()) => None,
+            Err(BucketError::Ourselves) => None,
+            Err(BucketError::NotInTheBucket{node_entry, bucket_distance}) => Some((node_entry, bucket_distance))
+        }.map(|(node_entry, bucket_distance)| {
+			trace!(target: "discovery", "Adding a new node {:?} into our bucket {}", &node_entry, bucket_distance);
 
+            let mut added_map = HashMap::with_capacity(1);
```
tomusdrw: nit: you can call it `added` and then just `TableUpdates { added, removed: Default::default() }`

util/network-devp2p/src/discovery.rs
```diff
@@ -200,41 +230,51 @@ impl<'a> Discovery<'a> {
...
-		};
-		if let Some(node) = ping {
-			self.try_ping(node, PingReason::Default);
-		}
-		Some(TableUpdates { added: added_map, removed: HashSet::new() })
+				} else {
+					None
+				}
+			};
+			node_to_ping.map(|node| self.try_ping(node, PingReason::Default));
```
tomusdrw: nit: Imho `if let Some()` was more readable for this (I don't like using `.map` for side-effects)

util/network-devp2p/src/discovery.rs
```diff
@@ -200,41 +230,53 @@ impl<'a> Discovery<'a> {
 		}
 	}
 
-	fn update_node(&mut self, e: NodeEntry) -> Option<TableUpdates> {
-		trace!(target: "discovery", "Inserting {:?}", &e);
+	fn update_bucket_record(&mut self, e: NodeEntry) -> Result<(), BucketError> {
```
cheme: Personal preference on returning an 'Option' here and not having 'BucketError' field (If I make no mistake BucketError::Ourself got same behavior as Ok except for a debug trace). But it is really only a personal preference (does not really matter).

util/network-devp2p/src/discovery.rs
```diff
@@ -55,6 +56,8 @@ const REQUEST_BACKOFF: [Duration; 4] = [
 
 const NODE_LAST_SEEN_TIMEOUT: Duration = Duration::from_secs(24*60*60);
 
+const OBSERVED_NODES_MAX_SIZE: usize = 10_000;
```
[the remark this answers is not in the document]
kirushik: `NODE_LAST_SEEN_TIMEOUT` is 24 hours, actually.
And yes, nothing in Discovery is persistent, the only way to persist some node info is to communicate `TableUpdate` as a result of an appropriate method call.

util/network-devp2p/src/discovery.rs
```diff
@@ -565,31 +625,41 @@ impl<'a> Discovery<'a> {
...
 
-		if self.is_a_valid_known_node(&node) {
-			self.respond_with_discovery(target, &node)?;
-		} else {
+		match self.check_validity(&node) {
+			NodeValidity::Ourselves => (), // It makes no sense to respond to the discovery request from ourselves
+			NodeValidity::ValidNode(_) => self.respond_with_discovery(target, &node)?,
 			// Make sure the request source is actually there and responds to pings before actually responding
-			self.try_ping(node, PingReason::FromDiscoveryRequest(target));
+			invalidity_reason => self.try_ping(node, PingReason::FromDiscoveryRequest(target, invalidity_reason))
```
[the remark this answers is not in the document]
kirushik: It can be either `ExpiredNode` or `UnknownNode`.
I can enumerate, true — but will it actually improve readability here, given that we also need to name the value in the pattern match anyway, since we consume it for the `PingReason`

util/network-devp2p/src/discovery.rs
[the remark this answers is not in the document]
cheme: :+1: makes a lot more sense to me :)

util/network-devp2p/src/discovery.rs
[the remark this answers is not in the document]
cheme: It is only for full text search of a variant (If I want to know what is done on 'ExpiredNode' I got the info directly, but it is true that searching on the Enum name could be a better practice). But yes it does only make sense if the rule is applied everywhere.

util/network-devp2p/src/discovery.rs
kirushik: Actually it's not true, and [discovery.rs:259](https://github.com/paritytech/parity-ethereum/pull/10305/files#diff-3193558172eb0ec141a7afd53b7a1118R259) we differentiate depending on the nature of the `BucketError`.
We can, theoretically, use `Option` to indicate the specific kind of error we want to handle there — but I don't think that will be easier to read than what I have now. (It's not obvious why would some errors while updating end up with `Some` and another with `None`.)

cheme: I meant making Ok and BucketError::ourselve the same variant (it requires dropping a debug trace), but as I said it is just a matter a personal preference.

5chdn:
Waiting for @TriplEight to confirm and merge

Issue #10080, opened 2018-12-18
Implement `eth_subscribe('syncing')`


amaurymartiny:
Didn't find this in the issues, so posting it here.

_Before filing a new issue, please **provide the following information**._

- **Parity Ethereum version**: 2.3.0
- **Operating system**: MacOS
- **Installation**: homebrew
- **Fully synchronized**: yes
- **Network**: ethereum / ropsten / kovan / all
- **Restarted**: yes

**Repro**:
create a WS connection and send `{"id":2,"jsonrpc":"2.0","method":"eth_subscribe","params":["syncing"]}`

**Expected**:
works

**Actual**: 
```
{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not implemented yet. Please create an issue on Github repo."},"id":2}
```

Tbaut:
This issue got spotted thanks to Fether but it also happens on full nodes.

tomusdrw:
It was intentionally not implemented, cause the definition of "syncing" is a bit vague :) I guess we should make it work using `is_major_sync` heuristic: on every block import check `is_major_importing` and trigger the notification if the flag flips.

PR #10311, opened 2019-02-07
RPC: Implements eth_subscribe("syncing")

seunlanlege:
closes https://github.com/paritytech/parity-ethereum/issues/10080


axelchalon:
oops, looks like we've been working on the same thing https://github.com/paritytech/parity-ethereum/pull/10312
thought I had communicated this with the assignment on the issue https://github.com/paritytech/parity-ethereum/issues/10080 sorry

rpc/src/v1/types/pubsub.rs
```diff
@@ -30,6 +30,17 @@ pub enum Result {
...
 	/// Transaction hash
 	TransactionHash(H256),
+	///
+	SyncState(PubSubSyncStatus)
+}
+
+#[derive(Debug, Serialize, Eq, PartialEq, Clone)]
+#[serde(rename_all="camelCase")]
+pub struct PubSubSyncStatus {
+	//	starting_block: U64,
```
tomusdrw: uncomment or remove and add missing docs

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -267,6 +314,10 @@ impl<C: Send + Sync + 'static> EthPubSub for EthPubSubClient<C> {
 				self.heads_subscribers.write().push(subscriber);
 				return;
 			},
+			(pubsub::Kind::Syncing, None) => {
+				self.sync_subscribers.write().push(subscriber);
```
tomusdrw: Should be added to unsubscribe as well I think

parity/rpc_apis.rs
```diff
@@ -322,8 +322,9 @@ impl FullDependencies {
 				}
 				Api::EthPubSub => {
 					if !for_generic_pubsub {
-						let client =
+						let mut client =
 							EthPubSubClient::new(self.client.clone(), self.executor.clone());
+						client.add_sync_notifier(self.sync.sync_notification());
```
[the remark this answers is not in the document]
tomusdrw: Ok, it seems we have different implementations depending on `full/light` node. We could either have different constructors or maybe attempt to avoid repetition of the entire code that adds the handler and for instannce just pass a closure that takes `QueueInfo` and dispatches the notification (unless there is an abstraction for queue info already then we should just use that and make `new` generic)

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -43,19 +45,55 @@ use types::filter::Filter as EthFilter;
...
+				handler.notify_syncing(pubsub::PubSubSyncStatus { is_syncing: is_verifying || is_syncing_state });
+				Ok(())
+			})
+		)
+	}
+}
+
+impl <C: LightChainClient> EthPubSubClient<C> {
+	/// add a light chain sync notification channel to the pubsub client
+	pub fn add_light_sync_notifier(&mut self, receiver: Notification<SyncState>) {
```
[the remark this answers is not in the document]
seunlanlege: unfortunately there isn't, but I could abstract it put

rpc/src/v1/types/pubsub.rs
seunlanlege: ok, as for this, what information do you think we need to include @tomusdrw ?

ethcore/sync/src/api.rs
```diff
@@ -352,6 +369,27 @@ impl EthSync {
...
+			// spawn task that constantly updates EthSync.is_major_sync
+			let notifications = sync.write().sync_notifications();
+			let moved_client = Arc::downgrade(&params.chain);
+			let moved_is_major_syncing = is_major_syncing.clone();
+
+			params.executor.spawn(notifications.for_each(move |sync_status| {
+				if let Some(queue_info) = moved_client.upgrade().map(|client| client.queue_info()) {
+					let is_syncing_state = match sync_status { SyncState::Idle | SyncState::NewBlocks => false, _ => true };
+					let is_verifying = queue_info.unverified_queue_size + queue_info.verified_queue_size > 3;
+					moved_is_major_syncing.store(is_verifying || is_syncing_state, Ordering::SeqCst);
```
seunlanlege: is `Ordering::SeqCst` the right one here?

ethcore/sync/src/api.rs
```diff
@@ -417,6 +456,14 @@ impl SyncProvider for EthSync {
 	fn transactions_stats(&self) -> BTreeMap<H256, TransactionStats> {
 		self.eth_handler.sync.transactions_stats()
 	}
+
+	fn sync_notification(&self) -> Notification<SyncState> {
+		self.eth_handler.sync.write().sync_notifications()
+	}
+
+	fn is_major_syncing(&self) -> bool {
+		self.is_major_syncing.load(Ordering::SeqCst)
```
seunlanlege: or here?

seunlanlege:
@tomusdrw please could you take a look at this again?

ethcore/sync/src/lib.rs
```diff
@@ -36,6 +36,8 @@ extern crate parking_lot;
 extern crate rand;
 extern crate rlp;
 extern crate triehash_ethereum;
+extern crate parity_runtime;
```
tomusdrw: Sort please

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -275,6 +279,7 @@ pub struct LightSync<L: AsLightClient> {
 	client: Arc<L>,
 	rng: Mutex<OsRng>,
 	state: Mutex<SyncStateWrapper>,
+	senders: RwLock<Vec<mpsc::UnboundedSender<ChainSyncState>>>,
 	// We duplicate this state tracking to avoid deadlocks in `is_major_importing`.
 	is_idle: Mutex<bool>,
```
tomusdrw: Is this still needed?

ethcore/sync/src/light_sync/mod.rs
```diff
@@ -699,6 +725,10 @@ pub trait SyncInfo {
 
 	/// Whether major sync is underway.
 	fn is_major_importing(&self) -> bool;
```
tomusdrw: IS this still used?

rpc/src/v1/tests/helpers/sync_provider.rs
```diff
@@ -34,6 +35,8 @@ pub struct Config {
 pub struct TestSyncProvider {
 	/// Sync status.
 	pub status: RwLock<SyncStatus>,
+	/// is major importing?
+	is_importing: RwLock<bool>
```
tomusdrw: nit: missing comma

rpc/src/v1/types/pubsub.rs
tomusdrw: I think just `is_syncing` flag is fine initially, but returning some additional details would be cool as well (if easily available). please coordinate with @axelchalon or @amaurymartiny to check if anything else is required by Fether.

axelchalon:
The eth_subscribe(syncing) pubsub isn't really "required" for Fether; currently we're polling the RPC eth_syncing and it does the job. We use the `currentBlock`, `highestBlock`, `startingBlock` returned by the RPC to display the syncing percentage to the user.

I imagine if we want to switch to using the pubsub eth_subscribe(syncing) in Fether, then the pubsub should also return those values (also for the sake of consistency with the RPC eth_syncing).

However I'm not sure if we can switch to the eth_subscribe(syncing) in Fether because (at least on my PR, haven't tested on this one) the pubsub can emit values very many times / very fast; the granularity is a bit too small and I think this could bottleneck the receiving end

axelchalon:
Having eth_subscribe("syncing") return true/false depending on is_major_importing sounds OK to me!

ethcore/sync/src/light_sync/mod.rs
seunlanlege: yes

seunlanlege:
> However I'm not sure if we can switch to the eth_subscribe(syncing) in Fether because (at least on my PR, haven't tested on this one) the pubsub can emit values very many times / very fast; the granularity is a bit too small and I think this could bottleneck the receiving end

actually in this PR, events are emitted only when `SyncState` changes, which doesn't occur very frequently during syncing, infact i think you'd still need to poll `eth_syncing` to consistently show sync progress.

ethcore/sync/src/light_sync/mod.rs
seunlanlege: yes for light clients

seunlanlege:
@tomusdrw could you look at this again?

ethcore/sync/src/api.rs
```diff
@@ -142,8 +149,14 @@ pub trait SyncProvider: Send + Sync {
 	/// Get the enode if available.
 	fn enode(&self) -> Option<String>;
 
+	/// gets sync status notifications
+	fn sync_notification(&self) -> Notification<SyncState>;
+
 	/// Returns propagation count for pending transactions.
 	fn transactions_stats(&self) -> BTreeMap<H256, TransactionStats>;
+
+	/// gets
```
tomusdrw: missing docs

ethcore/sync/src/api.rs
tomusdrw: yup

ethcore/sync/src/api.rs
tomusdrw: yup

ethcore/sync/src/light_sync/mod.rs
tomusdrw: Can't light clients use the same mechanism instead? it seems that it triggers notifications as well, no?

rpc/src/v1/types/pubsub.rs
```diff
@@ -40,6 +55,7 @@ impl Serialize for Result {
 			Result::Header(ref header) => header.serialize(serializer),
 			Result::Log(ref log) => log.serialize(serializer),
 			Result::TransactionHash(ref hash) => hash.serialize(serializer),
+			Result::SyncState(ref sync) => sync.serialize(serializer)
```
tomusdrw: missing comma

rpc/src/v1/types/pubsub.rs
```diff
@@ -30,6 +30,21 @@ pub enum Result {
...
 	TransactionHash(H256),
+	/// SyncStatus
+	SyncState(PubSubSyncStatus)
+}
+
+#[derive(Debug, Serialize, Eq, PartialEq, Clone)]
+#[serde(rename_all="camelCase")]
+pub struct PubSubSyncStatus {
+	/// starting block number
+	pub	starting_block: U64,
```
tomusdrw: funky ident

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -50,6 +52,70 @@ pub struct EthPubSubClient<C> {
...
+		)
+	}
+}
+
+impl<C> EthPubSubClient<LightFetch<C>>
+	where
+		C: 'static + LightSyncProvider + LightNetworkDispatcher + ManageNetwork
+{
+	/// adds a sync notification channel to the pubsub client
+	pub fn add_light_sync_notifier(&mut self, receiver: Notification<SyncState>, sync: Arc<LightSyncInfo>) {
```
tomusdrw: I thought that we extract `QueueInfo` trait to unify the `add_light_sync_notifier` and `add_sync_notifier`, if we don't do that, then the other changes are just unnecessary.
Can we unify this completely the logic seems to be exactly the same?

If there is something that I'm missing we can also pass a closure that produces `PubSubSyncStatus` and then the mode-specific logic can be there.

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -50,6 +52,70 @@ pub struct EthPubSubClient<C> {
...
+
+				let is_syncing_state = match state { SyncState::Idle | SyncState::NewBlocks => false, _ => true };
+				let is_verifying = queue_info.unverified_queue_size + queue_info.verified_queue_size > 3;
+
+				let highest_block = U64::from(status.highest_block_number.unwrap_or(status.start_block_number));
+
+				let sync_state = pubsub::PubSubSyncStatus {
+					is_syncing: is_verifying || is_syncing_state,
+					starting_block: status.start_block_number.into(),
+					current_block: client.chain_info().best_block_number.into(),
```
tomusdrw: So this is being triggered only when `sync` changes status right? Then IMHO returning this additional data here doesn't make sense, cause it's just misleading for users - that stuff won't change at all, since we only notify about sync/notsync.

Either we need to trigger this notification more often (after every imported block for instance), but we discussed with @axelchalon that it might be overwhelming for subscribers.

I'd go for `true/false` returned from subscription and then the clients can poll at their discretion for the proper sync status. Note that we also have subscription for `newHeads` so the two subscriptions combined give you pretty good understanding what happens with the node.

Calling `queue_info` an `chain_info` from client seems pretty heavy, I'd prefer to not do it at all.
Regarding `sync.status()` - why can't we just pass all the details in the `state` already? It's just redundant call that requires some locks to be acquired while we can provide all necessary data with every event.

rpc/src/v1/types/pubsub.rs
```diff
@@ -37,14 +37,8 @@ pub enum Result {
 #[derive(Debug, Serialize, Eq, PartialEq, Clone)]
 #[serde(rename_all="camelCase")]
 pub struct PubSubSyncStatus {
```
tomusdrw: I suppose this should be a simple `true/false` instead of `{"isSyncing": true/false}`

parity/rpc_apis.rs
```diff
@@ -322,7 +324,22 @@ impl FullDependencies {
 					if !for_generic_pubsub {
 						let mut client =
 							EthPubSubClient::new(self.client.clone(), self.executor.clone());
-						client.add_sync_notifier(self.sync.sync_notification(), self.sync.clone());
+						let weak_client = Arc::downgrade(&self.client);
+
+						client.add_sync_notifier(self.sync.sync_notification(), move |state| {
+							if let Some(client) = weak_client.upgrade() {
```
tomusdrw: You can actually write this:
```
let client = weak_client.upgrade()?
...
Some(PubSubSyncStatus{ })
```

rpc/src/v1/types/pubsub.rs
seunlanlege: leaving it as a struct allows us add more fields along the line.

tomusdrw: but is not compatible with go ethereum's output. The spec says:
> the result can either be a boolean indicating that the synchronization has started (true), finished (false) or an object with various progress indicators.

But the example shows:
```
"result":{"syncing":true,"status":{"startingBlock":674427,"currentBlock":67400,"highestBlock":674432,"pulledStates":0,"knownStates":0}}
```

So either let's rename the field to `syncing` (from `isSyncing`) and add an empty `status`(?) or let's just use `true/false` directly

ethcore/sync/src/api.rs
```diff
@@ -28,6 +28,8 @@ use network::client_version::ClientVersion;
 
 use types::pruning_info::PruningInfo;
 use ethereum_types::{H256, H512, U256};
+use futures::sync::mpsc as futures_mpsc;
+use futures::{Stream};
```
niklasad1:
```suggestion
use futures::Stream;
```

ethcore/sync/src/api.rs
```diff
@@ -355,6 +372,27 @@ impl EthSync {
 			params.private_tx_handler.as_ref().cloned(),
 			priority_tasks_rx,
 		);
+
+		let is_major_syncing = Arc::new(AtomicBool::default());
```
niklasad1: I would prefer to explicitly set this to false instead of default, is harder to read IMO!

ethcore/sync/src/api.rs
```diff
@@ -355,6 +372,27 @@ impl EthSync {
...
+
+		{
+			// spawn task that constantly updates EthSync.is_major_sync
+			let notifications = sync.write().sync_notifications();
+			let moved_client = Arc::downgrade(&params.chain);
+			let moved_is_major_syncing = is_major_syncing.clone();
+
+			params.executor.spawn(notifications.for_each(move |sync_status| {
+				if let Some(queue_info) = moved_client.upgrade().map(|client| client.queue_info()) {
+					let is_syncing_state = match sync_status { SyncState::Idle | SyncState::NewBlocks => false, _ => true };
```
niklasad1: long line with pattern matching on single line, IMO this is more readable:

```rust
 let is_syncing_state = match sync_status {
        SyncState::Idle | SyncState::NewBlocks => false,
        _ => true,
 };
```

rpc/src/v1/types/pubsub.rs
```diff
@@ -30,6 +30,15 @@ pub enum Result {
 	Log(Log),
 	/// Transaction hash
 	TransactionHash(H256),
+	/// SyncStatus
+	SyncState(PubSubSyncStatus)
+}
+
+#[derive(Debug, Serialize, Eq, PartialEq, Clone)]
+#[serde(rename_all="camelCase")]
+pub struct PubSubSyncStatus {
```
niklasad1: docs please

rpc/src/v1/types/pubsub.rs
```diff
@@ -30,6 +30,15 @@ pub enum Result {
 	Log(Log),
 	/// Transaction hash
 	TransactionHash(H256),
+	/// SyncStatus
+	SyncState(PubSubSyncStatus)
+}
+
+#[derive(Debug, Serialize, Eq, PartialEq, Clone)]
+#[serde(rename_all="camelCase")]
```
niklasad1: docs please

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -306,7 +347,8 @@ impl<C: Send + Sync + 'static> EthPubSub for EthPubSubClient<C> {
 		let res = self.heads_subscribers.write().remove(&id).is_some();
 		let res2 = self.logs_subscribers.write().remove(&id).is_some();
 		let res3 = self.transactions_subscribers.write().remove(&id).is_some();
+		let res4 = self.sync_subscribers.write().remove(&id).is_some();
 
-		Ok(res || res2 || res3)
+		Ok(res || res2 || res3 || res4)
```
niklasad1: Unrelated to this PR:

hmm, are we trying to remove `the subscription` from all possible subscriptions because we don't which one it is?
Thus, if we have a some logic bug i.e, adding a subscription to the wrong one this will still return true?

rpc/src/v1/impls/eth_pubsub.rs
```diff
@@ -20,15 +20,17 @@ use std::sync::{Arc, Weak};
...
+use jsonrpc_pubsub::typed::{Sink, Subscriber};
+use jsonrpc_pubsub::SubscriptionId;
 
-use v1::helpers::{errors, limit_logs, Subscribers};
+use v1::helpers::{errors, limit_logs, Subscribers, };
 use v1::helpers::light_fetch::LightFetch;
 use v1::metadata::Metadata;
 use v1::traits::EthPubSub;
-use v1::types::{pubsub, RichHeader, Log};
+use v1::types::{pubsub, RichHeader, Log, U64};
```
niklasad1: rebase/merge to master `U64` should be imported from `ethereum_types`

soc1c:
Please rebase

aawaken:
Hello! I'm using latest parity ethereum client 2.0.5-beta. Unfortunately, I still get the error "eth_subscribe(["syncing"]): -32000: This request is not implemented yet. Please create an issue on Github repo."

My js code:

**provider.js:**
```
import Api from "@parity/api";

const provider = new Api.Provider.Ws("ws://localhost:8450");

export default provider;
```


**index.js:**
```
import React from "react";
import ReactDOM from "react-dom";

import App from "./App";
import provider from "./provider";

import light, {accounts$, balanceOf$, blockNumber$, post$, transactionCountOf$} from "@parity/light.js";
import Api from '@parity/api';

light.setProvider(provider);

const api = new Api(provider);

accounts$().subscribe(accounts => {
    console.log(accounts);
});
```

**package.json:**
```
{
  "name": "new",
  "version": "1.0.0",
  "description": "",
  "keywords": [],
  "main": "src/index.js",
  "dependencies": {
    "@parity/api": "^5.1.3",
    "@parity/light.js": "5.1.3",
    "ajv": "^6.10.0",
    "react": "16.4.2",
    "react-dom": "16.4.2",
    "react-scripts": "^3.0.0",
    "rxjs": "^6.2.2",
    "typescript": "^3.4.5"
  },
  "devDependencies": {},
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test --env=jsdom",
    "eject": "react-scripts eject"
  },
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  }
}

```

Missing something?

ordian:
@aawaken 
> Missing something?

This PR was merged into master (currently 2.6) and wasn't backported to 2.5 (usually, we only backport bug fixes).

aawaken:
> @aawaken
> 
> > Missing something?
> 
> This PR was merged into master (currently 2.6) and wasn't backported to 2.5 (usually, we only backport bug fixes).

Thanks @ordian Any suggestions when 2.6 will be released?

Best

PR #10314, opened 2019-02-08
Update to latest mem-db, hash-db and trie-db.

cheme:
This PR updates parity-ethereum to use parity-trie crate, this is part of [paritytech/trie#5](https://github.com/paritytech/trie/issues/5) and follows https://github.com/paritytech/parity-common/pull/106 .

A few point of attention in this PR:
* HashDBRef new trait makes the triedb construction a bit awkward (ptr lifetime), there is certainly a way to make it look better, and possibly a change in the trait but I would put that out of scope of this pr.
* memorydb construction through utility function (I put it in journaldb but it is probably not the best place).
* [paritytech/trie#2](https://github.com/paritytech/trie/pull/2) makes 'keys' function a requirement of parity-eth only so I added some traits (in journaldb again as it is mainly use for it), I did not add the parametrics types as it is only related to eth at this point.
* triehash crate may be replace by trieroot but for now I also put that out of scope of this PR.


ethcore/light/src/cht.rs
```diff
@@ -73,7 +74,8 @@ impl<DB: HashDB<KeccakHasher, DBValue>> CHT<DB> {
 		if block_to_cht_number(num) != Some(self.number) { return Ok(None) }
 
 		let mut recorder = Recorder::with_depth(from_level);
-		let t = TrieDB::new(&self.db, &self.root)?;
+    let db: &HashDB<_,_> = &self.db;
```
ngotchac: Indentation is off.

util/journaldb/src/traits.rs
```diff
@@ -21,13 +21,28 @@ use std::sync::Arc;
...
-use hashdb::HashDB;
+use hash_db::{HashDB, AsHashDB};
 use keccak_hasher::KeccakHasher;
 use kvdb::{self, DBTransaction, DBValue};
+use std::collections::HashMap;
+
+
+/// expose keys of a hashDB for debugging or tests (slow).
+pub trait KeyedHashDB: HashDB<KeccakHasher, DBValue> {
+  /// Primarily use for tests, highly inefficient.
```
ordian: indentation
```suggestion
	/// Primarily use for tests, highly inefficient.
```

ordian: If `keys()` is only used for tests, maybe put it behind `#[cfg(test)]`?

cheme: It is actually use on non purely test context (eg with evmbin for debug display of trie state), it could still be put behind a feature.

ordian: I couldn't find any uses of `keys()` besides tests, could you point me to the code?

ethcore/src/snapshot/tests/state.rs
```diff
@@ -24,23 +24,23 @@ use types::basic_account::BasicAccount;
 use snapshot::account;
 use snapshot::{chunk_state, Error as SnapshotError, Progress, StateRebuilder, SNAPSHOT_SUBPARTS};
 use snapshot::io::{PackedReader, PackedWriter, SnapshotReader, SnapshotWriter};
-use super::helpers::{compare_dbs, StateProducer};
+use super::helpers::{StateProducer};
```
sorpaas:
```suggestion
use super::helpers::StateProducer;
```

util/journaldb/src/traits.rs
cheme: I may be mistaking (maybe it is simply triedb iterator that is used, I will test the change in a bit).

cheme: So  you are right :+1: . I can build without it and it seems fine, but I am having a hard time making the test run, I believe it is related to the fact that we use it in other crates (having conditional trait method seems ok for `commit_batch` or `inject_batch`, but those are only use in journaldb). The same thing happens when you depend on a crate and expect the test function to be here (thus the 'test_helper' crates).

ordian: Let's merge the PR then. I'll open an issue to make `KeyedHashDB` test-only.

PR #10315, opened 2019-02-08
Bundle protocol and packet_id together in chain sync

elferdo:
At different places in the chain sync code we have to specify which packet of which protocol we want to send (or we received) over the network. Having to manage both in parallel opens up the door for potential incosistencies. 

Besides, types of packages are identified as numeric constants with no guarantee that two different constants will not clash with the same value.

This PR proposes a solution to bundle protocol and packet id together.


ethcore/sync/src/chain/syncpacketid.rs
```diff
@@ -0,0 +1,141 @@
...
+use api::{ETH_PROTOCOL, WARP_SYNC_PROTOCOL_ID};
+use network::{PacketId, ProtocolId};
+
+/// An enum that defines all known packet ids in the context of
+/// synchronization and provides a mechanism to convert from
+/// packet ids (of type PacketId or u8) directly read from the network
+/// to enum variants. This implicitly provides a mechanism to
+/// check whether a given packet id is known, and to prevent
+/// packet id clashes when defining new ids.
+enum_from_primitive! {
```
tomusdrw: I'm not sure if the macro is worth additional (seems like not very well maintained) dependency.
Can we implement `from_usize/from_u8` manually?

Alternatively could we use constants instead of enum?
```
const STATUS_PACKET = SyncPacketId(0x00, ETH_PROTOCOL);
```

ethcore/sync/src/chain/supplier.rs
```diff
@@ -25,103 +26,101 @@ use types::ids::BlockId;
 
 use sync_io::SyncIo;
 
+use super::syncpacketid::{PacketInfo, SyncPacketId, SyncPacketId::*};
```
tomusdrw: Avoid glob imports please, or at least keep them as local as possible (I'd import in the function that does `match`)

ethcore/sync/src/chain/handler.rs
```diff
@@ -67,36 +59,40 @@ impl SyncHandler {
...
+				BlockHeadersPacket => SyncHandler::on_peer_block_headers(sync, io, peer, &rlp),
+				BlockBodiesPacket => SyncHandler::on_peer_block_bodies(sync, io, peer, &rlp),
+				ReceiptsPacket => SyncHandler::on_peer_block_receipts(sync, io, peer, &rlp),
+				NewBlockPacket => SyncHandler::on_peer_new_block(sync, io, peer, &rlp),
+				NewBlockHashesPacket => SyncHandler::on_peer_new_hashes(sync, io, peer, &rlp),
+				SnapshotManifestPacket => SyncHandler::on_snapshot_manifest(sync, io, peer, &rlp),
+				SnapshotDataPacket => SyncHandler::on_snapshot_data(sync, io, peer, &rlp),
+				PrivateTransactionPacket => SyncHandler::on_private_transaction(sync, io, peer, &rlp),
+				SignedPrivateTransactionPacket => SyncHandler::on_signed_private_transaction(sync, io, peer, &rlp),
+				_ => {
```
tomusdrw: Is that even possible now? This should rather go to `else` branch of `from_u8`

ethcore/sync/src/api.rs
```diff
@@ -39,8 +39,8 @@ use std::net::{SocketAddr, AddrParseError};
 use std::str::FromStr;
 use parking_lot::{RwLock, Mutex};
 use chain::{ETH_PROTOCOL_VERSION_63, ETH_PROTOCOL_VERSION_62,
-	PAR_PROTOCOL_VERSION_1, PAR_PROTOCOL_VERSION_2, PAR_PROTOCOL_VERSION_3,
-	PRIVATE_TRANSACTION_PACKET, SIGNED_PRIVATE_TRANSACTION_PACKET};
+	PAR_PROTOCOL_VERSION_1, PAR_PROTOCOL_VERSION_2, PAR_PROTOCOL_VERSION_3};
+use chain::syncpacketid::SyncPacketId::*;
```
tomusdrw: Same comment for glob imports (here I'd make it explicit, we only need two packets)

ethcore/sync/src/chain/syncpacketid.rs
elferdo: I considered using constants and found one argument against them: we have to manually check that no two constants repeat the same value, or they don't end up claiming the same packet_id, that's why I thought an enum would be a good idea.

We can indeed implement the conversion manually, though I found it repetitious: adding one new Id would mean adding it to the enum and to the conversion methods: three places to touch.

Agreed, the package seems a bit outdated, even if it has a large number of downloads. If we drop the dependency, for cleanliness I would go for constans.

ethcore/sync/src/chain/supplier.rs
elferdo: Ok, I will change this.

ethcore/sync/src/api.rs
elferdo: You're right, I'll make the two explicit.

ethcore/sync/src/chain/handler.rs
elferdo: This function handles only incoming packets, but the enum defines outgoing packets as well. The match wouldn't be exhaustive without this. Maybe the debug message could be more explicit: _unknown incoming packet_.

ethcore/sync/src/chain/syncpacketid.rs
tomusdrw: Well now we need to maintain a map of `variant -> protocol_id` manually.
Perhaps we can have a custom (proc) macro for this to cater all cases?

ethcore/sync/src/chain/handler.rs
tomusdrw: I see, we still should print a message in case we get `None` (completely unknown `packet_id`) though.

tomusdrw: Ugh sorry: just noticed we do.

ethcore/sync/src/chain/syncpacketid.rs
elferdo: Yeah, that mapping is also boring, I'll have a look at macros.

ethcore/sync/src/chain/mod.rs
```diff
@@ -88,6 +88,7 @@
 //! All other messages are ignored.
 
 mod handler;
+pub mod syncpacketid;
```
ngotchac: I'm not a fan of the name, why not `sync_packets`?

elferdo: Mmmm, no particular reason, I'll change it.

elferdo:
Anything left to address here? I have a little macro proposal like @tomusdrw suggested but would rather make it another PR.

5chdn:
@tomusdrw @ngotchac is this good to go?

tomusdrw:
yeah, it's `mustntgrumble` :) If we work on removing the dep on `enum-primitive` then let's merge that and addressed that separately as @elferdo suggested.

elferdo:
Yes, the dependency is gone in my next PR :)

ngotchac:
@elferdo Do you already have the PR ready? Otherwise it's fine, but just create an issue for it so we don't forget :)

PR #10324, opened 2019-02-11
Add message to IO errors

Pzixel:
When I migrated from 2.0.X to 2.1.X I faced an error "Something went wrong". Unfortunately, the original error gets rewritten with useless message. This PR makes original error preserved instead.


Pzixel:
[clabot:check]

Pzixel:
image

There is 52 more issues in other modules. I think they should be addressed as well. There is nothing worse than useless "oops" error message.

PR #10330, opened 2019-02-12
fix(add helper for timestamp overflows)

niklasad1:


ethcore/src/verification/verification.rs
```diff
@@ -40,6 +40,35 @@ use types::{BlockNumber, header::Header};
 use types::transaction::SignedTransaction;
 use verification::queue::kind::blocks::Unverified;
 
+
+/// FIXME: @niklasad1 - remove this when and use `SystemTime::checked_add`
+/// when https://github.com/rust-lang/rust/issues/55940 is stabilized
+pub fn timestamp_checked_add(sys: SystemTime, d2: Duration) -> Result<SystemTime, BlockError> {
```
tomusdrw: Maybe let's rather check if `Duration` is within reasonable bounds instead? The logic below seems pretty complicated.

cheme:
I think we can also get this function used in util/network-devp2p/src/node_table.rs and whisper/src/message.rs

PR #10335, opened 2019-02-12
change docker image based on debian instead of ubuntu due to the chan…

gabreal:
…ges of the build container


5chdn:
Please rebase

gabreal:
@TriplEight  why hotfix?
sorry, but that merge just right after this pull request made a rebase impossible.

PR #10337, opened 2019-02-12
Fix Beta Backports

HCastano:
I can't push to the `a5-beta-backports` branch directly, so my fixes gotta get PR'd in.


ordian:
Do we really need to backport #9782?

HCastano:
@ordian I pulled it in for the `BlockChainReset` trait, but I you're right the whole thing might not need to go in

ordian:
The issue is resolved now. Closing.

PR #10340, opened 2019-02-13
Macros to define subprotocol packet ids in chain sync

elferdo:
This PR elaborates on comments by @tomusdrw in https://github.com/paritytech/parity-ethereum/pull/10315

Synchronization of a node's blockchain with other nodes happens through an exchange of packets over the network. In parity we support a number of different synchronization protocols that are transmitted and received as devp2p subprotocols. 

In devp2p, subprotocol and the particular message within it are identified with a single number instead of two, one for protocol and one for message. Care must be taken to ensure that each type of package within each subprotocol gets a globally unique id that does not collide with other subprotocols.

This PR is a first proposal of a derive-macro that can be applied to an enum and attributes for the enum variants. Code would possibly look like the following:
```
#[derive(SyncPackets)]
enum Packet {
   #[eth] StatusMessage = 0x01,
...
   #[par] GetSnapshotManifestPacket = 0x11,
}
```
Benefits of the approach:
* The enum takes care itself that ids are globally unique
* The derive-macro implements conversion method `from_u8` so that ids read from the network can be directly converted to the appropriate identifier.
* The provided `from_u8` method returns an `Option<Packet>`
* The derive-macro automatically provides functionality to query id and protocol from an instance of `Packet`
* Protocol and Id are specified in one single compact line

Drawbacks:
* Protocols are defined as attributes of a derive-macro. Defining new protocols involves changing the macro (this should not happen frequently, though)
* Each message needs to be individually tagged with its protocol attribute. Whole blocks would be more ergonomic.
* The derive-macro implements the trait `PacketInfo` and its method `protocol()`. The constants returned in the method must be already defined outside the scope of this macro.

The first version of the PR is lacking error checks.


elferdo:
Sorry, had some issues with git. I need to create a new PR.

PR #10341, opened 2019-02-13
Macros to define subprotocol packet ids in chain sync 

elferdo:
This PR elaborates on comments by @tomusdrw in https://github.com/paritytech/parity-ethereum/pull/10315                                                                                                                                                
                                                                                                                                                                                                                                                       
Synchronization of a node's blockchain with other nodes happens through an exchange of packets over the network. In parity we support a number of different synchronization protocols that are transmitted and received as devp2p subprotocols.        
                                                                                                                                                                                                                                                       
In devp2p, subprotocol and the particular message within it are identified with a single number instead of two, one for protocol and one for message. Care must be taken to ensure that each type of package within each subprotocol gets a globally unique id that does not collide with other subprotocols.                                                                                                                                                                                                
                                                                                                                                                                                                                                                       
This PR is a first proposal of a derive-macro that can be applied to an enum and attributes for the enum variants. Code would possibly look like the following:                                                                                        
```
#[derive(SyncPackets)]
enum Packet {
   #[protocol(ETH_PROTOCOL)] StatusMessage = 0x01,
...
   #[protocol(WARP_SYNC_PROTOCOL)] GetSnapshotManifestPacket = 0x11,
}
```                                                                                                                                                                                                                                                    
Benefits of the approach:                                                                                                                                                                                                                              
* The enum takes care itself that ids are globally unique                                                                                                                                                                                              
* The derive-macro implements conversion method `from_u8` so that ids read from the network can be directly converted to the appropriate identifier.                                                                                                   
* The provided `from_u8` method returns an `Option<Packet>`                                                                                                                                                                                            
* The derive-macro automatically provides functionality to query id and protocol from an instance of `Packet`                                                                                                                                          
* Protocol and Id are specified in one single compact line  
* Further attributes may be easily added, for example a version number.        
* Syntax errors are reported as compile errors.                                                                                                                                                                                   
                                                                                                                                                                                                                                                       
Drawbacks:                                                                                                                                                                                                                                                                                                                                      
* Each message needs to be individually tagged with its protocol attribute. Whole blocks would be more ergonomic.


ngotchac:
Couldn't we have something like:
```rust
#[derive(EnumMap)]
enum Packet {
   #[packet = ETH_PACKET]
  StatusMessage = 0x01,
...
  #[packet = WARP_SYNC_PACKET] 
  GetSnapshotManifestPacket = 0x11,
}
```
and `Packet::StatusMessage.packet()` would return `ETH_PACKET`?

elferdo:
You mean that the attribute `packet` itself defines a new method? So for instance `#[version = v63]` would automatically define `Packet::StatusMessage.version()`?

ngotchac:
That would be the idea yeah, if possible

elferdo:
I think the syntax would need to look somewhat different, because attributes for derive-macros must be predefined, but I think we can achieve something along those lines. I'll look into it.

For some reason I prefer closed solutions to open-ended ones, or at least try not to generalize too early, but I can't see any big downside to your idea right now. I would see, for example, that instead of `#[eth]` and `#[par]` we would provide a fixed set of attributes (protocol and version) for example: `#[protocol WARP_SYNC]` or `#[version v63]`. This would at least decouple the macro from the definition of the protocols.

util/syncpacketid_derive/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "syncpacketid_derive"
```
niklasad1: can you make this crate `rust 2018` and remove `extern crate`?

util/syncpacketid_derive/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+/// The SyncPackets derive-macro will provide an enum with this attribute:
+///
+/// * With a method "from_u8" which will optionally convert a u8 value to
+///   one of the variants or None if the value is unknown.
+///
+/// * With an implementation of a trait PacketInfo to get the packet id and
+///   the protocol from instances of the enum.
+#[proc_macro_derive(SyncPackets, attributes(eth, par))]
+pub fn sync_packets(input: TokenStream) -> TokenStream {
+	let ast = syn::parse(input).unwrap();
```
niklasad1: In general I would prefer to use `expect` instead of  `unwrap` to give a somewhat descriptive compile-error.

However, in this case I don't know really how when `parse` fails but `TokenStream parse fail; do something` would certainly help

elferdo: Absolutely, as it is right now this is rather a proof of concept. I wanted to first gather some feedback to see if we like this direction, and then I'll add error handling.

util/syncpacketid_derive/Cargo.toml
elferdo: Sure, I'll do that.

util/syncpacketid_derive/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+fn impl_sync_packets(ast: &syn::DeriveInput) -> proc_macro2::TokenStream {
+	let body = match ast.data {
+		syn::Data::Enum(ref e) => e,
+		_ => panic!("#[derive(SyncPackets)] is only defined for enums."),
+	};
+
+	let enum_name = &ast.ident;
+
+	let eths: Vec<_> = body.variants.iter()
+		.filter(|v| v.attrs[0].path.is_ident("eth"))
```
niklasad1:
```suggestion
		.filter(|v| v.attrs.get(0).expect("attribute is missing; annotate your enum patterns with #[eth] or #[par]).path.is_ident("eth"))
```

util/syncpacketid_derive/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+		.map(|v| &v.ident).collect();
+
+	let pars: Vec<_> = body.variants.iter()
+		.filter(|v| v.attrs[0].path.is_ident("par"))
```
niklasad1:
```suggestion
		.filter(|v| v.attrs.get(0).expect("attribute is missing; annotate your enum patterns with #[eth] or #[par]).path.is_ident("par")))
```

util/syncpacketid_derive/src/lib.rs
elferdo: Ok, it wasn't such a big deal, sorry XD

util/syncpacketid_derive/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+		.map(|v| &v.ident).collect();
+
+	let idents: Vec<_> = body.variants.iter().map(|v| &v.ident).collect();
+	let values: Vec<_> = body.variants.iter().map(|v| v.discriminant.clone().unwrap().1).collect();
```
niklasad1:
```suggestion
	let values: Vec<_> = body.variants.iter().map(|v| v.discriminant.clone().expect("enum pattern is not discriminant; should have assigned unique value such #[eth] Foo = 1).1.collect();
```

elferdo: Changed manually in my local checkout.

ngotchac:
@elferdo Can you update the PR description to reflect your recent changes?

ethcore/sync/src/chain/sync_packet.rs
```diff
@@ -22,96 +22,44 @@
 //! to convert to/from the packet id values transmitted over the
 //! wire.
 
-use api::{ETH_PROTOCOL, WARP_SYNC_PROTOCOL_ID};
-use network::{PacketId, ProtocolId};
+use syncpacketid_derive::SyncPackets;
```
ngotchac: Do we really need it to be a separate crate if it's specific to SyncPackets?

elferdo:
@ngotchac Oh, sure, sorry, I will update it right now.

ethcore/sync/src/chain/sync_packet.rs
elferdo: As I understand from [the documentation](https://doc.rust-lang.org/reference/procedural-macros.html#derive-mode-macros):

> Procedural macros must be defined in a crate with the crate type of proc-macro.

Besides (taken from the compiler):

> error: `proc-macro` crate types cannot export any items other than functions tagged with `#[proc_macro_derive]`

I placed it alongside `rlp-derive`, which seemed to be the right place.

ngotchac: Right, sorry, I wasn't aware

ngotchac: Maybe just rename it to `syncpacket` then

elferdo: No worries, I was also surprised that they had to be in their own crate.

I renamed it.

util/syncpacket/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "syncpacket"
+version = "0.1.0"
+authors = ["Fernando Herrero Carrón <fernando@parity.io>"]
```
tomusdrw: Please update to `Parity Technologies`

util/syncpacket/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+// https://blog.rust-lang.org/2018/12/21/Procedural-Macros-in-Rust-2018.html
+extern crate proc_macro;
+
+use proc_macro2::Ident;
+use quote::quote;
+use syn::{Data, DeriveInput, Meta, MetaList, NestedMeta};
+
+fn parse_protocol_arguments(args: MetaList) -> Ident {
+	if args.nested.len() != 1 {
+		panic!("protocol attribute should have exactly one argument");
```
tomusdrw: Can we turn those into compiler errors? See: https://github.com/paritytech/jsonrpc/pull/382/files

util/syncpacket/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
tomusdrw: Could you add some tests for the crate?

util/syncpacket/Cargo.toml
elferdo: Oops, sure!

util/syncpacket/src/lib.rs
elferdo: Sure, I didn't know about that, thanks for the pointer.

util/syncpacket/Cargo.toml
elferdo: I just pushed some tests in the case when syntax is right, then everything should work as expected. I will have a look at `syn` to see how we can test wrong syntax.

dvdplm:
@elferdo It'd be good to have some module level docs here with an example or two. Pretty please?

elferdo:
> @elferdo It'd be good to have some module level docs here with an example or two. Pretty please?

Sure! I'll address your two issues asap.

elferdo:
> Seems like a bug yeah?

The macro was designed with one specific use case, namely that of putting all the p2p protocol IDs into one place. One of the restrictions these IDs must meet is that they are unique, and by allowing only one enum the language itself will take care of that.
> 
> Code:
> 
> ```
> #[derive(SyncPackets, Clone, Copy, Debug, PartialEq)]
> enum Packets {
> 	#[protocol(MY_PROTOCOL0)] Packet0 = 0x00,
> 	#[protocol(MY_PROTOCOL1)] Packet1 = 0x01,
> }
> 
> #[derive(SyncPackets, Clone, Copy, Debug, PartialEq)]
> enum Bovine {
> 	#[protocol(BEEF)] Bull = 0x00,
> 	#[protocol(BEEF)] Cow = 0x01,
> }
> ```

I guess you don't want:
```
#[derive(SyncPackets, Clone, Copy, Debug, PartialEq)]
enum Packets {
	#[protocol(MY_PROTOCOL0)] Packet0 = 0x00,
	#[protocol(MY_PROTOCOL1)] Packet1 = 0x01,
	#[protocol(BEEF)] Bull = 0x02,
	#[protocol(BEEF)] Cow = 0x03,
}
```
right?

Where is the other enum used?

The solution to the problem is quite straightforward, I'm just wondering what the use case is.

dvdplm:
> Where is the other enum used?
> 
> The solution to the problem is quite straightforward, I'm just wondering what the use case is.

There isn't one, just me playing around with the code while reviewing it and noticing a surprising error message: I think the expectation is that a macro should work even when applied several times?

If it becomes terribly complicated to fix, I'm fine with some good docs explaining how to go about applying the macro to several `enums` in a project.

dvdplm:
Just to be clear: I think this is a terrific improvement and I'm prepared to merge it even if imperfect, just as long as it's properly doc'd. :)

elferdo:
> > Where is the other enum used?
> > The solution to the problem is quite straightforward, I'm just wondering what the use case is.
> 
> There isn't one, just me playing around with the code while reviewing it and noticing a surprising error message: I think the expectation is that a macro should work even when applied several times?
> 
> If it becomes terribly complicated to fix, I'm fine with some good docs explaining how to go about applying the macro to several `enums` in a project.

No, not really, I already have it in place. My worries are that it gets abused and then all of a sudden we end up having, again, uncontrolled IDs. Like it should be obvious how to properly update IDs, or add a new one, and it should be difficult to make it wrong.

Let's make it reusable, just in case. The code as it is should be obviously easy to maintain, and if someone wants it s/he can have it.

dvdplm:
> namely that of putting all the p2p protocol IDs into one place

So this is used in `libp2p`? Should we perhaps extract it to `parity-common` then so we can share the code?

dvdplm:
ping @elferdo

util/syncpacket/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+		},
+		// Quoted string
+		a @ _ => return Err(syn::Error::new_spanned(a, "protocol argument must be an unquoted identifier"))
+	}
+}
+
+/// Helper function to parse arguments to the protocol attribute.
+/// Syntax should be #[protocol(P)] PacketName = 0xNN,
+fn parse_protocol_attribute(input: &Attribute) -> Result<Ident> {
+	let argument = match input.parse_meta() {
```
dvdplm: Can't this be just `let argument = input.parse_meta()?;`?

util/syncpacket/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+		Ok(arg) => arg,
+		Err(err) => return Err(err)
+	};
+
+	// Arguments to invocation attributes are delivered as a list
+	match argument {
+		Meta::Word(_) => Err(syn::Error::new_spanned(input, "protocol attribute without argument")),
+		Meta::List(args) => parse_protocol_arguments(&args).map(|ok| ok.clone()),
```
dvdplm: Put the happy path first?

dvdplm: Should this check that there's a single argument in the list as well?

dvdplm:
@elferdo ping

util/syncpacket/src/lib.rs
elferdo: Yep, good one!

util/syncpacket/src/lib.rs
elferdo: This is checked in `parse_protocol_arguments`, do you think here would be a better place?

debris:
> Sure! I'll address your two issues asap.

@elferdo please address the issues and reopen the pr :)

Issue #9771, opened 2018-10-17
Keep snapcraft up to date


fubuloubu:
Looks like the last time someone updated Snapcraft was 22 days ago. These are the releases for each channel:
- stable: 2.0.6
- candidate: 2.1.1
- beta: 2.1.1
- edge: 2.2.0-5a1dc3e-20180925

From what I can tell, there are more recent versions of all of these we should be using. Are snap installs still supported?

5chdn:
Hi Bryant, unfortunately, we dropped support for Snapcraft. Please use the one-line installer or convince your linux distro maintainers to add parity to the repositories, like archlinux did two years ago :)

fubuloubu:
@elopio any help finding the right maintainer ^ ?

elopio:
I could maintain it, if the team thinks it's a good idea.

elopio:
I've requested the transfer here:
https://forum.snapcraft.io/t/transfer-parity-snap/8203
They might contact you to confirm.

PR #10343, opened 2019-02-13
snap: prefix version and populate candidate channel

5chdn:
- snap: prefix version with v*
  - compare docker tags: https://hub.docker.com/r/parity/parity/tags
- snap: populate candidate releases with beta snaps to avoid stale channel 
  - compare candidate channel: https://snapcraft.io/parity

### DIFF
--- a/scripts/gitlab/publish-snap.sh
+++ b/scripts/gitlab/publish-snap.sh
@@ -4,7 +4,7 @@ set -e # fail on any error
 set -u # treat unset variables as error
 
 # some necromancy:
-# gsub(/"/, "", $2) deletes "qoutes" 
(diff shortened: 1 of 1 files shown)


General-Beck:
@5chdn and how we will solve this question now?
https://github.com/paritytech/parity-ethereum/issues/9771#issuecomment-433696614

5chdn:
> @5chdn and how we will solve this question now?
> [#9771 (comment)](https://github.com/paritytech/parity-ethereum/issues/9771#issuecomment-433696614)

Leo is maintainer but we still have publishing access. So, our gitlab script directly creates a release on snapcraft.

PR #10347, opened 2019-02-13
[CI] test(parity-clib example)

niklasad1:
just for running it in the CI


soc1c:
And now? :smiley:

niklasad1:
I think the `atomic counter` should fix `UB` lemme prepare PR but is probably too slow and not stable enough for running in the CI

PR #10356, opened 2019-02-14
enable lto, panic = "abort" in release build

ordian:
* shaves 11 MB off binary size (from 44MB to 33MB for `RUSTFLAGS=" -C link-arg=-s" cargo build --release`, -8MB thanks to LTO)
* run time should be faster, but I haven't measured

Last time we disabled `panic = "abort"` in #8983 (cc @sorpaas) due to failing build.


ordian:
> This is awesome but it increases the build-time significantly (roughly 2x) for me 

Don't we `build --release` only on releasable branches (stable and beta)? 
Also, IIUC, `profile.release` shouldn't affect `cargo test --release` (should be covered by `profile.bench`).

ordian:
#### Regarding LTO:
`./test.sh` on CI on this branch:
> 7326.09user 2098.40system 14:40.37elapsed 1070%CPU  5.8 GB mem

`./test.sh` on CI for other PRs:
> 4002.26user 536.6.9system 11:39.47elapsed 648%CPU 4.3 GB mem
> 4006.64user 1955.58system 15:45.66elapsed 630%CPU 5.4 GB mem [link](https://gitlab.parity.io/parity/parity-ethereum/-/jobs/122294)
> 4752.03user 447.98system 18:36.40elapsed 465%CPU [link](https://gitlab.parity.io/parity/parity-ethereum/-/jobs/122420)

Looks like the overall time for building + running tests hasn't changed much. Max RSS does not seem to be too large. 
Alternatively, we can backport this PR to beta and stable without merging into master.

#### Now on to `panic = "abort"`:
@arkpar argues [here](https://github.com/paritytech/parity-ethereum/pull/8983#issuecomment-406297225) that:
> Database has to be resistant to power loss anyway.

But now, in addition to `KeyValueDB`, we have [blooms-db](https://github.com/paritytech/parity-ethereum/blob/fa570f297ebc27fddab1bbfdaa89d48761a81ef8/parity/db/rocksdb/mod.rs#L39-L43), which is not power loss resistant. So that's a valid concern, I guess.

arkpar:
When I last checked (around 2017) `panic="abort"` did have a substantial effect of performance. Block import was around 10% faster.
The fact that there are three databases now makes atomic updates much harder. It would have been much better to use column families instead, as these can be written with a single atomic transaction. 
Even with `panic="unwind"`the three databases my go out of sync because of the power loss or process termination. There needs to be some additional code that would detect this by e.g. comparing best block number or hash stored in all of the databases. In case of mismatch it would either have to rewind the main database to an older block, or rebuild the missing traces.

ordian:
Ok, it seems like we all agree on the LTO change, if someone from devops can confirm it won't kill our machines (@gabreal ?).

> Anyway, panic and abort in the middle of an import won't lead to **main** database corruption because transaction atomicity is guaranteed by rocksdb, and as far as I know all of the block import is done in a single transaction.

This matches what @debris said [here](https://github.com/paritytech/parity-ethereum/pull/8712/files#r195708234):
> Existence of blooms is not consensus critical. If one of the writes fail (or partially fail), the particular block will be reimported on the next launch of parity and the database will overwrite corrupted data

If we agree on `panic = "abort"`, should #8999 be reverted?
If we don't, I can revert that change, or, alternatively, port it only to beta, to see if it would lead to more bug reports. Any thoughts?

gabreal:
As far as I can see, there should be no problem of memory anymore, but be aware that the test job run time increases approx about 7min (for this job there is only one sample). @TriplEight might also like to know.

5chdn:
> Don't we `build --release` only on releasable branches

tests are run in release mode afaik.

TriplEight:
Thanks for mentioning @gabreal .
> tests are run in release mode afaik.
[Yes](https://github.com/paritytech/parity-ethereum/blob/83bcb819da65c4dcc3ce2d89758fec014781c150/test.sh#L53). And would be really good if we could avoid it to make this features work only for the release builds.

Currently merging this will slow down our already slow test pipelines.

Please read about` LTO`  [here](https://vfoley.xyz/rust-compile-speed-tips), really a good read, and there are some benchmarks of different features.
```
Compilers build and optimize compilation units individually. LTO is a mechanism for taking those individually-optimized object files and finding more opportunities for optimization when they are considered as a group. This is great if you want to have the fastest possible program, but the price you pay are higher compilation times. In Rust, the compilation unit is the crate; if the heavy computations are all done in your crate, then maybe LTO isn’t necessary.
```

Regarding `panic` handling, `panic = abort` just does not include unwind libraries to the release, that's how it becomes lighter. 
```
By removing support for unwinding, you'll get smaller binaries. You will lose the ability to catch panics. 
```
That way we are refusing bug stack traces from users.

ordian:
> Currently merging this will slow down our already slow test pipelines.

That's why I suggested to only backport it to stable and beta, w/o merging into master.

> That way we are refusing bug stack traces from users.

catch panics != stack traces, we'll see stack traces

TriplEight:
That way it makes sense @ordian .

TriplEight:
Just had a talk with Pierre. He suggested that we should change the 
```
[profile.test]
opt-level=0 # or 1
panic=unwind
lto=false
```
that way we can get rid of `--release` for all the tests, in general it should make tests faster.

ordian:
> in general it should make tests faster.

It would make tests faster to compile but slower to run, I think the reason we're using `cargo test --release` in the first place is because the overall (compile + run) time is faster than the one w/o `--release`, but it would be nice to see if it can be reduced with `opt-level=1`.

TriplEight:
I'll try this `opt-level=1` locally, will report

gabreal:
> I think the reason we're using `cargo test --release` in the first place is because the overall (compile + run) time is faster than the one w/o `--release`,

In fact I changed that a while ago to have the tests perform as close as possible as they are used in the wild.

ordian:
I've split the change into two commits, the second commit also removes the `std::process::abort()` call from the panic hook.

My suggestion is to backport the first commit to stable and both to beta.

TriplEight:
@gabreal @ordian a comparsion report: https://drive.google.com/open?id=159CBQGiKOPz9JBfda_n9bLFNF10qIif3Fyz5DoyWBjk

dvdplm:
To my ears it sounds like we want `lto = true` for release builds and `lto = false` for tests (but otherwise keep the same optimisation level as release for tests).

Panic handling seems more contentious and it sounds like we need more testing and hand-wringing on that one.

@ordian can you PR those two changes separately? The diff here is a little confusing with all the Clique code in.

ordian:
> To my ears it sounds like we want lto = true for release builds and lto = false for tests (but otherwise keep the same optimisation level as release for tests).

I'm not sure how to enable it this way, since we run tests with `--release` flag if I'm not mistaken. So I suggested to not this PR into master, but rather backport commits into stable and beta. What do you think?

> The diff here is a little confusing with all the Clique code in.

Yeah, that's due to force rebase on master.

dvdplm:
> since we run tests with `--release` flag

Forgive me, but I'm slow. Can't we stop running tests that way and instead switch to:

```
[profile.test]
# proper optimization levels to match --release here
lto = false
```
…i.e. switch CI to run `cargo test` instead of `cargo test --release`? I must be missing something!

TriplEight:
@dvdplm yes, this thread is hard.
>Can't we stop running tests that way and instead switch to

AFAIU we can do that.

TriplEight:
@dvdplm, @ordian  something went wrong, I suggest falling back. 
Pipeline with this change: https://gitlab.parity.io/parity/parity-ethereum/pipelines/39408/builds
Any other pipeline: https://gitlab.parity.io/parity/parity-ethereum/pipelines/39408/builds
Check out the time it took to test.

dvdplm:
@TriplEight the links are the same there, not sure where to look. Can fix? :)

TriplEight:
@dvdplm I edited in the GitHub UI

PR #10359, opened 2019-02-14
chore(bump parity-daemonize): require rust >= 1.31

niklasad1:
Bump `parity-daemonize` to make `parity` compile with rustc 1.31.0 or later.

Added a comment about a `clone()` that was necessary to make it compile however it shouldn't be needed AFAIK!

### DIFF
--- a/parity/main.rs
+++ b/parity/main.rs
@@ -39,13 +39,15 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::{process, env};
+
 use ansi_term::Colour;
 use ctrlc::CtrlC;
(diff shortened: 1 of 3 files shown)


parity/main.rs
```diff
@@ -195,7 +197,9 @@ fn main_direct(force_can_restart: bool) -> i32 {
 		conf.args.arg_chain = spec_override;
 	}
 
-	let handle = if let Some(ref pid) = conf.args.arg_daemon_pid_file {
+	// FIXME: `pid_file` shouldn't need to cloned here
+	// see: `https://github.com/paritytech/parity-daemonize/pull/13` for more info
```
ordian: It shouldn't be a big deal, since only only do this once at startup and this string the issually small.

PR #10362, opened 2019-02-15
Add fields to `memzero`'s Cargo.toml

HCastano:
Adds some fields that are required before publishing a crate to crates.io.

### DIFF
--- a/util/memzero/Cargo.toml
+++ b/util/memzero/Cargo.toml
@@ -2,7 +2,7 @@
 name = "memzero"
 version = "0.1.0"
 description = "A wrapper for zero-ing out memory when dropped"
-license = "MIT"
+license = "GPL-3.0"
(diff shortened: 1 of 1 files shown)


util/memzero/Cargo.toml
```diff
@@ -1,4 +1,9 @@
 [package]
 name = "memzero"
 version = "0.1.0"
+description = "A wrapper for zero-ing out memory when dropped"
```
tomusdrw: Can you also add `docs = "https://docs.rs/crate/memzero"`?

util/memzero/Cargo.toml
```diff
@@ -1,4 +1,9 @@
...
+license = "MIT"
```
tomusdrw: Please use `GPL-3`

Issue #10344, opened 2019-02-13
RPC randomly hangs for no reason


Pzixel:
- **Parity Ethereum version**: 2.2.9
- **Operating system**: Linux
- **Installation**: docker
- **Fully synchronized**: yes
- **Network**: private (POA, [chain.json specification](https://gist.github.com/Pzixel/21e796d78bb6084b86e575b079732e38))
- **Restarted**: yes

For some reason after several hours of intensive block generations parity hangs for no reason. 

Friday night I ran my batcher, that just call some contract in batches (50 items in batch).
Here are my logs:

```
2019-02-08 23:35:49.024 +03:00 [Information] "51,38%" done. Elapsed = 09:49:30.2509107.
2019-02-08 23:35:58.938 +03:00 [Information] "51,40%" done. Elapsed = 09:49:40.1648875.
2019-02-08 23:36:08.794 +03:00 [Information] "51,41%" done. Elapsed = 09:49:50.0207078.
2019-02-08 23:36:18.833 +03:00 [Information] "51,43%" done. Elapsed = 09:50:00.0597905.
2019-02-08 23:36:28.819 +03:00 [Information] "51,44%" done. Elapsed = 09:50:10.0459107.
2019-02-08 23:36:38.562 +03:00 [Information] "51,46%" done. Elapsed = 09:50:19.7880563.
2019-02-08 23:36:48.819 +03:00 [Information] "51,47%" done. Elapsed = 09:50:30.0456974.
2019-02-08 23:36:58.739 +03:00 [Information] "51,49%" done. Elapsed = 09:50:39.9653453.
2019-02-08 23:37:08.770 +03:00 [Information] "51,50%" done. Elapsed = 09:50:49.9966959.
2019-02-08 23:37:18.834 +03:00 [Information] "51,52%" done. Elapsed = 09:51:00.0605550.
2019-02-08 23:37:29.025 +03:00 [Information] "51,53%" done. Elapsed = 09:51:10.2512043.
2019-02-08 23:37:38.840 +03:00 [Information] "51,55%" done. Elapsed = 09:51:20.0662434.
2019-02-08 23:37:48.780 +03:00 [Information] "51,56%" done. Elapsed = 09:51:30.0061202.
2019-02-08 23:37:58.858 +03:00 [Information] "51,57%" done. Elapsed = 09:51:40.0840885.
[...]

PR #10375, opened 2019-02-18
tx pool: always accept local transactions

ascjones:
...unless sender and nonce are the same, in which case the the normal gas price comparison applies.

rel: #10344

### DIFF
--- a/miner/src/pool/scoring.rs
+++ b/miner/src/pool/scoring.rs
@@ -123,16 +123,17 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
 	}
 
 	fn should_replace(&self, old: &P, new: &P) -> scoring::Choice {
-		if old.priority().is_local() && new.priority().is_local() {
-			// accept local transactions over the limit
-			scoring::Choice::InsertNew
-		} else if old.sender() == new.sender() {
+		let both_local = old.priority().is_local() && new.priority().is_local();
+		if old.sender() == new.sender() {
 			// prefer earliest transaction
 			match new.nonce().cmp(&old.nonce()) {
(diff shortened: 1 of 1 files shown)


miner/src/pool/scoring.rs
```diff
@@ -123,16 +123,16 @@ impl<P> txpool::Scoring<P> for NonceAndGasPrice where P: ScoredTransaction + txp
 	}
 
 	fn should_replace(&self, old: &P, new: &P) -> scoring::Choice {
-		if old.sender() == new.sender() {
+		if old.priority().is_local() && new.priority().is_local() {
+			// accept local transactions over the limit
+			scoring::Choice::InsertNew
```
tomusdrw: Sorry for confusion, but now that I think about it we should probably handle the case when senders AND nonces are equal. In such case we need to replace transaction instead of accepting both of them

tomusdrw: Would be worth checking if we hit that particular branch with `should_replace` anyway, cause I think it might be handled separately by calling `choose` directly.

tomusdrw:
Looks good, marking as `patch` cause I consider this a bug (i.e. sending some transactions locally without high-enough limits can cause some transactions to be lost hence rendering the rest to be stuck forever).

Pzixel:
Why not just check for `both_local ` as first branch? Wouldn't require any changes in nonce matching branch then.

tomusdrw:
@Pzixel the case where both sender and nonce is equal would be handled incorrectly. There can only be one transaction with given `(sender, nonce)` pair, so we can't return `InsertNew` there but rather need to choose either one.

tomusdrw:
Closes #10344

PR #10383, opened 2019-02-19
fix(extract `timestamp_checked_add` as lib)

### LINKED ISSUE #10406: Replace `timestamp_checked_add` with `SystemTime::checked_add` when available in rustc
We have [our own](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/verification/verification.rs#L51-#L60) implementation of [SystemTime::checked_add](https://doc.rust-lang.org/std/time/struct.SystemTime.html#method.checked_add) because it is `unstable` until `Rust 1.34.0`

Thus, when `Rust 1.34.0` we should remove that implementation alternatively provide both via [feature flag timestamp_checked_add](https://doc.rust-lang.org/unstable-book/library-features/time-checked-add.html?highlight=time_che#time_checked_addl) to make `parity` compile with older versions than `Rust 1.34`

I have a pending PR https://github.com/paritytech/parity-ethereum/pull/10383 in regard to this which may redundant and it could be closed if we remove our helpers completely but there are some additional `checked_add` operations that could be extracted to a separate if that is the case

niklasad1:
Closes #10406 

* Split out `timestamp_checked_add` into a library and fix some suspect usage of `SystemTime`
* Conditional compilation only to use `our time library` when not "time_checked_add" is available in the standard library

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -42,11 +42,13 @@ use itertools::{self, Itertools};
 use rlp::{encode, Decodable, DecoderError, Encodable, RlpStream, Rlp};
 use ethereum_types::{H256, H520, Address, U128, U256};
 use parking_lot::{Mutex, RwLock};
-use time_utils::CheckedSystemTime;
 use types::BlockNumber;
 use types::header::{Header, ExtendedHeader};
 use types::ancestry_action::AncestryAction;
 use unexpected::{Mismatch, OutOfBounds};
+
+#[cfg(not(feature = "time_checked_add"))]
+use time_utils::CheckedSystemTime;
 
 mod finality;
 
@@ -577,10 +579,9 @@ fn verify_timestamp(step: &Step, header_step: u64) -> Result<(), BlockError> {
 			trace!(target: "engine", "verify_timestamp: block too early");
 
 			let now = SystemTime::now();
-			let found = CheckedSystemTime::checked_add(now, Duration::from_secs(oob.found))
-				.ok_or(BlockError::TimestampOverflow)?;
-			let max = oob.max.and_then(|m| CheckedSystemTime::checked_add(now, Duration::from_secs(m)));
-			let min = oob.max.and_then(|m| CheckedSystemTime::checked_add(now, Duration::from_secs(m)));
+			let found = now.checked_add(Duration::from_secs(oob.found)).ok_or(BlockError::TimestampOverflow)?;
+			let max = oob.max.and_then(|m| now.checked_add(Duration::from_secs(m)));
+			let min = oob.min.and_then(|m| now.checked_add(Duration::from_secs(m)));
 
 			let new_oob = OutOfBounds { min, max, found };
 
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -42,6 +42,7 @@ use itertools::{self, Itertools};
 use rlp::{encode, Decodable, DecoderError, Encodable, RlpStream, Rlp};
 use ethereum_types::{H256, H520, Address, U128, U256};
 use parking_lot::{Mutex, RwLock};
+use time_utils::CheckedSystemTime;
 use types::BlockNumber;
 use types::header::{Header, ExtendedHeader};
 use types::ancestry_action::AncestryAction;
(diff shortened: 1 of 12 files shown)


util/time-utils/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+	/// Returns `Some<SystemTime>` when the result is successful and `None` when it is not
+	fn checked_sub(self, _d: Duration) -> Option<SystemTime>;
+}
+
+impl CheckedSystemTime for SystemTime {
+	fn checked_add(self, dur: Duration) -> Option<SystemTime> {
+		let this_dur = self.duration_since(UNIX_EPOCH).ok()?;
+		let total_time = this_dur.checked_add(dur)?;
+
+		if total_time.as_secs() <= i32::max_value() as u64 {
```
seunlanlege: so basically we can only add a maximum of 68 years with this method, just curious why this limit.

niklasad1: Right, good questions Seun :P 

It is because SystemTime represents seconds as `i32, i64, u64 or Duration` (traditional OS's represent those as time_t i32 or i64).

This basically prevents the overflow in easy-way (but limits all platforms with limit of 2038) i.e, not have to rely on any complicated conditional compilation.

But this should be removed/deprecated when `Rust 1.33` is available (2019-02-28)

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -47,6 +47,9 @@ use types::header::{Header, ExtendedHeader};
 use types::ancestry_action::AncestryAction;
 use unexpected::{Mismatch, OutOfBounds};
 
+#[cfg(not(feature = "time_checked_add"))]
```
sorpaas: Is this referring to rust feature or crate feature? I think for rust feature, we may need to define it to be used here:

```
#![cfg_attr(feature = "time_checked_add", feature(time_checked_add)]
```

niklasad1: Yeah, it is a rust feature.

It is a good idea I will change it will definitely be much cleaner :+1:

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -626,7 +637,7 @@ trait AsMillis {
 
 impl AsMillis for Duration {
 	fn as_millis(&self) -> u64 {
-		self.as_secs()*1_000 + (self.subsec_nanos()/1_000_000) as u64
+		self.as_secs() * 1_000 + (self.subsec_nanos() / 1_000_000) as u64
```
ordian: [Duration::as_millis](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_millis) is available since rust 1.33, though it returns `u128`. Do we want to remove this trait? (Maybe out of scope of this PR)

niklasad1: I prefer to keep it out this PR. If people agree with `this` approach it should be easy just to move the `AsMillis` trait to the `time_util library`. However, the implementation above truncates `Duration` into u64 which we don't want in the library.

Seems to be ok for the use case in `authority_round`

ethcore/src/lib.rs
```diff
@@ -15,6 +15,7 @@
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
 #![warn(missing_docs, unused_extern_crates)]
+#![cfg_attr(feature = "time_checked_add", feature(time_checked_add))]
```
cheme: I did not find a use for this line (just tested stable and nightly with changing `checked_add` name and commenting this line: nightly pass, stable fail because of changed name).

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -574,8 +577,15 @@ fn verify_timestamp(step: &Step, header_step: u64) -> Result<(), BlockError> {
 			// NOTE This error might be returned only in early stage of verification (Stage 1).
 			// Returning it further won't recover the sync process.
 			trace!(target: "engine", "verify_timestamp: block too early");
-			let oob = oob.map(|n| SystemTime::now() + Duration::from_secs(n));
-			Err(BlockError::TemporarilyInvalid(oob).into())
+
+			let now = SystemTime::now();
+			let found = now.checked_add(Duration::from_secs(oob.found)).ok_or(BlockError::TimestampOverflow)?;
+			let max = oob.max.and_then(|m| now.checked_add(Duration::from_secs(m)));
```
cheme: I wonder if we should return error on failing `checked_add` here (same for min but max would have failed before). I do not know the authority round logic enough to say.
At the same time if it goes over max value we got a natural bound check from `checked_add` so it is probably not an issue.

niklasad1: I think we shouldn't because you will get an error message something like: 

`OutOfBounds { min: None, max: None, found: val }`

which should be more informative than `TimestampOverflow`

However I think it would be better represent those `OutOfBounds` as Duration instead because they are easier to reason about i.e, will always represent seconds as u64!

Thoughts?

ethcore/src/lib.rs
niklasad1: @sorpaas  grumble: https://github.com/paritytech/parity-ethereum/pull/10383#discussion_r265515435

ethcore/src/engines/authority_round/mod.rs
cheme: Maybe `saturating_add` would fit better, my fear was that we were removing a upper bound by making this change but thinking twice it goes directly into error so it does not matter and you are right it only depends upon what we want to display, so it is ok as it is.

niklasad1: Right, I get your point saturating_add is a good idea but is very messy for SystemTime

ethcore/src/lib.rs
niklasad1: I can remove those if @sorpaas agree

PR #10391, opened 2019-02-21
set correct nonce when node creates a public transaction wrapping a private transaction

### LINKED ISSUE #10390: Incorrect nonce when node creates a public transaction wrapping a private transaction
- **Parity Ethereum version**: master
- **Operating system**: MacOS
- **Fully synchronized**: Not applicable
- **Network**: private
- **Restarted**: yes

I'm testing Parity Ethereum’s private transactions feature.

I have a private contract successfully deployed. I want to make a state change for this private contract. I create a private transaction, signed by an External Owned Account (which has nonce = 0x1) by sending a signed tx to `private_sendTransaction` JSON-RPC endpoint.

Once validators signatures collected, the node creates a public transaction wrapping my private transaction. This public transaction is signed by `signer_account` configurated on this node. This `signer_account` has nonce = 0x0, but the public transaction crafted by the node has nonce = 0x1 caused by this piece of code:

https://github.com/paritytech/parity-ethereum/blob/master/ethcore/private-tx/src/lib.rs#L365-L371.

This implies that the public transaction is never executed.

I think this is not correct, node should use the current nonce value for `signer_account` and not the nonce from the original transaction?

adetante:
Set correct nonce value when creating a public transaction wrapping a private one.

fix #10390

### DIFF
--- a/ethcore/private-tx/src/lib.rs
+++ b/ethcore/private-tx/src/lib.rs
@@ -362,18 +362,20 @@ impl Provider {
 			signatures.push(signed_tx.signature());
 			let rsv: Vec<Signature> = signatures.into_iter().map(|sign| sign.into_electrum().into()).collect();
 			// Create public transaction
+			let signer_account = self.signer_account.ok_or_else(|| ErrorKind::SignerAccountNotSet)?;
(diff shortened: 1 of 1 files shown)


adetante:
[clabot:check]

grbIzl:
Looks good to me. Again, @adetante  thanks for finding and fixing!

PR #10399, opened 2019-02-21
revert some changes, could be buggy

TriplEight:


.gitlab-ci.yml
```diff
@@ -99,7 +99,7 @@ build-windows:
 publish-docker:
   stage:                           publish
   only:                            *releaseable_branches
-  <<:                              *no_git
+  # <<:                              *no_git
```
soc1c: please remove it or not

.gitlab-ci.yml
```diff
@@ -140,23 +140,6 @@ publish-onchain:
   tags:
     - linux-docker
 
-# configures aws for fast uploads/syncs
-.s3-before-script:                &s3-before-script
```
soc1c: why is this being removed?

.gitlab-ci.yml
General-Beck: remove

.gitlab-ci.yml
General-Beck: moved to docker parity/awscli @TriplEight ?

gabreal: it's [here](https://github.com/paritytech/scripts/blob/master/awscli/config) now

PR #10405, opened 2019-02-22
perf(ethcore): `micro-opt`

niklasad1:
Mostly fixes that changes `eagerly eval` to `lazy eval` shouldn't

### DIFF
--- a/ethcore/src/error.rs
+++ b/ethcore/src/error.rs
@@ -33,7 +33,7 @@ use engines::EngineError;
 
 pub use executed::{ExecutionError, CallError};
 
-#[derive(Debug, PartialEq, Clone, Copy, Eq)]
+#[derive(Debug, PartialEq, Clone, Eq)]
 /// Errors concerning block processing.
 pub enum BlockError {
 	/// Block has too many uncles.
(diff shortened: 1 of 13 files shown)


ethcore/src/state/mod.rs
```diff
@@ -83,7 +83,7 @@ pub enum ProvedExecution {
 	/// The transaction failed, but not due to a bad proof.
 	Failed(ExecutionError),
 	/// The transaction successfully completd with the given proof.
-	Complete(Executed),
+	Complete(Box<Executed>),
```
niklasad1: Executed is 368 bytes is quite big and must be passed via stack (I assume)

ethcore/src/error.rs
```diff
@@ -84,7 +84,7 @@ pub enum BlockError {
 	/// Timestamp header field is too far in future.
 	TemporarilyInvalid(OutOfBounds<SystemTime>),
 	/// Log bloom header field is invalid.
-	InvalidLogBloom(Mismatch<Bloom>),
+	InvalidLogBloom(Box<Mismatch<Bloom>>),
```
niklasad1: Bloom: 256 bytes big

folsen:
@niklasad1 is it possible to add some microbenchmarks to this as well to see what the change is? I feel like we could use more benchmarks across the board

niklasad1:
@folsen yeah I think so. I can try to add that but I think the only noticeable difference is the `changed enum`

PR #10430, opened 2019-02-27
More Backports for Stable 2.3.5

HCastano:


parity/main.rs
```diff
@@ -184,6 +186,11 @@ fn main_direct(force_can_restart: bool) -> i32 {
 		parity_ethereum::Configuration::parse_cli(&args).unwrap_or_else(|e| e.exit())
 	};
 
+	let logger = setup_log(&conf.logger_config()).unwrap_or_else(|e| {
```
HCastano: There's a warning here about this being unused. This is because it only started getting used in #10007 which isn't in stable yet. Not sure if I should leave the warning, or not backport this PR in yet

PR #10432, opened 2019-02-27
Remove calls to heapsize

cheme:
This pr remove calls to heapsize crate (the crate will still be in the dependency tree from other crates).
It replace heapsize calls by malloc_size call which are basically the same but defined into 'parity-util-mem' (we got additional option and a feature for only running memory estimation (not calling the allocator)).
This **cannot be merged** as it is and require first that https://github.com/paritytech/parity-common/pull/93 get merged and published and then that https://github.com/paritytech/trie/pull/14 get merged and published. See the patched crates in Cargo.toml.

Yet it can be reviewed.

Also it is currently set (in Cargo.toml) to use jemalloc and crate size estimation (that may not be the target).


dvdplm:
Note: https://github.com/paritytech/parity-common/pull/93 is now merged.

ethcore/light/src/cache.rs
```diff
@@ -157,12 +157,14 @@ impl Cache {
...
 	pub fn mem_used(&self) -> usize {
-		self.heap_size_of_children()
+		self.malloc_size_of()
 	}
 }
 
-impl HeapSizeOf for Cache {
-	fn heap_size_of_children(&self) -> usize {
+
+// TODO this is fast method: should feature gate an exhaustive implementation
```
dvdplm: Please file an issue and add a ref to this comment.

ethcore/light/src/lib.rs
```diff
@@ -63,7 +63,8 @@ extern crate parity_bytes as bytes;
 extern crate ethereum_types;
 extern crate ethcore;
 extern crate hash_db;
-extern crate heapsize;
+extern crate parity_util_mem as malloc_size_of;
+extern crate parity_util_mem as mem;
```
dvdplm: I think this is confusing, it will require devs to come here to look. I think I'd prefer explicit paths here.

ethcore/src/engines/validator_set/test.rs
```diff
@@ -19,20 +19,23 @@
 use std::str::FromStr;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering as AtomicOrdering};
+use mem::{MallocSizeOf};
```
dvdplm:
```suggestion
use mem::MallocSizeOf;
```

ethcore/src/lib.rs
```diff
@@ -97,6 +96,8 @@ extern crate patricia_trie_ethereum as ethtrie;
 extern crate rand;
 extern crate rayon;
 extern crate rlp;
+extern crate parity_util_mem as mem;
+extern crate parity_util_mem as malloc_size_of;
```
dvdplm: idem as above

util/journaldb/src/lib.rs
```diff
@@ -16,7 +16,8 @@
 
 //! `JournalDB` interface and implementation.
 
-extern crate heapsize;
+extern crate parity_util_mem as mem;
+extern crate parity_util_mem as malloc_size_of;
```
dvdplm: idem as above

dvdplm:
Needs rebase.

ethcore/evm/src/interpreter/shared_cache.rs
```diff
@@ -28,8 +28,8 @@ const DEFAULT_CACHE_SIZE: usize = 4 * 1024 * 1024;
 // stub for a HeapSizeOf implementation.
```
dvdplm: This comment is interesting. Is the `Bit` type there just to impl `HeapSizeOf`? If yes, do we still need it? 

Either way the comment needs updating. ;)

ethcore/light/src/client/header_chain.rs
```diff
@@ -202,14 +202,21 @@ pub enum HardcodedSync {
 	Deny,
 }
 
+#[derive(MallocSizeOf)]
 /// Header chain. See module docs for more details.
 pub struct HeaderChain {
+	#[ignore_malloc_size_of = "perf"]
```
dvdplm: If one doesn't know what the value assigned to `ignore_malloc_size_of` does (I didn't) then `"perf"` isn't a great help. Maybe `"ignored for perf reasons"` (or add a comment before the struct to explain: "skipping expensive allocation info here")

dvdplm:
@cheme I've run this branch on mainnet over the weekend. Looks good. Very odd that CI failed after last commit, seems completely unrelated. Launched a new run.

dvdplm:
So things left to do here:

- [ ] remove patches in `Cargo.toml` once all referenced crates are published
- [ ] set `parity-ethereum` to use `jemalloc`
- [ ] use estimation wherever possible
- [ ] await needed changes to `triehash` before merging

cheme:
I am going to address them (malloc_size_of alias when it is required by the derive macro). I thought I did address them but that was not the case :(

dvdplm:
@cheme merge whenever you're ready.

ethcore/sync/src/lib.rs
```diff
@@ -53,8 +53,10 @@ extern crate enum_primitive;
 extern crate macros;
 #[macro_use]
 extern crate log;
+extern crate parity_util_mem;
+extern crate parity_util_mem as mem;
```
ordian: these two lines should be removed, right?

cheme: alas, they are required by the derive macro (the second one, the first one is used in the imports), just double checked.

Issue #9339, opened 2018-08-11
Write example of parity-clib in python


tomaka:
See #8604 

This is a "dummy" issue so that a bounty can be put on it.

niklasad1:
The C library has been removed

Issue #9338, opened 2018-08-11
Write example for parity-clib in Go


tomaka:
See #8604 

This is a "dummy" issue so that a bounty can be put on it.

nlordell:
I would like to work on this.

niklasad1:
@nlordell 

Sorry, we previously removed in the `C library` so this issue is not possible to work on anymore.

PR #10443, opened 2019-03-02
Implement Python wrapper for parity-clib and example

mjkoo:
Follows `jni` example, uses `pyo3` to build an extension module exposing the low-level API and includes a Python module which exports a more pythonic interface.

Ref: https://github.com/paritytech/parity-ethereum/issues/9339 and https://github.com/paritytech/parity-ethereum/issues/8604

Interested in also tackling https://github.com/paritytech/parity-ethereum/issues/9338 (Golang example) if this is well-received.


parity-clib/Cargo.toml
```diff
@@ -14,6 +14,7 @@ futures = "0.1.6"
 jni = { version = "0.11", optional = true }
 panic_hook = { path = "../util/panic-hook" }
 parity-ethereum = { path = "../", default-features = false }
+pyo3 = { version = "0.6.0-alpha.4", optional = true, features = ["extension-module"] }
```
niklasad1: `pyo3` will require the nightly toolchain and currently we only use stable features

parity-clib/src/python.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Handle to a Parity config created with parity_config_from_cli
+#[pyclass]
+struct ConfigHandle {
+	inner: *mut c_void,
+}
+
+impl Drop for ConfigHandle {
+	fn drop(&mut self) {
+		if self.inner != ptr::null_mut() {
```
niklasad1:
```suggestion
		if !self.inner.is_null() {
```

parity-clib/src/python.rs
```diff
@@ -0,0 +1,219 @@
...
+	// Ensure we don't double-free config
+	config.inner = ptr::null_mut();
+	ret
+}
+
+/// Perform a RPC query against a running Parity instance, invoking callback on completion
+#[pyfunction]
+fn rpc_query(
+	_py: Python<'_>,
+	parity: &ParityHandle,
```
niklasad1: how is it ensured that this function can't end up with `undefined behavior`?

parity-clib/src/python.rs
```diff
@@ -0,0 +1,219 @@
...
+/// Subscribe to a websocket event, invoking callback when event received
+#[pyfunction]
+fn subscribe_ws(
+	_py: Python<'_>,
+	parity: &ParityHandle,
+	rpc: &str,
+	callback: PyObject,
+) -> PyResult<SubscriptionHandle> {
+	if parity.inner == ptr::null_mut() {
+		return Err(ValueError::py_err("Parity execution completed"));
```
niklasad1: hrm, this means that `ParityClient` has been shutdown or it failed during init right?

Can you change the error message something like: "Attempt to subscribe to WebSocket when ParityClient is not running"?

parity-clib/src/python.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Subscribe to a websocket event, invoking callback when event received
+#[pyfunction]
+fn subscribe_ws(
+	_py: Python<'_>,
+	parity: &ParityHandle,
+	rpc: &str,
+	callback: PyObject,
+) -> PyResult<SubscriptionHandle> {
+	if parity.inner == ptr::null_mut() {
```
niklasad1:
```suggestion
	if parity.inner.is_null() {
```

parity-clib/python/parity/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        """Configure and start Parity
+
+        :param options: Command line arguments to pass to Parity
+        :param logger_mode: Logger options to pass to Parity
+        :param logger_file: File to log to
+        """
+        config = _parity.config_from_cli(options)
+        self.handle = _parity.build(config, logger_mode, logger_file)
+
+    def rpc_query_cb(self, query, cb, timeout_ms=1000):
```
niklasad1: Personally I would rename this to  `rpc_query_async`

parity-clib/python/parity/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        """Perform a RPC query, return immediately, cb will be invoked with the result asynchronously
+
+        :param query: Query to perform
+        :param cb: Callback to invoke with results
+        :param timeout_ms: Timeout in milliseconds
+        """
+        _parity.rpc_query(self.handle, query, timeout_ms, cb)
+
+    def rpc_query(self, query, timeout_ms=1000):
```
niklasad1: `rpc_query_sync`

parity-clib/python/parity/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        """Perform a RPC query and return the result synchronously
+
+        :param query: Query to perform
+        :param timeout_ms: Timeout in milliseconds
+        :return: Result of the rpc call
+        """
+        cb = _CallbackGenerator()
+        self.rpc_query_cb(query, cb, timeout_ms)
+        return next(cb)
```
niklasad1: This will call the iterator of the Callback class?

parity-clib/src/python.rs
```diff
@@ -0,0 +1,219 @@
...
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+use std::mem;
+use std::os::raw::c_void;
+use std::ptr;
+use std::sync::Arc;
+
+use parity_ethereum::RunningClient;
+use pyo3::exceptions::{RuntimeError, ValueError};
+use pyo3::prelude::*;
```
niklasad1: Please use explicit imports

parity-clib/python/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.so
```
niklasad1: hrm, move this to https://github.com/paritytech/parity-ethereum/blob/master/.gitignore

parity-clib/python/parity/__init__.py
mjkoo: of the `_CallbackGenerator` class, yes

parity-clib/python/.gitignore
mjkoo: Already in there, removed from this file. Strange though, pretty sure I added this for a reason :thinking:

parity-clib/src/python.rs
mjkoo: Need some help here, what guarantees should I be ensuring? Is the concern about the behavior of the callback?

Depending on response java code may also need addressing.

mjkoo:
@niklasad1 thank you for review, partially addressed comments, had some follow up questions asked above.

Also experienced ICE when compiling with latest nightly, had to revert to nightly-2019-02-07 which is what I've been testing with.

Ack re: pyo3 and stable features, especially since this is triggering bug with current nightly toolchain. Will follow stabilization efforts over there, seems fairly close.

parity-clib/src/python.rs
niklasad1: Sorry, if I my comment wasn't clear enough but what I meant was how is it ensured that this function is memory safe (aliasing, dangling pointer or similar)?

My only concern is that this function is not marked as unsafe, can you explain why?

niklasad1:
FYI, the ICE is fixed now https://github.com/rust-lang/rust/issues/58634

debris:
I would love to get it in. What is the status of this pr? @niklasad1 do you still have some suggestions regarding this pr?

debris:
I would love to get it in. What is the status of this pr? @niklasad1 do you still have some suggestions regarding this pr?

niklasad1:
@debris it is currently blocked on https://github.com/PyO3/pyo3/issues/210 but I agree would be nice to get it in. We shall maybe look how hard it to fix that issue IMHO.

niklasad1:
We have removed `parity-clib`, thus closing this one.

PR #10451, opened 2019-03-05
CI aws git checkout

General-Beck:


.gitlab-ci.yml
```diff
@@ -209,6 +196,7 @@ build-android:
   image:                           parity/rust-android:gitlab-ci
   variables:
     CARGO_TARGET:                  armv7-linux-androideabi
+  dependencies:
```
TriplEight: This line prevents artifacts from other jobs to be downloaded before this job.
It makes sense to do the same for the other build jobs.

TriplEight: later

gabreal: would suggest to use
```
dependencies: []
```
to make it consistent with the `cache: {}` statement.

scripts/gitlab/publish-aws.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e # fail on any error
+set -u # treat unset variables as error
+
+echo "__________Push binaries to AWS S3____________"
+case "${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}" in
+    (beta|stable|nightly)
+      export BUCKET=releases.parity.io/ethereum;
```
gabreal: no need to export that variable, it is only used within this script.

.gitlab-ci.yml
TriplEight: indeed

PR #10452, opened 2019-03-06
Tests parallelized

TriplEight:
The example of new regular pipeline: https://gitlab.parity.io/parity/parity-ethereum/pipelines/32255
# what's better
- first stage now runs ~ 2 min
- `cargo test` now runs ~ 16 min
- whole pipeline takes ~ 19 min
- now it's good to launch docker cache

# what's changed
- removed android-build from regular pipeline
- split `test.sh` and `test-all.sh` by this whole lot of jobs
- some stuff was optimized and simplified
- `nproc` now counts threads for `cargo test` and `cpp-test`

# reflections
- it's really needed to overwrite the `cpp-test`
- makes sense to think how to parallelize `test.sh`, maybe mark unit tests


.gitlab-ci.yml
```diff
@@ -29,22 +29,56 @@ variables:
...
 
-test-linux:
+test-audit:
   stage:                           test
-  variables:
-    RUN_TESTS:                     all
   script:
-    - scripts/gitlab/test-all.sh
+    - set -e
+    - set -u
```
gabreal: Since `cargo audit` is not a script these commands will have no effect.

.gitlab-ci.yml
```diff
@@ -29,22 +29,56 @@ variables:
...
+    - linux-docker
+  allow_failure:                   true
+
+test-cpp:
+  stage:                           build
+  dependencies:
+    - test-rust-changes
+  script:
+    - ./scripts/gitlab/test-cpp.sh
+    - sccache -s
```
gabreal: what does this do and why is it run only after the building step?

.gitlab-ci.yml
```diff
@@ -29,22 +29,56 @@ variables:
...
-    - cargo audit
+    - ./scripts/gitlab/cargo_check.sh
+    - sccache -s
+  tags:
+    - linux-docker
+
+test-rust-changes:
+  stage:                           test
+  script:
+    - ./scripts/gitlab/rust_changes.sh
```
gabreal: How can it end the pipeline with success when no relevant files have changed?

.gitlab-ci.yml
```diff
@@ -29,22 +29,56 @@ variables:
...
+    - sccache -s
+  tags:
+    - linux-docker
+
+test-linux:
+  stage:                           build
+  dependencies:
+    - test-rust-changes
+  script:
+    - ./scripts/gitlab/test.sh
```
gabreal: it's just one command, so why not put it here directly. or if there should be a script with every job, then why not name them in a way that it is obvious which belong together?

.gitlab-ci.yml
```diff
@@ -188,6 +222,7 @@ publish-docs:
 
 build-android:
   stage:                           optional
+  only:                            *releaseable_branches
```
gabreal: If you do this, then untested android builds will block releases.

scripts/gitlab/test-cpp.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+echo "________Running test-cpp.sh________"
+set -e # fail on any error
+set -u # treat unset variables as error
+#use nproc `linux only
+THREADS=$(nproc)
```
gabreal: are you aware that this yields 128 on ci2?

.gitlab-ci.yml
TriplEight: indeed, thanks

.gitlab-ci.yml
TriplEight: that's for debug purposes here, when we'll introduce `sccache` via docker volumes it will help.
The command itself shows what's going on with caches.

.gitlab-ci.yml
General-Beck: MB switch stage to test?

TriplEight: ok, will place it to `test` phase

scripts/gitlab/test-cpp.sh
TriplEight: is it bad?

.gitlab-ci.yml
TriplEight: not sure yet, up to test it

TriplEight: but it's not supposed to mark the pipeline as successful, previously it stopped tests from being run

.gitlab-ci.yml
TriplEight: and in here, `test-cpp.sh` does not use cargo, I understand your point.
In my opinion it should be complied with cargo though.

.gitlab-ci.yml
TriplEight: makes sense to put the command directly here. And will take a look at the naming as well.

.gitlab-ci.yml
gabreal: well, you should not be able to merge if the test are not successful.

scripts/gitlab/test-cpp.sh
gabreal: I'm not sure, but in general I think that overcommiting nodes hurts performance. this would be the case when submitting several jobs. Also I think this can simply be tested before merging into parity-ethereum master branch.

accounts/ethkey/src/brain.rs
```diff
@@ -18,7 +18,7 @@ use keccak::Keccak256;
 use super::{KeyPair, Generator, Secret};
 use parity_wordlist;
 
-/// Simple brainwallet.
+/// Simple brainwallet.s
```
soc1c: Huch? :D

scripts/gitlab/test-cpp.sh
TriplEight: this will be a subject to test later

TriplEight:
for now we are going to skip checking changed files, we are going to check them Gitlab way. But before that caching should be introduced.

accounts/ethkey/src/brain.rs
TriplEight: xD
that was for testing if pipeline will run with Rust files changed

TriplEight:
ready for the final review.

.gitlab-ci.yml
```diff
@@ -24,39 +24,81 @@ variables:
...
+  before_script:
+    - sccache -s
+  after_script:
     - sccache -s
   tags:
     - linux-docker
 
-test-audit:
+
+cargo-check 0 3:
```
gabreal: nice trick :)

.gitlab-ci.yml
```diff
@@ -24,39 +24,81 @@ variables:
...
-    - set -e
-    - set -u
     - cargo audit
   tags:
     - linux-docker
 
+validate-chainspecs:
+  stage:                           test
+  script:
+    - sccache -s
```
gabreal: could reuse the .cargo_check template (maybe different name)

.gitlab-ci.yml
```diff
@@ -24,39 +24,81 @@ variables:
...
+    - ./scripts/gitlab/validate-chainspecs.sh
+    - sccache -s
+  tags:
+    - linux-docker
+
+test-cpp:
+  stage:                           build
+  dependencies:                    []
+  script:
+    - sccache -s
```
gabreal: see above

.gitlab-ci.yml
```diff
@@ -24,39 +24,81 @@ variables:
...
+    - ./scripts/gitlab/test-cpp.sh
+    - sccache -s
+  tags:
+    - linux-docker
+
+test-linux:
+  stage:                           build
+  dependencies:                    []
+  script:
+    - sccache -s
```
gabreal: see above

gabreal: and below :)

.gitlab-ci.yml
```diff
@@ -203,13 +245,13 @@ publish-docs:
...
   image:                           parity/rust-android:gitlab-ci
   variables:
     CARGO_TARGET:                  armv7-linux-androideabi
   dependencies:                    []
   script:
-    - scripts/gitlab/build-unix.sh
+    - scripts/gitlab/build-linux.sh
   tags:
     - linux-docker
   allow_failure:                   true
```
gabreal: who will resolve the issues when they are already merged?

scripts/gitlab/cargo-check.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e # fail on any error
+set -u # treat unset variables as error
+echo "________Running cargo_check.sh________"
+
+PARAMS="--target $CARGO_TARGET --locked"
+
+echo "________Validate build________"
+time cargo check $PARAMS --no-default-features
```
gabreal: is this now duplicate to https://github.com/paritytech/parity-ethereum/pull/10452/files#diff-96edf7a6f008de9e928d04e1ae5e12a5R48?

scripts/gitlab/test-cpp.sh
gabreal: I don't quite understand. You make the changes now and test them later?

scripts/gitlab/test-linux.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+echo "________Running test-linux.sh________"
+set -e # fail on any error
+set -u # treat unset variables as error
+
+FEATURES="json-tests,ci-skip-tests"
+OPTIONS="--release"
+#use nproc `linux only
+THREADS=$(nproc)
```
gabreal: same here

gabreal:
great job, pipeline looks way more detailed and clear now

scripts/gitlab/cargo-check.sh
TriplEight: yes, forgot to remove the file, thx

scripts/gitlab/test-cpp.sh
TriplEight: it shoudn't influence other pipelines much, but if it will, I'll change the param to something like `nproc/4`

.gitlab-ci.yml
TriplEight: fixed

scripts/gitlab/cargo-check.sh
TriplEight: deleted this file

.gitlab-ci.yml
TriplEight: updated the template

.gitlab-ci.yml
TriplEight: template answer :)

.gitlab-ci.yml
TriplEight: template answer!

.gitlab-ci.yml
TriplEight: yup, liked it, but without cache it's just a bit faster

scripts/gitlab/test-linux.sh
TriplEight: template answer!!

General-Beck: we can check it in a separate pipeline @TriplEight

.gitlab-ci.yml
```diff
@@ -12,6 +12,16 @@ variables:
   CARGO_HOME:                      "${CI_PROJECT_DIR}/.cargo"
   CARGO_TARGET:                    x86_64-unknown-linux-gnu
 
+.rust_changes:                    &rust_changes      #exceptions to search for modified files
```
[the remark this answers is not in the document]
TriplEight: Yes, it works.
If changes meet these conditions, the job where it's met sipmly does not run, which economies everything.
Interesting thing though, we really do not need any of the current jobs to run in this case.
Also there is a cornercase, bug if you want. Pipeline won't run if the commit will change say README and some rust file.
Will try with `only` later.

scripts/gitlab/rust-changes.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
+echo "________Running rust_changes.sh________"
 set -e # fail on any error
 set -u # treat unset variables as error
 
+echo "__________Checking if Rust files were changed__________"
 git log --graph --oneline --decorate=short -n 10
```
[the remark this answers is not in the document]
TriplEight: right now the checking feature is removed from the pipeline in order to introduce caching and look how it works.
After that I'll reintroduce this feature and most probably delete this file.

Issue #10306, opened 2019-02-06
PoA network partition upon change of contract-based ValidatorSet


mathiasfrey:
- **Parity Ethereum version**: 2.3.2 / 2.2.9
- **Operating system**:  Linux
- **Installation**: Dockerized: parity/parity:{{version}}
- **Fully synchronized**: yes
- **Network**: custom PoA
- **Restarted**: yes

We're trying to set up a PoA blockchain along the https://wiki.parity.io/Demo-PoA-tutorial with the following specifics:

```
// chain.json
"stepDuration": "5"
"transactionPermissionContractTransition": 60,
"transactionPermissionContract": <ADDR>,
```

```
// authority.toml
usd_per_tx = "0"
tx_queue_size = 32768
tx_queue_mem_limit = 8096
min_gas_price = 0
gas_floor_target = "20000000"
force_sealing = true
```

We start up the first node, deploy the provided contracts for validators (reporting) and permissions and scale up to 6 nodes with 1 authority and 5 non-authority nodes.

Then we change non-validators into validators step by step & slowly (30 blocks time in between). The network keeps being in sync, all nodes see each other as peers. 

This is how turning the first non-authority node into an authority node looks like. Pay attention to the benign behaviour reporting, probably a race condition as the change of validator set might not be propagated fast enough / takes effect too quickly. In any case, the mining goes on:

```
2019-02-06 17:34:56 UTC Imported #1997 0x229d…9780 (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2019-02-06 17:35:01 UTC Imported #1998 0x6518…4caa (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2019-02-06 17:35:06 UTC Imported #1999 0xc04d…9c5c (0 txs, 0.00 Mgas, 0 ms, 0.57 KiB)
2019-02-06 17:35:06 UTC Error serving connection: Error { kind: Parse(Header), cause: None }
2019-02-06 17:35:17 UTC    5/25 peers    724 KiB chain  396 KiB db  0 bytes queue  243 KiB sync  RPC:  0 conn,    4 req/s,  170 µs
2019-02-06 17:35:19 UTC Reported benign validator misbehaviour 0xf33c…b8c2
[...]

PR #10467, opened 2019-03-12
Ensure static validator set changes are recognized

HCastano:
Fixes the issue seen in #10306 when updating a fixed list of validators. I've also been working on a test to simulate this scenario but it's been a pain in the ass to write so I'll be opening up an issue and doing it in another PR.


andresilva:
We still need to debug #10306 for changes signaled through the contract though.

HCastano:
Yeah, contracts are still a bit of a mystery

PR #10479, opened 2019-03-15
simplify block module and usage

debris:
- removed redundant `IsBlock` trait
- removed unused `LockedBlock::hash` function
- removed `ClosedBlock::hash` function, cause it caused confustion (see miner.rs)
- removed unused `OpenBlock::env_info` function

### DIFF
--- a/ethcore/src/block.rs
+++ b/ethcore/src/block.rs
@@ -367,9 +367,6 @@ impl ops::Deref for SealedBlock {
 }
 
 impl ClosedBlock {
-	/// Get the hash of the header without seal arguments.
-	pub fn hash(&self) -> H256 { self.header.bare_hash() }
-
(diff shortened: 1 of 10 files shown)


ethcore/src/block.rs
```diff
@@ -423,18 +400,13 @@ impl LockedBlock {
 		self.block.header.set_receipts_root(
 			ordered_trie_root(self.block.receipts.iter().map(|r| r.rlp_bytes()))
 		);
-		// compute hash and cache it.
-		self.block.header.compute_hash();
```
ascjones: Assume this was redundant

debris: yes, it was!

andresilva: Seems like it will be computed when actually needed (in `seal` or `try_seal` below).

PR #10493, opened 2019-03-20
fix(rpc): fix a bunch of clippy lints

niklasad1:
Attempt to cleanup the `rpc` crate after https://github.com/paritytech/parity-ethereum/pull/10217 was merged which has a lot of needless `into()`

I have ignored a bunch of lints which I think is just annoying which can be found in `rpc/lib.rs`. I can revert that change if anyone doesn't agree.

Furthermore, I have fixed most `clippy::all` lints with ~15 exceptions!


rpc/src/v1/impls/light/parity.rs
```diff
@@ -191,7 +191,8 @@ where
 	}
 
 	fn phrase_to_address(&self, phrase: String) -> Result<H160> {
-		Ok(Brain::new(phrase).generate().unwrap().address().into())
+		// TODO(niklasad1): replace `unwrap`
+		Ok(Brain::new(phrase).generate().unwrap().address())
```
ordian:
```suggestion
		Ok(Brain::new(phrase).generate().expect("Brain::generate always returns Ok; qed").address())
```

rpc/src/v1/impls/light/parity.rs
```diff
@@ -191,7 +191,8 @@ where
 	}
 
 	fn phrase_to_address(&self, phrase: String) -> Result<H160> {
-		Ok(Brain::new(phrase).generate().unwrap().address().into())
+		// TODO(niklasad1): replace `unwrap`
```
ordian:
```suggestion
```

rpc/src/lib.rs
```diff
@@ -17,6 +17,28 @@
 //! Parity RPC.
 
 #![warn(missing_docs, unused_extern_crates)]
+#![cfg_attr(feature = "cargo-clippy", warn(clippy::all, clippy::pedantic))]
+#![cfg_attr(
+	feature = "cargo-clippy",
+	allow(
+		clippy::empty_enum,
+		clippy::redundant_field_names,
```
ordian: I think shorthand syntax has no drawbacks.

rpc/src/lib.rs
```diff
@@ -17,6 +17,28 @@
...
+		// things are often more readable this way
+		clippy::cast_lossless,
+		clippy::module_name_repetitions,
+		clippy::single_match_else,
+		clippy::type_complexity,
+		clippy::use_self,
+		clippy::zero_prefixed_literal,
```
ordian: Do we actually have zero prefixed literals?

rpc/src/v1/impls/eth.rs
```diff
@@ -518,8 +517,8 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
 
 			let info = SyncInfo {
 				starting_block: status.start_block_number.into(),
-				current_block: current_block.into(),
-				highest_block: highest_block.into(),
+				current_block: current_block,
```
HCastano: Redundant field name

rpc/src/v1/impls/eth.rs
```diff
@@ -518,8 +517,8 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> Eth for EthClient<
...
+				highest_block: highest_block,
```
HCastano: Same here

rpc/src/lib.rs
niklasad1: Yes, I agree but I don't think redundant field names is bad enough to fix!

Thoughts?

rpc/src/lib.rs
niklasad1: No, remove? 

Well, this a first attempt to get some consensus around lints to allow/warn/deny in the code base. To make it easier to use the same in all the crates

rpc/src/lib.rs
niklasad1: Fixed all these and remove the `ignore`

rpc/src/v1/informant.rs
```diff
@@ -135,7 +135,7 @@ impl<T: Default + Copy + Ord> StatsCalculator<T> {
 #[derive(Default, Debug)]
 pub struct RpcStats {
 	requests: RwLock<RateCalculator>,
-	roundtrips: RwLock<StatsCalculator<u32>>,
+	roundtrips: RwLock<StatsCalculator<u128>>,
```
HCastano: What Clippy lint does this change correspond to?

niklasad1: That's a good question it was this one https://rust-lang.github.io/rust-clippy/master/#duration_subsec

But I replaced our `own` as_micros() method with `Duration::as_micros` and don't truncate the result as `u32`

Issue #8269, opened 2018-03-30
Add filtering capability to `parity_pendingTransactions`


AdvancedStyle:
Is there a way to limit the information returned from parity_pendingTransactions over JSON RPC?

Currently it returns all transactions in mempool with full transaction data in JSON format....which takes a long time to parse.

Is it possible to get only the transaction hashes?

In my use case I'd like to be able to poll the mempool and then compare it against my previously list of mempool transactions and then only lookup the full details on the new ones (so probably lookup 5 or 10 new transactions every few seconds, as opposed to trying to pull details for 10000 transactions every time)

With only a list of transaction hashes this compare function can be performed quite quickly, but with not with parsing a massive JSON.

tomusdrw:
Have you considered using `eth_subscribe(newPendingTransactions)`? Currently there is no way to get limited set of data, but I think it's a pretty good idea to allow that.

ppeinsold:
Would love to have a way of filtering here.
Use Cases:
- Check if a transaction with given hash is still pending
- Get all pending transactions by sender
- Get all pending transactions by receiver

ordian:
closed by #10506.

PR #10506, opened 2019-03-21
Add filtering capability to `parity_pendingTransactions` (issue 8269)

lamafab:
Hi

This pull request adds filtering capabilities for `parity_pendingTransactions`, as requested in issue #8269.

**Changes**
- MinerService now has a new `ready_transactions_filtered` method, which can filter transactions by hash, sender or receiver.
- The regular `ready_transactions` method wraps over the filter, therefore `ready_transactions_filtered` can be called directly from the RPC interface without having to adjust code through the whole codebase.

**Considerations**
I'm having slight trouble testing the code; `ready_transactions` (the original implementation) always returns two pending transactions on me, never more. This makes testing inconvenient. It might be related to a detail in the test miner that I'm missing, basing my assumption on this method description:

> /// Get a list of all ready transactions either ordered by priority or unordered (cheaper).
> ///
> /// **Depending on the settings may look in transaction pool or only in pending block.**
> /// If you don't need a full set of transactions, you can add `max_len` and create only a limited set of
> /// transactions.

If someone can point me in the right direction I can write some complete test cases **(so far, NO testing is being done).**

See `fn filter_pending_transactions_by_tx_hash()` in  `ethcore/src/miner/miner.rs` for more.

Any feedback is welcome!


lamafab:
[clabot:check]

lamafab:
[clabot:check]

ethcore/src/miner/mod.rs
```diff
@@ -184,6 +184,14 @@ pub trait MinerService : Send + Sync {
 	fn ready_transactions<C>(&self, chain: &C, max_len: usize, ordering: PendingOrdering) -> Vec<Arc<VerifiedTransaction>>
 		where C: ChainInfo + Nonce + Sync;
 
+	/// Get a list of all ready transactions either ordered by priority or unordered (cheaper), optionally filtered by hash, sender or receiver.
+	///
+	/// Depending on the settings may look in transaction pool or only in pending block.
+	/// If you don't need a full set of transactions, you can add `max_len` and create only a limited set of
+	/// transactions.
+	fn ready_transactions_filtered<C>(&self, chain: &C, max_len: usize, tx_hash: Option<H256>, receiver: Option<H160>, sender: Option<H160>, ordering: PendingOrdering) -> Vec<Arc<VerifiedTransaction>>
```
tomusdrw: What about filtering by other stuff like gas price, value, gas, etc? I guess it might be more complicated, since you will most likely want greater than/lesser than operators.

I think I'd rather be for a `FilteringObject` instead of passing all parameters by index. So:

```rust
/// This will require a custom serde impl, to handle cases like:
/// { "gt": "0x..." }, { "lt": "0x..." }, "0x..."
/// The last case should resolve to `Equal` case.
enum Operator<T> {
  /// Don't filter by this field
  Any,
  /// Require the value to be equal
  Equal(T),
  /// Require the value to be greater than
  GreaterThan(T),
  /// Require the value to be less than
  LessThan(T),
}
#[serde(Serialize, Deserialize)]
struct ReadyTransactionsFilter {
   sender: Operator<Address>,
   receiver: Operator<Option<Address>>,
   gas: Operator<U256>,
   gas_price: Operator<U256>,
   value: Operator<U256>,
   // I don't see use cases for this, but we might add it for completeness.
   nonce: Operator<U256>,
   /// We can also unify this limitting parameter to be passed here,
   /// but it's not strictly required since it has different semantics, so your call on this one :)
   // max_results: usize,
}

/// then you could unify the methods:
fn ready_transactions(&self, chain, max_len, filter: Option<ReadyTransactionsFilter>)

**Comment by ascjones** [comment]
@lamafab needs merging with master

**Comment by lamafab** [comment]
@tomusdrw @ascjones How quickly do those changes need to be implemented? I'm currently going through exam week, but I can complete this PR, soon. However, I can not guarantee on exactly when...

Regarding testing: I will skip the direct `MiningService` tests and use the RPC API, as I did in an other PR.

File: ethcore/src/miner/miner.rs [hunk 2]
```diff
@@ -1018,6 +1018,15 @@ impl miner::MinerService for Miner {
 		-> Vec<Arc<VerifiedTransaction>>
 	where
 		C: ChainInfo + Nonce + Sync,
+	{
+		// No special filtering options applied (neither tx_hash, receiver or sender)
+		self.ready_transactions_filtered(chain, max_len, None, None, None, ordering)
+	}
+
+	fn ready_transactions_filtered<C>(&self, chain: &C, max_len: usize, tx_hash: Option<H256>, receiver: Option<H160>, sender: Option<H160>, ordering: miner::PendingOrdering)
```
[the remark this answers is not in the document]
lamafab: I included the `tx_hash` since it was part of the originally requested issue. But yes; `eth_getTransactionByHash` does the same thing, assuming it also works with pending transactions and not fully completed/verified ones (?) I'll take a deeper look at the code.

tomusdrw:
@lamafab take your time. Just let us know roughly when you plan to start working on this, so we know when it's ok to ping you.

Good luck with your exams!

> Regarding testing: I will skip the direct MiningService tests and use the RPC API, as I did in an other PR.

I'd still encourage you to write a simple test to the filtering logic, I'm happy to help with that, you can catch me on gitter or riot (@todr:matrix.parity.io)

ethcore/src/miner/miner.rs
```diff
@@ -1045,6 +1054,34 @@ impl miner::MinerService for Miner {
...
+					.filter(|tx| {
+						if let Some(sender) = sender {
+							tx.signed().sender() == sender
+						} else {
+							true // do not filter if None was passed
+						}
+					})
+					// Filter by receiver
+					.filter(|tx| {
+						if let Some(receiver) = receiver {
```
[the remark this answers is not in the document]
lamafab: > Why not have `receiver: Option<Option<Address>>` then?

Ok, I can extend the filter to contract creations.

ethcore/src/miner/mod.rs
lamafab: > What about filtering by other stuff like gas price, value, gas, etc? I guess it might be more complicated, since you will most likely want greater than/lesser than operators.

It sure looks more complicated :) But I can come up with something. Would you like to have all those changes included in one round or should I first complete the rest and follow up with the gas/price filtering implementation later on? I'm also curious about performance on this one, but let's see.

lamafab:
> Good luck with your exams!

Thanks!

> I'd still encourage you to write a simple test to the filtering logic, I'm happy to help with that, you can catch me on gitter or riot (@todr:matrix.parity.io)

Yes, I was planning on testing the filtering logic via the RPC API.

lamafab:
@tomusdrw @ascjones  Alright, I did it! The unified _FilterOptions_ are now implemented in this PR.

```
pub struct FilterOptions {
    pub sender: FilterOperator<Address>,
    pub receiver: FilterOperator<Address>,
    pub gas: FilterOperator<U256>,
    pub gas_price: FilterOperator<U256>,
    pub value: FilterOperator<U256>,
    pub nonce: FilterOperator<U256>,
}

pub enum FilterOperator<T> {
    Any,
    Eq(T),
    GreaterThan(T),
    LessThan(T),
    ContractCreation, // only used for `receiver`
}
```


**What this PR contains:**
- The `parity_pendingTransactions` RPC API can take optional filter operators like the following:
```
    {
        "sender": {
            "eq": "0x5f3dffcf347944d3739b0805c934d86c8621997f"
        },
        "gas": {
            "gt": "0x493e0"
        }
    }
```
- Since not all operators can be used equally on each filter, the allowed/disallowed operators are caught during the deserialization process with a custom implementation of `Deserialize` (those tests are included in the same location where _FilterOptions_ is defined).

**What's missing**
- One could argue that there should be _GreaterEqualThan_ and _LessEqualThan_ operators.
- While there are deserialization tests, there are no logic tests for the filtering. @tomusdrw I did a code base search and could not find a lot of useful information regarding testing that API (there are some very simple/oversimplified tests). Can you tell me where those logic tests should be placed and what testing data I should use (or to be more specific: how should the RPC API test server be created?). I can hardly use Gitter during the workday, so I'd appreciate it if you could give me some advice here.

Thank you very much (and for the patience!)

ethcore/Cargo.toml
```diff
@@ -61,6 +61,7 @@ rlp = "0.4.0"
 rlp_derive = { path = "../util/rlp-derive" }
 rustc-hex = "1.0"
 serde = "1.0"
+serde_json = "1.0"
```
tomusdrw: seems it's only for tests, can you move it to `dev-dependencies`?

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
+use ethereum_types::{Address, U256};
```
tomusdrw: We need a preamble in every file.

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+use std::fmt;
+use std::marker::PhantomData;
+
+/// This structure provides filtering options for the pending transactions RPC API call
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct FilterOptions {
+    /// Contains the operator to filter the sender value of the transaction
+    pub sender: FilterOperator<Address>,
+    /// Contains the operator to filter the receiver value of the transaction
+    pub receiver: FilterOperator<Address>,
```
tomusdrw: Could we do `FilterOperator<Option<Address>>`? That wouldn't require the special case for `FilterOperator` as it is currently.

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+            gas: FilterOperator::Any,
+            gas_price: FilterOperator::Any,
+            value: FilterOperator::Any,
+            nonce: FilterOperator::Any,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum FilterOperator<T> {
```
tomusdrw: Since it's also public it could use some basic docs.

ethcore/src/miner/filter_options.rs
tomusdrw: Could we also use `from/to` instead of `sender/receiver`? I think it's more inline with other RPC structs.

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+    Any,
+    Eq(T),
+    GreaterThan(T),
+    LessThan(T),
+    ContractCreation, // only used for `receiver`
```
tomusdrw: see suggestion about changing the `receiver` type

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+                    fn visit_map<M>(self, mut map: M) -> Result<Self::Value, M::Error>
+                    where
+                        M: MapAccess<'de>,
+                    {
+                        let mut counter = 0;
+                        let mut f_op = FilterOperator::Any;
+
+                        while let Some(key) = map.next_key()? {
+                            match key {
+                                "eq" => f_op = FilterOperator::Eq(map.next_value()?),
```
tomusdrw: Is it possible to handle this:
```
{
  "sender": "0x..."
}
```
same as `eq` for convenience?

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+                                "gt" => f_op = FilterOperator::GreaterThan(map.next_value()?),
+                                "lt" => f_op = FilterOperator::LessThan(map.next_value()?),
+                                "action" => {
```
tomusdrw: I'd ditch that in favour of matching to `null`

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+                            ))
+                        }
+                    }
+                }
+
+                Ok(filter)
+            }
+        }
+
+        impl<'de, T: Deserialize<'de>> Deserialize<'de> for FilterOperator<T> {
```
tomusdrw: Do we really need custom `Deserialize` for `FilterOperator`? It seems to be tagged enum representation with a bunch of renames, no?

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,823 @@
...
+                deserializer.deserialize_map(FilterOperatorVisitor { data: PhantomData })
+            }
+        }
+
+        deserializer.deserialize_map(FilterOptionsVisitor)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
tomusdrw: :heart:

ethcore/types/src/transaction/transaction.rs
```diff
@@ -462,6 +462,24 @@ impl SignedTransaction {
 		self.sender
 	}
 
+	/// Returns transaction receiver, if any
+	pub fn receiver(&self) -> Option<Address> {
```
tomusdrw: This should be moved to `UnverifiedTransaction`

ethcore/types/src/transaction/transaction.rs
```diff
@@ -462,6 +462,24 @@ impl SignedTransaction {
...
+		match self.transaction.unsigned.action {
+			Action::Create => None,
+			Action::Call(receiver) => Some(receiver),
+		}
+	}
+
+	// Reference to the plain text transaction
+	pub fn as_unsigned(&self) -> &Transaction {
```
tomusdrw: Since there are `Deref` implementations that method is not needed, you can directly access stuff from `Transaction` anyway.

ethcore/types/src/transaction/transaction.rs
```diff
@@ -462,6 +462,24 @@ impl SignedTransaction {
...
+		self.transaction.as_unsigned()
+	}
+
+	/// Returns transaction hash
+	pub fn hash(&self) -> H256 {
```
tomusdrw: Not needed, we have `fn hash` in `UnverifiedTransaction` and there is a `Deref` to it.

rpc/src/v1/impls/light/parity.rs
```diff
@@ -217,7 +218,7 @@ where
 			.map(Into::into)
 	}
 
-	fn pending_transactions(&self, limit: Option<usize>) -> Result<Vec<Transaction>> {
+	fn pending_transactions(&self, limit: Option<usize>, _sender: Option<FilterOptions>) -> Result<Vec<Transaction>> {
```
tomusdrw:
```suggestion
	fn pending_transactions(&self, limit: Option<usize>, _filter: Option<FilterOptions>) -> Result<Vec<Transaction>> {
```

ethcore/Cargo.toml
lamafab: Yes, I will do that.

ethcore/src/miner/filter_options.rs
lamafab: ... I'm not sure what you mean by this?

ethcore/src/miner/filter_options.rs
lamafab: Yes, this can simplify certain aspects. I will implement this.

seunlanlege:
Amazing work :handshake:

ethcore/src/miner/filter_options.rs
```diff
@@ -0,0 +1,869 @@
...
+            {
+                "nonce": {
+                    "abc": "0x0"
+                }
+            }
+        "#;
+        let res = serde_json::from_str::<FilterOptions>(json);
+        assert!(res.is_err());
+    }
+}
```
seunlanlege: needs a new line

lamafab:
Thanks for the nice feedback, everyone. I've added some very small changes (like adding a newline at the end of the file as @seunlanlege suggested).

PR #10514, opened 2019-03-23
fix(light): make `OnDemand` generic instead of using the concrete type

niklasad1:
Helps to mock `OnDemand` to facilitate for example `RPC integration tests`


ethcore/light/src/on_demand/mod.rs
```diff
@@ -94,6 +94,25 @@ pub mod error {
...
 
+/// Public interface for performing network requests `OnDemand`
+pub trait OnDemandRequester: Send + Sync {
+	/// Submit a strongly-typed batch of requests.
+	///
+	/// Fails if back-reference are not coherent.
+	fn request<T>(&self, ctx: &BasicContext, requests: T) -> Result<OnResponses<T>, basic_request::NoSuchOutput>
+	where
+		T: request::RequestAdapter,
+		Self: Sized;
```
ordian: I wonder why `Self: Sized` is required?

ascjones: Good question. Prevents being used as a trait object?

ordian: But this method is generic, I thought it's not allowed in a trait object already.

niklasad1: Thanks guys, it was not needed :+1: :1st_place_medal:

PR #10519, opened 2019-03-25
Add trace information to eth_estimateGas

### LINKED ISSUE #10360: Feature request: Include trace information when virtual calls throw exceptions
- **Parity Ethereum version**: 2.2.9-stable
- **Operating system**: Linux
- **Fully synchronized**: yes
- **Network**: ethereum, kovan
- **Restarted**: yes

Related to https://github.com/paritytech/parity-ethereum/issues/10147. The error message thrown when `eth_estimateGas` (and I assume `eth_call` and siblings) is very cryptic and could be improved by including diagnostic info and VM trace info if applicable.

fanatid:
Closes #10360

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1575,8 +1575,7 @@ impl Call for Client {
 		let cond = |gas| {
 			exec(gas)
 				.ok()
-				.map(|r| r.exception.is_none())
-				.unwrap_or(false)
+				.map_or(false, |r| r.exception.is_none())
 		};
 
 		if !cond(upper) {
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1583,8 +1583,8 @@ impl Call for Client {
 			upper = max_upper;
 			match exec(upper) {
 				Ok(v) => {
-					if v.exception.is_some() {
-						return Err(CallError::Exceptional(v.exception.unwrap()))
+					if let Some(exception) = v.exception {
+						return Err(CallError::Exceptional(exception))
 					}
 				},
 				Err(_e) => {
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1570,22 +1570,28 @@ impl Call for Client {
 			let schedule = machine.schedule(env_info.number);
 			Executive::new(&mut clone, &env_info, &machine, &schedule)
 				.transact_virtual(&tx, options())
-				.ok()
-				.map(|r| r.exception.is_none())
 		};
 
-		let cond = |gas| exec(gas).unwrap_or(false);
+		let cond = |gas| {
+			exec(gas)
+				.ok()
+				.map(|r| r.exception.is_none())
+				.unwrap_or(false)
+		};
 
 		if !cond(upper) {
 			upper = max_upper;
 			match exec(upper) {
-				Some(false) => return Err(CallError::Exceptional),
-				None => {
+				Ok(v) => {
+					if v.exception.is_some() {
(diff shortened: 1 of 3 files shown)


joshua-mir:
This should probably be in the `parity_` namespace as it breaks spec.

ethcore/src/client/client.rs
```diff
@@ -1570,22 +1570,28 @@ impl Call for Client {
...
-		let cond = |gas| exec(gas).unwrap_or(false);
-
 		if !cond(upper) {
 			upper = max_upper;
 			match exec(upper) {
-				Some(false) => return Err(CallError::Exceptional),
-				None => {
+				Ok(v) => {
+					if v.exception.is_some() {
+						return Err(CallError::Exceptional(v.exception.unwrap()))
```
ordian: Could you use an `if let Some(exception) = v.exception` statement to avoid `.unwrap()`?

fanatid: I had read in docs that `.unwrap()` can cause panic, but before I check with `.is_some()`, so thought that this is fine.
Sorry, it's not clear for me, how `.unwrap()` can be avoided with `if let Some(exception) = v.exception`... can you explain?

ordian: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=132afa5ce895d6d8d46882ee4df7d904

fanatid: Thank you for example.

rpc/src/v1/helpers/errors.rs
```diff
@@ -173,11 +173,11 @@ pub fn state_corrupt() -> Error {
 	internal("State corrupt", "")
 }
 
-pub fn exceptional() -> Error {
+pub fn exceptional<T: fmt::Debug>(data: T) -> Error {
```
ordian: I think `fmt::Display` implementation for `vm::Error` is more user-friendly.

fanatid: I'm I understand correct that if we switch to `fmt::Display` this formatter will be used?
https://github.com/paritytech/parity-ethereum/blob/9cb86061032a323710a1dd549e1e1f0a4e33eedd/ethcore/vm/src/error.rs#L100-L117

ordian: Yes, that's my suggestion, but I don't have a strong opinion about that.

rpc/src/v1/helpers/errors.rs
```diff
@@ -173,11 +173,11 @@ pub fn state_corrupt() -> Error {
 	internal("State corrupt", "")
 }
 
-pub fn exceptional() -> Error {
+pub fn exceptional<T: fmt::Display>(data: T) -> Error {
 	Error {
 		code: ErrorCode::ServerError(codes::EXCEPTION_ERROR),
 		message: "The execution failed due to an exception.".into(),
-		data: None,
+		data: Some(Value::String(format!("{:?}", data))),
```
ordian:
```suggestion
		data: Some(Value::String(data.to_string())),
```

ethcore/src/client/client.rs
```diff
@@ -1570,22 +1570,28 @@ impl Call for Client {
 			let schedule = machine.schedule(env_info.number);
 			Executive::new(&mut clone, &env_info, &machine, &schedule)
 				.transact_virtual(&tx, options())
+		};
+
+		let cond = |gas| {
+			exec(gas)
 				.ok()
 				.map(|r| r.exception.is_none())
+				.unwrap_or(false)
```
niklasad1: could be `map_or(false, |r| r.exception.is_none())`

Issue #9785, opened 2018-10-20
Enable/disable in stratum section stratum of config?


stone212:
Looking at this page: https://wiki.parity.io/Configuring-Parity-Ethereum.html

I see the --stratum option on the command line but I don't see a toml option to enable stratum.  Is this one of those things that is right in front of me that I can't see?

niklasad1:
@stone212 

It is sufficient to just add ```[stratum]``` in your `toml` file but we should add a enable/disable option to that for clarity!

Such as 
```toml
[stratum]
disable = true
```
```

**Comment by stone212** [comment]
@joshua-mir @niklasad1 

Just for clarity.  If you are already making a change I recommend making the one I posted.  It uses an already existing [category] and makes more logical sense (to me at least.)

**Comment by joshua-mir** [comment]
@stone212 Stratum is its own category because it has configuration options of its own. Ie, 

```
[stratum]
# Stratum Server will be running on port 8007.
port = 8007
# Secret for authorizing Stratum server for peers.
secret = "Yellow"
```

I agree that it might be better to have a enable/disable option though. Should I reopen as a feature request @niklasad1?

niklasad1:
@joshua-mir Yes, please do

joshua-mir:
Personally favor `disable = true/false` so existing configs with this quirk aren't broken. But the behavior is weird because it's disabled by default unless the [stratum] section is present 🤔

Genysys:
should a different toml file be added for stratum or would it be wiser to add them to the exisiting toml files?

joshua-mir:
Doesn't make sense not to just use the [stratum] section of the existing config - separate files should only be used for things like passwords

niklasad1:
Closed by #10521

PR #10521, opened 2019-03-25
Explicitly enable or disable Stratum in config file (Issue 9785)

lamafab:
Hi

This PR adds the requested feature from issue #9785.

**Changes**
- Users can now directly enable or disable Stratum in the configuration file:
```
[stratum]
disable = false
port = 8007
secret = "Yellow"
```
```
[stratum]
disable = true
```
- For backward compatibility reasons, Stratum is enabled by default **if** the `[stratum]` section is present and **not** disabled by setting `disable = true`.
- Contains additional tests for the Stratum section in the config file.

I've run those changes directly on my machine; it works as intended (assuming these are the expected requirements. IMO, it would be cleaner if Stratum was only enabled if we explicitly set a `enabled = true` option. This would break backward compatibility, however).

Any feedback is welcome.


lamafab:
Are the CPP errors in your CI related to my changes? It does look like another problem, unless I'm wrong.

```
Scanning dependencies of target libparity
[ 10%] Creating directories for 'libparity'
[ 20%] No download step for 'libparity'
[ 30%] No update step for 'libparity'
[ 40%] No patch step for 'libparity'
[ 50%] No configure step for 'libparity'
[ 60%] Performing build step for 'libparity'
CMake Error at /builds/parity/parity-ethereum/parity-clib/examples/cpp/build/libparity-prefix/src/libparity-stamp/libparity-build-.cmake:16 (message):
  Command failed: 1

   '/usr/bin/cmake' '-Dmake=' '-Dconfig=' '-P' '/builds/parity/parity-ethereum/parity-clib/examples/cpp/build/libparity-prefix/src/libparity-stamp/libparity-build--impl.cmake'

  See also

    /builds/parity/parity-ethereum/parity-clib/examples/cpp/build/libparity-prefix/src/libparity-stamp/libparity-build-*.log


CMakeFiles/libparity.dir/build.make:110: recipe for target 'libparity-prefix/src/libparity-stamp/libparity-build' failed
make[2]: *** [libparity-prefix/src/libparity-stamp/libparity-build] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/libparity.dir/all' failed
make[1]: *** [CMakeFiles/libparity.dir/all] Error 2
make: *** [all] Error 2
Makefile:83: recipe for target 'all' failed
```

niklasad1:
/cc @TriplEight any idea what's going on here?

I think we should enable `VERBOSE` build for the `cpp example` (I have no clue what's is going on)

```make VERBOSE=1 -j $THREADS```

TriplEight:
@niklasad1 here you go: https://github.com/paritytech/parity-ethereum/pull/10524

soc1c:
I just restarted the job and it worked, maybe hold back with hacking the pipelines and try restarting it in future? ;)

HCastano:
We should also update the [wiki](https://wiki.parity.io/Configuring-Parity-Ethereum) to reflect this change. @lamafab  do you mind putting in a PR here: https://github.com/paritytech/wiki?

lamafab:
@HCastano Sure, no problem. I will do that after work (this evening).

stone212:
@lamafab You said `Stratum is enabled by default if the [stratum] section is present and not disabled by setting disable = true`

Does this mean:

Stratum is 

1. enabled by default if the [stratum] section is present
2. **not** disabled by setting `disable = true`

Why would you have `disable = true` if it does not disable it?

lamafab:
@stone212 This decision was made for backwards compatibility reasons. Stratum is enabled if the `[stratum]` section is present, either with or without `disable = false`. If you set `disable = true`, then Stratum gets disabled.

Sorry for the confusion...

stone212:
@lamafab It's a good decision, I just wanted to make the language clear.  Your sentence could mean one thing or the exact opposite.  Good to have this change!

PR #10526, opened 2019-03-26
tx-pool: check transaction readiness before replacing

ascjones:
Depends on https://github.com/paritytech/parity-common/pull/116.

When the pool becomes full, before replacing an existing transaction with a new one: check whether a Future transaction is not replacing a Ready one.


ascjones:
Waiting for https://github.com/paritytech/parity-common/pull/116 to be merged and transaction-pool 2.0 to be released

miner/src/pool/replace.rs
```diff
@@ -0,0 +1,416 @@
...
+			let old_score = (old.priority(), old.gas_price());
+			let new_score = (new.priority(), new.gas_price());
+			if new_score > old_score {
+				let state = &self.client;
+				// calculate readiness based on state nonce + pooled txs from same sender
+				let is_ready = |replace: &ReplaceTransaction<T>| {
+					let state_nonce = state.account_nonce(replace.sender());
+					let nonce =
+						replace.pooled_by_sender.map_or(state_nonce, |txs| {
+							txs.iter().fold(state_nonce, |nonce, tx| {
```
tomusdrw: I think we should early-exit here on first unmatching nonce. Also regular `for` loop might be more readable in such case.

soc1c:
This seemingly broke a test, I tried 3x with same test getting stuck:

```
test tests::private::send_private_transaction ... test tests::private::send_private_transaction has been running for over 60 seconds
ERROR: Job failed: execution took longer than 3h10m0s seconds
```

PR #10527, opened 2019-03-26
Refactor cpp example

niklasad1:
What does this PR:
* Replace potential undefined behavior when loop for `while (callback->counter != 0)` with `mutex + counter`
* Use timeouts to shortcut 
* Remove light client from the example
* Execution should take ~2 min unless shutdown takes longer than `"usual"`
* Enable WebSocket subscription again

### DIFF
--- a/parity-clib/examples/cpp/main.cpp
+++ b/parity-clib/examples/cpp/main.cpp
@@ -16,24 +16,19 @@
 
 #include <atomic>
 #include <chrono>
-#include <parity.h>
 #include <regex>
 #include <string>
 #include <cstdint>
 #include <thread>
-
-void* parity_run(std::vector<const char*>);
-int64_t parity_subscribe_to_websocket(void*);
-int64_t parity_rpc_queries(void*);
+#include "parity_client.hpp"
+#include "websocket_subscription.hpp"
 
 const int64_t SUBSCRIPTION_ID_LEN = 18;
-const size_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
+const uint64_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
 const uint64_t CALLBACK_RPC = 1;
 const uint64_t CALLBACK_WS = 2;
 
-struct Callback {
-	int64_t type;
-};
+std::atomic<uint64_t> callback_counter;
 
 // list of rpc queries
 const std::vector<std::string> rpc_queries {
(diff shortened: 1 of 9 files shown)


parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
...
-
-		if (parity_subscribe_to_websocket(parity)) {
-			printf("ws_queries failed\r\n");
-			return 1;
-		}
-
-		if (parity != nullptr) {
-			parity_destroy(parity);
-		}
+	std::vector<const char*> config = {"--no-ipc" , "--jsonrpc-apis=all", "--chain", "kovan"};
```
dvdplm: Maybe not use Kovan here, it being decommissioned and all?

parity-clib/examples/cpp/main.cpp
```diff
@@ -31,13 +32,11 @@ const unsigned int CALLBACK_WS = 2;
 
 struct Callback {
 	unsigned int type;
-	long unsigned int counter;
 };
 
 // list of rpc queries
 const std::vector<std::string> rpc_queries {
 	"{\"method\":\"parity_versionInfo\",\"params\":[],\"id\":1,\"jsonrpc\":\"2.0\"}",
-	"{\"method\":\"eth_getTransactionReceipt\",\"params\":[\"0x444172bef57ad978655171a8af2cfd89baa02a97fcb773067aef7794d6913fff\"],\"id\":1,\"jsonrpc\":\"2.0\"}",
```
dvdplm: Why remove this?

parity-clib/examples/cpp/main.cpp
soc1c: We just deprecate it but it will certainly be around for a while.

dvdplm:
I don't have the C++ chops to review this. @Robbepop can you take a look please?

parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
...
-		if (parity_subscribe_to_websocket(parity)) {
-			printf("ws_queries failed\r\n");
-			return 1;
-		}
-
-		if (parity != nullptr) {
-			parity_destroy(parity);
-		}
+	std::vector<const char*> config = {"--no-ipc" , "--jsonrpc-apis=all", "--chain", "kovan"};
+	void* parity = parity_run(config);
```
DemiMarie-parity:
```suggestion
	void *parity = parity_run(config);
```

would be a better match to how this is actually parsed by the C++ compiler.  However, the use of a raw pointer ― _especially_ a `void *` ― is an anti-pattern in C++.  It would be much better to use a proper C++ class with a destructor, following RAII.  Otherwise, it is highly likely that resources will be leaked.  This is especially important in C++, as using exceptions in C++ is considered ideomatic.  In this case, the resources will be destroyed when `main()` returns and the program exits, but that is still a bad example.

parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
...
+	if (parity_rpc_queries(parity)) {
+		printf("rpc_queries failed\r\n");
```
DemiMarie-parity: This leaks `parity`.

parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
...
-			printf("rpc_queries failed\r\n");
-			return 1;
-		}
-
-		if (parity_subscribe_to_websocket(parity)) {
-			printf("ws_queries failed\r\n");
-			return 1;
-		}
+	if (parity_subscribe_to_websocket(parity)) {
+		printf("ws_queries failed\r\n");
```
DemiMarie-parity: Ditto

parity-clib/examples/cpp/main.cpp
```diff
@@ -109,16 +88,17 @@ int parity_rpc_queries(void* parity) {
 		return 1;
 	}
 
-	Callback cb { .type = CALLBACK_RPC, .counter = rpc_queries.size() };
+	Callback cb { .type = CALLBACK_RPC };
```
DemiMarie-parity: This was a non-standard extension to C++ last I checked, but this may have changed in C++17.

parity-clib/examples/cpp/main.cpp
```diff
@@ -129,7 +109,8 @@ int parity_subscribe_to_websocket(void* parity) {
 
 	std::vector<const void*> sessions;
 
-	Callback cb { .type = CALLBACK_WS, .counter = ws_subscriptions.size() };
+	Callback cb { .type = CALLBACK_WS };
+	callback_counter = ws_subscriptions.size();
 
 	for (auto sub : ws_subscriptions) {
 		void *const session = parity_subscribe_ws(parity, sub.c_str(), sub.length(), callback, &cb);
```
DemiMarie-parity: Again, the use of `void *` is an anti-pattern.  This should be wrapped in a type-safe object that cleans up the resources in its destructor.

parity-clib/examples/cpp/main.cpp
```diff
@@ -139,12 +120,11 @@ int parity_subscribe_to_websocket(void* parity) {
 		sessions.push_back(session);
 	}
 
-	while(cb.counter != 0);
-	std::this_thread::sleep_for(std::chrono::seconds(60));
+	std::this_thread::sleep_for(std::chrono::seconds(15));
 	for (auto session : sessions) {
 		parity_unsubscribe_ws(session);
```
DemiMarie-parity: Using RAII would make this unnecessary, as `parity_unsubscribe_ws` would automatically be called.

parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
 	"{\"method\":\"eth_subscribe\",\"params\":[\"newHeads\"],\"id\":1,\"jsonrpc\":\"2.0\"}"
 };
 
+std::atomic<int> callback_counter;
```
DemiMarie-parity: `int` is probably the wrong type here.  I suspect that `std::int64_t` is the correct one.

parity-clib/examples/cpp/main.cpp
niklasad1: yeah, you are right `designated initializers` is a C++20 thing

parity-clib/examples/cpp/main.cpp
niklasad1: It doesn't work well for `light client`

scripts/gitlab/test-cpp.sh
```diff
@@ -11,7 +11,6 @@ mkdir -p $DIR
 cd $DIR
 cmake ..
 make VERBOSE=1 -j $THREADS
```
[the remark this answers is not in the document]
Robbepop: Or alternatively use the `ninja` build system. Then you do not need to calculate the number of threads manually because it is done by ninja by default.

parity-clib/examples/cpp/main.cpp
niklasad1: [C++ 11](https://github.com/paritytech/parity-ethereum/blob/master/parity-clib/examples/cpp/CMakeLists.txt#L4) but might possible to bump it to c++17.

I think it is a `CI` related question

dvdplm:
@niklasad1 any chance we can get this in shape in time for 2.6?

niklasad1:
> @niklasad1 any chance we can get this in shape in time for 2.6?

Due date for 2.6 is (I guess next week)? 
I don't know I'm struggling a bit in substrate and I want complete that first.

dvdplm:
> Due date for 2.6 is (I guess next week)?

Cut-off for PR work is tomorrow, Thursday.

debris:
@niklasad1  what is the status of this pr?

parity-clib/examples/cpp/main.cpp
```diff
@@ -49,58 +48,38 @@ const std::vector<std::string> ws_subscriptions {
...
+
 // callback that gets invoked upon an event
 void callback(void* user_data, const char* response, size_t _len) {
 	Callback* cb = static_cast<Callback*>(user_data);
 	if (cb->type == CALLBACK_RPC) {
-		cb->counter -= 1;
+		callback_counter -= 1;
 	} else if (cb->type == CALLBACK_WS) {
 		std::regex is_subscription ("\\{\"jsonrpc\":\"2.0\",\"result\":\"0[xX][a-fA-F0-9]{16}\",\"id\":1\\}");
```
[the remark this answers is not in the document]
DemiMarie-parity: In some cases (where there is a constructor that takes a `std::initializer_list`), using `{...}` is not possible.  In these cases, using `((...))` also does the job.

parity-clib/examples/cpp/main.cpp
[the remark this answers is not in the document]
niklasad1: That's a good idea but `string_view` is not null-terminated and doesn't provide `c_str()`. Thus, it is possible to use [string_view::data](https://en.cppreference.com/w/cpp/string/basic_string_view/data) (dangerous)  or `std::string(str_view).c_str()` so I decided that it is not worth the effort.

niklasad1:
@Robbepop @DemiMarie-parity can you have look at this again? Thanks

niklasad1:
Closing as stale

PR #10529, opened 2019-03-27
Add SealingState; don't prepare block when not ready.

afck:
This replaces `Engine::seals_internally` with `Engine::sealing_state`, and returns an enum indicating whether the engine seals internally and whether it is currently ready to do so.

For engines that don't seal internally, `Miner::prepare_block` is only called when they are ready. This prevents preparing (and subsequently failling to seal) blocks all the time with `AuthorityRound`.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -60,7 +60,7 @@ use client::{
 	IoClient, BadBlocks,
 };
 use client::bad_blocks;
-use engines::{MAX_UNCLE_AGE, EthEngine, EpochTransition, ForkChoice, EngineError};
+use engines::{MAX_UNCLE_AGE, EthEngine, EpochTransition, ForkChoice, EngineError, SealingState};
 use engines::epoch::PendingTransition;
 use error::{
 	ImportError, ExecutionError, CallError, BlockError,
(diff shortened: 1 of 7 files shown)


afck:
I [tried hard](https://github.com/poanetwork/parity-ethereum/issues/70#issuecomment-466016955) to not miss anything, but I'm not very confident about the change, so let's only merge it if someone who is more familiar with the code is certain that it makes sense.

An enum better conveys the meaning of the value, but of course I'm happy to revert to the `Option<bool>`, if preferred.

afck:
[clabot:check]

niklasad1:
@afck Can you please rebase this branch because a `force-push` was pushed to master recently?

afck:
Sorry for the delay! I rebased it.

varasev:
@niklasad1 Could you (or someone else from Parity team) please review these changes? We made the same for our fork (https://github.com/poanetwork/parity-ethereum/pull/103) but not sure the changes are correct. It would be awesome if someone from Parity could have a look at these changes. It would help AuRa nodes not to waste CPU time when they are not validators. Thank you.

afck:
Sorry, I don't understand why one of the builds failed: Is that something about the build cache, or is it an actual test failure?

niklasad1:
@afck it is build cache error, I don't know why it failed but I restarted and it succeed!

PR #10545, opened 2019-03-28
fix(light account response): update `tx_queue` on every account request/response

niklasad1:
This PR changes so we update the `transaction queue` after every account request/response instead of having to wait every 10 minutes for the `tx cull timer to expire`.

It should be relatively cheap, requires to look a given key and loop through the transactions to find old ones and remove them!

However, should make the `light client` more responsive when more requests are performed

### DIFF
--- a/rpc/src/v1/helpers/light_fetch.rs
+++ b/rpc/src/v1/helpers/light_fetch.rs
@@ -40,12 +40,13 @@ use light::on_demand::{
 };
 use light::on_demand::error::Error as OnDemandError;
 use light::request::Field;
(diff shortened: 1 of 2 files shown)


niklasad1:
> We keep the cull timer (in parity/light_helpers/queue_cull.rs), do we still need it?

It is possible that the account is never `touched` after sending a transaction for example then the queue will never be culled if we remove the cull_timer.

However, I'm working on get rid of the `cull_timer` completely but I'll take of it in another PR

soc1c:
please rebase

PR #10551, opened 2019-03-29
clique: make state backfill time measurement more accurate

jwasinger:
Moves the start of the measurement before the expensive chain traversal loop.

### DIFF
--- a/ethcore/src/engines/clique/mod.rs
+++ b/ethcore/src/engines/clique/mod.rs
@@ -280,6 +280,13 @@ impl Clique {
 				let last_checkpoint_number = header.number() - header.number() % self.epoch_length as u64;
 				debug_assert_ne!(last_checkpoint_number, header.number());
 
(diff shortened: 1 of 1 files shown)


ethcore/src/engines/clique/mod.rs
```diff
@@ -277,6 +277,13 @@ impl Clique {
...
 			}
 			Some(c) => {
+				// Catching up state, note that we don't really store block state for intermediary blocks,
+				// for speed.
+				let backfill_start = time::Instant::now();
+				trace!(target: "engine",
+						"Back-filling block state. last_checkpoint_number: {}, target: {}({}).",
+						last_checkpoint_number, header.number(), header.hash());
+
 				let last_checkpoint_number = header.number() - header.number() % self.epoch_length as u64;
```
niklasad1: nit, move this line before the trace

jwasinger: Done

PR #10552, opened 2019-03-31
eth_getRawBlockByNumber for RPC API (part 1, issue 9057)

lamafab:
Hello

This PR contains (some) changes requested by issue #9057. The issue also requested a `eth_submitRawBlock` implementation. This feature has not been written yet -> I will follow up with that in a separate PR.

(The original request specified the **parity_** prefix instead of **eth_**. I assume that was an error and **eth_** is correct here?)

**Changes**
- `eth_getRawBlockByNumber` API for both the full and light client. This endpoint returns the raw block in RLP form, no extras included (non-Rich). The return value is tested on the official Ethereum test cases in _BlockchainTests/bcGasPricerTest/RPC_API_Test_ (the json file contains a "rlp" field).
- To avoid duplicating code, the `rich_block` method on `EthClient` in both the full and light client has been split into `encoded_block` and `rich_block` (`rich_block` calling `encoded_block` from within).

**Todo**
- As I already mentioned, `eth_submitRawBlock` will follow up in a separate RP.
- While the endpoint does accept a `include_txs: bool`parameter, it currently does not do anything (transactions are included, the default). The `rpc::v1::types::Block` type can easily exclude transactions, the `ethcore::types::encoded::Block` type however is a little more complex as it only contains the raw bytes. I could use `BlockView` to find the transactions and cut those out from the final form. I will do that, soon (unless someone can recommend a different way of doing this?). Should this parameter be removed for now?

**Open questions**
- Should there be a similar endpoint for `eth_getBlockByHash`, i.e. `eth_getRawBlockByHash`?
- While I did test the changes for the light client by hand, there do not seem to be any automated tests for it. Did I miss those somewhere? By searching the code base it seems to only test the full client RPC interface.


seunlanlege:
already implemented in #10609

PR #10558, opened 2019-04-02
Watch transactions pool

IntegralTeam:
Issue #9713
Still in progress, but would like some review on current approach.
What is not ready:
- [x] Implementation for LightChainClient
- [x] testing

I will add more if feedback will be positive


IntegralTeam:
[clabot:check]

miner/src/pool/listener.rs
```diff
@@ -116,6 +116,70 @@ impl txpool::Listener<Transaction> for Logger {
...
+		fmt.debug_struct("TransactionsPoolNotifier")
+			.field("listeners", &self.listeners.len())
+			.finish()
+	}
+}
+
+impl txpool::Listener<Transaction> for TransactionsPoolNotifier {
+	fn added(&mut self, tx: &Arc<Transaction>, _old: Option<&Arc<Transaction>>) {
+		let hash = tx.hash.clone();
+		self.notify(hash, "added".to_string());
```
tomusdrw: I think it would be better to have an enum like this:
```
#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
enum Status {
 Added,
 Rejected,
 ...
 Culled,
}
```
And the same status enum should be sent in the notification.

miner/src/pool/queue.rs
```diff
@@ -32,7 +33,7 @@ use pool::{
 };
 use pool::local_transactions::LocalTransactionsList;
 
-type Listener = (LocalTransactionsList, (listener::Notifier, listener::Logger));
+type Listener = (LocalTransactionsList, (listener::Notifier, (listener::Logger, listener::TransactionsPoolNotifier)));
```
tomusdrw: That starts to look pretty ugly, we could have a macro in `parity-common` to implement `Listener` for tuples of different sizes.
(Note it's not strictly related to this PR)

parity/rpc_apis.rs
```diff
@@ -106,6 +108,7 @@ impl FromStr for Api {
 			"signer" => Ok(Signer),
 			"traces" => Ok(Traces),
 			"web3" => Ok(Web3),
+			"transactions_pool" => Ok(TransactionsPool),
```
tomusdrw: Is this API specified anywhere? If not I'd prefer to keep it under `parity` prefix (`Parity` namespace should be fine)

parity/rpc_apis.rs
```diff
@@ -336,6 +340,14 @@ impl FullDependencies {
 						handler.extend_with(client.to_delegate());
 					}
 				}
+				Api::TransactionsPool => {
+					if !for_generic_pubsub {
+						let receiver = self.miner.get_tx_pool_receiver();
+						let client = TransactionsPoolClient::new(self.executor.clone());
+						client.run(receiver);
```
tomusdrw: Why not pass it in `new`? Now initialisation requires those two methods to be called precisely in order.

rpc/src/v1/impls/transactions_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// Transactions pool PubSub implementation.
+pub struct TransactionsPoolClient {
+    handler: Arc<TransactionsNotificationHandler>,
+    transactions_pool_subscribers: Arc<RwLock<Subscribers<Client>>>,
+}
+
+impl TransactionsPoolClient {
+    /// Creates new `TransactionsPoolClient`.
+    pub fn new(executor: Executor) -> Self {
```
tomusdrw: Please use tabs in this entire file.

rpc/src/v1/impls/transactions_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+        );
+
+        TransactionsPoolClient {
+            handler,
+            transactions_pool_subscribers,
+        }
+    }
+    pub fn run(&self, pool_receiver: mpsc::Receiver<(H256, String)>) {
+        let handler = self.handler.clone();
+        thread::spawn(move || loop {
```
tomusdrw: Can't we use future-aware channels instead to avoid spawning a new thread?

rpc/src/v1/impls/transactions_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+            let res = pool_receiver.recv();
+            if let Ok(res) = res {
+                handler.notify_transaction(res)
+            }
+        });
+    }
+
+}
+
+pub struct TransactionsNotificationHandler {
```
tomusdrw: Docs

rpc/src/v1/impls/transactions_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+        executor.spawn(subscriber
+            .notify(Ok(result))
+            .map(|_| ())
+            .map_err(|e| warn!(target: "rpc", "Unable to send notification: {}", e))
+        );
+    }
+
+    pub fn notify_transaction(&self, status: (H256, String)) {
+        for subscriber in self.transactions_pool_subscribers.read().values() {
+            let status = (status.0.clone(), status.1.clone());
```
tomusdrw:
```suggestion
            let status = status.clone();
```

or even inline it into the next line.

rpc/src/v1/traits/transactions_pool.rs
```diff
@@ -0,0 +1,20 @@
...
+
+/// Transactions Pool PUB-SUB rpc interface.
+#[rpc]
+pub trait TransactionsPool {
+    /// Pub/Sub Metadata
+    type Metadata;
+
+    /// Subscribe to Transactions Pool subscription.
+    #[pubsub(subscription = "parity_watchTransactionsPool", subscribe, name = "parity_watchTransactionsPool")]
+    fn subscribe(&self, Self::Metadata, typed::Subscriber<pubsub::Result>, Option<pubsub::Params>);
```
tomusdrw: If you don't want any params just leave this out:
```suggestion
    fn subscribe(&self, Self::Metadata, typed::Subscriber<pubsub::Result>);
```

rpc/src/v1/impls/transactions_pool.rs
```diff
@@ -0,0 +1,103 @@
...
+            Self::notify(&self.executor, subscriber, pubsub::Result::TransactionStatus(status));
+        }
+    }
+}
+
+impl TransactionsPool for TransactionsPoolClient {
+    type Metadata = Metadata;
+
+    fn subscribe(&self, _meta: Metadata, subscriber: Subscriber<pubsub::Result>, params: Option<pubsub::Params>) {
```
tomusdrw: using `pubsub::Result` here kind of suggests that we might get different values of this enum as well. I think it should be safe to just use the tuple directly:
```suggestion
    fn subscribe(&self, _meta: Metadata, subscriber: Subscriber<(H256, Status)>, params: Option<pubsub::Params>) {
```

miner/src/pool/listener.rs
```diff
@@ -116,6 +116,70 @@ impl txpool::Listener<Transaction> for Logger {
 	}
 }
 
+/// Transactions pool notifier
+#[derive(Default)]
+pub struct TransactionsPoolNotifier {
+	listeners: Vec<Arc<Mutex<mpsc::Sender<(H256, String)>>>>,
```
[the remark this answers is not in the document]
IntegralTeam: @tomusdrw If I delete the `Arc<Mutex<`, it leads to errors

```
   Compiling ethcore v1.12.0 (/home/user/Work/parity-ethereum/ethcore)
error[E0277]: `std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>` cannot be shared between threads safely
    --> ethcore/src/client/client.rs:1305:6
     |
1305 | impl snapshot::DatabaseRestore for Client {
     |      ^^^^^^^^^^^^^^^^^^^^^^^^^ `std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>` cannot be shared between threads safely
     |
     = help: the trait `std::marker::Sync` is not implemented for `std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>`
     = note: required because of the requirements on the impl of `std::marker::Sync` for `std::ptr::Unique<std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>>`
     = note: required because it appears within the type `alloc::raw_vec::RawVec<std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>>`
     = note: required because it appears within the type `std::vec::Vec<std::sync::mpsc::Sender<ethcore_miner::pool::listener::TxStatus>>`
     = note: required because it appears within the type `ethcore_miner::pool::listener::TransactionsPoolNotifier`
     = note: required because it appears within the type `(ethcore_miner::pool::listener::Logger, ethcore_miner::pool::listener::TransactionsPoolNotifier)`
     = note: required because it appears within the type `(ethcore_miner::pool::listener::Notifier, (ethcore_miner::pool::listener::Logger, ethcore_miner::pool::listener::TransactionsPoolNotifier))`
     = note: required because it appears within the type `(ethcore_miner::pool::local_transactions::LocalTransactionsList, (ethcore_miner::pool::listener::Notifier, (ethcore_miner::pool::listener::Logger, ethcore_miner::pool::listener::TransactionsPoolNotifier)))`
     = note: required because it appears within the type `transaction_pool::pool::Pool<ethcore_miner::pool::VerifiedTransaction, ethcore_miner::pool::scoring::NonceAndGasPrice, (ethcore_miner::pool::local_transactions::LocalTransactionsList, (ethcore_miner::pool::listener::Notifier, (ethcore_miner::pool::listener::Logger, ethcore_miner::pool::listener::TransactionsPoolNotifier)))>`
     = note: required because of the requirements on the impl of `std::marker::Sync` for `lock_api::rwlock::RwLock<parking_lot::RawRwLock, transaction_pool::pool::Pool<ethcore_miner::pool::VerifiedTransaction, ethcore_miner::pool::scoring::NonceAndGasPrice, (ethcore_miner::pool::local_transactions::LocalTransactionsList, (ethcore_miner::pool::listener::Notifier, (ethcore_miner::pool::listener::Logger, ethcore_miner::pool::listener::TransactionsPoolNotifier)))>>`
     = note: required because it appears within the type `ethcore_miner::pool::TransactionQueue`
     = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<ethcore_miner::pool::TransactionQueue>`
     = note: required because it appears within the type `miner::miner::Miner`
     = note: required because of the requirements on the impl of `std::marker::Sync` for `std::sync::Arc<miner::miner::Miner>`
     = note: required because it appears within the type `client::client::Importer`
     = note: required because it appears within the type `client::client::Client`
...
```
Found this: [Allowing Access from Multiple Threads with Sync](https://doc.rust-lang.org/stable/book/ch16-04-extensible-concurrency-sync-and-send.html#allowing-access-from-multiple-threads-with-sync)
And it seems to me that either Mutex is needed here, or I don't understand what am I doing wrong.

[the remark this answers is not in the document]
tomusdrw: Ah, I see. Yes, unfortunatelly we can't satisfy `Sync` with `std::mpsc::Sender`. To overcome this I'd go with `mpsc::unbounded_channel` from `futures` crate, this will also allow us to get rid of this extra thread responsible for polling the channel.

miner/src/pool/listener.rs
```diff
@@ -116,6 +117,69 @@ impl txpool::Listener<Transaction> for Logger {
...
+	}
+}
+
+impl txpool::Listener<Transaction> for TransactionsPoolNotifier {
+	fn added(&mut self, tx: &Arc<Transaction>, _old: Option<&Arc<Transaction>>) {
+		let hash = tx.hash.clone();
+		self.notify(hash, TxStatus::Added);
+	}
+
+	fn rejected(&mut self, tx: &Arc<Transaction>, _reason: &txpool::ErrorKind) {
```
tomusdrw: I just realised that this might be called really frequently, and currently we need to iterate over all senders and have a unique lock for this, so this will have drastic performance consequences.

Instead I think we should do something very similar that we do with `Notifer` in the same file, i.e:
1. We first aggregate all the notifications that should be sent
2. Then we send them all at once after batch import to the queue is finished.

It might be even worth to remove `Notifier` and convert this other notification to use this (channel-based) notifier instead.

```rust
struct Notifier {
  pending: HashMap<H256, TxStatus>,
  listeners: Vec<futures::sync::mpsc::UnboundedSender<Arc<HashMap<H256, TxStatus>>>>,
}

impl Notifier {
 fn notify(&mut self) {
   let to_send = Arc::new(std::mem::replace(&mut self.pending, HashMap::new));
   self.listeners.retain(|listener| listener.unbounded_send(to_send.clone()).is_ok());
 }
}

impl txpool::Listener<Transaction> for TransactionPoolNotifier {
  fn added(&mut self, ...) {
    self.pending.insert(tx.hash, TxStatus::Added);
  }
  ...
}
```

tomusdrw: @IntegralTeam Please let me know if you need any more explanations or help with this.

IntegralTeam: @tomusdrw Thank you, I've got what I need to do.

But, I think, it's not quite right to use `HashMap` here, because transaction firstly could be added with according `Added` status, and then before sending notification, same transaction could get status `Culled` and it will lead that `Added` status will be erased from `HashMap`

Maybe it will be better to use tuple vector `Vec<(H256, TxStatus)>`?

tomusdrw: @IntegralTeam good point, yeah if we never want to miss any event and have them in order, then `Vec` is indeed a good choice here.

IntegralTeam: Ok, let's do it this way.

miner/src/pool/queue.rs
```diff
@@ -303,6 +303,8 @@ impl TransactionQueue {
 		// Notify about imported transactions.
 		(self.pool.write().listener_mut().1).0.notify();
```
tomusdrw: This listener now seems obsolete, since it just duplicates the same data and is less efficient than the new channel-based one.

IntegralTeam: @tomusdrw 

> Looks nice now, I'd like to see the other `Notifier` rewritten to futures as well, but we can leave that as a separate PR - please create an issue for this if that happens to be the case.

Ok, then let`s do it in other PR

>This listener now seems obsolete, since it just duplicates the same data and is less efficient than the new channel-based one.

Maybe we can make one Notifier by combining old one `Notifier` and `TransactionsPoolNotifier` and store two different lists of listeners for `parity_watchTransactionsPool` and `eth_pubsub` in it?

IntegralTeam:
@tomusdrw Please review

ethcore/light/src/transaction_queue.rs
```diff
@@ -358,11 +364,22 @@ impl TransactionQueue {
...
 	/// Notifies all listeners about new pending transaction.
-	fn notify(&self, hashes: &[H256]) {
+	fn notify(&mut self, hashes: &[H256], status: TxStatus) {
 		for listener in &self.listeners {
 			listener(hashes)
 		}
+
+		let to_send: Vec<(H256, TxStatus)> = hashes.into_iter().map(|hash| (hash.clone(), status)).collect();
+
+		self.tx_statuses_listeners.retain(| listener| listener.unbounded_send(Arc::new(to_send.clone())).is_ok());
```
tomusdrw: That's incorrect, the reason we have `Arc` is to avoid cloning the entire `Vec`. So:
1. First create the `Arc`
2. Then clone `Arc` not `Vec`

IntegralTeam: @tomusdrw I fixed it

IntegralTeam:
@tomusdrw 
> Could you also create an issue to merge the two listeners?

Yesterday I already created an issuee https://github.com/paritytech/parity-ethereum/issues/10581
Can I change the title and description?

tomusdrw:
Perfect :ok_hand:

ethcore/light/src/transaction_queue.rs
```diff
@@ -358,11 +364,26 @@ impl TransactionQueue {
...
+	}
+
 	/// Notifies all listeners about new pending transaction.
-	fn notify(&self, hashes: &[H256]) {
+	fn notify(&mut self, hashes: &[H256], status: TxStatus) {
 		for listener in &self.listeners {
 			listener(hashes)
 		}
+
+		let to_send: Arc<Vec<(H256, TxStatus)>> = Arc::new(
```
seunlanlege: there's no need for `Arc` here.

`Vec<T>` implements `Clone` if `T: Clone`.

`T` here is a tuple, `(H256, TxStatus)` which also implements clone because `H256` and `TxStatus` implements `Clone`


Nice work overall

IntegralTeam: @seunlanlege Then here, too, do not need Arc?
https://github.com/paritytech/parity-ethereum/pull/10558/files/286ba53d13330a39a3fc66b2053e2ab9d5eb058c#diff-f34b6331b91a5d4893e05107b8ee0622R124
https://github.com/paritytech/parity-ethereum/pull/10558/files/286ba53d13330a39a3fc66b2053e2ab9d5eb058c#diff-f34b6331b91a5d4893e05107b8ee0622R140

tomusdrw: We do need `Arc`. There is a difference between cloning an `Arc` reference and the entire vector that might be huge.

seunlanlege: @tomusdrw yeah you're right. Didn't think of that.

PR #10559, opened 2019-04-02
fix(light cull): poll light cull instead of timer

niklasad1:
* Adds light transactions queue `cull polling` on every call on `eth_getTransactiobByHash` request
* Removes `cull service` which used a hardcoded timer every 10th minute


rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -523,6 +523,45 @@ where
 		}))
 	}
 
+	pub fn light_cull(&self, txq: Arc<RwLock<TransactionQueue>>) -> impl Future <Item = (), Error = Error> + Send {
+		let senders = txq.read().queued_senders();
+		if senders.is_empty() {
+			return Either::B(future::err(errors::no_light_transactions()));
```
niklasad1: This is very unlikely to happen but it could happen if there is a race with calls to `eth_getTransactionByHash` i.e, a `cull` finished just between the transaction_by_hash and light_cull.

Then don't do a needless network request

rpc/src/v1/impls/light/eth.rs
```diff
@@ -420,15 +420,24 @@ where
...
+		let in_txqueue = self.transaction_queue.read().get(&hash).is_some();
+
+		// The transaction is in the `local txqueue`
+		// Check if the nonce on the accounts are still fresh and return result from
+		// the `local txqueue` if it is still fresh.
+		// Otherwise make a network request to fetch the transaction from the network
+		if in_txqueue {
+			// Note, this will block (relies on HTTP timeout) to make sure `cull` will finish to avoid having to call
+			// `eth_getTransactionByHash` more than once to ensure the `txqueue` is up to `date` when it is called
+			if let Err(e) = self.fetcher().light_cull(self.transaction_queue.clone()).wait() {
```
niklasad1: This is probably the most controversial thing about this PR.

The `light_cull` could be spawned in an executor but then we would most likely not get result in this call (accessing an atomic variable is much faster than a network request) and we would have to call this at least one more time to get the result!

rpc/src/v1/helpers/errors.rs
```diff
@@ -57,6 +57,7 @@ mod codes {
 	pub const FETCH_ERROR: i64 = -32060;
 	pub const NO_LIGHT_PEERS: i64 = -32065;
 	pub const NO_PEERS: i64 = -32066;
+	pub const NO_LIGHT_TRANSACTIONS: i64 = -32066;
```
HCastano: I don't think this error code should be the same as the `NO_PEERS` one

niklasad1: Thanks, yeah that's copy-paste error forgot the change the actual error code

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -523,6 +523,45 @@ where
...
+
+		let account_request = sync.with_context(move |ctx| {
+			// fetch the nonce of each sender in the queue.
+			let nonce_reqs = senders.iter()
+				.map(|&address| request::Account { header: best_header.clone().into(), address })
+				.collect::<Vec<_>>();
+
+			// when they come in, update each sender to the new nonce.
+			on_demand.request(ctx, nonce_reqs)
+				.expect("No back-references; therefore all back-references are valid; qed")
```
HCastano: I don't understand this, can you explain it?

rpc/src/v1/helpers/light_fetch.rs
```diff
@@ -523,6 +523,45 @@ where
 		}))
 	}
 
+	pub fn light_cull(&self, txq: Arc<RwLock<TransactionQueue>>) -> impl Future <Item = (), Error = Error> + Send {
```
HCastano: Docs on public methods

rpc/src/v1/helpers/light_fetch.rs
HCastano: I think you should leave this is a comment within the code, just in case

rpc/src/v1/impls/light/eth.rs
```diff
@@ -420,15 +420,24 @@ where
...
 
 	fn transaction_by_hash(&self, hash: H256) -> BoxFuture<Option<Transaction>> {
-		{
-			let tx_queue = self.transaction_queue.read();
-			if let Some(tx) = tx_queue.get(&hash) {
+		let in_txqueue = self.transaction_queue.read().get(&hash).is_some();
+
+		// The transaction is in the `local txqueue`
+		// Check if the nonce on the accounts are still fresh and return result from
+		// the `local txqueue` if it is still fresh.
```
ordian: If `light_cull` call fails, we still return result from the `local txqueue`.  Should we change or document that?

niklasad1: I think it should return the local transaction if that's the case I tried to change to docs let me know if it is clear enough

ordian: I think the code is quite self-explanatory and the `If it fails return the local transaction` is slightly misleading (may read as we return the local transaction only in this case), so I suggest to remove this part.

rpc/src/v1/helpers/light_fetch.rs
niklasad1: It is a `OnDemand` thing where you can make a `Request` that depends on `back-reference` which is a way to make request until the header is resolved

But the request doesn't have a back-reference because the header is known

PR #10562, opened 2019-04-02
Update Issue Template to direct security issue to email

kirushik:


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,10 +1,12 @@
 _Before filing a new issue, please **provide the following information**._
 
+_If you think that your issue is an exploitable security vulnerability, please mail your bugreport to security@parity.io instead._
```
soc1c: maybe link `SECURITY.md`?

kirushik: Oh, great idea, will do (and will also revisit `SECURITY.md` to make sure it's up-to-date)

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,10 +1,14 @@
 _Before filing a new issue, please **provide the following information**._
 
+_If you think that your issue is an exploitable security vulnerability, please mail your bugreport to security@parity.io instead; your submission might be eligible for our Bug Bounty._
+_You can find mode info on the reporting process in https://github.com/paritytech/parity-ethereum/blob/master/SECURITY.md _
```
soc1c:
```suggestion
_You can find mode info on the reporting process in [SECURITY.md](https://github.com/paritytech/parity-ethereum/blob/master/SECURITY.md)_
```

kirushik: But people will see it in an editable textbox without any markdown formatting when opening a new GitHub issue.
Won't this change make it only harder to copy-paste the full SECURITY.md URL?

(But yes, I get that the counterpoint is that we probably want to reduce noise in the submitted issues, and people won't remove anything from this issue template when submitting...)

kirushik: Actually, I think we're spending too much effort on this now; I'll merge it as it is now.

PR #10566, opened 2019-04-04
Stop breaking out of loop if a non-canonical hash is found

HCastano:
Currently investigating #10085 with the help of @tomusdrw. We have a couple theories floating around, but we aren't sure how to reproduce them on our own.

We've narrowed down the cause to one of three function calls within `Blockchain::epoch_transition_for`. This PR handles the `None` case of what we think is the most suspect function call better, as well as adds some additional logging so we can track where we are the loop.

### DIFF
--- a/ethcore/blockchain/src/blockchain.rs
+++ b/ethcore/blockchain/src/blockchain.rs
@@ -970,6 +970,7 @@ impl BlockChain {
 	/// Iterate over all epoch transitions.
 	/// This will only return transitions within the canonical chain.
 	pub fn epoch_transitions(&self) -> EpochTransitionIter {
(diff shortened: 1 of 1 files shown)


HCastano:
I don't think we need _all_ the logging added here, but I think some additions in this area of the code wouldn't hurt (especially if they're at the `trace` level).

dvdplm:
@HCastano @soc1c What is the plan for this PR? To merge it or is it used for debugging only? If the latter, perhaps we can close it and leave the branch for those who need it?

HCastano:
@dvdplm Hey, sorry for the late reply, I haven't been checking Github all that often recently. It's basically just for debugging. It didn't end up solving the issue, but did provide some extra logs to look at. I think its fine if we close it without deleting it.

PR #10579, opened 2019-04-08
CI improvements

TriplEight:
- Rust nightly and beta tests are back and shiny, will run during the nightly build
- all test and build jobs changed to run with `--verbose --color=always`
- max cache size raised to 50 Gib
- log files are collected for linux jobs and uploaded to GitLab artifacts. But I didn't figure out how to have two `artifacts:` sections in one job, so for `build` stage logs will still be uploaded, but with the `&collect_artifacts` rule (`on_success` and `1 mos`).

> So I was testing pipelines with per-job `sccache` caching, thinking it was a last resort.
It turned out it was not necessary to resort to such extreme measures:
Random compilation failure bug seems to appear only when the max cache size allocated is exceeded, and fail happens when one job removes an arbitrary artifact and then the next job wants to remove the same one, but doesn't find it.

Another problem I thought was connected with caching, some tests were failing. At first I thought it's due to caching, but per-job isolation showed that it's not the caching problem.

As a bonus, we (with Niklas help) have found 2 flaky tests.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,7 +70,7 @@ variables:
         exit 0;
       fi
     # collect log files
-    - mkdir -p ./artifacts
+    - mkdir -p ./logs
     - find . -name "*.log"
     - find . -name "*.log" | xargs tar --append -f ./logs/logs_"${CI_JOB_NAME}_${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}".tar
   tags:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,7 +74,7 @@ cargo-check 0 3:
   <<:                              *docker-cache-status
   <<:                              *collect_logs
   script:
-    - time cargo check --target $CARGO_TARGET --locked --no-default-features --verbose --color=auto
+    - time cargo check --target $CARGO_TARGET --locked --no-default-features --verbose --color=always
     - sccache -s
 
 cargo-check 1 3:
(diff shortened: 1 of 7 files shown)


.gitlab-ci.yml
```diff
@@ -31,116 +31,164 @@ variables:
...
+    - sccache -s
+
 
 cargo-check 0 3:
   stage:                           test
   <<:                              *docker-cache-status
+  <<:                              *collect_logs
   script:
-    - time cargo check --target $CARGO_TARGET --locked --no-default-features
+    - time cargo check --target $CARGO_TARGET --locked --no-default-features --verbose --color=always
```
gabreal: why is it necessary to use `--color=always`? Doesn't cargo decide that automatically e.g. there are cases where one wouldn't want that like when output is redirected to a file.

.gitlab-ci.yml
```diff
@@ -31,116 +31,164 @@ variables:
...
   tags:
     - linux-docker
 
+.build-on-linux:                   &build-on-linux
+  stage:                           build
+  <<:                              *docker-cache-status
+  <<:                              *collect_artifacts
+  script:
+    - scripts/gitlab/build-linux.sh
+    - sccache -s
```
gabreal: why not put `sccache -s` in the `after_script` of `docker-cache-status`?

.gitlab-ci.yml
```diff
@@ -31,116 +31,164 @@ variables:
...
     - sccache -s
 
 cargo-check 1 3:
   stage:                           test
   <<:                              *docker-cache-status
+  <<:                              *collect_logs
   script:
-    - time cargo check --target $CARGO_TARGET --locked --manifest-path util/io/Cargo.toml --no-default-features
+    - time cargo check --target $CARGO_TARGET --locked --manifest-path util/io/Cargo.toml --no-default-features --verbose --color=always
     - sccache -s
```
gabreal: see above and below

.gitlab-ci.yml
TriplEight: with `--verbose` it stops working, so to not to lose readability i had to add it.

.gitlab-ci.yml
TriplEight: unfortunately, gitlab kills all the processes with the end of `script:` section and that way `sccache -s` does not show stats from `after_script` section.

TriplEight:
> looks also good to me. so there won't be tests for mac or win?

@gabreal this is not decided yet, and will require changes on hosts.

PR #10584, opened 2019-04-10
Allow CORS requests in Secret Store API

adetante:
On the same model as what is done for JSON-RPC, this PR allows to specify the list of domains to allow for cross-domain requests to Secret Store API.

With CLI flag ` --secretstore-http-cors=http://mydomain` or in the toml configuration file.

The default behavior is the same as the current one: all cross-domain requests are rejected.

### DIFF
--- a/secret-store/src/listener/http_listener.rs
+++ b/secret-store/src/listener/http_listener.rs
@@ -446,7 +446,7 @@ mod tests {
 		let key_server: Arc<KeyServer> = Arc::new(DummyKeyServer::default());
 		let address = NodeAddress { address: "127.0.0.1".into(), port: 9000 };
 		let runtime = Runtime::with_thread_count(1);
-		let listener = KeyServerHttpListener::start(address, Arc::downgrade(&key_server),
+		let listener = KeyServerHttpListener::start(address, None, Arc::downgrade(&key_server),
 			runtime.executor()).unwrap();
 		drop(listener);
 	}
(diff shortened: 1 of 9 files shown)


secret-store/src/listener/http_listener.rs
```diff
@@ -204,6 +211,24 @@ impl KeyServerHttpHandler {
 			},
 		}
 	}
+
+	fn cors(&self, req: &HttpRequest<Body>) -> Cors {
```
svyatonik: Could you, please, remove trailing tabs?

secret-store/src/listener/http_listener.rs
```diff
@@ -204,6 +211,24 @@ impl KeyServerHttpHandler {
 			},
 		}
 	}
+
+	fn cors(&self, req: &HttpRequest<Body>) -> Cors {		
+		match req.headers().get(header::ORIGIN) {
+			None => Cors::CorsAuthorized(None),
+			Some(origin) => {
+				let is_authorized = self.handler.cors.clone()
```
svyatonik:
```suggestion
				let is_authorized = self.handler.cors.as_ref()
```

secret-store/src/listener/http_listener.rs
```diff
@@ -213,69 +238,75 @@ impl Service for KeyServerHttpHandler {
...
+						warn!(target: "secretstore", "Ignoring invalid {}-request {}", req_method, req_uri);
+						HttpResponse::builder()
+							.status(HttpStatusCode::NOT_FOUND)
+							.body(Body::empty())
+							.expect("Nothing to parse, cannot fail; qed")
+					}
+				}))
 			}
-		}))
+		}
```
svyatonik: Trailing tabs

adetante:
You're right: I'm going to change my implementation to the jsonrpc-server-utils crate.
Thank you for the feedback.

adetante:
I updated based on your comments.
Thank you for reviewing

svyatonik:
Test failure seems unrelated - SS tests are working fine on my laptop.

PR #10587, opened 2019-04-11
fix(whisper): change expiry `unix_time + ttl + work`

niklasad1:
* Fix a bug introduced by #10383, subtraction instead of addition
* Fix another bug, expiry =  `time + ttl + work` instead of `work factor` to calculate expiry (https://github.com/ethereum/wiki/wiki/Whisper-PoC-2-Protocol-Spec#inserting-authoring-messages)

### DIFF
--- a/whisper/src/message.rs
+++ b/whisper/src/message.rs
@@ -264,17 +264,17 @@ impl Message {
 
 		let mut rng = {
 			let mut thread_rng = ::rand::thread_rng();
-
 			XorShiftRng::from_seed(thread_rng.gen::<[u32; 4]>())
 		};
 
 		assert!(params.ttl > 0);
 
 		let expiry = {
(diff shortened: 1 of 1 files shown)


whisper/src/message.rs
```diff
@@ -271,7 +271,7 @@ impl Message {
 		assert!(params.ttl > 0);
 
 		let expiry = {
-			let after_mining = SystemTime::now().checked_sub(Duration::from_millis(params.work))
+			let after_mining = SystemTime::now().checked_add(Duration::from_secs(params.ttl))
```
rphmeier: Should probably be `SystemTime::now() + params.work + params.ttl`? And the var name should be changed to reflect meaning of code

niklasad1: Yes, agree

PR #10588, opened 2019-04-12
Configs for CI targets

General-Beck:
This change puts together all the config files for cross-compilations for armhf, arm64 and android.
PR https://github.com/paritytech/scripts/pull/72 depends of this one.


TriplEight:
It was decided this change is not needed.

Issue #10603, opened 2019-04-21
Clique sealing failes with "Extra block data too long."


soc1c:
- **Parity Ethereum version**: 2.5.0
- **Operating system**: Linux
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: kotti
- **Restarted**: yes

```
10|parity  | 2019-04-21 10:26:54 UTC Starting Parity-Ethereum/v2.5.0-beta-b52ac20-20190408/x86_64-linux-gnu/rustc1.33.0
10|parity  | 2019-04-21 10:26:54 UTC Keys path /home/parity/.local/share/io.parity.ethereum/keys/kotti
10|parity  | 2019-04-21 10:26:54 UTC DB path /home/parity/.local/share/io.parity.ethereum/chains/kotti/db/717b12a127188bca
10|parity  | 2019-04-21 10:26:54 UTC State DB configuration: fast
10|parity  | 2019-04-21 10:26:54 UTC Operating mode: active
10|parity  | 2019-04-21 10:26:55 UTC Configured for Kotti Testnet using Clique engine

10|parity  | 2019-04-21 10:26:58 UTC Listening for new connections on 127.0.0.1:8546.
10|parity  | 2019-04-21 10:27:03 UTC Public node URL: enode://06333009fc9ef3c9e174768e495722a7f98fe7afd4660542e983005f85e556028410fd03278944f44cfe5437b1750b5e6bd1738f700fe7da3626d52010d2954c@10.0.0.4:30303
10|parity  | 2019-04-21 10:27:04 UTC ERROR: seal failed when given internally generated seal: Block error (Extra block data too long. Value 31 out of bounds. Min=32, Max=32)
10|parity  | 2019-04-21 10:27:06 UTC ERROR: seal failed when given internally generated seal: Block error (Extra block data too long. Value 31 out of bounds. Min=32, Max=32)
10|parity  | 2019-04-21 10:27:08 UTC ERROR: seal failed when given internally generated seal: Block error (Extra block data too long. Value 31 out of bounds. Min=32, Max=32)
10|parity  | 2019-04-21 10:27:10 UTC ERROR: seal failed when given internally generated seal: Block error (Extra block data too long. Value 31 out of bounds. Min=32, Max=32)
10|parity  | 2019-04-21 10:27:12 UTC ERROR: seal failed when given internally generated seal: Block error (Extra block data too long. Value 31 out of bounds. Min=32, Max=32)
[...]

PR #10605, opened 2019-04-22
Clique: zero-fill extradata when the supplied value is less than 32 bytes in length

jwasinger:
Attempt at a fix for https://github.com/paritytech/parity-ethereum/issues/10603 .  haven't verified yet.

### DIFF
--- a/ethcore/src/engines/clique/mod.rs
+++ b/ethcore/src/engines/clique/mod.rs
@@ -67,6 +67,7 @@ use std::time;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use block::ExecutedBlock;
+use bytes::Bytes;
 use client::{BlockId, EngineClient};
 use engines::clique::util::{extract_signers, recover_creator};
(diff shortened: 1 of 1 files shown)


ethcore/src/engines/clique/mod.rs
```diff
@@ -713,6 +714,14 @@ impl Engine<EthereumMachine> for Clique {
 						header.set_difficulty(DIFF_NOTURN);
 					}
 				}
+
+				let zero_padding_len = VANITY_LENGTH - header.extra_data().len();
+				if zero_padding_len > 0 {
```
niklasad1: Looks ok to me but this piece of code code be simplified by using `Vec::resize`

Something like 
```rust
let mut extra_data = header.extra_data().clone();
extra_data.resize(VANITY_LENGTH, 0);
header.set_extra_data(new_extra_data);
```

jwasinger:
Verified locally that it solves https://github.com/paritytech/parity-ethereum/issues/10603

PR #10608, opened 2019-04-23
adds rpc error message for --no-ancient-blocks

seunlanlege:


rpc/src/v1/helpers/errors.rs
```diff
@@ -221,18 +222,34 @@ pub fn cannot_submit_work(err: EthcoreError) -> Error {
...
+		Error {
+			code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST),
+			message: "Block information is incomplete while ancient block sync is still in progress, before \
+					it's finished we can't determine the existence of requested item.".into(),
+			data: None,
+		}
+	} else {
+		Error {
+			code: ErrorCode::ServerError(codes::UNSUPPORTED_REQUEST),
+			message: "Ancient block sync is still in progress".into(),
```
tomusdrw: Might be good to improve this message as well.
```suggestion
			message: "Requested block number is in a range that is not available yet, because the ancient block sync is still in progress.".into(),
```

parity/run.rs
```diff
@@ -745,6 +745,7 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
 		gas_price_percentile: cmd.gas_price_percentile,
 		poll_lifetime: cmd.poll_lifetime,
 		allow_missing_blocks: cmd.allow_missing_blocks,
+		no_ancient_blocks: cmd.download_old_blocks,
```
tomusdrw: shouldn't this be the other way around?
```suggestion
		no_ancient_blocks: !cmd.download_old_blocks,
```

ordian: Was going to suggest the same :)

seunlanlege: actually you're right, good catch. :sweat_smile:

gmajoulet:
Awesome change, thanks!

Issue #10302, opened 2019-02-05
Replace `error-chain`


niklasad1:
`error-chain` is not maintained anymore and depends on [Error::cause](https://doc.rust-lang.org/std/error/trait.Error.html#method.cause) which will be deprecated in Rust 1.33.

I don't know what the best solution to this but what I can think of:
- Manually implement it
- Migrate to `failure`

ascjones:
In #10510 I manually implemented `Error` and the `From` impls, but used the `derive_more` `Display` derive for that boilerplate. @niklasad1 let me know what you think of this approach and we could apply to other errors.

Here's one of the regexes I used for the variants inside `error_chain {}` for future use:
```
\#\[doc = \"(.+)\"]\n\s+(.*) \{\n.*\n.*display\(\"(.*)\".*\n.*\}\n

/// $1\n\t\t#[display(fmt = "$3", _0)]\n\t\t$2,
```

niklasad1:
nice work, yeah looks good to me.

could make sense to write a macro for the `fn source impl`:

```rust
impl error::Error for Error {
	fn source(&self) -> Option<&(error::Error + 'static)> {
		match self {
			Error::Io(e) => Some(e),
	                // more patterns
			_ => None,
		}
	}
}
```

ascjones:
At first I converted to using [err-derive](https://crates.io/crates/err-derive) which has essentially just copied `failure` macros and allows annotating nested errors with `#[cause]` for this purpose.

However I decided to user the more general and popular `derive_more`, rather than introducing another error library.

So we could just copy that macro from there or consider adding a dependency to it. In the meantime I figured it's not that bad to implement `source` manually.

niklasad1:
Ok, I see. 

Let's go with a manual implementation for now and see how much `extra code` it will require.

PR #10616, opened 2019-04-29
[ethcore] remove error_chain

ascjones:
rel: #10302 

Replaces `error_chain` with manual `Error` impl and derived `Display`.


Tokens001:
Please provide my address no body  can control

PR #10624, opened 2019-05-03
Updated Bn128PairingImpl to use optimized batch pairing

zac-williamson:
Hello!  

This is a small change to the `Bn128PairingImpl` method that takes advantage of the new `pairing_batch` method in the `bn` crate (added in [commit 67df95d06683b99d4308af20f9adc31497afcece](https://github.com/paritytech/bn/commit/67df95d06683b99d4308af20f9adc31497afcece).  

This provides an approximate 2x increase in the speed of the algorithm, when multiple pairings are being computed at once.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,11 +264,13 @@ dependencies = [
 [[package]]
 name = "bn"
 version = "0.4.4"
-source = "git+https://github.com/paritytech/bn#2a71dbde5ca93451c8da2135767896a64483759e"
+source = "git+https://github.com/paritytech/bn#6beba2ed6c9351622f9e948ccee4063846b2b39a"
 dependencies = [
  "byteorder 1.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "crunchy 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


sorpaas:
I think this needs to run `cargo update -p bn` so that https://github.com/paritytech/bn/pull/13 is included.

dvdplm:
@zac-williamson can you please rebase this so we can get passed the `cargo audit` failure? Thanks!

dvdplm:
> This provides an approximate 2x increase in the speed

@zac-williamson did you write any benchmarks for this? It'd be great to include those in the PR too.

zac-williamson:
> @zac-williamson can you please rebase this so we can get passed the `cargo audit` failure? Thanks!

Heya, sorry for the very slow reply, fixing this now. Will also post benchmarks

dvdplm:
@zac-williamson let us know if you want us to take over! :)

zac-williamson:
> @zac-williamson let us know if you want us to take over! :)

Heya! Sorry for not being responsive, I'm afraid other commitments have been preoccupying my time lately.

I ran the full parity benchmark suite and uploaded it here: https://gist.github.com/zac-williamson/71b67b994c8b7be4de263fe31dc2eb21 , still need to format it however.

I've also pushed a commit that should have updated the PR to use the latest version of BN. I notice that the cargo-check test is failing; I'm afraid Rust is not a language I'm very familiar with - do you know what work needs to be done to resolve this?

Many thanks,
Zac

dvdplm:
> do you know what work needs to be done to resolve this?

I think a rebase should fix the cargo-audit error. :)

dvdplm:
Closing in favour of #10765

PR #10638, opened 2019-05-08
ci: publish docs debug

gabreal:


scripts/gitlab/publish-docs.sh
```diff
@@ -51,8 +51,11 @@ commit_files() {
 
 upload_files() {
     echo "__________Upload files__________"
-    git push -q origin HEAD
-    git push -q -f --tags
+    # this version of git (2.7.4) will dump the token on failure
```
kirushik: Why do we even use token instead of a deploy SSH key?

scripts/gitlab/publish-docs.sh
```diff
@@ -51,8 +51,11 @@ commit_files() {
...
+    git push -q origin HEAD 2>&1 \
+      | sed -r "s|(${GITHUB_USER}):[a-f0-9]+@|\1:REDACTED@|g"
```
kirushik: I find it possible that git will change the failure format and this regexp will stop redacting the secret...
And this failure will be terribly hard to notice in advance...

gabreal: It's the most  simple pattern `username:token`

TriplEight:
LGTM, also I'd vote moving to gitlab/github pages later

gabreal:
thanks, I don't mind at all moving to something else when there is time to plan.

soc1c:
labels!

PR #10644, opened 2019-05-10
Update publishing

TriplEight:
- removed the listing of sccache debug log
- docker images with parity-ethereum is now being built by k8s
- Dockerfile was optimized and lost `openssl`
- for now, there are `docker tag`s and `docker push`es for the all cases possible, to be decided later which do we need
- publishing of the docker image should not run nightly
- stopped onchain updates for Kovan
- new artifacts dir for scripts: `tools`
-- less jobs require git clone and submodules
- show rust version for docker builds

Tests are OK, but I'm afraid `publish-snap` and `publish-onchain` can break.


.gitlab-ci.yml
```diff
@@ -141,6 +141,8 @@ build-android:
 build-linux:
   <<:                              *build-on-linux
   only:                            *releaseable_branches
+  after_script:
+    - cp -r scripts/docker/hub/* ./artifacts/
```
TriplEight: I'm a bit worried if `Dockerfile` and `check-sync.sh` will be continuously published from now on

.gitlab-ci.yml
```diff
@@ -192,14 +194,56 @@ build-windows:
...
     - build-linux
   tags:
-    - shell
+    - kubernetes-parity-build
+  environment:
+    name: parity-build
+  cache:                           {}
+  only:                            *releaseable_branches
+  except:
+    - master
```
TriplEight: This is supposed to mean it shouldn't run on master to not to publish nightly images.

.gitlab-ci.yml
```diff
@@ -192,14 +194,56 @@ build-windows:
...
+          docker push ${CONTAINER_IMAGE}:latest ;;
+        stable|*)
+          echo "Docker TAGs - ${CONTAINER_IMAGE}:${CI_COMMIT_REF_NAME}";
+          docker build --no-cache
+              --build-arg VCS_REF="${CI_COMMIT_SHORT_SHA}"
+              --build-arg BUILD_DATE="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+              --tag ${CONTAINER_IMAGE}:$CI_COMMIT_REF_NAME
+              --file artifacts/Dockerfile .;
+          docker push ${CONTAINER_IMAGE}:$CI_COMMIT_REF_NAME ;;
+      esac
```
TriplEight: I'm a bit worried this script came so big, but didn't figure out how to make it smaller.

.gitlab-ci.yml
```diff
@@ -192,14 +194,56 @@ build-windows:
...
+          docker push ${CONTAINER_IMAGE}:beta;
+          docker push ${CONTAINER_IMAGE}:latest ;;
+        stable|*)
+          echo "Docker TAGs - ${CONTAINER_IMAGE}:${CI_COMMIT_REF_NAME}";
+          docker build --no-cache
+              --build-arg VCS_REF="${CI_COMMIT_SHORT_SHA}"
+              --build-arg BUILD_DATE="$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+              --tag ${CONTAINER_IMAGE}:$CI_COMMIT_REF_NAME
+              --file artifacts/Dockerfile .;
+          docker push ${CONTAINER_IMAGE}:$CI_COMMIT_REF_NAME ;;
```
TriplEight: maybe it makes sense to test the image before the push?

scripts/docker/hub/Dockerfile
```diff
@@ -1,27 +1,40 @@
...
+	groupadd -g 1000 parity; \
+	useradd -m -u 1000 -g parity -s /bin/sh parity
 
 WORKDIR /home/parity
 
-# add parity-ethereum to docker image
+# add parity-ethereum binary to docker image
 COPY artifacts/x86_64-unknown-linux-gnu/parity /bin/parity
 
-COPY scripts/docker/hub/check_sync.sh /check_sync.sh
```
gabreal: I don't know who put it there, but is it not needed anymore?

.gitlab-ci.yml
gabreal: files can be synced explicitly for publishing

.gitlab-ci.yml
gabreal: releasable_branches contains no master - so it will already not run on commits to the master branch.

.gitlab-ci.yml
```diff
@@ -192,14 +194,56 @@ build-windows:
...
+    - test "$Docker_Hub_User_Parity" -a "$Docker_Hub_Pass_Parity"
+        || ( echo "no docker credentials provided"; exit 1 )
+    - docker login -u "$Docker_Hub_User_Parity" -p "$Docker_Hub_Pass_Parity"
+    - docker info
   script:
-    - scripts/gitlab/publish-docker.sh parity
+    # we stopped pushing nightlies to dockerhub, will push to own registry prb.
+    - case "${SCHEDULE_TAG:-${CI_COMMIT_REF_NAME}}" in
+        $SCHEDULE_TAG)
+          echo "Docker TAG - ${CONTAINER_IMAGE}:${SCHEDULE_TAG}";;
```
gabreal: so no docker build for schedules with $SCHEDULE_TAG like nightly?

.gitlab-ci.yml
gabreal: I would suggest to put a test into the Dockerfile as in the substrate dockerfile.

.gitlab-ci.yml
gabreal: so what's the reason for including it into the ci file? don't want to transport the script via artifacts?

.gitlab-ci.yml
TriplEight: Yes, that's the idea. Most probably they are not needed to anyone.
Maybe we'll reintroduce nightly builds later, when our own repository will start working.

.gitlab-ci.yml
TriplEight: good point

.gitlab-ci.yml
TriplEight: indeed, thanks

.gitlab-ci.yml
TriplEight: what do you mean?

scripts/docker/hub/Dockerfile
TriplEight: returned, it was used somewhere, thanks.

.gitlab-ci.yml
TriplEight: changed to avoid just `nightly` schedule

.gitlab-ci.yml
gabreal: I just wanted to mention that if you don't want to publish them (doesn't make so much sense), you can specify the files explicitly that should be published i.e. either remove them from the artifacts directory later before syncing to s3 or copying only files to s3 that are needed.

TriplEight: indeed, will do it, thx

scripts/docker/hub/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
...
+echo ""
+echo "At some point in the future we will stop pushing new versions of 'parity/parity'."
+echo "Please use 'parity/ethereum' image instead, it is the new canonical location for "
+echo "'parity/parity' images."
+echo ""
+echo "We maintain the same set of tags there, so just replacing the image name"
+echo "everywhere would be safe."
+echo "#################################################################################"
+
+/bin/parity $@
```
gabreal: the `entrypoint.sh` was obsoleted before, why bring it back?
--- and if wasn't it `exec /bin/parity $@` to make parity the main process of the image?

gabreal: Why is the image now called `ethereum` instead of `parity-ethereum` like the repository?

gabreal: Concerning my first comment, afair the entrypoint script was introduced to have a variable binary. Like this there is really no benefit.

scripts/docker/hub/publish-docker.sh
```diff
@@ -0,0 +1,100 @@
...
+        docker build --no-cache \
+            --build-arg FROM_VERSION="${VERSION}-${CI_COMMIT_REF_NAME}" \
+            --build-arg VCS_REF="${CI_COMMIT_SHA}" \
+            --build-arg BUILD_DATE="$(date -u '+%Y-%m-%dT%H:%M:%SZ')" \
+            --tag "parity/parity:${VERSION}-${CI_COMMIT_REF_NAME}" \
+            --tag "parity/parity:stable" \
+            --file tools/parity.Dockerfile .;
+        docker push "parity/parity:${VERSION}-${CI_COMMIT_REF_NAME}";
+        docker push "parity/parity:stable";;
+    *)
```
gabreal: this would be the tag released versions then, right? don't you want to use the version and only the version for the tag?

TriplEight: tags and PR numbers will go there. So it will be like `parity/ethereum:2.5.0-beta` or `parity/ethereum:2.5.0-pr-10644`. Least is actually not desired to end up in registry, but for the testing purposes it'll be there for a while. That's why I covered all the cases.

scripts/docker/hub/entrypoint.sh
TriplEight: - just to send this deprecation message
- thanks, I'll check it then
- we decided with @kirushik that why not, it's still in Parity's repo, which makes it `parity/ethereum`
- since `COPY` instruction in the Dockerfile copies the concrete file, `$@` just take the arguments. Is there a way to do it better?

gabreal: That's the original `entrypoint.sh`:
https://github.com/paritytech/parity-ethereum/compare/v2.3.7...master#diff-c346274d88c3013298b1e18cbb88a149

gabreal: ah, you were asking for a better way of printing this message. well then still you can use `exec`.

gabreal: >     * we decided with @kirushik that why not, it's still in Parity's repo, which makes it `parity/ethereum`

so the repository `parity/parity` was renamed to `parity/parity-ethereum` because that was the new name of the software. what is the reason not to do the same for the dockerfile?

TriplEight: fixed and simplified, thanks.

scripts/docker/hub/parity.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+	io.parity.image.source="https://github.com/paritytech/parity-ethereum/blob/${VCS_REF}/\
+scripts/docker/hub/Dockerfile" \
+	io.parity.image.documentation="https://wiki.parity.io/Parity-Ethereum" \
+	io.parity.image.revision="${VCS_REF}" \
+	io.parity.image.created="${BUILD_DATE}"
+
+COPY tools/entrypoint.sh /entrypoint.sh
+
+# check if executable works in this container
+RUN parity --version
```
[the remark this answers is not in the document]
TriplEight: In this process there's no reason doing that, removed, thanks.

scripts/docker/hub/entrypoint.sh
TriplEight: It was made to be more clear. `ethereum` instead of `parity-ethereum` is only because it stays in `parity` docker namespace.
It will be better than `parity/parity` since we age going to publish more clients there: bitcoin, zcash and whatnot.

PR #10649, opened 2019-05-13
fix(compilation warnings)

niklasad1:


whisper/src/rpc/crypto.rs
```diff
@@ -77,7 +77,7 @@ impl EncryptionInstance {
 			EncryptionInner::AES(key, nonce, encode) => {
 				match encode {
 					AesEncode::AppendedNonce => {
-						let mut enc = Encryptor::aes_256_gcm(&*key).ok()?;
+						let enc = Encryptor::aes_256_gcm(&*key).ok()?;
```
ordian: This was fixed in https://github.com/paritytech/parity-ethereum/pull/10647/files#diff-5d5bfba94990222067b07edc3b22c811R80

niklasad1: rebased

PR #10650, opened 2019-05-13
Upgrade to parity-crypto 0.4

dvdplm:
Reverts using NonZeroU32 (introduced [here](https://github.com/paritytech/parity-ethereum/commit/b347599cf75fefd9fcf9be2b6ba6e8305f8e79ff)).

Ref. https://github.com/paritytech/parity-common/pull/149


parity/configuration.rs
```diff
@@ -144,8 +143,6 @@ impl Configuration {
 		let ipfs_conf = self.ipfs_config();
 		let secretstore_conf = self.secretstore_config()?;
 		let format = self.format()?;
-		let keys_iterations = NonZeroU32::new(self.args.arg_keys_iterations)
```
ordian: I think we should keep the check if it's zero and return an error in that case.

Issue #228, opened 2016-01-26
Ethash memory management


arkpar:
Don't keep unneeded cache items in memory. Limit size to 2

Issue #1337, opened 2016-06-19
Docker brson/multirust is deprecated


kartojal:
Your docker images and Dockerfiles are using a **deprecated** Rust installation https://raw.githubusercontent.com/brson/multirust/master/blastoff.sh , from [brson/multirust](https://github.com/brson/multirust) .

In their repo  [brson/multirust](https://github.com/brson/multirust), they have the following note at the readme.md: 

> Note: multirust is not actively maintained. Migrate to rustup.rs ([help](https://github.com/brson/multirust#migrating-to-rustup)).

Now it should be: `RUN curl https://sh.rustup.rs -sSf | sh -s -- -y ` for a silent installation, and add `/root/.cargo/bin` to the $PATH. Then you can execute `cargo build --release` and build parity.

rphmeier:
Thanks for the issue report! We just recently migrated our build instructions in the README to rustup, and we must have missed the dockerfiles.

PR #10657, opened 2019-05-14
refactor: Related #9459 - evmbin: replace untyped json! macro with fully typed serde serialization using Rust structs

ltfschoen:
Purpose: **Refactoring to replace untyped json! macro of #9489 with fully typed serde serialization using Rust structs for reusability and allows customization of how each field is encoded/decoded (i.e. camel case)**
* [x] - Move changes that are unrelated to the goal of this PR into a separate PR #10742 and PR #10740 
- [x] Mentioned related issue e.g. `Fixes #228` or `Related #1337`.
* [x] - Rename `multirust` (deprecated) to `rustup` 
* [x] - Initial update of evmbin Rust docs and code comments
* [x] - Add Response struct and implement it in response
* [x] - Remove serde imports that aren't used 
* [x] - Fix linting by converting from camelcase to snakecase properties as required
* [x] Is this PR even necessary? Hasn't https://github.com/paritytech/parity-ethereum/issues/9459 already been addressed in PR https://github.com/paritytech/parity-ethereum/pull/9489. Ans: Yes, see https://github.com/paritytech/parity-ethereum/pull/10657#issuecomment-492249700
- [x] Labeled PR with appropriate labels
- [x] PR adheres [the style guide](https://wiki.parity.io/Coding-guide)
* [x] Updated any rustdocs which may have changed
* [x] Implement serde json instead of `json!` wherever it is used apart from just TraceData
* [x] Replace untyped json! macro use with typed serde json `TraceData` in json.rs and std_json.rs. Status: resolved resolution of the `Clone` errors. 

<details><p>
**Question** (solved)
in json.rs i replaced use of the untyped json! macro with serde serialization, and i had to use `informant.clone()` here https://github.com/paritytech/parity-ethereum/pull/10657/files#diff-98c535329ae4c9199c5b7ddd6b06b5baR85, so i had to add `Clone` to the struct here https://github.com/paritytech/parity-ethereum/pull/10657/files#diff-98c535329ae4c9199c5b7ddd6b06b5baR30, and when i ran the tests they passed.

but when i did the same thing of using informant.clone() in the std_json.rs file here https://github.com/paritytech/parity-ethereum/pull/10657/files#diff-7c765744595c52ae7a64296e4264971cR206, and added `Clone` to the associated struct here https://github.com/paritytech/parity-ethereum/pull/10657/files#diff-7c765744595c52ae7a64296e4264971cR55, i get heaps of errors wrong number of type arguments: expected 3, found 2.
but when i add the additional first type as either `Clone` or `clone::std:clone::Clone` to address all the errors in commit https://github.com/paritytech/parity-ethereum/pull/10657/commits/42ebcf7e30c3fbbbe561865be9f5f190df5624ff, when i run the tests again with `cargo test -p evmbin --release -- --nocapture` i get a new set of errors, and heaps of them that include: the trait `std::clone::Clone` cannot be made into an object, and doesn't have a size known at compile-time (which i suspect i'd resolve with lifetimes).
i could continue trying to fix those errors, but i think i've done something wrong.

  * **Answer** 	based on discussion with @tomusdrw, instead of `stack: informant.clone().stack,` where we clone the entire informant and have to add `Informant<Clone`, it could instead be `informant.stack.clone()`, however we can do better by not cloning the stack and storage and rather borrowing, it involves some lifetimes, but doesn't require cloning at all.
</p></details>

* [x] Fix move out of borrowed content error by cloning instance of Informant struct with `Clone`
* [x] Fix failing tests and add new
```
cargo test -p evmbin --release -- --nocapture
```

Check that it works by copying create2callPrecompiles.json and teststate.json from https://github.com/paritytech/parity-ethereum/pull/10742 and run:
```
./target/release/parity-evm state-test ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople --json

./target/release/parity-evm state-test ./evmbin/res/teststate.json --only add12 --chain eip150 --json
```
* [x] - Remove debugging


evmbin/src/display/json.rs
```diff
@@ -18,6 +18,7 @@
 
 use std::collections::HashMap;
 use std::mem;
+use serde::{Serialize, Deserialize};
```
ltfschoen: shouldn't i be using these imports (i.e. `impl Serialize for Response`). see https://serde.rs/impl-serialize.html

ltfschoen: good question!

ordian:
@ltfschoen I think the #9459 was already resolved by #9489, but adding more docs is welcome.

ltfschoen:
Closing as this task has already been completed!

sorpaas:
@ordian i still think some of this PR can be useful -- replacing macro with fully typed serialization is a little bit better IMO. My opinion is that, not only we may be using it somewhere else in the future, but also, it allows us to customize how each field are encoded/decoded (camel case!) so it's a little bit more rusty. :)

WDYT?

ordian:
@sorpaas I agree the refactoring to use rust structs is better than untyped `json!`, though I think the original issue was resolved. Sorry I didn't mean we should close this PR.

evmbin/src/display/json.rs
```diff
@@ -59,25 +95,37 @@ impl Informant {
...
+				gas: &format!("{:#x}", gas_used.saturating_add(informant.gas_cost)),
+				gas_cost: &format!("{:#x}", informant.gas_cost),
+				memory: &format!("0x{}", informant.memory.to_hex()),
+				stack: &informant.stack,
+				storage: &informant.storage,
+				depth: informant.depth,
+			}
+		;
+
+		serde_json::to_string(&trace_data).unwrap()
```
ordian: Calling `unwrap` is discouraged according to our [style guide](https://wiki.parity.io/Substrate-Style-Guide).
```suggestion
		serde_json::to_string(&trace_data).expect("serialization cannot fail; qed")
```

evmbin/src/display/std_json.rs
```diff
@@ -178,17 +240,23 @@ impl<Trace: Writer, Out: Writer> trace::VMTracer for Informant<Trace, Out> {
...
+					op: instruction,
+					op_name: info.map(|i| i.name).unwrap_or(""),
+					gas: &format!("{:#x}", current_gas),
+					stack: &informant.stack,
+					storage: &informant.storage,
+					depth: informant.depth,
+				}
+			;
+
+			let serialized_trace_data = serde_json::to_string(&trace_data).unwrap();
```
ordian:
```suggestion
			let serialized_trace_data = serde_json::to_string(&trace_data).expect("serialization cannot fail; qed");
```

evmbin/src/display/json.rs
```diff
@@ -59,25 +95,37 @@ impl Informant {
...
-			"op": informant.instruction,
-			"opName": info.map(|i| i.name).unwrap_or(""),
-			"gas": format!("{:#x}", gas_used.saturating_add(informant.gas_cost)),
-			"gasCost": format!("{:#x}", informant.gas_cost),
-			"memory": format!("0x{}", informant.memory.to_hex()),
-			"stack": informant.stack,
-			"storage": informant.storage,
-			"depth": informant.depth,
-		}).to_string()
+		// Reference: https://serde.rs/attr-skip-serializing.html
```
dvdplm: Not sure what this comment means. Can you elaborate?

evmbin/src/display/json.rs
```diff
@@ -59,25 +95,37 @@ impl Informant {
...
 }
 
 impl vm::Informant for Informant {
 	type Sink = ();
 
 	fn before_test(&mut self, name: &str, action: &str) {
-		println!("{}", json!({"action": action, "test": name}));
+		let init_message =
+			InitMessage {
+				action: &format!("{}", action),
```
dvdplm: This looks odd. Why do you need to `format!` the params here?

evmbin/src/display/json.rs
```diff
@@ -59,25 +95,37 @@ impl Informant {
...
+				test: &format!("{}", name),
+			}
+		;
+
+		println!("{:?}", init_message);
```
dvdplm: Could this be a log instead, maybe `info!(…)`?

evmbin/src/display/json.rs
```diff
@@ -93,26 +141,30 @@ impl vm::Informant for Informant {
...
 				}
 
-				let success_msg = json!({
-					"output": format!("0x{}", success.output.to_hex()),
-					"gasUsed": format!("{:#x}", success.gas_used),
-					"time": display::as_micros(&success.time),
-				});
+				let message_success =
+					MessageSuccess {
+						output: &format!("0x{}", success.output.to_hex()),
```
dvdplm:
```suggestion
						output: &format!("{:#x}", success.output.to_hex()),
```

evmbin/src/display/json.rs
```diff
@@ -93,26 +141,30 @@ impl vm::Informant for Informant {
...
+						gas_used: &format!("{:#x}", success.gas_used),
+						time: &display::as_micros(&success.time),
+					}
+				;
 
-				println!("{}", success_msg)
+				println!("{:?}", message_success);
```
dvdplm: Maybe `info!`?

evmbin/src/display/std_json.rs
```diff
@@ -124,47 +174,59 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
 				Self::dump_state_into(trace_sink, success.state_root, &success.end_state);
 
-				let out_data = json!({
-					"output": format!("0x{}", success.output.to_hex()),
-					"gasUsed": format!("{:#x}", success.gas_used),
-					"time": display::as_micros(&success.time),
-				});
+				let out_data_success =
+					OutDataSuccess {
+						output: &format!("0x{}", success.output.to_hex()),
```
dvdplm:
```suggestion
						output: &format!("{:#x}", success.output.to_hex()),
```

evmbin/src/display/std_json.rs
```diff
@@ -178,17 +240,23 @@ impl<Trace: Writer, Out: Writer> trace::VMTracer for Informant<Trace, Out> {
...
-				"opName": info.map(|i| i.name).unwrap_or(""),
-				"gas": format!("{:#x}", current_gas),
-				"stack": informant.stack,
-				"storage": informant.storage,
-				"depth": informant.depth,
-			});
-
-			writeln!(&mut informant.trace_sink, "{}", trace_data).expect("The sink must be writeable.");
+
+			// Reference: https://serde.rs/attr-skip-serializing.html
```
dvdplm: Unclear (to me) comment.

evmbin/src/display/json.rs
tomusdrw: We want the output to go to `stdout` and it should be in a raw form (we generate JSON)

evmbin/src/display/json.rs
tomusdrw: I think you mean to skip `.to_hex()` too, no?

evmbin/src/display/json.rs
tomusdrw: Same as earlier, we don't want any additional formatting (time, module, etc) and need raw JSON on stdout.

evmbin/src/display/json.rs
ltfschoen: It was just a link to a reference but it's not important so i've removed it in commit 8ab7a78189efb3c318827eb873bd5f4c8df67e51

evmbin/src/display/std_json.rs
ltfschoen: It was just a link to a reference but it's not important so i've removed it in commit 8ab7a78189efb3c318827eb873bd5f4c8df67e51

evmbin/src/display/json.rs
ordian: Actually, the current output is a bit different than it used to be: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e97605d28a4ab5e2757c2b17f85e6f61.
I don't know if it really matters though.

tomusdrw: @ordian Great catch. Yeah - it's wrong now, it has to be proper JSON as the name of the file suggests.

evmbin/src/display/json.rs
ltfschoen: I've been trying to get this to compile for a while with a few different approaches. I've pushed the different approaches together in this commit https://github.com/paritytech/parity-ethereum/commit/5fbea495c8adcb1cb8de08abdbcb1c13d0e2c0d3#diff-98c535329ae4c9199c5b7ddd6b06b5ba of separate branch 'luke-9459-evmbin-serde-hex-format'.

i've been trying to "implement LowerHex for Vec<u8>" but when i build with `cargo build -p evmbin --release` i still get errors like the following that I haven't been able to overcome:
```
106 |             write!(f, "{:x}", val)
    |                               ^^^ the trait `std::fmt::LowerHex` is not implemented for `std::vec::Vec<u8>`

199 |                         output: &format!("{:#x}", success.output),        
    |                                                   ^^^^^^^^^^^^^^ the trait `std::fmt::LowerHex` is not implemented for `std::vec::Vec<u8>`
```

evmbin/src/display/json.rs
ltfschoen: @dvc94ch thanks, i just noticed we're using the env_logger crate https://docs.rs/env_logger/0.6.1/env_logger/ as the logging implementation https://docs.rs/log/0.4.6/log/#available-logging-implementations that we can use to do `info!(..)` 

@tomusdrw thanks, so it looks like we should use raw string literals https://doc.rust-lang.org/reference/tokens.html#raw-string-literals as used in this example, so it would become something like the following:
```
let var1 = "test1";
let json = format!(r#"{{"type": "type1", "type2": {}}}"#, var1);
println!("{}", json) // => {"type2": "type1", "type2": test1}
```

But then to use the env_logger crate so it goes to std_out (instead of std_err by default)
https://github.com/sebasmagri/env_logger#configuring-log-target we'd need to change it to something like:
```
use std::env;
use env_logger::{Builder, Target};

let mut builder = Builder::from_default_env();
builder.target(Target::Stdout);

builder.init();

... 

let raw_json = format!(r#"{:?}"#, init_message);
info!("{:?}", raw_json)
```

ordian: @ltfschoen I think we should use something like this
```suggestion
		let s = serde_json::to_string(&init_message).expect("serialization cannot fail; qed");
		println!("{}", s);
```
here and below

evmbin/src/display/json.rs
ltfschoen: @dvdplm thanks! i've pushed changes in this commit c743d338be2326a41925e53b69a03d881c875d38

evmbin/src/display/json.rs
ltfschoen: What do you think about adding a newline between each log and tab indenting the actual log dump under headings that correspond to the struct name. i.e.

```
info!("\nInit Message: \n\t{:?}", init_message);
```

evmbin/src/display/json.rs
dvdplm: Ok, it's not a big deal I think, thank you for trying! Apparently `LowerHex` for byte slices (or vectors) have a long history of discussion and isn't likely to happen soon.

evmbin/src/display/json.rs
```diff
@@ -93,26 +141,30 @@ impl vm::Informant for Informant {
...
+				let message_failure =
+					MessageFailure {
+						error: &failure.error.to_string(),
+						gas_used: &format!("{:#x}", failure.gas_used),
+						time: &display::as_micros(&failure.time),
+					}
+				;
 
-				println!("{}", failure_msg)
+				println!("{:?}", message_failure);
```
ordian:
```suggestion
				let s = serde_json::to_string(&message_failure).expect("serialization cannot fail; qed");
				println!("{}", s);
```

evmbin/src/display/std_json.rs
```diff
@@ -109,11 +156,14 @@ impl<Trace: Writer, Out: Writer> Informant<Trace, Out> {
...
-			});
-			writeln!(trace_sink, "{}", dump_data).expect("The sink must be writeable.");
+			let dump_data =
+				DumpData {
+					root: &root,
+					accounts: end_state,
+				}
+			;
+
+			writeln!(trace_sink, "{:?}", dump_data).expect("The sink must be writeable.");
```
ordian:
```suggestion
			let s = serde_json::to_string(&dump_data).expect("serialization cannot fail; qed");
			writeln!(trace_sink, "{}", s).expect("The sink must be writeable.");
```

evmbin/src/display/std_json.rs
```diff
@@ -124,47 +174,59 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
-		});
+		let out_data_before_test =
+			OutDataBeforeTest {
+				action: &format!("{}", action),
+				test: &format!("{}", name),
+			}
+		;
 
-		writeln!(&mut self.out_sink, "{}", out_data).expect("The sink must be writeable.");
+		writeln!(&mut self.out_sink, "{:?}", out_data_before_test).expect("The sink must be writeable.");
```
ordian:
```suggestion
		let s = serde_json::to_string(&out_data_before_test).expect("serialization cannot fail; qed");
		writeln!(&mut self.out_sink, "{}", s).expect("The sink must be writeable.");
```

evmbin/src/display/std_json.rs
```diff
@@ -124,47 +174,59 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
 			Ok(success) => {
-				let trace_data = json!({"stateRoot": success.state_root});
-				writeln!(trace_sink, "{}", trace_data)
+				let trace_data_state_root =
+					TraceDataStateRoot {
+						state_root: &success.state_root,
+					}
+				;
+
+				writeln!(trace_sink, "{:?}", trace_data_state_root)
```
ordian:
```suggestion
				let s = serde_json::to_string(&trace_data_state_root).expect("serialization cannot fail; qed");
				writeln!(trace_sink, "{}", s)
```

evmbin/src/display/std_json.rs
```diff
@@ -124,47 +174,59 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
+				let out_data_success =
+					OutDataSuccess {
+						output: &format!("0x{}", success.output.to_hex()),
+						gas_used: &format!("{:#x}", success.gas_used),
+						time: &display::as_micros(&success.time),
+					}
+				;
 
-				writeln!(out_sink, "{}", out_data).expect("The sink must be writeable.");
+				writeln!(out_sink, "{:?}", out_data_success).expect("The sink must be writeable.");
```
ordian:
```suggestion
				let s = serde_json::to_string(&out_data_success).expect("serialization cannot fail; qed");
				writeln!(out_sink, "{}", s).expect("The sink must be writeable.");
```

evmbin/src/display/std_json.rs
```diff
@@ -124,47 +174,59 @@ impl<Trace: Writer, Out: Writer> vm::Informant for Informant<Trace, Out> {
...
+						error: &failure.error.to_string(),
+						gas_used: &format!("{:#x}", failure.gas_used),
+						time: &display::as_micros(&failure.time),
+					}
+				;
 
 				Self::dump_state_into(trace_sink, failure.state_root, &failure.end_state);
 
-				writeln!(out_sink, "{}", out_data).expect("The sink must be writeable.");
+				writeln!(out_sink, "{:?}", out_data_failure).expect("The sink must be writeable.");
```
ordian:
```suggestion
				let s = serde_json::to_string(&out_data_failure).expect("serialization cannot fail; qed");
				writeln!(out_sink, "{}", s).expect("The sink must be writeable.");
```

evmbin/src/display/json.rs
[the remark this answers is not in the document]
ltfschoen: Thanks, I picked that up too in this PR https://github.com/paritytech/parity-ethereum/pull/10761. But I'll put the serialization changes together here though as you've suggested.

ordian:
@dvdplm could you take a look again?

accounts/ethstore/src/json/crypto.rs
```diff
@@ -41,7 +41,7 @@ impl str::FromStr for Crypto {
 
 impl From<Crypto> for String {
 	fn from(c: Crypto) -> Self {
-		serde_json::to_string(&c).expect("serialization cannot fail, cause all crypto keys are strings")
+		serde_json::to_string(&c).expect("Serialization cannot fail, cause all crypto keys are strings")
```
dvdplm:
```suggestion
		serde_json::to_string(&c).expect("Serialization cannot fail, because all crypto keys are strings")
```

PR #10658, opened 2019-05-14
docs: evmbin - Update Rust docs

ltfschoen:


evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
...
+//!
+//! The Parity EVM interpreter binary is an additional tool in the Parity
+//! Ethereum toolchain. It is an EVM implementation for Parity Ethereum that
+//! may be used to run a standalone version of the EVM interpreter.
+//!
+//! ## Usage
+//!
+//! Build Parity Ethereum from source, then start it with:
+//! ```bash
+//! cd evmbin && cargo build --release && cd ..
```
soc1c: something like

```
cargo build -p evmbin
```

and is much simpler

ltfschoen: i've pushed a commit addressing your comment to simplify it

ltfschoen:
Credit to @tomusdrw who explained evmbin too me so I was able to add some of the comments

evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
...
+//!
+//! The Parity EVM interpreter binary is an additional tool in the Parity
+//! Ethereum toolchain. It is an EVM implementation for Parity Ethereum that
+//! may be used to run a standalone version of the EVM interpreter.
+//!
+//! ## Usage
+//!
+//! Build Parity Ethereum from source, then start it with:
+//! ```bash
+//! cd evmbin && cargo build -p evmbin && cd ..
```
tomusdrw:
```suggestion
//! cargo build -p evmbin
```

That should suffice, cding should not be required.

evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Parity EVM interpreter binary.
+//! Parity EVM Interpreter Binary.
+//!
+//! ## Overview
+//!
+//! The Parity EVM interpreter binary is an additional tool in the Parity
```
dvdplm:
```suggestion
//! The Parity EVM interpreter binary is a tool in the Parity
```

evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
...
+//! Ethereum toolchain. It is an EVM implementation for Parity Ethereum that
+//! may be used to run a standalone version of the EVM interpreter.
```
dvdplm:
```suggestion
//! is used to run a standalone version of the EVM interpreter.
```

evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
...
+//!
+//! ## Usage
+//!
+//! Build Parity Ethereum from source, then start it with:
```
dvdplm: So it is not included in regular releases? That is a bit unusual so perhaps worth expanding on, like "The `evmbin` tool is not distributed with regular Parity Ethereum releases so you need to build it from source like so: …`?

evmbin/src/main.rs
dvdplm: Guessing here, but most users of `evmbin` would want to build it with `--release`, so it'd become `cargo build --release -p evmbin`.

evmbin/src/main.rs
```diff
@@ -14,7 +14,21 @@
...
+//! ```bash
+//! cd evmbin && cargo build -p evmbin && cd ..
+//! ./target/debug/parity-evm --help
```
dvdplm: If built with `--release` the path becomes `./target/release/parity-evm` (and we might want to suggest they copy the executable to, dunno, `~/bin`?).

evmbin/src/main.rs
```diff
@@ -268,13 +282,17 @@ struct Args {
 }
 
 impl Args {
+	/// Set the Gas Limit. Defaults to max value to allow code to run for whatever time is required.
```
dvdplm:
```suggestion
	/// Set the gas limit. Defaults to max value to allow code to run for whatever time is required.
```

evmbin/src/main.rs
```diff
@@ -268,13 +282,17 @@ struct Args {
...
 	pub fn gas(&self) -> Result<U256, String> {
 		match self.flag_gas {
 			Some(ref gas) => gas.parse().map_err(to_string),
 			None => Ok(U256::from(u64::max_value())),
 		}
 	}
 
+	/// Set the Gas Price. Defaults to zero to allow the code to run even if an account with no balance
```
dvdplm:
```suggestion
	/// Set the gas price. Defaults to zero to allow the code to run even if an account with no balance
```

evmbin/src/main.rs
soc1c: installing binaries is usually out of scope here

evmbin/src/main.rs
ltfschoen: i've pushed a commit fixing this, thanks!

evmbin/src/main.rs
ltfschoen: sorry, i've pushed a commit fixing it now so it's:
```
//! ```bash
//! cargo build -p evmbin --release
//! ./target/release/parity-evm --help
//! ```
```

evmbin/src/main.rs
ltfschoen: If they're using macOS or Linux and they want it in their PATH, we could suggest that they create a symlink i.e. `sudo ln -s <ABSOLUTE_PATH>/parity-ethereum/target/release/parity-evm /usr/local/bin`, assuming that `echo $PATH` indicates that /usr/local/bin is in their path, and then they can run `parity-evm --help` from any directory.
I'd have to look into what steps are required for Windows.

ltfschoen: Or should we just add a general suggestion like: _"Optionally if you want to be able to just run `parity-evm` from any directory then either copy the executable file itself to a directory that is in your PATH variable, or just create a symlink."_

evmbin/src/main.rs
ltfschoen: Are you happy with the changes @dvdplm?

evmbin/src/main.rs
ltfschoen: Is that ok @soc1c ?

evmbin/src/main.rs
ltfschoen: I've changed it to:
```
//! ```bash
//! cargo build -p evmbin --release
//! ./target/release/parity-evm --help
//! ```
```

Is that ok now @dvdplm @tomusdrw  ?

evmbin/src/main.rs
dvdplm: Personally I'd avoid symlinks to the build directory because files therein are clobbered anytime I build. Maybe a user of `evmbin` doesn't build as often as I do, but who knows. I think the wording could just be: "if you wish to run parity-evm from any directory copy the executable from ./target/release to a directory in your $PATH, e.g. ~/bin or /usr/local/bin.".

ltfschoen: Thanks, i've pushed a commit with this change

soc1c: again, I would like to emphasize that explaining users how to use binaries is out of scope for our documentation.

ltfschoen: @soc1c I've pushed another commit removing the explanation about how to use binaries

PR #10674, opened 2019-05-17
[hardware-wallet] use prost! for protobuf handling

ordian:
Makes cargo audit happy. Haven't actually tested on hardware wallet, but it seems to compile.
https://github.com/paritytech/trezor-sys/compare/prost

### DIFF
--- a/accounts/hw/src/trezor.rs
+++ b/accounts/hw/src/trezor.rs
@@ -29,7 +29,7 @@ use ethkey::Signature;
 use hidapi;
 use libusb;
 use parking_lot::{Mutex, RwLock};
-use prost::{DecodeError, Message};
+use prost::{DecodeError as ProstDecodeError, Message};
 use super::{DeviceDirection, WalletInfo, TransactionInfo, KeyPath, Wallet, Device, is_valid_hid_device};
 use trezor_sys::{EthereumAddress, PinMatrixAck, MessageType, EthereumTxRequest, EthereumSignTx, EthereumGetAddress, EthereumTxAck, ButtonAck};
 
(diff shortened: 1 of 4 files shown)


niklasad1:
cool, unfortunately I don't have access to a `Ledger`. Try to ping somebody in Berlin office to test it maybe @soc1c can? :)

accounts/hw/src/trezor.rs
```diff
@@ -118,6 +118,12 @@ enum HidVersion {
 	V2,
 }
 
+fn encode_message<M: Message>(message: M) -> Vec<u8> {
+	let mut result = Vec::new();
```
dvdplm: Does `message` know how long it is? Can you use `with_capacity()` here?

accounts/hw/src/trezor.rs
```diff
@@ -29,9 +29,9 @@ use ethkey::Signature;
 use hidapi;
 use libusb;
 use parking_lot::{Mutex, RwLock};
-use protobuf::{self, Message, ProtobufEnum};
+use prost::{DecodeError, Message};
```
dvdplm: Maybe a silly suggestion, but there's `rlp::DecodeError` all over the codebase so maybe a rename here is in order, e.g. `ProstDecodeError`?

dvdplm:
There's a new `protobuf` version out that might help here: https://github.com/stepancheg/rust-protobuf/issues/411

dvdplm:
Fix [back ported to `1.7.5` now](https://github.com/stepancheg/rust-protobuf/issues/411#issuecomment-493797088). Maybe that's a better route?

accounts/hw/src/trezor.rs
```diff
@@ -118,6 +118,12 @@ enum HidVersion {
 	V2,
 }
 
+fn encode_message<M: Message>(message: M) -> Vec<u8> {
+	let mut result = Vec::with_capacity(message.encoded_len());
+	message.encode(&mut result).expect("Vec has unlimited capacity; qed");
```
niklasad1: Nitpick, I don't agree with `unlimited capacity` because `Vec::alloc` will panic when capacity is bigger than `isize::max_value` on 32-bit platforms and `usize::max_value` on 64-bit platforms.

However, I think that is `only` theoretically and would be fine with some like `message could be represented as Vec there encodable; qed`

PR #10676, opened 2019-05-19
[CI] allow cargo audit to fail

ordian:
cargo audit is something that is orthogonal to a PR and shouldn't prevent it from being merged.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,7 +93,7 @@ cargo-audit:
   <<:                              *docker-cache-status
   script:
     - cargo audit
-  allow_failure:                   true
+  allow_failure:                   true # failed cargo audit shouldn't prevent a PR from being merged
 
 validate-chainspecs:
   stage:                           test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,6 +93,7 @@ cargo-audit:
   <<:                              *docker-cache-status
   script:
     - cargo audit
+  allow_failure:                   true
 
 validate-chainspecs:
   stage:                           test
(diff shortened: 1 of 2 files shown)


dvdplm:
Maybe it's an unpopular opinion but I think we should keep `cargo audit` as it is now (i.e. let it block merges). I agree that orthogonal concerns blocking our workflow is annoying and frustrating; the logic is impeccable.
However, I worry that without it we will become complacent with failures and leave it "for another day". This time we got a PR with a fix within hours, which is exactly the kind of turn-around time we want for anything security related. It is  a relatively rare occurrence and I think the frustration is worth it.

dvdplm:
```
ID:	 RUSTSEC-2019-0003
Crate:	 protobuf
Version: 1.7.5
Date:	 2019-06-08
URL:	 https://github.com/stepancheg/rust-protobuf/issues/411
Title:	 Out of Memory in stream::read_raw_bytes_into()
Solution: upgrade to: >= 1.7.5, >= 2.6.0
```

Sigh. So now we have a bug in `cargo-audit` too?

dvdplm:
On my machine and `cargo-audit 0.5.2` the check passes. Upgrading to latest (`0.6.1`) also passes. Can we upgrade to latest perhaps?

ordian:
Arguments for making cargo audit optional:
- cargo audit failure is highly unlikely to be introduced by a PR, since when a new dependency is introduced, it usually has the latest version 
- cargo audit can fail at any point in time, e.g. after a PR is merged (when a vulnerability is discovered)
- sometimes it can take some time make cargo audit happy (like e.g. upgrading tokio or ethereum-types)
- cargo audit has bugs too (^^^)
- we still can see if it fails
Screenshot from 2019-05-20 10-37-53

dvdplm:
@ordian you are not wrong (at all) and I'm not going to make a fuss about this but I still suspect we'll let things slip through. I'd rather keep it.

PR #10682, opened 2019-05-21
Don't panic if extra_data is longer than VANITY_LENGTH

dvdplm:
Seems like `extra_data().len()` can be longer than 32 (was 39 here) so this fixes that panic.

### DIFF
--- a/ethcore/src/engines/clique/mod.rs
+++ b/ethcore/src/engines/clique/mod.rs
@@ -713,7 +713,7 @@ impl Engine<EthereumMachine> for Clique {
 					}
 				}
 
-				let zero_padding_len = VANITY_LENGTH - header.extra_data().len();
+				let zero_padding_len = VANITY_LENGTH.checked_sub(header.extra_data().len()).unwrap_or(0);
 				if zero_padding_len > 0 {
 					let mut resized_extra_data = header.extra_data().clone();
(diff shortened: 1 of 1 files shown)


dvdplm:
> wrap in `release mode`

Isn't that potentially even worse though?

niklasad1:
> Isn't that potentially even worse though?

Yes, unless it is `signed a.k.a isize` but in this case it seems to be an usize

ethcore/src/engines/clique/mod.rs
```diff
@@ -713,7 +713,7 @@ impl Engine<EthereumMachine> for Clique {
 					}
 				}
 
-				let zero_padding_len = VANITY_LENGTH - header.extra_data().len();
+				let zero_padding_len = VANITY_LENGTH.checked_sub(header.extra_data().len()).unwrap_or(0);
```
dvdplm:
```suggestion
				let zero_padding_len = VANITY_LENGTH.saturating_sub(header.extra_data().len());
```

niklasad1:
> Isn't that potentially even worse though?

Yes, that is correct for example if `len  == 39` then it will wrap to a large value and if the will succeed!

Could potentially cast it to an `isize` instead but then it possible to the cast the `usize` to a negative value and it will be `addition` instead!

Thus, `saturating_sub` is all good :)

Issue #10648, opened 2019-05-13
refactor `Clique and StepService` 


niklasad1:
The issue is the following [here](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/engines/clique/mod.rs#L189-#L209)

```bash
warning[E0382]: assign to part of moved value: `engine`
   --> ethcore/src/engines/clique/mod.rs:204:4
    |
190 |         let mut engine = Clique {
    |             ---------- move occurs because `engine` has type `engines::clique::Clique`, which does not implement the `Copy` trait
...
201 |         let res = Arc::new(engine);
    |                            ------ value moved here
...
204 |             engine.step_service = Some(StepService::start(Arc::downgrade(&res) as Weak<Engine<_>>));
    |             ^^^^^^^^^^^^^^^^^^^ value partially assigned here after move
    = warning: this error has been downgraded to a warning for backwards compatibility with previous releases
    = warning: this represents potential undefined behavior in your code and this warning will become a hard error in the future
```

To elaborate a bit, basically we mutate an already value moved into the `Arc` and this should not compile IMHO.

So, I would suggest to move the `StepService` into `Clique` instead because it is not safe/possible to mutate an `Arc` when it is aliased (i.e, more than 1 is pointing to it)

Thoughts?

dvdplm:
Yeah, seems like `engine` ends up in two `Arcs`, one assigned to `step_service` and the other returned  by `Clique::new`? I'm also surprised it compiles tbh.

PR #10683, opened 2019-05-21
Fix compiler warning (that will become an error)

dvdplm:
Fixes https://github.com/paritytech/parity-ethereum/issues/10648

Depends on https://github.com/paritytech/parity-ethereum/pull/10689

I'm not sure this fix is Good™ but rustc seems happy enough.

~~There's a deeper issue which may or may not be related to this: the Engine is not shutdown properly and the `StepService` thread keeps running indefinitely after Ctrl-C (so `update_sealing()` is called repeatedly for 300sec). I don't think this is related to Clique as I've seen this happen on mainnet as well, but I wonder if the effects of it are worse for a PoA network where the node can create new blocks all on its own?~~

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -2545,16 +2545,6 @@ impl ProvingBlockChainClient for Client {
 
 impl SnapshotClient for Client {}
 
-impl Drop for Client {
-	fn drop(&mut self) {
-		if let Some(c) = Arc::get_mut(&mut self.engine) {
-			c.stop()
-		} else {
-			warn!(target: "shutdown", "unable to get mut ref for engine for shutdown.");
-		}
-	}
-}
-
 /// Returns `LocalizedReceipt` given `LocalizedTransaction`
 /// and a vector of receipts from given block up to transaction index.
 fn transaction_receipt(
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -25,7 +25,7 @@ use blockchain::{BlockReceipts, BlockChain, BlockChainDB, BlockProvider, TreeRou
 use bytes::Bytes;
 use call_contract::{CallContract, RegistryInfo};
 use ethcore_miner::pool::VerifiedTransaction;
-use ethereum_types::{H256, Address, U256};
+use ethereum_types::{H256, H264, Address, U256};
 use evm::Schedule;
 use hash::keccak;
 use io::IoChannel;
@@ -86,7 +86,7 @@ pub use types::blockchain_info::BlockChainInfo;
 pub use types::block_status::BlockStatus;
 pub use blockchain::CacheSize as BlockChainCacheSize;
 pub use verification::QueueInfo as BlockQueueInfo;
-use db::Writable;
+use db::{Writable, Readable, keys::BlockDetails};
 
 use_contract!(registry, "res/contracts/registrar.json");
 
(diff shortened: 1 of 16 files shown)


ethcore/src/engines/clique/mod.rs
```diff
@@ -198,13 +198,15 @@ impl Clique {
...
 		};
 
-		let res = Arc::new(engine);
+		let mut engine= Arc::new(engine);
 
 		if our_params.period > 0 {
-			engine.step_service = Some(StepService::start(Arc::downgrade(&res) as Weak<Engine<_>>));
+			let engine_weak = Arc::downgrade(&engine.clone());
+			let step_service = StepService::start(engine_weak as Weak<Engine<_>>);
+			Arc::get_mut(&mut engine).map(|e| e.step_service = Some(step_service));
```
niklasad1: I don't think this will work because we got two references to the same value, https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut

dvdplm: Well, it compiles with no warning and runs (seemingly) ok! It feels iffy for sure though.

Do you think it works because of the 5sec delay in the thread started by `StepService`?

ethcore/src/engines/clique/mod.rs
```diff
@@ -198,13 +198,15 @@ impl Clique {
 			step_service: None,
 		};
 
-		let res = Arc::new(engine);
+		let mut engine= Arc::new(engine);
 
 		if our_params.period > 0 {
-			engine.step_service = Some(StepService::start(Arc::downgrade(&res) as Weak<Engine<_>>));
+			let engine_weak = Arc::downgrade(&engine.clone());
```
tomusdrw: `clone()` doesn't seem to be needed here
```suggestion
			let engine_weak = Arc::downgrade(&engine);
```

ethcore/src/engines/clique/mod.rs
tomusdrw: Hmm, indeed this shouldn't work - can we replace `map` with `expect` to make sure it's always there (cause that's what we assume).
Maybe let's separate creation of `StepService` from actually starting the inner thread? So that we can do:
```
let mut engine = ...;
let step_service = StepService::new();
engine.step_service = step_service;
let engine = Arc::new(engine);
let weak = Arc::downgrade(&engine);
engine.step_service.start(weak);
```

ethcore/src/engines/clique/mod.rs
```diff
@@ -713,7 +715,7 @@ impl Engine<EthereumMachine> for Clique {
 					}
 				}
 
-				let zero_padding_len = VANITY_LENGTH - header.extra_data().len();
+				let zero_padding_len = VANITY_LENGTH.checked_sub(header.extra_data().len()).unwrap_or(0);
```
tomusdrw: `checked_sub()...unwrap_or(0)` seems to do the same as just `saturating_sub()`

ethcore/src/engines/clique/mod.rs
dvdplm: @tomusdrw excellent idea, I'll see where it takes me. Thanks!

dvdplm: `engine.step_service.start(weak);` does not work as long as `step_service.start()` takes `&mut self` (can't move out of an `Arc`) so need some tinkering. Hmm.

tomusdrw: @dvdplm because you can't store the thread handle? Maybe put it to `Mutex`, since it's not hot code anyway, and we just need to store it.

dvdplm: > because you can't store the thread handle?

Precisely. Using a `Mutex` most likely works; possibly reducing `StepService` to an `Arc<AtomicBool>` (i.e. not storing the thread handle at all) and let the thread run wild and free could work.

I am however pausing work on this for the moment because shutdown is currently broken elsewhere and I need to fix that before coming back to this. :/

niklasad1: Ok, it could potentially be this https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/engines/clique/mod.rs#L749? 

I expect it `Client` to aliased there too, have you seen the `warn` print?

dvdplm: @niklasad1 with [the other shutdown bug](https://github.com/paritytech/parity-ethereum/pull/10689) I do see it. Without it, that code is never called.

dvdplm:
@tomusdrw @niklasad1 hoisting [the commit message](https://github.com/paritytech/parity-ethereum/pull/10683/commits/4b6fb23f077c8cb38aa2963fc20033bdca7c81c2) here so it's easier to read:

 Fix warning, second attempt

The idea here is to avoid using `Arc::get_mut()` (which does not work: fails every time here) and instead trust `drop()` to do the right thing.

This is a conservative change. I think this can be reformed further, e.g. by `impl Drop for StepService` and halt the thread there, or even skip `join()`ing the thread entirely and trust the `AtomicBool` to signal shutdown. I also have doubts abut the `Option<StepService>`: seems a bit much to have an `Option` there and it makes things cumbersome.

dvdplm:
To add to the above, I don't think it's valid to run görli with `period` set to `0` so I think it's best to remove the conditional code that makes the code a bit convoluted here. I'll have a stab at that in a follow-up PR.

ethcore/src/client/client.rs
```diff
@@ -2545,16 +2545,6 @@ impl ProvingBlockChainClient for Client {
 
 impl SnapshotClient for Client {}
 
-impl Drop for Client {
-	fn drop(&mut self) {
-		if let Some(c) = Arc::get_mut(&mut self.engine) {
-			c.stop()
```
dvdplm: AFAICT `Clique` is the only engine that actually impls `stop()`. And the `Arc::get_mut` here fails every time.

ethcore/src/engines/clique/mod.rs
```diff
@@ -168,7 +168,7 @@ pub struct Clique {
 	block_state_by_hash: RwLock<LruCache<H256, CliqueBlockState>>,
 	proposals: RwLock<HashMap<Address, VoteType>>,
 	signer: RwLock<Option<Box<EngineSigner>>>,
-	step_service: Option<Arc<StepService>>,
+	step_service: Option<StepService>,
```
tomusdrw: If we disallow `period=0` we could get rid of the `Option` here as well.

ethcore/src/engines/clique/mod.rs
```diff
@@ -744,14 +756,6 @@ impl Engine<EthereumMachine> for Clique {
 		}
 	}
 
-	fn stop(&mut self) {
```
tomusdrw: Do we actually need `stop` on the `Engine` trait then? Is it used anywhere?

ethcore/src/client/client.rs
niklasad1: Can you change https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/engines/mod.rs#L429 to `fn stop(&self)` it is broken, because `Engine` is stored as `Arc<Engine>` and doesn't support mutation.

I'm sorry that I didn't catch when reviewing it :(

ethcore/src/engines/clique/mod.rs
niklasad1: > Can you change https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/engines/mod.rs#L429 to fn stop(&self) it is broken, because Engine is stored as Arc<Engine> and doesn't support mutation.

It is not used and it is broken. Either remove it or change it to `fn stop(&self)`

dvdplm: Yes, that's planned for the follow up PR. :)

(What, you a mind reader @tomusdrw ???)

ethcore/src/engines/clique/mod.rs
dvdplm: Yes, that's planned for the follow up PR. :)

niklasad1:
@dvdplm @s3krit I would really want this and https://github.com/paritytech/parity-ethereum/pull/10691/files backported, does it make sense?

dvdplm:
Wow, so they are missing on 2.5? That is unfortunate. :/

dvdplm:
Now I’m curious: how did you notice? And yes, ofc we should back port them. I wonder if adding the labels is enough or...?

niklasad1:
@dvdplm just by coincidence, I tried to build the `stable` branch with the nightly toolchain which failed because of the `compiler warning` that you fixed in this PR

PR #10690, opened 2019-05-22
fix(clique step_service): remove needless `Arc`

niklasad1:


niklasad1:
It shouldn't compile then because `Clique` is wrapped in an Arc directly in the `constructor`, which doesn't provide mutation unless `risky Arc::get_mut()`. Thus, I think `stop()` never is called otherwise it shouldn't compile.

Basically, this is all wrong either use `drop` or change to `stop(&self)` and wrap step_service in an Mutex/RwLock.

Closing in favor of #10683

PR #10691, opened 2019-05-23
Refactor Clique stepping

dvdplm:
As a continuation of https://github.com/paritytech/parity-ethereum/pull/10683, this PR tries to simplify the stepping. The three commits present three stages of simplification.

1. Tidy things up and use `Drop` instead of homegrown `stop()` to shutdown the `StepService`: https://github.com/paritytech/parity-ethereum/pull/10691/commits/fe55764596e3699d70185c754582511ee8451395
2. Move the `StepService` functionality inside `Clique::new`; shut it down using `impl Drop for Clique`: https://github.com/paritytech/parity-ethereum/pull/10691/commits/709e8fe47bde1512fe3c380423ef4c07daba6073
3. Don't try to shut down the thread at all, just let it run until the process exits: https://github.com/paritytech/parity-ethereum/pull/10691/commits/56eb2ac77014af2185319dcf1eb94e4c08f68cd5


ethcore/src/engines/clique/mod.rs
```diff
@@ -181,33 +178,38 @@ pub struct Clique {
...
-		}
+		let engine = Arc::new(engine);
+		let weak_eng = Arc::downgrade(&engine);
+
+		thread::Builder::new().name("StepService".into())
+			.spawn(move || {
+				thread::sleep(INITIAL_DELAY);
+				loop {
+					trace!(target: "miner", "StepService: triggering sealing");
+					weak_eng.upgrade().map(|x| x.step());
```
tomusdrw: Shall we print a warning and actually terminate the thread?

ethcore/src/engines/clique/mod.rs
```diff
@@ -181,33 +178,38 @@ pub struct Clique {
...
-			Ok(engine)
-		} else {
-			Ok(Arc::new(engine))
-		}
+		let engine = Arc::new(engine);
+		let weak_eng = Arc::downgrade(&engine);
+
+		thread::Builder::new().name("StepService".into())
+			.spawn(move || {
+				thread::sleep(INITIAL_DELAY);
```
tomusdrw: Why is this delay needed? Seems smelly...

ethcore/src/engines/clique/mod.rs
```diff
@@ -181,33 +178,38 @@ pub struct Clique {
...
 }
 
 impl Clique {
 	/// Initialize Clique engine from empty state.
 	pub fn new(params: CliqueParams, machine: EthereumMachine) -> Result<Arc<Self>, Error> {
-		let mut engine = Clique {
+		/// Pause before starting to step Clique
+		const INITIAL_DELAY: Duration = Duration::from_secs(5);
+		/// Step Clique at most every 2 seconds
+		const SEALING_FREQ: Duration = Duration::from_secs(2);
```
tomusdrw: That seems way too big, if you configure step to be `5seconds` you will end up being one second (20%) late in a worst case.

ethcore/src/engines/clique/mod.rs
```diff
@@ -181,33 +178,38 @@ pub struct Clique {
...
 			signer: Default::default(),
 			machine,
-			step_service: None,
 		};
-		if params.period > 0 {
-			engine.step_service = Some(StepService::new());
-			let engine = Arc::new(engine);
-			let weak_eng = Arc::downgrade(&engine);
-			if let Some(step_service) = &engine.step_service {
-				step_service.start(weak_eng);
```
tomusdrw: Shall we remove `start` from `step_service` if it's not used any more?

ethcore/src/engines/clique/mod.rs
tomusdrw: Now I think about it, it might be better to actually take the time calling `step()` took into account, i.e.:
Otherwise if `step` takes a long time (due to lock contention or sth), we might get quite out of sync.
```rust
loop {
  let mut next_step_at = Instant::now() + SEALING_FREQ;
  weak_eng.upgrade().map(|x| x.step());
  let now = Instant::now();
  if now < next_step_at {
    thread::sleep(next_step_at - now);
  }
}
```

dvdplm: You mean if the `upgrade()` fails? I read the code as "it probably doesn't matter if it skips a step" but I don't actually know. @niklasad1 ?

ethcore/src/engines/clique/mod.rs
dvdplm: Indeed, my guess is it's just to avoid triggering seals during startup to avoid some spam in the logs maybe? Seems safe to remove, but @niklasad1 knows better.

ethcore/src/engines/clique/mod.rs
tomusdrw: @dvdplm if `upgrade` fails once it will never recover (cause it means that `Engine` has been dropped), so the loop will be infite, might be better to print a warning and terminate the thread.

ethcore/src/engines/clique/mod.rs
tomusdrw: Don't we have `engine.start()`? We could probably use this as a signal to that thread to start working.

ethcore/src/engines/clique/mod.rs
dvdplm: @niklasad1 Did the stepping frequency come from the spec? I tried using 200ms here and got plenty of `TRACE miner  requires_reseal: reseal too early` in the logs.

ethcore/src/engines/clique/mod.rs
dvdplm: All of ethcore/src/engines/clique/step_service.rs is gone in this PR!

ethcore/src/engines/clique/mod.rs
tomusdrw: @dvdplm that's because of other setting from `miner` (`--min-reseal-period`) - it prevents constructing new blocks too often (usualy it's due to new transactions coming into the pool). But it seems pretty weird that apparently `step()` actually constructs new block every time? Then shortening the time will be pretty grim from performance perspective especially in networks where blocks are at it's limits.

ethcore/src/engines/clique/mod.rs
dvdplm: > if upgrade fails once it will never recover (cause it means that Engine has been dropped),

Got it. I misread the semantics of `upgrade`, thought it could recover. However, my question becomes: if the Engine is dropped, doesn't that mean that the process is exiting (so the thread is getting reaped anyway)? Put differently: is there a case where the Engine is dropped but we're not shutting down? I thought not.

dvdplm: > take the time calling `step()` took into account,

Very good pt, fixing! :)

tomusdrw: > However, my question becomes: if the Engine is dropped, doesn't that mean that the process is exiting (so the thread is getting reaped anyway)? Put differently: is there a case where the Engine is dropped but we're not shutting down? I thought not.

Yeah, that should be the case, but IMHO it's better to handle that explicitly - that will make the code more future proof.
Imagine that in the future we move some other stuff to this thread and not terminating it will prevent other drops.

ethcore/src/engines/clique/mod.rs
dvdplm: We do not have `engine.start()` (or I don't know where to look!)

tomusdrw: Oh, you're right. I thought that since we had `stop` we have `start` as well. We could either introduce it or maybe trigger `start()` on first `generate_seal`? I believe that function should be called for every engine by the miner module, no?

tomusdrw: Or we could use `register_client` as the trigger?

dvdplm: I went with removing the delay altogether. IMO it muddles up the code (why 5sec? is it important or just an optimization? etc) for little gain. @niklasad1 Thoughts on this? :)

dvdplm:
@niklasad1 ping

ethcore/src/engines/clique/mod.rs
niklasad1: @dvdplm It is not part of the spec AFAIK, as Tomek said I think it is used for sealing timing.

ethcore/src/engines/clique/mod.rs
```diff
@@ -181,33 +178,44 @@ pub struct Clique {
...
-		}
+		let engine = Arc::new(engine);
+		let weak_eng = Arc::downgrade(&engine);
+
+		thread::Builder::new().name("StepService".into())
+			.spawn(move || {
+				loop {
+ 					let next_step_at = Instant::now() + SEALING_FREQ;
+					trace!(target: "miner", "StepService: triggering sealing");
+					if let Some(eng) = weak_eng.upgrade() { eng.step() }
```
tomusdrw: I'm not sure about the collapse here, anything wrong with:
```rust
if let Some(eng) = weak_eng.upgrade() {
  eng.step();
} else {
  warn...
  break;
}
```
?

dvdplm: Nothing wrong. Fixed. :)

PR #10700, opened 2019-05-29
introduce MissingParent Error, fixes #10699

### LINKED ISSUE #10699: Provide Dedicated Error re #10085
(draft)

The highly sporadic "stage 3 verification error" reported in #10085 is reported in detail as:

'Error(Engine(RequiresClient)'

which is not correct.

A dedicated error-message should be provided, describing the real error-cause.

(I'm doing this, have it more or less already available)

lazaridiscom:
draft, provided based on code/code-doc review.

fixes #10699

### DIFF
--- a/ethcore/src/engines/mod.rs
+++ b/ethcore/src/engines/mod.rs
@@ -89,6 +89,8 @@ pub enum EngineError {
 	InvalidEngine,
 	/// Requires signer ref, but none registered.
 	RequiresSigner,
+	/// Missing Parent Epoch
+	MissingParent,
 	/// Checkpoint is missing
 	CliqueMissingCheckpoint(H256),
 	/// Missing vanity data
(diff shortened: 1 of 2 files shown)


ethcore/src/engines/mod.rs
```diff
@@ -89,6 +89,8 @@ pub enum EngineError {
 	InvalidEngine,
 	/// Requires signer ref, but none registered.
 	RequiresSigner,
+	/// Missing Parent Epoch
+	MissingParent,
```
dvdplm: Does it make sense to include the parent hash in the error?

lazaridiscom: > Does it make sense to include the parent hash in the error?

Yes, but i avoid strictly to introduce new functionality whilst fixing things (usually I don't touch even the whitespace). Thus I adopted for now the existent code. Less stress and makes review easier.

In a next iteration I'll go for "refine the relevant debug output / error messages" as a whole (e.g. print out block-no/hash in a uniform way, not only for this error, but for others & debug-log output, too ).

dvdplm: Fair enough, thank you for your work here! :)

lazaridiscom:
debug output refinement in follow-up issue: #10701

PR #10705, opened 2019-05-31
DevP2p: Get node IP address and udp port from Socket, if not included in PING packet

### LINKED ISSUE #10613: Node Discover V4 compatibility issue: PING request parity doesn't reply with PONG
... and in a result `parity` might not be compatible with `geth` bootnode
AS `geth` bootnode doesn't respond back to FINDNODE and internally logs error "unknown node"

- **Parity Ethereum version**: v2.5.0-beta-b52ac20-20190408/x86_64-linux-gnu/rustc1.33.0
- **Operating system**: Linux
- **Network**: private
- **Restarted**: yes

Issue might me related to code
https://github.com/paritytech/parity-ethereum/blob/master/util/network-devp2p/src/discovery.rs#L535

`ip-10-101-2-4:$ bootnode -nodekey bootnode.key -verbosity 9 -addr :30301 -writeaddress`
```
c6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e
```
`ip-10-101-2-4:$ bootnode -nodekey bootnode.key -verbosity 9 -addr :30301`
```
INFO [04-26|14:22:09.746] New local node record                    seq=1 id=9af0873613686cd1 ip=<nil> udp=0 tcp=0
TRACE[04-26|14:26:57.351] << PING/v4                               id=bed6bb55be4f44dc addr=10.111.2.225:30303 err=nil
TRACE[04-26|14:26:57.351] >> PONG/v4                               id=bed6bb55be4f44dc addr=10.111.2.225:30303 err=nil
TRACE[04-26|14:26:57.352] >> PING/v4                               id=bed6bb55be4f44dc addr=10.111.2.225:30303 err=nil
TRACE[04-26|14:26:58.219] << FINDNODE/v4                           id=bed6bb55be4f44dc addr=10.111.2.225:30303 err="unknown node"
TRACE[04-26|14:27:07.527] << FINDNODE/v4                           id=bed6bb55be4f44dc addr=10.111.2.225:30303 err="unknown node"
TRACE[04-26|14:27:17.438] << FINDNODE/v4                           id=bed6bb55be4f44dc addr=10.111.2.225:30303 err="unknown node"
TRACE[04-26|14:27:27.650] << FINDNODE/v4                           id=bed6bb55be4f44dc addr=10.111.2.225:30303 err="unknown node"
```

`ip-10-111-2-225:$ parity --base-path=./node --config=./config/gateway.toml -l discovery=trace`
```
Loading config file from ./config/gateway.toml
2019-04-26 14:26:50 UTC main INFO parity_ethereum::run  Starting Parity-Ethereum/v2.5.0-beta-b52ac20-20190408/x86_64-linux-gnu/rustc1.33.0
2019-04-26 14:26:50 UTC main INFO parity_ethereum::run  Keys path ./node/keys
2019-04-26 14:26:50 UTC main INFO parity_ethereum::run  DB path ./node/chains/db/94cd4c55fa93fb4b
2019-04-26 14:26:50 UTC main INFO parity_ethereum::run  State DB configuration: fast
2019-04-26 14:26:50 UTC main INFO parity_ethereum::run  Operating mode: active
2019-04-26 14:26:51 UTC main INFO ethcore_service::service  Configured for using AuthorityRound engine
2019-04-26 14:26:57 UTC IO Worker #2 INFO network  Public node URL: enode://cf849dcf68ba78181c256af456d57d6a5323b22617adf4ea04803eca23d8daa224dfef74f81daf3c90208bd54f3ecc6d21e151988e08941ac92db4c3a37ce4d2@10.111.2.225:30303
2019-04-26 14:26:57 UTC IO Worker #0 TRACE discovery  Sent Ping to NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 } ; node_id=0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e
2019-04-26 14:26:57 UTC IO Worker #2 TRACE discovery  Got Pong from V4(10.101.2.4:30301) ; node_id=0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e
2019-04-26 14:26:57 UTC IO Worker #2 TRACE discovery  Inserting NodeEntry { id: 0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e, endpoint: NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 } }
2019-04-26 14:26:57 UTC IO Worker #2 TRACE discovery  Adding a new node NodeEntry { id: 0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e, endpoint: NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 } } into our bucket 253
2019-04-26 14:26:57 UTC IO Worker #2 TRACE discovery  Got Ping from V4(10.101.2.4:30301)
2019-04-26 14:26:57 UTC IO Worker #3 TRACE discovery  Starting discovery
2019-04-26 14:26:58 UTC IO Worker #3 TRACE discovery  Starting round Some(0)
2019-04-26 14:26:58 UTC IO Worker #3 TRACE discovery  Sent FindNode to NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 }
2019-04-26 14:26:58 UTC IO Worker #3 TRACE discovery  Starting round Some(1)
2019-04-26 14:26:58 UTC IO Worker #3 TRACE discovery  Completing discovery
2019-04-26 14:27:00 UTC IO Worker #2 DEBUG discovery  Removing expired FIND NODE request for node_id=0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e
2019-04-26 14:27:00 UTC IO Worker #2 TRACE discovery  Requests to node NodeEntry { id: 0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e, endpoint: NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 } } timed out 1 consecutive time(s)
2019-04-26 14:27:07 UTC IO Worker #3 TRACE discovery  Starting discovery
2019-04-26 14:27:07 UTC IO Worker #0 TRACE discovery  Starting round Some(0)
2019-04-26 14:27:07 UTC IO Worker #0 TRACE discovery  Sent FindNode to NodeEndpoint { address: V4(10.101.2.4:30301), udp_port: 30301 }
2019-04-26 14:27:07 UTC IO Worker #0 TRACE discovery  Starting round Some(1)
2019-04-26 14:27:07 UTC IO Worker #0 TRACE discovery  Completing discovery
2019-04-26 14:27:09 UTC IO Worker #1 DEBUG discovery  Removing expired FIND NODE request for node_id=0xc6fea65e3dd8b8cd5102816d8a3f3e7c98e8952357adb12970c77c15144c151f7e2fa86bee7ae7a980cfbf04eda58a737800e55a44a357f966a473eef8b38c2e
```

seunlanlege:
closes #10613

### DIFF
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -527,7 +527,7 @@ impl<'a> Discovery<'a> {
 		} else {
 			let mut address = from.clone();
 			// address here is the node's tcp port. If we are unable to get the `NodeEndpoint` from the `ping_from`
-			// field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
+			// rlp field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
 			address.set_port(0);
 			NodeEndpoint {
 				address,
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -169,6 +169,9 @@ pub struct Discovery<'a> {
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
 
+	// these nodes can't be used for syncing.
+	bootnodes: HashSet<NodeId>,
+
 	// Sometimes we don't want to add nodes to the NodeTable, but still want to
 	// keep track of them to avoid excessive pinging (happens when an unknown node sends
 	// a discovery request to us -- the node might be on a different net).
@@ -200,6 +203,7 @@ impl<'a> Discovery<'a> {
 			discovery_id: NodeId::new(),
 			discovery_nodes: HashSet::new(),
 			node_buckets: (0..ADDRESS_BITS).map(|_| NodeBucket::new()).collect(),
+			bootnodes: HashSet::new(),
 			other_observed_nodes: LruCache::new(OBSERVED_NODES_MAX_SIZE),
 			in_flight_pings: HashMap::new(),
 			in_flight_find_nodes: HashMap::new(),
@@ -518,11 +522,16 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let ping_from = if let Ok(rlp) = rlp.at(1) {
-			NodeEndpoint::from_rlp(&rlp)?;
+		let ping_from = if let Ok(node_endpoint) = NodeEndpoint::from_rlp(&rlp.at(1)?) {
+			node_endpoint
 		} else {
+			let mut address = from.clone();
+			// address here is the node's tcp port. If we are unable to get the `NodeEndpoint` from the `ping_from`
+			// field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
+			self.bootnodes.insert(node_id.clone());
+			address.set_port(0);
 			NodeEndpoint {
-				address: from.clone(),
+				address,
 				udp_port: from.port()
 			}
 		};
@@ -547,7 +556,7 @@ impl<'a> Discovery<'a> {
 		self.send_packet(PACKET_PONG, from, &response.drain())?;
 
 		let entry = NodeEntry { id: *node_id, endpoint: pong_to.clone() };
-		if !entry.endpoint.is_valid() {
+		if !entry.endpoint.is_valid() && !self.bootnodes.contains(node_id) {
 			debug!(target: "discovery", "Got bad address: {:?}", entry);
 		} else if !self.is_allowed(&entry) {
 			debug!(target: "discovery", "Address not allowed: {:?}", entry);
@@ -609,8 +618,10 @@ impl<'a> Discovery<'a> {
 					NodeValidity::Ourselves => (),
 				}
 				Ok(None)
-			} else {
+			} else if !self.bootnodes.contains(node_id) {
 				Ok(self.update_node(node))
+			} else {
+				Ok(None)
 			}
 		} else {
 			debug!(target: "discovery", "Got unexpected Pong from {:?} ; request not found", &from);
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -743,7 +743,7 @@ impl<'a> Discovery<'a> {
 		trace!(target: "discovery", "Got {} Neighbours from {:?}", results_count, &from);
 		for r in rlp.at(0)?.iter() {
 			let endpoint = NodeEndpoint::from_rlp(&r)?;
-			if !endpoint.is_valid_sync_node() {
+			if !endpoint.is_valid_discovery_node() {
 				debug!(target: "discovery", "Bad address: {:?}", endpoint);
 				continue;
 			}
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -518,7 +518,14 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_from = if let Ok(rlp) = rlp.at(1) {
+			NodeEndpoint::from_rlp(&rlp)?;
+		} else {
+			NodeEndpoint {
+				address: from.clone(),
+				udp_port: from.port()
+			}
+		};
 		let ping_to = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
 		let timestamp: u64 = rlp.val_at(3)?;
 		self.check_timestamp(timestamp)?;
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -269,7 +269,7 @@ impl<'a> Discovery<'a> {
 
 			let node_to_ping = {
 				let bucket = &mut self.node_buckets[bucket_distance];
-				bucket.nodes.push_front(BucketEntry::new(node_entry));
+				bucket.nodes.push_front(BucketEntry::new(node_entry.clone()));
 				if bucket.nodes.len() > BUCKET_SIZE {
 					select_bucket_ping(bucket.nodes.iter())
 				} else {
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -261,7 +261,7 @@ impl<'a> Discovery<'a> {
             Ok(()) => None,
             Err(BucketError::Ourselves) => None,
             Err(BucketError::NotInTheBucket{node_entry, bucket_distance}) => Some((node_entry, bucket_distance))
-        }.map(|(node_entry, bucket_distance)| {
+        }.and_then(|(node_entry, bucket_distance)| {
 			trace!(target: "discovery", "Adding a new node {:?} into our bucket {}", &node_entry, bucket_distance);
 
             let mut added = HashMap::with_capacity(1);
@@ -279,7 +279,12 @@ impl<'a> Discovery<'a> {
 			if let Some(node) = node_to_ping {
 				self.try_ping(node, PingReason::Default);
 			};
-            TableUpdates{added, removed: HashSet::new()}
+
+            if !self.bootnodes.contains(&node_entry.id) {
+				Some(TableUpdates { added, removed: HashSet::new() })
+			} else {
+				None
+			}
         })
 	}
 
@@ -618,10 +623,8 @@ impl<'a> Discovery<'a> {
 					NodeValidity::Ourselves => (),
 				}
 				Ok(None)
-			} else if !self.bootnodes.contains(node_id) {
-				Ok(self.update_node(node))
 			} else {
-				Ok(None)
+				Ok(self.update_node(node))
 			}
 		} else {
 			debug!(target: "discovery", "Got unexpected Pong from {:?} ; request not found", &from);
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -522,17 +522,17 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let (ping_from, bootnode) = if let Ok(node_endpoint) = NodeEndpoint::from_rlp(&rlp.at(1)?) {
-			(node_endpoint, false)
+		let ping_from = if let Ok(node_endpoint) = NodeEndpoint::from_rlp(&rlp.at(1)?) {
+			node_endpoint
 		} else {
 			let mut address = from.clone();
 			// address here is the node's tcp port. If we are unable to get the `NodeEndpoint` from the `ping_from`
 			// field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
 			address.set_port(0);
-			(NodeEndpoint {
+			NodeEndpoint {
 				address,
 				udp_port: from.port()
-			}, true)
+			}
 		};
 		let ping_to = NodeEndpoint::from_rlp(&rlp.at(2)?)?;
 		let timestamp: u64 = rlp.val_at(3)?;
@@ -555,7 +555,7 @@ impl<'a> Discovery<'a> {
 		self.send_packet(PACKET_PONG, from, &response.drain())?;
 
 		let entry = NodeEntry { id: *node_id, endpoint: pong_to.clone() };
-		if !entry.endpoint.is_valid_sync_node() && !bootnode {
+		if !entry.endpoint.is_valid_discovery_node() {
 			debug!(target: "discovery", "Got bad address: {:?}", entry);
 		} else if !self.is_allowed(&entry) {
 			debug!(target: "discovery", "Address not allowed: {:?}", entry);
--- a/util/network-devp2p/src/discovery.rs
+++ b/util/network-devp2p/src/discovery.rs
@@ -555,7 +555,7 @@ impl<'a> Discovery<'a> {
 		self.send_packet(PACKET_PONG, from, &response.drain())?;
 
 		let entry = NodeEntry { id: *node_id, endpoint: pong_to.clone() };
-		if !entry.endpoint.is_valid() && !bootnode {
+		if !entry.endpoint.is_valid_sync_node() && !bootnode {
 			debug!(target: "discovery", "Got bad address: {:?}", entry);
 		} else if !self.is_allowed(&entry) {
 			debug!(target: "discovery", "Address not allowed: {:?}", entry);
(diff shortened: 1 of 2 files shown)


dvdplm:
cc @fjl

fjl:
The 'to' field in Ping shouldn't be used at all to determine where to send the response. I'd recommend send Pong back to the UDP packet source address unconditionally. 

The only thing 'to' is useful for is checking whether the remote node is aware of its own endpoint.

util/network-devp2p/src/discovery.rs
```diff
@@ -518,7 +518,14 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_from = if let Ok(rlp) = rlp.at(1) {
+			NodeEndpoint::from_rlp(&rlp)?;
```
tomusdrw: Is this correct? I think it will always return `Ok` since we have more elements in the RLP, but will fail on creating `NodeEndpoint` since the values are `null`.

Any chance we can have a test for this?

ngotchac: Yep, it should handle the `Err` case of `NodeEndpoint::from_rlp`, not the one of `rlp.at(1)`.

util/network-devp2p/src/discovery.rs
```diff
@@ -518,7 +518,14 @@ impl<'a> Discovery<'a> {
...
+		} else {
+			NodeEndpoint {
+				address: from.clone(),
+				udp_port: from.port()
```
ngotchac: I'm not 100% sure this would be correct. The issue is that here the TCP port would be set as the UDP port, and the node will be added as a syncing node even though we can infer from the empty `ping_from` that it isn't.
We should have NodeEnpoint as: IP from `from`, UDP port from `from` and TCP port as `null` or `0`.
However, having this will fail the `endpoint.is_valid` method, because of the null TCP port. So we should also handle this case as a _bootnode only_ endpoint IMO

fjl:
Sorry, misread the code a bit. The PR claims to fix #10613, I guess the fix is that early return is avoided if the 'from' field is invalid?

seunlanlege:
@fjl yes

util/network-devp2p/src/discovery.rs
```diff
@@ -169,6 +169,9 @@ pub struct Discovery<'a> {
 	discovery_nodes: HashSet<NodeId>,
 	node_buckets: Vec<NodeBucket>,
 
+	// these nodes can't be used for syncing.
+	bootnodes: HashSet<NodeId>,
```
tomusdrw: I still don't understand why do we need this? Those nodes doesn't seem to be added to `NodeBucket`s so they never going to be returned by `FIND_NODES` query.

So it seems that we are solving the problem here, but only partially by accepting `PING`s from nodes without TCP port, but not handling them properly as part of the discoverable network.

util/network-devp2p/src/discovery.rs
```diff
@@ -275,7 +279,12 @@ impl<'a> Discovery<'a> {
 			if let Some(node) = node_to_ping {
 				self.try_ping(node, PingReason::Default);
 			};
-            TableUpdates{added, removed: HashSet::new()}
+
+            if !self.bootnodes.contains(&node_entry.id) {
```
tomusdrw: spaces :fire:

util/network-devp2p/src/discovery.rs
tomusdrw: Also note that this collection is unbounded, so we will slowly leak memory here. It's exploitable by sending a bunch of PINGs from random ports.

seunlanlege: @tomusdrw the node will be added to `NodeBucket`

https://github.com/paritytech/parity-ethereum/pull/10705/files/783a66df3bf69b52f4372fcd32efc7f538451b6b#diff-3193558172eb0ec141a7afd53b7a1118R272

we're simply preventing the node from being added to the `Host::nodes` which is used for syncing.

and yes, the set is unbounded. wanted to open discussion on the PR on what conditions it should be cleared.

tomusdrw: I see, but I don't get why we need a separate set for this, why can't we have a `has_tcp_port()` method on the `entry.endpoint` itself similar to how we have `is_valid` now?

util/network-devp2p/src/discovery.rs
```diff
@@ -275,7 +279,12 @@ impl<'a> Discovery<'a> {
...
+				Some(TableUpdates { added, removed: HashSet::new() })
```
tomusdrw: Instead we could just set `added` to be an empty `HashMap` I suppose. (i.e. the condition could be moved higher to just prevent adding to `added` collection.

util/network-devp2p/src/discovery.rs
tomusdrw: Maybe we could also rename the methods to `is_valid_sync_node()` & `is_valid_discovery_node()` for clarity.

util/network-devp2p/src/discovery.rs
seunlanlege: that would unnecessarily call `update_node` with an empty `TableUpdate`

util/network-devp2p/src/node_table.rs
```diff
@@ -104,6 +104,10 @@ impl NodeEndpoint {
 		self.to_rlp(rlp);
 	}
 
+	pub fn is_valid_sync_node(&self) -> bool {
+		self.address.port() != 0
+	}
```
ngotchac: That's not really the definition of `valid`, it should include the checks of UDP port and IP address.
Maybe have `is_valid_sync_node` and `is_valid_discovery_node` (removing `is_valid`) would help make this distinction, with `is_valid_sync_node` being `is_valid_discovery_node() && tcp_port != 0`

util/network-devp2p/src/node_table.rs
```diff
@@ -104,9 +104,13 @@ impl NodeEndpoint {
 		self.to_rlp(rlp);
 	}
 
+	pub fn is_valid_sync_node(&self) -> bool {
```
tomusdrw: Missing docs

util/network-devp2p/src/node_table.rs
```diff
@@ -104,9 +104,13 @@ impl NodeEndpoint {
...
+		self.is_valid_discovery_node() && self.address.port() != 0
+	}
+
 	/// Validates that the port is not 0 and address IP is specified
-	pub fn is_valid(&self) -> bool {
-		self.udp_port != 0 && self.address.port() != 0 &&
+	pub fn is_valid_discovery_node(&self) -> bool {
```
tomusdrw: Docs could do with an update.

util/network-devp2p/src/discovery.rs
```diff
@@ -540,7 +555,7 @@ impl<'a> Discovery<'a> {
 		self.send_packet(PACKET_PONG, from, &response.drain())?;
 
 		let entry = NodeEntry { id: *node_id, endpoint: pong_to.clone() };
-		if !entry.endpoint.is_valid() {
+		if !entry.endpoint.is_valid_sync_node() && !bootnode {
```
ngotchac: I think we could just use `is_valid_bootnode` here, right?

util/network-devp2p/src/discovery.rs
```diff
@@ -728,7 +743,7 @@ impl<'a> Discovery<'a> {
 		trace!(target: "discovery", "Got {} Neighbours from {:?}", results_count, &from);
 		for r in rlp.at(0)?.iter() {
 			let endpoint = NodeEndpoint::from_rlp(&r)?;
-			if !endpoint.is_valid() {
+			if !endpoint.is_valid_sync_node() {
```
ngotchac: Shouldn't it be `is_valid_bootnode` here?

util/network-devp2p/src/discovery.rs
seunlanlege: `endpoint.is_valid_sync_node == endpoint.is_valid`

the extra check just asserts that if there was an error in getiing `NodeEndpoint` out of the packet, its a bootnode otherwise a peer sent us a bad address.

util/network-devp2p/src/discovery.rs
seunlanlege: `endpoint.is_valid_sync_node == endpoint.is_valid`

ngotchac: Well that's kind of my point. Here we are receiving neighbors, and I guess it would be possible that some of those are actually _bootnodes_. I didn't think of that earlier, but here I think you should revisit `let endpoint = NodeEndpoint::from_rlp(&r)?;` and `endpoint.is_valid_sync_node` to take this case into account: we want to add those to our discovery list, without adding them to our syncing list.

util/network-devp2p/src/discovery.rs
ngotchac: Well if you think about it, what you are checking here is whether the received entry's endpoint is a _syncing node_ or _bootnode_ really. As every _syncing node_ are actually _bootnodes_, you can just check whether the node is a _bootnode_, and skip the `bootnode` `bool` var, yeah?

seunlanlege: makes sense

util/network-devp2p/src/discovery.rs
```diff
@@ -728,7 +743,7 @@ impl<'a> Discovery<'a> {
 		trace!(target: "discovery", "Got {} Neighbours from {:?}", results_count, &from);
 		for r in rlp.at(0)?.iter() {
 			let endpoint = NodeEndpoint::from_rlp(&r)?;
```
ngotchac: Won't this still fail on bootnodes?

seunlanlege: yes, it most likely will, but we have no way of getting the socket address here.

seunlanlege: @ngotchac

ngotchac: I'm wondering what would happen if a node is sent here as a neighbour with a UDP port but no TCP port? It shouldn't fail, as you can still get discovery from it. If it fails, something needs to be done so it doesn't.

ngotchac: My bad, I thought this would error on empty TCP port. All good then.

util/network-devp2p/src/discovery.rs
```diff
@@ -518,7 +522,18 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_from = if let Ok(node_endpoint) = NodeEndpoint::from_rlp(&rlp.at(1)?) {
+			node_endpoint
+		} else {
+			let mut address = from.clone();
+			// address here is the node's tcp port. If we are unable to get the `NodeEndpoint` from the `ping_from`
+			// field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
```
dvdplm:
```suggestion
			// rlp field then this is most likely a BootNode, set the tcp port to 0 because it can not be used for syncing.
```

util/network-devp2p/src/node_table.rs
```diff
@@ -104,10 +104,16 @@ impl NodeEndpoint {
 		self.to_rlp(rlp);
 	}
 
-	/// Validates that the port is not 0 and address IP is specified
-	pub fn is_valid(&self) -> bool {
-		self.udp_port != 0 && self.address.port() != 0 &&
-		match self.address {
+	/// validates that the tcp port is not 0 && is_valid_discovery_node
```
dvdplm:
```suggestion
	/// Validates that the tcp port is not 0 and that the node is a valid discovery node (i.e. `is_valid_discovery_node()` is true).
```

util/network-devp2p/src/node_table.rs
```diff
@@ -104,10 +104,16 @@ impl NodeEndpoint {
...
+	/// sync happens over tcp
```
dvdplm:
```suggestion
	/// Sync happens over tcp.
```

util/network-devp2p/src/discovery.rs
```diff
@@ -518,7 +522,18 @@ impl<'a> Discovery<'a> {
 
 	fn on_ping(&mut self, rlp: &Rlp, node_id: &NodeId, from: &SocketAddr, echo_hash: &[u8]) -> Result<Option<TableUpdates>, Error> {
 		trace!(target: "discovery", "Got Ping from {:?}", &from);
-		let ping_from = NodeEndpoint::from_rlp(&rlp.at(1)?)?;
+		let ping_from = if let Ok(node_endpoint) = NodeEndpoint::from_rlp(&rlp.at(1)?) {
+			node_endpoint
+		} else {
```
dvdplm: I'd add a comment here on where such pings come from and why we're not simply rejecting them.

util/network-devp2p/src/node_table.rs
```diff
@@ -104,10 +104,16 @@ impl NodeEndpoint {
...
+	pub fn is_valid_sync_node(&self) -> bool {
+		self.is_valid_discovery_node() && self.address.port() != 0
+	}
+
+	/// Validates that the udp port is not 0 and address IP is specified
```
dvdplm:
```suggestion
	/// Validates that the udp port is not 0 and address IP is specified.
```

util/network-devp2p/src/node_table.rs
```diff
@@ -104,10 +104,16 @@ impl NodeEndpoint {
...
+	/// peer discovery happens over udp
```
dvdplm:
```suggestion
	/// Peer discovery happens over udp.
```

holiman:
Could you LMK when this is rolled out on the bootnodes? Currently, I get timeouts trying to check them
```
INFO [06-20|11:18:13.129] New local node record                    seq=1 id=9e0289db6e78a121 ip=127.0.0.1 udp=35546 tcp=0
node didn't respond: RPC timeout
INFO [06-20|11:18:13.670] New local node record                    seq=1 id=b183029189ff0be4 ip=127.0.0.1 udp=44773 tcp=0
node didn't respond: RPC timeout
INFO [06-20|11:18:14.209] New local node record                    seq=1 id=6eb9f21b386f3185 ip=127.0.0.1 udp=60007 tcp=0
node didn't respond: RPC timeout
INFO [06-20|11:18:14.747] New local node record                    seq=1 id=6bba547ab04d6e52 ip=127.0.0.1 udp=56096 tcp=0
node didn't respond: RPC timeout
```

dvdplm:
> Could you LMK when this is rolled out on the bootnodes?

@holiman should be up now. Thank you for reporting this! :)

PR #10710, opened 2019-06-01
[devp2p] Fix warnings and re-org imports

dvdplm:
Recent nigtlies warns about Trait objs without `dyn` deprecation.
Also re-organise imports.


util/network-devp2p/src/lib.rs
```diff
@@ -60,45 +60,49 @@
...
+extern crate lru_cache;
+extern crate mio;
+extern crate parity_bytes;
+extern crate parity_crypto as crypto;
+extern crate parity_path;
+extern crate parity_snappy as snappy;
+extern crate parking_lot;
+extern crate rand;
+extern crate rlp;
+//TODO: remove this
```
ordian: was it referring to `ansi_term`?

dvdplm: Yeah (although it sort of applies to `rustc-hex` version 1 as well!)

Issue #10698, opened 2019-05-29
[ethcore-network-devp2p] get rid of rust-crypto dependency


ordian:
Also, handle errors more gracefully: cc https://github.com/paritytech/parity-ethereum/pull/10670#discussion_r285012838.

PR #10714, opened 2019-06-03
[devp2p] Don't use `rust-crypto`

dvdplm:
~~Builds on #10710 so please review that first to reduce noise~~

Fixes https://github.com/paritytech/parity-ethereum/issues/10698 and is essentially a copy&paste from @cheme s "crypto-only` branch (https://github.com/paritytech/parity-ethereum/compare/master...cheme:crypto-only).

Attempts to kick the crypto debate forward by replacing `rust-crypto` with `parity-crypto` (which in turn uses the "mini crates" from RustCrypto). While much of the same criticism of `rust-crypto` still applies, I still think we're better off with code that is maintained than being stuck on code from 2016.

Using `parity-crypto` may also help to keep our crypto code in one place (and @cheme branch goes a lot further in this, but I think it's good to do this step-by-step, so expect further PRs in this sense).


util/network-devp2p/src/connection.rs
```diff
@@ -317,14 +315,13 @@ impl EncryptedConnection {
...
-		let iv = vec![0u8; 16];
-		let encoder = CtrMode::new(AesSafe256Encryptor::new(&key_material[32..64]), iv);
-		let iv = vec![0u8; 16];
-		let decoder = CtrMode::new(AesSafe256Encryptor::new(&key_material[32..64]), iv);
-
-		let key_material_keccak = keccak(&key_material);
-		(&mut key_material[32..64]).copy_from_slice(key_material_keccak.as_bytes());
-		let mac_encoder = EcbEncryptor::new(AesSafe256Encryptor::new(&key_material[32..64]), NoPadding);
+		// TODO: clarify this: ecdh::agree creates a **NEW** secret right? And AesCtr256 keeps an internal counter, right?
+		// Using a 0 IV with CTR is fine as long as the same IV is never reused with the same key. This is not the case here.
```
cheme: if ecdh creates a new secret which will be the symetric key used only for this session the 0 iv is only use once with this secret, so we are in the case of same iv use for different key.

util/network-devp2p/src/connection.rs
```diff
@@ -360,51 +357,50 @@ impl EncryptedConnection {
...
 		let mut header = header.out();
-		let padding = (16 - (payload.len() % 16)) % 16;
 		header.resize(16, 0u8);
-
-		let mut packet = vec![0u8; 32 + payload.len() + padding + 16];
-		self.encoder.encrypt(&mut RefReadBuffer::new(&header), &mut RefWriteBuffer::new(&mut packet), false).expect("Invalid length or padding");
-		EncryptedConnection::update_mac(&mut self.egress_mac, &mut self.mac_encoder,  &packet[0..16]);
-		self.egress_mac.clone().finalize(&mut packet[16..32]);
-		self.encoder.encrypt(&mut RefReadBuffer::new(payload), &mut RefWriteBuffer::new(&mut packet[32..(32 + len)]), padding == 0).expect("Invalid length or padding");
+		let header_len = header.len();
```
cheme: it was resized the line above, so I would think it is 16 always (we could use a constant for it).

util/network-devp2p/src/connection.rs
```diff
@@ -360,51 +357,50 @@ impl EncryptedConnection {
 		if len > MAX_PAYLOAD_SIZE {
 			bail!(ErrorKind::OversizedPacket);
 		}
+
 		header.append_raw(&[(len >> 16) as u8, (len >> 8) as u8, len as u8], 1);
 		header.append_raw(&[0xc2u8, 0x80u8, 0x80u8], 1);
-		//TODO: get rid of vectors here
+		let padding = (16 - (len % 16)) % 16;
+
+		let mut packet = vec![0u8; 32 + len + padding + 16];
```
cheme: I would propose to switch 32 to 16 + 16 as writing is done in two steps bellow.

util/network-devp2p/src/connection.rs
```diff
@@ -317,14 +315,13 @@ impl EncryptedConnection {
...
-
-		let key_material_keccak = keccak(&key_material);
-		(&mut key_material[32..64]).copy_from_slice(key_material_keccak.as_bytes());
-		let mac_encoder = EcbEncryptor::new(AesSafe256Encryptor::new(&key_material[32..64]), NoPadding);
+		// TODO: clarify this: ecdh::agree creates a **NEW** secret right? And AesCtr256 keeps an internal counter, right?
+		// Using a 0 IV with CTR is fine as long as the same IV is never reused with the same key. This is not the case here.
+		let encoder = AesCtr256::new(&key_material[32..64], &NULL_IV)?;
+		let decoder = AesCtr256::new(&key_material[32..64], &NULL_IV)?;
+        let key_material_keccak = keccak(&key_material);
+        (&mut key_material[32..64]).copy_from_slice(key_material_keccak.as_bytes());
```
cheme: This last copy seems useless (except to show a repeating pattern with previous lines, which can make sense and then we rely on optimization). Not sure if it is worth removing.

dvdplm: Yeah it seems like we could use `key_material_keccak` everywhere we use `key_material[32..64]` below. I'll leave it as it is because I don't understand it fully.

util/network-devp2p/src/connection.rs
```diff
@@ -317,14 +315,15 @@ impl EncryptedConnection {
...
-		let key_material_keccak = keccak(&key_material);
-		(&mut key_material[32..64]).copy_from_slice(key_material_keccak.as_bytes());
-		let mac_encoder = EcbEncryptor::new(AesSafe256Encryptor::new(&key_material[32..64]), NoPadding);
+		// Using a 0 IV with CTR is fine as long as the same IV is never reused with the same key.
+		// This is the case here: ecdh creates a new secret which will be the symmetric key used
+		// only for this session the 0 IV is only use once with this secret, so we are in the case
+		// of same IV use for different key.
+		let encoder = AesCtr256::new(&key_material[32..64], &NULL_IV)?;
+		let decoder = AesCtr256::new(&key_material[32..64], &NULL_IV)?;
+        let key_material_keccak = keccak(&key_material);
```
ordian: illegal whitespaces detected (and below)

util/network-devp2p/src/connection.rs
```diff
@@ -355,56 +354,55 @@ impl EncryptedConnection {
...
+		const HEADER_LEN: usize = 16;
 		let mut header = RlpStream::new();
 		let len = payload.len();
 		if len > MAX_PAYLOAD_SIZE {
 			bail!(ErrorKind::OversizedPacket);
 		}
+
 		header.append_raw(&[(len >> 16) as u8, (len >> 8) as u8, len as u8], 1);
 		header.append_raw(&[0xc2u8, 0x80u8, 0x80u8], 1);
-		//TODO: get rid of vectors here
```
ordian: did we actually do that?

cheme: we could put packet as a field of this struct, that would be pretty straightforward, but I am really not sure about the lifetime of this one (and if we want to keep memory of unusually large packet allocated).

dvdplm: I removed the comment because like @cheme I was unsure of what the original author of the comment actually intended with it: is to be read like "it sure would be nice to allocate a bit less here" or "this is a terrible performance problem". Such comments seem unhelpful to me so I removed it.

cheme:
My first impression is that this part of the code is not really a bottleneck. A good test would be to try to compare a standard build of this fork with a build with aes-ni activated (since aes-ni activation really show some huge performance changes).

util/network-devp2p/src/connection.rs
dvdplm: Sigh. I think this comes from applying review suggestions. :/

dvdplm:
> I don't know what are the performance implications of that change? And can it be a bottleneck really?

I do not think so but the truth is we just don't know much about how the network layer is performing, or how that performance impacts the overall performance of the client in general (i.e. is a 10% slowdown or speedup a 1% slowdown or speedup in "normal" operation? more? less?).

The purpose of the PR is to slooooowly start tidying up our usage of crypto and use `parity-crypto` as much as possible. @cheme did a lot of more good work on this but we sort of lost sight of it and it became (somewhat) stale. My plan was to revive that work piece by piece and this was the first step.

dvdplm:
> with aes-ni activated

My understanding was that this was activated autiomatically ([Crate switches between implementations automatically at compile time. (i.e. it does not use run-time feature detection)](https://docs.rs/aes/0.3.2/aes/)), but maybe I'm wrong. Hmm.

dvdplm:
The `aesni` crate is **not** activated by default or automatically. We need to build the whole client with the proper `RUSTFLAGS` or switch to runtime detection with the `is_x86_feature_detected!("aes")` macro.

PR #10719, opened 2019-06-05
Use RUSTFLAGS to set the optimization level

dvdplm:
Cargo has a [quirk](https://github.com/rust-lang/cargo/issues/4438) in how configuration settings are propagated when `cargo test` runs: local code respect the settings in `[profile.test]` but all dependencies use the `[profile.dev]` settings. Here we force `opt-level=3` for all dependencies.


Cargo.toml
```diff
@@ -117,12 +117,6 @@ path = "parity/lib.rs"
 path = "parity/main.rs"
 name = "parity"
 
-[profile.dev]
-
-[profile.test]
-lto = false
-opt-level = 3 # makes tests slower to compile, but faster to run
-
```
TriplEight: This might work: RUSTFLAGS="-C opt-level=3 -C overflow-checks=off -C incremental=false -C debuginfo=2" time cargo test -v
##

These are the settings for release mode I think:

# The release profile, used for `cargo build --release` (and the dependencies
# for `cargo test --release`, including the local library or binary).
[profile.release]
opt-level = 3
debug = false
rpath = false
lto = false
debug-assertions = false
codegen-units = 16
panic = 'unwind'
incremental = false
overflow-checks = false

TriplEight:
>dvdplm
So cargo test --release will use the settings from [profile.release] which with androniks change includes lto = true, which in turn makes compilation take a really long time (but binary size is smaller). So for tests, where we do care about fast run speeds, we want the code optimizations from release mode but not lto = true.


>The first idea was to stop passing --release to cargo test and instead just add opt-level = 3 to [profile.test]. That didn't work because cargo applies opt-level =3 only to the local code but not to dependencies. For dependencies it uses [profile.dev] (which is kind of crazy and probably a bug). To get around that we tried passing in the optimization level with RUST_FLAGS but as Kiril points out, we probably need more flags than just -C opt-level = 3.

dvdplm:
@TriplEight pushed #d654d255f to see what happens.

TriplEight:
I think it's still "not enough" optimized. 
Since you are passing opt-level=3 to everything including deps, it must be faster, but it's still not. 
There have to be something missing.

scripts/gitlab/test-linux.sh
```diff
@@ -14,4 +14,4 @@ rustup default $1
 rustup show
 
 echo "________Running Parity Full Test Suite________"
-time cargo test $OPTIONS --features "$FEATURES" --locked --all --target $CARGO_TARGET --verbose --color=always -- --test-threads $THREADS
+CARGO_INCREMENTAL=0 RUSTFLAGS="-C opt-level=3 -C overflow-checks=off -C debuginfo=2" time cargo test $OPTIONS --features "$FEATURES" --locked --all --target $CARGO_TARGET --verbose --color=always -- --test-threads $THREADS
```
ordian: why disable overflow-checks for tests? they can point to a bug more precisely
also, not introduced by this PR, but shall we remove `--color=always`?

dvdplm: It's the fruit of some sleuthing as to what `--release` actually means and `overflow-checks=off` is one of the implied settings.

ordian: you mean more compiled artifacts can be reused if this option is set?

.gitlab-ci.yml
```diff
@@ -127,6 +127,8 @@ test-linux:
   script:
     - ./scripts/gitlab/test-linux.sh stable
     - sccache -s
+  tags:
```
dvdplm: NOTE TO SELF: remove before merge

scripts/gitlab/test-linux.sh
TriplEight: But `overflow-checks` didn't seem to work as a flag, did  it?
What's wrong with `--color=always`? It was set explicitly for readability of logs in CI. Because somewhy if you set `--verbose`, the `--color` defaults to `never`.

dvdplm: > But `overflow-checks` didn't seem to work as a flag, did it?

It **does** work as a flag. What is odd about it is that `overflow-checks = false` is valid in `Cargo.toml` but **not** as a commandline opt (i.e. `-C overflow-checks=false` is an error).

> What's wrong with --color=always

I think @ordian wants the escape codes out of the raw logs. I'm indifferent.

TriplEight: So it should be `-C overflow-checks=off` to work as a flag?

Yes, `--color` shits codes in raw logs. Btw, we could get rid of `--verbose` if it's not needed any more.

dvdplm: > So it should be `-C overflow-checks=off` to work as a flag?

Yes.

> Btw, we could get rid of --verbose if it's not needed any more.

One thing at a time. :)

Issue #10688, opened 2019-05-22
Bad block detected: TimestampOverflow


sebaponti:
- **Parity Ethereum version**: 2.4.6-stable
- **Operating system**: Linux Ubuntu 16.04.6
- **Installation**: Docker image
- **Fully synchronized**: yes
- **Network**: Custom private network
- **Restarted**: no

I have configured a private network with PoA Aura and the follow chain spec:
```json
{
  "name": "Chain Name",
  "engine": {
    "authorityRound": {
      "params": {
        "blockReward": 0,
        "stepDuration": 5,
        "maximumEmptySteps": 24,
        "emptyStepsTransition": 0,
        "maximumUncleCountTransition": 0,
        "maximumUncleCount": 0,
        "validators": {
          "multi": {
            "0": {
              "list": [
                "0x..."
              ]
            },
            "5": {
              "safeContract": "0x0000000000000000000000000000000000000005"
            }
          }
        }
      }
    }
  },
  "params": {
    "gasLimitBoundDivisor": "0x400",
    "maximumExtraDataSize": "0x20",
    "minGasLimit": "0x1388",
    "networkID": "0x1E61",
    "eip140Transition": 0,
    "eip211Transition": 0,
    "eip214Transition": 0,
    "eip658Transition": 0,
    "eip145Transition": 0,
    "eip1014Transition": 0,
    "eip1052Transition": 0
  },
...
```
From block 6 onwards I have added 2 more validators and after a good working time for some reason 2 of those 3 validators started to show a warning message: 
```bash
5/22/2019 8:51:38 AM2019-05-22 11:51:38 UTC IO Worker #0 WARN client  Stage 1 block verification failed for 0x0eec…b457: Error(Block(TimestampOverflow), State { next_error: None, backtrace: InternalBacktrace { backtrace: Some(stack backtrace:
5/22/2019 8:51:38 AM   0:     0x56176aa8b19d - <no info>
5/22/2019 8:51:38 AM   1:     0x56176aa8a562 - <no info>
5/22/2019 8:51:38 AM   2:     0x56176a7e3ec8 - <no info>
5/22/2019 8:51:38 AM   3:     0x56176a7e41f5 - <no info>
5/22/2019 8:51:38 AM   4:     0x56176a2b3f9e - <no info>
5/22/2019 8:51:38 AM   5:     0x56176a228376 - <no info>
5/22/2019 8:51:38 AM   6:     0x56176a14b5d5 - <no info>
[...]

PR #10720, opened 2019-06-05
fix: aura don't add `SystemTime::now()`

niklasad1:
Attempt to close https://github.com/paritytech/parity-ethereum/issues/10688

In AURA we add`SystemTime::now` again which cases a Timestamp Overflow which this fixes

### DIFF
--- a/ethcore/src/engines/authority_round/mod.rs
+++ b/ethcore/src/engines/authority_round/mod.rs
@@ -22,7 +22,7 @@ use std::iter::FromIterator;
 use std::ops::Deref;
 use std::sync::atomic::{AtomicUsize, AtomicBool, Ordering as AtomicOrdering};
 use std::sync::{Weak, Arc};
-use std::time::{UNIX_EPOCH, SystemTime, Duration};
+use std::time::{UNIX_EPOCH, Duration};
 
 use block::*;
 use client::EngineClient;
@@ -42,13 +42,11 @@ use itertools::{self, Itertools};
 use rlp::{encode, Decodable, DecoderError, Encodable, RlpStream, Rlp};
 use ethereum_types::{H256, H520, Address, U128, U256};
 use parking_lot::{Mutex, RwLock};
+use time_utils::CheckedSystemTime;
 use types::BlockNumber;
 use types::header::{Header, ExtendedHeader};
 use types::ancestry_action::AncestryAction;
 use unexpected::{Mismatch, OutOfBounds};
-
-#[cfg(not(time_checked_add))]
-use time_utils::CheckedSystemTime;
 
 mod finality;
 
@@ -584,10 +582,10 @@ fn verify_timestamp(step: &Step, header_step: u64) -> Result<(), BlockError> {
 			// Returning it further won't recover the sync process.
 			trace!(target: "engine", "verify_timestamp: block too early");
 
-			let now = SystemTime::now();
-			let found = now.checked_add(Duration::from_secs(oob.found)).ok_or(BlockError::TimestampOverflow)?;
-			let max = oob.max.and_then(|m| now.checked_add(Duration::from_secs(m)));
-			let min = oob.min.and_then(|m| now.checked_add(Duration::from_secs(m)));
+			let found = CheckedSystemTime::checked_add(UNIX_EPOCH, Duration::from_secs(oob.found))
+				.ok_or(BlockError::TimestampOverflow)?;
+			let max = oob.max.and_then(|m| CheckedSystemTime::checked_add(UNIX_EPOCH, Duration::from_secs(m)));
+			let min = oob.min.and_then(|m| CheckedSystemTime::checked_add(UNIX_EPOCH, Duration::from_secs(m)));
 
 			let new_oob = OutOfBounds { min, max, found };
 
--- a/ethcore/src/verification/verification.rs
+++ b/ethcore/src/verification/verification.rs
@@ -40,7 +40,6 @@ use types::{BlockNumber, header::Header};
 use types::transaction::SignedTransaction;
 use verification::queue::kind::blocks::Unverified;
 
-#[cfg(not(time_checked_add))]
 use time_utils::CheckedSystemTime;
 
 /// Preprocessed block data gathered in `verify_block_unordered` call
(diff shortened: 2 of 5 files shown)


ethcore/src/error.rs
```diff
@@ -134,12 +134,12 @@ pub enum BlockError {
...
 /// Newtype for Display impl to show seconds
 #[derive(Debug, Clone, From, PartialEq, Eq)]
-pub struct OutOfBoundsTime(OutOfBounds<SystemTime>);
+pub struct OutOfBoundsTime(OutOfBounds<Duration>);
 
 impl fmt::Display for OutOfBoundsTime {
 	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
 		let seconds = self.0
-			.map(|st| st.elapsed().unwrap_or_default().as_secs());
+			.map(|st| st.as_secs());
```
niklasad1: This is not the same anymore, because before it printed out the time elapsed from that the `SystemTime` was created

The change just prints out it the duration as secs

dvdplm:
> A follow-up PR can be made and remove `time-utils-library` because we are at `Rust 1.35` and it `SystemTime::checked_**` was stabilized in `Rust 1.34`

Can you file an issue please?

ethcore/src/verification/verification.rs
```diff
@@ -308,10 +305,13 @@ pub fn verify_header_params(header: &Header, engine: &EthEngine, is_full: bool,
...
-		let max_time = SystemTime::now() + ACCEPTABLE_DRIFT;
-		let invalid_threshold = max_time + ACCEPTABLE_DRIFT * 9;
-		let timestamp = UNIX_EPOCH.checked_add(Duration::from_secs(header.timestamp()))
-			.ok_or(BlockError::TimestampOverflow)?;
+		let now = SystemTime::now();
+		let now = now.duration_since(UNIX_EPOCH).expect("now is after UNIX_EPOCH; qed");
+		let max_time = now.checked_add(ACCEPTABLE_DRIFT)
+			.unwrap_or(Duration::from_secs(u64::max_value()));
+		let invalid_threshold = max_time.checked_add(ACCEPTABLE_DRIFT * 9)
+			.unwrap_or(Duration::from_secs(u64::max_value()));
```
folsen: This is not the same behavior as before is it? It used to error out and now it just silently proceeds with `u64::max_value()`, is that an issue?

niklasad1: Yeah, I think the only change is that if `timestamp == u64::max_val() && max_time == u64::max_val() && invalid_threshold == u64::max_val() ` then we won't detect an error. 

However, before we returned `TimestampOverflow` when timestamp >  `i32::max_value`

Also, it should not be possible to force `SystemClock` to u64::max_value because it is represented i32/i64 on the systems I know of

Might be better to keep `max_time` and `invalid_threshold` as Options and return some other error when both are None?

folsen: As long as it doesn't introduce any consensus-relevant difference, e.g. an old chain with a block with a forged timestamp used to be invalid there and is valid now (or vice versa) so that there could be a split depending on if you reject or if you cap out.

niklasad1: No, as long as we are not near `2038` it should be safe :)

niklasad1:
@folsen 

> As long as it doesn't introduce any consensus-relevant difference, e.g. an old chain with a block with a forged timestamp used to be invalid there and is valid now (or vice versa) so that there could be a split depending on if you reject or if you cap out.

After thinking about this a bit I think it is possible to have consensus issues.

Before: we reject timestamps if it is bigger than `i32::max_value` (i.e, converting it with SystemTime::now().checked_add(timestamp). However it is not always true because `SystemTime` is platform dependent (usually i32 or i64). If we use our own [CheckedSystemTime](https://github.com/paritytech/parity-ethereum/blob/master/util/time-utils/src/lib.rs#L20-#L26) it will always be i32 but rust versions after `1.34` will use the `stdlib which is platform dependent`

In this PR: Timestamp is represented as Duration and it will be platform independent (secs as `u64` internally)

Thus, both might cause consensus issues.

I can revert the `Duration` stuff and just remove the issue in `verify_timestamp`?

dvdplm:
@andresilva do you have any comment on this?

folsen:
@niklasad1 I guess we also have to weigh it with how likely we think that any such blocks exist in reality. Given we can't see all chains using Aura, it's hard to judge, but we also have the nice property that Aura chains aren't really adversarial in the same way that a PoW chain is. In PoW you could cause the fork if you could mine a block with the wrong timestamp (by manipulating the system or the block directly). But in an Aura chain we have better assumptions on validator honesty but also have the property of most/all of them running a single client.

niklasad1:
> `CheckedSystemTime`'s aim seems to be to fix the resolution to `i32` so that we don't have consensus issues, not sure if replacing this is with `Duration` is worth it or not. cc @tomusdrw

Yes, expect we have the following

```rust
#[cfg(not(time_checked_add))]
use time_utils::CheckedSystemTime;
```

rust >= rust 1.34 it will use the standard library and be platform dependent.
Thus, might have consensus issues.

If I revert Duration and remove `#[cfg(not(time_checked_add))]` good to go?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -584,10 +582,10 @@ fn verify_timestamp(step: &Step, header_step: u64) -> Result<(), BlockError> {
 			// Returning it further won't recover the sync process.
 			trace!(target: "engine", "verify_timestamp: block too early");
 
-			let now = SystemTime::now();
-			let found = now.checked_add(Duration::from_secs(oob.found)).ok_or(BlockError::TimestampOverflow)?;
-			let max = oob.max.and_then(|m| now.checked_add(Duration::from_secs(m)));
-			let min = oob.min.and_then(|m| now.checked_add(Duration::from_secs(m)));
+			let found = CheckedSystemTime::checked_add(UNIX_EPOCH, Duration::from_secs(oob.found))
```
niklasad1: We could also saturate `oob.found` to `i32::max_value()` which could eliminate this completely?!

dvdplm:
@niklasad1 is this good to go or are you still poking at it?

niklasad1:
@dvdplm good to go from my side

Issue #10712, opened 2019-06-01
Provide `MissingParent` Error-Recovery


lazaridiscom:
moved to: https://github.com/lazaridiscom/parity-ethereum/issues/2

lazaridiscom:
# Important

overworked a bit, messed some commits up, will restore first thing after wake up

lazaridiscom:
all good, see reproduction section above

lazaridiscom:
# Question

Which functionality is used internally, in order to switch the blockchain to a specific path/fork?

joshua-mir:
https://github.com/paritytech/parity-ethereum/blob/dae5d75dd65d6a51873ecd9c1c4f159f49f2a9f3/ethcore/blockchain/src/blockchain.rs#L1049

You can see the important functions here, a route is created and then the block is inserted with the route - the blockchain reevaluates the `best_block` every insertion.

lazaridiscom:
> /// If the block is already known, does nothing.

problem is the blocks are already in the database.

joshua-mir:
The sentiment among the core-devs seems to be that this train of though is a little counterproductive. 

To elaborate we shouldn't be treating existing blocks as invalid after importing them in any case, as that would mean we imported them without validation at some point, any further chains built on top are ignored. More succinctly -> #10085 itself is the issue we should be looking into solving as it gets us into a scenario we explicitly don't want to handle, we should be fixing that, not handling the repurcussions of it afterwards.

dvdplm:
So, I've tried to use the repro code to see if it can help us pinpoint the problem and unfortunately I cannot trigger the error. Blocks #6942 and #6943 seems to be imported properly and while #6882 does cause `2019-06-06 18:55:43 UTC Verifier #2   TRACE blockchain  #6882 0x28fd…5a04 - 0x273b…0377 HASH UNEQUAL, #6882` it does not trigger a panic or error.

Not sure if I'm doing something wrong here or if the repro does not work on all hosts or…?

lazaridiscom:
@dvdplm, My repo contains the patch which avoids the Error being raised, and additional debug output. Either use the original poanet parity, or enable this line to have the error back:

https://github.com/lazaridiscom/parity-ethereum/blob/aura-pos/ethcore/blockchain/src/blockchain.rs#L1020

the "HASH UNEQUAL" location got additional debug, its an condition where 6942 has two instances.

lazaridiscom:
[...]

PR #10729, opened 2019-06-10
Stop breaking out of loop if a non-canonical hash is found

dvdplm:
Re-opens https://github.com/paritytech/parity-ethereum/pull/10566

### DIFF
--- a/ethcore/blockchain/src/blockchain.rs
+++ b/ethcore/blockchain/src/blockchain.rs
@@ -1010,7 +1010,14 @@ impl BlockChain {
 						.take_while(|t| t.block_number <= details.number)
 						.last()
 				},
-				Some(h) => trace!(target: "blockchain", "Block #{}: Found non-canonical block hash {} (expected {})", details.number, h, hash),
+				Some(h) => {
+					warn!(target: "blockchain", "Block #{}: Found non-canonical block hash {} (expected {})", details.number, h, hash);
+
+					trace!(target: "blockchain", "#{} {} != {} HASH UNEQUAL, #{}", details.number, hash, h, self.block_number(&h).unwrap_or_default() );
+					trace!(target: "blockchain", "      Version A {}: #{:#?}", hash, self.block_details(&hash));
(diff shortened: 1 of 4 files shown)


ethcore/blockchain/src/blockchain.rs
```diff
@@ -1000,11 +1003,15 @@ impl BlockChain {
...
-					.take_while(|t| t.block_number <= details.number)
-					.last()
+			match self.block_hash(details.number) {
+				Some(h) if h == hash => {
+					return self.epoch_transitions()
+						.map(|(_, t)| t)
+						.take_while(|t| t.block_number <= details.number)
+						.last()
+				},
+				Some(h) => trace!(target: "blockchain", "Found non-canonical block hash {}", h),
```
tomusdrw: Can we print `h vs hash` here?

lazaridiscom:
Be aware that this "Stop breaking out of loop" seems to do (at least as it looks for now) more harm than good.

tomusdrw:
@lazaridiscom can you elaborate what you mean? This fixes a lookup of canonical epoch on non-canonical chains that are longer than the current canon chain.

As you've seen when you analyzed #10085 (thanks a lot for your input!) currently we see such long non-canonical chains produced on AuRa.

However, the issue in general is that it should happen really rarely, and the fact that we can't recover from this situation is only partially related to the fix in this PR.

What happens with chains in #10085 is that we see two forks that both reach finality and the never re-organize between one and the other - that should never happen in the first place and we're working on a fix.

lazaridiscom:
Oh, that was just general input, re my reproduction setup. I think I understand your elaborations (despite my limited domain-knowledge)

I'll continue my path tomorrow - this was too much today.

(Btw: not that I have a "bad habit" to edit the text of the issues I file. This means that there is sometimes new stuff in the issues, without pings beeing send!)

lazaridiscom:
@all, https://github.com/paritytech/parity-ethereum/issues/10712 got a draft commit, which will ideally lead tomorrow to a complete fix. cu!

PR #10733, opened 2019-06-10
Print warnings when using dangerous settings for ValidatorSet

dvdplm:
AuRa should not be used with a single validator or with an even number. Print appropriate warnings.


ethcore/src/engines/authority_round/finality.rs
```diff
@@ -110,7 +111,14 @@ impl RollingFinality {
 	/// Returns a list of all newly finalized headers.
 	// TODO: optimize with smallvec.
 	pub fn push_hash(&mut self, head: H256, signers: Vec<Address>) -> Result<Vec<H256>, UnknownValidator> {
-		if signers.iter().any(|s| !self.signers.contains(s)) { return Err(UnknownValidator) }
+		// TODO: seems bad to iterate over signers twice like this.
```
tomusdrw: It's alright. There will usually be only one element in the list (block author), there might be a bit more if there are empty steps.

ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
...
 	/// Create a new `SimpleList`.
 	pub fn new(validators: Vec<Address>) -> Self {
-		SimpleList {
-			validators: validators,
+		let validator_count = validators.len();
+		if validator_count == 1 {
+			warn!(target: "engine", "Running AuRa with a single validator is not supported.");
 		}
+		if validator_count % 2 == 0 {
+			warn!(target: "engine", "Running AuRa with an even number of validators is not supported.");
```
tomusdrw: Well it is supported, but not recommended (because of possible network split)

ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
 impl SimpleList {
 	/// Create a new `SimpleList`.
 	pub fn new(validators: Vec<Address>) -> Self {
-		SimpleList {
-			validators: validators,
+		let validator_count = validators.len();
+		if validator_count == 1 {
+			warn!(target: "engine", "Running AuRa with a single validator is not supported.");
```
tomusdrw: In principle it's fine, but we should emphasize that in such setup we have instant finality and it might be better to use some database instead of a blockchain.

ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
 impl SimpleList {
```
tomusdrw: is simple list used for hardcoded list in specfile as well? (i.e. the case where we don't have a contract, but rather a fixed list from spec)

ethcore/blockchain/src/blockchain.rs
```diff
@@ -1000,11 +1003,22 @@ impl BlockChain {
...
+				Some(h) if h == hash => {
+					return self.epoch_transitions()
+						.map(|(_, t)| t)
+						.take_while(|t| t.block_number <= details.number)
+						.last()
+				},
+				Some(h) => {
+					warn!(target: "blockchain", "Block #{}: Found non-canonical block hash {} (expected {})", details.number, h, hash);
+
+					trace!(target: "blockchain", "Block #{} Mismatched hashes. Ancestor {} != Own {} – Own block #{}", details.number, hash, h, self.block_number(&h).unwrap_or_default() );
```
tomusdrw: doesn't make much sense to call `block_number` again, we know that it's equal to `details.number`, cause that's how we retrieved the hash. Displaying `block_details` of both blocks is enough.

ethcore/blockchain/src/blockchain.rs
```diff
@@ -1000,11 +1003,22 @@ impl BlockChain {
...
+					trace!(target: "blockchain", "      Ancestor {}: #{:#?}", hash, self.block_details(&hash));
```
tomusdrw:
```suggestion
					trace!(target: "blockchain", "      Ancestor {}: #{:#?}", hash, details);
```

ethcore/blockchain/src/blockchain.rs
```diff
@@ -988,7 +989,9 @@ impl BlockChain {
 	pub fn epoch_transition_for(&self, parent_hash: H256) -> Option<EpochTransition> {
 		// slow path: loop back block by block
 		for hash in self.ancestry_iter(parent_hash)? {
+			trace!(target: "blockchain", "Got parent hash {} from ancestry_iter", hash);
 			let details = self.block_details(&hash)?;
+			trace!(target: "blockchain", "Block #{}: Got block details", details.number);
```
tomusdrw: I'd print the `hash` in this log entry as well - it's easier to analyse the logs if you can just filter out one message and don't need the context (i.e. the line above with a hash in this case)

ethcore/blockchain/src/blockchain.rs
```diff
@@ -963,6 +963,7 @@ impl BlockChain {
 	/// Iterate over all epoch transitions.
 	/// This will only return transitions within the canonical chain.
 	pub fn epoch_transitions(&self) -> EpochTransitionIter {
+		trace!(target: "blockchain", "Iterating over all epoch transitions");
```
tomusdrw:
```suggestion
		debug!(target: "blockchain", "Iterating over all epoch transitions");
```

Might be helpful if we collect only `debug` logs initially.

ethcore/src/engines/validator_set/simple_list.rs
dvdplm: Yes, I believe so: https://github.com/paritytech/parity-ethereum/blob/7c335e87642b848f61ab3ee7bbff5964b53d6f11/ethcore/src/engines/validator_set/mod.rs#L49

ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
...
 	/// Create a new `SimpleList`.
 	pub fn new(validators: Vec<Address>) -> Self {
-		SimpleList {
-			validators: validators,
+		let validator_count = validators.len();
+		if validator_count == 1 {
+			warn!(target: "engine", "Running AuRa with a single validator implies instant finality. Use a database?");
 		}
+		if validator_count % 2 == 0 {
+			warn!(target: "engine", "Running AuRa with an even number of validators is not recommented (risk of network split).");
```
ltfschoen: typo `recommented` should be `recommended`

ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
...
 	/// Create a new `SimpleList`.
 	pub fn new(validators: Vec<Address>) -> Self {
-		SimpleList {
-			validators: validators,
+		let validator_count = validators.len();
+		if validator_count == 1 {
+			warn!(target: "engine", "Running AuRa with a single validator implies instant finality. Use a database?");
 		}
+		if validator_count % 2 == 0 {
+			warn!(target: "engine", "Running AuRa with an even number of validators is not recommended (risk of network split).");
```
tomusdrw: Just realised that it might get pretty spammy on misconfigured networks using validator set contract, can we verify that and maybe throttle if that's the case?

dvdplm: Hmm, I think we create `SimpleLists` fairly rarely and if they are running with a not-so-great config shouldn't we tell them about it? On `kovan` this message is shown twice when there are 4 validators; is that too spammy?

dvdplm: @tomusdrw is your concern here still valid?

tomusdrw: All good!

PR #10734, opened 2019-06-10
Log validator set changes in EpochManager

dvdplm:
Builds on https://github.com/paritytech/parity-ethereum/pull/10733


ethcore/src/engines/validator_set/simple_list.rs
```diff
@@ -33,9 +33,14 @@ pub struct SimpleList {
...
 	/// Create a new `SimpleList`.
 	pub fn new(validators: Vec<Address>) -> Self {
-		SimpleList {
-			validators: validators,
+		let validator_count = validators.len();
+		if validator_count == 1 {
+			warn!(target: "engine", "Running AuRa with a single validator implies instant finality. Use a database?");
 		}
+		if validator_count % 2 == 0 {
+			warn!(target: "engine", "Running AuRa with an even number of validators is not recommented (risk of network split).");
```
ltfschoen:
```suggestion
			warn!(target: "engine", "Running AuRa with an even number of validators is not recommended (risk of network split).");
```

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1298,8 +1306,13 @@ impl Engine<EthereumMachine> for AuthorityRound {
...
 				// - This specific check is only relevant if you're importing (since it checks
 				//   against wall clock)
-				if let Ok((_, set_number)) = self.epoch_set(header) {
-					self.validators.report_benign(header.author(), set_number, header.number());
+				if let Ok((set, set_number)) = self.epoch_set(header) {
+					let me = self.address();
+					if set.contains(header.parent_hash(), &me) {
+						self.validators.report_benign(header.author(), set_number, header.number());
+					} else {
+						debug!(target: "engine", "Not reporting benign misbehaviour (cause: InvalidSeal) because we're not part of the validator set, Own address: {}", me);
```
tomusdrw: I'd make it a `warn` - cause it means the node is configured as validator `(--engine-signer)`, but it's not on the list and that's most likely some misconfiguration.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -806,19 +813,25 @@ impl AuthorityRound {
 			return;
 		}
 
-		if let (true, Some(me)) = (current_step > parent_step + 1, self.signer.read().as_ref().map(|s| s.address())) {
+		if let (true, me) = (current_step > parent_step + 1, self.address()) {
 			debug!(target: "engine", "Author {} built block with step gap. current step: {}, parent step: {}",
```
tomusdrw: This will print out the warning even for non-authoirities now, maybe let's check `me.is_zero()` to avoid printing this for regular nodes?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -806,19 +813,25 @@ impl AuthorityRound {
...
+					if skipped_primary != me {
+						// Stop reporting once validators start repeating.
+						if !reported.insert(skipped_primary) { break; }
+						self.validators.report_benign(&skipped_primary, set_number, header.number());
+					} else {
+						trace!(target: "engine", "Primary that skipped is self, not self-reporting. Own address: {}", me);
+					}
+				}
+			} else {
+				warn!(target: "engine", "We are not part of the validator set so not reporting (skipped steps). Own address: {}", self.address());
```
tomusdrw: This definitely shouldn't be printed if `engine_signer` is not configured.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -806,19 +813,25 @@ impl AuthorityRound {
...
-			for step in parent_step + 1..current_step {
-				let skipped_primary = step_proposer(validators, header.parent_hash(), step);
-				// Do not report this signer.
-				if skipped_primary != me {
-					// Stop reporting once validators start repeating.
-					if !reported.insert(skipped_primary) { break; }
-					self.validators.report_benign(&skipped_primary, set_number, header.number());
- 				}
- 			}
+			if validators.contains(header.parent_hash(), &me) {
```
tomusdrw: Since now `me` can be `Address::zero` if signer is not configured I wonder if that wouldn't break some assumptions here if `Address:zero` is indeed a member of `validators` - while I know this shouldn't happen I think I kind of preferred the code when it was using `Option<>`

ethcore/src/engines/authority_round/mod.rs
dvdplm: My thinking here was: something called `report_skipped()` which I assumed shouldn't ever happen unless `engine_signer` is a-ok, so if they're set up to sign, but not part of the validator set then perhaps there's a bug worth knowing about?

tomusdrw:
I just looked at the code and it seems there is some confusion about who can/should issue reports.
I believe the original intention of the code was to report on every node that has `--engine-signer` configured even though it's not part of the validator set and let the contract decide.

I'd split `engine_signer` into: `block_signer` & `reporting_signer` (doesn't need CLI flag initially, but just for code clarity) and then use appropriate signers for different things. So it means that we would allow reporting even if it's not part of the validator set.

PR #10742, opened 2019-06-12
refactor: Refactor evmbin CLI

ltfschoen:
* teststate.json uses the `Test` struct and each name contains the `State` struct (see ethjson package)
* create2callPrecompiles.json https://github.com/ethereum/tests/blob/725dbc73a54649e22a00330bd0f4d6699a5060e5/GeneralStateTests/stCreate2/create2callPrecompiles.json#L2 state test JSON file from Ethereum Foundation that's now in the /res directory, as doesn't return a state root mistmatch error like the teststate.json file does.
* Run `parity-evm` binary using state test for example:
```
cargo build -p evmbin --release
 ./target/release/parity-evm state-test ./evmbin/res/teststate.json --only add12 --chain eip150
 ./target/release/parity-evm state-test ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople
```
* Run stats:
```
./target/release/parity-evm stats --to "0000000000000000000000000000000000000004" --from "0000000000000000000000000000000000000003" --code "05" --input "06" --gas "21" --gas-price "2" --chain "./evmbin/res/testchain.json" --json

Output:
{"depth":1,"gas":"0x21","gasCost":"0x0","memory":"0x","op":5,"opName":"SDIV","pc":0,"stack":[],"storage":{}}
{"error":"EVM: Stack underflow SDIV 2/0","gasUsed":"0x21","time":87}
```
Note: operation SDIV from code: 05 does division and expects two items on the stack, try changing to code: 01 for ADD (addition). For more info do a search for "SDIV" across the codebase and read about all the different EVM bytecode instructions.
* Build docs: `cargo doc -p evmbin --release --open --no-deps --document-private-items`


evmbin/src/info.rs
```diff
@@ -102,13 +102,21 @@ pub fn run_action<T: Informant>(
 
 /// Execute given Transaction and verify resulting state root.
 pub fn run_transaction<T: Informant>(
+	// Chain specification name associated with the transaction
```
dvdplm: Should these be doc-comments? I'm unsure of how the docs are rendered for private fields of a public struct.

evmbin/src/main.rs
```diff
@@ -78,39 +78,42 @@ EVM implementation for Parity.
...
-    parity-evm state-test <file> [--json --std-json --std-dump-json --only NAME --chain CHAIN --std-out-only --std-err-only]
+    parity-evm state-test <file> [--chain CHAIN --only NAME --json --std-json --std-dump-json --std-out-only --std-err-only]
     parity-evm stats [options]
     parity-evm stats-jsontests-vm <file>
     parity-evm [options]
     parity-evm [-h | --help]
 
 Commands:
-    state-test         Run a state test from a json file.
+    state-test         Run a state test on a provided state test JSON file path.
```
dvdplm:
```suggestion
    state-test         Run a state test on a provided state test JSON file.
```

evmbin/src/main.rs
```diff
@@ -78,39 +78,42 @@ EVM implementation for Parity.
...
     --code CODE        Contract code as hex (without 0x).
     --to ADDRESS       Recipient address (without 0x).
     --from ADDRESS     Sender address (without 0x).
     --input DATA       Input data as hex (without 0x).
     --gas GAS          Supplied gas as hex (without 0x).
-    --gas-price WEI    Supplied gas price as hex (without 0x).
+    --gas-price GWEI   Supplied gas price as hex (without 0x).
 
 State test options:
+    --chain CHAIN      Run only from specific chain name (i.e. EIP150, EIP158,
```
dvdplm:
```suggestion
    --chain CHAIN      Run only from specific chain name (i.e. one of EIP150, EIP158,
```

evmbin/src/main.rs
```diff
@@ -78,39 +78,42 @@ EVM implementation for Parity.
...
+                       Frontier, Homestead, Byzantium, Constantinople,
+                       ConstantinopleFix, EIP158ToByzantiumAt5, FrontierToHomesteadAt5,
+                       HomesteadToDaoAt5, HomesteadToEIP150At5).
     --only NAME        Runs only a single test matching the name.
-    --chain CHAIN      Run only tests from specific chain.
 
 General options:
+    --chain PATH       Chain spec file path.
```
dvdplm:
```suggestion
    --chain PATH       Path to chain spec file.
```

evmbin/src/main.rs
```diff
@@ -139,119 +142,153 @@ fn main() {
...
+			// (i.e. list of transactions and associated state roots hashes corresponding each chain).
 			for (idx, state) in states.into_iter().enumerate() {
 				let post_root = state.hash.into();
 				let transaction = multitransaction.select(&state.indexes).into();
 
+				// Determine the type of trie with state root to create in the database.
+				// The database is a key-value datastore implemented as a database-backend
+				// modified Merkle tree.
+				// Use a secure trie database specification when CLI option `--std-dump-json`
+				// is specified, otherwise use secure trie with FAT trie database.
```
dvdplm:
```suggestion
				// is specified, otherwise use secure trie with fat trie database.
```

evmbin/src/main.rs
```diff
@@ -139,119 +142,153 @@ fn main() {
...
 				};
-				if args.flag_json {
-					info::run_transaction(&name, idx, &spec, &pre, post_root, &env_info, transaction, display::json::Informant::default(), trie_spec)
-				} else if args.flag_std_dump_json || args.flag_std_json {
+
+				// Execute the given transaction and verify resulting state root
+				// for CLI option `--std-dump-json` or `--std-json`.
+				if args.flag_std_dump_json || args.flag_std_json {
 					if args.flag_std_err_only {
+						// Use Standard JSON informant with err only
```
dvdplm: s/err/stderr/ ?

evmbin/src/main.rs
```diff
@@ -139,119 +142,153 @@ fn main() {
...
 						info::run_transaction(&name, idx, &spec, &pre, post_root, &env_info, transaction, display::std_json::Informant::err_only(), trie_spec)
 					} else if args.flag_std_out_only {
+						// Use Standard JSON informant with out only
```
dvdplm: s/out/stdout/

evmbin/src/info.rs
ltfschoen: Thanks. I'm looking into it. If I update main.rs so that info.rs is public: `pub mod info;` then it's shown in the rustdocs, but if I make this a doc comment (i.e. `/// Chain ...`) and then build the docs with `cargo doc -p evmbin --release --open --no-deps` i get the following error.

```
error: documentation comments cannot be applied to method arguments
   --> evmbin/./src/info.rs:105:2
    |
105 |     /// Chain specification name associated with the transaction
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doc comments are not allowed here
```

ltfschoen: @dvdplm I tried to make the input parameters into this `run_transaction` into a struct called `InputData`, since we can use doc comments on structs. The approach that I tried has been pushed in this commit https://github.com/paritytech/parity-ethereum/commit/8a6e6445e125ff1b0a60062da9327a55d3457667 (a separate branch to try it out), however I'm not sure how to overcome the `^^ undeclared lifetime
` error. In the `InputData` struct i'm not sure how to declare the lifetime parameter in addition to the generic type `T` (i.e. change `pub struct InputData<T> {` to `pub struct InputData<'a, T> {` or something?), and the existing code `mut informant: T,` is giving the following error and i'm not sure how to overcome that either:
```
122 |     mut informant: T,
    |     ^^^ expected identifier, found keyword
help: you can escape reserved keywords to use them as identifiers
    |
122 |     r#mut informant: T,
```

evmbin/src/main.rs
ltfschoen: i've pushed commit https://github.com/paritytech/parity-ethereum/pull/10742/commits/0662a6d56baa0398865c7e13824e04dfc86c1e04 to address this

evmbin/src/main.rs
ltfschoen: i've pushed commit https://github.com/paritytech/parity-ethereum/pull/10742/commits/0662a6d56baa0398865c7e13824e04dfc86c1e04 to address this

evmbin/src/main.rs
```diff
@@ -68,8 +68,8 @@ use bytes::Bytes;
 use ethcore::{spec, json_tests, TrieSpec};
 use vm::{ActionParams, CallType};
 
-mod info;
-mod display;
+pub mod info;
```
sorpaas: Is it actually needed for this to be `pub`?

evmbin/src/main.rs
```diff
@@ -78,39 +78,42 @@ EVM implementation for Parity.
...
+                       in tsv format.
 
 Transaction options:
     --code CODE        Contract code as hex (without 0x).
     --to ADDRESS       Recipient address (without 0x).
     --from ADDRESS     Sender address (without 0x).
     --input DATA       Input data as hex (without 0x).
     --gas GAS          Supplied gas as hex (without 0x).
-    --gas-price WEI    Supplied gas price as hex (without 0x).
+    --gas-price GWEI   Supplied gas price as hex (without 0x).
```
sorpaas: This may be a backward incompatible change. I have a feeling maybe keeping `WEI` is better because this is supposed to be really raw interface (how do you specify a test requiring 2.1 GWEI gas?).

evmbin/src/main.rs
```diff
@@ -405,15 +457,118 @@ mod tests {
...
 		assert_eq!(args.flag_std_dump_json, true);
-		assert_eq!(args.flag_chain, Some("homestead".to_owned()));
-		assert_eq!(args.flag_only, Some("add11".to_owned()));
+		assert_eq!(args.flag_std_out_only, true);
+		assert_eq!(args.flag_std_err_only, true);
 	}
+
+	#[test]
+	fn should_parse_state_test_command_from_state_test_json_file() {
+		let s = r#"{
```
sorpaas: nit: putting this long string into a file and use `include_str!` is probably better.

evmbin/src/main.rs
ltfschoen: I had to make it `pub` otherwise it doesn't show you any of the documentation contained in that `info` module when you build and open the the docs for the evmbin package with `cargo doc -p evmbin --release --open --no-deps`

evmbin/src/main.rs
ltfschoen: thanks i've pushed update in commit 5575db7140ce2101ef4a15197c000a7f20f4fe80

evmbin/src/info.rs
dvdplm: See https://github.com/paritytech/parity-ethereum/commit/8a6e6445e125ff1b0a60062da9327a55d3457667#r33959733 – but I admit I'm not sure of the context here and why we need lifetimes on these values (prolly me being ignorant).

evmbin/src/main.rs
dvdplm: Hmm. It's a shame if we have to expose a larger API surface just so we can show docs... If `info` is not public then we probably should not make it so just to document it. Maybe [this](https://stackoverflow.com/questions/29378113/how-can-i-include-private-modules-when-generating-documentation-via-cargo) is useful here?

ltfschoen: Thanks, yes that link worked great. I've pushed changes in this commit 41b2285b7 that reverts the changes so those modules aren't public again. Their docs may be viewed with `cargo doc -p evmbin --release --open --no-deps --document-private-items`

evmbin/src/main.rs
ltfschoen: Thanks. I've pushed commit cd742a6245f96eff4988cf7762b47f9ed76fdcb9 that uses `include_str!` and I've updated the tests

evmbin/src/main.rs
```diff
@@ -405,15 +476,127 @@ mod tests {
...
+				// Hash of latest transaction index in the chain
+				let post_root = H256::from_str("0xde1d3953b508913c6e3e9bd412cd50daf60bb177517e5d1e8ccb0dab193aed03").unwrap();
+				let trie_spec = TrieSpec::Secure; // TrieSpec::Fat for --std_dump_json
+				let transaction: transaction::SignedTransaction = multitransaction.select(&state.indexes).into();
+				info::run_transaction(&name, idx, &spec, &pre, post_root, &env_info, transaction, informant, trie_spec)
+			}
+		}
+	}
+
+	// TODO - add test that passes without failing with `State root mismatch`
```
sorpaas: Please create an issue for those TODOs and add the issue number here.

ltfschoen: I've pushed commit b9db81e634fc6a3ba5159102f5ca1df07341f64a to provide a single TODO with a link to a new issue https://github.com/paritytech/parity-ethereum/issues/10768

evmbin/src/info.rs
ltfschoen: Thanks for the tips! I somehow managed to get it to compile and I've created a separate PR https://github.com/paritytech/parity-ethereum/pull/10769. However there's quite a bit of code duplication and I haven't figured out how to approach resolving yet.

ordian:
@ltfschoen a merge with master would fix cargo audit failure (and resolve the merge conflict).

evmbin/src/main.rs
```diff
@@ -407,15 +535,128 @@ mod tests {
...
+	// Simulate using CLI command `state-test` and option `--json` (JSON informant)
+	// to execute a given transaction and verify its resulting state root
+	// using create2callPrecompiles.json
+	//
+	// ```
+	// cargo build -p evmbin;
+	// ./target/debug/parity-evm state-test \
+	//   ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople --json
+	// ```
+	fn should_verify_state_root_using_constantinople_state_test_json_file() {
```
debris: missing `#[test]` and assertions

evmbin/src/main.rs
```diff
@@ -407,15 +535,128 @@ mod tests {
...
+
+	// Simulate using CLI command `state-test` and option `--json` (JSON informant)
+	// to execute a given transaction and verify its resulting state root
+	// using teststate.json
+	//
+	// ```
+	// cargo build -p evmbin;
+	// ./target/debug/parity-evm state-test ./evmbin/res/teststate.json --only add11 --chain EIP150 --json
+	// ```
+	fn should_verify_state_root_using_sample_state_test_json_file() {
```
debris: missing `#[test]` and assertions

evmbin/src/main.rs
```diff
@@ -407,15 +535,128 @@ mod tests {
...
+	#[test]
+	fn should_parse_specific_state_tests_from_test_state_test_json_file() {
+		let state_tests = include_str!("../res/teststate.json");
+		let _deserialized_state_tests: SampleStateTests = serde_json::from_str(state_tests)
+			.expect("Serialization cannot fail; qed");
+		let state_test_add11: State = _deserialized_state_tests.add11;
+		let state_test_add12: State = _deserialized_state_tests.add12;
 	}
+
+	fn should_parse_specific_state_test_from_constantinople_state_test_json_file() {
```
debris: missing `#[test]` and assertions

debris:
@dvdplm not entirely, tests in `evmbin/src/main.rs` are still not finished

dvdplm:
> @dvdplm not entirely, tests in `evmbin/src/main.rs` are still not finished

Ouch, sorry about that. @ordian can you take a look at that please? :/

Issue #10378, opened 2019-02-19
`parity --config=dev` doesn't shutdown correctly


vrde:
- **Parity Ethereum version**: 2.3.2-beta-678138f-20190203
- **Operating system**: Linux
- **Installation**: one-line installer
- **Fully synchronized**: no
- **Network**: development network
- **Restarted**: yes

When I run `parity --config=dev` and try to kill it with `ctrl+c` it hangs forever, see the following log:

```
➜  ~ parity --config=dev
2019-02-19 01:06:08  Starting Parity-Ethereum/v2.3.2-beta-678138f-20190203/x86_64-linux-gnu/rustc1.31.1
2019-02-19 01:06:08  Keys path /home/vrde/.local/share/io.parity.ethereum/keys/DevelopmentChain
2019-02-19 01:06:08  DB path /home/vrde/.local/share/io.parity.ethereum/chains/DevelopmentChain/db/1484bce8c021f2ca
2019-02-19 01:06:08  State DB configuration: fast
2019-02-19 01:06:08  Operating mode: active
2019-02-19 01:06:08  Configured for DevelopmentChain using InstantSeal engine
2019-02-19 01:06:13  Public node URL: enode://60366b486fae829c6424e900bdeb0bfab17bca0dcbbb456a5f8715d36de823c741b85d62af448ef93e4f163d0682959df5c9380a95f8c9291520ed90171f54c0@192.168.0.3:30303
^C2019-02-19 01:06:36  Finishing work, please wait...
2019-02-19 01:07:36  Shutdown is taking longer than expected.
^C^\[1]    12958 quit (core dumped)  parity --config=dev
```

Edit: I've just noticed #10364. The issue looks the same. Mine might be easier to reproduce.

joshua-mir:
Can you try shutting down with `--logging shutdown`?

joshua-mir:
(I know that I was messing around with dev on beta 2.3.3 earlier today and didn't see this issue so there must be more to it than just running on dev)

vrde:
```
➜  ~ parity --config=dev --logging shutdown
2019-02-19 01:16:22  main INFO parity_ethereum::run  Starting Parity-Ethereum/v2.3.2-beta-678138f-20190203/x86_64-linux-gnu/rustc1.31.1
2019-02-19 01:16:22  main INFO parity_ethereum::run  Keys path /home/vrde/.local/share/io.parity.ethereum/keys/DevelopmentChain
[...]

Issue #10364, opened 2019-02-15
parity ethereum client doesn't always shutdown gracefully


c0deright:
- **Parity Ethereum version**:  2.2.10-stable
- **Operating system**: Linux
- **Installation**: binary
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

Sometimes stopping parity results in the issue described here: https://github.com/paritytech/parity-ethereum/issues/9101#issuecomment-454746413

Sometimes when stopping parity even with shutdown tracing turned on the process exits immediately and nothing about a shutdown is being logged at all.

So we have 3 outcomes when stopping parity:

* clean shutdown (nothing being logged at all, almost instant)
* clean shutdown (shutdown being logged, taking 1-10 seconds)
* unclean shutdown (_Shutdown is taking longer than expected / Shutdown timeout reached, exiting uncleanly_)

How to debug this further?

c0deright:
```
2019-02-15 12:25:03  Verifier #1 INFO import  Imported #7223288 0xe31c…3dcf (84 txs, 7.99 Mgas, 675 ms, 14.60 KiB)
2019-02-15 12:25:08  IO Worker #3 INFO import     5/ 5 peers     18 MiB chain  115 MiB db  0 bytes queue   43 KiB sync  RPC:  0 conn,    0 req/s, 3244 µs
2019-02-15 12:25:23  Verifier #1 INFO import  Imported #7223289 0x70d4…8380 (103 txs, 7.99 Mgas, 1599 ms, 16.89 KiB)
2019-02-15 12:25:38  IO Worker #2 INFO import     5/ 5 peers     18 MiB chain  115 MiB db  0 bytes queue   43 KiB sync  RPC:  0 conn,    0 req/s, 3244 µs
2019-02-15 12:25:57  main INFO parity_ethereum::run  Finishing work, please wait...
2019-02-15 12:25:57  main TRACE shutdown  [IoService] Closing...
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closing...
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closed
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closing...
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closed
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closing...
2019-02-15 12:25:57   TRACE shutdown  [IoWorker] Closed
[...]

Issue #9101, opened 2018-07-12
Parity exits uncleanly on archive node during upgrade (timeout and shutdown at ~5m; restarting normally takes 22s)


XertroV:
> I'm running:
>
> - **Which Parity version?**: 1.10.6 or 1.10.8 -> 10.10.9
> - **Which operating system?**: Linux
> - **How installed?**: via binaries
> - **Are you fully synchronized?**: yes
> - **Which network are you connected to?**: ethereum
> - **Did you try to restart the node?**: n/a

While debugging another issue (RPC threads dying) I noticed that during the auto-upgrade to 1.10.9 my node shut itself down uncleanly :/

Apparently the timeout for detecting a "longer than expected" shutdown is not very long at all, and unsuitable for archive nodes.

Logs: (note, right hand side is truncated, but don't think there's much important missing)

```
Jul 10 18:21:45 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:21:45 UTC   52/500 peers      1 GiB chain 1015 MiB db  0 bytes queue  694 KiB sync  RPC:  0 conn, 11 req
Jul 10 18:21:46 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:21:46 UTC Snapshot: 328686 accounts 0 blocks 316371381 bytes
Jul 10 18:21:58 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:21:58 UTC Snapshot: 328686 accounts 0 blocks 316371381 bytes
Jul 10 18:22:01 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:22:01 UTC Attempting to get parity binary af91…5fa5
Jul 10 18:22:01 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:22:01 UTC Imported #5940363 e6ca…a20f (91 txs, 7.89 Mgas, 322.01 ms, 19.43 KiB)
Jul 10 18:22:02 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:22:02 UTC Fetched latest version (1.10.9-stable-23a9…ff51) OK to /tmp/HCniD5cVjwoI
Jul 10 18:22:02 eth-aws-nv-node-03 parity[4897]: 2018-07-10 18:22:02 UTC Installed updated binary to /home/ubuntu/.local/share/io.parity.ethereum-updates/parity-1.10.9
[...]

PR #10744, opened 2019-06-13
Add a way to signal shutdown to snapshotting threads

dvdplm:
Use an atomic bool in `Progress` to signal snapshotting threads that it's time to shutdown

Possibly related: https://github.com/paritytech/parity-ethereum/issues/10364, https://github.com/paritytech/parity-ethereum/issues/10378, https://github.com/paritytech/parity-ethereum/issues/9101

### DIFF
--- a/ethcore/src/snapshot/mod.rs
+++ b/ethcore/src/snapshot/mod.rs
@@ -107,7 +107,7 @@ impl Default for SnapshotConfiguration {
 	fn default() -> Self {
 		SnapshotConfiguration {
 			no_periodic: false,
-			processing_threads: ::std::cmp::max(1, num_cpus::get() / 2),
+			processing_threads: ::std::cmp::max(1, num_cpus::get_physical() / 2),
 		}
 	}
 }
--- a/ethcore/src/snapshot/mod.rs
+++ b/ethcore/src/snapshot/mod.rs
@@ -199,10 +199,6 @@ pub fn take_snapshot<W: SnapshotWriter + Send>(
 		let mut state_hashes = Vec::new();
 
 		for guard in state_guards {
-			if p.abort.load(Ordering::SeqCst) {
-				trace!(target: "snapshot", "[snapshot::take_snapshot] aborting");
-				return Err(Error::AbortSnapshot);
-			}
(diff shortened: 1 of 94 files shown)


ethcore/src/snapshot/mod.rs
```diff
@@ -119,6 +119,7 @@ pub struct Progress {
 	blocks: AtomicUsize,
 	size: AtomicUsize, // Todo [rob] use Atomicu64 when it stabilizes.
```
ordian: not introduced by the PR, but `AtomicU64` is [stable](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html) now

ethcore/src/snapshot/mod.rs
```diff
@@ -119,6 +119,7 @@ pub struct Progress {
...
 	done: AtomicBool,
+	abort: AtomicBool, // TODO: why no Arcs here?
```
ordian: because `&AtomicBool` is thead-safe already and `Progress` outlives the caller?

ethcore/src/snapshot/service.rs
```diff
@@ -479,15 +479,12 @@ impl Service {
...
 		if let Err(e) = res {
 			if client.chain_info().best_block_number >= num + client.pruning_history() {
-				// "Cancelled" is mincing words a bit -- what really happened
-				// is that the state we were snapshotting got pruned out
-				// before we could finish.
-				info!("Periodic snapshot failed: block state pruned.\
-					Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
+				// The state we were snapshotting was pruned before we could finish.
+				info!("Periodic snapshot failed: block state pruned. Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
+				// TODO: this looks odd to me. I don't think it's dangerous but I also don't understand why we're not returning an error here.
```
ordian: yeah, looks odd to me too

ethcore/src/snapshot/mod.rs
dvdplm: Is it obvious to you why `AtomicU64` is better? Just avoid ending up with an unexpected u32?

PR #10752, opened 2019-06-14
Use fewer threads for snapshotting

dvdplm:
When taking a snapshot the current default number of threads is equal to half the number of **logical** CPUs in the system. On HT enabled CPUs this value seems a bit high, e.g. 6 snapshotting threads on a 6/12 core/hyperthread CPU. Maybe a better default value is half the number of physical cores?


ordian:
for the record: https://github.com/paritytech/parity-ethereum/pull/9239

PR #10755, opened 2019-06-17
updater: fix static id hashes initialization

ordian:
Might help with https://github.com/paritytech/parity-ethereum/issues/10085#issuecomment-502594021.

The bug was introduced in #10670. 
Previously we used to pad short bytes with zeros when converting them to H256:
https://github.com/paritytech/parity-common/blob/e16441a8421d2e39a2e91a40f30a6b01cbaee00e/fixed-hash/src/hash.rs#L105-L111
But in the latest version of fixed-hash there is a length check and a panic.

I'm not really sure if there are other places in our codebase where we rely on the hash string being resized automatically when converting to `H256`.

### DIFF
--- a/updater/src/updater.rs
+++ b/updater/src/updater.rs
@@ -182,7 +182,7 @@ lazy_static! {
 
 
 // Pads the bytes with zeros or truncates the last bytes to H256::len_bytes()
-// before the convertion to match the previous behavior.
+// before the conversion to match the previous behavior.
 fn h256_from_str_resizing(s: &str) -> H256 {
 	let mut bytes = s.as_bytes().to_vec();
 	bytes.resize(H256::len_bytes(), 0);
--- a/updater/src/updater.rs
+++ b/updater/src/updater.rs
@@ -159,7 +159,7 @@ pub struct Updater<O = OperationsContractClient, F = fetch::Client, T = StdTimeP
 const CLIENT_ID: &str = "parity";
 
 lazy_static! {
-	static ref CLIENT_ID_HASH: H256 = H256::from_slice(CLIENT_ID.as_bytes());
+	static ref CLIENT_ID_HASH: H256 = h256_from_str_resizing(CLIENT_ID);
 }
 
 lazy_static! {
@@ -177,7 +177,16 @@ lazy_static! {
 }
 
 lazy_static! {
-	static ref PLATFORM_ID_HASH: H256 = H256::from_slice(PLATFORM.as_bytes());
+	static ref PLATFORM_ID_HASH: H256 = h256_from_str_resizing(&PLATFORM);
+}
+
+
+// Pads the bytes with zeros or truncates the last bytes to H256::len_bytes()
+// before the convertion to match the previous behavior.
+fn h256_from_str_resizing(s: &str) -> H256 {
+	let mut bytes = s.as_bytes().to_vec();
+	bytes.resize(H256::len_bytes(), 0);
+	H256::from_slice(&bytes)
 }
 
 /// Client trait for getting latest release information from operations contract.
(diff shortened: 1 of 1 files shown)


updater/src/updater.rs
```diff
@@ -177,7 +177,16 @@ lazy_static! {
 }
 
 lazy_static! {
-	static ref PLATFORM_ID_HASH: H256 = H256::from_slice(PLATFORM.as_bytes());
+	static ref PLATFORM_ID_HASH: H256 = h256_from_str_resizing(&PLATFORM);
+}
+
+
+// Pads the bytes with zeros or truncates the last bytes to H256::len_bytes()
+// before the convertion to match the previous behavior.
```
dvdplm: Typo: “conversion”

ordian:
```suggestion
// before the conversion to match the previous behavior.
```

ordian:
> Can't H256::from_slice only accepts 32-lenth slices?

It currently does and that is the problem, since `CLIEND_ID` (`"parity"`) and `PLATFORM` are less then 32 bytes.

ngotchac:
Yeah, I meant as the argument type? So that we could catch this at compile time.

ordian:
> Can't H256::from_slice only accepts 32-lenth slices?

Ah, you mean `&[u8; 32]`? That would probably make it harder to use this API when we generate a `Vec` for a hex string or use a subslice of larger type like `data[..32]`.

ngotchac:
Yeah sure, but we could also have a `from_unknown_slice` method or so for such cases that would pad it as it used to do

ordian:
> Yeah sure, but we could also have a from_unknown_slice method or so for such cases that would pad it as it used to do

Fair enough, but you can create it from `&[u8; 32]` slice already by `H256(*slice)` :P
Maybe this is what we should use instead though.

ngotchac:
Yeah sure.. But what I meant is that it would be better to not have a possible `panic` as we do now :p  Maybe renaming `from_slice` to `from_unsafe_slice` then :p

ordian:
I get what you're trying to say, but it would also require changing [conversions to uint](https://github.com/paritytech/parity-common/blob/9d2de86e70f8e84358d0819a92666acb4f6ab9c4/uint/src/uint.rs#L1092) for consistency, which would in turn imply a bunch of breaking changes and new releases, which I'm not sure is justified atm :p

dvdplm:
The change to panic with the wrong size is pretty new and before it was resizing the slice to the proper size, so old code might expect that behaviour. I agree with the change, but runtime panics like this are bad. Presumably any new code written will have tests in place to catch the panics but that's far from satisfying. The name `from_slice` is not helping here, it looks pretty innocent and "safe" (and I'm the first to admit I rarely read the rust docs for our own code where this is indeed [mentioned](https://docs.rs/primitive-types/0.4.0/primitive_types/struct.H256.html#method.from_slice)).

We can:

1. rename it to something "dangerous", like `from_slice_but_watchout_it_panics()`
2. make it safe on the type level, like @ngotchac suggests
3. revert `from_slice` to behave like before, silently taking care of padding/truncating

Personally I'm tending to 3) tbh, but I'd like to hear more voices.

ngotchac:
But why did we change the behavior in the first place?

dvdplm:
> But why did we change the behavior in the first place?

https://github.com/paritytech/parity-common/commit/ef4a8d8dfe31795144284b42395c4a512b57f133

ngotchac:
Yeah, I would vote for having `from_fixed_bytes` and `from_bytes_padding`/`from_padded_bytes` or something similar, so we'll have to update each references and be sure that it's correctly used.

PR #10756, opened 2019-06-17
Enable aesni

dvdplm:
Build parity-ethereum with SSE2, SSSE3 and AES enabled for `x86_64` targets so that the `aesni` crate is picked up at compile time.

Builds on https://github.com/paritytech/parity-ethereum/pull/10714


ordian:
wait, if we specify both `RUSTFLAGS` and `rustflags` in `.cargo/config`, doesn't the latter get overwritten by the former?

dvdplm:
This does not work, something's wrong here.

When I build with `RUSTFLAGS` I can verify the build with `objdump -d path/to/artifact | rg aesimc` and I see e.g. `1002108d0:	66 0f 38 db ca 	aesimc	%xmm2, %xmm1` but with `rustflags` set to the **exact** same string in `./cargo/config` I get this error instead:

```
   Compiling aes-ctr v0.3.0
   Compiling aes v0.3.2
error[E0463]: can't find crate for `aesni`
  --> /Users/dvd/.cargo/registry/src/github.com-1ecc6299db9ec823/aes-0.3.2/src/lib.rs:63:1
   |
63 | extern crate aesni;
   | ^^^^^^^^^^^^^^^^^^^ can't find crate
error[E0463]: can't find crate for `aesni`

  --> /Users/dvd/.cargo/registry/src/github.com-1ecc6299db9ec823/aes-ctr-0.3.0/src/lib.rs:62:1
   |
62 | extern crate aesni;
   | ^^^^^^^^^^^^^^^^^^^ can't find crate

```

Something's definitely weird. I can reproduce it with plain `rustc` as well:

```
$ rustc --crate-name aes /Users/dvd/.cargo/registry/src/github.com-1ecc6299db9ec823/aes-0.3.2/src/lib.rs --color always --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C metadata=13d9028e470a76bd -C extra-filename=-13d9028e470a76bd --out-dir /Users/dvd/dev/parity/parity-ethereum/target/release/deps -L dependency=/Users/dvd/dev/parity/parity-ethereum/target/release/deps --extern block_cipher_trait=/Users/dvd/dev/parity/parity-ethereum/target/release/deps/libblock_cipher_trait-cde3daa108f21a6f.rlib --cap-lints allow -C target-feature=+sse2,+aes,+ssse3
error[E0463]: can't find crate for `aesni`
  --> /Users/dvd/.cargo/registry/src/github.com-1ecc6299db9ec823/aes-0.3.2/src/lib.rs:63:1
   |
63 | extern crate aesni;
   | ^^^^^^^^^^^^^^^^^^^ can't find crate

**Comment by dvdplm** [comment]
e

> if we specify both `RUSTFLAGS` and `rustflags` in `.cargo/config`, doesn't the latter get overwritten by the former?

Yes, `RUSTFLAGS` should take the priority for the values that overlap.

**Comment by dvdplm** [comment]
Oh, this must be https://github.com/rust-lang/cargo/issues/6858

**Comment by ordian** [comment]
> Yes, RUSTFLAGS should take the priority for the values that overlap.

I think it overrides all flags, not just the ones that overlap.

**Comment by dvdplm** [comment]
> I think it overrides all flags, not just the ones that overlap.

You are correct, using `RUSTFLAGS` in any form clobbers the `rustflags` setting from `.cargo/config`. This is a pretty serious foot gun: it's quite possible that some future PR introduce `RUSTFLAGS` usage somewhere and without knowing it we're not using `aesimc` anymore.

**Comment by ordian** [comment]
#10714 is merged, could you rebase on/merge with master and resolve conflicts please?

**Comment by ordian** [comment]
> You are correct, using RUSTFLAGS in any form clobbers the rustflags setting from .cargo/config. 

We still set them [in build-linux.sh](https://github.com/paritytech/parity-ethereum/blob/2d693be73534674efc96cd6f2f6a4dfc65de6fa9/scripts/gitlab/build-linux.sh#L13) and [in test-linux.sh](https://github.com/paritytech/parity-ethereum/blob/2d693be73534674efc96cd6f2f6a4dfc65de6fa9/scripts/gitlab/test-linux.sh#L18).

**Comment by dvdplm** [comment]
> We still set them [in build-linux.sh](https://github.com/paritytech/parity-ethereum/blob/2d693be73534674efc96cd6f2f6a4dfc65de6fa9/scripts/gitlab/build-linux.sh#L13) and [in test-linux.sh](https://github.com/paritytech/parity-ethereum/blob/2d693be73534674efc96cd6f2f6a4dfc65de6fa9/scripts/gitlab/test-linux.sh#L18).

Yeah I know. So what is the best way to go here?

1. we can set the flags in both `.cargo/config` and in the release/test scripts – easy to get wrong
1. we can scrap this PR and not use `aesimc` until something changes in cargo or rustc that enables us to avoid the clobbering
1. only use `RUSTFLAGS` and not enable `aesimc` for devs, but only for release/test builds (i.e. set it in `build-*.sh` and `test-*.sh` and not use `.cargo/config` at all)

Do we have any other options?

**Comment by cheme** [comment]
The third option seems to be the more realistic to me (but I can be a bit pessimistic :)

**Comment by dvdplm** [comment]
> The third option seems to be the more realistic to me (but I can be a bit pessimistic :)

Hmm. One big downside is that devs will not run the same code as users, which can be a terrible timesink when fixing bugs.

**Comment by ordian** [comment]
I would vote for the third option as well.

> One big downside is that devs will not run the same code as users, which can be a terrible timesink when fixing bugs.

If we make clear for everyone which flags we use for release builds, devs can set the flags themselves, or use `build-linux.sh` directly.

**Comment by dvdplm** [comment]
Hmm. I think I'd prefer 1) then: we make a best effort to keep the two in sync and if `.cargo/config` goes out of sync, then it's not the end of the world because releases are built with `RUSTFLAGS`.

File: scripts/gitlab/build-linux.sh [hunk 1]
```diff
@@ -11,7 +11,7 @@ echo "CC:               " $CC
 echo "CXX:              " $CXX
 #strip ON
 export RUSTFLAGS=" -C link-arg=-s"
-# Linker for crosscomile
+# Linker for crosscompilation
```
dvdplm: I'm not sure what the lines below actually do. Can this be removed? cc @TriplEight

scripts/gitlab/build-windows.sh
```diff
@@ -14,6 +14,9 @@ echo "RUSTC_WRAPPER:    " $RUSTC_WRAPPER
 echo "SCCACHE_DIR:      " $SCCACHE_DIR
 
 echo "_____ Building target: "$CARGO_TARGET" _____"
+# NOTE: if you change this please remember to also update .cargo/config
+export RUSTFLAGS=" -Ctarget-feature=+aes,+sse2,+ssse3 -Ctarget-feature=+crt-static"
```
dvdplm: Before this PR we had `+crt-static` only for the `x86_64-pc-windows-msvc` target, but I'm not sure we care about 32-bit windows?

dvdplm:
@TriplEight is there an easy way to trigger a release build from here so I can double check the resulting binaries to ensure that AES in hardware is activated?

scripts/gitlab/build-linux.sh
TriplEight: you mean lines 15 and 16? 
```
echo "_____ Linker _____"
cat .cargo/config
```
It's just the info about the linker.

scripts/gitlab/build-windows.sh
TriplEight: No, we do not care about 32-bit windows.

scripts/gitlab/build-linux.sh
dvdplm: Right, but I think that with this PR it is not so anymore so probably best to remove it

.cargo/config
```diff
@@ -1,3 +1,26 @@
...
+#	scripts/build-linux.sh
+#	scripts/build-windows.sh
+
+# Using 'cfg` is broken, see https://github.com/rust-lang/cargo/issues/6858
+#[target.'cfg(target_arch = "x86_64")']
+#rustflags = ["-Ctarget-feature=+aes,+sse2,+ssse3"]
+
+# …so instead we list all target triples (Tier 1 64-bit platforms)
+[target.x86_64-unknown-linux-gnu]
+# Enable aesimc.
```
cheme: Small nit, easimc is one of the instruction resulting from eas-ni activation, so I would prefer a comment along the line of "Enable aes-ni instruction for RustCrypto dependency"

PR #10758, opened 2019-06-18
Initialize private tx logger only if private tx functionality is enabled

### LINKED ISSUE #10726: Improve startup message "Cannot read logs…"
At startup we print this:

```
2019-06-08 18:07:27  Cannot read logs: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

Print something nicer and more useful.

grbIzl:
Private tx logger tried to open log file even for the case, when private tx functionality was disabled.

Closes #10726

### DIFF
--- a/parity/configuration.rs
+++ b/parity/configuration.rs
@@ -1461,7 +1461,7 @@ mod tests {
 			private_provider_conf: ProviderConfig {
 				validator_accounts: Default::default(),
 				signer_account: Default::default(),
-				logs_path: Some(Directories::default().base),
+				logs_path: None,
 			},
 			private_encryptor_conf: Default::default(),
 			private_tx_enabled: false,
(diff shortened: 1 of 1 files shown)


ordian:
test failure seems legit

grbIzl:
> Should it be backported or was it introduced recently?

It was introduced in https://github.com/paritytech/parity-ethereum/pull/10056 As far as I can seee, it was not released yet

parity/configuration.rs
```diff
@@ -1458,7 +1461,7 @@ mod tests {
 			private_provider_conf: ProviderConfig {
 				validator_accounts: Default::default(),
 				signer_account: Default::default(),
-				logs_path: Some(Directories::default().base),
+				logs_path: None,
```
dvdplm: Seems like `ProviderConfig` should impl `Default` too?

PR #10760, opened 2019-06-19
[WIP] Historical Eras: keep full state for certain blocks and prune the rest - alternative to full archive node in some circumstances

XertroV:
This patch adds a new configuration item: pruning_historical_eras. It is a List<(u64,u64)> where List is some list-like type; both Vec and ElasticArray1024 are used atm. Any block heights within the specified ranges (inclusive) will not have their state entries removed. 

Testing whether blocks should be pruned or done is done in `mark_canonical`.

`read_overlay` will also, after the main overlay is loaded, go back and scan through historical eras to attempt to load from backing db.

`state_at` now checks if a block is within pruning history or whether it's within the ranges in historical_eras

`journaldb::new` now takes an additional parameter for historical eras (only used in overlayrecentdb)

Note: I've only tested this with account balances so far, not contract state, but seems to work. The testing has just been manual (sync node to mainnet and query account balance at certain heights)

Full commit history: https://github.com/XertroV/parity/tree/pruning_historical_eras

### Broad Overview:

- implemented only for OverlayRecentDB currently
- When marking canonical blocks we test whether the height exists in specified ranges
- if a height is within these ranges we make these changes to the mark_canonical process:
  - do not delete the db_key from the backing database
  - do not apply canon_deletions (remains an empty Vec)
  - do apply overlay_deletions
- we also add an additional check when calling Client::state_at to query the DB when a call specifies an ancient block

### Todo (improvements for this PR):

- revert formatting changes to code made automatically by jetbrains rust plugin :/
- delete/revert logging changes and `println!()`'s 
- ensure `read_overlay` doesn't scan the same era twice; either if it's in a range or a range overlaps with state already loaded into journal (recent era)
- I think atm the historical eras that are specified - while loaded from backing db on node init - are loaded into RAM and kept in RAM when node is syncing; would be good to keep on disk if possible to avoid overloading RAM
- fix unclean shutdown after RPC calls are made (unsure if this affects WS calls)

### Possible Future Improvements:

- Add an admin RPC call to register 
- Currently this works by block-height; while selective history would be nice (e.g. account balances + certain smart contract state) this seems like it'd be very complex with current architecture.
- Recalculate lost state when new ranges are added (e.g. I have a full node syncd to ~7900000 but add 4400000-4500000 as historical eras)
  - Q: can we do this via snapshots that are already taken? are snapshots of state taken automatically atm for full nodes (non-archive)?
- additional config item for loading historical states into RAM or keep on disk

### Risks:

- I don't understand the DB code well, would appreciate any explanation / docs / feedback on that to avoid bugs or bad implementation
- Potential to have way too much state in RAM - should use backing DB most of the time if possible
- others?

### Notes and Questions:

- Where should I use ElasticArray? I've seen it used occasionally and started using it while trying to solve another bug; can probs be replaced with Vec (which I was using originally).
- Is there any good docs on how the DB works? (particularly an explanation of insertions, deletions, journal entries, canonical entries, relationship to backing db)
- I am currently unsure if the patch - as it stands - also stores historical state for blocks/states that would otherwise be pruned (e.g. for a block that is an uncle now but was a head once; i.e. full state calculated and stored in journal). Any feedback here would be appreciated
- Is the design sensible?
- Any big gotchas I have missed or introduced?
- **Currently the node does not shutdown cleaning if you make an RPC request** - but does when no RPC request is made. I've had a look through various --logging=trace output (but more filtered). I can't see how/why this was introduced, but is obviously an issue.
- Is there a better way to do the config entry? It feels like a list of tuples is the smallest general structure, but there aren't any other config items like that (requiring a new parsing function) and makes it more ambiguous because there's no other items to compare the value to.

### PR Todo items:

- [ ] You mentioned a related issue if this PR related to it, e.g. `Fixes #228` or `Related #1337`.
- [ ] Your PR adheres [the style guide](https://wiki.parity.io/Coding-guide)
  - In particular, mind the maximal line length.
  - There is no commented code checked in unless necessary.
  - Any panickers have a proof or removed.
- [ ] You updated any rustdocs which may have changed


XertroV:
It occurs to me that a mix of this PR and the ArchiveDB (that's simple enough to implement) might be able to do a quick and dirty "full archive node from block X", sort of like `mark_canonical` just returning `Ok(())` after a certain height. Would still need to prune the overlay/journal tho.

dvdplm:
>     * Where should I use ElasticArray?

The reasons for using `ElasticArray` in Parity have mostly been superseded as `Vec` performance has improved over time and as far as I know there is now no good reason to use it (and there's`smallvec` for the cases where `Vec` isn't good enough).

dvdplm:
This PR looks like a very good start and while I'm still working on gathering opinions internally I wanted to get back to you and let you know that so far nobody has come up with strong objections. And thank you for your effort here!

XertroV:
@dvdplm - awesome! My main priority before moving beyond WIP is figuring out how to ensure historical eras are stored only on disk (rather than RAM). I see pruning non-canonical eras on disk as a nice to have rather than crucial feature since disk is (relatively) cheap.

And thanks for info re ElasticArray; should have removed all those usages now.

dvdplm:
cc @tomusdrw

dvdplm:
@XertroV how are things going for you? Any progress? :)

jeansouza:
Any new on this? This feature will be really useful! :)

XertroV:
@dvdplm @jeanspuza sadly no progress updates. At the time I was hoping/waiting for some feedback re questions in the PR about DB stuff since I didn't really get it and didn't want to stumble through (particularly because DB stuff is probably pretty important to get right).

jeansouza:
> @dvdplm @jeanspuza sadly no progress updates. At the time I was hoping/waiting for some feedback re questions in the PR about DB stuff since I didn't really get it and didn't want to stumble through (particularly because DB stuff is probably pretty important to get right).

Sad indeed! This would be a very useful feature.

XertroV:
I agree :P

dvdplm:
k, so first of all apologies for the lack of feedback here. There are currently no good documentation for the innards of how the db and state storage works so providing that kind of feedback is a pretty major undertaking. The best option is, alas, to read the code.

I'll put this PR "on ice" for the moment; if/when you are ready to pick up the work again let us know (and if not, it might serve as a starting point for someone else in the future).

XertroV:
@dvdplm sounds good and thanks for the explanation. 

I'm hoping I'll have time over the next few months to revisit this, but depends on a few things. 

If I do come back I'll try to document as I go. It's a little unintuitive tho so will probably ask for light reviews to make sure I'm not going off in the wrong direction

securitygeneration:
Came here looking for this exact functionality. Looking forward to it, any ETA? Let me know if you need help testing :)

tjayrush:
As an interested observer, may I ask this question: "Why would someone want this?" Would someone want to specify exactly where their own smart contracts interacted with the chain and thereby preserve state (for querying) at just those blocks? 

Also -- would the chain in this case act as an archive node only during those block ranges or would it act as an archive node ongoing from when it was configured?

dvdplm:
> "Why would someone want this?" Would someone want to specify exactly where their own smart contracts interacted with the chain and thereby preserve state (for querying) at just those blocks?

I think there are several use cases. Not sure about why you'd want to "snapshot" history at a specific point for your smart contract – the real state of the accounts involved is the most recent one so knowing what it looked like way back when is perhaps of limited use?
One use case could be to have a "poor man's sharding" of archive nodes: running an archive node is a major undertaking and requires many terrabytes of (fast) storage; maybe it's easier and cheaper to divvy up the work among nodes such that node A stores 0 –> 2M blocks, node B stores 2M+1 –> 4M etc

But yeah, it's a great question and I'm also interested in hearing other use cases!

jeansouza:
> As an interested observer, may I ask this question: "Why would someone want this?" Would someone want to specify exactly where their own smart contracts interacted with the chain and thereby preserve state (for querying) at just those blocks?
> 
> Also -- would the chain in this case act as an archive node only during those block ranges or would it act as an archive node ongoing from when it was configured?

I use the archive node to read internal transfers of new transactions, so I don't need all the past data, because it's slow and expensive to sync everything. That's the reason that this feature would help me a lot.

guifel:
Sad that this PR was frozen. It seems such an obvious feature. Many projects need only recent state data starting from block X when the project started or when the smart contract was deployed. 
Currently you need an expensive 6TB SSD and carry state data back from 2015 even if you're only interested in data from 2020 onward.

PR #10765, opened 2019-06-20
Updated Bn128PairingImpl to use optimized batch pairing 

seunlanlege:
original PR #10624

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,11 +264,13 @@ dependencies = [
 [[package]]
 name = "bn"
 version = "0.4.4"
-source = "git+https://github.com/paritytech/bn#2a71dbde5ca93451c8da2135767896a64483759e"
+source = "git+https://github.com/paritytech/bn#6beba2ed6c9351622f9e948ccee4063846b2b39a"
 dependencies = [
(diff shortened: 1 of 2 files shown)


andresilva:
What's different from the original PR? Also I think we have benchmarks for this, can we run them before and after the change to check the improvement?

seunlanlege:
@andresilva this PR has been rebased with the master branch, 

benchmarks for this branch 
https://gist.github.com/zac-williamson/71b67b994c8b7be4de263fe31dc2eb21

andresilva:
The diff between this PR and #10624 is the same. Also do I understand correctly that all bn128 benchmarks actually got slower with this change?

seunlanlege:
> This provides an approximate 2x increase in the speed of the algorithm, when multiple pairings are being computed at once.

from the original author.

dvdplm:
@andre yes, it’s the same PR but as the original author was busy we’re taking over. Mainly to double check the perf improvement claim.

ordian:
I've got the following results: https://gist.github.com/ordian/eeb56b65c44580aa59f60013bea2322b

andresilva:
@ordian Thanks for re-running the benchmarks. :+1:

PR #10767, opened 2019-06-20
Move Engine::register_client to be before other I/O handler registration

sorpaas:
This moves `Engine::register_client` to be right after client instantiation. Previously we register some I/O handlers first. Those I/O handlers may be triggered before the service finished `Engine::register_client`, thus results in `RequiresClient` error.

Might help #10085, but I'm not sure, as I have never reproduced it.


dvdplm:
@ordian did you mean to approve this PR?

Issue #10770, opened 2019-06-21
Istanbul Hard Fork Tracking


sorpaas:
Tracking issue for Istanbul EIPs implementation progress.

Accepted EIPs:

- [x] [EIP-152](https://github.com/ethereum/EIPs/pull/2129): Add Blake2 compression function F precompile (#10994)
- [x] [EIP-1108](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1108.md): Reduce alt_bn128 precompile gas costs (#10998 )
- [x] [EIP-1344](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1344.md): Add ChainID opcode (https://github.com/paritytech/parity-ethereum/issues/10982)
- [x] [EIP-1884](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1884.md): Repricing for trie-size-dependent opcodes (https://github.com/paritytech/parity-ethereum/issues/10990)
- [x] [EIP-2028](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2028.md): Calldata gas cost reduction (https://github.com/paritytech/parity-ethereum/issues/10986)
- [x] [EIP-2200](https://github.com/ethereum/EIPs/pull/2200): Rebalance net-metered SSTORE gas cost with consideration of SLOAD gas cost change (#10191)

Hard fork meta [EIP 1679](http://eips.ethereum.org/EIPS/eip-1679).

holiman:
@sorpaas is it too early for you guys to know how the genesis/config will look like? Will there be an istanbulBlock, or a separate thing for each eip ?

joshua-mir:
@holiman almost certainly a different item for each eip

soc1c:
https://github.com/goerli/testnet/issues/60#issuecomment-532200530


- Görli block `1561651` (Oct 30, final decision)
- Ropsten block `6485846` (Oct 2, not final yet)

PR #10771, opened 2019-06-21
EIP-1702: Generalized Account Versioning Scheme

sorpaas:
rel #10770 

This is a draft implementation of EIP-1702.

* Snapshots and light clients should be backward compatible. Older clients stops working once the network has its first non-legacy version account.
* The actual versioning is defined by `vm::Schedule::versions`, where we supports `PWasm` (which uses the pWasm execution engine) only (because it's the only useful target right now).
* For future `EVM` real usage, we'll define feature flags in `VersionedSchedule` (for example, `VersionedSchedule::EVM::eip1283`) that states a particular feature should be enabled in the current version.
* For frontend config, one can specify `wasm_version` together with `wasm_activation_transition`, so that pWasm is activated using an account version, instead of code prefix.


dvdplm:
>     * `pWasm` currently always create new contracts of version zero. This is a known issue.

Can you elaborate on this for my education?

sorpaas:
@dvdplm This is fixed in https://github.com/paritytech/parity-ethereum/pull/10771/commits/4e448f148cd4edea558107e4457d67733fef0e22. Let me know if you still need more info.

And for spec-related questions, maybe consider putting it under https://github.com/sorpaas/EIPs/issues/2!

ethcore/src/snapshot/account.rs
```diff
@@ -84,10 +85,14 @@ pub fn to_fat_rlps(
 	let mut leftover: Option<Vec<u8>> = None;
 	loop {
 		account_stream.append(account_hash);
-		account_stream.begin_list(5);
```
grbIzl: Is it possible to make one general begin_unbounded_list for all added rlp parameters? It looks weird, when it's first limited 5-6 elements list and unlimited after it.

ethcore/src/snapshot/account.rs
```diff
@@ -107,6 +112,10 @@ pub fn to_fat_rlps(
 			}
 		}
 
+		if !use_short_version {
+			account_stream.append(&acc.code_version);
+		}
+
 		account_stream.begin_unbounded_list();
```
grbIzl: Unbounded list starts here

ethcore/src/snapshot/account.rs
```diff
@@ -170,6 +179,12 @@ pub fn from_fat_rlp(
 		return Ok((ACC_EMPTY, None));
 	}
 
+	let use_short_version = match rlp.item_count()? {
+		5 => true,
```
grbIzl: I think, that it's a risky logic, to rely on item count in rlp packet. May be to pass this parameter explicitly?

ethcore/types/src/basic_account.rs
```diff
@@ -29,4 +29,48 @@ pub struct BasicAccount {
...
 	pub code_hash: H256,
+	/// Code version of the account.
+	pub code_version: U256,
+}
+
+impl rlp::Encodable for BasicAccount {
+	fn rlp_append(&self, stream: &mut rlp::RlpStream) {
+		let use_short_version = self.code_version == U256::zero();
+
+		match use_short_version {
```
grbIzl: Please also use begin_unbounded_list here.

ethcore/types/src/basic_account.rs
```diff
@@ -29,4 +29,48 @@ pub struct BasicAccount {
...
+
+		if !use_short_version {
+			stream.append(&self.code_version);
+		}
+	}
+}
+
+impl rlp::Decodable for BasicAccount {
+	fn decode(rlp: &rlp::Rlp) -> Result<Self, rlp::DecoderError> {
+		let use_short_version = match rlp.item_count()? {
```
grbIzl: The same comment about fragile logic based on item count

ethcore/src/snapshot/account.rs
sorpaas: I'm trying to maintain backward compatibility here. If we change the snapshotting format when `use_short_version` is `true`, then old clients will break. Or do you have better ideas?

ethcore/src/snapshot/account.rs
sorpaas: I don't get what you mean here. Which parameter do you think we can pass explicitly?

ethcore/types/src/basic_account.rs
sorpaas: Using `rlp.item_count()` is the only method to know whether we're at version `0` or above version `0`. I don't agree it's fragile -- `BasicAccount` is guaranteed to be either have 4 items (version `0` case) or 5 items (version non-`0` case). And we handle all errors explicitly.

ethcore/src/snapshot/account.rs
grbIzl: The only idea, that I've got, that doesn't break compatibility and doesn't increment the whole packet's length, is to pass the list instead of one of the parameters (for example, push (nonce, code_version) instead of nonce). And to define during decoding, is it list on this place in the packet or not. It will allow to get rid off the most of boolean flags in these two methods but it will be still "hacky" (not obvious). I mean something like this:
Encoding:
....
account_stream.begin_list(5);
account_stream.begin_list(2);
account_stream.append(&acc.nonce);
account_stream.append(&acc.code_version);
account_stream.append(&acc.balance);
...
Decoding:
let account_descr = rlp.val_at(0)?;
let mut nonce = 0;
let mut code_version = U256::zero();
if account_descr.is_list() {
	nonce = account_descr.val_at(0)?;
	code_version = account_descr.val_at(1)?;
} else {
	nonce = account_descr.as_val()?;
}
What do you think?

sorpaas: I still don't see what's the advantage of this over just checking the RLP list length. That should be reliable and we're already using it in places like transaction decoding: https://github.com/paritytech/parity-ethereum/blob/825c7990f29514d699e22c6e78f73710e53e983d/ethcore/types/src/transaction/transaction.rs#L121-L134

ethcore/src/snapshot/account.rs
```diff
@@ -84,10 +85,14 @@ pub fn to_fat_rlps(
 	let mut leftover: Option<Vec<u8>> = None;
 	loop {
 		account_stream.append(account_hash);
-		account_stream.begin_list(5);
+		let use_short_version = acc.code_version == U256::zero();
```
ordian:
```suggestion
		let use_short_version = acc.code_version.is_zero();
```

ethcore/src/factory.rs
```diff
@@ -31,11 +32,25 @@ pub struct VmFactory {
 }
 
 impl VmFactory {
-	pub fn create(&self, params: ActionParams, schedule: &Schedule, depth: usize) -> Box<dyn Exec> {
-		if schedule.wasm.is_some() && params.code.as_ref().map_or(false, |code| code.len() > 4 && &code[0..4] == WASM_MAGIC_NUMBER) {
-			Box::new(WasmInterpreter::new(params))
+	pub fn create(&self, params: ActionParams, schedule: &Schedule, depth: usize) -> Option<Box<dyn Exec>> {
+		if params.code_version == U256::zero() {
```
ordian:
```suggestion
		if params.code_version.is_zero() {
```

ethcore/src/executive.rs
```diff
@@ -451,12 +451,15 @@ impl<'a> CallCreateExecutive<'a> {
...
-					}
+				let out = match exec {
+					Some(exec) => {
+						let mut ext = Self::as_externalities(state, self.info, self.machine, self.schedule, self.depth, self.stack_depth, self.static_flag, &origin_info, &mut unconfirmed_substate, OutputPolicy::Return, tracer, vm_tracer);
+						match exec.exec(&mut ext) {
+							Ok(val) => Ok(val.finalize(ext)),
+							Err(err) => Err(err),
+						}
+					},
+					None => Ok(Err(vm::Error::OutOfGas)),
```
ordian: I'm not sure this is the correct error, is it?

sorpaas: This is the correct error -- `None` means the validation phrase fails. In that case, according to the spec we should return out of gas.

ethcore/src/snapshot/account.rs
```diff
@@ -246,6 +268,7 @@ mod tests {
 			balance: 123456789.into(),
 			storage_root: KECCAK_NULL_RLP,
 			code_hash: KECCAK_EMPTY,
+			code_version: 0.into(),
```
grbIzl: Can you please add one more test for encoding accounts with not zero code-version?

sorpaas:
I have added a frontend config `wasmVersion` (used together with `wasmActivationTransition` it enables pwasm via account versioning, rather than code prefix). I think the PR should be ready for review.

For testing, it's still not yet done and I think we may want to have a separate PR to do it comprehensively. (And collaborate with aleth, if possible.)

PR #9753, opened 2018-10-15
sync: retry different peer after empty subchain heads response

ascjones:
Further to #9531, I've experienced another intermittent cause of the ancient block sync retracting:

- Ancient block queue fills up, reset downloads (`blocks` is empty)
- Start a new sync round; request subchain heads
- Receive empty/useless response; we end up [here](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L324)
- When syncing old blocks this will return `Ok`, which will end up [giving the request to the same peer again](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/handler.rs#L104)
- Since `blocks` is empty and we didn't download any blocks this round, we retract a block for the next round (and later exponentially)
- That dodgy peer returns an empty/useless response again, and this process repeats which quickly causes the sync to retract to 0.

Fix is to return `Err(BlockDownloaderImportError::Useless)` in this case, which will result in that peer [being deactivated](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/handler.rs#L100) for this round.

*Edit*: this appears to have been causing the behaviour detailed here: https://github.com/paritytech/parity-ethereum/issues/9306#issuecomment-414756251

Have tested this with full ancient block sync successfully twice.

Issue #10763, opened 2019-06-19
Archive node is not syncing for >24 hours (stuck at block #993996)


ArseniiPetrovich:
- **Parity Ethereum version**: 2.4.5, 2.4.7
- **Operating system**: Linux
- **Installation**: pre-built binary
- **Fully synchronized**: no
- **Network**: Ethereum Mainnet
- **Restarted**: yes

**Parity config**:
```
[parity]
base_path = "parity_data"

[network]
reserved_peers="bootnodes.txt"
allow_ips = "all"
warp = false
max_peers = 30

[rpc]
apis = ["web3", "eth", "net", "parity", "traces"]
processing_threads = 50
server_threads = 200
port = 8545
interface = "0.0.0.0"

[websockets]
port = 8546
interface = "0.0.0.0"
max_connections = 1000
apis = ["web3","eth","net","parity", "pubsub", "traces"]
origins = ["all"]
hosts = ["all"]

[ui]
disable = true

[snapshots]
disable_periodic = true

[footprint]
tracing = "on"
pruning = "archive"
cache_size = 65536

[misc]
log_file = "/data/logs/parity.log"
```

**Expected behavior:** Parity slowly sync all the blockchain data
**Actual behavior:** Parity stuck on block #993996

The logs are:
```
Jun 19 12:40:29 ip-10-0-0-99.poa.internal parity[17243]: 2019-06-19 12:40:29 UTC Syncing  #993996 0x61d6…de17     0.00 blk/s    0.0 t…   13 µs
Jun 19 12:40:34 ip-10-0-0-99.poa.internal parity[17243]: 2019-06-19 12:40:34 UTC Syncing  #993996 0x61d6…de17     0.00 blk/s    0.0 t…   13 µs
Jun 19 12:40:39 ip-10-0-0-99.poa.internal parity[17243]: 2019-06-19 12:40:39 UTC Syncing  #993996 0x61d6…de17     0.00 blk/s    0.0 t…   13 µs
Jun 19 12:40:44 ip-10-0-0-99.poa.internal parity[17243]: 2019-06-19 12:40:44 UTC Syncing  #993996 0x61d6…de17     0.00 blk/s    0.0 t…   13 µs
Jun 19 12:40:49 ip-10-0-0-99.poa.internal parity[17243]: 2019-06-19 12:40:49 UTC Syncing  #993996 0x61d6…de17     0.00 blk/s    0.0 t…   13 µs
[...]

Issue #10724, opened 2019-06-05
Archive node loses peers (stop syncing blocks)


APshenkin:
- **Parity Ethereum version**: 2.4.6
- **Operating system**: Linux
- **Installation**: docker
- **Fully synchronized**: yes
- **Network**: ethereum mainnet
- **Restarted**: yes

Machine spec: 8 CPU 64GB ram, parity cache-size= 32GB

We have launched archive node and complete its sync. Today we update the node from 2.4.5. to 2.4.6 and run some software that fetch blocks information.

It's working, but peers connection is unstable and after some time node stop syncing blocks (it just displays Syncing 0.00 0.00 blk/s )

Below is examples of peers pump and down logs and last blocks (reversed order)

peers info
```


2019-06-05 20:32:52 UTC   12/25 peers     60 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 34380 µs
--
2019-06-05 20:32:12 UTC   10/25 peers     54 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 32475 µs
2019-06-05 20:31:40 UTC   10/25 peers     53 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 32475 µs
2019-06-05 20:31:05 UTC    9/25 peers     53 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 24494 µs
2019-06-05 20:30:34 UTC    9/25 peers     53 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 17362 µs
2019-06-05 20:29:55 UTC    9/25 peers     53 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    1 req/s, 18661 µs
2019-06-05 20:29:20 UTC    9/25 peers     53 MiB chain   17 MiB db  0 bytes queue   79 KiB sync  RPC:  1 conn,    7 req/s, 17362 µs
2019-06-05 20:28:49 UTC    9/25 peers     52 MiB chain   16 MiB db  0 bytes queue   82 KiB sync  RPC:  1 conn,    1 req/s, 32642 µs
[...]

Issue #10626, opened 2019-05-03
Losing peers and falling out of sync


dafky2000:
- **Parity Ethereum version**: 2.4.5
- **Operating system**: Linux
- **Installation**: Arch Package Repo
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

Having issues losing peers since yesterday morning (approx 08:00 - 09:00 UTC). Parity starts up fine, syncs to the latest block, then proceeds to drop peers and fall out of sync regularily.

Likely related?: Sometimes, syncing will stop completely with 0 blocks `Qed`. During these pauses, we seem to accumulate peers normally but syncing remains "stuck" - I have not caught this since enabling trace logging.

During both cases, restart parity then everything resumes correctly for a short time.

---

Full archive + tracing started with: 
```bash
parity --base-path /mnt/hardraid0/io.parity.ethereum/ --mode active --tracing on --pruning archive --db-compaction ssd --cache-size 8192 --min-peers 128 --max-peers 256 --no-periodic-snapshot --jsonrpc-interface 10.0.0.4 --ws-interface 10.0.0.4 --ws-hosts all -l sync=trace
```
- **Log**: [parity_peerdrop.log](https://github.com/paritytech/parity-ethereum/files/3142470/parity_peerdrop.log)

mewwts:
Having this issue on 2.4.5 running `parity --jsonrpc-threads=4 --jsonrpc-server-threads=4` with the following config file 
```
[parity]
auto_update = "none"
no_download = true
base_path = "/mnt/parity-data/parity-data/"
no_persistent_txqueue = true

[network]
nat = "extip:159.89.158.41"
warp = false
allow_ips = "public"
min_peers=40
max_peers=80

[ipc]
disable = true

[rpc]
interface = "all"
apis = ["safe"]

[websockets]
interface = "all"
apis = ["safe"]

[dapps]
disable = true

[footprint]
tracing = "on"
pruning = "archive"
cache_size=12288
```
These nodes are running on 4 vCPUs, 16GB ram.

mewwts:
I wonder if there could me a memory leak somewhere. Here's my memory consumption on two 16GB nodes.
![Screenshot_2019-05-05 DigitalOcean - parity-3](https://user-images.githubusercontent.com/1848670/57192203-2ca99200-6f2e-11e9-8678-334ce800d09a.png)
[...]

PR #10772, opened 2019-06-24
Do not drop the peer with None difficulty

grbIzl:
Several issues about archive nodes behavior were created with similar description: archive node starts loosing peers and stops sync. Sometimes the sync is restored after some period, sometimes no.
See: #10724 #10763 #10626 

The analysis shows one suspicious place, that leads to loosing peers (cut from the logs follows):
...
2019-06-24 06:27:03  IO Worker #1 TRACE sync  Skipping peer 312, force=false, td=Some(10728254029006356954158), our td=10728254029006356954158, state=Idle
...
2019-06-24 06:27:04  IO Worker #0 DEBUG sync  312 -> Dispatching packet: 1
2019-06-24 06:27:04  IO Worker #0 TRACE sync  312 -> NewHashes (1 entries)
2019-06-24 06:27:04  IO Worker #0 TRACE sync  New hash block already queued 0xef843f3c9be62b47228ae6b7819713117e3449dbdf73f8e644e057fe420fe289
2019-06-24 06:27:04  IO Worker #0 TRACE sync  Considering peer 312, force=false, td=None, our td=10728254029006356954158, latest=0xef84…e289, have_latest=true, state=Idle
2019-06-24 06:27:04  IO Worker #0 TRACE sync  peer 312 is not suitable for requesting old blocks, syncing_difficulty=10728254029006356954158, peer_difficulty=None
2019-06-24 06:27:04  IO Worker #0 TRACE sync  Deactivating peer 312
....

As you can see, the peer 312 was deactivated because its difficulty was defined as None. From my point of view it's an erroneous decision and this peer should be treated as legit. For example, compare this logic to the same several LOCs before: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/chain/mod.rs#L963

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -153,7 +153,7 @@ build-android:
 
 build-linux:
   <<:                              *build-on-linux
-  only:                            *releaseable_branches
+  # only:                            *releaseable_branches
 
 build-linux-i386:
   <<:                              *build-on-linux
--- a/ethcore/sync/src/chain/mod.rs
+++ b/ethcore/sync/src/chain/mod.rs
@@ -994,7 +994,7 @@ impl ChainSync {
 					}
 
 					// Only ask for old blocks if the peer has an equal or higher difficulty
-					let equal_or_higher_difficulty = peer_difficulty.map_or(false, |pd| pd >= syncing_difficulty);
+					let equal_or_higher_difficulty = peer_difficulty.map_or(true, |pd| pd >= syncing_difficulty);
 
 					if force || equal_or_higher_difficulty {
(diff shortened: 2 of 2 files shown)


grbIzl:
> I believe this is okay, though I would like a _warp_ sync + ancient blocks to be run to check there is no regression to getting stuck on a "bad" block, which was fixed by #9753.

I've looked through the fix #9753 and it seems to be not affected by this change. You can find the log of running node with this fix in this comment: https://github.com/paritytech/parity-ethereum/issues/10724#issuecomment-505052717

PR #10791, opened 2019-06-26
Removed machine abstraction from ethcore

debris:
we use only one `Machine`


dvdplm:
What was the rationale for being abstract over the machine? Can you give me some context here?

debris:
iirc, `trait Machine` was supposed to provide abstraction necessary to implement blockchains that are significantly different to `Ethereum`... but we've decided to build substrate instead :)

Issue #10793, opened 2019-06-26
Probable deadlock after UnlinkedAncientBlockChain


dvdplm:
If snapshot restoration fails with a `UnlinkedAncientBlockChain` (and you see `Encountered error during snapshot restoration: Snapshot error Unlinked ancient blocks chain` in the console) the client is stuck and Ctrl+c is not enough to shut down properly.

The error can be returned only from two places in [this method](https://github.com/paritytech/parity-ethereum/blob/d2120ded56d3035cce220550aba12e51d7d216bf/ethcore/src/snapshot/service.rs#L351).

Shutting down prints the expected `Finishing work, please wait...` (from [here](https://github.com/paritytech/parity-ethereum/blob/d2120ded56d3035cce220550aba12e51d7d216bf/parity/run.rs#L899)) but the unclean exit handling code in `wait_for_drop()` does not run. This is an indication that we have a deadlock when in this state.

PR #10796, opened 2019-06-26
Better logging when backfilling ancient blocks fail

### LINKED ISSUE #10792: Log progress when restoring snapshots
Currently we print e.g. `Snapshot initializing (1220 chunks restored)` but it'd be better to see `Snapshot initializing (1220/2222 chunks restored, 55% done)`

dvdplm:
**Handle databases with holes**

After snapshots have been downloaded and imported into the new DB we try to salvage existing blocks from the current DB before switching to the new DB ( this happens in `migrate_blocks()`). 

The current code assumes that the blocks in the current DB form a chain all the way down to the genesis block (i.e. they are "ancient blocks"), which is not necessarily true. There are situations where the current DB have some amount of blocks in it that do not stretch all the way back to genesis.

I think the way this situation can occur is:

- start a node with an empty DB
- a snapshot is downloaded and the chain populated with [n..m] blocks, where m is close to the tip and n = m - 10k blocks
- maybe the node stays up and imports block and is generally fine
- the node is stopped and naturally falls out of sync
- the node is started again and as it's too far off the tip, a new snapshot is downloaded
- the previous db will have blocks [i..j] where i != 0 (it'll be 10k blocks before wherever the tip was in the FIRST snapshot) and j is farther behind the current tip - 10k

Before this PR the final step of the second snapshot will fail with an`UnlinkedAncientBlockChain` error because the previous DB's blocks do not stretch all the way back to 0.

With this PR, the attempt to salvage the old blocks is simply abbandoned and the new snapshot is allowed to complete, effectively tossing away the content of the previous DB.

I suspect that it might be possible to do better than that, but best do this step by step.

The diff is rather noisy but the key change of the PR starts [here](https://github.com/paritytech/parity-ethereum/pull/10796/files#diff-09408288a391322d847f21fbec62b925R361).

ref. https://github.com/paritytech/parity-ethereum/issues/10793

Print total blocks imported, closes #10792


ethcore/src/snapshot/service.rs
```diff
@@ -361,11 +352,27 @@ impl Service {
...
 		// so the useful set of blocks is defined as:
 		// [0 ... min(new.first_block, best_ancient_block or best_block)]
+		//
+		// If, for whatever reason, the old db does not have ancient blocks (i.e.
+		// `best_ancient_block` is `None`) AND a non-zero `first_block`, such that the old db looks
+		// like [old_first_block..old_best_block] (which may or may not partially overlap with
+		// [new_first_block..new_best_block]) we do the conservative thing and do not migrate the
+		// old blocks.
 		let find_range = || -> Option<(H256, H256)> {
+			// TODO: In theory, if the current best_block is > new first_block (i.e. they overlap)
```
tomusdrw: I'd keep the comment, but drop the `TODO: ` prefix :)

ethcore/src/snapshot/service.rs
```diff
@@ -361,11 +352,27 @@ impl Service {
...
+			//  we could salvage them but what if there's been a re-org at the boundary and the two
+			//  chains do not match anymore? Would it actually work?
```
tomusdrw: I'd rather drop the old blocks if we are unsure. We should definitely verify if at least couple of latest block we had in the old db matches the new one.

ethcore/src/snapshot/service.rs
```diff
@@ -695,8 +721,9 @@ impl Service {
 			Ok(()) |
 			Err(Error::Snapshot(SnapshotError::RestorationAborted)) => (),
 			Err(e) => {
+				// TODO: after this we're sometimes deadlocked
```
tomusdrw: We should actually always deadlock - because we already have `self.restoration.lock()` acquired in line 720. parking_lot locks are not re-entrant.

Instead of calling `abort_restore()` we should rather use `*restoration = None` here. and maybe add a trace line if you care about it.
Alternatively we could have `abort_restore` take the locked `restoration`. IMHO if we do that for  some methods already we should do that for all of them.

ethcore/src/snapshot/service.rs
dvdplm: For PoW, is there a limit on the length of reorgs?

tomusdrw: Not really, it's only probabilistic. But usually the limiting factor is the state pruning, I don't remember though what value it's set to.

ethcore/src/snapshot/service.rs
dvdplm: > We should actually always deadlock

Well, for some reason we do not. It's pretty rare, I've seen it twice.

> Instead of calling abort_restore() we should rather use *restoration = None

That's what `abort_restore()` does though: [here](https://github.com/paritytech/parity-ethereum/blob/72c58b5319d1d7d9299f7aeded0ff4e6043e38fa/ethcore/src/snapshot/service.rs#L853). Why is it better to do it here?

tomusdrw: > Why is it better to do it here?

To avoid locking twice?

As I said the second ption should be `abort_restore_with_restoration(locked_res: &mut Option<Restoration>)` so we can keep the logic there, but just pass the existing lock.

dvdplm: So the reason I removed `*restoration = None` from here and replaced it with a call to `abort_restore()` was to collect all shutdown actions in one place. Not that `abort_restore()` is doing very much atm, but I still think it's good to have a single point of abortion.
is locking twice really a concern though: this is an error handler and I don't think we care much about performance?

> `abort_restore_with_restoration(locked_res…)`

The `restoration` is [a member of the struct](https://github.com/paritytech/parity-ethereum/blob/72c58b5319d1d7d9299f7aeded0ff4e6043e38fa/ethcore/src/snapshot/service.rs#L234) here, so I'm not sure what the point of passing it as a param tbh. Now you got me confused!

tomusdrw: Since `parking_lot::Mutex` is not re-entrant locking it twice in the same thread can (or does) lead to a deadlock. So the call like that:
```
fn some(&self) {
  let mut restorating = self.restoration.lock();
  ....
  self.abort_restore();
}

fn abort_restore(&self) {
  *self.restoration.lock() = None
}
```
is guaranteed to deadlock, I thought always, but I suspect it might be just random.

In the same file I suppose we already had deadlock issues, so the `with_restoration` pattern emerged, where instead of locking resources locally in a particular function you get passed a (mutable) reference to the locked resource:
```
let mut restoration = self.restoration.lock();
self.feed_chunk_with_restoration(&mut restoration)
```
I propose to use exactly the same patter for `abort_restore` - I'm totally fine grouping all the shutdown actions there, but to prevent potential double-locking and potential deadlocks we can pass the locked resource, so the first example becomes:

```
fn some(&self) {
  let mut restoration = self.restoration.lock();
  ....
  self.abort_restore(&mut restoration);
}

fn abort_restore(&self, res: &mut Option<Restoration>) {
  *res = None
}

Comment (dvdplm) [reply ethcore/src/snapshot/service.rs]: I see, thank you for explaining.

Is this an option?

```
		let r = {
			let mut restoration = self.restoration.lock();
			self.feed_chunk_with_restoration(&mut restoration, hash, chunk, is_state)
		};
		match r {
			…
			Err(e) => {
				…
				self.abort_restore();
				…
			}
		}

```

Comment (ordian) [reply ethcore/src/snapshot/service.rs]: According to [docs](https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html#method.lock):
> Attempts to lock a mutex in the thread which already holds the lock will result in a deadlock.

So I'm not sure either why it didn't always deadlock. The proposed fix looks good.

File: ethcore/src/snapshot/service.rs [hunk 4]
```diff
@@ -690,14 +717,20 @@ impl Service {
...
+		match r {
 			Ok(()) |
 			Err(Error::Snapshot(SnapshotError::RestorationAborted)) => (),
 			Err(e) => {
+				// TODO: after this we're sometimes deadlocked
 				warn!("Encountered error during snapshot restoration: {}", e);
-				*self.restoration.lock() = None;
-				*self.status.lock() = RestorationStatus::Failed;
+				self.abort_restore();
+				if let Some(mut status) = self.status.try_lock_for(std::time::Duration::from_millis(10)) {
```
ordian: I'm not sure if `self.status.lock` can lead to a deadlock unless we acquire locks in different order. Could you elaborate on rationale for introducing `try_lock_for`?

ethcore/src/snapshot/service.rs
```diff
@@ -361,11 +353,27 @@ impl Service {
...
+		// `best_ancient_block` is `None`) AND a non-zero `first_block`, such that the old db looks
+		// like [old_first_block..old_best_block] (which may or may not partially overlap with
+		// [new_first_block..new_best_block]) we do the conservative thing and do not migrate the
+		// old blocks.
 		let find_range = || -> Option<(H256, H256)> {
+			// In theory, if the current best_block is > new first_block (i.e. ranges overlap)
+			// we could salvage them but what if there's been a re-org at the boundary and the two
+			// chains do not match anymore? We'd have to check the existing blocks carefully.
+			if cur_chain_info.ancient_block_number.is_none() && cur_chain_info.first_block_number.unwrap_or(0) > 0 {
+				warn!(target: "blockchain", "blocks in the current DB do not stretch back to genesis; can't salvage them into the new DB. In current DB first block : {:?}/#{:?}, best block: {:?}, #{:?}",
```
ordian: not sure if this should be a `warn!`, maybe `info!` is enough?

ethcore/src/snapshot/service.rs
```diff
@@ -412,7 +424,11 @@ impl Service {
 					next_chain.insert_unordered_block(&mut batch, block, block_receipts, Some(parent_total_difficulty), false, true);
 					count += 1;
 				},
-				_ => break,
+				_ => {
+					error!(target: "snapshot", "migrate_blocks: failed to find receipts and parent total difficulty. Block #{}, parent_hash={:?}, parent_total_difficulty={:?}",
+						block_number, parent_hash, parent_total_difficulty);
+					break
```
ordian: should we return an error here instead of breaking out of the loop?

ethcore/src/snapshot/service.rs
dvdplm: Rationale: this is non-crucial so if it can't take the lock that's ok. `self.status` is updated in several places and I did not want to spend too much time making sure they were ok.

ethcore/src/snapshot/service.rs
dvdplm: Well, we're throwing away their data... but sure I can change to `info!`.

ethcore/src/snapshot/service.rs
dvdplm: The error is returned further down (when `target_hash != parent_hash`) but honestly I don't know why we want to flush to the DB right before returning an error. Maybe @ngotchac knows?

ngotchac: Hm no reason really, I think initially it might have been so you don't have to return the same error twice, but I guess it's the only case where it would error bellow so you could do it right there.

ethcore/src/snapshot/service.rs
```diff
@@ -361,11 +353,27 @@ impl Service {
...
-		// If we are fully synced neither `best_ancient_block` nor `first_block` is set, and we can assume that the whole range from [genesis, best_block] is imported.
-		// The new database only contains the tip of the chain ([first_block, best_block]),
+		// If we are fully synced neither `best_ancient_block` nor `first_block` is set, and we can
+		// assume that the whole range from [genesis, best_block] is imported.
+		// The new database only contains the tip of the chain ([new_first_block, new_best_block]),
 		// so the useful set of blocks is defined as:
 		// [0 ... min(new.first_block, best_ancient_block or best_block)]
+		//
+		// If, for whatever reason, the old db does not have ancient blocks (i.e.
+		// `best_ancient_block` is `None`) AND a non-zero `first_block`, such that the old db looks
```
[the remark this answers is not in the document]
dvdplm: Right, I should move the parens I think.

PR #10799, opened 2019-06-27
updated project to ansi_term 0.11

debris:


ethcore/service/Cargo.toml
```diff
@@ -5,8 +5,7 @@ version = "0.1.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 
 [dependencies]
-ansi_term = "0.10"
-derive_more = "0.14.0"
```
debris: unrelated, unused dependency

debris:
> So what's new in 0.11?

tbh, I didn't check

> Like, what drove this change?

Some of our libraries were already using `0.11`, so I just migrated the rest to remove duplicated dependency

PR #10801, opened 2019-06-27
updated price-info to edition 2018

debris:
also removed redundant `Error` type and unused dev-dependency `parking_lot`


dvdplm:
I've seen `WARN price_info  Failed to auto-update latest ETH price: Fetch(Timeout)` at client startup quite often lately and I think it's spurious as the network is fine otherwise. While you're at it, can you perhaps see if we can increase the timeout a little bit? Or retry?

debris:
> I've seen WARN price_info Failed to auto-update latest ETH price: Fetch(Timeout) at client startup quite often lately and I think it's spurious as the network is fine otherwise. While you're at it, can you perhaps see if we can increase the timeout a little bit? Or retry?

I'll prepare it in a separate PR :)

PR #10803, opened 2019-06-27
Fix a few typos and unused warnings.

afck:


ethcore/src/lib.rs
```diff
@@ -97,7 +95,6 @@ extern crate rand;
 extern crate rayon;
 extern crate rlp;
 extern crate parity_util_mem;
-extern crate parity_util_mem as mem;
```
dvdplm: Not sure this is correct, I think it is needed. cc @cheme

cheme: Those are related to the malloc_size_of_derive macro, but if CI pass it is fine to remove it.

PR #10806, opened 2019-06-27
removed redundant fmt::Display implementations

debris:
removed a bunch of code that is not used for years. Also `Debug` printing using `{:#?}` is more readable

### DIFF
--- a/ethcore/types/src/account_diff.rs
+++ b/ethcore/types/src/account_diff.rs
@@ -16,12 +16,9 @@
 
 //! Diff between two accounts.
 
-use std::cmp::*;
-use std::fmt;
 use std::collections::BTreeMap;
-use std::convert::TryFrom;
-use ethereum_types::{BigEndianHash as _, H256, U256};
 use bytes::Bytes;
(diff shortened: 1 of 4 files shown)


ethcore/types/src/account_diff.rs
```diff
@@ -16,12 +16,9 @@
 
 //! Diff between two accounts.
 
-use std::cmp::*;
-use std::fmt;
 use std::collections::BTreeMap;
-use std::convert::TryFrom;
-use ethereum_types::{BigEndianHash as _, H256, U256};
```
ordian: BigEndianHash is used [here](https://github.com/paritytech/parity-ethereum/blob/a729ca3223f44773ba27a1a5f87f87a73dbefd89/ethcore/types/src/account_diff.rs#L108)

debris: this code has been removed as well

debris:
> Test failures seem legit. You sure about the BigEndianHash removal?

It's a `StateDiff::get` function used in one of the tests. I will look into it

PR #10761, opened 2019-06-19
refactor: evmbin: CLI flag for coloured log levels. CLI flag for output to log file 

ltfschoen:
Extension of PR https://github.com/paritytech/parity-ethereum/pull/10657.
Not to be merged prior to https://github.com/paritytech/parity-ethereum/pull/10657 and https://github.com/paritytech/parity-ethereum/pull/10742

- [x] Temporarily added evmbin/res/create2callPrecompiles.json  and evmbin/res/teststate.json  just as input files for testing the code with CLI commands
- [x] Use fern crate for log level colours
- [x] Add CLI flag `--logging` to choose log level (consistent with parity-ethereum CLI)
- [x] Add CLI flag `--logging-to-file` to save outputs to output.log
- [x] Restrict use of logging levels in production (i.e. `cargo build -p evmbin --release`) to `info!` in Cargo.toml. In development users may use `debug!` or `trace!`. 
- [x] **Question 1** I made changes that to std_json.rs to add colours and log levels, but the tests failed so I reverted the changes in commit https://github.com/paritytech/parity-ethereum/commit/a687150c41af5fb6c8d12ca7cdde4588bf860e81. Were those changes on the right track? 
**ANS: Note that we cannot convert writing to the stdout buffer with `writeln!` in std_json.rs to writing directly to stdout by use of macros with color like `info!`. See the module doc comments in std_json.rs**

The changes may be tested as follows:

**std-json**
```
cargo build -p evmbin --release;
./target/release/parity-evm state-test ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople --logging 2 --std-json --std-dump-json
```

**json**
```
cargo build -p evmbin --release;
./target/release/parity-evm state-test ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople --logging 2 --json
```

**output to a file output.log**
```
cargo build -p evmbin --release;
./target/release/parity-evm state-test ./evmbin/res/create2callPrecompiles.json --only create2callPrecompiles --chain Constantinople --logging 2 --json --logging-to-file
```

PR #10810, opened 2019-06-28
feat: whisper: CLI flag for coloured log levels. CLI flag for output to log file

ltfschoen:
* [x] Usage of fern crate for log levels like in https://github.com/paritytech/parity-ethereum/pull/10761
* [x] Rename CLI flag to `--log-level` instead of `--log`
* [x] Add CLI flag `--log-to-file` to save output to output.log

### Production

```
cargo test -p whisper-cli --release
cargo build -p whisper-cli --release
./target/release/whisper --port 0 --whisper-pool-size 4 --address 127.0.0.1 --rpc-port 8545 --rpc-address 127.0.0.1
```

### Development

```
cargo test -p whisper-cli
cargo build -p whisper-cli
./target/debug/whisper --port 0 --whisper-pool-size 4 --address 127.0.0.1 --rpc-port 8545 --rpc-address 127.0.0.1
```

### DIFF
--- a/whisper/README.md
+++ b/whisper/README.md
@@ -17,9 +17,10 @@ Options:
 	-p, --port PORT                Specify which P2P port to use [default: random].
 	-a, --address ADDRESS          Specify which P2P address to use [default: 127.0.0.1].
 	-s, --secret KEYFILE           Specify which file contains the key to generate the enode.
-    -P, --rpc-port PORT            Specify which RPC port to use [default: 8545].
-    -A, --rpc-address ADDRESS      Specify which RPC address to use [default: 127.0.0.1].
-	-l, --log LEVEL                Specify the logging level. Must conform to the same format as RUST_LOG [default: Error].
+	-P, --rpc-port PORT            Specify which RPC port to use [default: 8545].
+	-A, --rpc-address ADDRESS      Specify which RPC address to use [default: 127.0.0.1].
+	-l, --logging LEVEL            Specify the logging level. Must conform to the same format as RUST_LOG [default: 0]. Log level verbosity configuration choice from highest to lowest priority (error 0, warn 1, info 2, debug 3, trace 4).
+	-L, --logging-to-file          Record logs to output.log file.
 	-h, --help                     Display this message and exit.
 ```
 
--- a/whisper/cli/src/config/logger.rs
+++ b/whisper/cli/src/config/logger.rs
@@ -26,8 +26,7 @@ fn setup_logger(verbosity: u32, logging_to_file: bool) -> Result<(), fern::InitE
 		.debug(Color::BrightBlack)
 		.trace(Color::BrightBlack);
 
-	let colors_level = colors_line.clone()
-		.info(Color::Green);
+	let colors_level = colors_line.clone().info(Color::Green);
 
 	let mut base_config = fern::Dispatch::new();
 
--- a/whisper/cli/src/config/logger.rs
+++ b/whisper/cli/src/config/logger.rs
@@ -19,76 +19,76 @@ use rlog::{LevelFilter};
 use std::io;
 
 fn setup_logger(verbosity: u32, logging_to_file: bool) -> Result<(), fern::InitError> {
-  let colors_line = ColoredLevelConfig::new()
-    .error(Color::Red)
-    .warn(Color::Yellow)
-    .info(Color::BrightBlack)
-    .debug(Color::BrightBlack)
-    .trace(Color::BrightBlack);
+	let colors_line = ColoredLevelConfig::new()
+		.error(Color::Red)
+		.warn(Color::Yellow)
+		.info(Color::BrightBlack)
+		.debug(Color::BrightBlack)
+		.trace(Color::BrightBlack);
 
-  let colors_level = colors_line.clone()
-    .info(Color::Green);
+	let colors_level = colors_line.clone()
+		.info(Color::Green);
 
-  let mut base_config = fern::Dispatch::new();
+	let mut base_config = fern::Dispatch::new();
 
-  base_config = match verbosity {
-    0 => {
-      base_config
-      .level(LevelFilter::Error)
-      .level_for("pretty_colored", LevelFilter::Error)
-    }
-    1 => {
-      base_config
-      .level(LevelFilter::Warn)
-      .level_for("pretty_colored", LevelFilter::Warn)
-    },
-    2 => {
-      base_config
-      .level(LevelFilter::Info)
-      .level_for("pretty_colored", LevelFilter::Info)
-    },
-    3 => {
-      base_config
-      .level(LevelFilter::Debug)
-      .level_for("pretty_colored", LevelFilter::Debug)
-    },
-    _ => {
-      base_config
-      .level(LevelFilter::Trace)
-      .level_for("pretty_colored", LevelFilter::Trace)
-    },
-  };
+	base_config = match verbosity {
+		0 => {
+			base_config
+			.level(LevelFilter::Error)
+			.level_for("pretty_colored", LevelFilter::Error)
+		}
+		1 => {
+			base_config
+			.level(LevelFilter::Warn)
+			.level_for("pretty_colored", LevelFilter::Warn)
+		},
+		2 => {
+			base_config
+			.level(LevelFilter::Info)
+			.level_for("pretty_colored", LevelFilter::Info)
+		},
+		3 => {
+			base_config
+			.level(LevelFilter::Debug)
+			.level_for("pretty_colored", LevelFilter::Debug)
+		},
+		_ => {
+			base_config
+			.level(LevelFilter::Trace)
+			.level_for("pretty_colored", LevelFilter::Trace)
+		},
+	};
 
-  let format_config = fern::Dispatch::new()
-    .format(move |out, message, record| {
-      out.finish(format_args!(
-        "{color_line}[{date}][{target}][{level}{color_line}] {message}\x1B[0m",
-        color_line = format_args!("\x1B[{}m", colors_line.get_color(&record.level()).to_fg_str()),
-        date = chrono::Local::now().format("[%Y-%m-%d][%H:%M:%S]"),
-        level = colors_level.color(record.level()),
-        target = record.target(),
-        message = message
-      ))
-    })
-    .chain(io::stdout());
+	let format_config = fern::Dispatch::new()
+		.format(move |out, message, record| {
+			out.finish(format_args!(
+				"{color_line}[{date}][{target}][{level}{color_line}] {message}\x1B[0m",
+				color_line = format_args!("\x1B[{}m", colors_line.get_color(&record.level()).to_fg_str()),
+				date = chrono::Local::now().format("[%Y-%m-%d][%H:%M:%S]"),
+				level = colors_level.color(record.level()),
+				target = record.target(),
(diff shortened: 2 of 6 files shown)


whisper/cli/src/main.rs
```diff
@@ -68,7 +83,8 @@ Options:
 	-s, --secret KEYFILE           Specify which file contains the key to generate the enode.
 	-P, --rpc-port PORT            Specify which RPC port to use [default: 8545].
 	-A, --rpc-address ADDRESS      Specify which RPC address to use [default: 127.0.0.1].
-	-l, --log LEVEL                Specify the logging level. Must conform to the same format as RUST_LOG [default: Error].
+	-l, --log-level LEVEL          Specify the logging level. Must conform to the same format as RUST_LOG [default: 4]. Log level verbosity configuration choice from highest to lowest priority (error 0, warn 1, info 2, debug 3, trace 4).
```
niklasad1: If we shall change these, it should be the same as in `parity-ethereum` IMHO

https://github.com/paritytech/parity-ethereum/blob/master/parity/cli/mod.rs#L851-#L858

niklasad1: `trace!` is definitely too verbose to be enabled by default

ltfschoen: Ok, I just had a look at that. I think `--logging` sound more like it takes a boolean. But parity-ethereum would get used a lot, so changing that flag to `--log-level` instead wouldn't add much value but it may require a lot of people to have to change their scripts, which may be annoying. So I'll change it to `--logging` instead of `--log-level` and you've suggested

ltfschoen: I'll change it back to `error!` level logging by default.

ltfschoen: @niklasad1 I've pushed two commits in this PR addressing your two comments (i.e. switch to default level of Error (0) and rename `--log-level` to `--logging`.

I've further updated another similar PR for evmbin for consistency with parity-ethereum and this PR in this commit https://github.com/paritytech/parity-ethereum/pull/10761/commits/4d9b74477362394ceebc14583545dc196328378b

whisper/cli/src/config/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+use fern::colors::{Color, ColoredLevelConfig};
+use rlog::{LevelFilter};
+use std::io;
+
+fn setup_logger(verbosity: u32, logging_to_file: bool) -> Result<(), fern::InitError> {
+  let colors_line = ColoredLevelConfig::new()
```
dvdplm: Is indentation ok here? Looks like it's using 2 spaces?

whisper/cli/src/config/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+    .error(Color::Red)
+    .warn(Color::Yellow)
+    .info(Color::BrightBlack)
+    .debug(Color::BrightBlack)
+    .trace(Color::BrightBlack);
+
+  let colors_level = colors_line.clone()
+    .info(Color::Green);
```
dvdplm: why newline here?

whisper/cli/src/config/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+/// - Use of logging level macros from highest priority to lowest: `error!`, `warn!`, `info!`, `debug!` and `trace!`.
+/// - [Compile time filters](https://docs.rs/log/0.4.1/log/#compile-time-filters) that override the CLI logging levels
+/// are configured in Cargo.toml. In production the max logging level may differ from in development.
+/// - Output to output.log when logging_to_file is true.
+pub fn init_logger(logging: &str, logging_to_file: bool) -> () {
+  let verbosity: u32 = logging.parse::<u32>().expect("parsing cannot fail; qed");
+
+  match setup_logger(verbosity, logging_to_file) {
+    Ok(_) => {
+      println!("Success initializing logger. Verbosity: {:?}. Log to file: {}", verbosity, &logging_to_file);
```
dvdplm: Not sure we care about this. If it works I think it should be silent. Also, why `&logging_to_file`: why ref?

whisper/cli/src/main.rs
```diff
@@ -16,16 +16,29 @@
 
 //! Whisper command line interface
 //!
+//! ## Overview
+//!
 //! Spawns an Ethereum network instance and attaches the Whisper protocol RPCs to it.
 //!
+//! ## Usage
+//!
+//! Whisper is not distributed with regular Parity Ethereum releases
```
dvdplm: You sure about this? We do build it as part of the CI pipeline for sure.

whisper/cli/src/main.rs
```diff
@@ -226,7 +243,10 @@ fn execute<S, I>(command: I) -> Result<(), Error> where I: IntoIterator<Item=S>,
 	let pool_size = args.flag_whisper_pool_size * POOL_UNIT;
 	let rpc_url = format!("{}:{}", args.flag_rpc_address, args.flag_rpc_port);
 
-	initialize_logger(args.flag_log);
+	// Default values are defined in the CLI configuration
+	let logging_level = args.clone().flag_logging;
+	let logging_to_file = args.clone().flag_logging_to_file;
+	config::logger::init_logger(&logging_level, logging_to_file);
```
dvdplm: Why not `config::logger::init_logger(args.flag_logging, args.flag.logging_to_file)`?

whisper/README.md
```diff
@@ -17,9 +17,10 @@ Options:
 	-p, --port PORT                Specify which P2P port to use [default: random].
 	-a, --address ADDRESS          Specify which P2P address to use [default: 127.0.0.1].
 	-s, --secret KEYFILE           Specify which file contains the key to generate the enode.
-    -P, --rpc-port PORT            Specify which RPC port to use [default: 8545].
-    -A, --rpc-address ADDRESS      Specify which RPC address to use [default: 127.0.0.1].
-	-l, --log LEVEL                Specify the logging level. Must conform to the same format as RUST_LOG [default: Error].
+	-P, --rpc-port PORT            Specify which RPC port to use [default: 8545].
+	-A, --rpc-address ADDRESS      Specify which RPC address to use [default: 127.0.0.1].
+	-l, --logging LEVEL            Specify the logging level. Must conform to the same format as RUST_LOG [default: 0]. Log level verbosity configuration choice from highest to lowest priority (error 0, warn 1, info 2, debug 3, trace 4).
```
dvdplm: I think we should discourage the use of numeric log levels (I didn't even know they were allowed) and instead only present the string-based level names: `error`, `warn`, `info`, `debug` and `trace`.

whisper/cli/src/config/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+      base_config
+      .level(LevelFilter::Info)
+      .level_for("pretty_colored", LevelFilter::Info)
+    },
+    3 => {
+      base_config
+      .level(LevelFilter::Debug)
+      .level_for("pretty_colored", LevelFilter::Debug)
+    },
+    _ => {
```
dvdplm: Does this mean the default level is `trace`? Should be `info`.

whisper/cli/src/config/logger.rs
```diff
@@ -0,0 +1,114 @@
...
+    .debug(Color::BrightBlack)
+    .trace(Color::BrightBlack);
+
+  let colors_level = colors_line.clone()
+    .info(Color::Green);
+
+  let mut base_config = fern::Dispatch::new();
+
+  base_config = match verbosity {
+    0 => {
```
dvdplm: Use `const` for these? Or strings. Not sure why this takes a `u32` tbh.

whisper/cli/src/config/logger.rs
ltfschoen: i've fixed indentation in commit 120eb4ce775bf751cbbf19bfcf12629eacddd3e8

whisper/cli/src/main.rs
niklasad1: It is not true, https://vanity-service.parity.io/parity-binaries?format=markdown&version=v2.5.4

whisper/cli/src/config/logger.rs
ltfschoen: i've pushed change in commit c793f2017bec5fac9af835463522bc309abfa6ed

whisper/cli/src/main.rs
ltfschoen: i've pushed a change in this commit showing both approaches bc1a3e62dd27724c44d1d69d701ed9848b9520a6

whisper/cli/src/main.rs
ltfschoen: Yes that works! I've pushed changes in commit c1646d94c39644501286662dcf173f05ed38b503

whisper/cli/src/config/logger.rs
ltfschoen: No, funnily enough I discovered that the default is set in the docopt CLI configuration option comments, i.e. `  [default: 0]` below means the default for the `--logging` value is `0` (i.e. Error)
```
	-l, --logging LEVEL            Specify the logging level. Must conform to the same format as RUST_LOG [default: 0]
```

ltfschoen: Actually, yes, if the user specifies a different value it's use trace

whisper/cli/src/config/logger.rs
ltfschoen: I've created a separate branch and tried to change it to string in this commit https://github.com/paritytech/parity-ethereum/commit/6ba4ee4556a6f9689cc79b20d650e6b51b671495

However I get the following error, and I do not understand why it doesn't work, as I've used this technique to before with enums:
```
$ cargo build -p whisper-cli
error: expected one of `::`, `=>`, `if`, or `|`, found `.`
  --> whisper/cli/src/config/logger.rs:58:18
   |
58 |         LogLevel::Error.value() => {
   |                        ^ expected one of `::`, `=>`, `if`, or `|` here

warning: unused import: `LevelFilter`
  --> whisper/cli/src/config/logger.rs:18:12
   |
18 | use rlog::{LevelFilter};
   |            ^^^^^^^^^^^
   |

warning: unreachable expression
  --> whisper/cli/src/config/logger.rs:57:2
   |
57 |       base_config = match verbosity {
   |  _____^
58 | |         LogLevel::Error.value() => {
59 | |             base_config
60 | |             .level(LevelFilter::Error)
...  |
87 | |         },
88 | |     };
   | |_____^
   |
   = note: #[warn(unreachable_code)] on by default

```

whisper/README.md
ltfschoen: See my response to your review comment https://github.com/paritytech/parity-ethereum/pull/10810#discussion_r299821365 that addresses this too

whisper/cli/src/config/logger.rs
ltfschoen: Note that once I figure out how to get this to compile, I'll have to apply the same changes to the evmbin CLI PR https://github.com/paritytech/parity-ethereum/pull/10761/files

ltfschoen:
Closing as whisper was moved to a separate repo before this PR was merged so the same PR has been created in the new whisper repo here https://github.com/paritytech/whisper/pull/1

PR #10812, opened 2019-06-28
refactor: whisper: Add type aliases and update rustdocs in message.rs

ltfschoen:
* [x] Add types aliases to message.rs and update Rustdocs
~~* [ ] Add missing memzero crate~~ - N/A
~~* [ ] Fix use of parity_util_mem instead of memzero~~ - N/A

## Preview changes

```
cargo doc -p parity-whisper --no-deps --open
```

### DIFF
--- a/whisper/src/message.rs
+++ b/whisper/src/message.rs
@@ -92,7 +92,6 @@ impl EnvelopeTopic {
 		for i in 0..3 {
 			let mut topic_idx = topics_data[i] as BloomTopicIndex;
 
-			// What does `1 << i` refer to
 			if topics_data[3] & (1 << i) != 0 {
 				topic_idx += 256;
 			}
--- a/whisper/src/message.rs
+++ b/whisper/src/message.rs
@@ -27,11 +27,38 @@ use tiny_keccak::{keccak256, Keccak};
 #[cfg(not(time_checked_add))]
 use time_utils::CheckedSystemTime;
 
+/// Bloom of topics.
+type Bloom = H512;
+/// Topic data index within a bloom.
+type BloomTopicIndex = usize;
+/// List of envelope topics.
+type EnvelopeTopics = SmallVec<[EnvelopeTopic; 4]>;
+/// Envelope topic data.
+type EnvelopeTopicData = u8;
+/// List of envelope topics data.
+type EnvelopeTopicsData = [EnvelopeTopicData; 4];
+/// Expiry timestamp of an envelope.
+type EnvelopeExpiryTimestamp = u64;
+/// Message contained within an envelope
+type EnvelopeMessage = Vec<u8>;
+/// Arbitrary value used to target lower PoW hash.
+type EnvelopeNonce = u64;
+/// Envelope nonce in bytes.
+type EnvelopeNonceBytes = [u8; 8];
+/// Envelope proving work duration in milliseconds.
+type EnvelopeProvingWorkDuration = u64;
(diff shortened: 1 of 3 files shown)


whisper/src/message.rs
```diff
@@ -44,68 +71,69 @@ pub fn work_factor_proved(size: u64, ttl: u64, hash: H256) -> f64 {
...
 
-		let data = &self.0;
+		let topics_data = &self.0;
 		for i in 0..3 {
-			let mut idx = data[i] as usize;
+			let mut topic_idx = topics_data[i] as BloomTopicIndex;
 
-			if data[3] & (1 << i) != 0 {
-				idx += 256;
+			// What does `1 << i` refer to
```
niklasad1: 1 << i:   is the same 2^i 

so 2^0, 2^1, 2^2.. 2^n .. -> 1, 2, 4, 8,,

niklasad1:
@ltfschoen 

> [x] Add missing memzero crate
   [x] Fix use of parity_util_mem instead of memzero

I can't find those changes, are they missing?

ltfschoen:
> @ltfschoen
> 
> > [x] Add missing memzero crate
> > [x] Fix use of parity_util_mem instead of memzero
> 
> I can't find those changes, are they missing?

Sorry for the confusion. I'm confused as well. I think we can ignore those dot points now. What I think happened this morning is that I got an error suggesting that `use memzero` needed me to add `extern crate memzero`, but I think VSCode may have added `use memzero` automatically or something. I think what's happened is that I tried adding the memzero dependency to the Cargo.toml and then adding `extern crate memzero`, but then I must have realised that we were using `parity_util_mem` instead. So then I must have reverted my memzero changes and the error went away.

whisper/src/message.rs
dvdplm: I for one decidedly prefer using constants or explicit numbers whenever possible.

whisper/src/message.rs
```diff
@@ -291,18 +324,23 @@ impl Message {
 			digest
 		};
 
+		// Find the best nonce based on using updating the digest with
```
dvdplm: "…using updating…" – not sure what this means, can you elaborate?

whisper/src/message.rs
```diff
@@ -344,9 +382,9 @@ impl Message {
 		Message::from_components(envelope, encoded_size, hash, now)
 	}
 
-	// create message from envelope, hash, and encoded size.
-	// does checks for validity.
-	fn from_components(envelope: Envelope, size: usize, hash: H256, now: SystemTime)
+	// Create message from envelope, hash, and encoded size.
+	// Does checks for validity.
```
dvdplm: Maybe?
```suggestion
	// Performs validity checks
```

whisper/src/message.rs
```diff
@@ -27,11 +27,38 @@ use tiny_keccak::{keccak256, Keccak};
...
+type EnvelopeProvingHash = H256;
+/// Envelope work that has been proved by the proving hash.
+type EnvelopeProvenWork = f64;
+/// Time-to-live of an envelope in seconds.
+type EnvelopeTTLDuration = u64;
+
 /// Work-factor proved. Takes 3 parameters: size of message, time to live,
 /// and hash.
 ///
 /// Panics if size or TTL is zero.
```
dvdplm: Document what the return value actually means.

dvdplm:
@ltfschoen merging this now, feel free to address remaining grumbles in a diff PR.

PR #10816, opened 2019-06-28
replace memzero with zeroize crate

ordian:
`parity_util_mem::Memzero` is about to be removed: https://github.com/paritytech/parity-common/pull/183.

Now keys are allocated on the stack instead of the heap, let me know if that's undesirable.


ordian:
> what about the other crates using parity_util_mem?

they use `MallocSizeOf` (heapsize replacement) functionality of that crate, not `Memzero`

PR #10820, opened 2019-06-30
refactor: whisper aes-gcm: Add type aliases and update rustdocs in AES-GCM module

ltfschoen:
* [x] Add types aliases to AES-GCM module and update Rustdocs
* **HELP** How many bits for the `Nonce` type definition (i.e. currently rustdocs says bits are between ???, noting that 1 to 2^64 is the value used for for 128-bit block ciphers, but we are using 256-bit)
* **HELP** What is range of bits for plaintext `Plaintext` type definition (i.e. currently the rustdocs says bits are between ???, noting that 0 to 2^39 - 256 is the value used for for 128-bit block ciphers, but we are using 256-bit.
* **HELP** Is the Authentication Tag Length between 0 and 256 bits

## Preview changes

```
cargo doc -p parity-whisper --open  --no-deps --document-private-items
```

Note: On Debian 9 it may be necessary to run `RUSTFLAGS="-C target-feature=+aes,+ssse3" cargo doc -p parity-whisper --open  --no-deps --document-private-items`

### DIFF
--- a/whisper/src/aes_gcm/mod.rs
+++ b/whisper/src/aes_gcm/mod.rs
@@ -50,7 +50,7 @@ type AuthenticationTagLength = usize;
 /// Plaintext message data to seal.
 type Plaintext = Vec<u8>;
 
-/// Cyphertext.
+/// Ciphertext.
 type Ciphertext = Vec<u8>;
 
 /// AES-GCM encryptor provides the authenticated encryption operation.
--- a/whisper/src/aes_gcm/mod.rs
+++ b/whisper/src/aes_gcm/mod.rs
@@ -14,9 +14,50 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
+//! # AES-GCM Module
+//!
+//! ### Overview
+//!
+//! Whisper's network protocol is secured using the Galois/Counter Mode of
+//! Operation (GCM) block cipher that provides confidentiality and message
+//! authentication assurance to inputs on each invocation (cf. [[2]]). 
+//!
+//! It provides authenticated encryption and signing ("sealing") of data,
+//! and authenticated decryption ("opening") of data that has been protected
+//! with Authenticated Encryption with Associated Data (AEAD).
+//! 
+//! It is constructed from the approved symmetric key block cipher Advanced
+//! Encryption Standard (AES) with a block size of 256-bits.
+//!
+//! [1]: https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf
+//! [2]: http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf
+
 use ring;
 
-/// AES GCM encryptor.
+/// Cryptographic secret key to seal the plaintext message, whose length is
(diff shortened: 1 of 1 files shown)


whisper/src/aes_gcm/mod.rs
```diff
@@ -14,17 +14,54 @@
...
 use ring;
 
-/// AES GCM encryptor.
+/// Cryptographic secret key to seal the plaintext message, whose length is
+/// appropriate for the underlying block cipher.
+type SecretKey = [u8; 32];
+
+/// Initialization Vector (IV) is a nonce used during the sealing process.
+/// It must be distinct for each invocation of the encryption
+/// operation for a fixed value of the Secret Key. It is authenticated but it is
```
dvdplm: Not sure what you mean by the nonce being authenticated tbh.

ltfschoen: It's actually written in http://luca-giuzzi.unibs.it/corsi/Support/papers-cryptography/gcm-spec.pdf on Page 5 of 43 as "The IV is authenticated, and it is not necessary
to include it in the AAD field."

ltfschoen:
> The `SecretKey` type should be changed to a zeroized type using the `parity-mem-util` crate but that can be another PR if you prefer.

I had a look at https://docs.rs/parity-util-mem/0.2.0/parity_util_mem/.
I tried using v0.2 instead of v0.1 that's currently in the Cargo.toml

**Cargo.toml**
```
parity-util-mem = "0.2"
```

**mod.rs**
```
extern crate parity_util_mem as malloc_size_of_is_0;

...

use parity_util_mem::malloc_size_of_is_0;

...

malloc_size_of_is_0!(SecretKey);

```

But I get error:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
  --> whisper/src/aes_gcm/mod.rs:44:1
   |
44 | parity_util_mem::malloc_size_of_is_0!(SecretKey);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ impl doesn't use types inside crate
   |
   = note: the impl does not reference only types defined in this crate
   = note: define and implement a trait or new type instead
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

```

Or should I be using v0.1's `Memzero`, as we're already using in whisper/src/rpc/mod.rs, i.e.

**Cargo.toml**
```
parity-util-mem = "0.1"
```

**mod.rs**
```
use parity_util_mem::Memzero;

...

let key = Memzero::from(rng.gen::<[u8; 32]>());
```

dvdplm:
> Or should I be using v0.1's `Memzero`

Yeah, best do that. We'll do a full sweep when we upgrade to `0.2`.

whisper/src/aes_gcm/mod.rs
dvdplm: I guess I meant to say "I do not understand what is meant by that", not that it is wrong! If you understand it and feel that most code readers will understand is also, then feel free to leave it. If you, like me, suspect most readers will not, then remove it. Your call.

ltfschoen:
Closing as whisper was moved to a separate repo before this PR was merged so the same PR has been created in the new whisper repo here https://github.com/paritytech/whisper/pull/2

PR #10823, opened 2019-06-30
Update wordlist to v1.3

fubuloubu:
`parity-wordlist` was updated to v1.3 for a small issue with the `rand` package that was updated here: https://github.com/paritytech/wordlist/pull/11


dvdplm:
lgtm but we need to await the publication of `1.3`.

cc @tomusdrw

dvdplm:
1.3 is published

dvdplm:
@fubuloubu as `Cargo.lock` is checked in I think you need to run `cargo update -p parity-wordlist` and commit to make CI happy.

fubuloubu:
Oh yeah, thanks!

PR #10830, opened 2019-07-02
Allow --nat extip:your.host.here.org

### LINKED ISSUE #10604: Cannot specify hostname for node URL
- **Parity Ethereum version**: 2.5.0
- **Operating system**: Linux
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

The `--nat` switch accepts `none`, `upnp` and `extip:<IP>`. `extip` accepts only IP address, not DNS host names. Currently this fails due to Rust inference. I don't see any reason why IPs are preferred over host names. When your infrastructure is more dynamic (e.g. teh cloud) host names are usually more stable than IPs.

dvdplm:
Closes #10604

### DIFF
--- a/parity/configuration.rs
+++ b/parity/configuration.rs
@@ -725,9 +725,8 @@ impl Configuration {
 		let port = self.args.arg_ports_shift + self.args.arg_port;
 		let listen_address = SocketAddr::new(self.interface(&self.args.arg_interface).parse().unwrap(), port);
 		let public_address = if self.args.arg_nat.starts_with("extip:") {
-			use regex::Regex;
-			let re = Regex::new(r":.*$").expect("regex is valid");
-			let host = format!("{}:{}", re.replace_all(&self.args.arg_nat[6..], ""), port);
+			let host = self.args.arg_nat[6..].split(':').next().expect("split has at least one part; qed");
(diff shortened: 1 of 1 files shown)


parity/configuration.rs
```diff
@@ -725,9 +725,19 @@ impl Configuration {
 		let port = self.args.arg_ports_shift + self.args.arg_port;
 		let listen_address = SocketAddr::new(self.interface(&self.args.arg_interface).parse().unwrap(), port);
 		let public_address = if self.args.arg_nat.starts_with("extip:") {
-			let host = &self.args.arg_nat[6..];
-			let host = host.parse().map_err(|_| format!("Invalid host given with `--nat extip:{}`", host))?;
-			Some(SocketAddr::new(host, port))
+			use regex::Regex;
+			let re = Regex::new(r":.*$").expect("regex is valid");
+			let host = format!("{}:{}", re.replace_all(&self.args.arg_nat[6..], ""), port);
```
ordian: we can extract the part before colon w/o using regex, e.g. like this:
```rust
let trimmed = self.args.arg_nat.trim_start_matches("extip:");
trimmed.split(':').next().expect("split has at least one part; qed")
```

Issue #10695, opened 2019-05-27
"Unable to sign consensus message: SStore(Invalid Password)" error in PoA network 


adilimroz:
- Parity Ethereum version: 2.4.4 and other newer versions
- Operating system: Linux
- Installation: From source
- Network - Custom poa network

**Steps to reproduce**

1. Create a PoA network with 2 authority nodes.
2. Run the nodes with parity-ethereum  v2.4.4 (or any version newer than 2.4.4).

![Screenshot from 2019-05-27 13-40-06](https://user-images.githubusercontent.com/5220489/58405434-09818680-8085-11e9-8522-1232ab13f775.png)

Note: 
Password file is properly formatted with each password in new line. 
Error comes only with the authority nodes.

hklee93:
- Parity Ethereum version: v2.4.7
- Operating system: Linux (arm64)
- Installation: From snap
- Network - Private PoA network

**Steps to reproduce**

1. Create a PoA network with 2 authority nodes (basically following https://wiki.parity.io/Demo-PoA-tutorial.html, but with actual 3 different physical machines). 

- On node0, there's a validator.
- On node0, there's a regular user. 
- On node1, there's a validator.
- On node1, there's a regular user.
- On node2, there's a regular user.

2. Followed the exact same steps up to #3, and on #4, when I create node.pwds file on both nodes and run each node with `parity --config node0.toml` and `parity --config node1.toml`, I get `2019-06-24 18:36:07 UTC Unable to sign consensus message: SStore(InvalidPassword)` error on both node0 and node1. This is not shown on node2, which is a non-validator node. 

3. Note:
- When I make a transaction on node0, which is a validator node, I have no issue.
- When I make a transaction on node1, which is also a validator node, I get `{"jsonrpc":"2.0","error":{"code":-32021,"message":"Account password is invalid or account does not exist.","data":"SStore(InvalidAccount)"},"id":0}`, which I think it shoulnd't be doing.
- When I make a transaction on node2, which is a non-validator node, I also get `{"jsonrpc":"2.0","error":{"code":-32021,"message":"Account password is invalid or account does not exist.","data":"SStore(InvalidAccount)"},"id":0}`, and I'm not sure if it's normal to only be able to make transactions on validator nodes. I thought making a transaction and creating a block were different. Correct me if I'm wrong.

grbIzl:
[...]

PR #10831, opened 2019-07-02
Remove excessive warning

grbIzl:
This warning is a excessive, because calling sign method of EngineSigner with incorrect password is legit situation. For example, it's being done on the start in order to define proper engine-signer account (all passwords are looped through and tried for signing test message, as result warnings about all incorrect passwords are shown in the console).
Also unifies the logic of  EngineSigner because all other components in this layer (rpc/v1/helpers) silently return the result or error.

Relates to #10695

### DIFF
--- a/rpc/src/v1/helpers/engine_signer.rs
+++ b/rpc/src/v1/helpers/engine_signer.rs
@@ -37,10 +37,7 @@ impl ethcore::engines::EngineSigner for EngineSigner {
 	fn sign(&self, message: ethkey::Message) -> Result<ethkey::Signature, ethkey::Error> {
 		match self.accounts.sign(self.address, Some(self.password.clone()), message) {
 			Ok(ok) => Ok(ok),
-			Err(e) => {
-				warn!("Unable to sign consensus message: {:?}", e);
-				Err(ethkey::Error::InvalidSecret)
-			},
+			Err(e) => Err(ethkey::Error::InvalidSecret),
 		}
 	}
 


dvdplm:
Did you `git blame` to see why it was added in the first place? Also: would s/warn!/trace!/ improve things?

grbIzl:
> Did you `git blame` to see why it was added in the first place? 

Yes, indeed. This warning was introduced by @tomusdrw in his https://github.com/paritytech/parity-ethereum/pull/10213 Before it we tried to sign test message using direct sign call from AccountProvider:

in run.rs:
...
if !passwords.iter().any(|p| miner.set_author(engine_signer, Some(p.to_owned())).is_ok()) {
	return Err(format!("No valid password for the consensus signer {}. {}", engine_signer, VERIFY_PASSWORD_HINT));
...
in miner.rs:
...
fn set_author(&self, address: Address, password: Option<Password>) -> Result<(), AccountError> {
	self.params.write().author = address;

	if self.engine.seals_internally().is_some() && password.is_some() {
		if let Some(ref ap) = self.accounts {
			let password = password.unwrap_or_else(|| Password::from(String::new()));
			// Sign test message
			ap.sign(address.clone(), Some(password.clone()), Default::default())?;
...

@tomusdrw  created EngineSigner in RPC in order to communicate with accounts through it and used its sign. I don't know, why he added a global warning in sign method. All account provider methods and RPC helpers just return error and let the caller log it if needed. 

>Also: would s/warn!/trace!/ improve things?

Obviously it would, but IMO it doesn't make sense for such component to log anything. See before, IMO the user of this component should take care about logging the error. And we have such approach for all other components.

grbIzl:
> Looks good, but can we make sure we have a warning when we actually try to use `EngineSigner` from within `Engine` when new blocks are being sealed?

You mean this warning: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/engines/authority_round/mod.rs#L1196 ?

joshua-mir:
Added patch labels and bumping milestone because this missed being backported

Issue #9114, opened 2018-07-13
Remove engine to client circular reference


sorpaas:
In `ethcore`, the `Engine` trait is usually invoked by `Client`. However, in some cases, an `Engine` would also need to query additional state, and requires a reference back to `Client`. We have `Engine::register_client` and `Error::RequiresClient`. Because this is circular, it may be better if we can remove it.

Currently we only use this feature for Aura engine in two places:

* In `on_close_block`, where we need to calculate how many empty steps have passed.
* In `EpochManager::zoom_to` function, where we need to get the epoch transition for the parent block.

The first one can be easily fixed by using the new AncestryIterator, possibly combined with `set_metadata`.

The second one is a little bit more complicated -- we would need to add an extra field to `BlockDetails`, recording the current epoch transition directly, and then use the same method we allowed engine to fetch finalized info to fetch current epoch transition. The complication arises because this time we cannot avoid a DB migration. But it indeed has its benefits -- the `epoch_transition_for` function can now be always `O(1)`.

adria0:
Closing the issue due to its stale state

PR #10833, opened 2019-07-02
Break circular dependency between Client and Engine (part 1)

dvdplm:
ref. https://github.com/paritytech/parity-ethereum/issues/9114

Don't use the `Client` to [look up the parent block](https://github.com/paritytech/parity-ethereum/pull/10833/files#diff-3b2a71ed5fccb75dfe326f3b122487cdL1244) so we can [calculate the parent step](https://github.com/paritytech/parity-ethereum/pull/10833/files#diff-3b2a71ed5fccb75dfe326f3b122487cdR1247). Instead, store the parent header in the `OpenBlock` and [pass it](https://github.com/paritytech/parity-ethereum/pull/10833/files#diff-e2e171033e005c3b329818aaff213767R313) to `on_close_block()`.

It is unsatisfying to store a whole `Header` in all `OpenBlocks`/`ClosedBlocks` just because Aura need the parent step (a `u64`), but all calls to `OpenBlock::new()` already pass in the parent `Header` (so no new lookup/decoding). I tried to store a reference instead but `reopen()` is problematic.

### DIFF
--- a/ethcore/src/block.rs
+++ b/ethcore/src/block.rs
@@ -52,7 +52,7 @@ use vm::{EnvInfo, LastHashes};
 use hash::keccak;
 use rlp::{RlpStream, Encodable, encode_list};
 use types::transaction::{SignedTransaction, Error as TransactionError};
-use types::header::{Header, ExtendedHeader};
+use types::header::Header;
 use types::receipt::{Receipt, TransactionOutcome};
 
 /// Block that is ready for transactions to be added.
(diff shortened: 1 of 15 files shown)


ethcore/src/block.rs
```diff
@@ -195,7 +195,7 @@ impl<'x> OpenBlock<'x> {
 		engine.populate_from_parent(&mut r.block.header, parent);
 
 		engine.machine().on_new_block(&mut r.block)?;
-		engine.on_new_block(&mut r.block, is_epoch_begin, &mut ancestry.into_iter())?;
```
dvdplm: The ancestry was never used so I'm not sure what it was used for tbh.

debris: good catch!

sorpaas: This was for engines that need to compute additional information regarding finalisation (for example, finalise blocks based on ancestor finalisation information).

Looks like we at least don’t need it right now!

PR #10838, opened 2019-07-04
Extricate PodAccount and state Account to own crates

dvdplm:
Splitting up `ethcore` into smaller chunks is a big chore and this is a timid attempt at that. There are plenty of cross dependencies and there's an example of this in this PR where the tests of `state-account` crate depend on the entire `ethcore`. Putting this up here to get a feel for what people think of the approach in general.


ethcore/src/account_db.rs
```diff
@@ -143,15 +143,12 @@ impl<'db> AccountDBMut<'db> {
...
 	pub fn from_hash(db: &'db mut dyn HashDB<KeccakHasher, DBValue>, address_hash: H256) -> Self {
-		AccountDBMut {
-			db: db,
-			address_hash: address_hash,
-		}
+		AccountDBMut { db, address_hash }
 	}
 
-	#[cfg(test)]
+	/// Create an `AccountDB` from an `AccountDBMut`.
```
ordian: Maybe add a comment this is only used for tests.

PR #10839, opened 2019-07-04
Extract AccountDB to account-db

dvdplm:
Builds on https://github.com/paritytech/parity-ethereum/pull/10838

PR #10840, opened 2019-07-04
Move more code into state-account

dvdplm:
Moves `ethcore/state` into a new `account-state` crate.

State management has a piece that is tightly coupled to other `ethcore` modules, mainly `Machine` and `Excutive` (and related types and support code). 
In order to extract the state management code from `ethcore` I have left one module – [`executive_state`](https://github.com/paritytech/parity-ethereum/pull/10840/files#diff-3e54da02528737bfd0849f5e9b7cbb5aR141) – in `ethcore` that isolates the code where the state code is too entangled with the rest of `ethcore` to be tackled at this time. The tests for state management are also still in `ethcore` because of the tight coupling with other `ethcore` modules. While this is rather ungainly and unsatisfactory I think it's an improvement to have this code in an isolated unit that makes it obvious where the coupling takes place.

The end result is that we now have 6k loc less in `ethcore` and the following new crates:

- `trace`
- `factories`
- `account-state`
- `pod`
- `account-db` from https://github.com/paritytech/parity-ethereum/pull/10839

We also have a somewhat [smaller API surface](https://github.com/paritytech/parity-ethereum/pull/10840/files#diff-4b97685c858e68be56d9d669bd20caaaL165) on `ethcore`.

The diff is very noisy and for that I apologise. The only really new code is in [`executive_state.rs`](https://github.com/paritytech/parity-ethereum/pull/10840/files#diff-3e54da02528737bfd0849f5e9b7cbb5aR141) and the rest is mostly cleanup and munging. None of the `TODO`s that show up as "green" are mine (but still worthy of consideration ofc). I have used this branch to sync Goerli.

Builds on https://github.com/paritytech/parity-ethereum/pull/10839


ethcore/factories/src/lib.rs
```diff
@@ -31,8 +31,16 @@ pub struct VmFactory {
...
+	pub fn create(
+		&self,
+		params: ActionParams,
+		schedule: &Schedule,
+		depth: usize
+	) -> Box<dyn Exec> {
+		if schedule.wasm.is_some()
+			&& params.code.as_ref().map_or(false, |code| {
+			code.len() > 4 && &code[0..4] == WASM_MAGIC_NUMBER
+		}) {
```
debris: 👏 

I truly hate those oneliners

ethcore/src/client/evm_test_client.rs
```diff
@@ -65,12 +69,21 @@ use ethjson::spec::ForkSpec;
...
-fn no_dump_state(_: &state::State<state_db::StateDB>) -> Option<pod_state::PodState> {
+fn no_dump_state(_: &State<state_db::StateDB>) -> Option<PodState> {
+	None
+}
+
+#[cfg(feature = "to-pod-full")]
+fn dump_state(state: &State<state_db::StateDB>) -> Option<PodState> {
+	state.to_pod_full().ok()
+}
+#[cfg(not(feature = "to-pod-full"))]
```
debris: why do we need this feature?

ethcore/src/error.rs
```diff
@@ -220,6 +220,9 @@ pub enum Error {
 	/// A convenient variant for String.
 	#[display(fmt = "{}", _0)]
 	Msg(String),
+	/// State errors
+	#[display(fmt = "State error ({})", _0)]
+	State(account_state::Error),
```
debris: do we still need a `Error::Trie` if we have this one here?

ethcore/trace/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+description = "Transaction tracing"
```
debris: 👍

ethcore/trace/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+name = "trace"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+common-types = { path = "../types"}
```
debris: Can we move here all trace types that this module uses from `common-types` crate?

ethcore/trace/src/db.rs
```diff
@@ -18,17 +18,22 @@
...
+	Key, Writable, Readable, CacheUpdatePolicy,
+};
+use ethcore_blockchain::BlockChainDB;
 use ethereum_types::{H256, H264};
-use kvdb::{DBTransaction};
+use kvdb::DBTransaction;
 use parking_lot::RwLock;
-use types::BlockNumber;
+use common_types::BlockNumber;
+use ethcore_blockchain::{BlockChain, BlockProvider, TransactionAddress};
```
debris: can be merged with line 26

ethcore/trace/src/db.rs
```diff
@@ -332,19 +337,43 @@ impl<T> TraceDatabase for TraceDB<T> where T: DatabaseExtras {
 	}
 }
 
+/// Bridge between TraceDb and Blockchain.
+impl DatabaseExtras for BlockChain {
```
debris: I'm not 100% sure whether this trait impl belongs here. The whole idea was that tracedb and blockchaindb crates are independent and don't know about each other. Placing this implementation here makes blockchaindb one of the tracedb dependencies

ethcore/trace/src/types/flat.rs
```diff
@@ -96,7 +97,7 @@ impl Into<Vec<FlatTrace>> for FlatTransactionTraces {
 
 /// Represents all traces produced by transactions in a single block.
 #[derive(Debug, PartialEq, Clone, Default, RlpEncodableWrapper, RlpDecodableWrapper, MallocSizeOf)]
-pub struct FlatBlockTraces(Vec<FlatTransactionTraces>);
+pub struct FlatBlockTraces(pub(crate) Vec<FlatTransactionTraces>);
```
debris: why is it `pub(crate)` now?

rpc/src/lib.rs
```diff
@@ -103,6 +105,8 @@ extern crate rand_xorshift;
 #[cfg(test)]
 extern crate ethjson;
 #[cfg(test)]
+extern crate patricia_trie_ethereum as ethtrie;
+#[cfg(test)]
 extern crate transaction_pool as txpool;
```
debris: is `tx_pool` used only in tests?

ethcore/src/client/evm_test_client.rs
dvdplm: It's used in a few places, notably in `evm_test_client` where it's used from the outside in `private-tx`. It can most likely be removed by some light refactoring.

ethcore/trace/src/types/flat.rs
dvdplm: I honestly don't remember. It was necessary at some point during the refactor but afaict it works fine now without it. Fixing.

rpc/src/lib.rs
dvdplm: No, but `miner/src/lib.rs` imports its own. `private-tx` does the same.

ethcore/trace/Cargo.toml
dvdplm: Not sure I understand what you mean. The only type from `common-types` that is used in the `trace` crate is `BlockNumber`:

```
$ rg -F "use common_types" ethcore/trace/
ethcore/trace/src/lib.rs
19:use common_types::BlockNumber;

ethcore/trace/src/db.rs
29:use common_types::BlockNumber;
364:	use common_types::BlockNumber;

ethcore/trace/src/import.rs
19:use common_types::BlockNumber;

ethcore/trace/src/types/localized.rs
21:use common_types::BlockNumber;
```

dvdplm: or did you mean moving types from `trace` to `common-types` maybe?

debris: Ah, I forgot that `BlockNumber` type is defined in `common_types`

> or did you mean moving types from trace to common-types maybe? 

The other way round, I was thinking of moving `struct Filter` which is located in `ethcore/types/src/trace_filter.rs`, but I just double checked and I can see now, that it is actually only used by client. So please ignore my previous comment 🙃

ethcore/trace/src/types/flat.rs
dvdplm: I found it: it's a new error (i.e. a warning is now promoted to error):

```
error: private struct constructors are not usable through re-exports in outer modules
   --> ethcore/trace/src/types/flat.rs:226:4
    |
226 |             FlatTransactionTraces(vec![flat_trace]),
    |             ^^^^^^^^^^^^^^^^^^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #39207 <https://github.com/rust-lang/rust/issues/39207>
```

ethcore/trace/Cargo.toml
dvdplm: > Ah, I forgot that `BlockNumber` type is defined in `common_types`

Btw I noticed we define __two__ `BlockNumber` types, the other is [here](https://github.com/paritytech/parity-ethereum/blob/7c335e87642b848f61ab3ee7bbff5964b53d6f11/rpc/src/v1/helpers/poll_filter.rs#L28).

ethcore/trace/src/db.rs
dvdplm: Great idea, fixing.

ethcore/trace/src/types/flat.rs
dvdplm: Interestingly that error shows up only if I run `cargo test -p trace` and not with `cargo test --all`. Hmm.

ethcore/src/error.rs
dvdplm: Hmm, if I remove it most compilation errors are indeed from state-related code but there are still a bunch from snapshot code that also pokes at the trie directly.

ethcore/src/client/evm_test_client.rs
dvdplm: removed the feature

ethcore/src/error.rs
dvdplm: I looked into this and it seems fairly involved. Changing `account-state` to use a standard `Result` with its own `Error` is easy but there's plenty of code in `ethcore` that returns `vm::Result`, which means that I need a `From<account_state::Error> for vm::Error` which obviously doesn't work.
I think the "fix" is to change `ethcore` to stop returning `vm::Result`. I'd rather not do that refactoring in this PR.

dvdplm: (and besides, there's still `snapshot` that still needs use `TrieError` so we wouldn't get rid of it here anyway)

ethcore/account-state/Cargo.toml
```diff
@@ -1,28 +1,37 @@
...
 version = "0.1.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
 common-types = { path = "../types"}
+derive_more = "0.15.0"
 ethereum-types = "0.6.0"
 ethtrie = { package = "patricia-trie-ethereum", path = "../../util/patricia-trie-ethereum" }
+evm = { path = "../evm" }
```
tomusdrw: In future it would be good to clean up some of the dependencies. I don't think the `Ethereum accounts` should require `evm` dependency.

ethcore/account-state/src/substate.rs
```diff
@@ -68,8 +70,9 @@ impl Substate {
 
 #[cfg(test)]
 mod tests {
-	use super::{Substate, Address};
-	use types::log_entry::LogEntry;
+	use ethereum_types::Address;
+	use common_types::log_entry::LogEntry;
```
tomusdrw: Was the idea to sort them? Seems sort before rename didn't work that good.

ethcore/blockchain/src/database_extras.rs
```diff
@@ -14,14 +14,27 @@
...
-use trace::DatabaseExtras as TraceDatabaseExtras;
-use types::BlockNumber;
+use ethcore_db::keys::TransactionAddress;
 
-impl TraceDatabaseExtras for BlockChain {
+use crate::blockchain::{BlockProvider, BlockChain};
+
+/// `DatabaseExtras` provides an interface to query extra data which is not stored in TraceDB,
+/// but necessary to work correctly.
+pub trait DatabaseExtras {
```
tomusdrw: To be fair IMHO this trait should be part of the `TraceDB` modules, cause it seems it's required there.
IMHO any place where we have a trait and an implementation for a non-dummy struct right below suggests some issue with code split.

ethcore/factories/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+description = "Collection of factories"
```
tomusdrw: Imho it's a bit too generic (both name and description), as it might attract some more code in the future and will end up messy.

Is there a way to make it more specific? I believe it's used for trie and VM factories, no? Maybe worth splitting the two?
Or just change the name to `trie-vm-factories` to avoid attracting more "factory-like" code to this crate.

ethcore/trace/src/db.rs
```diff
@@ -18,17 +18,21 @@
...
+	cache_manager::CacheManager,
+	Key, Writable, Readable, CacheUpdatePolicy,
+};
 use ethereum_types::{H256, H264};
-use kvdb::{DBTransaction};
+use kvdb::DBTransaction;
 use parking_lot::RwLock;
-use types::BlockNumber;
+use common_types::BlockNumber;
+use ethcore_blockchain::{BlockChainDB, DatabaseExtras};
```
tomusdrw: Worth sorting?

ethcore/trace/Cargo.toml
tomusdrw: It's a typedef anyway, so it really doesn't matter that much, since you can just use `u64` (which I believe we do in many places) and won't even notice that you don't use "the correct" type.

ethcore/trace/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+ethcore-blockchain = { path = "../blockchain" }
+ethcore-db = { path = "../db" }
+ethereum-types = "0.6"
+evm = { path = "../evm" }
```
tomusdrw: Again requiring entire EVM seems odd, would be worth to remove it in the future.

ethcore/trace/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+description = "Transaction tracing"
+name = "trace"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+common-types = { path = "../types"}
+ethcore-blockchain = { path = "../blockchain" }
```
tomusdrw: Do we really need the entire blockchain or just some trait definitions?

ethcore/trace/src/types/flat.rs
tomusdrw: @dvdplm Perhaps new crates need to be added to the workspace? Can you double check that the tests are actually run on the CI?

ethcore/blockchain/src/database_extras.rs
dvdplm: @debris asked for it to be moved here to avoid tight coupling. Personally I think it is better off here, but if you feel strongly otherwise I can move it back.

tomusdrw: Well now, we couple `trace` with `blockchain`. I think it would be best to have the trait in `trace` but keep the implementation outside of `blockchain` crate as well (the impl should live in a place where it's actually instantiated) - that way you can avoid coupling at all.

ethcore/account-state/src/substate.rs
dvdplm: I think it's my IDE that first groups imports in std, external crates, local crates and then sorts each group, but yeah looks odd.

ethcore/trace/Cargo.toml
dvdplm: Currently we use `BlockChainDB` and the contentious `DatabaseExtras`, e.g.:

```
pub struct TraceDB<T> where T: DatabaseExtras {
	/// cache
	traces: RwLock<HashMap<H256, FlatBlockTraces>>,
	/// hashes of cached traces
	cache_manager: RwLock<CacheManager<H256>>,
	/// db
	db: Arc<dyn BlockChainDB>,
	/// tracing enabled
	enabled: bool,
	/// extras
	extras: Arc<T>,
}
```

ethcore/trace/Cargo.toml
dvdplm: We use `CallType` from the `evm`, that's all. Seems like a good opportunity for a refactor indeed.

ethcore/account-state/Cargo.toml
dvdplm: We use `CallType` from the `evm`. Maybe we can move that to `common_types`.

ethcore/blockchain/src/database_extras.rs
debris: > @debris asked for it to be moved here to avoid tight coupling. Personally I think it is better off here, but if you feel strongly otherwise I can move it back.

not exactly. I wanted this trait to remain in trace library, but only an implementation of it to be moved elsewhere. Something like this:

- blockchain - defines `BlockchainDB`
- trace - defines `TraceDB` and `DatabaseExtras`
- ethcore - defines a structure that holds `BlockchainDB` (e.g. Client) and implements `DatabaseExtras` for it

ethcore/src/client/evm_test_client.rs
```diff
@@ -100,7 +108,7 @@ impl<'a> EvmTestClient<'a> {
 	}
 
 	/// Change default function for dump state (default does not dump)
-	pub fn set_dump_state_fn(&mut self, dump_state: fn(&state::State<state_db::StateDB>) -> Option<pod_state::PodState>) {
+	pub fn set_dump_state(&mut self) {
 		self.dump_state = dump_state;
```
debris: this function looks a bit weird

dvdplm:
@tomusdrw @debris Merging this but I'll make sure the following gets addressed in follow up PRs:

1. sort out the trace <–> blockchain interdependencies
cf. https://github.com/paritytech/parity-ethereum/pull/10840#discussion_r301080741

2. attempt to avoid pulling in EVM in `trace` and `account-state`

3. make the conditional "dump state" suck a bit less

ethcore/blockchain/src/database_extras.rs
dvdplm: Addressed [here](https://github.com/paritytech/parity-ethereum/pull/10868).

ethcore/trace/Cargo.toml
dvdplm: What is your thinking here? Are you thinking we should move trait to own crates?

ethcore/trace/src/types/flat.rs
dvdplm: @tomusdrw The failures do show up in CI (but not locally with `--all`). :/

PR #10843, opened 2019-07-05
make fn submit_seal more idiomatic

debris:
and more readable


ethcore/src/miner/miner.rs
```diff
@@ -1306,31 +1306,39 @@ impl miner::MinerService for Miner {
...
+			})?;
 
-		result.and_then(|sealed| {
-			let n = sealed.header.number();
-			let h = sealed.header.hash();
-			info!(target: "miner", "Submitted block imported OK. #{}: {}", Colour::White.bold().paint(format!("{}", n)), Colour::White.bold().paint(format!("{:x}", h)));
-			Ok(sealed)
-		})
+		trace!(target: "miner", "Submitted block {hash}={bare_hash} with seal {seal:?}",
+			   hash = block_hash,
```
ordian: mixed tabs and spaces, a proper way to indent multiline function call with tabs is 
```rust
call(
	arg1,
	arg2,
);
```

ethcore/src/miner/miner.rs
```diff
@@ -1306,31 +1306,39 @@ impl miner::MinerService for Miner {
...
+			.map_err(|e| {
+				warn!(target: "miner", "Mined solution rejected: {}", e);
+				Error::PowInvalid
+			})?;
+
+		let n = sealed.header.number();
+		let h = sealed.header.hash();
+
+		info!(target: "miner", "Submitted block imported OK. #{number}: {hash}",
+			  number = Colour::White.bold().paint(n.to_string()),
```
ordian: ditto

PR #10848, opened 2019-07-05
bump crossbeam

debris:
- reduces the number of duplicated dependencies

### DIFF
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -34,7 +34,7 @@ use externalities::*;
 use trace::{self, Tracer, VMTracer};
 use types::transaction::{Action, SignedTransaction};
 use transaction_ext::Transaction;
-use crossbeam;
+use crossbeam_utils::thread;
 pub use executed::{Executed, ExecutionResult};
 
 #[cfg(debug_assertions)]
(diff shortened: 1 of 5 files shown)


ethcore/src/executive.rs
```diff
@@ -977,11 +977,18 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
+				scope.builder()
+					.stack_size(stack_size)
+					.spawn(|_| {
+						self.call_with_stack_depth(params, substate, stack_depth, tracer, vm_tracer)
+					})
+					.expect("Sub-thread creation cannot fail; the host might run out of resources; qed")
+					.join()
+			})
+			.expect("Sub-thread never panics; qed")
+			.expect("Sub-thread never panics; qed")
```
debris: crossbeamapi has changed a bit here,
- We have to call `join` inside `fn scope` callback, cause the `join` handle lifetime is now the same as scope lifetime :(
- `fn scope` returns error, so that's why we have double `expect` here

dvdplm: > fn scope returns error, 

What are the conditions under which it errors?

debris: > If the child thread panics, an error is returned.

from https://docs.rs/crossbeam/0.7.1/crossbeam/thread/struct.ScopedJoinHandle.html

and

> If any of the joined threads has panicked, an Err is returned containing errors from panicked threads.

from https://docs.rs/crossbeam/0.7.1/crossbeam/fn.scope.html

PR #10849, opened 2019-07-05
removed redundant fork choice abstraction

debris:
fork_choice is always based on total difficulty, so it does not need to be abstracted

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -40,7 +40,7 @@ use types::encoded;
 use types::filter::Filter;
 use types::log_entry::LocalizedLogEntry;
 use types::receipt::{Receipt, LocalizedReceipt};
-use types::{BlockNumber, header::{Header, ExtendedHeader}};
+use types::{BlockNumber, header::Header};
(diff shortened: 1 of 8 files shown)


ethcore/src/client/client.rs
```diff
@@ -501,32 +501,24 @@ impl Importer {
 		let traces = block.traces.drain();
 		let best_hash = chain.best_block_hash();
 
-		let new = ExtendedHeader {
-			header: header.clone(),
```
debris: redundant clone

ethcore/src/client/client.rs
```diff
@@ -501,32 +501,24 @@ impl Importer {
...
+				.expect("Parent block is in the database; qed")
+				.total_difficulty;
+			parent_total_difficulty + header.difficulty()
 		};
 
-		let best = {
-			let header = chain.block_header_data(&best_hash)
-				.expect("Best block is in the database; qed")
-				.decode()
-				.expect("Stored block header is valid RLP; qed");
```
debris: redundant database query and block header decoding

PR #10853, opened 2019-07-06
Replace all calls of mem::uninitialized with mem::MaybeUninit

### LINKED ISSUE #10842: Remove all uses of mem::uninitialized
1.36 is stable and we're using `mem::uninitialized` in a few places. Replace with `MaybeUninit`.

drdgvhbh:
This PR replaces all calls of `mem::uninitialized` with `mem::MaybeUninit::uninit().assume_init`.

This fixes #10842

### DIFF
--- a/ethash/src/compute.rs
+++ b/ethash/src/compute.rs
@@ -141,19 +141,22 @@ pub fn quick_get_difficulty(header_hash: &H256, nonce: u64, mix_hash: &H256, pro
 			//
 			// This cannot be elided by the compiler as it doesn't know the implementation of
 			// `keccak_512`.
-			let mut buf: [u8; 64 + 32] = mem::MaybeUninit::uninit().assume_init();
+			let mut buf = mem::MaybeUninit::<[u8; 64 + 32]>::uninit();
 
-			ptr::copy_nonoverlapping(header_hash.as_ptr(), buf.as_mut_ptr(), 32);
-			ptr::copy_nonoverlapping(&nonce as *const u64 as *const u8, buf[32..].as_mut_ptr(), 8);
+			ptr::copy_nonoverlapping(header_hash.as_ptr(), buf.as_mut_ptr() as *mut u8, 32);
+			ptr::copy_nonoverlapping(&nonce as *const u64 as *const u8, buf.assume_init()[32..].as_mut_ptr(), 8);
 
-			keccak_512::unchecked(buf.as_mut_ptr(), 64, buf.as_ptr(), 40);
-			ptr::copy_nonoverlapping(mix_hash.as_ptr(), buf[64..].as_mut_ptr(), 32);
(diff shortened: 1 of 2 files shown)


drdgvhbh:
[clabot:check]

ethash/src/compute.rs
```diff
@@ -141,7 +141,7 @@ pub fn quick_get_difficulty(header_hash: &H256, nonce: u64, mix_hash: &H256, pro
 			//
 			// This cannot be elided by the compiler as it doesn't know the implementation of
 			// `keccak_512`.
-			let mut buf: [u8; 64 + 32] = mem::uninitialized();
+			let mut buf: [u8; 64 + 32] = mem::MaybeUninit::uninit().assume_init();
```
ordian: This is undefined behavior (as it was before), see https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant. The correct way to use `MaybeUninit` is to initialize the value to type `MaybeUninit<[u8; 64 + 32]>`, then actually initialize it (with `copy_nonoverlapping`) and then call `assume_init` to extract the underlying type (and ensure drop).

ethash/src/compute.rs
```diff
@@ -141,19 +141,22 @@ pub fn quick_get_difficulty(header_hash: &H256, nonce: u64, mix_hash: &H256, pro
...
-			ptr::copy_nonoverlapping(mix_hash.as_ptr(), buf[64..].as_mut_ptr(), 32);
+			keccak_512::unchecked(buf.as_mut_ptr() as *mut u8, 64, buf.assume_init().as_ptr(), 40);
+			ptr::copy_nonoverlapping(mix_hash.as_ptr(), buf.assume_init()[64..].as_mut_ptr(), 32);
 
 			// This is initialized in `keccak_256`
-			let mut hash: [u8; 32] = mem::MaybeUninit::uninit().assume_init();
-			keccak_256::unchecked(hash.as_mut_ptr(), hash.len(), buf.as_ptr(), buf.len());
+			let mut hash = mem::MaybeUninit::<[u8; 32]>::uninit();
+			keccak_256::unchecked(hash.as_mut_ptr() as *mut u8,
+								  hash.assume_init().len(),
```
drdgvhbh: Is this considered safe? Taking the length of an "assume_init" unitialized array.

dvdplm: Isn't that calling `assume_init()` before initializing the memory (which is UB)? Why not use `32`?

drdgvhbh: Yeah thats what I thought of doing. How does the previous code work though. It was using `len` on uninitialized memory as well

dvdplm: Isn't that kind of the thing with UB: it works until it doesn't!

ordian:
closing if favor of #10861

Issue #10821, opened 2019-06-30
unreasonably high memory usage (without crash) and won't shut down


iFA88:
Greetings, sadly my `Parity-Ethereum/v2.4.6-stable-94164e1-20190514/x86_64-linux-gnu/rustc1.34.1` node eats unreasonably high memory.
Node log and process statistics in CSV : https://www.fusionsolutions.io/doc/memlog.tar.gz

Start parameters are:
```
--ipc-apis all --reserved-peers /own/config/archiveEthNode.txt --no-serve-light --no-periodic-snapshot --jsonrpc-allow-missing-blocks --no-persistent-txqueue --jsonrpc-server-threads 8 --ipc-path=/own/sockets/ethNode.ipc --min-gas-price=10000000 --tx-queue-mem-limit=4096 --tx-queue-size=256000 --reseal-on-txs=all --force-sealing --base-path "/mnt/node-1/eth" --rpcport 8548 --port 30306 --no-ws --no-secretstore --cache-size 4096 --log-file /own/log/nodes/eth/parity_eth_$DATE.log"
```
The memory usage will not be higher as 12gb.

On `16:20:20` have killed the process with `KILLSIG`, this is the only way that I can shut down the process.

I glad help with any trace parameters or statistics.

dvdplm:
How did you collect the memory stats showed in the csv?

iFA88:
Like in python:
```
import psutil
process = psutil.Process(PID)
print(process.memory_info().rss)
```
Gives the same as the `htop`:
![kép](https://user-images.githubusercontent.com/18205395/60398445-670f6400-9b58-11e9-8ac6-c0db67650c9f.png)

dvdplm:
So it's RSS, perfect! :)

The number of pending txs is pretty high, is that a normal amount in your setup?

iFA88:
Yeah i'm parsing pending transactions to my DB. Check the my start parameters :)

dvdplm:
Other than staying in sync, what is the node doing? I.e. what kind of RPC traffic is it used for?

iFA88:
I'm using for every new block this RPC's: `trace_block` `eth_getBlockByNumber` `eth_getUncleByBlockHashAndIndex` `eth_blockNumber` `eth_getTransactionReceipt`.

dvdplm:
[...]

PR #10864, opened 2019-07-08
Treat blocks, that already downloaded, as synced

grbIzl:
This PR aims to fix the following scenario in the block sync:

1. Node downloads the next block number N and starts importing it. Due to block's complexity or some configuration issues block's import takes more time (I've seen this issue, when import took 5-10 seconds).
2. Meanwhile the next sync round has begun and Parity tries to sync block N once more time, obviously import of this block has no success with error AlreadyQueued. As a result the sync round has completed with 0 blocks synced
3. As no new blocks were synced during sync round Parity starts descending in block number, treating this situation as error during blocks download: https://github.com/paritytech/parity-ethereum/blob/master/ethcore/sync/src/block_sync.rs#L429 and repeats sync for (already downloaded) blocks
4. Attempt to sync block number N-1 returns the same error and causes attempt to download block N-2 and so on
5. This downfall only finishes, when (if) block number N is imported and Parity catches up the current best block
6. As a result huge DDOS of GetBlock requests outcomes from the node. This also can affect current block's verification and exacerbates the issue.

Suggested solution:
Treat downloaded blocks on step 2 and further, that exist already in the chain or in the queue, as actually  synced on this round and (as result) do not start error handling procedure with blocks descending. 

Comments: 

- Relates to #10724. 
- Also such DDOS can cause the hanging of other node (see #10821 ), that replies to these massive GetBlocks requests.
- This issue has similarities to the described in https://github.com/paritytech/parity-ethereum/pull/9531 but has different scenario and root causes are not connected.


grbIzl:
I put this PR to onice in order to perform more testing. Please do not merge it before it's done.

dvdplm:
@tomusdrw Can you take a look at this please? What are the consequences of considering blocks that are currently getting processed as imported?

grbIzl:
This change breaks fork's handling logic in block sync. More accurate approach is required

grbIzl:
> Since we can only import one block a the time, can't we have an atomic operation that actually makes sure that we shouldn't re-try importing it, cause it's already being imported?

It already works so. Probably I was not able to emphasize this subtle difference between importing and syncing of blocks. Saying syncing, I meant downloading of block's header, body etc. Import happens for synced block and it is indeed atomic and we don't allow multiple re-imports of the block. The problem, that this PR addresses, is multiple re-sync of already synced blocks. 

>Alternatively we could return a different error code than AlraedyQueued to distinguish this situation.

I came to the nearly the same solution.

lazaridiscom:
As a general rule, Error-Codes should reflect reality. Thus, changing "AlreadyQueued" to the real error-code like e.g. "ImportOngoing", is the first step to take. The act on this error-code, in an non-breaking manner.

Changing the logic, place atomic operations etc. should be avoided, because this can introduce unexpected behavior.

lazaridiscom:
> (I've seen this issue, when import took 5-10 seconds

possibly related to #10085

grbIzl:
PR is verified. Closing it in favor of the newly created #11264 with clear history.

PR #10866, opened 2019-07-08
Update README.md and Changelogs

s3krit:
* README.md referred to an out-of-date beta release so has been updated.
* Changelogs were quite out of date, so new ones containing the missing changes have been created.


s3krit:
> Are we getting rid of commit messages in the changelog, or is this just how your scripts currently work?

Hmm, interesting point. The script I use to build the changelogs simply takes the text from the release description for each release. Since the format of these isn't yet consistent, it appears it left out some important information (namely, the full list of commits/links to their relevant PRs). I'll fix this up and push my changes.

PR #10868, opened 2019-07-08
Move DatabaseExtras back to trace

dvdplm:
Move `DatabaseExtras` back to `trace` and remove the impl for `BlockChain`. Instead add a `BlockChainWithExtras` type to `ethcore` and impl `DatabaseExtras` for it.

ref. https://github.com/paritytech/parity-ethereum/pull/10840#discussion_r301080741


ethcore/src/client/client.rs
```diff
@@ -171,6 +171,35 @@ struct Importer {
...
+					block_hash,
+					index: tx_position
+				};
+				self.0.transaction(&tx_address)
+			})
+			.map(|tx| tx.hash())
+	}
+}
+
+impl From<Arc<BlockChain>> for BlockChainWithExtras {
```
tomusdrw: I'd be in favour of `derive_more::From` to avoid this boilerplate.

ethcore/src/client/client.rs
```diff
@@ -171,6 +171,35 @@ struct Importer {
...
 }
 
+
+/// Newtype around an `Arc<BlockChain>` so we can glue together `BlockChain` (from the `blockchain`
+/// crate) and `DatabaseExtras` (from the `trace` crate).
+struct BlockChainWithExtras(Arc<BlockChain>);
+
+impl trace::DatabaseExtras for BlockChainWithExtras {
+	fn block_hash(&self, block_number: BlockNumber) -> Option<H256> {
+		(self.0.as_ref() as &dyn BlockProvider).block_hash(block_number)
```
tomusdrw:
```suggestion
		BlockProvider::block_hash(&self.0, block_number)
```

ethcore/src/client/client.rs
```diff
@@ -171,6 +171,35 @@ struct Importer {
...
+	}
+
+	fn transaction_hash(&self, block_number: BlockNumber, tx_position: usize) -> Option<H256> {
+		(self.0.as_ref() as &dyn BlockProvider).block_hash(block_number)
```
tomusdrw:
```suggestion
		BlockProvider::block_hash(&self.0, block_number)
```

ethcore/trace/Cargo.toml
```diff
@@ -6,7 +6,6 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-common-types = { path = "../types"}
 ethcore-blockchain = { path = "../blockchain" }
```
tomusdrw: Hmm, so we still can't get rid of the blockchain depenency? Then I don't think it makes much sense.

ethcore/trace/src/db.rs
```diff
@@ -57,6 +57,16 @@ impl Key<FlatBlockTraces> for H256 {
 	}
 }
 
+/// `DatabaseExtras` provides an interface to query extra data which is not stored in TraceDB,
+/// but necessary to work correctly.
+pub trait DatabaseExtras {
```
tomusdrw: If we still need the `blockchain` dependenccy we can probably just implement `DatabaseExtras` for any `BlockProvider` here easily. There is no need for a wrapper struct afaict.

ethcore/trace/Cargo.toml
dvdplm: No, we cannot: https://github.com/paritytech/parity-ethereum/blob/44cc442d12348641395531980bf913659561d012/ethcore/trace/src/db.rs#L72

ethcore/src/client/client.rs
dvdplm:
```
error[E0277]: the trait bound `std::sync::Arc<blockchain::BlockChain>: blockchain::BlockProvider` is not satisfied
   --> ethcore/src/client/client.rs:181:3
    |
181 |         BlockProvider::block_hash(&self.0, block_number)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `blockchain::BlockProvider` is not implemented for `std::sync::Arc<blockchain::BlockChain>`
    |
    = note: required by `blockchain::BlockProvider::block_hash`
```

seunlanlege: I guess `self.0` is of type `std::sync::Arc<blockchain::BlockChain>` you'll need to deref.

`BlockProvider::block_hash(&*self.0, block_number)`

ethcore/trace/Cargo.toml
debris: `dyn BlockChainWithExtras`? :)

dvdplm: `dyn BlockChainWithExtrasAndExtraMayo` ;)

Sure, we can work on disentangling things further, but I think we can get more bang for our effort elsewhere at this time. I'm fine adding an issue to make `trace` independent of `blockchain` but for the time being I think we should pick the lesser of two evils, so the question is: which is worst, this PR or the current master? Personally I think they both smell a little so I don't really care which it is. @tomusdrw seem to feel this PR is worse. What is your opinion @debris, merge or close?

tomusdrw: Since we depend on `blockchain` the wrapper type is unecessary since we can write the implementation right next to the `DatabaseExtras` definition. I'd just add this in `traces`:
```rust
impl<T: BlockProvider> trace::DatabaseExtras for T {
	fn block_hash(&self, block_number: BlockNumber) -> Option<H256> {
		BlockProvider::block_hash(&self, block_number)
	}

	fn transaction_hash(&self, block_number: BlockNumber, tx_position: usize) -> Option<H256> {
		self.block_hash(block_number)
			.and_then(|block_hash| {
				let tx_address = TransactionAddress {
					block_hash,
					index: tx_position
				};
				self.transaction(&tx_address)
			})
			.map(|tx| tx.hash())
	}
}
```
and be done with it.

dvdplm:
Updated as per @tomusdrw's grumble here: https://github.com/paritytech/parity-ethereum/pull/10868#discussion_r302451378

ethcore/trace/src/db.rs
```diff
@@ -57,6 +57,34 @@ impl Key<FlatBlockTraces> for H256 {
...
+	/// Returns hash of given block number.
+	fn block_hash(&self, block_number: BlockNumber) -> Option<H256>;
+
+	/// Returns hash of transaction at given position.
+	fn transaction_hash(&self, block_number: BlockNumber, tx_position: usize) -> Option<H256>;
+}
+
+impl<T: BlockProvider> DatabaseExtras for T {
+	fn block_hash(&self, block_number: BlockNumber) -> Option<H256> {
+		self.block_hash( block_number)
```
tomusdrw: Does it actually call the right version of `block_hash`? Or just recursively calls itself?

dvdplm: I just assumed the tests would ensure that, but looking at the test mocks we have I am doubtful too, so best be explicit. Good catch.

tomusdrw: I think rust compiler gives a warning in such cases, so hopefuly we're fine. But indeed might be worth a test or at least explicit notation.

dvdplm: I think it does too, but to clear up the doubt (and save any readers some brain-friction): https://github.com/paritytech/parity-ethereum/pull/10868/commits/2d06a4e89aa5ca0cc014ad972ab4634cfe1ce65c

PR #10874, opened 2019-07-11
Integration of EIP1962 precompile

shamatar:
This PR incorporates reference implementation for [EIP1962](http://eips.ethereum.org/EIPS/eip-1962) into Parity client. 

Due to the structure of EIP functionality this integration delegates most of the work to the [implementation](https://github.com/matter-labs/eip1962):
- Implementation has a pricer that may return an error on malformed inputs or computationally insane parameters, in this case `U256::max_value()` is returned
- Implementation will do the parsing and input validation (and was fuzzy-tested), in this case error is cast into client-specific one


shamatar:
[clabot:check]

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
 	}
 }
 
+impl Implementation for EIP1962 {
+	/// Can fail in many cases, so leave it for an implementation
```
dvdplm: No need for a doc-comment here. Also: I'm not sure what "…so leave it for an implementation" means, can you clarify?

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+	fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), &'static str> {
+		let result = eth_pairings::public_interface::API::run(&input);
```
dvdplm: `eth_pairings::public_interface::API` is an odd sequence of names, why not just `eth_pairings::run`? I mean, if it's callable from the outside it's part of the api and part of the public interface so it seems tautological to repeat "api" and "public interface".

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+		if result.is_err() {
+			trace!("EIP1962 error: {:?}", result.err().expect("is error"));
```
dvdplm: Better to use `error!` here.

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+			return Err("Precompile call was unsuccessful");
```
dvdplm: Can you include the actual error in the string?

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+		}
+		let result: Vec<u8> = result.expect("some result");
```
dvdplm: All `expect`s should motivate the reason why it is safe to unwrap, e.g. "we just checked if is_err so we know it's ok; qed"

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+		if result.len() == 0 {
+			return Err("Precompile call returned no data");
+		}
+		if result.len() == 1 {
+			// manually made BE of Uint256(1)
```
dvdplm:
```suggestion
			// manually make BE of Uint256(1)
```

ethcore/builtin/src/lib.rs
```diff
@@ -548,14 +572,39 @@ impl Bn128Pairing {
...
+			let mut buf = [0u8; 32];
+			buf[31] = result[0];
+			output.write(0, &buf);
+		} else {
+			output.write(0, &result[..]);
```
dvdplm: Is this guaranteed to be of a certain length (32 bytes?)? If so, can you do `&result[..32]` here?

ethcore/builtin/src/lib.rs
```diff
@@ -990,6 +1039,110 @@ mod tests {
...
+		};
+
+		// test for returning max gas on invalid input
+		{
+			let input = FromHex::from_hex("00").unwrap();
+			let expected_cost = U256::max_value();
+			assert_eq!(f.cost(&input[..]), expected_cost.into());
+		}
+
+		// test for a valid return value for some BLS12 test vector
```
dvdplm: Can you add the source for this test vector to the comment?

ethcore/builtin/src/lib.rs
shamatar: A single byte is returned iff a pairing operation is performed, so I manually put this byte into last byte of 32 bytes buffer to simulate serialization of U256. In case of other operations returned length is not equal to one, so I write directly to the output.

ethcore/builtin/src/lib.rs
shamatar: Sure, I was not sure if it would be handy

ethcore/builtin/src/lib.rs
sorpaas:
```suggestion
		let result = eth_pairings::public_interface::API::run(&input).map_err(|e| ...insert anything here you want to display e...)?;
```

PR #10875, opened 2019-07-11
Configuration map of block reward contract addresses

vkomenda:
The original issue fixed by this PR's changeset: https://github.com/poanetwork/parity-ethereum/issues/144.

### DIFF
--- a/json/src/spec/authority_round.rs
+++ b/json/src/spec/authority_round.rs
@@ -14,7 +14,29 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Authority params deserialization.
+//! Authority Round parameter deserialization.
+//!
+//! Here is an example of input parameters where the step duration is constant at 5 seconds, the set
+//! of validators is decided by the contract at address `0x10..01` starting from block 0, and where
+//! the address of the contract that computes block rewards is set to `0x20..02` for blocks 0
+//! through 41 and to `0x30.03` for all blocks starting from block 42.
+//!
+//! ```ignore
+//! "params": {
+//!     "stepDuration": "5",
+//!     "validators": {
+//!         "multi": {
+//!             "0": {
+//!                 "contract": "0x1000000000000000000000000000000000000001"
+//!             }
+//!         }
+//!     },
+//!     "blockRewardContractTransitions": {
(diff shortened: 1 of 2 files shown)


vkomenda:
[clabot:check]

afck:
Looks like that test is also failing on master.

ordian:
please revert submodules update (ethcore/res/ethereum/tests and ethcore/res/wasm-tests)

vkomenda:
Please review.

vkomenda:
@dvdplm, thanks for your feedback! I've added tests and docs. Please check if you are happy with them.

json/src/spec/authority_round.rs
```diff
@@ -46,14 +68,19 @@ pub struct AuthorityRoundParams {
 	/// a single block reward contract transition and is compatible with the multiple address option
 	/// `block_reward_contract_transitions` below.
 	pub block_reward_contract_transition: Option<Uint>,
-	/// Block reward contract address (setting the block reward contract overrides the static block
-	/// reward definition). This option allows to add a single block reward contract address and is
-	/// compatible with the multiple address option `block_reward_contract_transitions` below.
+	/// Block reward contract address which overrides the `block_reward` setting. This option allows
```
andogro: ~~This option allows to add~~
This provides the option to add

vkomenda: I'm a bit mixed up. Can you clarify? The sentence loses its meaning after "and" if we rewrite it this way. Originally: this option allows and this option is compatible. With the change: this provides an option and this is compatible. In the latter case I'm not sure what "this" is referring to.

andogro: Just attempting to make a complete sentence here rather than change the meaning- maybe would be better to say "this option allows for a single block reward contract..." or "this option allows the user to add" or another variation that completes the sentence and preserves the meaning?

dvdplm: I think @andogro is saying that the wording "…allows to add…" is grammatically a bit off and that it'd improve the text if you reworded it with "…provides the option to add…" (or "set") here and on line 67 above. :)

vkomenda: Would the issue be fixed by changing "...allows to add..." into "...allows one to add..."?

dvdplm: Sure that works too.

PR #10877, opened 2019-07-11
Transaction propagation improvements

ordian:
Implement suggestions by @tomusdrw:
* propagates only new transactions when we get an import notification [here](https://github.com/paritytech/parity-ethereum/pull/10877/files#diff-05d235380ad5b27b1e5ee6a84511a5e8R658), but send it no more than once per 20ms

TODO
- [ ] check if we need to tweak sending transactions on `TX_TIMER` at all

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -153,7 +153,7 @@ build-android:
 
 build-linux:
   <<:                              *build-on-linux
-  only:                            *releaseable_branches
+#  only:                            *releaseable_branches
 
 build-linux-i386:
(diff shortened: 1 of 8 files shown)


grbIzl:
>disconnect from peers that send us too many transactions that are rejected after

Do we have an issue or use case for it? I've never seen such myself.

ordian:
> Do we have an issue or use case for it? I've never seen such myself.

No, not so far. I was just testing different hypotheses.

ordian:
Closing as stale for now.

PR #10880, opened 2019-07-12
Move more types out of ethcore

dvdplm:
This PR is a small step on the way to be able to extract the `Engine` trait from ethcore. The next step after this is probably to remove `Machine` from the trait but there are several tricky types that need moving/deep thought too:

- [ExecutedBlock](https://github.com/paritytech/parity-ethereum/pull/10880): embeds `State<StateDB>` which means it can't be moved to `common-types` (circular ref.)
- [Proof](https://github.com/paritytech/parity-ethereum/blob/fdc7b0fdaa386faed5a74bc61b8ffba2ac7d7f9f/ethcore/src/engines/mod.rs#L238): has a troublesome `WithState` variant which implies a dependency on `Machine`.
- [ConstructedVerifier](https://github.com/paritytech/parity-ethereum/blob/fdc7b0fdaa386faed5a74bc61b8ffba2ac7d7f9f/ethcore/src/engines/mod.rs#L246): uses a trait that needs to be extracted somewhere sensible.
- [ClientIoMessage](https://github.com/paritytech/parity-ethereum/blob/dbdb57a8c00148df1018db9e7649b12335c303c2/ethcore/src/client/io_message.rs#L25): this one is probably the trickiest to move outside of `ethcore` because of how the `Execute` variant is used.

…and probably others.

**Reviewers guide**

- common-types now has [plenty more dependencies](https://github.com/paritytech/parity-ethereum/pull/10880/files#diff-d61cfa86935fa4fae74c503844301633R8)
- common-types now owns `PreverifiedBlock` and several machine related types: `AuxiliaryData`, `AuxiliaryRequest`, `Call`, `CommonParams` (here the crate gained way more code than I'd like: something we can perhaps fix with a trait), [all error types](https://github.com/paritytech/parity-ethereum/pull/10880/files#diff-f032a0c16208db4a716f695445dbcbe3R19)
- having independent errors simplifies [error handling in `account-state`](https://github.com/paritytech/parity-ethereum/pull/10880/files#diff-f032a0c16208db4a716f695445dbcbe3L17) (and potentially others down the line)
- the `params()` function in the `Engine` trait does not provide a default implementation anymore: the idea is that future PRs will remove the dependency on `Machine` all-together.
- despite the high SNR here there are no changes to logic or other "dangerous" changes.


ethcore/types/src/engines/machine.rs
```diff
@@ -0,0 +1,46 @@
...
+	/// Needs the receipts.
+	Receipts,
+	/// Needs both body and receipts.
+	Both,
+}
+
+/// Auxiliary data fetcher for an Ethereum machine. In Ethereum-like machines
+/// there are two kinds of auxiliary data: bodies and receipts.
+#[derive(Default, Clone)]
+pub struct AuxiliaryData<'a> {
```
debris: Can we keep those data structures together with `Machine`? I don't think that putting all plain data structures in the same crate is a good idea. imo this crate should define only the basic few (header, block, blockid..) and nothing else

ethcore/types/src/engines/mod.rs
```diff
@@ -26,3 +52,33 @@ pub enum ForkChoice {
 	/// Choose the current best block.
 	Old,
 }
+
+/// Ethash-specific extensions.
+#[derive(Debug, Clone)]
+pub struct EthashExtensions {
```
debris: same here

ethcore/src/ethereum/ethash_eng.rs
```diff
@@ -23,19 +23,23 @@ use ethereum_types::{H256, H64, U256};
 use ethjson;
 use hash::{KECCAK_EMPTY_LIST_RLP};
 use rlp::Rlp;
-use types::header::Header;
-use types::BlockNumber;
+use types::{
```
debris: why `ethash_eng.rs`?

dvdplm: Honestly no good reason: my IDE was confused by a crate and module with the same name. OTOH it’s odd that we have an engine outside the engines folder so my plan was to move it anyway.

ethcore/src/client/traits.rs
```diff
@@ -25,32 +25,34 @@ use ethereum_types::{H256, U256, Address};
...
 use engines::Engine;
-use error::{Error, EthcoreResult};
-use executed::CallError;
+use EthcoreResult;
 use executive::Executed;
 use account_state::state::StateInfo;
 use trace::LocalizedTrace;
-use verification::queue::QueueInfo as BlockQueueInfo;
-use verification::queue::kind::blocks::Unverified;
+use verification::queue::kind::blocks::Unverified; // todo this is reexported from common_types
```
ordian: do you want to resolve the TODO in this PR?

ethcore/src/verification/queue/mod.rs
```diff
@@ -26,14 +26,17 @@ use parity_util_mem::{MallocSizeOf, MallocSizeOfExt};
...
 use engines::Engine;
 use client::ClientIoMessage;
 use len_caching_lock::LenCachingMutex;
+use types::errors::{BlockError, EthcoreError as Error};
 
 use self::kind::{BlockLike, Kind};
 
-pub use types::verification_queue_info::VerificationQueueInfo as QueueInfo;
+pub use types::{
+	errors::ImportError, // todo: does not need this re-export
```
ordian: same question

ethcore/types/Cargo.toml
```diff
@@ -5,15 +5,20 @@ version = "0.1.0"
...
-parity-util-mem = "0.1"
 ethkey = { path = "../../accounts/ethkey" }
 keccak-hash = "0.2.0"
 parity-bytes = "0.1"
+parity-util-mem = "0.1"
+parity-snappy = "0.1"
+patricia-trie-ethereum = { path = "../../util/patricia-trie-ethereum" }
 rlp = "0.4.0"
 rlp_derive = { path = "../../util/rlp-derive" }
+rustc-hex= "1.0" # todo: only needed for CommonParams to do from_hex on a const. Remove?
```
ordian: I would remove it

ethcore/types/Cargo.toml
```diff
@@ -5,15 +5,20 @@ version = "0.1.0"
...
 unexpected = { path = "../../util/unexpected" }
+vm = { path = "../vm"}
```
ordian: is it used somewhere?

ethcore/src/lib.rs
```diff
@@ -178,3 +176,5 @@ pub mod test_helpers;
 pub use executive::contract_address;
 pub use evm::CreateContractAddress;
 pub use trie::TrieSpec;
+/// Ethcore Result
+pub type EthcoreResult<T> = Result<T, types::errors::EthcoreError>;
```
ordian: this type is already defined in [`ethcore/src/error.rs`](https://github.com/paritytech/parity-ethereum/pull/10880/files#diff-4342081aac4ee027883e1efc84bedea4R22), right?

ethcore/types/src/engines/mod.rs
```diff
@@ -16,7 +16,33 @@
...
+	/// The engine does not seal internally.
+	External,
+}
+
+/// The number of generations back that uncles can be.
+pub const MAX_UNCLE_AGE: usize = 6;
+
+/// Default EIP-210 contract code.
+/// As defined in https://github.com/ethereum/EIPs/pull/210
+// todo: why can't I pre-hex this so we don't have to import rustc-hex?
```
ordian: why can't you? :)

ethcore/types/src/errors/block_error.rs
```diff
@@ -0,0 +1,143 @@
+use std::{
```
ordian: missing license header (and in others `*_error.rs` files), also consider removing `_error` suffix, as it's already in `error` module.

debris:
I'll rebase this pr and fix review suggestions to get it in soon ;)

ethcore/types/Cargo.toml
debris: it is :(

ethcore/types/src/engines/mod.rs
debris: resolved

ethcore/src/client/traits.rs
debris: let's do it in a separate one

ethcore/src/ethereum/denominations.rs
```diff
@@ -1,37 +0,0 @@
-// Copyright 2015-2019 Parity Technologies (UK) Ltd.
-// This file is part of Parity Ethereum.
```
ordian: was it removed as unused?

ethcore/src/engines/mod.rs
```diff
@@ -32,127 +33,39 @@ pub use self::instant_seal::{InstantSeal, InstantSealParams};
...
-
-impl error::Error for EngineError {
-	fn description(&self) -> &str {
-		"Engine error"
-	}
-}
-
 /// Seal type.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Seal {
```
ordian: Feels like `Seal` and `SealingState` should go together, no?

ethcore/types/Cargo.toml
```diff
@@ -5,15 +5,21 @@ version = "0.1.0"
...
+parity-snappy = "0.1"
+patricia-trie-ethereum = { path = "../../util/patricia-trie-ethereum" }
 rlp = "0.4.0"
 rlp_derive = { path = "../../util/rlp-derive" }
 unexpected = { path = "../../util/unexpected" }
+vm = { path = "../vm"}
 
 [dev-dependencies]
-rustc-hex= "1.0"
+rustc-hex= "1.0" # todo: only needed for CommonParams to do from_hex on a const. Remove?
```
ordian: that's fine for tests

ethcore/src/ethereum/denominations.rs
debris: yes

ethcore/types/Cargo.toml
debris: oh, yes, I should've removed the comment

ethcore/src/client/traits.rs
dvdplm: It’s fine to do in separate pr: it’s a matter of checking what uses the type outside ethcore.

ethcore/types/Cargo.toml
dvdplm: It is yes, and it’s annoying. :/

ethcore/types/src/engines/machine.rs
dvdplm: You are right and I agree that with this pr the common-types crate becomes a bit too “heavy”. We need to move some types out of the way on the path to disentangle machine from engine though but that’s not done in this pr so maybe best to undo this change here and redo it in the next pr (or do it better).

ethcore/src/lib.rs
dvdplm: I must have forgot to remove `error.rs`. My bad.

Issue #8417, opened 2018-04-17
Return appropriate error when data>300kb for own transactions


Tbaut:
Following discussions at https://github.com/paritytech/parity/issues/8301

The limitation currently is imposed on the network layer (see MAX_TRANSACTION_SIZE) and we just ignore incoming transactions larger than 300kb, probably would be good to:

- return error for own transactions exceeding that limit (additional check in verification pipeline)
- make it configurable (spec?)

debris:
@grbIzl could you look into this?

grbIzl:
Sure!

folsen:
@grbIzl Did your PR fix this issue? If so please close.

grbIzl:
Closed by #8473

PR #10887, opened 2019-07-15
fn queue_transactions accepts UnverifiedTransactions

debris:
importing transactions on a client layer is a mess, this pr tries too simplify it a bit and reduce the number of redundant transaction encoding and decoding. example scenario before this pr and after this pr

#### before

light client, relay:

- `ethcore/light/src/net/mod.rs` decodes transaction
- `ethcore/sync/src/api.rs` encodes transaction
- `ethcore/src/client/client.rs ` decodes transaction
- `miner/src/pool/verifier.rs` encodes transaction
- `ethcore/src/machine.rs` decodes transaction

#### now

light client, relay

- `ethcore/light/src/net/mod.rs` decodes transaction
- `miner/src/pool/verifier.rs` encodes transaction

it's still not perfect, but definitely better than it was before

### DIFF
--- a/ethcore/light/src/net/mod.rs
+++ b/ethcore/light/src/net/mod.rs
@@ -246,7 +246,7 @@ pub trait Handler: Send + Sync {
 	/// Called when a peer makes an announcement.
 	fn on_announcement(&self, _ctx: &dyn EventContext, _announcement: &Announcement) { }
 	/// Called when a peer requests relay of some transactions.
-	fn on_transactions(&self, _ctx: &dyn EventContext, _relay: &[UnverifiedTransaction]) { }
+	fn on_transactions(&self, _ctx: &dyn EventContext, _relay: Vec<UnverifiedTransaction>) { }
 	/// Called when a peer responds to requests.
 	/// Responses not guaranteed to contain valid data and are not yet checked against
 	/// the requests they correspond to.
@@ -1138,7 +1138,7 @@ impl LightProtocol {
 				peer,
 				io,
 				proto: self,
-			}, &txs);
+			}, txs.clone());
 		}
 
 		Ok(())
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -2217,22 +2217,22 @@ impl BlockChainClient for Client {
 }
 
 impl IoClient for Client {
-	fn queue_transactions(&self, transactions: Vec<Bytes>, peer_id: usize) {
+	fn queue_transactions(&self, transactions: Vec<UnverifiedTransaction>, peer_id: usize) {
 		trace_time!("queue_transactions");
 		let len = transactions.len();
 		self.queue_transactions.queue(&self.io_channel.read(), len, move |client| {
 			trace_time!("import_queued_transactions");
-
-			let txs: Vec<UnverifiedTransaction> = transactions
-				.iter()
-				.filter_map(|bytes| client.engine.decode_transaction(bytes).ok())
-				.collect();
-
+			let hashes: Vec<_> = transactions.iter().map(UnverifiedTransaction::hash).collect();
 			client.notify(|notify| {
-				notify.transactions_received(&txs, peer_id);
+				notify.transactions_received(&hashes, peer_id);
 			});
 
-			client.importer.miner.import_external_transactions(client, txs);
+			// TODO: queue_transactions.queue, should accept FnOnce, so it can capture outer variables
+			for res in client.importer.miner.import_external_transactions(client, transactions.clone()) {
+				if let Err(e) = res {
+					debug!(target: "client", "Import of external transaction has failed: {}", e);
+				}
+			}
 		}).unwrap_or_else(|e| {
 			debug!(target: "client", "Ignoring {} transactions: {}", len, e);
 		});
--- a/ethcore/src/engines/mod.rs
+++ b/ethcore/src/engines/mod.rs
@@ -541,9 +541,13 @@ pub trait Engine: Sync + Send {
 		self.machine().verify_transaction_basic(t, header)
 	}
 
-	/// Performs pre-validation of RLP decoded transaction before other processing
-	fn decode_transaction(&self, transaction: &[u8]) -> Result<UnverifiedTransaction, transaction::Error> {
-		self.machine().decode_transaction(transaction)
+	/// Verifies len of transaction RLP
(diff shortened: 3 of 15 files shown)


ethcore/src/client/client.rs
```diff
@@ -2217,22 +2217,22 @@ impl BlockChainClient for Client {
...
-	fn queue_transactions(&self, transactions: Vec<Bytes>, peer_id: usize) {
+	fn queue_transactions(&self, transactions: Vec<UnverifiedTransaction>, peer_id: usize) {
 		trace_time!("queue_transactions");
 		let len = transactions.len();
 		self.queue_transactions.queue(&self.io_channel.read(), len, move |client| {
 			trace_time!("import_queued_transactions");
-
-			let txs: Vec<UnverifiedTransaction> = transactions
-				.iter()
-				.filter_map(|bytes| client.engine.decode_transaction(bytes).ok())
```
debris: this verification is performed by a miner in `import_external_transactions`

ethcore/src/client/client.rs
```diff
@@ -2217,22 +2217,22 @@ impl BlockChainClient for Client {
...
-				.collect();
-
+			let hashes: Vec<_> = transactions.iter().map(UnverifiedTransaction::hash).collect();
 			client.notify(|notify| {
-				notify.transactions_received(&txs, peer_id);
+				notify.transactions_received(&hashes, peer_id);
```
debris: maybe we should move this notification after miner imports the transactions? cause those transactions may be invalid

ethcore/src/engines/mod.rs
```diff
@@ -541,9 +541,13 @@ pub trait Engine: Sync + Send {
 		self.machine().verify_transaction_basic(t, header)
 	}
 
-	/// Performs pre-validation of RLP decoded transaction before other processing
-	fn decode_transaction(&self, transaction: &[u8]) -> Result<UnverifiedTransaction, transaction::Error> {
-		self.machine().decode_transaction(transaction)
+	/// Verifies len of transaction RLP
+	fn verify_transaction_len(&self, transaction: &[u8]) -> Result<(), transaction::Error> {
+		if transaction.len() > self.params().max_transaction_size {
```
debris: `max_transaction_size` is configurable since #8417 but I don't know if this is really needed... maybe we could just remove it?

ethcore/src/client/client.rs
ordian: AFAIR this notification only updates `last_sent_hashes` of a peer, so that we don't propagate the same transaction that the peer sent us

ethcore/src/client/client.rs
```diff
@@ -2217,22 +2217,22 @@ impl BlockChainClient for Client {
...
 			});
 
-			client.importer.miner.import_external_transactions(client, txs);
+			// TODO: queue_transactions.queue, should accept FnOnce, so it can capture outer variables
```
ordian: could you open an issue?

ethcore/src/client/client.rs
ordian: So we previous ignored invalid rlps and now we return an err in `on_peer_transactions`? This seems like an improvement to me.

debris: > So we previous ignored invalid rlps and now we return an err in on_peer_transactions? 

Yes

> This seems like an improvement to me. But maybe it was done on purpose?

No, it was a regression introduced very recently

ethcore/light/src/net/mod.rs
```diff
@@ -246,7 +246,7 @@ pub trait Handler: Send + Sync {
 	/// Called when a peer makes an announcement.
 	fn on_announcement(&self, _ctx: &dyn EventContext, _announcement: &Announcement) { }
 	/// Called when a peer requests relay of some transactions.
-	fn on_transactions(&self, _ctx: &dyn EventContext, _relay: &[UnverifiedTransaction]) { }
+	fn on_transactions(&self, _ctx: &dyn EventContext, _relay: Vec<UnverifiedTransaction>) { }
```
tomusdrw: This looks like a regression. Luckily it's only a light client, so we shouldn't have many transactions but you may end up cloning a lot of data.

ethcore/src/client/client.rs
```diff
@@ -2217,22 +2217,22 @@ impl BlockChainClient for Client {
...
+			for res in client.importer.miner.import_external_transactions(client, transactions.clone()) {
```
tomusdrw: This is unacceptable, it means you clone ~4MB a second.

ethcore/src/client/client.rs
tomusdrw: I wanted to do it initially, but it requires reworking `IoWorkers`. Currently every `IoMessage` is passed to every handler. We just have one handler for `ClientIoMessage`, I guess we could maybe hack around it with some internal mutability though.

ethcore/src/engines/mod.rs
tomusdrw: No we can't just remove it. We need a way to tell users that their transaction is not being accepted to the pool and will never be propagated. We can obviously hardcode that into a miner, but it feels like a regression.

ethcore/sync/src/chain/handler.rs
```diff
@@ -673,8 +673,7 @@ impl SyncHandler {
 		trace!(target: "sync", "{:02} -> Transactions ({} entries)", peer_id, item_count);
 		let mut transactions = Vec::with_capacity(item_count);
 		for i in 0 .. item_count {
-			let rlp = r.at(i)?;
-			let tx = rlp.as_raw().to_vec();
+			let tx = r.val_at(i)?;
```
tomusdrw: The whole point of passing `Bytes` was to avoid doing any intensive work (like hashing transactions) synchronously in the network `IoHandlers`.
The whole refactoring was done after a careful profiling of issues that happens on the mainnet where `network` lock could be acquired for a very long time.

ethcore/src/client/client.rs
tomusdrw: This closure is invoked in a `IoWorker` thread pool and the whole thing was done to prevent blocking networking task for  useless stuff like decoding transactions.

tomusdrw:
Regarding light client: note that all transactions always come from the RPC and we expect there is only few of them. Additional decoding-encoding is way lower priority than full node pipeline where you can have 16MBs packets of transactions received many times a second.

debris:
I'll rework this code and open a new pr

PR #10890, opened 2019-07-15
simplify create_address_scheme

debris:
`Engine::create_address_scheme` always returns the same value


ethcore/private-tx/Cargo.toml
```diff
@@ -42,6 +42,7 @@ tiny-keccak = "1.4"
 trace = { path = "../trace" }
 transaction-pool = "2.0"
 url = "1"
+vm = { path = "../vm" }
```
ordian: FWIW we may want to extract `vm::CreateContractAddress` and `ethcore::contract_address` in a separate crate.
parity (see https://github.com/paritytech/parity-common/issues/186).

PR #10891, opened 2019-07-15
remove verify_transaction_unordered from engine

debris:
one more method removed from the engine


ethcore/src/engines/mod.rs
```diff
@@ -515,18 +515,6 @@ pub trait Engine: Sync + Send {
...
 	}
 
-	/// Verify a particular transaction is valid.
-	///
-	/// Unordered verification doesn't rely on the transaction execution order,
-	/// i.e. it should only verify stuff that doesn't assume any previous transactions
-	/// has already been verified and executed.
-	///
-	/// NOTE This function consumes an `UnverifiedTransaction` and produces `SignedTransaction`
-	/// which implies that a heavy check of the signature is performed here.
```
ordian: maybe we should keep the comment

debris:
@tomusdrw I agree! My main goal was just removing yet another verification method from an engine. I will work on it!

Issue #10671, opened 2019-05-16
Local transactions with same nonce but significantly higher gas price are rejected


marko429:
- **Parity Ethereum version**: 2.4.5
- **Operating system**: Linux
- **Installation**: Arch repo
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

Actual: When sending local transaction with a gas price, say 5 Gwei, and then a moment later another transaction with the same nonce, but significantly higher gas price 7 Gwei, Parity returns error 32010 "There are too many transactions in the queue. Your transaction was dropped due to limit. Try increasing the fee." 

Expected: local transaction with significantly higher fee and same nonce should have been accepted. This was actually the case in prior versions, as long as the gas price increase was high enough. Not sure when the regression happened, but 5 -> 7 Gwei increase would have been accepted in earlier versions.

joshua-mir:
Are you certain your transaction is being ignored due to it's gasPrice and not due to your txqueue being full for the particular sender making the transaction?

marko429:
Pretty sure. It's my own local transaction and the only one.

marko429:
I did some further testing. Downgrading to 2.3.8 seems to fix the issue.

or2008:
having the same issue on 2.5.3. downgrading to 2.3.8 did fix the problem.

ordian:
cc @ascjones

PR #10901, opened 2019-07-18
tx-pool: accept local tx with higher gas price when pool full

ascjones:
Attempt at resolving https://github.com/paritytech/parity-ethereum/issues/10671.

I believe what was happening:

1. Pool becomes full (memory or transaction limit exceeded)
2. Local tx(1) submitted with gas price *x* 
3. tx(1) accepted, replaces *worst* transaction (because it is better, local priority and/or higher gas price)
4. Local tx(2) submitted with same sender and nonce as tx(1), but a higher gas price.
5. tx(2) **rejected** (as described in #10671) because it falls foul of the logic introduced in #10526: a *ready* transaction cannot be replaced by a *non-ready* transaction. The existence of tx(1) meant that the new tx was being calculated as *not ready*, and thus rejected as it is in this case.

This fix will accept any transaction which has the same sender/nonce and a higher gas price.

### DIFF
--- a/miner/src/pool/replace.rs
+++ b/miner/src/pool/replace.rs
@@ -71,13 +71,20 @@ where
 			let old_score = (old.priority(), old.gas_price());
 			let new_score = (new.priority(), new.gas_price());
 			if new_score > old_score {
+				// check for an existing transaction with the same nonce
+				if let Some(txs) = new.pooled_by_sender {
+					if let Ok(index) = txs.binary_search_by(|old| self.scoring.compare(old, new)) {
+						return self.scoring.choose(&txs[index], new)
+					}
+				}
+
 				let state = &self.client;
 				// calculate readiness based on state nonce + pooled txs from same sender
 				let is_ready = |replace: &ReplaceTransaction<T>| {
 					let mut nonce = state.account_nonce(replace.sender());
 					if let Some(txs) = replace.pooled_by_sender {
 						for tx in txs.iter() {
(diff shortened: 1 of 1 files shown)


dvdplm:
@tomusdrw can you take a look at this please? :)

miner/src/pool/replace.rs
```diff
@@ -71,6 +71,13 @@ where
 			let old_score = (old.priority(), old.gas_price());
 			let new_score = (new.priority(), new.gas_price());
 			if new_score > old_score {
+				// check for an existing transaction with the same nonce
```
tomusdrw:
```suggestion
				// Check if this is a replacement transaction.
				// 
				// With replacement transactions we can safely return `InsertNew` here, because
				// we don't need to remove `old` (worst transaction in the pool) since `new` will replace
			    // some other transaction in the pool so we will never go above limit anyway.
```

miner/src/pool/replace.rs
```diff
@@ -71,6 +71,13 @@ where
...
+				if let Some(txs) = new.pooled_by_sender {
+					if let Ok(index) = txs.binary_search_by(|old| self.scoring.compare(old, new)) {
+						return self.scoring.choose(&txs[index], new)
```
tomusdrw: We should convert `ReplaceOld` to `InsertNew`, to prevent removing the worst transaction from the pool - since we detected that this is a replacement transaction, it will, well, replace some other transaction anyway.

PR #10902, opened 2019-07-18
Step duration map configuration parameter ported from the POA Network fork

vkomenda:
The original PR: https://github.com/poanetwork/parity-ethereum/pull/152.

Summary: Instead of a single step duration, this PR allows specifying time intervals with given step durations.

### DIFF
--- a/ethcore/types/src/errors/engine_error.rs
+++ b/ethcore/types/src/errors/engine_error.rs
@@ -67,6 +67,8 @@ pub enum EngineError {
 	CliqueTooRecentlySigned(Address),
 	/// Custom
 	Custom(String),
+	/// Configuration error.
+	Config,
 }
 
 impl fmt::Display for EngineError {
@@ -83,6 +85,7 @@ impl fmt::Display for EngineError {
 			CliqueWrongAuthorCheckpoint(ref oob) => format!("Unexpected checkpoint author: {}", oob),
 			CliqueFaultyRecoveredSigners(ref mis) => format!("Faulty recovered signers {:?}", mis),
 			CliqueTooRecentlySigned(ref address) => format!("The signer: {} has signed a block too recently", address),
+			Config => format!("Configuration error"),
 			Custom(ref s) => s.clone(),
 			DoubleVote(ref address) => format!("Author {} issued too many blocks.", address),
 			NotProposer(ref mis) => format!("Author is not a current proposer: {}", mis),
--- a/ethcore/types/src/errors/engine_error.rs
+++ b/ethcore/types/src/errors/engine_error.rs
@@ -67,8 +67,6 @@ pub enum EngineError {
 	CliqueTooRecentlySigned(Address),
 	/// Custom
 	Custom(String),
-	/// Configuration error.
-	Config,
 }
 
 impl fmt::Display for EngineError {
@@ -85,7 +83,6 @@ impl fmt::Display for EngineError {
 			CliqueWrongAuthorCheckpoint(ref oob) => format!("Unexpected checkpoint author: {}", oob),
 			CliqueFaultyRecoveredSigners(ref mis) => format!("Faulty recovered signers {:?}", mis),
 			CliqueTooRecentlySigned(ref address) => format!("The signer: {} has signed a block too recently", address),
-			Config => format!("Configuration error"),
 			Custom(ref s) => s.clone(),
 			DoubleVote(ref address) => format!("Author {} issued too many blocks.", address),
 			NotProposer(ref mis) => format!("Author is not a current proposer: {}", mis),
--- a/json/src/spec/step_duration.rs
+++ b/json/src/spec/step_duration.rs
@@ -1,0 +1,36 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Step duration configuration parameter
+
+use std::collections::BTreeMap;
+
+use serde::Deserialize;
+
+use crate::uint::Uint;
+
+/// Step duration can be specified either as a `Uint` (in seconds), in which case it will be
+/// constant, or as a list of pairs consisting of a block number and a duration, in which case the
+/// duration of a step will be determined by a mapping arising from that list.
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum StepDuration {
+	/// Duration of all steps.
+	Single(Uint),
+	/// Step duration transitions: a mapping of timestamp to step durations.
+	Transitions(BTreeMap<Uint, Uint>),
+}
--- a/json/src/spec/step_duration.rs
+++ b/json/src/spec/step_duration.rs
@@ -23,8 +23,8 @@ use serde::Deserialize;
 use crate::uint::Uint;
 
 /// Step duration can be specified either as a `Uint` (in seconds), in which case it will be
-/// constant, or as a list of pairs consisting of a block number and a duration, in which case the
-/// duration of a step will be determined by a mapping arising from that list.
+/// constant, or as a list of pairs consisting of a timestamp of type `Uint` and a duration, in
+/// which case the duration of a step will be determined by a mapping arising from that list.
 #[derive(Debug, PartialEq, Deserialize)]
 #[serde(deny_unknown_fields)]
 #[serde(untagged)]
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -192,6 +192,8 @@ struct Step {
 impl Step {
 	fn load(&self) -> u64 { self.inner.load(AtomicOrdering::SeqCst) }
 
+	/// Finds the remaining duration of the current step. Panics if there was a counter under- or
+	/// overflow.
 	fn duration_remaining(&self) -> Duration {
 		self.opt_duration_remaining().unwrap_or_else(|| {
 			let ctr = self.load();
@@ -1016,6 +1018,12 @@ impl AuthorityRound {
 	fn address(&self) -> Option<Address> {
 		self.signer.read().as_ref().map(|s| s.address() )
 	}
+
+	/// AuRa step getter for testing purposes.
+	#[allow(unused)]
+	fn permissioned_step(&self) -> &PermissionedStep {
+		self.step.as_ref()
+	}
 }
 
 fn unix_now() -> Duration {
@@ -1733,12 +1741,33 @@ impl Engine for AuthorityRound {
 	}
 }
 
+/// A helper accumulator function mapping a step duration and a step duration transition timestamp
+/// to the corresponding step number and the correct starting second of the step.
+fn next_step_time_dur(
+	prev_step: TransitionStep,
+	prev_time: TransitionTimestamp,
+	prev_dur: StepDuration,
+	time: TransitionTimestamp,
+) ->
+	Option<(TransitionStep, TransitionTimestamp)>
+{
+	let step_diff = time.checked_add(prev_dur)?
+		.checked_sub(1)?
+		.checked_sub(prev_time)?
+		.checked_div(prev_dur)?;
+	Some((
+		prev_step.checked_add(step_diff)?,
+		step_diff.checked_mul(prev_dur)?.checked_add(time)?,
+	))
+}
+
 #[cfg(test)]
 mod tests {
 	use std::collections::BTreeMap;
 	use std::str::FromStr;
 	use std::sync::Arc;
 	use std::sync::atomic::{AtomicUsize, AtomicU64, Ordering as AtomicOrdering};
+	use std::time::Duration;
 	use keccak_hash::keccak;
 	use accounts::AccountProvider;
 	use ethereum_types::{Address, H520, H256, U256};
@@ -2045,6 +2074,36 @@ mod tests {
 			durations: [(0, 0, 1)].to_vec().into_iter().collect(),
 		};
 		step.duration_remaining();
+	}
+
+	#[test]
+	fn test_change_step_duration() {
+		use super::Step;
+		use std::thread;
+
+		let now = super::unix_now().as_secs();
+		let step = Step {
+			calibrate: true,
+			inner: AtomicU64::new(::std::u64::MAX),
+			durations: [
+				(0, 0, 1),
+				(now, now, 2),
+				(now + 1, now + 2, 4)
+			].to_vec().into_iter().collect(),
+		};
+		// calibrated step `now`
+		step.calibrate();
+		let duration_remaining = step.duration_remaining();
+		assert_eq!(step.inner.load(AtomicOrdering::SeqCst), now);
+		assert!(duration_remaining <= Duration::from_secs(2));
+		thread::sleep(duration_remaining);
+		step.increment();
+		// calibrated step `now + 1`
+		step.calibrate();
+		let duration_remaining = step.duration_remaining();
+		assert_eq!(step.inner.load(AtomicOrdering::SeqCst), now + 1);
+		assert!(duration_remaining > Duration::from_secs(2));
+		assert!(duration_remaining <= Duration::from_secs(4));
 	}
 
 	#[test]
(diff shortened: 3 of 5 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -97,13 +99,26 @@ const U16_MAX: usize = ::std::u16::MAX as usize;
...
-			warn!(target: "engine", "step_duration is too high ({}), setting it to {}", step_duration_usize, U16_MAX);
-		}
+		let map_step_duration = |u: ethjson::uint::Uint| {
+			let mut step_duration_usize: usize = u.into();
+			if step_duration_usize == 0 {
+				panic!("AuthorityRoundParams: step duration cannot be 0");
+			}
+			if step_duration_usize > U16_MAX {
+				step_duration_usize = U16_MAX;
+				warn!(target: "engine", "step duration is too high ({}), setting it to {}", step_duration_usize, U16_MAX);
```
ordian: not introduced in this PR, but we should probably swap this line with the line above, otherwise we will print `U16_MAX` two times.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -97,13 +99,26 @@ const U16_MAX: usize = ::std::u16::MAX as usize;
...
+			}
+			step_duration_usize as u16
+		};
+		let step_durations: BTreeMap<u64, u16> = match p.step_duration {
+			StepDuration::Single(u) => iter::once((0, map_step_duration(u))).collect(),
+			StepDuration::Transitions(tr) => {
```
tomusdrw: Do we verify there is at least one transition?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +140,96 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
-				let ctr = self.load();
-				error!(target: "engine", "Step counter is too high: {}, aborting", ctr);
-				panic!("step counter is too high: {}", ctr)
-			},
+		let mut prev_dur = u64::from(self.durations[&0]);
+		let mut prev_step = 0u64;
+		let mut prev_time = 0u64;
+		let next_step = self.load().checked_add(1)?;
+		for (time, dur) in self.durations.iter().skip(1) {
+			let step_diff = time.checked_add(prev_dur)?.checked_sub(1)?.checked_sub(prev_time)?.checked_div(prev_dur)?;
```
tomusdrw: Can't we precompute timestamp->step conversion and store them in the `BTreeMap` instead of calculating it every time?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +140,96 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
+
 	fn duration_remaining(&self) -> Duration {
+		self.opt_duration_remaining().unwrap_or_else(|| {
+			let ctr = self.load();
+			error!(target: "engine", "Step counter under- or overflow: {}, aborting", ctr);
+			panic!("step counter under- or overflow: {}", ctr)
+		})
+	}
+
+	fn opt_duration_remaining(&self) -> Option<Duration> {
```
tomusdrw: Would be good to have a comment of what (and how) we actually try to achieve in this function. It's not super clear to me. I suspect we want to calculate the remaining duration of current step taking into account potential transitions.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +140,96 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
-			None => {
-				let ctr = self.load();
-				error!(target: "engine", "Step counter is too high: {}, aborting", ctr);
-				panic!("step counter is too high: {}", ctr)
-			},
+		let mut prev_dur = u64::from(self.durations[&0]);
+		let mut prev_step = 0u64;
+		let mut prev_time = 0u64;
+		let next_step = self.load().checked_add(1)?;
+		for (time, dur) in self.durations.iter().skip(1) {
```
tomusdrw: Is this loop supposed to find the transition timestamp and duration of current step? If so I think it could be greately simplified (and optimized) by:
1. Converting timestamps into step transitions + keeping the timestamps
2. Bin searching in the step transitions to find current which transition we fall into and just use the pre-computed starting timestamp.

Instead of bin-searching I think we should rather start from the end, as that's a most likely case that we will fall into.
So to summarize, we need:
`Vec<(StepTransition, TimestampTransition, StepDuration)>`
Then you start from the end and check if `next_step > StepTransition` if it is you use `TimestampTransition + (next_step - StepTransition) * StepDuration` to compute the expected timestamp of the next step.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +140,96 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
+			if step >= next_step {
+				break;
+			}
+			prev_step = step;
+			prev_time = step_diff.checked_mul(prev_dur)?.checked_add(prev_time)?;
+			prev_dur = u64::from(*dur);
+		}
+		let time = prev_time.checked_add(next_step.checked_sub(prev_step)?.checked_mul(prev_dur)?)?;
+		let step_end = Duration::from_secs(time);
+		if step_end > now {
```
tomusdrw: You can use `saturating_sub` instead of this `if-else`

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +140,96 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
+			}
 		}
 	}
 
+	fn opt_calibrate(&self) -> Option<()> {
+		let now = unix_now().as_secs();
+		let mut prev_dur = u64::from(self.durations[&0]);
+		let mut prev_step = 0u64;
+		let mut prev_time = 0u64;
+		for (time, dur) in self.durations.range(..now).skip(1) {
```
tomusdrw: This logic seems convoluted and duplicated, please at least refactor it to a single place.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -679,23 +736,25 @@ impl<'a, A: ?Sized, B> Deref for CowLike<'a, A, B> where B: AsRef<A> {
 impl AuthorityRound {
 	/// Create a new instance of AuthorityRound engine.
 	pub fn new(our_params: AuthorityRoundParams, machine: Machine) -> Result<Arc<Self>, Error> {
-		if our_params.step_duration == 0 {
-			error!(target: "engine", "Authority Round step duration can't be zero, aborting");
-			panic!("authority_round: step duration can't be zero")
+		if !our_params.step_durations.contains_key(&0) {
+			error!(target: "engine", "Authority Round step 0 duration is undefined, aborting");
+			panic!("authority_round: step 0 duration is undefined")
```
tomusdrw: Why do we panic instead of returning a proper error?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -59,12 +60,13 @@ mod finality;
...
-	/// Time to wait before next block or authority switching,
-	/// in seconds.
+	/// A map defining intervals of blocks with the given times (in seconds) to wait before next
+	/// block or authority switching. The keys in the map are timestamps of starting blocks of those
+	/// periods. The entry at `0` should be defined.
 	///
-	/// Deliberately typed as u16 as too high of a value leads
-	/// to slow block issuance.
-	pub step_duration: u16,
+	/// Wait times (durations) are deliberately typed as `u16` since larger values lead to slow
```
tomusdrw: We can still store `u64` to avoid casting in the functions that use it and just validate when creating the params or the engine.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -2387,7 +2448,7 @@ mod tests {
 		let (_spec, tap, accounts) = setup_empty_steps();
 		let engine = build_aura(|p| {
 			p.validators = Box::new(SimpleList::new(accounts.clone()));
-			p.step_duration = 4;
+			p.step_durations = [(0, 4)].to_vec().into_iter().collect();
```
tomusdrw: We need more tests for the actual transitions.

ethcore/src/engines/authority_round/mod.rs
vkomenda: Yes, we are already checking that the durations are in the range when receiving `AuthorityRoundParams`. I don't mind storing `u64` in the map.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +148,87 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
+		let (mut prev_dur, mut prev_step, mut prev_time) =
+			self.durations.first().expect("durations cannot be empty");
+		for (step, time, dur) in self.durations.iter().skip(1) {
+ 			if *step >= next_step {
+				break;
+			}
+			prev_step = *step;
+			prev_time = *time;
+			prev_dur = *dur;
+		}
```
afck: Maybe simpler:
```rust
let (prev_dur, prev_step, prev_time) = self.durations.iter()
    .take_while(|(_, step, _)| step < next_step)
    .last()
    .expect("durations cannot be empty")
    .clone();
```
(Also in `opt_calibrate`.)

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -125,54 +148,87 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
...
 		}
 	}
 
+	/// Calibrates the AuRa step number given according to the current time.
+	fn opt_calibrate(&self) -> Option<()> {
+		let now = unix_now().as_secs();
+		let (mut prev_step, mut prev_time, mut prev_dur) =
+			self.durations.first().expect("durations cannot be empty");
+		for (step, time, dur) in self.durations.iter().skip(1) {
+			if *time >= now {
```
afck: Should this actually be strict? If `time == now`, I imagine we consider the new step to have started?

ethcore/src/engines/authority_round/mod.rs
vkomenda: It's not a question I can answer because the panic is already there. I can change it to returning an error though.

ethcore/src/engines/authority_round/mod.rs
vkomenda: I've added a test. Please let me know if you require more.

ethcore/src/engines/authority_round/mod.rs
vkomenda: You are right. `==` works better for `Vec`.

ethcore/src/engines/authority_round/mod.rs
vkomenda: Actually, it's `(step, _, _)`, which brings the question: do you think there should be a struct instead of the triple?

afck: Ah, sorry! But then it should also be:
```rust
let (prev_step, prev_dur, prev_time) = …
```
A struct would avoid that kind of confusion, yes, but it's also a bit more verbose. Your call.

vkomenda: It's `(step, time, dur)` and I dropped `prev_` as redundant.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -906,6 +971,12 @@ impl AuthorityRound {
 	fn address(&self) -> Option<Address> {
 		self.signer.read().as_ref().map(|s| s.address() )
 	}
+
+	/// AuRa step getter for testing purposes.
+	#[allow(unused)]
```
ordian: Is it actually used in tests? Why not `#[cfg(test)]`?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -679,23 +726,41 @@ impl<'a, A: ?Sized, B> Deref for CowLike<'a, A, B> where B: AsRef<A> {
...
+			error!(target: "engine", "Authority Round step 0 duration is undefined, aborting");
+			return Err(Error::Engine(EngineError::Config));
+		}
+		if our_params.step_durations.values().any(|v| *v == 0) {
+			error!(target: "engine", "Authority Round step duration cannot be 0");
+			return Err(Error::Engine(EngineError::Config));
 		}
 		let should_timeout = our_params.start_step.is_none();
-		let initial_step = our_params.start_step.unwrap_or_else(|| (unix_now().as_secs() / (our_params.step_duration as u64)));
+		let initial_step = our_params.start_step.unwrap_or(0);
```
ordian: I've noticed the `or_else` value has changed, why is it correct?

ethcore/types/src/errors/engine_error.rs
```diff
@@ -83,6 +85,7 @@ impl fmt::Display for EngineError {
 			CliqueWrongAuthorCheckpoint(ref oob) => format!("Unexpected checkpoint author: {}", oob),
 			CliqueFaultyRecoveredSigners(ref mis) => format!("Faulty recovered signers {:?}", mis),
 			CliqueTooRecentlySigned(ref address) => format!("The signer: {} has signed a block too recently", address),
+			Config => format!("Configuration error"),
```
ordian: Not sure if this error adds any information, we could we Custom error with a message instead.

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1615,11 +1688,32 @@ impl Engine for AuthorityRound {
 	}
 }
 
+/// A helper accumulator function mapping a step duration and a step duration transition timestamp
+/// to the corresponding step number and the correct starting second of the step.
+fn next_step_time_dur(
```
ordian:
```suggestion
fn next_step_time_duration(
```

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -679,23 +726,41 @@ impl<'a, A: ?Sized, B> Deref for CowLike<'a, A, B> where B: AsRef<A> {
...
-		let initial_step = our_params.start_step.unwrap_or_else(|| (unix_now().as_secs() / (our_params.step_duration as u64)));
+		let initial_step = our_params.start_step.unwrap_or(0);
+
+		let mut durations = Vec::new();
+		let mut prev_step = 0u64;
+		let mut prev_time = 0u64;
+		let mut prev_dur = our_params.step_durations[&0];
+		durations.push((prev_step, prev_time, prev_dur));
+		for (time, dur) in our_params.step_durations.iter().skip(1) {
+			let (step, time) = next_step_time_dur(prev_step, prev_time, prev_dur, *time)
```
ordian:
```suggestion
			let (step, time) = next_step_time_duration(prev_step, prev_time, prev_dur, *time)
```

ethcore/src/engines/authority_round/mod.rs
afck: The computation of the initial step is more complicated now, so we call `calibrate()` below instead, to avoid duplicating that code.

ethcore/src/engines/authority_round/mod.rs
vkomenda: Right, it's dead code and will be removed.

vkomenda:
Are you happy with the way comments have been addressed? Is there anything else that could be improved?

json/src/spec/step_duration.rs
```diff
@@ -0,0 +1,33 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Step duration configuration parameter
+
+use std::collections::BTreeMap;
+use uint::Uint;
+
+/// Step duration can be specified either as a `Uint` (in seconds), in which case it will be
+/// constant, or as a list of pairs consisting of a block number and a duration, in which case the
```
ordian:
```suggestion
/// constant, or as a list of pairs consisting of a timestamp (Unit time) and a duration, in which case the
```

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -57,14 +58,19 @@ use unexpected::{Mismatch, OutOfBounds};
 
 mod finality;
 
+type TransitionStep = u64;
+type TransitionTimestamp = u64;
+type StepDuration = u64;
```
ordian: was it changed to `u64` to avoid casts?

vkomenda: We have triples of `u64` which needed some explanation in https://github.com/paritytech/parity-ethereum/blob/7863f8381a3790a3e62c605b6fc27c6c6b7fc5fe/ethcore/src/engines/authority_round/mod.rs#L157.

@ordian Would you prefer a struct type instead of plain triples?

ordian: @vkomenda not sure if the fields would be named like `transition_step, transition_timestamp, step_duration`, it would be any clearer, but I don't have a strong preference here.

vkomenda: > @vkomenda not sure if the fields would be named like `transition_step, transition_timestamp, step_duration`, it would be any clearer, but I don't have a strong preference here.

That might actually help to avoid confusing different fields when using the triple. I'll add that struct.

ethcore/src/engines/authority_round/mod.rs
afck: I think `Duration` has no `saturating_sub`. But we could convert `now` to seconds and back, if you prefer.

ethcore/src/engines/authority_round/mod.rs
afck: But then I wonder whether `u16::MAX` is a good upper bound. This has nothing to do with the number of bits and should just be a threshold above which step durations don't make sense anymore. `u16::MAX` corresponds to about 18 hours. Should we set the limit to 24 hours (86400 seconds)?

ethcore/src/engines/authority_round/mod.rs
afck: In fact, we should verify that there is a transition for time `0`, i.e. an initial step duration.

afck:
I rebased on `master`.

afck:
Something seems to be wrong with CI.
```
ERROR: Job failed (system failure): Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? (executor_docker.go:739:120s)
```

dvdplm:
@afck Most likely yes and caused by a docker-related change that is missing on this branch. If you merge in master it should be alright. Sorry for the hassle!

varasev:
Seems all checks now have passed. Could we merge it then?

PR #10904, opened 2019-07-19
get rid of hidden mutability of Spec

debris:
changes in this pr:
- change `Spec::state_root_memo` from `RwLock<H256>`  to `H256`
- cleanup how `SpecHardcodedSync` is being deserialized from json and printed


ethcore/src/spec/spec.rs
```diff
@@ -163,13 +250,14 @@ impl Clone for Spec {
...
 			constructors: self.constructors.clone(),
-			state_root_memo: RwLock::new(*self.state_root_memo.read()),
+			state_root_memo: self.state_root_memo,
 			genesis_state: self.genesis_state.clone(),
 		}
 	}
 }
 
 /// Part of `Spec`. Describes the hardcoded synchronization parameters.
+#[derive(Debug, Clone)]
```
debris: `Clone` replaces `impl Clone` bolierplate.
`Debug` is has replaced `to_json` which was used to print this structure

ethcore/src/spec/spec.rs
```diff
@@ -511,11 +487,18 @@ impl Spec {
...
+			&self.constructors,
+			&*self.engine,
+			self.author,
+			self.timestamp,
+			self.difficulty,
 			&Default::default(),
 			BasicBackend(journaldb::new_memory_db()),
 		)?;
 
+		self.state_root_memo = root;
```
debris: `run_constructors` no longer mutates `self.state_root_memo`, so we need to set the root manutally

ethcore/src/spec/spec.rs
```diff
@@ -226,59 +296,51 @@ fn load_from(spec_params: SpecParams, s: ethjson::spec::Spec) -> Result<Spec, Er
...
+	let timestamp = g.timestamp;
+	let difficulty = g.difficulty;
+	let constructors: Vec<_> = s.accounts
+		.constructors()
+		.into_iter()
+		.map(|(a, c)| (a.into(), c.into()))
+		.collect();
+	let genesis_state: PodState = s.accounts.into();
+
+	let (state_root_memo, _) = run_constructors(
```
debris: by running constructors always before `Spec` is created, we ensure that `spec_root_memo` is always correct

ethcore/src/spec/spec.rs
```diff
@@ -540,7 +515,20 @@ impl Spec {
...
+			&self.constructors,
+			&*self.engine,
+			self.author,
+			self.timestamp,
+			self.difficulty,
+			factories,
+			db
+		)?;
+		if root != self.state_root() {
+			return Err("Invalid state root".into());
```
debris: this should actually never occur, unless someone modifies this code and at some point overwrites `state_root_memo`. Maybe I should change it to `assert_eq!`?

json/src/spec/hardcoded_sync.rs
```diff
@@ -18,14 +18,15 @@
...
 
 /// Spec hardcoded sync.
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Deserialize)]
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "camelCase")]
 pub struct HardcodedSync {
 	/// Hexadecimal of the RLP encoding of the header of the block to start synchronization from.
-	pub header: String,
+	pub header: Bytes,
```
debris: it should be `Bytes`, not `String`. Someone did it out of laziness and, because of that we had to deal unparsed string in `ethcore`

rpc/src/v1/tests/eth.rs
```diff
@@ -68,7 +68,6 @@ fn make_spec(chain: &BlockChain) -> Spec {
 	let state = chain.pre_state.clone().into();
 	spec.set_genesis_state(state).expect("unable to set genesis state");
 	spec.overwrite_genesis_params(genesis);
-	assert!(spec.is_state_root_valid());
```
debris: it's always valid now

parity/export_hardcoded_sync.rs
```diff
@@ -96,7 +96,7 @@ pub fn execute(cmd: ExportHsyncCmd) -> Result<String, String> {
 	let hs = service.client().read_hardcoded_sync()
 		.map_err(|e| format!("Error reading hardcoded sync: {}", e))?;
 	if let Some(hs) = hs {
-		Ok(::serde_json::to_string_pretty(&hs.to_json()).expect("generated JSON is always valid"))
+		Ok(format!("{:#?}", hs))
```
debris: printing debug is as readable and pretty as printing json in this case

ordian: I guess Json is used to embed `hardcodedSync` easily into a chain spec.

ordian: (so I would rather keep the previous behavior)

debris: then it's as easy as it was before,

before: 

```
{
  "header": "0x...",
  "total_difficulty": "0x...",
  "chts": [
    "0x...",
  ]
}
```

now
```
"HardcodedSync": {
  "header": 0x...,
  "total_difficulty": 0x...,
  "chts": [
    0x...,
  ]
}
```

ordian: notice the quotes around the hashes

debris: fixed

ethcore/src/spec/spec.rs
```diff
@@ -540,7 +526,18 @@ impl Spec {
...
+			&self.constructors,
+			&*self.engine,
+			self.author,
+			self.timestamp,
+			self.difficulty,
+			factories,
+			db
+		)?;
+
+		assert_eq!(root, self.state_root(), "Spec's state root has not been precomputed correctly.");
```
tomusdrw: Is it safe / expected to panic here?

debris: yes, cause `self.state_root()` is always created during init with the same constructors

Issue #9636, opened 2018-09-24
rpc test randomly failing


debris:
job: https://gitlab.parity.io/parity/parity-ethereum/-/jobs/100436
pr: #9633

```
---- v1::tests::mocked::signing::should_check_status_of_request_when_its_resolved stdout ----
thread 'v1::tests::mocked::signing::should_check_status_of_request_when_its_resolved' panicked at 'assertion failed: `(left == right)` (left: `Some("{\"jsonrpc\":\"2.0\",\"result\":null,\"id\":1}")`, right: `Some("{\"jsonrpc\":\"2.0\",\"result\":\"0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001\",\"id\":1}")`, diff: ` Some("{\"jsonrpc\":\"2.0\",\"result\":null,\"id\":1}") Some("{\"jsonrpc\":\"2.0\",\"result\":\"0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001\",\"id\":1}") `)', rpc/src/v1/tests/mocked/signing.rs:207:2
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:221
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:475
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:345
   7: parity_rpc::v1::tests::mocked::signing::should_check_status_of_request_when_its_resolved
   8: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1451
             at /checkout/src/libcore/ops/function.rs:223
             at /checkout/src/liballoc/boxed.rs:642
   9: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:105

```

PR #10909, opened 2019-07-22
Add a 2/3 quorum option to Authority Round.

afck:
This prevents the "Attack of the Clones": https://arxiv.org/pdf/1902.10244.pdf

PR for POA's `aura-pos` branch: https://github.com/poanetwork/parity-ethereum/pull/132
Original issue: https://github.com/poanetwork/parity-ethereum/issues/108

### DIFF
--- a/ethcore/src/engines/authority_round/finality.rs
+++ b/ethcore/src/engines/authority_round/finality.rs
@@ -20,6 +20,7 @@ use std::collections::{VecDeque};
 use std::collections::hash_map::{HashMap, Entry};
 
 use ethereum_types::{H256, Address};
+use types::BlockNumber;
 
 use engines::validator_set::SimpleList;
 
@@ -30,21 +31,24 @@ pub struct UnknownValidator;
 /// Rolling finality checker for authority round consensus.
 /// Stores a chain of unfinalized hashes that can be pushed onto.
 pub struct RollingFinality {
-	headers: VecDeque<(H256, Vec<Address>)>,
+	headers: VecDeque<(H256, BlockNumber, Vec<Address>)>,
 	signers: SimpleList,
 	sign_count: HashMap<Address, usize>,
 	last_pushed: Option<H256>,
+	/// First block for which a 2/3 quorum (instead of 1/2) is required.
+	quorum_2_3_transition: BlockNumber,
 }
 
 impl RollingFinality {
 	/// Create a blank finality checker under the given validator set.
-	pub fn blank(signers: Vec<Address>) -> Self {
+	pub fn blank(signers: Vec<Address>, quorum_2_3_transition: BlockNumber) -> Self {
 		trace!(target: "finality", "Instantiating blank RollingFinality with {} signers: {:?}", signers.len(), signers);
 		RollingFinality {
 			headers: VecDeque::new(),
 			signers: SimpleList::new(signers),
 			sign_count: HashMap::new(),
 			last_pushed: None,
+			quorum_2_3_transition,
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


afck:
`cargo audit` failed because of `memoffset`, which is an indirect dependency via `crossbeam-epoch`: https://github.com/crossbeam-rs/crossbeam/pull/402

afck:
I can't reproduce that test failure locally: `cargo test` in the `rpc` directory passes for me.

ordian:
> I can't reproduce that test failure locally: `cargo test` in the `rpc` directory passes for me.

That's a known issue: https://github.com/paritytech/parity-ethereum/issues/9636, it fails spuriously. The job is restarted.

ethcore/src/engines/authority_round/finality.rs
```diff
@@ -110,97 +104,127 @@ impl RollingFinality {
...
+		}
+	}
+
+	/// Removes the signers from the sign count.
+	fn remove_signers(&mut self, signers: &[Address]) {
+		for signer in signers {
+			match self.sign_count.entry(*signer) {
+				Entry::Occupied(mut entry) => {
+					// decrement count for this signer and purge on zero.
+					*entry.get_mut() -= 1;
```
dvdplm: Maybe I'm paranoid but if we ever call `remove_signers()` with an `Occupied` entry that is `0`, then this will underflow and the signer will not be removed. Maybe you can do `if *entry.get() == 1` before this and remove it if it is (and decrement if it isn't)?

ethcore/src/engines/authority_round/finality.rs
```diff
@@ -110,97 +104,127 @@ impl RollingFinality {
...
+
+					if *entry.get() == 0 {
+						entry.remove();
+					}
+				}
+				Entry::Vacant(_) => {
+					panic!("all hashes in `header` should have entries in `sign_count` for their signers; qed");
```
dvdplm: Isn't it safe to simply do nothing here? Panicking seems a bit radical perhaps?

afck:
I addressed both comments, but I'm not sure about them:
Both instances represent serious bugs within this module (`engines::authority_round::finality`): Whenever `remove_signers` is called, those signers are expected to be present in the map, with an entry ≥ 1.

dvdplm:
@afck the diff is looking a bit odd here, not sure what might have happened but it seems like a bunch of unrelated commits got added. Makes it a bit hard to review atm, can you take a look at that please?

afck:
Sorry! Fixed now.

json/src/spec/authority_round.rs
```diff
@@ -58,6 +58,8 @@ pub struct AuthorityRoundParams {
 	pub maximum_empty_steps: Option<Uint>,
 	/// Strict validation of empty steps transition block.
 	pub strict_empty_steps_transition: Option<Uint>,
+	/// First block for which a 2/3 quorum (instead of 1/2) is required.
+	pub quorum_2_3_transition: Option<Uint>,
```
ordian: Note that it will be called `quorum23Transition` in a json spec, which looks a bit weird.
What do you think about `two_thirds_majority_transition`?

ethcore/src/engines/authority_round/finality.rs
```diff
@@ -110,97 +104,127 @@ impl RollingFinality {
...
+				Entry::Occupied(mut entry) => {
+					// decrement count for this signer and purge on zero.
+					if *entry.get() <= 1 {
+						entry.remove();
+					} else {
+						*entry.get_mut() -= 1;
+					}
+				}
+				Entry::Vacant(_) => {
+					error!(target: "engine", "Called `remove_signers` with a vacant entry. This is a bug!");
```
ordian: As a fail-fast proponent, I don't agree with @dvdplm here and think the previous behavior (panic) was better.

dvdplm: fine, let's revert to panics then. :)

json/src/spec/authority_round.rs
afck: Done.

ethcore/src/engines/authority_round/finality.rs
afck: Done.

dvdplm:
This lgtm; let's wait for @andresilva to give it a once-over and then merge it.

afck:
I rebased, and added a few lines to the module documentation.

andresilva:
Thanks for the PR and sorry for taking so long to review. It was an easy review at this point though. :)

PR #10916, opened 2019-07-24
Additional arithmetic EVM opcode benchmarks

vkomenda:
These benchmarks apply to demanding arithmetic operations, in particular to `MULMOD`. A helper function `run_code` was added in order to keep the benchmark code succinct.

### DIFF
--- a/ethcore/evm/benches/basic.rs
+++ b/ethcore/evm/benches/basic.rs
@@ -32,6 +32,7 @@ extern crate rustc_hex;
 use criterion::{Criterion, Bencher, black_box};
 use std::str::FromStr;
 use std::sync::Arc;
+use bytes::Bytes;
 use ethereum_types::{U256, Address};
 use vm::{ActionParams, Result, GasLeft, Ext};
 use vm::tests::FakeExt;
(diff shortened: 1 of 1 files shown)


ethcore/evm/benches/basic.rs
```diff
@@ -207,3 +230,188 @@ fn result(r: Result<evm::GasLeft>) -> U256 {
...
+}
+
+/// Compute mulmod(U256::MAX, U256::MAX, 1) 500 times.
+fn mulmod1_500(b: &mut Criterion) {
+	b.bench_function("mulmod modulo 1, 500 times", |b| {
+		run_code(b, "6101f45b6001900360017fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff80095080600357".from_hex().unwrap());
+	});
+}
+
+/// Compute mulmod(U256::MAX, U256::MAX, 1) 1000 times.
```
ordian: What's the point of having a loop in the bytecode with different number of iterations, just to make sure it scales linearly? IMHO we should be benchmarking a single operation and criterion can take care of running it in a loop. It would make sense to have a loop if we were using a JIT compiler, but we're not.

vkomenda: Without a loop in the smart contract you get a significant noise from loading the contract. You are right, different numbers of iterations are to check that scaling is linear.

PR #10920, opened 2019-07-25
RPC method for clearing the engine signer

vkomenda:
This PR adds a new RPC method for clearing the engine signer value.

Original PR: https://github.com/poanetwork/parity-ethereum/pull/114

See also https://github.com/poanetwork/posdao-test-setup/pull/41#issuecomment-477628212

### DIFF
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -295,6 +295,9 @@ pub trait Engine: Sync + Send {
 	/// Register a component which signs consensus messages.
 	fn set_signer(&self, _signer: Box<dyn EngineSigner>) {}
 
+	/// Unregisters the engine signer address to stop signing consensus messages.
+	fn clear_signer(&self) {}
+
 	/// Sign using the EngineSigner, to be used for consensus tx signing.
 	fn sign(&self, _hash: H256) -> Result<Signature, Error> { unimplemented!() }
 
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -293,10 +293,7 @@ pub trait Engine: Sync + Send {
 	fn handle_message(&self, _message: &[u8]) -> Result<(), EngineError> { Err(EngineError::UnexpectedMessage) }
 
 	/// Register a component which signs consensus messages.
-	fn set_signer(&self, _signer: Box<dyn EngineSigner>) {}
-
-	/// Unregisters the engine signer address to stop signing consensus messages.
-	fn clear_signer(&self) {}
+	fn set_signer(&self, _signer: Option<Box<dyn EngineSigner>>) {}
 
 	/// Sign using the EngineSigner, to be used for consensus tx signing.
 	fn sign(&self, _hash: H256) -> Result<Signature, Error> { unimplemented!() }
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -1622,12 +1622,8 @@ impl Engine for AuthorityRound {
 		self.validators.register_client(client);
 	}
 
-	fn set_signer(&self, signer: Box<dyn EngineSigner>) {
-		*self.signer.write() = Some(signer);
-	}
-
-	fn clear_signer(&self) {
-		*self.signer.write() = Default::default();
+	fn set_signer(&self, signer: Option<Box<dyn EngineSigner>>) {
+		*self.signer.write() = signer;
 	}
 
 	fn sign(&self, hash: H256) -> Result<Signature, Error> {
@@ -1763,7 +1759,7 @@ mod tests {
 		let b1 = OpenBlock::new(engine, Default::default(), false, db1, &genesis_header, last_hashes.clone(), addr1, (3141562.into(), 31415620.into()), vec![], false).unwrap();
 		let b1 = b1.close_and_lock().unwrap();
 
-		engine.set_signer(Box::new((tap.clone(), addr1, "1".into())));
+		engine.set_signer(Some(Box::new((tap.clone(), addr1, "1".into()))));
 		if let Seal::Regular(seal) = engine.generate_seal(&b1, &genesis_header) {
 			assert!(b1.clone().try_seal(engine, seal).is_ok());
 			// Second proposal is forbidden.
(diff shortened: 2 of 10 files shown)


ethcore/src/engines/authority_round/mod.rs
```diff
@@ -1721,12 +1725,68 @@ mod tests {
 
 		engine.set_signer(Box::new((tap, addr2, "2".into())));
 		if let Seal::Regular(seal) = engine.generate_seal(&b2, &genesis_header) {
+			// FIXME: This branch is unreachable because the call to `generate_seal` above always
```
niklasad1: Nice catch

I think it is better to have `if-else` or something similar just to be explicit for what is expected behavior. 

```rustc
if let Seal::Regular(seal) = engine.generate_seal(&b1, &genesis_header) {
...
} else {
   assert_eq!(1, 0, "Seal::Regular is expected here")
}
assert_eq!(engine.generate_seal(&b2, &genesis_header), Seal::None), "Engine is not stepped; thus no blocks accepted at the same time")
```

However, I'm not convinced if we need the second block for this test perhaps just remove it.

/cc @debris

ethcore/src/miner/miner.rs
```diff
@@ -908,6 +908,14 @@ impl miner::MinerService for Miner {
 		}
 	}
 
+	fn clear_author(&self) {
+		self.params.write().author = Default::default();
+		if self.engine.sealing_state() == SealingState::Ready {
+			self.sealing.lock().enabled = false;
+			self.engine.clear_signer();
```
niklasad1: I don't understand why you are only `clearing` engine signer when `SealingState::Ready`? Can you explain thanks?

ethcore/src/engines/authority_round/mod.rs
vkomenda: Yes, it depends on what the expected behaviour was supposed to be. Wouldn't the test will always fail in the `else` branch that you proposed? The last assertion after `if-else` is unreachable.

ethcore/src/miner/miner.rs
vkomenda: I agree with you it's convoluted and I'm open to suggestions.

1. Should we disable sealing unconditionally?
2. Or should we not disable sealing at all in `clear_author`, possibly requiring a separate call to `clear_signer`?

ethcore/src/engines/authority_round/mod.rs
niklasad1: yes, if we would two identical ones without stepping the engine (as it is currently) then it would fail.

But I'm more in favor of just to test one block in the test `generates_seal_and_does_not_double_propose` or simply remove it in favor of `your added test "generates_seal_iff_sealer_is_set"`. It seems redundant now..

vkomenda: I agree, `generates_seal_and_does_not_double_propose` does seem a bit redundant in view of having the same test being part of `generates_seal_iff_sealer_is_set`. I'll keep it for now and remove only the test of the second block.

ethcore/src/miner/miner.rs
vkomenda: I will change this to clearing the signer when the sealing state is other than `SealingState::External`. This will make `clear_author` consistent with `set_author`.

vkomenda:
Please review.

ethcore/engine/src/engine.rs
```diff
@@ -295,6 +295,9 @@ pub trait Engine: Sync + Send {
 	/// Register a component which signs consensus messages.
 	fn set_signer(&self, _signer: Box<dyn EngineSigner>) {}
 
+	/// Unregisters the engine signer address to stop signing consensus messages.
+	fn clear_signer(&self) {}
```
dvdplm: I wish there was a way to achieve this without burdening the `Engine` trait further. I fear it would require quite a few changes though. :/

dvdplm: What if we could use the `set_signer()` method for this instead? If the `signer` argument were an `Option<Box<dyn EngineSigner>>` couldn't we move the logic for clearing the signer to the engine impl?

ethcore/src/miner/mod.rs
```diff
@@ -133,6 +133,9 @@ pub trait MinerService : Send + Sync {
 	/// On chains where sealing is done externally (e.g. PoW) we provide only reward beneficiary.
 	fn set_author(&self, author: Author);
 
+	/// Clears the engine signer and stops signing.
+	fn clear_author(&self);
```
dvdplm: Do you think the same "trick" would work here too? I.e. avoid adding a new trait method and instead change `set_author` to take an `Option<Author>` instead?

afck: Done.
I made it a `T: Into<Option<Author>>`, so that you can pass in just an `Author`, and we don't have to update all the call sites.
Also, I added `*self.signer.write() = None` to `MinerService::set_author`, which seemed like an omission. Or am I misreading this?

afck: (And: Greetings from Solna! :slightly_smiling_face: :sweden:)

dvdplm:
@niklasad1 can you take a look at this as well please?

ethcore/engines/clique/src/lib.rs
```diff
@@ -762,9 +762,8 @@ impl Engine for Clique {
 		}
 	}
 
-	fn set_signer(&self, signer: Box<dyn EngineSigner>) {
-		trace!(target: "engine", "set_signer: {}", signer.address());
```
niklasad1: we might keep this log, I think we could have something like:

```rustc
		if let Some(signer) = signer.as_ref() {
			trace!(target: "engine", "set_signer: {:?}", signer.address());
		} else {
		  // maybe read self.signer.read() and print if the signer was set to Some(signer) before	
          trace!(target: "engine", "clear_signer");
		}

```

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1751,31 +1751,72 @@ mod tests {
...
-		let b2 = b2.close_and_lock().unwrap();
 
-		engine.set_signer(Box::new((tap.clone(), addr1, "1".into())));
+		engine.set_signer(Some(Box::new((tap.clone(), addr1, "1".into()))));
 		if let Seal::Regular(seal) = engine.generate_seal(&b1, &genesis_header) {
 			assert!(b1.clone().try_seal(engine, seal).is_ok());
 			// Second proposal is forbidden.
 			assert!(engine.generate_seal(&b1, &genesis_header) == Seal::None);
+		} else {
+			panic!("block 1 not sealed");
```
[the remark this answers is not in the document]
afck: Since we're doing quite a lot with the matched seal that might be a bit awkward in this case:
```rust
assert_matches!(engine.generate_seal(&b1, &genesis_header), Seal::Regular(seal)
    if b1.clone().try_seal(engine, seal).is_ok()
    && engine.generate_seal(&b1, &genesis_header) == Seal::None);
```

ethcore/engines/clique/src/lib.rs
niklasad1: do you mind putting it in an `if log_enabled!(target: "engine", log::Level::Trace)` block so we don't perform these operations when the log is disabled?

Thanks

niklasad1:
@afck Yeah, it's a CI issue. Restarted and it succeeded.

Issue #10908, opened 2019-07-22
Hive test failures on addition of new ethereum consensus tests


adilimroz:
- **Parity Ethereum version**: https://github.com/paritytech/parity-ethereum/tree/7f707fa524baaaccac499d69f9e452b380c31ffa

_Your issue description goes here below. Try to include **actual** vs. **expected behavior** and **steps to reproduce** the issue._

[New Tests](https://github.com/ethereum/tests/tree/develop/BlockchainTests/GeneralStateTests/stRevertTest) has been added in ethereum consensus test against which parity-ethereum client is failing. 

![Screenshot from 2019-07-22 13-43-05](https://user-images.githubusercontent.com/5220489/61616663-b728a480-ac86-11e9-819a-ca2e3847b8b9.png)
![Screenshot from 2019-07-22 13-43-19](https://user-images.githubusercontent.com/5220489/61616689-c60f5700-ac86-11e9-9a35-0d25ce2e771b.png)

PR #10923, opened 2019-07-28
new ethereum consensus tests

debris:
ref #10908

discussion explaining changes I made: https://github.com/ethereum/aleth/pull/5664

my limited knowledge of current state of executive and how tests are loaded doesn't allow me to fix all failing tests

tests which are still failing

- `RevertPrecompiledTouchExactOOG_d31g1v0_Byzantium`
- `RevertPrecompiledTouchExactOOG_d31g1v0_Constantinople`
- `RevertPrecompiledTouchExactOOG_d31g1v0_ConstantinopleFix`
- `RevertPrecompiledTouchExactOOG_d7g1v0_Byzantium` 
- `RevertPrecompiledTouchExactOOG_d7g1v0_Constantinople` 
- `RevertPrecompiledTouchExactOOG_d7g1v0_ConstantinopleFix`

----

my questions:

- what is [`"network": "EIP158"`](https://github.com/ethereum/tests/blob/a8f6d53bdd77bedcc08badb86079931c500e7f2f/BlockchainTests/GeneralStateTests/stRevertTest/RevertPrecompiledTouchExactOOG_d31g1v0.json#L896) and why do we load it from json test file `eip161_test.json`
- how is this network different to byzantium and constantinople

cc @sorpaas @tomusdrw

### DIFF
--- a/ethcore/src/executive.rs
+++ b/ethcore/src/executive.rs
@@ -346,6 +346,10 @@ impl<'a> CallCreateExecutive<'a> {
 				| Err(vm::Error::OutOfBounds)
 				| Err(vm::Error::Reverted)
 				| Ok(FinalizationResult { apply_state: false, .. }) => {
+					let ripemd = Address::from_low_u64_be(3);
+					if un_substate.touched.contains(&ripemd) {
+						substate.touched.insert(ripemd);
+					}
 					state.revert_to_checkpoint();
 			},
 			Ok(_) | Err(vm::Error::Internal(_)) => {


ethcore/src/executive.rs
```diff
@@ -346,6 +346,10 @@ impl<'a> CallCreateExecutive<'a> {
 				| Err(vm::Error::OutOfBounds)
 				| Err(vm::Error::Reverted)
 				| Ok(FinalizationResult { apply_state: false, .. }) => {
+					let ripemd = Address::from_low_u64_be(3);
```
debris: If we agree that this is a right place for this 'fix', I'll add comments and move this value to the constant

sorpaas:
@debris "EIP158" refers to the Spurious Dragon hard fork (http://eips.ethereum.org/EIPS/eip-607).

debris:
> Looks good, but like Wei, I'm not convinced about the location of the code.

This is also very confusing to me. Why should it be placed in `[cfg(test)]` if this change is required by `_Byzantium` and `_Constaninople` tests? Why are those tests altered/added now? If they fail, does it imply that we are consensus issue between parity and geth? If it does not imply that, why do we even need them? Are they named correctly? <- just my thoughts, would be good if someone helped me understand the problem deeper

ordian:
This is related: https://github.com/ethereum/go-ethereum/pull/3341/files#r89548318

debris:
> This is related: https://github.com/ethereum/go-ethereum/pull/3341/files#r89548318

If above is true, why are the tests failing and why do we need to implement anything?

dvdplm:
@debris what is the status of this? Can we move forward on it?

dvdplm:
Closing as stale.

Issue #10570, opened 2019-04-06
Expanse Constantinople and Duration Limit Reduction at block 1,860,000 (Has 10k EXP Bounty)


chrisfranko:
- **Parity Ethereum version**: 0.0.0
- **Operating system**: OSX
- **Installation**: built from source
- **Fully synchronized**: yes
- **Network**: expanse
- **Restarted**: no

Hello, at block 1,860,000 expanse will enable the Constantinople EIP's, as well as XIP3 (our new name for expips).

XIP3 is similar to XIP2 where we reduced the duration Limit to 30 seconds from 60 seconds. In XIP3 we are reducing the duration Limit again from 30 seconds to 15 seconds but this time keeping the block reward at 4 EXP. 

The have a 10k EXP bounty for this. The reason I wanted to do a bounty is because last time I attempted this on my own (im terrible at rust) and someone had to help. This shouldnt be too terribly involved.

In our GETH implementation we also added a feature to ignore chain reorgs from synced nodes that are > 50 blocks in the past and dont add any new work. Its a passive feature but if someone could help me implement that into parity it would help keep the geth and parity networks from experiencing chain splits if those conditions were to be met.

chrisfranko:
This bounty is still open.

chrisfranko:
rip

PR #10926, opened 2019-07-30
Disable unsyncable expanse chain

joshua-mir:
Until the necessary features described in #10570 are implemented, expanse appears unsyncable - should probably be removed unless someone wants to do that.

### DIFF
--- a/parity/cli/mod.rs
+++ b/parity/cli/mod.rs
@@ -301,7 +301,7 @@ usage! {
 			ARG arg_chain: (String) = "foundation", or |c: &Config| c.parity.as_ref()?.chain.clone(),
 			"--chain=[CHAIN]",
 			"Specify the blockchain type. CHAIN may be either a JSON chain specification file or ethereum, classic, poacore, volta, ewc, musicoin, ellaism, mix, callisto, morden, ropsten, kovan, rinkeby, goerli, kotti, poasokol, testnet, or dev.",
-		
+
 			ARG arg_keys_path: (String) = "$BASE/keys", or |c: &Config| c.parity.as_ref()?.keys_path.clone(),
 			"--keys-path=[PATH]",
 			"Specify the path for JSON key files to be found",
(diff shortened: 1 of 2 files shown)


ethcore/res/ethereum/expanse.json
```diff
@@ -1,95 +0,0 @@
...
-				"difficultyHardforkTransition": "0x59d9",
-				"difficultyHardforkBoundDivisor": "0x0200",
-				"bombDefuseTransition": "0x30d40",
-				"eip100bTransition": "0xC3500",
-				"metropolisDifficultyIncrementDivisor": "0x1E",
-				"difficultyBombDelays": {
-					"0xC3500": 3000000
-				},
-				"expip2Transition": "0xC3500",
-				"expip2DurationLimit": "0x1E"
```
debris: to remove it fully, we would have to also remove code using those 2 variables

joshua-mir:
Agree somewhat - maybe I  should revert everything except https://github.com/paritytech/parity-ethereum/pull/10926/commits/4b6e31f45a6aef72ba2f42bac4c0413bac78a545

dvdplm:
>     * remove expanse completely, including the code using transitions. I think we should do this only if we do not plan to bring back expanse in the future

@folsen thoughts on this? IMO we should not keep code around because "maybe someday we'll want it back" but I lack some context here.

folsen:
@dvdplm I generally agree with you and prefer to remove if we think it's unlikely it'll come back. However, I know one of the devs has updated this before and there is some bounty out there to help get this support back in place. I'm not sure how long it's been, given some adequate grace period I would then completely remove. Should it be that someone wants to bring it back, one can always look at history.

dvdplm:
Ok, let's disable it for the time being and come back and do the rest of the deed in 60 days (I'll set an alarm).

dvdplm:
@debris can we get a thumbs up here pls?

dvdplm:
@joshua-mir need a rebase.

joshua-mir:
We're seeing #10952 in CI 😅 causing a test failure. Looks random.

PR #10928, opened 2019-07-30
Set the block gas limit to the value returned by a contract call

vkomenda:
This PR allows to have programmable block gas limit.

See also:
- original PR: https://github.com/poanetwork/parity-ethereum/pull/140
- original issue https://github.com/poanetwork/parity-ethereum/issues/119

### DIFF
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -347,6 +347,12 @@ pub trait Engine: Sync + Send {
 		Ok(*header.author())
 	}
 
+	/// Overrides the block gas limit. Whenever this returns `Some` for a header, the next block's gas limit must be
+	/// exactly that value.
+	fn gas_limit_override(&self, _header: &Header) -> Option<U256> {
+		None
+	}
+
 	/// Get the general parameters of the chain.
 	fn params(&self) -> &CommonParams;
 
@@ -396,6 +402,11 @@ pub trait Engine: Sync + Send {
 	/// Performs pre-validation of RLP decoded transaction before other processing
 	fn decode_transaction(&self, transaction: &[u8]) -> Result<UnverifiedTransaction, transaction::Error> {
 		self.machine().decode_transaction(transaction)
+	}
+
+	/// The configured minimum gas limit.
+	fn min_gas_limit(&self) -> U256 {
+		self.params().min_gas_limit
 	}
 }
 
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -52,6 +52,7 @@ use machine::{
 use macros::map;
 use keccak_hash::keccak;
 use log::{info, debug, error, trace, warn};
+use lru_cache::LruCache;
 use engine::signer::EngineSigner;
 use parity_crypto::publickey::Signature;
 use io::{IoContext, IoHandler, TimerToken, IoService};
@@ -130,6 +131,9 @@ pub struct AuthorityRoundParams {
 }
 
 const U16_MAX: usize = ::std::u16::MAX as usize;
+
+/// The number of recent block hashes for which the gas limit override is memoized.
+const GAS_LIMIT_OVERRIDE_CACHE_CAPACITY: usize = 10;
 
 impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 	fn from(p: ethjson::spec::AuthorityRoundParams) -> Self {
@@ -577,6 +581,8 @@ pub struct AuthorityRound {
 	randomness_contract_address: BTreeMap<u64, Address>,
 	/// The addresses of contracts that determine the block gas limit.
 	block_gas_limit_contract_transitions: BTreeMap<u64, Address>,
+	/// Memoized gas limit overrides, by block hash.
+	gas_limit_override_cache: Mutex<LruCache<H256, Option<U256>>>,
 }
 
 // header-chain validator.
@@ -880,6 +886,7 @@ impl AuthorityRound {
 				received_step_hashes: RwLock::new(Default::default()),
 				randomness_contract_address: our_params.randomness_contract_address,
 				block_gas_limit_contract_transitions: our_params.block_gas_limit_contract_transitions,
+				gas_limit_override_cache: Mutex::new(LruCache::new(GAS_LIMIT_OVERRIDE_CACHE_CAPACITY)),
 			});
 
 		// Do not initialize timeouts for tests.
@@ -1872,7 +1879,12 @@ impl Engine for AuthorityRound {
 				return None;
 			}
 		};
-		block_gas_limit(full_client, header, address)
+		if let Some(limit) = self.gas_limit_override_cache.lock().get_mut(&header.hash()) {
+			return *limit;
+		}
+		let limit = block_gas_limit(full_client, header, address);
+		self.gas_limit_override_cache.lock().insert(header.hash(), limit);
+		limit
 	}
 }
 
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -42,6 +42,7 @@ use std::u64;
 
 use client_traits::{EngineClient, ForceUpdateSealing, TransactionRequest};
 use engine::{Engine, ConstructedVerifier};
+use block_gas_limit::block_gas_limit;
 use block_reward::{self, BlockRewardContract, RewardKind};
 use ethjson;
 use machine::{
@@ -78,7 +79,6 @@ use common_types::{
 	transaction::SignedTransaction,
 };
 use unexpected::{Mismatch, OutOfBounds};
-
 use validator_set::{ValidatorSet, SimpleList, new_validator_set};
 
 mod finality;
@@ -124,6 +124,9 @@ pub struct AuthorityRoundParams {
 	pub strict_empty_steps_transition: u64,
 	/// If set, enables random number contract integration. It maps the transition block to the contract address.
 	pub randomness_contract_address: BTreeMap<u64, Address>,
+	/// The addresses of contracts that determine the block gas limit with their associated block
+	/// numbers.
+	pub block_gas_limit_contract_transitions: BTreeMap<u64, Address>,
 }
 
 const U16_MAX: usize = ::std::u16::MAX as usize;
@@ -180,6 +183,12 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 				(block.as_u64(), addr.into())
 			}).collect()
 		});
+		let block_gas_limit_contract_transitions: BTreeMap<_, _> =
+			p.block_gas_limit_contract_transitions
+			.unwrap_or_default()
+			.into_iter()
+			.map(|(block_num, address)| (block_num.into(), address.into()))
+			.collect();
 		AuthorityRoundParams {
 			step_durations,
 			validators: new_validator_set(p.validators),
@@ -196,6 +205,7 @@ impl From<ethjson::spec::AuthorityRoundParams> for AuthorityRoundParams {
 			two_thirds_majority_transition: p.two_thirds_majority_transition.map_or_else(BlockNumber::max_value, Into::into),
 			strict_empty_steps_transition: p.strict_empty_steps_transition.map_or(0, Into::into),
 			randomness_contract_address,
+			block_gas_limit_contract_transitions,
 		}
 	}
 }
@@ -565,6 +575,8 @@ pub struct AuthorityRound {
 	received_step_hashes: RwLock<BTreeMap<(u64, Address), H256>>,
 	/// If set, enables random number contract integration. It maps the transition block to the contract address.
 	randomness_contract_address: BTreeMap<u64, Address>,
+	/// The addresses of contracts that determine the block gas limit.
+	block_gas_limit_contract_transitions: BTreeMap<u64, Address>,
 }
 
 // header-chain validator.
@@ -867,6 +879,7 @@ impl AuthorityRound {
 				machine,
 				received_step_hashes: RwLock::new(Default::default()),
 				randomness_contract_address: our_params.randomness_contract_address,
+				block_gas_limit_contract_transitions: our_params.block_gas_limit_contract_transitions,
 			});
 
 		// Do not initialize timeouts for tests.
@@ -1218,6 +1231,14 @@ impl Engine for AuthorityRound {
 
 		let score = calculate_score(parent_step, current_step, current_empty_steps_len);
 		header.set_difficulty(score);
+		if let Some(gas_limit) = self.gas_limit_override(header) {
+			trace!(target: "engine", "Setting gas limit to {} for block {}.", gas_limit, header.number());
+			let parent_gas_limit = *parent.gas_limit();
+			header.set_gas_limit(gas_limit);
+			if parent_gas_limit != gas_limit {
+				info!(target: "engine", "Block gas limit was changed from {} to {}.", parent_gas_limit, gas_limit);
+			}
+		}
 	}
 
 	fn sealing_state(&self) -> SealingState {
@@ -1834,6 +1855,25 @@ impl Engine for AuthorityRound {
 	fn params(&self) -> &CommonParams {
 		self.machine.params()
 	}
+
+	fn gas_limit_override(&self, header: &Header) -> Option<U256> {
+		let (_, &address) = self.block_gas_limit_contract_transitions.range(..=header.number()).last()?;
+		let client = match self.client.read().as_ref().and_then(|weak| weak.upgrade()) {
+			Some(client) => client,
+			None => {
+				error!(target: "engine", "Unable to prepare block: missing client ref.");
+				return None;
+			}
+		};
+		let full_client = match client.as_full_client() {
+			Some(full_client) => full_client,
+			None => {
+				error!(target: "engine", "Failed to upgrade to BlockchainClient.");
+				return None;
+			}
+		};
+		block_gas_limit(full_client, header, address)
+	}
 }
 
 /// A helper accumulator function mapping a step duration and a step duration transition timestamp
@@ -1910,6 +1950,7 @@ mod tests {
 			strict_empty_steps_transition: 0,
 			two_thirds_majority_transition: 0,
 			randomness_contract_address: BTreeMap::new(),
+			block_gas_limit_contract_transitions: BTreeMap::new(),
 		};
 
 		// mutate aura params
--- a/ethcore/verification/src/verification.rs
+++ b/ethcore/verification/src/verification.rs
@@ -61,6 +61,14 @@ pub fn verify_block_basic(block: &Unverified, engine: &dyn Engine, check_seal: b
 		}
 	}
 
+	if let Some(gas_limit) = engine.gas_limit_override(&block.header) {
+		if *block.header.gas_limit() != gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(
+				OutOfBounds { min: Some(gas_limit), max: Some(gas_limit), found: *block.header.gas_limit() }
+			)));
+		}
+	}
+
 	for t in &block.transactions {
 		engine.verify_transaction_basic(t, &block.header)?;
 	}
@@ -284,6 +292,25 @@ pub(crate) fn verify_header_params(header: &Header, engine: &dyn Engine, check_s
 			found: *header.gas_used()
 		})));
 	}
+	if engine.gas_limit_override(header).is_none() {
+		let min_gas_limit = engine.min_gas_limit();
+		if header.gas_limit() < &min_gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
+				min: Some(min_gas_limit),
+				max: None,
+				found: *header.gas_limit()
+			})));
+		}
+		if let Some(limit) = engine.maximum_gas_limit() {
+			if header.gas_limit() > &limit {
+				return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
+					min: None,
+					max: Some(limit),
+					found: *header.gas_limit()
+				})));
+			}
+		}
+	}
 	let maximum_extra_data_size = engine.maximum_extra_data_size();
 	if header.number() != 0 && header.extra_data().len() > maximum_extra_data_size {
 		return Err(From::from(BlockError::ExtraDataOutOfBounds(OutOfBounds {
@@ -341,8 +368,6 @@ fn verify_parent(header: &Header, parent: &Header, engine: &dyn Engine) -> Resul
 	assert!(header.parent_hash().is_zero() || &parent.hash() == header.parent_hash(),
 			"Parent hash should already have been verified; qed");
 
-	let gas_limit_divisor = engine.params().gas_limit_bound_divisor;
-
 	if !engine.is_timestamp_valid(header.timestamp(), parent.timestamp()) {
 		let now = SystemTime::now();
 		let min = CheckedSystemTime::checked_add(now, Duration::from_secs(parent.timestamp().saturating_add(1)))
@@ -365,31 +390,8 @@ fn verify_parent(header: &Header, parent: &Header, engine: &dyn Engine) -> Resul
 			found: header.number()
 		}).into());
 	}
-	if let Some(gas_limit) = engine.gas_limit_override(header) {
-		if *header.gas_limit() != gas_limit {
-			return Err(BlockError::InvalidGasLimit(
-				OutOfBounds { min: Some(gas_limit), max: Some(gas_limit), found: *header.gas_limit() }
-			).into());
-		}
-	} else {
-		let min_gas_limit = engine.params().min_gas_limit;
-		if header.gas_limit() < &min_gas_limit {
-			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
-				min: Some(min_gas_limit),
-				max: None,
-				found: *header.gas_limit()
-			})));
-		}
-		if let Some(limit) = engine.maximum_gas_limit() {
-			if header.gas_limit() > &limit {
-				return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
-					min: None,
-					max: Some(limit),
-					found: *header.gas_limit()
-				})));
-			}
-		}
-
+	if engine.gas_limit_override(header).is_none() {
+		let gas_limit_divisor = engine.params().gas_limit_bound_divisor;
 		let parent_gas_limit = *parent.gas_limit();
 		let min_gas = parent_gas_limit - parent_gas_limit / gas_limit_divisor;
 		let max_gas = parent_gas_limit + parent_gas_limit / gas_limit_divisor;
@@ -658,9 +660,8 @@ mod tests {
 		header.set_uncles_hash(good_uncles_hash.clone());
 		check_ok(basic_test(&create_test_block_with_data(&header, &good_transactions, &good_uncles), engine));
 
-		header = good.clone();
 		header.set_gas_limit(min_gas_limit - 1);
-		check_fail(family_test(&create_test_block(&header), engine, &bc),
+		check_fail(basic_test(&create_test_block(&header), engine),
 			InvalidGasLimit(OutOfBounds { min: Some(min_gas_limit), max: None, found: header.gas_limit().clone() }));
 
 		header = good.clone();
--- a/ethcore/verification/src/verification.rs
+++ b/ethcore/verification/src/verification.rs
@@ -284,23 +284,6 @@ pub(crate) fn verify_header_params(header: &Header, engine: &dyn Engine, check_s
 			found: *header.gas_used()
 		})));
 	}
-	let min_gas_limit = engine.params().min_gas_limit;
-	if header.gas_limit() < &min_gas_limit {
-		return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
-			min: Some(min_gas_limit),
-			max: None,
-			found: *header.gas_limit()
-		})));
-	}
-	if let Some(limit) = engine.maximum_gas_limit() {
-		if header.gas_limit() > &limit {
-			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds {
-				min: None,
-				max: Some(limit),
-				found: *header.gas_limit()
-			})));
-		}
-	}
 	let maximum_extra_data_size = engine.maximum_extra_data_size();
 	if header.number() != 0 && header.extra_data().len() > maximum_extra_data_size {
 		return Err(From::from(BlockError::ExtraDataOutOfBounds(OutOfBounds {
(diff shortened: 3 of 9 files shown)


vkomenda:
I've rebased the PR. Please review :)

dvdplm:
@vkomenda sorry about the churn on master recently. The test failure seem to be a simple import fix (`Header` from `common-types`).

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1658,6 +1669,36 @@ impl Engine for AuthorityRound {
...
 		self.machine.params()
 	}
+
+	fn gas_limit_override(&self, header: &Header) -> Option<U256> {
+		let (_, &address) = self.machine.params().block_gas_limit_contract.range(..=header.number()).last()?;
+
+		let client = match self.client.read().as_ref().and_then(|weak| weak.upgrade()) {
+			Some(client) => client,
+			None => {
+				debug!(target: "engine", "Unable to prepare block: missing client ref.");
```
dvdplm: Should this be `error!`?

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1658,6 +1669,36 @@ impl Engine for AuthorityRound {
...
+				return None;
+			}
+		};
+		let full_client = match client.as_full_client() {
+			Some(full_client) => full_client,
+			None => {
+				debug!(target: "engine", "Failed to upgrade to BlockchainClient.");
```
dvdplm: Should this be `error!`?

ethcore/src/verification/verification.rs
```diff
@@ -264,13 +264,21 @@ pub fn verify_header_params(header: &Header, engine: &dyn Engine, is_full: bool,
...
+			)));
+		}
+	} else {
+		let min_gas_limit = engine.params().min_gas_limit;
+		if header.gas_limit() < &min_gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: Some(min_gas_limit), max: None, found: *header.gas_limit() })));
+		}
+		if let Some(limit) = engine.maximum_gas_limit() {
+			if header.gas_limit() > &limit {
+				return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: None, max: Some(limit), found: *header.gas_limit() })));
```
dvdplm: Like above.

ethcore/src/verification/verification.rs
```diff
@@ -264,13 +264,21 @@ pub fn verify_header_params(header: &Header, engine: &dyn Engine, is_full: bool,
...
+	if let Some(gas_limit) = engine.gas_limit_override(header) {
+		if *header.gas_limit() != gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(
+				OutOfBounds { min: Some(gas_limit), max: Some(gas_limit), found: *header.gas_limit() }
+			)));
+		}
+	} else {
+		let min_gas_limit = engine.params().min_gas_limit;
+		if header.gas_limit() < &min_gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: Some(min_gas_limit), max: None, found: *header.gas_limit() })));
```
dvdplm: Long line here. Also, if `From` is impld for `BlockError`, shouldn't the conversion be more ergonomic than this?

json/src/spec/params.rs
```diff
@@ -130,12 +135,36 @@ pub struct Params {
 	pub kip6_transition: Option<Uint>,
 }
 
+/// A contract that determines block gas limits can be configured either as a single address, or as a map, assigning
+/// to each starting block number the contract address that should be used from that block on.
```
dvdplm: Maybe a dumb q but wouldn't the code be a bit simpler if the "single address" case was just a map with a single entry?

ethcore/call-contract/src/call_contract.rs
```diff
@@ -18,12 +18,18 @@
...
 use ethereum_types::Address;
+use types::header::Header;
 use types::ids::BlockId;
 
 /// Provides `call_contract` method
 pub trait CallContract {
 	/// Like `call`, but with various defaults. Designed to be used for calling contracts.
 	fn call_contract(&self, id: BlockId, address: Address, data: Bytes) -> Result<Bytes, String>;
+
+	/// Makes a constant call to a contract, at the beginning of the block corresponding to the given header, i.e. with
```
dvdplm: Maybe dumb q: what is "constant" about the call? Isn't it called for every block of the override is in effect?

ethcore/src/client/client.rs
```diff
@@ -1511,6 +1511,49 @@ impl CallContract for Client {
...
+			error!(target: "client", "Failed to call contract at {:?}'s child", header.parent_hash());
+			format!("Failed to call contract at {:?}'s child", header.parent_hash())
+		})?;
+		let root = parent.state_root();
+		let nonce = self.engine.account_start_nonce(header.number());
+		let mut state = State::from_existing(db, root, nonce, self.factories.clone())
+			.map_err(|_| CallError::StatePruned.to_string())?;
+
+		let from = Address::default();
+		let transaction = transaction::Transaction {
```
dvdplm: I think this merits a comment: looks like it's a "fake" tx used to satisfy the type signature of `do_virtual_call`? Maybe it can be extracted to its own method to make the intent clear?

ethcore/src/verification/verification.rs
```diff
@@ -264,13 +264,21 @@ pub fn verify_header_params(header: &Header, engine: &dyn Engine, is_full: bool,
...
-		if header.gas_limit() > &limit {
-			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: None, max: Some(limit), found: *header.gas_limit() })));
+	if let Some(gas_limit) = engine.gas_limit_override(header) {
+		if *header.gas_limit() != gas_limit {
+			return Err(From::from(BlockError::InvalidGasLimit(
+				OutOfBounds { min: Some(gas_limit), max: Some(gas_limit), found: *header.gas_limit() }
+			)));
+		}
+	} else {
+		let min_gas_limit = engine.params().min_gas_limit;
```
dvdplm: Seems like there should be an `engine.min_gas_limit()` method (or perhaps `engine.gas_limits() -> (u64, u64)`?).

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -63,8 +63,11 @@ use common_types::{
 	snapshot::Snapshotting,
 };
 use unexpected::{Mismatch, OutOfBounds};
-
 use validator_set::{ValidatorSet, SimpleList, new_validator_set};
+use ethabi::FunctionOutputDecoder;
+use ethabi_contract::use_contract;
+
+use_contract!(block_gas_limit, "../../res/contracts/block_gas_limit.json");
```
tomusdrw: I think this should rather go to a separate crate that only contains this contract to avoid polluting the engine.

ethcore/src/client/client.rs
```diff
@@ -1511,6 +1511,49 @@ impl CallContract for Client {
 			.map_err(|e| format!("{:?}", e))
 			.map(|executed| executed.output)
 	}
+
+	fn call_contract_before(&self, header: &Header, address: Address, data: Bytes) -> Result<Bytes, String> {
```
tomusdrw: Why is that even needed? Why can't use use `call_contract` and pass `BlockId` with `parent_hash`?

ethcore/src/verification/verification.rs
```diff
@@ -264,13 +264,21 @@ pub fn verify_header_params(header: &Header, engine: &dyn Engine, is_full: bool,
...
 		return Err(From::from(BlockError::TooMuchGasUsed(OutOfBounds { max: Some(*header.gas_limit()), min: None, found: *header.gas_used() })));
 	}
-	let min_gas_limit = engine.params().min_gas_limit;
-	if header.gas_limit() < &min_gas_limit {
-		return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: Some(min_gas_limit), max: None, found: *header.gas_limit() })));
-	}
-	if let Some(limit) = engine.maximum_gas_limit() {
-		if header.gas_limit() > &limit {
-			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: None, max: Some(limit), found: *header.gas_limit() })));
+	if let Some(gas_limit) = engine.gas_limit_override(header) {
```
tomusdrw: I'm pretty sure it won't really work correctly. What guarantees do we have that `parent_hash` is already imported? State-touching checks are only allowed further down in verification afair.

ethcore/src/verification/verification.rs
```diff
@@ -329,11 +335,14 @@ fn verify_parent(header: &Header, parent: &Header, engine: &dyn Engine) -> Resul
 		return Err(BlockError::RidiculousNumber(OutOfBounds { min: Some(1), max: None, found: header.number() }).into());
 	}
 
-	let parent_gas_limit = *parent.gas_limit();
-	let min_gas = parent_gas_limit - parent_gas_limit / gas_limit_divisor;
-	let max_gas = parent_gas_limit + parent_gas_limit / gas_limit_divisor;
-	if header.gas_limit() <= &min_gas || header.gas_limit() >= &max_gas {
-		return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: Some(min_gas), max: Some(max_gas), found: *header.gas_limit() })));
+	if engine.gas_limit_override(header).is_none() {
```
tomusdrw: That's pretty wasteful to call the contract twice during verification (one for `verify_header` and second time here), why not move the other gas_limit verification in here?

json/src/spec/params.rs
vkomenda: Not a dumb question at all! I don't know of a sensible way to allow the user to choose from two configuration options: a single address and a map. If we simply remove the single address option, you would always need to define a map.

ethcore/src/verification/verification.rs
vkomenda: I'll add a method although it's redundant because the field `min_gas_limit` is public.

vkomenda:
> @vkomenda sorry about the churn on master recently. The test failure seem to be a simple import fix (`Header` from `common-types`).

@dvdplm, which module has the error? I couldn't find it with `cargo test`.

ethcore/src/client/client.rs
vkomenda: @dvdplm Well, this is interesting. This function has nothing to do with the PR. It is from the stable  branch: https://github.com/paritytech/parity-ethereum/blob/bf74eb5cc891f6c8820aaee8ad11ac6029c26ab6/ethcore/src/client/client.rs#L1246.

Would you like to refactor it?

json/src/spec/params.rs
dvdplm: > you would always need to define a map.

I think that is fine. I think simpler code is worth a little extra user-hassle because runtime bugs are very expensive to fix while a config problem is easy to fix and easy to document?

dvdplm:
> which module has the error? I couldn't find it with `cargo test`.

oh I bet you don't see the CI output right? It's here:

```
error[E0412]: cannot find type `Header` in this scope
   --> ethcore/private-tx/src/key_server_keys.rs:164:44
    |
164 |         fn call_contract_before(&self, _header: &Header, _address: Address, _data: Bytes) -> Result<Bytes, String> {
    |                                                  ^^^^^^ not found in this scope
help: possible candidates are found in other modules, you can import them into scope
    |
140 |     use ethjson::blockchain::Header;
    |
140 |     use ethjson::blockchain::header::Header;
    |
140 |     use types::encoded::Header;
    |
140 |     use types::header::Header;
    |

For more information about this error, try `rustc --explain E0412`.
error: Could not compile `ethcore-private-tx`.
```

FYI you might want to run `cargo test --all --release` in the repo root to make sure you run all tests. :)

ethcore/src/client/client.rs
vkomenda: We can definitely use `call` later on in this function. I'm not sure about replacing `call_contract_before` with `call_contract` because `state` is defined differently in those two functions.

vkomenda: It looks like the use of `mut state` in `call_contract_before` is incorrect because it is then dropped. So whatever changes `call` makes to it are forgotten.

@afck, what are your thoughts on using `call_contract` instead of `call_contract_before` (which should fix the use of mutable state)?

json/src/spec/params.rs
vkomenda: @varasev, do we currently use a single block gas limit contract config option? If we don't, can we remove its support from the code?

varasev: @vkomenda no, we don't. It's enough to only have a map.

ethcore/src/verification/verification.rs
dvdplm: `min_gas_limit` being public is… not that great! There's probably some historical reason for that.

The point I was trying to make was this: I'm a bit hesitant about this PR because it adds code that is useful only to Aura. Ideally we should architect things such that engine-specific behaviour stays in that engine's code. One idea I had was to have a single `gas_limits()` method on `Engine` which would let the Aura implementation to run the logic it needs without "polluting" other engines.

vkomenda: This can be done, I think.
@varasev, is it OK to move `blockGasLimitContract` from the general engine parameter set to the AuRa parameter set? I would also recommend renaming it as `blockGasLimitContracts` since we are changing it to support only a map.

ethcore/src/client/client.rs
afck: You're probably right: We should just pass in the block number into `block_gas_limit` as a parameter.

Originally this was motivated by a smart contract (written before this Rust code) that had a [`limitBlockGas` getter](https://github.com/poanetwork/posdao-contracts/blob/3b5db5f74082da96c161dec0010105078d2f4b81/contracts/TxPermission.sol#L181) that returned whether the gas limit should be reduced for the _current_ block. So we needed a (["not exactly sane"](https://github.com/poanetwork/parity-ethereum/pull/140#issue-284967844)) way to call this with a state that didn't exist yet: where the block number is the number of the block that we're currently creating. The [new `blockGasLimit` implementation](https://github.com/poanetwork/posdao-contracts/blob/2f3449b765494af346503a3de52f0fdfbe425eef/contracts/TxPermission.sol#L223) has a similar problem: It calls a few methods that use the current block number.
@varasev: Would it be feasible to just pass the block number into `blockGasLimit` as an argument instead?

> It looks like the use of `mut state` in `call_contract_before` is incorrect because it is then dropped.

[`Client::call_contract`](https://github.com/paritytech/parity-ethereum/blob/3420c2bac08353f4269b9589decf8a9351275e64/ethcore/src/client/client.rs#L1401) drops the state as well: It ends up [via `call`](https://github.com/paritytech/parity-ethereum/blob/3420c2bac08353f4269b9589decf8a9351275e64/ethcore/src/client/client.rs#L1482) in [`do_virtual_call`](https://github.com/paritytech/parity-ethereum/blob/3420c2bac08353f4269b9589decf8a9351275e64/ethcore/src/client/client.rs#L1169), where it is used to [create an `Executive` instance](https://github.com/paritytech/parity-ethereum/blob/3420c2bac08353f4269b9589decf8a9351275e64/ethcore/src/client/client.rs#L1194) and a diff. It's not persisted, however, and that's intended: These methods are meant to run calls _locally_ and return the call's output. They don't create transactions and thus can't modify the actual EVM state.

ethcore/call-contract/src/call_contract.rs
afck: It's "constant" in the sense that it doesn't create a transaction and modify the blockchain's actual state. Not sure what the right word for this is.

ethcore/src/verification/verification.rs
varasev: Yes, it's OK to move, but don't rename, please, because the map still assumes that there is only one contract at a time.

If you move this parameter, please also move it in our https://github.com/poanetwork/parity-ethereum/tree/aura-pos branch.

vkomenda: @varasev OK, I'll copy that in `aura-pos`. Are you sure about the name? Wouldn't plural look better alongside `blockRewardContractTransitions`?

ethcore/src/client/client.rs
varasev: > Would it be feasible to just pass the block number into `blockGasLimit` as an argument instead?

No, the `blockGasLimit` calls other getters of other contracts to determine the gas limit for the current block. The returned values of those getters depend on the current state of the current block.

ethcore/src/verification/verification.rs
varasev: Then that should be named `blockGasLimitContractTransitions`

vkomenda: Makes sense.

vkomenda: Done in https://github.com/poanetwork/parity-ethereum/pull/189.

ethcore/call-contract/src/call_contract.rs
vkomenda: Maybe "stateless" or "immutable"? But technically the contract ABI can allow a contract call that modifies the state. I think assuming that the call modifies the state would be more complete.

afck: Not sure what the best wording is. In [this blog post](https://blog.b9lab.com/calls-vs-transactions-in-ethereum-smart-contracts-62d6b17d0bc2?gi=3fb53ba75e85) it's called "local" and "read-only call", as opposed to "verified" and "potentially state-changing transaction".

(To clarify: This _can_ be used with calls that _would_ modify the state. It's just that it throws away the modified state instead of creating a transaction which, when mined, would actually change the blockchain's EVM state.)

afck:
The test was failing because it expected `verify_block_basic` to check the gas limit, which we moved to `verify_block_family` now. (And because of a missing `header = good.clone();`.)
Is it okay for `verify_block_basic` to not check the gas limit anymore?

Otherwise we'd have to add something like `fn use_gas_limit_contract(&self, block: &BlockNumber) -> bool` to `Engine` that doesn't call the contract and only looks into the configuration to find out whether it _would_ use the contract at that block number, and check the gas limit in `verify_header_params` (used in `verify_block_basic`) only if `use_gas_limit_contract` returns `false`.

@varasev: The meaning of `blockGasLimit` has changed now. It is supposed to return the _next_ block's gas limit. Is that practical in `posdao-contracts`?

varasev:
> The meaning of blockGasLimit has changed now. It is supposed to return the next block's gas limit. Is that practical in posdao-contracts?

Ok

varasev:
> The meaning of `blockGasLimit` has changed now. It is supposed to return the next block's gas limit. Is that practical in `posdao-contracts`?

What about block `#1`? As far as I understand, the engine will call `blockGasLimit` to get the gas limit for the next block. It means that when it is called on the block `#1`, the `blockGasLimit` will return a limit for block `#2`. What limit will be for block #1 then?

afck:
It will actually be called "after the genesis block", I think. So it will be called at the point where the `block.number` is still 0 and the EVM state is the state between blocks 0 and 1.

ethcore/machine/src/machine.rs
```diff
@@ -96,6 +96,11 @@ impl Machine {
 		self.ethash_extensions.as_ref()
 	}
 
+	/// Get a reference to the transaction filter, if present.
+	pub fn tx_filter(&self) -> Option<&Arc<TransactionFilter>> {
```
VladLupashevskyi: Why do we need to have it here, since it's not used anywhere, or have I missed smth?

ethcore/machine/src/tx_filter.rs
```diff
@@ -68,6 +68,11 @@ impl TransactionFilter {
 		)
 	}
 
+	/// Get a reference to the contract address.
+	pub fn contract_address(&self) -> &Address {
```
VladLupashevskyi: The same here

ethcore/machine/src/machine.rs
vkomenda: Sorry, this is perhaps an unwanted effect of a rebase.

ethcore/machine/src/tx_filter.rs
vkomenda: Agreed. This method can be removed as well as `Machine::tx_filter`.

vkomenda:
> Don't we need to let the miner know about gas limit transitions? It seems to me that following the logic of having contract for gas limit configurations should mean that the miner would have to change the limit on the fly, otherwise you cannot be sure that transactions will get mined after such a transition until the validator node configuration is updated.

We set the gas limit once for every block at the time of header construction in [`Engine::populate_from_parent`](https://github.com/paritytech/parity-ethereum/pull/10928/files#diff-79ef4a3dce58a9d303d49c5aee781e17R1081). Do you think the miner can mine transactions in a block before its header has been constructed?

VladLupashevskyi:
@vkomenda I just have noticed that `gas_range_target` is not updated with the value from contract and it's used in block opening here:

https://github.com/paritytech/parity-ethereum/blob/bceb1d569147cf113520b904bf84255d100b4d1d/ethcore/src/client/client.rs#L2332

Which would lead to issues that miner wouldn't allow tx to be included to the block if its gas cost is more then it's set for `OpenBlock `, however I have checked and tx is included without any issues.

ethcore/verification/src/verification.rs
```diff
@@ -268,15 +268,6 @@ pub(crate) fn verify_header_params(header: &Header, engine: &dyn Engine, is_full
...
 	}
-	let min_gas_limit = engine.params().min_gas_limit;
-	if header.gas_limit() < &min_gas_limit {
-		return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: Some(min_gas_limit), max: None, found: *header.gas_limit() })));
-	}
-	if let Some(limit) = engine.maximum_gas_limit() {
-		if header.gas_limit() > &limit {
-			return Err(From::from(BlockError::InvalidGasLimit(OutOfBounds { min: None, max: Some(limit), found: *header.gas_limit() })));
-		}
-	}
```
afck: Would it be better to keep these checks, whenever the gas limit contract returns `None`?

vkomenda: They've moved to line 331 below.

afck:
I can't reproduce that test failure. For me, `cargo test --release --all` works.

afck:
I rebased again, due to merge conflicts. Please let me know if there are any further changes you would like me to make to the PR.

ethcore/engine/src/engine.rs
```diff
@@ -331,6 +331,12 @@ pub trait Engine: Sync + Send {
 		Ok(*header.author())
 	}
 
+	/// Overrides the block gas limit. Whenever this returns `Some` for a header, the next block's gas limit must be
+	/// exactly that value.
+	fn gas_limit_override(&self, _header: &Header) -> Option<U256> {
```
dvdplm: So I think I made this suggestion before but perhaps it got lost.

In order to keep the `Engine` trait from growing ever fatter, how about we have a `fn gas_limits(&self) -> Option<(U256, U256)>` that return the min/max limit. The default impl could return  `(self.params().min_gas_limit, self.params().max_gas_limit)` and then Aura can provide its own impl with the limit override logic tucked away nicely. This would require refactoring other code around the project but lets us remove `maximum_gas_limit(&self)` and saves us from adding two new methods.

Not sure if you tried this and it doesn't work or if you feel it's a bad idea?

afck: Sorry, I missed that!

I'm not sure about this idea: When the engine changes the gas limit, we don't want to do [these checks](https://github.com/paritytech/parity-ethereum/pull/10928/files#diff-4eb43f14344f39e06bd7ceb98f5b3353R345):

At the moment, it seems to be impossible to produce a valid block if the interval based on the gas limit divisor doesn't overlap with the interval based on the engine's min and max gas limits.
Maybe engines that use a gas limit contract should just use a gas limit divisor of 1, but that would still prevent them from more than doubling the gas limit from one block to the next.

But we could make `gas_limits` return a `bool` as well, that tells us whether it's an override. Or it could return an enum?
```rust
enum GasLimits {
    MinMax(U256, U256),
    Override(U256),
}
```

dvdplm: Hmm, so why do you need to know that there was an override? Why isn't it enough to know what the bounds must be at that block? In this PR there are 3 error cases: gas too low, too high, overridden but mismatching – why is it important to know that the error came from an override?

afck: There seem to be two different intervals for the block gas limit:

* [the engine's min/max](https://github.com/paritytech/parity-ethereum/pull/10928/files#diff-4eb43f14344f39e06bd7ceb98f5b3353R331) and
* [the range calculated using the gas limit divisor](https://github.com/paritytech/parity-ethereum/pull/10928/files#diff-4eb43f14344f39e06bd7ceb98f5b3353R347), which allows us to change the limit only by a certain percentage compared to _the previous block_.

The second limitation can conflict with the first, if the engine makes sudden changes to the gas limit. So I think it should be disabled if there is an override.

I'm not really sure at all what's the best way to approach this. To give you some context, the gist is:
In [POSDAO](https://forum.poa.network/c/posdao), there will be rather expensive implicit calls to the governance contracts once per week, to distribute rewards and select new validators. While these are ongoing, we want to limit the gas available for transactions, so that the total CPU usage per block remains roughly constant. (For details see: https://github.com/poanetwork/parity-ethereum/issues/119#issue-430516284)

ethcore/verification/src/verification.rs
```diff
@@ -332,12 +321,36 @@ fn verify_parent(header: &Header, parent: &Header, engine: &dyn Engine) -> Resul
...
+				OutOfBounds { min: Some(min_gas_limit), max: None, found: *header.gas_limit() }
+			).into());
+		}
+		if let Some(limit) = engine.maximum_gas_limit() {
+			if header.gas_limit() > &limit {
+				return Err(BlockError::InvalidGasLimit(
+					OutOfBounds { min: None, max: Some(limit), found: *header.gas_limit() }
+				).into());
+			}
+		}
```
debris: this function is called `verify_parent`, but this logic has nothing to do with the parent header

debris: what if you move this verification to engine's `fn verify_block_basic`? If you do that you will not need to add additional methods to the `Engine` trait

afck: Done.
I also moved the checks back to `verify_header_params` that were originally there.
Since I'm now calling `gas_limit_override` three times instead of once, I added a cache to memoize the results, to avoid executing the same contract call multiple times.

debris: Thanks for changes, but that's not exactly what I wanted 😅 

I believe that files:
- `ethcore/verification/src/verification.rs`
- `ethcore/engine/src/engine.rs`

can and should not be modified by this pr.

All the changes can be encapsulated in `ethcore/engines/authority-round/src/lib.rs` file

afck: How could this be achieved without modifying `ethcore/verification/src/verification.rs`? After all, it's an exception to block verification to allow for gas limits set by the engine.

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1730,7 +1737,12 @@ impl Engine for AuthorityRound {
 				return None;
 			}
 		};
-		block_gas_limit(full_client, header, address)
+		if let Some(limit) = self.gas_limit_override_cache.lock().get_mut(&header.hash()) {
```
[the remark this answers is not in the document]
afck: I agree, but [LruCache](https://docs.rs/lru-cache/0.1.2/lru_cache/struct.LruCache.html) seems to only come with `get_mut` for some reason; there's no `get`.

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -507,6 +511,8 @@ pub struct AuthorityRound {
 	received_step_hashes: RwLock<BTreeMap<(u64, Address), H256>>,
 	/// The addresses of contracts that determine the block gas limit.
 	block_gas_limit_contract_transitions: BTreeMap<u64, Address>,
+	/// Memoized gas limit overrides, by block hash.
+	gas_limit_override_cache: Mutex<LruCache<H256, Option<U256>>>,
```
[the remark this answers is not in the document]
afck: Since there's only `get_mut`, we always need to obtain a write lock.

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -122,6 +123,9 @@ pub struct AuthorityRoundParams {
 
 const U16_MAX: usize = ::std::u16::MAX as usize;
 
+/// The number of recent block hashes for which the gas limit override is memoized.
+const GAS_LIMIT_OVERRIDE_CACHE_CAPACITY: usize = 10;
```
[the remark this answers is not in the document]
afck: Yes, in POSDAO we'll have overrides for longer sequences of blocks.
But the main reason I added the cache is that with the current implementation in `verification.rs`, the gas limit is accessed three times for the current block and once for each uncle. So ideally, all of those values would still be cached.
If we expect blocks to rarely have more than 9 uncles, 10 should be fine. (And it takes very little memory anyway.)
Once a block has been verified and all of its nephews have been imported, it can drop out of the cache.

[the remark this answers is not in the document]
dvdplm: Ok, that's a great explanation, thank you. Maybe extend the comment with that info? But I guess it depends on what happens with the rest of the code: if you manage to move it out of `verification` and into the engine, maybe this need to change.

ethcore/verification/src/verification.rs
afck: What we could do is move all the gas limit related verification code from `verification.rs` to the engines, and make the current behavior the `Engine` trait's default implementation. Do you think that would be better?

ethcore/engines/authority-round/src/lib.rs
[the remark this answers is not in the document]
dvdplm: Maybe you don't need a `LruCache` and a plain old `RwLock<HasMap<…>>` could do the trick?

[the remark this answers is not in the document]
afck: But isn't it a waste of memory to keep all old blocks' block gas limits around forever?

varasev:
Hi guys, please consider @afck's last answers. Are there still some things we need to improve in this PR?

afck:
Another test failure that I can't reproduce locally.

ethcore/engines/authority-round/src/lib.rs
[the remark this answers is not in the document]
seunlanlege: wait so is `Mutex` _faster_  than `RwLock` ?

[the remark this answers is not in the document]
afck: I don't know, but I'd expect so: It's definitely simpler than an `RwLock` because it never has to deal with multiple locks.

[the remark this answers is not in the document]
dvdplm: > wait so is `Mutex` _faster_ than `RwLock` ?

It's complicated and depends on the context. If your code needs several threads to write to the resource under the lock, a `Mutex` is probably the way to go. On the other hand, if your code is a mix of reads and writes, using an `RwLock` lets your reads go ahead unhindered by the write lock. Only benchmarks (using realistic workloads) can tell you exactly which read/write mix is the best for your single case. It's a big topic. :)

Here, given the API of `lru-cache`, my suggestion to use an `RwLock` doesn't work at all ofc.

ordian:
It seems that the code is reviewed well. Please resolve the merge conflicts, so that we can proceed with merge :)

afck:
Thanks! I rebased it.

PR #10929, opened 2019-07-30
journaldb changes

debris:
The goal of this pr is to simplify a bit journaldb interface and usage

changes:

- edition 2018
- removed traits `KeyedHashDB` and `AsKeyedHashDB` as they are redundant
- moved `fn keys()` from `KeyedHashDB` to `JournalDB` as it is used only to check journal entries in ethcore block tests
- remove `KeyedHashDB` implementation for `ProofCheck` and `Proving` as it was never used
- `commit_batch` and `inject_batch`  are no longer a part of trait declaration, but they are independent functions hidden under `cfg(test)` gate
- `JournalDB` trait is now declared in `lib.rs` file
- `overlaydb` is no longer a `pub` module
- removed unused `revert` and `commit_refs` from `OverlayDB`


util/journaldb/src/lib.rs
```diff
@@ -16,51 +16,86 @@
 
 //! `JournalDB` interface and implementation.
 
-extern crate parity_util_mem;
-extern crate parity_util_mem as mem;
 extern crate parity_util_mem as malloc_size_of;
```
ordian: Have you considered renaming it in `Cargo.toml`? (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml)

debris: oh, I didn't know it's possible! thanks :)

PR #10931, opened 2019-08-01
Better error message for rpc gas price errors

### LINKED ISSUE #10348: "Transaction Gas price too low. It does not satisfy your node's minimal gas price" error always prints 0 as the `minimal` price
- **Parity Ethereum version**: 2.2.10
- **Operating system**: Linux
- **Fully synchronized**: yes
- **Network**: ethereum mainnet
- **Restarted**: yes

**Expected Behavior**
`eth_sendRawTransaction` with gas price of 5000000000 wei goes through or returns a different error message.

**Actual Behavior**
`eth_sendRawTransaction` returns
```json
{"jsonrpc":"2.0","error":{"code":-32010,"message":"Transaction gas price is too low. It does not satisfy your node's minimal gas price (minimal: 0, got: 5000000000). Try increasing the gas price."}
```

Checking the node's minimal gas price definitely returns `0x250e87c` or 38856828 wei, which is definitely still less than 5000000000 wei.

```shell
curl --data '{"jsonrpc": "2.0", "id": 0, "method": "parity_minGasPrice"}' -H "Content-Type: application/json" -X POST localhost:8545
```

```json
{"jsonrpc":"2.0","result":"0x250e87c","id":0}
```

I've never programmed in Rust, but looking at the [source code of the error](https://github.com/paritytech/parity-ethereum/blob/d5c19f8719dbcaeabafe1ed99cd0486c52ea0c48/rpc/src/v1/helpers/errors.rs#L396) and the [default option](https://github.com/paritytech/parity-ethereum/blob/d89b8d904ff28382d3692290a936706304627255/miner/src/pool/verifier.rs#L40), it looks like maybe the option is never getting set, or maybe there's some printing floating point error?

seunlanlege:
Previously if you were trying to replace a transaction in the transaction pool, (i.e same nonce) but you don't supply a higher gas price; you get this error telling you that your "transaction's gas price does satisfy node's minimal gas price"
```
Transaction gas price is too low. It does not satisfy your node's minimal gas price (minimal: 0, got: 5000000000). Try increasing the gas price.
```

This error message has been changed to:

```
Transaction gas price 38856828wei is too low. There is another transaction with the same nonce in the queue with a gas price: 5000000000wei. Try increasing the gas price or incrementing the nonce.
```

closes #10348


rpc/src/v1/helpers/errors.rs
```diff
@@ -423,6 +423,9 @@ pub fn transaction_message(error: &TransactionError) -> String {
 		TooCheapToReplace => {
 			"Transaction gas price is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.".into()
 		}
+		TxTooCheapToReplace { prev, new } => {
```
ordian: We already have `TooCheapToReplace` error, do you think we can replace the old one with the new one?

seunlanlege: The old one is constructed here: 

https://github.com/paritytech/parity-ethereum/blob/6c7d0fef4e30e03920a174eea0344260d59a1077/miner/src/pool/queue.rs#L594-L598

from `transaction_pool::Error` which holds the transaction hash, not the gas prices

https://github.com/paritytech/parity-common/blob/57555f49b6873193437624404121f17adae2723b/transaction-pool/src/error.rs#L21-L28

ordian: Maybe we could make the `prev` and `new` `Option`al, so that we print more info if it's available, while keeping one error type. Or just document while there are two error types for the same error.

ethcore/types/src/transaction/error.rs
```diff
@@ -42,6 +42,14 @@ pub enum Error {
 		/// Transaction gas price
 		got: U256,
 	},
+	/// Transaction has too low fee
+	/// (there is already a transaction with the same sender-nonce but higher gas price)
```
dvdplm: We need to document this better: what is the difference between `TooCheapToReplace` and `TxTooCheapToReplace` – it is not clear to when we return the one or the other.

dvdplm:
test failures seem legit

miner/src/pool/tests/mod.rs
```diff
@@ -91,10 +91,10 @@ fn should_return_correct_nonces_when_dropped_because_of_limit() {
...
 	assert_eq!(res, vec![Ok(()), Ok(())]);
 	assert_eq!(res2, vec![
-			   // The error here indicates reaching the limit
-			   // and minimal effective gas price taken into account.
-			   Err(transaction::Error::InsufficientGasPrice { minimal: 2.into(), got: 1.into() }),
-			   Ok(())
+        // The error here indicates reaching the limit
+        // and minimal effective gas price taken into account.
+			   Err(transaction::Error::TooCheapToReplace { prev: 2.into(), new: 1.into() }),
+        Ok(())
```
ordian: spaces

rpc/src/v1/helpers/errors.rs
```diff
@@ -420,8 +420,8 @@ pub fn transaction_message(error: &TransactionError) -> String {
 	match *error {
 		AlreadyImported => "Transaction with the same hash was already imported.".into(),
 		Old => "Transaction nonce is too low. Try incrementing the nonce.".into(),
-		TooCheapToReplace => {
-			"Transaction gas price is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.".into()
+		TooCheapToReplace { prev, new } => {
+			format!("Transaction gas price {:?}wei is too low. There is another transaction with same nonce in the queue with gas price: {:?}wei. Try increasing the gas price or incrementing the nonce.", new, prev)
```
ordian: IMO we should print `"{:?}wei"` and "with gas price: {:?}wei"` part only if new and prev are `Some`.

PR #10940, opened 2019-08-03
Change the return type of step_inner function.

Leeleo3x:
This PR simplifies the return result of step_inner function.

Using `Result<Never, InterpreterResult>` as return result will cause 
unnecessary overhead (mostly additional memory operations). Since 
step_inner is a hot function, I propose to simplify the return type of 
step_inner function and handle the error results manually.

Some benchmark results I have:

Without Optimization (Original Version):
test rng_u256          ... bench:  68,395,922 ns/iter (+/- 19,520,348)
test rng_usize         ... bench:  46,822,703 ns/iter (+/- 19,869,026)
test simple_loop_u256  ... bench:  18,807,019 ns/iter (+/- 5,683,907)
test simple_loop_usize ... bench:  13,812,568 ns/iter (+/- 3,310,862)

With Optimization:
test rng_u256          ... bench:  53,654,255 ns/iter (+/- 12,985,727)
test rng_usize         ... bench:  36,481,886 ns/iter (+/- 13,044,046)
test simple_loop_u256  ... bench:  15,010,101 ns/iter (+/- 5,306,132)
test simple_loop_usize ... bench:  10,584,103 ns/iter (+/- 2,439,375)

Some perf results I have:
 
image

### DIFF
--- a/ethcore/evm/src/interpreter/mod.rs
+++ b/ethcore/evm/src/interpreter/mod.rs
@@ -341,13 +341,13 @@ impl<Cost: CostType> Interpreter<Cost> {
 
 				// Calculate gas cost
 				let requirements = match self.gasometer.as_mut().expect(GASOMETER_PROOF).requirements(ext, instruction, info, &self.stack, self.mem.size()) {
-                    Ok(t) => t,
+					Ok(t) => t,
 					Err(e) => return InterpreterResult::Done(Err(e)),
 				};
 				if self.do_trace {
 					ext.trace_prepare_execute(self.reader.position - 1, opcode, requirements.gas_cost.as_u256(), Self::mem_written(instruction, &self.stack), Self::store_written(instruction, &self.stack));
 				}
-                if let Err(e) = self.gasometer.as_mut().expect(GASOMETER_PROOF).verify_gas(&requirements.gas_cost) {
+				if let Err(e) = self.gasometer.as_mut().expect(GASOMETER_PROOF).verify_gas(&requirements.gas_cost) {
 					return InterpreterResult::Done(Err(e));
 				}
 				self.mem.expand(requirements.memory_required_size);
@@ -365,7 +365,7 @@ impl<Cost: CostType> Interpreter<Cost> {
 					Ok(x) => x,
 				};
 				evm_debug!({ self.informant.after_instruction(instruction) });
-                result
+				result
 			},
 		};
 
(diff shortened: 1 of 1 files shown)


Leeleo3x:
[clabot:check]

dvdplm:
Do you mind including the benchmark code in the PR?

Removing the error conversion and all uses of `?` makes the code less readable. Can you elaborate on why the return type change necessitates changing the error handling like this?

Leeleo3x:
The benchmark I'm using is included in current Parity code base. To run benchmarks:
```
cargo bench --package evmbin
```

I do understand `?` operator makes the code more readable. However, the `?` operator calls `into_result` function and wraps the result into `Result` type which causes additional memory operations. This is also because the size of `InterpreterResult` is large and need to be stored in memory to move around.

ordian:
> Do you mind including the benchmark code in the PR?

I think the benches are from `evmbin/benches/mod.rs`.

> Removing the error conversion and all uses of ? makes the code less readable. Can you elaborate on why the return type change necessitates changing the error handling like this?

That's a known issue https://github.com/rust-lang/rust/issues/37939

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -336,22 +328,28 @@ impl<Cost: CostType> Interpreter<Cost> {
...
 				self.last_stack_ret_len = info.ret;
-				self.verify_instruction(ext, instruction, info)?;
+				if let Err(e) = self.verify_instruction(ext, instruction, info) {
+					return InterpreterResult::Done(Err(e));
+				};
 
 				// Calculate gas cost
-				let requirements = self.gasometer.as_mut().expect(GASOMETER_PROOF).requirements(ext, instruction, info, &self.stack, self.mem.size())?;
+				let requirements = match self.gasometer.as_mut().expect(GASOMETER_PROOF).requirements(ext, instruction, info, &self.stack, self.mem.size()) {
+                    Ok(t) => t,
```
ordian: spaces

ethcore/evm/src/interpreter/mod.rs
```diff
@@ -393,28 +392,31 @@ impl<Cost: CostType> Interpreter<Cost> {
 					self.valid_jump_destinations = Some(self.cache.jump_destinations(&self.params.code_hash, &self.reader.code));
 				}
 				let jump_destinations = self.valid_jump_destinations.as_ref().expect("jump_destinations are initialized on first jump; qed");
-				let pos = self.verify_jump(position, jump_destinations)?;
+				let pos = match self.verify_jump(position, jump_destinations) {
+					Ok(x) => x,
+                    Err(e) => return InterpreterResult::Done(Err(e))
```
ordian: and here

dvdplm:
> That's a known issue [rust-lang/rust#37939](https://github.com/rust-lang/rust/issues/37939)

I had no idea, thank you for educating me.

PR #10948, opened 2019-08-06
unify loading spec && further spec cleanups

debris:


ethcore/src/spec/spec.rs
```diff
@@ -186,17 +186,16 @@ fn run_constructors<T: Backend>(
 
 /// Parameters for a block chain; includes both those intrinsic to the design of the
 /// chain and those to be interpreted by the active chain engine.
+#[derive(Clone)]
```
ordian: Do we use `spec.clone()` at all?
```suggestion
```

ethcore/src/spec/bundle.rs
```diff
@@ -0,0 +1,166 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Ethereum protocol module.
+//!
+//! Contains all Ethereum network specific stuff, such as denominations and
```
ordian: These docs need some love. We have removed denominations recently, also it's not only ethereum specific.

ethcore/src/spec/spec.rs
dvdplm: I think this replaces the test-only `Clone` impl ([here](https://github.com/paritytech/parity-ethereum/pull/10948/files#diff-8c15897d65990126e87e06b4feefef7dL234))

ordian: @dvdplm yeah, I saw that, but it code doesn't seem to use a clone in tests.
Otherwise, it can be replaced with 
```suggestion
#[cfg_attr(test, derive(Clone))]
```

ethcore/src/spec/bundle.rs
debris: hahaha, sure!

debris:
updated, please review again :)

ethcore/src/spec/chain.rs
```diff
@@ -0,0 +1,165 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Ethereum protocol module.
```
ordian:
```suggestion
//! Chain specification module.
```

dvdplm: Ouch missed this. I'll make this change elsewhere.

PR #10949, opened 2019-08-06
Extract Machine from ethcore

dvdplm:
In the pursuit of a slimmer `ethcore` crate, this PR moves the `Machine` and related types/code to a new `machine` crate, ~3kloc of Rust. 
I am not entirely sure that my choice of what to move where is the best possible way of slicing things up, but I have made a few failed attempts to disentangle `Machine` from `Engine` and at the end of the day I think that `Machine` is the lower level type here and so easier to move.

The end goal, other than a slimmer `ethcore` crate ofc, is to be able to move the `Engine` trait outside of `ethcore` which in turn would enable us to move the engine impls to own crates as well. I believe that would be beneficial both for maintaining existing code and for external contributors to work on new/tweaked engines.

The PR is noisy and for that I apologise. There is no new logic added or changed, other than some slight refactoring of the way tests instantiate `Machine`s. The `ethcore::client` module re-exports plenty of types for no apparent reason and removing those led to significant fallout in other places, contributing to the noise. I do believe that re-exports of that kind creates more confusion than convenience.

The most relevant changes of this PR is the addition of the `machine` crate and a new `client-traits` crate (which for now is tiny, but could grow if/when we extract more traits from `ethcore::client::traits`). Extracting the `machine` crate implies moving plenty of other code as well: `Executive`, `Externalities`, `Substate` and `TxFilter`.


ethcore/light/src/on_demand/mod.rs
```diff
@@ -40,7 +40,8 @@ use cache::Cache;
 use request::{self as basic_request, Request as NetworkRequest};
 use self::request::CheckedRequest;
 
-pub use ethcore::executed::ExecutionResult;
+use machine::executed::ExecutionResult; // todo[dvdplm] why was this re-exported?
```
debris: I also don't know and I also don't think it should be

ethcore/machine/src/test_helpers.rs
```diff
@@ -0,0 +1,60 @@
...
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+// Creating test Machines
+
+use common_types::engines::params::CommonParams;
+use ethjson;
+use crate::Machine;
+
+
+pub fn load_machine(reader: &[u8]) -> Machine {
```
debris: this will cause conflicts with #10948 . not an issue, just observation :)

ethcore/machine/src/tx_filter.rs
```diff
@@ -16,22 +16,24 @@
...
 	engines::params::CommonParams,
 	transaction::{Action, SignedTransaction}
 };
-use hash::KECCAK_EMPTY;
+use keccak_hash::KECCAK_EMPTY;
 
-use_contract!(transact_acl_deprecated, "res/contracts/tx_acl_deprecated.json");
-use_contract!(transact_acl, "res/contracts/tx_acl.json");
+use_contract!(transact_acl_deprecated, "../res/contracts/tx_acl_deprecated.json");
+use_contract!(transact_acl, "../res/contracts/tx_acl.json");
```
debris: maybe let's move those json files to `machine/res` dir

ethcore/src/client/mod.rs
```diff
@@ -35,19 +35,18 @@ pub use self::io_message::ClientIoMessage;
...
-	BlockChainReset
+	BlockChainReset, BlockChainClient, EngineClient, ProvingBlockChainClient, IoClient
 };
 pub use account_state::state::StateInfo;
-pub use self::traits::{BlockChainClient, EngineClient, ProvingBlockChainClient, IoClient};
 
-pub use types::ids::*;
-pub use types::trace_filter::Filter as TraceFilter;
-pub use types::pruning_info::PruningInfo;
-pub use types::call_analytics::CallAnalytics;
```
debris: removing those reexports is great!

ethcore/src/lib.rs
```diff
@@ -153,19 +155,19 @@ pub mod block;
 pub mod client;
 pub mod engines;
 pub mod ethereum;
-pub mod executed;
-pub mod executive;
+//pub mod executed;
+//pub mod executive;
 pub mod executive_state;
-pub mod machine;
+//pub mod machine;
```
debris: comments to remove

ethcore/machine/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+// todo[dvdplm]: document all modules
+
+// todo[dvdplm]: fix ethabi for use with edition 2018
```
debris: todo to fix

ethcore/machine/src/test_helpers.rs
dvdplm: Yes, it'll be terrible!

ethcore/src/client/mod.rs
dvdplm: I really dislike re-exports. I get they can be convenient at times, and that there are valid use-cases for them but most of the time I think they should be avoided (even forbidden if it's just done for convenience). There are places where we pull in all of `ethcore` mostly just to gain access to re-exported types, which is not so great...

PR #10954, opened 2019-08-07
Verify transaction against its block during import

### LINKED ISSUE #10411: POA: chain id is not verified during transition submit

- **Parity Ethereum version**: 2.2.7-stable
- **Operating system**: Linux
- **Installation**:  one-line installer 
- **Network**: private-poa
- **Restarted**: no

Issue Description:
I build a private poa network with chainid other than ethereum's chainid 1. If I submit a transaction with a wrong chainid, for example chainid=1. A block will be created on the current node because chainid was not being verified during transition submittion. However this block will never being successfully boardcased to other nodes because verification will be failed during sync. Error log is:

 IO Worker #0 WARN client Stage 1 block verification failed for 0x9b03…9db9: Error(Transaction(InvalidChainId), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })

I expected the wrong chainid transition will be rejected at the first place. Otherwise the chaindata on nodes are broken as this screen shot below: block height are not synced.

screen shot 2019-02-25 at 12 13 26 am

grbIzl:
Currently transaction's basic verification is not performed during its import from the pool (transaction's verify_basic method is not called).
As a result, incorrect transaction (for example, with incorrect chain id) can be added into the block and mined. Such blocks won't be accepted by other nodes (this transaction's verification will be called for the new block on other nodes). But incorrect block will stay on originator's node.

Solution: perform engine's verify_transaction_basic before adding transaction to the block during import (this will call transaction's verify_basic).

Drawback: this additional verification will increase import's time.

Closes #10411

### DIFF
--- a/ethcore/src/miner/pool_client.rs
+++ b/ethcore/src/miner/pool_client.rs
@@ -120,6 +120,11 @@ impl<'a, C: 'a> PoolClient<'a, C> where
 	pub fn verify_signed(&self, tx: &SignedTransaction) -> Result<(), transaction::Error> {
 		self.engine.machine().verify_transaction(&tx, &self.best_block_header, self.chain)
 	}
+
+	/// Verifies transaction against its block (before its import into this block)
+	pub fn verify_for_pending_block(&self, transaction: &UnverifiedTransaction, header: &Header) -> Result<(), transaction::Error> {
+		self.engine.machine().verify_transaction_basic(transaction, header)
+	}
 }
 
 impl<'a, C: 'a> fmt::Debug for PoolClient<'a, C> {
--- a/ethcore/src/miner/pool_client.rs
+++ b/ethcore/src/miner/pool_client.rs
@@ -138,8 +138,10 @@ impl<'a, C: 'a> pool::client::Client for PoolClient<'a, C> where
 	}
 
 	fn verify_transaction(&self, tx: UnverifiedTransaction)-> Result<SignedTransaction, transaction::Error> {
+		self.engine.verify_transaction_basic(&tx, &self.best_block_header)?;
 		let tx = tx.verify_unordered()?;
-		self.verify_for_pending_block(&tx, &self.best_block_header)?;
+
+		self.engine.machine().verify_transaction(&tx, &self.best_block_header, self.chain)?;
 		Ok(tx)
 	}
 
--- a/ethcore/src/miner/pool_client.rs
+++ b/ethcore/src/miner/pool_client.rs
@@ -114,16 +114,13 @@ impl<'a, C: 'a> PoolClient<'a, C> where
 		}
 	}
 
-	/// Verifies if signed transaction is executable.
+	/// Verifies transaction against its block (before its import into this block)
+	/// Also Verifies if signed transaction is executable.
 	///
 	/// This should perform any verifications that rely on chain status.
-	pub fn verify_signed(&self, tx: &SignedTransaction) -> Result<(), transaction::Error> {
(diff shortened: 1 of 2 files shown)


dvdplm:
Have you measured the increased import time with this change?

grbIzl:
> Have you measured the increased import time with this change?

With the fix:
2019-08-08 13:46:50  Total pushed 1713 transactions in 354 ms
Avg transaction's import time: 0.206 ms

Without fix:
Total pushed 1636 transactions in 320 ms
Avg transaction's import time: 0.196 ms

Only the time of the transactions' import to the block was measured.

dvdplm:
> With the fix:
> Total pushed 1713 transactions in 354 ms
> Avg transaction's import time: 0.206 ms
> 
> Without fix:
> Total pushed 1636 transactions in 320 ms
> Avg transaction's import time: 0.196 ms

Are they different transactions? If yes, can you measure with the same txs?

grbIzl:
> Are they different transactions? If yes, can you measure with the same txs?

No, it was actually the same tx sent > 1000 times in the loop.

curl --data '{"jsonrpc":"2.0","method":"personal_sendTransaction","params":[{"from":"0x004ec07d2329997267Ec62b4166639513386F32E","to":"0x00Bd138aBD70e2F00903268F3Db08f2D25677C9e","value":"0xde0b6b3a7640000"}, "user"],"id":0}' -H "Content-Type: application/json" -X POST localhost:8540

ethcore/src/miner/pool_client.rs
```diff
@@ -120,6 +120,11 @@ impl<'a, C: 'a> PoolClient<'a, C> where
 	pub fn verify_signed(&self, tx: &SignedTransaction) -> Result<(), transaction::Error> {
 		self.engine.machine().verify_transaction(&tx, &self.best_block_header, self.chain)
 	}
+
+	/// Verifies transaction against its block (before its import into this block)
+	pub fn verify_for_pending_block(&self, transaction: &UnverifiedTransaction, header: &Header) -> Result<(), transaction::Error> {
+		self.engine.machine().verify_transaction_basic(transaction, header)
```
tomusdrw: Why not call `verify_transaction` here as well (and verify against `header` not `best_block_header`)? So that we can actually simplify the miners code and perform only one check.

grbIzl: Sorry, I don't understand your suggestion. Which verify_transaction method do you mean? verify_transaction from machine doesn't call required verification methods. verify_transaction from pool client doesn't check against custom header, only for best block's

grbIzl: Thanks! Got it.

ethcore/src/miner/pool_client.rs
```diff
@@ -136,11 +138,8 @@ impl<'a, C: 'a> pool::client::Client for PoolClient<'a, C> where
 	}
 
 	fn verify_transaction(&self, tx: UnverifiedTransaction)-> Result<SignedTransaction, transaction::Error> {
-		self.engine.verify_transaction_basic(&tx, &self.best_block_header)?;
 		let tx = tx.verify_unordered()?;
-
-		self.verify_signed(&tx)?;
-
+		self.verify_for_pending_block(&tx, &self.best_block_header)?;
```
tomusdrw: Ah, unfortunately we can't change the order here. `basic` needs to be performed before `verify_unordered`, otherwise we are risking a DoS attack vector.

Maybe just use `engine.machine()` methods here directly.

grbIzl: Ok

PR #10958, opened 2019-08-10
Extract the Engine trait

dvdplm:
Another big, noisy PR. This time the `Engine` trait and associated types are extracted to [an `engine` crate](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-cd228108908a02064ea5ca4ebc710626R20). Along with this many of the client traits have been moved from `ethcore/client/traits.rs` into the [`client_traits` crate](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-de317cf28117c4c1c13bdeabf6196857).

To some degree the PR makes arbitrary decisions on which types/traits to move. The criteria used is not so much about what belongs where, but rather given that we want a free standing `Engine`, what else *has* to move? One example of such unclean division is how the `engine` crate now hosts [two traits related to snapshotting](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-6415efe306dc7594afc85e1873b7b8e6R33).

The only [relevant logic change](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-ad151464b04d9bf02d2d22eb6bbbe6f8R245) is in `Ethash`: the `epoch_verifier()` method requires a type that implements `EpochVerifier` that can be stuck inside a `ConstructedVerifier` enum and before this PR this was achieved by an impl on `Arc<Ethash>`. When the `Engine` trait is no longer local, this is not possible. Instead there is now a local `EpochVerifier` struct that impls the `EpochVerifier` trait, much like the other engines do. This leads to somewhat convoluted code but otoh it saves us the doubly-arc-wrapped setup we had before.

Based on https://github.com/paritytech/parity-ethereum/pull/10949


ethcore/Cargo.toml
```diff
@@ -85,6 +88,8 @@ rlp_compress = { path = "../util/rlp-compress" }
 serde_json = "1.0"
 tempdir = "0.3"
 trie-standardmap = "0.12.4"
+engine = { path = "./engine", features = ["test-helpers"] }
```
tomusdrw: sort?

ethcore/client-traits/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+call-contract = { package = "ethcore-call-contract", path = "../call-contract" }
+ethereum-types = "0.6.0"
+ethcore-db = { path = "../db" }
+common-types = { path = "../types" }
+bytes = { package = "parity-bytes", version = "0.1.0" }
+stats = { path = "../../util/stats" }
+vm = { path = "../vm" }
+trace = { path = "../trace" }
+ethcore-miner = { path = "../../miner" }
+kvdb = "0.1.0"
```
tomusdrw: sort?

ethcore/engine/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0"
+
+[dependencies]
+blockchain = { package = "ethcore-blockchain", path = "../blockchain" }
+client-traits = { path = "../client-traits" }
+common-types = { path = "../types" }
+builtin = { path = "../builtin", package = "ethcore-builtin" }
```
tomusdrw: sort?

ethcore/engine/src/snapshot.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
```
tomusdrw: Is there any reason why `snapshots` are part of engine?

ethcore/src/engines/authority_round/mod.rs
```diff
@@ -24,15 +24,18 @@ use std::sync::atomic::{AtomicUsize, AtomicBool, Ordering as AtomicOrdering};
 use std::sync::{Weak, Arc};
 use std::time::{UNIX_EPOCH, Duration};
 
-use block::*;
-use client::EngineClient;
-use engines::{Engine, Seal, ConstructedVerifier};
+use client_traits::EngineClient;
```
tomusdrw: I guess the next step would be to move specific engines implementations into it's own crates? ;)

ethcore/src/engines/ethash.rs
```diff
@@ -344,25 +374,30 @@ impl Engine for Arc<Ethash> {
...
-			res = H256(result.value));
-		if mix != seal.mix_hash {
-			return Err(From::from(BlockError::MismatchedH256SealElement(Mismatch { expected: mix, found: seal.mix_hash })));
-		}
-		if &difficulty < header.difficulty() {
-			return Err(From::from(BlockError::InvalidProofOfWork(OutOfBounds { min: Some(header.difficulty().clone()), max: None, found: difficulty })));
-		}
-		Ok(())
+		verify_block_unordered(&self.pow, header)
+//		let seal = Seal::parse_seal(header.seal())?;
```
tomusdrw: Remove please

ethcore/src/engines/ethash.rs
```diff
@@ -207,17 +241,18 @@ impl Ethash {
 // for any block in the chain.
 // in the future, we might move the Ethash epoch
 // caching onto this mechanism as well.
-// NOTE[dvdplm]: the reason we impl this for Arc<Ethash> and not plain Ethash is the
```
tomusdrw: :tada:

ethcore/src/engines/ethash.rs
```diff
@@ -207,17 +241,18 @@ impl Ethash {
...
+
+impl engine::EpochVerifier for EpochVerifier {
 	fn verify_heavy(&self, header: &Header) -> Result<(), Error> {
-		self.verify_block_unordered(header).into()
+		verify_block_unordered(&self.pow, header)
 	}
 }
 
-impl Engine for Arc<Ethash> {
+//impl Engine for Arc<Ethash> {
```
tomusdrw: Remove

ethcore/src/engines/mod.rs
```diff
@@ -25,423 +25,10 @@ mod null_engine;
...
 
 pub use self::authority_round::AuthorityRound;
 pub use self::basic_authority::BasicAuthority;
 pub use self::instant_seal::{InstantSeal, InstantSealParams};
 pub use self::null_engine::NullEngine;
-pub use self::signer::EngineSigner;
 pub use self::clique::Clique;
 pub use self::ethash::{Ethash, Seal as EthashSeal};
-
-// TODO [ToDr] Remove re-export (#10130)
```
tomusdrw: :tada:

ethcore/src/engines/authority_round/mod.rs
dvdplm: You nailed it. :)

ethcore/engine/src/snapshot.rs
dvdplm: No good reason that I can think of. However the `Engine` trait does expose the `SnapshotComponent` trait (as a return type), which in turn pulls in a bunch of other stuff. Quite vexing but yeah, this didn't feel like the right PR to sort that out.

tomusdrw: Fair enough.

ethcore/src/engines/authority_round/mod.rs
afck: We're actually already doing that with our WIP Honey Badger engine:
https://github.com/poanetwork/parity-ethereum/tree/hbbft/ethcore/hbbft_engine
Saves a lot of compile time!

dvdplm: @afck See https://github.com/paritytech/parity-ethereum/pull/10966 – once you get rid of the `ethcore` dependency compilation times are pretty great.

PR #10964, opened 2019-08-13
Fix ethcore/benches build.

afck:


afck:
I think this was overlooked in #10949. It would probaby make sense to add something like `cargo check --benches` to CI.

PR #10966, opened 2019-08-13
Extract engines to own crates

dvdplm:
Extracts all engines to their own crates.

The main possibly contentious change here is the decoupling of `Engine` and the snapshotting code. Before this PR the `engine` crate had to host a few traits it really has no business with: `SnapshotComponents` and `Rebuilder`. Here, the two traits are moved back to `ethcore` and when the `Client` initiates a snapshot there is a [`chunker` function](https://github.com/paritytech/parity-ethereum/pull/10966/files#diff-6f2331322c25086c5cbee5a32449ea43R31) that hands back the appropriate `dyn SnapshotComponent`. 

Based on https://github.com/paritytech/parity-ethereum/pull/10958


ethcore/src/snapshot/consensus/mod.rs
```diff
@@ -22,3 +22,18 @@ mod work;
...
 pub use self::work::*;
+
+use ethash_engine::{MAX_SNAPSHOT_BLOCKS, SNAPSHOT_BLOCKS};
+use snapshot::SnapshotComponents;
+
+/// Create a factory for building snapshot chunks and restoring from them.
+/// `None` indicates that the engine doesn't support snapshot creation.
+pub fn chunker(engine_name: &str) -> Option<Box<dyn SnapshotComponents>> {
+	match engine_name {
+		"AuthorityRound" => Some(Box::new(PoaSnapshot)),
```
tomusdrw: Hmm - this feels really fragile. We do import all engines to `ethcore` anyway, right? Can't we at least use constants?

dvdplm: It is, and it's the crucial point point of this PR. The thing to note is that we have two modes of performing a snapshot: PoA and PoW. Neither is tied to the specific engine used. That means there's an opportunity to decouple the two and get rid of the `snapshot_components()` method of the `Engine` trait. 
That method of the `Engine` trait is what makes it problematic to extract engines from `ethcore` because snapshots depend on `Client`.
In this PR we do import all engines to `ethcore`, but in [the next one]() `spec` is extracted. 

Using constants in a clean way is surprisingly messy. Callers of `chunker()` have a `dyn Engine` available so the obvious solution to add an associated `NAME` const to the `Engine` trait doesn't work (because trait object safety…). I did make some feeble attempts to have the `name()` method return a `TypeId` instead of a `&str` but it turned ugly fast.
Of course I could add an `ENGINES` local const here and match on that, but it didn't seem like a real improvement tbh. Happy to do that if you feel it's better.

ethcore/engines/validator-set/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+description = "Manage validators and sets of validators"
```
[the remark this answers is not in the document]
dvdplm: Well, it's a grab-bag of functionality really: creation and validation of contracts, contract calling, epoch management, proofs and proving, trait&type defs. I wouldn't say it's only state retrieval.

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1595,13 +1597,7 @@ impl Engine for AuthorityRound {
...
 	}
 
-	fn snapshot_components(&self) -> Option<Box<dyn (SnapshotComponents)>> {
-		if self.immediate_transitions {
-			None
-		} else {
-			Some(Box::new(::snapshot::PoaSnapshot))
-		}
-	}
+	fn supports_warp(&self) -> bool { !self.immediate_transitions }
```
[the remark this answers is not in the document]
dvdplm: TLDR is: logic is the same but moved here.

Before this PR we had `snapshot_components()` (with a default impl returning `None`) and this method was [implemented in terms of that](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-828b7601a264bb7a19bbaef294131f47R317). In this PR the `snapshot_components()` method is removed from the `Engine` trait so `supports_warp()` needs to stand on it's own. 
In the case of `AuthorityRound`, snapshotting is not possible if `immediate_transitions` is turned on; before this PR [that logic was in the `snapshot_components()` impl](https://github.com/paritytech/parity-ethereum/pull/10958/files#diff-3b2a71ed5fccb75dfe326f3b122487cdL1595).

ethcore/executive-state/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+description = "Execute transactions and modify State"
```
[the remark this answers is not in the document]
dvdplm: Maybe "Execute transactions producing a receipt and an optional trace."?

ethcore/src/snapshot/consensus/mod.rs
dvdplm: Had another stab at this: added an `EngineType` to `common_types` and changed `Engine.name()` to return that. Cleans up the match statement here and is neater. It still means we have to maintain a list of the engines we support but it is a bit better.

ethcore/engines/validator-set/Cargo.toml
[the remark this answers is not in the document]
tomusdrw: Can we extend the description then to be more specific?

ethcore/src/snapshot/consensus/mod.rs
tomusdrw: I was rather thinking about something like this (it's not cool if we need to enumerate all engines):
```
/// authority_round.rs
const NAME: &'str = "AuthorityRound";

// snapshot/consensus/mod.rs
match name {
   authority_round::NAME => ...
}
```

However now I thought about a different approach:
```
enum Snapshotting {
  Unsupported,
  Pow { snapshot_blocks, max_snapshot_blocks },
  Poa,
}
/// + a function in Engine
Engine::supports_snapshots() -> Snapshotting;
```
So you don't even need to enumerate engines at all, and they just say what kind of snapshotting they support.

dvdplm: `authority_round::NAME => ...` still means we need to manually enumerate them so that feels like more of the same.

The `Snapshotting` works though, good idea, will try that! :)

ethcore/executive-state/Cargo.toml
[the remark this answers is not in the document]
dvdplm: Yeah, so the [core function of the crate is `execute()`](https://github.com/paritytech/parity-ethereum/blob/663f13e01eec635f64f21c0d344bfad113082e16/ethcore/executive-state/src/lib.rs#L240) which instantiates an `Executive` and calls `transact()` on it passing the state.

ethcore/src/snapshot/consensus/mod.rs
```diff
@@ -22,3 +22,16 @@ mod work;
...
+
+use snapshot::SnapshotComponents;
+use types::snapshot::Snapshotting::{self, *};
+
+/// Create a factory for building snapshot chunks and restoring from them.
+/// `None` indicates that the engine doesn't support snapshot creation.
+pub fn chunker(snapshot_type: Snapshotting) -> Option<Box<dyn SnapshotComponents>> {
+	match snapshot_type {
+		PoA => Some(Box::new(PoaSnapshot)),
+		PoW{ blocks, max_restore_blocks } => Some(Box::new(PowSnapshot::new(blocks, max_restore_blocks))),
```
tomusdrw:
```suggestion
		PoW { blocks, max_restore_blocks } => Some(Box::new(PowSnapshot::new(blocks, max_restore_blocks))),
```

PR #10969, opened 2019-08-13
EIP-1108: Add deactivate_at to builtin contracts

Shadowfiend:
For [EIP-1108](https://eips.ethereum.org/EIPS/eip-1108), the gas costs of certain builtin operations related to alt_bn128 are being reduced. I looked at how these are currently implemented in Parity and asked a couple of questions in the gitter channel. The approach this PR suggests is to add a `deactivate_at` to builtin contracts that allows certain contracts to expire. The expectation for Istanbul would be to use this functionality to deactivate the current alt_bn128 builtins and activate new ones with adjusted gas costs at the chosen fork block.

This is my first PR for Parity and Rust, so apologies if anything is off. I believe I've held to the style guide, but I'm not sure about the ideal next steps in terms of test writing, though I'm happy to dig in as needed. Also open to feedback that this is not a good approach, of course.

Two flags:
- The additions to `is_active` may make for a line that's too long per the style guide; happy to wrap that if preferred.
- `deactivate_at` in `builtin/src/lib.rs` is an `Option<u64>` for expressiveness. If options are too expensive, this should be fairly straightforward to convert to a boolean + `u64` combo.

Opening this as a draft given my own uncertainty :)

### DIFF
--- a/json/src/spec/builtin.rs
+++ b/json/src/spec/builtin.rs
@@ -69,6 +69,8 @@ pub struct Builtin {
 	pub pricing: Pricing,
 	/// Activation block.
 	pub activate_at: Option<Uint>,
+	/// Deactivation block.
+	pub deactivate_at: Option<Uint>,
 }
 
 #[cfg(test)]
@@ -101,5 +103,19 @@ mod tests {
 		assert_eq!(deserialized.name, "late_start");
 		assert_eq!(deserialized.pricing, Pricing::Modexp(Modexp { divisor: 5 }));
 		assert_eq!(deserialized.activate_at, Some(Uint(100000.into())));
+	}
+
+	#[test]
+	fn deactivate_at() {
+		let s = r#"{
+			"name": "early_finish",
+			"deactivate_at": 200000,
+			"pricing": { "modexp": { "divisor": 5 } }
+		}"#;
+
+		let deserialized: Builtin = serde_json::from_str(s).unwrap();
+		assert_eq!(deserialized.name, "early_finish");
+		assert_eq!(deserialized.pricing, Pricing::Modexp(Modexp { divisor: 5 }));
+		assert_eq!(deserialized.deactivate_at, Some(Uint(200000.into())));
 	}
 }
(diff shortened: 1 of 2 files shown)


json/src/spec/builtin.rs
```diff
@@ -102,4 +104,18 @@ mod tests {
...
+		let s = r#"{
+			"name": "early_finish",
+			"deactivate_at": 200000,
+			"pricing": { "modexp": { "divisor": 5 } }
+		}"#;
+
+		let deserialized: Builtin = serde_json::from_str(s).unwrap();
+		assert_eq!(deserialized.name, "early_finish");
+		assert_eq!(deserialized.pricing, Pricing::Modexp(Modexp { divisor: 5 }));
+		assert_eq!(deserialized.activate_at, Some(Uint(200000.into())));
```
ordian:
```suggestion
		assert_eq!(deserialized.deactivate_at, Some(Uint(200000.into())));
```

Shadowfiend:
Fixed the above issue and some author info… Will handle the CLA tomorrow.

Shadowfiend:
Ah yes, of course. That makes a lot of sense.

To clarify, where would you see the `eip1108Transition` parameter as being? It seems like one approach would be adding `eip1108Base`, `eip1108Word`, `eip1108Divisor`, and `eip1108Pair` to the respective `Pricing` structs, alongside an `eip1108ActivationBlock` parameter, and then have the respective `Pricer` implementations switch as appropriate. That seems like it would require the pricers to also take the current block, though, and starts getting into a fairly large change that might exceed my capacity as a codebase and language noob <_<

Shadowfiend:
[clabot:check]

Shadowfiend:
Going to close this PR for now as the approach won't work. I'll consider moving forward with the solution I've outlined above, but want to re-emphasize that I'm new both to the Parity codebase *and* Rust as a language, so I will probably need some heavy support. Also happy to let someone with deeper expertise here take the lead, given [the AllCoreDevs decision](https://github.com/ethereum/pm/blob/master/All%20Core%20Devs%20Meetings/Meeting%2068.md#decisions-made) to have all Istanbul EIPs in by Friday.

niklasad1:
@Shadowfiend have you made any progress on this? Or shall I take over it?

PR #10971, opened 2019-08-14
Switching sccache from local to Redis

General-Beck:
All the images which are used in parity-ethereum build process are now being published to our Gitilab docker container registry.
These images now are making use of `sccache` with Redis to avoid the data race condition when pruning old cache pieces.
Metadata was also edited to show the current addresses of the images.
[test](https://gitlab.parity.io/General-Beck/parity-ethereum/tree/sccache_redis_test)
 
Depends on https://github.com/paritytech/scripts/pull/91


.gitlab-ci.yml
```diff
@@ -41,22 +41,7 @@ variables:
...
-      then
-        echo "_____All crate-types:_____";
-        grep 'parse_arguments.*--crate-type' sccache_debug.log | sed -re 's/.*"--crate-type", "([^"]+)".*/\1/' | sort | uniq -c;
-        echo "_____Non-cacheable reasons:_____";
-        grep CannotCache sccache_debug.log | sed -re 's/.*CannotCache\((.+)\).*/\1/' | sort | uniq -c;
-      else
-        echo "_____No logs from sccache_____";
-        exit 0;
-      fi
+    - SCCACHE_REDIS=redis://:${REDIS}@172.17.0.1/0 sccache --start-server
```
dvdplm: Is it wise to use the explicit ip address here?

kirushik: That's a default gateway address of a Docker container-internal network (so in practice it's a mapping of host's `127.0.0.1` into a local subnet).
I guess it's safe enough, provided that we're not going to mess with Docker's networking too much in this setup.

As far as I understood from my chat with @General-Beck earlier today, there's no easy way to teach `sccache` to identify that IP by hostname, and there's no _non-hacky_ lookup flow to be used instead either.

.gitlab-ci.yml
```diff
@@ -37,26 +37,11 @@ variables:
...
-    - if test -e sccache_debug.log;
-      then
-        echo "_____All crate-types:_____";
-        grep 'parse_arguments.*--crate-type' sccache_debug.log | sed -re 's/.*"--crate-type", "([^"]+)".*/\1/' | sort | uniq -c;
-        echo "_____Non-cacheable reasons:_____";
-        grep CannotCache sccache_debug.log | sed -re 's/.*CannotCache\((.+)\).*/\1/' | sort | uniq -c;
-      else
-        echo "_____No logs from sccache_____";
-        exit 0;
-      fi
```
gabreal: so debugging will be disabled completely?

TriplEight: Yep, sccache debugging was actually helpful only once.
But what's important and bugs me, is that debug log level somehow hinders `sccache` to communicate with redis.
Also, had to remove `sccache -s` as it was failing the job because it couldn't get the info from Redis. I tried some ideas, but think that in order to do that, we should install sccache to CI1 and CI2 which is not really necessary, at least for now.

dvdplm:
@General-Beck what is the status of this? Do you need any help sorting out the tests?

General-Beck:
@dvdplm I do the latest checks and tests. I think tomorrow we will launch

PR #10972, opened 2019-08-14
Update to latest trie version.

cheme:
This PR update to latest trie crates (and parity-util-mem 0.2).
Those crates introduces some changes in codec:
- trait changes: a Layout trait defines that we uses extension node plus the hash and codec to use.
- codec changes: the partial slice used internally in the trie do not anymore contain a header or is encoded (this explains that the first byte is skipped in decoding of children slice and that two methods are added to encode the slices).


util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -16,26 +16,53 @@
...
+
+/// encode a partial value
+fn encode_partial<'a>(partial: Partial<'a>, is_leaf: bool) -> impl Iterator<Item = u8> + 'a {
+	let enc_type = if is_leaf {0x20} else {0};
+	let first = if (partial.0).0 > 0 {
+		0x10 + enc_type + (partial.0).1
+	} else {
+		enc_type
+	};
+	std::iter::once(first).chain(partial.1.iter().map(|v|*v))
```
ordian: Can it be implemented via `encode_partial_it` somehow, the code looks similar?

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -16,26 +16,53 @@
...
+}
+
+/// encode a partial value
+fn encode_partial_it<'a>(mut partial: impl Iterator<Item = u8> + 'a, odd: bool, is_leaf: bool) -> impl Iterator<Item = u8> + 'a {
```
ordian: `_iter` is a more rust-idiomatic suffix

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -107,28 +150,36 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
+		if let Some(value) = maybe_value {
 			stream.append(&&*value);
 		} else {
 			stream.append_empty_data();
 		}
 		stream.drain()
 	}
+
+	fn branch_node_nibbled(
+		_partial:	impl Iterator<Item = u8>,
```
ordian:
```suggestion
		_partial: impl Iterator<Item = u8>,
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -107,28 +150,36 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
+		_nb_nibble: usize,
+		_children: impl Iterator<Item = impl Borrow<Option<ChildReference<<KeccakHasher as Hasher>::Out>>>>,
+		_maybe_value: Option<&[u8]>) -> Vec<u8> {
+		unreachable!()
```
ordian: Could you add a comment inside `unreachable` why it is so:
```rust
unreachable!("you proof goes here")
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -71,32 +108,38 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
-		stream.append(&partial);
+		stream.append_iter(encode_partial(partial, true));
 		stream.append(&value);
 		stream.drain()
 	}
 
-	fn ext_node(partial: &[u8], child_ref: ChildReference<<KeccakHasher as Hasher>::Out>) -> Vec<u8> {
+	fn extension_node(
+		partial: impl Iterator<Item = u8>,
+		nb_nibble: usize,
```
ordian: could you apply the same renaming here as you did in substrate, e.g. `s/nb/number`, etc?

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -16,29 +16,63 @@
...
+
+/// Encode a partial value with an iterator as input.
+fn encode_partial_from_iterator_iter<'a>(mut partial: impl Iterator<Item = u8> + 'a, odd: bool, is_leaf: bool) -> impl Iterator<Item = u8> + 'a {
+	let first = if odd { partial.next().unwrap_or(0) } else { 0 }; 
+	encode_partial_inner_iter(first, partial, odd, is_leaf)
+}
+
+/// Encode a partial value with an iterator as input.
+fn encode_partial_inner_iter<'a>(
+	first: u8,
```
ordian:
```suggestion
	first_byte: u8,
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -16,29 +16,63 @@
...
+	partial_remaining: impl Iterator<Item = u8> + 'a,
+	odd: bool,
+	is_leaf: bool,
+) -> impl Iterator<Item = u8> + 'a {
+	let encoded_type = if is_leaf {0x20} else {0};
+	let first = if odd {
+		let first_byte = first;
```
ordian:
```suggestion
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -70,65 +113,81 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
 		}
 		stream.drain()
 	}
+
+	fn branch_node_nibbled(
+		_partial: impl Iterator<Item = u8>,
+		_number_nibble: usize,
+		_children: impl Iterator<Item = impl Borrow<Option<ChildReference<<KeccakHasher as Hasher>::Out>>>>,
+		_maybe_value: Option<&[u8]>) -> Vec<u8> {
+		unreachable!("This codec is only use with a trie Layout that uses extension node.")
```
dvdplm:
```suggestion
		unreachable!("This codec is only used with a trie Layout that uses extension nodes.")
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -70,65 +113,81 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
 			_ => Err(DecoderError::Custom("Rlp is not valid."))
 		}
 	}
+
 	fn try_decode_hash(data: &[u8]) -> Option<<KeccakHasher as Hasher>::Out> {
-		let r = Rlp::new(data);
-		if r.is_data() && r.size() == KeccakHasher::LENGTH {
-			Some(r.as_val().expect("Hash is the correct size; qed"))
+
+		if data.len() == KeccakHasher::LENGTH {
```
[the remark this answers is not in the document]
cheme: That is a change from upstream trie crate, the `try_decode_hash` is using directly the hash data, and we skip the header in `decode` function at 
`nodes[i] = Some(&value.as_raw()[1..]);`. (I will change this line, writing this comment make me realize there is  a cleanest way of doing this and without the implied malleability from skipping a byte).

The change is related to the fact that when trie uses this payload as an encoded children (`try_decode_hash` return none) trie code previously need to manage this byte.

So now this is done in `decode`. Thinking again that is not ideal and it would have been more direct to return either a hash or a new smaller slice from `try_decode` function.
Which is not perfect either (if the type of children could be encoded heading byte for instance).

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -70,65 +113,81 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
+			let mut r = <KeccakHasher as Hasher>::Out::default();
+			r.as_mut().copy_from_slice(data);
```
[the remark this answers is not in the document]
cheme: I think that could be performance improvement from refactoring a bit the codec trait (avoiding returning Vec in others codec function codec functions would also be a huge one).
But I do not know of any simple/immediate solution here.

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -47,19 +84,36 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
-				(slice, true) => Ok(Node::Leaf(slice, r.at(1)?.data()?)),
-				(slice, false) => Ok(Node::Extension(slice, r.at(1)?.as_raw())),
+			Prototype::List(2) => {
+				let enc_nibble = r.at(0)?.data()?;
+				let from_encoded = if enc_nibble.is_empty() {
+					(NibbleSlice::new(&[]), false)
+				} else {
+					// check leaf bit from header.
+					let is_leaf = enc_nibble[0] & 32 == 32;
+					// check from headre bit if odd partial (only a niblle of heder info)
```
dvdplm:
```suggestion
					// Check the header bit to see if we're dealing with an odd partial (only a nibble of header info)
```

util/patricia-trie-ethereum/src/rlp_node_codec.rs
```diff
@@ -47,19 +84,36 @@ impl NodeCodec<KeccakHasher> for RlpNodeCodec<KeccakHasher> {
...
+					// or even partial (a full byte to skip).
```
dvdplm:
```suggestion
					// or an even partial (skip a full byte).
```

PR #10957, opened 2019-08-09
Add support for Energy Web Foundation's new chains

ngyam:
- Support for --chain volta (ewf's new test network)
- Support for --chain ewc/energyweb (ewf's new production network)
- Remove support for Tobalaba (old deprecated test network)

PR #10863, opened 2019-07-08
Docker images renaming

TriplEight:
Changing the builders images names to have them standardized.
Please merge only after https://github.com/paritytech/scripts/pull/72 .
In order to build with the fresh images everyone should rebase their branches.

PR #10973, opened 2019-08-14
2.6.1-beta

s3krit:
* Add support for Energy Web Foundation's new chains (#10957)
* Kaspersky AV whitelisting (#10919)
* Avast whitelist script (#10900)
* Docker images renaming (#10863)
* Remove excessive warning (#10831)
* Allow --nat extip:your.host.here.org (#10830)
* When updating the client or when called from RPC, sleep should mean sleep (#10814)
* added new ropsten-bootnode and removed old one (#10794)
* ethkey no longer uses byteorder (#10786)
* Do not drop the peer with None difficulty (#10772)
* docs: Update Readme with TOC, Contributor Guideline. Update Cargo package descriptions (#10652)
*  [Cargo.lock] cargo update -p crossbeam-epoch (#10921)


miner/src/pool/replace.rs
```diff
@@ -71,6 +71,20 @@ where
 			let old_score = (old.priority(), old.gas_price());
 			let new_score = (new.priority(), new.gas_price());
 			if new_score > old_score {
+				// Check if this is a replacement transaction.
+				//
+				// With replacement transactions we can safely return `InsertNew` here, because
+				// we don't need to remove `old` (worst transaction in the pool) since `new` will replace
+			    // some other transaction in the pool so we will never go above limit anyway.
```
ordian: oh no, some spaces snuck into the master!

s3krit: 😱

PR #10976, opened 2019-08-15
[evmbin] fix compilation

dvdplm:
`evmbin` uses the `json-tests` feature from `ethcore` but after recent refactorings the needed helpers have moved.
Also update the crate to edition 2018.


ordian:
why patch labels, `evmbin` should compile on stable and beta, right?

dvdplm:
It was actually broken for me on master.

PR #10978, opened 2019-08-16
Extract spec to own crate

dvdplm:
Move the engine spec management code outside of `ethcore`. Nothing much to see here, it's mostly just reshuffling code and fixing imports.

Based on https://github.com/paritytech/parity-ethereum/pull/10966.


ethcore/executive-state/src/lib.rs
```diff
@@ -1542,7 +1542,7 @@ mod tests {
 	fn should_not_panic_on_state_diff_with_storage() {
 		let mut state = get_temp_state();
 		let a = Address::from_low_u64_be(0xa);
-		state.init_code(&a, b"abcdefg".to_vec()).unwrap();
+		state.init_code(&a, b"abcdefg".to_vec()).unwrap();;
```
ordian: `;;`? :smile:

dvdplm: I've added a ton of these in these PRs, I have no idea how it happened. :/

PR #10979, opened 2019-08-16
[ipfs] Convert to edition 2018

dvdplm:
Also: whitespace


ipfs/src/error.rs
```diff
@@ -15,7 +15,8 @@
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
 use {multihash, cid, http};
```
niklasad1: needless `use` because those are `crates` right?

PR #10980, opened 2019-08-18
Fix rlp decode for inline trie nodes.

cheme:
The test from #10972 last commit was incorrect, and fails on inline trie node.


util/patricia-trie-ethereum/src/lib.rs
```diff
@@ -145,3 +145,41 @@ pub type TrieFactory = trie::TrieFactory<Layout>;
 pub type TrieError = trie::TrieError<H256, DecoderError>;
 /// Convenience type alias for Keccak/Rlp flavoured trie results
 pub type Result<T> = trie::Result<T, H256, DecoderError>;
+
+#[test]
```
ordian: maybe put in `mod tests` with `#[cfg(test)]`?

PR #10981, opened 2019-08-19
Make ClientIoMessage generic over the Client

dvdplm:
The one big remaining issue in ethcore is the `ClientIoMessage` type which has an `Execute` variant that [embeds a `Client`](https://github.com/paritytech/parity-ethereum/pull/10981/files#diff-9a402a2cf87ab501d6a73278cfe12c73R38). This is a blocker for several things, e.g. making a freestanding `ethcore/light`, extracting `verification` etc.
This PR attempts to disentangle this type by making it generic and extracts `verification` from `ethcore`.

There are some questionable changes made here, in particular wrt to the traits used/modified. Careful review of these areas is much appreciated.

1. the [`ImportBlock` trait](https://github.com/paritytech/parity-ethereum/pull/10981/files#diff-de317cf28117c4c1c13bdeabf6196857R177) gains a method despite the `FIXME` that seems to indicate that the trait should not even exist
1. the new [`Tick` trait](https://github.com/paritytech/parity-ethereum/pull/10981/files#diff-de317cf28117c4c1c13bdeabf6196857R198)
1. hijacking the `SnapshotClient` trait (maybe the `SnapshotWriter` trait should be combined with this?)
1. ~~making the light client's [fetcher type param `'static`](https://github.com/paritytech/parity-ethereum/pull/10981/files#diff-223b03b977353484865ff436905fac0cR165) – better way?~~

After this PR we can build the eth light client without depending on `Client`.

Based on https://github.com/paritytech/parity-ethereum/pull/10978.

I have synced mainnet with this branch.


ethcore/verification/src/lib.rs
```diff
@@ -18,6 +18,8 @@
 
 use call_contract::CallContract;
 use client_traits::BlockInfo;
+// The MallocSizeOf derive looks for this in the root
+use parity_util_mem as malloc_size_of;
```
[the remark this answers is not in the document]
dvdplm: I think I prefer to keep this explicit for the time being.

ethcore/client-traits/src/io_message.rs
```diff
@@ -16,13 +16,12 @@
```
niklasad1: it is a bit odd to have these types in `client-traits`, is there a better location for it?

ethcore/client-traits/src/lib.rs
```diff
@@ -173,6 +177,10 @@ pub trait EngineClient: Sync + Send + ChainInfo {
 pub trait ImportBlock {
 	/// Import a block into the blockchain.
 	fn import_block(&self, block: Unverified) -> EthcoreResult<H256>;
+
+	// todo[dvdplm]: probably a bad place for this method
```
niklasad1: seems reasonable to me at least or maybe introduce another trait `ImportVerifiedBlock`

ethcore/client-traits/src/lib.rs
```diff
@@ -173,6 +177,10 @@ pub trait EngineClient: Sync + Send + ChainInfo {
...
+	/// Triggered by a message from a block queue when the block is ready for insertion
```
niklasad1: perhaps document that it returns the number of imported blocks

additionally, the type that implements it needs access to the block queue but it is implied by the method signature I guess.

ethcore/client-traits/src/lib.rs
```diff
@@ -187,6 +195,13 @@ pub trait IoClient: Sync + Send {
 	fn queue_consensus_message(&self, message: Bytes);
 }
 
+pub trait Tick {
```
niklasad1: nitpick, add trait docs

ethcore/client-traits/src/lib.rs
```diff
@@ -423,3 +441,36 @@ pub trait ProvingBlockChainClient: BlockChainClient {
...
+
+/// External database restoration handler
+pub trait DatabaseRestore: Send + Sync {
+	/// Restart with a new backend. Takes ownership of passed database and moves it to a new location.
+	fn restore_db(&self, new_db: &str) -> Result<(), EthcoreError>;
+}
+
+/// Trait alias for the Client traits used when taking snapshots
+pub trait SnapshotClient: BlockChainClient + BlockInfo + DatabaseRestore + BlockChainReset {
+	fn take_snapshot<W: SnapshotWriter + Send>(
```
niklasad1: nitpick, docs

ethcore/src/client/client.rs
```diff
@@ -2581,7 +2495,62 @@ impl ProvingBlockChainClient for Client {
 	}
 }
 
-impl SnapshotClient for Client {}
+impl SnapshotClient for Client {
+	/// Take a snapshot at the given block.
```
niklasad1: move the documentation to the trait definition

ethcore/client-traits/src/lib.rs
```diff
@@ -423,3 +441,36 @@ pub trait ProvingBlockChainClient: BlockChainClient {
...
 	fn epoch_signal(&self, hash: H256) -> Option<Vec<u8>>;
 }
+
+/// External database restoration handler
+pub trait DatabaseRestore: Send + Sync {
+	/// Restart with a new backend. Takes ownership of passed database and moves it to a new location.
+	fn restore_db(&self, new_db: &str) -> Result<(), EthcoreError>;
+}
+
+/// Trait alias for the Client traits used when taking snapshots
```
niklasad1: this is not a trait alias because it has a method?

ethcore/client-traits/src/lib.rs
dvdplm: The reason I was hesitant is the comment above saying `// FIXME Why these methods belong to BlockChainClient and not MiningBlockChainClient?` – I don't really know who/when left it and I think it's not relevant any more (even before my changes) but I'm also not 100% sure.

ethcore/client-traits/src/io_message.rs
dvdplm: You are not wrong. :)

I could move it to `common-types`. Would that be better?

ethcore/client-traits/src/lib.rs
niklasad1: Right, the FIXME is a bit vague.  I think it refers to [this](https://github.com/paritytech/parity-ethereum/blob/0dfe3c9ffacfd4b3f1d6bf46031cb26ff2244b2e/ethcore/client-traits/src/lib.rs#L212-#L213).

`BlockChainClient` depends on `ImportBlock` thus the FIXME has nothing to do with `ImportBlock` itself but that it is a super trait to `BlockChainClient`. I suggest we remove the FIXME.

ethcore/light/src/client/service.rs
```diff
@@ -58,15 +60,15 @@ impl fmt::Display for Error {
 }
 
 /// Light client service.
-pub struct Service<T> {
+pub struct Service<T: 'static> {
 	client: Arc<Client<T>>,
-	io_service: IoService<ClientIoMessage>,
+	io_service: IoService<ClientIoMessage<()>>,
```
niklasad1: why is this `()`?

does it work because we only care about `ClientIoMessage::BlockVerified` for the light client?

dvdplm: Great question. So before this PR `light` depends on `ethcore` essentially only for `ClientIoMessage`, and then only because it has a `Execute` variant that embeds a `Callback` which is a `Fn` that takes an `ethcore::Client` as a param… So the dependency was actually not "real" but merely ceremonial, to satisfy the types as they were set up.
Once I made the `Callback` generic over the Client (and all the other types using it) @ordian spotted that there's actually not necessary to use anything other than `()` in `light`.

Now, we all agree that `ClientIoMessage<()>` looks plenty odd and hopefully someone will come along and refactor this further. ;)
In particular I know that @ngotchac is working on changes to the IO code that may, as a desired side-effect, result in a better message passing infrastructure, one that does not need a `Client` (or whatever stand-in we use).

niklasad1: Ok, I see.
> In particular I know that @ngotchac is working on changes to the IO code that may, as a desired side-effect, result in a better message passing infrastructure, one that does not need a Client (or whatever stand-in we use).

Cool :+1:

Issue #10982, opened 2019-08-21
EIP-1344: Add ChainID opcode


dvdplm:
Implement [EIP-1344](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1344.md).

PR #10983, opened 2019-08-21
EIP-1344 Add CHAINID op-code

dvdplm:
Part of https://github.com/paritytech/parity-ethereum/issues/10770

Closes https://github.com/paritytech/parity-ethereum/issues/10982

Based on https://github.com/paritytech/parity-ethereum/pull/10978


ethcore/machine/src/externalities.rs
```diff
@@ -479,6 +479,7 @@ mod tests {
 			last_hashes: Arc::new(vec![]),
 			gas_used: 0.into(),
 			gas_limit: 0.into(),
+			chain_id: 0.into(),
```
sorpaas: We can avoid changing the structure altogether. `Externalities` contains a reference to `Machine`, which allows retrieving the complete config parameter via `Machine::params` already. What we may be able to do here to simplify thing is to create a new function in externalities, and then make chain id opcode call that externalities.

ethcore/wasm/src/tests.rs
```diff
@@ -801,6 +801,7 @@ fn externs() {
 			gas_limit: 0x777777777777u64.into(),
 			last_hashes: Default::default(),
 			gas_used: 0.into(),
+			chain_id: 0.into(), // todo[dvdplm]: check if this is ok
```
sorpaas: This should be okay. For all tests we can currently just assume all chain id opcode returns zero, until new test format is defined.

ethcore/machine/src/externalities.rs
dvdplm: You mean extending [`vm::Ext` trait](https://github.com/paritytech/parity-ethereum/blob/141f6a047ece33cf54bea130707553a075d152b2/ethcore/vm/src/ext.rs#L67) to include a `chain_id()` method and then implement it on [`machine::Externalities`](https://github.com/paritytech/parity-ethereum/blob/bcd43d6b67b62507ca8429f0e1f5a63ae735eeef/ethcore/machine/src/externalities.rs#L131) using the ref to `Machine`?

sorpaas: Yes exactly.

grbIzl:
(Stupid question) Do we need to add the corresponding instruction into wasm as well?

sorpaas:
@grbIzl Let's not complicate it. Unless there're needs for pWasm, the only reason to implement it is for Istanbul. There're many other ways one can have the functionality of `CHAIN_ID` without even needing a hard fork.

sorpaas:
@dvdplm Can you do a rebase?

ethcore/evm/src/lib.rs
```diff
@@ -33,6 +33,8 @@ extern crate log;
 
 #[cfg(test)]
 extern crate rustc_hex;
```
niklasad1: `rustc_hex` is not used anymore, remove?!

ethcore/evm/src/tests.rs
```diff
@@ -19,17 +19,23 @@ use std::str::FromStr;
 use std::hash::Hash;
 use std::sync::Arc;
 use std::collections::{HashMap, HashSet};
-use rustc_hex::FromHex;
+//use rustc_hex::FromHex;
```
niklasad1: remove?

dvdplm:
@sorpaas rebased, should be easier to review now.

ethcore/vm/src/schedule.rs
```diff
@@ -291,6 +294,7 @@ impl Schedule {
 	/// Schedule for the Istanbul fork of the Ethereum main net.
 	pub fn new_istanbul() -> Schedule {
 		let mut schedule = Self::new_constantinople();
+        schedule.have_chain_id = true;
```
ordian:
```suggestion
		schedule.have_chain_id = true;
```

dvdplm: This is the dumb OSX merge conflict review tool I'm using. It ignores whitespace settings. :/

ethcore/evm/src/tests.rs
```diff
@@ -19,17 +19,22 @@ use std::str::FromStr;
...
 
 evm_test!{test_add: test_add_int}
 fn test_add(factory: super::Factory) {
 	let address = Address::from_str("0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6").unwrap();
-	let code = "7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff01600055".from_hex().unwrap();
+	// 7f       PUSH32 ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
+	// 7f       PUSH32 ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff ff
+	// 01       ADD
+	// 60 00    PUSH 0
+	// 55       SSTORE
```
ordian: Would be nice to have a macro/builder for bytecode instead of having comments, but that's outside of the scope of the PR.

dvdplm: In the meantime https://ethervm.io/decompile can be helpful.

ethcore/machine/src/externalities.rs
```diff
@@ -416,6 +416,10 @@ impl<'a, T: 'a, V: 'a, B: 'a> Ext for Externalities<'a, T, V, B>
 		self.env_info
 	}
 
+	fn chain_id(&self) -> u64 {
+		self.machine.params().chain_id
```
sorpaas: Just a note that this has the behavior that if EIP-155 is not enabled on a chain, its `CHAIN_ID` will instead return the network id. This is never specified in EIP-1344, but looks like acceptable behavior for now.

PR #10987, opened 2019-08-23
EIP 2028: transaction gas lowered from 68 to 16

### LINKED ISSUE #10986: EIP-2028: Calldata gas cost reduction
Reduce the gas cost of Calldata (GTXDATANONZERO) from its current value of 68 gas per byte to 16 gas per byte. EIP: https://eips.ethereum.org/EIPS/eip-2028

Part of #10770

dvdplm:
Closes #10986

Note: there are no tests in this PR. Not sure what the process is for adding state tests (or whatever it is we need).

### DIFF
--- a/ethcore/evm/src/tests.rs
+++ b/ethcore/evm/src/tests.rs
@@ -262,7 +262,6 @@ fn test_calldataload(factory: super::Factory) {
 
 	assert_eq!(gas_left, U256::from(79_991));
 	assert_store(&ext, 0, "23ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff23");
-
 }
 
 evm_test!{test_author: test_author_int}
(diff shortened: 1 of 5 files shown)


sorpaas:
Martin's note on transaction pool. Haven't entirely checked whether this is relevant to us (but most likely it is):

> One note, which I already made for aleth, but I see now is applicable for other nodes aswell: Regarding implementing calldata reduction, there's a minor snag with transaction pools -- if they start accepting istanbul-priced txs too early, that's a DoS vector. So in geth we made the pools stateful, and when istanbul hits we flip the switch to use another formula to calculate the intrinsic gas (and thus validity of) a transaction.

dvdplm:
@sorpaas what is "intrinsic gas"?

sorpaas:
@dvdplm it refers to the gas deducted before code execution starts.

PR #10991, opened 2019-08-27
Fix compilation on recent nightlies

dvdplm:
For reasons unknown (to me) recent rustc nightlies can't find the `format!`-macro in `core` when there is an import named `core`. This PR removes the package rename and uses the full `jsonrpc_core` crate name instead.


niklasad1:
yeah, I had the same problem yesterday. 

I think it some name conflict with [Rust core library](https://doc.rust-lang.org/stable/core/index.html)

Issue #10990, opened 2019-08-26
EIP 1884 repricing trie-size dependent opcodes


dvdplm:
Implement [EIP 1884](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1884.md).

Part of https://github.com/paritytech/parity-ethereum/issues/10770

PR #10992, opened 2019-08-27
EIP 1884 Re-pricing of trie-size dependent operations

dvdplm:
Closes https://github.com/paritytech/parity-ethereum/issues/10990

Based on https://github.com/paritytech/parity-ethereum/pull/10983


ethcore/evm/src/interpreter/mod.rs
```diff
@@ -843,6 +845,12 @@ impl<Cost: CostType> Interpreter<Cost> {
 			instructions::GASLIMIT => {
 				self.stack.push(ext.env_info().gas_limit.clone());
 			},
+			instructions::CHAINID => {
+				self.stack.push(ext.chain_id().into())
+			},
+			instructions::SELFBALANCE => {
+				self.stack.push(ext.balance(&self.params.origin)?);
```
dvdplm: @sorpaas is `self.params.origin` what I need here, i.e. is that what corresponds to `this`?

PR #10993, opened 2019-08-27
fix compile warnings

niklasad1:
This PR eliminates a bunch of compiler warnings most of them were `bare-trait-object` i.e., replaced by `dyn trait`.

Tried `cargo fix` first but then I fixed the rest of them manually


rpc/src/lib.rs
```diff
@@ -85,7 +85,7 @@ extern crate parity_version as version;
 extern crate eip_712;
 extern crate rlp;
 extern crate account_state;
-extern crate spec;
```
ordian:
```rust
error[E0432]: unresolved imports `spec`, `spec`
  --> rpc/src/v1/tests/eth.rs:25:5
   |
25 | use spec::{Genesis, Spec, self};
   |     ^^^^                  ^^^^ no `spec` in the root
   |     |
   |     unresolved import
   |     help: a similar path exists: `ethjson::spec`
```
I guess this should be `#[cfg(test)]`

PR #10995, opened 2019-08-27
Fix parallel transactions race-condition

### LINKED ISSUE #9660: --config=dev doesn't seal properly when transactions are sent in parallel

- **Parity Ethereum version**: v2.0.6-stable-7658d22-20180921
- **Operating system**: Linux
- **Installation**: docker image
- **Fully synchronized**: yes
- **Network**: dev config.
- **Restarted**: yes
- **Used Options**: `--jsonrpc-cors "*" --jsonrpc-apis all --jsonrpc-hosts=all --base-path=/data/ --config dev --unlock 0x00a329c0648769a73afac7f9381e08fb43dbea72 --password=/pw --jsonrpc-interface "0.0.0.0"`

I have a client sending in transactions to parity via the HTTP RPC method `eth_sendRawTransaction`. The pseudo code for the logs later is roughly:

```
logger.debug('Sending tx: {hash}'))
web3.eth.sendRawTransaction(rawTxn)
logger.debug('Sent tx: {hash}')
```
Also the client is listening for events using filters. I send in the transaction at nearly the same time. Because I use the instant seal engine I would expect parity to instantly produce blocks for the two transaction, either by including them in one block or by producing two blocks shortly after each other. 
What happens is that the first transaction gets included, but the second transaction hangs somewhere and gets included only if another transaction comes in. 
You can see that in the logs I collected. 

```
client        | 2018-09-27 14:36:00,470[DEBUG] Sending tx: 0xc838d956c720eae34b04752a8b2bc361762512d0fdb983fde7bebabfa06b63c5
client        | 2018-09-27 14:36:00,472[DEBUG] Sending tx: 0x637e268e8da6ff7403ad2c03ce723afd619c174349077c04163229202f314036
parity       | 2018-09-27 14:36:00 UTC Imported #26 0x051e…ae52 (1 txs, 0.06 Mgas, 0 ms, 0.69 KiB) + another 1 block(s) containing 1 tx(s)
parity       | 2018-09-27 14:36:00 UTC Transaction mined (hash 0xc838d956c720eae34b04752a8b2bc361762512d0fdb983fde7bebabfa06b63c5)
client        | 2018-09-27 14:36:00,479[DEBUG] Sent tx: 0xc838d956c720eae34b04752a8b2bc361762512d0fdb983fde7bebabfa06b63c5
client        | 2018-09-27 14:36:00,481[DEBUG] Sent tx: 0x637e268e8da6ff7403ad2c03ce723afd619c174349077c04163229202f314036
client        | 2018-09-27 14:36:00,515[DEBUG] new entries for filter Filter for 0x5: ['0xc838d956c720eae34b04752a8b2bc361762512d0fdb983fde7bebabfa06b63c5']
parity       | 2018-09-27 14:36:03 UTC Imported #27 0x341b…37f8 (2 txs, 0.34 Mgas, 0 ms, 0.79 KiB) + another 1 block(s) containing 2 tx(s)
parity       | 2018-09-27 14:36:03 UTC Transaction mined (hash 0x1a757833012244c503bdd74e16b91be46ab8cde50e97a4622e3dda46373d7488)
parity       | 2018-09-27 14:36:03 UTC Transaction mined (hash 0x637e268e8da6ff7403ad2c03ce723afd619c174349077c04163229202f314036)
parity       | 2018-09-27 14:36:03 UTC Transaction mined (hash 0x2d4e9421e1d9d60c6c371c7561235200aad8f8dca582b39f66ca01b7680f4c8c)
client        | 2018-09-27 14:36:03,321[DEBUG] new entries for filter Filter for 0x2: ['0x637e268e8da6ff7403ad2c03ce723afd619c174349077c04163229202f314036']
```

As a comparison, this is how it looks if I send in the transaction after each other:

```
client        | 2018-09-27 14:44:21,384[DEBUG] Sending tx: 0x7f0e1baac164517fbabbea8cad2d47098a2d61cc5a769a87dc65164a341c814a
parity       | 2018-09-27 14:44:21 UTC Transaction mined (hash 0x7f0e1baac164517fbabbea8cad2d47098a2d61cc5a769a87dc65164a341c814a)
parity       | 2018-09-27 14:44:21 UTC Imported #26 0xa2c7…6569 (1 txs, 0.06 Mgas, 0 ms, 0.69 KiB) + another 1 block(s) containing 1 tx(s)
client        | 2018-09-27 14:44:21,391[DEBUG] Sent tx: 0x7f0e1baac164517fbabbea8cad2d47098a2d61cc5a769a87dc65164a341c814a
client        | 2018-09-27 14:44:21,392[DEBUG] Sending tx: 0x3bed1cc2f1dfa6308ab6434c80faf75016be10fafe98e2236c55a30a939c89d0
parity       | 2018-09-27 14:44:21 UTC Transaction mined (hash 0x3bed1cc2f1dfa6308ab6434c80faf75016be10fafe98e2236c55a30a939c89d0)
client        | 2018-09-27 14:44:21,400[DEBUG] Sent tx: 0x3bed1cc2f1dfa6308ab6434c80faf75016be10fafe98e2236c55a30a939c89d0
client        | 2018-09-27 14:44:21,567[DEBUG] new entries for filter Filter for 0x2: ['0x3bed1cc2f1dfa6308ab6434c80faf75016be10fafe98e2236c55a30a939c89d0']
client        | 2018-09-27 14:44:21,729[DEBUG] new entries for filter Filter for 0x5: ['0x7f0e1baac164517fbabbea8cad2d47098a2d61cc5a769a87dc65164a341c814a']
```

This was an issue before, see https://github.com/paritytech/parity-ethereum/issues/2589, https://github.com/paritytech/parity-ethereum/issues/4494, and it seems it was not fully fixed.

ngotchac:
Closes #9660 

This PR should fix the race-condition issue that happens when sending multiple parallel transactions to a node running a dev chain.
To reproduce the issue, simply run a node with:
```
> parity --config dev --tx-queue-size 60000 --no-persistent-txqueue
```
and run this NodeJS script https://gist.github.com/ngotchac/8b85622f6396001f243fa2323184104d which sends a random number of transactions in parallel until it hangs.

The issue seems to be that the pending transactions queue was not totally cleared on culled transaction, and multiple blocks at the same height could be produced by the _instant sealing_ engine.

With this PR, we ensure that only one block per height is produced, and to try to seal a new block after importing a new (internal) _best block_ if there are some pending transactions.


ethcore/src/miner/mod.rs
```diff
@@ -84,7 +84,7 @@ pub trait MinerService : Send + Sync {
 
 	/// Update current pending block
 	fn update_sealing<C>(&self, chain: &C)
-		where C: BlockChain + CallContract + BlockProducer + SealedBlockImporter + Nonce + Sync;
+		where C: BlockChainClient;
```
tomusdrw: I think originally this was the bound, but we have moved to more relaxed traits to avoid tight coupling with `Client`. I'd say it would be good to keep it this way.

miner/src/pool/queue.rs
```diff
@@ -458,6 +458,7 @@ impl TransactionQueue {
 			let state_readiness = ready::State::new(client.clone(), stale_id, nonce_cap);
 			removed += self.pool.write().cull(Some(chunk), state_readiness);
 		}
+		self.cached_pending.write().clear();
```
tomusdrw: This is not strictly needed:
1. The cache will be cleared if you request for a different block number
2. The cache will be cleared if you add any new transactions succesfully
3. The old values are still valid (for given (historical) `block_number`), so for instance we can still use it for propagation.

That said, I don't think it makes any difference.

ethcore/src/miner/miner.rs
```diff
@@ -876,6 +877,13 @@ impl Miner {
 			SealingState::NotReady => { self.maybe_enable_sealing(); },
 		}
 	}
+
+	/// Call `update_sealing` if needed
+	pub fn maybe_update_sealing<C: miner::BlockChainClient>(&self, chain: &C) {
+		if self.transaction_queue.has_local_pending_transactions() {
```
tomusdrw: I'm not sure about this. The implementation is very specific for `NullEngine` (i.e. checks local transactions), while we have plenty of other conditions that affect when to seal. Why can't we just call `update_sealing` instead?

miner/src/pool/queue.rs
ngotchac: Actually, I still think it's needed. I've seen this happening (with one sender):

1. A block is created and imported into the chain, which includes transaction w/ nonce N.
2. A transaction is sent with nonce N+1, which triggers a new block creation.
3. The new block tries to import pending transactions, but the transaction queue still includes tx w/ nonce N. As the previous block has been imported, the tx w/ nonce N cannot be imported, and so the new pending block has 0 transactions.
4. The transaction queue is culled, which leaves tx w/ nonce N + 1, which triggers a new block, but at the same height as the previous one, so the cache has already been created with 0 transaction.

ngotchac: I think if we changed the _transaction pool_ [here](https://github.com/paritytech/parity-common/blob/master/transaction-pool/src/pool.rs#L626) to take into account the case `Readiness::Stale` ([see here](https://github.com/paritytech/parity-ethereum/blob/master/miner/src/pool/ready.rs#L96)) we could fix this issue. What do you think @tomusdrw ?

tomusdrw: @ngotchac Indeed! I was sure that we just skip `Stale` transactions in the `PendingIterator`, but it seems we don't. Definitely [here](https://github.com/paritytech/parity-common/blob/master/transaction-pool/src/pool.rs#L626) we should just check next-best in case we get `Stale`.

ethcore/engines/instant-seal/src/lib.rs
```diff
@@ -80,10 +78,10 @@ impl Engine for InstantSeal {
 		if block.transactions.is_empty() {
 			Seal::None
 		} else {
-			let mut sealed_blocks = self.sealed_blocks.lock();
 			let block_number = block.header.number();
-			if !sealed_blocks.contains(&block_number) {
-				sealed_blocks.insert(block_number);
+			let last_sealed_block = self.last_sealed_block.load(Ordering::SeqCst);
+			if block_number > last_sealed_block {
+				self.last_sealed_block.store(block_number, Ordering::SeqCst);
```
tomusdrw: We have a race condition here. We should do `compare_and_swap` and check if we were the first ones to update the block - if not we should return `Seal::None`.

ngotchac: Yeah, it's too bad there isn't `fetch_max` in stable yet... Wouldn't it make more sense to use a Mutex/RwLock then?

tomusdrw: Actually `compare_and_swap` is just fine - we just don't create a new block if we know that some other thread already did. `fetch_max` would just make it in one atomic call vs two (if it actually translates to an atomic call), but it's negligible.

ngotchac:
Waiting for https://github.com/paritytech/parity-common/pull/218 to be merged.

miner/src/pool/queue.rs
ngotchac: cf. https://github.com/paritytech/parity-common/pull/218

ethcore/src/miner/miner.rs
```diff
@@ -876,6 +877,14 @@ impl Miner {
 			SealingState::NotReady => { self.maybe_enable_sealing(); },
 		}
 	}
+
+	/// Call `update_sealing` if needed
```
dvdplm: Maybe document this a bit more thoroughly: what are the conditions used to determine "needed" etc?

ethcore/src/miner/miner.rs
```diff
@@ -876,6 +877,14 @@ impl Miner {
...
+	pub fn maybe_update_sealing<C: miner::BlockChainClient>(&self, chain: &C) {
+		use miner::MinerService;
```
dvdplm: Please put on top.

ethcore/src/miner/miner.rs
```diff
@@ -1429,15 +1438,22 @@ impl miner::MinerService for Miner {
 						service_transaction_checker.as_ref(),
 					);
 					queue.cull(client);
+					if is_internal_import {
```
dvdplm: For my education, what is an "internal" import in this context?

ethcore/engines/instant-seal/Cargo.toml
```diff
@@ -13,6 +13,7 @@ ethjson = { path = "../../../json" }
 ethereum-types = "0.6.0"
 keccak-hash = "0.2.0"
 machine = { path = "../../machine" }
+parking_lot = "0.8"
```
ordian:
```suggestion
```

Cargo.lock
```diff
@@ -2128,6 +2128,7 @@ dependencies = [
  "ethjson 0.1.0",
  "keccak-hash 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "machine 0.1.0",
+ "parking_lot 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
```
ordian:
```suggestion
```

ethcore/src/miner/miner.rs
ngotchac: I think here it's a block that has been created after having sent transactions to your node, so triggered by a local RPC call rather than external transactions through the DevP2P network.

dvdplm: Should it be named `is_local_import` you think?

dvdplm:
> Waiting for [paritytech/parity-common#218](https://github.com/paritytech/parity-common/pull/218) to be merged.

Do you need a release&publish of this?

ethcore/src/miner/miner.rs
ngotchac: Actually it's for import blocks that have been sealed internally, sorry. So I think we should keep it that way.

ordian:
Awesome work! Race conditions are hard to debug.

ethcore/src/miner/miner.rs
tomusdrw: I think this is still valid.

ngotchac: Ah sorry, I thought you were talking about the next line, where I first used `prepare_and_update_sealing`. I just updated it.

tomusdrw: I guess then there is no point in adding `maybe_update_sealing` to the traits and just call `update_sealing` directly?

PR #10999, opened 2019-08-28
Convert `std::test` benchmarks to use Criterion

### LINKED ISSUE #10975: Convert all benchmarks to criterion
Criterion works with rust stable, meaning we can `cargo check --all --benches` in our CI to prevent compilation regressions.

tobz1000:
Fixes #10975.

I've replaced the `ethcore` bench check with a check for all benches, hopefully as desired in the issue.

Query: I've stuck to 2015-style imports in the modified source files, to minimise changes. Is this the correct style, or should modified files be updated to 2018-style imports?

Also modified `ethash` manifest so that benches should check okay with `cargo check --all --benches`.


ethash/Cargo.toml
```diff
@@ -29,7 +29,9 @@ bench = []
 [[bench]]
 name = "basic"
 harness = false
+required-features = ["bench"]
 
 [[bench]]
 name = "progpow"
 harness = false
+required-features = ["bench"]
```
tobz1000: Looking again, this probably isn't the right thing to do, as it will just skip the benchmark. A compile error might be preferable.

tobz1000:
[clabot:check]

ethash/Cargo.toml
ordian: Yeah, skipping benchmarks is not what we're trying to achieve here.
We can check ethash benches separately:
```
cargo check --all --benches --exclude ethash
```
and then
```
pushd ethash && cargo check --benches --features=bench && popd
```

evmbin/benches/mod.rs
```diff
@@ -20,72 +20,80 @@
 //! rustup run nightly cargo bench
```
ordian:
```suggestion
//! cargo bench
```

ordian:
merge with master should fix cargo audit

evmbin/Cargo.toml
```diff
@@ -29,7 +29,12 @@ trace = { path = "../ethcore/trace" }
 vm = { path = "../ethcore/vm" }
 
 [dev-dependencies]
+criterion = "0.2"
```
ordian: let's upgrade to 0.3 here and elsewhere

PR #11006, opened 2019-08-29
v2.5.7 stable

s3krit:
* check mem diff (#11002)


ethcore/src/trace/executive_tracer.rs
```diff
@@ -18,6 +18,7 @@
 
 use ethereum_types::{U256, Address};
 use vm::{Error as VmError, ActionParams};
+use log::{debug, warn};
```
dvdplm: Is `debug` used?

ordian: yes, but `stable` branch uses 2015 edition with `#[macro_use] extern crate` style imports

Issue #10998, opened 2019-08-28
EIP 1108: Reduce alt_bn128 precompile gas costs


niklasad1:
Tracking issue for https://eips.ethereum.org/EIPS/eip-1108

https://github.com/paritytech/parity-ethereum/issues/10770

PR #11008, opened 2019-08-29
EIP 1108: Reduce alt_bn128 precompile gas costs

niklasad1:
Attempt to close https://github.com/paritytech/parity-ethereum/issues/10998 and implement https://eips.ethereum.org/EIPS/eip-1108

Warning, this is not an elegant PR because `ethcore/builtins` reads the values directly from JSON and I tried to implement it as best effort.

If this is good enough, I will update the `chain spec` accordingly


ethcore/builtin/src/lib.rs
```diff
@@ -186,16 +218,30 @@ impl From<ethjson::spec::Builtin> for Builtin {
...
+						base: pricer.eip1108_transition_base,
+						pair: pricer.eip1108_transition_pair,
+					},
+				})
+			}
+			ethjson::spec::Pricing::AltBn128ConstOperations(pricer) => {
+				Box::new(AltBn128ConstOperations {
+						price: pricer.price,
+						eip1108_transition_price: pricer.eip1108_transition_price,
+						eip1108_transition_at: b.eip1108_transition.map_or(u64::max_value(), Into::into)
```
niklasad1: deny-by-default

ethcore/builtin/src/lib.rs
```diff
@@ -186,16 +218,30 @@ impl From<ethjson::spec::Builtin> for Builtin {
 			}
 			ethjson::spec::Pricing::AltBn128Pairing(pricer) => {
 				Box::new(AltBn128PairingPricer {
-					base: pricer.base,
-					pair: pricer.pair,
+					price: AltBn128PairingPrice {
+						base: pricer.base,
+						pair: pricer.pair,
+					},
+					eip1108_transition_at: b.eip1108_transition.map_or(u64::max_value(), Into::into),
```
niklasad1: deny-by-default

json/src/spec/builtin.rs
```diff
@@ -44,6 +54,10 @@ pub struct AltBn128Pairing {
 	pub base: usize,
 	/// Price per point pair.
 	pub pair: usize,
+	/// EIP 1108 transition base price
+	pub eip1108_transition_base: usize,
+	/// EIP 1108 transition price per point,
```
dvdplm: Did you mean?
```suggestion
	/// EIP 1108 transition price per point pair
```

dvdplm:
> If this is good enough, I will update the `chain spec` accordingly

Can you elaborate on this? What further changes to the chainspec do we need?

As for "good enough" I'm interested in hearing if Wei has a better idea. I think juggling these transitions is always a bit inelegant but maybe he has a better approach.

@niklasad1 how would you refactor the [`Pricer` trait](https://github.com/paritytech/parity-ethereum/pull/11008/files#diff-e23044c0bf96889bae2af5f7e449f3e2R45) to avoid the extra param?

sorpaas:
@dvdplm A better approach I can think of maybe something like this: we define a new type `builtinMulti` (as a field in `json::Account`). It takes a hashmap, where keys are block number, and values are `Builtin`. We then activate/deactivate builtins based on the block number. The resulting config will look something like this:

```
"accounts": {
  "0000000000000000000000000000000000000008": { "builtinMulti": {
    "0": { "name": "alt_bn128_pairing", "activate_at": 2000000, "pricing": { "alt_bn128_pairing": { "base": 100000, "pair": 80000 } } } },
    "12345678": { "name": "alt_bn128_pairing", "activate_at": 2000000, "pricing": { "alt_bn128_pairing": { "base": 900, "pair": 1000 } } } },
  }
}
```

However, this will probably take some time to refactor, and given our time constraints, I think @niklasad1's approach should be good enough!

ethcore/types/src/engines/params.rs
```diff
@@ -31,8 +31,8 @@ const MAX_TRANSACTION_SIZE: usize = 300 * 1024;
 ///
 /// we define a "bugfix" hard fork as any hard fork which
 /// you would put on-by-default in a new chain.
-// NOTE [dvdplm]: `Clone` is needed only for tests.
-#[derive(Debug, PartialEq, Default, Clone)]
+#[cfg_attr(test, derive(Clone))]
```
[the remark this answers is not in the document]
niklasad1: Ok, it should be possible to do `#[cfg_attr(any(test, feature = "test-helpers"), derive(Clone))]` then

ethcore/builtin/src/lib.rs
```diff
@@ -477,7 +523,7 @@ impl Implementation for Bn128Pairing {
 	///     - any of even points does not belong to the twisted bn128 curve over the field F_p^2 = F_p[i] / (i^2 + 1)
 	fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), &'static str> {
 		if input.len() % 192 != 0 {
-			return Err("Invalid input length, must be multiple of 192 (3 * (32*2))".into())
+			return Err("Invalid input length, must be multiple of 192 (3 * (32*2))")
```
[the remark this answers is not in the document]
niklasad1: That won't work because it returns `&'static str` and changing it to `&format` doesn't work either. Ok to ignore?

niklasad1:
@dvdplm 

> Can you elaborate on this? What further changes to the chainspec do we need?

We need to update all other chain specs with the `eip1108_transition params` 

> As for "good enough" I'm interested in hearing if Wei has a better idea. I think juggling these transitions is always a bit inelegant but maybe he has a better approach.

Yes, it is not pretty I agree

> how would you refactor the Pricer trait to avoid the extra param?

I think it should be possible to revert it completely if we could handle it within `JSON (builtinMulti as Wei suggested)`. Otherwise, I have no good idea without reworking the entire thing.

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
dvdplm: Annoying but yes, let's ignore.

dvdplm:
> update all other chain specs with the `eip1108_transition params`

Right, of course. Ok, overall lgtm. Holler when you're done with this and I'll give it a once-over again. :)

niklasad1:
I'm done 👍

ethcore/builtin/src/lib.rs
```diff
@@ -38,9 +38,11 @@ trait Implementation: Send + Sync {
 }
 
 /// A gas pricing scheme for built-in contracts.
+// NOTE(niklasad1): maybe refactor this trait because only `bn128_operations` are using
```
dvdplm: Please add an issue and include a link here.

dvdplm:
@sorpaas looks good to you? Let's merge this! :)

holiman:
Wow, nice job on the PR number; 11008 for 1108!

dvdplm:
Glad someone finally noticed! ;)

phahulin:
In future releases it might be a good idea to provide instructions about required updates in chainspec or other breaking changes in Changelog, for those who run custom networks

dvdplm:
@phahulin I agree, we did not handle that optimally. Apologies. OTOH it's non-trivial to keep track of exactly what changes affect which (advanced) users; do you have any concrete suggestions for us, other than writing more detailed changelogs?

PR #11009, opened 2019-08-29
EIP-152: Blake2 F compression function

### LINKED ISSUE #10994: EIP-152
[EIP-152](https://github.com/ethereum/EIPs/pull/2129):  Add Blake2 compression function F precompile

seunlanlege:
WIP

will close #10994


ordian:
The is also https://github.com/oconnor663/blake2_simd, but none of them seem to expose the compress function. Would be nice to ask for making it pub/extract into a separate low-level crate.
And the impl for EIP-152 should probably be in a separate crate, not EIP-712.

util/EIP-712/src/blake2_f.rs
```diff
@@ -0,0 +1,223 @@
...
+	[9, 5, 2, 10, 0, 7, 4, 15, 14, 11, 6, 3, 1, 12, 8, 13],
+	[2, 6, 0, 8, 12, 10, 11, 3, 4, 7, 15, 1, 13, 5, 14, 9],
+	[12, 1, 14, 4, 5, 15, 13, 10, 0, 6, 9, 8, 7, 3, 2, 11],
+	[13, 7, 12, 3, 11, 14, 1, 9, 5, 15, 8, 2, 0, 4, 6, 10],
+	[6, 14, 11, 0, 15, 9, 3, 8, 12, 13, 1, 10, 2, 7, 4, 5],
+	[10, 8, 7, 1, 2, 4, 6, 5, 15, 9, 3, 13, 11, 14, 12, 0],
+];
+
+
+/// IV is an initialization vector for BLAKE2b
```
dvdplm:
```suggestion
/// IV is the initialization vector for BLAKE2b
```

I think this is the IV mandated by the spec yes?

util/EIP-712/src/blake2_f.rs
```diff
@@ -0,0 +1,223 @@
...
+const IV: [u64; 8] = [
+	0x6a09e667f3bcc908, 0xbb67ae8584caa73b, 0x3c6ef372fe94f82b, 0xa54ff53a5f1d36f1,
+	0x510e527fade682d1, 0x9b05688c2b3e6c1f, 0x1f83d9abfb41bd6b, 0x5be0cd19137e2179,
+];
+
+/// F is a compression function for BLAKE2b. It takes as an argument the state
```
dvdplm:
```suggestion
/// F is the compression function for BLAKE2b. It takes as an argument the state
```

util/EIP-712/src/blake2_f.rs
```diff
@@ -0,0 +1,223 @@
...
+	v13 ^= c1;
+
+	if f {
+		v14 ^= 0xffffffffffffffff;
+	}
+
+	for i in 0..rounds {
+		let s = &(PRECOMPUTED[i % 10]);
+
+		v0 = v0.overflowing_add(m[s[0]]).0;
```
dvdplm: Consider using the `unroll` crate rather than manually unroll the loop.

util/EIP-712/src/blake2_f.rs
```diff
@@ -0,0 +1,223 @@
+/// the precomputed values for BLAKE2b
+/// there are 10 16-byte arrays - one for each round
+/// the entries are calculated from the sigma constants.
+const PRECOMPUTED: [[usize; 16]; 10] = [
```
dvdplm: Why is this `usize` rather than `u64`?

util/EIP-712/src/blake2_f.rs
```diff
@@ -0,0 +1,223 @@
...
+const IV: [u64; 8] = [
+	0x6a09e667f3bcc908, 0xbb67ae8584caa73b, 0x3c6ef372fe94f82b, 0xa54ff53a5f1d36f1,
+	0x510e527fade682d1, 0x9b05688c2b3e6c1f, 0x1f83d9abfb41bd6b, 0x5be0cd19137e2179,
+];
+
+/// F is a compression function for BLAKE2b. It takes as an argument the state
+/// vector `h`, message block vector `m`, offset counter `t`, final
+/// block indicator flag `f`, and number of rounds `rounds`. The state vector
+/// provided as the first parameter is modified by the function.
+fn blake2_f(h: &mut [u64; 8], m: [u64; 16], c: [u64; 2], f: bool, rounds: usize) {
```
dvdplm: The doc-comment mentions the "offset counter `t`" but the actual args have no `t`. I think it corresponds to `c` but I don't quite understand why it's an array. The spec says "2w-bit offset counter "t", but yeah, not sure I understand that either. Can you help me understand?

dvdplm: `rounds` is `usize` here, but the EIP defines the rounds to be `u32` (be). Maybe we should just use a `u32` here?

util/EIP-712/src/blake2_f.rs
seunlanlege: slices can only be indexed by usize

util/EIP-712/src/blake2_f.rs
seunlanlege: > "2w-bit offset counter "t"
`w` means u64

so `t` is an array of 2 u64 bits: `[u64; 2]`

seunlanlege: > rounds is usize here, but the EIP defines the rounds to be u32 (be). Maybe we should just use a u32 here?
So the reason `rounds` is usize is because it'll be used to index the `SIGMA` slice and array indexing requires `usize`.

util/EIP-712/src/blake2_f.rs
dvdplm: ofc, ignore me.

seunlanlege:
closing in favour of #11017

PR #11010, opened 2019-08-29
Extract snapshot to own crate

dvdplm:
Extract `snapshot` to its own crate. Nothing particularly interesting to see here: shuffling files around and move types&traits to where they need to be. Adds a few more generics to the `snapshot` crate to disentangle it from `ethcore::Client`.
One needed refactoring is not present here: putting back `SnapshotClient` and `SnapshotWriter` in `snapshot`. This is taken care of in #11012 as it turned out to be more involved than I thought.

I have warp synced mainnet and I have observed snapshot chunks being created with this branch.

@niklasad1 Maybe this enables snapshots for the light client?


ethcore/client-traits/src/lib.rs
```diff
@@ -474,3 +475,38 @@ pub trait SnapshotWriter {
 	/// with the chunks written.
 	fn finish(self, manifest: common_types::snapshot::ManifestData) -> std::io::Result<()> where Self: Sized;
 }
+
+
+/// Represents what has to be handled by actor listening to chain events
+pub trait ChainNotify : Send + Sync {
```
niklasad1:
```suggestion
pub trait ChainNotify: Send + Sync {
```

ethcore/snapshot/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+description = "Snapshot TODO"
```
niklasad1: add a better description

dvdplm:
```suggestion
description = "Take and restore snapshots of the blockchain and read/write it in chunks from/to disk"
```

ethcore/snapshot/src/watcher.rs
```diff
@@ -86,17 +86,16 @@ impl Watcher {
 	/// Create a new `Watcher` which will trigger a snapshot event
 	/// once every `period` blocks, but only after that block is
 	/// `history` blocks old.
-	pub fn new<F>(client: Arc<Client>, sync_status: F, channel: IoChannel<ClientIoMessage<Client>>, period: u64, history: u64) -> Self
-		where F: 'static + Send + Sync + Fn() -> bool
+	pub fn new<F, C>(client: Arc<dyn BlockInfo>, sync_status: F, channel: IoChannel<ClientIoMessage<C>>, period: u64, history: u64) -> Self
```
ordian: Why doesn't it use the same `C` from `Mutex<IoChannel<ClientIoMessage<C>>>`, i.e. why introduce a separate generic type in the constructor?

dvdplm: I think they are the same? Maybe you misread the diff thinking `new()` was part of the impl of `Broadcast for Mutex<IoChannel<ClientIoMessage<C>>>`? Or I'm misunderstanding! `Watcher` is not parametrized.

dvdplm:
> Can you elaborate why the `Send + Senc` requirement wasn't needed before in several places and is now?

I wish I had something intelligent to say here, but tbh I followed the compilers' orders and she asked for `Send + Sync` (and `'static` in some places) and I obeyed. 
I explained it to myself by thinking that the concrete type `Client` was `Send + Sync` (à la "if a type is composed entirely of Send or Sync types, then it is Send or Sync") and so when replacing it with a generic, it needs the extra restrictions whenever the compiler sees that it is shared between threads.

ethcore/snapshot/src/watcher.rs
ordian: Ah, right, I misread the diff, sorry :/

PR #11012, opened 2019-08-30
Move snapshot related traits to their proper place

dvdplm:
This PR is a continuation of https://github.com/paritytech/parity-ethereum/pull/11010 that puts `SnapshotClient` and `SnapshotWriter` back where they belong in the `snapshot` crate.
This move creates a tricky situation for tests, where the direct dependency of the tests on `ethcore::Client` (and other stuff from `ethcore`) cause the trait impls to be "lost". 
As a work-around this PR moves all snapshot tests to their own crate, `snapshot-tests`.

Salient points:
- moves the `SnapshotClient` and `SnapshotWriter` traits from `client-traits` to the `snapshot` crate where they belong
- moves all tests from `snapshot` to `snapshot-tests`
- there are a few places in the `snapshot` where types or methods have been made `pub` only to accommodate tests (e.g. [here](https://github.com/paritytech/parity-ethereum/pull/11012/files#diff-262e9bb6549c6f70b158a8c213623ed8R38)); in places where it makes sense to do so, I have used `#[cfg(any(test, feature = "test-helpers"))]` (e.g. [here](https://github.com/paritytech/parity-ethereum/pull/11012/files#diff-f0ca3d8125e35de42b032b8fd1527270R67)). This is unfortunate but I could not think of a better way.
- to make `cargo test --all` actually work a crate has to be a workspace member **or** part of the dependency tree; for `snapshot-tests` this means it has to be listed in the `[dev-dependencies]` for `snapshot`.

Continuation of #11010


ethcore/snapshot/snapshot-tests/src/helpers.rs
```diff
@@ -22,6 +22,18 @@ use std::sync::atomic::AtomicBool;
...
+use blockchain::{BlockChain, BlockChainDB};
+use client_traits::ChainInfo;
+use common_types::{
+	ids::BlockId,
+	basic_account::BasicAccount,
+	errors::EthcoreError
+};
+use engine::Engine;
+use ethcore::client::Client;
+=======
```
ordian: there are some merge conflicts here and the tests are passing, which is alarming

ethcore/snapshot/snapshot-tests/src/lib.rs
```diff
@@ -16,25 +16,43 @@
...
 mod state;
+#[cfg(test)]
 mod service;
+#[cfg(test)]
+mod watcher;
 
+<<<<<<< HEAD:ethcore/snapshot/snapshot-tests/src/lib.rs
+#[cfg(test)]
+mod helpers;
+=======
```
ordian: merge conflicts

ethcore/snapshot/snapshot-tests/src/proof_of_authority.rs
```diff
@@ -40,6 +62,7 @@ use ethabi_contract::use_contract;
 use lazy_static::lazy_static;
 
 use_contract!(test_validator_set, "../res/contracts/test_validator_set.json");
+>>>>>>> master:ethcore/snapshot/src/tests/proof_of_authority.rs
```
ordian: ditto

ethcore/snapshot/src/service.rs
```diff
@@ -200,7 +216,7 @@ impl Restoration {
 	}
 
 	// is everything done?
-	fn is_done(&self) -> bool {
+	pub fn is_done(&self) -> bool {
```
ordian: let's make the comments rustdocs (`///`) on pub functions

ethcore/snapshot/snapshot-tests/src/state.rs
```diff
@@ -96,7 +113,11 @@ fn snap_and_restore() {
 		new_db
 	};
 
+<<<<<<< HEAD:ethcore/snapshot/snapshot-tests/src/state.rs
+	let new_db = journaldb::new(db, Algorithm::OverlayRecent,  ethcore_db::COL_STATE);
+=======
 	let new_db = journaldb::new(db, Algorithm::OverlayRecent, ethcore_db::COL_STATE);
+>>>>>>> master:ethcore/snapshot/src/tests/state.rs
```
ordian: here too

ethcore/snapshot/snapshot-tests/src/proof_of_work.rs
```diff
@@ -19,18 +19,28 @@
 use std::sync::atomic::AtomicBool;
 use tempdir::TempDir;
 use common_types::{
+<<<<<<< HEAD:ethcore/snapshot/snapshot-tests/src/proof_of_work.rs
+	errors::{EthcoreError as Error, SnapshotError},
+=======
 	errors::EthcoreError as Error,
+>>>>>>> master:ethcore/snapshot/src/tests/proof_of_work.rs
```
ordian: same

ethcore/snapshot/snapshot-tests/src/service.rs
```diff
@@ -22,17 +22,30 @@ use std::sync::Arc;
...
+use snapshot::{
+	chunk_state, chunk_secondary, SnapshotService,
+	io::{PackedReader, PackedWriter, SnapshotReader, SnapshotWriter},
+	service::{Service, ServiceParams, Guard, Restoration, RestorationParams},
+=======
 use crate::{
 	chunk_state, chunk_secondary, SnapshotService,
 	io::{PackedReader, PackedWriter, SnapshotReader},
 	service::{Service, ServiceParams},
+>>>>>>> master:ethcore/snapshot/src/tests/service.rs
```
ordian: and there

ethcore/snapshot/Cargo.toml
```diff
@@ -52,3 +52,9 @@ lazy_static = { version = "1.3" }
 spec = { path = "../spec" }
 tempdir = "0.3"
 trie-standardmap = "0.15.0"
+# Note[dvdplm]: Ensure the snapshot tests are included in the dependency tree, which in turn means that
+# `cargo test --all` runs the tests.
```
ordian: this is unfortunate, but I didn't come up with something better

ethcore/snapshot/snapshot-tests/src/lib.rs
```diff
@@ -16,25 +16,37 @@
 
 //! Snapshot tests.
 
+#[cfg(test)]
```
niklasad1: why not just `#![cfg(test)]` since it is just used for tests?

ethcore/snapshot/src/watcher.rs
```diff
@@ -53,8 +55,8 @@ impl<F> Oracle for StandardOracle<F>
 	}
 }
 
-// helper trait for broadcasting a block to take a snapshot at.
-trait Broadcast: Send + Sync {
+/// Helper trait for broadcasting a block to take a snapshot at.
```
niklasad1: :+1:

ethcore/snapshot/src/watcher.rs
```diff
@@ -53,8 +55,8 @@ impl<F> Oracle for StandardOracle<F>
...
+pub trait Broadcast: Send + Sync {
 	fn take_at(&self, num: Option<u64>);
```
niklasad1: Can you add docs for this too?

ethcore/snapshot/src/watcher.rs
```diff
@@ -86,7 +88,14 @@ impl Watcher {
 	/// Create a new `Watcher` which will trigger a snapshot event
 	/// once every `period` blocks, but only after that block is
 	/// `history` blocks old.
-	pub fn new<F, C>(client: Arc<dyn BlockInfo>, sync_status: F, channel: IoChannel<ClientIoMessage<C>>, period: u64, history: u64) -> Self
+	pub fn new<F, C>(
+		client: Arc<dyn BlockInfo>,
+		sync_status: F,
+		channel:
```
niklasad1: format nit, should probably be `channel: IoChannel<ClientIoMessage<C>>,`

ethcore/snapshot/src/block.rs
```diff
@@ -30,7 +30,7 @@ use triehash::ordered_trie_root;
 const HEADER_FIELDS: usize = 8;
 const BLOCK_FIELDS: usize = 2;
 
-pub(crate) struct AbridgedBlock {
+pub struct AbridgedBlock {
```
niklasad1: nitpick: docs

ethcore/snapshot/src/lib.rs
```diff
@@ -57,22 +55,29 @@ use state_db::StateDB;
...
+pub use self::traits::{SnapshotService, SnapshotClient, SnapshotComponents, Rebuilder};
+pub use self::io::SnapshotWriter;
 pub use self::watcher::Watcher;
-pub use common_types::basic_account::BasicAccount;
+use common_types::basic_account::BasicAccount;
 
 pub mod io;
 pub mod service;
 
+#[cfg(any(test, feature = "test-helpers"))]
```
niklasad1: well, I think `feature = "test-helpers"` should be enough because `snapshot` doesn't contain any tests after this PR but I don't mind if we keep it to be on safe-side.

ethcore/snapshot/src/watcher.rs
```diff
@@ -29,10 +29,12 @@ use ethcore_io::IoChannel;
 use log::{trace, warn};
 use parking_lot::Mutex;
 
-// helper trait for transforming hashes to numbers and checking if syncing.
-trait Oracle: Send + Sync {
+/// Helper trait for transforming hashes to block numbers and checking if syncing.
+pub trait Oracle: Send + Sync {
```
niklasad1: maybe move the `new pub traits` to the `traits.rs`?

ethcore/snapshot/src/lib.rs
niklasad1: Also, I think we could have something like:

```rustc
#[cfg(not(any(test, feature = "test-helpers")))]
pub mod test_helpers {
   pub use watcher::Watcher;
    .... export everything we need
}
```
However, I don't know how painful it would be to change the import paths for the tests but it would reduce the `cfg stff` a bit

dvdplm: It makes the tests a wee bit more opaque, but overall I think your suggestion is better than what I had. 👍

PR #11017, opened 2019-09-02
Add blake2_f precompile

### LINKED ISSUE #10994: EIP-152

dvdplm:
As per EIP 152, adds a new precompile at `0x09` to expose the Blake2 compression function, aka the "F function".

Spec: https://tools.ietf.org/html/rfc7693#section-3.2
EIP-152: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md

Closes #10994

*Note*: the F compression function in this PR is not optimised (or rather, it's optimised for readability ;). This will be addressed in a separate PR as the intention here is to be 1. spec compliant and 2. get the precompile plumbing in place.

*Note-2*: The chainspecs are not yet updated a part from [`foundation.json`](https://github.com/paritytech/parity-ethereum/pull/11017/files#diff-09e1141f3610b5496b23dd6fa862d325R3961-R3969). Would appreciate a review of the format used there and I'll replicate it elsewhere.


json/src/spec/builtin.rs
```diff
@@ -18,6 +18,9 @@
 
 use uint::Uint;
 
+/// Fixed pricing: each call cost the same
```
ordian: The cost is based on rounds from input, this should be reflected in the name and the docs somehow.

ethcore/builtin/src/lib.rs
```diff
@@ -44,6 +45,19 @@ trait Pricer: Send + Sync {
...
 }
 
+/// A linear pricing model. This computes a price using a base cost and a cost per-word.
+pub type Fixed = u64;
+
+impl Pricer for Fixed {
+	fn cost(&self, input: &[u8], _at: u64) -> U256 {
+		use std::convert::TryInto;
+		let (rounds_bytes, _) = input.split_at(std::mem::size_of::<u32>());
+		// todo[dvdplm] what's the error handling strategy here? panic? return U256::MAX? Is the input len checked before this call?
```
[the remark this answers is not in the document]
dvdplm: `ModexpPricer` seem to return zero or `U256::max_value()`. Returning zero seems odd though: they get to run the precompile for free?
/cc @sorpaas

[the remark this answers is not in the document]
ordian: @dvdplm if we return `U256::max_value()`, then the error on empty input would be out of gas, but the spec requires otherwise
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-152.md#test-vector-0
`execute` will do the length check and return the appropriate error.

ethcore/builtin/src/lib.rs
```diff
@@ -44,6 +46,22 @@ trait Pricer: Send + Sync {
...
+pub type Blake2FPricer = u64;
+
+impl Pricer for Blake2FPricer {
+	fn cost(&self, input: &[u8], _at: u64) -> U256 {
+		use std::convert::TryInto;
+		let (rounds_bytes, _) = input.split_at(std::mem::size_of::<u32>());
+		// Returning zero if the conversion fails is fine because `execute()` will check the length
+		// and bail with the appropriate error.
+		let rounds = u32::from_be_bytes(rounds_bytes.try_into().unwrap_or_else(|_| [0u8; 4] ));
+		U256::from(*self * rounds as u64)
```
ordian: can it overflow? maybe let's cast it to `u128`?

ethcore/builtin/src/lib.rs
```diff
@@ -337,6 +366,54 @@ impl Implementation for Sha256 {
...
 }
 
+impl Implementation for Blake2F {
+	/// Format of `input`: `[4 bytes for rounds][64 bytes for h][128 bytes for m][8 bytes for t_0][8 bytes for t_1][1 byte for f]`
+	fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), &'static str> {
+		const BLAKE2_F_ARG_LEN: usize = 213;
+		const PROOF: &str = "Checked the length of the input above; qed";
+
+		if input.len() != BLAKE2_F_ARG_LEN {
+			// todo[dvdplm] what's the right 'target' here?
```
ordian: either `evm` or a new target `builtin` is fine

ethcore/builtin/Cargo.toml
```diff
@@ -14,7 +14,10 @@ keccak-hash = "0.2.0"
 log = "0.4"
 num = { version = "0.1", default-features = false, features = ["bigint"] }
 parity-bytes = "0.1"
-parity-crypto = "0.4.0"
+eip-152 = { path = "../../util/EIP-152" }
+parity-crypto = { version = "0.4.0", git = "https://github.com/paritytech/parity-common/", branch = "dp/feature/add-blake2_f-toparity-crypto"}
```
ordian: Do we want to use an `EIP-152` crate or move it to parity-crypto? If it's not used anywhere else, having a crate in `util` is IHMO fine.

ethcore/builtin/src/lib.rs
```diff
@@ -337,6 +366,54 @@ impl Implementation for Sha256 {
...
+				Some(0) => false,
+				_ => {
+					trace!(target: "evm", "incorrect final block indicator flag, was: {:?}", input.last());
+					return Err("incorrect final block indicator flag")
+				}
+			};
+
+		compress(&mut h, m, [t0, t1], f, rounds as usize);
+
+		output.write(0, unsafe { std::slice::from_raw_parts(h.as_ptr() as *const u8, h.len() * std::mem::size_of::<u64>())});
```
ordian: Please add a comment about safety.

ethcore/builtin/src/lib.rs
seunlanlege: There's no need to cast, a `u32` is 4 `u8`s. the call to `mem::size_of::<u32>` would return `4`
```suggestion
		U256::from(*self as u128 * rounds as u128)
```

seunlanlege:
```suggestion
		U256::from(*self as u128 * rounds as u128)
```

ethcore/builtin/Cargo.toml
dvdplm: Yes, we'll have `util/eip-152` (see https://github.com/paritytech/parity-ethereum/pull/11017/commits/263e3593e3ee2bf976b836fc2824a1d6a15728e3).

ethcore/builtin/src/lib.rs
```diff
@@ -526,7 +611,7 @@ impl Implementation for Bn128Pairing {
 		}
 
 		if let Err(err) = self.execute_with_error(input, output) {
-			trace!("Pairining error: {:?}", err);
+			trace!(target: "builtin", "Pairining error: {:?}", err);
```
niklasad1:
```suggestion
			trace!(target: "builtin", "Pairing error: {:?}", err);
```

ethcore/builtin/src/lib.rs
niklasad1: I think `builtin` is fine, can you remove the `todo`?

ethcore/builtin/src/lib.rs
```diff
@@ -600,8 +685,117 @@ mod tests {
 	use num::{BigUint, Zero, One};
 	use parity_bytes::BytesRef;
 	use rustc_hex::FromHex;
+	use hex_literal::hex; // todo[dvdplm] switch tests to hex_literal
```
niklasad1: well, `hex_literal` is used in the tests in this PR.

I guess it refers to the other tests right? Maybe clarify or create an issue for it

dvdplm: I'll port over all the other tests to use `hex!`. I am so fed up with `rustc-hex`. :)

ethcore/builtin/Cargo.toml
ordian: Should this line be removed then?
```suggestion
```

dvdplm:
I'd like a review from @sorpaas as well.

dvdplm:
> please open an issue to optimize the compress function.

https://github.com/paritytech/parity-ethereum/issues/11020 – grab it while it's hot! :)

util/EIP-152/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+	v[12] ^= t[0];
+	v[13] ^= t[1];
+
+	if f {
+		v[14] = !v[14]
+	}
+
+	for i in 0..rounds {
+		let s = &SIGMA[i % 10];
+		v[0] = v[0].wrapping_add(m[s[0]]);
```
ordian: we could perhaps update the code from https://github.com/paritytech/parity-ethereum/pull/11009/files#diff-ef6e79ce14a73aeb8448a91308cc6391R23-R63 for better readability

dvdplm: /cc @seunlanlege

ordian: https://github.com/paritytech/parity-ethereum/compare/dp/feature/eip-152-add-blake2-precompile...ao-more-readable-blake

util/EIP-152/src/lib.rs
```diff
@@ -218,28 +131,32 @@ mod tests {
...
+				"b63a380cb2897d521994a85234ee2c181b5f844d2c624c002677e9703449d2fba551b3a8333bcdf5f2f7e08993d53923de3d64fcc68c034e717b9293fed7a421",
 			),
-			(
-				"0000000048c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000001",
-				"08c9bcf367e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d282e6ad7f520e511f6c3e2b8c68059b9442be0454267ce079217e1319cde05b"
-			)
+			// Test vector 8 – u32::MAX rounds – too slow to run
+//			(
```
niklasad1: Is it annoying to refactor this to be a separate test for each test vector? 

Would be better to use:

```rustc
// ignore test because it is slow to run 
// run with `cargo test -- --ignored` if you want to test this 
#[ignore]
#[test]
```

Then we don't have to manually comment/uncomment

seunlanlege: Actually, I'd elect for removing it completely. Primarily because there's no possibility of an overflow

ethcore/builtin/src/lib.rs
```diff
@@ -337,6 +368,59 @@ impl Implementation for Sha256 {
...
+			trace!(target: "builtin", "input length for Blake2 F precompile should be exactly 213 bytes, was {}", input.len());
+			return Err("input length for Blake2 F precompile should be exactly 213 bytes")
+		}
+
+		let mut cursor = Cursor::new(input);
+		let rounds = cursor.read_u32::<BigEndian>().expect(PROOF);
+
+		// state vector, h
+		let mut h = [0u64; 8];
+		for i in 0..8 {
```
niklasad1: prefer iterators, over ranges

```rustc
    for x in h.iter_mut() {
        *x = cursor.read_u64::<LittleEndian>().expect(PROOF);
    }
```

ethcore/builtin/src/lib.rs
```diff
@@ -44,6 +48,22 @@ trait Pricer: Send + Sync {
...
+/// byte slice.
+pub type Blake2FPricer = u64;
+
+impl Pricer for Blake2FPricer {
+	fn cost(&self, input: &[u8], _at: u64) -> U256 {
+		use std::convert::TryInto;
+		let (rounds_bytes, _) = input.split_at(std::mem::size_of::<u32>());
+		// Returning zero if the conversion fails is fine because `execute()` will check the length
+		// and bail with the appropriate error.
+		let rounds = u32::from_be_bytes(rounds_bytes.try_into().unwrap_or_else(|_| [0u8; 4] ));
```
niklasad1: nitpick: needless closure, prefer `unwrap_or` or `unwrap_or_default` because `[u8; 4]` is just value

ethcore/builtin/src/lib.rs
```diff
@@ -337,6 +368,59 @@ impl Implementation for Sha256 {
...
+
+		// state vector, h
+		let mut h = [0u64; 8];
+		for i in 0..8 {
+			h[i] = cursor.read_u64::<LittleEndian>().expect(PROOF);
+		}
+
+		// message block vector, m
+		let mut m = [0u64; 16];
+		for i in 0..16 {
```
niklasad1: same here; prefer iterators, over ranges

json/src/spec/builtin.rs
```diff
@@ -65,6 +68,8 @@ pub struct AltBn128Pairing {
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "snake_case")]
 pub enum Pricing {
+	/// Fixed pricing.
```
ordian:
```suggestion
	/// Pricing for Blake2 compression function: each call costs the same amount per round.
```

json/src/spec/builtin.rs
```diff
@@ -18,6 +18,9 @@
 
 use uint::Uint;
 
+/// Pricing for Blake2 compression function: each call costs the same amount per round.
```
ordian:
```suggestion
/// Price per round of Blake2 compression.
```

util/EIP-152/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+pub fn compress(h: &mut [u64; 8], m: [u64; 16], t: [u64; 2], f: bool, rounds: usize) {
+	let mut v = [0u64; 16];
+	v[..h.len()].copy_from_slice(h);    // First half from state.
+	v[h.len()..].copy_from_slice(&IV);  // Second half from IV.
+
+	v[12] ^= t[0];
+	v[13] ^= t[1];
+
+	if f {
+		v[14] = !v[14] // Invert all bits is the last-block-flag is set.
```
niklasad1:
```suggestion
		v[14] = !v[14] // Invert all bits if the last-block-flag is set.
```

util/EIP-152/src/lib.rs
dvdplm: There's a chance the encoding format will change (see https://github.com/ethereum/EIPs/issues/152#issuecomment-526645573) so this one might become an error candidate. Otherwise I agree we should remove it.

niklasad1:
@dvdplm do we got any benches for this?

Would be nice but maybe it is intended to be part of #11020?

dvdplm:
Let's not merge this PR before I say so. There are problems with the EIP being discussed in the [gitter](https://gitter.im/ethereum/AllCoreDevs) and until we know what is going to happen we're better off not merging this.

ethcore/res/ethereum/foundation.json
```diff
@@ -3958,6 +3958,15 @@
...
 		},
+        "0x0000000000000000000000000000000000000009": {
+          "builtin": {
+            "name": "blake2_f",
+            "activate_at": "0xffffff",
+            "pricing": {
+              "blake2_f": 1
+            }
+          }
+        },
```
ordian: s/spaces/tabs

holiman:
> Let's not merge this PR before I say so. There are problems with the EIP being discussed in the gitter and until we know what is going to happen we're better off not merging this.

It was decided on ACD on friday that nothing changes on the implementation: https://github.com/ethereum/EIPs/issues/152#issuecomment-528919091 . 

> *    There seems to be some work going on in the background to make use of a BLAKE2b-variant with non-standard rounds. Unfortunately @mhluongo could not divulge further details.
> *    It is unknown what would be a reasonable upper bound for rounds for such a non-standard BLAKE2b. As a result the rounds parameter seemed to be left at 32-bit.
> *    It was agreed that the EIP should be updated to reflect it describes a very specific superset of BLAKE2b, but still a subset of BLAKE2. The EIP also should include the actual compression function and parameters (SIGMA, IV) – and not leave it as a reference to the RFC.
 > *   There was no agreement about introducing the m_len field, which is a devex improvement and gas saving. It would make sense for client implementers to judge the complexity and decide upon that.

Please get this merged so testing can continue. Btw: did you guys implement `Istanbul` in state tests yet?

ethcore/builtin/src/lib.rs
```diff
@@ -337,6 +368,62 @@ impl Implementation for Sha256 {
...
+			};
+
+		compress(&mut h, m, t, f, rounds as usize);
+
+		output.write(0, unsafe {
+			// `from_raw_parts` is safe under the assumption that `h` is
+			// 1. immutable,
+			// 2. comprised of `u64`
+			// …and because we're using the (known) length of `h`
+			slice::from_raw_parts(h.as_ptr() as *const u8, h.len() * size_of::<u64>())
```
sorpaas: It might be better to remove this unsafe. What I worry is someone compiling parity-ethereum to a big endian platform. There may not be many, but once people do that it breaks consensus. We may want to just specify the exact encoding how conversion from u64 array to bytearray happens.

ordian: There could be other problems with big endian, e.g. with ethash https://github.com/paritytech/parity-ethereum/blob/5baa7e8fb50512b0de11b0994fa1b136de289831/ethash/src/compute.rs#L273-L276

but in this case it should work with big endian if I'm not mistaken

PR #11019, opened 2019-09-03
[trace] introduce trace failed to Ext

ordian:
When a subcall fails, we need to clean up the tracing data (`last_mem_written`, `last_store_written`) in the `ExecutiveVMTracer`. Also to properly track subcalls, we maintain a stack of tracing data.

One approach of doing that is to [pass `mem_written`](https://github.com/paritytech/parity-ethereum/compare/master...ordian:ao-move-memdiff-to-trace-executed) directly to `trace_executed`. It [does](https://gist.github.com/ordian/a6b398650ae30df0dc7ebf51207f988d) introduce a perf regression.

This PR adds a method `trace_failed` to `Ext`, which is called when the execution of an instruction has failed. An alternative suggested by @tomusdrw would be to merge `trace_failed` into `trace_executed` by passing a `Result<(stack_push: &[U256], mem: &[u8]), ()>`, but I'm not sure if is any better.


ethcore/vm/src/ext.rs
```diff
@@ -166,6 +166,9 @@ pub trait Ext {
 	/// Prepare to trace an operation. Passthrough for the VM trace.
 	fn trace_prepare_execute(&mut self, _pc: usize, _instruction: u8, _gas_cost: U256, _mem_written: Option<(usize, usize)>, _store_written: Option<(U256, U256)>) {}
```
tomusdrw: Can we improve the docs here and underline the fact that this call is ALWAYS followed by either `trace_executed` or `trace_failed`. and that the implementation is responsible to clean up in either/both of those methods.

ordian:
> Why did you opt to remove the bounds checks here? For performance reasons? Unless it's a severe regression I think it'd be good to have.

If the bounds could still be incorrect, it would indicate another bug and a warning could easily be missed/ignored. But if we reach a consensus on that the checks should be added back, I don't mind.

dvdplm:
> But if we reach a consensus on that the checks should be added back, I don't mind.

@tomusdrw thought? Better to crash?

tomusdrw:
@dvdplm I'd rather bound the `o` and `o+s` with  `mem.len()` to be safe. The bound checking and a warning was introducing a lot of noise.
To be fair I'd rather not crash and have potentially incorrect data in tracing.

dvdplm:
https://github.com/paritytech/parity-ethereum/pull/11019/commits/f818b23a3e91546994b5985a10bfd080f15a0058 lgtm

PR #11022, opened 2019-09-04
[dependencies] bump rand 0.7

niklasad1:
After upgrading `ethereum-types` to `0.8` which internally depends on `rand 0.7` this PR tries the align we use `rand 0.7` in all our crates in parity-ethereum

It is still some other crates that depend on earlier versions of `rand` which this doesn't fix.


niklasad1:
I misread the vulnerability, should be backported to 0.6 so no rush I guess. Closing for now

accounts/ethkey/src/random.rs
```diff
@@ -21,14 +21,13 @@ use super::{Generator, KeyPair, SECP256K1};
...
-	type Error = ::std::io::Error;
+	type Error = std::io::Error;
 
 	fn generate(&mut self) -> Result<KeyPair, Self::Error> {
-		let mut rng = OsRng::new()?;
-		match rng.generate() {
-			Ok(pair) => Ok(pair),
-			Err(void) => match void {}, // LLVM unreachable
-		}
+		// TODO(niklasad1): Generator::generate always returns `Ok`, then use expect instead?!
```
ordian: I think the match with void type was done for optimization purposes.

PR #11025, opened 2019-09-06
Fix ethash benchmark imports.

afck:
This fixes `cargo check --all --all-targets`.

Otherwise the `ethash` benches wouldn't compile:
```
error[E0603]: module `progpow` is private
 --> ethash/benches/progpow.rs:9:13
  |
9 | use ethash::progpow;
  |             ^^^^^^^

error[E0603]: module `compute` is private
  --> ethash/benches/progpow.rs:14:13
   |
14 | use ethash::compute::light_compute;
```
(When building benches, the crate isn't compiled with the `bench` feature.)

### DIFF
--- a/ethash/src/lib.rs
+++ b/ethash/src/lib.rs
@@ -37,20 +37,16 @@ extern crate serde_json;
 #[cfg(test)]
 extern crate tempdir;
 
-#[cfg(feature = "bench")]
+// TODO: This should only be public for benchmarks.
 pub mod compute;
-#[cfg(not(feature = "bench"))]
-mod compute;
 
 mod seed_compute;
 mod cache;
 mod keccak;
 mod shared;
 
-#[cfg(feature = "bench")]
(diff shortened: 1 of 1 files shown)


ordian:
You can run ethash benches by passing `--features bench`, see https://github.com/paritytech/parity-ethereum/blob/680807e601be936e21086573e28189200f1c7167/.gitlab-ci.yml#L105

afck:
Right, so
```
cargo check --all --all-targets
```
isn't supposed to pass at all?
I guess there's no way to exclude the benches from `--all-targets`… :confused: 
Anyway, sorry for the confusion! Feel free to close this PR.

ordian:
> isn't supposed to pass at all?
I guess there's no way to exclude the benches from --all-targets… confused

Right, this is annoying, but we have multi-stage CI pipeline and there are other crates that have features, that are not exposed in the root `Cargo.toml`. So there is no one command that can check every target with every feature combination.

PR #11030, opened 2019-09-08
Resolve "Reverted report validator transactions get added in block #10580"

### LINKED ISSUE #10580: Reverted report validator transactions get added in block

VladLupashevskyi:
Continuation of #10602

I opened the new MR since I couldnt reopen the previous MR (probably because I force pushed).

Basically here I implemented CLI flags for selecting a preferred strategy for reports as mentioned [here](https://github.com/paritytech/parity-ethereum/pull/10602#issuecomment-502715665) with caching.

It's possible to use the following CLI flags:
- benign_reporting=[force|call|disable]
- malicious_reporting=[force|call|disable]
- report_cache_call_every=[blk_num]

The frequency in blocks for caching is set for both reporting types. I wasn't sure whether it makes sense to make it also separated, but it will be easy to add if needed.

Please take a look and let me know if I need to improve or update something :)

Closes #10580

### DIFF
--- a/ethcore/call-contract/src/call_contract.rs
+++ b/ethcore/call-contract/src/call_contract.rs
@@ -16,14 +16,63 @@
 
 //! Provides CallContract and RegistryInfo traits
 
+use std::fmt;
 use bytes::Bytes;
-use ethereum_types::Address;
+use derive_builder::Builder;
+use ethereum_types::{Address, U256};
 use types::ids::BlockId;
+
+#[derive(Debug)]
+/// Transaction call error
+pub enum CallError {
+	/// Call reverted
+	Reverted(String),
+	/// Other error
+	Other(String),
+}
+
+impl fmt::Display for CallError {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		fmt::Debug::fmt(&self, f)
+	}
+}
+
+#[derive(Builder, Default, Clone)]
+#[builder(setter(into))]
+/// Options to make a call to contract
+pub struct CallOptions {
+	/// Contract address
+	pub contract_address: Address,
+	/// Transaction sender
+	pub sender: Address,
+	/// Transaction data
+	pub data: Bytes,
+	/// Value in wei
+	pub value: U256,
+	/// Provided gas
+	pub gas: U256,
+	/// Gas price
+	pub gas_price: U256,
+}
+
+impl CallOptions {
+	/// Convenience method for creating the most common use case.
+	pub fn new(contract: Address, data: Bytes) -> Self {
+		CallOptionsBuilder::default().contract_address(contract).data(data).gas_price(U256::from(50_000_000)).build().unwrap()
+	}
+}
 
 /// Provides `call_contract` method
 pub trait CallContract {
-	/// Like `call`, but with various defaults. Designed to be used for calling contracts.
-	fn call_contract(&self, id: BlockId, address: Address, data: Bytes) -> Result<Bytes, String>;
+	/// Executes a transient call to a contract at given `BlockId`.
+	///
+	/// The constructed transaction must be executed on top of state at block `id`,
+	/// any changes introduce by the call must be discarded.
+	/// Returns:
+	/// - A return data from the contract if the call was successful,
+	/// - A `CallError::Reverted(msg)` error in case the call was reverted with an exception and message.
+	/// - A `CallError::Other(msg)` in case the call did not succeed for other reasons.
+	fn call_contract(&self, id: BlockId, call_options: CallOptions) -> Result<Bytes, CallError>;
 }
 
 /// Provides information on a blockchain service and it's registry
--- a/ethcore/call-contract/src/call_contract.rs
+++ b/ethcore/call-contract/src/call_contract.rs
@@ -50,6 +50,7 @@ pub struct CallOptions {
 	/// Value in wei
 	pub value: U256,
 	/// Provided gas
+	#[builder(default = "self.default_gas()")]
 	pub gas: U256,
 	/// Gas price
 	pub gas_price: U256,
@@ -58,8 +59,14 @@ pub struct CallOptions {
 impl CallOptions {
 	/// Convenience method for creating the most common use case.
 	pub fn new(contract: Address, data: Bytes) -> Self {
-		CallOptionsBuilder::default().contract_address(contract).data(data).gas(U256::from(50_000_000)).build().unwrap()
+		CallOptionsBuilder::default().contract_address(contract).data(data).build().unwrap()
 	}
+}
+
+impl CallOptionsBuilder {
+    fn default_gas(&self) -> U256 {
+        U256::from(50_000_000)
+    }
 }
 
 /// Provides `call_contract` method
--- a/ethcore/call-contract/src/call_contract.rs
+++ b/ethcore/call-contract/src/call_contract.rs
@@ -58,7 +58,7 @@ pub struct CallOptions {
 impl CallOptions {
 	/// Convenience method for creating the most common use case.
 	pub fn new(contract: Address, data: Bytes) -> Self {
-		CallOptionsBuilder::default().contract_address(contract).data(data).gas_price(U256::from(50_000_000)).build().unwrap()
+		CallOptionsBuilder::default().contract_address(contract).data(data).gas(U256::from(50_000_000)).build().unwrap()
 	}
 }
 
--- a/ethcore/call-contract/src/call_contract.rs
+++ b/ethcore/call-contract/src/call_contract.rs
@@ -38,7 +38,7 @@ impl fmt::Display for CallError {
 }
 
 #[derive(Builder, Default, Clone)]
-#[builder(setter(into))]
+#[builder(default)]
 /// Options to make a call to contract
 pub struct CallOptions {
 	/// Contract address
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -384,7 +384,7 @@ pub trait Engine: Sync + Send {
 	}
 
 	/// Set reporting
-	fn set_reporting(&self, benign_reporting: ReportingConfig, malicious_reporting: ReportingConfig) {}
+	fn set_reporting(&self, _benign_reporting: ReportingConfig, _malicious_reporting: ReportingConfig) {}
 }
 
 /// Verifier for all blocks within an epoch with self-contained state.
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -45,6 +45,7 @@ use machine::{
 use vm::{EnvInfo, Schedule, CallType, ActionValue};
 
 use crate::signer::EngineSigner;
+use validator_reporting_config::ReportingConfig;
 
 /// A system-calling closure. Enacts calls on a block's state from the system address.
 pub type SystemCall<'a> = dyn FnMut(Address, Vec<u8>) -> Result<Vec<u8>, String> + 'a;
@@ -381,6 +382,9 @@ pub trait Engine: Sync + Send {
 	fn decode_transaction(&self, transaction: &[u8]) -> Result<UnverifiedTransaction, transaction::Error> {
 		self.machine().decode_transaction(transaction)
 	}
+
+	/// Set reporting
+	fn set_reporting(&self, benign_reporting: ReportingConfig, malicious_reporting: ReportingConfig) {}
 }
 
 /// Verifier for all blocks within an epoch with self-contained state.
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -876,7 +876,13 @@ impl AuthorityRound {
 					if !reported.insert(skipped_primary) { break; }
 					trace!(target: "engine", "Reporting benign misbehaviour (cause: skipped step) at block #{}, epoch set number {}, step proposer={:#x}. Own address: {}",
 						header.number(), set_number, skipped_primary, me);
-					self.validators.report_benign(&skipped_primary, set_number, header.number(), &mut *self.benign_reporting.lock());
+					self.validators.report_benign(
+						&skipped_primary,
+						set_number,
+						header.number(),
+						&mut *self.benign_reporting.lock(),
+						me
+					);
 				} else {
 					trace!(target: "engine", "Primary that skipped is self, not self-reporting. Own address: {}", me);
 				}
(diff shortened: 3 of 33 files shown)


VladLupashevskyi:
Hmm. I cant figure out why tests are failing. It seems that error is `Bad proof` which is caused because of `ExecutionError::Internal` and this one is caused because of `IncompleteDatabase`. This is so strange since I haven't touched that pieces of the code.

Could somebody please throw a quick look at it, maybe there is something really simple that I've missed?

dvdplm:
> Could somebody please throw a quick look at it, maybe there is something really simple that I've missed?

Nothing obvious stands out to me. I can repeat the failures locally and see `Insufficient validation proof: Bad proof` from `zoom_to_after()`.

VladLupashevskyi:
@dvdplm Thank you for checking this out.

I've tried running test on different commits and it starts to fail since this one a9079db48cdc5bc320acb422536078f63e4f63f8. Will dig into finding what's causing the issue.

VladLupashevskyi:
I think it's ready for review :)

ethcore/call-contract/src/call_contract.rs
```diff
@@ -16,14 +16,70 @@
...
+}
+
+impl fmt::Display for CallError {
+	fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+		fmt::Debug::fmt(&self, f)
+	}
+}
+
+#[derive(Builder, Default, Clone)]
+#[builder(default)]
```
debris: I believe this using this macro once is not worth bringing 5 new external dependencies to the project

ethcore/call-contract/src/call_contract.rs
```diff
@@ -16,14 +16,70 @@
...
 use types::ids::BlockId;
 
+#[derive(Debug)]
+/// Transaction call error
+pub enum CallError {
+	/// Call reverted
+	Reverted(String),
+	/// Other error
+	Other(String),
+}
```
debris: api consumer doesn't doesn't care whether the call is reverted has been reverted or not, which, imo makes this new data structure redundant

ethcore/engine/src/engine.rs
```diff
@@ -381,6 +382,9 @@ pub trait Engine: Sync + Send {
 	fn decode_transaction(&self, transaction: &[u8]) -> Result<UnverifiedTransaction, transaction::Error> {
 		self.machine().decode_transaction(transaction)
 	}
+
+	/// Set reporting
+	fn set_reporting(&self, _benign_reporting: ReportingConfig, _malicious_reporting: ReportingConfig) {}
```
debris: this new method should not be here, cause:
a) it's engine specific
b) it's used only when engine is created, so it should be a part of engine creation process

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -482,6 +484,8 @@ pub struct AuthorityRound {
 	two_thirds_majority_transition: BlockNumber,
 	maximum_empty_steps: usize,
 	machine: Machine,
+	benign_reporting: Mutex<ValidatorReporting>,
+	malicious_reporting: Mutex<ValidatorReporting>,
```
debris: those values are set only once per lifetime of an engine - after it's creation, so they should not wrapped in the `Mutex` and should be set in `AuthorityRound::new` method

ethcore/engines/validator-set/reporting-config/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+description = "Validators reporting config"
+name = "validator-reporting-config"
```
debris: creating a new library for a single structure that is a config option for a `validator-set` library doesn't make sense. The content of this library should be a part of `validator-set` lib

ethcore/machine/src/tx_filter.rs
```diff
@@ -107,11 +107,12 @@ impl TransactionFilter {
 				match version_u64 {
 					2 => {
 						let (data, decoder) = transact_acl::functions::allowed_tx_types::call(sender, to, value);
-						client.call_contract(BlockId::Hash(*parent_hash), contract_address, data)
+						client.call_contract(BlockId::Hash(*parent_hash), CallOptions::new(contract_address, data))
+							.map_err(|e| e.to_string())
 							.and_then(|value| decoder.decode(&value).map_err(|e| e.to_string()))
 							.map(|(p, f)| (p.low_u32(), f))
 							.unwrap_or_else(|e| {
-								error!(target: "tx_filter", "Error calling tx permissions contract: {:?}", e);
```
debris: `{}` implies displaying less information than `{:?}`. Why was it changed?

ethcore/call-contract/src/call_contract.rs
VladLupashevskyi: I agree, that using this one once does not make really sense! That and `CallError` have been introduced by @tomusdrw suggestion in [this](https://github.com/paritytech/parity-ethereum/pull/10602#discussion_r294176475) comment. I understood that intention to use builder was because it was planned to be used in other places. And according to `CallError` usage I assumed this one was suggested as the base boilerplate and is gonna be extended in the future for other cases which actually depend on the error type. My bad - I didn't clarify that before implementation, and I don't mind getting rid of this (I think it's not a big deal with builder macro, but more unsure about `CallError`), but I would like first to clarify this in order to prevent reverting the changes back and forth.

ethcore/engine/src/engine.rs
VladLupashevskyi: Nice catch, thank you! :) I used `engine_signer` logic as an example to introduce this one, but I have missed that signer can be changed during the lifetime of the Engine, which we don't need to do in this case. So this one and mutexes shall go away.

ethcore/engines/validator-set/reporting-config/Cargo.toml
VladLupashevskyi: The main intention to move this into a separate module was the idea of not introducing other dependancies to the engine crate, which `validator-set` would bring with necessary to use just one struct from there. This was inspired by `call-contract` and other similar crates.

So you think it would be ok, to add `validator-set` to the Engine?

ethcore/machine/src/tx_filter.rs
VladLupashevskyi: Can't tell exactly now from top of my head, but I think this formatting is used because we map error to string before, so debug fmt is not really needed here anymore. Will double check this when I'm with my laptop next week.

ethcore/engines/validator-set/reporting-config/Cargo.toml
debris: > So you think it would be ok, to add `validator-set` to the Engine?

No, but if you use `ReportingConfig` in the construction phase of the `validator-set` engine, the `Engine` trait doesn't have to know about this `ReportingConfig` struct at all

ethcore/call-contract/src/call_contract.rs
debris: @tomusdrw is there any reason to have 2 separate enum variants for `CallError`? We don't use them at all

Also, do you think we should commit to `derive_builder` and if yes, should we also use it in other places?

debris:
Thank you @VladLupashevskyi for the answers! For now, let's change the way `validator-set` engine is initialized and let's wait for @tomusdrw answers regarding `call_contract.rs`. You are right that it doesn't make sense to make changes back and forth :)

VladLupashevskyi:
I was trying to change the way how `validator-set` engine is initialized and I found out that it will be a bit tricky to pass CLI args to it, since `AuthorityRound::new` finally uses `Spec::load` which is dependent on json specification and `SpecParams`, so I think the simplest way to do that is pass `ReportingConfig` there. However I'm not sure that it logically would fit there.

@debris What do you think about this approach, would it make sense to extend `SpecParams`?

ethcore/engines/validator-set/reporting-config/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+	Disable
+}
+
+impl ReportingConfig {
+	pub fn from_str_and_call_every(config_str: Option<String>, call_every: u64) -> Result<Self, String> {
+		if let Some(s) = config_str {
+			let s = s.as_str();
+			match s {
+				"force" => Ok(ReportingConfig::Force),
+				"call" => Ok(ReportingConfig::Call(call_every)),
```
dvdplm: Not sure if this fits but it seems like reporting frequency is a concern of the engine impl. If so, maybe you can get rid of the `u64` here and just have `Force`, `CallEvery` and `Disable`; then the engine will know what frequency to use for the `CallEvery` case?

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -61,13 +65,34 @@ impl ValidatorContract {
...
 
 		match client.as_full_client() {
 			Some(c) => {
+				let latest_block_num = c.block_number(BlockId::Latest).expect("Latest block is never None");
+				match reporting.config {
+					ReportingConfig::Disable => return Err("Reporting is disabled.".into()),
+					ReportingConfig::Force => {},
+					ReportingConfig::Call(blk_num) => {
+						let checked_last_block = reporting.checked_last_block;
+						if checked_last_block.is_none() || (latest_block_num - checked_last_block.expect("checked for None before")) >= blk_num {
```
dvdplm: Maybe easier to read?
```suggestion
						if latest_block_num - checked_last_block.unwrap_or(0) >= blk_num {
```

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -61,13 +65,34 @@ impl ValidatorContract {
...
+							reporting.update_cache(latest_block_num, false);
+                            let call_options = CallOptionsBuilder::default()
+                                .contract_address(self.contract_address)
+                                .data(data.clone())
+                                .sender(sender)
+                                .build().unwrap();
```
dvdplm: No `unwrap` please.

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -61,13 +65,34 @@ impl ValidatorContract {
...
+							c.call_contract(BlockId::Latest, call_options).map_err(|e| e.to_string())?;
```
dvdplm: Can we get rid of the stringified errors?

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -213,7 +238,7 @@ mod tests {
 		assert_eq!(client.chain_info().best_block_number, 1);
 		// Check if the unresponsive validator is `disliked`.
 		assert_eq!(
-			client.call_contract(BlockId::Latest, validator_contract, "d8f2e0bf".from_hex().unwrap()).unwrap().to_hex(),
+			client.call_contract(BlockId::Latest, CallOptions::new(validator_contract, "d8f2e0bf".from_hex().unwrap())).unwrap().to_hex(),
```
dvdplm: A comment showing where `"d8f2e0bf"` comes from?

ethcore/engines/validator-set/src/lib.rs
```diff
@@ -61,6 +63,29 @@ pub fn new_validator_set(spec: ValidatorSpec) -> Box<dyn ValidatorSet> {
...
+	cached_result: bool
+}
+
+impl ValidatorReporting {
+	pub fn new(config: ReportingConfig) -> Self {
+		ValidatorReporting{
+			config,
+			checked_last_block: Default::default(),
+			cached_result: Default::default()
+		}
```
dvdplm: `ValidatorReporting {config, ..Default::default() }`

ethcore/engines/validator-set/src/lib.rs
```diff
@@ -61,6 +63,29 @@ pub fn new_validator_set(spec: ValidatorSpec) -> Box<dyn ValidatorSet> {
 	}
 }
 
+#[derive(Default)]
+pub struct ValidatorReporting {
```
dvdplm: Document all public items in this file.

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -294,7 +295,7 @@ impl ValidatorSet for ValidatorSafeContract {
 			.ok_or_else(|| "No client!".into())
 			.and_then(|c| {
 				match c.as_full_client() {
-					Some(c) => c.call_contract(id, addr, data),
+					Some(c) => c.call_contract(id, CallOptions::new(addr, data)).map_err(|e| e.to_string()),
```
dvdplm: Any chance we can avoid the `to_string()` error here?

ethcore/machine/src/tx_filter.rs
```diff
@@ -109,11 +109,12 @@ impl TransactionFilter {
 				match version_u64 {
 					2 => {
 						let (data, decoder) = transact_acl::functions::allowed_tx_types::call(sender, to, value);
-						client.call_contract(BlockId::Hash(*parent_hash), contract_address, data)
+						client.call_contract(BlockId::Hash(*parent_hash), CallOptions::new(contract_address, data))
+							.map_err(|e| e.to_string())
```
dvdplm: :/

parity/cli/mod.rs
```diff
@@ -834,6 +834,19 @@ usage! {
 			"--max-round-blocks-to-import=[S]",
 			"Maximal number of blocks to import for each import round.",
 
+		["Reporting configuration"]
```
dvdplm:
```suggestion
		["Validator behaviour reporting"]
```

parity/cli/mod.rs
```diff
@@ -834,6 +834,19 @@ usage! {
...
+			ARG arg_benign_reporting: (Option<String>) = None, or |c: &Config| c.reporting.as_ref()?.benign_reporting.clone(),
+			"--benign_reporting=[force|call|disable]",
+			"AuRa benign reporting configuration.",
```
dvdplm: Document what the options mean. And defaults.

parity/cli/mod.rs
```diff
@@ -834,6 +834,19 @@ usage! {
...
+
+			ARG arg_malicious_reporting: (Option<String>) = None, or |c: &Config| c.reporting.as_ref()?.malicious_reporting.clone(),
+			"--malicious_reporting=[force|call|disable]",
+			"AuRa malicious reporting configuration.",
```
dvdplm: Document what the options mean. And defaults.

parity/cli/mod.rs
```diff
@@ -834,6 +834,19 @@ usage! {
...
+
+			ARG arg_report_cache_call_every: (u64) = 100u64, or |c: &Config| c.reporting.as_ref()?.cache_call_every.clone(),
+			"--report_cache_call_every=[BLOCK_NUM]",
+			"AuRa cache call report successfulness for specified amount if blocks.",
```
dvdplm: Not sure I understand what this means.

PR #11035, opened 2019-09-10
Benchmarks for block verification

dvdplm:
Uses real blocks from mainnet to benchmark the `verify_*` family of methods
in the `verification` module.

**NOTE** the functions benchmarked here are `pub` but the module is not exposed publicly. To work around this it is necessary to run the benchmarks with `cargo bench --features=bench`. [Prior art](https://github.com/paritytech/substrate/blob/master/srml/staking/src/lib.rs#L246).

Results:

```
verify_block_basic      time:   [309.11 us 310.68 us 312.20 us]
                        change: [+0.0444% +3.8884% +7.8211%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  6 (6.00%) high severe

Benchmarking verify_block_unordered: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 288.1s or reduce sample count to 10
verify_block_unordered  time:   [25.622 ms 25.870 ms 26.151 ms]
                        change: [-5.8548% -3.4823% -0.9301%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe

verify_block_family (partial)
                        time:   [348.14 ns 350.11 ns 352.17 ns]
                        change: [-3.3510% -0.3836% +2.6583%] (p = 0.82 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe

verify_block_family (full)
                        time:   [615.22 us 620.32 us 625.74 us]
                        change: [-6.1034% -3.6971% -0.3676%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

Also exposes the `TestBlockChain` in a test helper.


ethcore/verification/src/test_helpers.rs
```diff
@@ -0,0 +1,124 @@
...
+use common_types::{
+	BlockNumber,
+	encoded,
+	verification::Unverified,
+	log_entry::{LogEntry, LocalizedLogEntry},
+};
+use ethereum_types::{BloomRef, H256};
+use parity_bytes::Bytes;
+
+pub struct TestBlockChain {
```
niklasad1: This probably just moved but can you `derive(Default)` and remove the `default impl`?

Also, I think the `new` impl should be removed because it is the same the default but if is used a lot don't bother

Cargo.toml
```diff
@@ -122,7 +122,7 @@ name = "parity"
 
 [profile.release]
 debug = false
-lto = true
+lto = false
```
niklasad1: intended?

ethcore/src/block.rs
```diff
@@ -392,7 +392,10 @@ impl SealedBlock {
 		block_rlp.append(&self.block.header);
 		block_rlp.append_list(&self.block.transactions);
 		block_rlp.append_list(&self.block.uncles);
-		block_rlp.out()
+		let o= block_rlp.out();
```
niklasad1: shall this be removed? 
I suppose `target: "dp" == David Palm :smile: ?

ethcore/types/src/verification.rs
```diff
@@ -57,7 +57,7 @@ impl VerificationQueueInfo {
 }
 
 /// An unverified block.
-#[derive(PartialEq, Debug, MallocSizeOf)]
+#[derive(Clone, PartialEq, Debug, MallocSizeOf)]
```
niklasad1: `Clone` is used for benchmarks/tests? 

Why not `#[cfg_attr(any(foo, bar), derive(Clone)]`?

ethcore/verification/benches/verification.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use common_types::verification::Unverified;
+use criterion::{Criterion, criterion_group, criterion_main};
+use ethash::{EthashParams, Ethash};
+use ethereum_types::U256;
+use ethcore::client::TestBlockChainClient;
+use spec::new_constantinople_test_machine;
+use tempdir::TempDir;
+
+use ::verification::{
```
niklasad1:
```suggestion
use verification::{
```

ethcore/verification/benches/verification.rs
```diff
@@ -0,0 +1,163 @@
...
+			m.insert(4370000, 3000000000000000000u64.into());
+			m.insert(7280000, 2000000000000000000u64.into());
+			m
+		},
+		expip2_transition: u64::max_value(),
+		expip2_duration_limit: 30,
+		block_reward_contract_transition: 0,
+		block_reward_contract: None,
+		difficulty_bomb_delays: {
+			let mut m = BTreeMap::new();
```
niklasad1: or use the macro `map!`

ethcore/verification/benches/verification.rs
```diff
@@ -0,0 +1,163 @@
...
+	Ethash::new(
+		cache_dir.path(),
+		ethash_params,
+		machine,
+		None
+	)
+}
+
+fn block_verification(c: &mut Criterion) {
+	const PROOF: &'static str = "bytes from disk are ok";
```
niklasad1:
```suggestion
	const PROOF: &str = "bytes from disk are ok";
```

ethcore/verification/benches/verification.rs
```diff
@@ -0,0 +1,163 @@
...
+	let mut block_provider = TestBlockChain::new();
+	block_provider.insert(rlp_8481476.clone()); // block to verify
+	block_provider.insert(rlp_8481475.clone()); // parent
+	block_provider.insert(rlp_8481474.clone()); // uncle's parent
+
+	let client = TestBlockChainClient::default();
+	c.bench_function("verify_block_family (full)", |b| {
+		b.iter(|| {
+			let full = FullFamilyParams { block: &preverified, block_provider: &block_provider, client: &client };
+			let o= verification::verify_block_family::<TestBlockChainClient>(
```
niklasad1:
```suggestion
			let o = verification::verify_block_family::<TestBlockChainClient>(
```

.gitlab-ci.yml
```diff
@@ -101,8 +101,9 @@ cargo-check-benches:
   <<:                              *docker-cache-status
   script:
     - time (
-        cargo check --all --benches --exclude ethash --target $CARGO_TARGET --locked --verbose --color=always;
-        (cd ethash; time cargo check --benches --features bench --target $CARGO_TARGET --locked --verbose --color=always)
+        (cargo check --all --benches --exclude ethash --exclude verification --target $CARGO_TARGET --locked --verbose --color=always);
+        (pushd ethash; cargo check --benches --features bench --target $CARGO_TARGET --locked --verbose --color=always; popd);
```
ordian: Technically you don't need pushd/popd here as the commands are run in a subshell via `(...)`.
The problem with current approach is that it will ignore the result of the all but the last command, so e.g. if `cargo check --benches --exclude ethash --exclude verification` fails, but ethcore/verification succeeds, this check won't fail.
I think we should run each command as a separate step (`-`).

ethcore/types/src/verification.rs
dvdplm: Yes, that is the reason. It's more convenience than necessity though. In general though, why is it bad to derive `Clone`? Reading https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-common-traits it seems idiomatic to do so whenever it's convenient?

Cargo.toml
dvdplm: God no. So sloppy.

ethcore/verification/src/test_helpers.rs
dvdplm: https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-common-traits says "Note that it is common and expected for types to implement both Default and an empty new constructor. new is the constructor convention in Rust, and users expect it to exist, so if it is reasonable for the basic constructor to take no arguments, then it should, even if it is functionally identical to default." – which makes sense to me so I think I'll keep `new`.

ethcore/verification/benches/verification.rs
dvdplm: Unfortunately it's necessary:

```
--> ethcore/verification/benches/verification.rs:29:5
   |
29 | use verification::{
   |     ^^^^^^^^^^^^ ambiguous name
   |
   = note: `verification` could refer to a crate passed with `--extern`
   = help: use `::verification` to refer to this crate unambiguously
```

ethcore/verification/benches/verification.rs
dvdplm: eew, I hate that macro. I think explicit is ok here, no? :)

.gitlab-ci.yml
dvdplm: > I think we should run each command as a separate step (-).

Good point. Doing that.

ethcore/verification/benches/verification.rs
niklasad1: yes, it ok

ethcore/verification/src/test_helpers.rs
niklasad1: ok, then I learned something new. I thought default was for `"default"` and new with some params...

ethcore/types/src/verification.rs
niklasad1: nothing bad I guess but it might have the reasoning that such big structures should not be  cloned which are enforced by `rustc` but I don't know really just guessing!

ethcore/verification/src/test_helpers.rs
dvdplm: Yes, me too. I went looking at the guidelines for the comment about `Clone` and saw this.

PR #11036, opened 2019-09-11
[ethcore]: make it compile without `test-helpers` feature

niklasad1:
After [this change](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/client/client.rs#L2686) `rustc-hex` is not an optional dependency anymore which this PR fixes

As a bonus, I added docs to [DataFormat](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/types/src/data_format.rs) and replaced space indentation.

### DIFF
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -54,7 +54,7 @@ rand_xorshift = "0.1.1"
 rayon = "1.1"
 rlp = "0.4.0"
 rlp_derive = { path = "../util/rlp-derive" }
-rustc-hex = { version = "1", optional = true }
+rustc-hex = "1"
 serde = "1.0"
 serde_derive = "1.0"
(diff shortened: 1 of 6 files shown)


ethcore/Cargo.toml
```diff
@@ -54,7 +54,7 @@ rand_xorshift = "0.1.1"
 rayon = "1.1"
 rlp = "0.4.0"
 rlp_derive = { path = "../util/rlp-derive" }
-rustc-hex = { version = "1", optional = true }
+rustc-hex = "1"
```
dvdplm: Unless there's a better way to do hexing, can we use `v2` here?

niklasad1: well, I missed this one. Will fix it but looks like it needs some more type annotations..

PR #11039, opened 2019-09-12
[builtin]: support `multiple prices and activations` in chain spec

### LINKED ISSUE #11014: ethcore/builtins: refactor `Pricer` trait and introduce `JSON Address: builtinMulti`
[ethcore/builtins](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/builtin/src/lib.rs#L168-#L201) read the cost and other related data directly from JSON which works fine as long as there is only one value for each field.

However, [EIP 1108](https://eips.ethereum.org/EIPS/eip-1108) requires to reduce gas cost `bn128_operations` after a certain block number. Thus, it doesn't work that well and [#11008](https://github.com/paritytech/parity-ethereum/pull/11008) was a dirty fix.

It introduced the `current_block_number` in the [Pricer trait](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/builtin/src/lib.rs#L40-#L44)  (leaky abstraction) and has two different costs included in the `Builtin`.

So, a suggestion for this is as follows: 

We define a new type `builtinMulti` (as a field in `json::Account`). It takes a hashmap, where keys are block number, and values are `Builtin`. We then activate/deactivate builtins based on the block number. The resulting config will look something like this:

```
"accounts": {
  "0000000000000000000000000000000000000008": { "builtinMulti": {
    "0": { "name": "alt_bn128_pairing", "activate_at": 1000000, "pricing": { "alt_bn128_pairing": { "base": 100000, "pair": 80000 } } } },
    "12345678": { "name": "alt_bn128_pairing", "activate_at": 2000000, "pricing": { "alt_bn128_pairing": { "base": 900, "pair": 1000 } } } },
  }
}
```

The proposal could enhance the following:
1. Remove `block number` from `Pricer trait`
2. Revert pricing model `alt_bn128_const_operations`
3. Cleanup chain specs
4. Change costs via [is_active](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/builtin/src/lib.rs#L163) or something similar
5. Simplify further changes of builtin costs

### LINKED ISSUE #11063: Validating Aura node crashes on startup with 2.5.8
- **Parity Ethereum version**: 2.5.8
- **Fully synchronized**: yes
- **Network**: Volta Testnet (POA/Aura)
- **Restarted**: yes

Validator node crashed with this error message, when upgrading from 2.5.7 to 2.5.8 right at startup:
```
parity_1            | Loading config file from /parity/config/parity-signing.toml
parity_1            | Spec json is invalid: missing field `eip1108_transition_base` at line 142 column 21
parity_1            | Loading config file from /parity/config/parity-signing.toml 
```

Expectations: backward compatibility. it shouldn't be necessary to include the new EIPs if not needed

niklasad1:
This PR tries to close #11014 and close #11063

To summarize the important change with this PR is that the chain spec format for builtins which support old format (with single pricing/activation) and new format as `map` to support multiple pricing/activations follows:

```json	
"builtin": {
    "name": "single_activation_old_format",
    "pricing": { "linear": {"base": 60, "word": 12 } },
     "activate_at": 0
}

"builtin": {
    "name": "multi_prices_and_activation",
    "pricing": {
        "4370000": {
            "info": "Hardfork X enable EIP ZZ",
            "price": {
                "linear": {
                        "base": 500,
                        "word": 0
                }
            }
        },
        "36028797018963967": { 
           "info": "Hardfork Y enable EIP ZZ", 
            "price": {
                "linear": {
                        "base": 150,
                        "word": 0
                }
            }
        }
    }     
}
```


dvdplm:
Would something like this work?

```json
"builtin": {
	"name": "single_activation",
	"pricing": {
		"linear": {"base": 60, "word": 12 }, "activate_at": 87
	}
}

"builtin": {
    "name": "multi_prices_and_activation",
    "pricing": {
		[
			{ "linear": {"base": 60, "word": 12 }, "activate_at": 123 },
			{ "modexp": {"divisor": 3.14 }, "activate_at": 456123 },
		]
    }
}
```

niklasad1:
> Would something like this work?

Yes, except the floating-point/fraction part. I'm not really sure if it makes sense to change price model on the same `builtin` but I guess it is more future-proof than just changing values if it is needed.

dvdplm:
> makes sense to change price model on the same `builtin` but I guess it is more future-proof than just changing values if it is needed.

It is most likely completely non-sensical, but yeah, the world is a crazy place!

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +227,128 @@ impl Builtin {
...
-		let pricer: Box<dyn Pricer> = match b.pricing {
-			ethjson::spec::Pricing::Blake2F { gas_per_round } => {
-				Box::new(gas_per_round)
+		// TODO(niklasad1): change this to `try_from` and propogate the error accordingly
+		let native = EthereumBuiltin::from_str(&b.name).unwrap();
+		let pricer = match b.pricing {
+			ethjson::spec::builtin::Pricing::Single(pricer) => {
+				// TODO(niklasad1): change chain specs and don't enable by default!?
+				// Seems more reasonable to do it the otherway around but would be a breaking change
+				map![b.activate_at.map_or(0, Into::into) => pricer.into()]
```
dvdplm: So if `activate_at` is missing in the JSON, it's on? Do I read that right? Seems a bit dangerous, but if that's the current default we should probably not change it.

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +227,128 @@ impl Builtin {
...
+			}
+
+			ethjson::spec::builtin::Pricing::Multi(pricer) => {
+				assert!(!pricer.is_empty(), "chainspec contains multiple builtins; but no values found");
```
dvdplm: Can this be a `static_assert!`? Aren't we reading in the json file at compile time with `include_bytes!`?

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +227,128 @@ impl Builtin {
...
 					word: linear.word,
 				})
 			}
-			ethjson::spec::Pricing::Modexp(exp) => {
-				Box::new(ModexpPricer {
+			ethjson::spec::builtin::InnerPricing::Modexp(exp) => {
+				Pricing::Modexp(ModexpPricer {
 					divisor: if exp.divisor == 0 {
-						warn!(target: "builtin", "Zero modexp divisor specified. Falling back to default.");
+						warn!("Zero modexp divisor specified. Falling back to default.");
```
dvdplm:
```suggestion
						warn!("Zero modexp divisor specified. Falling back to default: 10.");
```

Also, why remove the `target: `? Is it not printed otherwise?

niklasad1: Probably a bad rebase...

ethcore/builtin/src/lib.rs
niklasad1: Yes, that is correct and I agree.

ethcore/builtin/src/lib.rs
niklasad1: That would be cool, I will check tomorrow

ordian: I don't think that's possible, rust doesn't have dependent types and you can't statically guarantee vec is not empty.

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +227,128 @@ impl Builtin {
...
 	fn from(b: ethjson::spec::Builtin) -> Self {
-		let pricer: Box<dyn Pricer> = match b.pricing {
-			ethjson::spec::Pricing::Blake2F { gas_per_round } => {
-				Box::new(gas_per_round)
+		// TODO(niklasad1): change this to `try_from` and propogate the error accordingly
+		let native = EthereumBuiltin::from_str(&b.name).unwrap();
+		let pricer = match b.pricing {
+			ethjson::spec::builtin::Pricing::Single(pricer) => {
+				// TODO(niklasad1): change chain specs and don't enable by default!?
+				// Seems more reasonable to do it the otherway around but would be a breaking change
```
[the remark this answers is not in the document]
niklasad1: Good observation,  AltBn128ConstOperations was only introduced in the previous hack but I can add them to be on the safe side

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +227,128 @@ impl Builtin {
...
 	}
 }
 
 impl From<ethjson::spec::Builtin> for Builtin {
 	fn from(b: ethjson::spec::Builtin) -> Self {
-		let pricer: Box<dyn Pricer> = match b.pricing {
-			ethjson::spec::Pricing::Blake2F { gas_per_round } => {
-				Box::new(gas_per_round)
+		// TODO(niklasad1): change this to `try_from` and propogate the error accordingly
+		let native = EthereumBuiltin::from_str(&b.name).unwrap();
```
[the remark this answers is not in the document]
niklasad1: I get the following error message for this:

```bash
Thread 'main' panicked at 'invalid builtin name: foo', ethcore/builtin/src/lib.rs:282
```

which I thought is equivalent to a manual panic? 

Otherwise, I have to change to have `Box<dyn Trait` instead of an enum.

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
niklasad1: Yepp, let's enable a warning for it

ethcore/builtin/src/lib.rs
niklasad1: well, can't you do in a build-script with a special deserialization function which denies empty vectors? 

However, we support loading custom chain specs at runtime so I think we can't guarantee that statically anyways,

ethcore/res/ethereum/classic.json
```diff
@@ -4415,45 +4415,49 @@
 				}
 			}
 		},
-		"0x0000000000000000000000000000000000000006": {
+		"0000000000000000000000000000000000000006": {
```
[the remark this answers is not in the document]
niklasad1: I would prefer `0x` thoughts? Previously I tried to use the same as in the file but here I think I just copy-pasted from another file (sloppy me)...

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
ordian: Ah, ok, so `unwrap` for `Result` panic with the proper message, but I would rather have an explicit panic instead of unwrap, since unwrap usually indicates that it cannot panic in practice, otherwise it's a bug.

ethcore/res/ethereum/classic.json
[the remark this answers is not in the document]
ordian: `0x` seems fine to me

ethcore/builtin/src/lib.rs
```diff
@@ -193,15 +195,29 @@ impl ModexpPricer {
...
+	///
+	/// Return the cost of the most recently activated pricer at the current block number.
+	///
+	/// If no pricer is actived `zero` is returned
+	///
+	/// If multiple `activation_at` has the same block number the last one is used
+	/// which follows the BTreeMap semantics
+	///
+	// NOTE(niklasad1): we could remove is_active and replace with `fn cost(...) -> Option<U256>`
+	// but it would require some changes that I don't want to do...
```
ordian: I actually prefer the current approach since it's more explicit. So if we won't change that, let's remove the NOTE.

ethcore/builtin/src/lib.rs
dvdplm: Maybe it shouldn't be a `Vec`? ;)

It doesn't matter here ofc, just curious. In theory it sounds like it should be possible to do static asserts on bytes from `include_bytes!`, but then again maybe it requires serde to run at compile time too...

ethcore/res/authority_round.json
```diff
@@ -50,44 +50,48 @@
...
 				"name": "alt_bn128_add",
-				"activate_at": 0,
-				"eip1108_transition": "0x7fffffffffffff",
-				"pricing": {
-					"alt_bn128_const_operations": {
-						"price": 500,
-						"eip1108_transition_price": 150
+				"pricing": [
+					{
+						"activate_at": 0,
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

ethcore/res/authority_round.json
```diff
@@ -50,44 +50,48 @@
...
 				"name": "alt_bn128_mul",
-				"activate_at": 0,
-				"eip1108_transition": "0x7fffffffffffff",
-				"pricing": {
-					"alt_bn128_const_operations": {
-						"price": 40000,
-						"eip1108_transition_price": 6000
+				"pricing": [
+					{
+						"activate_at": 0,
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

ethcore/res/authority_round.json
```diff
@@ -50,44 +50,48 @@
...
-				"eip1108_transition": "0x7fffffffffffff",
-				"pricing": {
-					"alt_bn128_pairing": {
-						"base": 100000,
-						"pair": 80000,
-						"eip1108_transition_base": 45000,
-						"eip1108_transition_pair": 34000
+				"pricing": [
+					{
+						"activate_at": 0,
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

ethcore/res/authority_round_block_reward_contract.json
```diff
@@ -53,44 +53,48 @@
...
 				"name": "alt_bn128_add",
-				"activate_at": 0,
-				"eip1108_transition": "0x7fffffffffffff",
-				"pricing": {
-					"alt_bn128_const_operations": {
-						"price": 500,
-						"eip1108_transition_price": 150
+				"pricing": [
+					{
+						"activate_at": 0,
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

ethcore/res/ethereum/callisto.json
```diff
@@ -76,44 +76,48 @@
...
-				"eip1108_transition": "0x7fffffffffffff",
-				"pricing": {
-					"alt_bn128_pairing": {
-						"base": 100000,
-						"pair": 80000,
-						"eip1108_transition_base": 45000,
-						"eip1108_transition_pair": 34000
+				"pricing": [
+					{
+						"activate_at": 20,
```
dvdplm: Is this `20` hex or dec? (This is why I like `0x`)

ethcore/res/ethereum/goerli.json
```diff
@@ -126,48 +126,52 @@
...
 				"name": "alt_bn128_add",
-				"activate_at": "0x00",
-				"eip1108_transition": "0x17d433",
-				"pricing": {
-					"alt_bn128_const_operations": {
-						"price": 500,
-						"eip1108_transition_price": 150
+				"pricing": [
+					{
+						"activate_at": "0x00",
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

ethcore/res/ethereum/goerli.json
```diff
@@ -126,48 +126,52 @@
...
-				"eip1108_transition": "0x17d433",
-				"pricing": {
-					"alt_bn128_pairing": {
-						"base": 100000,
-						"pair": 80000,
-						"eip1108_transition_base": 45000,
-						"eip1108_transition_pair": 34000
+				"pricing": [
+					{
+						"activate_at": "0x00",
```
dvdplm:
```suggestion
						"activate_at": "0x0",
```

json/src/spec/builtin.rs
```diff
@@ -56,17 +45,13 @@ pub struct AltBn128Pairing {
...
-	/// EIP 1108 transition price per point pair
-	pub eip1108_transition_pair: usize,
 }
 
 /// Pricing variants.
 #[derive(Debug, PartialEq, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "snake_case")]
-pub enum Pricing {
+pub enum InnerPricing {
```
dvdplm: Bikeshed, but I think the codebase in general prefers `SomethingInner` vs `InnerSomething`.

json/src/spec/builtin.rs
```diff
@@ -92,13 +75,35 @@ pub struct Builtin {
...
+	/// Multiple builtins
+	Multi(Vec<PricingWithActivation>),
+}
+
+/// Builtin price with which block to activate it on
+#[derive(Debug, PartialEq, Deserialize, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct PricingWithActivation {
+	/// Description of the activation
+	pub info: Option<String>,
```
dvdplm: 👍

json/src/spec/builtin.rs
```diff
@@ -92,13 +75,35 @@ pub struct Builtin {
...
-	/// EIP 1108
-	pub eip1108_transition: Option<Uint>,
+}
+
+/// Builtin price
+#[derive(Debug, PartialEq, Deserialize, Clone)]
+#[serde(rename_all = "snake_case")]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum Pricing {
```
dvdplm: What is the reason we can't make all chains be "Multi"? Like, what does having `Single`/`Multi` buy us? The only reason I can think of is backwards compatibility (which is a good reason). What would it take to make serde cope with both types and deserialize to a vec of `PricingWithActivation`?

json/src/spec/builtin.rs
```diff
@@ -92,13 +75,35 @@ pub struct Builtin {
...
+	/// Single builtin
+	Single(InnerPricing),
+	/// Multiple builtins
+	Multi(Vec<PricingWithActivation>),
+}
+
+/// Builtin price with which block to activate it on
+#[derive(Debug, PartialEq, Deserialize, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct PricingWithActivation {
```
dvdplm: Maybe name this `PricingAt`?

json/src/spec/spec.rs
```diff
@@ -23,6 +23,7 @@ use serde_json::Error;
 
 /// Fork spec definition
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize)]
+#[allow(missing_docs)]
```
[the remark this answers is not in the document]
dvdplm: Docs FTW. I'd love to see something like:
```
…
/// Homestead, active between #123  - #456, ~01/01/2016 - ~04/04/2017
Homestead,
…
```

json/src/spec/builtin.rs
```diff
@@ -108,7 +113,40 @@ mod tests {
 		}"#;
 		let deserialized: Builtin = serde_json::from_str(s).unwrap();
 		assert_eq!(deserialized.name, "ecrecover");
-		assert_eq!(deserialized.pricing, Pricing::Linear(Linear { base: 3000, word: 0 }));
+		assert_eq!(deserialized.pricing, Pricing::Single(InnerPricing::Linear(Linear { base: 3000, word: 0 })));
+		assert!(deserialized.activate_at.is_none());
+	}
+
+	#[test]
+	fn builtin_v2_deserialization() {
```
dvdplm:
```suggestion
	fn deserialize_multiple_pricings {
```

ethcore/res/ethereum/classic.json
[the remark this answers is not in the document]
niklasad1: let's take of them in another PR. 

so they get the attention they deserve

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
niklasad1: added an explicit panic

json/src/spec/builtin.rs
niklasad1: It is for backward compatibility only I would say.

I don't think it is possible to have both types and deserialize to `Vec<PricingWithActivation>` in a backward-compatible way. However, another option is to add another field `Vec<PricingWithActivation>` and deprecated the old one but I'm open to suggestions if you come with something better

dvdplm: I was sort of thinking about something akin to [`HashOrMap`](https://github.com/paritytech/parity-ethereum/blob/d9201aa6f2a03471c88f8c5638a7abe93d0db57c/json/src/spec/state.rs#L27-L38) that I used to deal with json like `"postState": "blabla"` vs `"postState" : { "stuff": "here"}`.
Not sure it'd work here but the situation there was similar: support two different types in the input JSON.

ethcore/res/spec_backward_compability.json
```diff
@@ -0,0 +1,202 @@
...
...000806040516020840160008287838a8c6187965a03f1925050508091505b50949350505050565b81548183558181151161200f5781836000526020600020918201910161200e91906120e8565b5b505050565b81548183558181151161203b5781836000526020600020918201910161203a91906120e8565b5b505050565b602060405190810160405280600081525090565b602060405190810160405280600081525090565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106120a957805160ff19168380011785556120d7565b828001600101855582156120d7579182015b828111156120d65782518255916020019190600101906120bb565b5b5090506120e491906120e8565b5090565b61210a91905b808211156121065760008160009055506001016120ee565b5090565b905600a165627a7a72305820f1129b699b3017535535a920e15503cd06af1f5c77637c0637cc29355b1dad3400290000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000040000000000000000000000000cb1437200aea736788f1fc56f327c0456c3598d00000000000000000000000074dd76e24b2cfb43c1b1a4498295d553d0843746000000000000000000000000eeb4ceee443f9e0d17bdbd6daa241681ee5e51c2000000000000000000000000a005caea55375ae20e3aaef746113535503abc19"
```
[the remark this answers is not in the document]
niklasad1: no, I just copied the entire file.

ethcore/builtin/src/lib.rs
```diff
@@ -105,26 +113,32 @@ struct AltBn128PairingPrice {
...
-		let price = if at >= self.eip1108_transition_at {
-			self.eip1108_transition_price
-		} else {
-			self.price
-		};
+/// Pricing for constant alt_bn128 operations
+#[derive(Debug, Copy, Clone)]
+pub struct AltBn128ConstOperations {
+	/// price
+	pub price: usize,
```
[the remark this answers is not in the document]
niklasad1: It is a bug, should be u64.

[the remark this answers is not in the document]
niklasad1: No, not really, I think I used it because the other pricing models used `usize`.

Are you thinking of `platform-dependent` issues when `price > u32::max_value()`?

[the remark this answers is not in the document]
ordian: yeah, just wondering why not `u64` in the first place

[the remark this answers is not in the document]
niklasad1: that's an excellent observation, shall I address it in this PR or create an issue?

json/src/spec/builtin.rs
niklasad1: I don't understand what you are after here. 
Pricing is similar to `HashOrMap` in your example but with an additional Multi. If you are referring to `Multi<Vec<PricingAt>` could be replaced `Vec<PricingAt>` that should be possible.

Otherwise, explain :wave:

json/src/spec/spec.rs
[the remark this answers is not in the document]
niklasad1: I was not sure what to write there but addressed in a separate PR [#11138](https://github.com/paritytech/parity-ethereum/pull/11138) instead

ethcore/builtin/src/lib.rs
```diff
@@ -211,89 +238,187 @@ impl Builtin {
...
-			pricer,
-			native: ethereum_builtin(&b.name),
-			activate_at: b.activate_at.map_or(0, Into::into),
+// Convert `JSON pricing` into Pricing
+//
+// Because `AltBn128XX` has two different pricings for `eip1108 transition` both need
+// to be taken into account.
+//
+// Returns `(Pricing, Some(Pricing))` if eip1108_transition is enabled with an activation
+// otherwise it returns `(Pricing, None)`
```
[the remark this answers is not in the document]
niklasad1: Yes, I agree. 
It would be good if we could remove the mess from #11008 but it might annoy people that recently updated their chain spec except maybe the type `Const/Fixed price` instead of Linear. If we decide to remove the hack we must add something in the changelog. 

Lesson: don't write hacks that are supposed to be backward compatible

json/src/spec/builtin.rs
dvdplm: > If you are referring to `Multi<Vec<PricingAt>` could be replaced `Vec<PricingAt>` that should be possible.

Yep, that's it: avoid having the `Single`/`Multi` distinction in the code and instead deal with input JSON being effectively different for the single/multi cases when deserializing directly. Worth doing here in your opinion?

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
dvdplm: I guess it depends a bit on whether we want this backported or not. I think we do want this backported (but unsure tbh) and in that case we should avoid breaking peoples setups. We could also opt to not backport it and have this go in to the 2.7 release right after the Istanbul HF, in which case it's fine to break things.

json/src/spec/builtin.rs
niklasad1: all right, I see. Yes, it is worth doing

niklasad1: well, I think I got confused by this again after investigated it or understood it wrong again sorry. 

The problem is that current chain spec has the following format for pricing:

```bash
// legacy
"pricing": { "linear": {"base": 60, "word": 12 }, "activate_at": 87}
// new
"pricing": [ ],
```
Having it deserialize to a `Vec` directly can't be done because it is the same type currently. Thus, `HashOrMap` example doesn't help because you still need to match on it to figure out whether it is `Array<Vec<T>>` or `Object(T)` at runtime.

As I see it we could do:
1. `provide a method `to_vec` which does the matching and then returns `Vec<PricingAt>`
2. create a new type `BuiltinV2` which just has an inner type `Vec/BTreeMap<PricingAt>`.

Have I got anything wrong or other suggestions?

dvdplm: > Have I got anything wrong or other suggestions?

Most likely I made it simpler than it is. Let me check out the code and play with it.

dvdplm: So, a tricky one for sure.
What I had in mind was having `pub struct Pricing(Vec<PricingAt>)` and then have a `#[serde(deserialize_with = "bla")]`. In `bla()` we'd check if it's a JSON `Array` or `Object` (or garbage, i.e. error) and the deserialize the right type:

 - if array, then assume it's a vec of `PricingAt`
 - if object, then check if it's a legacy `PriceInner` and if yes, deserialize it and build a `PriceAt` out of it; if no, deserialize a single `PricingAt`. Wrap in a `Vec` and return.

The point of it all is to move the messy input format handling into one spot and let the rest of the Rust code only ever see `Vec<PriceAt>`.

My serde is weak so I'm not sure how deep of a rabbit hole this is. I had a quick stab at it and failed so…

niklasad1: Ok, thanks for the explanation. I can give it a try, would be nice to include to fail empty `Vec` as then I guess

json/src/spec/account.rs
```diff
@@ -18,7 +18,7 @@
 
 use std::collections::BTreeMap;
 
-use crate::{bytes::Bytes, spec::builtin::Builtin, uint::Uint};
+use crate::{bytes::Bytes, spec::builtin::BuiltinCombat, uint::Uint};
```
dvdplm: Combat? Did you mean "compat"? I mean, it's a struggle for sure, I feel your pain! :)

niklasad1: Yes, thanks

json/src/spec/builtin.rs
niklasad1: I didn't get it to work, see https://github.com/niklasad1/serde-manual-deserialize/blob/master/src/lib.rs#L74-#L120

So, I created a `compatibility layer` instead...

ethcore/builtin/src/lib.rs
niklasad1: damn, it got lost again...  needs to be fixed!

ethcore/builtin/src/lib.rs
```diff
@@ -213,93 +240,120 @@ impl Builtin {
 
 	/// Whether the builtin is activated at the given block number.
 	pub fn is_active(&self, at: u64) -> bool {
-		at >= self.activate_at
+		self.pricer.keys().rev().any(|&other_at| at >= other_at)
```
dvdplm: `is_active()` is called from `machine` and `executive`; if this is a "hot" function maybe this onerous?

niklasad1: you mean like this would be `n times` slower because we might need to iterate through the entire `BTreeMap` in the worst case instead of a constant time operation.

But, I think it is very likely that one or two iterations of the `BTreeMap`, otherwise is the transitions enabled too early I guess.

Still, accessing the first element through an iterator 20 times slower than the previous version on my machine :cry: but still if we want more than one activation at runtime it will be slower regardless....

So, I don't know

niklasad1: I will investigate if it is easy to create a benchmark for this, in `machine`

ethcore/builtin/Cargo.toml
```diff
@@ -7,17 +7,18 @@ edition = "2018"
 
 [dependencies]
 bn = { git = "https://github.com/paritytech/bn", default-features = false }
+byteorder = "1.3.2"
+eip-152 = { path = "../../util/EIP-152" }
```
debris: not an issue, but I believe that all deps were supposed to be in order :)
if you're using vim, it is `ctrl-v` + `:sort`

json/src/spec/builtin.rs
```diff
@@ -16,50 +16,75 @@
...
+	pub price: u64,
 	/// EIP 1108 transition price
-	pub eip1108_transition_price: usize,
+	// for backward compatibility
+	pub eip1108_transition_price: Option<u64>,
+}
+
+impl AltBn128ConstOperations {
+	fn clear_eip1108(&mut self) {
+		self.eip1108_transition_price = None;
```
debris: Not sure what is the purpose of this function. This crate should be used only for parsing

json/src/spec/builtin.rs
```diff
@@ -16,50 +16,75 @@
...
 	/// EIP 1108 transition price per point pair
-	pub eip1108_transition_pair: usize,
+	// for backward compatibility
+	pub eip1108_transition_pair: Option<u64>,
+}
+
+impl AltBn128Pairing {
+	fn clear_eip1108(&mut self) {
+		self.eip1108_transition_base = None;
+		self.eip1108_transition_pair = None;
```
debris: Not sure what is the purpose of this function. This crate should be used only for parsing

ethcore/builtin/Cargo.toml
niklasad1: yeah, I did before but I must have messed it up when sorting out merge conflicts.

json/src/spec/builtin.rs
niklasad1: Yeah, that's not clear sorry (it's mess).  To give some context, see https://github.com/paritytech/parity-ethereum/pull/11039#discussion_r331566965

Thus, this crate if converts `old format` to `new_format` to avoid doing it elsewhere, ideally we should do it while deserializing and not have to do it at runtime to match `object` or `map`.

An edge case is when a `single activation` has `eip1108_transitions` then it's converted into a `BTreeMap` with two entries and sets the `eip1108 params` to `None` to make sure that those are used anymore.

niklasad1:
> Instead of having 2 different formats and doing some hacks while parsing, why can't we have just another pricing type that internally has a map

Right, I think that it will still be `hacky` because then we need to check if the map is of `length 1` and has another the field `activate_at` also the map still needs to have the same type so how could the deserialization work?

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
niklasad1: I think it will become too hard to maintain and a lot of ugliness. Let's keep the old format but deprecate the `eip1108 params` and don't backport?

ordian:
What's the status of this? The chainspec tests are failing and there is a merge conflict.

niklasad1:
> What's the status of this? The chainspec tests are failing and there is a merge conflict.

There a three additional things to be sorted out:
1. I reverted the `backward compatibility` w.r.t. `eip1108_params` that's why the chain specs fail but I want some comments if we agree to remove those or not? Further, discussion [here](https://github.com/paritytech/parity-ethereum/pull/11039#discussion_r331809690)
2. David's [grumble](https://github.com/paritytech/parity-ethereum/pull/11039#discussion_r334365455) on performance regression (
3. Mareks [grumble](https://github.com/paritytech/parity-ethereum/pull/11039#pullrequestreview-301861240) on `chain spec format` + `json crate does additional parsing`

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
dvdplm: > Let's keep the old format but deprecate the eip1108 params and don't backport?

Works for me. Ok by you @ordian ?

ethcore/builtin/src/lib.rs
niklasad1: I benched it locally,

- Single activation, ~10% slower
- 10 activations where block number lower than lowest block number in the map: ~30% slower (i.e, most iterate through entire map `O(n)`)
- 10 activations block number bigger than highest block number, ~10% slower

ordian: you could do a binary search with
`pricer.range(0..=at).last().is_some()`, not sure if it's faster for a small number of activations though

niklasad1: It is definitely faster for 10 elements seems slower for 1 element though.

Let me check

ethcore/builtin/src/lib.rs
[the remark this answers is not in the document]
niklasad1: > And by deprecate you don't mean remove, 

Sorry, I meant remove but I reverted it now, 

> so it will add complexity to the code

https://github.com/paritytech/parity-ethereum/pull/11039/commits/d58ef6198663e17423f04057d4e2652e85be789b is the complexity it introduces

> I would rather go for "annoy people that recently updated their chain spec" approach to break compatibility (but show them how to upgrade in the release notes), but only do it once with a future-proof API.

Yes, I agree

dvdplm:
>     2\. David's [grumble](https://github.com/paritytech/parity-ethereum/pull/11039#discussion_r334365455) on performance regression

Thank you for checking on the perf. Given your findings I think we're fine in this regard.

dvdplm:
> I reverted the backward compatibility w.r.t. eip1108_params that's why the chain specs fail but I want some comments if we agree to remove those or not? Further, discussion here

@ordian's argument is convincing: I say we annoy people now, once, and try to do our best to explain how to change the chainspec in the rel notes.

niklasad1:
> @niklasad1 if I understand correctly, the only unresolved question at this point if whether we want to keep eip1108-* fields in the spec for backwards compatibility. It seems like me and @dvdplm are for removing it. If you agree, let's remove it and merge the PR. It also would be good to backport the PR, so that we don't have to maintain two spec formats.

Yeah, I agree let's remove then and update chain specs then.

util/network-devp2p/src/host.rs
```diff
@@ -37,7 +37,6 @@ use mio::{
 use parity_path::restrict_permissions_owner;
 use parking_lot::{Mutex, RwLock};
 use rlp::{Encodable, RlpStream};
-use rustc_hex::ToHex;
```
niklasad1: sorry, this is unrelated

niklasad1:
> I'm not sure about omitting is_active check in some places, but other than that looks good.

Fair enough, I can revert those to be on the safe side, however, machine calls `is_active` three times instead of one. I can handle that in another PR that doesn't have to be backported

ethcore/builtin/src/lib.rs
```diff
@@ -198,111 +214,151 @@ impl ModexpPricer {
...
 
 impl Builtin {
 	/// Simple forwarder for cost.
+	///
+	/// Return the cost of the most recently activated pricer at the current block number.
+	///
+	/// If no pricer is actived `zero` is returned
+	///
+	/// If multiple `activation_at` has the same block number the last one is used
+	/// which follows the BTreeMap semantics
```
dvdplm:
```suggestion
	/// (follows `BTreeMap` semantics).
```

ethcore/builtin/src/lib.rs
```diff
@@ -198,111 +214,151 @@ impl ModexpPricer {
...
 	}
 
 	/// Simple forwarder for execute.
+	#[inline]
 	pub fn execute(&self, input: &[u8], output: &mut BytesRef) -> Result<(), &'static str> {
 		self.native.execute(input, output)
 	}
 
 	/// Whether the builtin is activated at the given block number.
+	#[inline]
```
dvdplm: Just curious: did you measure that these `#[inline]`s matter or is it hunch-based optimization?

niklasad1: I think those were already inlined thus I didn't see any measurable difference IIRC, do want the annotations removed?

dvdplm: No, it's fine, I was just curious.

PR #11042, opened 2019-09-12
Increase default gas limit to 11M

sandakersmann:
The default gas limit should be increased in Geth and Parity. This will signal to miners that developers think that new limit is safe. Some miners are also lazy and do not adjust defaults. Let's get it done!


sandakersmann:
What should I change to make this PR have the desired effect? I can push those changes if you give me some pointers :)

joshua-mir:
It's a fairly trivial change to these values:

https://github.com/paritytech/parity-ethereum/blob/b6415c61963bb1d415ff0e173a78cdb8d2c65fa0/parity/cli/mod.rs#L761-L767

I'm still unsure about what good values for the gas target and cap would be to set as a default for all parity clients. There's no "per-network node specification", and this also affects nodes running other networks that might be affected poorly by the change.

sandakersmann:
If other networks have issues they can adjust the limit. I think the limit should be set mainly with the Ethereum network in mind, and in Ethereum this change is required now to avoid a dangerous economic situation on the network.

joshua-mir:
Is there a specific reason you changed the musiccoin chainspec and a chainspec for one of the musiccoin tests? I understand changing the instantSeal/dev chain in this PR but I'm not really sure the musiccoin changes should be

sandakersmann:
No specific reason. I can remove the musiccoin changes if you want.

sandakersmann:
Should I remove them?

joshua-mir:
Yes please? 😅

dvdplm:
@sandakersmann Can you fix the failing tests please?

ordian:
Linking discussion in geth: https://github.com/ethereum/go-ethereum/pull/20066

dvdplm:
@sandakersmann sorry approving/disapproving this. We will await the outcome of the AllCoreDevs call before doing this. :)

dvdplm:
> How can I fix the failing test?

Apologies, you do not see the CI portion of this page. Never mind, we'll take care of it when the time comes (or you'd need to check out the repo and run the tests, spot the failing one, `configuration::tests::test_run_cmd`, and fix it).

PR #11047, opened 2019-09-13
[build]: remove `test-helper feat` from build

niklasad1:
We have unintentionally built parity with the feature `test-helpers` which should only be enabled for running tests which this PR fixes.

Could reduce the binary size but I don't think it would be much...

### DIFF
--- a/ethcore/Cargo.toml
+++ b/ethcore/Cargo.toml
@@ -97,7 +97,11 @@ parity = ["work-notify", "price-info", "stratum", "macros"]
 # but might be omitted for other dependent crates.
 work-notify = ["ethcore-miner/work-notify"]
 price-info = ["ethcore-miner/price-info"]
-stratum = ["ethcore-stratum"]
+stratum = [
+	"ethash",
+	"ethcore-stratum"
+]
+
 
 # Disables seal verification for mined blocks.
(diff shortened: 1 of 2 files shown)


dvdplm:
This needs backporting too I'm afraid.

ordian:
> This needs backporting too I'm afraid.

Why? ethash is not an optional dependency of ethcore on beta and stable

dvdplm:
> Why?

Because `rpc` builds `ethcore` with `test-helpers` enabled.

https://github.com/paritytech/parity-ethereum/pull/11047/files#diff-1089f02be9842f0dbbe6b280f927c886L42

dvdplm:
Backport PRs:

https://github.com/paritytech/parity-ethereum/pull/11048

https://github.com/paritytech/parity-ethereum/pull/11049

niklasad1:
> Eew. Did I do this? Most likely. Apologies.

Introduced by https://github.com/paritytech/parity-ethereum/commit/4938d5dde5c83403e6cffc7301ebd35cbdcf04a2#diff-1089f02be9842f0dbbe6b280f927c886 so you don't need to apologize but this kind of things are really hard to catch anyways.

PR #11054, opened 2019-09-13
Update JSON tests to d4f86ecf4aa7c

dvdplm:
This PR started out as an attempt to finish https://github.com/paritytech/parity-ethereum/pull/10923 but ended up being something else.

- update the json tests to the latest commit on the upstream `develop` branch, [d4f86ecf4aa7c](https://github.com/ethereum/tests/commit/d4f86ecf4aa7c44a40bc0c972fd3e25d63ef5d92)
- fix the json parsing code to accept the new test formats (see https://github.com/ethereum/tests/issues/637)
- cleans up the code for readability
- remove some dead code and unneeded re-exports
- skips all failing tests

Questions for reviewers:

- having two different mechanisms for skipping tests is confusing: what's a better way? Skipping whole test files is easy but coarse; skipping test-by-test is very cumbersome but precise. Do we need both?
- currently one test-skipping mechanism is hidden behind a feature (`ci-skip-tests`). Is this really what we want?
- what else can we do to make maintenance of json tests less painful than it currently is?

Open issues for fixing the failing tests:

- [ ] stRevertTest blockchain test failures, #11073
- [x] stReturnDataTest blockchain test failures, #11074
- [x] stSpecialTest blockchain test failures, #11075
- [x] stStaticCall blockchain test failures, #11076
- [x] stTransactionTest blockchain test failures, #11077
- [ ] stRevertTest/RevertPrecompiledTouch_storage state test failures, #11078
- [ ] stRevertTest/RevertPrecompiledTouchExactOOG state test failures, #11079
- [ ] stRevertTest/RevertPrecompiledTouch state test failures, #1080

Previous description, now outdated:

This PR reverts the controversial changes of the previous PR and skips the failing tests.

Maybe I misunderstand the [suggested workaround](https://github.com/paritytech/parity-ethereum/pull/10923#pullrequestreview-267525958) of putting the fix under `#[cfg(test)]` but it seems odd to run different code in production than we run in tests. 
Instead here I suggest we skip the failing tests with the argument that we do not wish to fix this issue (at least not at this time) because it does not affect us. If I am wrong, and I likely am, I look
forward to hearing why and what a better approach to updating the state tests is.

We need to update the state tests again very soon for Istanbul so there's no avoiding this discussion.

Branched off https://github.com/paritytech/parity-ethereum/pull/10923

ref #10908


niklasad1:
@dvdplm 

I got a bit confused because of `title update JSON tests to 725dbc73a` 

[725dbc73a](https://github.com/ethereum/tests/commit/725dbc73a54649e22a00330bd0f4d6699a5060e5) is used already on master and really old?

Is that correct? 

In general, I agree with the change in skipping the tests rather than changing production code but I'm that not familiar with these tests TBH.

dvdplm:
> [725dbc73a](https://github.com/ethereum/tests/commit/725dbc73a54649e22a00330bd0f4d6699a5060e5) is used already on master and is really old?

@niklasad1 Yes it's the same as master. I must have done something wrong when branching off https://github.com/paritytech/parity-ethereum/pull/10923

Good catch!

json/src/spec/state.rs
```diff
@@ -19,38 +19,74 @@
...
+pub enum HashOrMap {
+	/// When the `postState` is large, tests sometimes just include the state root of the last
+	/// successful block here.
+	Hash(H256),
+	/// The expected `postState` of a test
+	Map(BTreeMap<Address, Account>),
+}
+
+/// Recent JSON tests can be either a map or a string. See https://github.com/ethereum/tests/issues/637
+// todo[dvdplm]: what is the point of doing this for non-test code? Needed?
```
niklasad1: it is just for the `derive Clone` but let's merge them and have it implement `Clone`

json/src/spec/state.rs
```diff
@@ -19,38 +19,74 @@
...
 	/// Returns all builtins.
 	pub fn builtins(&self) -> BTreeMap<Address, Builtin> {
-		self.0
-			.iter()
-			.filter_map(|(add, ref acc)| acc.builtin.clone().map(|b| (add.clone(), b)))
-			.collect()
+		match &self.0 {
+			HashOrMap::Hash(_) => BTreeMap::default(),
+			HashOrMap::Map(map) => {
+				map.iter().filter_map(|(add, ref acc)| {
```
niklasad1: impl `Iterator` instead of having to do this matching every time?

ethcore/src/json_tests/chain.rs
```diff
@@ -89,16 +90,25 @@ pub fn json_chain_test<H: FnMut(&str, HookType)>(json_data: &[u8], start_stop_ho
 					config.check_seal = false;
 				}
 				config.history = 8;
+				config.queue.verifier_settings.num_verifiers = 1;
```
dvdplm: The default is `num_cpus` which includes HT and means we spin up a lot of threads for each test. I think 1 thread is enough here.

ethcore/src/json_tests/mod.rs
```diff
@@ -30,17 +30,7 @@ mod skip;
 mod difficulty;
 
 pub use self::test_common::HookType;
-
-pub use self::transaction::run_test_path as run_transaction_test_path;
-pub use self::transaction::run_test_file as run_transaction_test_file;
 pub use self::executive::run_test_path as run_executive_test_path;
 pub use self::executive::run_test_file as run_executive_test_file;
```
dvdplm: These were the only ones used outside the crate (by `evmbin` I think)

json/src/spec/state.rs
dvdplm: You mean like this?

```rust
pub enum HashOrMap {
  …
}

#[cfg(any(test, feature = "test-helpers"))]
impl Clone for HashOrMap {
  …
}
```

json/src/spec/state.rs
dvdplm: So I tried this; compiler yells at me. My attempt:

```
impl Iterator for State {
	type Item = <BTreeMap<Address, Account> as Iterator>::Item;
	fn next(&mut self) -> Option<Self::Item> {
		match self.0 {
			HashOrMap::Map(map) => map.iter().next(),
			HashOrMap::Hash(_) => None
		}
	}
}
```

The error:

```
   --> json/src/spec/state.rs:103:1
    |
103 | impl IntoIterator for State {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<I> std::iter::IntoIterator for I
              where I: std::iter::Iterator;
```

What am I doing wrong?

json/src/transaction.rs
```diff
@@ -36,11 +36,14 @@ pub struct Transaction {
...
+	#[serde(default)]
+	pub r: MaybeEmpty<Uint>,
 	/// S.
-	pub s: Uint,
+	#[serde(default)]
+	pub s: MaybeEmpty<Uint>,
 	/// V.
-	pub v: Uint,
+	#[serde(default)]
+	pub v: MaybeEmpty<Uint>,
```
ordian: This code is used not only in tests, right? Not sure if this is the right place for a fix.

json/src/spec/state.rs
niklasad1: I mean't just remove the conditional compliation and derive Clone

json/src/spec/state.rs
niklasad1: Bad suggestion sorry, don't bother with this I think you need another type to workaround this such as `StateIter` or alternatively `impl IntoIter for &State`

ethcore/res/ethereum/tests-issues/currents.json
```diff
@@ -1,4 +1,88 @@
...
+			"failing": "stStaticCall",
+			"subtests": [
+				"static_callBasic_d0g0v0_ConstantinopleFix",
+				"static_callBasic_d1g0v0_ConstantinopleFix",
+				"static_callBasic_d2g0v0_ConstantinopleFix",
+				"static_callBasic_d3g0v0_ConstantinopleFix"
+			]
+		},
+		{
+			"reference": "Issue https://github.com/paritytech/parity-ethereum/issues/11077",
```
niklasad1: :+1:

ethcore/src/json_tests/chain.rs
niklasad1: ok, have you notice any change in execution time? Seems fine

ethcore/src/json_tests/chain.rs
```diff
@@ -108,24 +118,30 @@ pub fn json_chain_test<H: FnMut(&str, HookType)>(json_data: &[u8], start_stop_ho
...
 	use super::json_chain_test;
 	use json_tests::HookType;
 
-	fn do_json_test<H: FnMut(&str, HookType)>(json_data: &[u8], h: &mut H) -> Vec<String> {
-		json_chain_test(json_data, h)
+	fn do_json_test<H: FnMut(&str, HookType)>(path: &Path, json_data: &[u8], h: &mut H) -> Vec<String> {
+		json_chain_test(path, json_data, h)
 	}
-
+	//todo[dvdplm] do these tests match all folders in `res/` or are there tests we're missing?
```
niklasad1: Yes, see [test! macro](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/json_tests/test_common.rs#L107-#L112)

ethcore/src/json_tests/chain.rs
```diff
@@ -172,14 +188,23 @@ mod block_tests {
 	declare_test!{BlockchainTests_GeneralStateTest_stRandom2, "BlockchainTests/GeneralStateTests/stRandom2/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stRecursiveCreate, "BlockchainTests/GeneralStateTests/stRecursiveCreate/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stRefundTest, "BlockchainTests/GeneralStateTests/stRefundTest/"}
-	declare_test!{BlockchainTests_GeneralStateTest_stReturnDataTest, "BlockchainTests/GeneralStateTests/stReturnDataTest/"}
+	// todo[dvdplm]: 1 test fails. Single test. Have not investigated. Ignored in currents.json.
```
niklasad1: Do we have an issue number for this?

niklasad1: Same for the remaining `TODOS`, insert issue number if we got it

ethcore/src/json_tests/state.rs
```diff
@@ -164,6 +163,11 @@ mod state_tests {
 	declare_test!{GeneralStateTest_stRecursiveCreate, "GeneralStateTests/stRecursiveCreate/"}
 	declare_test!{GeneralStateTest_stRefundTest, "GeneralStateTests/stRefundTest/"}
 	declare_test!{GeneralStateTest_stReturnDataTest, "GeneralStateTests/stReturnDataTest/"}
+	// todo[dvdplm]:
```
niklasad1: Do we got an issue number for this?

ethcore/src/json_tests/test_common.rs
```diff
@@ -30,43 +30,60 @@ pub enum HookType {
...
-	p: &Path, skip: &[&'static str],
-	runner: fn(json_data: &[u8], start_stop_hook: &mut H) -> Vec<String>,
+	path: &Path,
+	skip: &[&'static str],
+	runner: fn(path: &Path, json_data: &[u8], start_stop_hook: &mut H) -> Vec<String>,
 	start_stop_hook: &mut H
 ) {
+	if !skip.is_empty() {
+		// todo[dvdplm] it's really annoying to have to use flushln here. Should be `info!(target: "json-tests", …)`
+		flushln!("[run_test_path] Skipping tests in {}: {:?}", path.display(), skip);
```
niklasad1: why is `flushln!` needed? Isn't `info! output unbuffered` (i.e, wait for the entire output before continuing)?

ethcore/src/json_tests/chain.rs
dvdplm: > if all folders are `declared`

Right, that's what the todo is about.

dvdplm: I actually know all folders are not covered. I have **no** idea if they should be though, e.g. `LegacyTests`.

ethcore/src/json_tests/chain.rs
dvdplm: Probably best if I remove the todos now that we have real issues filed and linked to this PR?

ethcore/src/json_tests/test_common.rs
dvdplm: The json test logging needs a major overhaul and I did not want to spend too much time on that at this time. `println!` was swallowed, as was `info!`. I think it has to do with the setup where a string is `flush!` at the start of a test; at the end it is completed with an `ok` or `failed`. It really should be all `info!`s, `warn!` etc. I tried to port them all over but it did not work right away so I reverted and kept the weirdness without digging deeper.

ethcore/src/json_tests/chain.rs
```diff
@@ -89,16 +90,25 @@ pub fn json_chain_test<H: FnMut(&str, HookType)>(json_data: &[u8], start_stop_ho
...
 					&spec,
 					db,
 					Arc::new(Miner::new_for_tests(&spec, None)),
 					IoChannel::disconnected(),
-				).unwrap();
+				).expect("Failed to instantiate a new Client");
+
 				for b in blockchain.blocks_rlp() {
+					let bytes_len = b.len();
 					if let Ok(block) = Unverified::from_rlp(b) {
```
ordian: does it make sense to add a warning if it returns an `Err`?

ordian: Does it make sense to add a `warn!` in case it return an `Err`?

ethcore/src/json_tests/chain.rs
ordian: resolved
```suggestion
```

ethcore/src/json_tests/chain.rs
```diff
@@ -172,14 +188,23 @@ mod block_tests {
...
+	declare_test!{ BlockchainTests_GeneralStateTest_stReturnDataTest, "BlockchainTests/GeneralStateTests/stReturnDataTest/"}
+	// todo[dvdplm]:
+	//      "RevertPrecompiledTouch_storage" contains 4 tests, only two fails
+	//      "RevertPrecompiledTouchExactOOG" contains a ton of tests, only two fails
+	//      "RevertPrecompiledTouch" has 4 tests, 2 failures
+	//  Ignored in currents.json, have not investigated
 	declare_test!{BlockchainTests_GeneralStateTest_stRevertTest, "BlockchainTests/GeneralStateTests/stRevertTest/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stShift, "BlockchainTests/GeneralStateTests/stShift/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stSolidityTest, "BlockchainTests/GeneralStateTests/stSolidityTest/"}
+	//  todo[dvdplm]: Ignored in currents.json, single test, failure caused by Parity calculating the state root for the genesis block differently than expected by the test.
```
ordian:
```suggestion
```

ethcore/src/json_tests/chain.rs
```diff
@@ -172,14 +188,23 @@ mod block_tests {
...
 	declare_test!{BlockchainTests_GeneralStateTest_stSpecialTest, "BlockchainTests/GeneralStateTests/stSpecialTest/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stStackTests, "BlockchainTests/GeneralStateTests/stStackTests/"}
+	// todo[dvdplm]: Ignored in currents.json, 4 failing tests out of 4. Have not investigated.
```
ordian:
```suggestion
```

ethcore/src/json_tests/chain.rs
```diff
@@ -172,14 +188,23 @@ mod block_tests {
...
 	declare_test!{BlockchainTests_GeneralStateTest_stStaticCall, "BlockchainTests/GeneralStateTests/stStaticCall/"}
 	declare_test!{BlockchainTests_GeneralStateTest_stSystemOperationsTest, "BlockchainTests/GeneralStateTests/stSystemOperationsTest/"}
+	//	todo[dvdplm]: Ignored in currents.json, 128 failures; have not investigated
```
ordian:
```suggestion
```

ethcore/src/json_tests/chain.rs
niklasad1: Ok, let's keep it out of this PR but we could probably discuss it in a separate issue. I don't know TBH.

ethcore/src/json_tests/chain.rs
niklasad1: yeah, that would the best

ethcore/src/json_tests/test_common.rs
niklasad1: Ok, let's keep it out of this PR and create a separate issue? Nice, that you have investigated it though..

ethcore/src/json_tests/chain.rs
dvdplm: Nothing that noticeable I guess. On my machine tests spin up and tear down 12 verifier threads for each test so if nothing else this makes the logs less noisy.

dvdplm:
> Let's update to beta when it's ready [ethereum/tests#640](https://github.com/ethereum/tests/pull/640) and merge this PR.

Hmm. Not sure I want to wait for that tbh. Who knows what else awaits when we update. 

FWIW I updated and ran the tests from https://github.com/ethereum/tests/commit/86b8a272a6d8334fa88bcd65cf8626449c652a02 and we pass except for the Istanbul version of the same handful of tests that we skip in this PR.

dvdplm:
>     * currently one test-skipping mechanism is hidden behind a feature (`ci-skip-tests`). Is this really what we want?

I think we should remove the feature. The purpose of the feature is to let CI run green while still being annoying to devs. 
With the updated tests however we're in a situation where we either change our code like @debris suggestion in #10923 or we keep skipping those `revert` tests indefinitely. In that scenario, `ci-skip-tests` doesn't make much sense.

ethcore/spec/src/spec.rs
```diff
@@ -134,53 +131,56 @@ fn run_constructors<T: Backend>(
...
-			origin: from.clone(),
-			gas: U256::max_value(),
-			gas_price: Default::default(),
-			value: ActionValue::Transfer(Default::default()),
-			code: Some(Arc::new(constructor.clone())),
-			data: None,
-			call_type: CallType::None,
-			params_type: ParamsType::Embedded,
+	if constructors.is_empty() {
+		state.populate_from(genesis_state.clone());
```
ordian: I'm not sure if it should be called only in case of empty constructors

PR #11056, opened 2019-09-16
SIMD Implementation for EIP-152

seunlanlege:
This PR refactors the `EIP-152` crate in `util` to make use of simd instruction sets where available.

code was graciously lifted and refactored from https://github.com/oconnor663/blake2_simd/blob/master/blake2b/src/avx2.rs

 - [x] Benchmarks
 - [ ] SSE implementation (?)
 - [x] Better code documentation so anyone can follow what's going on, almost done.


util/EIP-152/src/avx.rs
```diff
@@ -0,0 +1,405 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
```
ordian: IANAL, but shouldn't we provide some attribution to blake2_simd here?

util/EIP-152/src/avx.rs
```diff
@@ -0,0 +1,405 @@
...
+                t1 = _mm256_unpackhi_epi64(m4, m1);
+                b0 = _mm256_blend_epi32(t0, t1, 0xF0);
+                g1(&mut a, &mut b, &mut c, &mut d, &mut b0);
+                t0 = _mm256_blend_epi32(m5, m0, 0x33);
+                t1 = _mm256_unpacklo_epi64(m7, m6);
+                b0 = _mm256_blend_epi32(t0, t1, 0xF0);
+                g2(&mut a, &mut b, &mut c, &mut d, &mut b0);
+                undiagonalize(&mut a, &mut b, &mut c, &mut d);
+            },
+            _ => unreachable!("numbers 0-10 covered; qed")
```
ordian: when we have benchmarks, would be nice to see if removing the unreachable statement by replacing `9 =>` with `_=>` makes a difference

util/EIP-152/src/lib.rs
```diff
@@ -38,62 +41,25 @@ const IV: [u64; 8] = [
...
-		g(&mut v, 1, 6, 11, 12, m[s[10]], m[s[11]]);
-		g(&mut v, 2, 7, 8, 13, m[s[12]], m[s[13]]);
-		g(&mut v, 3, 4, 9, 14, m[s[14]], m[s[15]]);
-	}
-
-	for i in 0..8 {
-		h[i] ^= v[i] ^ v[i + 8];
+/// blake2b compression function
+fn compress(h: &mut [u64; 8], m: [u64; 16], t: [u64; 2], f: bool, rounds: usize) {
+	if is_x86_feature_detected!("avx2") {
```
ordian: There is [ifunc trick](https://github.com/BurntSushi/rust-memchr/blob/1ec5ecce03c220c762dd9a8b08f7a3d95522b765/src/x86/mod.rs#L9-L60) to do cpu detection only on the first call.

util/EIP-152/src/lib.rs
```diff
@@ -119,8 +85,21 @@ mod tests {
...
+		let mut h_in = [
+			0x6a09e667f2bdc948_u64, 0xbb67ae8584caa73b_u64,
+			0x3c6ef372fe94f82b_u64, 0xa54ff53a5f1d36f1_u64,
+			0x510e527fade682d1_u64, 0x9b05688c2b3e6c1f_u64,
+			0x1f83d9abfb41bd6b_u64, 0x5be0cd19137e2179_u64,
+		];
 
+		// avx
+		unsafe {
+			avx::compress(&mut h_in, m, c, f, rounds);
```
ordian: can we run this test only if it's supported by the CI machine?

dvdplm:
> Having only avx2 simd impl is fine for now as it's supported since Haswell (Q3 2013) Intel CPUs, but long-term we should look into [simdeez](https://github.com/jackmott/simdeez), which derives SSE and portable impl automatically.

I have not looked deeply into simdeez (it does seem amazing though). Another alternative to consider is what rust-crypto does [here](https://github.com/RustCrypto/hashes/tree/master/blake2/src/simd) – I find their approach to be quite readable (which is high praise in SIMD-land…).

util/EIP-152/src/avx.rs
```diff
@@ -0,0 +1,405 @@
...
+    let m7 = _mm256_broadcastsi128_si256(loadu_128(msg_chunks.7));
+
+    let iv0 = a;
+    let iv1 = b;
+
+    let mut t0;
+    let mut t1;
+    let mut b0;
+
+    for i in 0..rounds {
```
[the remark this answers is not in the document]
ordian: `unroll!` only works with constants, `rounds` is a dynamic parameter

seunlanlege:
so after adding benchmarks, criterion shows that the avx version is 2 times slower than the portable version 😅

util/EIP-152/benches/bench.rs
```diff
@@ -0,0 +1,179 @@
...
+            511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b6162630000000000000000000000000000000\
+            000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\
+            000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\
+            0000000000000000000000000000000000000000000000300000000000000000000000000000001",
+                        "b63a380cb2897d521994a85234ee2c181b5f844d2c624c002677e9703449d2fba551b3a8333bcdf5f2f7e0\
+            8993d53923de3d64fcc68c034e717b9293fed7a421",
+                    ).to_parts()
+                ];
+
+                for (mut state, message, count, f, rounds, expected) in vectors {
```
ordian: there is too much noise in what we're measuring
let's make it a benchmark with constant input from `test_blake2_f` except for number of rounds
(https://bheisler.github.io/criterion.rs/book/user_guide/benchmarking_with_inputs.html#benchmarking-with-a-range-of-values)
not need for `assert_eq!`, we're not trying to find bugs here, adding `black_box` if necessary to avoid the compiler to optimize the call away

seunlanlege:
how to handle conditional compilation on android? :thinking:

niklasad1:
> how to handle conditional compilation on android? :thinking: 

`#[cfg(target_os = "android")]`

util/EIP-152/src/avx.rs
```diff
@@ -0,0 +1,418 @@
...
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! AVX2 implementation of the blake2b compression function
+#![allow(dead_code)]
```
ordian:
```suggestion
#![cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_feature = "avx2"))]
```
maybe this will help with android, but I'm not sure

util/EIP-152/src/avx.rs
ordian: also, this file should be named avx2.rs, not avx.rs

util/EIP-152/src/lib.rs
```diff
@@ -38,58 +43,43 @@ const IV: [u64; 8] = [
...
+// Blake2bF fn pointer type
+type Blake2bFn = fn(&mut [u64; 8], [u64; 16], [u64; 2], bool, usize);
+
+// holds the pointer to the compress function
+static FN: AtomicPtr<()> = AtomicPtr::new(detect as FnRaw);
+
+// called the the first time, will check if avx is supported
+// and store the appropriate fn pointer in `FN`
+fn detect(state: &mut [u64; 8], message: [u64; 16], count: [u64; 2], f: bool, rounds: usize) {
+	let fun = if is_x86_feature_detected!("avx2") {
```
ordian: please tabify this file (and others)

util/EIP-152/src/lib.rs
```diff
@@ -38,58 +43,43 @@ const IV: [u64; 8] = [
...
-		g(&mut v, 1, 6, 11, 12, m[s[10]], m[s[11]]);
-		g(&mut v, 2, 7, 8, 13, m[s[12]], m[s[13]]);
-		g(&mut v, 3, 4, 9, 14, m[s[14]], m[s[15]]);
+// raw fn pointer type, can be cast to any fn type
+type FnRaw = *mut ();
+// Blake2bF fn pointer type
+type Blake2bFn = fn(&mut [u64; 8], [u64; 16], [u64; 2], bool, usize);
+
+// holds the pointer to the compress function
+static FN: AtomicPtr<()> = AtomicPtr::new(detect as FnRaw);
```
ordian: Did the `ifunk` trick actually make a difference in benchmarks?

seunlanlege: It's hard to say. Sometimes it's better sometimes it's worse.

seunlanlege:
Looks like rpc tests are failing

niklasad1:
Restarted the CI, seems fine now those RPC tests sometimes fail...

util/EIP-152/src/lib.rs
ordian: If there is no clear benefit, let's remove it then to avoid maintenance burden.
Also please fix the funny indentation (see github diff).

seunlanlege:
@dvdplm @ordian @niklasad1 

hey guys, lets get this merged.

util/EIP-152/benches/bench.rs
```diff
@@ -0,0 +1,173 @@
+use criterion::{Criterion, criterion_group, criterion_main, black_box, Throughput, BenchmarkId};
```
ordian: this file is missing license header

util/EIP-152/src/lib.rs
```diff
@@ -139,8 +134,8 @@ mod tests {
...
 			),
 			(   // test vector 5
-				"0000000c48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000001",
-				"ba80a53f981c4d0d6a2797b69f12f6e94c212f14685ac4b74b12bb6fdbffa2d17d87c5392aab792dc252d5de4533cc9518d38aa8dbf1925ab92386edd4009923",
+			    "0000000c48c9bdf267e6096a3ba7ca8485ae67bb2bf894fe72f36e3cf1361d5f3af54fa5d182e6ad7f520e511f6c3e2b8c68059b6bbd41fbabd9831f79217e1319cde05b61626300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000001",
+			    "ba80a53f981c4d0d6a2797b69f12f6e94c212f14685ac4b74b12bb6fdbffa2d17d87c5392aab792dc252d5de4533cc9518d38aa8dbf1925ab92386edd4009923",
```
ordian: please revert spaces change here and above

util/EIP-152/src/lib.rs
```diff
@@ -38,58 +42,30 @@ const IV: [u64; 8] = [
...
+		} else {
+			return portable::compress(state, message, count, f, rounds)
+		};
 	}
 
-	for i in 0..8 {
-		h[i] ^= v[i] ^ v[i + 8];
-	}
+	#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+	return portable::compress(state, message, count, f, rounds);
```
niklasad1:
```suggestion
	portable::compress(state, message, count, f, rounds);
```

util/EIP-152/src/avx2.rs
```diff
@@ -0,0 +1,417 @@
...
+
+
+/// The Blake2b compression function F. See https://tools.ietf.org/html/rfc7693#section-3.2
+/// Takes as an argument the state vector `state`, message block vector `message`, offset counter, final
+/// block indicator flag `f`, and number of rounds `rounds`. The state vector provided as the first
+/// parameter is modified by the function.
+#[target_feature(enable = "avx2")]
+pub unsafe fn compress(state: &mut [u64; 8], message: [u64; 16], count: [u64; 2], f: bool, rounds: usize) {
+	// get a mutable reference to state[0..4], state[4..]
+	let (state_low, state_high) = mut_array_refs!(state, 4, 4);
```
niklasad1: what is the rationale of using `array_refs/mut_array_refs` instead of `slice::split_at/slice::split_at_mut`?

seunlanlege: `split_at_mut` returns `&[T]` `array_refs` returns `[T; len]`

seunlanlege: In order to load the state into an `_m256i` I need `[u64; 4]`

niklasad1: Ok, I see

util/EIP-152/src/avx2.rs
```diff
@@ -0,0 +1,417 @@
...
+	*a = _mm256_permute4x64_epi64(*a, _MM_SHUFFLE!(0, 3, 2, 1));
+	*d = _mm256_permute4x64_epi64(*d, _MM_SHUFFLE!(1, 0, 3, 2));
+	*c = _mm256_permute4x64_epi64(*c, _MM_SHUFFLE!(2, 1, 0, 3));
+}
+
+
+/// The Blake2b compression function F. See https://tools.ietf.org/html/rfc7693#section-3.2
+/// Takes as an argument the state vector `state`, message block vector `message`, offset counter, final
+/// block indicator flag `f`, and number of rounds `rounds`. The state vector provided as the first
+/// parameter is modified by the function.
```
dvdplm: It would be good to include an overview of how the AVX2 impl works here too. It looks very different from the pseudo-code in the spec so I think it's useful to have to give future readers a chance to follow the code.

util/EIP-152/benches/bench.rs
```diff
@@ -0,0 +1,190 @@
...
+					];
+					let count = [3, 0];
+					let f = true;
+
+					unsafe {
+						let fun = FN.load(Ordering::Relaxed);
+						mem::transmute::<FnRaw, Blake2bF>
+							(fun)
+							(
+								black_box(&mut state),
```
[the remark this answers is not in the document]
seunlanlege: `blackbox` reduces noise from comiler optimizartions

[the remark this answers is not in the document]
dvdplm: Not sure what that means, can you educate me? How would I measure the impact on noise (other than seeing a change in the benchmark numbers)?

util/EIP-152/LICENSE
```diff
@@ -0,0 +1,25 @@
+This program is copyright 2019 Parity Technologies Limited and its licensors.
+
+                         GNU GENERAL PUBLIC LICENSE
+                            Version 3, 29 June 2007
+
+Some portions of the program (“the Software”) are Copyright (c) 2018 Jack O'Connor
```
dvdplm: 👍

util/EIP-152/benches/bench.rs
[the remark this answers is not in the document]
seunlanlege: > A function that is opaque to the optimizer, used to prevent the compiler from optimizing away computations in a benchmark.

From what I can tell, `black_box` prevents compiler optimizations from interfering with the benchmark, removing it gives me this message: 

Screenshot from 2019-10-01 15-21-42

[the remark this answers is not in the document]
dvdplm: I get that message both with and without using `black_box`. I think that message is simply Criterion-speak for "noticed some changes compared to last run, but too small to report as improvment/regression". See https://docs.rs/criterion/0.2.4/criterion/struct.Criterion.html#method.noise_threshold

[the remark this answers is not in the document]
dvdplm: Maybe this is a better ref: https://github.com/bheisler/criterion.rs/blob/master/book/src/user_guide/command_line_output.md#change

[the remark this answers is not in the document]
dvdplm: /cc @ordian can you guide us here? :)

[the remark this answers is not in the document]
ordian: `black_box` is a hint to the compiler to prevent it from optimizing the code away. The message you're seeing about `Change within noise threshold` is likely to be caused by other factors, like other processes running on your cpu, cpu throttling after compiling some rust code, etc. Unless you're seeing this consistently only with `black_box` removed, which I doubt.

IMHO, it's fine to keep `black_box` even if the compiler doesn't optimize the code away w/o it as of now.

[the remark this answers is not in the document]
dvdplm: > IMHO, it's fine to keep `black_box` even if the compiler doesn't optimize the code away w/o it as of now.

Two questions on this:

1. usually I see `black_box` used around the *function* under test, not its params. I think both usages are fine, but I think Criterions `black_box` is not zero-overhead so perhaps it introduces less noise to apply it only once rather than one time for each param? i.e. use it like this `black_box(myfun(1, "abc"))` rather than `myfun(black_box(1), black_box("abc"))`?

2. removing the blackboxes I see a sizeable perf increase here, 10 - 25%. It seems a bit too little to be explained with a const optimization and a bit much for `black_box` overhead. I did not run with a silent machine though so not sure how to interpret this. :/ Not really a question I guess.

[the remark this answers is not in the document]
ordian: 1. I don't know what's the best practice here, but the getting started page of criterion https://bheisler.github.io/criterion.rs/book/getting_started.html uses black_box on params. 
2. Sometimes I see measurable perf, but then I realize that rls was running previously. So making sure it's a silent run and the results are reproducible is important.

PR #11059, opened 2019-09-16
[ethcore]: reduce re-exports

### LINKED ISSUE #10130: Remove re-exports from ethcore

niklasad1:
Last piece of refactoring to close #10130 after work done by @dvdplm et. al

After this PR, we have the following re-exports from other crates in `non-testbuilds`:
- evm::VMType (client)
- ethcore_miner::local_accounts::LocalAccounts (miner)
- ethcore_miner::pool::PendingOrdering (miner)

I think the miner re-exports make sense (closely tied to the module) and `VMType` (related to module but not closely)


ethcore/src/client/config.rs
```diff
@@ -16,15 +16,15 @@
...
 use snapshot::SnapshotConfiguration;
+use trace::Config as TraceConfig;
 use types::client_types::Mode;
+use verification::QueueConfig;
 
-pub use std::time::Duration;
-pub use blockchain::Config as BlockChainConfig;
-pub use trace::Config as TraceConfig;
 pub use evm::VMType;
+pub use verification::VerifierType;
```
tomusdrw: What about this one?

ethcore/src/client/config.rs
```diff
@@ -16,15 +16,15 @@
 
 use std::str::FromStr;
 
-use verification::{VerifierType, QueueConfig};
+use blockchain::Config as BlockChainConfig;
 use journaldb;
 use snapshot::SnapshotConfiguration;
+use trace::Config as TraceConfig;
 use types::client_types::Mode;
+use verification::QueueConfig;
```
dvdplm:
```suggestion
use verification::{VerifierType, QueueConfig};
```
(and remove line 27)

PR #11065, opened 2019-09-18
[chainspecs]: make `eip1108` optional

### LINKED ISSUE #11063: Validating Aura node crashes on startup with 2.5.8

niklasad1:
Closes #11063

This PR makes the `eip1108 fields` as `Option` such that serde deserialization succeeds without being specified.

Also a `json file` is added to test it.


ethcore/builtin/src/lib.rs
```diff
@@ -83,16 +83,15 @@ impl Pricer for Linear {
...
+	eip1108_transition_price: Option<usize>,
 }
 
 impl Pricer for AltBn128ConstOperations {
 	fn cost(&self, _input: &[u8], at: u64) -> U256 {
-		if at >= self.eip1108_transition_at {
-			self.eip1108_transition_price.into()
-		} else {
-			self.price.into()
+		match (self.eip1108_transition_at, self.eip1108_transition_price) {
```
niklasad1: this became a little more complex unfortunately

However, it will only enable the `eip` iff both fields are specified in the chain spec. Do we want a warning/trace if only one field is specified (probably a bug in the case)?

niklasad1: Also this matches are `not DRY` but it becomes messy to add a struct `Eip1108` then we would need another trait object in there.

niklasad1: Also this two matches are `not DRY` but it would require a trait object or enum such as `Eip1108` to replace `eip_transition_at` and `eip_price` (price types will be different). I prefer not to do because most of this will be fixed by https://github.com/paritytech/parity-ethereum/pull/11039

ordian:
@niklasad1 could you elaborate on how this relates to #11039? Would this be a spec breaking change and #11039 not, just refactoring the internal impl?

niklasad1:
> @niklasad1 could you elaborate on how this relates to #11039? Would this be a spec breaking change and #11039 not, just refactoring the internal impl?

#11008 was an unintentional breaking change this PR is just a hack to fix it. 

#11039 is re-write of this mess to handle different prices and price models in more `future proof` way for builtins. Currently, it is a breaking change but I can investigate if it is possible to make it backward compatible and `deprecate (with a warning)` the old format.

Thoughts?

ordian:
We probably want to backport one future-proof change, ideally backwards (with 2.5.7 format) compatible.
If it's not possible to be backwards compatible, then what is the point of merging this PR only to be replaced by #11039 later (just delaying the breakage?)?

So I would propose to put this PR on hold until we see what would the accepted pricing formats in #11039 will be.

niklasad1:
Yeah, agree.

Issue #9089, opened 2018-07-10
Using gasprice oracle for a custom network


phahulin:
> I'm running:
>
> - **Which Parity version?**: 1.10.6
> - **Which operating system?**: Linux
> - **How installed?**: from source
> - **Are you fully synchronized?**: -
> - **Which network are you connected to?**: -
> - **Did you try to restart the node?**: -

It looks like currently gasprice oracle can only be used for foundation chain

https://github.com/paritytech/parity/blob/da5de4a6ffd275961cff124798caaef7b9d57483/parity/configuration.rs#L655-L657

Maybe I'm missing something. Docs mention that
> --usd-per-eth=[SOURCE]
        USD value of a single ETH. SOURCE may be either an amount
        in USD, a **web service** or 'auto' to use each web service in
        turn and fallback on the last known good value. (default:
        auto)

but it doesn't seem to work even if the condition above is changed and parity recompiled (I tried `--usd-per-eth=http://localhost:8000`), because validation fails on this line:
https://github.com/paritytech/parity/blob/da5de4a6ffd275961cff124798caaef7b9d57483/parity/configuration.rs#L659
with error `Invalid transaciton price 's' given. Must be a decimal number.`

It would be nice to be able to use gasprice oracle for any custom network by providing url of oracle http service.

Also, when recalculating `wei_per_gas`
https://github.com/paritytech/parity/blob/da5de4a6ffd275961cff124798caaef7b9d57483/parity/configuration.rs#L644-L649
precision might be lost.

Baranowski:
I just briefly skimmed through the code, and this seems easy. If you don't mind, I would like to take a stab at this as my first project in parity and in Rust in general.

One thing I would appreciate help with is a detailed (ideally, specific command lines and expected outputs) instruction how to test my patch. Or, even better, suggestion how to add an automated test, if feasible. Thanks in advance!

dvdplm:
@Baranowski welcome, and we're very happy to accept contributions! We'll do out very best to be responsive ofc, feel free to post any questions here and maybe @phahulin would be willing to assist as well? :)

joshua-mir:
[...]

PR #11066, opened 2019-09-18
Support --usd-per-eth=http://... (#9089)

Baranowski:
Support gasprice oracles for custom networks

### DIFF
--- a/ethcore/src/miner/miner.rs
+++ b/ethcore/src/miner/miner.rs
@@ -1886,6 +1886,7 @@ mod tests {
 				},
 				fetch,
 				p,
+				"dummy".to_owned(),
 			)
 		)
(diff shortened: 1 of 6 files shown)


Baranowski:
Sorry, I just tried for the second time to go through the CLA, and it's more legalese than I'm comfortable signing. I should've read it before submitting the PR, my bad.

dvdplm:
> more legalese than I'm comfortable signing. I should've read it before submitting the PR, my bad.

That is unfortunate. :/

Do not hesitate to ask questions about the CLA (if you have any ofc). Its purpose is to ensure the code remains unencumbered for posterity and anything we can do to make it more readable and understandable would be very welcome.

PR #11061, opened 2019-09-16
[sync]: remove unused dependencies or make dev

niklasad1:
cleanup dependencies in the `sync` crate

PR #11067, opened 2019-09-18
[sync]: rust 2018

niklasad1:
Address https://github.com/paritytech/parity-ethereum/pull/11061#pullrequestreview-288839718

#### Will not be addressed in this PR:

I think there are additional cleanups to do in the crate such:
- replace `enum_from_primitive` with a proc macro for `SyncPacket`
- rename/unify types such as `TransactionStats`, `LightSync` and similar, because several things are named the same thing which is confusing at least to me
- split `sync` into different crates for light and full sync

### DIFF
--- a/ethcore/sync/Cargo.toml
+++ b/ethcore/sync/Cargo.toml
@@ -4,26 +4,27 @@ name = "ethcore-sync"
 version = "1.12.0"
 license = "GPL-3.0"
 authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
 
 [lib]
 
 [dependencies]
 client-traits = { path = "../client-traits" }
-common-types = { path = "../types" }
+types = { package = "common-types", path = "../types" }
 enum_primitive = "0.1.1"
 ethcore-io = { path = "../../util/io" }
-ethcore-light = { path = "../light" }
-ethcore-network = { path = "../../util/network" }
-ethcore-network-devp2p = { path = "../../util/network-devp2p" }
+light = { package = "ethcore-light", path = "../light" }
+network = { package = "ethcore-network", path = "../../util/network" }
+devp2p = { package = "ethcore-network-devp2p", path = "../../util/network-devp2p" }
 ethcore-private-tx = { path = "../private-tx" }
 ethereum-types = "0.6.0"
 ethkey = { path = "../../accounts/ethkey" }
 fastmap = { path = "../../util/fastmap" }
 futures = "0.1"
-keccak-hash = "0.2.0"
+hash = { package = "keccak-hash", version = "0.2.0"}
 log = "0.4"
 macros = { path = "../../util/macros" }
-parity-bytes = "0.1"
+bytes = { package = "parity-bytes", version = "0.1" }
 parity-runtime = { path = "../../util/runtime" }
 parity-util-mem = "0.2.0"
 parking_lot = "0.8"
--- a/ethcore/sync/Cargo.toml
+++ b/ethcore/sync/Cargo.toml
@@ -20,7 +20,7 @@ ethereum-types = "0.6.0"
 ethkey = { path = "../../accounts/ethkey" }
 fastmap = { path = "../../util/fastmap" }
 futures = "0.1"
-hash = { package = "keccak-hash", version = "0.2.0"}
+keccak-hash = "0.2.0"
 light = { package = "ethcore-light", path = "../light" }
 log = "0.4"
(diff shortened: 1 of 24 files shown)


ethcore/sync/Cargo.toml
```diff
@@ -4,26 +4,27 @@ name = "ethcore-sync"
...
 license = "GPL-3.0"
 authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
 
 [lib]
 
 [dependencies]
 client-traits = { path = "../client-traits" }
-common-types = { path = "../types" }
+types = { package = "common-types", path = "../types" }
```
dvdplm: nit: I for one prefer to use the full name for this one and update the imports. It's fine like this too, just saying. I do prefer what you did below with `light`/`network`/`devp2p` so yeah, I guess I'm inconsistent.
:)

ethcore/sync/Cargo.toml
```diff
@@ -4,26 +4,27 @@ name = "ethcore-sync"
...
+light = { package = "ethcore-light", path = "../light" }
+network = { package = "ethcore-network", path = "../../util/network" }
+devp2p = { package = "ethcore-network-devp2p", path = "../../util/network-devp2p" }
 ethcore-private-tx = { path = "../private-tx" }
 ethereum-types = "0.6.0"
 ethkey = { path = "../../accounts/ethkey" }
 fastmap = { path = "../../util/fastmap" }
 futures = "0.1"
-keccak-hash = "0.2.0"
+hash = { package = "keccak-hash", version = "0.2.0"}
```
dvdplm: Here like above I prefer the full name because `hash` is such a commonly used word.

ethcore/sync/src/lib.rs
```diff
@@ -21,47 +21,10 @@
...
-extern crate macros;
-#[macro_use]
-extern crate log;
-extern crate parity_util_mem;
-#[macro_use]
-extern crate parity_util_mem as malloc_size_of;
-#[macro_use]
-extern crate trace_time;
+// needed to make the procedural macro `MallocSizeOf` to work
+#[macro_use] extern crate parity_util_mem as malloc_size_of;
```
dvdplm: @cheme I've run into this as well in other crates. Are we using it wrong or is this a "known issue"?

ethcore/sync/src/chain/supplier.rs
```diff
@@ -426,7 +436,7 @@ mod test {
 			rlp.append(&if reverse {1u32} else {0u32});
 			rlp.out()
 		}
-		fn to_header_vec(rlp: ::chain::RlpResponseResult) -> Vec<SyncHeader> {
+		fn to_header_vec(rlp: crate::chain::RlpResponseResult) -> Vec<SyncHeader> {
```
dvdplm: Add a `use crate::chain::RlpResponseResult` maybe?

ethcore/sync/src/chain/propagator.rs
```diff
@@ -335,17 +336,26 @@ impl SyncPropagator {
...
+	use network::client_version::ClientVersion;
 	use parking_lot::RwLock;
 	use rlp::Rlp;
-	use std::collections::VecDeque;
-	use tests::{
-		helpers::TestIo,
-		snapshot::TestSnapshotService,
-	};
-
-	use super::{*, super::{*, tests::*}};
```
dvdplm: Lovely to see these go.

dvdplm:
>     * replace `enum_from_primitive` with a proc macro for `SyncPacket`

We had a PR open that did exactly this and was really close to being finished. Maybe now's the time to pick that up and push it over the finish line?

dvdplm:
>     * split `sync` into different crates for light and full sync

Can you elaborate on this? Does `light` use `sync` atm? Sounds like a great idea on paper.

ethcore/sync/src/lib.rs
niklasad1: I think the crate need `proc-macro = true` in the manifests to work?

ethcore/sync/Cargo.toml
niklasad1: sure, sounds ok to me

niklasad1:
> Can you elaborate on this? Does light use sync atm? Sounds like a great idea on paper.

We have two different types for `LightSync` and `FullSync a.k.a EthSync` if we could extract the traits and handle `light protocol` handler for `EthSync` it should be doable IIRC. I can have a look myself at this :P

PR #11069, opened 2019-09-19
[ethash] remove manual unrolling 

ordian:
Manual unrolling is no longer necessary, since LLVM is good at that as well as autovectorizing, esp. with `-C target-cpu=native` https://rust.godbolt.org/z/6P6y8Q. Criterion detected no change in performance (but would be good if someone else could confirm this on another machine).

* remove manuall `unroll!`ing
* specify `required-features` for benches
* add a keccak512 bench

### DIFF
--- a/ethash/Cargo.toml
+++ b/ethash/Cargo.toml
@@ -6,7 +6,6 @@ authors = ["Parity Technologies <admin@parity.io>"]
 
 [dependencies]
 common-types = { path = "../ethcore/types" }
-crunchy = "0.1.0"
 either = "1.0.0"
 ethereum-types = "0.6.0"
 keccak-hash = "0.2.0"
--- a/ethash/Cargo.toml
+++ b/ethash/Cargo.toml
@@ -28,7 +28,9 @@ bench = []
 [[bench]]
 name = "basic"
 harness = false
+required-features = ['bench']
 
 [[bench]]
 name = "progpow"
 harness = false
+required-features = ['bench']
--- a/ethash/src/cache.rs
+++ b/ethash/src/cache.rs
@@ -21,7 +21,7 @@ use memmap::MmapMut;
 use parking_lot::Mutex;
 use seed_compute::SeedHashCompute;
 
-use shared::{ETHASH_CACHE_ROUNDS, NODE_BYTES, NODE_DWORDS, Node, epoch, get_cache_size, to_hex};
+use shared::{ETHASH_CACHE_ROUNDS, NODE_BYTES, Node, epoch, get_cache_size, to_hex};
 
 use std::borrow::Cow;
 use std::fs;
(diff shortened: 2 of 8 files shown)


ethash/Cargo.toml
```diff
@@ -29,7 +28,9 @@ bench = []
 [[bench]]
 name = "basic"
 harness = false
+required-features = ['bench']
 
 [[bench]]
 name = "progpow"
 harness = false
+required-features = ['bench']
```
dvdplm: 👍

ethash/src/cache.rs
```diff
@@ -331,11 +326,8 @@ unsafe fn initialize_memory(memory: *mut Node, num_nodes: usize, ident: &H256) {
 				let mut data: Node = nodes.get_unchecked(data_idx).clone();
 				let rhs: &Node = nodes.get_unchecked(idx);
 
-				unroll! {
-					for w in 0..8 {
-						*data.as_dwords_mut().get_unchecked_mut(w) ^=
-							*rhs.as_dwords().get_unchecked(w);
-					}
+				for (a, b) in data.as_dwords_mut().iter_mut().zip(rhs.as_dwords()) {
```
niklasad1: `NodeDoubleWords type` is of length 8? Do this imply bounce checking with the use of `iterator`?

ethash/src/compute.rs
```diff
@@ -239,24 +239,16 @@ fn hash_compute(light: &Light, full_size: usize, header_hash: &H256, nonce: u64)
...
-					for w in 0..16 {
-						mix[n].as_words_mut()[w] =
-							fnv_hash(
-								mix[n].as_words()[w],
-								tmp_node.as_words()[w],
-							);
-					}
-				}
+			// NODE_WORDS
+			for (a, b) in mix[n].as_words_mut().iter_mut().zip(tmp_node.as_words()) {
```
niklasad1: NodeDoubleWords type is of length 16? Do this imply bounce checking with the use of iterator?

ethash/src/compute.rs
```diff
@@ -326,10 +315,8 @@ pub fn calculate_dag_item(node_index: u32, cache: &[Node]) -> Node {
 			num_parent_nodes as u32;
 		let parent = &cache[parent_index as usize];
 
-		unroll! {
-			for w in 0..16 {
-				ret.as_words_mut()[w] = fnv_hash(ret.as_words()[w], parent.as_words()[w]);
-			}
+		for (a, b) in ret.as_words_mut().iter_mut().zip(parent.as_words()) {
```
niklasad1: NodeWords type is of length 16? Do this imply bounce checking with the use of iterator?

ethash/src/cache.rs
ordian: https://github.com/paritytech/parity-ethereum/blob/2627288311e952ff3b6b9675dfeda24f75b8249e/ethash/src/shared.rs#L29-L31

Bounds checking is eliminated by LLVM here (see the godbolt link above)

Issue #11075, opened 2019-09-20
[json-tests] stSpecialTest blockchain test failures


dvdplm:
After https://github.com/paritytech/parity-ethereum/pull/11054  the following JSON tests fail in [`stSpecialTest`](https://github.com/ethereum/tests/tree/f386934ce509fa0a4dbd1c68421b3361b01b2efa/BlockchainTests/GeneralStateTests/stSpecialTest) and are added to the list of skipped tests in `ethcore/res/ethereum/tests-issues/currents.json`.

```
tx_e1c174e2_d0g0v0_ConstantinopleFix
```

Fix the tests and "un-skip" them.

**Investigation**

The issue here is about the genesis block not matching the test expectation. We end up with a genesis block with hash `0xacad0c9ee590092cffee3b38edf7778f4819198c1a21ded24da4373772ae397d` and the test block expects the parent hash to be `0x39bbf39462cacb35fc779266aedc833436eb930d63ab9498a593c10ff96d9750` and so the block import fails with an `UnknownParent` error. 
The genesis block is constructed using the state from the `pre` section of the test file. The `state_root` we calculate from this data, `0xf92a34b36c86ed9642040927cd2a7ab7e0b7eda763d8dc4fcaa58af34526e74f`, does not match what the test expects, `0xd328574142f4eaad1125fa855bc8b12cffb2f29e0d0e5ac2d003a5d60c71b61c`. 

This is likely what causes the genesis blocks `hash` to be wrong.

Likely the same bug as #11074.

Issue #11077, opened 2019-09-20
[json-tests] stTransactionTest blockchain test failures


dvdplm:
After https://github.com/paritytech/parity-ethereum/pull/11054  the following JSON tests fail in [`stTransactionTest`](https://github.com/ethereum/tests/tree/f386934ce509fa0a4dbd1c68421b3361b01b2efa/BlockchainTests/GeneralStateTests/stTransactionTest) and are added to the list of skipped tests in `ethcore/res/ethereum/tests-issues/currents.json`.

```
[...]

Issue #11076, opened 2019-09-20
[json-tests] stStaticCall blockchain test failures


dvdplm:
After https://github.com/paritytech/parity-ethereum/pull/11054  the following JSON tests fail in [`stStaticCall`](https://github.com/ethereum/tests/blob/f386934ce509fa0a4dbd1c68421b3361b01b2efa/BlockchainTests/GeneralStateTests/stStaticCall/static_callBasic.json#L388) and are added to the list of skipped tests in `ethcore/res/ethereum/tests-issues/currents.json`.

```
"static_callBasic_d0g0v0_ConstantinopleFix",
"static_callBasic_d1g0v0_ConstantinopleFix",
"static_callBasic_d2g0v0_ConstantinopleFix",
"static_callBasic_d3g0v0_ConstantinopleFix"
```

Fix the tests and "un-skip" them.

ordian:
Closed by #11083.

PR #11083, opened 2019-09-23
[json-tests] populate state from genesis pod state

### LINKED ISSUE #11704: How to config something in openethereum  v3.0.0  like parity config.toml ?
image
I'm running openethereum with command ./target/release/openethereum

ordian:
Fixes #11704.
In the failing test we had
```json
                "storage" : {
                    "0x00" : "0x00"
                }
```
And our test code for constructing `state_root` didn't handle [removing zero valued storage entries](https://github.com/paritytech/parity-ethereum/blob/2627288311e952ff3b6b9675dfeda24f75b8249e/ethcore/account-state/src/account.rs#L511-L516) from trie. 
[`populate_from`](https://github.com/paritytech/parity-ethereum/blob/00124b5a4bf8a38ad6c660b462a5e5addf13cab3/ethcore/account-state/src/state.rs#L782) is used to update the state cache so that `commit` logic works and another `state.commit()` is added to work with 0 constructors case.

### DIFF
--- a/ethcore/spec/src/spec.rs
+++ b/ethcore/spec/src/spec.rs
@@ -131,6 +131,7 @@ fn run_constructors<T: Backend>(
 	let start_nonce = engine.account_start_nonce(0);
 
 	let mut state = State::from_existing(db, root, start_nonce, factories.clone())?;
+	state.populate_from(genesis_state.clone());
 
 	// Execute contract constructors.
 	let env_info = EnvInfo {
@@ -177,6 +178,7 @@ fn run_constructors<T: Backend>(
 
 		let _ = state.commit()?;
 	}
+	let _ = state.commit()?;
 
 	Ok(state.drop())
 }
--- a/ethcore/res/ethereum/tests-issues/currents.json
+++ b/ethcore/res/ethereum/tests-issues/currents.json
@@ -22,13 +22,6 @@
 			]
 		},
 		{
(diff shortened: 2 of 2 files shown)


ethcore/spec/src/spec.rs
```diff
@@ -131,6 +131,7 @@ fn run_constructors<T: Backend>(
 	let start_nonce = engine.account_start_nonce(0);
 
 	let mut state = State::from_existing(db, root, start_nonce, factories.clone())?;
+	state.populate_from(genesis_state.clone());
 
 	// Execute contract constructors.
 	let env_info = EnvInfo {
```
dvdplm: So this is not used when `self.constructors` is empty, nor is `from` below. Maybe wrap the whole block in an `if !constructors.empty()` and move the `state.commit()` outside?

dvdplm:
Test failure looks spurious. Retrying.

ethcore/spec/src/spec.rs
ordian: I don't have a strong opinion about this change, but it's out of scope of this PR, feel free to commit it directly to your branch after this PR is merged.

dvdplm:
Can confirm that this fixes #11075 as well.

dvdplm:
…and both #11076 and #11077 are fixed too.

Issue #11082, opened 2019-09-23
Race Condition and incorrectly dropped consensus messages


dforsten:
In ethcore/src/client/client.rs the implementation of the `queue()` function of IoChannelQueue there is no guarantee that `self.currently_queued` is incremented before it is being decremented.

This potentially results in an underflow of `self.currently_queued` and incorrectly dropped consensus messages if new consensus messages are being processed on another thread at the same time:
```
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG sync  0 -> Dispatching packet: 21
2019-09-19 09:17:54 UTC IO Worker #0 TRACE sync  Received consensus packet from 0
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG poa  Ignoring the message, error queueing: The queue is full (18446744073709551615)
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG sync  0 -> Dispatching packet: 21
2019-09-19 09:17:54 UTC IO Worker #0 TRACE sync  Received consensus packet from 0
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG poa  Ignoring the message, error queueing: The queue is full (18446744073709551615)
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG sync  0 -> Dispatching packet: 21
2019-09-19 09:17:54 UTC IO Worker #0 TRACE sync  Received consensus packet from 0
2019-09-19 09:17:54 UTC IO Worker #0 DEBUG poa  Ignoring the message, error queueing: The queue is full (18446744073709551615)
2019-09-19 09:17:54 UTC IO Worker #2 DEBUG sync  0 -> Dispatching packet: 21
2019-09-19 09:17:54 UTC IO Worker #2 TRACE sync  Received consensus packet from 0
```

We have encountered that issue quite frequently when deploying Honey Badger validators, leading to the Honey Badger Consensus Engine to become stuck in configurations with low node counts.

Removing the queue size check promptly fixed the issue. One possible fix for the underflow is to increment `self.currently_queued` *before* calling `channel.send()`, and decrementing it after `channel.send()` returns an error.

dvdplm:
Thank you for the report. Are you interested in submitting a PR to fix this? :)

dforsten:
I will see to get a PR submitted later today.

dforsten:
Fixed with PR #11086, thank you!

PR #11086, opened 2019-09-23
[client]: Fix for incorrectly dropped consensus messages (#11082)

dforsten:
Fixes a race condition causing the currently_queued counter to underflow and consensus messages getting dropped incorrectly as a consequence.

The initial solution suggested in #11082 was dropped in favor of converting the atomic counter to a signed integer for robustness. There is no definite guarantee that if the send() message returns an error the message will not get executed at a later point, which would cause another potential underflow.

The fix requires a slight limitation of the maximum size of the queue, a 63bit positive integer range should however be well large enough for its purposes.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -16,10 +16,11 @@
 
 use std::cmp;
 use std::collections::{BTreeMap, HashSet, VecDeque};
+use std::convert::TryFrom;
 use std::io::{BufRead, BufReader};
 use std::str::from_utf8;
 use std::sync::{Arc, Weak};
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering as AtomicOrdering};
+use std::sync::atomic::{AtomicBool, AtomicI64, Ordering as AtomicOrdering};
 use std::time::{Duration, Instant};
 
 use ansi_term::Colour;
@@ -2739,12 +2740,17 @@ fn transaction_receipt(
 
 /// Queue some items to be processed by IO client.
 struct IoChannelQueue {
-	currently_queued: Arc<AtomicUsize>,
-	limit: usize,
+	/// Using a *signed* integer for counting currently queued messages since the
+	/// order in which the counter is incremented and decremented is not defined.
+	/// Using an unsigned integer can (and will) result in integer underflow,
+	/// incorrectly rejecting messages and returning a FullQueue error.
+	currently_queued: Arc<AtomicI64>,
+	limit: i64,
 }
 
 impl IoChannelQueue {
 	pub fn new(limit: usize) -> Self {
+		let limit = i64::try_from(limit).unwrap_or(i64::max_value());
 		IoChannelQueue {
 			currently_queued: Default::default(),
 			limit,
@@ -2756,8 +2762,11 @@ impl IoChannelQueue {
 	{
 		let queue_size = self.currently_queued.load(AtomicOrdering::Relaxed);
 		if queue_size >= self.limit {
-			return Err(EthcoreError::FullQueue(self.limit))
+			let err_limit = usize::try_from(self.limit).unwrap_or(usize::max_value());
+			return Err(EthcoreError::FullQueue(err_limit))
 		};
+
+		let count = i64::try_from(count).unwrap_or(i64::max_value());
 
 		let currently_queued = self.currently_queued.clone();
 		let _ok = channel.send(ClientIoMessage::execute(move |client| {


ethcore/src/client/client.rs
```diff
@@ -2756,9 +2758,12 @@ impl IoChannelQueue {
...
 		if queue_size >= self.limit {
-			return Err(EthcoreError::FullQueue(self.limit))
+			let err_limit = usize::try_from(self.limit).unwrap_or(usize::max_value());
+			return Err(EthcoreError::FullQueue(err_limit))
 		};
 
+		let count = i64::try_from(count).unwrap_or(i64::max_value());
+
 		let currently_queued = self.currently_queued.clone();
 		let _ok = channel.send(ClientIoMessage::execute(move |client| {
```
dvdplm: Probably a silly question but why not check for an error here and only increment if the `send()` was `Ok`?

```
let send_result =  channel.send(ClientIoMessage::execute(move |client| {
    …
}));
match send_result {
    Ok(()) => self.currently_queued.fetch_add(count, AtomicOrdering::SeqCst),
    Err(e) => return Err(EthcoreError::Io(e)),
}
Ok(())
```

ethcore/src/client/client.rs
```diff
@@ -2739,12 +2740,13 @@ fn transaction_receipt(
...
 struct IoChannelQueue {
-	currently_queued: Arc<AtomicUsize>,
-	limit: usize,
+	currently_queued: Arc<AtomicI64>,
+	limit: i64,
 }
 
 impl IoChannelQueue {
 	pub fn new(limit: usize) -> Self {
+		let limit = i64::try_from(limit).unwrap_or(i64::max_value());
```
dvdplm: Did you consider changing the input arg to `isize` rather than casting it? Seems like there are 3 call sites in total (in `Client::new()`) so perhaps changing the type is not too bad?

ethcore/src/client/client.rs
```diff
@@ -2739,12 +2740,13 @@ fn transaction_receipt(
...
 /// Queue some items to be processed by IO client.
 struct IoChannelQueue {
-	currently_queued: Arc<AtomicUsize>,
-	limit: usize,
+	currently_queued: Arc<AtomicI64>,
+	limit: i64,
 }
 
 impl IoChannelQueue {
 	pub fn new(limit: usize) -> Self {
```
dvdplm: Not your code, but maybe this can be a private method?

ethcore/src/client/client.rs
dforsten: That is how the code looked like a couple of months back, it was changed to avoid that match statement (not by me).
But it essentially does the same as the current implementation, just in a slightly different style, has no effect on the behavior.

ethcore/src/client/client.rs
dforsten: Changing the input arg type does spread into a couple of corners in the code, especially into client settings.
It is an option of course - on the other hand it should only be configurable as a positive number, making the setting itself a signed integer might be confusing to the user. 
Using a signed integer in IoChannelQueue is a local implementation strategy to ensure the queue size counter checks work correctly in a multi-threaded environment.

ordian:
> There is no definite guarantee that if the send() message returns an error the message will not get executed at a later point, which would cause another potential underflow.

Could you elaborate how's that possible? I thought if `send(msg)` returns an error, it means the `msg` won't be executed.

Using a signed integer is a clever trick, but seems more like a hack than a proper solution to me.

ordian:
> The initial solution suggested in #11082 was dropped in favor of converting the atomic counter to a signed integer for robustness.

Have you been able to reproduce the issue with the initial solution?

dforsten:
> > There is no definite guarantee that if the send() message returns an error the message will not get executed at a later point, which would cause another potential underflow.
> 
> Could you elaborate how's that possible? I thought if `send(msg)` returns an error, it means the `msg` won't be executed.

Reading the implementation of send(msg) I am not at all certain that if it returns an error the message will not be executed. If there is an absolute guarantee that this is the case then we can implement the solution I originally suggested.

dvdplm:
> Reading the implementation of send(msg) I am not at all certain that if it returns an error the message will not be executed. If there is an absolute guarantee that this is the case then we can implement the solution I originally suggested.

Looking at the [send()](https://github.com/paritytech/parity-ethereum/blob/49f219451bf517bd360b67fb2872b22eaf54aa08/util/io/src/service_mio.rs#L380) impl it seems like one – `send_sync()` – of the two code paths never returns an error but even in that case, e.g. a failed call to `Weak::upgrade()`, would *not* send the message.
Do you know which of the two code paths you end up in when you've encountered the problem?

dforsten:
The code section I am most unsure if it will return an error even if the message will eventually be sent is this one in channel.rs of mio:
```
    pub fn try_send(&self, t: T) -> Result<(), TrySendError<T>> {
        self.tx.try_send(t)
            .map_err(From::from)
            .and_then(|_| {
                self.ctl.inc()?;
                Ok(())
            })
    }
```

`try_send()` is ok, but I am unclear if when `self.ctl.inc()?;` fails the message actually gets de-queued or sent at a later point.

dvdplm:
> I am unclear if when `self.ctl.inc()?;` fails the message actually gets de-queued or sent at a later point.

Help me check if I understand this right:

Afaict if [`try_send()`](https://github.com/tokio-rs/mio/blob/v0.6.x/src/channel.rs#L119) succeeds then the message is sent; if `self.ctl.inc()?;` fails after that we'd see an error bubble back up to the caller and we'd have a situation where the closure returns an `Err` but the message was actually sent (so we'd see a decrement). In that scenario we'd [bail](https://github.com/paritytech/parity-ethereum/blob/0051c26acfedd26591224155973ea175ca10f39a/ethcore/src/client/client.rs#L2766) before doing the increment.
Did I get that right? There is no de-queueing and/or re-trying happening though, or am I missing something?

When `self.ctl.inc()?;` fails it returns a `std::io::Error` so it's not something we can match on I think. If it returned a more specific error variant we could perhaps look for that specifically.

I think I understand what you mean by the original solution not being the right one: we'd increment before sending, send the message and still get an error back. Now we'd do two decrements (one from the message closure and the other in the `Err => ` branch in the caller) and could underflow.

Sounds like your second solution with the signed int is the least worst. :/

dvdplm:
> Sounds like your second solution with the signed int is the least worst. :/

With this PR we could end up with negative `currently_queued`: maybe `queue()` should check for that before sending and reset to `0` if that is the case?

ethcore/src/client/client.rs
```diff
@@ -2739,12 +2740,13 @@ fn transaction_receipt(
 
 /// Queue some items to be processed by IO client.
 struct IoChannelQueue {
-	currently_queued: Arc<AtomicUsize>,
-	limit: usize,
+	currently_queued: Arc<AtomicI64>,
```
ordian: Please add docs why we're using signed integers here, so it won't get accidentally reverted in the future.

dvdplm:
> Long-term we want to rewrite the whole mio-thing,

Indeed. And on that note it's a bit weird to send in an "item count" from the outside to the queue. The queue itself should keep count of how many messages it has pending/sent/failed etc. The only callsite where `count` is not `1` is when we enqueue transactions.

ethcore/src/client/client.rs
dforsten: I added documentation to the `currently_queued` field to clarify why signed integers are being used here.

dforsten:
> > Sounds like your second solution with the signed int is the least worst. :/
> 
> With this PR we could end up with negative `currently_queued`: maybe `queue()` should check for that before sending and reset to `0` if that is the case?

The only effect a negative number would have is that the queue size limit would effectively become larger. Before this fix it would straight away fail in that case, and not forward consensus messages at all.

dvdplm:
> The only effect a negative number would have is that the queue size limit would effectively become larger. Before this fix it would straight away fail in that case, and not forward consensus messages at all.

Yes, I am clear on that. It'd still be good to avoid negative numbers I think. If you disagree, do you mind explaining why?

ordian:
> Yes, I am clear on that. It'd still be good to avoid negative numbers I think. If you disagree, do you mind explaining why?

Negative numbers could appear only temporarily due to data race, and if we reset them to zero,  `currently_queued` will be incorrectly indefinitely growing and reach the limit at some point.

ordian:
> There is no definite guarantee that if the send() message returns an error the message will not get executed at a later point, which would cause another potential underflow.

But with this in mind, both the old and the signed implementation are still incorrect and could underflow.

dforsten:
> > There is no definite guarantee that if the send() message returns an error the message will not get executed at a later point, which would cause another potential underflow.
> 
> But with this in mind, both the old and the signed implementation are still incorrect and could underflow.

That is correct, if send() returns an Error without dropping the message the counter will decrement by one each time. With signed numbers the effect is an immediate underflow, with 64bit signed numbers it will take a very long time until it underflows.

If that case happens a lot or constantly the validator node should be considered faulty though since consensus messages will likely take a long/indefinite time be processed.

dforsten:
Is there still anything I should change on my end to get this PR merged?

PR #11092, opened 2019-09-26
Correct EIP-712 encoding

### LINKED ISSUE #11087: Incorrect EIP-712 Hashing for Arrays of Structs Inside Structs
- **Parity Ethereum version**: 2.5.7
- **Operating system**: Windows / MacOS / Linux
- **Fully synchronized**: n/a
- **Network**: n/a
- **Restarted**: n/a

Had an issue trying to use the EIP-712 code to encode some typed data. The hash of the typed  data differs from https://github.com/ethereumjs/ethereumjs-util which results in an incorrect recovery address. The JS one seems to match up with solidity so this appears to be a parity issue.

Here is an example of a failing payload: https://github.com/ptescher/parity-ethereum/blob/fix-eip-712-arrays/util/EIP-712/src/encode.rs#L600

seunlanlege:
Previously when building a type hash for a given struct, the `build_dependencies` function was unaware that structs could have fields that are arrays of custom types e.g

```
struct Person {
    address[] wallets
    string name
}

struct Group {
    string name
    Person[] members
}
```

so when it tried to encode the `Group` type it produced an incorrect encoding

`Group(string name, Person[] members)`

instead of:

`Group(string name, Person[] members)Person(string name, address[] wallets)`

which caused: #11087

this PR addresses this issue and closes #11087


util/EIP-712/src/encode.rs
```diff
@@ -56,11 +56,18 @@ fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes
 			deps.insert(item);
 
 			for field in fields {
+				// check if this field is an array type
+				let field_type = if field.type_.contains('[') {
+					let index = field.type_.find('[').expect("alrady checked in if statement; qed");
```
ordian: we could get rid of `expect` with `if let Some(index) = field.type_.find('[') {`

seunlanlege: ah good catch

seunlanlege:
> Unrelated q: this crate uses `lunarity 0.1`. Is there a point in upgrading to `0.2`?

lunarity 0.2 does bump up performance

util/EIP-712/src/encode.rs
```diff
@@ -56,11 +56,17 @@ fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes
 			deps.insert(item);
 
 			for field in fields {
+				// check if this field is an array type
+				let field_type = if let Some (index) = field.type_.find('[') {
```
ordian:
```suggestion
				let field_type = if let Some(index) = field.type_.find('[') {
```
funny space after `Some`

dvdplm:
> lunarity 0.2 does bump up performance

Can you add a ticket for it please? :)

seunlanlege:
already bumped it

seunlanlege:
rpc failing tests strike again.

niklasad1:
@seunlanlege done, but in the future, you can do it by yourself by just logging in to gitlab with your github account. Otherwise, ping the CI-team and they can help you :)

util/EIP-712/src/encode.rs
```diff
@@ -56,11 +56,17 @@ fn build_dependencies<'a>(message_type: &'a str, message_types: &'a MessageTypes
...
+				// check if this field is an array type
+				let field_type = if let Some(index) = field.type_.find('[') {
+					&field.type_[..index]
+				} else {
+					&field.type_
+				};
 				// seen this type before? or not a custom type skip
-				if deps.contains(&*field.type_) || !message_types.contains_key(&*field.type_) {
-					continue;
+				if deps.contains(field_type) || !message_types.contains_key(field_type) {
```
niklasad1: maybe remove needless `continue` and swap the if condition:

```suggestion
				if !deps.contains(field_type) || message_types.contains_key(field_type) { types.insert(field_type); }
```

util/EIP-712/src/parser.rs
```diff
@@ -68,81 +67,70 @@ impl fmt::Display for Type {
...
+			}
+			_ => return Err(ErrorKind::UnexpectedToken(lexer.slice().to_owned(), field_type.to_owned()))?
+		};
 
-		Ok(token.ok_or_else(|| ErrorKind::NonExistentType)?)
+		token = Some(type_);
+		lexer.advance();
 	}
+
+	Ok(token.ok_or_else(|| ErrorKind::NonExistentType)?)
```
niklasad1: could be ok_or instead because it is just a value.

PR #11096, opened 2019-09-26
[CI] check evmbin build

ordian:
To prevent #11095 from happening.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -96,6 +96,13 @@ cargo-check 2 3:
     - time cargo check --target $CARGO_TARGET --locked --manifest-path util/io/Cargo.toml --features "mio" --verbose --color=always
     - sccache -s
 
+cargo-check-evmbin:
+  stage:                           test
+  <<:                              *docker-cache-status
+  script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -96,6 +96,13 @@ cargo-check 2 3:
     - time cargo check --target $CARGO_TARGET --locked --manifest-path util/io/Cargo.toml --features "mio" --verbose --color=always
     - sccache -s
 
+cargo-check-evmbin:
+  stage:                           test
+  <<:                              *docker-cache-status
+  script:
+    - time cargo check -p evmbin --target $CARGO_TARGET --locked --verbose --color=always
```
dvdplm: I would have thought that having `evmbin` in the `members` of the workspace was sufficient and that `cargo check --all` would run on all members listed. It's not and I'm not sure if that's a bug or not.

ordian: It seems different, e.g. `cargo check --all` passed on beta, but `cargo check -p evmbin`. Probably have to do with how features are picked up.

PR #11102, opened 2019-09-30
Refactor return type of `BlockChainClient::code` #7098

### LINKED ISSUE #7098: Refactor return type of `BlockChainClient::code`
Currently it is defined like this:

```Rust
/// Get address code at given block's state.
fn code(&self, address: &Address, id: BlockId) -> Option<Option<Bytes>>;
```

Technically it's OK, but may be the source of confusion or even misinterpretation during code review. Taking into account potential importance of this method, I may suggest refactoring return type to something more verbose.

Because absence of code is not a error per se, we should not use `Result` here. We may introduce an `enum` that would either return a value or provide a reason why it wasn't there. As far as I understand, value may be missing due to nonexistent address, or due to actual absence of the code. 

So, it may look something like this:
```Rust
enum AddressCode {
    AddressNotFound,
    Empty,
    Some(Bytes),
}
```

botika:
Fix #7098

### DIFF
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -67,6 +67,15 @@ pub enum StateOrBlock {
 	Block(BlockId)
 }
 
+/// Result to be used during get address code at given block's state
+pub enum StateResult<T> {
+	/// State is missing
+	Missing,
+
+	/// State is some
+	Some(T),
+}
+
 impl From<Box<dyn StateInfo>> for StateOrBlock {
 	fn from(info: Box<dyn StateInfo>) -> StateOrBlock {
 		StateOrBlock::State(info)
@@ -231,12 +240,14 @@ pub trait BlockChainClient : Sync + Send + AccountData + BlockChain + CallContra
 	fn block_hash(&self, id: BlockId) -> Option<H256>;
 
 	/// Get address code at given block's state.
-	fn code(&self, address: &Address, state: StateOrBlock) -> Option<Option<Bytes>>;
+	fn code(&self, address: &Address, state: StateOrBlock) -> StateResult<Option<Bytes>>;
 
 	/// Get address code at the latest block's state.
 	fn latest_code(&self, address: &Address) -> Option<Bytes> {
-		self.code(address, BlockId::Latest.into())
-			.expect("code will return Some if given BlockId::Latest; qed")
+		match self.code(address, BlockId::Latest.into()) {
+			StateResult::Missing => panic!("code will return Some if given BlockId::Latest; qed"),
+			StateResult::Some(t) => t,
+		}
 	}
 
 	/// Get a reference to the `BlockProvider`.
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -68,6 +68,7 @@ pub enum StateOrBlock {
 }
 
 /// Result to be used during get address code at given block's state
+// todo[botika] move to `common-types`
 pub enum StateResult<T> {
 	/// State is missing
 	Missing,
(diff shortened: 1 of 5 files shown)


joshua-mir:
Thanks for the PR! Don't mind the CI failures atm, it looks like that is broken on our end 😅

botika:
It seems that the `parity/parity-ci-linux:latest` image does not exist in the registries.

ethcore/client-traits/src/lib.rs
```diff
@@ -231,12 +236,14 @@ pub trait BlockChainClient : Sync + Send + AccountData + BlockChain + CallContra
 	fn block_hash(&self, id: BlockId) -> Option<H256>;
 
 	/// Get address code at given block's state.
-	fn code(&self, address: &Address, state: StateOrBlock) -> Option<Option<Bytes>>;
+	fn code(&self, address: &Address, state: StateOrBlock) -> StateResult<Option<Bytes>>;
```
niklasad1: `StateResult` seems to re-inventing the `Result type` why not `Result<Bytes, &'static str>` instead? Or is there any particular reason why `StateResult` is needed?

dvdplm: It would be `Result, Option<Bytes>, something>` which doesn't really help understanding much. The name `StateResult` does make me think of `Result` but I'm not sure what a better name is. 
The point of it is, if I understand the issue correctly, to distinguish and communicate three possible states: "all good, here's the code", "all good, but there is no code there", and "not ok, there should have been code there but wasn't"

joshua-mir:
Hey @botika could you please rebase on master? It's necessary for our CI fix to apply 🙏

ethcore/client-traits/src/lib.rs
ordian: We could have a type alias like `type StateResult = Result<Option<Bytes>, StateError>; type StateError = ();`, but at this point it seems like bikeshedding.

niklasad1: yeah alright, thanks. Let's keep it as it is then.

botika:
In my computer it works correctly, 
```
running 2 tests
test v1::tests::mocked::signing::should_check_status_of_request ... ok
test v1::tests::mocked::signing::should_check_status_of_request_when_its_resolved ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 296 filtered out
```
change is a simple alias for the standard Option enumeration. So I think the fault is in the thread management of the test, although I can't prove it because I can't make it fail on my computer.

I don't know, what do you think?

niklasad1:
> change is a simple alias for the standard Option enumeration. So I think the fault is in the thread management of the test, although I can't prove it because I can't make it fail on my computer.
I don't know, what do you think?

The RPC tests sometimes fail in the CI, I have restarted the job. Seems unrelated to this

ethcore/client-traits/src/lib.rs
```diff
@@ -67,6 +67,11 @@ pub enum StateOrBlock {
 	Block(BlockId)
 }
 
+pub enum StateResult<T> {
```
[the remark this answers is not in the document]
botika: Documented!:sweat_smile: 
Ok when this PR is merged, I will make another one with these changes.

ethcore/client-traits/src/lib.rs
```diff
@@ -67,6 +67,15 @@ pub enum StateOrBlock {
 	Block(BlockId)
 }
 
+/// Result to be used during get address code at given block's state
+pub enum StateResult<T> {
```
dvdplm:
```suggestion
// todo[botika] move to `common-types`
pub enum StateResult<T> {
```

PR #11103, opened 2019-09-30
Make EIP712Domain Fields Optional

ptescher:
According to
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-
of-domainseparator “Protocol designers only need to include the fields
that make sense for their signing domain.”

### DIFF
--- a/util/EIP-712/src/eip712.rs
+++ b/util/EIP-712/src/eip712.rs
@@ -19,8 +19,7 @@ use serde_json::{Value};
 use std::collections::HashMap;
 use ethereum_types::{U256, H256, Address};
 use regex::Regex;
-use validator::Validate;
-use validator::ValidationErrors;
+use validator::{Validate, ValidationError, ValidationErrors};
 use lazy_static::lazy_static;
 
 pub(crate) type MessageTypes = HashMap<String, Vec<FieldType>>;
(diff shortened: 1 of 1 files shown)


ptescher:
[clabot:check]

joshua-mir:
Hey @ptescher could you rebase this on master? We've just fixed CI.

ptescher:
@dvdplm I noticed that the spec says that at least one field must be present, so I added a validator for that. As far as a malicious payload I believe the prefix https://github.com/paritytech/parity-ethereum/blob/master/util/EIP-712/src/encode.rs#L245 takes care of that.

I also added a test.

@joshua-mir I rebased.

util/EIP-712/src/eip712.rs
```diff
@@ -175,4 +186,43 @@ mod tests {
 		let data = from_str::<EIP712>(string).unwrap();
 		assert_eq!(data.validate().is_err(), true);
 	}
+
+	#[test]
+	fn test_invalid_domain() {
```
dvdplm:
```suggestion
	fn domain_needs_at_least_one_field() {
```

dvdplm: Dare I ask for a test with a valid domain too? :)

ptescher: Done!

util/EIP-712/src/eip712.rs
```diff
@@ -32,16 +31,27 @@ lazy_static! {
...
+	pub(crate) version: Option<String>,
+	#[serde(skip_serializing_if="Option::is_none")]
+	pub(crate) chain_id: Option<U256>,
+	#[serde(skip_serializing_if="Option::is_none")]
+	pub(crate) verifying_contract: Option<Address>,
 	#[serde(skip_serializing_if="Option::is_none")]
 	pub(crate) salt: Option<H256>,
 }
+
+fn validate_domain(domain: &EIP712Domain) -> Result<(), ValidationError> {
```
seunlanlege: 👏🏽

ptescher:
@seunlanlege ah whoops, fixed that too

PR #11105, opened 2019-10-01
use images from our registry

TriplEight:
Changes docker images addresses to our own registry.
Here: https://github.com/paritytech/scripts/blob/master/.gitlab-ci.yml#L61 I've started to move service images from dockerhub to our internal repository. 
The repository is `registry.parity.io/parity/infrastructure/scripts` and image names didn't change.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,7 +4,7 @@ stages:
   - publish
   - optional
 
-image:                             parity/parity-ci-linux:latest
+image:                             ${REGISTRY}/parity-ci-linux:latest
 variables:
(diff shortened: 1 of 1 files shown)


dvdplm:
@TriplEight can you elaborate on this for my education? :)

joshua-mir:
IIUC our CI images aren't hosted on the public registry, but we don't want to publicly share our private registry, so we put it's location in a environment variable. 

(for any users: the scripts to generate the images are in this repo)

s3krit:
@dvdplm we have our own Docker image repository. This change makes it so we use that rather than dockerhub.

PR #11108, opened 2019-10-01
[spec] fix rinkeby spec

ordian:
This caused genesis hash mismatch.


dvdplm:
Ah... while I was working on the json-tests I actually noticed that we had this odd 1wei balance for the builtins. Apparently this is the workaround we used years back when we had a consensus issue with geth and the ripemd bultin contract was removed due to it being considered empty. Sending 1wei to the builtins made them non-empty and so they were not removed...

PR #11110, opened 2019-10-01
Deduplicate registrar contract & calling logic

### LINKED ISSUE #7821: remove registrar logic from the Client

tobz1000:
Fixes #7821.

A summary of this refactor:

- `RegistryInfo` is removed, and `RegistrarClient` used in its place.
- `RegistrarClient` has a bound on `CallContract`, instead of implementing contract-calling directly.
- Removed `Registrar` wrapping type.
- Removed `BlockChainClient::registrar_address`, since `RegistrarClient` is a bound for that trait anyway.
- Async-ness has been removed from `RegistrarClient`, since there weren't any async implementations anyway.
  - If async contract-calling is ever implemented in future, refactoring back may be easier, since it would probably use `async`/`await` :)
  - `URLHint::Resolve` is now also synchronous as a result of this. There's possibly more that could be refactored in this way, but it's probably best left to another PR.

The previous implementation of `RegistryInfo::registry_address` conflated failure and a receiving null address by returning `None` in either case. Additionally, one caller of `RegistrarClient::get_address` was zero-checking the result (but not conflating it with call failure). So I decided to have `get_address` return `Result<Option<Address>, String>`, which allows the previous behaviour in all cases.

I've kept the behaviour the same for all previous usages of `RegistryInfo::registry_address` except for `ServiceTransactionChecker::check_address`, which will now return a different error message in the case of call failure.

These changes haven't been tested practically.


util/registrar/src/registrar.rs
```diff
@@ -14,59 +14,41 @@
...
+pub trait RegistrarClient: CallContract + Send + Sync {
+	/// Get address of the registrar itself
+	fn registrar_address(&self) -> Option<Address>;
 
-/// Registrar is dedicated interface to access the registrar contract
-/// which in turn generates an address when a client requests one
-pub struct Registrar {
-	client: Arc<dyn RegistrarClient<Call=Asynchronous>>,
-}
+	fn get_address<'a>(&self, key: &'a str, block: BlockId) -> Result<Option<Address>, String> {
```
dvdplm: Missing docs?

dvdplm:
FYI I think the CI failures are spurious.

util/registrar/src/registrar.rs
tomusdrw: Also is the lifetime really needed? It should be elided by the compiler.

parity/run.rs
```diff
@@ -687,29 +682,13 @@ fn execute_impl<Cr, Rr>(cmd: RunCmd, logger: Arc<RotatingLogger>, on_client_rq:
...
-	};
-
 	// the updater service
 	let updater_fetch = fetch.clone();
 	let updater = Updater::new(
 		&Arc::downgrade(&(service.client() as Arc<dyn BlockChainClient>)),
 		&Arc::downgrade(&sync_provider),
 		update_policy,
-		hash_fetch::Client::with_fetch(contract_client.clone(), updater_fetch, runtime.executor())
+		hash_fetch::Client::with_fetch(client.clone(), updater_fetch, runtime.executor())
```
dvdplm: I wonder if this should be a `Weak<dyn RegistrarClient>` so we don't hold on to a hard ref to the `Client`?

tomusdrw: Good catch, it should most likely be a `Weak` reference given that other stuff is passed as `Weak` too.

tobz1000: It looks like the `BlockChainClient` and `SyncProvider` references handle lack of backing client without emitting an error (using a default value, or just not performing the operation).

It would be easy enough to let the `hash_fetch::Client` fail if the backing client isn't there, but this would then be indistinguishable from some other error occurring.

Should we distinguish between the client disappearing, and other errors with the fetcher? (And presumably take no action if the client isn't present.)

tobz1000: The latest commit `164c61d` uses a weak reference to the client and fails when it's not present (so doesn't distinguish).

util/registrar/src/registrar.rs
tobz1000: Added a doc line, and removed the line on the trait specifying that it's for the latest block.

parity/run.rs
dvdplm: I think you're saying that if [this](https://github.com/tobz1000/parity-ethereum/blob/2f0ef2e2dbc4458856c4f57b1ae7b6cf95aaf372/updater/hash-fetch/src/urlhint.rs#L164) fails the the upgrade would fail? I think what you did is fine but I admit I'm not familiar with the code.
/cc @tomusdrw

tobz1000: Yes - compare that to [this](https://github.com/tobz1000/parity-ethereum/blob/2f0ef2e2dbc4458856c4f57b1ae7b6cf95aaf372/updater/src/updater.rs#L689) (performs no action if Weak upgrade fails) and [this](https://github.com/tobz1000/parity-ethereum/blob/2f0ef2e2dbc4458856c4f57b1ae7b6cf95aaf372/updater/src/updater.rs#L511) (ultimately performs no action). Neither of these will emit any error if the client is missing.

dvdplm: Thank you for explaining. It’s obviously not ideal to swallow errors like this but the changes here do not make things worse so let’s merge.

dvdplm:
@tobz1000 I think we must have missed something. When I run parity with the Goerli network after this was merged I see this in the logs:

```
2019-10-04 11:34:50  Syncing #1407080 0xffe1…4c93   166.90 blk/s  135.4 tx/s   15.8 Mgas/s      0+ 2078 Qed  #1409168    1/25 peers   0 bytes chain 54 MiB db 4 MiB queue 663 KiB sync  RPC:  0 conn,    0 req/s,    0 µs
2019-10-04 11:34:52  Error occurred while refreshing service transaction cache: Registrar address not defined.
2019-10-04 11:34:52  Imported #1410873 0xaf9c…9644 (0 txs, 0.00 Mgas, 1 ms, 0.59 KiB) + another 2 block(s) containing 2 tx(s)
2019-10-04 11:34:59  Error occurred while refreshing service transaction cache: Registrar address not defined.
2019-10-04 11:34:59  Imported #1410874 0x547c…6181 (0 txs, 0.00 Mgas, 0 ms, 0.59 KiB)
2019-10-04 11:34:59  Error occurred while refreshing service transaction cache: Registrar address not defined.
```

The error comes from [here](https://github.com/paritytech/parity-ethereum/blob/b6415c61963bb1d415ff0e173a78cdb8d2c65fa0/ethcore/src/miner/miner.rs#L1456) but I have not dug in further.

Do you know what might cause this?

tobz1000:
> Do you know what might cause this?

It looks like I missed a behavioural change that I introduced. Previously, `ServiceTransactionChecker::refresh_cache` would return `Ok(false)` for any failure in getting the contract address. After the refactor, it returns `Ok(false)` for a null address, and `Err(...)` for any other error.

Based on its usage (where you point out the error site in `miner.js`), `refresh_cache` should also return `Ok(false)` in the case that the registrar address is unavailable.

To fix this, we can a) revert the behaviour entirely by swallowing any error from `client.get_address`, or b) check if `client.registrar_address().is_none()` first. I think b) will give the desired behaviour, but still emit errors if calling the registrar contract fails.

dvdplm:
> b) will give the desired behaviour, with the added benefit of emitting errors if calling the registrar contract fails.

Yes, let's do this. And thank you for the explanation, it makes sense.

Do you have the bandwidth atm to PR this? Currently the logs are a bit spammy... :)

tobz1000:
I've opened #11126 to fix this. I wasn't able to reproduce the issue, but this should do it.

Sorry about that!

PR #11121, opened 2019-10-03
Move StateResult to `common-types`

botika:
To move `StateOrBlock` you should also move the `StateInfo` trait from `acount-state` to `common-types` to avoid cyclic dependence. 

In which module of ` common-types` should I put `StateInfo`?


dvdplm:
Ah I see. Don’t move `StateOrBlock` then, I suspected we left it there for a reason.

dvdplm:
@botika there are tests failing in `rpc/`, probably just imports to sort out.

Issue #11119, opened 2019-10-02
panicked: attempt to calculate the remainder with a divisor of zero


soc1c:
- **Parity Ethereum version**: `v2.5.9-stable-06c7096054-20190926`
- **Operating system**: Arch Linux
- **Installation**: repositories `pacman -S parity-ethereum`
- **Fully synchronized**: no
- **Network**: https://github.com/eth-classic/mordor
- **Restarted**: no

**expected behaviour**: synchronize blocks one by one.

**actual behaviour**: multiple panics on block import.

```
~/mordor master one@acrux                                                                                                                                                                       
❯ parity --chain parity.json                                                                                                                                                                    
2019-10-02 21:56:52  Starting Parity-Ethereum/v2.5.9-stable-06c7096054-20190926/x86_64-linux-gnu/rustc1.37.0                                                                                    
2019-10-02 21:56:52  Keys path /home/one/.local/share/io.parity.ethereum/keys/mordor                                                                                                            
2019-10-02 21:56:52  DB path /home/one/.local/share/io.parity.ethereum/chains/mordor/db/72f707e6f596a43b                                                                                        
2019-10-02 21:56:52  State DB configuration: fast                                                                                                                                                                                
2019-10-02 21:56:52  Operating mode: active                                                                                                                                                                                      
2019-10-02 21:56:52  Configured for Mordor Classic Testnet using Ethash engine                                                                                                                                                   
2019-10-02 21:56:52  Listening for new connections on 127.0.0.1:8546.                                                                                                                                                            
2019-10-02 21:56:57  Public node URL: enode://a59e33ccd2b3e52d578f1fbd70c6f9babda2650f0760d6ff3b37742fdcdfdb3defba5d56d315b40c46b70198c7621e63ffa3f987389c7118634b0fefbbdfa7fd@10.18.62.7:30303                                  
2019-10-02 21:57:27     1/25 peers   832 bytes chain 128 bytes db 0 bytes queue 1008 bytes sync  RPC:  0 conn,    0 req/s,    0 µs                                                                                               
2019-10-02 21:57:57     2/25 peers   832 bytes chain 128 bytes db 0 bytes queue 2 KiB sync  RPC:  0 conn,    0 req/s,    0 µs                                                                                                    
2019-10-02 21:58:27     2/25 peers   832 bytes chain 128 bytes db 0 bytes queue 2 KiB sync  RPC:  0 conn,    0 req/s,    0 µs                                                                                                    
[...]

PR #11123, opened 2019-10-04
[ethash] chainspec validate `ecip1017EraRounds` non-zero

niklasad1:
See https://github.com/paritytech/parity-ethereum/issues/11119 for background information.

This PR does the following:

1. Modifies chain spec validation to require ecip1017EraRounds to be non-zero. For a more human-friendly error message. (Could maybe be bigger than 1 instead, however, it is technically a breaking change but it makes parity useless)
2. Document `fn ecip1017_eras_block_reward` and add explicit panic for `hidden` panics that may occur by `divide by zero`

We could maybe just return `zero era rounds` when  `ecip1017EraRounds < 2`

Thoughts?

### DIFF
--- a/ethcore/engines/ethash/src/lib.rs
+++ b/ethcore/engines/ethash/src/lib.rs
@@ -472,7 +472,7 @@ impl Ethash {
 /// This function will panic if `era_rounds` is less than `2`
 fn ecip1017_eras_block_reward(era_rounds: u64, mut reward: U256, block_number: u64) -> (u64, U256) {
 	// NOTE(niklasad1): all numbers is divisible by 1, it will cause the if below
-	// to succeed except the first block. Thus, `era_rounds - 1 == 0` and cause `divide by zero`
+	// to succeed except for the first block. Thus, `era_rounds - 1 == 0` and cause `divide by zero`
 	//
 	// TODO(niklasad): We could set `eras` to zero if this occurs, but I'm not sure of the implications
 	assert!(era_rounds > 1, "ecip1017EraRounds must be bigger than 1");
--- a/ethcore/engines/ethash/src/lib.rs
+++ b/ethcore/engines/ethash/src/lib.rs
@@ -464,7 +464,18 @@ impl Ethash {
 	}
 }
 
-fn ecip1017_eras_block_reward(era_rounds: u64, mut reward: U256, block_number:u64) -> (u64, U256) {
+
+/// Calculates the number of eras and reward
+///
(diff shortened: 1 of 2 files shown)


ethcore/engines/ethash/src/lib.rs
```diff
@@ -464,7 +464,18 @@ impl Ethash {
...
+/// Calculates the number of eras and reward
+///
+/// # Panics
+///
+/// This function will panic if `era_rounds` is less than `2`
+fn ecip1017_eras_block_reward(era_rounds: u64, mut reward: U256, block_number: u64) -> (u64, U256) {
+	// NOTE(niklasad1): all numbers is divisible by 1, it will cause the if below
+	// to succeed except the first block. Thus, `era_rounds - 1 == 0` and cause `divide by zero`
+	//
+	// TODO(niklasad): We could set `eras` to zero if this occurs, but I'm not sure of the implications
```
niklasad1: thoughts?

ethcore/engines/ethash/src/lib.rs
```diff
@@ -464,7 +464,18 @@ impl Ethash {
...
-fn ecip1017_eras_block_reward(era_rounds: u64, mut reward: U256, block_number:u64) -> (u64, U256) {
+
+/// Calculates the number of eras and reward
+///
+/// # Panics
+///
+/// This function will panic if `era_rounds` is less than `2`
+fn ecip1017_eras_block_reward(era_rounds: u64, mut reward: U256, block_number: u64) -> (u64, U256) {
+	// NOTE(niklasad1): all numbers is divisible by 1, it will cause the if below
+	// to succeed except the first block. Thus, `era_rounds - 1 == 0` and cause `divide by zero`
```
niklasad1:
```suggestion
	// to succeed except for the first block. Thus, `era_rounds - 1 == 0` and cause `divide by zero`
```

ethcore/engines/ethash/src/lib.rs
ordian: I think it's fine to assume blocks in one rounds is at least 2 (it's set to 5_000_000 for ETC if I'm not mistaken).

ordian: Let's remove the TODO and we're good to merge.

niklasad1: the note too?

ordian: the note is good for future readers

PR #11126, opened 2019-10-04
ServiceTransactionChecker::refresh_cache: allow registrar unavailable

tobz1000:
Return `Ok(false)` instead of `Err(...)`. Fixes regression from PR #11110.


dvdplm:
To check try running `cargo run  -- --chain=goerli`

Issue #10096, opened 2018-12-21
eth_call to pending state of fully synced full history Parity node returns error


vbaranov:
- **Parity Ethereum version**: 2.2.1-beta
- **Operating system**: MacOS
- **Installation**: one-line installer
- **Fully synchronized**: yes
- **Network**: POA core
- **Restarted**: yes

### Description of the Issue

`eth_call` to `pending` state with fully synced Parity full history node returns **This request is not supported because your node is running with state pruning. Run with --pruning=archive.**

```
MacBook-Pro-Viktor:~ viktor$ curl -X POST -H 'Content-Type: application/json'  --data '{"jsonrpc":"2.0","method":"eth_call","params":[{"to": "0x4316f704989157162de90997ce1a05cff3118680", "data": "0x8da5cb5b"}, "pending"],"id":676681188149}' http://localhost:8545
{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":676681188149}
```

<img width="723" alt="screen shot 2018-12-21 at 13 16 03" src="https://user-images.githubusercontent.com/4341812/50338687-2c33d000-0526-11e9-9651-572ef21ea1df.png">

#### Parameters of archive node:

```
2018-12-21 13:48:10  Starting Parity-Ethereum/v2.2.1-beta-5c56fc5-20181114/x86_64-macos/rustc1.30.1
2018-12-21 13:48:10  Keys path ./parity_data/keys/Core
2018-12-21 13:48:10  DB path ./parity_data/chains/Core/db/39bb23edadd3d2d8
2018-12-21 13:48:10  State DB configuration: archive +Fat +Trace
2018-12-21 13:48:10  Operating mode: active
2018-12-21 13:48:10  Warning: Warp Sync is disabled because Fat DB is turned on.
2018-12-21 13:48:10  Configured for Core using AuthorityRound engine
```

- no CLI parameters
- [spec.json](https://github.com/poanetwork/poa-chain-spec/blob/core/spec.json)
- [bootnodes.txt](https://github.com/poanetwork/poa-chain-spec/blob/core/bootnodes.txt)
- [node.toml](https://gist.github.com/vbaranov/2e352c02e670afd3655bff0393c16c21#file-node-toml)

```
[parity]
chain = "./spec.json"
base_path = "./parity_data"

[network]
reserved_peers="./bootnodes.txt"
port = 30303

[rpc]
[...]

Issue #9707, opened 2018-10-05
eth_call 'pending' state error


vasinkd:
- **Parity Ethereum version**: v2.0.5-stable
- **Operating system**: Ubuntu 16.04
- **Installation**: one-line installer
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: no

I run parity in a default mode (--pruning=fast) and want to find out my balance on erc20-contract. I run RPC-command:
```
{'jsonrpc': '2.0',
 'id': 330186953061876261,
 'method': 'eth_call',
 'params': [{'to': '0x0affa06e7fbe5bc9a764c979aa66e8256a631f02',
             'data': '0x70a08231000000000000000000000000948f20d662dc17a4ddc16abda082c2b49e2295ea'},
 'pending']}  
```
which (I believe) used to work with previous versions of Parity.
I receive the following response:
```
{'jsonrpc': '2.0',
 'error': {'code': -32000,
           'message': 'This request is not supported because your node is running with state pruning. Run with --pruning=archive.'},
 'id': 330186953061876261}
```
I don't get it: why I can't perform such a request with --pruning=fast. I need a pending state, not the historical data. BTW, everything is Ok with state "latest".

joshua-mir:
I believe it's something to do with this code:

https://github.com/paritytech/parity-ethereum/blob/7ba5652bea42f7a3bde9e8bdf917154c9ae11906/rpc/src/v1/impls/parity.rs#L452

@tomusdrw mentioned that elsewhere in the code we fallback to latest in pending isn't available; the error code here appears to be misplaced 😅

ethernomad:
This bug still exists in Parity 2.2.9

tomusdrw:
We should either fallback to latest consistently or the error message should be improved.
To produce pending block please run with `--force-sealing`.

ethernomad:
Thankyou @tomusdrw

epheph:
This bug still exists in v2.5.9

ordian:
Closed by #11127.

PR #11127, opened 2019-10-04
Change how RPCs eth_call and eth_estimateGas handle "Pending"

dvdplm:
Before this PR we would return a rather confusing error when calling `eth_call` and `eth_estimateGas` with `"Pending"`, e.g.:

```
{"jsonrpc":"2.0","error":{"code":-32000,"message":"This request is not supported because your node is running with state pruning. Run with --pruning=archive."},"id":"e237678f6648ed12ff05a74933d06d17"}
```

In reality what is going on is that users often use `"Pending"` when they really mean `"Latest"` (e.g. MyCrypto…) and when the block in question is not actually pending. This changes our behaviour for these two RPC calls to fall back to `"Latest"` when the query with `"Pending"` fails.
Note that we already behave this way for many other RPCs:

	- eth_call (after this PR)
	- eth_estimateGas (after this PR)
	- eth_getBalance
	- eth_getCode
	- eth_getStorageAt

Closes https://github.com/paritytech/parity-ethereum/issues/10096 https://github.com/paritytech/parity-ethereum/issues/9707

### DIFF
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -274,7 +274,7 @@ pub trait BlockChainClient:
 
 	/// Get a list of all storage keys in the block `id`, if fat DB is in operation, otherwise `None`.
 	/// If `after` is set the list starts with the following item.
-	fn list_storage(&self, id: BlockId, account: &Address, after: Option<&H256>, count: u64) -> Option<Vec<H256>>;
+	fn list_storage(&self, id: BlockId, account: &Address, after: Option<&H256>, count: Option<u64>) -> Option<Vec<H256>>;
 
 	/// Get transaction with given hash.
 	fn transaction(&self, id: TransactionId) -> Option<LocalizedTransaction>;
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -487,17 +487,17 @@ pub trait ChainNotify: Send + Sync {
 
 /// Provides a method for importing/exporting blocks
 pub trait ImportExportBlocks {
-    /// Export blocks to destination, with the given from, to and format argument.
-    /// destination could be a file or stdout.
-    /// If the format is hex, each block is written on a new line.
-    /// For binary exports, all block data is written to the same line.
+	/// Export blocks to destination, with the given from, to and format argument.
+	/// destination could be a file or stdout.
+	/// If the format is hex, each block is written on a new line.
+	/// For binary exports, all block data is written to the same line.
 	fn export_blocks<'a>(
-        &self,
-        destination: Box<dyn std::io::Write + 'a>,
-        from: BlockId,
-        to: BlockId,
-        format: Option<DataFormat>
-    ) -> Result<(), String>;
+		&self,
+		destination: Box<dyn std::io::Write + 'a>,
+		from: BlockId,
+		to: BlockId,
+		format: Option<DataFormat>
+	) -> Result<(), String>;
 
 	/// Import blocks from destination, with the given format argument
 	/// Source could be a file or stdout.
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -412,7 +412,7 @@ pub trait StateClient {
 	/// Type representing chain state
 	type State: StateInfo;
 
-	/// Get a copy of the best block's state.
+	/// Get a copy of the best block's state and header.
 	fn latest_state_and_header(&self) -> (Self::State, Header);
 
 	/// Attempt to get a copy of a specific block's final state.
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -413,7 +413,7 @@ pub trait StateClient {
 	type State: StateInfo;
 
 	/// Get a copy of the best block's state.
-	fn latest_state(&self) -> Self::State;
+	fn latest_state_and_header(&self) -> (Self::State, Header);
 
 	/// Attempt to get a copy of a specific block's final state.
 	///
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1805,7 +1805,7 @@ impl BlockChainClient for Client {
 		Some(accounts)
 	}
 
-	fn list_storage(&self, id: BlockId, account: &Address, after: Option<&H256>, count: u64) -> Option<Vec<H256>> {
+	fn list_storage(&self, id: BlockId, account: &Address, after: Option<&H256>, count: Option<u64>) -> Option<Vec<H256>> {
 		if !self.factories.trie.is_fat() {
 			trace!(target: "fatdb", "list_storage: Not a fat DB");
 			return None;
@@ -1846,9 +1846,16 @@ impl BlockChainClient for Client {
 			}
 		}
 
-		let keys = iter.filter_map(|item| {
-			item.ok().map(|(key, _)| H256::from_slice(&key))
-		}).take(count as usize).collect();
+		let keys = {
+			let f = iter.filter_map(|item| {
+				item.ok().map(|(key, _)| H256::from_slice(&key))
+			});
+			if let Some(count) = count {
+				f.take(count as usize).collect()
+			} else {
+				f.collect()
+			}
+		};
 
 		Some(keys)
 	}
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1033,15 +1033,16 @@ impl Client {
 	}
 
 	/// Get a copy of the best block's state.
-	pub fn latest_state(&self) -> State<StateDB> {
+	pub fn latest_state_and_header(&self) -> (State<StateDB>, Header) {
 		let header = self.best_block_header();
-		State::from_existing(
+		let state = State::from_existing(
 			self.state_db.read().boxed_clone_canon(&header.hash()),
 			*header.state_root(),
 			self.engine.account_start_nonce(header.number()),
 			self.factories.clone()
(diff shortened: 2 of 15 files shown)


rpc/src/v1/impls/eth.rs
```diff
@@ -451,11 +456,38 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
 				self.miner
 					.pending_state(info.best_block_number)
 					.map(|s| Box::new(s) as Box<dyn StateInfo>)
-					.unwrap_or(Box::new(self.client.latest_state()) as Box<dyn StateInfo>)
+					.unwrap_or_else(|| {
+						// todo[dvdplm]: this strikes me as odd. If I call `get_state` with
```
tomusdrw: Well, that's exactly why it's `StateOrBlock`. Afair we can't use `Block` as it's only for states that are in the db already. Similarly we can use `State` for past blocks, as it's not exposed from the client (cause it's not needed).

rpc/src/v1/impls/eth.rs
```diff
@@ -451,11 +456,38 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
...
+					})
 					.into()
 			}
 		}
 	}
+
+	/// Get the state and header of best pending block. On failure, fall back to the best imported
+	/// blocks state&header.
+	// todo[dvdplm]: better name?
+	// todo[dvdplm]: Scary thought: what if the state is not pending but the header is? We'd return the state from
```
tomusdrw: That's a good point. We use `best_block_number` here, which feels a bit incorrect as the block may be imported in the meantime. Would be way better to get the `hash` of latest block and consistently use `hash` to extract the information we need.

rpc/src/v1/impls/eth.rs
```diff
@@ -439,6 +440,10 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
 		Ok(Some(block))
 	}
 
+	/// Get state for the given block number. Returns either the State or a block from which state
+	/// can be retrieved.
+	/// Note: When passing `BlockNumber::Pending` we fall back to the state of the current best block
```
tomusdrw: I'm actually not sure if fallback behavior is desired. I'd rather return a more meaningful error which underlines that:
1. Pending state is only available if mining or using `--force-sealing`.
2. Pending state is discouraged cause it's not available for light clients.

If a dapp requests `pending` state and the client doesn't really have one it seems more like some errornous setup for me. In the past we did fall back to `latest` afair (at least for some methods) - maybe there is still a CLI flag that controls this behavior?

dvdplm: > it seems more like some errornous setup for me

Me too. However this is the way MyCrypto works (and many other users too, given the amount of bug reports we get on this)  and they have been unable to fix their code for quite a long while. The behaviour of `get_state()` in this regard is unchanged in this PR and does indeed fall back to `"latest"`. This PR adds this fallback behaviour to `eth_call()` and `eth_estimateGas` as well.

rpc/src/v1/impls/eth.rs
dvdplm: My `todo` is unclear, apologies. My point is: if I call `get_state(BlockNumber::Latest)` I get `StateOrBlock::Block` but if I call `get_state(BlockNumber::Pending)` and I'm not mining so I hit the fallback path then I'll get a `StateOrBlock::State` back. 
It felt a bit inconsistent, is all.

rpc/src/v1/impls/eth.rs
tomusdrw: Fair enough!

ethcore/src/client/client.rs
```diff
@@ -1052,7 +1053,10 @@ impl Client {
 	pub fn state_at(&self, id: BlockId) -> Option<State<StateDB>> {
 		// fast path for latest state.
 		match id.clone() {
```
niklasad1: could be replaced with `if let BlockId::Latest = id { ... }`

ethcore/src/client/client.rs
```diff
@@ -1088,7 +1092,8 @@ impl Client {
 
 	/// Get a copy of the best block's state.
 	pub fn state(&self) -> Box<dyn StateInfo> {
```
niklasad1:
```suggestion
	pub fn state(&self) -> impl StateInfo {
```

rpc/src/v1/impls/eth.rs
```diff
@@ -451,11 +457,45 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
...
 		}
 	}
+
+	/// Get the state and header of best pending block. On failure, fall back to the best imported
+	/// blocks state&header.
+	fn pending_state_and_header_with_fallback(&self) -> (T, Header) {
+		let best_block_number = self.client.chain_info().best_block_number;
+
+		let (maybe_state, maybe_header) = (
+				// todo:[dvdplm]: this takes the queue lock twice and is inelegant. It would be
```
niklasad1: Ok, I think that makes sense.

Or maybe just provide the `iterator API` instead of `map_existing_pending_block`

rpc/src/v1/impls/eth.rs
```diff
@@ -451,11 +457,45 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
...
+				//  better to move `map_existing_pending_block()` to the trait and impl the
+				//  `pending_*` methods in terms of it rather than pollute it with more special-cased
+				//  methods. That does not work currently because of the generic `T` returned here.
+				//  (Tbh I wonder if the whole `MinerService` trait actually carries its weight.)
+				self.miner.pending_state(best_block_number),
```
niklasad1: we bail early to avoid locking twice

rpc/src/v1/impls/eth.rs
```diff
@@ -451,11 +457,45 @@ impl<C, SN: ?Sized, S: ?Sized, M, EM, T: StateInfo + 'static> EthClient<C, SN, S
...
-					.unwrap_or(Box::new(self.client.latest_state()) as Box<dyn StateInfo>)
+					.unwrap_or_else(|| {
+						// todo[dvdplm]: this strikes me as odd. If I call `get_state` with
+						// `BlockNumber::Latest` I get a `StateOrBlock::Block` back, but if I call
+						// it with `BlockNumber::Pending` and the state is not there so the fallback
+						// is triggered, then I get a `StateOrBlock::State` back. Shouldn't it be
+						// consistent?
+						warn!("Asked for best pending state, but none found. Falling back to latest state");
+						let (state, _) = self.client.latest_state_and_header();
+						Box::new(state)  as Box<dyn StateInfo>
```
niklasad1:
```suggestion
						Box::new(state) as Box<dyn StateInfo>
```

ethcore/client-traits/src/lib.rs
```diff
@@ -413,7 +413,7 @@ pub trait StateClient {
 	type State: StateInfo;
 
 	/// Get a copy of the best block's state.
-	fn latest_state(&self) -> Self::State;
+	fn latest_state_and_header(&self) -> (Self::State, Header);
```
niklasad1: I'm not a big fan of methods with `several responsibilities`. Why not a separate method `fn latest_header()`?

Is this some kind of optimization that I don't understand? 

Also, seems that we call this method a couple of this without caring about the header. I think the header type is quite big and might carry some cost to return by value when it's not needed.

dvdplm: Looking up [latest state](https://github.com/paritytech/parity-ethereum/blob/4fd1ec643f1c64eef1770ac45deac6013b298bfd/ethcore/src/client/client.rs#L1036-L1045) first looks up latest header, so when we need both we:

- look up the header
- use the header to fetch the state
- toss away the header
- look up the header again

So while I agree with you that it's generally best to have methods do a single thing, I think returning both pieces of data is warranted here.

ethcore/client-traits/src/lib.rs
```diff
@@ -413,7 +413,7 @@ pub trait StateClient {
 	type State: StateInfo;
 
 	/// Get a copy of the best block's state.
```
tomusdrw: Docs could do with an update.

ethcore/client-traits/src/lib.rs
tomusdrw: I think it was made to ensure atomicity of the operation - i.e. that the we don't import the block in the meantime and the state does not match header.

rpc/src/v1/impls/eth.rs
tomusdrw: hmm, right. But do you feel it breaks something?

rpc/src/v1/impls/eth.rs
dvdplm: You mean

```rust
if let Some(s) = self.miner.pending_state(best_block_number) {
	if let Some(h) = self.miner.pending_block_header(best_block_number) {
		(Some(s), Some(h))
	} else {
		(None, None)
	}
} else {
	(None, None)
}
```

rpc/src/v1/impls/eth.rs
dvdplm: No, I don't think it's a bona fide bug, just a wart.

rpc/src/v1/impls/eth.rs
niklasad1: yeah, but the variant you implemented was nicer

PR #11128, opened 2019-10-04
Fix some random typos, formatting/whitespace

dvdplm:
Random tweaks.

### DIFF
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -487,17 +487,17 @@ pub trait ChainNotify: Send + Sync {
 
 /// Provides a method for importing/exporting blocks
 pub trait ImportExportBlocks {
-    /// Export blocks to destination, with the given from, to and format argument.
-    /// destination could be a file or stdout.
-    /// If the format is hex, each block is written on a new line.
-    /// For binary exports, all block data is written to the same line.
+	/// Export blocks to destination, with the given from, to and format argument.
+	/// destination could be a file or stdout.
+	/// If the format is hex, each block is written on a new line.
+	/// For binary exports, all block data is written to the same line.
 	fn export_blocks<'a>(
-        &self,
-        destination: Box<dyn std::io::Write + 'a>,
-        from: BlockId,
-        to: BlockId,
-        format: Option<DataFormat>
-    ) -> Result<(), String>;
+		&self,
+		destination: Box<dyn std::io::Write + 'a>,
+		from: BlockId,
+		to: BlockId,
+		format: Option<DataFormat>
+	) -> Result<(), String>;
 
 	/// Import blocks from destination, with the given format argument
 	/// Source could be a file or stdout.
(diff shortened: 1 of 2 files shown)


ethcore/client-traits/src/lib.rs
```diff
@@ -487,17 +487,17 @@ pub trait ChainNotify: Send + Sync {
...
+	/// For binary exports, all block data is written to the same line.
 	fn export_blocks<'a>(
-        &self,
-        destination: Box<dyn std::io::Write + 'a>,
-        from: BlockId,
-        to: BlockId,
-        format: Option<DataFormat>
-    ) -> Result<(), String>;
+		&self,
+		destination: Box<dyn std::io::Write + 'a>,
```
niklasad1: Out-of-scope: I think the lifetimes in this trait is not needed.

dvdplm: Happy to toss it in here while we're all here. :)

ordian: Looks like they're needed after all :)
```rust
error[E0597]: `out` does not live long enough
   --> ethcore/src/tests/client.rs:408:12
    |
408 |         Box::new(&mut out),
    |         ---------^^^^^^^^-
    |         |        |
    |         |        borrowed value does not live long enough
    |         cast requires that `out` is borrowed for `'static`
...
438 | }
    | - `out` dropped here while still borrowed

error[E0502]: cannot borrow `out` as immutable because it is also borrowed as mutable
   --> ethcore/src/tests/client.rs:414:26
    |
408 |         Box::new(&mut out),
    |         ------------------
    |         |        |
    |         |        mutable borrow occurs here
    |         cast requires that `out` is borrowed for `'static`
...
414 |     let written = from_utf8(&out)
    |                             ^^^^ immutable borrow occurs here

error[E0502]: cannot borrow `out` as immutable because it is also borrowed as mutable
   --> ethcore/src/tests/client.rs:431:34
    |
408 |         Box::new(&mut out),
    |         ------------------
    |         |        |
    |         |        mutable borrow occurs here
    |         cast requires that `out` is borrowed for `'static`
...
431 |     client.import_blocks(Box::new(&*out), Some(DataFormat::Hex)).unwrap();
    |                                     ^^^ immutable borrow occurs here

error[E0597]: `out` does not live long enough
   --> ethcore/src/tests/client.rs:447:12
    |
447 |         Box::new(&mut out),
    |         ---------^^^^^^^^-
    |         |        |
    |         |        borrowed value does not live long enough
    |         cast requires that `out` is borrowed for `'static`
...
469 | }
    | - `out` dropped here while still borrowed

error[E0502]: cannot borrow `out` as immutable because it is also borrowed as mutable
   --> ethcore/src/tests/client.rs:462:34
    |
447 |         Box::new(&mut out),
    |         ------------------
    |         |        |
    |         |        mutable borrow occurs here
    |         cast requires that `out` is borrowed for `'static`
...
462 |     client.import_blocks(Box::new(&*out), Some(DataFormat::Binary)).unwrap();
    |                                     ^^^ immutable borrow occurs here

error: aborting due to 5 previous errors
```

dvdplm: Damn compiler, always knows best huh? K, reverted.

niklasad1: lol, sorry for ruining your Friday night

Issue #9405, opened 2018-08-23
Alltools docker image


holiman:
On docker [hub](https://hub.docker.com/r/parity/parity/), there seems to be only `latest` and `stable` built with any regularity (aside from release-builds). Afaiu, they're built off [this](https://github.com/paritytech/parity-ethereum/blob/master/docker/hub/Dockerfile) dockerfile. 

I would be very greatful if you could add another docker image to docker hub, `alltools`, which also contains `parity-evm` (and any other binaries that the build process can generate). A similar thing exists for geth, with `ethereum/client-go:alltools-latest `

holiman:
Here's an example dockerfile that contains `parity-evm`:

```

# Taken from https://github.com/paritytech/parity-ethereum/blob/master/docker/alpine/Dockerfile
FROM alpine:edge AS builder

# show backtraces
ENV RUST_BACKTRACE 1

RUN apk add --no-cache \
  build-base \
  cargo \
  cmake \
  eudev-dev \
  linux-headers \
  perl \
  rust git

RUN git clone https://github.com/paritytech/parity-ethereum /parity
WORKDIR /parity
RUN cargo build --release --target x86_64-alpine-linux-musl --verbose -p evmbin
RUN strip target/x86_64-alpine-linux-musl/release/parity-evm


FROM alpine:edge

# show backtraces
ENV RUST_BACKTRACE 1

RUN apk add --no-cache \
  libstdc++ \
  eudev-libs \
  libgcc

RUN addgroup -g 1000 parity \
  && adduser -u 1000 -G parity -s /bin/sh -D parity

USER parity
COPY --chown=parity:parity --from=builder /parity/target/x86_64-alpine-linux-musl/release/parity-evm ./

#ENTRYPOINT ["./parity"]

```

5chdn:
Actually, it's worth to add all of our tools to the parity docker image instead of maintaining multiple docker images.

holiman:
Definitely! Anyone who wants a smaller image can just use that complete image as the basis of a multistage build, and pick out the cherries.

holiman:
Any progress here  ? Would be really nice to have...

sachincool:
@joshua-mir hey, Can I pick this up, I tried to do the same thing as said above in the example of `ethereum/client-go:alltools-latest` by simply Copying everything from the build `release/*` 

As mentioned above, We can make multiple Stages in the multi stage build where the source builder is same, making multiple Enviroment Versions super fast and easy eg:
[...]

PR #11135, opened 2019-10-06
feat: Add alltools version in dockerfile, fix: #9405

sachincool:
Running `docker build -t parity/parity:dev . --target=dev` for the alltools version 
Running `docker build -t parity/parity:latest . --target=release` for the small production level image

This can be eaisly used to push to dockerhub simultaneously dev and latest images from the same CI build. because of the already cached Steps of builder Stage

Before you submitting, please check that:

- [x] You added a brief description of the PR, e.g.:
  - What does it do?
  - What important points reviewers should know?
  - Is there something left for follow-up PRs?
- [ ] You labeled the PR with appropriate labels if you have permissions to do so.
- [x] You mentioned a related issue if this PR related to it, e.g. `Fixes #228` or `Related #1337`.
- [ ] You asked any particular reviewers to review. If you aren't sure, start with GH suggestions.
- [ ] Your PR adheres [the style guide](https://wiki.parity.io/Coding-guide)
  - In particular, mind the maximal line length.
  - There is no commented code checked in unless necessary.
  - Any panickers have a proof or removed.
- [ ] You updated any rustdocs which may have changed


sachincool:
[clabot:check]

joshua-mir:
Thanks for the PR! (cc @TriplEight @s3krit?)

sachincool:
I understood, but If adding `--add` in the carbo build increases the building time of the image, shall I create a seperate `Dockerfile.dev`? what would be best here?

joshua-mir:
Afaict on my local machine there was very little difference in the build time when adding `--all` because the additional tools are already dependencies of parity, for the most part, but that is only anecdotal.

dvdplm:
@TriplEight @s3krit can you please weigh in on this?

TriplEight:
First of all thank you very much for the contribution! That's a neat idea do make dockerfiles like that.

Have to say that totally agree with @joshua-mir here.

It doesn't make sense to copy the whole `target` dir there with incremental build artifacts, it won't make it `all-tools`, will just make it bigger.

I do not think this better to make two separate dockerfiles named `release.Dockerfile` and `dev.Dockerfile` or even make it 3 stage builder pattern, than try to understand where's the difference in these dupes.

Doubt that this `scripts/docker/alpine/Dockerfile` along with `centos` and everything besides the `hub` dir (which is used in CI) are actually being used by someone.

joshua-mir:
> Doubt that this scripts/docker/alpine/Dockerfile along with centos and everything besides the hub dir (which is used in CI) are actually being used by someone.
.@TriplEight I can confirm that quite a few users use our non-hub dockerfiles because I regularly recommend them to people 😅

TriplEight:
> @TriplEight I can confirm that quite a few users use our non-hub dockerfiles because I regularly recommend them to people sweat_smile

We should update them then.

PR #11138, opened 2019-10-07
[json]: add docs to `hardfork specification`

niklasad1:
Fix https://github.com/paritytech/parity-ethereum/pull/11039#discussion_r331434940

### DIFF
--- a/json/src/spec/spec.rs
+++ b/json/src/spec/spec.rs
@@ -24,17 +24,29 @@ use serde_json::Error;
 /// Fork spec definition
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize)]
 pub enum ForkSpec {
-	EIP150,
-	EIP158,
+	/// Frontier (#1, 2015-07-30)
 	Frontier,
+	/// Homestead (#1,150,000, 2016-03-14)
 	Homestead,
+	/// EIP 150 Tangerine Whistle: Gas cost changes for IO-heavy operation (#2,463,000, 2016-10-18)
+	EIP150,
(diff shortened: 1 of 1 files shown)


json/src/spec/spec.rs
```diff
@@ -24,17 +24,29 @@ use serde_json::Error;
...
 	Frontier,
+	/// Homestead (#1,150,000, 2016-03-14)
 	Homestead,
+	/// EIP 150 Tangerine Whistle: Gas cost changes for IO-heavy operation (#2,463,000, 2016-10-18)
+	EIP150,
+	/// EIP 158/EIP 161 Spurious Dragon: State trie clearing (#2,675,000, 2016-11-22)
+	EIP158,
+	/// Byzantium Metropolis phase 1 (#4,370,000, 2017-10-16)
 	Byzantium,
+	/// Constantinople Metropolis phase (#7,280,000, 2019-02-28)
```
ordian: phase 2?

json/src/spec/spec.rs
```diff
@@ -24,17 +24,29 @@ use serde_json::Error;
 /// Fork spec definition
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Deserialize)]
 pub enum ForkSpec {
-	EIP150,
-	EIP158,
```
ordian: `ForkSpec` derives `Ord`, I don't know if changing the order here breaks something or not?

json/src/spec/spec.rs
niklasad1: yes

json/src/spec/spec.rs
niklasad1: reverted

PR #11139, opened 2019-10-07
Update ethereum types to 0.8.0 version

grbIzl:
The version of the crate and linked crates (like keccak-hash) were updated across the repository

Updates: 
- ethbloom to `0.8`
- keccak-hash to `0.4`
- ethereum-types to `0.8`
- ethabi = "9.0.1"
- ethabi-contract = "9.0.0"
- ethabi-derive = "9.0.1"


grbIzl:
@niklasad1 commit e898ad3 was made in order to avoid clashes between rand 0.6 and 0.7. You may want to revert it with global switch to rand 0.7

rpc/src/v1/helpers/subscribers.rs
```diff
@@ -78,7 +79,8 @@ impl<T> Default for Subscribers<T> {
 
 impl<T> Subscribers<T> {
 	fn next_id(&mut self) -> Id {
-		let data = H64::random_using(&mut self.rand);
+		let mut data = H64::default();
+		self.rand.fill_bytes(&mut data.as_bytes_mut());
```
niklasad1: ahh all right, so we should be able to revert this once both are rand 0.7 then?

rpc/src/v1/tests/mocked/eth_pubsub.rs
```diff
@@ -62,28 +62,28 @@ fn should_subscribe_to_new_heads() {
 
 	// Subscribe
 	let request = r#"{"jsonrpc": "2.0", "method": "eth_subscribe", "params": ["newHeads"], "id": 1}"#;
-	let response = r#"{"jsonrpc":"2.0","result":"0x43ca64edf03768e1","id":1}"#;
+	let response = r#"{"jsonrpc":"2.0","result":"0x4333966aca52ad0b","id":1}"#;
```
niklasad1: why change this?

rpc/src/v1/helpers/subscribers.rs
dvdplm: I'm out of the loop, forgive the silly q: How does this change with `rand 0.7`?

rpc/src/v1/tests/mocked/eth_pubsub.rs
dvdplm: Did we change PRNG maybe? :/

rpc/src/v1/tests/mocked/eth_pubsub.rs
```diff
@@ -62,28 +62,28 @@ fn should_subscribe_to_new_heads() {
...
...40","gasUsed":"0x0","hash":"0xdf04a98bb0c6fa8441bd429822f65a46d0cb553f6bcef602b973e65c81497f8e","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x0000000000000000000000000000000000000000","number":"0x3","parentHash":"0x44e5ecf454ea99af9d8a8f2ca0daba96964c90de05db7a78f59b84ae9e749706","receiptsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","sealFields":[],"sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x1c9","stateRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","timestamp":"0x0","transactionsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"},"subscription":"0x4333966aca52ad0b"}}"#;
```
dvdplm: What changed here?

rpc/src/v1/helpers/subscribers.rs
niklasad1: This is related to the changes of `OsRng` where version `0.6` is not compatible with version `0.7`. I don't really know exactly what was caused it but I think the `trait OsRngImpl` was removed in `0.7` when `OsRng` started to depend on `getrandom` crate rather than implement it.

Thus, because we now use `rand 0.7` in `ethereum-types v0.8` and `rand 0.6` in `rpc` it `let data = H64::random_using(&mut self.rand);` doesn't work anymore.

rpc/src/v1/tests/mocked/eth_pubsub.rs
niklasad1: Yes, it looks like [fixed-hash](https://github.com/paritytech/parity-common/blob/master/fixed-hash/src/hash.rs#L510) is using `rand::distributions::Standard::sample` and `OsRng::fill_bytes` is using `getrandom` but I haven't understood the internals...

rpc/src/v1/helpers/subscribers.rs
dvdplm: ok, let's add a `todo` here so we don't forget.

grbIzl: Yes, @niklasad1 is right. I'll add todo here

rpc/src/v1/tests/mocked/eth_pubsub.rs
grbIzl: I changed the way, how next_id is generated. As a result the initial value calculated from the test seed was also changed in tests (where deterministic PRNG used).

PR #11140, opened 2019-10-08
Use TryFrom instead of From+panic for Builtin

### LINKED ISSUE #11134: [ethcore-builtin]: use `TryFrom` instead of `From`
Instead of panicking when faulty builtin names are entered in the chain spec, replace the  [From implementation](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/builtin/src/lib.rs#L218-#L268) with `TryFrom` and propagate the error accordingly

rokob:
Fixes #11134 

Currently there is a panic if a bad builtin name is entered in the chain spec. This PR replaces that panic by returning an Error instead of panicking in that situation and then using TryFrom instead of From to create the Builder type.

I'm not sure what compiler version you are using as I got errors and a ton of change from running rustfmt, so I finagled it to run over just my changes but there might still be some stylistic issues.

Also, the code is a bit convoluted but I wanted to put this up to make sure this is what you are looking for before cleaning it up a bit.

### DIFF
--- a/ethcore/builtin/src/lib.rs
+++ b/ethcore/builtin/src/lib.rs
@@ -18,12 +18,14 @@
 
 use std::{
 	cmp::{max, min},
-	io::{self, Read, Cursor},
+	convert::TryFrom,
+	io::{self, Cursor, Read},
 	mem::size_of,
 };
 
 use bn;
 use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
+use common_types::errors::EthcoreError as Error;
 use ethereum_types::{H256, U256};
 use ethjson;
 use ethkey::{Signature, recover as ec_recover};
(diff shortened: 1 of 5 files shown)


rokob:
[clabot:check]

niklasad1:
> I'm not sure what compiler version you are using as I got errors and a ton of change from running rustfmt, so I finagled it to run over just my changes but there might still be some stylistic issues.

We are not using rustfmt that is the reason why you get a errors I guess. Officially, we use the latest stable compiler

ethcore/builtin/src/lib.rs
```diff
@@ -1265,15 +1266,15 @@ mod tests {
...
+		let b = Builtin::try_from(ethjson::spec::Builtin {
 			name: "identity".to_owned(),
 			pricing: ethjson::spec::Pricing::Linear(ethjson::spec::Linear {
 				base: 10,
 				word: 20,
 			}),
 			activate_at: None,
 			eip1108_transition: None,
-		});
+		}).expect("bad spec");
```
niklasad1:
```suggestion
		}).expect("known builtin");
```

ethcore/builtin/src/lib.rs
```diff
@@ -1298,39 +1299,39 @@ mod tests {
 			}),
 			activate_at: Some(Uint(U256::from(10))),
 			eip1108_transition: Some(Uint(U256::from(20))),
-		});
+		}).expect("bad spec");
```
niklasad1:
```suggestion
		}).expect("known builtin");
```

ethcore/builtin/src/lib.rs
```diff
@@ -1298,39 +1299,39 @@ mod tests {
...
+		let b = Builtin::try_from(ethjson::spec::Builtin {
 			name: "alt_bn128_add".to_owned(),
 			pricing: ethjson::spec::Pricing::AltBn128ConstOperations(ethjson::spec::builtin::AltBn128ConstOperations {
 				price: 500,
 				eip1108_transition_price: 150,
 			}),
 			activate_at: Some(Uint(U256::from(10))),
 			eip1108_transition: Some(Uint(U256::from(20))),
-		});
+		}).expect("bad spec");
```
niklasad1:
```suggestion
		}).expect("known builtin");
```

ethcore/builtin/src/lib.rs
```diff
@@ -1298,39 +1299,39 @@ mod tests {
...
+		let b = Builtin::try_from(ethjson::spec::Builtin {
 			name: "alt_bn128_mul".to_owned(),
 			pricing: ethjson::spec::Pricing::AltBn128ConstOperations(ethjson::spec::builtin::AltBn128ConstOperations {
 				price: 40_000,
 				eip1108_transition_price: 6000,
 			}),
 			activate_at: Some(Uint(U256::from(10))),
 			eip1108_transition: Some(Uint(U256::from(20))),
-		});
+		}).expect("bad spec");
```
niklasad1:
```suggestion
		}).expect("known builtin");
```

ethcore/builtin/src/lib.rs
```diff
@@ -215,61 +217,58 @@ impl Builtin {
...
+				},
+			}),
 			ethjson::spec::Pricing::AltBn128ConstOperations(pricer) => {
 				Box::new(AltBn128ConstOperations {
-						price: pricer.price,
-						eip1108_transition_price: pricer.eip1108_transition_price,
-						eip1108_transition_at: b.eip1108_transition.map_or(u64::max_value(), Into::into)
+					price: pricer.price,
+					eip1108_transition_price: pricer.eip1108_transition_price,
+					eip1108_transition_at: b
```
niklasad1: formatting nit

ethcore/builtin/src/lib.rs
```diff
@@ -679,6 +679,7 @@ impl Bn128Pairing {
 
 #[cfg(test)]
 mod tests {
+    use std::convert::TryFrom;
```
niklasad1: indentation replace with tab please

ethcore/machine/src/test_helpers.rs
```diff
@@ -16,14 +16,16 @@
...
 use common_types::engines::params::CommonParams;
+use ethcore_builtin::Builtin;
 use ethjson;
 use crate::Machine;
 
 pub fn load_machine(reader: &[u8]) -> Machine {
 	let spec = ethjson::spec::Spec::load(reader).expect("chain spec is invalid");
 
-	let builtins = spec.accounts.builtins().into_iter().map(|p| (p.0.into(), From::from(p.1))).collect();
+	let builtins = spec.accounts.builtins().into_iter().map(|p| (p.0.into(), Builtin::try_from(p.1).expect("chain spec is invalid"))).collect();
```
niklasad1: It would be cool if we could display all errors in the `builtin chain` spec if there were more than one error.

Currently, the error output looks like:
```bash
Thread 'main' panicked at 'Chain spec ethereum/foundation is invalid.: Msg("invalid builtin name: foo")'
```

ethcore/builtin/src/lib.rs
```diff
@@ -215,61 +217,58 @@ impl Builtin {
 	}
 }
 
-impl From<ethjson::spec::Builtin> for Builtin {
-	fn from(b: ethjson::spec::Builtin) -> Self {
+impl TryFrom<ethjson::spec::Builtin> for Builtin {
+	type Error = Error;
```
dvdplm: nitpick: I think this would have been more readable: `type Error = EthcoreError`.

ethcore/builtin/src/lib.rs
```diff
@@ -215,61 +217,58 @@ impl Builtin {
...
-						pair: pricer.pair,
-					},
-					eip1108_transition_at: b.eip1108_transition.map_or(u64::max_value(), Into::into),
-					eip1108_transition_price: AltBn128PairingPrice {
-						base: pricer.eip1108_transition_base,
-						pair: pricer.eip1108_transition_pair,
-					},
-				})
-			}
+			ethjson::spec::Pricing::Blake2F { gas_per_round } => Box::new(gas_per_round),
```
dvdplm: One suggestion to cut down on the noise a little bit would be to add a local `use ethjson::spec::Pricing::*;` right before the `match`. Entirely optional though! :)

ethcore/builtin/src/lib.rs
rokob: Good point

ethcore/builtin/src/lib.rs
rokob: I actually ended up with a bunch of formatting changes here that were unnecessary so I reverted those.

ethcore/machine/src/test_helpers.rs
rokob: The `.into_iter().map(...).collect()` if the return type is `Result<_, _>` will also short circuit the map whenever there is an error. I'm not sure of a good way around that other than to create a new error variant that holds a vector of errors.

I tried to keep the same behavior as before as we will panic on the expect where it used to inside From::from.

rokob:
I got rid of the extraneous lines that were just formatting noise and not part of this actual change.

ethcore/machine/src/test_helpers.rs
niklasad1: right, let's keep it out from this PR then

PR #11145, opened 2019-10-08
Fix sccache statistics

General-Beck:
small fixes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -63,35 +63,35 @@ cargo-check 0 3:
   <<:                              *docker-cache-status
   script:
     - time cargo check --target $CARGO_TARGET --locked --no-default-features --verbose --color=always
-    - sccache -s
+    - sccache --show-stats
 
 cargo-check 1 3:
   stage:                           test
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -63,21 +63,21 @@ cargo-check 0 3:
...
-    - sccache --stop-server
+    - sccache -s
 
 cargo-check 2 3:
   stage:                           test
   <<:                              *docker-cache-status
   script:
     - time cargo check --target $CARGO_TARGET --locked --manifest-path util/io/Cargo.toml --features "mio" --verbose --color=always
-    - sccache --stop-server
+    - sccache -s
```
dvdplm: What's wrong with `--stop-server`?

General-Beck: i's  not work in darwin. and this is more correct

General-Beck: -s show stats only. --stop-server - stop server and show stats

dvdplm:
```suggestion
    - sccache --show-stats
```

More readable imo!

PR #11160, opened 2019-10-10
Aura: Report malice on sibling blocks from the same validator

afck:
Calls `report_malicious` if a validator creates multiple sibling blocks in the same step.

This was originally written by @vkomenda (https://github.com/poanetwork/parity-ethereum/pull/165), then squashed for easier rebasing on master. Cleanup of `received_step_hashes` was moved to `verify_block_family`, since `on_prepare_block` does not exist on master, and a unit test was added.

Original commit messages:
* added the map of received block header hashes
* do not return an error and remove older received block records
* optimised older record removal
* block hash comparison optimisation and the weak client ref fix
* SIBLING_MALICE_DETECTION_PERIOD constant
* review comments
* using step numbers instead of block numbers

### DIFF
--- a/ethcore/engines/authority-round/src/lib.rs
+++ b/ethcore/engines/authority-round/src/lib.rs
@@ -482,6 +482,8 @@ pub struct AuthorityRound {
 	two_thirds_majority_transition: BlockNumber,
 	maximum_empty_steps: usize,
 	machine: Machine,
+	/// History of step hashes recently received from peers.
+	received_step_hashes: RwLock<BTreeMap<(u64, Address), H256>>,
 }
 
 // header-chain validator.
@@ -572,7 +574,7 @@ fn header_expected_seal_fields(header: &Header, empty_steps_transition: u64) -> 
 
 fn header_step(header: &Header, empty_steps_transition: u64) -> Result<u64, ::rlp::DecoderError> {
 	Rlp::new(&header.seal().get(0).unwrap_or_else(||
-		panic!("was either checked with verify_block_basic or is genesis; has {} fields; qed (Make sure the spec
+		panic!("was either checked with verify_block_basic or is genesis; has {} fields; qed (Make sure the spec \
 				file has a correct genesis seal)", header_expected_seal_fields(header, empty_steps_transition))
 	))
 	.as_val()
@@ -751,6 +753,7 @@ impl AuthorityRound {
 				two_thirds_majority_transition: our_params.two_thirds_majority_transition,
 				strict_empty_steps_transition: our_params.strict_empty_steps_transition,
 				machine,
+				received_step_hashes: RwLock::new(Default::default()),
 			});
 
 		// Do not initialize timeouts for tests.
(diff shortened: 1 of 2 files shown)


ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1377,6 +1380,25 @@ impl Engine for AuthorityRound {
 			Err(EngineError::DoubleVote(*header.author()))?;
 		}
 
+		// Report malice if the validator produced other sibling blocks in the same step.
+		let received_step_key = (step, *header.author());
+		let new_hash = header.hash();
+		if self.received_step_hashes.read().get(&received_step_key).into_iter().any(|h| *h != new_hash) {
```
dvdplm: I must be missing something, `self.received_step_hashes.read().get(&received_step_key)` returns a single `H256` yes? Why not `if self.received_step_hashes.read().get(&received_step_key) != Some(&new_hash)`?

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1377,6 +1380,25 @@ impl Engine for AuthorityRound {
...
+			trace!(target: "engine", "Validator {} produced sibling blocks in the same step", header.author());
+			self.validators.report_malicious(header.author(), set_number, header.number(), Default::default());
+		} else {
+			self.received_step_hashes.write().insert(received_step_key, new_hash);
+		}
+
+		// Remove older step hash records.
+		const SIBLING_MALICE_DETECTION_PERIOD: u64 = 100;
```
dvdplm: It'd be nice to see a comment on why `100` was chosen.

ethcore/engines/authority-round/src/lib.rs
afck: `a != Some(b)` is `true` if `a` is `None`.
`a.into_iter().any(|h| h != b)` is `false` if `a` is `None`.

We could use `a.map_or(false, |h| h != b)`. Not sure if that's more readable. After all, we want to say something like: "If the map already contains a hash, _and_ that's different from `new_hash`, report the block author."

afck:
> an expert Aura code reader

It would be _really_ great to get the opinion of one of those!

As far as I understand, this PR fixes an omission: A validator does get reported as malicious if they create a block whose step is not greater than its parent's step. However, they do not get reported if they create two sibling blocks in the same step, which shouldn't be allowed in Aura either.

There's no good reason for the number 100, I think: It just needs to be large enough so that it's likely that any single-slot sibling block pair that could still do damage to consensus would be detected and reported. And not large enough to waste too much memory. (In that respect, 100 seems like a conservative choice to me.)

I'll add some documentation and comments, as far as I can.

ethcore/engines/authority-round/src/lib.rs
niklasad1: IMHO, `map_or` is more readable in this context because when using one element it may be confusing to the reader

afck:
I think it might make sense to use the number of validators _N_ instead of _100_ here, or maybe _2 N_: That would mean we are always able to detect multiple blocks per step in the current and the previous full cycle of steps. After a full cycle, blocks are likely to be finalized, since every validator has already built on top of them, and an attack would not be effective.

If you're okay with that, I'll change the `100` to `2 * validators.count()`.

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1377,6 +1391,25 @@ impl Engine for AuthorityRound {
...
+		let received_step_key = (step, *header.author());
+		let new_hash = header.hash();
+		if self.received_step_hashes.read().get(&received_step_key).map_or(false, |h| *h != new_hash) {
+			trace!(target: "engine", "Validator {} produced sibling blocks in the same step", header.author());
+			self.validators.report_malicious(header.author(), set_number, header.number(), Default::default());
+		} else {
+			self.received_step_hashes.write().insert(received_step_key, new_hash);
+		}
+
+		// Remove older step hash records.
```
dvdplm:
```suggestion
		// Remove hash records older than 100 steps (picked as a reasonable trade-off between memory consumption and fault-tolerance).
```

afck:
Great, thank you!
(So I don't replace the 100 with `2 * validators.count()`?)

dvdplm:
Oh missed that, sure that’s not a bad idea.

afck:
Another runner system failure: `Cannot connect to the Docker daemon at unix:///var/run/docker.sock.`.

dvdplm:
Best merge before the beast wakes up. ;)

PR #11161, opened 2019-10-10
Cleanup stratum a bit

dvdplm:
Salvage some code from https://github.com/paritytech/parity-ethereum/pull/10884 + some cleanup and typos.

### DIFF
--- a/miner/stratum/src/lib.rs
+++ b/miner/stratum/src/lib.rs
@@ -109,10 +109,6 @@ impl PushWorkHandler for Stratum {
 	fn push_work_all(&self, payload: String) {
 		self.implementation.push_work_all(payload, &self.tcp_dispatcher)
 	}
-
-	fn push_work(&self, payloads: Vec<String>) -> Result<(), Error> {
-		self.implementation.push_work(payloads, &self.tcp_dispatcher)
-	}
 }
 
 impl Drop for Stratum {
@@ -160,7 +156,7 @@ impl StratumImpl {
 
 	/// rpc method `mining.authorize`
 	fn authorize(&self, params: Params, meta: SocketMetadata) -> RpcResult {
-		params.parse::<(String, String)>().map(|(worker_id, secret)|{
+		params.parse::<(String, String)>().map(|(worker_id, secret)| {
 			if let Some(valid_secret) = self.secret {
 				let hash = keccak(secret);
 				if hash != valid_secret {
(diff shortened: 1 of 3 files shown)


miner/stratum/src/lib.rs
```diff
@@ -204,31 +204,33 @@ impl StratumImpl {
...
 
-			let mut hup_peers = HashSet::with_capacity(0); // most of the cases won't be needed, hence avoid allocation
+			// most of the cases won't be needed, hence avoid allocation
+			let mut hup_peers = HashSet::with_capacity(0);
 			let workers_msg = format!("{{ \"id\": {}, \"method\": \"mining.notify\", \"params\": {} }}", next_request_id, payload);
 			trace!(target: "stratum", "pushing work for {} workers (payload: '{}')", workers.len(), &workers_msg);
-			for (ref addr, _) in workers.iter() {
+			for (addr, _) in workers.iter() {
 				trace!(target: "stratum", "pusing work to {}", addr);
 				match tcp_dispatcher.push_message(addr, workers_msg.clone()) {
```
niklasad1:
```suggestion
				if let Err(err) = tcp_dispatcher.push_message(addr, workers_msg.clone()) {
```

miner/stratum/src/traits.rs
```diff
@@ -55,7 +55,7 @@ pub trait JobDispatcher: Send + Sync {
 /// Interface that can handle requests to push job for workers
 pub trait PushWorkHandler: Send + Sync {
 	/// push the same work package for all workers (`payload`: json of pow-specific set of work specification)
-	fn push_work_all(&self, payload: String) -> Result<(), Error>;
```
niklasad1: The trait looks a bit weird after this change because now `push_work_all` is `infallible` and `push_work` can fail.

I do realize it is because we don't want to `short-circuit` if one message fails when we need to send `x more` but still I guess it would be possible to keep all failed messages in a buffer and return `Err` if !buffer.is_empty()

dvdplm: > The trait looks a bit weird

It does, but it's better to get the weirdness "in your face" than hidden behind a comfy `Result` that is always `Ok`.

> it would be possible to keep all failed messages in a buffer

Yeah, and while that would be more satisfying I still wonder what use calling code can make with such an error? Honest question: would it be useful?

niklasad1: > Yeah, and while that would be more satisfying I still wonder what use calling code can make with such an error? Honest question: would it be useful?

I don't think so currently it is only called from [here](https://github.com/paritytech/parity-ethereum/blob/master/ethcore/src/miner/stratum.rs#L226) AFAIK and would require more changes then.

Also, seems like `push_work` is not used except in the tests.

We can merge this and investigate my points above separately

miner/stratum/src/lib.rs
```diff
@@ -204,70 +200,48 @@ impl StratumImpl {
...
 
 		if !hup_peers.is_empty() {
 			let mut workers = self.workers.write();
-			for hup_peer in hup_peers { workers.remove(&hup_peer); }
-		}
-
-		Ok(())
-	}
-
-	fn push_work(&self, payloads: Vec<String>, tcp_dispatcher: &Dispatcher) -> Result<(), Error> {
```
niklasad1: :+1:

PR #592, opened 2016-03-04
Fixed potential deadlock on startup

arkpar:

PR #11172, opened 2019-10-15
ethcore client: fix a double Read Lock bug in fn Client::logs()

BurtonQin:
***Description***
This PR fixes a double Read lock bug in ethcore client.
The two Read locks on self.chain reside in the same function Client::logs(), at the beginning and the end line of the function, respectively. The lock self.chain is a parking_lot::RwLock.

This bug is similar to #592

According to parking_lot::RwLock:
“readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock.”
“attempts to recursively acquire a read lock within a single thread may result in a deadlock.”

Therefore, once a function (e.g. restore_db, etc) requires a write lock in between the execution of function logs(), a double lock may happen.

***How I fix***
Reuse the chain and remove the second lock.

### DIFF
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -2038,7 +2038,7 @@ impl BlockChainClient for Client {
 			blocks
 		};
 
-		Ok(self.chain.read().logs(blocks, |entry| filter.matches(entry), filter.limit))
+		Ok(chain.logs(blocks, |entry| filter.matches(entry), filter.limit))
 	}
 
 	fn filter_traces(&self, filter: TraceFilter) -> Option<Vec<LocalizedTrace>> {


BurtonQin:
[clabot:check]

dvdplm:
Thank you for spotting this. Just curious: did you observe the deadlock in the wild?

niklasad1:
@dvdplm it is not a deadlock because it is a `RwLock` where multiple readers are ok

dvdplm:
>  it is not a deadlock because it is a `RwLock` where multiple readers are ok when there is no write acquired, which is the case here

But the second read-lock taken here is unnecessary right? And if there is a write lock attempted in another thread, the second read-lock here could deadlock? (Seems like the only place we take the write lock is [here](https://github.com/paritytech/parity-ethereum/blob/8f0f6d9efdd10e1bb5d35e285a129fdce74536d1/ethcore/src/client/client.rs#L1286), so it should be a rare problem I guess)

niklasad1:
> But the second read-lock taken here is unnecessary right? 

Yes

> And if there is a write lock attempted in another thread, the second read-lock here could deadlock? (Seems like the only place we take the write lock is here, so it should be a rare problem I guess)

Right, that it is correct. My bad it didn't read the PR description properly RwLock uses `task-fair locking`

PR #11178, opened 2019-10-17
Pause pruning while snapshotting

dvdplm:
WIP

### DIFF
--- a/ethcore/snapshot/src/service.rs
+++ b/ethcore/snapshot/src/service.rs
@@ -212,7 +212,7 @@ impl Restoration {
 		}
 
 		self.guard.disarm();
-		trace!(target: "snapshot", "restoration finalised correctly");
+		trace!(target: "snapshot", "Restoration finalised correctly");
 		Ok(())
 	}
 
@@ -413,7 +413,10 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 			Some(x) => x,
 			None => return Ok(0),
 		};
-
+		info!(target: "snapshot", "Migrating blocks from old db to new. Start: #{}/{:?}, Target: #{}/{:?}",
+			self.client.block_number(BlockId::Hash(start_hash)).unwrap_or_default(), start_hash,
+			self.client.block_number(BlockId::Hash(target_hash)).unwrap_or_default(), target_hash,
+		);
 		let mut batch = DBTransaction::new();
 		let mut parent_hash = start_hash;
 		while parent_hash != target_hash {
@@ -455,10 +458,10 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 				next_chain.commit();
 				next_db.key_value().flush().expect("DB flush failed.");
 				batch = DBTransaction::new();
-			}
 
-			if block_number % 10_000 == 0 {
-				info!(target: "snapshot", "Block restoration at #{}", block_number);
+				if block_number % 10_000 == 0 {
+					info!(target: "snapshot", "Block restoration at #{}", block_number);
+				}
 			}
 		}
 
@@ -497,6 +500,7 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 		}
 
 		info!("Taking snapshot at #{}", num);
+		let start_time = std::time::Instant::now();
 		self.progress.reset();
 
 		let temp_dir = self.temp_snapshot_dir();
@@ -506,21 +510,21 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 
 		let writer = LooseWriter::new(temp_dir.clone())?;
 
-
 		let guard = Guard::new(temp_dir.clone());
 		let res = client.take_snapshot(writer, BlockId::Number(num), &self.progress);
 		self.taking_snapshot.store(false, Ordering::SeqCst);
 		if let Err(e) = res {
+			// TODO[dvdplm]: when freezing pruning this condition is always met, causing this message to be logged at shutdown.
 			if client.chain_info().best_block_number >= num + client.pruning_history() {
-				// The state we were snapshotting was pruned before we could finish.
-				info!("Periodic snapshot failed: block state pruned. Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
+				info!("Periodic snapshot failed: the state at #{} was pruned before we could finish. Run with a longer `--pruning-history` (currently {}) or with `--no-periodic-snapshot`?", num, client.pruning_history());
+				trace!(target: "snapshot", "snapshot error: {:?}", e);
 				return Err(e);
 			} else {
 				return Err(e);
 			}
 		}
-
-		info!("Finished taking snapshot at #{}", num);
+		let end_time = std::time::Instant::now();
+		info!("Finished taking snapshot at #{}, in {:#?}", num, end_time - start_time);
 
 		let mut reader = self.reader.write();
 
@@ -528,11 +532,12 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 		*reader = None;
 
 		if snapshot_dir.exists() {
+			trace!(target: "snapshot", "Removing previous snapshot at {:?}", &snapshot_dir);
 			fs::remove_dir_all(&snapshot_dir)?;
 		}
 
 		fs::rename(temp_dir, &snapshot_dir)?;
-
+		trace!(target: "snapshot", "Moved new snapshot into place at {:?}", &snapshot_dir);
 		*reader = Some(LooseReader::new(snapshot_dir)?);
 
 		guard.disarm();
@@ -655,13 +660,18 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 		Ok(())
 	}
 
-	/// Import a previous chunk at the given path. Returns whether the block was imported or not
-	fn import_prev_chunk(&self, restoration: &mut Option<Restoration>, manifest: &ManifestData, file: io::Result<fs::DirEntry>) -> Result<bool, Error> {
+	/// Import a previous chunk at the given path. Returns whether the chunk was imported or not
+	fn import_prev_chunk(
+		&self,
+		restoration: &mut Option<Restoration>,
+		manifest: &ManifestData,
+		file: io::Result<fs::DirEntry>
+	) -> Result<bool, Error> {
 		let file = file?;
 		let path = file.path();
 
 		let mut file = File::open(path.clone())?;
-		let mut buffer = Vec::new();
+		let mut buffer = Vec::new(); // todo[dvdplm] estimate size and allocate up-front.
 		file.read_to_end(&mut buffer)?;
 
 		let hash = keccak(&buffer);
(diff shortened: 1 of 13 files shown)


ethcore/src/client/client.rs
```diff
@@ -204,6 +204,9 @@ pub struct Client {
 	/// Database pruning strategy to use for StateDB
 	pruning: journaldb::Algorithm,
 
+	/// Don't prune the state we're currently snapshotting
+	snapshotting_at: RwLock<AtomicU64>,
```
ordian: you don't need an `RwLock` around atomic

dvdplm: I know, but I need to mutate it from a `&self`. What's better, a `Mutex` (but it's written once and read many times)? Or `RwLock<u64>`?

ordian: Atomics allow modification while having `&self`. This is like an `RwLock<u64>`, but more performant.

ethcore/snapshot/src/service.rs
```diff
@@ -497,45 +500,45 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
...
-		if let Err(e) = res {
-			if client.chain_info().best_block_number >= num + client.pruning_history() {
-				// The state we were snapshotting was pruned before we could finish.
-				info!("Periodic snapshot failed: block state pruned. Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
-				return Err(e);
-			} else {
-				return Err(e);
+			if res.is_err() {
+				return res
 			}
```
ordian:
```suggestion
```

ethcore/snapshot/src/service.rs
```diff
@@ -497,45 +500,45 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
...
-		self.taking_snapshot.store(false, Ordering::SeqCst);
-		if let Err(e) = res {
-			if client.chain_info().best_block_number >= num + client.pruning_history() {
-				// The state we were snapshotting was pruned before we could finish.
-				info!("Periodic snapshot failed: block state pruned. Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
-				return Err(e);
-			} else {
-				return Err(e);
+			if res.is_err() {
+				return res
```
ordian:
```suggestion
```

ethcore/snapshot/src/service.rs
```diff
@@ -497,45 +500,45 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
...
-		let res = client.take_snapshot(writer, BlockId::Number(num), &self.progress);
-		self.taking_snapshot.store(false, Ordering::SeqCst);
-		if let Err(e) = res {
-			if client.chain_info().best_block_number >= num + client.pruning_history() {
-				// The state we were snapshotting was pruned before we could finish.
-				info!("Periodic snapshot failed: block state pruned. Run with a longer `--pruning-history` or with `--no-periodic-snapshot`");
-				return Err(e);
-			} else {
-				return Err(e);
+			if res.is_err() {
```
ordian:
```suggestion
```

ethcore/snapshot/src/service.rs
```diff
@@ -497,45 +500,45 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
...
-		let temp_dir = self.temp_snapshot_dir();
-		let snapshot_dir = self.snapshot_dir();
+			let _ = fs::remove_dir_all(&temp_dir); // expected to fail
 
-		let _ = fs::remove_dir_all(&temp_dir);
+			let writer = LooseWriter::new(temp_dir.clone())?;
 
-		let writer = LooseWriter::new(temp_dir.clone())?;
+			let guard = Guard::new(temp_dir.clone());
+			let res = client.take_snapshot(writer, BlockId::Number(num), &self.progress);
```
ordian:
```suggestion
			let _ = client.take_snapshot(writer, BlockId::Number(num), &self.progress)?;
```

parity/run.rs
```diff
@@ -68,11 +68,12 @@ use signer;
...
-// how often to take periodic snapshots.
-const SNAPSHOT_PERIOD: u64 = 5000;
+// How often we attempt to take a snapshot: only snapshot on blocknumbers that are multiples of this.
+// const SNAPSHOT_PERIOD: u64 = 5000;
+const SNAPSHOT_PERIOD: u64 = 50;
 
-// how many blocks to wait before starting a periodic snapshot.
-const SNAPSHOT_HISTORY: u64 = 100;
+// Start snapshots `history` blocks from the tip.
+ const SNAPSHOT_HISTORY: u64 = 100;
```
ordian:
```suggestion
const SNAPSHOT_HISTORY: u64 = 100;
```

ethcore/snapshot/src/watcher.rs
```diff
@@ -104,20 +102,36 @@ impl Watcher {
...
 			.filter_map(|h| self.oracle.to_number(h))
+			// …only keep the new blocks that have numbers bigger than period + history
+			// todo:    this seems nonsensical: period is always 5000 and history is always 100, so
+			//          this filters out blocknumbers that are lower than 5100; what's the point of that?
 			.filter(|&num| num >= self.period + self.history)
+			// Subtract `history` (i.e. `SNAPSHOT_HISTORY`, i.e. 100) from the block numbers.
+			// todo:    why? Here we back off 100 blocks such that the final block
+			//          number from which we start the snapshot is "(new) highest block" - 100)
+			//          Maybe we do this to avoid IO contention on the tip? Or for reorgs? If the
+			//          latter, it should be maybe be higher?
```
ordian: my bet is on reorgs, but maybe @rphmeier or @arkpar could comment on that

rphmeier: yes, it's about reorgs.

ethcore/types/src/snapshot.rs
```diff
@@ -39,31 +41,53 @@ pub enum Snapshotting {
 }
 
 /// A progress indicator for snapshots.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Progress {
 	/// Number of accounts processed so far
-	pub accounts: AtomicUsize,
+	accounts: AtomicUsize,
```
rphmeier: with this many fields, it's almost certainly better to use a single parking lot mutex wrapping a single struct containing all this data. maybe beyond the scope

arkpar:
> Make a copy of the data needed for the snapshot into a new rocksdb database and make a snapshot of that instead. This would likely require a very large chunk of disk space be fairly involved.

That was one of the goals for parity-db. 
The canonical state stored not in the trie, but directly in the database will take a fraction of the the space. It will also help greatly with query performance.

ethcore/types/src/snapshot.rs
dvdplm: Good idea, I'll make the change in a diff PR as it gets a bit noisy.

dvdplm:
@rphmeier @arkpar thank you for your input. I'll collect a few more opinions but for the time being I think pausing pruning is the least worst choice we have for the short term. 
Hopefully it'll give us enough time to plan the next steps, parity-db or other.

ethcore/src/client/client.rs
```diff
@@ -2533,48 +2548,61 @@ impl SnapshotClient for Client {
...
-		)?;
-		Ok(())
+		trace!(target: "snapshot", "Snapshot requested at block {:?}. Using block #{}/{:?}. Earliest block: #{}, earliest state era #{}. Using {} threads.",
+			at, actual_block_nr, block_hash, self.pruning_info().earliest_chain, earliest_era, processing_threads,
+		);
+		// Stop pruning from happening while the snapshot is under way.
+		self.snapshotting_at.store(actual_block_nr, Ordering::SeqCst);
+		{
+			scopeguard::defer! {{
+				trace!(target: "snapshot", "(defer) Unfreezing pruning, setting snapshotting_at back to 0");
```
ngotchac: Should _defer_ be a comment?

ethcore/src/client/client.rs
```diff
@@ -204,6 +204,9 @@ pub struct Client {
 	/// Database pruning strategy to use for StateDB
 	pruning: journaldb::Algorithm,
 
+	/// Don't prune the state we're currently snapshotting
+	snapshotting_at: AtomicU64,
```
ngotchac: Would it make more sense for it to be a `RwLock<Option<usize>>` ? Even though we would never take a snapshot at 0.

ethcore/src/client/client.rs
```diff
@@ -979,17 +985,26 @@ impl Client {
...
+				Some(earliest_era) if earliest_era + self.history <= latest_era => {
+					let freeze_at = self.snapshotting_at.load(Ordering::SeqCst);
+					if freeze_at > 0 && freeze_at == earliest_era {
+						trace!(target: "pruning", "Pruning is frozen at era {}; earliest era={}, latest era={}, journal_size={}, mem_used={}. Not pruning.",
+						       freeze_at, earliest_era, latest_era, state_db.journal_db().journal_size(), state_db.journal_db().mem_used());
+						break;
+					}
+					trace!(target: "pruning", "Pruning state for ancient era #{}; latest era={}, journal_size={}, mem_used={}.",
+					       earliest_era, latest_era, state_db.journal_db().journal_size(), state_db.journal_db().mem_used());
+					// todo[dvdplm] reinstate this before merge, logging mem is expensive:
```
ordian: unresolved todo?

ethcore/types/src/snapshot.rs
```diff
@@ -73,10 +97,37 @@ impl Progress {
...
+		}
+		let delta_acc = self.accounts.load(Ordering::Relaxed)
+			.saturating_sub(self.prev_accounts.load(Ordering::Relaxed));
+		let delta_bytes = self.bytes.load(Ordering::Relaxed)
+			.saturating_sub(self.prev_bytes.load(Ordering::Relaxed));
+		(delta_acc as f64 / dt, delta_bytes as f64 / dt)
+	}
+
+	/// Update state progress counters and set the last tick.
+	pub fn update(&self, accounts_delta: usize, bytes_delta: usize) {
```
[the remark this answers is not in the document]
dvdplm: My thinking was that as `update()` is called much more often than `rate()`, it made sense to do less computation here and leave it for the reader. 
There's a follow up PR to this one where all of `Progress` is in a `RwLock` and no atomics are used, so it ends up looking like [this](https://github.com/paritytech/parity-ethereum/pull/11197/files#diff-44822d6cb72087de8c76ba6ee20891a0R102).

ethcore/src/client/client.rs
dvdplm: Definitely. Or not be there at all.

ethcore/src/client/client.rs
dvdplm: I considered it. I think using an atomic here is faster (but not sure tbh) but it definitely reads better imo. It is written to once when a snapshot starts, and once again when it ends and in between only read. My thinking for omitting the `Option` here was that even if someone down the line would ask for as snapshot at block `0` due to a bug, nothing bad would happen, so the `Option` doesn't buy me much in terms of safety.

ngotchac: Yeah I'm not to worried about a potential bug, it was mainly for readability

PR #11179, opened 2019-10-17
[receipt]: add `sender` & `receiver` to `RichReceipts`

### LINKED ISSUE #11109: Transaction receipt is missing information.
- **Parity Ethereum version**: 2.5.9
- **Operating system**: Windows
- **Installation**: Docker
- **Fully synchronized**: yes
- **Network**: ethereum
- **Restarted**: yes

When fetching a transaction receipt immediately after getting the transaction hash in response to a `eth_sendRawTransaction` call, the receipt has null for a lot of values that should never be null.

An example receipt response for a simple 0 value transfer:
```json
{
    "jsonrpc": "2.0",
    "result": {
        "blockHash": null,
        "blockNumber": null,
        "contractAddress": null,
        "cumulativeGasUsed": "0x5208",
        "from": null,
        "gasUsed": "0x5208",
        "logs": [],
        "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
        "root": null,
        "status": "0x1",
        "to": null,
        "transactionHash": "0x2083491971605336023db17f2446d2af1561d4a6f788b4f2a1742b0ce70694f7",
        "transactionIndex": "0x0"
    },
    "id": null
}
```
While `blockHash` and `blockNumber` are expected to be null, `from` and `to` should _not_ be null.  The `root` being null is also bizarre, I would expect it to return the pending block root `root`, though I can appreciate why that is `null`.

The problem persists with a non-0 value transfer as well.

niklasad1:
Currently, we represent `Receipts` a bit different in the code base and currently we have `RpcReceipt`, `Receipt (only four fields)`, `RichReceipt`, and `LocalizedReceipt`.

When a `full-node` requests a transaction in the `rpc interface` we either respond with `RichReceipt` or `LocalizedReceipt`. `RichReceipt` was missing `sender address` and `receiver` address which this fixes

Attempt to close #11109


rpc/src/v1/tests/mocked/parity.rs
```diff
@@ -510,7 +510,7 @@ fn rpc_parity_block_receipts() {
...
+	let response = r#"{"jsonrpc":"2.0","result":[{"blockHash":"0x0000000000000000000000000000000000000000000000000000000000000003","blockNumber":"0x0","contractAddress":null,"cumulativeGasUsed":"0x5208","from":"0x0000000000000000000000000000000000000009","gasUsed":"0x5208","logs":[],"logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001","status":null,"to":null,"transactionHash":"0x0000000000000000000000000000000000000000000000000000000000000001","transactionIndex":"0x0"}],"id":1}"#;
```
niklasad1: root was removed from the serialization when it missing (is None)

ethcore/types/src/receipt.rs
```diff
@@ -129,6 +129,10 @@ pub struct RichReceipt {
 	pub log_bloom: Bloom,
 	/// Transaction outcome.
 	pub outcome: TransactionOutcome,
+	/// Receiver address
+	pub to: Option<H160>,
```
niklasad1: This is an `Option` because it missing when it is `Action::Create` transaction

tomusdrw: Might be good to add this as regular comment.

PR #11186, opened 2019-10-21
Make InstantSeal Instant again

### LINKED ISSUE #11173: instantSeal engine wont seal more than one block
- **Parity Ethereum version**: >=2.5.8
- **Operating system**: Archlinux (kernel 5.3.6)
- **Installation**: pacman (for <=2.5.9) and built from source (for 2.6.4)
- **Fully synchronized**: N/A
- **Network**: custom instantSeal ([Chain file](https://gist.github.com/tristan/64cc72a211910000f71eefba3dff8876))
- **Restarted**: yes

After upgrading to 2.5.9 I noticed that my instantSeal network would no longer seal more than one block. i.e. The first transaction submitted correctly creates a new block, but after submitting a second transaction nothing happens.

I created a script to test this which can be found [here](https://gist.github.com/tristan/6d02fca748413c4855a9210c46d817a6). The script checks the parity version, creates the relevant chain file, starts parity, then attempts to submit a number of transactions, checking they have been included in a block before continuing to the next. Using parity 2.5.7, the script runs to completion as expected. Upgrading to 2.5.8 and later (including 2.6.4), the script will get stuck forever waiting for the 2nd block to be sealed.

From the logs (with `--logging own_tx=trace,rpc=trace,chain=trace,engine=trace,miner=trace,state=trace`) i see a bunch of things around the first transaction (which i don't really follow), but for the 2nd transaction all i see is the `Imported to the pool` message (from own_tx) and nothing else.

I've tried a bunch of different options including, different networkId, different starting nonces for my base account, different base accounts, sending the second transaction from a different account, none of which changed the outcome.

Parity is started using the following arguments
```
/usr/bin/parity 
    --port $PORT 
    --no-color
    --chain $TMP_DIR/chain.json 
    --author 0102030405060708090001020304050607080900 
    --tracing on 
    --node-key 74a32e8a5ea986ebb4becf4746a09b2bb7abb8ee09f6ead3bc4c0d8968fc27ed
    --base-path $TMP_DIR/data
    --jsonrpc-port $JSONRPC_PORT 
    --jsonrpc-hosts all 
    --ws-interface local 
    --ws-port $WS_PORT 
    --ws-origins all 
    --ws-hosts all 
    --min-gas-price 10000000000 
    --logging "own_tx=trace,rpc=trace,chain=trace,engine=trace,miner=trace,state=trace"
```
The chain file used can be seen [here](https://gist.github.com/tristan/64cc72a211910000f71eefba3dff8876).

seunlanlege:
[this](https://github.com/paritytech/parity-ethereum/pull/10995/files#diff-662498a077300e2e5277a2281eb3be28R1432-R1434) call to `update_sealing` calls `requires_reseal`

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L1266-L1276

and `requires_reseal` does a few checks and updates `SealingWork.next_allowed_reseal`

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L637-L663

for `InstantSeal` `sealing_state()` is always `SealingState::Ready` 

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/engines/instant-seal/src/lib.rs#L75

so `should_disable_sealing` would be `false` for `InstantSeal` and so `next_allowed_reseal` is updated based on `reseal_min_period`

Now `update_sealing` calls `Miner.seal_and_import_block_internally` which in turn calls `ImportSealedBlock::import_sealed_block` which in turn calls `Miner.chain_new_blocks`.

 `Miner.chain_new_blocks` resets `next_allowed_reseal` to `Instant::now()` if a block was successfully imported.

https://github.com/paritytech/parity-ethereum/blob/6993ec95312f3ee446345ecb2591f0c66a9b4fcb/ethcore/src/miner/miner.rs#L1415-L1420

Previously `update_sealing` was called immediately after an internal import with the hopes of sealing a new block with local transactions (#9660) but it did so without the knowledge of the transaction_queue. And so if there were no transactions to seal, `next_allowed_reseal` is updated, but `chain_new_blocks` is never called to reset it back to `Instant::now()`

And now the next time a transaction is sent, it won't trigger a block sealing because `reseal_allowed()` returns `false`

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L1009-L1011

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L233-L238

This PR adds a new method to the `Engine` trait `should_reseal_on_update` and checks if there are local pending transactions before calling `update_sealing` after a block import.

`should_reseal_on_update` tells the miner that the engine is interested in sealing a new block if there are local pending transactions. 

closes #11173

### DIFF
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -146,12 +146,12 @@ pub trait TransactionInfo {
 /// Provides various blockchain information, like block header, chain state etc.
 pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 
-/// do we want to force update sealing
+/// Do we want to force update sealing?
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub enum ForceUpdateSealing {
-	/// Ideally
+	/// Ideally you want to use `No` at all times as `Yes` skips `reseal_required` checks.
 	Yes,
-	/// yes
+	/// Don't skip `reseal_required` checks
 	No
 }
 /// Client facilities used by internally sealing Engines.
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -149,7 +149,7 @@ pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 /// Client facilities used by internally sealing Engines.
 pub trait EngineClient: Sync + Send + ChainInfo {
 	/// Make a new block and seal it.
-	fn update_sealing(&self);
+	fn update_sealing(&self) -> bool;
 
 	/// Submit a seal for a block in the mining queue.
 	fn submit_seal(&self, block_hash: H256, seal: Vec<Bytes>);
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -146,10 +146,18 @@ pub trait TransactionInfo {
 /// Provides various blockchain information, like block header, chain state etc.
 pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 
+/// do we want to force update sealing
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum ForceUpdateSealing {
+	/// Ideally
+	Yes,
+	/// yes
+	No
+}
 /// Client facilities used by internally sealing Engines.
 pub trait EngineClient: Sync + Send + ChainInfo {
 	/// Make a new block and seal it.
-	fn update_sealing(&self);
+	fn update_sealing(&self, force: ForceUpdateSealing);
 
 	/// Submit a seal for a block in the mining queue.
 	fn submit_seal(&self, block_hash: H256, seal: Vec<Bytes>);
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -149,6 +149,7 @@ pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 /// Client facilities used by internally sealing Engines.
 pub trait EngineClient: Sync + Send + ChainInfo {
 	/// Make a new block and seal it.
+	/// returns true if a block was imported and sealed.
 	fn update_sealing(&self) -> bool;
 
 	/// Submit a seal for a block in the mining queue.
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -149,8 +149,7 @@ pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 /// Client facilities used by internally sealing Engines.
 pub trait EngineClient: Sync + Send + ChainInfo {
 	/// Make a new block and seal it.
-	/// returns true if a block was imported and sealed.
-	fn update_sealing(&self) -> bool;
+	fn update_sealing(&self);
 
 	/// Submit a seal for a block in the mining queue.
 	fn submit_seal(&self, block_hash: H256, seal: Vec<Bytes>);
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -45,7 +45,6 @@ use machine::{
 use vm::{EnvInfo, Schedule, CallType, ActionValue};
 
 use crate::signer::EngineSigner;
-use std::time::Duration;
 
 /// A system-calling closure. Enacts calls on a block's state from the system address.
 pub type SystemCall<'a> = dyn FnMut(Address, Vec<u8>) -> Result<Vec<u8>, String> + 'a;
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -45,6 +45,7 @@ use machine::{
 use vm::{EnvInfo, Schedule, CallType, ActionValue};
 
 use crate::signer::EngineSigner;
+use std::time::Duration;
 
 /// A system-calling closure. Enacts calls on a block's state from the system address.
 pub type SystemCall<'a> = dyn FnMut(Address, Vec<u8>) -> Result<Vec<u8>, String> + 'a;
@@ -188,10 +189,13 @@ pub trait Engine: Sync + Send {
 	/// Returns the engine's current sealing state.
 	fn sealing_state(&self) -> SealingState { SealingState::External }
 
-	/// Called in `chain_new_blocks` if there are local pending txs
+	/// Called in `miner.chain_new_blocks` if the engine wishes to `update_sealing`
+	/// after a block was recently sealed, and there are local pending tx in the pool.
 	///
-	/// Does nothing by default
-	fn maybe_update_sealing(&self) {}
+	/// returns false by default
+	fn should_reseal_on_update(&self) -> bool {
+		false
+	}
 
 	/// Attempt to seal the block internally.
 	///
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -189,7 +189,7 @@ pub trait Engine: Sync + Send {
 	fn sealing_state(&self) -> SealingState { SealingState::External }
 
 	/// Called in `miner.chain_new_blocks` if the engine wishes to `update_sealing`
-	/// after a block was recently sealed, and there are local pending tx in the pool.
+	/// after a block was recently sealed.
 	///
 	/// returns false by default
 	fn should_reseal_on_update(&self) -> bool {
--- a/ethcore/engine/src/engine.rs
+++ b/ethcore/engine/src/engine.rs
@@ -188,6 +188,11 @@ pub trait Engine: Sync + Send {
 	/// Returns the engine's current sealing state.
 	fn sealing_state(&self) -> SealingState { SealingState::External }
 
+	/// Called in `chain_new_blocks` if there are local pending txs
+	///
+	/// Does nothing by default
+	fn maybe_update_sealing(&self) {}
+
 	/// Attempt to seal the block internally.
 	///
 	/// If `Some` is returned, then you get a valid seal.
--- a/ethcore/engines/instant-seal/Cargo.toml
+++ b/ethcore/engines/instant-seal/Cargo.toml
@@ -7,10 +7,10 @@ edition = "2018"
 license = "GPL-3.0"
 
 [dependencies]
+client-traits = { path = "../../client-traits" }
 common-types = { path = "../../types" }
 engine = { path = "../../engine" }
 ethjson = { path = "../../../json" }
-client-traits = { path = "../../client-traits" }
 ethereum-types = "0.8.0"
 keccak-hash = "0.4.0"
 parking_lot = "0.9"
--- a/ethcore/engines/instant-seal/Cargo.toml
+++ b/ethcore/engines/instant-seal/Cargo.toml
@@ -13,6 +13,8 @@ ethjson = { path = "../../../json" }
 client-traits = { path = "../../client-traits" }
 ethereum-types = "0.8.0"
 keccak-hash = "0.4.0"
+parking_lot = "0.9"
+crossbeam-channel = "0.3.9"
 machine = { path = "../../machine" }
 trace = { path = "../../trace" }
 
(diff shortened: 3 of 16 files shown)


seunlanlege:
> If this is a fix to some issues we have with InstantSeal, I don't think it's the right direction. It would be better to re-architecture the way engines and miners interact and new block production is triggered (perhaps using Futures, like in substrate) to clean this up.

I completely agree with this.

dvdplm:
> Previously the InstantSeal engine would only seal a block once and then never until reseal_allowed()

So as per the issue reported in #11173 (repeatable) something changed in v2.5.8+ – have you found out what the change was that "broke" this? I know you had a hypothesis that [this](https://github.com/paritytech/parity-ethereum/pull/10995/files#diff-662498a077300e2e5277a2281eb3be28R1432-R1434) code was involved but I do not understand the connection between that and the code snippets in the description above. Can you elaborate on that?

Is it like this: the fix in https://github.com/paritytech/parity-ethereum/pull/10995 is good but uncovered this quirk where our default config values do not work with `InstantSeal`? Or…?

Either way, we need a fix here of some sort, code or docs, so we don't have to spend days digging through code to help users. Do we need better defaults in the `instant_seal.json`? Is that where per-engine config should happen? Currently `InstantSeal` has no parameters at all but maybe it should?

seunlanlege:
[this](https://github.com/paritytech/parity-ethereum/pull/10995/files#diff-662498a077300e2e5277a2281eb3be28R1432-R1434) call to `update_sealing` calls `requires_reseal`

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L1266-L1276

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L637-L663

for `InstantSeal` `sealing_state()` is always `SealingState::Ready` 

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/engines/instant-seal/src/lib.rs#L75

so `should_disable_sealing` would be `false` for `InstantSeal` and so `next_allowed_reseal` is updated based on `reseal_min_period`

And now the next time a transaction is sent, it won't trigger a block sealing because `reseal_allowed()` returns `false`

https://github.com/paritytech/parity-ethereum/blob/6960d35abb9f11c27b47226e402ec617bfe585f1/ethcore/src/miner/miner.rs#L1009-L1011


Hopefully this makes sense.

seunlanlege:
> Currently InstantSeal has no parameters at all but maybe it should?

I worry about having to specify params to an engine to do what its name says it does. 

> Do we need better defaults in the instant_seal.json? Is that where per-engine config should happen?

With the way the code is structured, the `Miner` "speaks for" the `Engine` as to when sealing should occur, so the only way to achieve this is to let the engine itself decide for itself if sealing should occur or not.

tomusdrw:
Thanks for the explanations @seunlanlege. Your hypothesis makes a lot of sense and I see the purpose of this PR better now.

However I'd be more inclined to just introduce a thread to `InstantSeal` that would periodically (every `reseal_min_period`?) trigger sealing, just as we have for `Clique`.

Why I don't like this approach I feel it's cleaner and safer than overriding `reseal_min_period` for this specific engine. Long-term it would be best to re-work the miner-engine interaction to actually have `engine` controlling the sealing times as it seems to be becoming a common pattern.

dvdplm:
> I worry about having to specify params to an engine to do what its name says it does.

Good point. (And thank you for the great description, it sort of makes a little bit more sense now, maybe, much appreciated.)

@tomusdrw Regarding the "stepper thread" approach: what if `reseal_min_period` is `0`, or very short? Or are you saying we should enforce a minimum resealing period?

tomusdrw:
> Regarding the "stepper thread" approach: what if reseal_min_period is 0, or very short? Or are you saying we should enforce a minimum resealing period?

No, not at all. `InstantSeal` is creating blocks as soon as there are any transactions (i.e. it doesn't create empty blocks). So having `reseal_min_period` set to `0` just means that blocks are created as soon as transactions are created. If we want to change `InstantSeal` to produce empty blocks as well, I'd say it would be sensible to rather produce a new block every `reseal_max_period`.

dvdplm:
Hmm. I am still confused, apologies. In Clique the stepper thread ends up calling `update_sealing()` every 2 seconds (`SEALING_FREQ`) – if we do the same for `InstantSeal` wouldn't it end up sealing blocks at most every 2 seconds?
The purpose of `reseal_min_period` is to accumulate more txs I guess?

tomusdrw:
@dvdplm Yes, but block sealing is also triggered in other situation (like transaction import). If we call `update_sealing` every 2 seconds the blocks will be actually be created at least every 2 seconds, cause otherwise incoming transactions should trigger sealing as well (but at most `reseal_min_period`).

Actually `reseal_min_period` is more related to `PoW`, where finding a block seal take siginifcant amount of time. From one perspective we want to include as much transactions in a block as possible, but at the same time re-setting the mining equpiment has some constant cost, so we don't want to send new header hash too often to avoid just idling. Hence `reseal_min_period` says: "Create new pending block, but not if we just created one recently."

seunlanlege:
@tomusdrw could you take a look at this?

ethcore/src/miner/miner.rs
```diff
@@ -1441,7 +1446,7 @@ impl miner::MinerService for Miner {
 				let engine = self.engine.clone();
 				let accounts = self.accounts.clone();
 				let service_transaction_checker = self.service_transaction_checker.clone();
-
+				let cloned_tx_queue = self.transaction_queue.clone();
```
tomusdrw:
```suggestion
```

ethcore/src/miner/miner.rs
```diff
@@ -1451,8 +1456,11 @@ impl miner::MinerService for Miner {
 						service_transaction_checker.as_ref(),
 					);
 					queue.cull(client);
-					if is_internal_import {
-						chain.update_sealing();
+					if cloned_tx_queue.all_transaction_hashes().len() > 0 {
```
tomusdrw:
```suggestion
					if queue.has_local_pending_transactions() {
```

ethcore/src/miner/miner.rs
```diff
@@ -1461,8 +1469,11 @@ impl miner::MinerService for Miner {
 				}
 			} else {
 				self.transaction_queue.cull(client);
-				if is_internal_import {
-					self.update_sealing(chain);
+				if self.transaction_queue.all_transaction_hashes().len() > 0 {
```
tomusdrw:
```suggestion
				if self.transaction_queue.has_local_pending_transactions() {
```

ethcore/src/miner/miner.rs
```diff
@@ -1451,8 +1456,11 @@ impl miner::MinerService for Miner {
 						service_transaction_checker.as_ref(),
 					);
 					queue.cull(client);
-					if is_internal_import {
-						chain.update_sealing();
+					if cloned_tx_queue.all_transaction_hashes().len() > 0 {
+						if !chain.update_sealing() {
```
tomusdrw: But this is not really a desired behavior for all engines. It's just a property of `InstantSeal` that it should produce blocks as long as there are transactions in the pool.

Adding this snip of code here will actually cause all engines to produce pending blocks way too often and that's really wasteful. I think we should rather make this behavior engine-specific.

ethcore/client-traits/src/lib.rs
```diff
@@ -149,7 +149,7 @@ pub trait BlockChain: ChainInfo + BlockInfo + TransactionInfo {}
 /// Client facilities used by internally sealing Engines.
 pub trait EngineClient: Sync + Send + ChainInfo {
 	/// Make a new block and seal it.
-	fn update_sealing(&self);
+	fn update_sealing(&self) -> bool;
```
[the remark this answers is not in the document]
tomusdrw: It's actually not used anywhere afaict.

ethcore/src/miner/miner.rs
```diff
@@ -1451,7 +1452,9 @@ impl miner::MinerService for Miner {
 						service_transaction_checker.as_ref(),
 					);
 					queue.cull(client);
-					if is_internal_import {
+					if queue.has_local_pending_transactions() &&
```
tomusdrw: We can get rid of `queue.has_local_pending_transactions()`

ethcore/src/miner/miner.rs
```diff
@@ -1461,7 +1464,9 @@ impl miner::MinerService for Miner {
 				}
 			} else {
 				self.transaction_queue.cull(client);
-				if is_internal_import {
+				if self.transaction_queue.has_local_pending_transactions() &&
```
tomusdrw: Same.

ethcore/engine/src/engine.rs
```diff
@@ -188,6 +188,14 @@ pub trait Engine: Sync + Send {
 	/// Returns the engine's current sealing state.
 	fn sealing_state(&self) -> SealingState { SealingState::External }
 
+	/// Called in `miner.chain_new_blocks` if the engine wishes to `update_sealing`
+	/// after a block was recently sealed, and there are local pending tx in the pool.
```
tomusdrw: I think we don't need to guard this by additional check with the transaction pool. Just triggering a reseal will cause the `generate_seal` method to check if there are transactions or not:
https://github.com/paritytech/parity-ethereum/blob/81ca599f2adf7737c4c57f49fce94a304c6fc242/ethcore/engines/instant-seal/src/lib.rs#L78

tomusdrw: ^^ Still worth updating.

ethcore/engines/instant-seal/Cargo.toml
```diff
@@ -10,8 +10,11 @@ license = "GPL-3.0"
 common-types = { path = "../../types" }
 engine = { path = "../../engine" }
 ethjson = { path = "../../../json" }
+client-traits = { path = "../../client-traits" }
```
tomusdrw: Sort?

ethcore/src/miner/miner.rs
```diff
@@ -1412,22 +1416,20 @@ impl miner::MinerService for Miner {
...
 
-		if has_new_best_block || (imported.len() > 0 && self.options.reseal_on_uncle) {
+		if has_new_best_block {
 			// Reset `next_allowed_reseal` in case a block is imported.
 			// Even if min_period is high, we will always attempt to create
 			// new pending block.
 			self.sealing.lock().next_allowed_reseal = Instant::now();
 
-			if !is_internal_import {
+			if !is_internal_import && imported.len() > 0 && self.options.reseal_on_uncle {
```
tomusdrw: Perhaps this stuff is not needed? Seems that my patch might have been made on outdated version of master.

seunlanlege:
needs one more approval cc @niklasad1 @ordian

ethcore/src/miner/miner.rs
```diff
@@ -1305,13 +1308,14 @@ impl miner::MinerService for Miner {
 				if self.seal_and_import_block_internally(chain, block) {
 					trace!(target: "miner", "update_sealing: imported internally sealed block");
 				}
+				return
```
ordian: this return doesn't change anything, right?

ethcore/src/miner/miner.rs
```diff
@@ -1451,8 +1454,9 @@ impl miner::MinerService for Miner {
 						service_transaction_checker.as_ref(),
 					);
 					queue.cull(client);
-					if is_internal_import {
```
ordian: was this previously true only for instant seal engine? otherwise, should it be something like
```rust
				if engine.should_reseal_on_update() {
						// force update_sealing here to skip `reseal_required` checks
						chain.update_sealing(ForceUpdateSealing::Yes);
				} else if is_internal_import {
						chain.update_sealing(ForceUpdateSealing::No);
				}
```

ethcore/src/miner/miner.rs
seunlanlege: yeah it doesn't

ethcore/src/miner/miner.rs
seunlanlege: yeah it was only for `InstantSeal`.

PR #11192, opened 2019-10-22
Type annotation for next_key() matching of json filter options

### LINKED ISSUE #11189: Can no longer filter transactions with RPC API
- **Parity Ethereum version**: v2.7.0-nightly-6960d35ab-20191019
- **Operating system**: Ubuntu 19.04
- **Fully synchronized**: no
- **Network**: ethereum
- **Restarted**: yes

Yesterday I was contacted by a stranger about a PR I did a few months ago #10506. The RPC API `parity_pendingTransactions` was able to accept filtering parameters in the following form:

```
{
    "from": {
        "eq": "0x5f3dffcf347944d3739b0805c934d86c8621997f"
    },
    "to": {
        "eq": "0xe8b2d01ffa0a15736b2370b6e5064f9702c891b6"
    },
}
```

This no longer seems to be the case. It's also not documented in the [docs](https://wiki.parity.io/JSONRPC-parity-module#parity_pendingtransactions). The changed I made are still part of the codebase.

**How to reproduce**
1. Run the parity node
2. Make the following RPC request:
```
curl --data '{"method":"parity_pendingTransactions","params":[{"from":{"eq":"0x5f3dffcf347944d3739b0805c934d86c8621997f"},"to": {"eq":"0xe8b2d01ffa0a15736b2370b6e5064f9702c891b6"}}],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
```
3. The node returns the following error message:
```
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params: invalid type: map, expected usize."},"id":1}
```

**Expected behavior**
The node should return the filtered transactions (if any).

Please clarify whether this feature is actually intended to work/should have been implemented.

Thanks.

lamafab:
Fixes #11189. Very small and easy changes.

**Example**
```
"params": [
  15,
  {
    "to": {
        "eq": "0xe8b2d01ffa0a15736b2370b6e5064f9702c891b6"
    },
    "gas": {
        "gt": "0x493e0"
    },
  }
]
```

**`curl` example:**
```
curl --data '{"method":"parity_pendingTransactions","id": 1,"params":[5,{"from":{"eq":"0x5f3dffcf347944d3739b0805c934d86c8621997f"},"to": {"eq":"0xe8b2d01ffa0a15736b2370b6e5064f9702c891b6"}}],"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST localhost:8545
```

EDIT:
PR for documentation https://github.com/paritytech/wiki/pull/345

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ To start Parity Ethereum as a regular user using `systemd` init:
 
 ## 4. Testing <a id="chapter-004"></a>
 
-You can run tests with the following commands:
+Download the required test files: `git submodule update --init --recursive`. You can run tests with the following commands:
 
 * **All** packages
   ```
--- a/ethcore/src/miner/filter_options.rs
+++ b/ethcore/src/miner/filter_options.rs
@@ -160,8 +160,8 @@ impl<'de> Deserialize<'de> for FilterOptions {
                 M: MapAccess<'de>,
             {
                 let mut filter = FilterOptions::default();
-                while let Some(key) = map.next_key()? {
-                    match key {
+                while let Some(key) = map.next_key::<String>()? {
+                    match key.as_str() {
                         "from" => {
                             filter.from = map.validate_from()?;
                         },
@@ -221,8 +221,8 @@ impl<'de> Deserialize<'de> for FilterOptions {
                         let mut counter = 0;
                         let mut f_op = Wrapper::O(FilterOperator::Any);
 
-                        while let Some(key) = map.next_key()? {
-                            match key {
+                        while let Some(key) = map.next_key::<String>()? {
+                            match key.as_str() {
                                 "eq" => f_op = W::O(FilterOperator::Eq(map.next_value()?)),
                                 "gt" => f_op = W::O(FilterOperator::GreaterThan(map.next_value()?)),
                                 "lt" => f_op = W::O(FilterOperator::LessThan(map.next_value()?)),
--- a/rpc/src/v1/tests/mocked/parity.rs
+++ b/rpc/src/v1/tests/mocked/parity.rs
@@ -316,11 +316,44 @@ fn rpc_parity_unsigned_transactions_count_when_signer_disabled() {
 }
 
 #[test]
-fn rpc_parity_pending_transactions() {
+fn rpc_parity_pending_transactions_without_limit_without_filter() {
 	let deps = Dependencies::new();
 	let io = deps.default_client();
 
 	let request = r#"{"jsonrpc": "2.0", "method": "parity_pendingTransactions", "params":[], "id": 1}"#;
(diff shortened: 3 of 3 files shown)


lamafab:
Should the tests just cover JSON validity or actually contain filtering logic? I remember trying to do that but had trouble creating sample test transactions. I was told we'll create those eventually, but never got down to it.

(do note that the filter_options already have simple unit tests for JSON validity, but those do not cover the full RPC call, as shown by this bug).

niklasad1:
> Should the tests just cover JSON validity or actually contain filtering logic? I remember trying to do that but had trouble creating sample test transactions. I was told we'll create those eventually, but never got down to it.

All right, I see it requires some additional mock code to create some blocks with `to` and `from` I suppose. I think that should rather trivial something like:

```rust
	use v1::helpers::fake_sign;
	use types::transaction::*;
	use std::str::FromStr;

	let from: Address = Address::from_str("5f3dffcf347944d3739b0805c934d86c8621997f").unwrap();

	let deps = Dependencies::new();
	{
		let miner = deps.miner.clone();
		for i in 0..10 {
			let tx = Transaction::default().fake_sign(from);
			miner.pending_transactions.lock().insert([i as u8; 32].into(), tx);
		}
	}
	let io = deps.default_client();

	let request = r#"{"jsonrpc": "2.0", "method": "parity_pendingTransactions", "params":[1, {"from": {"eq": "0x5f3dffcf347944d3739b0805c934d86c8621997f"}}], "id":1}"#;

```

lamafab:
@niklasad1 I just updated the PR. This includes some basic JSON deserialization tests for the pending transactions RPC API. There should be logic tests, which I can try to implement when I have some spare time, but I can't right now.

Also, in the README.md I mentioned to pull the git submodules required for running tests.

README.md
```diff
@@ -151,7 +151,7 @@ To start Parity Ethereum as a regular user using `systemd` init:
 
 ## 4. Testing <a id="chapter-004"></a>
 
-You can run tests with the following commands:
+Download the required test files: `git submodule update --init --recursive`. You can run tests with the following commands:
```
dvdplm: Does `cargo test --all` fail without this? I thought it was needed only for `cargo test --all --features=json-tests` but maybe I'm wrong?

niklasad1: Yes, it fails we have [this macro](https://github.com/paritytech/parity-ethereum/blob/master/rpc/src/v1/tests/mod.rs#L31-#L41) which requires the `ethereum tests submodule` which is used in a couple of `rpc tests`

lamafab: I don't know if with `--all`, but it did for me when running `cargo test -p parity-json`

PR #11197, opened 2019-10-23
Use a lock instead of atomics for snapshot Progress

dvdplm:
After https://github.com/paritytech/parity-ethereum/pull/11178 `snapshot::Progress` is a bit crowded. This PR sticks the whole struct inside a `RwLock`.

Not very interesting code here, but [this](https://github.com/paritytech/parity-ethereum/pull/11197/files#diff-efbf5065ea836912f2329b8f6ef5f644R266-R278) change could use an extra eye or two to ensure I got the lock scope right.


ethcore/types/src/snapshot.rs
```diff
@@ -44,89 +43,68 @@ pub enum Snapshotting {
...
-		self.prev_bytes.store(
-			self.bytes.fetch_add(bytes_delta, Ordering::SeqCst),
-			Ordering::SeqCst
-		);
+	pub fn update(&mut self, accounts_delta: u64, bytes_delta: u64) {
+		self.last_tick = Instant::now();
+		self.prev_accounts = self.accounts;
+		self.prev_bytes = self.bytes;
+		self.accounts += accounts_delta;
+		self.bytes +=bytes_delta;
```
ordian:
```suggestion
		self.bytes += bytes_delta;
```

dvdplm:
> While switching to locks makes the code more readable

Interesting you think that (I don't, tbh), @ngotchac had a similar comment. Not that it matters much ofc.

I definitely agree that locks are much harder to reason about.

ethcore/types/src/snapshot.rs
```diff
@@ -44,89 +43,68 @@ pub enum Snapshotting {
 #[derive(Debug)]
 pub struct Progress {
 	/// Number of accounts processed so far
-	accounts: AtomicUsize,
+	accounts: u64,
```
niklasad1: yeah, I like `u64` instead `usize` here

parity/snapshot_cmd.rs
```diff
@@ -257,20 +258,25 @@ impl SnapshotCommand {
...
-					let bytes = ::informant::format_bytes(cur_size as usize);
-					info!("Snapshot: {} accounts (state), {} blocks, {} bytes", p.accounts(), p.blocks(), bytes);
+			loop {
+				{
+					let progress = p.read();
+					if !progress.done() {
+						let cur_size = progress.bytes();
+						if cur_size != last_size {
+							last_size = cur_size;
+							let bytes = ::informant::format_bytes(cur_size as usize);
```
niklasad1: this will truncate `u64` when `usize::max_value() < u64::max_value()`, I don't know it is issue or if we just should change `informant::format_bytes` to take an `u64` instead

dvdplm: The `usize`s here come from the io methods deeper down the stack that actually write bytes; they all return `usize` for the nr of bytes written. I wanted to be consistent and went with `u64` but we can go the other way too if preferred. I don't have a strong opinion tbh.

niklasad1: ok, I don't have strong opinions of this but I think `truncation casting` deserves a comment because it is not clear to me at least whether it is intended or bug.

parity/snapshot_cmd.rs
```diff
@@ -257,20 +258,25 @@ impl SnapshotCommand {
...
 			::std::thread::sleep(Duration::from_secs(5));
-
 			let mut last_size = 0;
-			while !p.done() {
-				let cur_size = p.bytes();
-				if cur_size != last_size {
-					last_size = cur_size;
-					let bytes = ::informant::format_bytes(cur_size as usize);
-					info!("Snapshot: {} accounts (state), {} blocks, {} bytes", p.accounts(), p.blocks(), bytes);
+			loop {
```
niklasad1: I wonder how much slower this loop would execute w.r.t. reading a `RwLock` instead of `AtomicUsize` in each iteration?

parity/snapshot_cmd.rs
dvdplm: I went the other way in #ee4a0a9b0 ad changed `format_bytes()` to take a `u64` instead.

parity/snapshot_cmd.rs
dvdplm: Hopefully using a `RwLock` is faster, but I agree, a benchmark providing a hard number would be a good thing. The reason it's (hopefully) faster is there were several atomics getting written to with `SeqCst` ordering in several places, so having a single lock is hopefully faster.

PR #11198, opened 2019-10-23
Improve the C API and add a C++ wrapper

DemiMarie-parity:
This improves the C API and adds a C++ wrapper.  The new API uses proper
opaque structs instead of `void`, and is better documented.  The C++
wrapper is designed for modern C++17, and uses exceptions to signal
errors.


DemiMarie-parity:
This breaks the CI because the C++ wrapper I wrote is not compatible with old versions of GCC.  Given that C++14 and C++17 have already been released, I don’t think that limiting ourselves to C++11 is reasonable.

parity-clib/parity.h
```diff
@@ -88,92 +120,112 @@ int parity_config_from_cli(char const* const* args, size_t const* arg_lens, size
...
+/// `parity_start` destroys the configuration object as well (even on failure).
+///
+/// It is safe to pass NULL here, in which case this function has no effect.
+void parity_config_destroy(struct parity_config *cfg);
 
-/// Starts the parity client in background threads. Returns a pointer to a struct that represents
-/// the running client. Can also return NULL if the execution completes instantly.
+/// Starts the parity client in a background thread. Returns a pointer to a
+/// struct that represents the running client. Can also return NULL if the
+/// execution completes instantly.
```
dvdplm: When would this happen? Like, is `NULL` an error condition or can it be legitimate?

parity-clib/parity.h
```diff
@@ -88,92 +120,112 @@ int parity_config_from_cli(char const* const* args, size_t const* arg_lens, size
...
-///	- ud			: Specific user defined data that can used in the callback
+///	- response		: Callback to invoke when a websocket event
+/// occurs
+///	- ud			: Specific user defined data that can used in
+/// the callback
 ///
 ///	- On success	: The function returns an object to the current session
-///					  which can be used cancel the subscription
+///					which can be used cancel the
+/// subscription
```
dvdplm: odd whitespace?

parity-clib/parity.h
```diff
@@ -88,92 +120,112 @@ int parity_config_from_cli(char const* const* args, size_t const* arg_lens, size
...
-int parity_unsubscribe_ws(const void *const session);
+int parity_unsubscribe_ws(const struct parity_subscription *const session);
 
 /// Sets a callback to call when a panic happens in the Rust code.
 ///
-/// The callback takes as parameter the custom param (the one passed to this function), plus the
-/// panic message. You are expected to log the panic message somehow, in order to communicate it to
-/// the user. A panic always indicates a bug in Parity.
+/// The callback takes as parameter the void * `param` passed to this function
+/// and the panic message. You are expected to log the panic message somehow, in
```
dvdplm:
```suggestion
/// and the panic message. You are expected to log the panic message somehow in
```

parity-clib/parity.h
```diff
@@ -88,92 +120,112 @@ int parity_config_from_cli(char const* const* args, size_t const* arg_lens, size
...
+/// order to communicate it to the user. A panic almost always indicates either
```
dvdplm:
```suggestion
/// order to communicate it to the user. A panic almost always indicates 
```

dvdplm:
> is not compatible with old versions of GCC.

Can you work with thew devops team to get that sorted?

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
-		if (parity != nullptr) {
-			parity_destroy(parity);
-		}
-	}
-	return 0;
+  // run full-client
+  {
+    std::vector<const char *> config = {"--no-ipc", "--jsonrpc-apis=all",
+                                        "--chain", "kovan"};
+    struct parity_ethereum *parity = parity_run(config);
```
niklasad1: why is not the newly introduced `ParityEthereum class` used here instead of the `C API` bindings?

parity-clib/parity.hpp
```diff
@@ -0,0 +1,155 @@
...
+  ParityEthereum(ParityConfig config, ParityLogger logger,
+                 std::function<void(std::string_view)> new_chain_spec_callback)
+      : parity_ethereum_instance(nullptr),
+        callback(std::make_unique<decltype(new_chain_spec_callback)>(
+            std::move(new_chain_spec_callback))) {
+    struct ::ParityParams params = {
+        config.config,
+        [](void *custom, const char *new_chain, size_t new_chain_len) {
+          auto view = std::string_view(new_chain, new_chain_len);
+          reinterpret_cast<decltype(new_chain_spec_callback) *>(custom)->
```
niklasad1: can you explain why `reinterpret_cast` is needed here? Should be avoided if possible

parity-clib/examples/cpp/main.cpp
```diff
@@ -16,22 +16,23 @@
...
 #include <regex>
 #include <string>
 #include <thread>
 
-void* parity_run(std::vector<const char*>);
-int parity_subscribe_to_websocket(void*);
-int parity_rpc_queries(void*);
+static parity_ethereum *parity_run(std::vector<const char *>);
+static int parity_subscribe_to_websocket(parity_ethereum *);
+static int parity_rpc_queries(parity_ethereum *);
```
Robbepop: We generally should avoid using non-fixed sized types such as `int` and instead use types defined in https://en.cppreference.com/w/cpp/header/cstdint or https://en.cppreference.com/w/cpp/header/cstddef

parity-clib/examples/cpp/main.cpp
```diff
@@ -16,22 +16,23 @@
...
 #include <parity.h>
+#include <parity.hpp>
 #include <regex>
 #include <string>
 #include <thread>
 
-void* parity_run(std::vector<const char*>);
-int parity_subscribe_to_websocket(void*);
-int parity_rpc_queries(void*);
+static parity_ethereum *parity_run(std::vector<const char *>);
```
Robbepop: We should wrap the `parity_ethereum` type provided by the C interface as a C++ class.
There we can define a destructor that does all the required clean-up through calls to the C-API.
This way we can simply return a value-type of the C++ `parity_ethereum` type.
Also for a C++ interface we should generally avoid `std::vector` as input parameters and instead prefer having an iterator (or ranges in C++20) based API. Also there is the `std::string_view` type in C++17 that you might want to use instead of `const char*`. It is quite similar to Rust's `&str`.

If you want to return an optional instance (because I saw returns of `nullptr`) then you should instead return `std::optional<parity_ethereum>` with the C++ wrapper.
If `parity_ethereum` is a sub-class and you want to return a base-class of it, then prefer returning `std::unique_ptr<parity_ethereum>`.

parity-clib/examples/cpp/main.cpp
```diff
@@ -16,22 +16,23 @@
...
 const int SUBSCRIPTION_ID_LEN = 18;
 const size_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
 const unsigned int CALLBACK_RPC = 1;
 const unsigned int CALLBACK_WS = 2;
 
 struct Callback {
-	unsigned int type;
-	long unsigned int counter;
+  unsigned int type;
+  long unsigned int counter;
```
Robbepop: Prefer using types defined in https://en.cppreference.com/w/cpp/header/cstdint.

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
-		}
-
-		if (parity != nullptr) {
-			parity_destroy(parity);
-		}
-	}
-	return 0;
+  // run full-client
+  {
+    std::vector<const char *> config = {"--no-ipc", "--jsonrpc-apis=all",
```
Robbepop: you could use `std::string_view` here instead of `const char*`.

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
+                                        "--chain", "kovan"};
+    struct parity_ethereum *parity = parity_run(config);
+    if (parity_rpc_queries(parity)) {
+      printf("rpc_queries failed\r\n");
```
Robbepop: Not sure if done here but make sure to use C++'s version of `printf` (found in <cstdio>) here instead of C's version.

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
+    }
+
+    if (parity != nullptr) {
+      parity_destroy(parity);
+    }
+  }
+
+  // run light-client
+  {
+    std::vector<const char *> light_config = {
```
Robbepop: again could use `std::string_view`

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
+        "--no-ipc", "--light", "--jsonrpc-apis=all", "--chain", "kovan"};
+    struct parity_ethereum *parity = parity_run(light_config);
+
+    if (parity_rpc_queries(parity)) {
+      printf("rpc_queries failed\r\n");
+      return 1;
```
Robbepop: Why don't we return proper error codes instead of raw numbers such as `1`?

parity-clib/examples/cpp/main.cpp
```diff
@@ -50,136 +51,141 @@ const std::vector<std::string> ws_subscriptions {
...
+    }
+
+    if (parity_subscribe_to_websocket(parity)) {
+      printf("ws_queries failed\r\n");
+      return 1;
+    }
+
+    if (parity != nullptr) {
+      parity_destroy(parity);
+    }
```
Robbepop: We should really make use of a RAII class instead of directly call destructors.

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: @niklasad1 because I have not had time yet.  That’s something I will do today.

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: @Robbepop That comes next.

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: The RAII part is done.  I chose to throw C++ exceptions instead of using `optional` for return values.

DemiMarie-parity:
@Robbepop The code is now much more idiomatic C++.  It uses RAII to manage resources, and relies on C++11 move semantics and C++17 types such as `std::string_view`.

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
+#ifdef __clang__
+#pragma clang diagnostic error "-Wvexing-parse"
```
Robbepop: Do we require this? Are we dedicated to using `clang` ?

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
...
-int parity_subscribe_to_websocket(void*);
-int parity_rpc_queries(void*);
+void parity_subscribe_to_websocket(ParityEthereum &ethereum);
+void parity_rpc_queries(ParityEthereum &);
+parity::ethereum::ParityEthereum parity_run(const std::vector<std::string> &);
 
 const int SUBSCRIPTION_ID_LEN = 18;
 const size_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
 const unsigned int CALLBACK_RPC = 1;
 const unsigned int CALLBACK_WS = 2;
```
Robbepop: I know that this code does not technically belong to your PR but please convert them to `constexpr` because we can - and also make use of proper sized int types as you did in the rest of the code. :)

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
...
+    } else if (type == CALLBACK_WS) {
+      std::match_results<std::string_view::iterator> results;
+      std::regex is_subscription(
+          "\\{\"jsonrpc\":\"2.0\",\"result\":\"0[xX][a-fA-"
+          "F0-9]{16}\",\"id\":1\\}");
+      if (std::regex_match(response.begin(), response.end(), results, is_subscription)) {
+        counter -= 1;
+      }
+    } else {
+      abort();
```
Robbepop: Could we use http://www.cplusplus.com/reference/cstdlib/exit/ here instead?

Robbepop: I am not sure about this but would it make more sense to have this as a lambda instead?

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
...
-const std::vector<std::string> rpc_queries {
-	"{\"method\":\"parity_versionInfo\",\"params\":[],\"id\":1,\"jsonrpc\":\"2.0\"}",
-	"{\"method\":\"eth_getTransactionReceipt\",\"params\":[\"0x444172bef57ad978655171a8af2cfd89baa02a97fcb773067aef7794d6913fff\"],\"id\":1,\"jsonrpc\":\"2.0\"}",
-	"{\"method\":\"eth_estimateGas\",\"params\":[{\"from\":\"0x0066Dc48bb833d2B59f730F33952B3c29fE926F5\"}],\"id\":1,\"jsonrpc\":\"2.0\"}",
-	"{\"method\":\"eth_getBalance\",\"params\":[\"0x0066Dc48bb833d2B59f730F33952B3c29fE926F5\"],\"id\":1,\"jsonrpc\":\"2.0\"}"
+const std::vector<std::string> rpc_queries{
+    "{\"method\":\"parity_versionInfo\",\"params\":[],\"id\":1,\"jsonrpc\":\"2.0\"}"s,
+    "{\"method\":\"eth_getTransactionReceipt\",\"params\":[\"0x444172bef57ad978655171a8af2cfd89baa02a97fcb773067aef7794d6913fff\"],\"id\":1,\"jsonrpc\":\"2.0\"}"s,
+    "{\"method\":\"eth_estimateGas\",\"params\":[{\"from\":\"0x0066Dc48bb833d2B59f730F33952B3c29fE926F5\"}],\"id\":1,\"jsonrpc\":\"2.0\"}"s,
+    "{\"method\":\"eth_getBalance\",\"params\":[\"0x0066Dc48bb833d2B59f730F33952B3c29fE926F5\"],\"id\":1,\"jsonrpc\":\"2.0\"}"s,
```
Robbepop: All of these would be cleaner with https://www.geeksforgeeks.org/raw-string-literal-c/.

Robbepop: Also could we maybe make them `constexpr std::vector<std::string_view>` ?

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
...
-		}
-	}
+void callback(std::string_view buf) { (void)buf; }
+} // namespace
+
+int main(int _argc, char **_argv) {
+  (void)_argc, (void)_argv;
+  using parity::ethereum::ParityEthereum;
+  // run full-client
+  if (0) {
```
Robbepop: ? :upside_down_face: 

also please prefer `false` or `true` in `bool` expressions.

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,134 @@
...
+    ;
 }
 
-int parity_rpc_queries(void* parity) {
-	if (!parity) {
-		return 1;
-	}
+void parity_subscribe_to_websocket(ParityEthereum &parity) {
+  // MUST outlive the std::vector below
+  Callback cb{.type = CALLBACK_WS, .counter = ws_subscriptions.size()};
```
Robbepop: As far as I know designated initializers are a C++20 feature and unfortunately it is still 2019. :(
https://blog.therocode.net/2018/11/exploring-cpp-20-designated-initializers

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: We do not require this.  I use `clang` myself, but this code should work with any conforming C++17 compiler.  This code is harmless, but I can go ahead and remove it if you want.

Robbepop: Unknown pragmas are going to be ignored by other compilers. So it doesn't hurt. I just wonder if we could encode this behaviour into CMake maybe?

parity-clib/examples/cpp/main.cpp
```diff
@@ -26,30 +26,33 @@
...
-const int SUBSCRIPTION_ID_LEN = 18;
-const size_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
-const unsigned int CALLBACK_RPC = 1;
-const unsigned int CALLBACK_WS = 2;
+constexpr uint32_t SUBSCRIPTION_ID_LEN = 18;
+constexpr uint32_t TIMEOUT_ONE_MIN_AS_MILLIS = 60 * 1000;
+enum class parity_callback_type : size_t {
+   CALLBACK_RPC = 1,
+   CALLBACK_WS = 2,
+};
```
Robbepop: Nice move making it an `enum class`! We should also respect naming of `enum class` members to be small snake case. :)

parity-clib/examples/cpp/main.cpp
```diff
@@ -26,30 +26,33 @@
...
 
 struct Callback {
-  size_t type;
+  parity_callback_type type;
```
Robbepop: generally I try to avoid variable declarations without initializers. For this I especially like the newer `auto` based initialization scheme so that you cannot forget initializing a variable:

For example:
```
auto type = parity_callback_type{ /* however it needs to be initialized */ };
```

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
+#ifndef PARITY_H_INCLUDED
+#define PARITY_H_INCLUDED 1
 #include <stddef.h>
+#include <stdint.h>
+// Rust assumes that size_t and uintptr_t are the same.  Check that that is the
+// case to avoid very obscure errors.
+#ifdef __cplusplus
+#if __cplusplus >= 201103L
+#include <type_traits>
+static_assert(std::is_same<uintptr_t, size_t>::value,
```
tomaka: Isn't that a tautology? I think it's defined to be always the case.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
+              "Rust code requires that uintptr_t and size_t are the same");
+#endif
+extern "C" {
+#elif __STDC_VERSION__ >= 201112L
+_Static_assert(sizeof(uintptr_t) == sizeof(size_t),
+               "Rust code requires that uintptr_t and size_t are the same");
+#endif
+
+/// An opaque struct that represents a Parity configuration.
+struct parity_config;
```
tomaka: It is "safe" to use `struct foo;` and `foo*` rather than using `void*`?
From my experience, any attempt to be type-safe in C is a bad idea that generally backfires.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
-/// to `len`. The strings don't need to be zero-terminated.
+/// The first two parameters are string pointers and string lengths. They must
+/// have a length equal to `len`. The strings don't need to be zero-terminated.
 ///
-/// On success, the produced object will be written to the `void*` pointed by `out`.
+/// On success, the produced object will be written to the `struct parity_config
+/// *` pointed by `out`.
 ///
-/// Returns 0 on success, and non-zero on error.
+/// Returns false on success, and true on error.
```
tomaka: Idiomatic C usually involves returning integers, so that you can pass back error codes.
I didn't return error codes because that was too cumbersome to define, but ideally that's IMO what we should do.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
 ///
 /// # Example
 ///
 /// ```no_run
 /// void* cfg;
 /// const char *args[] = {"--light", "--can-restart"};
-/// size_t str_lens[] = {7, 13};
+/// uintptr_t str_lens[] = {7, 13};
```
tomaka: `uintptr_t` is defined as the integer representation of a pointer, while `size_t` is defined as a number of bytes. Here we're passing string lengths, so `size_t` is the correct thing here.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
 /// if (parity_config_from_cli(args, str_lens, 2, &cfg) != 0) {
-///		return 1;
+/// 		return 1;
 /// }
 /// ```
 ///
-int parity_config_from_cli(char const* const* args, size_t const* arg_lens, size_t len, void** out);
+bool parity_config_from_cli(char const *const *args, uintptr_t const *arg_lens,
```
tomaka: Same remark.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
-/// 				call `parity_start` (even on failure).
+/// **Important**: The configuration object passed inside `cfg` is destroyed
+/// when you call `parity_start` (even on failure).
 ///
-/// On success, the produced object will be written to the `void*` pointed by `out`.
+/// On success, the produced object will be written to the `void*` pointed by
+/// `out`.  If the provided command-line arguments didn’t require starting a
+/// node, `*out` will be unchanged.
 ///
 /// Returns 0 on success, and non-zero on error.
```
tomaka: I don't really understand why you kept the integer here, but not above.

parity-clib/parity.h
```diff
@@ -14,166 +14,270 @@
...
+/// and the panic message. You are expected to log the panic message and exit.
+/// If this function returns, Parity Ethereum will exit the process itself.
 ///
-/// Note that this method sets the panic hook for the whole program, and not just for Parity. In
-/// other words, if you use multiple Rust libraries at once (and not just Parity), then a panic
-/// in any Rust code will call this callback as well.
+/// It is not possible to recover from a panic.  Calling `longjmp` or throwing a
+/// C++ exception results in undefined behavior.  It *is* permissible to block
+/// for an arbitrary amount of time in this callback.  Due to Rust’s memory
+/// safety, it is unlikely that memory has been corrupted, so you can (and
```
tomaka: Data might be in an incorrect state though, so I'd advise to **not** save data to disk before exiting.

parity-clib/parity.hpp
```diff
@@ -0,0 +1,228 @@
...
+  ParityConfig &operator=(ParityConfig &&other) noexcept {
+    if (this != &other) {
+      this->config = other.config;
+      other.config = nullptr;
+    }
+    return *this;
+  }
+  ~ParityConfig() {
+    if (this->config)
+      assert(false && "ParityConfig objects must be moved into a "
```
tomaka: This will trigger if an exception is thrown before the user moves the `ParityCOnfig` into a `ParityParams`.
What's wrong with just calling `parity_config_destroy` here?

parity-clib/parity.hpp
```diff
@@ -0,0 +1,228 @@
...
+      throw std::runtime_error("Failed to start Parity Ethereum");
+  }
+  ParityEthereum(ParityEthereum &&other) noexcept
+      : parity_ethereum_instance(nullptr), callback(nullptr) {
+    this->parity_ethereum_instance = other.parity_ethereum_instance;
+    other.parity_ethereum_instance = nullptr;
+  }
+  ParityEthereum &operator=(ParityEthereum &&other) noexcept {
+    if (this != &other) {
+      this->parity_ethereum_instance = other.parity_ethereum_instance;
```
tomaka: I think you should call `parity_destroy(this->parity_ethereum_instance)` beforehand here. Same in the method above.

parity-clib/parity.hpp
```diff
@@ -0,0 +1,228 @@
...
+                     new callback_ptr(copied_callback)))
+      throw std::runtime_error("Parity RPC failed");
+  }
+  parity_subscription subscribe(
+      const std::string_view buffer,
+      parity_rpc_callback callback) const {
+    if (::parity_subscription *session = ::parity_subscribe_ws(
+            this->parity_ethereum_instance, buffer.data(), buffer.size(),
+            parity_subscription_callback,
+            new parity_rpc_callback(callback),
```
tomaka: If `parity_subscribe_ws` returns null, you have a memory leak here.

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: Not really.  `__clang__` will only be defined by clang, so this will certainly be ignored by other compilers.

parity-clib/parity.h
DemiMarie-parity: It should be, IMO.  `void*` and `struct foo*` have the same ABI on all platforms I know of.

How does it backfire in your experience?

parity-clib/parity.h
DemiMarie-parity: I didn’t think of that.

parity-clib/parity.h
DemiMarie-parity: You are indeed correct here.

parity-clib/parity.h
DemiMarie-parity: No good reason.

parity-clib/parity.h
DemiMarie-parity: That is a good point.

parity-clib/parity.hpp
DemiMarie-parity: Nothing, but there is no corresponding `parity_logger_destroy`.

parity-clib/parity.h
tomaka: I don't have a strong opinion on `void*` vs `struct foo*` so feel free to discard. My question was mostly about whether `void*` is truly equivalent to `struct foo*`.

With `struct foo;` it would for example be possible for some to write `struct foo { int a; };` and try to use `foo->a`. But I would consider that as being in the "don't do that" category, rather than the "safety" category.

As for the backfire: I don't think there's anything wrong with what this PR does. Trying to squeeze oriented-object programming into C (a language that isn't made for that) tends to lead to very weird or annoying APIs (think GTK for example), so we shouldn't try *too hard* to be strongly-typed.

parity-clib/parity.hpp
DemiMarie-parity: Done

parity-clib/parity.hpp
DemiMarie-parity: Fixed

parity-clib/parity.hpp
DemiMarie-parity: Fixed by using explicit destructor callbacks.

parity-clib/parity.h
DemiMarie-parity: Fixed with explicit error codes.

parity-clib/parity.h
DemiMarie-parity: Fixed, with static assertions that `uintptr_t` and `size_t` are the same (which Rust assumes).

parity-clib/parity.h
DemiMarie-parity: Fixed.

parity-clib/parity.h
DemiMarie-parity: Fixed.

parity-clib/parity.h
DemiMarie-parity: Agreed.  IMO, the main use of the C API is for wrappers in higher-level languages.

parity-clib/examples/cpp/CMakeLists.txt
```diff
@@ -1,18 +1,22 @@
...
-    COMMAND cargo build -p parity-clib --verbose --color=always    # Note: use --release in a real project
+    BUILD_COMMAND ""    # Note: use --release in a real project
+    COMMAND cargo build -p parity-clib --verbose --color=always # Note: use --release in a real project
     BINARY_DIR "${CMAKE_SOURCE_DIR}/../../../target"
     INSTALL_COMMAND ""
     LOG_BUILD ON)
 
 add_dependencies(parity-example libparity)
-target_link_libraries(parity-example "${CMAKE_SOURCE_DIR}/../../../target/debug/${CMAKE_SHARED_LIBRARY_PREFIX}parity${CMAKE_SHARED_LIBRARY_SUFFIX}")
+target_link_libraries(parity-example "${CMAKE_SOURCE_DIR}/../../../target/release/${CMAKE_SHARED_LIBRARY_PREFIX}parity${CMAKE_SHARED_LIBRARY_SUFFIX}")
```
niklasad1: this should be `debug` unless use you `cargo build --release`

parity-clib/examples/cpp/CMakeLists.txt
```diff
@@ -7,16 +7,19 @@ set (CMAKE_CXX_EXTENSIONS OFF) # Enforce C++17
 set (CMAKE_EXPORT_COMPILE_COMMANDS ON) # generate a compilation database
 add_executable(parity-example main.cpp)
 add_compile_options("-Wall -Wextra -Werror")
-
+set(parity_shared_library_path "${CMAKE_SOURCE_DIR}/../../../target/release/${CMAKE_SHARED_LIBRARY_PREFIX}parity${CMAKE_SHARED_LIBRARY_SUFFIX}")
 ExternalProject_Add(
-    libparity
+    libparity.so
```
tomaka: I don't think that's cross-platform!

DemiMarie-parity: It doesn’t actually matter ― it is just an opaque string to CMake.  That said, I can revert it.

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: I abort here because this should never happen, and I would rather hard-kill the process than risk a use-after-free.

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
-		if (parity != nullptr) {
-			parity_destroy(parity);
-		}
-	}
-	return 0;
+void callback(std::string_view buf) { (void)buf; }
+} // namespace
+
+int main(int _argc, char **_argv) {
+  (void)_argc, (void)_argv;
```
niklasad1: why is this needed?

parity-clib/examples/cpp/CMakeLists.txt
```diff
@@ -1,18 +1,25 @@
...
-
+add_compile_options("-Wall -Wextra -Werror")
+set(parity_shared_library_path "${CMAKE_SOURCE_DIR}/../../../target/release/${CMAKE_SHARED_LIBRARY_PREFIX}parity${CMAKE_SHARED_LIBRARY_SUFFIX}")
 ExternalProject_Add(
     libparity
     DOWNLOAD_COMMAND ""
     CONFIGURE_COMMAND ""
-    BUILD_COMMAND ""
-    COMMAND cargo build -p parity-clib --verbose --color=always    # Note: use --release in a real project
+    BUILD_COMMAND cargo build -p parity-clib --release --verbose --color=always
```
niklasad1: I'm not sure we want `--release` here because it is part of the `CI` pipeline and might significantly slower

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
-		if (parity_rpc(parity, query.c_str(), query.length(), TIMEOUT_ONE_MIN_AS_MILLIS, callback, &cb) != 0) {
-			return 1;
-		}
-	}
-
-	while(cb.counter != 0);
-	return 0;
+namespace {
+void parity_rpc_queries(ParityEthereum &parity) {
+  Callback cb{.type = parity_callback_type::callback_rpc, .counter = rpc_queries.size()};
```
niklasad1: designated initializers is a C++20 thing?

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: That was left from when I was trying to debug the segmentation faults, and didn’t want to sync the Kovan testnet.  Very restrictive firewall rules on my system would have stopped it anyway, but I still didn’t want it to try.  Now fixed.

parity-clib/examples/cpp/main.cpp
```diff
@@ -16,22 +16,23 @@
```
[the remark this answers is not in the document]
DemiMarie-parity: Agreed!

tomaka:
I must admit that I'd be much more confident in this PR if someone who knows idiomatic C/C++ very well could review this.

DemiMarie-parity:
@Robbepop can you re-review this?  I can add missing documentation.

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,136 @@
...
-const unsigned int CALLBACK_RPC = 1;
-const unsigned int CALLBACK_WS = 2;
+namespace {
+using namespace std::literals::string_literals;
+using parity::ethereum::parity_subscription;
+using parity::ethereum::ParityEthereum;
+
+void parity_subscribe_to_websocket(ParityEthereum &ethereum);
+void parity_rpc_queries(ParityEthereum &);
+parity::ethereum::ParityEthereum parity_run(const std::vector<std::string> &);
```
Robbepop: I am not sure how strict we are on ourselves here but I want to point something out for this API:
If I were to translate this API directly to Rust it would look something similar to this:
```rust
fn parity_run(args: &Vec<String>) -> parity::ethereum::ParityEthereum { ... }
```
If this was a Rust codebase I wouldn't accept such as an API and instead propose something like:
```rust
fn parity_run<'a, A>(args: A) -> parity::ethereum::ParityEthereum
where
    A: IntoIterator<Item = &'a str> + 'a,
{ ... }
```
So iterators instead of reference to `Vec` and `str` instead of `String` since it doesn't seem that we are going to take ownership of those `String` instances anyways. Also we might think about returning a `Result` instead but I don't know if the C++ code throws.
This Rust code has somewhat of a translation into C++ as the following:
```c++
template<typename ArgsIt>
auto parity_run(ArgsIt begin, ArgsIt end) -> parity::ethereum::ParityEthereum;
```
The good thing is that we are generic over an input range (iterator). The bad thing is that since C++ is missing concepts (before C++20) we cannot annotate type interface (e.g. Rust trait bounds) so the underlying implementation decides what is done here.

Now since in C++ templates are sometimes avoided in interfaces etc. we might want to think about another non-template solutions.

If we actually want to re-use and take ownership of the arguments we should do:
```c++
auto parity_run(args: std::vector<String>) -> parity::ethereum::ParityEthereum;
```
or
```c++
auto parity_run(args: std::vector<std::string_view>) -> parity::ethereum::ParityEthereum;
```

If we cannot use a template here we should still not use `std::vector<String>&` (reference to vec) for the same reasons we avoid doing so in Rust. What we can do instead is to simply use `std::vector<std::string_view>`. The big problem in C++ is that as until recently C++ had no type to express what is expressed in Rust through slices (e.g. `&[i32]`). There were proposals for an `std::array_view` type, but the (renamed) `std::span` type just got into C++20. So we cannot use it, yet.

If done without generics a generic way to deal with such an interface in C++ is to use a concrete iterator.
```c++
auto parity_run(begin: MyArgsIter, end: MyArgsIter) -> parity::ethereum::ParityEthereum;
```
However, this is super super ugly.
Conclusion: C++20 is the fix but it isn't there yet. In theory we should use template if we can. Otherwise we should think about the best possible interface from the above mentioned. Maybe you can come up with another one.

tomaka: > If we cannot use a template here we should still not use std::vector<String>& (reference to vec) for the same reasons we avoid doing so in Rust.

For what it's worth, `const foo&` is very idiomatic in C++. Semantically I really isn't the same as `&Foo` in Rust.

dvdplm:
@DemiMarie-parity I'd really like to see this land soon. We don't need absolute perfection and I think we already have a decent improvement over status quo?

DemiMarie-parity:
@dvdplm I would love to see this merged too.  @Robbepop can you take another look?

Robbepop:
> @dvdplm I would love to see this merged too. @Robbepop can you take another look?

I discussed this with @dvdplm and you some time ago and I think we should generally move forward with this PR, merge it, make a notice of its unstability, see what is broken when it is being used and fix it.
So according to this plan we've had some time ago: Let's just merge it and see what breaks?

dvdplm:
@DemiMarie-parity ptal at the CI failures
@niklasad1 @tomaka care to take one last look if there's anything you feel must change before we merge?

scripts/gitlab/test-cpp.sh
```diff
@@ -2,16 +2,15 @@
...
 THREADS=$(nproc)
-export CC="sccache gcc"
-export CXX="sccache g++"
 
 echo "________Running the C++ example________"
 DIR=parity-clib/examples/cpp/build
 mkdir -p $DIR
 cd $DIR
-cmake ..
+cmake -DCMAKE_C{,XX}_COMPILER_LAUNCHER=sccache
```
niklasad1: Seems to be executed from the wrong directory

```suggestion
cmake .. -DCMAKE_C{,XX}_COMPILER_LAUNCHER=sccache
```

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
-	for (auto query : rpc_queries) {
-		if (parity_rpc(parity, query.c_str(), query.length(), TIMEOUT_ONE_MIN_AS_MILLIS, callback, &cb) != 0) {
-			return 1;
-		}
-	}
-
-	while(cb.counter != 0);
-	return 0;
+namespace {
+void parity_rpc_queries(ParityEthereum &parity) {
```
niklasad1: Doesn't have to be addressing in this PR: 

However, technically I think `cb.counter != 0` could end up with `UB` but since we are not running this in the CI anymore we could just those indefinitely or use a timer to short circuit it and exit with the error.

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
-
-	while(cb.counter != 0);
-	std::this_thread::sleep_for(std::chrono::seconds(60));
-	for (auto session : sessions) {
-		parity_unsubscribe_ws(session);
-	}
-	return 0;
+  while (cb.counter != 0) {
+  }
+  std::this_thread::sleep_for(std::chrono::seconds(60));
```
niklasad1: why is this `sleep` needed?

niklasad1: I don't think this `sleep` necessary

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
+void parity_rpc_queries(ParityEthereum &parity) {
+  Callback cb{parity_callback_type::callback_rpc, rpc_queries.size()};
+  auto cb_func = std::function(cb);
+
+  try {
+    for (const auto &query : rpc_queries)
+      parity.rpc(query, TIMEOUT_ONE_MIN_AS_MILLIS, cb_func);
+  } catch (std::exception &exn) {
+    std::cerr << exn.what() << std::endl;
+    while (cb.counter != 0) {
```
niklasad1: Can you explain why `while (cb.counter != 0)` is waited for when exceptions occur? 

I expect this to loop forever when an exception occurs because we will only get exceptions when the query fails?!

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
-		cb->counter -= 1;
-	} else if (cb->type == CALLBACK_WS) {
-		std::regex is_subscription ("\\{\"jsonrpc\":\"2.0\",\"result\":\"0[xX][a-fA-F0-9]{16}\",\"id\":1\\}");
-		if (std::regex_match(response, is_subscription) == true) {
-			cb->counter -= 1;
-		}
-	}
-}
+const std::vector<std::string> cli_args{"--no-ipc"s, "--jsonrpc-apis=all"s,
+                                        "--chain"s, "kovan"s};
```
niklasad1: doesn't have to be addressed in this PR: but I suppose we should run `goerli` instead `kovan`

parity-clib/examples/cpp/main.cpp
```diff
@@ -14,172 +14,138 @@
...
+const std::vector<std::string> light_config{
+    "--no-ipc"s, "--light"s, "--jsonrpc-apis=all"s, "--chain"s, "kovan"s};
```
niklasad1: doesn't have to be addressed in this PR: but I suppose we should run `goerli` instead `kovan`

parity-clib/examples/cpp/main.cpp
DemiMarie-parity: `while (cb.counter != 0)` is waited for to avoid UB (use after free).  That said, the new design may make this unnecessary.

dvdplm:
Closing as there are no known use-cases for this.

PR #11206, opened 2019-10-25
Upgrade jsonrpc to latest

dvdplm:


niklasad1:
should this be backported?

ordian:
> should this be backported?

no, 14.0.3 fixed a regression introduced in 14.0.x release, and stable and beta are using older versions

PR #11208, opened 2019-10-27
Use logging instead of flush! and flushln! for json-tests

### LINKED ISSUE #11084: [json-tests] Sort out the logging as tests run
The logging of activity during json-test runs is a bit odd, with copious use of `flushln!` and `flush!` which makes it hard to follow what is printed where.
Switch to using proper logging with `info!`/`warn!`/`debug!`/`trace` as appropriate. Use `target: "json-tests"` to make filtering easy.

rakanalh:
Fixes #11084 

Replace `flush!` and `flushln!` macro with:
- `warn!` for failures, skipped tests 
- `trace!` for messages about tests
- `info!` for successful runs

### DIFF
--- a/ethcore/src/json_tests/chain.rs
+++ b/ethcore/src/json_tests/chain.rs
@@ -56,7 +56,7 @@ pub fn json_chain_test<H: FnMut(&str, HookType)>(path: &Path, json_data: &[u8], 
 			let mut fail_unless = |cond: bool| {
 				if !cond && !fail {
 					failed.push(name.clone());
-					flushln!("FAIL");
+					warn!(target: "json-tests", "FAIL");
 					fail = true;
 					true
 				} else {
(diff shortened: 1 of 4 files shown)


rakanalh:
[clabot:check]

rakanalh:
@dvdplm, thanks for looking into this.

The original issue said:
```
The logging of activity during json-test runs is a bit odd, with copious use of flushln! and flush! which makes it hard to follow what is printed where.
Switch to using proper logging with info!/warn!/debug!/trace as appropriate. Use target: "json-tests" to make filtering easy.
```

Let me check if i understand your previous comment correctly. Should i somehow create a string vector, append those strings from the `flush!` and `flushln!` there and print the items of the vector at the end of the tests?

dvdplm:
> append those strings from the `flush!` and `flushln!` there and print the items of the vector at the end of the tests?

Yeah, that was sort of what I was thinking. It is not very elegant either though, so if you can think of something better I'd be very interested. Seems like it might require some refactoring though.

dvdplm:
Closing as stale, but feel free to re-open if/when you are ready to work on it again.

PR #11209, opened 2019-10-27
Use provided usd-per-eth value if an endpoint is specified

### LINKED ISSUE #9089: Using gasprice oracle for a custom network

rakanalh:
Fixes #9089 

This PR will use `GasPriceConfig::Calibrated` when the user provides an endpoint to an API endpoint which provides the Eth->USD price for a custom network.

The change is inspired by previous work of [Baranowski](https://github.com/Baranowski)

### DIFF
--- a/miner/price-info/src/lib.rs
+++ b/miner/price-info/src/lib.rs
@@ -57,8 +57,7 @@ impl<F> cmp::PartialEq for Client<F> {
 
 impl<F: Fetch> Client<F> {
 	/// Creates a new instance of the `Client` given a `fetch::Client`.
-	pub fn new(fetch: F, pool: Executor) -> Client<F> {
-		let api_endpoint = "https://api.etherscan.io/api?module=stats&action=ethprice".to_owned();
+    pub fn new(fetch: F, pool: Executor, api_endpoint: String) -> Client<F> {
 		Client { pool, api_endpoint, fetch }
 	}
 
@@ -105,11 +104,11 @@ mod test {
 	use super::Client;
 
 	fn price_info_ok(response: &str, executor: Executor) -> Client<FakeFetch<String>> {
-		Client::new(FakeFetch::new(Some(response.to_owned())), executor)
+		Client::new(FakeFetch::new(Some(response.to_owned())), executor, "fake_endpoint".to_owned())
 	}
 
 	fn price_info_not_found(executor: Executor) -> Client<FakeFetch<String>> {
-		Client::new(FakeFetch::new(None::<String>), executor)
+		Client::new(FakeFetch::new(None::<String>), executor, "fake_endpoint".to_owned())
 	}
 
 	#[test]
--- a/miner/price-info/src/lib.rs
+++ b/miner/price-info/src/lib.rs
@@ -57,7 +57,7 @@ impl<F> cmp::PartialEq for Client<F> {
 
 impl<F: Fetch> Client<F> {
 	/// Creates a new instance of the `Client` given a `fetch::Client`.
-    pub fn new(fetch: F, pool: Executor, api_endpoint: String) -> Client<F> {
+	pub fn new(fetch: F, pool: Executor, api_endpoint: String) -> Client<F> {
 		Client { pool, api_endpoint, fetch }
 	}
 
--- a/parity/configuration.rs
+++ b/parity/configuration.rs
@@ -58,6 +58,7 @@ use network::{IpFilter};
 
 const DEFAULT_MAX_PEERS: u16 = 50;
 const DEFAULT_MIN_PEERS: u16 = 25;
+pub const ETHERSCAN_ETH_PRICE_ENDPOINT: &str = "https://api.etherscan.io/api?module=stats&action=ethprice";
 
 #[derive(Debug, PartialEq)]
 pub enum Cmd {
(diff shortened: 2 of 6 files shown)


parity/configuration.rs
```diff
@@ -666,23 +666,32 @@ impl Configuration {
...
+					Colour::Yellow.bold().paint(format!("{}", wei_per_gas))
+				);
 
-		Ok(GasPricerConfig::Fixed(wei_per_gas))
+				return Ok(GasPricerConfig::Fixed(wei_per_gas))
+			},
+			Err(_) => {
+				let mut arg_usd_per_eth_endpoint = self.args.arg_usd_per_eth.as_str();
+				if "auto" ==  arg_usd_per_eth_endpoint {
+					arg_usd_per_eth_endpoint = "https://api.etherscan.io/api?module=stats&action=ethprice"
```
rakanalh: I believe this default value should go into some kind of a defaults module... will try to check if there is one.

miner/price-info/src/lib.rs
```diff
@@ -57,8 +57,7 @@ impl<F> cmp::PartialEq for Client<F> {
 
 impl<F: Fetch> Client<F> {
 	/// Creates a new instance of the `Client` given a `fetch::Client`.
-	pub fn new(fetch: F, pool: Executor) -> Client<F> {
-		let api_endpoint = "https://api.etherscan.io/api?module=stats&action=ethprice".to_owned();
+    pub fn new(fetch: F, pool: Executor, api_endpoint: String) -> Client<F> {
```
dvdplm: odd looking whitespace here

parity/helpers.rs
```diff
@@ -146,7 +146,7 @@ pub fn to_addresses(s: &Option<String>) -> Result<Vec<Address>, String> {
 
 /// Tries to parse string as a price.
 pub fn to_price(s: &str) -> Result<f32, String> {
-	s.parse::<f32>().map_err(|_| format!("Invalid transaciton price 's' given. Must be a decimal number."))
+	s.parse::<f32>().map_err(|_| format!("Invalid transaciton price {:?} given. Must be a decimal number.", s))
```
dvdplm: 👍

parity/configuration.rs
dvdplm: yeah, it's a bit nasty to have hardcoded strings like this. Even a `const` would be an improvement.

parity/helpers.rs
niklasad1:
```suggestion
	s.parse::<f32>().map_err(|_| format!("Invalid transaction price {:?} given. Must be a decimal number.", s))
```

parity/configuration.rs
niklasad1: I'm not super intimate with this piece,  however, I think it would be more clear to explicit check `if "auto" == self.args.arg_usd_per_eth.as_str()` before matching/parsing the string as a floating-point number.

In addition, I don't think it makes sense to add an `API endpoint` to  `GasPricerConfig::Calibrated` because it is already stored in `PriceInfoClient`? Why is it needed?

parity/params.rs
```diff
@@ -264,7 +266,7 @@ impl GasPricerConfig {
 	pub fn to_gas_pricer(&self, fetch: FetchClient, p: Executor) -> GasPricer {
 		match *self {
 			GasPricerConfig::Fixed(u) => GasPricer::Fixed(u),
-			GasPricerConfig::Calibrated { usd_per_tx, recalibration_period, .. } => {
+			GasPricerConfig::Calibrated { usd_per_tx, recalibration_period, ref api_endpoint } => {
```
niklasad1: `api_endpoint`is needed here because it is used to create `PriceInfoClient`

parity/configuration.rs
```diff
@@ -666,23 +666,32 @@ impl Configuration {
...
-			Colour::Yellow.bold().paint(format!("{}", wei_per_gas))
-		);
+				info!(
+					"Using a fixed conversion rate of Ξ1 = {} ({} wei/gas)",
+					Colour::White.bold().paint(format!("US${:.2}", usd_per_eth_parsed)),
+					Colour::Yellow.bold().paint(format!("{}", wei_per_gas))
+				);
 
-		Ok(GasPricerConfig::Fixed(wei_per_gas))
+				return Ok(GasPricerConfig::Fixed(wei_per_gas))
```
niklasad1: explicit return not needed

```suggestion
				Ok(GasPricerConfig::Fixed(wei_per_gas))
```

parity/configuration.rs
```diff
@@ -666,23 +666,32 @@ impl Configuration {
...
+			},
+			Err(_) => {
+				let mut arg_usd_per_eth_endpoint = self.args.arg_usd_per_eth.as_str();
+				if "auto" ==  arg_usd_per_eth_endpoint {
+					arg_usd_per_eth_endpoint = "https://api.etherscan.io/api?module=stats&action=ethprice"
+				}
+				return Ok(GasPricerConfig::Calibrated {
```
niklasad1: explicit return not needed

```suggestion
				Ok(GasPricerConfig::Calibrated {
```

parity/configuration.rs
```diff
@@ -666,23 +666,32 @@ impl Configuration {
...
+					Colour::White.bold().paint(format!("US${:.2}", usd_per_eth_parsed)),
+					Colour::Yellow.bold().paint(format!("{}", wei_per_gas))
+				);
 
-		Ok(GasPricerConfig::Fixed(wei_per_gas))
+				return Ok(GasPricerConfig::Fixed(wei_per_gas))
+			},
+			Err(_) => {
+				let mut arg_usd_per_eth_endpoint = self.args.arg_usd_per_eth.as_str();
+				if "auto" ==  arg_usd_per_eth_endpoint {
```
niklasad1:
```suggestion
				if "auto" == arg_usd_per_eth_endpoint {
```

niklasad1:
```bash
--usd-per-eth=[SOURCE]                                                                                                                                                                                  
        USD value of a single ETH. SOURCE may be either an amount in USD, a web service or 'auto' to use each web                                                                                           
        service in turn and fallback on the last known good value. (default: auto)  
```

Well, it doesn't seem that `auto` means that falling back to the last good value anymore, just use `default web service (etherscan)`?! 

Can somebody explain what the difference between `--usd-per-eth auto` and `--usd-per-eth https://api.etherscan.io/api\?module\=stats\&action\=price
`?

parity/configuration.rs
rakanalh: Used if-let and defined the endpoint as a constant.

rakanalh:
Thanks for reviewing @dvdplm and @niklasad1. I addressed your comments.

rakanalh:
> Can somebody explain what's the difference between `--usd-per-eth auto` and `--usd-per-eth https://api.etherscan.io/api\?module\=stats\&action\=price `?

According to the previous code, they are both the same. Because the previous client was being instantiated using the default endpoint defined in the client itself.

parity/configuration.rs
```diff
@@ -666,23 +667,31 @@ impl Configuration {
...
+			info!(
+				"Using a fixed conversion rate of Ξ1 = {} ({} wei/gas)",
+				Colour::White.bold().paint(format!("US${:.2}", usd_per_eth_parsed)),
+				Colour::Yellow.bold().paint(format!("{}", wei_per_gas))
+			);
 
-		Ok(GasPricerConfig::Fixed(wei_per_gas))
+			Ok(GasPricerConfig::Fixed(wei_per_gas))
+		} else {
+			let arg_usd_per_eth_endpoint = self.args.arg_usd_per_eth.as_str();
```
niklasad1: I don't think this variable makes the code more readable, maybe just remove it and then `api_endpoint: self.args.arg_usd_per_eth.clone()`

niklasad1:
> According to the previous code, they are both the same. Because the previous client was being instantiated using the default endpoint defined in the client itself.

Yeah, I was referring to that the documentation seems to be incorrect but not related to your changes. Initially, I think `auto` was meant to use the last known value if for example `etherscan` or the internet connection was down. I don't think `auto` is good name anymore we should probably rename it to `etherscan` or update the documentation

rakanalh:
@niklasad1 seems like the CI docker is having issues. I'll amend the commit and force push again once  it's working again.

niklasad1:
> seems like the CI docker is having issues. I'll amend the commit and force push again once it's working again.

Don't bother I'll restart the CI jobs for you

parity/params.rs
```diff
@@ -256,6 +257,7 @@ impl Default for GasPricerConfig {
 		GasPricerConfig::Calibrated {
 			usd_per_tx: 0.0001f32,
 			recalibration_period: Duration::from_secs(3600),
+			api_endpoint: "https://api.etherscan.io/api?module=stats&action=ethprice".to_owned()
```
dvdplm: Can't we use `ETHERSCAN_ETH_PRICE_ENDPOINT` here?

rakanalh: Done

PR #11211, opened 2019-10-28
[informant]: `MillisecondDuration` -> `as_millis()`

niklasad1:
This commit removes the trait `MillisecondDuration` and replaces it with `Duration::as_millis` instead.

### DIFF
--- a/parity/blockchain.rs
+++ b/parity/blockchain.rs
@@ -294,14 +294,13 @@ fn execute_import_light(cmd: ImportBlockchain) -> Result<(), String> {
 	}
 	client.flush_queue();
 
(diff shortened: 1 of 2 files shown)


parity/blockchain.rs
```diff
@@ -294,13 +294,14 @@ fn execute_import_light(cmd: ImportBlockchain) -> Result<(), String> {
 	}
 	client.flush_queue();
 
-	let ms = timer.elapsed().as_milliseconds();
+	let elapsed_secs = timer.elapsed().as_secs();
+	let elapsed_millis = timer.elapsed().as_millis();
```
ordian: let's extract `timer.elapsed()` into a variable here and below?

PR #11219, opened 2019-10-29
Snapshot restoration overhaul

dvdplm:
This PR does several things and is quite noisy. 
The general thrust is to make snapshot syncing easier to understand and debug and in the meanwhile makes a few tweaks to better adapt to the current harsh snapshotting environment.

- Separate snapshot sync logging from general sync, `-l snapshot_sync` and improve logging messages in general.
- Add documentation and comments where missing/lacking.
- Ensure we pick the snapshot with the highest blocknumber whenever there's a tie.

	Selecting a peer to sync a snapshot from gives the preference to snapshots that are seeded by 3 or more peers. On todays ethereum it is rare that nodes have snapshots from the same point in the chain. This means we always have a single peer to download from. Before this PR we did not sort the peers based on how recent their snapshot is when there's a tie (i.e. a single peer has a snapshot). Fixed [here](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-4cbf12eb53ed0891acc41ffd155a8930R882-R883).
	
- Do not skip snapshots further away than 30k block from the highest block seen.
	Currently we look for peers that seed snapshots that are close to the highest block seen on the network (where "close" means within 30k blocks of the tip). When a node starts up we wait for some time (5sec, increased here to 10sec) to let peers connect. If we find a suitable peer to sync a snapshot from at the end of that delay, we start the download; if none is found we start slow-syncing (or, if `--warp-barrier` is used we stall).
	Before this PR, when looking for a suitable snapshot we use the highest block seen on the network to check if a peer has a snapshot that is within 30k blocks of that highest block number. This means that in a situation where all available snapshots are old, we will often fail to start a snapshot at all. What's worse is that the longer we delay starting a snapshot sync (to let more peers connect, in the hope of finding a good snapshot), the more likely we are to have seen a high block and thus the less likely we become to accept a snapshot. This behaviour is quite surprising and tricky to reason about.
	[Here this highest-seen blocknumber comparison is removed](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-4cbf12eb53ed0891acc41ffd155a8930L834-L836) and we switch to the simpler: [pick the best suitable snapshot we find in 10sec](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-4cbf12eb53ed0891acc41ffd155a8930R862-R874).
	
- ~~A small but possibly controversial change: before this PR when a remote peers sends us a snapshot chunk we've already seen, we disconnect the peer and stop the download. After this PR we ignore the chunk and keep going instead ([change](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-2994ac0084ed75748d0d34fd97dfd543R116-R117)). I do not think this is problematic but I have never seen a dupe happen so can't really be sure.~~

- Don't warp-sync twice

	With the highest-seen blocknumber criteria gone, a node can end up using a snapshot that is a fair bit away from the tip, which means that after completion it needs to slow-sync for quite a bit to catch up to the tip. Before [this change](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-4cbf12eb53ed0891acc41ffd155a8930R849-R854) it can happen that a restored node finds a new snapshot and starts warp syncing again; this PR adds a check to see if our own best block is higher than the `--warp-barrier` and if so, does not start another sync.

- Avoid excessive looping over `Vec`s. 

	[Use a `HashSet` instead of a `Vec`](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-2994ac0084ed75748d0d34fd97dfd543R39-R44) to store the pending chunks. [Remove items](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-2994ac0084ed75748d0d34fd97dfd543R121-R135) from the set as chunks are processed. [Calculate and store the total number of chunks](https://github.com/paritytech/parity-ethereum/pull/11219/files#diff-2994ac0084ed75748d0d34fd97dfd543R106) in the `Snapshot`  struct instead of counting pending chunks each time.


ethcore/sync/src/snapshot_sync.rs
```diff
@@ -57,111 +73,117 @@ impl Snapshot {
...
+	/// Check if the the chunk is known, i.e. downloaded already or currently downloading; if so add
+	/// it to the `completed_chunks` set.
+	/// Returns a `ChunkType` with the hash of the chunk.
 	pub fn validate_chunk(&mut self, chunk: &[u8]) -> Result<ChunkType, ()> {
 		let hash = keccak(chunk);
 		if self.completed_chunks.contains(&hash) {
-			trace!(target: "sync", "Ignored proccessed chunk: {:x}", hash);
-			return Err(());
+			trace!(target: "snapshot_sync", "Already proccessed chunk {:x}. Ignoring.", hash);
+			return Ok(ChunkType::Dupe(hash));
```
ngotchac: I'm not sure about this change. It would mean that a peer can just continuously send the same chunk over and over, potentially eating our bandwidth, without getting disconnected. Is it actually happening in the wild that peers would send twice the same chunk and get disconnected from for it?

ethcore/sync/src/sync_io.rs
```diff
@@ -127,12 +125,22 @@ impl<'s> SyncIo for NetSyncIo<'s> {
...
 
-	fn peer_session_info(&self, peer_id: PeerId) -> Option<SessionInfo> {
-		self.network.session_info(peer_id)
+	fn peer_version(&self, peer_id: PeerId) -> ClientVersion {
+		self.network.peer_client_version(peer_id)
 	}
 
-	fn is_expired(&self) -> bool {
-		self.network.is_expired()
+	fn peer_enode(&self, peer_id: PeerId) -> String {
```
ngotchac: Why not returning `Option<String>` instead of some `???`?

ngotchac: Is this used anywhere?

ngotchac:
I left a couple of comments above. One thing I didn't mention is the change to accept far-away snapshots. I'm not quite convinced this would be a good idea: as a user, if I warp-sync my node I would expect to be able to use it a few minutes after (syncing to the top of the chain). Without any limits, I could actually spend quite some time syncing a snapshot and then a couple of days to the top of the chain. And if there's a new and better snapshot while I'm syncing the old one, I'm basically screwed.

Cargo.toml
```diff
@@ -121,7 +121,7 @@ name = "parity"
 
 [profile.release]
 debug = false
-lto = true
+lto = false
```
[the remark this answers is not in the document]
dvdplm: You're right, I should do that and stop polluting all my PRs like this. Apologies.

ethcore/sync/src/chain/handler.rs
```diff
@@ -495,56 +497,66 @@ impl SyncHandler {
...
 				if let (Some(hash), RestorationStatus::Failed) = (sync.snapshot.snapshot_hash(), status) {
-					trace!(target: "sync", "Noting snapshot hash {} as bad", hash);
+					// todo[dvdplm]: how do we know we `Failed` because of them and not because
+					// something happened on our end, say disk was full?
+					// On a second look I think `feed_chunk_with_restoration()`, the
+					// `SnapshotService` implementation, is indeed tossing away some errors. I think
+					// it should be refactored to distinguish the two cases of: restoration failed
+					// because the snapshot data is bad (corrupted, partial, malicious) and errors
+					// caused by the local env (out of disk seems like the obvious one).
+					debug!(target: "snapshot_sync", "Marking snapshot manifest hash {} as bad", hash);
```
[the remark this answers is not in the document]
dvdplm: They are not tossed, my wording is poor. It's true that this is not a big deal as they'll restart anyway. I'll remove the todo.

ethcore/sync/src/snapshot_sync.rs
dvdplm: > Is it actually happening in the wild that peers would send twice the same chunk and get disconnected from for it?

Indeed that is the question. I have never seen it happen and I have never heard of anyone that reported this. It seems like an odd thing to add a check for unless it actually happened.

Wrt to the dos potential here: any malicious node could send around fake manifests and spray the network with bad snapshots and users will not know the blocks are bad until they've imported them all. Protecting against the case of duplicate chunks seems like worrying about the wrong thing if you see what I mean?
On the other hand I also don't see how a non-malicious user could end up sending a duplicate unless there's a bug in our code.

(Somewhat related, on Ropsten there are snapshots at `#6500000` that are bad, most likely taken from a node on the wrong fork; it's quite frustrating to see your node go through the whole restoration and start syncing from `#6500001` and fail.)

I don't have a strong opinion here. @ordian what do you think? Should I undo this change?

dvdplm:
> the change to accept far-away snapshots.

Yeah, this is the most important change here. Hopefully they will not be "far far away". The idea is: use the best possible snapshot we can find.

> as a user, if I warp-sync my node I would expect to be able to use it a few minutes after

That would be nice, but that is simply not the reality today. Obviously we can hope that after the [Pause pruning while snapshotting](https://github.com/paritytech/parity-ethereum/pull/11178) PR things will improve, but I just don't think we should expect a return to the good old days where many nodes produce snapshots every 5k blocks. Today this is not the case.

I also think that the previous logic was a bit wonky, where the choice of peer to download from was highly dependent on seeing a high block on the chain. This led to a lot of head scratching wondering why a decent snapshot was rejected all of a sudden. I do not want to revert to that.

> And if there's a new and better snapshot while I'm syncing the old one, I'm basically screwed.

Why do you think that? You will not see the new better snapshot for sure (this was the case before this PR too) but once you've restored the one you were working on you could always restart the node with a higher `--warp-barrier` and do it again. Not saying that's good obviously, but it's not terrible either. Note that users don't really have a way to know what snapshots are out there so they don't really know what to expect (and I think we should consider adding a parity command or an oracle that give this information).

ethcore/sync/src/chain/mod.rs
```diff
@@ -172,19 +172,30 @@ const MAX_NEW_BLOCK_AGE: BlockNumber = 20;
...
+/// Time to wait for a peer to start being useful to us in some form. After this they are
+/// disconnected.
+const STATUS_TIMEOUT: Duration = Duration::from_secs(10);
 const HEADERS_TIMEOUT: Duration = Duration::from_secs(15);
 const BODIES_TIMEOUT: Duration = Duration::from_secs(20);
 const RECEIPTS_TIMEOUT: Duration = Duration::from_secs(10);
 const FORK_HEADER_TIMEOUT: Duration = Duration::from_secs(3);
-const SNAPSHOT_MANIFEST_TIMEOUT: Duration = Duration::from_secs(5);
+/// Max time to wait for a peer to send us a snapshot manifest.
+const SNAPSHOT_MANIFEST_TIMEOUT: Duration = Duration::from_secs(15);
```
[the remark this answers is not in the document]
dvdplm: The manifest files are fairly small, ~100-150kb. It's an rlp-encoded  [`ManifestData`(https://github.com/paritytech/parity-ethereum/blob/41f583cc4d355bf113e4c883623225be1a0de75a/ethcore/types/src/snapshot.rs#L135), containing the hashes of the block/state chunks you see on disk in your `snapshot/` folder.
Why tripled? I simply wanted to give all possible snapshotting peers a decent chance to be considered.

ethcore/sync/src/snapshot_sync.rs
dvdplm: Fair enough, reverted.

ethcore/snapshot/src/service.rs
```diff
@@ -676,7 +676,6 @@ impl<C> Service<C> where C: SnapshotClient + ChainInfo {
 		} else if manifest.state_hashes.contains(&hash) {
 			true
 		} else {
-			warn!(target: "snapshot", "Hash of the content of {:?} not present in the manifest block/state hashes.", path);
```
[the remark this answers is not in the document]
ordian: either way is fine :)

ethcore/sync/src/chain/mod.rs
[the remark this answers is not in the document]
dvdplm: It kind of boils down to the same thing though right?
- a manifest request is sent to a peer and the `asking` is set to `SnapshotManifest`
- every now and then we check if they are still in `PeerAsking::SnapshotManifest`
- when the peer responds with the manifest rlp we reset the `asking` 
- …unless of course the timeout is triggered before they respond, in which case they are disconnected

Did I get that right?

Either way I updated the doc string, not sure why it's not showing up here.

ethcore/sync/src/snapshot_sync.rs
```diff
@@ -22,20 +22,43 @@ use keccak_hash::keccak;
...
+	// See https://github.com/paritytech/parity-common/issues/255
+	#[ignore_malloc_size_of = "no impl for IndexSet (yet)"]
+	pending_state_chunks: IndexSet<H256>,
+	/// List of hashes of the block chunks we need to complete the warp sync from this snapshot.
+	/// These hashes are contained in the Manifest we downloaded from the peer(s).
+	/// Note: this is an ordered set so that state restoration happens in order, which keeps
+	/// memory usage down.
+	// See https://github.com/paritytech/parity-common/issues/255
+	#[ignore_malloc_size_of = "no impl for IndexSet (yet)"]
+	pending_block_chunks: IndexSet<H256>,
```
ngotchac: I'm not sure it's worth the extra dependency on `indexmap`. The only benefit is removing the chunk for the set when it's done.

dvdplm:
@ngotchac reverted to 5 sec. :)

dvdplm:
@s3krit this one may be tricky to backport, just a heads up.

PR #11225, opened 2019-10-31
Insert explicit warning into the panic hook

grbIzl:
This additional warning should help to debug the use case, when stderr is setup to put output somewhere else (not in Parity log).
In particular, I had this issue, when Parity was running as daemon and one of its threads panicked. The log looked very innocent, no panic was shown and it was not easy to understand the reason, why Parity quits.

### DIFF
--- a/parity/main.rs
+++ b/parity/main.rs
@@ -285,6 +285,7 @@ fn main_direct(force_can_restart: bool) -> i32 {
 					let e = exit.clone();
 					let exiting = exiting.clone();
 					move |panic_msg| {
+						warn!("Panic occured, see stderr for details");
 						eprintln!("{}", panic_msg);
 						if !exiting.swap(true, Ordering::SeqCst) {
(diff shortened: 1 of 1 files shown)


dvdplm:
Shouldn't the daemonizer script be responsible for setting up stdout/stderr? How did you run parity on your end? `parity daemon`? Or…?

grbIzl:
> Shouldn't the daemonizer script be responsible for setting up stdout/stderr? How did you run parity on your end? `parity daemon`? Or…?

The script does setup this. But I was not looking for the panic as there were no any signs of it in the logs. We considered all other things, that can cause such behavior (OOM, misconfig, cache-size etc). And it was not my setup btw, so it also made things harder.

grbIzl:
> I would still suggest moving `eprintln!` to a `warn!` below as stderr is not redirected to a parity log when running it as a (systemd-) daemon.

I just don't want to break the current behavior, as something can depend on it.

Issue #11224, opened 2019-10-31
Docker centos build fails


luosprz:
- **Parity Ethereum version**: 2.7.0-nightly
- **Operating system**: Linux
- **Installation**: docker built from source

I'm trying to build the centos docker version of Parity, based on the template in the master branch repo: https://github.com/paritytech/parity-ethereum/tree/master/scripts/docker.

The README.md seems to be outdated after move `docker` directory to `scripts` directory. After run the `build.sh` in the right directory:
```
git clone https://github.com/paritytech/parity-ethereum.git
cd parity-ethereum/scripts
./docker/centos/build.sh
```
I got the following cmake error:
```
cmake version 3.12.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
error: could not find `Cargo.toml` in `/build/parity-ethereum` or any parent directory
The command '/bin/sh -c yum -y update &&     yum install -y systemd-devel git make gcc-c++ gcc file binutils &&     curl -L "https://cmake.org/files/v3.12/cmake-3.12.0-Linux-x86_64.tar.gz" -o cmake.tar.gz &&     tar -xzf cmake.tar.gz &&     cp -r cmake-3.12.0-Linux-x86_64/* /usr/ &&     curl https://sh.rustup.rs -sSf | sh -s -- -y &&     PATH=/root/.cargo/bin:$PATH &&     RUST_BACKTRACE=1 &&     rustc -vV &&     cargo -V &&     gcc -v &&     g++ -v &&     cmake --version &&     cd parity-ethereum &&     cargo build --verbose --release --features final &&     strip /build/parity-ethereum/target/release/parity &&     file /build/parity-ethereum/target/release/parity' returned a non-zero code: 101
```

This is because the [Dockerfile.build](https://github.com/paritytech/parity-ethereum/blob/master/scripts/docker/centos/Dockerfile.build#L5) is adding from the scripts directory instead from the previous path with the parity source code.

## How to fix

Update [build.sh](https://github.com/paritytech/parity-ethereum/blob/master/scripts/docker/centos/build.sh) shell script replacing the old path `docker/centos` to the new one `scripts/docker/centos`:
```
sed -i 's,docker/centos,scripts/docker/centos,g' build.sh
```

Run from `parity-ethereum` directory:

```
git clone https://github.com/paritytech/parity-ethereum.git
cd parity-ethereum/
./scripts/docker/centos/build.sh
```

PR #11226, opened 2019-10-31
Fix docker centos build

luosprz:
- Update shell script to build docker centos image with the new path.
  - Update README.md with updated docker centos build instructions.

Closes [#11224](https://github.com/paritytech/parity-ethereum/issues/11224)


luosprz:
[clabot:check]

PR #11227, opened 2019-10-31
[ethcore]: apply filter when `PendingSet::AlwaysQueue` in `ready_transactions_filtered`

niklasad1:
Fixes that `the filter` is applied for when parity is executed with ` --relay-set= cheap` i.e, fetch transactions from the transaction queue

The important changes are:
1. Apply filter after calling `self.transaction_queue.pending(...` same for `from_pending`
2. Moved the methods related to filtering in `filter_options` to avoid doing several times
3. Added some basic tests in `ethcore/miner/miner`
4. First, I attempted to mock this with RPC tests but it wouldn't test the real `filtering logic` so I didn't bother.
5. Fixed a bunch of whitespaces, replaced spaces with tabs

The semantics will be different between `from_queue` and `from_pending`,

- `from_queue`: will receive a Vec<VerifiedTransaction> with at most `max_len` and then the filter is applied, to avoid allocating a huge vector if there are many transactions in the queue
- `from_pending`: will iterate through the pending transactions and apply filter until max_len transactions are collected or all have been visited.


ethcore/src/miner/filter_options.rs
```diff
@@ -14,59 +14,105 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
 
-use ethereum_types::{Address, U256};
-use serde::de::{Deserialize, Deserializer, Error, MapAccess, Visitor};
 use std::fmt;
 use std::marker::PhantomData;
 
+use ethereum_types::{Address, U256};
```
niklasad1: most changes are whitespaces

ethcore/src/miner/miner.rs
```diff
@@ -1114,77 +1115,26 @@ impl miner::MinerService for Miner {
 					max_len,
 					ordering,
 				},
-			)
+			);
+
+			pending.retain(|tx| {
```
niklasad1: this is the `real fix`

dvdplm: Would have been hard to spot otherwise! :D

dvdplm: just curious, why not tack on the `retain` call to the call to `pending()`?

ethcore/src/miner/miner.rs
```diff
@@ -1114,77 +1115,26 @@ impl miner::MinerService for Miner {
...
+				filter.as_ref().map_or(true, |filter| {
+					filter.matches(tx.signed())
+				})
+			});
+
+			pending
```
dvdplm: Just remove the semi-column and the `pending` here to let implicit returns do its thing?

ethcore/src/miner/miner.rs
```diff
@@ -1114,77 +1115,26 @@ impl miner::MinerService for Miner {
...
-					.filter(|tx| {
-						filter.as_ref().map_or(true, |filter| {
-							match_common_filter(&filter.value, &(*tx.signed()).value)
-						})
-					})
-					// Filter by nonce
-					.filter(|tx| {
-						filter.as_ref().map_or(true, |filter| {
-							match_common_filter(&filter.nonce, &(*tx.signed()).nonce)
+							filter.matches(tx.signed())
```
dvdplm: 👍

ethcore/src/miner/miner.rs
niklasad1: [Vec::retain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain) doesn't return anything i.e, takes `&mut self`

ethcore/src/miner/filter_options.rs
```diff
@@ -14,59 +14,107 @@
...
-    pub gas: FilterOperator<U256>,
-    /// Contains the operator to filter the gas price value of the transaction
-    pub gas_price: FilterOperator<U256>,
-    /// Contains the operator to filter the transaction value
-    pub value: FilterOperator<U256>,
-    /// Contains the operator to filter the nonce value of the transaction
-    pub nonce: FilterOperator<U256>,
+	/// Contains the operator to filter the from value of the transaction
+	pub from: FilterOperator<Address>,
+	/// Contains the operator to filter the to value of the transaction
```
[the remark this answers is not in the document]
niklasad1: is probably better the rename `from` -> `sender` and `to` -> `receiver` to be uniform with the transaction types but these are not changed in this PR but yeah.

ethcore/src/miner/miner.rs
niklasad1: See [my earlier comment](https://github.com/paritytech/parity-ethereum/pull/11227/files#r343253293)

PR #11232, opened 2019-11-01
Encapsulate access to the client for secret store

grbIzl:
Hopefully it's the last step, that will allow to separate SecretStore to its own repository.
Currently Secret Store uses Client in order to retrieve information from the chain, mostly for its service contracts (like check of permissions). This functionality should be hidden behind abstract trait, that should be implemented inside ethcore.
Details about the solution:
### Before this change:
- SecretStore server is started (from parity main) using method with parameters-pointers to Client, Miner etc:
`pub fn start(client: Arc<Client>, sync: Arc<dyn SyncProvider>, miner: Arc<Miner>, self_key_pair: Arc<dyn NodeKeyPair>, mut config: ServiceConfiguration,
	db: Arc<dyn KeyValueDB>, executor: Executor) -> Result<Box<dyn KeyServer>, Error>`
- SecretStore module has build dependencies to ethcore libraries
- Parity module has thin layer inside to start secret store (mostly with required parameters)

### After this change: 
- SecretStore server is started (from parity main) using method with parameter-pointer to SecretStoreChain structure, declared in SecretStore module:
`pub fn start(trusted_client: Arc<dyn SecretStoreChain>, self_key_pair: Arc<dyn SigningKeyPair>, mut config: ServiceConfiguration,
	db: Arc<dyn KeyValueDB>, executor: Executor) -> Result<Box<dyn KeyServer>, Error>`
- This newly added structure doesn't have dependencies to ethcore in its API. So, as a result, SecretStore module doesn't have any dependencies to ethcore or any other internal Parity libraries. It has only dependencies to public crates from crates.io
- Parity module has thicker layer inside to start secret store, that contains implementation of SecretStoreChain (this implementation has dependencies to ethcore, but as it's the internal implementation inside parity, it's ok)

Tasks:
- [x] Move client related calls from all SecretStore modules to TrustedClient
- [x] Rework\duplicate used ethcore\common_types
- [x] Rename TrustedClient to (?) ClientWrapper
- [x] Move TrustedClient implementation into ethcore and leave only trait declaration for SecretStore
- [x] Check tests
- [x] Wrap Registar and CallContract interfaces
- [ ] Solve parity runtime dependency - will be resolved separately after https://github.com/paritytech/parity-common/pull/271


dvdplm:
>     * Solve parity runtime dependency

This one could be pretty involved I think. Do you have any thoughts on it (yet)?

secret-store/Cargo.toml
```diff
@@ -42,10 +35,8 @@ jsonrpc-server-utils = "14.0.3"
...
 [dev-dependencies]
 env_logger = "0.5"
-ethkey = { path = "../accounts/ethkey" }
-ethcore = { path = "../ethcore", features = ["test-helpers"] }
 tempdir = "0.3"
 kvdb-rocksdb = "0.1.5"
 
 [features]
-accounts = ["ethcore-accounts"]
+accounts = []
```
svyatonik: I think this feature isn't required anymore, right?

niklasad1: hmm, how is the `SS RPC interface` suppose to work without accounts? Or is it disabled by default?

secret-store/Cargo.toml
```diff
@@ -42,10 +35,5 @@ jsonrpc-server-utils = "14.0.3"
...
 [dev-dependencies]
 env_logger = "0.5"
-ethkey = { path = "../accounts/ethkey" }
-ethcore = { path = "../ethcore", features = ["test-helpers"] }
 tempdir = "0.3"
-kvdb-rocksdb = "0.1.5"
-
-[features]
-accounts = ["ethcore-accounts"]
+kvdb-rocksdb = "0.1.5"
```
ordian: please upgrade to `0.2` as in #11248

grbIzl:
> I thought [paritytech/parity-common#271](https://github.com/paritytech/parity-common/pull/271) needs to be merged first?

We discussed it @dvdplm and decided to try to separate secret store code into the new repository locally, in order to make sure, that we don't need anything except runtime (that is in PR 271). But in order to proceed to this, I need this preparation PR to be merged.

parity/secretstore/nodekeypair.rs
```diff
@@ -0,0 +1,57 @@
...
+	address: Address,
+	public: Public,
+	password: Password,
+}
+
+impl KeyStoreNodeKeyPair {
+	pub fn new(account_provider: Arc<AccountProvider>, address: Address, password: Password) -> Result<Self, EthKeyError> {
+		let public = account_provider.account_public(address.clone(), &password).map_err(|e| EthKeyError::Custom(format!("{}", e)))?;
+		Ok(KeyStoreNodeKeyPair {
+			account_provider: account_provider,
```
dvdplm:
```suggestion
			account_provider,
```

parity/secretstore/nodekeypair.rs
```diff
@@ -0,0 +1,57 @@
...
+			address: address,
```
dvdplm:
```suggestion
			address,
```

parity/secretstore/nodekeypair.rs
```diff
@@ -0,0 +1,57 @@
...
+			public: public,
```
dvdplm:
```suggestion
			public,
```

parity/secretstore/nodekeypair.rs
```diff
@@ -0,0 +1,57 @@
...
+			password: password,
```
dvdplm:
```suggestion
			password,
```

PR #11245, opened 2019-11-06
Add POSDAO transition and malice report queue.

afck:
This adds a `posdaoTransition` parameter to enable:

* Calling the `emitInitiateChange` function of the validator set contract.
The [POSDAO](https://forum.poa.network/t/posdao-white-paper/2208) network needs one of the validators to call this, so that a log event is added to a block when the validator set changes.
* A malice report queue.
Misbehaving validators can be reported to the validator set contract, by making a transaction with a call to `reportMalicious`. However, transactions may fail to get included and may even become obsolete, if some other transaction was made with their nonce. The queue keeps track of malice reports, and retries them with updated nonces until they succeed.
* Setting the gas price for malice reports to zero: Validator nodes shouldn't need to have positive balance in their signing accounts.

This is based on https://github.com/poanetwork/parity-ethereum/pull/151, https://github.com/poanetwork/parity-ethereum/pull/129, and https://github.com/poanetwork/parity-ethereum/commit/424c593932a273c09aea2243f4b7fcf66d6e1388.

It shares some code with https://github.com/paritytech/parity-ethereum/pull/10946.


afck:
Those test failures seem to be problems with the CI setup again.

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -300,6 +383,88 @@ impl ValidatorSet for ValidatorSafeContract {
...
+		let client = self.client.read().as_ref().and_then(Weak::upgrade).ok_or("No client!")?;
+		let client = client.as_full_client().ok_or("No full client!")?;
+
+		let latest_block = header.number().max(client.block_number(BlockId::Latest).unwrap_or(0)) + 1;
+		self.filter_report_queue(our_address, latest_block)?;
+
+		let mut queue = self.queued_reports.lock().unwrap();
+
+		if queue.len() > MAX_QUEUED_REPORTS {
+			warn!(target: "engine", "Removing report from report cache, even though it has not been finalized");
```
dvdplm:
```suggestion
			warn!(target: "engine", "Removing {} reports from report cache, even though it has not been finalized", queue.len() - MAX_QUEUED_REPORTS);
```
maybe?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -283,6 +317,55 @@ impl ValidatorSafeContract {
 			Some(matched_event) => Some(SimpleList::new(matched_event.new_set))
 		}
 	}
+
+	fn filter_report_queue(&self, our_address: &Address, latest_block: BlockNumber) -> Result<(), EthcoreError>{
+		let client = self.client.read().as_ref().and_then(Weak::upgrade).ok_or("No client!")?;
+		let client = client.as_full_client().ok_or("No full client!")?;
+
+		self.queued_reports.lock().unwrap().retain(|&(malicious_validator_address, block, ref _data)| {
+			trace!(target: "engine", "Checking if report can be removed from cache");
```
dvdplm:
```suggestion
			trace!(target: "engine", "Checking if report should be removed from cache");
```
?
And maybe include a way to know which report we're checking here? Include the malicious validator address and the block?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -300,6 +383,88 @@ impl ValidatorSet for ValidatorSafeContract {
...
+		if !emit_initiate_change_callable {
+			trace!(target: "engine", "New block #{} issued ― no need to call emitInitiateChange()", header.number());
+		} else {
+			trace!(target: "engine", "New block issued #{} ― calling emitInitiateChange()", header.number());
+			let (data, _decoder) = validator_set::functions::emit_initiate_change::call();
+			transactions.push((self.contract_address, data));
+		}
+
+		// Retry all pending reports.
+		self.filter_report_queue(header.author(), header.number())?;
```
dvdplm: How come `filter_report_queue()` isn't implemented as a member of the `queued_reports` itself? So the line below could be something like `let queued_reports = self.queued_reports.lock().unwrap().filter_reports()` (or something to the same effect)?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -283,6 +317,55 @@ impl ValidatorSafeContract {
...
+
+		self.queued_reports.lock().unwrap().retain(|&(malicious_validator_address, block, ref _data)| {
+			trace!(target: "engine", "Checking if report can be removed from cache");
+			if block > latest_block {
+				return false; // Report cannot be used, as it is for a block that isn’t in the current chain
+			}
+			if latest_block > 100 && latest_block - 100 > block {
+				return false; // Report is too old and cannot be used
+			}
+			// Check if the validator is already banned...
```
dvdplm: Just thinking out aloud here, but isn't it worthwhile to cut down on the number of contract calls? Could it make a single call like `does_this_guy_need_to_be_reported::call()? In other words: it doesn't seem like we care about the reason a report should be dropped.
Taking that one step further: would it be more performant to just send off all the reports we have and let the contract decide if the report is a duplicate or already processed?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -47,6 +49,11 @@ use super::simple_list::SimpleList;
 
 use_contract!(validator_set, "res/validator_set.json");
 
+/// The maximum number of reports to keep queued.
+const MAX_QUEUED_REPORTS: usize = 10;
+/// The maximum number of returned malice reports to resend.
```
dvdplm: I don't understand this comment. Can you expand/elaborate a bit?

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -48,31 +49,64 @@ pub struct ValidatorContract {
...
+		let data = validator_report::functions::report_benign::encode_input(*address, block);
+		let gas_price = self.report_gas_price(latest.number());
+		self.transact(data, gas_price)?;
+		warn!(target: "engine", "Benign report for validator {} at block {}", address, block);
+		Ok(())
+	}
+
+		/// Returns the gas price for report transactions.
+		///
+		/// After `posdaoTransition`, this is zero. Otherwise it is the default (`None`).
```
dvdplm: whitespace

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -48,31 +49,64 @@ pub struct ValidatorContract {
...
+		let client = self.client.read().as_ref().and_then(Weak::upgrade).ok_or("No client!")?;
+		let full_client = client.as_full_client().ok_or("No full client!")?;
+		let tx_request = TransactionRequest::call(self.contract_address, data).opt_gas_price(gas_price);
+		match full_client.transact(tx_request) {
+			Ok(()) | Err(transaction::Error::AlreadyImported) => Ok(()),
+			Err(e) => Err(e.to_string())?,
+		}
+	}
+
+	fn do_report_malicious(&self, address: &Address, block: BlockNumber, proof: Bytes) -> Result<(), EthcoreError> {
```
dvdplm: Maybe?
```suggestion
	fn enqueue_malicious_behaviour_report(&self, address: &Address, block: BlockNumber, proof: Bytes) -> Result<(), EthcoreError> {
```

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -583,6 +588,9 @@ pub struct AuthorityRound {
 	block_gas_limit_contract_transitions: BTreeMap<u64, Address>,
 	/// Memoized gas limit overrides, by block hash.
 	gas_limit_override_cache: Mutex<LruCache<H256, Option<U256>>>,
+	/// The block number at which the consensus engine switches from AuRa to AuRa with POSDAO
+	/// modifications.
```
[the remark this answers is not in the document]
varasev: The link could be https://www.xdaichain.com/for-validators/posdao-whitepaper

ethcore/engines/validator-set/src/safe_contract.rs
afck: Sure, I'd be happy to replace them with a single call per report, or even with a single call for the whole list. I don't know how expensive constant local contract calls are, and what the overhead of several simple vs. a single complex call is.

@varasev: What do you think about this API change?

ethcore/engines/validator-set/src/contract.rs
afck: But we not only enqueue, but also send a transaction:
```rust
self.transact(data, gas_price)?;
```
The queue is only a fallback, so that we retry if the transaction fails to get included in a block.

afck:
Sorry for the indentation confusion! (Trying with vim-IndentConsistencyCop now.)
I tried to address all comments so far.

ethcore/engines/validator-set/src/safe_contract.rs
afck: I extracted the report queue into a new struct; please let me know whether that's what you had in mind. (I put it in a separate commit, so we can revert it if you don't agree.)

afck:
I can't reproduce those test failures locally.

dvdplm:
> I can't reproduce those test failures locally.

It's a flaky test, apologies. Restarted the job.

ethcore/engines/validator-set/src/safe_contract.rs
varasev: OK, I added a separate getter to `ValidatorSet` contract, so let's replace [these lines](https://github.com/poanetwork/parity-ethereum/blob/dbdd9c5f0f0cbe274b7cc6962decc997e3ebe319/ethcore/engines/validator-set/src/safe_contract.rs#L592-L629) with a single call to [that getter](https://github.com/poanetwork/posdao-contracts/blob/2700b23fb4e6ad2a80802c72175c2a18b2149cb7/contracts/ValidatorSetAuRa.sol#L562-L586).

afck: Done.

varasev:
Hi all - It seems all suggestions have been addressed. Are there any other proposals we could take into account here? Or this PR could be approved?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -200,14 +211,37 @@ fn prove_initial(contract_address: Address, header: &Header, caller: &Call) -> R
...
+		let full_client = client.as_full_client().ok_or("No full client!")?;
+
+		let tx_request = TransactionRequest::call(self.contract_address, data).gas_price(U256::zero()).nonce(nonce);
+		match full_client.transact(tx_request) {
+			Ok(()) | Err(transaction::Error::AlreadyImported) => Ok(()),
+			Err(e) => Err(e)?,
 		}
 	}
 
+	pub(crate) fn queue_report(&self, addr: Address, block: BlockNumber, data: Vec<u8>) {
```
dvdplm:
```suggestion
	pub(crate) fn enqueue_report(&self, addr: Address, block: BlockNumber, data: Vec<u8>) {
```

varasev: Done in https://github.com/paritytech/parity-ethereum/pull/11245/commits/f7c6325a9a0f7cc9bbe4000a9e6a2778ccb8129f

ethcore/engines/authority-round/src/lib.rs
```diff
@@ -1116,6 +1126,56 @@ impl AuthorityRound {
...
+			return Ok(Vec::new());
+		}
+
+		let opt_signer = self.signer.read();
+		let signer = match opt_signer.as_ref() {
+			Some(signer) => signer,
+			None => return Ok(Vec::new()), // We are not a validator, so we shouldn't call the contracts.
+		};
+		let our_addr = signer.address();
+		let client = self.client.read().as_ref().and_then(|weak| weak.upgrade()).ok_or_else(|| {
```
ordian: should it use the newly added [`upgrade_client_or`](https://github.com/paritytech/parity-ethereum/pull/11408/files#diff-79ef4a3dce58a9d303d49c5aee781e17R1124) method?

ethcore/engines/validator-set/src/safe_contract.rs
```diff
@@ -16,9 +16,10 @@
 
 /// Validator set maintained in a contract, updated using `getValidators` method.
 
-use std::sync::{Weak, Arc};
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex, Weak};
```
ordian: why not `parking_lot::Mutex`? we're using `parking_lot::RwLock` here already, this would allow us to get rid of `unwrap`s

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -48,31 +49,64 @@ pub struct ValidatorContract {
...
-
-		match client.as_full_client() {
-			Some(c) => {
-				c.transact(TransactionRequest::call(self.contract_address, data))
-					.map_err(|e| format!("Transaction import error: {}", e))?;
-				Ok(())
-			},
-			None => Err("No full client!".into()),
+	fn transact(&self, data: Bytes, gas_price: Option<U256>) -> Result<(), String> {
+		let client = self.client.read().as_ref().and_then(Weak::upgrade).ok_or("No client!")?;
```
ordian: do we need to upgrade the client ref here? it seems that we do that in `do_report_malicious` and `do_report_benign` already, so we could just pass it here, right?

afck:
I addressed the comments, rebased, and fixed the bug @varasev found: The nonce was not incremented between the randomness and the POSDAO calls after the refactoring (moving the transaction generation into separate methods).
Let's run [posdao-test-setup](https://github.com/poanetwork/posdao-test-setup/) as well before merging, to make sure.

Happy to discuss this PR in a call (see Discord)!

varasev:
> Let's run `posdao-test-setup` as well before merging, to make sure.

Yes, I will retest that

varasev:
> The nonce was not incremented between the randomness and the POSDAO calls after the refactoring (moving the transaction generation into separate methods). Let's run `posdao-test-setup` as well before merging, to make sure.

That works 👌

afck:
Works for me as well. :+1:

afck:
We made the changes we discussed yesterday. Please take a look at the two new commits.

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -165,13 +165,17 @@ impl ValidatorSet for ValidatorContract {
 	fn report_malicious(&self, address: &Address, _set_block: BlockNumber, block: BlockNumber, proof: Bytes) {
 		if let Err(s) = self.do_report_malicious(address, block, proof) {
 			warn!(target: "engine", "Validator {} could not be reported ({}) on block {}", address, s, block);
+		} else {
+			info!(target: "engine", "Reporting malicious validator {} on block {}", address, block);
```
varasev: Seems we already have the warning message inside the `do_report_malicious` function.

ethcore/engines/validator-set/src/contract.rs
```diff
@@ -165,13 +165,17 @@ impl ValidatorSet for ValidatorContract {
...
 		}
 	}
 
 	fn report_benign(&self, address: &Address, _set_block: BlockNumber, block: BlockNumber) {
 		trace!(target: "engine", "validator set recording benign misbehaviour at block #{} by {:#x}", block, address);
 		if let Err(s) = self.do_report_benign(address, block) {
 			warn!(target: "engine", "Validator {} could not be reported ({}) on block {}", address, s, block);
+		} else {
+			info!(target: "engine", "Reporting validator {} on block {} (benign fault)", address, block);
```
varasev: Seems we already have the warning message inside the `do_report_benign` function.

afck:
You're right, @varasev, we completely missed that yesterday! I reverted that part of the change.

varasev:
@dvdplm Will it be correct to contribute to Parity docs through https://github.com/paritytech/wiki repository after our changes are released? Or there are other places we should add the docs to?

dvdplm:
> Will it be correct to contribute to Parity docs through https://github.com/paritytech/wiki repository after our changes are released? Or there are other places we should add the docs to?

That is a great question. I don't think we have a comprehensive plan for the client docs once we migrate away from the Parity org, so until we do I'd say the parity wiki is the least worst option?

ethcore/res/validator_contract.sol
```diff
@@ -1,62 +1,103 @@
 // Source for the test AuRa validator set contract.
```
dvdplm:
```suggestion
// Source for the test AuRa validator set contract. DO NOT USE IN PRODUCTION.
```

Maybe?

ethcore/res/validator_contract.sol
```diff
@@ -66,39 +107,13 @@ contract TestList {
...
-		if (isValidatorBanned(_maliciousMiningAddress)) {
-			return false; // already banned
-		}
-		address[] memory reports = maliceReportedForBlock(_maliciousMiningAddress, _blockNumber);
-		for (uint256 i = 0; i < reports.length; i++) {
-			if (reports[i] == _reportingMiningAddress) {
-				return false; // already reported
-			}
+		if (isValidatorBanned[_maliciousValidator]) {
+			// We shouldn't report of the malicious validator
```
dvdplm:
```suggestion
			// We shouldn't report the malicious validator
```

varasev:
> I'd say the parity wiki is the least worst option?

Ok, thanks

PR #11249, opened 2019-11-12
simplify verification

debris:
This pr attempts to simplify a bit a header/block verification process.

Changes:

- I removed `NoopVerifier` as it was never used
- after removal of `NoopVerifier`, the only remaining implementation of `Verifier` was `CanonVerifier` so I also removed it as well as the trait itself
- last param of `fn verify_block_family` is no longer optional. (it was always called with `Some`)
- I split `fn verify_header_params` into 2 functions:
  - `fn verify_header_params`
  - `fn verify_header_time` which does only timestamp verification of the header
- I formatted several extremely long lines so they are now more readable


debris:
> What was the reason you did this? Like, what's the gain having them split in two?

The function was taking an additional argument - `bool`, which was used to decide whether time of the header should be checked. By splitting it into two functions I reduced the complexity of the code.

> Did you notice any difference in the benches? I don't expect much of an impact, but you never know.

No, I forgot to alter the `verify_block_family` function call in one of the tests. It's fixed in d90e057

PR #11250, opened 2019-11-12
remove client mode

debris:
This is quite opinionated pr as it removes one of the parity features. Since 2016 parity allows users to set an operation mode of the node, that is: `online`, `passive`, `dark` or `off`. I believe that our users (who are mostly power users), never use any other option that `online`. Therefore I believe it is feasible to remove this feature, especially that it's implementation is not very clear, nor idiomatic

### DIFF
--- a/ethcore/client-traits/src/lib.rs
+++ b/ethcore/client-traits/src/lib.rs
@@ -31,7 +31,6 @@ use common_types::{
 	BlockNumber,
 	call_analytics::CallAnalytics,
 	chain_notify::{NewBlocks, ChainMessageType},
-	client_types::Mode,
 	encoded,
 	engines::{epoch::Transition as EpochTransition, machine::Executed},
 	errors::{EthcoreError, EthcoreResult},
@@ -369,12 +368,6 @@ pub trait BlockChainClient:
 
 	/// Get the preferred chain ID to sign on
 	fn signing_chain_id(&self) -> Option<u64>;
-
-	/// Get the mode.
-	fn mode(&self) -> Mode;
-
-	/// Set the mode.
-	fn set_mode(&self, mode: Mode);
 
 	/// Get the chain spec name.
 	fn spec_name(&self) -> String;
--- a/ethcore/src/client/client.rs
+++ b/ethcore/src/client/client.rs
@@ -1572,7 +1572,6 @@ impl BlockChainClient for Client {
 	}
 
 	fn disable(&self) {
-		//self.set_mode(Mode::Off);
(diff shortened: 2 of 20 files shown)


ordian:
should it be removed from `parity/cli/mod.rs` as well?

dvdplm:
I know I've used `mode=offline` in the last few months but I can't remember why I did it (it's in my shell history). Do you know why it was added in the first place? One thing that is really odd about it is that it's "sticky": setting it persists across restarts. Is that what you mean by it being not idiomatic?
@joshua-mir do you know of anyone ever using this?

joshua-mir:
I kinda disagree with this change - I do know of significant categories of users who use `passive` and `offline` - it seems used pretty regularly for people doing IOT stuff, embedded, and resource-limited environments, especially in combination with `--light`.

`passive` is really popular with people running "temporary" rpc endpoints as well.

debris:
>  Is that what you mean by it being not idiomatic?

No. I'm referring to how this feature is implemented. `enum Mode` as it is implemented right now is used in a similar way to C enums and it doesn't prevent anyone from actually using `Client` incorrectly. And since there are no tests for this code, we can't even be sure if it's working accordingly to the intensions of the author.

> do you know of anyone ever using this?

When I think of this a bit more, `offline` might be used by someone who wants to run a single node chain for test purposes... but I can't think of any case when someone would like to change run mode in the runtime via rpc.

debris:
> I kinda disagree with this change - I do know of significant categories of users who use passive and dark - it seems used pretty regularly for people doing IOT stuff, embedded, and resource-limited environments, especially in combination with --light.

I believe that mode is not even implemented for light clients 😞 

https://github.com/paritytech/parity-ethereum/blob/93fbbb9aaf161f21471050a2a3257f820c029a73/rpc/src/v1/impls/light/parity_set.rs#L124-L126

> `offline` is somewhat popular with people running "temporary" rpc endpoints as well.

what do you mean by saying temporary rpc endpoints?

joshua-mir:
> what do you mean by saying temporary rpc endpoints?

Clone the db, spin up a node in `offline` mode, make your rpc requests, and then shut everything down.

dvdplm:
> spin up a node in `offline` mode, make your rpc requests,

Ah, that's what I did too: spun up a silent node while benchmarking RPCs.

Ok, @debris, sounds like we're not removing this then. We should definitely clean up the code though, you up for that? Here or in a separate PR?

debris:
A separate pr is better

